PR #163, opened 2018-02-25
Changed the default home location

Baccata:
* Mill's default home folder is now $HOME/.mill/ammonite
  This lets users provide global settings by defining a predef.sc file
  in that home folder.

* Re-enabled the --home cli param to manually set the home folder


lihaoyi:
Looks good to me

PR #165, opened 2018-02-25
Scala js multi modules support

rockjam:
This PR adds support for multi-module builds with scala js. Also fixes issues with running tests on multi-module builds

### DIFF
--- a/scalajslib/src/mill/scalajslib/ScalaJSModule.scala
+++ b/scalajslib/src/mill/scalajslib/ScalaJSModule.scala
@@ -1,7 +1,7 @@
 package mill
 package scalajslib
 
-import ammonite.ops.{Path, ls, mkdir, rm}
+import ammonite.ops.{Path, exists, ls, mkdir, rm}
 import coursier.Cache
 import coursier.maven.MavenRepository
 import mill.eval.PathRef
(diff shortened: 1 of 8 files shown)


scalajslib/src/mill/scalajslib/ScalaJSModule.scala
```diff
@@ -17,7 +17,7 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
   trait Tests extends TestScalaJSModule {
     override def scalaVersion = outer.scalaVersion()
     override def scalaJSVersion = outer.scalaJSVersion()
-    override def moduleDeps = Seq(outer)
+    override def moduleDeps = outer +: outer.moduleDeps
```
rockjam: We need this, cause otherwise we don't have `sjsir` files of dependent modules in tests

lihaoyi: Why don't we have the `sjsir` files of dependent modules though? Shouldn't they all be aggregated as part of the `runClasspath`?

rockjam: Hm, yt turns out, that we don't use sjsir files from `runClasspath` in link:
https://github.com/lihaoyi/mill/blob/8a3af216f7ada18d25550b4d4c171d082050bc85/scalajslib/src/mill/scalajslib/ScalaJSModule.scala#L121

We only use libraries(jars) from runtime classpath, whereas sjsir files come from `input`

rockjam: @lihaoyi Do you think we should take sjsir files from runClasspath instead of what I did here? Now I think it's better soultion

PR #168, opened 2018-02-26
Add dsl for SCM (now called VersionControl)

MasseGuillaume:
The scm url syntax is a source of confusion for developper. I added VersionControl.github() to simplify this process. We can add other common VersionControl url scheme like Git, Bazar, etc.

### DIFF
--- a/docs/example-2/build.sc
+++ b/docs/example-2/build.sc
@@ -10,10 +10,7 @@ object foo extends PublishModule{
     organization = "com.lihaoyi",
     url = "https://github.com/lihaoyi/example",
     licenses = Seq(License.MIT),
-    scm = SCM(
-      "git://github.com/lihaoyi/example.git",
-      "scm:git://github.com/lihaoyi/example.git"
-    ),
+    versionControl = VersionControl.github("lihaoyi", "example"),
     developers = Seq(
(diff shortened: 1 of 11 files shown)


scalalib/src/mill/scalalib/publish/settings.scala
```diff
@@ -58,11 +58,25 @@ case class License(
     distribution: String = "repo"
 )
 
-case class SCM(
+object SCM {
+  @deprecated("use VersionControl", "0.1.0")
```
rockjam: Not sure we should deprecate this

MasseGuillaume: I prefer VersionControl, it's self-explanatory.

lihaoyi:
looks ok to me

docs/example-2/build.sc
```diff
@@ -10,10 +10,7 @@ object foo extends PublishModule{
     organization = "com.lihaoyi",
     url = "https://github.com/lihaoyi/example",
     licenses = Seq(License.MIT),
-    scm = SCM(
-      "git://github.com/lihaoyi/example.git",
-      "scm:git://github.com/lihaoyi/example.git"
-    ),
+    versionControl = VersionControl.github("lihaoyi", "example"),
```
japgolly: Fantastic

PR #169, opened 2018-02-26
use outputStream for test logs

rockjam:
This should fix color output for scalatest and other testing frameworks


lihaoyi:
Looks good

PR #195, opened 2018-03-04
Allows users to customize Node.js (Scala.js' NodeJSEnv) in ScalaJSModule

robby-phd:
This pull request allows users to configure Node.js in ScalaJSModule such as for passing arguments to increase heap memory, executable, etc., as allowed in Scala.js NodeJSEnv.

### DIFF
--- a/scalajslib/test/src/mill/scalajslib/NodeJSConfigTests.scala
+++ b/scalajslib/test/src/mill/scalajslib/NodeJSConfigTests.scala
@@ -14,8 +14,9 @@ object NodeJSConfigTests extends TestSuite {
   val scalaVersion = "2.12.4"
   val scalaJSVersion = "0.6.22"
   val utestVersion = "0.6.3"
-  val scalaTestVersion = "3.0.4"
-  val nodeArg2G = "--max-old-space-size=2048"
+  val nodeArgsEmpty = List()
+  val nodeArgs2G = List("--max-old-space-size=2048")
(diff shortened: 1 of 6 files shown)


rockjam:
@robby-phd Looks good, just one thing. Could we add a test case to check node.js configuration? If it's not hard to inspect somehow

scalajslib/test/src/mill/scalajslib/NodeJSConfigTests.scala
```diff
@@ -0,0 +1,118 @@
...
+    } yield (scala, nodeArgs)
+
+    object helloJsWorld extends Cross[BuildModule](matrix:_*)
+    class BuildModule(val crossScalaVersion: String, nodeArgs: List[String]) extends HelloJSWorldModule {
+      override def artifactName = "hello-js-world"
+      def scalaJSVersion = NodeJSConfigTests.scalaJSVersion
+      override def nodeJSConfig = T { NodeJSConfig(args = nodeArgs) }
+    }
+
+    object buildUTest extends Cross[BuildModuleUtest](matrix:_*)
```
rockjam: As it is dedicated spec for node js configuration, I think we don't need utest/scalatest builds. We'll be good with one test framework(utest) and 4 test cases:
```
test
test2G
run
run2G
```

rockjam: This will simplify build configuration and tests too.

rockjam:
Thank you

PR #196, opened 2018-03-04
Add out/{module_name} as a content root in generated IntelliJ project

nornagon:
The "/.." thing here feels janky to me, but I wasn't sure how to better obtain the "out/{module_name}" dir.

Without this, IntelliJ doesn't notice generated sources, presumably because they are not beneath any "content root", but I'm not sure exactly what the format of these .iml files are because I couldn't find any documentation 😕

### DIFF
--- a/scalalib/src/mill/scalalib/GenIdea.scala
+++ b/scalalib/src/mill/scalalib/GenIdea.scala
@@ -190,6 +190,10 @@ object GenIdea {
         mod.compile.ctx.segments
       )
       val Seq(scalaVersion: String) = evaluator.evaluate(Agg(mod.scalaVersion)).values
+      val generatedSourceOutPath = Evaluator.resolveDestPaths(
+        evaluator.outPath,
+        mod.generatedSources.ctx.segments
+      )
 
       val elem = moduleXmlTemplate(
         mod.millModuleBasePath.value,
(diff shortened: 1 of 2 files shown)


scalalib/src/mill/scalalib/GenIdea.scala
```diff
@@ -298,6 +298,7 @@ object GenIdea {
       <component name="NewModuleRootManager">
         <output url={"file://$MODULE_DIR$/" + relify(outputPath) + "/dest/classes"} />
         <exclude-output />
+        <content url={"file://$MODULE_DIR$/" + relify(outputPath) + "/.."} />
```
lihaoyi: You can do `outputPath / up` here

lihaoyi: Why do you need `/...` from the `outputPath` though? The `outputPath` itself should already be the folder that wraps the output paths of the module's Targets

nornagon:
The outputPath is out/foo/compile, but the generated source is at
out/foo/generatedSource, which isn’t a subdir of outputPath

lihaoyi:
Is there some way to pass in `out/foo/` directly? I'd prefer that rather than fishing around with `up`, since that tends to break if the layout changes

nornagon:
Yeah, that would be better. I’ll take a closer look later, was hoping you’d
have a suggestion :) thanks for taking a look!

nornagon:
Updated to be more targeted—instead of adding `out/{mod_name}` as a content root, it now adds `out/{mod_name}/generatedSources/dest`, which is more precisely the effect needed. Also, it computes that path in a more sane way without any silly `/..` hacks :)

lihaoyi:
great

PR #198, opened 2018-03-04
fix optional xml tags in POM

rockjam:
Fixes bug with printing of optional values in pom.xml. In 0.1.4 there is corrupted pom generated like: 
```
    <scm>
        <connection>Some(scm:git:git://github.com/rockjam/license-headers.git)</connection>
        .optionnal
        <developerConnection>Some(scm:git:ssh://git@github.com:rockjam/license-headers.git)</developerConnection>
        .optionnal
        <tag>None</tag>
        .optionnal
        <url>Some(https://github.com/rockjam/license-headers)</url>
        .optionnal
    </scm>
```


lihaoyi:
lgtm once you have a simple unit test for pom-generation to avoid regression

PR #203, opened 2018-03-05
Improve fullOpt behavior

### LINKED ISSUE #189: Scala.js: FullOpt + Closure compiler support
Hi!

I've ported a Scala/Scala.js app to mill, which was easier than expected. Mill works very well for me, very nice to compose with plain Scala, and it responds much quicker than SBT did in watch mode. So thank you!

The last part missing for me however is having the fullOpt output run through the closure compiler (which is what `sbt fullOptJS` does). `foo.scalaJSLinkerClasspath` indicates that all the necessary closure dependencies are available to the imported sjs toolkit. 

I imagine that if you're familiar with Scala.js and Mill internals it's more or less a configuration option to pass to the SJS linker? Or is there more voodoo involved?

I'd be happy to try if anyone can point me in the right direction.

Dennis4b:
Specifically, fix 2 issues in order to better mirror the behavior of "sbt fullOptJS"

- Invoke the Google Closure Compiler after generating the optimized Javascript output

- Set scala.scalajs.LinkingInfo.developmentMode = false (and productionMode = true) when
building fullOpt in order to allow Scala.js code to distinguish between development
and production mode.

Fixes #189

### DIFF
--- a/scalajslib/jsbridges/0.6/src/mill/scalajslib/bridge/ScalaJSBridge.scala
+++ b/scalajslib/jsbridges/0.6/src/mill/scalajslib/bridge/ScalaJSBridge.scala
@@ -7,7 +7,7 @@ import java.io.File
 import org.scalajs.core.tools.io.IRFileCache.IRContainer
 import org.scalajs.core.tools.io._
 import org.scalajs.core.tools.jsdep.ResolvedJSDependency
-import org.scalajs.core.tools.linker.{ModuleInitializer, StandardLinker}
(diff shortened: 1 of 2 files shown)


rockjam:
@Dennis4b some tests are failing. Maybe there is something with CI. can you push once again to trigger CI?

rockjam:
The PR iteself looks good. We can merge after CI goes green

lihaoyi:
Looks green-ish, merging it

PR #204, opened 2018-03-06
Update VersionControl ref to SCM in publish docs

swoogles:


swoogles:
I've recommitted this a few different ways, and for some reason it keeps ninja-including that line 226?
When I go to commit the change, it shows exactly one changed line. Yet when it completes, it includes the second line.

Might just be expected behavior with the hooks you've defined.

swoogles:
D'oh! I didn't have the latest version. Ignore this.

PR #206, opened 2018-03-06
WIP: Scala native

ajrnz:
The PR adds scala-native support to mill. It is integrated to mill via `mill.scalanativelib.ScalaNativeModule` which can be extended in the same way as `ScalaModule`

It's built against a scala-native 0.3.8-SNAPSHOT PR (scala-native/scala-native#1234) which would need to be compiled and published locally.

In addition to the standard `scalaVersion` you must also supply `scalaNativeVersion`.

An optimized binary can be built using `releaseMode = ReleaseMode.Release`

Two test frameworks are currently supported `utest` and `scalatest`. Support for `scalacheck` should be possible when the relevant artifacts have been published.

`HelloNativeWorldTests.scala` builds and tests a simple project (based on `HelloJSWorldTests.scala`)

An example `build.sc` would look like:

```
import mill._
import mill.define._
import mill.scalanativelib._

object nativehello extends ScalaNativeModule {
  def scalaVersion = "2.11.12"
  def scalaNativeVersion = "0.3.7"
  def logLevel = NativeLogLevel.Info // optional
  def releaseMode = ReleaseMode.Debug // optional
}
```

This code code will require minor modifications to remove the -SNAPSHOT versioning once the release version of scala-native 0.3.8 is released.


lolgab:
How the work is on this PL? Now Scala Native 0.3.7 with the new API is out so it could be very useful!

rockjam:
@lolgab I plan to continue work on this PR this week if @ajrnz don't mind. Yes, 0.3.7 with new API will be useful here.

lolgab:
Very good to know, thank you :-) 
Yes, I meant that before the official release of the API the Scala Native integration could not be published, but now that 0.3.7 is out the are no more excuses :-) I was about to fork the PR but I asked here before... I noticed that the PR is behind some commits compared to master, so I thought it needs a merge.

ajrnz:
Actually. I'm just about ready to push another update. I currently have it working against 0.3.7. The support for testing frameworks has been the most challenging part. I currently have utest, and scalatest mostly working. 

I'll push an update for comment...

rockjam:
@ajrnz Oh, that's great. I think if testing support is the most challenging part - we could implement it as separate PR, what do you think?

ajrnz:
I think I've done all the hard work now. It works, it just needs a little tidying. Probably better to have people look at it. 

I just need to rebase from 0.2.0 to master and then I can update the PR.

rockjam:
That sounds great, thank you!

rockjam:
Don't hesitate to ask, if you need any help

ajrnz:
Ok you should be able to use this against the published version of scala-native (0.3.7). It uses the new build API.

Tests should work but there may be a few rough edges.

Still needs the following:
* tests
* documentation (I'll write something when people are happy with the code/design etc.

https://github.com/ajrnz/scala-native-example-app can be used to for an example of how to use the module, pending documentation.

lolgab:
It seems also that ScalaNativeModule doesn't change the mainClass according to `mainClass`.

ajrnz:
@lolgab can you please try this repo https://github.com/ajrnz/scala-native-example-app and tell me what happens?

Also try adding,

    def nativeLinkStubs = true

lolgab:
Adding `def nativeLinkStubs = true` doesn't change anything.
I tried your repository and it works nicely.

ajrnz:
Hmm strange, it should be being included. Time for some test projects...

ajrnz:
In the current state it can now build and test using published scala-native 0.3.7.

It's capable of building and testing (10 test failures) upickle for example. See: https://github.com/ajrnz/upickle/tree/scala-native. You'll need to locally publish your own scalacheck 1.14.0 scala-native artifact until it's published on maven central)

@lihaoyi, @rockjam It would be good to get some feedback

I haven't really used it in anger yet but I have an idea for a simple project which I'll try when I get a chance

lihaoyi:
@ajrnz could you update the PR description with a summary of what the major changes and moving parts are in this PR?

ajrnz:
@lihaoyi Updated. Let me know if you need anything else

lihaoyi:
I'll take a look; @densh do you think you could glance over this and see if what andrew is doing makes sense to you, from a scala-native perspective?

lolgab:
In my setup it prints by default 735 lines of informations (useless to a normal user like me) about the linking stage.
```
/usr/bin/clang-4.0
	-c
	/home/lorenzo/scala/provaNative/out/build/nativeWorkdir/dest/__extern.ll
	-o
	/home/lorenzo/scala/provaNative/out/build/nativeWorkdir/dest/__extern.ll.o
	-O0
	-I/usr/local/include
	-Qunused-arguments
[debug] running
/usr/bin/clang-4.0
	-c
	/home/lorenzo/scala/provaNative/out/build/nativeWorkdir/dest/scala.util.ll
	-o
	/home/lorenzo/scala/provaNative/out/build/nativeWorkdir/dest/scala.util.ll.o
	-O0
	-I/usr/local/include
	-Qunused-arguments
...
...
```

lihaoyi:
This looks ok to me. @densh said he'll take a look

scalanativelib/src/mill/scalanativelib/testinterface/ComRunner.scala
```diff
@@ -0,0 +1,128 @@
...
+
+case class MessageOnlyException(msg: String) extends Exception(msg)
+
+/**
+ * Represents a distant program with whom we communicate over the network.
+ * @param bin    The program to run
+ * @param args   Arguments to pass to the program
+ * @param logger Logger to log to.
+ */
+class ComRunner(bin: File,
```
densh: We're going to refactor test runner out of sbt plugin in `0.3.8` (see https://github.com/scala-native/scala-native/pull/1234). It was an unfortunate oversight not to do it sooner.

scalanativelib/scalanativebridges/0.3/src/mill/scalanativelib/bridge/ScalaNativeBridge.scala
```diff
@@ -0,0 +1,71 @@
...
+import scala.scalanative.build.{Config, GC, Logger, Mode, Discover, Build}
+import ammonite.ops.Path
+import mill.scalanativelib.NativeConfig
+
+class ScalaNativeBridge extends mill.scalanativelib.ScalaNativeBridge {
+  val logger =
+    Logger(debugFn = msg => Unit, //err.println(msg),
+      infoFn  = msg => out.println(msg),
+      warnFn  = msg => out.println(msg),
+      errorFn = msg => err.println(msg))
```
densh: It would be nice to have some way to trigger debug messages being shown or not, maybe add a `nativeVerbose: Boolean` as a extra setting to control this. @lihaoyi is this the idiomatic way to do logging in mill?

scalanativelib/src/mill/scalanativelib/ScalaNativeModule.scala
```diff
@@ -0,0 +1,275 @@
...
+      Seq(Cache.ivy2Local, MavenRepository("https://repo1.maven.org/maven2")),
+      Lib.depToDependency(_, scalaVersion(), platformSuffix()),
+      toolsIvyDeps()
+    ).map(t => (scalaNativeBridgeClasspath().toSeq ++ t.toSeq).map(_.path))
+  }
+
+  override def scalacPluginIvyDeps = super.scalacPluginIvyDeps() ++
+    Agg(ivy"org.scala-native:nscplugin_${scalaVersion()}:${scalaNativeVersion()}")
+
+  def releaseMode = T { false }
```
densh: We reserve right to introduce more modes, that's why this setting is a string in the build api and sbt plugin. I'd recommend to keep it a string here as well to avoid future problems if more modes are introduced.

scalanativelib/src/mill/scalanativelib/testinterface/ComRunner.scala
```diff
@@ -0,0 +1,128 @@
...
+
+  private def log(message: Log): Unit =
+    println(message)
+//    message.level match {
+//      case Level.Info  => logger.info(message.message)
+//      case Level.Warn  => logger.warn(message.message)
+//      case Level.Error => logger.error(message.message)
+//      case Level.Trace => message.throwable.foreach(logger.trace(_))
+//      case Level.Debug => logger.debug(message.message)
+//    }
```
densh: We have "no commented out code" policy in SN, is this left as an oversight?

scalanativelib/src/mill/scalanativelib/testinterface/ComRunner.scala
ajrnz: @densh Thanks for the review

scalanativelib/scalanativebridges/0.3/src/mill/scalanativelib/bridge/ScalaNativeBridge.scala
ajrnz: I agree it. It is currently way to verbose. I'm not really sure what the best way to handle this is. Happy to take suggestions @lihaoyi

scalanativelib/src/mill/scalanativelib/ScalaNativeModule.scala
ajrnz: OK. I'd been meaning to look at this. Will change.

scalanativelib/src/mill/scalanativelib/testinterface/ComRunner.scala
ajrnz: Just to remind me that the current situation is not ideal and it needs sorting out. Will be removed when the logging situation is resolved.

Thanks!

scalanativelib/scalanativebridges/0.3/src/mill/scalanativelib/bridge/ScalaNativeBridge.scala
lihaoyi: We could define any flags or configuraiton on `ScalaNativeModule` and plumb it down here and where-ever necessary. Then people can override it to configure

scalanativelib/src/mill/scalanativelib/ScalaNativeModule.scala
lolgab: To me it is better to have a sealed trait instead of a string. So if more modes are introduced there are no problems in the future. It is how I did with scala js `ModuleKind`s miming the scalajs api.

ajrnz:
Ok the latest push should sort out the situation regarding release mode and logging configuration.

I'll check out scala-native/scala-native#1234 to see that it does everything required. I'd definitely prefer not to commit copied code.

lihaoyi:
@densh says the modularization of Scala Native's SBT test interfaces is scheduled to publish late may to early june, so unless someone has a time-critical need for this to go in I'm happy to just hold off a few weeks and cleanup/merge when Scala Native side is in better shape

PR #210, opened 2018-03-08
fix #174 skip interfaces when finding tests

### LINKED ISSUE #174: Tries to run trait as test
In the output of my tests I see the following red line:
```
...
Can't find a constructor for class my.project.testutil.BaseAfterAll
...
1 targets failed
core.test.test my.project.testutil.BaseAfterAll SuiteSelector
```

this is defined as 

```scala
trait BaseAfterAll extends AfterAll{
  override def afterAll: Unit = ()
}
```

The test fingerprint picks this up because it extends the specs2 `SpecificationStructure` but since this is a trait it should be skipped

francisdb:
fix #174 skip interfaces when finding tests


lihaoyi:
lgtm

PR #212, opened 2018-03-08
Support to switch between NoModule and CommonJSModule ModuleKinds

lolgab:
ScalaJS can generate CommonJS modules to use with bundlers like [Browserify](http://browserify.org) and [WebPack](https://webpack.js.org).
Now you can select the `CommonJSModule` mode in a `ScalaJSModule`:
```scala
object Foo extends ScalaJSModule {
  def scalaJSVersion = "0.6.22"
  def scalaVersion = "2.12.4"
  def moduleKind = ModuleKind.CommonJSModule
}
```
The default is `ModuleKind.NoModule` which exports to the global scope.


rockjam:
Thanks for the contribution. May we have a test for module kinds?

lihaoyi:
I'd be happy to merge this without tests for now, it seems like something unlikely for us to break accidentally

rockjam:
@lihaoyi I'm fine with that

PR #215, opened 2018-03-10
Support for Java 9 (and 10)

robby-phd:
This pull request addresses #156 by upgrading ammonite to (at least) [1.0.5-1-819bc80](https://github.com/lihaoyi/Ammonite/commit/819bc803fd0b95993246560a2a4a595998545c73).

Integrating this pull request builds a mill release that's compatible with Java 9 and 10, which can be used, for example, to include tests in travis using oraclejdk9 (in a separate pull request).

PR #217, opened 2018-03-10
Changes to make {clientserver,main,scalalib,scalajslib}.test work under Java 9

robby-phd:
While pull request #215 enables mill to run under Java 9 and 10, it turns out more changes are needed to make `{clientserver,main,scalalib,scalajslib}.test` work under Java 9-10.

Basically, creation of `URLClassLoader` has to take into account "exported rt.jar" (see lihaoyi/Ammonite#761); a helper object method is added to facilitate this.

Note that the following tests are disabled under Java 9 or above (require further investigations):

* Tests using Scala 2.10 and 2.11 (cannot find xsbt.CompilerInterface)

* Tests using ScalaJsUtil.runJs (null when retrieving nashorn script engine)

Integrating this pull request builds a mill release that can be used to run `test-mill-0.sh` using oraclejdk9 in travis.

### DIFF
--- a/core/src/mill/util/ClassLoader.scala
+++ b/core/src/mill/util/ClassLoader.scala
@@ -1,0 +1,14 @@
+package mill.util
+
+import java.net.URL
+
+import io.github.retronym.java9rtexport.Export
+
+import scala.reflect.internal.util.ScalaClassLoader.URLClassLoader
+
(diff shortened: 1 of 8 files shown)


lihaoyi:
Sounds good to me

PR #219, opened 2018-03-11
Test mill using Java 9 (in addition to Java 8)

robby-phd:


lihaoyi:
Seems like the last unstable commit release hasn't made it to maven central, will re-trigger this in an hour or so

Issue #157, opened 2018-02-24
Doesn't work on Windows


philwalk:
I added the following build.sc below an existing sbt project called TaxPrep:

```
// build.sc
import mill._
import mill.scalalib._

object TaxPrep extends ScalaModule {
  def scalaVersion = "2.12.4"
}
```
This was my first experiment after installing mill in a Windows/Cygwin environment, as per the instructions, and it installed with no problems.   When I try to compile the TaxPrep project, I get the following error message (sun/java/scala portions of the stack dump were removed for brevity):

```
$ mill TaxPrep.compile
Compiling (synthetic)/ammonite/predef/interpBridge.sc
Compiling (synthetic)/ammonite/predef/DefaultPredef.sc
Compiling C:\Users\philwalk\taxes\build.sc
[25/25] TaxPrep.compile
error: error while loading Object, Missing dependency 'object scala in compiler mirror', required by C:\opt\jdk\jre\lib\rt.jar(java/lang/Object.class)
error: scala.reflect.internal.MissingRequirementError: object scala in compiler mirror not found.
        at mill.scalaworker.ScalaWorker.compileZincBridge(ScalaWorker.scala:96)
        at mill.scalaworker.ScalaWorker.compileScala(ScalaWorker.scala:131)
        at mill.scalalib.ScalaModule.$anonfun$compile$2(ScalaModule.scala:130)
        at mill.define.ApplyerGenerated.$anonfun$zipMap$9(ApplicativeGenerated.scala:21)
        at mill.define.Task$MappedDest.evaluate(Task.scala:358)
        at mill.eval.Evaluator.$anonfun$evaluateGroup$14(Evaluator.scala:297)
        at mill.eval.Evaluator.$anonfun$evaluateGroup$13(Evaluator.scala:297)
        at mill.eval.Evaluator.$anonfun$evaluateGroup$9(Evaluator.scala:296)
        at mill.eval.Evaluator.$anonfun$evaluateGroup$9$adapted(Evaluator.scala:254)
        at mill.eval.Evaluator.evaluateGroup(Evaluator.scala:254)
        at mill.eval.Evaluator.evaluateGroupCached(Evaluator.scala:179)
        at mill.eval.Evaluator.$anonfun$evaluate$2(Evaluator.scala:86)
        at mill.eval.Evaluator.$anonfun$evaluate$2$adapted(Evaluator.scala:83)
        at mill.eval.Evaluator.evaluate(Evaluator.scala:83)
[...]

PR #228, opened 2018-03-15
Fixes for Windows non-interactive (client/server) mode, assembly, and posix permission

robby-phd:
More fixes for #157.

I've tested the non-interactive mode on Windows 10 with MSYS2 to build a mill release and to run `{main, scalajslib}.test` on Java 8 & 9.

Once merged, we can test the non-interactive mode on AppVeyor.

Notes:

* The echo server in `clientserver.test` that uses memory locks (instead of the file locks used by mill non-interactive) still does not work on Windows (sometimes deadlocking or throwing exceptions; I'll let others investigate the memory locks)

* `scalalib.test` still does not work due to Windows path length limit

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -236,7 +236,7 @@ object dev extends MillModule{
 
     write(outputPath, prependShellScript())
 
-    if (!System.getProperty("os.name").toLowerCase.startsWith("windows")) {
+    if (!scala.util.Properties.isWin) {
       val perms = java.nio.file.Files.getPosixFilePermissions(outputPath.toNIO)
       perms.add(PosixFilePermission.GROUP_EXECUTE)
       perms.add(PosixFilePermission.OWNER_EXECUTE)
--- a/build.sc
+++ b/build.sc
@@ -236,11 +236,13 @@ object dev extends MillModule{
 
     write(outputPath, prependShellScript())
 
-    val perms = java.nio.file.Files.getPosixFilePermissions(outputPath.toNIO)
-    perms.add(PosixFilePermission.GROUP_EXECUTE)
-    perms.add(PosixFilePermission.OWNER_EXECUTE)
-    perms.add(PosixFilePermission.OTHERS_EXECUTE)
-    java.nio.file.Files.setPosixFilePermissions(outputPath.toNIO, perms)
+    if (!System.getProperty("os.name").toLowerCase.startsWith("windows")) {
+      val perms = java.nio.file.Files.getPosixFilePermissions(outputPath.toNIO)
+      perms.add(PosixFilePermission.GROUP_EXECUTE)
+      perms.add(PosixFilePermission.OWNER_EXECUTE)
+      perms.add(PosixFilePermission.OTHERS_EXECUTE)
+      java.nio.file.Files.setPosixFilePermissions(outputPath.toNIO, perms)
+    }
     PathRef(outputPath)
   }
 
--- a/clientserver/src/mill/clientserver/Client.java
+++ b/clientserver/src/mill/clientserver/Client.java
@@ -121,7 +121,7 @@ public class Client {
         while(ioSocket == null && System.currentTimeMillis() - retryStart < 1000){
             try{
                 ioSocket = ClientServer.isWindows?
-                        new Win32NamedPipeSocket("\\\\.\\pipe\\" + new File(lockBase).getName())
+                        new Win32NamedPipeSocket(ClientServer.WIN32_PIPE_PREFIX + new File(lockBase).getName())
                         : new UnixDomainSocket(lockBase + "/io");
             }catch(Throwable e){
                 Thread.sleep(1);
--- a/clientserver/src/mill/clientserver/Client.java
+++ b/clientserver/src/mill/clientserver/Client.java
@@ -1,9 +1,11 @@
 package mill.clientserver;
 
 import io.github.retronym.java9rtexport.Export;
-import org.scalasbt.ipcsocket.UnixDomainSocket;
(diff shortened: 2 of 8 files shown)


rockjam:
I'm not very familiar with client/server code in mill, so I guess @lihaoyi is a better match for review.

lihaoyi:
@rockjam sure I'll take a look

build.sc
```diff
@@ -236,11 +236,13 @@ object dev extends MillModule{
 
     write(outputPath, prependShellScript())
 
-    val perms = java.nio.file.Files.getPosixFilePermissions(outputPath.toNIO)
-    perms.add(PosixFilePermission.GROUP_EXECUTE)
-    perms.add(PosixFilePermission.OWNER_EXECUTE)
-    perms.add(PosixFilePermission.OTHERS_EXECUTE)
-    java.nio.file.Files.setPosixFilePermissions(outputPath.toNIO, perms)
+    if (!System.getProperty("os.name").toLowerCase.startsWith("windows")) {
```
lihaoyi: Does windows filesystem support a similar set of group/owner/other execute permissions? Or are those concepts just non-existent on windows?

clientserver/src/mill/clientserver/Client.java
```diff
@@ -100,16 +103,26 @@ public static int run(String lockBase,
...
 
-        UnixDomainSocket ioSocket = null;
+        Socket ioSocket = null;
 
         long retryStart = System.currentTimeMillis();
         while(ioSocket == null && System.currentTimeMillis() - retryStart < 1000){
             try{
-                ioSocket = new UnixDomainSocket(lockBase + "/io");
+                ioSocket = ClientServer.isWindows?
+                        new Win32NamedPipeSocket("\\\\.\\pipe\\" + new File(lockBase).getName())
```
lihaoyi: Can we put the `\\\\.\\pipe\\` string into a static constant in `ClientServer` so we don't have to copy paste it? That will also give us a place to add a short comment with a link to the windows named-pipe doc where this magic-string comes from

clientserver/src/mill/clientserver/Server.scala
```diff
@@ -37,18 +37,32 @@ class Server[T](lockBase: String,
...
-          val ioSocket = new UnixDomainServerSocket(lockBase + "/io")
-          val sockOpt = Server.interruptWith(
-            acceptTimeout,
-            new UnixDomainSocket(lockBase + "/io").close(),
-            ioSocket.accept()
-          )
+          val (serverSocket, sockOpt) = if (ClientServer.isWindows) {
+            val socketName = """\\.\pipe\""" + new File(lockBase).getName
+            val ioSocket = new Win32NamedPipeServerSocket(socketName)
+            (ioSocket, Server.interruptWith(
```
lihaoyi: Looks like most of the `Server.interruptWith` block is the same, except for the difference in the `new FooSocket` logic; could you limit the if-statement to initializing the ioSocket & a `close` callback, and leave the `sockOpt` definition outside the if-statement?

core/src/mill/util/OS.scala
```diff
@@ -0,0 +1,5 @@
+package mill.util
+
+object OS {
+  val isWindows: Boolean = System.getProperty("os.name").toLowerCase.startsWith("windows")
```
lihaoyi: Do we need this if we already have `scala.util.Properties.isWin`? Same for the version of this in `build.sc`

(Note that we should not use `scala.util.Properties.isWin` in the Mill client, since that only uses Java APIs to avoid classloading/initialization slowness)

clientserver/src/mill/clientserver/Server.scala
```diff
@@ -106,7 +120,13 @@ class Server[T](lockBase: String,
 
     t.interrupt()
     t.stop()
-    clientSocket.close()
+
+    if (ClientServer.isWindows) {
+      // Closing Win32NamedPipeSocket can take a second or more!
```
lihaoyi: If the previous `Win32NamedPipeSocket` hasn't fully closed, is it really OK for us to return from `handleRun` and possibly open a new `Win32NamedPipeSocket`? Maybe it is, but if so you should add a comment because it's not obvious from the code that having two sockets open at the same time is OK

lihaoyi:
@robby-phd overall looks good, left some comment

build.sc
robby-phd: It throws `UnsuportedOperationException` on Windows.

clientserver/src/mill/clientserver/Client.java
robby-phd: Sure.

clientserver/src/mill/clientserver/Server.scala
robby-phd: Sure.

core/src/mill/util/OS.scala
robby-phd: Somehow that escaped me. I'll change to use Scala's `isWin` except on the client.

clientserver/src/mill/clientserver/Server.scala
robby-phd: Yes, this is a hack. I tested this by calling mill (client) several times on quick successions and it seems to work OK (in my machine); I wonder whether closing the server socket helps speed up the process. Without that, there is a very perceivable lag (often ~3-5 seconds) between the last output from mill server and when the mill client exits. I can put in more comments.

robby-phd:
@lihaoyi merge?

lihaoyi:
@robby-phd Looks good to me, feel free to merge

PR #221, opened 2018-03-12
Fixes for non-interactive (client/server) mode on Java 9

robby-phd:
More fixes for #156.

PR #239, opened 2018-03-18
Java 9 optimization by caching rt.jar

robby-phd:
This PR adds Java 9 optimization by caching rt.jar.

On interactive mode, `mill.util.ClassLoader` is adjusted to have thread-local home path context enclosing the execution of `MainRunner` (if home context is not set, then `mill.util.ClassLoader.create` falls back to previous behavior). This approach was chosen in order to avoid making pervasive changes to many callers of `mill.util.ClassLoader.create` which would mean propagation of mill's home path originating from `Main.main0` such as to `Jvm`, `ScalaModule`, etc.

On client/server mode, this is provided by upgrading to Ammonite 1.0.5-7-f032887 which includes lihaoyi/Ammonite#773 (and changes made previously in #221).

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -69,7 +69,7 @@ trait MillModule extends MillPublishModule{ outer =>
 
 object clientserver extends MillModule{
   def ivyDeps = Agg(
-    ivy"com.lihaoyi:::ammonite:1.0.5-4-c0cdbaf",
+    ivy"com.lihaoyi:::ammonite:1.0.5-7-f032887",
     ivy"org.scala-sbt.ipcsocket:ipcsocket:1.0.0"
   )
   val test = new Tests(implicitly)
@@ -84,7 +84,7 @@ object core extends MillModule {
 
   def ivyDeps = Agg(
     ivy"com.lihaoyi::sourcecode:0.1.4",
-    ivy"com.lihaoyi:::ammonite:1.0.5-4-c0cdbaf",
+    ivy"com.lihaoyi:::ammonite:1.0.5-7-f032887",
     ivy"jline:jline:2.14.5"
   )
 
--- a/core/src/mill/util/ClassLoader.scala
+++ b/core/src/mill/util/ClassLoader.scala
@@ -2,9 +2,17 @@ package mill.util
 
 import java.net.{URL, URLClassLoader}
 
+import ammonite.ops.Path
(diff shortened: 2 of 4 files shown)


lihaoyi:
@robby-phd I'm not super comfortable with using a threadlocal for this; could we instead do the necessary plumbing to plumb the mill-classloader-home to the right place?

As far as I can tell you should be able to do this in a not-too-gross way by making `millClassloaderHome` available as part of a trait mixed into [mill.util.Ctx](https://github.com/lihaoyi/mill/blob/a9d4eea6f5b81cdbef84724e612bf07954e2673c/core/src/mill/util/Ctx.scala#L34), and then making `mill.util.ClassLoader.create` take an implicit of that trait.

Also, I don't understand why there's a different code path between `--interactive` and non-interactive modes; could you explain that a bit?

robby-phd:
@lihaoyi Heh, I somehow knew you would say that. TBH, I'm also uncomfortable with it, but I tried to propagate mill's home at first (similar to you what you described, although I was only propagating mill's home path), and I'm uncomfortable making pervasive changes to mill's codebase because without your inputs. So, I did the simplest thing and made the pull request anyway to solicit your reaction on this. 😃

Now that you said that it's better to do that ... 🚀

On the Client/Server, when adding support for Java 9 for that, I previously got an exception when forking the server process because it could not load `javax.script.ScriptEngineManager`, which necessitates passing rt.jar to the server's classpath, of which I of course then cached (note that this was after I added the rt.jar export in Ammonite). I'll see if the two code paths can be merged after implicitly propagating Ctx.

robby-phd:
@lihaoyi Ready for another review. Not as bad as I thought as apparently there are already a lot of implicit Ctx plumbing. I also removed the Client's side rt.jar caching; it is still necessary to add "java.class.path" to the server's classpath however.

robby-phd:
@lihaoyi The plumbing is really done now (I hope).

lihaoyi:
@robby-phd any chance you could help me take a look at https://travis-ci.org/lihaoyi/mill/builds/358707048? It seems the PR is failing on Java 9 only and I have no idea what's going wrong (Java 8 jobs seem fine)

lihaoyi:
It seems Ammonite is incorrectly caching the `build.sc` file when it should be invalidated, leading to `NoSuchMethodError`s

PR #243, opened 2018-03-19
Batch (.bat) file generation for assembly, launcher, and release

robby-phd:
This PR adds batch file generation for Windows, thus, enabling Window users to run mill without a sh environment (no repl support at the moment; see below). The resulting assembly and release .bat files are similar to the sh version where their corresponding jar file is embedded in it.

Notes:

* Shifting of command line argument for interactive mode is moved to Main.main because it's difficult to do in batch for arguments with special characters 

* Support for ANSI color is disabled by default in Windows 10 Console; it requires a native API call to enable it. It's best to use terminal replacement such as [ConEmu](https://conemu.github.io).

* mill repl does not work because Ammonite repl assumes [sh/Unix environment](https://github.com/lihaoyi/Ammonite/blob/ac754ee65a2760bf6c442ab9d5a30777757ed063/terminal/src/main/scala/ammonite/terminal/Utils.scala#L94). I think there is still value to offer non-repl .bat version as mill CLI tools are quite nice to use and the client/server mode now provides quick turnaround to probe mill build.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -261,8 +261,9 @@ object dev extends MillModule{
   }
 
   def assembly = T{
-    mv(super.assembly().path, T.ctx().dest / 'mill)
-    PathRef(T.ctx().dest / 'mill)
+    val filename = if (scala.util.Properties.isWin) "mill.bat" else "mill"
+    mv(super.assembly().path, T.ctx().dest / filename)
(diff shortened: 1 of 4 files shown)


lihaoyi:
@robby-phd this looks good to me, feel free to merge when ready

PR #247, opened 2018-03-19
Support scala milestone releases (with fixes for 2.13.0-M3)

rockjam:
This PR lets us use scala milestone releases, like `2.13.0-M3` in mill. Didn't work before

### DIFF
--- a/scalalib/src/mill/scalalib/ScalaModule.scala
+++ b/scalalib/src/mill/scalalib/ScalaModule.scala
@@ -98,13 +98,19 @@ trait ScalaModule extends mill.Module with TaskModule { outer =>
 
   def platformSuffix = T{ "" }
 
-  def scalaCompilerBridgeSources = T{
+  def scalaCompilerBridgeSources = T {
+    val (scalaVersion0, scalaBinaryVersion0) =
+      if (scalaVersion().startsWith("2.13."))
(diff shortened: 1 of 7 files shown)


scalalib/src/mill/scalalib/ScalaModule.scala
```diff
@@ -98,13 +98,19 @@ trait ScalaModule extends mill.Module with TaskModule { outer =>
 
   def platformSuffix = T{ "" }
 
-  def scalaCompilerBridgeSources = T{
+  def scalaCompilerBridgeSources = T {
```
rockjam: We do the same special treatment as here: https://github.com/sbt/zinc/blob/1.x/build.sbt#L391

lihaoyi: @rockjam seems like a pretty hack way of doing it; is there some way where we won't have to change the code when 2.13.0-final is out? Could we look for the trailing `-Mn` isntead of the leading `2.13.`?

rockjam: sure

rockjam: Fixed it

lihaoyi:
Looks good to me!

Issue #227, opened 2018-03-14
Add `mill clean`


francisdb:
Would be more friendly for newcomers than `rm -rf out/` (always dangerous for typos and depends on work dir)

antonkulaga:
It would be great to have it as a project-based command. I often get caching-linking problems with mill ScalaJS projects and it is way faster to clean the scalajs project out by mill command rather than going to appropriate place in out with bash

PR #253, opened 2018-03-21
Adding clean as a default task

msoule:


msoule:
First time submitting a PR for this project. I have a few questions:

1. Unit tests for MainModule? I didn't see any.
2. Is it acceptable that I used `ammonite.ops.%%`?
3. Can `ammonite.ops.%%` throw an exception?
4. I opted for multi clean targets because it gave me an elegant way to do clean all (by passing no arguments). Is there a use case for nested clean or transitive dependency clean?
5. What are the styling standards for this project?

msoule:
Also, one quirk is that clean does not clean up after itself. So after running clean you are left with `out/clean`. Not sure if there is anything I can do about that.

msoule:
Fixes #227

main/src/mill/main/MainModule.scala
```diff
@@ -173,4 +175,17 @@ trait MainModule extends mill.Module{
...
     }
   }
+
+  /**
+    * Deletes the given targets from the out directory. Providing no targets will clean everything.
+    */
+  def clean(targets: String*) = mill.T.command {
+    val result = targets match {
+      case Nil =>
+        List(ammonite.ops.%%('rm, "-rf", OutDir)(ammonite.ops.pwd).out.string.trim())
```
lihaoyi: probably should use `ammonite.ops.rm` here

main/src/mill/main/MainModule.scala
```diff
@@ -173,4 +175,17 @@ trait MainModule extends mill.Module{
...
+      case _ =>
+        targets.map(cleanTarget => ammonite.ops.%%('rm, "-rf", s"$OutDir/$cleanTarget")(ammonite.ops.pwd).out.string.trim())
```
lihaoyi: This won't work for targets like `foo.bar.baz`

main/src/mill/main/MainModule.scala
msoule: Aw okay didn't even know that was a thing.

main/src/mill/main/MainModule.scala
msoule: Yeah I wanted to ask about this first to confirm that it is desirable. Will mark as WIP until I've added this.

main/src/mill/main/MainModule.scala
msoule: Good point. I did note that other functions in this file do the following:

```
    val resolved = RunScript.resolveTasks(
      mill.main.ResolveMetadata, evaluator, targets, multiSelect = true
    )
```
And then I extract the names from `resolved`. Would that be sufficient? I will play with it later.

francisdb: would it not be enough to validate that what you pass to rm -rf contains `out`?

francisdb: And maybe that the parent folder contains build.sc?

msoule: Yes you're right. I guess I'm wondering if there is also value is validating that the passed targets are indeed valid targets.

francisdb: You probably want to convert the path to a absolute path first before applying tests.

main/src/mill/main/MainModule.scala
```diff
@@ -173,4 +179,34 @@ trait MainModule extends mill.Module{
       }
     }
   }
+
+  /**
+    * Deletes the given targets from the out directory. Providing no targets will clean everything.
+    */
+  def clean(targets: String*) = mill.T.command {
+    val result = Try {
+      if (ammonite.ops.ls(ammonite.ops.pwd).contains(ammonite.ops.pwd / BuildFile)) {
```
lihaoyi: This can just be `exists(pwd / "build.sc")`

main/src/mill/main/MainModule.scala
```diff
@@ -173,4 +179,34 @@ trait MainModule extends mill.Module{
...
+        targets match {
+          case Nil =>
+            remove(List(ammonite.ops.pwd / OutDir))
+          case _ =>
+            remove(targets
+              .map(_.replace(".", "/"))
```
lihaoyi: You should use `resolveDestPaths` (https://github.com/lihaoyi/mill/blob/master/core/src/mill/eval/Evaluator.scala#L342) to figure out what targets to clean, to ensure it stays in sync with any future changes

main/src/mill/main/MainModule.scala
```diff
@@ -173,4 +179,34 @@ trait MainModule extends mill.Module{
       }
     }
   }
+
+  /**
+    * Deletes the given targets from the out directory. Providing no targets will clean everything.
+    */
+  def clean(targets: String*) = mill.T.command {
+    val result = Try {
```
lihaoyi: No idea why you are wrapping this in a `Try`; you just need to return a simple `Result` object

main/src/mill/main/MainModule.scala
```diff
@@ -173,4 +179,34 @@ trait MainModule extends mill.Module{
...
+      }
+    }
+    result match {
+      case Success(r) => r
+      case Failure(error) => Result.Exception(error, new Result.OuterStack(error.getStackTrace))
+    }
+  }
+
+  private def remove(paths: List[Path]): Result[String] = {
+    paths.filter(ammonite.ops.exists).foreach(ammonite.ops.rm)
```
lihaoyi: You don't need the `filter`; `rm` already just ignores a path if there' nothing to remove

lihaoyi:
This should have at least one unit test somewhere before merging

msoule:
@lihaoyi Where are the unit tests for the other functions in this file? Are there any?

main/src/mill/main/MainModule.scala
msoule: The ammonite functions call things which can throw exceptions. Is it acceptable to let those exceptions be thrown?

main/src/mill/main/MainModule.scala
msoule: Oh nice.

main/src/mill/main/MainModule.scala
msoule: Will do.

msoule:
Also @lihaoyi Can you explain the build failures I'm seeing. The same test passes locally.

lihaoyi:
Closing due to inactivity

msoule:
@lihaoyi Can you answer some of the questions I have above?

Issue #214, opened 2018-03-09
Option to exclude transitive ivy dependencies


raskam:
Currently `Dep` only has an option to configure coursier attributes, but there's no way to configure coursier exclusions.

Right now I'm using implicits to achieve this:

```scala
  implicit class RichDep(dep: Dep) {
    def exclude(exclusions: (String, String)*): Dep = {
      dep match {
        case d: Dep.Java => d.copy(dep = d.dep.copy(exclusions = d.dep.exclusions ++ exclusions))
        case d: Dep.Scala => d.copy(dep = d.dep.copy(exclusions = d.dep.exclusions ++ exclusions))
        case d: Dep.Point => d.copy(dep = d.dep.copy(exclusions = d.dep.exclusions ++ exclusions))
      }
    }
  }
```

And usage:

```
ivy"org.quartz-scheduler:quartz:${V.quartzV}"
  .exclude(
    "com.mchange" -> "c3p0",
    "com.zaxxer" -> "HikariCP-java6"
  )
```

Better would be of course to modify the `Dep` directly. I could make it into PR if you like the way it is implemented.

lihaoyi:
@raskam you can always fall back to passing in a `coursier.Dependency` to `ivyDeps`; could you try configuring one with the exclusion you want and seeing if that works?

PR #254, opened 2018-03-21
add exclusions and custom configuration to dependencies

shengc:
partly borrow the idea from #214, but also reflect the exclusions in ivy and pom generation.

### DIFF
--- a/docs/pages/2 - Configuring Mill.md
+++ b/docs/pages/2 - Configuring Mill.md
@@ -71,6 +71,24 @@ def repositories = super.repositories ++ Seq(
 )
 ```
 
+To add custom resolvers to the initial bootstrap of the build, you can create a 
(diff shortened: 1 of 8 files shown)


lihaoyi:
@shengc can we have a unit test for this?

shengc:
@lihaoyi sure, I have to do this after hours. Is the unit test supposed to be under `integration/test`, or somewhere else ?

lihaoyi:
@shengc probably one unit test in https://github.com/lihaoyi/mill/tree/master/scalalib/test/src/mill/scalalib to test that resolving does the right thing and the classpath is properly missing the jar that got excluded

shengc:
@lihaoyi tests are in.

lihaoyi:
Looks good to me, thanks!

PR #263, opened 2018-03-28
use repositories from the mill build to bootstrap idea proj generation

shengc:
this is for users behind proxy to be able to generate idea project using `mill`


lihaoyi:
@shengc looks good to me, thanks!

PR #264, opened 2018-03-28
universal script generation for assembly, launcher, and release that run on linux and windows

LolHens:
As a follow-up to lihaoyi/Ammonite#781 this pr generates scripts that can run on linux shell and windows cmd.
It merges the shell and batch scripts introduced in #243.
Feedback is always welcome!

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -205,34 +205,34 @@ object integration extends MillModule{
   def forkArgs = testArgs()
 }
 
-def launcherScript(isWin: Boolean,
-                   jvmArgs: Seq[String],
-                   classPath: Agg[String]) = {
-  def universalScript(shellCommands: String,
-                      cmdCommands: String,
-                      shebang: Boolean = true): String = {
+private def universalScript(shellCommands: String,
+                            cmdCommands: String,
+                            shebang: Boolean = false): String = {
+  Seq(
+    if (shebang) "#!/usr/bin/env sh" else "",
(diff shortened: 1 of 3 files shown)


LolHens:
The error you are running into indeed seems very strange. I ran the version I just pushed on my windows system and on a redhat linux machine and it worked fine.
I had to change the approach because before I was prepending a "::" to every line of the shell script so that the cmd skips them however this does not work at all for switch case statements. The new approach tells the cmd to just jump over the shell script and should work for any script.
I removed the releaseAll and releaseBatch and now uploadToGithub only generates "mill".
Also I set the shebang line to be off by default.
Now I am just waiting to see what the tests will say :)

LolHens:
Unfortunately I don't have access to that machine right now so I ran the tests on a local debian installation but I couldn't get all of the tests to succeed there.
I made a new commit but changed nothing just to restart the build and it seems to have worked where it previously failed at publishVersion in build.sc:360.
I think there might be some other bug involved that is not related to this pr.
Could you try again to execute the generated scripts?

robby-phd:
I just tried it again locally on macOS and linux, and tests passed (didn't clean out `~/.mill` before). The changes look good to me, thanks.

LolHens:
Very nice, thanks! Should I make a pr to change the generated script in ammonite according to this so that the code is equivalent?

robby-phd:
That would be great!

lihaoyi:
Oh ok, got it

PR #271, opened 2018-04-01
Add a TwirlModule to compile Twirl templates

Mogztter:
Based on @pleira implementation

This is a work in progress:

 * [x] Find a better solution to solve the `java.lang.ClassNotFoundException: scala.util.parsing.input.Positional`
 * [x] An exception is thrown when `TwirlCompiler.compile` function is invoked using Reflection
 * [x] Unable to call the `compile` method with a primitive boolean using reflection: https://github.com/playframework/twirl/blob/f4ed321dee30c977b7f303ef1229169180beab45/compiler/src/main/java/play/japi/twirl/compiler/TwirlCompiler.java#L38
 * [x] Find a way to use the Scala API because the Java API is only available in Twirl 1.3+

For now the main issue is that an exception is thrown when `TwirlCompiler.compile` is invoked.
Oddly tests are running fine in Intellij IDEA. In this case the reflection is calling `JavaMirrors.JavaVanillaMethodMirror.apply(args: Any*)`. But when I run the tests using `mill twirllib.test`, the reflection is calling `JavaMirrors.JavaTransformingMethodMirror.apply(args: Any*)`.

I've also tried to use Java reflection but without success:

```scala
val clt = cl.loadClass("play.twirl.compiler.TwirlCompiler")
val module = cl.loadClass("play.twirl.compiler.TwirlCompiler$").getField("MODULE$")
val compileMethod = clt.getMethods.find(_.getName == "compile").get
compileMethod.invoke(module, 
    source,
    sourceDirectory,
    generatedDirectory,
    formatterType,
    additionalImports,
    constructorAnnotations,
    codec,
    inclusiveDot) // error: the result type of an implicit conversion must be more specific than AnyRef
```

Help, feedback or review will be greatly appreciated :smile:

### DIFF
--- a/main/test/src/mill/define/CacherTests.scala
+++ b/main/test/src/mill/define/CacherTests.scala
@@ -17,7 +17,7 @@ object CacherTests extends TestSuite{
   }
   object Middle extends Middle
   trait Middle extends Base{
-    def value = T{ super.value() + 2}
+    override def value = T{ super.value() + 2}
     def overriden = T{ super.value()}
   }
   object Terminal extends  Terminal
--- a/twirllib/src/mill/twirllib/TwirlWorker.scala
+++ b/twirllib/src/mill/twirllib/TwirlWorker.scala
@@ -3,66 +3,104 @@ package twirllib
 
 import java.io.File
 import java.net.URLClassLoader
-import ammonite.ops.{Path,ls}
 
+import ammonite.ops.{Path, ls}
+import mill.eval.PathRef
+import mill.scalalib.CompilationResult
 
-class TwirlWorker  {
+import scala.io.Codec
+import scala.reflect.runtime.universe._
+
+class TwirlWorker {
 
   private var twirlInstanceCache = Option.empty[(Long, TwirlWorkerApi)]
 
   private def twirl(twirlClasspath: Agg[Path]) = {
-    val classloaderSig =
-      twirlClasspath.map(p => p.toString().hashCode + p.mtime.toMillis).sum
+    val classloaderSig = twirlClasspath.map(p => p.toString().hashCode + p.mtime.toMillis).sum
     twirlInstanceCache match {
       case Some((sig, instance)) if sig == classloaderSig => instance
       case _ =>
-        val cl = new URLClassLoader(twirlClasspath.map(_.toIO.toURI.toURL).toArray)
-        val clt = cl.loadClass("play.twirl.compiler.TwirlCompiler")
-        val instance = clt.getDeclaredConstructor().newInstance()
-          .asInstanceOf[TwirlWorkerApi]
+        val instance = new TwirlWorkerApi {
+          override def compileTwirl(source: File,
+                                    sourceDirectory: File,
+                                    generatedDirectory: File,
+                                    formatterType: String,
+                                    additionalImports: Seq[String] = Nil,
+                                    constructorAnnotations: Seq[String] = Nil,
+                                    codec: Codec = Codec(scala.util.Properties.sourceEncoding),
+                                    inclusiveDot: Boolean = false) {
+            val cl = new URLClassLoader(twirlClasspath.map(_.toIO.toURI.toURL).toArray)
+            val runtime = runtimeMirror(cl)
+            val moduleMirror = runtime.reflectModule(runtime.staticModule("play.twirl.compiler.TwirlCompiler"))
+            val moduleInstance = moduleMirror.instance
+            val instanceMirror = runtime.reflect(moduleInstance)
+            val compileSymbol = instanceMirror.symbol.typeSignature.member(TermName("compile")).asMethod
+            // FIXME: type erasure, unable to call "play.twirl.compiler.TwirlCompiler.compile" function
+            // https://github.com/playframework/twirl/blob/dd56444cf9ea2f95ef3961d3af911561f846df50/compiler/src/main/scala/play/twirl/compiler/TwirlCompiler.scala#L167
+
+            // scala.MatchError: List() (of class scala.collection.immutable.Nil$)
+            // at scala.reflect.runtime.JavaMirrors$JavaMirror$DerivedValueClassMetadata.unboxer$lzycompute(JavaMirrors.scala:270)
+            // at scala.reflect.runtime.JavaMirrors$JavaMirror$DerivedValueClassMetadata.unboxer(JavaMirrors.scala:269)
+            // at scala.reflect.runtime.JavaMirrors$JavaMirror$MethodMetadata.paramUnboxers(JavaMirrors.scala:416)
+            // at scala.reflect.runtime.JavaMirrors$JavaMirror$JavaTransformingMethodMirror.apply(JavaMirrors.scala:435)
+            instanceMirror.reflectMethod(compileSymbol)(source,
+              sourceDirectory,
+              generatedDirectory,
+              formatterType,
+              additionalImports,
+              constructorAnnotations,
+              codec,
+              inclusiveDot)
+          }
+        }
         twirlInstanceCache = Some((classloaderSig, instance))
         instance
     }
   }
-  
-  def compile(twirlClasspath: Agg[Path],
-              sourceDirectories: Seq[Path],
-              dest: Path) : Unit = {
-    val compiler = twirl(twirlClasspath)
-    def compileTwirlDir(inputDir: Path) = {
-      val twirlFiles = ls.rec(inputDir).filter(_.name.matches(".*.scala.(html|xml|js|txt)"))
-      twirlFiles.map { tFile =>
-        val extFormat = twirlExtensionFormat(tFile.name)
-        compiler.compile( tFile.toIO,
-                          inputDir.toIO,
-                          dest.toIO,
-                          s"play.twirl.api.${extFormat}"
-        )
-      }
-    }
-    sourceDirectories.map(compileTwirlDir)
-  }
- 
- private def twirlExtensionFormat(name: String) =
-        if (name.endsWith("html"))     "HtmlFormat"
-        else if (name.endsWith("xml")) "XmlFormat"
-        else if (name.endsWith("js"))  "JavaScriptFormat"
-        else "TxtFormat"
 
+  def compile(twirlClasspath: Agg[Path], sourceDirectories: Seq[Path], dest: Path)
+             (implicit ctx: mill.util.Ctx): mill.eval.Result[CompilationResult] = {
+    val compiler = twirl(twirlClasspath)
+
+    def compileTwirlDir(inputDir: Path) {
+      ls.rec(inputDir).filter(_.name.matches(".*.scala.(html|xml|js|txt)"))
+        .foreach { template =>
+          val extFormat = twirlExtensionFormat(template.name)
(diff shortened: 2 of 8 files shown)


main/test/src/mill/define/CacherTests.scala
```diff
@@ -17,7 +17,7 @@ object CacherTests extends TestSuite{
   }
   object Middle extends Middle
   trait Middle extends Base{
-    def value = T{ super.value() + 2}
+    override def value = T{ super.value() + 2}
```
Mogztter: Intellij IDEA is complaining about this line (preventing me from running the tests) but Mill can compile the code just fine :neutral_face:

rockjam: Yep, that's fine in mill, because there is a compiler plugin that makes `override` word optional

Mogztter: Thanks for the explanation.
Do you know if there is a way to configure intellij IDEA in order to use this compiler plugin for this file ?

rockjam: Technically it should be possible, If I understand correctly, you need to include http://search.maven.org/#artifactdetails%7Ccom.lihaoyi%7Cmill-moduledefs_2.12%7C0.1.7%7Cjar in classpath and add "-P:auto-override-plugin". to scalac options.

rockjam:
@Mogztter I'll take a look at those `ClassNotFoundException` issues when time permits. It's not clear what's wrong from first look

Mogztter:
>  I'll take a look at those ClassNotFoundException issues when time permits. It's not clear what's wrong from first look

@rockjam Ok. but I'm no longer sure that this issue can be reproduced with `mill twirllib.test` maybe it's an issue with the Intellij IDEA tests' classpath. I will do some more testing tomorrow 😉

Mogztter:
I'm now using the Java API (available in Twirl 1.3+):

https://github.com/playframework/twirl/blob/f4ed321dee30c977b7f303ef1229169180beab45/compiler/src/main/java/play/japi/twirl/compiler/TwirlCompiler.java#L31

Though, I'm unable to call a method with a boolean primitive type using reflection, so `codec` and `inclusiveDot` parameters will not be available.

rockjam:
> is it possible to do remote debugging ?

Honestly, I didn't try. You can try
`JAVA_OPTS="-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=1044" mill -i twirllib.testLocal`

Mogztter:
@rockjam I've created a simple project where I'm using Java reflection to call the Twirl compiler and it's working fine: https://github.com/Mogztter/mill-twirl-compiler/blob/master/app/src/Main.scala

If I'm using the same code inside the `TwirlWorker`, the following exception is thrown: `java.lang.IllegalArgumentException: argument type mismatch`.

I've checked both classpath and I didn't see any difference.
Please not that I had to add the following dependency `scala-parser-combinators_2.12-1.1.0.jar` in https://github.com/Mogztter/mill-twirl-compiler/blob/master/build.sc#L10

Do you reproduce this issue ? I must be doing something wrong but I can't see what... thanks again for your help :+1:

Mogztter:
Indeed it will make the integration easier but I don't think there is a CLI.
The Twirl source code is here: https://github.com/playframework/twirl

I'm really clueless about this exception since I can't reproduce it outside Mill.

rockjam:
I'll try code from your PR to figure out what's going on with it in next couple days

Mogztter:
@rockjam Did you have a chance to reproduce the issue ?

Mogztter:
Rebased on master

rockjam:
@Mogztter sorry, didn't have a chance to look at it yet. I'll try to take a look this weekend. Is it ok for you?

Mogztter:
Sure thanks for taking the time.
In the mean time I will fix the build failure 😀

rockjam:
@Mogztter I couldn't reproduce the`ClassNotFoundException`  issue. I left some comments regarding PR itself. Apart from that it looks fine

Mogztter:
@rockjam Thanks for your review!

> I couldn't reproduce theClassNotFoundException issue.

You were able to call the Scala API https://github.com/playframework/twirl/blob/881b41e58350320570e6d3786cca87db5d04ecb4/compiler/src/main/scala/play/twirl/compiler/TwirlCompiler.scala#L167 ?

>  I left some comments regarding PR itself. Apart from that it looks fine

Ok I will fix them :+1:

rockjam:
@Mogztter I actually did succeed with scala API of twirl. Here is a commit that uses scala API, but passes default values of methods to compile method. You can take a look https://github.com/rockjam/mill/commit/0daf9caaa72f419e95fda7db11cbe03437f7dd56

I also included scala parser-combinator library in classpath, since otherwise it failed.

rockjam:
@Mogztter and here is the second take that removes reflective calls to default methods. But I'm not sure that @lihaoyi will aprove such tricks with classloading, since we mix mill's classloader with classloader of twirl, and we may go into trouble when `mill` and `twirl` will have different scala versions. https://github.com/rockjam/mill/commit/dc546f60ef89981ebfda2669fb1345d405d4f47b

Mogztter:
>  I actually did succeed with scala API of twirl. Here is a commit that uses scala API, but passes default values(from reflected TwirlCompiler class) of parameters to compile method. You can take a look here: rockjam/mill@0daf9ca

Oh I see... 🤔 
Is it possible to invoke the "default value function" with the method parameter ? I mean if `additionalImports` or `constructorAnnotations` is defined in the TwirlWorker, I want these parameters to be passed to the TwirlCompiler API.

> I also included scala parser-combinator library in classpath, since otherwise it failed. with ClassNotFoundException

So it's not just me 😌 

> nd here is the second take that removes reflective calls to default methods. But I'm not sure that @lihaoyi will aprove such tricks with classloading, since we mix mill's classloader with classloader of twirl, and we may go into trouble when mill and twirl will have different scala versions. rockjam/mill@dc546f6

Indeed, we should not mess with the classloader 😉 
I think the first solution is good as long as we can pass method parameters to the TwirlCompiler API, namely:

* `additionalImports`
* `constructorAnnotations`
* `codec`
* `inclusiveDot`

Mogztter:
Sorry I've been busy lately, I will try to go back to this pull request next week.
The next step will be to try to compile a simple Play! application with a Twirl template (to make sure that everything is working fine together) :smile:

rockjam:
Sounds great. We could add a sample play app in our integration test suite after this PR gets merged

Mogztter:
> Sounds great. We could add a sample play app in our integration test suite after this PR gets merged

That would be awesome :+1: 
But I think we should do it first because we will for sure find some limitations with the current implementation.

We can use this template https://github.com/playframework/play-scala-seed.g8 ?

rockjam:
I guess we could. Or we could port some real play application, built with mill. But let's do it in separate PR

Mogztter:
Ok 👌

I've integrated your commit and rebased on master so I think we are ready.
I can squash the commits if you want ?

One last question about the API, do you think we should keep the `twirl` prefix ?

> Or we could port some real play application, built with mill

Yes I guess we could. A popular Play application on GitHub is Lichess: https://github.com/ornicar/lila

But it will be quite challenging to build it with Mill 😋

lihaoyi:
No need to squash, Github can squash during the merge through the UI whenever @rockjam is ready.

Lichess seems like a bit ambitious, probably should start with something smaller like Hello World =P

rockjam:
> One last question about the API, do you think we should keep the twirl prefix ?

Do you mean twirl prefix in package names? If yes, lets keep it.

I will take a look at this PR this weekend

Mogztter:
@rockjam No I meant the prefix in function names: 

* `twirlVersion` -> `version`
* `twirlSources` -> `sources`
* `twirlClasspath` -> `classpath`
* `compileTwirl` -> `compile`

Also I noticed that `mill dev.assembly` does not contain `TwirlModule`, I guess we need to add `twirllib` here https://github.com/lihaoyi/mill/blob/032ae8b5138863c1523cf34a612ba1b19870713b/build.sc#L294

Mogztter:
Ok should we rename `compileTwirl` tout `twirlCompile` ?

> Lichess seems like a bit ambitious, probably should start with something smaller like Hello World =P

I agree.
We should probably also need to implement a routes compiler : https://github.com/playframework/playframework/blob/master/framework/src/sbt-plugin/src/main/scala/play/sbt/routes/RoutesCompiler.scala

rockjam:
@Mogztter created separate issue for play routes compiler: https://github.com/lihaoyi/mill/issues/324

> Ok should we rename compileTwirl tout twirlCompile ?

I don't have strong preference. `compileTwirl` sounds fine

twirllib/src/mill/twirllib/TwirlWorker.scala
```diff
@@ -0,0 +1,109 @@
...
+                                    formatterType: String,
+                                    additionalImports: Seq[String] = Nil,
+                                    constructorAnnotations: Seq[String] = Nil,
+                                    codec: Codec = Codec(Properties.sourceEncoding),
+                                    inclusiveDot: Boolean = false) {
+            val o = compileMethod.invoke(null, source,
+              sourceDirectory,
+              generatedDirectory,
+              formatterType,
+              defaultAdditionalImportsMethod.invoke(additionalImports),
```
rockjam: How does it work, can you explain a little bit? Why do we invoke method with `additionalImports` parameter?

Mogztter: @rockjam Good catch! It's not working :disappointed: 
`defaultAdditionalImportsMethod` is a method that does not take any argument and the call is wrong (ie. the first argument must be the object):

```scala
defaultAdditionalImportsMethod.invoke(null, additionalImports)
```

If I replace `defaultAdditionalImportsMethod.invoke(null)` with `additionalImports` then the reflection does not work anymore and the following exception is thrown:

```
java.lang.IllegalArgumentException: argument type mismatch
```

Back to square one I guess :cry: 
So it's working but we cannot define/override `additionalImports`, `constructorAnnotations`, `codec`, or `inclusiveDot`.

Mogztter: Interestingly, `additionalImports` is an instance of `scala.collection.immutable.$colon$colon` not `scala.collection.immutable.Seq[String]`...

rockjam: We can keep the defaults for these arguments for now, and merge this PR like that. And we can address modifying of these arguments in subsequent PR. How about that?

Mogztter: Ok, I will revert my changes to invoke the method without argument

Mogztter: @rockjam Done! I've created (private) functions in `TwirlModule` and add a comment so when we know how to override these arguments, everything will be ready.

rockjam:
@Mogztter travis fails for some reason. Can you rebase it on master to trigger CI?

Mogztter:
@rockjam Done!

rockjam:
@Mogztter thank you

PR #278, opened 2018-04-03
Move interactiveSubprocess to Util

guillaumebort:
Currently forking an interactive subprocess is not possible because of the way standard IO are captured by mill. I have the feeling that something is missing here because it is probably a common use case.

In the meantime I have found that `JVM.baseInteractiveSubprocess` does the job but it is a bit strange to call something in `JVM` to fork a non JVM process, so I propose to move it to `Util`.

If you think that there are better ways to solve this, especially if you think there is a way to make ammonite ops work for interactive process in mill I could give it a try!

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -303,7 +303,7 @@ object dev extends MillModule{
       case wd0 +: rest =>
         val wd = Path(wd0, pwd)
         mkdir(wd)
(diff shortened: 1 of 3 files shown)


guillaumebort:
Ok this is probably not needed as `scala.sys.process` works as expected after all. I'm still wondering how it could work with `ammonite.ops`.

PR #283, opened 2018-04-07
Merge April 1 - 6, 2018 changes on master

robby-phd:
Branch `master` was recently force pushed and it removed some changes made on April 1-6.
If the removal is unintentional, this PR restores/merges those changes with current `master`.


lihaoyi:
turned on master branch protection to avoid further screwups

robby-phd:
No problem. Glad that I still have some of the git history locally.
Note that this only restores/merges up to b920052 (April 3).

PR #287, opened 2018-04-07
Pass scalac plugins to scaladoc

### LINKED ISSUE #282: Macro expansion seems to fail when running Scaladoc
Related to #275 (awesome fix by the way, thank you!) — macros seem to be expanded properly when executing `compile`, `run`, or `console`, but scaladoc is complaining about them being missing. This also causes e.g. `publishLocal` to fail since it relies on `docJar`. Minimal example:
`build.sc`:
```scala
import mill._, mill.scalalib._

  def scalaVersion = "2.11.8"

  def ivyDeps = Agg(
    ivy"com.github.julien-truffaut::monocle-macro::1.4.0"
  )
  def scalacPluginIvyDeps = super.scalacPluginIvyDeps() ++ Agg(
    ivy"org.scalamacros:::paradise:2.1.0"
  )

  override def mainClass = Some("Main")
}
```
`foo/src/Main.scala`:
```scala
import monocle.macros._
@Lenses case class Foo(bar: String)
object Main {
  final def main(args: Array[String]): Unit = {
    println(Foo.bar.get(Foo("bar")))
  }
}
```
Running `mill foo.run` prints out `bar` as expected. However, `mill foo.docJar` produces:
```
[15/15] foo.docJar 
/Users/julian/Documents/scratch/test/foo/src/Main.scala:5: error: value bar is not a member of object Foo
    println(Foo.bar.get(Foo("bar")))
                ^
model contains 3 documentable templates
one error found
1 targets failed
foo.docJar ammonite.ops.InteractiveShelloutException
    mill.modules.Jvm$.subprocess(Jvm.scala:180)
    mill.scalalib.ScalaModule.$anonfun$docJar$2(ScalaModule.scala:226)
    mill.scalalib.ScalaModule$$Lambda$2307/1935861231.apply(Unknown Source)
    mill.define.ApplyerGenerated$$Lambda$2232/2014319819.apply(Unknown Source)
    mill.define.Task$MappedDest.evaluate(Task.scala:348)
```
The failure is being thrown during the Scaladoc `subprocess` call in `docJar`. I tried adding `scalacPluginClasspath().map(_.path)` to the classpath in the subprocess call but the problem remained.

julianmichael:
Fixes #282. Passing the flags as @lihaoyi suggested seemed to do the trick (thanks!), and docJar and publishing now seem to be working. I added a small test case with the example from the issue, which fails before the change and succeeds after.

### DIFF
--- a/scalalib/src/mill/scalalib/ScalaModule.scala
+++ b/scalalib/src/mill/scalalib/ScalaModule.scala
@@ -127,7 +127,8 @@ trait ScalaModule extends JavaModule { outer =>
       if p.isFile
     } yield p.toNIO.toString
 
-    val options = Seq("-d", javadocDir.toNIO.toString, "-usejavacp")
+    val pluginOptions = scalacPluginClasspath().map(pluginPathRef => s"-Xplugin:${pluginPathRef.path}")
+    val options = Seq("-d", javadocDir.toNIO.toString, "-usejavacp") ++ pluginOptions
 
     if (files.nonEmpty) subprocess(
       "scala.tools.nsc.ScalaDoc",
--- a/scalalib/test/resources/hello-world-macros/core/src/Main.scala
+++ b/scalalib/test/resources/hello-world-macros/core/src/Main.scala
@@ -1,0 +1,5 @@
+import monocle.macros._
+@Lenses case class Foo(bar: String)
+object Main extends App {
+  println(Foo.bar.get(Foo("bar")))
+}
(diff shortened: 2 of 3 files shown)


lihaoyi:
Geez more compiler interface problems :/

@julianmichael can you try poking/shaking it to see if you can make any progress? I might be able to help dig into it and get it fixed, but probably not anytime soon

julianmichael:
I don't think I have the time unfortunately... plus this works as a stopgap solution for me for now (I'm not using JDK 9, which...seems to be the problem?). I will try to come back to this at some point though.

lihaoyi:
Otherwise don't worry about it, just leave this here and I'll find time to figure it out

lihaoyi:
@robby-phd any idea why these tests appear to be sad on java 9?

lihaoyi:
merged in 3eb623542a664c1a924d4cad4ee1cdaef8581f8c

PR #294, opened 2018-04-10
can skip doc and src jars publish

shengc:


lihaoyi:
Could this be a flag on `publishLocal` instead?

shengc:
Yes I will tweak the PR tonight.

shengc:
@lihaoyi does it look good to you now ?

lihaoyi:
weren't you going to make it be a flag on `publishLocal` rather than their own, separate targets?

shengc:
Is it a hiccup ? ci build was fine.

lihaoyi:
kicked the ci build and it failed with the same output

shengc:
it passed build on `travis-ci`, `cygwin` and `mysys2` , but failed on windows shell. This is where the trace reports what is broken with the PR introduced,
https://github.com/shengc/mill/blob/master/core/src/mill/define/Module.scala#L67

I don't have a windows machine to debug this. From the surface value of the error, it is not clear how the error could happen.

lihaoyi:
Neither do I; perhaps you can add printlns and iterate on travis?

robby-phd:
I previously encountered a similar exception under Java 9; try adding the following line after default's publishLocal:

```cmd
rd /s /q %USERPROFILE%\.mill &&
```

lihaoyi:
closing due to inactivity

PR #296, opened 2018-04-11
Fix docs re: GenIdea/idea command

artempyanykh:
I found that the module name has changed, but the docs still refer to `GenIdeaModule`


rockjam:
Thank you

PR #298, opened 2018-04-15
Mark test module as test sources; change sbt module root

### LINKED ISSUE #201: `src/main` folders, etc interpreted as a package names
When intellij project files are regenerated, it seems to set `src` as the sources root, and not e.g. src/main/scala, which makes all new packages created have a prefix of `main.scala.whatever....`

rockjam:
This PR fixes #201
It's achieved by introducing `intellijModulePath`, which is different from `millSourcePath` for sbt/maven module's tests.
For instance:
module path in Intellij is: `module`
module's test path in Intellij is: `module/src/test`.

Meanwhile `millSourcePath` didn't change

### DIFF
--- a/scalalib/src/mill/scalalib/GenIdeaImpl.scala
+++ b/scalalib/src/mill/scalalib/GenIdeaImpl.scala
@@ -217,7 +217,7 @@ object GenIdeaImpl {
       val isTest = mod.isInstanceOf[TestModule]
 
       val elem = moduleXmlTemplate(
-        mod.millModuleBasePath.value,
+        mod.intellijModulePath,
         scalaVersionOpt,
         Strict.Agg.from(resourcesPathRefs.map(_.path)),
         Strict.Agg.from(normalSourcePaths),
--- a/scalalib/src/mill/scalalib/GenIdeaImpl.scala
+++ b/scalalib/src/mill/scalalib/GenIdeaImpl.scala
@@ -214,6 +214,8 @@ object GenIdeaImpl {
         mod.generatedSources.ctx.segments
       )
 
+      val isTest = mod.isInstanceOf[TestModule]
+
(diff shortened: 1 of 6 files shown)


scalalib/src/mill/scalalib/GenIdeaImpl.scala
```diff
@@ -321,7 +324,9 @@ object GenIdeaImpl {
                         compileOutputPath: Path,
                         generatedSourceOutputPath: Path,
                         libNames: Strict.Agg[String],
-                        depNames: Strict.Agg[String]) = {
+                        depNames: Strict.Agg[String],
+                        isTest: Boolean
```
rockjam: This allows us to display test sources nicely in sources tree, 
image

and split them from ordinary sources in search

scalalib/src/mill/scalalib/MiscModule.scala
```diff
@@ -65,13 +65,13 @@ trait MavenModule extends JavaModule{outer =>
   )
   override def resources = T.sources{ millSourcePath / 'src / 'main / 'resources }
   trait Tests extends super.Tests with MavenTests {
-    override def millSourcePath = outer.millSourcePath
+    override def millSourcePath = outer.millSourcePath / 'src / 'test
```
rockjam: I thought that we can make `parentModule/src/test` as a root for test module. @lihaoyi  Don't you mind that?

rockjam: Or maybe we can do special handling of base path in GenIdea itself, cause this change may break current builds, when you rely on path of your test module. This happened in Play Json build.

rockjam:
@lihaoyi ready for review

lihaoyi:
Looks good to me. Once you flesh out the PR description feel free to squash and merge

PR #306, opened 2018-04-30
Resolves #305, use an English locale when excuting java command line

### LINKED ISSUE #305: JavaCompileJarTests rely on localized message
The following line is using a localized message:
https://github.com/lihaoyi/mill/blob/d86494099d7fca87ab1a8bf0b02ae305ad210d3b/main/test/src/mill/eval/JavaCompileJarTests.scala#L134

With a French locale, the exception message is: "impossible de trouver ou charger la classe principale".
In other words, if you are running the tests suite with a non English locale, the build will fail.

I think we should add the `"-Duser.language=en"` argument to the `java` command.

Mogztter:
Otherwise we cannot rely on the exception message because the error returned by the java command line will be localized.


rockjam:
Good, will merge after CI

PR #307, opened 2018-04-30
don't fail command when there is no build.sc file in directory

rockjam:
This PR allows users to run commands from mill's MainModule without build.sc file

`mill version` gives following output:

```
➜  mill-empty mill version
No build file found, you should create build.sc to do something useful
[1/1] version
0.2.0-13-d86494-DIRTYaeada95f
```

But this can also lead to errors when we run `GenIdea/idea` with empty build file - it will surely fail.

### DIFF
--- a/main/src/mill/main/RunScript.scala
+++ b/main/src/mill/main/RunScript.scala
@@ -41,7 +41,7 @@ object RunScript{
           case Right(interp) =>
             interp.watch(path)
             val eval =
-              for(rootModule <- evaluateRootModule(wd, path, interp))
+              for(rootModule <- evaluateRootModule(wd, path, interp, log))
               yield Evaluator.State(
                 rootModule,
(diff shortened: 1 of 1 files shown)


main/src/mill/main/RunScript.scala
```diff
@@ -72,14 +72,19 @@ object RunScript{
...
+                        ): Res[mill.define.BaseModule] = {
 
     val (pkg, wrapper) = Util.pathToPackageWrapper(Seq(), path relativeTo wd)
 
     for {
       scriptTxt <-
         try Res.Success(Util.normalizeNewlines(read(path)))
-        catch { case e: NoSuchFileException => Res.Failure("Script file not found: " + path) }
+        catch { case _: NoSuchFileException =>
+          log.info("No build file found, you should create build.sc to do something useful")
```
rockjam: Better wording will be very welcome

lihaoyi:
Looks good to me

PR #308, opened 2018-05-01
Scalafmt support

rockjam:
This PR introduces the module that provides `scalafmt` support. It features stateful worker that caches file signatures internally to avoid formatting sources that were already formatted. Caching is also aware of changes in `.scalafmt.conf`, in this case we reformat everything from scratch.

It would be great if we could discuss this PR to figure out the "idiomatic" way to write such mutable tasks/commands that change sources, but don't provide any valuable result apart from side-effect


scalalib/src/mill/scalalib/scalafmt/ScalafmtModule.scala
```diff
@@ -0,0 +1,36 @@
...
+    worker().reformat(
+      filesToFormat(sources()),
+      scalafmtConfig().head,
+      scalafmtDeps().map(_.path)
+    )
+  }
+
+  def scalafmtVersion: T[String] = "1.5.1"
+
+  def scalafmtConfig: Sources = T.sources(pwd / ".scalafmt.conf")
```
rockjam: Don't we have an alternative for `Sources` for a single file?

scalalib/src/mill/scalalib/scalafmt/ScalafmtModule.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  def scalafmtDeps = resolveDeps(
```
rockjam: Should we use certain scala version to resolve deps to stay stable no matter what version of scala mill uses?

scalalib/src/mill/scalalib/scalafmt/ScalafmtModule.scala
lihaoyi: I don't think so; we could add one if it comes up repeatedly, but for now just picking the first item in the `Agg[PathRef]` may be fine

scalalib/src/mill/scalalib/scalafmt/ScalafmtModule.scala
lihaoyi: I don't have any opinion, though I suppose scalafmt is meant to do the same thing regardless of scalaVersion?

lihaoyi:
looks good to me i guess

scalalib/src/mill/scalalib/scalafmt/ScalafmtModule.scala
rockjam: Yes, version of scala doesn't affect the result of formatting with scalafmt. What I meant is the situation like this: 
At some point, scala 2.13 is released. We decide to migrate mill to 2.13, but we won't be able to do that until scalafmt for 2.13 is released.

rockjam:
Also, do we keep scalafmt support in scalalib, or should I move it to some other subproject?

scalalib/src/mill/scalalib/scalafmt/ScalafmtWorker.scala
```diff
@@ -0,0 +1,58 @@
...
+
+import ammonite.ops.{Path, exists}
+import mill._
+import mill.define.{Discover, ExternalModule, Worker}
+import mill.modules.Jvm
+import mill.util.Ctx
+
+import scala.collection.mutable
+
+object ScalafmtWorkerModule extends ExternalModule {
```
rockjam: It wasn't immediately clear that worker must live in a separate module to be a single one in a build. 
Also, I copy-pasted `extends ExternalModule` from scala and scalajs workers. Why should it be an `ExternalModule`, not just a `Module` ?
When I tried `extends Module` complilation failed with:

```
[error] /Users/rockjam/projects/mill/scalalib/src/mill/scalalib/scalafmt/ScalafmtWorker.scala:11:54: Modules, Targets and Commands can only be defined within a mill Module
[error] private[scalafmt] object ScalafmtWorker extends mill.Module {
```

lihaoyi: @rockjam being an `ExternalModule` allows a task `myTask` to be run via `mill.scalalib.scalafmt.ScalafmtWorkerModule/myTask`. Normal `Module` objects must be defined in the `build.sc` file and be run via their object selector from the root of `build.sc`. 

If you want to let people run ScalaFmt without needing to extend anything, you could make a task in the `ExternalModule` that lets people run e.g. `mill.scalalib.scalafmt.ScalafmtWorkerModule/fmt __.sources` to format the `sources` of all their `ScalaModule`s. See how the `mill.scalalib.PublishModule/publishAll` command works as an example.

As for being separate Mill submodules, we only needed to put things like `ScalaWorker`/`ScalaJSWorker` in there because we needed code that runs with some heavyweight dependencies on the classpath that might collide (e.g. different Scala.js versions): the dependencies are isolated within a classloader and interacted with via reflection. If you don't have any heavyweight dependencies, or you do but you can just interact with them directly via reflection/subprocesses, then there's no need to isolate them in a separate Mill build module

scalalib/test/src/mill/scalalib/scalafmt/ScalafmtTests.scala
```diff
@@ -0,0 +1,105 @@
...
+          val Right(_) = eval.apply(reformatCommand)
+
+          val firstReformat = getProjectFiles(ScalafmtTestModule.core, eval)
+
+          assert(
+            firstReformat("Main.scala").modifyTime > before("Main.scala").modifyTime,
+            firstReformat("Main.scala").content != before("Main.scala").content,
+            firstReformat("Person.scala").modifyTime > before("Person.scala").modifyTime,
+            firstReformat("Person.scala").content != before("Person.scala").content,
+            // resources files aren't modified
```
rockjam: I wanted to separate different parts of the test with nested test cases(like 'firstTime, 'cached, 'afterChange), but it didn't work, because I needed shared test evaluator, and if I did nested tests - test evaluator was created every time for each test. I couldn't create `TestEvaluator` outside of the `Tests` block either because it requires `implicit TestPath`, that is not available outside of the `Tests` block.
So, is it possible to do this with `utest`?

rockjam:
@lihaoyi added some tests, made external module for reformatting all the sources and added some lines into documentation. Ready for review

scalalib/test/src/mill/scalalib/scalafmt/ScalafmtTests.scala
lihaoyi: It's not easy to do with uTest, leaving them in the same test seems ok for now

PR #309, opened 2018-05-02
fix #233 add append and exclude rules to assembly

### LINKED ISSUE #233: assembly should allow configuring the merge strategy
The `assembly` task doesn't allow to define a merge strategy for duplicate files across jars. This is needed - for example - for typesafe config `reference.conf` files, which usually need to be concatenated.

This issue doesn't allow me to use Mill on a project which uses akka, because the default akka `reference.conf` is not present in the final jar.

It would be cool for Mill to have a way of defining a merge strategy similarly to how sbt-assembly does it (choose between discard/concat/first/...).

This would also provide an API to do things like #232 in a more principled way.

rockjam:
This PR introduces ability to define rules for duplicated files in assembly. Currently, there are two rules: `Append` and `Exclude`

TODO:
- [x] Tests on append/exclude rules
- [ ] Update docs about duplication rules


rockjam:
Special thanks to @alexarchambault . Code mostly based on [assembly code from coursier](https://github.com/coursier/coursier/blob/5c6d719598ae6300ec90447997e264977c5990d1/cli/src/main/scala-2.12/coursier/cli/util/Assembly.scala#L38)

rockjam:
@mlerner Want to join the review?

main/src/mill/modules/Jvm.scala
```diff
@@ -15,7 +17,35 @@ import mill.util.{Ctx, IO}
 import mill.util.Loose.Agg
 
 import scala.collection.mutable
+import scala.collection.JavaConverters._
 
+object Assembly {
```
rockjam: Is it right place to have this rules?
For people who want to define their own rules it will be:

```
import mill.modules.Assembly._
val myRules = Seq(Rule.Append("foo.conf"), Rule.Exclude("bar.txt"))
```

rockjam:
I found on use-case that won't work in the current implementation. Since we make assembly in two steps: `upstreamAssembly` and then `assembly` itself, in `assembly` step we don't keep track on files that were already written in `upstreamAssembly`. So for instance, if `reference.conf` is present in module we are making `assembly` for, it will rewrite `reference.conf` written(and concatenated) in `upstreamAssembly`. I will later write test for this case and try to fix it

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -108,6 +108,8 @@ trait JavaModule extends mill.Module with TaskModule { outer =>
     }
   }
 
+  def assemblyRules: T[Seq[Assembly.Rule]] = T { Assembly.defaultRules }
```
rockjam: Do we need to keep it as Target, or we can go with plain def here? Cause if we can go with def, we can remove serialization code for Assembly.Rule

rockjam:
PR is ready for review. I will add section to docs if the API for assembly rules is fine.

scalalib/src/mill/scalalib/JavaModule.scala
lihaoyi: I don't have any real preference. If we go with def, we could also include rules that are not serializable, e.g. containing lambdas. I leave that up to you

scalalib/test/src/mill/scalalib/HelloWorldTests.scala
```diff
@@ -45,11 +48,79 @@ object HelloWorldTests extends TestSuite {
   }
 
   object HelloWorldWithMain extends HelloBase {
-    object core extends HelloWorldModule{
-      def mainClass = Some("Main")
+    object core extends HelloWorldModuleWithMain
+  }
+
+  val akkaHttpDeps = Agg(ivy"com.typesafe.akka::akka-http:10.0.13")
```
lihaoyi: We really need to break up the tests in this file :/ Maybe not in this PR, but if you could do that in a subsequent one that would be great

main/src/mill/modules/Jvm.scala
```diff
@@ -257,20 +261,52 @@ object Jvm {
...
+    for(entry <- classpathIterator(inputPaths)) {
+      val mapping = entry.mapping
+
+      def append() = concatEntries += mapping -> (entry :: concatEntries.getOrElse(mapping, Nil))
+
+      rulesMap.get(mapping) match {
+        case Some(_:Assembly.Rule.Append) =>
+          append()
+        case Some(_:Assembly.Rule.Exclude) =>
+          ctx.log.info(s"Excluding entry: ${mapping}")
```
lihaoyi: I think we can skip the logging here; I find the logging from SBT's assembly task more annoying than useful

main/src/mill/modules/Jvm.scala
```diff
@@ -257,20 +261,52 @@ object Jvm {
     manifest.write(manifestOut)
     manifestOut.close()
 
-    def isSignatureFile(mapping: String): Boolean =
-      Set("sf", "rsa", "dsa").exists(ext => mapping.toLowerCase.endsWith(s".$ext"))
```
lihaoyi: Should we maintain this behavior in the new default `assemblyRules`? IIRC we added this because these files were causing problems for someone (there's an issue)

main/src/mill/modules/Jvm.scala
```diff
@@ -295,25 +331,45 @@ object Jvm {
     PathRef(output)
   }
 
+  sealed trait AssemblyEntry {
+    def mapping: String
+    def is: InputStream
```
lihaoyi: Let's call this `inputStream` rather than `.is`, for better greppability =)

main/src/mill/modules/Jvm.scala
```diff
@@ -257,20 +261,52 @@ object Jvm {
...
+                writeEntry(path, entry.is, append = false)
+              }
+              seen += mapping
+            }
+          }
       }
-      file.close()
     }
+
+    concatEntries.foreach { case (mapping, entries) =>
```
lihaoyi: Would it be possible to unify the `Some(_: Assembly.Rule.Append)` logic with the `None` logic, rather than having them go down a totally different code path? Apart from complexity, I think this would also re-arrange the order of files in the jar, which while harmless is unnecessary and makes debugging harder (e.g. when you `jar tf` two jars to compare contents)

Perhaps we could make the `None` case go through `concatEntries` as well, just over-writing a key rather than appending to it. If we then make `concatEntries` a `LinkedHashMap`, that should let us preserve the original ordering of keys

lihaoyi:
@rockjam left a review

scalalib/test/src/mill/scalalib/HelloWorldTests.scala
rockjam: Agreed

main/src/mill/modules/Jvm.scala
rockjam: Yep, I added it in default rules actually, right here: https://github.com/lihaoyi/mill/pull/309/files#diff-8987430aa3e4d7f9373fa4b3c3d5f2d0R13

main/src/mill/modules/Jvm.scala
rockjam: The thing is - we process `concatEntries` outside the `for` comprehension that goes through all the classpath entries. If we process both "write once" and append outside `for` that means  we have to keep all the files in memory, before we write them. I guess we want to avoid this, right?

lihaoyi: @rockjam aren't we're only keeping the list of files in `concatEntries`, and not the file contents themselves, since each `AssemblyEntry` is just a thin wrapper that allows you to open an `InputStream` on-demand? As long as we're not keeping their byte contents, and not holding open all their file-handles at the same time, I think keeping the wrappers in memory should be acceptable. e.g. Ammonite only has `16710` file within it's assembly, and even an order of magnitude more 

Or we could just leave this as is; I guess a bit of reshuffling isn't the end of the world either.

scalalib/test/src/mill/scalalib/HelloWorldTests.scala
```diff
@@ -45,11 +48,79 @@ object HelloWorldTests extends TestSuite {
...
+      def assemblyRules = T { Seq(Assembly.Rule.Append("reference.conf")) }
+    }
+    object model extends HelloWorldModule
+  }
+
+  object HelloWorldMultiExclude extends HelloBase {
+    object core extends HelloWorldModuleWithMain {
+      def moduleDeps = Seq(model)
+
+      def assemblyRules = T { Seq(Assembly.Rule.Exclude("reference.conf")) }
```
lihaoyi: Could we have one unit test for the behavior of anchored regexes? e.g. making sure`^reference.conf` matches only `reference.conf` and not `foo/reference.conf`, and that `reference.conf$` matches only `foo/reference.conf` but not `foo/reference.confetti`

lihaoyi:
Whatever we end up doing, it's probably worth running some basic benchmarks using `System.currentTimeMillis` to e.g. assemble the Mill jar and make sure we didn't regress the speed

scalalib/src/mill/scalalib/JavaModule.scala
rockjam: If we make it def, instead of `T`, and the users change them while working on their build, will it reevaluate for sure? I tend to think that it will, because we invalidate caches on build change

rockjam:
@lihaoyi Updated PR according to your review. Also moved all that append and exclude logic from Jvm to Assembly object - seems to simplify `createAssembly` code a little bit.

lihaoyi:
@rockjam looks good to me

rockjam:
@lihaoyi  By the way, made some primitive benchmarking. Did three runs for each version.

Before append rules:
1. 9119 millis
2. 6620 millis
3. 6719 millis

With append rules PR:
1. 9417 millis
2. 7322 millis
3. 7770 millist

So, there is some slowdown introduced by this PR. Do we keep it as is, or it's better to find a way to reduce this slowdown?

rockjam:
Hm, travis shows build as failed, even though it's all steps are fine. Release steps are canceled, but it always happens, if I remember correctly.  
image

lihaoyi:
@rockjam those benchmark numbers look fine; a 10%-ish slowdown is still several times faster than SBT due to our layered jar-building.

Travis is probably just being flaky; feel free to merge

PR #312, opened 2018-05-05
Fix #311: sources and resources paths for Intellij test modules

### LINKED ISSUE #311: Generated Intellij project: output path is shared between production, tests
I have created a mill build with a few modules, each of them with a nested test module. I have generated and opened an Intellij IDEA project with `mill.scalalib.GenIdea/idea`. If I try to run anything from within IDEA (rather than from mill), I get the following error for each test module :
```
Output path <project path>/out/myModule/test/compile/dest/classes is shared between: Module 'myModule.test' production, Module 'myModule.test' tests
```

This is apparently because the test modules have an output path, but don't have a test output path. Since mill test modules only contain test code, I solved this by removing the regular output path, and using the same path for the test output path instead (basically, replace `<output>` with `<output-test>`). It seems to work for me.

I'll make a PR for this (it seems like it would be a trivial change in `GenIdeaImpl`), hopefully I'm not missing something obvious here.

guilgaly:
Se issue #311; the fix is base on what I see in the .iml files after manually editing the modules in Intellij IDEA.

### DIFF
--- a/scalalib/src/mill/scalalib/GenIdeaImpl.scala
+++ b/scalalib/src/mill/scalalib/GenIdeaImpl.scala
@@ -329,7 +329,13 @@ object GenIdeaImpl {
                        ) = {
     <module type="JAVA_MODULE" version="4">
       <component name="NewModuleRootManager">
-        <output url={"file://$MODULE_DIR$/" + relify(compileOutputPath) + "/dest/classes"} />
+        {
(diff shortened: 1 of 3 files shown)


rockjam:
@guilgaly Thank you for contribution. Will take a look at PR later today

rockjam:
Looks good to me

PR #316, opened 2018-05-11
Optimize calculation of commit count

jkstrauss:
Instead of count all commits back to the beginning of time twice, and
then subtracting, just make one call to let git figure it out.

This is my first time trying to contribute, so any comments are very welcome.


lihaoyi:
Looks good to me

PR #318, opened 2018-05-11
Update 4 - Tasks.md

h2000:
T.source { -> T.sources {


h2000:
Found more places to fix :)

PR #321, opened 2018-05-14
Initial support for creating WAR files

benjaminfrank:
This PR adds the ability to create WAR files from java/scala projects. I'd like
to get some feedback on a couple of points I noticed during the implementation
and testing.

The PR currently implements 2 flavours of the `packageWar` task:

PackageWarSimple - always recreate the war file in a single task irrespective
of what really changed (class files, classpath, upstream modules, etc). This is
a very straight forward implementaton and seems to be among the fastest ones
but creates a lot of IO.

PackageWar - uses a persitent task and individual task for upstream changes to
minimize the amount of files that need to be copied to reduce IO. Most time
consuming task is (usually) the calculation of classpath entries that need to
be copied as calculating the individual md5sums is quite expensive in contrast
to simply copying (factor of >10). Using the filesize and mtime to compute
changes in the classpath looks performance wise comparable to copying. Both
options are implemented and configurable. Biggest drawback is the complexity of
the implementation.

Some performance numbers (manually taken an a recent MBP; just couple of runs):
A current project with only very few scala files, a few resources but >100
(transitive) dependencies on the classpath which translates into a ~90MB war
when changing a single line of code roughly takes:

- simple copying ~500ms
- individuals tasks + classpath changes basd on md5: ~7000ms
- individuals tasks + classpath changes basd on filesize and mtime: ~500ms

For me the main question are:
- is it worth trading some complexity and potential inaccuracy (filessize and
  mtime instead of md5) for reduced IO?
- is it worthwhile to make the computation of changes configurable or should we
  be opinionated and tell people to simply override the comparasion in case
  they need to make absolutely sure (md5/sha256/etc)?
- how to improve naming of tasks and consistent naming of them
- how to deal with name/path clashes? I could not find any info on this but
  putting a resource with the same name as a dependency into a folder called
  `lib` will result in one overriding the other.

Any feedback is welcome.

Disclosure: I (and my colleague @daniel-kullmann, who helped to develop/test)
are working for @SAP and developed this during work and are allowed to
contribute this back to mill.


benjaminfrank:
not sure why the java tests are failing due to a dependency that works in other runs...any help is welcome

rockjam:
@benjaminfrank triggered CI, let's see what's going to happen. Maybe it's a transient error.

benjaminfrank:
@lihaoyi @rockjam as this PR is open for quite some time and we had a short discussion about how to best handle plugins, I would like to re-open the idea of having some mill-contrib like repository as more and more plugins for mill are showing up. I would see two main benefits of a contrib repo for plugins: it would be a good pace to discover plugins (external repos can be linked in the readme as well) and the core can develop independent from the plugins. although the latter would ultimately require some encoding of the mill version into the plugin version at some point.

lihaoyi:
closing due to inactivity

benjaminfrank:
yes. Sorry. I totally forgot that I opened this PR.
A colleague of mine currently implements DistZip support. When he is done I will add a heavily simplified version of this PR to bootstrap the development of a sbt-native-packager equivalent.

lefou:
Bump. Any updates on this?

PR #322, opened 2018-05-14
Java container support

benjaminfrank:
Based on #321 and mainly for gathering early feedback.
It currently is lacking tests (although I am getting familiar with how to test mill, so I should be able to write some tomorrow) and the whole publishing shebang.
Main question is if this should be a separate module or be part of the core. It has no external dependencies. It also misses proper windows support and has some rough edges wrt pid handling.


lihaoyi:
closing due to inactivity

PR #323, opened 2018-05-14
Initial support for jsDependencies

benjaminfrank:
ScalaJS artifacts can bundle .js files (using `jsDependencies += ProvidedJS / "file.js"` in sbt).
Any application that has such a dependency will not be able run unless all those .js files are present when the application loads. For this the scalajs sbt plugin creates a `*-deps.js` file containing all .js files as part of the `fastOpt`/`fullOpt` phases.

While this mechanism is already superseded by scala-js-bundler some (older) artifacts still require this method of bundling .js files.

I'd like to get some early feedback on this especially as this is a separate module. As this way of bundling .js files is no longer recommended (see https://www.scala-js.org/doc/project/dependencies.html Also the sbt-plugin is on life-support only) and it pulls in quite a few external dependencies I think it should be an explicit opt-in for scalajs applications that really require this mechanism. This is also the reason why this plugin does not support the addition of js dependencies for a build but only the assembly of the `-deps.js` for the (transitive) dependencies. I am also a bit unhappy about the overloading of `fastOpt`/`fullOpt` as it somewhat hides the second generated js file.

Most of the code was simply adjusted from the sbt-plugin (https://github.com/scala-js/jsdependencies) with a lot of help from @sjrd

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -238,6 +238,11 @@ object contrib extends MillModule {
     def moduleDeps = Seq(scalalib)
   }
 
+  object scalajsdependencies extends MillModule {
+    def moduleDeps = Seq(scalajslib)
+    def ivyDeps = Agg(ivy"org.scala-js::jsdependencies-core:1.0.0-M3")
+  }
+
 }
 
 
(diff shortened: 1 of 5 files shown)


benjaminfrank:
build failure seems to be some dependency issue (wrong checksum) on windows unrelated to what this PR modifies.

benjaminfrank:
updated the code to latest master, added some documentation and tests.
there is an NPE when I trigger `fullOpt` in the test which is somehow related to resolving dependencies via coursier. it kind of looks familiar but i will need some more time to hunt it down. 
the code itself works fine for one off my projects though.

lihaoyi:
closing due to inactivity

PR #326, opened 2018-05-15
Fixed dev.{assembly, launcher} reaching max CLI arguments in Windows

robby-phd:
This PR fixes mill dev.assembly (and .launcher) in Windows where the max num of CLI arg char limit is not as generous as in macOS or Linux, thus causing `Argument list too long` error when running the generated batch file.

The fix generates `mill.vmoptions` file along the corresponding batch file; when invoked, the script looks for the VM options file in the same directory as the batch file.

Notes:

* This only works for Java 9 or above; an error is thrown otherwise. (There does not seem to be a way to pass options file to `java` before Java 9.)

* The generated `mill.vmoptions` is not returned as the Task result of dev.{assembly, launcher} because the tasks only return a single pathref. Returning the .vmoptions file requires changing JavaModule (thus, requiring bootstrapping); it does not seem necessary because they, in general, do not produce long args like dev.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -10,7 +10,7 @@ clone_folder: c:\mill
 environment:
   matrix:
   - COMPILER: default
-    JAVA_HOME: C:\Program Files\Java\jdk9
+    JAVA_HOME: C:\Program Files\Java\jdk1.8.0
   - COMPILER: cygwin
     CYGWIN_DIR: cygwin64
     JAVA_HOME: C:\Program Files\Java\jdk9
@@ -18,10 +18,7 @@ environment:
     MSYS2_ARCH: x86_64
     MSYS2_DIR: msys64
     MSYSTEM: MINGW64
-    JAVA_HOME: C:\Program Files\Java\jdk1.8.0
-
-cache:
-  - '%LOCALAPPDATA%\Coursier\cache -> build.sc'
+    JAVA_HOME: C:\Program Files\Java\jdk9
 
 install:
   - SET MILL_URL=https://github.com/lihaoyi/mill/releases/download/0.1.7/0.1.7-97-00b10e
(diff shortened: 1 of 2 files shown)


.appveyor.yml
```diff
@@ -10,18 +10,15 @@ clone_folder: c:\mill
 environment:
   matrix:
   - COMPILER: default
-    JAVA_HOME: C:\Program Files\Java\jdk9
+    JAVA_HOME: C:\Program Files\Java\jdk1.8.0
```
lihaoyi: What's this change for?

robby-phd: I changed the msys2 env to test dev.assembly, which requires changing to Java 9. I think it's good to test mill on both Java 8 and 9 in Windows so I changed the default one to Java 8.

.appveyor.yml
```diff
@@ -10,18 +10,15 @@ clone_folder: c:\mill
...
     CYGWIN_DIR: cygwin64
     JAVA_HOME: C:\Program Files\Java\jdk9
   - COMPILER: msys2
     MSYS2_ARCH: x86_64
     MSYS2_DIR: msys64
     MSYSTEM: MINGW64
-    JAVA_HOME: C:\Program Files\Java\jdk1.8.0
-
-cache:
-  - '%LOCALAPPDATA%\Coursier\cache -> build.sc'
```
robby-phd: Also, disabled caching to see if that helps with weird `MatchError`s due to `wrong checksum`, e.g., https://ci.appveyor.com/project/lihaoyi/mill/build/1.0.475/job/3n4rcqtqpeji55ur

jkstrauss:
@robby-phd: Would it make sense to package the class path inside of a single proxy jar that has the class-path set for pre-Java 9? Please see https://docs.oracle.com/javase/tutorial/deployment/jar/downman.html

robby-phd:
There does not seem to be an attribute to pass other VM options in the manifest other than classpath.
While the classpath is long, Java properties MILL_SCALA_WORKER, MILL_BUILD_LIBRARIES, etc. used in the resulting dev.assembly are longer (these are the ones stored in the generated `mill.vmoptions`). At any rate, if you find a way to make it work for pre-Java 9, that'd be better as it should work for Java 9, hopefully. (PR?)

PR #330, opened 2018-05-16
Allow hyphens in target names

### LINKED ISSUE #166: Using hyphen inside module names
At the moment, it seems to be no way to include hyphens in the module names which is a common practice to name sbt modules, for example, I create the folder `playsonify-test` and then, define it in the build definition as `object playsonify-test extends ...` or using backtics around the name with no luck.

It would be useful to be able to do this.

jkstrauss:
This resolves #166.

I decided to be conservative, and only allow hyphens.

Hyphens were already legal characters for mill identifiers, but were not matched due to their names being "javafied", e.g., `hyphenated-target` becomes `hyphenated$minustarget`. Therefore, there is not much downside just allowing them.

If any other characters are to be legalized they will need to be added to `identChars` in `mill.util.ParseArgs.parseSelector`.

### DIFF
--- a/core/src/mill/define/Module.scala
+++ b/core/src/mill/define/Module.scala
@@ -3,9 +3,10 @@ package mill.define
 import java.lang.reflect.Modifier
 
 import ammonite.ops.Path
+import mill.util.ParseArgs
 
 import scala.language.experimental.macros
-import scala.reflect.ClassTag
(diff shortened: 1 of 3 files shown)


jkstrauss:
Holding off on this for the moment. I think there may be a better way to do this.

Issue #301, opened 2018-04-17
Repl command shouldn't eval ~/.ammonite/predef.sc 


narma:
My `~/.ammonite/predef.sc` is project independent file which contains imports of libraries for common daily use and it contain conflict libraries and/or versions with project libraries/versions use.

Probably this behaviour should be configurable at least.

UPD: I should clarify that ammonite's `predef.sc` is used when I execute `mill -i module.repl`. In my scenario this is not preferred behavior. I don't want global ammonite's predef in my project's REPL.

rockjam:
Should be fine in 0.2.0. @narma may we close it?

narma:
@Baccata I should clarify that ammonite's `predef.sc` is used when I execute `mill -i module.repl`. In my scenario this is not preferred behavior. I don't want global ammonite's predef in my project's REPL.

Baccata:
I'm no maintainer, but I no the codebase a bit and maybe I could have helped. As @rockjam (who is a maintainer) says, this should be fixed in version >= 0.2.0. Which version are you running ?

rockjam:
@Baccata we were actually talking about two different things.
First I thought that the issue is that `predef` of ammonite is used when you run `mill` itself, but it turns out that it happens only in `repl` command. This may for instance lead to a situation when there are conflicting versions of libraries in the classpath - one from your project's classpath and another from ammonite's `predef`

lihaoyi:
I think I'm going to call this a wontfix; Mill's Ammonite using `~/.ammonite/predef.sc` behaves the same as using Ammonite from SBT. If you wish to override the default, you can do so by `override`ing `def repl` and passing in the `--predef foo.sc` flag to point your predef at whatever file you wish

PR #333, opened 2018-05-17
Add configurable repl options

narma:
Related to #301 

It simplify use case described in this issue #301
instead of overriding whole `repl` function
you just need override one `replOptions` on your choice.


lihaoyi:
Could we instead make the `repl` command take the `replOptions: String*` as varargs? That way people can manually pass in whatever `replOptions` they want directly from the command line, and people would still be able to `override def repl(replOptions: String*) = super.repl(Seq("foo", "bar", "baz"))` if they wanted to override it in the build file

narma:
@lihaoyi 👍 just updated PR according your suggestion

Issue #314, opened 2018-05-08
Port sbt-updates to mill


guilgaly:
The [sbt-updates plugin](https://github.com/rtimush/sbt-updates) allows to check if there are newer versions available for a project's dependencies (from the project's configured Maven and Ivy repos). It should be pretty straightforward to reuse a lot of code from the SBT plugin (all the logic for understanding and comparing version numbers isn't SBT-specific and is licensed under BSD 3-clause).

I can try to do it when I have some time, if it's OK to submit this for the main mill repo (otherwise we can close this issue and I'll try to do something similar to [mill-dgraph](https://github.com/ajrnz/mill-dgraph)).

rockjam:
Sound good to me to have it in mill main repo. Maybe coursier have a functionality to check for newer version, I don't know for sure

guilgaly:
I'll definitely check that out, but I don't expect so, since "newer" is a bit subjective (as Maven/Ivy don't really enforce a versioning scheme)... In practice, sbt-updates just tries to parse the version string against a bunch of known versioning schemes (like semver) and ignores anything it doesn't understand.

But even if we could just use coursier to resolve available versions, that would already be nice (sbt-updates does its own thing to download artefact metadata and parse the list of available versions). Anyway, I'll see what I can find.

alexarchambault:
Yeah, there's some basic stuff for that in coursier, around [here](https://github.com/coursier/coursier/blob/8427be2d453581b86e6b3aae8c581e6521e90e12/core/shared/src/main/scala/coursier/maven/MavenRepository.scala#L160) for Maven repositories, and [here](https://github.com/coursier/coursier/blob/8427be2d453581b86e6b3aae8c581e6521e90e12/core/shared/src/main/scala/coursier/ivy/IvyRepository.scala#L171) for Ivy repositories (although just Maven should cover most cases for mill).

The Maven method could be made public, and its signature possibly added to `coursier.core.Repository` (and be later implemented for other repos…). It's tested indirectly in coursier by resolving versions like `x.y.+` around [here](https://github.com/coursier/coursier/blob/8427be2d453581b86e6b3aae8c581e6521e90e12/tests/shared/src/test/scala/coursier/test/CentralTests.scala#L383).

Send a PR to do that in coursier if you'd like, and I'll do a release right after it…

guilgaly:
@alexarchambault Thanks, I'll have a look!

lihaoyi:
[...]

PR #340, opened 2018-05-19
Issue #314; port sbt-updates to mill

guilgaly:
See issue #314.

Usage: `mill mill.scalalib.Dependency/updates`.

Known limitations (I'd rather resolve them in one or more follow-up PRs):

- Maven-only (missing implementation to check for updates from Ivy repos)
- Won't find versions from a Maven repo missing the proper `maven-metadata.xml` file (I think it can happen, and Coursier appears to have a fallback solution that could be made public, as suggested by @alexarchambault in the issue discussion)
- Missing ability to restrict check to specific modules (currently checks for updates in all `JavaModule`s in the build)
- Doesn't check for updates to the `$ivy` dependencies in the build definition

TODO (in this PR):

- [x] Properly format output
- [x] Allow listing pre-release versions
- [x] Add documentation
- [x] Add tests
- [x] Rewrite versions parser with fastparse

Also worth noting: I borrowed some code from sbt-updates, under the BSD 3-clause license. This project is under the MIT license. Since the BSD license requires retaining the original license text, I copied it directly at the top of the relevant files, I think that should be fine?


build.sc
```diff
@@ -131,7 +131,8 @@ object scalalib extends MillModule {
   def moduleDeps = Seq(main)
 
   def ivyDeps = Agg(
-    ivy"org.scala-sbt:test-interface:1.0"
+    ivy"org.scala-sbt:test-interface:1.0",
+    ivy"org.scala-lang.modules::scala-parser-combinators:1.1.0"
```
guilgaly: Parser combinators are used for parsing the dependencies version numbers

scalalib/src/mill/scalalib/dependency/metadata/MetadataLoader.scala
```diff
@@ -0,0 +1,7 @@
+package mill.scalalib.dependency.metadata
```
guilgaly: `MetadataLoader.scala`, `MavenMetadataLoader.scala` and `MetadataLoaderFactory.scala` exist in sbt-updates, be I rewrote evrything in thoses classes, so I didn't copy the BSD license text in those files

build.sc
hepin1989: not fastparse?

scalalib/src/mill/scalalib/dependency/versions/Version.scala
```diff
@@ -0,0 +1,249 @@
+/*
```
lihaoyi: Let's get rid of the license headers since we haven't been using them throughout most of the codebase; any code contributed ends up licensed under the top-level license in the repo

build.sc
lihaoyi: Yeah if you're going to use a combinator library just use fastparse, after all we already have it as a transitive dependency through Ammonite

scalalib/src/mill/scalalib/dependency/versions/Version.scala
lihaoyi: actually, if the bsd license requires copying the license blurb, then that's probably fine.

build.sc
guilgaly: I took the parsing code straight from from sbt-updates, which uses scala-parser-combinators ; but if we already have fastparse as a dependency, I can certainly try to rewrite that code with it. 👍

scalalib/src/mill/scalalib/dependency/versions/Version.scala
guilgaly: Yes, it's a requirement of all variants of the BSD license (clause number 2 for the 3-clause variant here).

build.sc
guilgaly: @lihaoyi turned out to be pretty easy to rewrite with fastparse (as an added bonus, I think it's now a bit more readable than the original code which used regex-based parsers). 👍

lihaoyi:
@guilgaly let me know if/when this is ready for review and we can take another look

guilgaly:
@lihaoyi I think this can be reviewed as-is, if you agree that the limitations I listed in the description are acceptable for now. 👍

antonkulaga:
This feature will be really useful, looking forward to see it merged!

lihaoyi:
looks good to me

PR #342, opened 2018-05-20
Preserve caches between interactive and client/server mode

lihaoyi:
We were incorrectly duplicating the JDK classpath as part of the application classpath when we spawned the Mill server from the Mill client. This makes the transmission of application classpath to the Mill server explicit via an environment variable, so we don't end up including random things from the client classloader hierarchy that we didn't expect


main/client/src/mill/main/client/MillClientMain.java
```diff
@@ -11,20 +11,8 @@
...
-        while(current != null){
-            if (current instanceof java.net.URLClassLoader) {
-                URL[] urls = ((java.net.URLClassLoader) current).getURLs();
-                for (URL url: urls) {
-                    selfJars.add(new File(url.toURI()).getCanonicalPath());
-                }
-            }
-            current = current.getParent();
-        }
-        if (Util.isJava9OrAbove) {
```
lihaoyi: @robby-phd do you remember what this was for, or why we need it?

robby-phd: It was always there [before](https://github.com/lihaoyi/mill/blob/122122729d8567e389f434cfb1c074d3119b0088/clientserver/src/mill/clientserver/Client.java#L14-L25) I worked on changing mill client; I assume it's to propagate classpath to the spawned server.

lihaoyi:
Ok, maybe in that case I can get rid of it then

main/client/src/mill/main/client/MillClientMain.java
lihaoyi: @robby-phd how about the `System.getProperty("java.class.path")` logic? It looks like you added it in https://github.com/lihaoyi/mill/commit/a9d4eea6f5b81cdbef84724e612bf07954e2673c#diff-00b5b77d5f332c03ac7aa4d9ac51c941R29

PR #346, opened 2018-05-22
Solves 345 : optional signing

Baccata:
* Made gpgPassphrase optional for publishing
* Added a flag to remove signing of published artifacts altogether


lihaoyi:
@rockjam want to review this? Since this is mostly your code

rockjam:
Sure, will do

scalalib/src/mill/scalalib/PublishModule.scala
```diff
@@ -73,32 +73,35 @@ trait PublishModule extends JavaModule { outer =>
   }
 
   def publish(sonatypeCreds: String,
-              gpgPassphrase: String,
+              gpgPassphrase: String = "",
```
rockjam: Is it possible to make this argument optional? Same applies to `publishAll` command

scalalib/src/mill/scalalib/PublishModule.scala
```diff
@@ -73,32 +73,35 @@ trait PublishModule extends JavaModule { outer =>
...
+              signed: Boolean = true,
               release: Boolean): define.Command[Unit] = T.command {
     val PublishModule.PublishData(artifactInfo, artifacts) = publishArtifacts()
     new SonatypePublisher(
       sonatypeUri,
       sonatypeSnapshotUri,
       sonatypeCreds,
-      gpgPassphrase,
+      Some(gpgPassphrase).filter(_.nonEmpty),
```
rockjam: If having `gpgPassphrase` parameter optional is not possible, I guess we could have default value as `null` and wrap it in `Option(gpgPassphrase)` instead of filtering out an empty string, what do you think?

scalalib/src/mill/scalalib/publish/SonatypePublisher.scala
```diff
@@ -135,10 +136,15 @@ class SonatypePublisher(uri: String,
...
-  private def poorMansSign(file: Path, passphrase: String): Path = {
+  private def poorMansSign(file: Path, maybePassphrase: Option[String]): Path = {
     val fileName = file.toString
     import ammonite.ops.ImplicitWd._
-    %("gpg", "--passphrase", passphrase, "--batch", "--yes", "-a", "-b", fileName)
+    maybePassphrase match {
+      case Some(passphrase) =>
+        %("gpg", "--passphrase", passphrase, "--batch", "--yes", "-a", "-b", fileName)
+      case None =>
+        %("gpg", "--batch", "--yes", "-a", "-b", fileName)
```
rockjam: Is it useful to sign artifacts without `gpgPassphrase` ?

rockjam:
@Baccata left some comments

scalalib/src/mill/scalalib/PublishModule.scala
Baccata: I don't think it's possible to write a ScoptReader for "absence of values" tbh. However your suggestion to default to null is better than empty string. Addressed : ec362d8

scalalib/src/mill/scalalib/PublishModule.scala
Baccata: Addressed ec362d8

scalalib/src/mill/scalalib/publish/SonatypePublisher.scala
Baccata: I honestly don't know if it's useful, but I'm following Haoyi's comment  : 

https://github.com/lihaoyi/mill/issues/345#issuecomment-390849979

rockjam: Oh, I didn't notice that. All good then

rockjam:
@Baccata everything seems to be fine. Will merge once CI goes green

PR #348, opened 2018-05-23
Adding note on cleaning .idea files before regenerating

bbarker:


lihaoyi:
is this really necessary? I've never had to do this before, though since IDEA 2018 I've started needing to close &re-open the project

guilgaly:
From what I've seen, when you run `mill.scalalib.GenIdea/idea`:
- the tasks reads the Java SDK configured in the existing Intellij project (to keep the same one in the newly generated project files)
- everything else is replaced (or should be, otherwise it's probably a bug in `GenIdeaImpl`?)
- Intellij asks to close & reopen the project since the project files have changed behind its back

So I don't think there should be any need to manually delete the project files; and if you do, you'll loose the benefit of keeping the already configured Java SDK (if yours isn't a JDK 8 named `1.8 (1)`, since that's the default value).

bbarker:
Maybe not, but it seemed to help me. I tried regenerating and closing and
opening idea about 4 times before this helped. No dependencies were working
in the idea project though, so maybe it had gotten into a bad state.

bbarker:
I'll close for now, until I understand the issue better.

PR #349, opened 2018-05-24
Add a `visualize` command to render portions of the build graph as SVG

lihaoyi:
```
out/dev/launcher/dest/run -i visualize __.compile __.test
out/dev/launcher/dest/run -i visualize core.__
```

screenshot 2018-05-24 08 27 05

We perform a transitive reduction on the dependency graph before rendering, which preserves the overall structure of the graph but removes a lot of noise making it easier to see what depends on what.

This bundles the graphviz & jgrapht dependencies in a separate module in order to avoid bloating the dependencies of the core Mill executable, and prevent appveyor CI problems caused by too-long-classpaths being passed at the command line.

To support that module, I moved the mill module ivy resolution from `scalalib` to `main`, since everyone (not just java/scala projects) can now make use of it via `visualize`. 

`TestRunner` is moved from `scalalib.worker` to `scalalib` since it doesn't need any of the Zinc dependencies.


robby-phd:
Looks like the dev.assembly classpath with the changes in the PR now exceeds the Windows char limit for CLI args (I assume the PR adds more libs). I tried to put the classpath in the .vmoptions file before but Java doesn't want to accept `-cp` somehow. I didn't try `-classpath` or `--class-path` though; the relevant code is at https://github.com/lihaoyi/mill/blob/visualize/build.sc#L291 if you want to give it a try (or perhaps disabling the breaking build task in AppVeyor for now).

PR #351, opened 2018-05-27
Improve Intellij Idea support

Baccata:
Improves the Intellij Idea support in various ways :

* Cherrypicks the idea conf that needs deleting rather than deleting the whole .idea directory. That directory contains elements of configuration like VCS reference that were annoying to set again every time mill regenerated idea config.
* Attempts to retrieve libraries that the build depends on by inspecting the classloader of the top module
* Attempts at grouping jars and sources together in order to have both in the same idea files, which appears to give better jump to definition
* Hacks the library names for the libraries the build depends on, in order to match Intellij's ammonite support and not show red to the user about the library that has successfuly been resolved. Also allows to jump to the library sources from the magic import.

### DIFF
--- a/scalalib/src/mill/scalalib/GenIdeaImpl.scala
+++ b/scalalib/src/mill/scalalib/GenIdeaImpl.scala
@@ -1,13 +1,14 @@
 package mill.scalalib
 
 import ammonite.ops._
-import coursier.Repository
+import ammonite.runtime.SpecialClassLoader
+import coursier.{Cache, CoursierPaths, Repository}
 import mill.define._
 import mill.eval.{Evaluator, PathRef, Result}
-import mill.{T, scalalib}
 import mill.util.Ctx.{Home, Log}
-import mill.util.{Loose, PrintLogger, Strict}
 import mill.util.Strict.Agg
+import mill.util.{Loose, Strict}
+import mill.{T, scalalib}
 
 import scala.util.Try
 
@@ -34,7 +35,8 @@ object GenIdeaImpl {
 
     val jdkInfo = extractCurrentJdk(pwd / ".idea" / "misc.xml").getOrElse(("JDK_1_8", "1.8 (1)"))
 
-    rm! pwd/".idea"
+    rm! pwd/".idea"/"libraries"
+    rm! pwd/".idea"/"scala_compiler.xml"
     rm! pwd/".idea_modules"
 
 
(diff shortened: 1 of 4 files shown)


scalalib/src/mill/scalalib/GenIdeaImpl.scala
```diff
@@ -34,7 +35,8 @@ object GenIdeaImpl {
 
     val jdkInfo = extractCurrentJdk(pwd / ".idea" / "misc.xml").getOrElse(("JDK_1_8", "1.8 (1)"))
 
-    rm! pwd/".idea"
```
Baccata: this change makes the user's life more comfortable by not asking them to reset various pieces of mill-unrelated configuration

scalalib/src/mill/scalalib/GenIdeaImpl.scala
```diff
@@ -81,6 +83,19 @@ object GenIdeaImpl {
           res.items.toList.map(_.path)
       }
 
+    val buildDepsPaths = Try(evaluator
```
Baccata: attempts to retrieve magic imports

scalalib/src/mill/scalalib/GenIdeaImpl.scala
```diff
@@ -178,16 +241,15 @@ object GenIdeaImpl {
...
 
-    val libraries = allResolved.map{path =>
+    val libraries = resolvedLibraries(allResolved).map{ resolved =>
+      import resolved.path
       val url = "jar://" + path + "!/"
-      val name = pathToLibName(path)
-      Tuple2(".idea"/'libraries/s"$name.xml", libraryXmlTemplate(name, url))
-    }
-
-    val buildLibraries = buildLibraryPaths.map{path =>
```
Baccata: `buildLibraryPaths` is contained by `allResolved`

scalalib/test/src/mill/scalalib/GenIdeaTests.scala
```diff
@@ -43,8 +43,6 @@ object GenIdeaTests extends TestSuite {
           millSourcePath / "generated" / ".idea_modules" /"mill-build.iml",
         "gen-idea/idea/libraries/scala-library-2.12.4.jar.xml" ->
           millSourcePath / "generated" / ".idea" / "libraries" / "scala-library-2.12.4.jar.xml",
-        "gen-idea/idea/libraries/scala-library-2.12.4-sources.jar.xml" ->
```
Baccata: sources are grouped in the same file as the jar

lihaoyi:
@rockjam feel free to review this one, i might not get to it soon

scalalib/src/mill/scalalib/GenIdeaImpl.scala
```diff
@@ -148,13 +161,64 @@ object GenIdeaImpl {
...
+        val pom = path / up / s"$withoutExt.pom"
+        val sources = Some(path / up / s"$withoutExt-sources.jar")
+          .filter(_.toIO.exists())
+        Some(CoursierResolved(path, pom, sources))
+      } else Some(OtherResolved(path))
+    }
+
+    // Hack so that Intellij does not complain about unresolved magic
+    // imports in build.sc when in fact they are resolved
+    def sbtLibraryNameFromPom(pom : Path) : String = {
```
Baccata: I should maybe remove this, Intellij seems to be flaky as hell with this. It worked on a java library, doesn't seem to work on scala ones ...

Baccata: Changed to ed958b7. This allows users to avoid squiggly lines by defining their artifacts with ":" rather than "::" (which implies adding `_2.12` to the artifact name)

hepin1989:
Thank you very much. @Baccata

PR #353, opened 2018-05-28
Check for availability of Java compiler

daniel-kullmann:
Throws an exception with a sensible error message when the Java compiler
is not available.

I spent some time trying to figure out where that NullPointerException was coming from, so I added a check for that case.


lihaoyi:
looks good to me

PR #354, opened 2018-05-29
Fix MILL_CLASSPATH for Windows

jkstrauss:
Use MILL_CLASSPATH system property for Windows. It looks like it was an oversight when it was originally introduced. Also, since this make the command line arguments exceed the limit for Windows, use a vm options file for starting the server.


lihaoyi:
Looks good i guess

PR #359, opened 2018-05-31
run scalafmt on mill's sources

rockjam:
I ran `mill mill.scalalib.scalafmt.ScalafmtModule/reformatAll __.sources`(pretty long command, btw) on mill's sources with standard scalafmt config. I know that it's pretty tedious to review this PR, but it's one-time only, and I think it worth to have a unified formatting for project to simplify contribution for newcomers and being more confident about how code should look like. 

I started with scalafmt's defaults, but I'd happily adjust configuration if there are any suggestions.

### DIFF
--- a/core/src/mill/define/Applicative.scala
+++ b/core/src/mill/define/Applicative.scala
@@ -14,14 +14,14 @@ import scala.reflect.macros.blackbox.Context
   * Applier.zipMap(applyable1, applyable2){ (a1, a2, ctx) => ... a1 ... a2 ... }
   */
 object Applicative {
-  trait ApplyHandler[M[+_]]{
+  trait ApplyHandler[M[+ _]] {
     def apply[T](t: M[T]): T
   }
-  object ApplyHandler{
+  object ApplyHandler {
     @compileTimeOnly("Target#apply() can only be used with a T{...} block")
-    implicit def defaultApplyHandler[M[+_]]: ApplyHandler[M] = ???
+    implicit def defaultApplyHandler[M[+ _]]: ApplyHandler[M] = ???
   }
-  trait Applyable[M[+_], +T]{
+  trait Applyable[M[+ _], +T] {
     def self: M[T]
     def apply()(implicit handler: ApplyHandler[M]): T = handler(self)
   }
@@ -32,20 +32,24 @@ object Applicative {
     def ctx()(implicit c: Ctx) = c
     def underlying[A](v: W[A]): T[_]
 
-    def zipMap[R]()(cb: Ctx => Z[R]) = mapCtx(zip()){ (_, ctx) => cb(ctx)}
+    def zipMap[R]()(cb: Ctx => Z[R]) = mapCtx(zip()) { (_, ctx) =>
+      cb(ctx)
+    }
     def zipMap[A, R](a: T[A])(f: (A, Ctx) => Z[R]) = mapCtx(a)(f)
     def zip(): T[Unit]
     def zip[A](a: T[A]): T[Tuple1[A]]
   }
 
-  def impl[M[_], T: c.WeakTypeTag, Ctx: c.WeakTypeTag](c: Context)
-                                                      (t: c.Expr[T]): c.Expr[M[T]] = {
-    impl0(c)(t.tree)(implicitly[c.WeakTypeTag[T]], implicitly[c.WeakTypeTag[Ctx]])
+  def impl[M[_], T: c.WeakTypeTag, Ctx: c.WeakTypeTag](c: Context)(
+      t: c.Expr[T]): c.Expr[M[T]] = {
+    impl0(c)(t.tree)(implicitly[c.WeakTypeTag[T]],
+                     implicitly[c.WeakTypeTag[Ctx]])
   }
-  def impl0[M[_], T: c.WeakTypeTag, Ctx: c.WeakTypeTag](c: Context)
-                                                       (t: c.Tree): c.Expr[M[T]] = {
+  def impl0[M[_], T: c.WeakTypeTag, Ctx: c.WeakTypeTag](c: Context)(
+      t: c.Tree): c.Expr[M[T]] = {
     import c.universe._
-    def rec(t: Tree): Iterator[c.Tree] = Iterator(t) ++ t.children.flatMap(rec(_))
+    def rec(t: Tree): Iterator[c.Tree] =
+      Iterator(t) ++ t.children.flatMap(rec(_))
 
     val bound = collection.mutable.Buffer.empty[(c.Tree, ValDef)]
     val targetApplySym = typeOf[Applyable[Nothing, _]].member(TermName("apply"))
@@ -60,11 +64,10 @@ object Applicative {
 
     val transformed = c.internal.typingTransform(t) {
       case (t @ q"$fun.apply()($handler)", api) if t.symbol == targetApplySym =>
-
         val localDefs = rec(fun).filter(_.isDef).map(_.symbol).toSet
         val banned = rec(t).filter(x => defs(x.symbol) && !localDefs(x.symbol))
 
-        if (banned.hasNext){
+        if (banned.hasNext) {
           val banned0 = banned.next()
           c.abort(
             banned0.pos,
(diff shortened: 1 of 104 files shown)


core/src/mill/define/Module.scala
```diff
@@ -82,12 +85,13 @@ object Module{
     // script/REPL runner always wraps user code in a wrapper object/trait
     def reflectNestedObjects[T: ClassTag] = {
       (reflect[T] ++
-        outer
-          .getClass
-          .getClasses
+        outer.getClass.getClasses
           .filter(implicitly[ClassTag[T]].runtimeClass isAssignableFrom _)
-          .flatMap(c => c.getFields.find(_.getName == "MODULE$").map(_.get(c).asInstanceOf[T]))
```
lihaoyi: Perhaps we could allow this to remain on one line by configuring the max line length to 100?

core/src/mill/define/BaseModule.scala
```diff
@@ -2,52 +2,53 @@ package mill.define
...
-      Ctx.External(external0),
-      Ctx.Foreign(foreign0),
-      millFile0
-    )
-  ){
+                          foreign0: Boolean = false)(
+    implicit millModuleEnclosing0: sourcecode.Enclosing,
+    millModuleLine0: sourcecode.Line,
+    millName0: sourcecode.Name,
+    millFile0: sourcecode.File)
```
lihaoyi: I really do not like this formatting =( it seems worse in every way compared to the original:

- Mixed indentation
- Mixed open/closing-paren alignment style

There are lots of other cases in this diff which are similarly changed. Could we configure scalafmt to maintain the old style?

core/src/mill/define/Applicative.scala
```diff
@@ -14,14 +14,14 @@ import scala.reflect.macros.blackbox.Context
...
     def apply[T](t: M[T]): T
   }
-  object ApplyHandler{
+  object ApplyHandler {
     @compileTimeOnly("Target#apply() can only be used with a T{...} block")
-    implicit def defaultApplyHandler[M[+_]]: ApplyHandler[M] = ???
+    implicit def defaultApplyHandler[M[+ _]]: ApplyHandler[M] = ???
   }
-  trait Applyable[M[+_], +T]{
+  trait Applyable[M[+ _], +T] {
```
lihaoyi: Any idea why scalafmt wants a space between the `+` and the `_`, but not between `+` and `T`?

core/src/mill/define/Applicative.scala
```diff
@@ -32,20 +32,24 @@ object Applicative {
     def ctx()(implicit c: Ctx) = c
     def underlying[A](v: W[A]): T[_]
 
-    def zipMap[R]()(cb: Ctx => Z[R]) = mapCtx(zip()){ (_, ctx) => cb(ctx)}
+    def zipMap[R]()(cb: Ctx => Z[R]) = mapCtx(zip()) { (_, ctx) =>
+      cb(ctx)
+    }
```
lihaoyi: Honestly I don't think this is worth new-lining; the original is already pretty short & concise as-is, and the two new lines have basically nothing on them

core/src/mill/define/BaseModule.scala
```diff
@@ -2,52 +2,53 @@ package mill.define
...
 
-
-abstract class ExternalModule(implicit millModuleEnclosing0: sourcecode.Enclosing,
-                              millModuleLine0: sourcecode.Line,
-                              millName0: sourcecode.Name)
-  extends BaseModule(ammonite.ops.pwd, external0 = true, foreign0 = false){
+abstract class ExternalModule(
+    implicit millModuleEnclosing0: sourcecode.Enclosing,
+    millModuleLine0: sourcecode.Line,
+    millName0: sourcecode.Name)
```
lihaoyi: In general I would like definitions to never include a newline after the opening paren. I especially don't like the fact that after this reformatting, sometimes there's a newline and sometimes there isn't...

core/src/mill/define/Task.scala
```diff
@@ -44,28 +44,30 @@ abstract class Task[+T] extends Task.Ops[T] with Applyable[Task, T]{
...
-object Target extends TargetGenerated with Applicative.Applyer[Task, Task, Result, mill.util.Ctx] {
+object Target
+    extends TargetGenerated
+    with Applicative.Applyer[Task, Task, Result, mill.util.Ctx] {
 
-  implicit def apply[T](t: T)
-                       (implicit rw: RW[T],
-                        ctx: mill.define.Ctx): Target[T] = macro targetImpl[T]
+  implicit def apply[T](t: T)(implicit rw: RW[T],
+                              ctx: mill.define.Ctx): Target[T] =
```
lihaoyi: I think this is inferior to the original. Having everything laid out vertically in one nice column is better than having the first row be horizontally and then subsequent args laid out vertically

main/src/mill/main/RunScript.scala
```diff
@@ -22,94 +22,112 @@ import scala.reflect.ClassTag
...
+      for (s <- evalState)
+        yield
+          new Evaluator[Any](home,
+                             wd / 'out,
+                             wd / 'out,
+                             s.rootModule,
+                             log,
+                             s.classLoaderSig,
+                             s.workerCache,
+                             env)
```
lihaoyi: If we want to multi-line this, it should be
```scala
new Evaluator[Any](
  home,
  wd / 'out,
  wd / 'out,
  s.rootModule,
  log,
  s.classLoaderSig,
  s.workerCache,
  env
)
```

main/src/mill/main/RunScript.scala
```diff
@@ -22,94 +22,112 @@ import scala.reflect.ClassTag
...
             val eval =
-              for(rootModule <- evaluateRootModule(wd, path, interp, log))
-              yield Evaluator.State(
-                rootModule,
-                rootModule.getClass.getClassLoader.asInstanceOf[SpecialClassLoader].classpathSignature,
-                mutable.Map.empty[Segments, (Int, Any)],
-                interp.watchedFiles
-              )
+              for (rootModule <- evaluateRootModule(wd, path, interp, log))
+                yield
```
lihaoyi: I would prefer `for` and `yield` to be vertically aligned, rather than having the `yield` indented

core/src/mill/define/Cross.scala
```diff
@@ -42,18 +41,17 @@ object Cross{
...
-class Cross[T](cases: Any*)
-              (implicit ci: Cross.Factory[T],
-               ctx: mill.define.Ctx) extends mill.define.Module()(ctx) {
+class Cross[T](cases: Any*)(implicit ci: Cross.Factory[T], ctx: mill.define.Ctx)
+    extends mill.define.Module()(ctx) {
 
   override lazy val millModuleDirectChildren =
     this.millInternal.reflectNestedObjects[Module] ++
-    items.collect{case (k, v: mill.define.Module) => v}
+      items.collect { case (k, v: mill.define.Module) => v }
```
lihaoyi: I'd prefer expressions split over multiple lines with infix operators to be vertical aligned, rather than having the subsequent expressions be indented relative to the first

core/src/mill/define/BaseModule.scala
sake92: Dangling parens are now enabled by default (v1.6.0): https://github.com/scalameta/scalafmt/issues/1151  
Shouldn't there be a `.scalafmt.conf` in mill's root?

rockjam: There will be `.scalafmt.conf`, as a first attempt, I ran scalafmt on default configuration.

lihaoyi:
closing due to inactivity

rockjam:
Yep, hope to revisit it later

core/src/mill/define/BaseModule.scala
rockjam: @dwijnand I believe it was 1.5.1

dwijnand: Hmm but didn't https://github.com/scalameta/scalafmt/issues/1151 ship in 1.6.0-RC1?

PR #362, opened 2018-06-04
Allow hyphens in module and task names

jkstrauss:
This solves issue #166 .

I decided to be conservative, and only allow hyphens in names. Any other characters could be added very easily, but I am not sure if there is a need for them, especially since many of them are not normally used in file names.

Please give me any thoughts about how I constructed the tests.


lihaoyi:
Looks good to me

PR #364, opened 2018-06-06
Allow configuration of TestRunner's ScalaWorkerModule

aosagie:
Hi,
My environment proxies everything and has some internal repositories that we're mandated to use. We can currently compile (thanks to this issue's resolution: https://github.com/lihaoyi/mill/issues/246) but we can't run tests.


lihaoyi:
Looks good thanks!

Kidist-Abraham:
@aosagie I was also following that solution and I got some problems. How were you able to extend ScalaWorkerModule? I am getting an error saying 'not found: type ScalaWorkerModule' when I extend  ScalaWorkerModule.

PR #365, opened 2018-06-06
Fixed typos, wordings, formatted examples

sake92:
While reading the docs, I noticed some minor inconsistencies, which are fixed in this PR:
- use same import syntax in examples
- fixed code formatting (missing space after class name or `T` target def)
- using "its" for possessive, instead of "it's" 
- inheritance instead of inherit**e**nce
- inserted/changed some words to have more meaningful description


lihaoyi:
thanks!

PR #376, opened 2018-06-24
Add scala JS version suffix to ivy deps for published JS ivy.xml

julianmichael:
When compiling a JS library, the platform suffix (e.g., `_sjs0.6`) is correctly appended before the scala version to dependencies during resolution. However, when publishing, the artifact ID listed in the published `ixy.xml` is missing the suffix, causing dependency resolution to fail for downstream clients using the published JS library.

Maybe in the longer term it'd be good to just have one code path producing the artifact names so they wouldn't ever end up disagreeing like this.

But anyway, this super simple patch works for me for now.

### DIFF
--- a/scalalib/src/mill/scalalib/ScalaModule.scala
+++ b/scalalib/src/mill/scalalib/ScalaModule.scala
@@ -38,7 +38,12 @@ trait ScalaModule extends JavaModule { outer =>
   }
 
   override def resolvePublishDependency: Task[Dep => publish.Dependency] = T.task{
-    publish.Artifact.fromDep(_: Dep, scalaVersion(), Lib.scalaBinaryVersion(scalaVersion()))
+    publish.Artifact.fromDep(
+      _: Dep,
+      scalaVersion(),
+      Lib.scalaBinaryVersion(scalaVersion()),
+      platformSuffix()
+    )
   }
 
   override def finalMainClassOpt: T[Either[String, String]] = T{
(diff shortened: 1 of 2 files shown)


lihaoyi:
using `::` for dependencies e.g. `com.lihaoyi::upickle::0.6.6` should fix this. Are you sure you have the second `::` in place?

julianmichael:
Yes. The build file I'm using is [here](https://github.com/julianmichael/spacro/blob/master/build.sc). Then I'm doing `publishLocal` and trying to use it in a client project defined [here](https://github.com/julianmichael/spacro/blob/master/spacro-sample/build.sc). I'm not super familiar with how the ivy macro thing works, but it seems to me that the issue is that when resolving dependencies at compile time, `ScalaModule` uses `[resolveCoursierDependency](https://github.com/lihaoyi/mill/blob/master/scalalib/src/mill/scalalib/ScalaModule.scala#L36)`, which looks at the platform suffix for the Scala.js version, while when resolving dependencies to include in `ivy.xml`, it uses `[resolvePublishDependency](https://github.com/lihaoyi/mill/blob/master/scalalib/src/mill/scalalib/ScalaModule.scala#L40)`, which ignores the platform suffix.

So I just changed the latter to take the platform suffix as an argument & add it into the artifact ID. That fixed the issue for me.

lihaoyi:
hmm ok, will take a look, thanks!

lihaoyi:
looks ok to me

PR #384, opened 2018-07-11
Fix bug with 'u' at the beginning of path

### LINKED ISSUE #378: Mill doesn't seem to work on Windows when the user's name starts with lowercase u
There seems to be a problem caused by the fact that slash-u starts a Unicode escape sequence.
Here is a screen-grab:
mill in cygwin screenshot

jkstrauss:
This fixes  #378 

I forced this error for testing purposes by applying the following patch on Windows:

    diff --git a/core/src/mill/util/Ctx.scala b/core/src/mill/util/Ctx.scala
    index 6c8b2af..ce3fe0c 100644
    --- a/core/src/mill/util/Ctx.scala
    +++ b/core/src/mill/util/Ctx.scala
    @@ -33,7 +33,7 @@ object Ctx{
         def args: IndexedSeq[_]
       }
     
    -  def defaultHome = ammonite.ops.home / ".mill" / "ammonite"
    +  def defaultHome = ammonite.ops.home / ".mill" / "uammonite"
     
     }
     class Ctx(val args: IndexedSeq[_],

### DIFF
--- a/main/src/mill/MillMain.scala
+++ b/main/src/mill/MillMain.scala
@@ -94,14 +94,13 @@ object MillMain {
           stderr.println("Build repl needs to be run with the -i/--interactive flag")
           (false, stateCache)
         }else{
-          val tqs = "\"\"\""
           val config =
             if(!repl) cliConfig
             else cliConfig.copy(
               predefCode =
                 s"""import $$file.build, build._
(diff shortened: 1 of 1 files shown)


jkstrauss:
@lihaoyi Any reason this is sitting?

lihaoyi:
i thought the build was red, but it looks like tha thas been fixed

PR #391, opened 2018-07-26
Plugins do not belong on the compiler classpath

smarter:
They will be classloaded by the compiler itself based on -Xplugin.


xeno-by:
There may be a benefit to putting plugins on the compiler classpath: https://github.com/scala/scala-dev/issues/458. For semanticdb-scalac, we observed a significant performance improvement: https://github.com/xeno-by/scalameta/commit/9d19cc34ba8fed0b03b0137c3b99e85622adbe11#diff-7a43ed35a87fba85b3e188877f40f0b6.

smarter:
@xeno-by There's already a setting to let the compiler cache the plugin classloaders, but it's off currently: https://github.com/scala/scala/issues/6412 You should ask the scalac folks to consider turning it on by default for 2.13.

lihaoyi:
I'm reverting this change because master seems to misbehave in Java 9+ once classloaders are properly isolated, and adding the plugin classpath back onto the compiler classpath seems to fix it. No idea why

PR #394, opened 2018-07-30
Fix bullets in manual

jkstrauss:
Really a small thing, but the bullets do not show correctly in my addition to the Mill manual with regard to Module/Task Names.

@lihaoyi: Is there any way to preview the manual? Looking at the Markdown on Github shows correctly without this change. See [2 - Configuring Mill.md](https://github.com/lihaoyi/mill/blob/master/docs/pages/2%20-%20Configuring%20Mill.md).

I would like to make sure that my change actually corrects the problem. Right now, I am just imitating what I see elsewhere, i.e., empty line before the bullets, and no indentation.

### DIFF
--- a/docs/pages/2 - Configuring Mill.md
+++ b/docs/pages/2 - Configuring Mill.md
@@ -395,9 +395,10 @@ object unhyphenatedModule extends Module {
 ```
 
 Mill modules and tasks may be composed of any of the following characters types:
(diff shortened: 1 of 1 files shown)


jkstrauss:
@lihaoyi I see that the docs are fixed now, but how would I have been able to test it?

PR #404, opened 2018-08-08
Visualize Plan

jkstrauss:
The purpose of this PR is too visualize all tasks that would be resolved by "plan", and not just the tasks returned by "resolve", which is what "visualize" does.

However, it will still not work if any external modules are returned by plan, and I cannot figure out how to either filter them out, or better yet, resolve them.

Any suggestions are welcome.


lihaoyi:
looks good to me

PR #409, opened 2018-08-14
Visualize plan

jkstrauss:
Make VisualizePlan.svg available in docs and use simpler plan to fit better.


jkstrauss:
@lihaoyi How should I handle the merges/PRs now?

PR #411, opened 2018-08-22
Upgrade ammonite to 1.1.2-30-53edc31

smarter:
This is mainly to get https://github.com/lihaoyi/Ammonite/pull/851 which
should reduce the amount of unnecessary work done by incremental
compilation in the Mill build. This requires some code changes since
this means we now depend on a more recent version of coursier, as
a side-effect this means that we do not depend on scalaz anymore.

Also use the same ammonite version in the Mill build and in
ScalaModule#ammoniteReplClasspath.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -85,7 +85,8 @@ object main extends MillModule {
     )
 
     def ivyDeps = Agg(
-      ivy"com.lihaoyi:::ammonite:1.1.2-6-27842d9",
+      // Keep synchronized with ammonite in Versions.scala
+      ivy"com.lihaoyi:::ammonite:1.1.2-30-53edc31",
       // Necessary so we can share the JNA classes throughout the build process
       ivy"net.java.dev.jna:jna:4.5.0",
       ivy"net.java.dev.jna:jna-platform:4.5.0"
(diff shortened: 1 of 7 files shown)


smarter:
Looks like the tests are failing because the snapshot is not published for 2.11.8:
```scala
not found: https://repo1.maven.org/maven2/com/lihaoyi/ammonite_2.11.8/1.1.2-28-e8b4996/ammonite_2.11.8-1.1.2-28-e8b4996.pom
```
@lihaoyi Can the snapshot be published for 2.10/2.11 too? Otherwise I'll revert the change to ScalaModule and keep 1.1.2 there.

smarter:
There's one test failure that doesn't seem spurious, I have no idea what it means:
```scala
[406/408] dev.upstreamAssembly 
1 targets failed
publishVersion ammonite.ops.ShelloutException: CommandResult 128
fatal: bad object f424a857d8cd4435feb878b7cf9253f84814a93e
    ammonite.ops.Shellout$.executeStream(Shellout.scala:101)
    ammonite.ops.Shellout$.$anonfun$$percent$percent$1(Shellout.scala:15)
    ammonite.$file.build.$anonfun$publishVersion$4(build.sc:504)
    mill.define.Task$MappedDest.evaluate(Task.scala:348)
```
any clue @lihaoyi ?

smarter:
It didn't trigger in https://github.com/lihaoyi/mill/pull/414 which contains the current PR, so I guess not. Should we merge anyway?

lihaoyi:
re-triggering the job seems to make it fail reliably, at least the last 2-3 times i tried; probably worth investigating before merge

lihaoyi:
or maybe you can just make sure #414 really-truly works and just merge that

smarter:
I've restarted #414 a couple of times and it always came back green, so let's go with that

PR #412, opened 2018-08-22
Fix paths for ivy-style repos in grepJar

smarter:


lihaoyi:
dunno, looks ok i guess

smarter:
You can check that this makes sense by doing something like `ls ~/.ivy2/cache/org.scala-lang/scala-library/` and seeing the jars and srcs folder.

PR #414, opened 2018-08-22
Fix incremental compilation when a Scala project depends on a Java project

smarter:
This PR depends on https://github.com/lihaoyi/mill/pull/411

---

Before this PR, JavaModule#compile simply called javac
unconditionally, thus generating new classfiles every time. But if a
Scala project depends on a Java project, this will throw off the
incremental compilation algorithm which will unnecessarily recompile
files. To avoid this we now use Zinc to compile Java projects too (as a
bonus this means that Java compilation becomes incremental). This
required some refactoring in ZincWorkerImpl to be able to compile stuff
without having to pass Scala-specific options.

The issue solved by this commit could be reproduced by running in the
Mill repository:
```shell
$ mill main.compile
$ mill -i
@ main.compile()
```
and observing that before this commit, the `main.compile()` call ended
up recompiling code.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -85,7 +85,8 @@ object main extends MillModule {
     )
 
     def ivyDeps = Agg(
-      ivy"com.lihaoyi:::ammonite:1.1.2-6-27842d9",
+      // Keep synchronized with ammonite in Versions.scala
+      ivy"com.lihaoyi:::ammonite:1.1.2-30-53edc31",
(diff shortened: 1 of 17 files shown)


smarter:
@lihaoyi Please please don't use "Squash and merge" when merging this :).

lihaoyi:
looks good to me i guess, feel free to merge whenever

PR #419, opened 2018-08-28
Use the same Scala version everywhere

smarter:
Using both 2.12.4 and 2.12.6 in different places didn't seem justified.

Also add a buildVersions.sc to mirror Versions.scala (I don't think
there's a way to have a source file shared between the build and the
project itself).

### DIFF
--- a/scalalib/src/mill/scalalib/Versions.scala
+++ b/scalalib/src/mill/scalalib/Versions.scala
@@ -1,8 +1,8 @@
 package mill.scalalib
 
+// Keep this synchronized with buildVersions.sc
 object Versions {
-  // Keep synchronized with ammonite dependency in core in build.sc
   val ammonite = "1.1.2-30-53edc31"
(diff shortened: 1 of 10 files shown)


lihaoyi:
sure

scalalib/src/mill/scalalib/Versions.scala
```diff
@@ -1,8 +1,8 @@
 package mill.scalalib
 
+// Keep this synchronized with buildVersions.sc
 object Versions {
```
lihaoyi: Rather than duplicating this from the `.sc` file, could we inject these from the Mill build into the Mill executable as environment variables? We already do that in a number of places, e.g. `-DMILL_VERSION `, so this would fit right in

lihaoyi:
Sure we can use buildinfo, though that’s pretty orthogonal to this PR and
i’d rather you migrate all the env vars to buildinfo variables all at once
in a separate commit

lihaoyi:
Closing due to inactivity

PR #433, opened 2018-09-16
Dedicated scalaDoc plugins and options

Mogztter:
Add dedicated functions to define documentation options and plugins used by the `docJar` task during the scaladoc generation:

> Documentation Options
–doc-title <title>
    Define the overall title of the documentation, typically the name of the library being documented.
–doc-version <version>
    Define the overall version number of the documentation, typically the version of the library being documented.
–doc-source-url <url>
    Define a URL to be concatenated with source locations for link to source files. 

Source: https://www.scala-lang.org/old/sites/default/files/linuxsoft_archives/docu/files/tools/scaladoc.html

This feature is backward compatible, by default:

* `scalaDocPluginIvyDeps` resolves to `scalacPluginIvyDeps`
* `scalaDocOptions` resolves to `scalacOptions`

In other words it's not mandatory to define `scalaDocPluginIvyDeps` or `scalaDocOptions`.

It's possible to add extra options when generation the scaladoc:

```scala
def scalacOptions = T(Seq("-Ywarn-unused", "-Xfatal-warnings"))
def scalaDocOptions = super.scalaDocOptions() ++ Seq("-doc-version", "1.2.3") 
// Mill will use the scalac options in addition with the "-doc-version" option when generating the scaladoc
```

It's also possible to completely override `scalac` options:

```scala
def scalaDocOptions = T(Seq("-doc-title", "Mill", "-doc-version", "1.2.3"))
// Mill will only use "-doc-title" and "-doc-version" options when generating the scaladoc
```

Or to filter `scalac` options:

```scala
def scalaDocOptions = super.scalaDocOptions().filter(_ != "-Xfatal-warnings")
// Mill will use the scalac options, except "-Xfatal-warnings", when generating the scaladoc
```

### DIFF
--- a/scalalib/src/mill/scalalib/ScalaModule.scala
+++ b/scalalib/src/mill/scalalib/ScalaModule.scala
@@ -61,11 +61,11 @@ trait ScalaModule extends JavaModule { outer =>
 
   def scalacPluginIvyDeps = T{ Agg.empty[Dep] }
 
-  def scalaDocPluginIvyDeps = scalacPluginIvyDeps
+  def scalaDocPluginIvyDeps = T{ scalacPluginIvyDeps() }
 
   def scalacOptions = T{ Seq.empty[String] }
 
-  def scalaDocOptions = scalacOptions
+  def scalaDocOptions = T{ scalacOptions() }
 
   private val Milestone213 = raw"""2.13.(\d+)-M(\d+)""".r
 
--- a/scalalib/src/mill/scalalib/ScalaModule.scala
+++ b/scalalib/src/mill/scalalib/ScalaModule.scala
@@ -3,8 +3,7 @@ package scalalib
 
 import ammonite.ops._
 import coursier.Repository
-import mill.define.Task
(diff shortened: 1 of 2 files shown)


scalalib/src/mill/scalalib/ScalaModule.scala
```diff
@@ -62,8 +61,12 @@ trait ScalaModule extends JavaModule { outer =>
 
   def scalacPluginIvyDeps = T{ Agg.empty[Dep] }
 
+  def scalaDocPluginIvyDeps = scalacPluginIvyDeps
+
   def scalacOptions = T{ Seq.empty[String] }
 
+  def scalaDocOptions = scalacOptions
```
lihaoyi: This (and above) should probably be `T{ scalacOptions() }` for consistency with the rest of the code

Mogztter: Done :+1:

Mogztter:
Thanks 👍

Issue #410, opened 2018-08-19
Support for webpack in Scala.js modules.


crimzie:
Some libraries targeting Scala.js require the 'sbt-scalajs-bundler' sbt plugin in dependent projects, because they themselves depend on npm packages. To use this kind of libraries with mill one has to bundle those javascript dependencies by either applying webpack by hand, or by writing some automation, and both ways that takes deep knowledge of npm, webpack, and scala.js intricacies.

It would be nice if mill handled this by default.

lihaoyi:
Feel free to send a PR adding such a module to `contrib/`!

crimzie:
👌

MarkusA380:
For anyone coming from Google, error messages like:

```scalajsenv.js:26 Uncaught ReferenceError: exports is not defined```

are related to this. Once the PR is completed these issues [should be resolved].(https://github.com/scalacenter/scalajs-bundler/issues/193#issuecomment-352534260).

lefou:
The related PR #410 which has unresolved review comments was closed due to inactivity.

Is anyone working on this right now or has some working solution to share?

nafg:
I posted a gist there of what I got working within a project of mine. I think it works correctly but I haven't tested it heavily. I'll post it here as well:

https://gist.github.com/nafg/edb39f43daf2cd7ca8300b4252ace709

nafg:
Another option: https://github.com/iterative-works/mill-webpack

lefou:
> Another option: https://github.com/iterative-works/mill-webpack

Looks unmaintained to me. Last commit is from May 2020 and it's still using Mill 0.6.x. A lot has changing is Scala.js land since then.

lolgab:
There is https://github.com/nafg/mill-bundler which solves this very problem.
I think this issue can be closed.

lefou:
https://mill-build.com/mill/Thirdparty_Plugins.html#_bundler

PR #437, opened 2018-09-23
add contrib.ScalaJSWebpackModule that replicates sbt's scalajs-bundler (#410)

crimzie:
This PR completes issue #410 .

### DIFF
--- a/contrib/webpacklib/src/mill/contrib/ScalaJSWebpackModule.scala
+++ b/contrib/webpacklib/src/mill/contrib/ScalaJSWebpackModule.scala
@@ -1,0 +1,166 @@
+package mill
+package contrib
+
+import java.io
+import java.util.zip
+
+import ammonite.ops
+import mill.define.{Target, Task}
+import mill.scalajslib._
+import mill.scalalib._
+import mill.util.Ctx
+import ujson.Js
+
+import scala.collection.mutable
+
+trait ScalaJSWebpackModule extends ScalaJSModule {
+  def webpackVersion: Target[String] = "4.17.1"
+
+  def webpackCliVersion: Target[String] = "3.1.0"
+
+  def webpackDevServerVersion: Target[String] = "3.1.7"
+
+  case class JsDeps(
+      compileDependencies: List[(String, String)],
+      compileDevDependencies: List[(String, String)],
+      jsSources: Map[String, String]) {
+    def ++(that: JsDeps): JsDeps =
+      JsDeps(
+        compileDependencies ++ that.compileDependencies,
+        compileDevDependencies ++ that.compileDevDependencies,
+        jsSources ++ that.jsSources)
+  }
+
+  object JsDeps {
+    def apply(json: mutable.Map[String, Js.Value]): JsDeps = {
+      def read(key: String): List[(String, String)] =
+        json.get(key).fold(List.empty[(String, String)]) {
+          _.arr.flatMap {
+            _.obj.headOption.map { case (s, v) => s -> v.str }
+          }.toList
+        }
+
+      JsDeps(
+        read("compileDependencies") ++ read("compile-dependencies"),
+        read("compileDevDependencies") ++ read("compile-devDependencies"),
+        Map.empty)
+    }
+  }
+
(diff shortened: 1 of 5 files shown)


lihaoyi:
Anyone want to help me review this? As long as someone takes a look I'm happy for it to go in

contrib/webpacklib/src/mill/contrib/ScalaJSWebpackModule.scala
```diff
@@ -0,0 +1,166 @@
...
+              case z if z.getName.endsWith(".js") &&
+                !z.getName.startsWith("scala/") =>
+                JsDeps(Nil, Nil, Map(z.getName -> read(stream).toString))
+            }
+            .to[Seq]
+        stream.close()
+        deps
+      }
+      .toSeq
+      .foldLeft(JsDeps(Nil, Nil, Map.empty)) { _ ++ _ }
```
DavidGregory084: I feel like this mega-expression could be broken down a bit

DavidGregory084:
Looks good to me, just one big method calling chain I would break down a bit

(caveat: I've never used Scala.js for anything serious)

contrib/webpacklib/src/mill/contrib/ScalaJSWebpackModule.scala
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  def webpack: Task[(ops.Path, ops.Path, Boolean) => Unit] = T.task {
+    (src: ops.Path, dst: ops.Path, opt: Boolean) =>
+      val outjs = dst / src.segments.last
+      val deps = sbtBundlerDeps()
+      val cfg = "webpack.config.js"
+      ops.cp(src, outjs)
+      writeBundleSources().apply(deps, dst)
+      writeWpConfig().apply(dst, cfg, outjs.toString, "out-bundle.js", opt)
```
raquo: My wishlist:
(1) configurable bundle name and location
(2) ability to provide a path to custom webpack config file

Regarding (2), I think Mill should not be configuring webpack directly. I think letting users configure webpack via plain webpack config files is much more straightforward than via an indirect, limited DSL. Mill's role should be simply to pass any necessary information such as the paths of Scala.js output files to the config.

gavares: Not having a replacement for scaljs-bundler is just about the only thing keeping me from using mill currently. Having this change merged would be fantastic.

That said, I agree with @raquo that users should at least have the option to configure webpack directly. It's fairly complicated and the plugin ecosystem for webpack is so diverse that a cusom DSL will never keep pace or satisfy all the needs that users will have.

@crimzie , as I said I'm really looking forward to having this committed. Let me know if you could use any help.

crimzie: Hi, @raquo @gavares ! 

I was pretty busy lately and couldn't respond any earlier .

As I wrote in original issue #410, there is a problem of existing libs pulling in dependencies on js sources via sbt-scalajs-bundler. This here is a minimalistic solution for that sole problem. It wasn't originally intended as a full-fledged tool that would let users add their own js dependencies like sbt-scalajs-bundler does and as such doesn't provide customization options.

Yet I agree with your arguments and will rewrite it with an option for user to provide config, just need to find some time to do it :)

lihaoyi:
closing due to inactivity

lefou:
@crimzie Any plan to revieve this one?

nafg:
Updated version that I've used a little: https://gist.github.com/nafg/6ecce298a0a20f1e4a259cdae5634060

sequencer:
@nafg is possible for you to upstream this?

nafg:
I'll try to post my version of the code soon

sequencer:
Thank you!

nafg:
Sorry for the delay. Here is what I have currently. I believe it works but mill is not currently my main build tool so it hasn't been used for production purposes.

https://gist.github.com/nafg/edb39f43daf2cd7ca8300b4252ace709

PR #438, opened 2018-09-24
BuildInfo: Support more basic types

benjaminfrank:
Support more basic types in BuildInfo.

### DIFF
--- a/contrib/buildinfo/src/mill/contrib/BuildInfo.scala
+++ b/contrib/buildinfo/src/mill/contrib/BuildInfo.scala
@@ -13,9 +13,7 @@ trait BuildInfo extends ScalaModule {
 
   def buildInfoObjectName: String = "BuildInfo"
 
-  def buildInfoMembers: T[Map[String, String]] = T {
-    Map.empty[String, String]
-  }
+  def buildInfoMembers: Map[String, Any] = Map.empty[String, Any]
 
   private def generateBuildInfo(members: Map[String, Any])(implicit dest: Ctx.Dest): Seq[PathRef] =
   if(!members.isEmpty){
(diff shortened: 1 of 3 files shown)


contrib/buildinfo/src/mill/contrib/BuildInfo.scala
```diff
@@ -13,17 +13,23 @@ trait BuildInfo extends ScalaModule {
 
   def buildInfoObjectName: String = "BuildInfo"
 
-  def buildInfoMembers: T[Map[String, String]] = T {
-    Map.empty[String, String]
-  }
+  def buildInfoMembers: Map[String, Any] = Map.empty[String, Any]
```
Ichoran: This seems really type-unsafe.  I really wouldn't want this in unless you wrapped the types in an ADT.

benjaminfrank: I had an at ADT first but found the usability lacking: to support implicit conversion you had to put the type arguments to your map or you had to manually wrap everything yourself.
While the current approach only supports the basic types and for everything else basically relies on `toString´.
Is there a particular scenario you'd be afraid of that could happen where this leads to an issue?

contrib/buildinfo/test/src/mill/contrib/BuildInfoTests.scala
```diff
@@ -17,30 +17,55 @@ import utest.framework.TestPath
...
+  val doubleVal: Double = 1.79769313486231570e+308d
+  val charVal: Char = 'Æ'
+  val stringVal: String = "2.12.4"
+
+  val packageName = "foo"
+  val objectName = "bar"
   trait BuildInfoModule extends TestUtil.BaseModule with scalalib.ScalaModule with BuildInfo {
     def millSourcePath =  TestUtil.getSrcPathBase() / millOuterCtx.enclosing.split('.')
-    def scalaVersion = scalaVersionString
+    def scalaVersion = "2.12.6"
```
Ichoran: Why is this hard-coded?

benjaminfrank: it's the build for the test and needs a scalaversion.

benjaminfrank:
@rockjam @lihaoyi any opinion om the typesafety issue here?
only drawback I found when using an ADT was that the Map requires an type annottaion when using implict conversion, e.g., 
``` scala
def buildInfoMembers = 
  Map[String, ADT](
    "intVal" -> 1)
```
or using the ADT directly, e.g.,
``` scala
def buildInfoMembers  = 
  Map.empty(
    "intVal" -> ADTInt(1))
```

lihaoyi:
I don't care much either way, but if you want to avoid the type ascription you can always define your own `BuildInfo(args: (String, ADT)*)` function so people can `def buildInfoMembers = BuildInfo("intVal" -> 1)` and have the implicits kick in correctly

benjaminfrank:
hm didn't think of that. I will give it a try later. thanks

lihaoyi:
closing due to inactivity

PR #441, opened 2018-09-28
Fixes #440: fix for allowing subgroups of a given staging profile

### LINKED ISSUE #440: Publishing to subgroups
I may have identified a bug, though as always, more likely a user error. When the org/groupId is set to `w.x.y.z` in the `build.sc` file and I own the groupId `w.x.y` on Maven/Sontatype/whatever.

I'm getting this error now: `Could not find staging profile for groupId: edu.cornell.ncrn.ced2ar`. I've uploaded to this groupId before using maven/pom, but the issue might be that the top level groupId associated with the profile is actually just edu.cornell.ncrn. 

So I thought on this line ( https://github.com/lihaoyi/mill/blob/master/scalalib/src/mill/scalalib/publish/SonatypeHttpApi.scala#L39 ), maybe it should be checking that profile("name").str is a substring of groupId (rather than equal to it). But I tested it out locally (hopefully correctly using mill/out/dev/launcher/dest/run) using the change `.find(profile => profile("name").str.startsWith(groupId)` and got the same error. So, I'm not sure what to investigate next.

bbarker:
Probably due to my raging cold, but I had it backwards in the issue.

### DIFF
--- a/scalalib/src/mill/scalalib/publish/SonatypeHttpApi.scala
+++ b/scalalib/src/mill/scalalib/publish/SonatypeHttpApi.scala
@@ -36,7 +36,8 @@ class SonatypeHttpApi(uri: String, credentials: String) {
       ujson
         .read(response.body)("data")
         .arr
-        .find(profile => profile("name").str == groupId)
+        .find(profile =>
+          groupId.split('.').startsWith(profile("name").str.split('.')))
         .map(_("resourceURI").str.toString)
 
     resourceUri.getOrElse(


lihaoyi:
Maybe we could split it on `.` and use `Seq#startsWith`? We don't want `com.lihaoyi` to be treated as a prefix of `com.lihaoyi2`

bbarker:
Nice catch - I tested it and it seems to work (updated commit in PR)

PR #458, opened 2018-10-07
Configure Idea project generator per module

lefou:
This PR is a first sketch how to skip some modules in Idea project generator.

The motivation is simple:
* Idea cannot handle multiple projects with the same base directory
* When working with Cross to enable cross compile projects of any kind, Idea selects one of the cross versions (almost) randomly per directory, which is undesired

This is a quickly hacked PR to sketch one possible solution to tackle the need to customize project generation. I'm open for input / change requests. If we can reach a consensus about the general direction, we can beautify, e.g. move the `IdeaConfigModule` into it's own file. Also tests would be nice, but I'm not familiar with the integration test setup of mill yet.

### DIFF
--- a/scalalib/src/mill/scalalib/GenIdeaImpl.scala
+++ b/scalalib/src/mill/scalalib/GenIdeaImpl.scala
@@ -28,7 +28,7 @@ object GenIdea extends ExternalModule {
 }
 
 /**
-  * Marker trait to manipulate idea project generation.
+  * Module specific control over Idea project file generation.
   */
 trait IdeaConfigModule extends mill.Module {
   def skipIdea: Boolean = false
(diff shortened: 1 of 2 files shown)


lefou:
What about "grouping" idea related def's by `trait JavaModule extends IdeaConfigModule`?

lihaoyi:
Currently idea-related defs aren't standalone; they live mixed in with the JavaModule defs, as bits and pieces refactored out to be conveniently consumed by intellij. Maybe in future when we have more idea-related stuff we can extract it out into it's own trait

scalalib/src/mill/scalalib/GenIdeaImpl.scala
```diff
@@ -225,8 +226,9 @@ object GenIdeaImpl {
       Tuple2(
         ".idea"/"modules.xml",
         allModulesXmlTemplate(
-          for((path, mod) <- modules)
-            yield moduleName(path)
+          modules
+            .filter(!_._2.skipIdea)
```
megri: This could use a collect instead to avoid iterating over the modules twice
```
.collect {
  case (path, mod) if !mod.skipIdea => moduleName(path)
}
```

PR #459, opened 2018-10-09
Add a filter parameter (default true for all files) to createJar().

ucbjrl:
Add createJar() filter test.

### DIFF
--- a/main/src/mill/modules/Jvm.scala
+++ b/main/src/mill/modules/Jvm.scala
@@ -209,7 +209,9 @@ object Jvm {
     m
   }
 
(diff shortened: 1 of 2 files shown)


lefou:
Just curious, what is the motivation behind the filter?
Some ScalaDoc wouldn't be bad.

ucbjrl:
When creating source or doc jars, there may be files in the source directories that should not be included in the jar (notably `.DS_Store` files on Macs). I think the decision on which files to be excluded should be left up to the client of createJar() and not hard-coded in the createJar() implementation.

Good point on the lack of ScalaDoc. I'll generate some.

PR #461, opened 2018-10-10
Added mill-osgi module to documentation, contrib modules

lefou:
Module is still in development, but can already produce proper bundles.


lefou:
Thanks! Merged.

PR #464, opened 2018-10-16
Add tut contrib module

DavidGregory084:
- [x] Add tut module
- [x] Add tests
- [x] Add docs


contrib/tut/src/mill/contrib/tut/TutModule.scala
```diff
@@ -0,0 +1,44 @@
...
+  def tutClasspath: T[Agg[PathRef]] = tutJar() ++ compileClasspath()
+  def tutScalacPluginIvyDeps: T[Agg[Dep]] = scalacPluginIvyDeps()
+  def tutNameFilter: T[Regex] = T { """.*\.(md|markdown|txt|htm|html)""".r }
+  def tutScalacOptions: T[Seq[String]] = scalacOptions()
+  def tutVersion: T[String] = "0.6.7"
+
+  def tutJar: T[Agg[PathRef]] = T {
+    Lib.resolveDependencies(
+      repositories :+ MavenRepository(s"https://dl.bintray.com/tpolecat/maven"),
+      Lib.depToDependency(_, "2.12.4"),
```
DavidGregory084: Is there a "millScalaVersion" somewhere that we can use for definitions like this?

DavidGregory084: Actually I don't think that makes sense here - we want Tut to run with the `scalaVersion` of the module

DavidGregory084:
I want to add the `localClasspath` to the `tutClasspath` but this causes the same issue that @nh13 experience on their coverage PR

I'm trying to figure out what causes that problem

DavidGregory084:
It seems like this is due to the system property MILL_VERSION being unset - this is passed directly to coursier as a version for mill-scalalib-worker

nh13:
@DavidGregory084 I'll definitely keep an eye on this.  I had issues with ivy dependency resolution too in the latest version of my PR :/.

DavidGregory084:
@nh13 I had to set the testArgs in the module in order to avert the issue

DavidGregory084:
I'm still running into some funny business with the classpath not being used by the scala interpreter. I think there might be some SBT specific stuff in there somewhere, maybe this: https://github.com/tpolecat/tut/blob/series/0.6.x/modules/core/src/main/scala/tut/FileIO.scala#L79

I'll keep chipping away in any case

build.sc
```diff
@@ -250,6 +250,10 @@ object contrib extends MillModule {
     }
    }
 
+  object tut extends MillModule {
+    def moduleDeps = Seq(scalalib)
+    def testArgs = Seq("-DMILL_VERSION=" + build.publishVersion()._1)
```
nh13: should this be `_.2` like it is in other parts of the build file?

DavidGregory084: I tried that, but unfortunately in a project with local changes it just meant that Coursier was unable to resolve the dependency (as MILL_VERSION is some unpublished version of mill)

DavidGregory084:
Ok, I think this is ready to go now

I was disappointed to have to shell out to call Tut but there was no obvious way to change it to make sure it used the class path of the context class loader in the interpreter (I may come back to that later).

I also added a test using libraries and scalac plugins to make sure that worked

lefou:
I'd love to see the already good task documentation also as ScalaDoc. After all, mill is scala source based and at least when used in IntelliJ, the ScalaDoc would be immediately visible when configuring the module.

DavidGregory084:
Great idea!

lefou:
LGTM.

@DavidGregory084 Do you plan to add the tasks ScalaDoc? As soon as you give green light, I can merge this PR.

DavidGregory084:
@lefou I should be able to take a look at that this evening, or you can merge this one now and I'll send a follow up PR to do that for all the contrib modules? I don't mind which :)

lefou:
@DavidGregory084 As this PR needs a conflict resolve, I'd rather wait for your rebase and ScalaDoc commit than merge as is.

DavidGregory084:
Okay @lefou I think this is ready to go pending your review

contrib/tut/src/mill/contrib/tut/TutModule.scala
```diff
@@ -6,23 +6,96 @@ import coursier.MavenRepository
...
+      "-Ywarn-unused:imports",
+      "-Ywarn-unused-import",
+      "-Ywarn-dead-code",
+      "-Xfatal-warnings"
+    ))
+
+  /**
+    * The version of Tut to use.
+    */
   def tutVersion: T[String] = "0.6.7"
```
lefou: To follow the same pattern as in other mill modules (e.g. ScalaModule), we should not hard code any versions. Hence this field should be mandatory.

DavidGregory084: Yep, seems reasonable

DavidGregory084: OK, the latest commit removes the default version here

lefou:
Thank you!

PR #482, opened 2018-11-06
Fix broken link

buaya91:
The link to docs is broken, not sure if this is the correct one, I also notice there are docs in this repo, are they hosted somewhere?


lihaoyi:
This isn't the fix, lemme take a look

PR #484, opened 2018-11-08
tiny PR fixing ScalafmtModule usability

vitaliis:
removed parentheses from ScalafmtModule.reformat (details below)

Tasks in `Mill` (e.g. compile) don't have parentheses. Having them for reformat introduces an issue in usability when trying to apply this task.

I just spent some time figuring out why `reformat` is not executed having this code:

```
  override def compile = T {
    reformat()
    super.compile()
  }
```

while the same code sample works in https://github.com/rockjam/mill-basic-stuff/blob/master/build.sc#L45

### Current workaround

With current code one has to write of the following lines to actually execute reformat:
```
reformat().apply()
// OR
reformat()()
```
Which is inconvenient and currently not in line with the rest.

### DIFF
--- a/scalalib/src/mill/scalalib/scalafmt/ScalafmtModule.scala
+++ b/scalalib/src/mill/scalalib/scalafmt/ScalafmtModule.scala
@@ -6,7 +6,7 @@ import mill.scalalib._
 
 trait ScalafmtModule extends JavaModule {
 
-  def reformat(): Command[Unit] = T.command {
+  def reformat: Command[Unit] = T.command {
     ScalafmtWorkerModule
       .worker()
       .reformat(


lefou:
Actually, reformat is (currently) a `T.command` and by convention, commands are defined with parenthesis. Commands will not be cached and will always re-compute. Also, they can take parameters. See [Commands section of the mill website](http://www.lihaoyi.com/mill/page/tasks.html#commands).

vitaliis:
And simple `target` doesn't fit because `sources()` doesn't give you a set of changed sources, but all of them. Is it correct?

And using `input` task would be too much here? Reformatting all the files on every execution.

PR #485, opened 2018-11-08
Pass long Mill props via file

sake92:
When developing Mill on Windows, there is an unavoidable issue of command line limit of about 8000 characters.  
The only way to bypass it is to have less parameters. This change writes all of Mill's `LongProps` to a `mill.properties` file. Then it passes that file path as system property called `MILL_OPTIONS_PATH`...

IMHO, it's not ok to build different launchers on different OS-es, e.g. `launcher + prependShellScript` builds one artifact on Linux and **different** on Windows..   
I think we should build an executable JAR with just one entry point. And have scripts for Bash and Batch, just like Gradle does... That would simplify things a lot I think.


sake92:
Superseded by #791

PR #486, opened 2018-11-23
Add Maven/SBT-style integration test folders

leonardehrenfried:
This PR adds the folders under `src/it/` to the `SbtModule`.

These folders are used to separate the integration tests (slow) from the normal unit tests (fast).

The documentation is a little sparse but I'm not sure where exactly I should add it. Any pointers would be welcome.

### DIFF
--- a/docs/pages/3 - Common Project Layouts.md
+++ b/docs/pages/3 - Common Project Layouts.md
@@ -217,6 +217,8 @@ foo/
             scala/
         test/
             scala/
+        it/
+            scala/
 ```
 
 Useful if you want to migrate an existing project built with SBT without having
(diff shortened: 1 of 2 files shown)


leonardehrenfried:
The windows build has failed with the following error:

```
java.nio.file.FileSystemException: C:\mill\target\workspace\mill\eval\EvaluationTests\Checker\eval
  uator\evaluateSingle\diamond\down\log: The process cannot access the file because it is being used
   by another process.
    sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:86)
```
Is this a known issue?

lefou:
As far as I understand, integration tests are no standard setup, or at least they must be explicitly enabled (see https://www.scala-sbt.org/1.x/docs/Testing.html#Integration+Tests).
I tend to not merge this one, as I think it's seldomly used and is pretty easy to add in the `build.sc` if needed. Also, this PR is pretty old and inactive and needs some updates.

PR #488, opened 2018-11-25
A small tweak to the marketing

propensive:


nafg:
Are you so pessimistic about it? :laughing:

PR #489, opened 2018-11-29
Add Gitlab to VersionControl

leonardehrenfried:


leonardehrenfried:
Naturally, I disagree with this assessment.

Since there is already a convenience method for Github, I don't see a reason for not having one for Gitlab, too.

Besides the ideological argument, this is low complexity code that doesn't change any behaviour of mill.

lefou:
Fair enough. Mill is hosted on Github, that could explain the special case. I'd be fine with removing that one, too. After all, defining some publishing metadata shortcuts isn't the core of a build tool. Best place might be a plugin. But I will not have the final word on this PR. /cc @lihaoyi

SRGOM:
Might as well add bitbucket, gogs, gitorious, nonameuselessshitgit as well. 

Github, as I said, has two orders of magnitude more activity. And deserves a special case. I'd rather see a new protocol added (mercurial) which will actually be useful than somebody using this project to make stupid political statements.

lihaoyi:
@SRGOM please do not be rude.

I don't mind adding this for convenience. If someone wants to add bitbucket, phabricator, or other things, that's fine too.

PR #494, opened 2018-12-04
Avoid unnecessary dependency downloading by providing fetches per cache policy

gehnaphore:
Also adds basic ticker logging to indicate when they are downloading.


alexarchambault:
I confirm this is the right thing to do (~chaining the cache policies of `coursier.CachePolicy.default`), the methods on `coursier.Cache` should soon accept a sequence of `CachePolicy` rather than just one `CachePolicy`, with `coursier.CachePolicy.default` as a default value.

alexarchambault:
One can get unnecessary network requests with the current default parameters for `cachePolicy` in the methods of `coursier.Cache`.

lefou:
Thank you!

Issue #508, opened 2018-12-13
build.sc exceeds JVM code size limits


kongo2002:
Hi,

I just tried to migrate a rather large SBT project to `mill`. The experience is great so far! However I do run into problems now that I am about to add all project's modules (approx 100 modules).

I tried to move some methods and stuff around but now I cannot continue without running into errors like these:

```
Could not write class ammonite/$file/build$ because it exceeds JVM code size limits. Method ammonite/main/Router$Overrides's code too large!
Compilation Failed
```

```
Could not write class ammonite/$file/build$ because it exceeds JVM code size limits. Method null's code too large!
Compilation Failed
```

```
Could not write class ammonite/$file/build$ because it exceeds JVM code size limits. Method mill/util/Router$EntryPoint's code too large!
Compilation Failed
```

As you can see the method names do vary. 

I will try to come up with a reproducible example but for now I can say the modules are at max 3 levels deep and are basically plain `SbtModule` instances:

```scala
object base extends Module {
  object sub extends Module {
    object inner1 extends SbtModule {
      def moduleDeps = Seq(some.other.module, some.more.module)
    }
    object inner2 extends SbtModule {
      def moduleDeps = Seq(base.sub.inner1, even.more.module)
    }
    // ...
  }
}
```

Cheers
Gregor

kongo2002:
Thanks for the quick response! Sadly this does not resolve the issue:

```
Could not write class ammonite/$file/build$ because it exceeds JVM code size limits. Method millDiscover$lzycompute's code too large!
Compilation Failed
```

I've added this to the `conf.sc` like you mentioned:

```scala
interp.configureCompiler(_.settings.Ydelambdafy.tryToSetColon(List("inline")))
```

lihaoyi:
Is it a different exception than before? Might still be progress even if it still doesn't work! Now we just need to break up `millDiscover` method rather than breaking up all the other methods your original thing reported

kongo2002:
That's true indeed. I did try a couple of times and now it's always `millDiscover$lzycompute` in the exception.

kongo2002:
I am not sure if my scala macro skills are solid enough to fix this one by myself ;-)

However I was able to come up with a minimal reproducible example (simply put in an empty directory):

```scala
// build.sc
import $file.conf
import mill._, scalalib._

trait TModule extends SbtModule {
  def scalaVersion = "2.12.7"
}

object foo extends Module {
  object common extends Module {
    object one extends TModule {
[...]

PR #509, opened 2018-12-13
Discover - break overridesRoutes into fixed size chunks

kongo2002:
I now did try to tackle #508 - as I said before I have not too much experience in scala macros so there might be a much nicer solution to this problem.

However this change successfully builds the example I posted before.

### DIFF
--- a/integration/test/resources/large-project/build.sc
+++ b/integration/test/resources/large-project/build.sc
@@ -1,0 +1,294 @@
+import $file.conf
+import mill._, scalalib._
+
+trait TModule extends SbtModule {
+  def scalaVersion = "2.12.7"
+}
+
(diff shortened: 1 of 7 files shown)


kongo2002:
The added `LargeProjectTests` fails without the fix and succeeds afterwards. I hope I put it in the right place there :)

lihaoyi:
@kongo2002 looks good. One more thing: could you try running the test without `Ydelambdafy=inline` and see if anything breaks?

kongo2002:
Yes, it does work even without the `Ydelambdafy=inline` - I removed this from the test as well.

lihaoyi:
looks good, thanks!

PR #516, opened 2018-12-19
Accept sysprops via cmdline and forward to mill server

lefou:
See https://github.com/lihaoyi/mill/issues/498
See https://github.com/lihaoyi/mill/issues/499

### DIFF
--- a/main/src/MillMain.scala
+++ b/main/src/MillMain.scala
@@ -24,7 +24,7 @@ object MillMain {
       System.err,
       System.getenv().asScala.toMap,
       b => (),
-      System.getProperties().asScala.toMap
+      initialSystemProperties = Map()
     )
     System.exit(if (result) 0 else 1)
   }
@@ -135,6 +135,8 @@ object MillMain {
           stderr.println("Build repl needs to be run with the -i/--interactive flag")
           (false, stateCache)
         }else{
+          val systemProps = initialSystemProperties ++ extraSystemProperties
+
           val config =
             if(!repl) cliConfig
             else cliConfig.copy(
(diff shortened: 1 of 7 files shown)


lefou:
Any idea, why the travis build failed?

```
1 targets failed
all 1 targets failed
publishVersion ammonite.ops.ShelloutException: CommandResult 128
fatal: bad object 3ef5f3e450546c58545f973eba31bdfaab39dffc
    ammonite.ops.Shellout$.executeStream(Shellout.scala:101)
    ammonite.ops.Shellout$.$anonfun$$percent$percent$1(Shellout.scala:15)
    ammonite.$file.build.$anonfun$publishVersion$4(build.sc:525)
    mill.define.Task$MappedDest.evaluate(Task.scala:337)
The command "$CI_SCRIPT" exited with 1.
```

lihaoyi:
That looks like typical travis flakiness; retrying it a few times has made it work for me in the past

lefou:
The only reason why I still not merged this PR is the fact that it copies the whole system property set, instead of just those explicitly given. I think, at least some syspros out of the `java.` scope and probably many others, which are typically controlled by the JVM should not be set explicitly. But If we want to keep to support the `JAVA_OPTS` support, we don't know which properties were explicitly set. In contract to the `-D` cmdline option, which is perfectly under our control.

lihaoyi:
@lefou I'm fine with a hardcoded list of "builtin" properties we do not propagate. 

I don't really understand the problem with `JAVA_OPTS` you are describing

lefou:
I now no longer copy all sys props. Properties given with `-D` will now properly propagated to a running server. There are no tests for that behavior, as I have no idea how to best implement such a test with the current integration test machinery. @lihaoyi Could you please review and comment, if merging without tests is OK?

PR #518, opened 2018-12-20
Added some docs to result type

lefou:
@lihaoyi Can you please check if these are correct?


lihaoyi:
Looks good to me. Note the difference between `Failed` which generally is intentionally used, and `Exception` which refers to things that blow up generally-unintentionally

lefou:
With `Failed` you mean `Failure`, don't you?

lihaoyi:
yea

Issue #529, opened 2019-01-21
Cannot select PGP key when signing files


lxohi:
Hi,

I'm trying to publish [my project](https://github.com/ant-design-binding/ant-design-binding) to Maven Central using mill.

Setting up is quite fast & easy. But then I discovered that I cannot select the key used in PGP signing meanwhile there are multiple key pairs in my local machine.

I think maybe add a protected ```Option[String]``` member to ```PublishModule``` that specifies the key name and pass it all the way to [here](https://github.com/lihaoyi/mill/blob/9ba4cb69331386dfde9bac69dc2d5b22401face3/scalalib/src/publish/SonatypePublisher.scala#L138) and then use it as the ```-u``` param of the gpg command will solve this issue.

What do you think about this change? If it's good, may I create a PR for this?

lihaoyi:
sure

lihaoyi:
@lxohi I would prefer just passing in the key name as a flag to `PublishModule.publishAll`

lxohi:
OK. I'll send a PR later that passing the key name as a flag.

lxohi:
Closed due to #530 was already merged.

PR #530, opened 2019-01-22
PublishModule: add --gpgKeyName flag

lxohi:
PR for #529 

Add ```--gpgKeyName``` flag to ```.publish``` & ```.publishAll```

I was not modify any test cases because I cannot find any related with ```PublishModule```.
New flag has been manually executed & verified on my local machine.


lxohi:
Note that ```mill integration.test``` will fail on the ```master``` branch so this CI probably will fail too.

PR #544, opened 2019-02-11
Create a Play! module to compile the router

jeantil:
Hi, 

This pull request reboots #421, it creates a `playlib` module to provide a `RouterModule` able to process the play routes file and generate the corresponding source code. The next step would be to create a `PlayModule` to account for the custom directory layout imposed by Play!.

I took #421, rebased it on master then reworked it to use the same dynamic worker mechanism used by the `scalajslib` module.

This massively simplified the manipulation of the Play! router and the translation of its return values.


contrib/playlib/src/mill/playlib/RouterModule.scala
```diff
@@ -0,0 +1,120 @@
...
+    */
+  def playVersion: T[String]
+
+  /**
+    * The scala version to use for this project. The version defaults to
+    * the value chosen by play but some version are crossbuilt for multiple
+    * versions of scala.
+    * Override it if the default doesn't match your current usage.
+    */
+  override def scalaVersion: T[String] = T {
```
lefou: I'm not sure, I like that default here. What about other modules, that also like to override `scalaVersion`.

docs/pages/9 - Contrib Modules.md
```diff
@@ -165,76 +165,68 @@ object app extends ScalaModule with TwirlModule {
 } 
 ``` 
 
+### Play Framework
+
+Play framework routes generation support.
+
+ 
+To declare a module that needs to generate Play Framework routes , you must mix-in the
```
lefou:
```suggestion
To declare a module that needs to generate Play Framework routes, you must mix-in the 
```

docs/pages/9 - Contrib Modules.md
```diff
@@ -165,76 +165,68 @@ object app extends ScalaModule with TwirlModule {
...
-This task will compile `*.scala.html` templates (and others, like `*.scala.txt`) into the `out/app/compileTwirl/dest` 
+This task will compile `routes` templates into the `out/app/compileRouter/dest` 
 directory. This directory must be added to the generated sources of the module to be compiled and made accessible from the rest of the code:
 ```scala
-object app extends ScalaModule with TwirlModule {
-  def twirlVersion = "1.3.15"
-  def generatedSources = T{ Seq(compileTwirl().classes) }
+object app extends ScalaModule with RouterModule {
+  def playVersion= "2.7.0"
+  def generatedSources = T{ Seq(compileRouter().classes) }
```
lefou: Can't we provide that behavior by default? If not, we should at least suggest a cumulative addition.
 ```suggestion
  def generatedSources = T{ super.generatedSources() ++ Seq(compileRouter().classes) }
```

File: contrib/playlib/src/mill/playlib/RouterModule.scala (hunk 1 above)
Comment (jeantil) [reply contrib/playlib/src/mill/playlib/RouterModule.scala]: The play sbt plugin provides a default scala version which is why I added this default behaviour. 
Since all modules inherit from ScalaModule the last mixed-in will "define" the actual scalaVersion used and the end user can always override in his own object. 

With that said I don't specifically care either way as I always define my own explicit scalaVersion in my app build definition.

File: docs/pages/9 - Contrib Modules.md (hunk 3 above)
Comment (jeantil) [reply docs/pages/9 - Contrib Modules.md]: We can probably provide that by default, I followed the structure of the TwirlModule and didn't add the behaviour by default. I don't mind adding it.

Comment (lefou) [reply docs/pages/9 - Contrib Modules.md]: With "behavior" I meant the addition of the `generatedSources`. It looks like as they need to be added in all cases.

Comment (lefou) [reply docs/pages/9 - Contrib Modules.md]: Ups, commented at same time... What I wanted to say, if we always need to add those generatedSources, than we should provide a reasonable default.

Comment (jeantil) [reply docs/pages/9 - Contrib Modules.md]: >With "behavior" I meant the addition of the `generatedSources`
I got that :) As I said I matched the way the TwirlModule did things, as far as I know this behavior would be needed for Twirl too.

File: docs/pages/9 - Contrib Modules.md (hunk 2 above)
Comment (jeantil) [reply docs/pages/9 - Contrib Modules.md]: (i'll fixup the fixes and push once I have resolved all comments if that's acceptable. I'll mark :+1: on all comments I have processed)

File: docs/pages/9 - Contrib Modules.md (hunk 3 above)
Comment (jeantil) [reply docs/pages/9 - Contrib Modules.md]: I'll add the default to the RouterModule. I guess I could open a separate PR to add it to the twirl module too :)

**Comment by jeantil** [comment]
>Also, as the typical play framework app development is highly coupled to a running sbt instance, we should probably line out, how the development workflow looks like with mill. E.g. should/can we start mill with `--watch` option?

This PR only aims to add a route compiler module which coupled with the twirl module lets you completely compile most play applications. I don't intend to address running the app in this PR.

I intend to do a follow-up PR to provide a PlayModule which configures the build to match the default play application directory structure, includes the RouterModule since you cannot write a useful play application without a route file, and include a basic run facility (most likely akin to the `stage` command in sbt).
I'll leave out the TwirlModule by default as you can definitely write useful play apps without twirl.

These two should enable a basic play workflow. If I have some more time, I'll look into running in dev mode but that's highly coupled to SBT and could require significantly more code if we want the same classloading behavior.

**Comment by jeantil** [comment]
Perhaps @Mogztter who initiated the first PR has an opinion :)

File: contrib/playlib/src/mill/playlib/RouterModule.scala [hunk 4]
```diff
@@ -0,0 +1,122 @@
...
+
+  override def generatedSources = T{ super.generatedSources() ++ Seq(compileRouter().classes) }
+
+  /**
+    * The [[PathRef]] to the main routes file.
+    *
+    * This is the default path for play projects and it should be fine but you
+    * can override it if needed.
+    */
+  def routesFile: T[PathRef] = T {
```
lefou: I think, this should be a `T.sources`, but am not 100% sure, because it is only a file and not a directory.

http://www.lihaoyi.com/mill/page/tasks.html#sources

Mogztter:
Oh wow thanks @jeantil for taking care of it!

> This PR only aims to add a route compiler module which coupled with the twirl module lets you completely compile most play applications. I don't intend to address running the app in this PR.

I think it's the right approach, one step at a time :+1: 

> I intend to do a follow-up PR to provide a PlayModule which configures the build to match the default play application directory structure, includes the RouterModule since you cannot write a useful play application without a route file, and include a basic run facility (most likely akin to the `stage` command in sbt).

:+1: 
I would have done two pull requests, one to provide `PlayModule` and one with the run facility.

> I'll leave out the TwirlModule by default as you can definitely write useful play apps without twirl.

`TwirlModule` can also be used outside Play! so we should definitely not include this module by default.
However we should make sure that it can play nice with the `PlayModule`.

> These two should enable a basic play workflow. If I have some more time, I'll look into running in dev mode but that's highly coupled to SBT and could require significantly more code if we want the same classloading behavior.

Indeed, I didn't take a closer look but I guess it will not be an easy task (but a challenging one!).

contrib/playlib/src/mill/playlib/RouterModule.scala
Mogztter: I think it might be a bit too early to define a default value. I would let the version "undefined" so the user must provide a version.
And we can always come back to it and define a default value if we feel like it's tedious to declare the Scala version.

contrib/playlib/src/mill/playlib/RouterModule.scala
jeantil: It may be necessary, I have to investigate how the play routes compiler handles route file inclusion (https://www.playframework.com/documentation/2.6.x/SBTSubProjects) as it seems that the included file is not compiled at the moment. I'll update the pr when I understand how this works.

lefou: Without the `.sources` it will at least never recompile after a change, as mill will cache it until next clean (which evicts the cache). The `.sources` forces mill to always run and check for changes.

docs/pages/9 - Contrib Modules.md
```diff
@@ -233,7 +233,65 @@ These imports will always be added to every template.  You don't need to list th
...
+import $ivy.`com.lihaoyi::mill-contrib-playlib:0.3.3`,  mill.playlib._
+object app extends RouterModule {
+// ...
+} 
+``` 
+
+#### Configuration options
+
+  * `def playVersion: T[String]` (mandatory) - The version of play to use to compile the routes file.
+  * `def scalaVersion: T[String]` - The scalaVersion in use in your project. (Defaults to the default scalaVersion for a given play version.)
```
lefou: This has no longer a default.

```suggestion
  * `def scalaVersion: T[String]` - The scalaVersion in use in your project.
```

contrib/playlib/src/mill/playlib/RouterModule.scala
jeantil: ok I had a look and it seems T.sources is going to be needed and I do have to handle a list of files for it to behave correctly

jeantil: T.sources only seems to accept directories. in my case the directory is shared by multiple files only some of which are consumed by the routes compiler. If I understand correctly this may end up triggering unneeded compilations if some other files change around the.

I tried working around this by : 

* defining a setting `routesDirectory` which is a T.sources
* defining a private setting `routesFiles` which is a T[PathRef] and selects the proper files from the `routesDirectory` (it also makes sure they appear in the proper order so the main routes file is processed last)
* making the compileRouter task depend on the `routesFiles` target

If I understand correctly, any change to any files under `routesDirectory` will trigger a trivial reevaluation of routesFiles. The output of routesFiles will not change unless a route file has changed which will prevent useless compilations and should cost less than fully compiling the router.
I hope I got this right as it would be very inconvenient to have to clean the cache manually to force a router recompile after the

lefou: It should work as you have outlined. I agree, we should not be forced to implement such workarounds in mill for such a common use case. Maybe @lihaoyi can shed some light how to best tackle single source files?

contrib/playlib/src/mill/playlib/RouterModule.scala
```diff
@@ -51,9 +65,10 @@ trait RouterModule extends mill.Module with ScalaModule {
   /**
     * The routes compiler type to be used. Can only be one of:
     * <ul>
-    *   <li>[[RouteCompilerType.InjectedGenerator]]
-    *   <li>[[RouteCompilerType.StaticGenerator]]
+    * <li>[[RouteCompilerType.InjectedGenerator]]
```
lefou: Just a hint: instead of a JavaDoc-style list, in Scaladoc you can simply use `- `.

https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html

jeantil:
Having addressed most of the review comments, I'll rebase and squash the fixups.

lefou:
Nice work! Thanks to both of you @Mogztter  and @jeantil .

Mogztter:
@jeantil did all the heavy lifting :muscle: 
Well done! :champagne:

jeantil:
thank you :)

Issue #476, opened 2018-11-02
Provide a mill wrapper to be included into repositories (`./mill`)


lefou:
The wrapper should be:
* cross-platform
* small
* easy to inspect
* avoid containing bytecode (if needed, it could ad-hoc compile source code)

Use cases:
* Easy checkout and build without need to install mill
* Provide always the correct version of mill
* Easy to use mill in CI environments

lihaoyi-databricks:
I'd be happy to just call it `./mill`. The leading `./` ensures there'll be no ambiguity with the PATH `mill` anyway.

I don't see any reason why the mill wrapper would need to be written in anything but Bash; it just needs to poke at some files and maybe download something.

lihaoyi:
yeah bash works, and a mill.bat works too

lefou:
I created project https://github.com/lefou/millw

But a Windows Batch version is still missing.

PR #558, opened 2019-02-23
Added a simple mill wrapper script (sh)

lefou:
See also https://github.com/lihaoyi/mill/issues/476

### DIFF
--- a/mill
+++ b/mill
@@ -14,7 +14,7 @@ if [ "x$1" = "x--mill-version" ] ; then
     MILL_VERSION="$1"
     shift
   else
-    echo "You specified -v without a version."
+    echo "You specified --mill-version without a version."
     echo "Please provide a version that matches one provided on"
     echo "https://github.com/lihaoyi/mill/releases"
     false
--- a/mill
+++ b/mill
@@ -1,0 +1,45 @@
+#!/usr/bin/env sh
+
+# This is a wrapper script, that automatically download mill from GitHub release pages
+# You can give the required mill version with --mill-version parameter
+# If no version is given, it falls back to the value of DEFAULT_MILL_VERSION
+
+DEFAULT_MILL_VERSION=0.2.7-10-db8525
+
+set -e
+
+if [ "x$1" = "x--mill-version" ] ; then
+  shift
+  if [ "x$1" != "x" ] ; then
+    MILL_VERSION="$1"
+    shift
+  else
+    echo "You specified -v without a version."
+    echo "Please provide a version that matches one provided on"
+    echo "https://github.com/lihaoyi/mill/releases"
+    false
+  fi
+fi
+
+if [ "x${MILL_VERSION}" = "x" ] ; then
+  MILL_VERSION="${DEFAULT_MILL_VERSION}"
+fi
+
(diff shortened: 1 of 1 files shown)


mill
```diff
@@ -0,0 +1,45 @@
...
+
+# This is a wrapper script, that automatically download mill from GitHub release pages
+# You can give the required mill version with --mill-version parameter
+# If no version is given, it falls back to the value of DEFAULT_MILL_VERSION
+
+DEFAULT_MILL_VERSION=0.2.7-10-db8525
+
+set -e
+
+if [ "x$1" = "x--mill-version" ] ; then
```
lihaoyi: My Bash is a bit rusty, but why did you add the `x` prefix before the LHS and RHS of the comparison?

mill
```diff
@@ -0,0 +1,45 @@
...
+  shift
+  if [ "x$1" != "x" ] ; then
+    MILL_VERSION="$1"
+    shift
+  else
+    echo "You specified -v without a version."
```
lihaoyi: Should this be `"You specified --mill-version without a version."`

mill
```diff
@@ -0,0 +1,45 @@
...
+    echo "Please provide a version that matches one provided on"
+    echo "https://github.com/lihaoyi/mill/releases"
```
lihaoyi: Does this only support stable versions, or unstable versions too? It could be handy for our own development to allow a flag to easily pass in an unstable version so we can test our own code that hasn't been tagged with a numbered release

mill
lihaoyi: Also, to be clear, this `--mill-version` param must be the first argument, before any other parameters, and before `--interactive`/`-i`? If so we can give a nice error message if the user passes in `--mill-version` in another position, similar to what we do for `--interactive`/`-i`, by creating a dummy command-line argument in Mill's parser and error-ing out if that argument is provided. See https://github.com/lihaoyi/mill/blob/master/main/src/MillMain.scala#L94

lefou: The "x$i" thing is an old habit. Can't tell you anymore, why that was needed, but 20 years ago, almost all scripts looked like this. Could be some prevention of some unwanted executions and the `test` tool, who knows?

Too keep the script simple, I parse that option only when it's the first argument. I'm a bit uncomfortable with an dummy `--mill-version` option in mill. It could also check if the given version is in fact the runtime version, as a safety measure. Then there would be no need to remove it.

mill
lefou: You are right, I changed it from a simple `-v` to `--mill-version` to avoid prospective collisions.

mill
lefou: It supports all downloadable version, also these n-commits after-release downloads. In fact, I wrote the script exactly for that use case, called it `mill-launcher` and then realized, that it's already the solution for #476.

mill
lefou: After a short web-search I found older/other shells could not compare an empty string properly, and some couldn't handle a `!` or `(` in `$1`. And someone said Solaris 10 still has issues.

lefou:
I created a dedicated project for the wrapper, as the wrapped should be in general decoupled from the current mill version and also already supports older mill versions.

 https://github.com/lefou/millw

PR #561, opened 2019-02-25
Add a shutdown target which allows user to kill background server

lbialy:
Never got any response in gitter so I just went through the codebase and found a place to put that `System.exit(1)` in :smiley: 

Rationale for this is: in a Jenkins pipeline (and also when working locally on several projects at once) one would like to be able to manage lifetimes of mill servers manually. For instance first Jenkins pipeline stage would implicitly spawn server, then other stages would reuse existing server and reap the benefit of running in warm jvm and after that in pipeline's finally block one would issue `mill shutdown` to prevent resource leaks.

Give me a shout if you see this functionality in any other way. It would help a lot in my attempt to migrate build to mill. Thanks!

### DIFF
--- a/main/src/main/MainModule.scala
+++ b/main/src/main/MainModule.scala
@@ -247,6 +247,14 @@ trait MainModule extends mill.Module{
     }
   }
 
+  /**
+    * Shuts down mill's background server
+    */
+  def shutdown() = mill.T.command {
+    T.ctx().log.info("Shutting down Mill server...")
+    System.exit(0)
+  }
+
   private type VizWorker = (LinkedBlockingQueue[(scala.Seq[_], scala.Seq[_], os.Path)],
     LinkedBlockingQueue[Result[scala.Seq[PathRef]]])
 


lbialy:
I guess I will probably have to edit docs, but I'll wait for your opinions first.

lihaoyi:
looks fine to me i guess

lefou:
> Never got any response in gitter so I just went through the codebase and found a place to put that `System.exit(1)` in smiley

As I already answered on gitter, PRs are welcome.

Before merge, I'd like to check how the server process communicates the shutdown to the client. Esp. in case of a `mill all target1 shutdown` invocation.

lefou:
Sorry for the delay. I now tried to shutdown the server.

What I found interesting is, that the mill process with `shutdown` target exits with `1`, whereas properly shutting down should result in a `0`  as there is no error at all.

lbialy:
Well, obviously :man_facepalming: I don't know what I was thinking. I'll fix that. Any other requirements before this can be merged? Documentation changes?

lihaoyi:
Merging due to inactivity

lefou:
To my understanding, that PR was still WIP. There's the issue with the wrong exit code, which @lbialy was going to fix.

PR #562, opened 2019-02-25
Introduced JavaModule.docOptions to add extra procesing options

lefou:
E.g. disabling picky JavaDoc linter with `-Xdoclint:none`.

The method is called `docOptions` instead of `javadocOptions` so that other modules like `ScalaModule` can reuse it.

### DIFF
--- a/scalalib/src/JavaModule.scala
+++ b/scalalib/src/JavaModule.scala
@@ -298,36 +298,41 @@ trait JavaModule extends mill.Module with TaskModule { outer =>
   }
 
   /**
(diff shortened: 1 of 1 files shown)


lihaoyi:
Seems like a reasonable idea. Note that we already have https://github.com/lihaoyi/mill/pull/433. Whether you want to consolidate them or not is up to you (I don't mind either way)

lefou:
Thanks for the pointer. In that case, I will rename to `javadocOptions` so that options for polyglot projects can co-exists.

lihaoyi:
lgtm

PR #564, opened 2019-03-02
Add section to documentation about ivy classifiers

nvander1:


lihaoyi-databricks:
I didn't know this syntax worked at all; can someone else help verify before I merge?

nvander1:
This is how I discovered the syntax:

https://github.com/coursier/coursier/blob/0b59a618ac4beeb50e64a7ab90a312ad92a4ea4b/modules/core/shared/src/main/scala/coursier/util/Parse.scala#L103

lefou:
I checked. This works. Thanks.

PR #565, opened 2019-03-05
Modify manual install to always fetch latest

nvander1:
The manual install snippet is changed to grab the latest version of mill by hitting Github's api.


lihaoyi-databricks:
Seems a bit too hacky v.s. the existing workflow of updating the readme for every tagged version, which we'll have to do anyway (e.g. to update changelogs)

nvander1:
That makes sense.

PR #569, opened 2019-03-08
Support new dotty version

molikto:


molikto:
@lefou the new dotty version bridge is in Java. I think they are in Scala before, so this code didn't work on new dotty version

PR #570, opened 2019-03-08
Implementation of --java-home flag.

markehammons:
With these changes, mill will now launch a background jvm based on the jvm it was launched with or based on the java home that was passed to it with the `-jh` or `--java-home` flag.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -8,6 +8,7 @@ import mill._
 import mill.scalalib._
 import publish._
 import mill.modules.Jvm.createAssembly
+
 trait MillPublishModule extends PublishModule{
 
   def artifactName = "mill-" + super.artifactName()
--- a/main/src/modules/Jvm.scala
+++ b/main/src/modules/Jvm.scala
@@ -6,7 +6,7 @@ import java.net.URI
 import java.nio.file.{FileSystems, Files, StandardOpenOption}
 import java.nio.file.attribute.PosixFilePermission
 import java.util.Collections
-import java.util.jar.{JarEntry, JarFile, JarOutputStream}
+import java.util.jar.{Attributes, JarEntry, JarFile, JarOutputStream, Manifest}
 
 import coursier.{Dependency, Fetch, Repository, Resolution}
 import coursier.util.{Gather, Task}
@@ -19,6 +19,7 @@ import mill.api.Loose.Agg
 
 import scala.collection.mutable
 import scala.collection.JavaConverters._
+import upickle.default.{macroRW, ReadWriter => RW}
 
 object Jvm {
   /**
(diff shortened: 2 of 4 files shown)


markehammons:
This has not been tested on windows yet.

markehammons:
I'm adding this functionality because of my project [here](https://github.com/markehammons/Wayland-McWayface_JVM-edition/tree/sans_kludge). I'd like to have mill compatibility for this project, but I have to use a nightly build of java 13 with project panama built in in order to build and run my app. Before this change, mill's worker processes would use the system jvm, and for very understandable reasons I don't want to make a nightly build of java 13 from [shipilev's build server](https://builds.shipilev.net/openjdk-panama/?C=M&O=D) my system jvm.

This change lets me use mill with this project without swapping out my system jvm by simply writing `mill -jh ~/bin/jdk/ wmcwf.compile`

build.sc
```diff
@@ -389,16 +389,31 @@ def launcherScript(shellJvmArgs: Seq[String],
...
-
-      s"""case "$$1" in
-         |  -i | --interactive )
-         |    ${java("mill.MillMain")}
-         |    ;;
-         |  *)
-         |    ${java("mill.main.client.MillClientMain")}
-         |    ;;
-         |esac""".stripMargin
+        s"""exec $$JVM $jvmArgsStr $$JAVA_OPTS -cp "${shellClassPath.mkString(":")}" $mainClass "$$@""""
```
lefou: This is, what I have in mind.

```suggestion
        s"""if [ -z "$JAVA_HOME" ] ; then
           |  JAVACMD="java"
           |else
           |  JAVACMD="$JAVA_HOME/bin/java"
           |fi
           |exec $$JAVACMD $jvmArgsStr $$JAVA_OPTS -cp "${shellClassPath.mkString(":")}" $mainClass "$$@"""".stripMargin
```

markehammons:
I wasn't thinking about making --java-home have to come first. I didn't think about changing -i so that it didn't short circuit argument evaluation.

I can make --java-home not required to be first. Adding support for JAVA_HOME would be good, but I still think the flags would be a good idea too.

lefou:
> I wasn't thinking about making --java-home have to come first. I didn't think about changing -i so that it didn't short circuit argument evaluation.

The current PR essentially starts mill, if the first parameter _is not_ `--java-home` or `-jh`.

> I can make --java-home not required to be first. Adding support for JAVA_HOME would be good, but I still think the flags would be a good idea too.

But that means, you have to more or less parse the complete command line in the script. That's possibly the reason, why `--interactive` must come as first parameter, too.

Anyway, I'd like to immediately merge the Java part of this PR, which implements the `java.home` sensitivity in the mill server selection. It's definitely a safety measure to avoid trouble. Do you mind to split your commit?

markehammons:
They've been split now.

lefou:
Thanks. I added that part to master, including proper JAVA_HOME handling.

This PR is now only about the new `--java-home` option.

PR #571, opened 2019-03-09
Add support for Scoverage

nvander1:
This pull request will add support generating test coverage reports in mill builds via scoverage.

I've created a `ScoverageModule` trait that I've used myself: see [this thread on scoverage](https://github.com/scoverage/scalac-scoverage-plugin/issues/243)

In order to make my changes compile, I've added an ivy dependency on scoverage into the scalalib module in build.sc. I don't think this is a good approach as that dependency should be resolved by a `ScoverageWorker` which is yet to be implemented.

I would like some guidance on how to offload this dependency to a worker instead of the main target.
Additionally, I still need to implement some tests for this.


scalalib/src/scoverage/ScoverageModule.scala
```diff
@@ -0,0 +1,48 @@
+package mill.lib.scoverage
```
lefou: I think, it would be better to add under `mill.contrib.scoverage`.

Documentation then goes to the `Contrib Module` pages.

scalalib/src/scoverage/ScoverageModule.scala
```diff
@@ -0,0 +1,48 @@
...
+    def scalaVersion = outer.scalaVersion()
+    def compileIvyDeps = outer.compileIvyDeps()
+    def ivyDeps = outer.ivyDeps() ++
+      Agg(ivy"org.scoverage::scalac-scoverage-runtime:1.3.1")
+    def scalacPluginIvyDeps = outer.scalacPluginIvyDeps() ++
+      Agg(ivy"org.scoverage::scalac-scoverage-plugin:1.3.1")
+    def scalacOptions = outer.scalacOptions() ++
+      Seq(s"-P:scoverage:dataDir:$dataDir")
+
+    def html() = T.command {
```
lefou: My first thought was, that `html` is a bit too general wording. Maybe something like `report` is more self-explanatory?

scalalib/src/scoverage/ScoverageModule.scala
```diff
@@ -0,0 +1,48 @@
...
+        val coverage = deserialize(coverageFileObj)
+        coverage(invoked(findMeasurementFiles(dataDir)))
+        val Seq(PathRef(sourceFolderPath, _, _)) = sources()
+        val sourceFolders = Seq(sourceFolderPath.toIO)
+        val htmlFolder = new java.io.File(s"${dataDir}/html")
+        htmlFolder.mkdir()
+        new ScoverageHtmlWriter(sourceFolders, htmlFolder, None)
+          .write(coverage)
+      } else {
+        T.ctx().log.error(s"Cannot write scoverage report. Directory ${dataDir} does not exist!")
```
lefou: That means what? What should I do, when I see this error message? Who is supposed to create the data dir for me? Does that mean, scoverage failed?

scalalib/src/scoverage/ScoverageModule.scala
```diff
@@ -0,0 +1,48 @@
...
+trait ScoverageModule { outer: ScalaModule =>
+
+  trait ScoverageCompile extends ScalaModule {
+    def dataDir: String
+    def sources = outer.sources
+    def resources = outer.resources
+    def scalaVersion = outer.scalaVersion()
+    def compileIvyDeps = outer.compileIvyDeps()
+    def ivyDeps = outer.ivyDeps() ++
+      Agg(ivy"org.scoverage::scalac-scoverage-runtime:1.3.1")
```
lefou: Besides the fact, that you want to move this part into the worker, the duplicate appearance of the hard coded version number indicates, that it is better moved into a dedicated target.

scalalib/src/scoverage/ScoverageModule.scala
nvander1: Re: package location
d41d1d51faa3259c1af5989dfe5a12c44e2bdb85

scalalib/src/scoverage/ScoverageModule.scala
lefou: With "it" I mean, it might be a good idea to add a `def scoverageVersion: T[String]` target and use that in the dependencies.

scalalib/src/scoverage/ScoverageModule.scala
nvander1: `dataDir` is where scoverage will output it's measurement files + also any reports it generates.

After reading more about the Task context api, I think it may be better to use the allocated `out/foo/scoverage/...` directory tree for a `object foo extends ScalaModule` to store the measurements and reports.

Is there a way to access a `out/foo` from within the body of the `ScalaModule` object? I know I can access `T.ctx().dest` from within a task/command, but this seems different.

scalalib/src/scoverage/ScoverageModule.scala
nvander1: Ah, I see. I think you're right.

scalalib/src/scoverage/ScoverageModule.scala
nvander1: I think `htmlReport` as this leaves room to implement the other report formats that scoverage has like `xmlReport` for example

scalalib/src/scoverage/ScoverageModule.scala
nvander1: Regarding the worker, after digging in the source tree here, it seems I can dig into scalanativelib, scalalib, and scalajslib to see how they are resolving their dependencies + injecting them into the classpath

scalalib/src/scoverage/ScoverageModule.scala
nvander1: I believe I can use anonymous tasks actually to handle this issue. I'll give it a go

nvander1:
Update:

I've now gotten mill to build a relase with ./ci/publish-local.sh.
There are some basic tests that pass.
Working on getting the release I've built to compile an example and generate reports.

nvander1:
@lefou What do you think of the changes + documentation added here?

contrib/scoverage/src/ScoverageModule.scala
```diff
@@ -0,0 +1,110 @@
...
+        .htmlReport(dataDir().toString, selfDir().toString)
+    }
+  }
+
+  trait ScoverageTests extends outer.Tests {
+    // This ensures that the runtime is added to the ivy deps
+    // specified by the user. Unfortunately it also clutters
+    // any calls to agg inside of ScoverageTests with this dependency,
+    // seems like a reasonable trade-off since it is only for test sources.
+    def Agg(items: Dep*) = T { Loose.Agg(outer.scoverageRuntimeDep()) ++ Loose.Agg(items: _*) }
```
lefou: Sorry, I'm not sure if I understand what is/should going on here. What if the user also uses `Loose.Agg` to define the deps?

docs/pages/9 - Contrib Modules.md
```diff
@@ -368,6 +368,48 @@ which will work with the following directory structure:
...
+
+In addition to the normal tasks available to your Scala module, Scoverage
+modules introduce a few new tasks and changes the behavior of an existing one.
+
+```
+mill foo.scoverage.compile      # compiles your module with test instrumentation
+                                # (you don't have to run this manually, running the test task will force its invocation)
+
+mill foo.test                   # tests your project and collects metrics on code coverage
+mill foo.scoverage.htmlReport   # uses the metrics collected by a previous test run to generate a coverage report in html format
```
lefou: Does the `htmlReport` target trigger the test run, or will it fail, if no test ran before?

contrib/scoverage/src/ScoverageModule.scala
```diff
@@ -12,52 +12,44 @@ import mill.moduledefs.Cacher
...
- *      def ivyDeps = Agg(ivy"com.lihaoyi::utest:0.6.0")
- *      def testFrameworks = Seq("utest.runner.Framework")
- *    }
- *  }
- *  }}}
+ * In addition to the normal tasks available to your Scala module, Scoverage
+ * Modules introduce a few new tasks and changes the behavior of an existing one.
  *
- *  You can compile the example like normally via
+ * <pre>
```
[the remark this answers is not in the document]
nvander1: I was trying to do a code block that wasn't highlighted as scala. Is there a better way to do so?

contrib/scoverage/src/ScoverageModule.scala
nvander1: Without this `def Agg()` If the user defines their own ivyDeps in test like so:

```scala
object foo extends ScoverageModule {
  def scoverageVersion = "1.3.1"
  def scalaVersion = "2.12.8"
  object test extends ScoverageTests {
    def ivyDeps = Agg(...)
   }
```

Then `test` will not have the necessary `org.scoverage::scalac-scoverage-runtime:1.3.1`. 


One option is to require the user to do the following:
```scala
object foo extends ScoverageModule {
  def scoverageVersion = "1.3.1"
  def scalaVersion = "2.12.8"
  object test extends ScoverageTests {
    def ivyDeps = super.ivyDeps() ++ Agg(...)
   }
```
I dislike this solution since it's not immediately clear to the user why they should have to specify this themselves.

With my current implementation (shadowing Loose.Agg with `def Agg()` in ScoverageTests),
the following code will always include the necessary scoverage runtime dependency:

```scala
object foo extends ScoverageModule {
  def scoverageVersion = "1.3.1"
  def scalaVersion = "2.12.8"
  object test extends ScoverageTests {
    def ivyDeps = Agg(...)  // this is calling `ScoverageTests.Agg`
   }
```

However, you're right in noting that if the user uses `Loose.Agg` themselves it won't use my method of the same name.

Do you have any thoughts on how to always include the `scoverage-runtime` dependency without requiring the user to specify it themselves?

docs/pages/9 - Contrib Modules.md
nvander1: It doesn't trigger the test run currently. But it will generate a report with all coverage reported as 0% if no tests have been run, otherwise it will use the measurement data from the most recent test run. I can force the invocation of the test run in `htmlReport` if you think that will be less surprising to a user.

lefou: Can't tell you what I like more. So we can leave it, as it gives more flexibility.

contrib/scoverage/src/ScoverageModule.scala
lefou: One idea is to hook the dependency later in, e.g. in `JavaModule.compileClasspath`. Though, when doing it there, also adding a way to disable that magic is at least worth a thought.

contrib/scoverage/src/ScoverageModule.scala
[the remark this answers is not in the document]
lefou: I wouldn't use a code block for normal text or a bullet list.

nvander1:
90% sure I handled the merge conflict in the docs appropriately.

contrib/scoverage/src/ScoverageModule.scala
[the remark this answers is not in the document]
nvander1: Noted: d265f3d2ba90b48b8c25a88d58db377faad61fcb

contrib/scoverage/src/ScoverageModule.scala
nvander1: That's much cleaner, you're right. This should implement that fix:  https://github.com/lihaoyi/mill/pull/571/commits/14030fb3763eddfe433c7a7781d35c9d61d121af

lefou:
> 90% sure I handled the merge conflict in the docs appropriately.

Almost. We try to keep the modules in alphabetical order.

Also, there are test failures:

* in CI_SCRIPT=ci/test-mill-bootstrap.sh
* in CI_SCRIPT=ci/test-mill-0.sh

```
[1/1] core.test.ivyDeps 
X mill.contrib.scoverage.HelloWorldTests.HelloWorld.core.test.ivyDeps 40ms 
  utest.AssertionError: #1: result == Agg(
  result: mill.api.Loose.Agg[mill.scalalib.Dep] = Agg(Dep(Dependency(org.scalatest:scalatest,3.0.5,d
  efault(compile),Set(),Attributes(,),false,true),Binary(false),false))
    utest.asserts.Asserts$.assertImpl(Asserts.scala:114)
    mill.contrib.scoverage.HelloWorldTests$.$anonfun$tests$26(HelloWorldTests.scala:84)
    mill.contrib.scoverage.HelloWorldTests$.$anonfun$tests$26$adapted(HelloWorldTests.scala:81)
    mill.contrib.scoverage.HelloWorldTests$.workspaceTest(HelloWorldTests.scala:35)
    mill.contrib.scoverage.HelloWorldTests$.$anonfun$tests$25(HelloWorldTests.scala:38)
1 targets failed
all 1 targets failed
contrib.scoverage.test.test mill.contrib.scoverage.HelloWorldTests mill.contrib.scoverage.HelloWorldTests.HelloWorld.core.test.ivyDeps
```

nvander1:
I'm not certain the cause of the remaining build failure. It's on `CI_SCRIPT=ci/test-mill-bootstrap.sh`. It looks like its a compile error on the JDK 9:

```

**Comment by nvander1** [comment]
I'm going to close and repon to trigger a build.

**Comment by Baccata** [comment]
I'm interested in this PR. Haven't dived really far to pinpoint the bug, but I can reproduce the java9 issue locally by running the failing CI script. 

Running that script (with JAVA_HOME pointing to a java9 JDK) results in a `mill-2` executable being created under your home. You can get the error by running placing yourself in the mill clone and running : 

```
 ~/mill-2 contrib.scoverage.test 
```

I might try to dive into it at some point but thought I'd share the info in the meantime.

File: contrib/scoverage/src/ScoverageModule.scala [hunk 8]
```diff
@@ -0,0 +1,115 @@
...
+    override def runClasspath = T {
+      super.runClasspath() ++
+      resolveDeps(T.task{Agg(outer.scoverageRuntimeDep())})()
+    }
+
+    // Need the sources compiled with scoverage instrumentation to run.
+    override def moduleDeps: Seq[JavaModule] = Seq(outer.scoverage)
+  }
+
+  def test: ScoverageTests
```
Baccata: Is reserving this name an absolute necessity ? As a user, I don't want to be forced to use this name for ScoverageTests

contrib/scoverage/src/ScoverageModule.scala
```diff
@@ -0,0 +1,115 @@
...
+ * - mill foo.scoverage.compile      # compiles your module with test instrumentation
+ *                                 # (you don't have to run this manually, running the test task will force its invocation)
+ *
+ * - mill foo.test                   # tests your project and collects metrics on code coverage
+ * - mill foo.scoverage.htmlReport   # uses the metrics collected by a previous test run to generate a coverage report in html format
+ *
+ * The measurement data is available at `out/foo/scoverage/data/`,
+ * And the html report is saved in `out/foo/scoverage/htmlReport/`.
+ */
+trait ScoverageModule extends ScalaModule { outer: ScalaModule =>
```
Baccata: Any reason for extending ScalaModule here ? Considering you have a self-type, extending `mill.define.Module` should be sufficient to define tasks

lefou:
Thanks for the insights, @Baccata . I haven't found the time to investigate into the build failure myself. Any help is much appreciated. This PR is fine to merge, as soon as we fix the test suite issues.

Baccata:
Oh well, I had no recollection of this : https://github.com/lihaoyi/mill/issues/302 ... 

Maybe we could condition the failing test with `!ammonite.util.Util.java9OrAbove`, as it is the case in many other tests in the codebase, and move on ? It'd be great to have the help of somebody who knows their stuff.

contrib/scoverage/src/ScoverageModule.scala
lefou: I'd rather keep the `extends` to make it easier to use `ScoverageModule`. I assume the motivation for the self-type it the named handle.

contrib/scoverage/src/ScoverageModule.scala
lefou: Yeah, good catch. Especially, if you want to keep "normal" tests and scoverage-enabled tests side-by-side.

contrib/scoverage/src/ScoverageModule.scala
Baccata: I think it might create problems when trying to make it work with scalajs modules (which scoverage supports), as having "extends ScalajsModule with ScoverageModule" could/would have unfortunate results because of inheritance

lefou: Oh, didn't know that. Though I don't see what the issue could be, as it has to be a `ScalaModule` either way.

lefou: Maybe, I misunderstood your request. Do you mean to just remove the extends, but keep the `outer: ScalaModule`, or do you suggest to remove both?

Baccata: The former : remove the extend, keep the self-type

nvander1: The extending was to make it easier to use. Since you can write this
```scala
object foo extends ScoverageModule
```
instead of
```scala
object foo extends ScalaModule with ScoverageModule
```
 This shouldn't cause issues with the current implementation:
Because ScalaJsModule already extends ScalaModule
```scala
object foo extends ScalaJsModule with ScoverageModule
```

It will just be a stylistic choice - one that I don't have a strong preference.
FWIW:
`PublishModule extends JavaModule` and its common to see
```scala
object baz extends ScalaModule with PublishModule
```
And ScalaModule already inherits from JavaModule.

nvander1: So it won't cause problems with inheritence, but if others prefer something else stylistically that's fine with me.

contrib/scoverage/src/ScoverageModule.scala
nvander1: https://github.com/lihaoyi/mill/pull/571/commits/dc966e282a0b7d57bde7e926aa99ef5d500c3f2f

Removed.

nvander1:
:crossed_fingers: Built the classloader using java.net.URLClassLoader to see if that will fix it.

contrib/scoverage/src/ScoverageModule.scala
Baccata: Right I'm in front of a computer now and had a deeper. Inheritance would be problematic if ScalaJsModule was overriding any of the methods implemented in ScalaModule, which doesn't seem to be the case (I wrongfully thought it was).

nvander1:
Ah nope it didn't fix it. Definitely need someone who has a better understanding of the classloader stuff.

Baccata:
@nvander1, I think it'd be okay to revert your latest commit and just condition the test so that it's only executed on java8 : there is a fair amount of tests that do the same thing in mill : 

* https://github.com/lihaoyi/mill/search?q=%21ammonite.util.Util.java9OrAbove&unscoped_q=%21ammonite.util.Util.java9OrAbove

* https://github.com/lihaoyi/mill/search?q=disableInJava9OrAbove&unscoped_q=disableInJava9OrAbove

nvander1:
So I found also errors in the actual htmlReport task. It was caused by an incorrect construction of the ScoverageReportWorker. It should be fixed now, but I had issues writing a unit for it, because to test it involves running the unit tests to get the coverage data. Running unit tests on the HelloWorld project inside of a unit test seems wonky because the directories aren't where I expect them to be.

nvander1:
It's actually still not ready to merge:
image

nvander1:
Got it. parrot
image

nvander1:
:green_heart:

lihaoyi:
Tests are green, and @lefou has given his blessing, so I'm merging this

nvander1:
Thanks for all the help on this one everyone!! :)

PR #573, opened 2019-03-17
Bump graphviz and JGraphT versions

andrelfpinto:


lihaoyi:
dunno looks ok I guess???

PR #579, opened 2019-03-28
GenIdea - handle projects not in the main build.sc file

ajrnz:
Projects which are imported from other files and used as module dependencies are not found by GenIdea and this causes an exception. This PR transitively adds all the modules to the lookup map.


lihaoyi:
no idea but ok

PR #585, opened 2019-04-04
Documentation around artifact name for publishing

kag0:
Added a line to the documentation on how to change the artifact id.  
Updated the scaladoc on JavaModule to explain the difference between artifactName and artifactId.


lefou:
Thanks

PR #587, opened 2019-04-08
Quick formatting fixes for Play module documentation

kag0:
The markdown parser for lihaoyi.com requires an empty line before lists.


lefou:
Thanks!

PR #588, opened 2019-04-09
Flyway module

kag0:
Module for running flyway migrations and operations. Similar to the old SBT flyway module.

TODO:
- [x] Documentation
- [x] clean and baseline commands
- [x] Some kind of tests?

I know adding a db dependency for not very significant tests might be a bit much, so if anyone objects to that we can just take it out.


.travis.yml
```diff
@@ -24,6 +24,10 @@ matrix:
   - stage: build
     env: CI_SCRIPT=ci/test-mill-0.sh
     jdk: oraclejdk8
+  - stage: build
+    env: CI_SCRIPT=ci/test-mill-0.5.sh
```
lefou: Could you rename that script file to something more expressive, e.g. `test-mill-postgresql.sh`. It would it make easier to grasp which script does what.

contrib/flyway/src/FlywayModule.scala
```diff
@@ -0,0 +1,72 @@
...
+  private def strToOptPair[A](key: String, v: String) =
+    Option(v)
+      .filter {
+        case a: String => a.nonEmpty
+        case _ => true
+      }
+      .map(key -> _)
+
+  private implicit val flywayWriter = writer[String].comap[Flyway](_.toString)
+  def flywayInstance = T.command {
```
lefou: This command looks like a good candidate for a worker. http://www.lihaoyi.com/mill/page/tasks.html#workers

lefou:
As for the integration tests, you could also use a H2 database, which is completely Java, and you could setup the test environment in the test.

kag0:
> Does Flyway support reading DB credentials from a separate file? Asking, because DB credentials in the buildfile have a high potential to end up under (public) version control.

It does, although this doesn't support that yet. I'm thinking users would read from files or environment variables in the tasks for the credentials. Perhaps a reminder in the docs about checking in credentials is in order.

docs/pages/9 - Contrib Modules.md
```diff
@@ -35,6 +35,43 @@ object project extends BuildInfo {
...
+  
+### Flyway
+
+Enables you to configure and run [Flyway](https://flywaydb.org/) commands from your mill build file.
+The flyway module currently supports the most common flyway use cases with file based migrations.
+
+Configure flyway by overriding settings in your module. For example
+
+```scala
+object foo extends JavaModule with FlywayModule {
```
lefou: Can you please add proper imports to the code snippet.

kag0: For sure, it should get published as `mill.contrib.flyway.FlywayModule` right?

lefou: Yes, that in line with the other contrib packages.

lefou:
Great, I'm happy to merge. Can you please rebase and resolve the conflicts.

PR #589, opened 2019-04-09
contrib BuildInfo plugin: fix generatedSources, it must be a folder

vitaliis:
generatedSources must be a folder,
but till now it was a reference to a generated file
This commit fixes this issue.

There are 2 reasons for this fix:
(1) Because generatedSources must be folders by the definition:
Folders containing source files that are generated
(2) If you use intellij IDEA and build your project from the IDE,
building module, which includes BuildInfo plugin, fails saying:

Error:scalac: Error: /home/vitalii/risk42/repo/risk42-workspace2/out/infinity/foundations/buildInfo/dest/BuildInfo.scala does not exist or is not a directory
scala.reflect.internal.FatalError: /home/vitalii/risk42/repo/risk42-workspace2/out/infinity/foundations/buildInfo/dest/BuildInfo.scala does not exist or is not a directory
	at scala.tools.nsc.settings.MutableSettings$OutputDirs.checkDir(MutableSettings.scala:286)
	at scala.tools.nsc.settings.MutableSettings$OutputDirs.add(MutableSettings.scala:276)
	at xsbt.CachedCompiler0.$anonfun$new$1(CompilerInterface.scala:68)
	at xsbt.CachedCompiler0.$anonfun$new$1$adapted(CompilerInterface.scala:66)
	at scala.collection.IndexedSeqOptimized.foreach(IndexedSeqOptimized.scala:36)
	at scala.collection.IndexedSeqOptimized.foreach$(IndexedSeqOptimized.scala:33)
	at scala.collection.mutable.ArrayOps$ofRef.foreach(ArrayOps.scala:198)
	at xsbt.CachedCompiler0.<init>(CompilerInterface.scala:66)
	at xsbt.CompilerInterface.newCompiler(CompilerInterface.scala:22)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    ...

### DIFF
--- a/contrib/buildinfo/src/BuildInfo.scala
+++ b/contrib/buildinfo/src/BuildInfo.scala
@@ -1,8 +1,8 @@
 package mill.contrib.buildinfo
 
 import mill.T
-import mill.define.Target
-import mill.api.{Ctx, Logger, PathRef}
+import mill.api.Logger
+import mill.api.PathRef
 import mill.scalalib.ScalaModule
 
 trait BuildInfo extends ScalaModule {
@@ -15,33 +15,35 @@ trait BuildInfo extends ScalaModule {
     Map.empty[String, String]
   }
 
-  private def generateBuildInfo(members: Map[String, Any])(implicit dest: Ctx.Dest, log: Ctx.Log): Seq[PathRef] =
-    if (!members.isEmpty) {
-      val outputFile = dest.dest / "BuildInfo.scala"
+  def generatedBuildInfo: T[Seq[PathRef]] = T {
+    val logger: Logger = T.ctx.log
+    val members: Map[String, String] = buildInfoMembers()
+    if (members.nonEmpty) {
+      val outputFile = T.ctx.dest / "BuildInfo.scala"
       val internalMembers =
         members
           .map {
             case (name, value) => s"""  def ${name} = "${value}""""
           }
           .mkString("\n")
-      log.log.debug(s"Generating object [${buildInfoPackageName.map(_ + ".").getOrElse("")}${buildInfoObjectName}] with [${members.size}] members to [${outputFile}]")
+      logger.debug(s"Generating object [${buildInfoPackageName.map(_ + ".").getOrElse("")}${buildInfoObjectName}] with [${members.size}] members to [${outputFile}]")
       os.write(
         outputFile,
-        s"""|${buildInfoPackageName.map(p => s"package ${p}").getOrElse("")}
-          |object ${buildInfoObjectName} {
-          |$internalMembers
-          |}""".stripMargin
+        s"""|${buildInfoPackageName.map(packageName => s"package ${packageName}\n").getOrElse("")}
+            |object ${buildInfoObjectName} {
+            |$internalMembers
+            |}""".stripMargin
       )
       Seq(PathRef(outputFile))
     } else {
-      log.log.debug("No build info member defined, skipping code generation")
+      logger.debug("No build info member defined, skipping code generation")
       Seq.empty[PathRef]
     }
-
-  def buildInfo = T {
-    generateBuildInfo(buildInfoMembers())
   }
 
-  override def generatedSources: Target[Seq[PathRef]] = T { super.generatedSources() ++ buildInfo() }
+  override def generatedSources = T {
+    super.generatedSources() ++
+      generatedBuildInfo().map(pathRef => PathRef(pathRef.path / os.up))
+  }
 
 }
--- a/contrib/buildinfo/src/BuildInfo.scala
+++ b/contrib/buildinfo/src/BuildInfo.scala
@@ -15,7 +15,7 @@ trait BuildInfo extends ScalaModule {
     Map.empty[String, String]
   }
 
-  def generatedBuildInfo: T[Seq[PathRef]] = T {
+  def generatedBuildInfo: T[(Seq[PathRef], PathRef)] = T {
     val logger: Logger = T.ctx().log
     val members: Map[String, String] = buildInfoMembers()
     if (members.nonEmpty) {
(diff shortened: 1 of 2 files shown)


contrib/buildinfo/test/src/BuildInfoTests.scala
```diff
@@ -104,7 +103,17 @@ object BuildInfoTests extends TestSuite {
...
+        )
+      }
+
+      "generatedSources must be a folder" - workspaceTest(BuildInfo) { eval =>
+        val generatedSourcesPath = eval.outPath / 'generatedBuildInfo / 'dest
+        val Right((result, evalCount)) = eval.apply(BuildInfo.generatedSources)
+        assert(
+          result.size == 1,
+          result.head.path == generatedSourcesPath
+        )
```
lefou: I fail to see, where this test checks, that the result is a folder.

contrib/buildinfo/src/BuildInfo.scala
```diff
@@ -15,33 +15,35 @@ trait BuildInfo extends ScalaModule {
...
     }
-
-  def buildInfo = T {
-    generateBuildInfo(buildInfoMembers())
   }
 
-  override def generatedSources: Target[Seq[PathRef]] = T { super.generatedSources() ++ buildInfo() }
+  override def generatedSources = T {
+    super.generatedSources() ++
+      generatedBuildInfo().map(pathRef => PathRef(pathRef.path / os.up))
```
lefou: Looks like "guessing" the best directory here. What if `generateBuildInfo` also generates sub-directories for each package?

contrib/buildinfo/src/BuildInfo.scala
```diff
@@ -15,33 +15,35 @@ trait BuildInfo extends ScalaModule {
     Map.empty[String, String]
   }
 
-  private def generateBuildInfo(members: Map[String, Any])(implicit dest: Ctx.Dest, log: Ctx.Log): Seq[PathRef] =
-    if (!members.isEmpty) {
-      val outputFile = dest.dest / "BuildInfo.scala"
+  def generatedBuildInfo: T[Seq[PathRef]] = T {
+    val logger: Logger = T.ctx.log
```
lefou: According to the documentation, it should be `T.ctx().log`.

contrib/buildinfo/test/src/BuildInfoTests.scala
```diff
@@ -104,7 +103,17 @@ object BuildInfoTests extends TestSuite {
         val Right((result, evalCount)) = eval.apply(BuildInfo.run(runResult.toString))
         assert(
           os.exists(runResult),
-          os.read(runResult) == scalaVersionString)
+          os.read(runResult) == scalaVersionString
+        )
+      }
+
+      "generatedSources must be a folder" - workspaceTest(BuildInfo) { eval =>
```
lefou: I fail to see, where this test checks, that the result is a folder.

contrib/buildinfo/src/BuildInfo.scala
vitaliis: done

contrib/buildinfo/src/BuildInfo.scala
vitaliis: Right, but I haven't found a better way to specify a folder. Do you have any ideas how to find and  "specify" the 'dest/ folder here?

contrib/buildinfo/test/src/BuildInfoTests.scala
vitaliis: Ok, it was implicitly done by checking against the `generatedSourcesPath`, which is a folder. But that's true, it can be done better ;-) I added an explicit check.

contrib/buildinfo/src/BuildInfo.scala
lefou: What about returning a `(PathRef, Seq[PathRef])` instead, the first is the directory, the second the generated resources. That way, when generating nothing, the dir is still present, but the seq might be empty.

lefou: By "returning" I refer to the `generateBuildInfo` target, of course.

vitaliis: Will change it when I'm back from work

vitaliis:
Noticed, sadly :anguished:, that the build is kind of flaky.

contrib/buildinfo/src/BuildInfo.scala
vitaliis: I was thinking, maybe one can use `myTask.ctx().dest` and indeed there is such a thing as `generatedBuildInfo.ctx` (notice the absense of brackets) of type `mill.define.NamedTask#ctx`. This property in turn is of type `mill.define.Ctx`, which contains a lot of different properties, but not `dest`. What do you think, does it make sense to add the dest property there? So that it can be used and referenced by others. Or is there a better place for this?

I will implement it for now as you suggested, by returning a tuple. But for the future this could be a solution as well. Because returning and destructing tuple is not the nicest thing, considering that the task should return only one value.

lefou: Imagine, someone is overriding `generateBuildInfo`, then of course, "dest" might be located somewhere else, so you must dynamically bind to the actual target or better, use the return value.

In fact, a tuple IS one return value. But if you prefer it you can use a case class instead (`case class Generated(dir: Path, files: Seq[Path])`).

vitaliis: I know, I tried it with case classes as well.

The only thing is that you have to wrap Path into a PathRef to have it serializable.

Latest changes already pushed.

lefou:
Thanks!

PR #590, opened 2019-04-10
[Fixes #442] Improve Scalafmt configuration options

### LINKED ISSUE #442: ScalafmtModule hardcodes scala version
https://github.com/lihaoyi/mill/blob/146d58bc5d98da1d142758f974d4d5de6f5948bf/scalalib/src/mill/scalalib/scalafmt/ScalafmtModule.scala#L27

I suppose we could have `ScalafmtModule` extend `ScalaModule` and use `scalaVersion()`?  

My use case is trying to use `def scalafmtVersion = "1.6.0-RC4"`, which requires scala `2.12.6`, as well as overriding `scalafmtDeps` as follows to include the reflect library:

```
  def scalafmtDeps: T[Agg[PathRef]] = T {
    Lib.resolveDependencies(
      zincWorker.repositories,
      Lib.depToDependency(_, scalaVersion()),
      Seq(ivy"com.geirsson::scalafmt-cli:${scalafmtVersion()}", ivy"org.scala-lang:scala-reflect:${scalaVersion()}")
    )
  }
```

I'd be happy to submit a PR if folks agree on the fix.

guilgaly:
See issue #442 .

This change notably makes it easier to use the latest Scalafmt 2.0 RC version (`2.0.0-RC5`), by:

- allowing to set the Scala version used for Scalafmt, and bumping up the default
- allowing to set the Scalafmt group ID, and trying to determine the correct one by default (it changed starting with version `2.0.0-RC2`)

**Notes:**

- Without this change, the alternative is to just override `scalafmtDeps` as a whole.
- In the future, I think it might be interesting to try to use `scalafmt-dynamic` to load a Scalafmt version specified in the `.scalafmt.conf` file (just like what Intellij does now, starting with Intellij 2019.1), but this is much simpler.

### DIFF
--- a/scalalib/src/scalafmt/ScalafmtModule.scala
+++ b/scalalib/src/scalafmt/ScalafmtModule.scala
@@ -18,13 +18,23 @@ trait ScalafmtModule extends JavaModule {
 
   def scalafmtVersion: T[String] = "1.5.1"
 
+  def scalafmtGroupId: T[String] = T {
+    val v = scalafmtVersion()
+    if (v.startsWith("0.") || v.startsWith("1.") || v == "2.0.0-RC1")
+      "com.geirsson"
+    else
+      "org.scalameta"
+  }
+
+  def scalafmtScalaVersion: T[String] = "2.12.8"
+
   def scalafmtConfig: Sources = T.sources(os.pwd / ".scalafmt.conf")
 
   def scalafmtDeps: T[Agg[PathRef]] = T {
     Lib.resolveDependencies(
       zincWorker.repositories,
-      Lib.depToDependency(_, "2.12.4"),
-      Seq(ivy"com.geirsson::scalafmt-cli:${scalafmtVersion()}")
+      Lib.depToDependency(_, scalafmtScalaVersion()),
+      Seq(ivy"${scalafmtGroupId()}::scalafmt-cli:${scalafmtVersion()}")
     )
   }
 
--- a/scalalib/test/src/scalafmt/ScalafmtTests.scala
+++ b/scalalib/test/src/scalafmt/ScalafmtTests.scala
@@ -11,11 +11,19 @@ object ScalafmtTests extends TestSuite {
   trait TestBase extends TestUtil.BaseModule {
     def millSourcePath =
       TestUtil.getSrcPathBase() / millOuterCtx.enclosing.split('.')
+    def core: ScalaModule with ScalafmtModule
   }
 
-  object ScalafmtTestModule extends TestBase {
+  object DefaultScalafmtTestModule extends TestBase {
     object core extends ScalaModule with ScalafmtModule {
       def scalaVersion = "2.12.4"
+    }
+  }
+
+  object Scalafmt2TestModule extends TestBase {
+    object core extends ScalaModule with ScalafmtModule {
+      def scalaVersion = "2.12.4"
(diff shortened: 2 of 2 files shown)


scalalib/src/scalafmt/ScalafmtModule.scala
```diff
@@ -18,13 +18,23 @@ trait ScalafmtModule extends JavaModule {
...
 
+  def scalafmtGroupId: T[String] = T {
+    val v = scalafmtVersion()
+    if (v.startsWith("0.") || v.startsWith("1.") || v == " 2.0.0-RC1")
+      "com.geirsson"
+    else
+      "org.scalameta"
+  }
+
+  def scalafmtScalaVersion: T[String] = "2.12.8"
```
lefou: Hardcoded versions have a high risk to blow up eventually. Wouldn't it be better to leave it abstract, and let the user decide which version to use? In most cases, the user module also implements `ScalaModule`, so a good default could be `ScalaModule.scalaVersion`. We could add a hint in the scaladoc.

guilgaly: Using `ScalaModule.scalaVersion` by default also seems problematic to me, because Scalafmt` is currently only published for Scala 2.11 and 2.12 but can be used to format code built (or cross-built) with Scala 2.10 or a 2.13 milestone/RC (or even with Dotty, I think). Also, probably a smaller problem, but there is no benefit to downloading Scalafmt and its dependencies twice (once for 2.11, once for 2.12) just because you have a project that cross-builds for Scala 2.11 and 2.12.

Just leaving this abstract avoids putting a brittle choice in mill itself, but I don't think it's very user friendly - I think the user will typically only care about choosing the Scalafmt version, not the Scala version used for running Scalafmt, so that's not very intuitive to have to configure that?

I think the better long term solution might be to use [scalafmt-dynamic](https://scalameta.org/scalafmt/docs/installation.html#standalone-library), as that now seems to be the recommended way to integrate with Scalafmt. I haven't looked into this in details so far, but from what I understand, if we do this, we will just use it like any library we depend on. For the Scalafmt version configuration, we could just document that people should set the `version` key in their `.scalafmt.conf` file (and scalafmt-dynamic will take care of resolving the correct Scalafmt version). Of course that implies adding another dependency to mill, and there may be other tradeoffs I haven't noticed.

lihaoyi: @olafurpg what do you think we should do here?

olafurpg: scalafmt-dynamic is the recommended library API and it handles stuff like
- respecting "version" setting in .scalafmt.conf
- picking the correct organization name for different version (we moved from com.geirsson to org.scalameta in v2)
- simple and documented API

See docs https://scalameta.org/scalafmt/docs/installation.html#standalone-library

Most importantly, editor integrations like IntelliJ and Metals already use scalafmt-dynamic so it would be great to do the same in Mill. This ensure the same scalafmt version is used in the editor and the build.

The dynamic module still has a few remaining issues, primarily around concurrency, but I'm optimistic those will get fixed sometime this summer.

guilgaly: @lihaoyi @lefou I've started to work on using scalafmt-dynamic, and it actually seems pretty easy to do. I can submit a pull request for that soon if you want.

PR #594, opened 2019-04-15
Revert Zinc to 1.2.5 to match release notes

ahirreddy:
Zinc was previously on on pre-release milestone (1.3.0-M1)

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -187,7 +187,7 @@ object scalalib extends MillModule {
 
     def ivyDeps = Agg(
       // Keep synchronized with zinc in Versions.scala
-      ivy"org.scala-sbt::zinc:1.3.0-M1"
+      ivy"org.scala-sbt::zinc:1.2.5"
     )
     def testArgs = T{Seq(
       "-DMILL_SCALA_WORKER=" + runClasspath().map(_.path).mkString(",")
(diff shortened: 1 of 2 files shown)


lefou:
@lihaoyi , do you have any reason against this? (You bumped the version.) I have not tried all corner cases, but tests suceed. I'm in favour of merging this one, as it also fixed the out-of-sync relation to `Versions.scala`.

lihaoyi:
@lefou this was an experiment to try and minimize some nondeterministic crashes in our Bazel build, which re-uses Mill's ZincWorkerImpl. Unfortunately, I don't think reverting the Zinc version solved it, so we can probably close this. @ahirreddy is that right?

lefou:
Hm, either we merge or we fix the Changelog entry and the value of `Versions.zinc` to the currently used zinc version.

vitaliis:
Had a question as well, why they are out of sync... :+1: for adjusting this.

lihaoyi:
The original motivation for trying to get on the 1.3.0-Mx series was to see if Zinc's new compile-to-jar functionality works, since it could save us time zipping stuff into jars.

Turns out it doesn't haha

lefou:
Ok then, I'll merge it.

lefou:
@ahirreddy Thank you.

PR #595, opened 2019-04-15
Beginning of Bloop integration 

Baccata:
This is the first of a series of PRs that aims at giving mill modules the ability to delegate some tasks to bloop. I have some more complex stuff lined up, but thought it'd be easier to split the work into smaller PRs. 

I initially had started the integration against the [bloop repository](https://github.com/scalacenter/bloop/blob/master/integrations/mill-bloop/src/main/scala/bloop/integrations/mill/MillBloop.scala), but mill's API evolving faster makes it hard to ensure that the integration remains up to date, and I believe this as its place here (happy to have the debate) 

This first PR does the following: 

* ports the work currently present in the bloop repo, allowing for the generation of the bloop config from mill, in a way fairly similar to how intellij config is currently generated.  
* adds a `BloopModule` trait that can be mixed in a mill's JavaModule, giving the ability to call `mill module.bloop.config`  ~~and`mill module.bloop.compile`~~. (edit : compiling with bloop will occur in a subsequent PR)
* updates the versions of the libraries involved to their latest (ujson-circe)  

The aim of this first PR is to lay the basis for the integration. It does not address scalajs/scalanative specific usecases. In subsequent PRs, I'm planning to use a mill worker to maintain a bsp connection between mill and bloop to allow the implementation of more complex tasks (testing in particular, the results of which cannot be easily retrieved via mere shellout). 

The `BloopModule` trait does not extend `ScalaModule` (nor `JavaModule`). This is to build on the existing modules and let the users decide whether the want to override their `compile` tasks by calling `bloop.compile`, or whether they want to retain the freedom to call either tasks. 

~~I haven't added unit-tests yet, and would appreciate some guidance around testing : should I just follow what's been done around the GenIdea module ?~~ Unit tests added

### DIFF
--- a/contrib/bloop/src/mill.contrib.bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill.contrib.bloop/BloopImpl.scala
@@ -76,11 +76,41 @@ class BloopImpl(ev: () => Evaluator, wd: Path) extends ExternalModule {
     sources.toMap
   }
 
-  protected[bloop] def name(m: JavaModule) = m.millModuleSegments.render
+  protected def name(m: JavaModule) = m.millModuleSegments.render
 
-  /**
-    * Computes the bloop configuration for a mill module.
-    */
+  //////////////////////////////////////////////////////////////////////////////
+  // SemanticDB related configuration
+  //////////////////////////////////////////////////////////////////////////////
+
+  // Version of the semanticDB plugin.
+  def semanticDBVersion: String = "4.1.0"
+
+  // Scala versions supported by semantic db. Needs to be updated when
+  // bumping semanticDBVersion.
+  // See [https://github.com/scalameta/metals/blob/333ab6fc00fb3542bcabd0dac51b91b72798768a/build.sbt#L121]
+  def semanticDBSupported = Set(
+    "2.12.8",
+    "2.12.7",
+    "2.12.6",
+    "2.12.5",
+    "2.12.4",
+    "2.11.12",
+    "2.11.11",
+    "2.11.10",
+    "2.11.9"
+  )
+
+  // Recommended for metals usage.
+  def semanticDBOptions = List(
+    s"-P:semanticdb:sourceroot:$pwd",
+    "-P:semanticdb:synthetics:on",
+    "-P:semanticdb:failures:warning"
+  )
+
+  //////////////////////////////////////////////////////////////////////////////
+  // Computation of the bloop configuration for a specific module
+  //////////////////////////////////////////////////////////////////////////////
+
   def bloopConfig(bloopDir: Path,
                   module: JavaModule): Task[(BloopConfig.File, Path)] = {
     import _root_.bloop.config.Config
@@ -96,20 +126,20 @@ class BloopImpl(ev: () => Evaluator, wd: Path) extends ExternalModule {
 
     val scalaConfig = module match {
       case s: ScalaModule =>
+        val semanticDb = s.resolveDeps(s.scalaVersion.map {
+          case scalaV if semanticDBSupported(scalaV) =>
+            Agg(ivy"org.scalameta:semanticdb-scalac_$scalaV:$semanticDBVersion")
+          case _ => Agg()
+        })
+
         T.task {
-          val pluginOptions = s.scalacPluginClasspath().map { pathRef =>
+          val pluginCp = semanticDb() ++ s.scalacPluginClasspath()
+          val pluginOptions = pluginCp.map { pathRef =>
             s"-Xplugin:${pathRef.path}"
           }
 
-          // Recommended for metals usage.
-          val semanticDbOptions = List(
-            s"-P:semanticdb:sourceroot:$pwd",
-            "-P:semanticdb:synthetics:on",
-            "-P:semanticdb:failures:warning"
-          )
           val allScalacOptions =
-            (s.scalacOptions() ++ pluginOptions ++ semanticDbOptions).toList
-
+            (s.scalacOptions() ++ pluginOptions ++ semanticDBOptions).toList
           Some(
             BloopConfig.Scala(
               organization = "org.scala-lang",
--- a/contrib/bloop/src/mill.contrib.bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill.contrib.bloop/BloopImpl.scala
@@ -5,38 +5,55 @@ import bloop.config.ConfigEncoderDecoders._
 import bloop.config.{Config => BloopConfig}
 import mill._
 import mill.api.Loose
-import mill.define._
+import mill.define.{Module => MillModule, _}
 import mill.eval.Evaluator
 import mill.scalalib._
+import os.pwd
 
-trait BloopModule extends Module with CirceCompat { self: JavaModule =>
+/**
+  * Implementation of the Bloop related tasks. Inherited by the
+  * `mill.contrib.Bloop` object, and usable in tests by passing
+  * a custom evaluator.
+  */
+class BloopImpl(ev: () => Evaluator, wd: Path) extends ExternalModule {
 
-  object bloop extends Module {
-
-    def config = T {
-      // Forcing the generation of the config for all modules here
-      val _ = BloopModule.install() // NB: looks like it has to be assigned to be executed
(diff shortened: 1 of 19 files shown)


Baccata:
Pinging @jvican for obvious reasons 
@robby-phd this might be of interest to you as well

Baccata:
@lefou sure, will do 👍. Could you also provide guidelines regarding testing ? What tests would you expect to see in this PR ? Should I follow the GenIdea tests ?

lefou:
For testing, we need to cover the file generator (which is like the `GenIdeaTests` test) and the compilation with bloop (like `HelloWorldTests` or `HelloJavaTests`). We should avoid testing bloop itself, but should ensure we cover communication and error handling.

lefou:
If you need further help with initial test setup, have a look at other contrib project or ask on gitter.

Baccata:
> and the compilation with bloop (like HelloWorldTests or HelloJavaTests 

Gotcha. In the interest of keeping PRs small(ish), I'll remove the compilation task and will add it to a subsequent PR with the relevant tests. Thanks for the pointers.

main/test/src/util/TestEvaluator.scala
```diff
@@ -12,12 +12,12 @@ object TestEvaluator{
   val externalOutPath = os.pwd / 'target / 'external
 
 
-  def static(module: TestUtil.BaseModule)(implicit fullName: sourcecode.FullName) = {
+  def static(module: => TestUtil.BaseModule)(implicit fullName: sourcecode.FullName) = {
```
Baccata: This is needed to accommodate a cyclic reference  (`BloopImpl` requires an evaluator contains a path-dependant trait )

Baccata:
@lefou, I'm happy for this PR to be reviewed

contrib/bloop/src/mill.contrib.bloop/BloopImpl.scala
```diff
@@ -0,0 +1,339 @@
...
+    * }
+    * }}}
+    */
+  trait Module extends MillModule with CirceCompat { self: JavaModule =>
+
+    object bloop extends MillModule {
+
+      def config = T {
+        // Forcing the generation of the config for all modules here
+        val installed: Seq[(String, Path)] = install()
```
Baccata: This call the reason for the path-dependant trait. Refencing the global `mill.contrib.Bloop` module would disallow overrides and make testing fairly impossible.

contrib/bloop/src/mill.contrib.bloop/BloopImpl.scala
```diff
@@ -0,0 +1,339 @@
...
+    val scalaLibraryIvyDeps = module match {
+      case x: ScalaModule => x.scalaLibraryIvyDeps
+      case _              => T.task { Loose.Agg.empty[Dep] }
+    }
+
+    /**
+      * Resolves artifacts using coursier and creates the corresponding
+      * bloop config.
+      */
+    def artifacts(repos: Seq[coursier.Repository],
```
Baccata: Calling coursier directly as the logic needs to retain the `dep => path` relation

docs/pages/9 - Contrib Modules.md
```diff
@@ -912,3 +912,62 @@ Publishing to custom local Maven repository
...
+This plugin generates [bloop](https://scalacenter.github.io/bloop/) configuration 
+from your build file, which lets you use the bloop CLI for compiling, and makes 
+your scala code editable in [Metals](https://scalameta.org/metals/)   
+
+
+#### Quickstart:
+```scala
+// build.sc (or any other .sc file it depends on, including predef) 
+// Don't forget to replace VERSION
+import $ivy.`com.lihaoyi::mill-contrib-bloop:VERSION`
```
olafurpg: Is it possible to programmatically run this setup? For example `mill --extra-predef /tmp/metals.sc mill.contrib.Bloop/install`. It would be nice if we could support mill in the "import build" command in Metals, the same way it works for sbt.

Baccata: Just had a go at it : ` mill --predef /tmp/metals.sc mill.contrib.Bloop/install` would work

contrib/bloop/src/mill.contrib.bloop/BloopImpl.scala
```diff
@@ -0,0 +1,339 @@
...
+  }
+
+  protected def name(m: JavaModule) = m.millModuleSegments.render
+
+  //////////////////////////////////////////////////////////////////////////////
+  // SemanticDB related configuration
+  //////////////////////////////////////////////////////////////////////////////
+
+  // Version of the semanticDB plugin.
+  def semanticDBVersion: String = "4.1.0"
```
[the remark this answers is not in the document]
Baccata: Bumped: a6bbfe8

docs/pages/9 - Contrib Modules.md
olafurpg: Cool, we're working on adding automatic "import build" for Gradle in Metals, which requires abstracting some sbt-specifics away. I'm optimistic it won't be too much effort to add automatic "import build" for Mill after that.

robby-phd:
@Baccata Similar to the summary on Metals support, it would be great if the documentation also summarizes IntelliJ integration (as an alternative to mill's GenIdea) with regards to importing mill project into IntelliJ, compilation on change via Bloop, running/debugging apps/tests, etc., and perhaps refer to the documentation on IntelliJ's support for BSP/Bloop for more information.

Baccata:
> Similar to the summary on Metals support, it would be great if the documentation also summarizes IntelliJ integration (as an alternative to mill's GenIdea) with regards to importing mill project into IntelliJ, compilation on change via Bloop, running/debugging apps/tests, etc., and perhaps refer to the documentation on IntelliJ's support for BSP/Bloop for more information.

In all frankness, I'm not planning on testing any of that myself... at least not before the Intellij-Bloop integration gets out of EAP and reaches mainstream releases. I'd rather someone who's actively interested in those things try them out and see if they work, and amend the documentation after doing so.

robby-phd:
Hmm... I'm currently using IntelliJ 2019.1.1 with IntelliJ Scala 2019.1.7, and I can see that it has BSP and Bloop in the build tool configuration.  I think those are mainstream releases; anyway, just a suggestion.

contrib/bloop/src/mill.contrib.bloop/BloopImpl.scala
```diff
@@ -0,0 +1,356 @@
...
+      val rootModule = eval.rootModule
+      rootModule.millInternal.segmentsToModules.values.collect {
+        case m: scalalib.JavaModule => m
+      }.toSeq
+    } else Seq()
+  }
+
+  /**
+    * Computes sources files paths for the whole project. Cached in a way
+    * that doesnot get invalidated upon sourcefile change. Mainly called
```
lefou:
```suggestion
    * that does not get invalidated upon sourcefile change. Mainly called
```

contrib/bloop/src/mill.contrib.bloop/BloopImpl.scala
```diff
@@ -0,0 +1,356 @@
...
+
+      def writeTransitiveConfig = T {
+        Task.traverse(jm.transitiveModuleDeps)(_.bloop.writeConfig)
+      }
+    }
+  }
+
+  private val bloopDir = wd / ".bloop"
+
+  private def computeModules = {
```
lefou: Explicit return type would be nice.

contrib/bloop/src/mill.contrib.bloop/BloopImpl.scala
Baccata: 049f5ca

contrib/bloop/src/mill.contrib.bloop/BloopImpl.scala
Baccata: 049f5ca

lefou:
Great, I'm going to merge. @Baccata What do you think about your commit history? Does keeping all commits make sense, or should I squash?

Baccata:
I won't be offended if you squash them, most of the commits are clean but I did mess a couple of them

PR #600, opened 2019-04-21
[Fixes #442] Switch from scalafmt-cli to scalafmt-dynamic

### LINKED ISSUE #442: ScalafmtModule hardcodes scala version

guilgaly:
This could replace pull request #590.

I completely removed the `scalafmtVersion` key because it can now be configured directly in the `.scalafmt.conf` configuration file. If a version isn't provided in the configuration file, the fallback is the default value provided by the `scalafmt-dynamic` library (which is the same as the version of the library). Alternatively, we could keep the `scalafmtVersion` key to allow specifying a different fallback version, but then we should probably make it optional to avoid hardcoding our own default when there is already one in `scalafmt-dynamic`?

I have added an empty configuration file in the resources,  used if the project doesn't include one, because `scalafmt-dynamic` requires providing an actual config file even though Scalafmt has a reasonable set of defaults.

Finally, by default, `scalafmt-dynamic` writes messages to `System.err` but we can implement our own `ScalafmtReporter` to handle printing messages ourselves, if there is a reason to prefer that.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -143,7 +143,8 @@ object scalalib extends MillModule {
   def moduleDeps = Seq(main, scalalib.api)
 
   def ivyDeps = Agg(
-    ivy"org.scala-sbt:test-interface:1.0"
+    ivy"org.scala-sbt:test-interface:1.0",
+    ivy"org.scalameta::scalafmt-dynamic:2.0.0-RC6"
   )
 
   def genTask(m: ScalaModule) = T.task{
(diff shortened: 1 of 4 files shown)


lihaoyi:
This looks good to me

PR #612, opened 2019-05-22
Fixed bloop config classpath element when no ivyDep is set

Baccata:
Realised that the classpath wasn't being set when no ivyDep was present 😅 

Also fixed modules not re-installing after change to build.sc 

- [x] /!\ Don't merge just yet, I need to double check some things

### DIFF
--- a/contrib/bloop/src/mill.contrib.bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill.contrib.bloop/BloopImpl.scala
@@ -17,13 +17,21 @@ import os.pwd
   */
 class BloopImpl(ev: () => Evaluator, wd: Path) extends ExternalModule { outer =>
 
+  private val bloopDir = wd / ".bloop"
+
   /**
     * Generates bloop configuration files reflecting the build,
     * under pwd/.bloop.
     */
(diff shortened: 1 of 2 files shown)


contrib/bloop/src/mill.contrib.bloop/BloopImpl.scala
```diff
@@ -92,13 +92,13 @@ class BloopImpl(ev: () => Evaluator, wd: Path) extends ExternalModule { outer =>
     * that does not get invalidated upon sourcefile change. Mainly called
     * from module#sources in bloopInstall
     */
-  def moduleSourceMap: Target[Map[String, Seq[Path]]] = T {
+  def moduleSourceMap = T.input {
```
Baccata: I was thinking that the cache would be invalidated upon the build changing, but it appear to not be the case for global tasks. Making this an input to ensure that `install` picks up new modules every time the user adds one.

Baccata:
@lefou I'm happy with the current state, deleting the obsolete config files rather after the new ones are generated proves a lot better than deleting all, then generating.

lefou:
@Baccata Thank you!

Issue #613, opened 2019-05-23
Scoverage contrib module not running unit tests and throwing an exception


nh13:
I am trying to use the new [scoverage contrib module](http://www.lihaoyi.com/mill/page/contrib-modules.html#scoverage) (#571) and have encountered two issues.  @nvander1.  I am completely open (and expecting) I have done something wrong.

Repo: https://github.com/fulcrumgenomics/commons
Commit: https://github.com/fulcrumgenomics/commons/commit/23ae8deab907f66bf7ec30e391feb2ce47b667dc

### Unit tests don't run
```
mill commons.test
```

Log output:
<details>

```
[41/71] commons.scoverage.compile 
Compiling compiler interface...
warning: there were three deprecation warnings (since 2.12.0); re-run with -deprecation for details
warning: there were three feature warnings; re-run with -feature for details
two warnings found
[info] Compiling 17 Scala sources and 2 Java sources to /Users/nhomer/git/fg/commons/out/commons/scoverage/compile/dest/classes ...
[info] [info] Cleaning datadir [/Users/nhomer/git/fg/commons/out/commons/scoverage/data]
[info] [info] Beginning coverage instrumentation
[info] [info] Instrumentation completed [1377 statements]
[info] [info] Wrote instrumentation file [/Users/nhomer/git/fg/commons/out/commons/scoverage/data/scoverage.coverage.xml]
[info] [info] Will write measurement data to [/Users/nhomer/git/fg/commons/out/commons/scoverage/data]
[warn] there were three deprecation warnings (since 2.12.0); re-run with -deprecation for details
[warn] there were two unchecked warnings; re-run with -unchecked for details
[warn] there were three feature warnings; re-run with -feature for details
[warn] three warnings found
[info] Done compiling.
[71/71] commons.test.test 
```

</details>

I don't see any unit tests run.

### Exception creating the coverage report

```
mill commons.scoverage.htmlReport
```

Log output:
<details>

```
[11/11] commons.scoverage.htmlReport 
1 targets failed
commons.scoverage.htmlReport scala.MatchError: Vector(PathRef(/Users/nhomer/git/fg/commons/src/main/scala,false,-898194901), PathRef(/Users/nhomer/git/fg/commons/src/main/java,false,-1272352833)) (of class scala.collection.immutable.Vector)
    mill.contrib.scoverage.worker.ScoverageReportWorkerImpl.htmlReport(ScoverageReportWorkerImpl.scala:14)
[...]

PR #615, opened 2019-05-24
Scoverage multiple source folders

nvander1:
Previously, the scoverage report worker would fail on a match error if more than one source directory existed, as found in #613 by @nh13.


nvander1:
@lefou

lihaoyi:
looks good to me i guess

lefou:
@nvander1 I was busy last week. Yes, looks good to me. Thanks.

PR #618, opened 2019-05-27
Scoverage can output an XML report

nh13:
* use the _.scoverage.xmlReport task

@nvander1 and @lefou thank-you again for the solution in #615.  

This adds the task to output an XML report, which is required by [codecov for upload](https://docs.codecov.io/reference#section-acceptable-report-formats).  This will work with only a single module, whereas if we have multiple modules in a repo, we will need to add another PR to utilize `scoverage.report.CoverageAggregator`.  I may get to that when I need it, but for now, this is a good start.


lefou:
Thanks!

PR #619, opened 2019-05-28
Added tasks allowing to delegate compile/run to bloop.

Baccata:
@lefou opening this early as I'd like to discuss how testing should be done. So far the PRs were merely writing configuration for bloop to pick up, and didn't involve running bloop in any way. 

This PR adds the necessary code to give the user the option to have `mill foo.compile` call `bloop foo compile`, thus re-using the results of the compilation triggered by their editor when they use metals. This also makes the user benefit from parallel compilation, as in a situation where a user would call `mill foo.compile` when `foo` depends on two unrelated `bar` and `baz` modules, bloop would compile `bar` and `baz` in parallel. 

The thing I'm not sure about is how to test this in continuous-integration, as it requires another tool to be installed on the machine. The "easiest" option would be to install `bloop` in the CI setup step, but I'm worried it be intrusive and would create an undesirable precedent. A harder option would be to call upon the "bloop launcher" which would dynamically retrieve bloop and start it, possibly from the test. 

What do you think ?

### DIFF
--- a/contrib/bloop/src/mill.contrib.bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill.contrib.bloop/BloopImpl.scala
@@ -7,9 +7,11 @@ import mill._
 import mill.api.Loose
 import mill.define.{Module => MillModule, _}
 import mill.eval.Evaluator
+import mill.modules.Jvm
 import mill.scalajslib.ScalaJSModule
 import mill.scalajslib.api.ModuleKind
 import mill.scalalib._
+import mill.scalalib.api.CompilationResult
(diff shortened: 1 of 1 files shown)


lefou:
To the test setup: I guess either way would be OK to increase the coverage. Having some control over the bloop instance in the test process itself is of course a nice feature, as it allows not event the CI but also each developer to run bloop integration tests, even if she has not or a different bloop version installed locally. But it would make the setup more complex.
I'd say, who codes, decides. So, pick what you want. We can still improve later.

Baccata:
Closing for now, will re-open when I find time to finish it.

PR #623, opened 2019-05-31
Inherit moduleDeps from outer in Scoverage

### LINKED ISSUE #622: Scoverage can't compile modules with moduleDeps
`mill amodule.scoverage.compile` will not work for any `amodule` that has defined `moduleDeps`. I suspect the [`scoverage`](https://github.com/lihaoyi/mill/blob/master/contrib/scoverage/src/ScoverageModule.scala#L86) module needs to `def moduleDeps = outer.moduleDeps`.

build file to reproduce: 
(also need to create a class in foo and import it in the other project)
```scala
// build.sc

import $ivy.`com.lihaoyi::mill-contrib-buildinfo:0.4.0`
import mill.contrib.scoverage.ScoverageModule

  def scalaVersion = "2.12.7"
}

object scoverageSadness extends ScalaModule with ScoverageModule {
  def scoverageVersion = "1.3.1"
  def scalaVersion = "2.12.7"

  def moduleDeps = Seq(foo)

  object test extends ScoverageTests {
    def ivyDeps = Agg(ivy"org.scalatest::scalatest:3.0.5")
    def testFrameworks = Seq("org.scalatest.tools.Framework")
  }
}
```

run `scoverageSadness.compile` to confirm it works, then run `scoverageSadness.scoverage.compile` to confirm not found errors on imports.

Does that diagnosis sound right @nvander1?

PS: thanks for the scoverage PR, very handy.

nvander1:
Thanks to @kag0 for finding this bug! Fixes #622. @lefou @lihaoyi

### DIFF
--- a/contrib/scoverage/src/ScoverageModule.scala
+++ b/contrib/scoverage/src/ScoverageModule.scala
@@ -87,6 +87,7 @@ trait ScoverageModule extends ScalaModule { outer: ScalaModule =>
     def selfDir = T { T.ctx().dest / os.up / os.up }
     def dataDir = T { selfDir() / "data" }
 
+    def moduleDeps = outer.moduleDeps
     def sources = outer.sources
     def resources = outer.resources
     def scalaVersion = outer.scalaVersion()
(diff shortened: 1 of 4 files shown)


lihaoyi:
no idea

PR #632, opened 2019-06-16
Add all JAR version attributes to JavaModule

clintval:
For one of my projects, I need to set the package version information in my JAR file's manifest. Would you be open to enabling that feature within Mill?

This is my first attempt at contributing to Mill so please let me know how I can improve this PR!

See this link for more information on Jar Package Version Information:

- https://docs.oracle.com/javase/tutorial/deployment/jar/packageman.html

Thank you for your time, and also for providing such an excellent build tool.

How this can be used:

```scala
import mill._, scalalib._
import mill.modules.Jvm.PackageVersionInfo

object commons extends ScalaModule {
  def scalaVersion = "2.12.4"
  
  override def packageVersionInfo = new PackageVersionInfo(
    specificationTitle = Some("Scala Utility Classes"),
    specificationVersion = Some("1.2"),
    specificationVendor = Some("Example Tech, Inc."),
    implementationTitle = Some("scala.util"),
    implementationVersion = Some("build57"),
    implementationVendor = Some("Example Tech, Inc.")
   )
}
```

### DIFF
--- a/main/src/modules/Jvm.scala
+++ b/main/src/modules/Jvm.scala
@@ -16,6 +16,7 @@ import mill.eval.{PathRef, Result}
 import mill.util.Ctx
 import mill.api.IO
 import mill.api.Loose.Agg
+import upickle.default.{ReadWriter => RW, macroRW}
 
 import scala.collection.mutable
 import scala.collection.JavaConverters._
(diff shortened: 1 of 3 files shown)


nh13:
This looks really nice, but a JAR manifest can have custom attributes (header: value). How does this support these, or allow them for the future? Sbt for example does, see “Add manifest attributes”: https://www.scala-sbt.org/1.x/docs/Howto-Package.html 

Should these attributes be inferred from other attributes, or be defs in Module? They seem quite important and may be broadly used. 

Also, you might want to look at the POM if you publish with mill and see if anything is missing versus your sbt build. That’s for another PR/issue though.

clintval:
@nh13, you raise a great point I had not considered.

The solution I implemented does not provide a method for setting arbitrary headers in the JAR Manifest, but it also does not exclude the ability do implement such a feature in the future (besides setting a precedent that commonly used Manifest headers are grouped by case class).

I see two paths:

1. Group the commonly used [Package Version Information](https://docs.oracle.com/javase/tutorial/deployment/jar/packageman.html) headers in a case class like in this PR, and then in the future, have a method for adding other *custom* Manifest headers via `Map`, or similar, such as how `sbt` implements this.
2. Have all Manifest headers (except `Main-Class`) be set via `Map` or similar.

I am open to feedback and want to see an implementation that does not restrict the API going forward!

lefou:
I like a generic solution (e.g. a string map) most, as it is simple and maps best to the concept of a manifest. I could also imagine some kind of view that gives access and helps setting commonly used manifest headers, but which operates on the underlying map. It could also help with encoding of list values.

Please note that in more specific contexts like OSGi, where we want to deal with a predefined set of manifest entries, a dedicated case class might be a good idea (and it fact i think it is: https://github.com/lefou/mill-osgi/blob/042b61b6fa5cd3777eac2c1694810e72213a6e15/core/src/de/tobiasroeser/mill/osgi/OsgiHeaders.scala). But as we are still on a general purpose level, the chances are high, that we always miss someones most essential headers.

lefou:
@clintval Oh, and yes, we are glad you opened a PR. Welcome to mill!

ajrnz:
BTW: I have an implementation of a fairly generic Jar manifest which lets you add entries to the manifest main and grouped entries. You just override `manifest` in your `ScalaModule`/`JavaModule` to add entries.

Will try to push a WIP PR in the next couple of days

clintval:
Thanks @lefou, I appreciate it!

@ajrnz, excited to see, I'll hold off on making progress on this feature to see what you have!

ajrnz:
Ok I've pushed PR #634 to let you override/modify a manifest. Let me know what you think

lefou:
@ajrnz Nice, I like #634 .

@clintval Do you think #634 fulfills your needs, or is there something missing?

clintval:
@lefou, #634 definitely fulfills my needs! Thank you so much! I'll close this draft PR in favor of the new implementation.

PR #634, opened 2019-06-18
add ability to define jar manifest

ajrnz:
Add a `manifest` target to `JavaModule` (and `JarManifest` object). To let the user set/modify the manifest.

Usage:

```scala
  def manifest = T{
    super.manifest().add(
      "Manifest-Version" -> "1.0",
      "Rundeck-Plugin-File-Version" -> pluginVersion(),
      "Rundeck-Plugin-Classnames" -> pluginClassNames().mkString(","),
      "Rundeck-Plugin-Name" -> "Rundeck Telegram Plugin"
    )
  }
```

or to override:

```scala
  def manifest = T{
    JarManifest().add(
      "Manifest-Version" -> "1.0",
      "Rundeck-Plugin-File-Version" -> pluginVersion(),
      "Rundeck-Plugin-Classnames" -> pluginClassNames().mkString(","),
      "Rundeck-Plugin-Name" -> "Rundeck Telegram Plugin"
    )
  }
```

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -8,6 +8,8 @@ import mill._
 import mill.scalalib._
 import publish._
 import mill.modules.Jvm.createAssembly
+import mill.modules.Assembly
+
 trait MillPublishModule extends PublishModule{
 
   def artifactName = "mill-" + super.artifactName()
@@ -605,12 +607,15 @@ def assembly = T{
   os.move(
     createAssembly(
       devRunClasspath,
+      None,
       prependShellScript = launcherScript(
         shellArgs,
         cmdArgs,
         Agg("$0"),
         Agg("%~dpnx0")
-      )
+      ),
+      None,
+      Assembly.defaultRules
     ).path,
     T.ctx().dest / filename
   )
--- a/build.sc
+++ b/build.sc
@@ -8,7 +8,6 @@ import mill._
 import mill.scalalib._
 import publish._
 import mill.modules.Jvm.createAssembly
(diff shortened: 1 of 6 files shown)


lefou:
I restarted all travis jobs, let's see if that already solves the failures.

ajrnz:
The error looked like some kind of git problem. I repushed a slightly modified commit at that seems to have solved most of the problem. Still a couple of job failing. No idea why when others work.

lefou:
Do these tests (e.g. `scalalib.test.test`) succeed on your machine? Asking because I can reproduce that failure.

ajrnz:
Ah ok, I can reproduce it. I spent a while trying to figure out what is happening but got nowhere. Will have another go when I get a chance

ajrnz:
No problem. There is a something wrong but it's very hard to track down as there is very little logging of each action that takes place. It need to insert some debugging code to find out where the problem lies. It's not easy to see how it is happening just from inspection. The jars produced seem fine however... I need to find some time to look in more detail

TheElectronWill:
I've got [some working fixes](https://github.com/lihaoyi/mill/compare/master...TheElectronWill:jar-manifest) for jar manifest support. Is it better to open another PR from my branch?

ajrnz:
Updated the PR. Thanks. LGTM

lefou:
Thanks for the contribution! Nice collaboration!

PR #636, opened 2019-06-21
Support 'eclipse-plugin' type dependencies

kag0:
Resolve one particular package type from https://github.com/lihaoyi/mill/issues/635


lefou:
Thanks. I think we can safely support that type, which is essentially a JAR. If we later support explicit setting of type, we can reduce the list of automatically supported types, if there is any reason to do that.

PR #643, opened 2019-07-01
Fix build, split up release and assembly

lihaoyi:
We need to upgrade the mill version used in travis and appveyor to the current version that people are likely to be using

### DIFF
--- a/ci/publish-local.sh
+++ b/ci/publish-local.sh
@@ -2,6 +2,6 @@
 
 set -eux
 
-./mill -i all __.publishLocal release
+./mill -i all __.publishLocal executable
 
-mv out/release/dest/mill ~/mill-release
+mv out/executable/dest/mill ~/mill-release
--- a/ci/publish-local.sh
+++ b/ci/publish-local.sh
@@ -2,6 +2,6 @@
 
 set -eux
 
-mill -i all __.publishLocal release
+./mill -i all __.publishLocal release
 
 mv out/release/dest/mill ~/mill-release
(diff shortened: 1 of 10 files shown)


ci/test-mill-1.sh
```diff
@@ -6,4 +6,6 @@ set -eux
 git clean -xdf
 
 # Run tests
-mill integration.test "mill.integration.local.{JawnTests,BetterFilesTests,UpickleTests}"
+./mill integration.test "mill.integration.local.JawnTests"
+./mill integration.test "mill.integration.local.BetterFilesTests"
+./mill integration.test "mill.integration.local.UpickleTests"
```
lihaoyi: Not sure why this stuff is necessary, but it seems to make tests pass. Can investigate in a separate PR

ci/publish-local.sh
```diff
@@ -2,6 +2,6 @@
 
 set -eux
 
-mill -i all __.publishLocal release
+./mill -i all __.publishLocal executable
 
-mv out/release/dest/mill ~/mill-release
+mv out/executable/dest/mill ~/mill-release
```
lefou: Is there some reason why this is a `mv` and not a `cp`? Currently it violates the policy to not manually change everthing under `./out`. A sub-sequent run of `mill executable` will just skip because it is cached, although there is no executable anymore.

lihaoyi: @lefou no reason, should probably be a `cp`

lefou: Good to know, thanks.

Issue #620, opened 2019-05-28
Issue with custom repositories when trying to use Scoverage


jacarey:
Using two simple build files:

The following works fine:

```import coursier.maven.MavenRepository
import mill._
import mill.scalalib._
import ammonite.ops._
import mill.contrib.scoverage.ScoverageModule

private val dagrVersion             = "0.6.0-b2189cf-SNAPSHOT"
private val scalatestVersion        = "3.0.5"

object commons extends ScoverageModule {
    def scoverageVersion = "1.3.1"
    def scalaVersion = "2.12.2"

    def repositories = super.repositories ++ Seq(
      MavenRepository("https://oss.sonatype.org/content/repositories/public"),
    )

    def ivyDeps = Agg(
      ivy"com.fulcrumgenomics::dagr-tasks::$dagrVersion",
    )

    object test extends Tests {
      def ivyDeps = Agg(ivy"org.scalatest::scalatest::3.0.5")
      def testFrameworks = Seq("org.scalatest.tools.Framework")
    }
}
```

this does not

```
import coursier.maven.MavenRepository
import mill._
import mill.scalalib._
import ammonite.ops._
import mill.contrib.scoverage.ScoverageModule

private val dagrVersion             = "0.6.0-b2189cf-SNAPSHOT"
private val scalatestVersion        = "3.0.5"

object commons extends ScoverageModule {
    def scoverageVersion = "1.3.1"
    def scalaVersion = "2.12.2"

    def repositories = super.repositories ++ Seq(
      MavenRepository("https://oss.sonatype.org/content/repositories/public"),
    )

    def ivyDeps = Agg(
      ivy"com.fulcrumgenomics::dagr-tasks::$dagrVersion",
    )

    object test extends ScoverageTests {
      def ivyDeps = Agg(ivy"org.scalatest::scalatest::3.0.5")
      def testFrameworks = Seq("org.scalatest.tools.Framework")
    }
}
```

The only difference is that the second build file test extends ScoverageTests instead of Tests. 

Output from first build file:

```
mill commons.test
[31/56] commons.compile
warning: there were three deprecation warnings (since 2.12.0); re-run with -deprecation for details
warning: there were three feature warnings; re-run with -feature for details
two warnings found
[56/56] commons.test.test
```

Output from the second build file:
```
mill commons.test
[30/71] commons.scoverage.compileClasspath | Downloading [8/8] artifacts (~0/0 bytes)
1 targets failed
commons.scoverage.compileClasspath
[...]

Issue #640, opened 2019-06-25
Scoverage: Issue when also using `BuildInfoModule` 


lefou:
When I add `ScoverageModule` to a project, that also uses the `BuildInfo` module to genrate a `BuildInfo.scala` file, the file generator is not properly invoked.

The compile step fails, because there is no generated `BuildInfo.scala`.

```
$ mill core.test
[42/72] core.scoverage.compile 
[info] Compiling 2 Scala sources to /home/lefou/work/opensource/mill-osgi/out/core/scoverage/compile/dest/classes ...
[error] /home/lefou/work/opensource/mill-osgi/core/src/de/tobiasroeser/mill/osgi/OsgiBundleModule.scala:7:34: object internal is not a member of package de.tobiasroeser.mill.osgi
[error] import de.tobiasroeser.mill.osgi.internal.BuildInfo
[error]                                  ^
[error] /home/lefou/work/opensource/mill-osgi/core/src/de/tobiasroeser/mill/osgi/OsgiBundleModule.scala:149:116: not found: value BuildInfo
[error]       log.error(s"Your used mill version is most probably too old. In case of errors use (at least) mill version ${BuildInfo.millVersion}.")
[error]                                                                                                                    ^
[error] /home/lefou/work/opensource/mill-osgi/core/src/de/tobiasroeser/mill/osgi/OsgiBundleModule.scala:269:97: not found: value BuildInfo
[error]   protected[osgi] def checkMillVersion(millVersion: Option[String]): Boolean = checkMillVersion(BuildInfo.millVersion, millVersion)
[error]                                                                                                 ^
[error] three errors found
1 targets failed
core.scoverage.compile Compilation failed
```

Here is the target execution plan with scoverage:
```
$ mill.scalalib.ZincWorkerModule.scalalibClasspath
core.test.forkArgs
core.test.forkEnv
core.test.testFrameworks
core.test.resources
mill.scalalib.ZincWorkerModule.classpath
mill.scalalib.ZincWorkerModule.compilerInterfaceClasspath
mill.scalalib.ZincWorkerModule.worker
mill.scalalib.ZincWorkerModule.worker
core.scoverage.upstreamCompileOutput
core.sources
core.scoverage.generatedSources
core.scoverage.allSources
core.scoverage.allSourceFiles
core.scoverage.transitiveLocalClasspath
core.resources
core.scoverage.unmanagedClasspath
core.scalaVersion
core.scoverage.scalaVersion
[...]

PR #645, opened 2019-07-02
Fix scoverage: inherit repositories from outer project

krcz:
`ScoverageModule` submodule inherits `ivyDeps` from the outer project, not inheriting `repositories`, which creates problems with resolving when package is in custom repository (#620).

### DIFF
--- a/contrib/scoverage/src/ScoverageModule.scala
+++ b/contrib/scoverage/src/ScoverageModule.scala
@@ -91,6 +91,7 @@ trait ScoverageModule extends ScalaModule { outer: ScalaModule =>
     def sources = outer.sources
     def resources = outer.resources
     def scalaVersion = outer.scalaVersion()
+    def repositories = outer.repositories
     def compileIvyDeps = outer.compileIvyDeps()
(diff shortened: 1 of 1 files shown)


lefou:
Thank you. I'll merge when tests pass. But please notice, this is not the only issue with the scoverage module right now. See discussion of #640 if you are interested.

krcz:
@lefou I don't think I have good enough knowledge on mill internals to provide a good solution on that - I just started using it two weeks ago.

I believe generally the problem is that you need to compile all the sources with additional compiler options you don't want in the final build - and it seems that mill doesn't support tasks parametrized by some context, which is why `scoverage` attempts to clone the original build, but there are many edge cases where it doesn't work. In case of your issue I might try adding `def generatedSources = outer.generatedSources` as well, but I'm not sure if that's going to work.

nvander1:
@krcz Thanks for this fix! And you're right about the generatedSources I believe.

krcz:
@lefou Thank you for quick response and merge!

PR #649, opened 2019-07-06
Interpolates $MILL_VERSION in ivy imports

Baccata:
This overrides the default $ivy magic import by replacing the `$MILL_VERSION` string with mill's version, in order to facilitate the loading/update of contrib modules.

Ammonite had to be bumped to 1.6.9

NB : I've given it a manual test to see in the logs that coursier attempts to resolve the correct version (and fails, since the version is not yet published). It'd feel a bit weird to add that particular test, which would mean that pulling the latest master and running the tests on it would be red, since the version would then be published and coursier would be able to resolve it ... 

EDIT : apologies for the accidental whitespace removal in the readme. You can read the diff there without the whitespaces https://github.com/lihaoyi/mill/pull/649/files?w=1

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -2,7 +2,6 @@ import $file.ci.shared
 import $file.ci.upload
 import java.nio.file.attribute.PosixFilePermission
 import $ivy.`org.scalaj::scalaj-http:2.4.1`
-import $ivy.`com.lihaoyi::mill-contrib-bloop:0.4.2`
 
 import coursier.maven.MavenRepository
 import mill._
--- a/build.sc
+++ b/build.sc
@@ -2,6 +2,7 @@ import $file.ci.shared
 import $file.ci.upload
 import java.nio.file.attribute.PosixFilePermission
 import $ivy.`org.scalaj::scalaj-http:2.4.1`
+import $ivy.`com.lihaoyi::mill-contrib-bloop:0.4.2`
 
 import coursier.maven.MavenRepository
 import mill._
@@ -94,14 +95,30 @@ object main extends MillModule {
 
     def ivyDeps = Agg(
       // Keep synchronized with ammonite in Versions.scala
-      ivy"com.lihaoyi:::ammonite:1.6.7",
+      ivy"com.lihaoyi:::ammonite:1.6.9",
       // Necessary so we can share the JNA classes throughout the build process
       ivy"net.java.dev.jna:jna:4.5.0",
       ivy"net.java.dev.jna:jna-platform:4.5.0"
     )
 
     def generatedSources = T {
-      Seq(PathRef(shared.generateCoreSources(T.ctx().dest)))
+      val dest = T.ctx().dest
+      val version = publishVersion()
+      writeBuildInfo(dest, version)
+      shared.generateCoreSources(dest)
+      Seq(PathRef(dest))
+    }
+
+    def writeBuildInfo(dir : os.Path, version : String) = {
+      val code = s"""
+        |package mill
+        |
+        |object BuildInfo {
+        |  val millVersion = "$version"
+        |}
+      """.stripMargin.trim
+
+      os.write(dir / "BuildInfo.scala", code)
     }
   }
 
--- a/main/core/src/util/Compat.scala
+++ b/main/core/src/util/Compat.scala
@@ -1,0 +1,13 @@
+package mill.util
+
+import scala.reflect.macros.blackbox.Context
+
+object Compat{
+  def copyAnnotatedType(c: Context)
+                       (tpe: c.universe.AnnotatedType,
+                        newAnnots: List[c.universe.Annotation]) = {
+    import c.universe.compat._
+
+    c.universe.AnnotatedType(newAnnots, tpe.underlying)
+  }
+}
--- a/scalalib/src/Versions.scala
+++ b/scalalib/src/Versions.scala
@@ -2,7 +2,7 @@ package mill.scalalib
 
 object Versions {
   // Keep synchronized with ammonite dependency in core in build.sc
-  val ammonite = "1.6.7"
+  val ammonite = "1.6.9"
   // Keep synchronized with zinc dependency in scalalib.worker in build.sc
   val zinc = "1.2.5"
 }
--- a/scalalib/test/src/HelloWorldTests.scala
+++ b/scalalib/test/src/HelloWorldTests.scala
@@ -212,6 +212,7 @@ object HelloWorldTests extends TestSuite {
     object foo extends ScalaModule {
       def scalaVersion = "2.11.8"
       override def scalaOrganization = "org.typelevel"
+      override def ammoniteVersion = "1.6.7"
 
       def ivyDeps = Agg(
         ivy"com.github.julien-truffaut::monocle-macro::1.4.0"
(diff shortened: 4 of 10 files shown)


main/core/src/util/Compat.scala
```diff
@@ -0,0 +1,13 @@
+package mill.util
+
+import scala.reflect.macros.blackbox.Context
+
+object Compat{
```
Baccata: This was removed from Ammonite there https://github.com/lihaoyi/Ammonite/pull/981/files

Baccata:
Windows build is failing for unrelated reasons

Baccata:
@tgodzik, this would allow for the removal of the following hardcoded value (and also the logic to read millw config file) : 

https://github.com/scalameta/metals/blob/302b2cf35920e9b27a2c05a314f14bae083b66b4/metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala#L65

Baccata:
Build is failing due to ammonite 1.6.9 not having been published for the 2.11 series. Not sure whether failure or intended. I'll update relevant tests when I have a reply on that.

tgodzik:
> @tgodzik, this would allow for the removal of the following hardcoded value (and also the logic to read millw config file) :
> 
> https://github.com/scalameta/metals/blob/302b2cf35920e9b27a2c05a314f14bae083b66b4/metals/src/main/scala/scala/meta/internal/builds/MillBuildTool.scala#L65

That's really cool! I think we still need to leave the hardcoded version inside, since the default is to use embedded mill and we need a version that we are sure works. I am open to suggestions though if you have ideas on how to make it better.

Baccata:
@tgodzik good question, haha. My personal opinion would be to have the default be the latest release, which could theoretically be retrieved simply via `curl https://github.com/lihaoyi/mill/releases/latest` and parsing the redirection URL. 

I however completely understand the desire to be conservative and ensuring stable behaviour, and keeping the default is absolutely fine 😄

tgodzik:
It would be awesome if we could automate a PR whenever a new release of mill is available. That would be the ideal middle ground.  I will see if there is any kind of bot available.

scalalib/test/src/HelloWorldTests.scala
```diff
@@ -847,7 +847,6 @@ object HelloWorldTests extends TestSuite {
     'typeLevel - workspaceTest(HelloWorldTypeLevel){ eval =>
       val classPathsToCheck = Seq(
         HelloWorldTypeLevel.foo.runClasspath,
-        HelloWorldTypeLevel.foo.ammoniteReplClasspath,
```
Baccata: This check breaks because `HelloWorldTypeLevel` is a scala 2.11 ScalaModule and support for 2.11 has been dropped in Ammonite 1.6.9

lihaoyi:
Going forward I'd like to try and get everyone to standardize on a `./mill` script or `./millw` and `.mill-version`; if we do that then tools like that would be able to read the version of Mill required for any particular repo.

Baccata:
The builds seems to be red for an unrelated reason : 

### Travis 

``` 
integration.test.test mill.integration.local.CaffeineTests mill.integration.local.CaffeineTests.test
```

### Appveyor : 

``` 
/usr/local/bin/mill: line 10: C:\Program: No such file or directory
```

lihaoyi:
caffeine might just be flaky

scalalib/src/Versions.scala
```diff
@@ -2,7 +2,7 @@ package mill.scalalib
 
 object Versions {
```
lefou: And when we start to use `BuildInfo` module in our own build (see my comment to `MillModule`), we also might want to generate this file. _But this is only note, not a request, as that change is unrelated to the success of this PR._

scalalib/test/src/HelloWorldTests.scala
lefou: Without knowing what this line does, can we disable it only in case we run under Scala 2.11?

build.sc
```diff
@@ -94,14 +94,30 @@ object main extends MillModule {
...
     def generatedSources = T {
-      Seq(PathRef(shared.generateCoreSources(T.ctx().dest)))
+      val dest = T.ctx().dest
+      val version = publishVersion()
+      writeBuildInfo(dest, version)
+      shared.generateCoreSources(dest)
+      Seq(PathRef(dest))
+    }
+
+    def writeBuildInfo(dir : os.Path, version : String) = {
```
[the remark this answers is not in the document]
Baccata: I don't disagree with the principle, but BuildInfo being a contrib module, I don't think building mill should depend on it. 

Moreover, my personal opinion is that the `BuildInfo` sbt plugin only exists because people don't understand sbt enough to write that logic themselves. Mill having much simpler semantics, everybody can write the logic and I don't think there's a justification to have a "BuildInfo" plugin. 

That being said, you're a maintainer and I'm not, so if you're convinced this PR should make use of the buildInfo module, I'll abide (ironically though, I'll have to hardcode the version as this PR hasn't been merged yet )

scalalib/test/src/HelloWorldTests.scala
Baccata: Instead of ignoring the problem, I surfaced the ammonite version as a target in the scalalib trait : f386c83 

This will allow people still building against 2.11 to run ammonite on their projects via the `repl` command

build.sc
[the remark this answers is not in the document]
lefou: Good point. Yet, we repeat ourselves when we do it "manually". So, please treat my request as not mandatory.

scalalib/src/Versions.scala
lefou: Ignore this comment as well, it's unrelated to this PR.

lihaoyi:
> Good point. Yet, we repeat ourselves when we do it "manually". So, please treat my request as not mandatory.

I would be happy to add a dependency on `BuildInfo`, and eventually just promote it from `contrib/` to the core `main/` codebase. IIRC we have some spots where it would be nice to keep things in sync (e.g. Mill's Ammonite version and the `foo.repl` Ammonite version)

Or not, up to you guys

Baccata:
The only thing I'm really opposed to is : 

```
import $ivy.`com.lihaoyi::mill-contrib-buildinfo:0.4.2` // hardcoded version 
```

I can pick the "BuildInfo" usage in a subsequent PR, with the following sequence of actions : 

#### Option 1 

1. Have this one merged 
2. Publish mill's next version 
3. Bump the version of mill used to build mill
4. Add dependency on `import $ivy.`com.lihaoyi::mill-contrib-buildinfo:$MILL_VERSION`, and use it to replace the values that are currently hardcoded in the runtime code (as well as my ad-hoc build-info generated source)  

#### Option 2 

1. Move the BuildInfo trait into `scalalib` module (it doesn't require any additional dependency, might as well just do it) 
2. Publish mill's next version 
3. Bump the version of mill used to build mill
4. Replace the values that are currently hardcoded in the runtime code (as well as my ad-hoc build-info generated source if that work is done before this PR gets merged)

lefou:
@Baccata I'm for option 1, as BuildInfo module is already published as contrib module and used by others (including me).

Please note, that mill master is already build with mill 0.5.0, so the hardcoded version should be 0.5.0 for now.

Baccata:
> Please note, that mill master is already build with mill 0.5.0, so the hardcoded version should be 0.5.0 for now. 

Proving the necessity of this very PR 😄

lefou:
> > Please note, that mill master is already build with mill 0.5.0, so the hardcoded version should be 0.5.0 for now.
> 
> Proving the necessity of this very PR smile

I'm not sure I can follow. Using BuildInfo now should be already possible, the only thing we need to changed __after__ the next release, is changing the hardcoded import version to the variable one, which depends on this PR.

lefou:
Nonetheless we can just merge as is, and make the move to BuildInfo after the next release.

lefou:
Thanks again!

PR #655, opened 2019-07-10
Improve ergonomics of running and generating the launcher script

anatoliykmetyuk:
- Fix the `dev.run` command documentation in README

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -63,10 +63,10 @@ To run test suites:
 To manually test Mill on a small build, you can use the `scratch` folder:
 
 ```bash
-./mill -i dev.run scratch -w show thingy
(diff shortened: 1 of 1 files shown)


build.sc
```diff
@@ -540,6 +540,7 @@ object dev extends MillModule{
     } else {
       os.perms.set(outputPath, "rwxrwxrwx")
     }
+    T.ctx().log.info(s"Launcher script was written to $outputPath")
```
lihaoyi-databricks: We shouldn't need to log this path here; anyone who wants it can already use `mill show dev.launcher` to see the full path to the script, or the value returned by any other task they wish to run

anatoliykmetyuk: Ok, makes sense

Issue #591, opened 2019-04-11
For Dotty, download the binary version of the compiler bridge instead of compiling it from sources


smarter:
Starting with Dotty 0.13.0-RC1, the bridge is a Java-only project (https://github.com/lampepfl/dotty/tree/master/sbt-bridge/src/xsbt) and binary artifacts are published on Maven: https://repo1.maven.org/maven2/ch/epfl/lamp/dotty-sbt-bridge/0.13.0-RC1/. This means you can avoid all the complications caused by having to fetch it sources, compiling these sources, caching the result, etc. In particular, it means https://github.com/lihaoyi/mill/issues/389 ceases to be a problem.

The relevant code in mill is in https://github.com/lihaoyi/mill/blob/master/scalalib/worker/src/ZincWorkerImpl.scala

lefou:
What exactly is the issue here? Doesn't commit 981c4e1d6919e3a4ceea6a4c63335d7b69d9e8aa fix the Java-only compilation for the compiler bridge?

lefou:
For reference: https://github.com/lihaoyi/mill/pull/569

smarter:
This is a request for enhancement not a bug :). The current solution works fine, but as explained above, using a pre-built binary version of the bridge avoids a lot of complications (and is faster for the initial compilation too).

PR #659, opened 2019-07-15
Use the binary version of the compiler bridge when available

TheElectronWill:
Implement #591: for Dotty 0.13.0-RC1 or more recent, use the normal (non-src) compiler bridge jar.
I've checked that no sources are unpacked/compiled with Dotty `0.16.0-RC3`.

### DIFF
--- a/scalalib/src/ZincWorkerModule.scala
+++ b/scalalib/src/ZincWorkerModule.scala
@@ -7,15 +7,15 @@ import mill.T
 import mill.api.KeyedLockedCache
 import mill.define.{Discover, Worker}
 import mill.scalalib.Lib.resolveDependencies
-import mill.scalalib.api.Util.isDotty
+import mill.scalalib.api.Util.{isDotty, isBinaryBridgeAvailable}
 import mill.scalalib.api.ZincWorkerApi
 import mill.api.Loose
 import mill.util.JsonFormatters._
 
-object ZincWorkerModule extends mill.define.ExternalModule with ZincWorkerModule{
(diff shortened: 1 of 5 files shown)


lihaoyi:
@smarter could you help review this? I have no idea what the dotty compiler bridge is about

smarter:
> I have no idea what the dotty compiler bridge is about

The gist is that the dotty bridge no longer requires retrieving and compiling a source jar, because it's now written purely in Java and we publish a ready-to-use binary jar, this should simplify everyone's life!

scalalib/src/ZincWorkerModule.scala
```diff
@@ -69,8 +69,9 @@ trait ZincWorkerModule extends mill.Module{
     instance.asInstanceOf[mill.scalalib.api.ZincWorkerApi]
   }
 
-  def scalaCompilerBridgeSourceJar(scalaVersion: String,
-                                   scalaOrganization: String) = {
+  def scalaCompilerBridgeJar(scalaVersion: String,
```
smarter: Please add a documentation comment to this method which in particular should describe what it returns.

scalalib/worker/src/ZincWorkerImpl.scala
```diff
@@ -22,7 +22,7 @@ case class MockedLookup(am: File => Optional[CompileAnalysis]) extends PerClassp
 }
 
 class ZincWorkerImpl(compilerBridge: Either[
```
smarter: Here too, a documentation comment explaining the various things that `compilerBridge` can be would be useful.

Issue #603, opened 2019-05-10
Make it easier to create "optional" dependencies


lefou:
If you need to set the `optional` flag (Maven semantics) to a dependency, you can not do it with the `ivy` string context or it's returned `Dep` instance. Instead you have to manipulate the enclosed coursier dependency object, which is cumbersome.

I suggest adding a `optional` def to `mill.scalalib.Dep`.

Example:

```scala
ivy"org.slf4j:slf4j-api:1.7.25".optional(true)
```

lihaoyi:
sounds good

lefou:
Currently, published ignores the optional-flag.

PR #661, opened 2019-07-16
Ease the declaration of optional dependencies + handle them in maven and ivy publications

TheElectronWill:
Usage (see #603):
```scala
ivy"org.slf4j:slf4j-api:1.7.25".optional() // or .optional(true)
```

Regarding the publications:
- Optional dependencies are added to the `optional` ivy group. Is more work necessary here?
- `<optional>true</optional>` is added to optional maven dependencies


lihaoyi:
@alexarchambault does this look reasonable to you? If so I'm happy to merge once travis is green

alexarchambault:
That seems ok, yeah. It's a pity there are no non-regression tests to validate this is all fine (it's easy to make mistakes with this kind of stuff).

PR #670, opened 2019-08-01
Make mill wrapper backwards compatible

kag0:
resolves https://github.com/lihaoyi/mill/issues/669

### DIFF
--- a/mill
+++ b/mill
@@ -20,9 +20,13 @@ fi
 MILL_DOWNLOAD_PATH="$HOME/.mill/download"
 MILL_EXEC_PATH="${MILL_DOWNLOAD_PATH}/$MILL_VERSION"
 
+version_remainder="$MILL_VERSION"
+MILL_MAJOR_VERSION="${version_remainder%%.*}"; version_remainder="${version_remainder#*.}"
+MILL_MINOR_VERSION="${version_remainder%%.*}"; version_remainder="${version_remainder#*.}"
+
 if [ ! -x "$MILL_EXEC_PATH" ] ; then
   mkdir -p $MILL_DOWNLOAD_PATH
-  if [ "$MILL_VERSION" > "0.4.2" ] ; then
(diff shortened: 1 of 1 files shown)


mill
```diff
@@ -22,8 +22,11 @@ MILL_EXEC_PATH="${MILL_DOWNLOAD_PATH}/$MILL_VERSION"
 
 if [ ! -x "$MILL_EXEC_PATH" ] ; then
   mkdir -p $MILL_DOWNLOAD_PATH
+  if [ "$MILL_VERSION" > "0.4.2" ] ; then
```
lefou: Is this doing what you want? What happens, if we release mill `0.10.0`?

kag0: :grimacing: 
I think `0.10.0` is still larger than `0.4.2`, but it might be smaller than `0.4.22`.
I guess we should probably take a less naive approach and split the version numbers by section and compare them separately.

mill
```diff
@@ -20,9 +20,13 @@ fi
 MILL_DOWNLOAD_PATH="$HOME/.mill/download"
 MILL_EXEC_PATH="${MILL_DOWNLOAD_PATH}/$MILL_VERSION"
 
+version_remainder="$MILL_VERSION"
+MILL_MAJOR_VERSION="${version_remainder%%.*}"; version_remainder="${version_remainder#*.}"
```
lefou: Just to be sure, is this `sh` compatible or is this `bash` specific? It looks like it will only work with bash.

kag0: I think it's sh compatible (at least intellij shell plugin isn't giving me any POSIX warnings and it seems to run OK on bash in POSIX mode).
Substring removal is used here as well https://github.com/lihaoyi/mill/blob/master/mill#L26 so we'll need to change that if string manipulation is not compatible with our target.
Are we targeting POSIX shells or another?

lefou: Yep, POSIX shells. Thanks for checking.

lefou:
Can you please rebase to trigger CI? Your last rebase killed the CI job.

kag0:
@lefou rebased

lefou:
Great, thanks!

PR #678, opened 2019-08-20
Update semanticDBVersion and semanticDBSupported

gabro:
This allows Metals to support 2.12.9 on Mill builds


lefou:
Thanks!

gabro:
@lefou thanks for merging! Is there any planned release of Mill which will include this? Until then Metals won't work with Mill projects which use 2.12.9

lefou:
@gabro Releases are in general easy to do (in fact, each master commit gets released automatically), but currently, they are cut by @lihaoyi .

You can consume mill-0.5.0-18-d2b311, which is the first release containing this PR, from GitHub relase pages (binary jar) and Maven Central (all module and contrib plugins).

gabro:
@lefou Thanks for the info! I'll make sure mill-0.5.0-18-d2b311 works with Metals on a 2.12.9 Mill project while we wait for a new stable release.

PR #682, opened 2019-08-26
Properly support Dotty >= 0.18.1

smarter:
Dotty now uses the 2.13 standard library, so `withDottyCompat` needs to
use the correct suffix depending on the Dotty version.


lefou:
Some test fail, including the newly added one (`ci/test-mill-dev.sh`).

```
...
1 targets failed
all 1 targets failed
scalalib.test.test mill.scalalib.HelloWorldTests mill.scalalib.HelloWorldTests.dotty213
The command "$CI_SCRIPT" exited with 1.
```

smarter:
Should be good for review now.

PR #686, opened 2019-09-03
Bump flyway version

kag0:


kag0:
Is there something I should be doing to make appveyor pass?

lihaoyi-databricks:
I think appveyor is just borked these days, feel free to ignore it for now

lefou:
Out of curiosity, why is that custom logger needed?

kag0:
The old version had a stdout logger available for us to import in flyway core. The new version has moved that logger into flyway command line and into private scope. So just adding it ourselves prevents us from having to add another dependency and needing to make some classes outside our package namespace.

PR #691, opened 2019-09-10
Changed JSON encoding for StackTraceElement (filename is nullable)

lefou:
Fixes https://github.com/lihaoyi/mill/issues/690


lihaoyi-databricks:
Looks ok to me. This stuff is all internal so I doubt we need to worry about anyone else breaking due to the change in serialization

PR #697, opened 2019-09-21
Add JBuildInfo

carueda:


carueda:
Thanks @lefou for considering this addition and for the suggestion, which has just been incorporated. I'm just releasing the new 0.1.2 version to maven central.

PR #701, opened 2019-09-25
Bump Ammonite version

lihaoyi:


lefou:
Is it fixing some issues, or why do we depend on an untagged version?

lihaoyi:
was just seeing if i could propagate a new binary-incompatible version of upickle through my graph of projects; seems like some updates are needed, so may punt on this for now

PR #704, opened 2019-10-03
WIP Fix "htmlReport" does not write into dest directory

### LINKED ISSUE #651: Scoverage: htmlReport does not write into dest directory
All generated files are written to `out/project/scoverage/htmlRepost` whereas they should be written under `out/project/scoverage/htmlRepost/dest` (notice the `dest`). 

Mill version: 0.5.0

nvander1:
@lefou To fix #651

It's been a bit since I've seen the code. I can't recall why I have duplicated the worker modules for scoverage 1.3.1 and 1.4.0. I want to address it, but I don't think I should mix that address with this bugfix. I'll open a separate patch for that.

Also, going to try to modify the tests to verify these reports are generated in the right directories. I'll dig into the project to look for similar tests that do this, if you know of any modules to check as a starting point let me know :)

### DIFF
--- a/contrib/scoverage/worker/1.3.1/src/ScoverageReportWorkerImpl.scala
+++ b/contrib/scoverage/worker/1.3.1/src/ScoverageReportWorkerImpl.scala
@@ -17,7 +17,7 @@ class ScoverageReportWorkerImpl extends ScoverageReportWorkerApi {
     val coverage = deserialize(coverageFileObj)
     coverage(invoked(findMeasurementFiles(dataDir)))
     val sourceFolders = sources.map(_.path.toIO)
-    val folder = new java.io.File(s"${selfDir}/${reportType.folderName}")
+    val folder = new java.io.File(s"${selfDir}/${reportType.folderName}/dest")
     folder.mkdir()
     reportType match {
       case Html =>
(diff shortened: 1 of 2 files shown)


contrib/scoverage/worker/1.3.1/src/ScoverageReportWorkerImpl.scala
```diff
@@ -17,7 +17,7 @@ class ScoverageReportWorkerImpl extends ScoverageReportWorkerApi {
     val coverage = deserialize(coverageFileObj)
     coverage(invoked(findMeasurementFiles(dataDir)))
     val sourceFolders = sources.map(_.path.toIO)
-    val folder = new java.io.File(s"${selfDir}/${reportType.folderName}")
+    val folder = new java.io.File(s"${selfDir}/${reportType.folderName}/dest")
```
lefou: Best would be to take a `implicit ctx: Ctx.Dest` parameter in the `def buildReport` instead of the explicit `selfDir`. Then you can simply use `ctx.dest` as target folder. This would made the caller side easier too.

lefou: @nvander1 I opened a PR to your PR some time ago...

lefou:
This PR is stalled for some time. Closing because of inactivity.

I opened #735 as a successor.

PR #705, opened 2019-10-03
Remove semanticdb settings

Baccata:
Bloop 1.3.3 can assume the responsibility of finding the semanticdb plugin and setting the relevant options in the compiler, which means we don't have to do it ourselves anymore to accommodate metals users.

See https://github.com/scalameta/metals/pull/852

### DIFF
--- a/contrib/bloop/src/mill.contrib.bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill.contrib.bloop/BloopImpl.scala
@@ -129,33 +129,6 @@ class BloopImpl(ev: () => Evaluator, wd: Path) extends ExternalModule { outer =>
     bloopDir / s"${name(module)}.json"
 
   //////////////////////////////////////////////////////////////////////////////
-  // SemanticDB related configuration
-  //////////////////////////////////////////////////////////////////////////////
-
-  // Version of the semanticDB plugin.
-  def semanticDBVersion: String = "4.2.3"
-
-  // Scala versions supported by semantic db. Needs to be updated when
-  // bumping semanticDBVersion.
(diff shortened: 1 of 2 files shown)


lefou:
Thanks.

jodersky:
Unfortunately this appears to have been merged prematurely, before a new metals release supporting scalameta/metals#852 has been released.

Here is the timeline of events:

- September 5th: mill 0.5.1 is released
- September 23rd: metals 0.7.6 is released
- October 3rd: scalameta/metals#852 is merged into metals, unblocking this PR
- October 3rd: this PR is merged
- October 17th: mill 0.5.2 is released

The effect is that, as of now, the latest version of mill (0.5.2) does not fully work with the latest version of metals (0.7.6).

This means that if editor features relying on semanticdb are desired, any new features added to mill between 0.5.1 and 0.5.2 are unavailable.

A new release of metals will probably fix this.

Just leaving this comment here as the issue is subtle and affects only the official latest releases (i.e. it can't be reproduced when using latest snapshots of all components)

lefou:
@jodersky Good to know. Is there a workaround for those who need mill 0.5.2, e.g. manually adding the semanticdb dependency?

tgodzik:
> @jodersky Good to know. Is there a workaround for those who need mill 0.5.2, e.g. manually adding the semanticdb dependency?

Sorry about the issue, as a workaround you use one of the latest snapshots of Metals. They listed on the website here: https://scalameta.org/metals/docs/editors/vscode.html#using-latest-metals-snapshot

Baccata:
Manually adding the semanticdb dependency (and the couple related options) would indeed be another workaround.

jodersky:
Using a metals snapshot works great, thanks! I mostly wanted to leave a comment here just in case anyone else runs into this issue.

Issue #601, opened 2019-05-06
ScalaJSModule#Tests fails to resolve test framework when moduleKind is set


Baccata:
Found in mill 0.3.6 

### Minimal example to reproduce 

```scala
import mill._
import mill.scalalib._
import mill.scalajslib._
import ammonite.ops._

object pipo extends ScalaJSModule {

  def scalaVersion = "2.12.8"
  def scalaJSVersion ="0.6.26"

  object test extends Tests {
    override def ivyDeps = Agg(ivy"com.lihaoyi::utest::0.6.6")
    override def testFrameworks = Seq("utest.runner.MillFramework")
    // Commenting out the following line results in no-crash. 
    override def moduleKind = T(mill.scalajslib.api.ModuleKind.CommonJSModule)
  }
}
```

### Error message : 

```
oliviermelois@otana:/tmp/pipo$ mill pipo.test
Compiling /private/tmp/pipo/build.sc
[74/75] pipo.test.fastOptTest 
Linker: Compute reachability: 371294 us
Linker: Assemble LinkedClasses: 1376240 us
Basic Linking: 1964107 us
Emitter: Class tree cache stats: reused: 1054 -- invalidated: 1062
Emitter: Method tree cache stats: resued: 0 -- invalidated: 10560
Emitter (write output): 904164 us
[75/75] pipo.test.test 
Starting process: node
1 targets failed
pipo.test.test java.lang.RuntimeException: Failed to get framework
    mill.scalajslib.worker.ScalaJSWorkerImpl.$anonfun$getFramework$3(ScalaJSWorkerImpl.scala:75)
    scala.Option.getOrElse(Option.scala:138)
    mill.scalajslib.worker.ScalaJSWorkerImpl.getFramework(ScalaJSWorkerImpl.scala:75)
    mill.scalajslib.ScalaJSWorker.getFramework(ScalaJSWorkerApi.scala:63)
    mill.scalajslib.TestScalaJSModule.$anonfun$test$1(ScalaJSModule.scala:179)
    mill.define.ApplyerGenerated.$anonfun$zipMap$6(Applicative
```

Baccata:
Relevant convo from the sjs gitter : https://gitter.im/scala-js/scala-js?at=5cd01bc5e416b845193c9444

javaduke:
I'm running into the same issue when building my project, is there any fix for it?

gzm0:
[...]

PR #706, opened 2019-10-03
Fix for ScalaJS tests issue #601

javaduke:


javaduke:
This addresses issue #601

javaduke:
> @javaduke this seems to have broken some of the existing tests. Do you think you could help investigate the travis failures?

I fixed the issue but now travis fails with some weird error, not sure how to fix it...

lihaoyi:
@javaduke can you try rebasing and squashing your commits on top of the latest master? This is a travis bug and squashing normally makes it go away

javaduke:
Yay, that did the trick, Travis works now. The Appveyor still fails though, do you know how to fix it?

scalajslib/worker/0.6/src/ScalaJSWorkerImpl.scala
```diff
@@ -3,29 +3,33 @@ package scalajslib
...
 import org.scalajs.core.tools.jsdep.ResolvedJSDependency
 import org.scalajs.core.tools.linker.{ModuleInitializer, Semantics, StandardLinker, ModuleKind => ScalaJSModuleKind}
 import org.scalajs.core.tools.logging.ScalaConsoleLogger
 import org.scalajs.jsenv._
 import org.scalajs.testadapter.TestAdapter
-import mill.scalajslib.api.{JsEnvConfig, ModuleKind}
+
 class ScalaJSWorkerImpl extends mill.scalajslib.api.ScalaJSWorkerApi {
+    
+  var scalaJSModuleKind: ScalaJSModuleKind = null;
```
lihaoyi: @javaduke we should not have mutable variables if we can avoid it. If we need this field in `getFramework`, then we should pass it in

lihaoyi:
one more comment. Just ignore appveyor for now

scalajslib/worker/0.6/src/ScalaJSWorkerImpl.scala
javaduke: Good point, fixed.

PR #714, opened 2019-10-15
Add `scalaPBProtocPath` option to scalapblib

fduraffourg:
Add an option to ScalaPBModule to allow setting the protoc compiler. This is useful for systems that can't run the jar embedded protoc.


tdroxler:
Niiiiice, as a NixOS user this would be super useful  :+1:

lefou:
Thanks.

PR #721, opened 2019-10-20
make coursier download dependency sources

### LINKED ISSUE #720: Intellij Idea with bsp doesn't make coursier download dependency sources.
After #664, mill is able to generate bsp library,
However, there still exist some problem:
1. if mill is provided by binary distribution, idea cannot find the source mill, which make the build.sc all red.
2. idea intellij won't download dependency source makes too hard to inspect API.
maybe @adadima and @samvel1024 would give a little help to this.

sequencer:
Before this PR, invoking bsp won't download sources from maven center, since `resolveDeps` is set to `false` by default, this PR will ask mill to download all sources(including mill sources) to fix this issue.
I think this should also be handled by Intellij too, since sbt has configurations to let user decide downloading source or not, bsp in Intellij didn't give this option to user.

### DIFF
--- a/contrib/bsp/src/mill/contrib/bsp/MillBuildServer.scala
+++ b/contrib/bsp/src/mill/contrib/bsp/MillBuildServer.scala
@@ -154,17 +154,17 @@ class MillBuildServer(evaluator: Evaluator,
       for (targetId <- dependencySourcesParams.getTargets.asScala) {
         val millModule = targetIdToModule(targetId)
         var sources = evaluateInformativeTask(evaluator,
-                                              millModule.resolveDeps(millModule.transitiveIvyDeps),
+                                              millModule.resolveDeps(millModule.transitiveIvyDeps, true),
                                               Agg.empty[PathRef]) ++
           evaluateInformativeTask(evaluator,
-                                  millModule.resolveDeps(millModule.compileIvyDeps),
+                                  millModule.resolveDeps(millModule.compileIvyDeps, true),
                                   Agg.empty[PathRef]) ++
           evaluateInformativeTask(evaluator,
                                   millModule.unmanagedClasspath,
                                   Agg.empty[PathRef])
         millModule match {
           case _: ScalaModule => sources ++= evaluateInformativeTask(evaluator,
-                                                                     millModule.resolveDeps(millModule.asInstanceOf[ScalaModule].scalaLibraryIvyDeps),
+                                                                     millModule.resolveDeps(millModule.asInstanceOf[ScalaModule].scalaLibraryIvyDeps, true),
                                                                      Agg.empty[PathRef])
           case _: JavaModule => sources ++= List()
         }


samvel1024:
The compilation of the branch fails :(
```
[info] Compiling 1 Scala source to /Users/sme/meel/out/contrib/bsp/compile/dest/classes ...
[error] /Users/sme/meel/contrib/bsp/src/mill/contrib/bsp/MillBuildServer.scala:161:55: variable definition needs type because 'sources' is used as a named argument in its body.
[error]                                   Agg.empty[PathRef]) ++
[error]
```

sequencer:
I just test and compile success? Strange...

samvel1024:
Well, seems like so. Commit `3fdf0629` is failing when running `sh ci/publish-local.sh`

sequencer:
My fault, sorry about that bug...

lefou:
I can't tell if this PR is good to merge. It would be great, if someone who understands the issue and tried this PR could comment/review. I'd be glad to merge after positive feedback.

sequencer:
Oh, I know, I may need help from @samvel1024 to review.

lihaoyi:
Regardless of who reviews this, it’s going to need a better PR description

sequencer:
Currently I think this PR is just a hot-fix to contrib/bsp, there still need something to do in the Intellij.
@lihaoyi So sorry for that, is this PR description OK?

lihaoyi:
@sequencer no problem, that description is fine. Doesn't need to be super fancy, just needs to be something that something that an ignorant future maintainer can glance at and see what's happening. e.g. with the description you have now, I already have a ton more context in what the PR is doing than I did last time I looked.

I can't sign off on it though, since I know nothing about BSP. Hopefully someone who knows something about either Coursier (@alexarchambault!) or the BSP can chime in and let us know if it looks sensible!

alexarchambault:
That seems more related to BSP than coursier, right? This fetches source JARs instead of standard JARs, around the lines this PR touches.

contrib/bsp/src/mill/contrib/bsp/MillBuildServer.scala
```diff
@@ -154,17 +154,17 @@ class MillBuildServer(evaluator: Evaluator,
       for (targetId <- dependencySourcesParams.getTargets.asScala) {
         val millModule = targetIdToModule(targetId)
         var sources = evaluateInformativeTask(evaluator,
-                                              millModule.resolveDeps(millModule.transitiveIvyDeps),
+                                              millModule.resolveDeps(millModule.transitiveIvyDeps, true),
```
lefou: Isn't this resolving source jars instead of binary jars?

lefou: Oh, now I see, the `var` already has the name `sources`, so indeed, the patch might just do fix something.

lefou:
I tend to merge, once the travis-ci tests are green. Could you please rebase on current master to trigger a new build on travis?

PR #724, opened 2019-10-28
Remove contrib modules from main distribution

lihaoyi:
They're bloating things up to an unacceptable degree, with the main jar size up to 50mb from 37mb not too long ago.

The whole point of contrib modules is that they're not part of the main distribution, but can be `import $ivy`ed in, so let's make them follow that convention


lihaoyi:
looks like the main distribution is back down to 37.6 MB again

PR #733, opened 2019-11-20
Add task to only check formatting of Scala files

jodersky:
Adds a new task 'checkFormat' to the scalafmt module that only checks if all files are formatted. It is basically equivalent to the `--test` flag that can be passed to scalafmt on the command line, and can come in very handy in CI scripts.

Caching is preserved wrt to the 'reformat' action, so that only unformatted or changed files are checked.

### DIFF
--- a/scalalib/src/scalafmt/ScalafmtWorker.scala
+++ b/scalalib/src/scalafmt/ScalafmtWorker.scala
@@ -8,6 +8,7 @@ import mill.api.Ctx
 import org.scalafmt.interfaces.Scalafmt
 
 import scala.collection.mutable
+import mill.api.Result
 
 object ScalafmtWorkerModule extends ExternalModule {
   def worker: Worker[ScalafmtWorker] = T.worker { new ScalafmtWorker() }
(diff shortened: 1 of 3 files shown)


scalalib/src/scalafmt/ScalafmtWorker.scala
```diff
@@ -21,44 +22,84 @@ private[scalafmt] class ScalafmtWorker {
...
+                  scalafmtConfig: PathRef)(implicit ctx: Ctx): Result[Unit] = {
+
+    val misformatted = reformatAction(input, scalafmtConfig, dryRun = true)
+    if (misformatted.isEmpty) {
+      Result.Success(())
     } else {
-      ctx.log.info(s"Everything is formatted already")
+      val out = ctx.log.outputStream
+      for (u <- misformatted) {
+        out.println(s"${u.path}")
```
jodersky: Printing unformatted files directly to stdout plays nicely with redirection. For instance, if I want to save the list of all unformatted files to another file yet still see error messages, I could call mill in the following way:

`mill project.checkFormat > unformatted.list`

scalalib/src/scalafmt/ScalafmtWorker.scala
```diff
@@ -21,44 +22,84 @@ private[scalafmt] class ScalafmtWorker {
...
+      val configPath =
+        if (os.exists(scalafmtConfig.path))
+          scalafmtConfig.path.toNIO
+        else
+          JPaths.get(getClass.getResource("default.scalafmt.conf").toURI)
+
+      // keeps track of files that are misformatted
+      val misformatted = mutable.ListBuffer.empty[PathRef]
+
+      def markFormatted(path: PathRef) = {
```
kag0: nit: I'd invoke `markFormatted` from `checkFormat` and `reformat` instead of in the `foreach` loop.

lihaoyi:
Thanks @kag0 !

PR #737, opened 2019-12-03
Added CoursierModule as parent of JavaModule

lefou:
This enables the user to resolve dependencies without defining a Java/Scala module, e.g. in assembly modules.

Fixes https://github.com/lihaoyi/mill/issues/722


lihaoyi:
lgtm

PR #738, opened 2019-12-05
Add support for Scala.js 0.6.29+ and 1.0.0-RC1.

sjrd:
Drop support for the obsolete Scala.js 1.0.0-M2.

Starting from Scala.js 0.6.29, test artifacts must depend on `scalajs-test-bridge` instead of `scalajs-test-interface`, as the JS bridge has been moved to that artifact. This is the only change required for 0.6.29+.

In Scala.js 1.x, in addition to depending on `scalajs-test-bridge`, test artifacts must include a module initializer that initializes the bridge. This required the addition of a `testBridgeInit: Boolean` parameter to the `link` methods scattered here and there. The parameter is ignored in 0.6.

The support for 1.0.0-RC1 is suboptimal: we need to force emitting ECMAScript 5.1 code instead of the new default ES 2015. The only reason is that the tests of Mill itself rely on Nashorn to execute some Scala.js outputs, and Nashorn does not support ES 2015. Since there was no configuration mechanism for the ES mode, I decided to always force ES 5.1. It would be better to introduce a configuration option for that in the future.

---

I also believe that this implementation will support 1.0.0 final and later releases, but of course it is hard to test, so I cannot guarantee that.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -18,11 +18,11 @@ object Deps {
   }
 
   object Scalajs_1_0 {
-    val scalajsEnvJsdomNodejs =  ivy"org.scala-js::scalajs-env-jsdom-nodejs:1.0.0-M2"
-    val scalajsEnvNodejs =  ivy"org.scala-js::scalajs-env-nodejs:1.0.0-M2"
-    val scalajsEnvPhantomjs =  ivy"org.scala-js::scalajs-env-phantomjs:1.0.0-M2"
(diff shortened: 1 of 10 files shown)


lefou:
This looks good to me, but I'm not a ScalaJS power user. Unfortunately there are merge conflicts, which need to resolved before merging. A second review would be good, too.

sjrd:
Rebased and fixed the conflict :)

lihaoyi:
Looked through this. I have no idea, but I hope @sjrd knows what he's doing!

Issue #560, opened 2019-02-25
Sysout is never shown in server-mode


lefou:
When starting mill without `-i` system output is never shown to the user. Also nothing appears in `out/mill-worker-1/logs`.

lefou:
Fixed by #741

PR #741, opened 2019-12-09
Write mill server output to files and forward their content into client (like tail -f)

lefou:
Related to https://github.com/lihaoyi/mill/issues/740
Related to #560

### DIFF
--- a/main/client/src/mill/main/client/FileToStreamTailer.java
+++ b/main/client/src/mill/main/client/FileToStreamTailer.java
@@ -10,88 +10,88 @@ import java.util.Optional;
 
 public class FileToStreamTailer extends Thread implements AutoCloseable {
 
-	private final File file;
-	private final PrintStream stream;
-	private final int intervalMsec;
+    private final File file;
+    private final PrintStream stream;
+    private final int intervalMsec;
 
-	private Optional<BufferedReader> reader = Optional.empty();
-	// if true, we won't read the whole file, but only new lines
-	private boolean ignoreHead = true;
+    private Optional<BufferedReader> reader = Optional.empty();
+    // if true, we won't read the whole file, but only new lines
+    private boolean ignoreHead = true;
 
-	private volatile boolean keepReading = true;
-	private volatile boolean flush = false;
+    private volatile boolean keepReading = true;
+    private volatile boolean flush = false;
 
-	public FileToStreamTailer(File file, PrintStream stream, int intervalMsec) {
-		super("Tail");
-		this.intervalMsec = intervalMsec;
-		setDaemon(true);
-		this.file = file;
-		this.stream = stream;
-	}
+    public FileToStreamTailer(File file, PrintStream stream, int intervalMsec) {
+        super("Tail");
+        this.intervalMsec = intervalMsec;
+        setDaemon(true);
+        this.file = file;
+        this.stream = stream;
+    }
 
-	@Override
-	public void run() {
-		if (isInterrupted()) {
-			keepReading = false;
-		}
-		while (keepReading || flush) {
-			flush = false;
-			try {
-				if (!reader.isPresent()) {
-					try {
-						this.reader = Optional.of(new BufferedReader(new FileReader(file)));
-					} catch (FileNotFoundException e) {
-						// nothing to ignore if file is inially missing
-						ignoreHead = false;
-					}
-				}
-				reader.ifPresent(r -> {
-					// read lines
-					try {
-						String line;
-						while ((line = r.readLine()) != null) {
-							if (!ignoreHead) {
-								stream.println(line);
-							}
-						}
-						// we ignored once
-						this.ignoreHead = false;
-					} catch (FileNotFoundException e) {
-						// File vanished
-						this.reader = Optional.empty();
-						this.ignoreHead = false;
-					} catch (IOException e) {
-						// could not read line or file vanished
-					}
-				});
-			} finally {
-				if (keepReading) {
-					// wait
-					try {
-						Thread.sleep(intervalMsec);
-					} catch (InterruptedException e) {
-						// can't handle anyway
-					}
-				}
-			}
-		}
-	}
+    @Override
+    public void run() {
+        if (isInterrupted()) {
+            keepReading = false;
+        }
+        while (keepReading || flush) {
+            flush = false;
+            try {
+                if (!reader.isPresent()) {
+                    try {
+                        this.reader = Optional.of(new BufferedReader(new FileReader(file)));
+                    } catch (FileNotFoundException e) {
+                        // nothing to ignore if file is inially missing
+                        ignoreHead = false;
+                    }
+                }
+                reader.ifPresent(r -> {
+                    // read lines
+                    try {
+                        String line;
+                        while ((line = r.readLine()) != null) {
+                            if (!ignoreHead) {
+                                stream.println(line);
+                            }
+                        }
+                        // we ignored once
+                        this.ignoreHead = false;
+                    } catch (FileNotFoundException e) {
+                        // File vanished
+                        this.reader = Optional.empty();
+                        this.ignoreHead = false;
+                    } catch (IOException e) {
+                        // could not read line or file vanished
+                    }
+                });
+            } finally {
+                if (keepReading) {
+                    // wait
+                    try {
+                        Thread.sleep(intervalMsec);
+                    } catch (InterruptedException e) {
+                        // can't handle anyway
+                    }
+                }
+            }
+        }
+    }
 
-	@Override
-	public void interrupt() {
-		this.keepReading = false;
-		super.interrupt();
-	}
+    @Override
+    public void interrupt() {
+        this.keepReading = false;
+        super.interrupt();
+    }
 
-	/**
-	 * Force a next read, even if we interrupt the thread.
-	 */
-	public void flush() {
-		this.flush = true;
-	}
+    /**
+     * Force a next read, even if we interrupt the thread.
+     */
+    public void flush() {
+        this.flush = true;
+    }
 
-	@Override
-	public void close() throws Exception {
-		interrupt();
-	}
+    @Override
+    public void close() throws Exception {
+        interrupt();
+    }
 }
--- a/main/client/src/mill/main/client/FileToStreamTailer.java
+++ b/main/client/src/mill/main/client/FileToStreamTailer.java
@@ -56,10 +56,6 @@ public class FileToStreamTailer extends Thread implements AutoCloseable {
                         }
                         // we ignored once
                         this.ignoreHead = false;
-                    } catch (FileNotFoundException e) {
-                        // File vanished
-                        this.reader = Optional.empty();
-                        this.ignoreHead = false;
                     } catch (IOException e) {
                         // could not read line or file vanished
                     }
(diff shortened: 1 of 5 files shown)


lihaoyi:
The issue with this approach is that the server a particular client are
connecting to may not be a process that client spawned. After all, the
point of servers is to serve multiple clients. In that case only the first
client will get logs and subsequent clients sending commands to the server
will not

lefou:
Isn't each invocation of mill creating a new client? But the server stays the same? So, subsequent calls to `mill hello` should not show the output, right?

```scala
// build.sc
import mill._

def hello() = T.command {
  os.proc("echo", "hello").call(stdout = os.Inherit, stderr = os.Inherit)
  ()
}
```
Output: 
```
11:53:02 θ64° lefou:/tmp/mill-1 $ for i in {1..10}; do ~/mill-release hello; sleep 1; done
[1/1] hello 
hello
[1/1] hello 
hello
[1/1] hello 
hello
[1/1] hello 
hello
[1/1] hello 
hello
[1/1] hello 
hello
[1/1] hello 
hello
[1/1] hello 
hello
[1/1] hello 
hello
[1/1] hello 
hello
11:53:38 θ64° lefou:/tmp/mill-1 14s $
```

lihaoyi:
@lefou it seems that inheriting it pushes the output from the server to the console of the first run of `MillClientMain`. If you then open up a second terminal in the same folder and run `mill hello` again, the first terminal gets all the output.

You can also verify this by making the output asynchronous:

```scala
// build.sc
import mill._

def hello() = T.command {
  new Thread(() => {
    Thread.sleep(1000)
    os.proc("echo", "hello").call(stdout = os.Inherit, stderr = os.Inherit)
  }).start()
  println("hello0")
  ()
}
```

In which case you will see the output is printed to the terminal even after the MillClientMain has exited.

lefou:
@lihaoyi Thank you. Now I understand.

lefou:
@lihaoyi I implemented some primitive `tail -f` pendant which seems to work reasonable well. If a client process is connected, the output will be delegated to the client process STDOUT/ERR. This works around the issues you described. Also, all output is preserved in the `out/mill-worker.../{stdout,stderr}` files.

Your test example that starts a background thread is in some way not idiomatic mill code though. Maybe we should add some warning in the docs that the use of threads and futures in tasks doesn't integrate well with mills execution concept. Nevertheless, the output will end in the stdout file and IFF a client is connected at print time, it will show up in that client, regardless whether it was the originating client or not.

I've chosen a rather arbitrary value of 20ms as wait time for polling for new output. Do you have an opinion on that?

main/client/src/mill/main/client/MillClientMain.java
```diff
@@ -122,10 +134,15 @@ public void run() {
                             args,
                             env
                     );
+
+                    // Should we give the server process some extra time to flush the output files?
+                    // currently, we don't
```
lihaoyi: If we call `System.out.flush()` and `System.err.flush()` on the server before closing the `serverSocket`, I think that would be enough to ensure that both output streams are flushed to disk

main/client/src/mill/main/client/FileToStreamTailer.java
```diff
@@ -0,0 +1,97 @@
...
+
+	@Override
+	public void run() {
+		if (isInterrupted()) {
+			keepReading = false;
+		}
+		while (keepReading || flush) {
+			flush = false;
+			try {
+				if (!reader.isPresent()) {
```
lihaoyi: Honestly I think this might be easier to read as an `if-else` on `null`s; the asymmetry between `if(!reader.isPresent())` and `reader.ifPresent` is pretty awkward

main/client/src/mill/main/client/FileToStreamTailer.java
```diff
@@ -0,0 +1,97 @@
...
+						String line;
+						while ((line = r.readLine()) != null) {
+							if (!ignoreHead) {
+								stream.println(line);
+							}
+						}
+						// we ignored once
+						this.ignoreHead = false;
+					} catch (FileNotFoundException e) {
+						// File vanished
```
lihaoyi: In what scenarios would the "File is missing" and "File vanished" errors occur? I wonder if we can specify the problem away, e.g. by having the protocol be that someone *must* create the log files before these tailers are started, and the server waits until the client has finished tailing before picking up another client. Otherwise it seems hard to be confident we handle all scenarios where the log file an disappear from under you

main/client/src/mill/main/client/FileToStreamTailer.java
```diff
@@ -0,0 +1,97 @@
...
+
+public class FileToStreamTailer extends Thread implements AutoCloseable {
+
+	private final File file;
+	private final PrintStream stream;
+	private final int intervalMsec;
+
+	private Optional<BufferedReader> reader = Optional.empty();
+	// if true, we won't read the whole file, but only new lines
+	private boolean ignoreHead = true;
```
lihaoyi: Could we specify that the server creates a new log file for every run? That way we do not need to worry about ignoring anything, and can always read the log files starting from the top

lefou: Not exactly. Even if we create new logs for each server run, as the log files are read by the client, each subsequent client would output all older logs.

main/client/src/mill/main/client/FileToStreamTailer.java
lefou: There is no scenario in mill besides "the user deleted it". I will just drop the catch.

main/client/src/mill/main/client/MillClientMain.java
```diff
@@ -96,13 +99,22 @@ public static int main0(String[] args) throws Exception{
             String lockBase = "out/mill-worker-" + jvmHomeEncoding + "-" + index;
             new java.io.File(lockBase).mkdirs();
 
+            File stdout = new java.io.File(lockBase + "/stdout");
+            File stderr = new java.io.File(lockBase + "/stderr");
+            int refeshIntervalMsec = 20;
```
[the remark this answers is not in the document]
lefou: Of course. We can still raise it later, it needed.

lefou:
> Could we use 4-space indents instead of tabs? I believe that's what we use for the rest of the java files in here.

Sure. I'm going to add a `.editorconfig` in a separate commit.

> One consideration is that this dual-stream approach results in some raciness w.r.t. interleaving of stdout and stderr. e.g.
> 
> ```scala
> System.out.println("A");
> System.err.println("B");
> System.out.println("C");
> System.err.println("D");
> System.out.println("E");
> System.err.println("F");
> ```
> 
> May no longer appear in the console in the order `A B C D E F`. This normally comes up when exceptions are thrown, since the stack traces go to stderr while user `println`s go to stdout, and as a result these may be jumbled up.
> 
> We could mitigate this problem by using a single combined stream/logfile (via `.redirectErrorStream()`) when the output is a console (TTY), and only using dual-streams/logfiles in other scenarios (e.g. output is a pipe). We could also expose a flag to allow users to force one mode or the other.
> 
> That would be a bit of work, but I think is worth doing to allow both piping as well as preserving the integrity of the user experience.

I really don't like to mix up the two streams. As we process the output always line-wise, we could prefix the lines with "E:" and "O:" and when processing split them up to the error and output stream. That might be a bit slower, but we don't expect large amount of text anyway, do we?

main/client/src/mill/main/client/FileToStreamTailer.java
lefou: This isn't starting an if-else block. Only in case a reader is not present, it will be initialized.

The following `ifPresent` uses the (now always initialized) reader.

lefou:
Currently the separation between STDOUT and STDERR is good enough, IMHO. 

```scala
// build.sc
import mill._

def hello() = T.command {
  os.proc("echo", "hello A").call(stdout = os.Inherit, stderr = os.Inherit)
  os.proc("logger", "-s", "hello B").call(stdout = os.Inherit, stderr = os.Inherit)
  os.proc("echo", "hello C").call(stdout = os.Inherit, stderr = os.Inherit)
  os.proc("logger", "-s", "hello D").call(stdout = os.Inherit, stderr = os.Inherit)
  os.proc("echo", "hello E").call(stdout = os.Inherit, stderr = os.Inherit)
  os.proc("logger", "-s", "hello F").call(stdout = os.Inherit, stderr = os.Inherit)
  System.out.println("A");
  System.err.println("B");
  System.out.println("C");
  System.err.println("D");
  System.out.println("E");
  System.err.println("F");
  ()
}

```

```
$ ~/mill-release hello
Compiling /tmp/mill-1/build.sc
[1/1] hello 
hello A
<13>Dec 11 09:31:32 lefou: hello B
hello C
<13>Dec 11 09:31:32 lefou: hello D
hello E
<13>Dec 11 09:31:32 lefou: hello F
A
B
C
D
E
F
```

lefou:
I think the `ProxyOutputStream` and it's pumper are still in use. They handle all direct system output via the socket. This PR only handles additional output produced by sub-processes of the server process, e.g. a `os.proc` call.

lihaoyi:
I'm still seeing it come out jumbled when running on the scratch folder, with the code in this PR and the `build.sc` file contents you posted above:

```bash
$ ./mill -i dev.run scratch hello
Compiling (synthetic)/ammonite/predef/interpBridge.sc
Compiling (synthetic)/ammonite/predef/sourceBridge.sc
Compiling (synthetic)/ammonite/predef/frontEndBridge.sc
[508/508] dev.run
[1/1] hello
hello A
Dec 11 20:42:50  lihaoyi[35748] <Notice>: hello B
hello C
Dec 11 20:42:50  lihaoyi[35750] <Notice>: hello D
hello E
A
B
C
Dec 11 20:42:50  lihaoyi[35752] <Notice>: hello F
D
E
F
```

Of course it gets more jumbled if we increase the polling frequency. Overall, this is fundamentally a pretty racy setup, and e.g. having stack traces turning up in the wrong spot in the terminal due to this re-ordering is super confusing.

That's the whole reason I implemented the who ProxyOutputStream thing in the first place, to ensure the ordering of chunks of data from stdout/stderr are serialized into the correct order. I think we should preserve that behavior, at least for interactive console usage, and we can get the best of both worlds by switching between combined/separate stdout/stderr streams in interactive/non-interactive use at the expense of a small amount of code in the Java client

lefou:
Looks like it's because we send some output (A-F) directly via the socket to the client and the subprocess-output (hello A-hello F) via the intermediate log file.

lihaoyi:
If we want to do this, we should be relying on subprocess forwarding to send all output to the log files, without relying on the `ioSocket` mechanism we have now. I suspect even in that case, we will see out-of-order interleaving of stdout/stderr as long as we continue to receive them as separate streams.

lefou:
After having another look, I like the way the ProxyOutputStream works. I think it should be easy to just put it's outcome into a file and use the pumper on an FileInputStream on client side (with some modificaiton to handle temporary EOF). So, I guess we're technically able to handle interleaved output streams via a file.

The issue is, that currently not all output is taking the same path. E.g. sub-processes started by the server cannot be filtered though ProxyOutputStream if this is done in the short-lived client (initServer) which later disappears. We need to setup this in the server instance, e.g. via System.setOut/Err. This is currently not the case. I started to play with it, but haven't finished it yet to my satisfaction, because of lack of time.

But honestly, I would even prefer the current PR over the solution in master, which shows no output from sub-processes at all. These kind of missing output issues popped up multiple times on gitter.

lihaoyi-databricks:
@lefou ok, I'm fine with merging this as is. It's a strict improvement over the status quo, and if I want to do further work to e.g. forward the stdout/stderr streams through the same set of log files that can be a separate future PR

Feel free to merge when you are ready

PR #742, opened 2019-12-09
Re-try starting mill server after detecting a changed version

lefou:
Fixes https://github.com/lihaoyi/mill/issues/557


lihaoyi:
looks good to me

PR #744, opened 2019-12-15
Non overlapping foreign module paths for #743

### LINKED ISSUE #743: Foreign modules which alter millSourcePath can result in non-unique output paths
Foreign modules which alter `millSourcePath` can result in non-unique output paths. This is the cause of a unwanted recompilations of a library which had not been modified. The output path of `js` and `jvm` submodules was being mapped to the same directory causing mill to make the unnecessary recompilations.

The setup is like this:
```scala
object commonlib extends Module {
  object jvm extends ScalaModule
    def millSourcePath = commonlib.millSourcePath
    def sources = T.sources( millSourcePath / "src", millSourcePath / "src-jvm" )
  }
  
  object js extends ScalaJSModule {
    def millSourcePath = commonlib.millSourcePath
    def sources = T.sources( millSourcePath / "src", millSourcePath / "src-js" )
  }
}
```

so that you have source directories of the form
```
commonlib/src/
commonlib/src-jvm/
commonlib/src-js/
```

and targets
```
commonlib.js.compile
commonlib.jvm.compile
```

Then in another project you reference the `commonlib.js` and `commonlib.js` module.

```scala
import $file.^.common.{build => commonlibbuild}, commonlibbuild.commonlib

object client extends ScalaJSModule {
  def moduleDeps = Seq(commonlib.js)
  ...
}

object server extends ScalaModule {
  def moduleDeps = Seq(commonlib.jvm)
  ...
}
```

Then the output directories in this project for `commonlib.js` and `commonlib.js` will be identical.
This is because `ctx.millSourcePath` is used as part of the path for foreign modules - not the segmented path name (see `destSegments()` in `Evaluator.scala`). 

It would be better to encode this as just the labeled task segments relative to the root of the foreign module as is done with non-foreign modules however it is not obvious how to get the root of a foreign module when just given a task (see `destSegments()`). One option might be to change `foreign` boolean of `Ctx` to be an `Option[Path]` which points to the root of a foreign module. 

Happy to do a PR if the approach is ok.

ajrnz:
Calculates foreign modules paths so that they do not overlap when millSourcePath is overridden

Precalculates `Ctx.foreign` which is now a Option[Segments] based on relative path to the foreign module rather than using `millSourcePath` which would not be unique where `millSourcePath`is overridden (eg when have have `commonlib.js` & `commonlib.jvm` which share a root source path)

Fixes #743

### DIFF
--- a/main/core/src/define/BaseModule.scala
+++ b/main/core/src/define/BaseModule.scala
@@ -7,7 +7,7 @@ object BaseModule{
 
 abstract class BaseModule(millSourcePath0: os.Path,
                           external0: Boolean = false,
-                          foreign0 : Boolean = false)
+                          foreign0 : Option[Segments] = None)
                          (implicit millModuleEnclosing0: sourcecode.Enclosing,
                           millModuleLine0: sourcecode.Line,
                           millName0: sourcecode.Name,
@@ -41,7 +41,7 @@ abstract class BaseModule(millSourcePath0: os.Path,
 abstract class ExternalModule(implicit millModuleEnclosing0: sourcecode.Enclosing,
                               millModuleLine0: sourcecode.Line,
                               millName0: sourcecode.Name)
-  extends BaseModule(ammonite.ops.pwd, external0 = true, foreign0 = false)(
+  extends BaseModule(ammonite.ops.pwd, external0 = true, foreign0 = None)(
     implicitly, implicitly, implicitly, implicitly, Caller(())
   ){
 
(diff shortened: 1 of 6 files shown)


Baccata:
It's been a while they were written, but there are unit-tests associated to foreign-modules. You should add a check that covers your use-case. See : 

https://github.com/lihaoyi/mill/blob/9ba4cb69331386dfde9bac69dc2d5b22401face3/main/test/src/main/ForeignBuildsTest.scala#L18

https://github.com/lihaoyi/mill/blob/902f8e01bc4b2eeac86c34526a7d454dbfc88556/main/test/resources/examples/foreign/project/build.sc#L10

ajrnz:
Added

Baccata:
@lihaoyi as far as I'm concerned this PR looks good

PR #747, opened 2019-12-20
Fixed link and updated docs for mill-ensime plugin

lefou:
I'm not the author of the mentioned plugin, but the existing link was dead, so I replaced it with one, which I *think* is the right one.

Review suggestions: @davoclavo


davoclavo:
👍 

The original author of the `mill-ensime` plugin deleted their repo, so I'll maintain it going onwards.

lefou:
@davoclavo Great. Thanks!

PR #749, opened 2019-12-24
Add T.source to listen to a single source file/folder

lihaoyi:
https://github.com/lihaoyi/mill/issues/546

Largely copy-pasted from `T.sources`, with all the `Seq`-related stuff removed

Not super heavily tested, but I stuffed a single usage into `JavacCompileJavaTests` to make sure it at least gets some coverage. Seems to work.

Review by @lefou


lihaoyi:
I guess it would, but it's a nice symmetry with `T.sources`. Apart from the existing Scala/Java rules, I find use cases for `T.source` much more common than `T.sources`.

`T.source` does some things where the path can be defined by an incoming task, copied from `T.sources`. I don't know if it's a good idea or not, but I figured I'd reproduce the behavior for consistency

lefou:
We also need a documentation update to reflect the additional target type.

PR #752, opened 2019-12-30
Upgrade Bloop support to current 1.4.0-RC1, fixes #751

### LINKED ISSUE #751: Bloop/install fails with NoClassDefFoundError
Attempting to run `mill.contrib.Bloop/install` on a project now always fails with:

```
java.lang.NoClassDefFoundError: io/circe/RootEncoder
        at mill.contrib.bloop.BloopImpl$BloopOps$bloop$.$anonfun$config$3(BloopImpl.scala:81)
```

This happens after dependencies are resolved and downloaded, both on a bigger project of mine and a fresh one (example `build.sc` below):

```scala
import mill._, scalalib._

import $ivy.`com.lihaoyi::mill-contrib-bloop:$MILL_VERSION`

object myproject extends ScalaModule {
  def scalaVersion = "2.12.10"
}
```

Selected `scalaVersion` does not seem to affect anything, which is expected, because it's a buildscript's dependency that causes trouble, not a module's one.

krotton:
_(It's my first contribution here, so I don't know whom to address. Hello!)_

### TL;DR:
1. Bloop bumped to the newest listed version and working again (#751),
2. Using Bloop's own serialization routines and working with a plain string on Mill's side.

### Rationale:
Bloop integration hasn't been working since 0.5.2, as mentioned by @ryneflood in #751. This is due to some binary changes in Bloop's circe dependencies. I've fixed this by dropping usage of circe in Bloop integration and relying on Bloop config's own `read` and `write` methods that encapsulate the whole serialization logic, whatever it is. This is the better way to handle it anyway, as Bloop has just dropped circe and started using jsoniter_scala, can do so back and forth on a whim and *we don't need to care here*. I've also bumped the Bloop's version to the newest listed one: `1.4.0-RC1`. I'll change it to `1.4.0` when it's released.


Baccata:
I don't want the API to change, because having access to bloop's config object unlocks mill delegating compilation to bloop (I haven't PR'd it in because I haven't found the time to automate tests for it) 

Changing the output type to string is not a good solution. If bloop has changed the json library it uses, we should just ditch the circe compat and find the easiest way to convert jsoniter json to ujson. Alternatively, derive ujson codecs via macro for each type involved in the Bloop config type tree (which is a little bit of boilerplate, but might take less time)

krotton:
OK, understood. Actually I've started by implementing ujson codecs for Bloop's config classes, but that was a lot of boilerplate and didn't seem necessary (right now, the generated JSON is not used in any way). Even if you do want to save the JSON output directly for bloop to consume in the future, couldn't we just dump it as a file to `dest/`?

Baccata:
>  right now, the generated JSON is not used in any way

thing is, you don't know that for sure : users (like I) might depend on this value already. It's exposed in the API because people (like I) might find it useful . Your change is breaking these people. I appreciate when my code doesn't get broken. I'm pretty sure other users do too. 

> Actually I've started by implementing ujson codecs for Bloop's config classes, but that was a lot of boilerplate 

The only boilerplate required should be 1 line per case class involved in the bloop config tree, thanks to the very useful [macroRW](https://github.com/lihaoyi/upickle/tree/master/upickle/test/src/upickle/example/ExampleTests.scala#L172-L174)

krotton:
Wouldn't macro generation of codecs require Bloop config classes' source to be pulled in? Currently bloop-config is just a binary dependency.

Baccata:
> Wouldn't macro generation of codecs require Bloop config classes' source to be pulled in? Currently bloop-config is just a binary dependency.

The macro doesn't need the source code, you can use it on case classes provided by third party libraries 😉

krotton:
OK, I can change that then.

jodersky:
Any way we can help with this? It appears that the latest metals release now requires Bloop 1.4, and hence also the changes of this PR

jodersky:
I fixed conflicts ans added upickle serializers in #756

krotton:
Great, thanks! I couldn't find the time for the last few days. Closing this PR then.

PR #755, opened 2020-01-14
update docs related to Metals

ckipp01:
With the new stable release of Metals, 0.8.0, there is no need for a user to include the plugin in order to generate the necessary files. It will all happen automatically.


lefou:
Thanks!

PR #756, opened 2020-01-14
Upgrade and fix bloop integration

jodersky:
This builds on top of #752 (thanks @krotton), mainly adding upickle serializers for the bloop config task (as suggested by @Baccata).

I tested it locally, and witth scalameta/metals-vscode#175 also fixed, bloop once again works out-of-the-box for mill projects!


Baccata:
Looks great, many thanks @jodersky !

jodersky:
@lihaoyi or @lefou, would it be possible to create an official release, now that this fix is in?

Baccata:
@jodersky mill gets released for each merge to master. You should be able to set your mill version to `0.5.9-6-484657` and get these changes

jodersky:
Good to know, thanks!

lefou:
http://www.lihaoyi.com/mill/#overriding-mill-versions

PR #758, opened 2020-01-16
Auto-update dependencies: output info in machine-readable format

jodersky:
Mill has a convenient task to check for the availability of newer versions of dependencies `mill.scalalib.Dependency/updates`.

Unfortunately this task only prints the output (in human readable form), but does not persist it in a machine-readable format. This makes it very difficult to reuse the output in other tasks, essentially requiring regex parsing of strings.

The changes proposed here change the `updates` task to expose the underlying data structures representing the upgrade path. An additional task, `showUpdates` is added, implementing the original, human readable, output of updates.

### DIFF
--- a/scalalib/src/Dependency.scala
+++ b/scalalib/src/Dependency.scala
@@ -18,6 +18,10 @@ object Dependency extends ExternalModule {
         allowPreRelease)
     }
 
+  def showUpdates(ev: Evaluator, allowPreRelease: Boolean = false) = T.command{
+    DependencyUpdatesImpl.showAllUpdates(updates(ev, allowPreRelease)())
+  }
+
   implicit def millScoptEvaluatorReads[T]: EvaluatorScopt[T] =
     new mill.main.EvaluatorScopt[T]()
   lazy val millDiscover: Discover[Dependency.this.type] = Discover[this.type]
--- a/scalalib/src/dependency/versions/ModuleDependenciesVersions.scala
+++ b/scalalib/src/dependency/versions/ModuleDependenciesVersions.scala
@@ -3,7 +3,7 @@ package mill.scalalib.dependency.versions
 import mill.scalalib.JavaModule
 
 private[dependency] final case class ModuleDependenciesVersions(
-    module: JavaModule,
+    modulePath: String,
     dependencies: Seq[DependencyVersions])
 
 private[dependency] final case class DependencyVersions(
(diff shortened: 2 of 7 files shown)


scalalib/src/dependency/versions/ModuleDependenciesVersions.scala
```diff
@@ -3,10 +3,10 @@ package mill.scalalib.dependency.versions
 import mill.scalalib.JavaModule
 
 private[dependency] final case class ModuleDependenciesVersions(
-    module: JavaModule,
```
lefou: What's the motivation for this change?

scalalib/src/dependency/versions/ModuleDependenciesVersions.scala
```diff
@@ -3,10 +3,10 @@ package mill.scalalib.dependency.versions
...
+    modulePath: String,
     dependencies: Seq[DependencyVersions])
 
 private[dependency] final case class DependencyVersions(
-    dependency: coursier.Dependency,
```
lefou: Wouldn't be `coursier.Dependency` the better type for later inspection of possible updates?

scalalib/src/dependency/versions/ModuleDependenciesVersions.scala
jodersky: The issue is that `JavaModule` is not picklable, hence it can't be used in the updates task's result

scalalib/src/dependency/versions/ModuleDependenciesVersions.scala
jodersky: yes indeed. I wasn't sure if coursier should be exposed by mill, so I used a string instead. If it's ok to leak coursier, then I'd be happy to use that instead

lefou: I guess, mill already leaks `coursier.Dependency` as it's an public property of `mill.scalalib.Dep`. So I think it would be ok expose `coursier.Dependency` here. We currently have no other abstraction / API for managed dependencies.

scalalib/src/dependency/versions/ModuleDependenciesVersions.scala
lefou: Yeah, that makes sense.

jodersky:
@lefou, I addressed your comments

scalalib/src/Dependency.scala
```diff
@@ -18,6 +18,10 @@ object Dependency extends ExternalModule {
         allowPreRelease)
     }
 
+  def showUpdates(ev: Evaluator, allowPreRelease: Boolean = false) = T.command{
```
lefou: Can you please add some ScalaDoc to the new command. It would be great, if you could also update/add the comment of the changed `updates` command.

lefou:
Thank you!

PR #765, opened 2020-01-27
Add Scala Native 0.4.0-M2 support

lolgab:


scalalib/api/src/ZincWorkerApi.scala
```diff
@@ -85,23 +85,28 @@ object Util {
...
+  def scalaJSNativeBinaryVersion(version: String) = version match {
+      case ScalaJSNativeFullVersion(major, minor, patch, suffix) =>
+        if (suffix != null && suffix.startsWith("-M") && patch == "0")
+          version
+        else if (suffix != null && suffix.startsWith("-RC") && patch == "0")
           s"$major.$minor$suffix"
         else
-          major
+          s"$major.$minor"
+  }
```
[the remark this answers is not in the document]
lolgab: Thank you for your review @sjrd, I added a new commit that should fix this.

lefou:
Travis fails (even after some restarts), so I ran the test locally, but got failures too. Are there some additional requirements?

lefou:
> Are there some additional requirements?

Oh, never mind. I overlooked the setup script. Sorry, I can't reproduce a successful test setup for now. Maybe, others can test/comment? Or you can fix the travis setup?

lolgab:
Thank you @lefou.
I see some flakiness in Scala Native tests, because it seems the logger is trying to read the standard output before it is flushed (or something like that).
By the way it failed one time with the 0.3.9 version (that was already there) and another with the version 0.4.0-M2.
In this PR I added the Scala Native tests in the CI script because they were not there.
I think this PR doesn't change the status quo of test flakiness, but, at least, it doesn't seem to create regressions.

lolgab:
There was a test hanging forever because the Scala Native linker takes a long time (and a lot of memory) to optimise in release mode, when a lot of classes are compiled (in this case scalatest was used). So I deactivated the test in release mode, since the release mode itself is already covered somewhere else.

lolgab:
I rebased the PR on top of current master.
My last commit should have solved the wrong test that was failing before.

lolgab:
✅It is green on Travis 🎉

scalanativelib/test/src/HelloNativeWorldTests.scala
```diff
@@ -169,31 +175,35 @@ object HelloNativeWorldTests extends TestSuite {
...
       val paths = Evaluator.resolveDestPaths(
         helloWorldEvaluator.outPath,
         task.ctx.segments
       )
-      val log = os.read(paths.log)
-      assert(
-        evalCount > 0,
-        log.contains("Scala Native")
-      )
+      val stdout = ammonite.ops.%%("./out")(paths.out / 'dest).out.lines
```
lefou: I would be great if you could use `os.proc` instead of `ammonite.ops` here.

lolgab: ✅Done

lolgab: Added again the test on `evalCount > 0` I had removed by mistake.

lolgab:
> I'd like to have a green Travis-CI result. I'll restart flaky jobs (e.g. those envVars tests fail regularly on travis)

The envVars tests failed again. Can you rerun those jobs?

lefou:
Thank you!

lolgab:
Thank you very much @lefou and @sjrd.
So happy it was merged 🎉!

Issue #671, opened 2019-08-02
AppendEntry assembly's can cause incorrect syntax in .conf files


sloanesturz:
My mill build is breaking when I include a dependency on a jar that has a reference.conf file that lacks a trailing newline. This issue is best explained with an example.

Module A:
```
a {
  c = "foo"
}
```
Module B:
```
b {
  d = "wat"
}
```
If there is no newline at the end of a/reference.conf, you can get this, which misses a newline between `}` and `b`:

```
a {
  c = "foo"
} b {
  d = "wat"
}
```

Which typesafe config won't parse.

I believe I've tracked down the issue to the `mill.modules.Jvm` object's method `createAssembly`. An `AppendEntry` should probably add in a newline between each entry. Unfortunately, overriding this as a client of mill is a bit tricky because AppendEntry is a private case class.

``` Assembly.groupAssemblyEntries(inputPaths, assemblyRules).view
      .foreach {
        case (mapping, AppendEntry(entries)) =>
          val path = zipFs.getPath(mapping).toAbsolutePath
          val concatenated = new SequenceInputStream(
            Collections.enumeration(entries.map(_.inputStream).asJava)) // <- I want a newline at the end of each `_.inputStream`
          writeEntry(path, concatenated, append = true)
        case (mapping, WriteOnceEntry(entry)) =>
          val path = zipFs.getPath(mapping).toAbsolutePath
          writeEntry(path, entry.inputStream, append = false)
      }```

emstlk:
Have gotten the same problem, here is PR #768

lefou:
This bug is possibly someone else's feature. So a proper fix might introduce a option to configure the wanted behavior.

PR #768, opened 2020-02-06
Fix merge files in assembly

emstlk:


emstlk:
Updated pr for #671

main/src/modules/Jvm.scala
```diff
@@ -283,10 +283,13 @@ object Jvm {
 
     Assembly.groupAssemblyEntries(inputPaths, assemblyRules).view
       .foreach {
-        case (mapping, AppendEntry(entries)) =>
+        case (mapping, AppendEntry(entries, separator)) =>
           val path = zipFs.getPath(mapping).toAbsolutePath
+          val separated = entries.flatMap { e =>
+            List(e, JarFileEntry(e.mapping, () => new ByteArrayInputStream(separator.getBytes)))
+          }
```
lefou: Looks like the separator is also added after the last entry. I'd rather keep a naive separator semantic in terms of "separate" and avoid adding it to the end.

Imagine the case where exactly one "reference.conf" is found. The above impl. would change the file contents (and add a separator) instead of just keeping it as-is.

emstlk: Hm, yes I agree with you

lefou:
Thank you!

PR #771, opened 2020-02-11
Added options to include compile and runtime deps to ivyDepsTree

lefou:


scalalib/src/JavaModule.scala
```diff
@@ -394,6 +399,29 @@ trait JavaModule extends mill.Module
...
 
+  /**
+   * Command to print the transitive dependency tree to STDOUT.
+   *
+   * @param inverse Invert the tree representation, so that the root is on the bottom.
+   * @param withCompile Include the compile-time only dependencies (`compileIvyDeps`, provided scope) into the tree.
+   * @param withRuntime Include the runtime dependencies (`runIvyDeps`, runtime scope) into the tree.
+   */
+  def ivyDepsTree(inverse: Boolean = false, withCompile: Boolean = false, withRuntime: Boolean = false): Command[Unit] =
+    (withCompile, withRuntime) match {
```
lefou: We do the match before constructing the command, to avoid unnecessary dependencies.

PR #772, opened 2020-02-12
Allow disabling lenses on the ScalaPBModule

AlexITC:


contrib/scalapblib/src/ScalaPBModule.scala
```diff
@@ -33,6 +33,8 @@ trait ScalaPBModule extends ScalaModule {
 
   def scalaPBSingleLineToProtoString: T[Boolean] = T { false }
 
+  def scalaPBLenses: T[Boolean] = T { true }
```
lefou: Some ScalaDoc would be nice.

AlexITC: Done, thanks.

contrib/scalapblib/src/ScalaPBModule.scala
```diff
@@ -33,6 +33,7 @@ trait ScalaPBModule extends ScalaModule {
 
   def scalaPBSingleLineToProtoString: T[Boolean] = T { false }
 
+  // ScalaPB enables lenses by default, this option allows you to disable it
```
lefou:
```suggestion
  /** ScalaPB enables lenses by default, this option allows you to disable it. */
```

Otherwise, it is only a code comment, but not ScalaDoc.

lefou:
Thank you!

PR #773, opened 2020-02-14
Fix Scala Native for testing libraries

lolgab:
- Add missing test-runner in Scala Native tools

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -468,16 +468,8 @@ object scalanativelib extends MillModule {
 
   def testArgs = T{
     val mapping = Map(
-      "MILL_SCALANATIVE_WORKER_0_3" ->
-        worker("0.3").runClasspath()
-          .map(_.path)
-          .filter(_.toIO.exists)
-          .mkString(","),
-      "MILL_SCALANATIVE_WORKER_0_4" ->
-        worker("0.4").runClasspath()
-          .map(_.path)
-          .filter(_.toIO.exists)
-          .mkString(",")
+      "MILL_SCALANATIVE_WORKER_0_3" -> worker("0.3").compile().classes.path,
+      "MILL_SCALANATIVE_WORKER_0_4" -> worker("0.4").compile().classes.path
     )
     scalalib.worker.testArgs() ++
     scalalib.backgroundwrapper.testArgs() ++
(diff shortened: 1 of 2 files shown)


lolgab:
Last PR I did, I broke the Scala Native build for testing libraries.. That's because Scala Native, starting from version 0.3.8, has a separate artifact for running tests, `test-runner`. Upgrading I didn't add it.
Made also a cleanup in the `build.sc` file.

build.sc
```diff
@@ -488,15 +480,15 @@ object scalanativelib extends MillModule {
...
     def moduleDeps = Seq(scalanativelib.api)
     def ivyDeps = Agg(
-      ivy"org.scala-native::tools:${scalaNativeVersion()}",
-      ivy"org.scala-native::util:${scalaNativeVersion()}",
-      ivy"org.scala-native::nir:${scalaNativeVersion()}",
-      ivy"org.scala-native::test-runner:${scalaNativeVersion()}",
+      ivy"org.scala-native::tools:$scalaNativeVersion",
+      ivy"org.scala-native::util:$scalaNativeVersion",
+      ivy"org.scala-native::nir:$scalaNativeVersion",
+      ivy"org.scala-native::test-runner:$scalaNativeVersion",
```
lefou: Can you move these concrete deps up into the `Deps` object?

scalanativelib/src/ScalaNativeModule.scala
```diff
@@ -50,7 +50,8 @@ trait ScalaNativeModule extends ScalaModule { outer =>
     Seq(
       ivy"org.scala-native:tools_2.12:${scalaNativeVersion()}",
       ivy"org.scala-native:util_2.12:${scalaNativeVersion()}",
-      ivy"org.scala-native:nir_2.12:${scalaNativeVersion()}"
+      ivy"org.scala-native:nir_2.12:${scalaNativeVersion()}",
+      ivy"org.scala-native:test-runner_2.12:${scalaNativeVersion()}"
```
lefou: It looks like this "test" dependency ends up in a non-test project? My assumption would be to add a test dep only in test projects.

build.sc
lolgab: Sure :)

scalanativelib/src/ScalaNativeModule.scala
lolgab: These are the dependencies only for linking (they are not Scala Native dependencies but Scala JVM 2.12 ones). Even in Scala JS there are test dependencies in the LinkerClasspath:
https://github.com/lihaoyi/mill/blob/c7873b6f2d977324f215209854a1f415e2882622/scalajslib/src/ScalaJSModule.scala#L40

build.sc
lolgab: Done ✅

scalanativelib/src/ScalaNativeModule.scala
lefou: Thanks for clarification.

lefou:
Thanks!

lolgab:
> Thanks!

Thanks to you for having reviewed this so fast! :)
You are a great maintainer!

PR #775, opened 2020-02-14
Support customizing coursier dependency resolution (make OS specific classifiers work)

### LINKED ISSUE #767: Can't compile JavaFX projects (regression?)
More precisely, since commit a1247748926f9af95a2698de4692d3c167b8775a, a simple JavaFX project cannot compile and log the following errors:

```[19/29] javafx.compile
[info] Compiling 1 Java source to ./out/javafx/compile/dest/classes ...
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by sbt.internal.inc.javac.DiagnosticsReporter$PositionImpl$ (file:/~/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-sbt/zinc-compile-core_2.12/1.2.5/zinc-compile-core_2.12-1.2.5.jar) to field com.sun.tools.javac.api.ClientCodeWrapper$DiagnosticSourceUnwrapper.d
WARNING: Please consider reporting this to the maintainers of sbt.internal.inc.javac.DiagnosticsReporter$PositionImpl$
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
[error] ./javafx/src/ButtonApp.java:1:1: package javafx.application does not exist
[error] import javafx.application.Application;
[error] ./javafx/src/ButtonApp.java:2:1: package javafx.stage does not exist
[error] import javafx.stage.Stage;
…
```
Basically it cannot find the JavaFX packages.

Latest commit tested: 80eb7e0df5132dff7b081247176bfae597fe4264
Tested with java-11-graalvm and java-11-openjdk.
Was working as expected before a1247748926f9af95a2698de4692d3c167b8775a.

### Sample code

`cat build.sc`
```scala
import mill._, scalalib._

val javaFXVersion = "13.0.2"

object javafx extends JavaModule {
  def mainClass = Some("Main")
  def ivyDeps = Agg(ivy"org.openjfx:javafx-controls:$javaFXVersion")
}
```

`cat javafx/src/ButtonApp.java`
```java
import javafx.application.Application;
import javafx.stage.Stage;
import javafx.scene.Scene;
import javafx.scene.layout.VBox;
import javafx.scene.control.Button;

public class ButtonApp extends Application {
  public static void main(String[] args) {
    launch(args);
  }

  @Override public void start(Stage stage) {
    var button = new Button("Please Click Me!");
    button.setOnAction(e -> stage.close());
    var box = new VBox();
    box.getChildren().add(button);
    var scene = new Scene(box);
    stage.setScene(scene);
    stage.setTitle("Just a button");
    stage.show();
  }
}
```

`cat javafx/src/Main.java`
```java
public class Main {

  public static void main(String[] args) {
    ButtonApp.main(args);
  }
}
```

lefou:
This should fix #767.

TBH, I'm not sure we want this in mill. So I encourage a review and/or discussion.

On one side this fixes an issue with resolving transitive dependencies with os-specific classifiers (#767). But on the other side it makes dependency resolution less reproducible.

So a perfect outcome of this PR for me is also a rejection paired with some documentation how to deal with such situations.

### DIFF
--- a/main/src/modules/Jvm.scala
+++ b/main/src/modules/Jvm.scala
@@ -474,6 +474,7 @@ object Jvm {
     PathRef(outputPath)
   }
 
+
   /**
     * Resolve dependencies using Coursier.
     *
@@ -486,10 +487,11 @@ object Jvm {
                           force: IterableOnce[coursier.Dependency],
                           sources: Boolean = false,
                           mapDependencies: Option[Dependency => Dependency] = None,
-                          ctx: Option[mill.util.Ctx.Log] = None): Result[Agg[PathRef]] = {
+                          customizer: Option[coursier.core.Resolution => coursier.core.Resolution] = None,
+                          ctx: Option[mill.api.Ctx.Log] = None): Result[Agg[PathRef]] = {
 
     val (_, resolution) = resolveDependenciesMetadata(
-      repositories, deps, force, mapDependencies, ctx
+      repositories, deps, force, mapDependencies, customizer, ctx
     )
     val errs = resolution.errors
 
@@ -552,12 +554,12 @@ object Jvm {
     }
   }
 
-
   def resolveDependenciesMetadata(repositories: Seq[Repository],
                                   deps: IterableOnce[coursier.Dependency],
                                   force: IterableOnce[coursier.Dependency],
                                   mapDependencies: Option[Dependency => Dependency] = None,
-                                  ctx: Option[mill.util.Ctx.Log] = None) = {
+                                  customizer: Option[coursier.core.Resolution => coursier.core.Resolution] = None,
+                                  ctx: Option[mill.api.Ctx.Log] = None): (Seq[Dependency], Resolution) = {
 
     val cachePolicies = coursier.cache.CacheDefaults.cachePolicies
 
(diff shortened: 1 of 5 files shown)


lefou:
If we decide, to make resolution OS specific, I could imagine some dedicated target in `CoursierModule` to fine control OS specific resolution, e.g. to still allow cross platform projects.

lihaoyi:
Having a dedicated target on `CoursierModule` you can override seems like the right thing to do. 

It could default to `coursier.core.Activation.Os.fromProperties(sys.props.toMap)` by default, but you could override it if you want it to have some other value (e.g. you are on OSX but deploying to Linux) and you could override it with a `T.input(T.ctx.env("..."))` if you want it to be configurable from the command line as an environment variable

lefou:
I'm going to update this PR. This is the chance I have in mind: I'd like to add a `resolutionCustomizer` task to `CoursierModule`.

The idea is to override it in cases you want to tweak the coursier resolution process in your module.

For example, to make JavaFX compilation work, you need to override it like this:

```scala
import mill._
import mill.scalalib._

object javafx extends JavaModule {
  override def resolutionCustomizer = T.task {
    Some((_: coursier.core.Resolution).withOsInfo(coursier.core.Activation.Os.fromProperties(sys.props.toMap)))
  }
  def ivyDeps = Agg(ivy"org.openjfx:javafx-controls:13.0.2")
}
```

ajrnz:
I've just tested this PR out and it works correctly. Much nicer. Thanks.

lefou:
> I've just tested this PR out and it works correctly. Much nicer. Thanks.

Thanks for the feedback. I hesitate to merge this PR right now as I have no test set up for it. But it seems there is quiet a demand for this feature and (besides all other tests succeeding) your comment suggest that it works as intended. If @lihaoyi or @alexarchambault have no other objects regarding the proposed new `resolutionCustomizer` task in `CoursierModule`, I think we can merge as-is.

alexarchambault:
It's fine to allow users to customize the `Resolution`, IMO. The only possible issue is it ties the mill API to the one of coursier even more, which could be an issue if the API of coursier were to change. But it's less likely to happen since coursier switched to [data-class](https://github.com/alexarchambault/data-class).

lefou:
@alexarchambault Thank you for your input. I think I'm going to merge this PR.

As a matter of fact, we already depend on Coursier API, and I think it makes not much sense to introduce some abstraction for this concrete configuration point.

hmf:
Is this available in the `0.9.6-51-e4c838`? 
I tried it and it failed to execute with the error:

```
Error: JavaFX runtime components are missing, and are required to run this application
```

Please note that with unmanaged jars, the test works.

PR #782, opened 2020-03-02
Improve Idea project generator

lefou:
This pull requests contains the following features:
* GenIdea no longer includes source JARs in the module classpath, which removes many (all) "Problems" visible in "Project Structure" view
* Generated library files now conform to what IDEA expects
* Better guess dependency sources, when they come from Maven repositories


lefou:
Typical Travis CI hiccup. But mine succeeded: https://travis-ci.org/lefou/mill/builds/657258353

PR #791, opened 2020-03-15
Fix Windows launcher

sake92:
There are 2 problems this PR tries to fix:
1. production version of Mill, built by `assembly` task
2. development version of Mill, built by `dev.launcher` task

These problems manifest with java 1.8.  
In Hotspots 9+ there exists an option `-XX:VMOptionsFile`...

### Problem
- Download latest Mill (0.6.1-25-360107-assembly) and rename it to `mill.bat`
- unzip "example-1.zip" project from docs
- run it with downloaded Mill as `mill foo.compile`
First time you run it you'll get this message:
> Error: Could not create the Java Virtual Machine.
> Error: A fatal exception has occurred. Program will exit.
> Unrecognized VM option 'VMOptionsFile=C:\projects\TODO\mill\winfix\example-1\out\mill-worker-+9h7erhXUQxlj6SW7zBIF3jvnFg=-1\vmoptions'

If you kill it and **try again, it just hangs**...  
I couldnt attach VisualVM for some reason to debug further. 
I guess it cant start a server and pipe between client-server breaks!?

### Fix
Remove `-XX:VMOptionsFile` stuff.  
This is enough to get working Mill for Windows users! (even on java 1.8)

### Testing
Build Mill with `publish-local.sh` commands.  
Then use built Mill version on projects. See that it works. 😄 

## Development version (`dev.launcher`)
Here is where it gets interesting.

### Problem
- Clone Mill repository and build its dev version with `mill -i dev.launcher`
- try to run some command with it, you'll get:
>dev.launcher java.lang.Error: dev.launcher in Windows is only supported using Java 9 or above
>ammonite.$file.build$dev$.windowsVmOptions(build.sc:628)
(this was using the "mill.vmoptions" file passed via `-XX:VMOptionsFile`)

Why? Because `-XX:VMOptionsFile` doesnt exist in java 1.8...  
Why is it needed in first place? Because Windows has command-line **limitation of only ~8000 characters**!  
Most lengthy operations are `MILL_STUFF` and classpath, of course (lot of JARs from Coursier local repo...)

### Fix
1. introduce `MILL_OPTIONS_PATH=$path/mill.properties` to bypass long `-DMILL_..` props
2. introduce a Pathing JAR (see [here](http://todayguesswhat.blogspot.com/2011/03/jar-manifestmf-class-path-referencing.html) for more details) to bypass long `-cp` parameter

### Testing
Build `dev.launcher` and see it works in client-server mode.  
Only thing that isnt working on Windows is `-i` mode of `dev.launcher`:
>scala.reflect.internal.MissingRequirementError: object scala in compiler mirror not found...

---
Minor NOTE: Once you do `-i` it stays during that shell session (`SETLOCAL` wasn't used), so if you run `mill` after that it will still use interactive mode... This PR fixes that also.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -641,7 +641,7 @@ object dev extends MillModule{
 
   def prependShellScript = T{
     val (millArgs, otherArgs) = forkArgs().partition(arg => arg.startsWith("-DMILL") && !arg.startsWith("-DMILL_VERSION"))
-    // Pass dev.assembly VM options via file in Windows due to small max args limit
+    // Pass Mill options via file, due to small max args limit in Windows
     val vmOptionsFile = T.ctx.dest / "mill.properties"
     val millOptionsContent = millArgs.map(_.drop(2).replace("\\", "/")).mkString("\r\n") // drop -D prefix, replace \ with /
     os.write(vmOptionsFile, millOptionsContent)
--- a/build.sc
+++ b/build.sc
@@ -791,4 +791,3 @@ def uploadToGithub(authKey: String) = T.command{
 
   upload.apply(launcher().path, releaseTag, label, authKey)
 }
-
--- a/build.sc
+++ b/build.sc
@@ -643,9 +643,9 @@ object dev extends MillModule{
     val (millArgs, otherArgs) = forkArgs().partition(arg => arg.startsWith("-DMILL") && !arg.startsWith("-DMILL_VERSION"))
     // Pass dev.assembly VM options via file in Windows due to small max args limit
     val vmOptionsFile = T.ctx.dest / "mill.properties"
-    val millOptionsPath = // drop -D prefix, replace \ with /
-      os.write(vmOptionsFile, millArgs.map(_.drop(2).replace("\\", "/")).mkString("\r\n"))
-    val jvmArgs = otherArgs ++ List(s"-DMILL_OPTIONS_PATH=$millOptionsPath")
+    val millOptionsContent = millArgs.map(_.drop(2).replace("\\", "/")).mkString("\r\n") // drop -D prefix, replace \ with /
+    os.write(vmOptionsFile, millOptionsContent)
+    val jvmArgs = otherArgs ++ List(s"-DMILL_OPTIONS_PATH=$vmOptionsFile")
     val classpath = Agg(pathingJar().path.toString)
     launcherScript(
       jvmArgs,
--- a/build.sc
+++ b/build.sc
@@ -620,27 +620,13 @@ object dev extends MillModule{
       )
     ).distinct
 
-
-  // Pass dev.assembly VM options via file in Windows due to small max args limit
-  def windowsVmOptions(taskName: String, batch: os.Path, args: Seq[String])
-                      (implicit ctx: mill.util.Ctx) = {
-    if (System.getProperty("java.specification.version").startsWith("1.")) {
-      throw new Error(s"$taskName in Windows is only supported using Java 9 or above")
-    }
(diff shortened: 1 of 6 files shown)


build.sc
```diff
@@ -650,30 +636,43 @@ object dev extends MillModule{
     val isWin = scala.util.Properties.isWin
     val millPath = T.ctx.dest / (if (isWin) "mill.bat" else "mill")
     os.move(super.assembly().path, millPath)
-    if (isWin) windowsVmOptions("dev.launcher", millPath, forkArgs())
     PathRef(millPath)
   }
 
   def prependShellScript = T{
+    val (millArgs, otherArgs) = forkArgs().partition(arg => arg.startsWith("-DMILL") && !arg.startsWith("-DMILL_VERSION"))
```
sake92: We pass the `MILL_SOMETHING` properties via file, so that Windows doesn't complain!  
Everything except `MILL_VERSION` because it's used in lots of places.

build.sc
```diff
@@ -650,30 +636,43 @@ object dev extends MillModule{
...
-        Seq("""-XX:VMOptionsFile=%~dp0\mill.vmoptions""")
-      )
     launcherScript(
-      shellArgs,
-      cmdArgs,
+      jvmArgs,
+      jvmArgs,
       classpath,
-      classpath
+      Agg(pathingJar().path.toString) // TODO not working yet on Windows! see #791
```
sake92: On Windows, the classpath is only the "pathingJar", it contains all the dependencies.
It works in batch mode, interactive needs bit more debugging.

main/client/src/mill/main/client/MillClientMain.java
```diff
@@ -19,32 +19,43 @@
...
+        
+        String selfJars = "";
         List<String> vmOptions = new ArrayList<>();
-        l.add(System.getProperty("java.home") + File.separator + "bin" + File.separator + "java");
-        final Properties props = System.getProperties();
-        for(final String k: props.stringPropertyNames()){
+        String millOptionsPath = System.getProperty("MILL_OPTIONS_PATH");
+        if(millOptionsPath != null) {
+            // read MILL_CLASSPATH from file MILL_OPTIONS_PATH
+            Properties millProps = new Properties();
```
sake92: On Windows we read the classpath from the `MILL_OPTIONS_PATH` file instead of env var..

lihaoyi:
@sake92 Is this ready to review? If so I can take a look

sake92:
@lihaoyi yes please!

build.sc
sake92: @lihaoyi this only relates to the `dev.assembly` version..

lihaoyi:
@sake92 this looks good to me I think.

Before I merge, could you flesh out the PR description a bit to summarize (1) exactly which user-facing problems you are fixing, (2) what the root cause of those problems is, (3) how your PR solves those root causes and (4) how you tested it to verify your PR solves the things you think it solves? I'm still a bit fuzzy on the different problems with dev, assembly, dev.assembly, dev.launcher, etc. and which ones I should expect to behave differently due to this PR. This will also help future maintenance.

Once that's done I'm happy to merge!

sake92:
@lihaoyi I updated the description, hope it's clearer now.  
In summary:
- prod version will see almost no changes, except it'll work on Windows now also
- dev.launcher will use `mill.properties` instead of `-DMILL_PROP` properties.   
And on Windows it'll use "pathingJar" instead of "normal classpath"

lihaoyi:
This looks good, thanks for the detailed PR description!

PR #798, opened 2020-03-20
Bintray support

heksesang:


heksesang:
We might want to also add an option to override what package name it should use on Bintray. Currently it's using `artifact.id`, but we might want to change that so it has a `def bintrayPackage(artifact: Artifact) = artifact.id` in `BintrayPublishModule` which can be overridden if wanted.

lefou:
Can you please resolve the merge conflict and rebase the branch to latest master, to trigger a clean Travis build?

lefou:
Thanks for your contribution!

PR #800, opened 2020-03-22
New VersionFile module

heksesang:
A module to help set release version and update to next snapshot when publishing.

### DIFF
--- a/contrib/versionfile/src/mill/contrib/versionfile/VersionFileModule.scala
+++ b/contrib/versionfile/src/mill/contrib/versionfile/VersionFileModule.scala
@@ -6,12 +6,13 @@ trait VersionFileModule extends Module {
 
   implicit val wd = os.pwd
 
-  def versionFileName = "version"
-  def versionFileDirectory = wd
-  def versionFilePath = versionFileDirectory / versionFileName
-
-  def currentVersion = T.input { Version.of(os.read(versionFilePath)) }
+  /** The file containing the current version. */
+  def versionFile: define.Source = T.source(millSourcePath / "version")
+  /** The current version. */
+  def currentVersion: T[Version] = T { Version.of(os.read(versionFile().path)) }
+  /** The release version. */
   def releaseVersion = T { currentVersion().asRelease }
+  /** The next snapshot version. */
   def nextVersion(bump: String) = T.command { releaseVersion().bump(bump) }
 
   def setReleaseVersion = T {
@@ -20,7 +21,7 @@ trait VersionFileModule extends Module {
     T.ctx.log.info(commitMessage)
 
     os.write.over(
-      versionFilePath,
+      versionFile().path,
       releaseVersion().toString
     )
 
@@ -34,7 +35,7 @@ trait VersionFileModule extends Module {
     T.ctx.log.info(commitMessage)
 
     os.write.over(
-      versionFilePath,
+      versionFile().path,
       nextVersion(bump).toString
     )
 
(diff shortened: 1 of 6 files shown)


heksesang:
I'll try add the suggested changes. :)

Is there any documentation anywhere on what the different `T` types are and how they work?

And about the name, "versionFile" is the only of your suggestions that I think makes it clear that it's not actually doing any release action. Though, if we also add support for using the git tag instead of a version file, is that name a bit misleading again?

lefou:
The different Task types are documented here: http://www.lihaoyi.com/mill/page/tasks.html

Yeah, I guess "versionFile" for the current PR is the best. If you want to integrate the git idea it won't fit, I agree. But the git idea was only a pointer or suggestion. Do what you want.

heksesang:
I'll change it to versionFile and not add the git idea in this PR. The git idea should probably be made into some module, but maybe it should be its own module? That would make naming easier.

contrib/versionfile/src/mill/contrib/versionfile/VersionFileModule.scala
```diff
@@ -0,0 +1,44 @@
...
+
+trait VersionFileModule extends Module {
+
+  implicit val wd = os.pwd
+
+  def versionFileName = "version"
+  def versionFileDirectory = wd
+  def versionFilePath = versionFileDirectory / versionFileName
+
+  def currentVersion = T.input { Version.of(os.read(versionFilePath)) }
```
lefou:
```suggestion
  /** The file containing the current version. */
  def versionFile: Source = T.source(millSourcePath / "version")
  /** The current version. */
  def currentVersion: T[Version] = T { Version.of(os.read(versionFilePath().path)) }
```

That way, we avoid reading the file if it hasn't changed. It looks also a bit more idiomatic (if that is even a thing).

heksesang: Won't `millSourcePath / "version"` prevent one from having the version file at root level of the repository?

lefou: Yeah. It's an oversight from me. But `wd` isn't mill API. You can either override `millSourcePath` or you can use the value `millOuterCtx.millSourcePath`.

docs/pages/9 - Contrib Modules.md
```diff
@@ -845,3 +844,118 @@ These imports will always be added to every template.  You don't need to list th
...
+}
+```
+
+### Configure the version file
+This module reads and updates a version file. The default is the file `version` at the project root.
+
+If you prefer something else than the default, you can override:
+* `VersionFileModule.versionFileName` which defaults to `version`
+* `VersionFileModule.versionFileDirectory` which defaults to `os.pwd`
+* `VersionFileModule.versionFilePath` which defaults to `versionFileDirectory / versionFileName`
```
heksesang:
```suggestion
This module reads and updates a version file. The default is the file `version` at the module source root. If you prefer something other than the default, you can override `VersionFileModule.versionFile`.
```

contrib/versionfile/src/mill/contrib/versionfile/VersionFileModule.scala
```diff
@@ -0,0 +1,44 @@
...
+  def releaseVersion = T { currentVersion().asRelease }
+  def nextVersion(bump: String) = T.command { releaseVersion().bump(bump) }
```
heksesang:
```suggestion
  /** The release version. */
  def releaseVersion = T { currentVersion().asRelease }
  /** The next snapshot version. */
  def nextVersion(bump: String) = T.command { releaseVersion().bump(bump) }
```
Might as well add comments to these two too. :)

lefou:
Thank you!

PR #802, opened 2020-03-25
Added () to nextVersion(bump) calls in VersionFileModule.

heksesang:


lefou:
Thanks. Merging this, as it fixed a bug in the current codebase.

It would be cool if you can come up with another PR adding some tests. As discussed on gitter, it might be an option to abstract over the git-calls via mill commands, so it would be possible to change the used VCS or even don't use any at all. In the tests, you could instead hook some validation logic.

heksesang:
In the refactored version that I am working on, I have `def tag: T[Seq[os.proc]] = ...`, and to actually execute them you'd do `mill versionFile.execute --procs versionFile.tag` for example. The idea is to make the whole thing a bit more flexible, and also allow to have tests that check that the correct commands are generated without having to test the execution of them.

Just need to add the tests, might make a PR today or tomorrow if I have time.

PR #804, opened 2020-03-28
Support publishing of extra artifacts

lefou:
With the new `PublishModule.extraPublish` you can add additional artifacts/files to be published. 

Use cases: 

* Publish a jar with dependencies
* Publish a distribution archive
* Publish a native binary

### DIFF
--- a/scalalib/src/PublishModule.scala
+++ b/scalalib/src/PublishModule.scala
@@ -51,6 +51,9 @@ trait PublishModule extends JavaModule { outer =>
     Artifact(pomSettings().organization, artifactId(), publishVersion())
   }
 
+  /**
+    * Extra artifacts to publish.
+    */
   def extraPublish: T[Seq[PublishModule.ExtraPublish]] = T{ Seq.empty[PublishModule.ExtraPublish] }
 
   def publishLocal(): define.Command[Unit] = T.command {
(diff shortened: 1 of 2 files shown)


lefou:
@heksesang I'd like to have your feedback on this. Also, if we agree to merge this, it would be cool if you can also extends the other publish modules (bintray, artifactory) to also support it. I'm looking for a next mill release with much improved publishing support.

scalalib/src/PublishModule.scala
```diff
@@ -106,12 +112,24 @@ trait PublishModule extends JavaModule { outer =>
...
     implicit def jsonify: upickle.default.ReadWriter[PublishData] = upickle.default.macroRW
   }
 
+  /** An extra resource artifact to publish.
+    * @param file The artifact file
+    * @param ivyCategory The ivy catogory (e.g. "jars", "zips")
+    * @params The file suffix including the file extension (e.g. "-with-deps.jar", "-dist.zip").
+    *        It will be appended to the artifact id to construct the full file name.
+    */
+  case class ExtraPublish(file: PathRef, ivyCategory: String, suffix: String)
```
lefou: I've no idea, if `ivyCategory` is the right name for this kind of thing.

heksesang: Classifier, isn't that the right name? And shouldn't it be `@param suffix` there?

heksesang:
Well, looks good to me. It's definitely a useful feature. But it should automatically extend to artifactory and bintray as those use the publishArtifacts() target, so not sure what extra support you think needs to be added for those?

scalalib/src/PublishModule.scala
```diff
@@ -106,12 +112,24 @@ trait PublishModule extends JavaModule { outer =>
...
 
+  case class PublishData(meta: Artifact, payload: Seq[(PathRef, String)])
   object PublishData{
     implicit def jsonify: upickle.default.ReadWriter[PublishData] = upickle.default.macroRW
   }
 
+  /** An extra resource artifact to publish.
+    * @param file The artifact file
+    * @param ivyCategory The ivy catogory (e.g. "jars", "zips")
+    * @params The file suffix including the file extension (e.g. "-with-deps.jar", "-dist.zip").
```
lefou:
```suggestion
    * @param The file suffix including the file extension (e.g. "-with-deps.jar", "-dist.zip").
```

lefou: Oh, thanks. An oversight.

scalalib/src/PublishModule.scala
lefou: I'm pretty sure it's not the classifier. It's something only needed for ivy repositories because of there layout. I think it's probably the type in a plural form?

PR #807, opened 2020-04-02
Bump Ammonite to pull in interp.watchValue

lihaoyi:
Sister PR to https://github.com/lihaoyi/Ammonite/pull/1056


main/core/src/eval/Evaluator.scala
```diff
@@ -856,7 +856,7 @@ object Evaluator{
   case class State(rootModule: mill.define.BaseModule,
                    classLoaderSig: Seq[(Either[String, java.net.URL], Long)],
                    workerCache: mutable.Map[Segments, (Int, Any)],
-                   watched: Seq[(os.Path, Long)])
+                   watched: Seq[(() => Long, Long)])
```
lefou, line 859: I'd love to see some `@param` ScalaDoc here. `() => Long` is way less self-expressive than `os.Path`.

main/src/main/RunScript.scala
```diff
@@ -25,15 +25,15 @@ object RunScript{
   def runScript(home: os.Path,
                 wd: os.Path,
                 path: os.Path,
-                instantiateInterpreter: => Either[(Res.Failing, Seq[(os.Path, Long)]), ammonite.interp.Interpreter],
+                instantiateInterpreter: => Either[(Res.Failing, Seq[(() => Long, Long)]), ammonite.interp.Interpreter],
```
lefou, line 28: Also here, some `@param` scaladoc telling what that is supposed to be would be great.

main/core/src/eval/Evaluator.scala
lihaoyi: Yeah this `() => Long` thing is annoying. I'll go wrap it in a `Watchable` class or something so it's easier to read and update this PR with the updated version of Ammonite

PR #809, opened 2020-04-03
Add release-fast and release-full modes to scala-native

ajrnz:
These modes where introduced in scala-native 0.4.0

release == release-full
release-fast prioritizes link optimization time and executable size

### DIFF
--- a/scalanativelib/api/src/ScalaNativeWorkerApi.scala
+++ b/scalanativelib/api/src/ScalaNativeWorkerApi.scala
@@ -50,8 +50,10 @@ object NativeLogLevel {
 sealed abstract class ReleaseMode(val name: String)
 
 object ReleaseMode {
-  case object Debug extends ReleaseMode("debug")
-  case object Release extends ReleaseMode("release")
+  case object Debug extends ReleaseMode("debug")                // fast compile, little optimization
+  case object Release extends ReleaseMode("release")            // same as ReleaseFull for versions 0.3.x
+  case object ReleaseFast extends ReleaseMode("release-fast")   // runtime optimize, faster compile, smaller binary
+  case object ReleaseFull extends ReleaseMode("release-full")   // runtime optimize, prefer speed over compile time and size
 
   implicit def rw: RW[ReleaseMode] = macroRW
 }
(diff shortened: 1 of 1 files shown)


scalanativelib/api/src/ScalaNativeWorkerApi.scala
```diff
@@ -52,6 +52,8 @@ sealed abstract class ReleaseMode(val name: String)
 object ReleaseMode {
   case object Debug extends ReleaseMode("debug")
   case object Release extends ReleaseMode("release")
+  case object ReleaseFast extends ReleaseMode("release-fast")
+  case object ReleaseFull extends ReleaseMode("release-full")
```
lolgab, line 56: These two modes don't work with SN 0.3. Since when I updated mill I joined the SN 0.4 and 0.3 workers I think it's fine. When 0.4.0 is out we can remove the `Release` mode at all.

lefou: Would be great to have these information as comment in the code, so anybody can see it when deciding which mode to choose.

scalanativelib/api/src/ScalaNativeWorkerApi.scala
```diff
@@ -50,8 +50,10 @@ object NativeLogLevel {
 sealed abstract class ReleaseMode(val name: String)
 
 object ReleaseMode {
-  case object Debug extends ReleaseMode("debug")
-  case object Release extends ReleaseMode("release")
+  case object Debug extends ReleaseMode("debug")                // fast compile, little optimization
+  case object Release extends ReleaseMode("release")            // same as ReleaseFull for versions 0.3.x
+  case object ReleaseFast extends ReleaseMode("release-fast")   // runtime optimize, faster compile, smaller binary
+  case object ReleaseFull extends ReleaseMode("release-full")   // runtime optimize, prefer speed over compile time and size
```
lefou, line 56:
```suggestion
  /** Fast compile, little optimization. */
  case object Debug extends ReleaseMode("debug")
  /** Same as [[ReleaseFull]] for versions 0.3.x. Deprecated since ScalaNative 0.4. */
  case object Release extends ReleaseMode("release")
  /** Runtime optimize, faster compile, smaller binary. */
  case object ReleaseFast extends ReleaseMode("release-fast")
  /** Runtime optimize, prefer speed over compile time and size. */
  case object ReleaseFull extends ReleaseMode("release-full")
```

I had rather some ScalaDoc in mind, which is part of the API and most probably will guide the `build.sc` writer when editing in an IDE.

lefou:
Thank you!

PR #810, opened 2020-04-03
Refactoring of the version file module

heksesang:
Refactored the module, to make it more testable. As a result, it's more modular than the previous iteration. The git tagging/pushing and writing the versions to file are now separated into different tasks, making it easier to use just the parts you want and extend as you want.

Most likely all the git related stuff here will be moved to its own module at some point. Will probably do that when adding the support for pure git tagging that was mentioned in the previous pull request, as that seems like a better place for that stuff to reside.


lefou:
Looks good to me. Thanks!

PR #811, opened 2020-04-07
Added `PublishModule.publishM2Local` to publish to local Maven repositories

lefou:


lihaoyi:
I don't know much about m2 repository formats but this looks good to me

PR #812, opened 2020-04-11
Pass linking and compile options to test in Scala Native

lolgab:
- Pass nativeLinkingOptions and nativeCompileOptions from testOuter to test
- Substitute platformSuffix with `::` in ScalaNativeModule

## Motivation
With current mill if you need native linking or compile options they are not passed to the Scala Native test and it fails to link.
I'm not testing this because it would mean to add link-time dependencies to test mill


lihaoyi:
Looks good to me I think, but I'll have to trust that you know what you're doing :)

lolgab:
Thank you for the trust :)

PR #815, opened 2020-04-14
More robust MILL_CLASSPATH on Windows

sake92:
I ran into problems while trying `mill.bat` from various folders..  
This ensures you can use it from e.g. "C:\New folder\mill.bat" (notice the space)


lefou:
Looks good to me.

PR #816, opened 2020-04-15
Added check for isDotty in project gen.

heksesang:


lihaoyi:
@smarter does this look reasonable?

PR #822, opened 2020-04-19
fix doc Tasks  code

YingLiu4203:
Fix  doc Tasks code to make it workable
 - import broken/missing packages
-  use `os.proc` to run subprocess

Improve  Custom Tasks code by removing redundant placeholder `_` in `os.isFile` and `os.read.lines`


lihaoyi:
Thanks @YingLiu4203 !

PR #826, opened 2020-04-22
fix broken cmds and outputs in document page 1

YingLiu4203:
Checked every command and verified output in docs/pages/1 - Intro to Mill

- update output such as `mill resolve _`
- fix broken command such as `mill resolve foo.__`, `mill plan moduledefs.compileClasspath`. and `mill show visualize __.compile`
- replace all `core` module with the `foo` module in the Example file and update outputs.
- update the file content as shown in the Example file
- fix some typos

PR #827, opened 2020-04-22
add doc for visualizePlan of two modules

YingLiu4203:
`mill show visualizePlan _.compile` diagrams the relationships between the compile tasks of two modules. The old command `mill show visualize __.compile` was wrong and its diagram was deleted in previous doc PR #826. Now add the correct command and its output:

- add a sample `build.sc` file  with two modules. 
- add description.
- ddd the generated diagram.


lefou:
Thank you!

PR #835, opened 2020-04-25
Use sheband in launcherUniversalScript

### LINKED ISSUE #834: mill fails to run in fish shell
Installed mill in arch linux and when running it in fish shell I got this error:

```
$ mill
Failed to execute process '/usr/bin/mill'. Reason:
exec: Exec format error
The file '/usr/bin/mill' is marked as an executable but could not be run by the operating system.
```

Googling the error, I found this https://github.com/fish-shell/fish-shell/issues/4946 which suggest the file is missing a shebang...

Here is the first lines of `/usr/bin/mill`:

```
@ 2>/dev/null # 2>nul & echo off & goto BOF^M
:
if [ -z "$JAVA_HOME" ] ; then
  JAVACMD="java"
else
  JAVACMD="$JAVA_HOME/bin/java"
fi
case "$1" in
```

Adding `#!/bin/bash` fixes the problem

jtjeferreira:
fixes #834

### DIFF
--- a/main/src/modules/Jvm.scala
+++ b/main/src/modules/Jvm.scala
@@ -373,7 +373,7 @@ object Jvm {
                               shellClassPath: Agg[String],
                               cmdClassPath: Agg[String],
                               jvmArgs: Seq[String],
-                              shebang: Boolean = false) = {
+                              shebang: Boolean = true) = {
     universalScript(
       shellCommands =
         s"""exec java ${jvmArgs.mkString(" ")} $$JAVA_OPTS -cp "${shellClassPath.mkString(":")}" $mainClass "$$@"""",


main/src/modules/Jvm.scala
```diff
@@ -373,7 +373,7 @@ object Jvm {
                               shellClassPath: Agg[String],
                               cmdClassPath: Agg[String],
                               jvmArgs: Seq[String],
-                              shebang: Boolean = false) = {
+                              shebang: Boolean = true) = {
```
lefou, line 376: This is a default argument of a public method. Changing the default means potentially breaking all users. This is definitely the wrong position to fix the linked issue (#834). Instead, you should provide the value on the call site.

lihaoyi:
Yeah this has come up before. Ultimately there are more windows people than fish people, so making it work on windows takes priority

jtjeferreira:
> Yeah this has come up before. Ultimately there are more windows people than fish people, so making it work on windows takes priority. If we can find a way to make it work on fish while preserving windows compatibility, that would be one possible path forward.

fair enough. should I also close #834 ?

lefou:
> fair enough. should I also close #834 ?

It's a valid issue after all. Guess, we may keep it open to catch other people struggling with the same issue.

PR #839, opened 2020-04-29
Make T.input watchable for Path, PathRef and Seq of both

lefou:


lihaoyi:
Not sure if we want to do it this way; I assume if anyone wanted a `T.input` with a `Path` or `PathRef`, they'd probably be using `T.source`/`T.sources` anyway.

I think what we want to do here is to plumb the lambda that `T.input` captures and change the rest of the RunScript infrastructure that does watching/etc. to know how to handle either `PathRef`s or lambdas. We could even re-use the `ammonite.interp.Watchable` type, since that models exactly this and is used for exactly the same use case

PR #840, opened 2020-04-29
[WIP] swap out custom scheduler for Futures

lihaoyi:
Still pretty rough, but seems to work on my small scale tests.

Right now I'm not actually using Futures to pass values around: I'm simply using them to schedule computations, whose values end up the being stored in the same `results` dictionary we were using before for both parallel and sequential evaluators. I sprinkled some `synchronized` around to make it work for this prototype, but we can clean them up and use `ConcurrentHashMap` or similar later before merging

This seems to greatly improve utilization and scheduling over the hand-rolled one we had before. Below is the `mill-par-profile.json` for compiling the book www.handsonscala.com with `./mill -j 6 color.pdf`:

`0.6.2-16-b169e9`:

Screenshot 2020-04-29 18 34 51

This PR:

Screenshot 2020-04-29 18 34 45

The wall clock time dropped from ~87s to ~57s, with a much denser apparent chrome profile.


main/core/src/define/Task.scala
```diff
@@ -235,17 +235,29 @@ object Target extends TargetGenerated with Applicative.Applyer[Task, Task, Resul
...
-  def workerImpl[T: c.WeakTypeTag](c: Context)
-                                  (t: c.Expr[T])
-                                  (ctx: c.Expr[mill.define.Ctx]): c.Expr[Worker[T]] = {
+  def workerImpl2[T: c.WeakTypeTag](c: Context)
+                                   (t: c.Expr[T])
+                                   (ctx: c.Expr[mill.define.Ctx]): c.Expr[Worker[T]] = {
     import c.universe._
-    reify(
-      new Worker[T](Applicative.impl[Task, T, mill.api.Ctx](c)(t).splice, ctx.splice)
+    mill.moduledefs.Cacher.impl0[Worker[T]](c)(
```
lihaoyi, line 257: @lefou I think this is the root of your problems with the workers running multiply; we were not caching them on first evaluation, so multiple worker instances were being created. Caching them like we do for the rest of the tasks seems to make the problem go away

main/core/src/eval/Evaluator.scala
```diff
@@ -598,221 +547,89 @@ case class Evaluator(
...
+        evalLog.debug(s"End time: ${new java.util.Date()}")
+
+        evalLog.close()
+        timeLog.close()
+        Evaluator.Results(
+          goals.indexed.map(results(_).map(_._1)),
+          finished.flatMap(_.newEvaluated),
+          transitive,
+          failing,
+          Vector(),
```
lihaoyi, line 628: this is TODO, but maybe since we now have the `mill-par-profile.json` file I can just get rid of this functionality entirely

main/core/src/eval/Evaluator.scala
```diff
@@ -598,221 +547,89 @@ case class Evaluator(
...
-        os.close()
-        baos.close()
+        //      println("AWAITING futures")
+        val finished = futures.map(Await.result(_, duration.Duration.Inf))
+
+        val failing = new util.MultiBiMap.Mutable[Either[Task[_], Labelled[_]], Result.Failing[_]]
+        for ((k, vs) <- sortedGroups.items()) {
+          failing.addAll(
+            k,
+            vs.items.flatMap(results.get).collect { case f: Result.Failing[_] => f.map(_._1) }
```
lihaoyi, line 615: This is mostly copy-pasted from the sequential evaluator; I can move them into a helper function

main/core/src/eval/Evaluator.scala
```diff
@@ -598,221 +547,89 @@ case class Evaluator(
...
+        pprint.log(groupGraph, height=99999)
+        val terminals = sortedGroups.keys.toVector
+        //      println("INITIALIZING promises " + terminals.map(label).mkString(", "))
+        val promises = terminals
+          .map(k => (k, scala.concurrent.Promise[Any]))
+          .toMap
+
+        val taskFutures: Map[Terminal, Future[Any]] = promises.map { case (k, p) => (k, p.future) }
+
+        val results = mutable.LinkedHashMap.empty[Task[_], Result[(Any, Int)]]
```
lihaoyi, line 574: This guy should probably be made a ConcurrentHashMap, which would make all the `synchronized` blocks go away. I think this map is the only mutable data structure in the entire algorithm, except for the `workerCache` which we can also make concurrent

main/core/src/eval/Evaluator.scala
```diff
@@ -598,221 +547,89 @@ case class Evaluator(
...
+          val deps = interGroupDeps((k, sortedGroups.lookupKey(k))).map(_._1)
+          //          println("REGISTERING " + label(k) + " <- " + deps)
+          Future.sequence(deps.map(taskFutures(_)))
+            .map { upstreamValues =>
+              //              println("START " + label(k) + " " + deps.size)
+              //              println("THREAD " + Thread.currentThread().getName())
+              //              println("DEPS " + deps)
+              val startTime = System.currentTimeMillis()
+
+              val res = blocking(evaluateGroupCached(k, sortedGroups.lookupKey(k), results, "lol", _ => None, testReporter, logger))
```
lihaoyi: I'm re-using the same `evaluateGroupCached` function here for now, for maximal code re-use with the sequential code and the previous parallel implementation. The API isn't perfect, but it'll do for now

main/core/src/eval/Evaluator.scala
```diff
@@ -598,221 +547,89 @@ case class Evaluator(
...
+          val deps = interGroupDeps((k, sortedGroups.lookupKey(k))).map(_._1)
+          //          println("REGISTERING " + label(k) + " <- " + deps)
+          Future.sequence(deps.map(taskFutures(_)))
+            .map { upstreamValues =>
+              //              println("START " + label(k) + " " + deps.size)
+              //              println("THREAD " + Thread.currentThread().getName())
+              //              println("DEPS " + deps)
+              val startTime = System.currentTimeMillis()
+
+              val res = evaluateGroupCached(k, sortedGroups.lookupKey(k), results, "lol", _ => None, testReporter, logger)
```
lihaoyi, line 586: I'm re-using the same `evaluateGroupCached` function here, to re-use most of the caching logic with the previous parallel scheduler and the sequential implementation. The interface isn't quite what I want, but it'll do for now I think

lefou:
Yeah, looks great. I'll have a closer look tonight.

lihaoyi:
@lefou I did a round of cleanup, this should be ready for review

lihaoyi:
Latest profiles from my laptop:

`./mill -i dev.run ../hands-on-scala-programming/ -j 6 color.pdf`

Screenshot 2020-04-30 23 50 17

`./mill -i dev.run ../ammonite/ -j 8 __.compile`

Screenshot 2020-04-30 23 50 08

main/core/src/eval/Evaluator.scala
```diff
@@ -198,7 +299,7 @@ case class Evaluator(
         } yield (parsed, cached.valueHash)
 
         val workerCached: Option[Any] = labelledNamedTask.task.asWorker
-          .flatMap{w => workerCache.get(w.ctx.segments)}
+          .flatMap{w => synchronized{ workerCache.get(w.ctx.segments) }}
```
hepin1989, line 302: can this be removed ?

main/core/src/eval/Evaluator.scala
```diff
@@ -261,7 +362,7 @@ case class Evaluator(
                        inputsHash: Int,
                        labelledNamedTask: Labelled[_]) = {
     labelledNamedTask.task.asWorker match{
-      case Some(w) => workerCache(w.ctx.segments) = (inputsHash, v)
+      case Some(w) => synchronized{ workerCache(w.ctx.segments) = (inputsHash, v) }
```
hepin1989, line 365: can this `synchronized ` be removed?

PR #845, opened 2020-05-02
Rename .test with .testUncached to avoid duplicate resolution

lihaoyi:
This should avoid the problem with `__.test` running the `.test.test()` command twice

### DIFF
--- a/docs/pages/8 - Mill Internals.md
+++ b/docs/pages/8 - Mill Internals.md
@@ -104,7 +104,7 @@ macros which allow you to use `Task#apply()` within the block to "extract" a
 value.
 
 ```scala
-def test() = T.command {
+def testUncached() = T.command {
   TestRunner.apply(
    "mill.UTestFramework",
    runDepClasspath().map(_.path) :+ compile().path,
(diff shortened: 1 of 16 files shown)


lefou:
I don't like that name change. Also it will break quite a lot of setups including mine.

Also, I think it's only an workaround. We still have the issue, that command might end up being build concurrently resulting in build failures. Just not in the case of `mil __.test`.

lihaoyi:
@lefou this is a somewhat orthogonal change; apart from working around the problem with parallel `__.test`, it also works around the problem with sequential `__.test` running the same thing more than once, and also works around the problem with `mill resolve __.test` picking up things multiple times. There are probably other places as well

Trying to fix the problem in the parallel evaluator is possible, but even if we did that we'd still need to fix it in a bunch of other places. While this is a bit of a workaround, I think this sort of squishiness is somewhat inherent in our name-based query syntax.

It shouldn't be too hard to fix your setup to accommodate the name change I think? I think some amount of breakage needs to be acceptable, as long as fixing it isn't too hard. After all, the upgrade to 2.13 broke a bunch of my own stuff as well.

lefou:
Fair enough. (I still find the new name really ugly.)
Would be good to have tickets to track these other issues.

lihaoyi:
haha yeah the new name is really ugly. I thought about renaming it `run`, but then that would conflict with the existing `test.run` mechanism for running `main` methods, that I actually do make use of for "manual" non-test-framework testing. That would be a more annoying compatibility breakage than renaming `test -> testUncached` in a few places.

I'm hoping people won't be actively using `.testUncached` much, since most people would just use `mill __.test` and it should do the right thing...

PR #850, opened 2020-05-05
Fix Sonatype requests retrying too fast

joan38:


lefou:
In essence, we just wait a bit longer each time after a previous 5xx error?

joan38:
Yes exponentially.
Sonatype is sometimes taking time to commit the fact that a repo is closed. The issue could have been also resolved by waiting more between retries or by increasing the number of retries but I thought I'll make a proper exponential backoff to account for all situations.

lefou:
Make sense. Thank you!

joan38:
Do I need any more action on this wrt CI?

joan38:
Great Thanks

Issue #345, opened 2018-05-21
Make gpg passphrase optional in publish method. 


Baccata:
My team uses a private artifact repository (artifactory) which exposes the same API as sonatype. 

However, the signing of the artifacts is not required. I'd like to make the gpg passphrase optional in `PublishModule#publish`, or have another task (`publishWithoutSigning`) that would not require the passphrase.

Thoughts ?

rockjam:
What happens when you pass null?

Baccata:
```
1 targets failed
scribe.recursion.publish java.lang.NullPointerException
    java.lang.ProcessBuilder.start(ProcessBuilder.java:1012)
    ammonite.ops.Shellout$.executeInteractive(Shellout.scala:27)
    ammonite.ops.Shellout$.$anonfun$$percent$1(Shellout.scala:14)
    ammonite.ops.Shellout$.$anonfun$$percent$1$adapted(Shellout.scala:14)
    ammonite.ops.Command.applyDynamic(Shellout.scala:118)
    mill.scalalib.publish.SonatypePublisher.poorMansSign(SonatypePublisher.scala:141)
```

rockjam:
I think that having optional gpg passphrase is reasonable

lihaoyi:
Maybe we could give `publish` a separate flag `--signed false`? The reason is that there are three cases here:

- Signed with passphrase
- Signed without passphrase
- Unsigned

If we're going to change this we might as well make sure we handle all three cases

Baccata:
Closing this as the problem it describes is solved. There is still a problem with publishing to artifactory with the default PublishModule. I need to have a closer look at how it's API differs from sonatype's API to formulate a new issue.

ddelautre:
@Baccata Did you succeed publishing to Artifactory?

Baccata:
@ddelautre,  yeah, here's a snippet I used. It might need updating to the latest mill version, but it should do the trick : 

```scala 
import java.math.BigInteger
import java.security.MessageDigest

import ammonite.ops._
import mill.scalalib.publish.{ SonatypeHttpApi, _ }
import mill.util.Logger
import scalaj.http.HttpResponse

// Basically a copy of https://github.com/lihaoyi/mill/blob/0.2.2/scalalib/src/mill/scalalib/publish/SonatypePublisher.scala
// to avoid requiring a gpg passphrase.
class ArtifactoryPublisher(uri: String, snapshotUri: String, credentials: String, log: Logger) {

[...]

PR #851, opened 2020-05-05
Ability to pass GPG arguments to publish

joan38:
Not sure if we are ok to break the interface here. If not I can probably add back `gpgPassphrase` and `gpgKeyName` with a `@deprecated` and wire that to `gpgArgs`.

### DIFF
--- a/scalalib/src/publish/SonatypePublisher.scala
+++ b/scalalib/src/publish/SonatypePublisher.scala
@@ -9,9 +9,8 @@ import os.Shellable
 class SonatypePublisher(uri: String,
                         snapshotUri: String,
                         credentials: String,
-                        gpgPassphrase: Option[String],
-                        gpgKeyName: Option[String],
                         signed: Boolean,
+                        gpgArgs: Seq[String],
                         readTimeout: Int,
                         connectTimeout: Int,
                         log: Logger,
@@ -25,7 +24,6 @@ class SonatypePublisher(uri: String,
   }
 
   def publishAll(release: Boolean, artifacts: (Seq[(os.Path, String)], Artifact)*): Unit = {
-
     val mappings = for ((fileMapping0, artifact) <- artifacts) yield {
       val publishPath = Seq(
         artifact.group.replace(".", "/"),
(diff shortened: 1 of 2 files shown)


scalalib/src/publish/SonatypePublisher.scala
```diff
@@ -154,12 +152,9 @@ class SonatypePublisher(uri: String,
...
 
   // http://central.sonatype.org/pages/working-with-pgp-signatures.html#signing-a-file
-  private def poorMansSign(file: os.Path, maybePassphrase: Option[String], maybeKeyName: Option[String]): os.Path = {
+  private def gpgSigned(file: os.Path, args: Seq[String]): os.Path = {
     val fileName = file.toString
-    val optionFlag = (flag: String, ov: Option[String]) => ov.map(flag :: _ :: Nil).getOrElse(Nil)
-    val command = "gpg" ::
-      optionFlag("--passphrase", maybePassphrase) ++ optionFlag("-u", maybeKeyName) ++
-        Seq("--batch", "--yes", "-a", "-b", fileName)
+    val command = Seq("gpg", "--batch", "--yes", "-a", "-b", fileName) ++ args
```
lefou, line 157: Wondering, if we should move these defaults into the default param to `gpgArgs` and document it's meaning. It's not a change request, just a thought. WDYT?

joan38: I was thinking the same but then people need to be aware that if there specify the `--gpgArgs` they need to provide these too to keep them.

lefou: We could add a hint in the scaladoc, too. But having these hard-coded args screams for future maintenance work, as seen in the past (#345, #530, #753). We could also rename the parameter to `alternativeGpgArgs` or `overrideGpgArgs` to indicate the fact that it replaced all defaults.

lefou: What about an accessor to the defaults to make reusing easy:

```scala
val defaultGpgArgs = ...
def publish(...,  gpgArgs: Seq[String] = defaultGpgArgs, ...)
```

And a simple example in the docs, showing how to add a passphrase using the defaultsArgs + the passphrase args

joan38: Sounds good to me. What do you think @jodersky?

jodersky: Looks good! I agree with @lefou's suggestion too. Thanks for working on this!

joan38: Change pushed

lihaoyi:
This has broken publishing on the master branch:

```
Unknown arguments: "--gpgPassphrase" "[secure]"
```

@joan38 can you take a look?

joan38:
@lihaoyi You need to pass `--gpgArgs '--passphrase [secure]'`

PR #854, opened 2020-05-05
Fix ScalaPB and upgrade coursier

### LINKED ISSUE #846: scalapblib resolves scalapbc_2.12 in 2.13 builds of mill
In the post-0.6.2 builds of mill that are built on 2.13, `compileScalaPB` fails because the `ScalaPBModule` uses 2.12.4 as target version when it resolves the `scalapbc` dependency:

```scala
def scalaPBClasspath: T[Loose.Agg[PathRef]] = T {
  resolveDependencies(
    Seq(
      coursier.LocalRepositories.ivy2Local,
      MavenRepository("https://repo1.maven.org/maven2")
    ),
    Lib.depToDependency(_, "2.12.4"),
    Seq(ivy"com.thesamet.scalapb::scalapbc:${scalaPBVersion()}")
  )
}

```

This should be fixable by replacing the hardcoded `"2.12.4"` with `scalaVersion()`, I think.

### LINKED ISSUE #853: Non-deterministic ordering of jars from $ivy imports
I'm not sure if this issue concerns mill, ammonite or coursier but I am opening it here because I encountered it when working with mill. Please feel free to redirect/transfer this issue to the correct place.

---

I just ran into a non-determinism bug concerning the ordering of source and binary jars resolved through an `$ivy` import.

Essentially, when adding a dependency through an `i$vy` import, both the regular jar and source jars are downloaded via coursier and made available on the classpath. Unfortunately the ordering of source vs regular jars is non-deterministic and can vary between environments, which can lead to extremely subtle bugs. For instance, here is a build.sc extract that triggered my investigation:

```scala
import $ivy.`com.google.api-client:google-api-client:1.30.9`

def gversion() = T.command {

  val url = classOf[com.google.api.client.googleapis.GoogleUtils].getResource("google-api-client.properties")
  println("url: " + url)

  val stream = classOf[com.google.api.client.googleapis.GoogleUtils].getResourceAsStream("google-api-client.properties")
  val content = scala.io.Source.fromInputStream(stream).mkString
  println("content: " + content)

}
```

Running `mill -i gversion` on machine A gives:

```
url: jar:file:$HOME/.cache/coursier/v1/https/repo1.maven.org/maven2/com/google/api-client/google-api-client/1.30.9/google-api-client-1.30.9.jar!/com/google/api/client/googleapis/google-api-client.properties
content: google-api-client.version=1.30.9
```

and on machine B:

```
url: jar:file:$HOME/.cache/coursier/v1/https/repo1.maven.org/maven2/com/google/api-client/google-api-client/1.30.9/google-api-client-1.30.9-sources.jar!/com/google/api/client/googleapis/google-api-client.properties
content: google-api-client.version=${project.version}
```

As you can see, machine A reads the properties resource from the binary jar whereas machine B reads it from the source jar. AFAICT the ordering depends on the filesystem type which may vary from platform to platform (in my case I was actually wondering why my build was behaving differently on Docker compared to by almost identical workstation).

My suggestion to fixing this would be to either
1) add only the binary jars to the classpath and drop the sources
2) if dropping sources is not an option, then order jars deterministically, prioritizing binary jars

I'd be happy to give the fix a shot if someone can point me to the correct location.

Thanks!

jodersky:
This PR contains two commits, each of which are independent (see individual commit messages).

In essence, this fixes #846 and fixes #853

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -41,7 +41,7 @@ object Deps {
   val acyclic = ivy"com.lihaoyi::acyclic:0.2.0"
   val ammonite = ivy"com.lihaoyi:::ammonite:2.1.1"
   val bloopConfig = ivy"ch.epfl.scala::bloop-config:1.4.0-RC1"
-  val coursier = ivy"io.get-coursier::coursier:2.0.0-RC5-3"
+  val coursier = ivy"io.get-coursier::coursier:2.0.0-RC6-14"
   val flywayCore = ivy"org.flywaydb:flyway-core:6.0.1"
   val graphvizJava = ivy"guru.nidi:graphviz-java:0.8.3"
   val ipcsocket = ivy"org.scala-sbt.ipcsocket:ipcsocket:1.0.0"
--- a/contrib/scalapblib/src/ScalaPBModule.scala
+++ b/contrib/scalapblib/src/ScalaPBModule.scala
@@ -64,7 +64,7 @@ trait ScalaPBModule extends ScalaModule {
         coursier.LocalRepositories.ivy2Local,
         MavenRepository("https://repo1.maven.org/maven2")
       ),
-      Lib.depToDependency(_, "2.12.4"),
+      Lib.depToDependency(_, "2.13.1"),
       Seq(ivy"com.thesamet.scalapb::scalapbc:${scalaPBVersion()}")
     )
   }


lefou:
Thank you. It looks to me, that the coursier update make Travis-CI unhappy. Lots of "concurrent download" errors. If you split your commits into separate pull requests, we can handle them independently.

jodersky:
ok no worries, I'll split them up

PR #856, opened 2020-05-06
Adds a way to skip bloop configuration

Baccata:
When a project has a large amount of cross-compilation and bloop is
only used as a gateway to metals, one might want to skip the generation
of the configuration for some specific scala versions or target
platforms

### DIFF
--- a/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
@@ -63,7 +63,7 @@ class BloopImpl(ev: () => Evaluator, wd: Path) extends ExternalModule { outer =>
     }
 
     /**
-      * Setting to true enables the bloop configuration generation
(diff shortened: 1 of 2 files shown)


contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
```diff
@@ -206,6 +216,8 @@ class BloopImpl(ev: () => Evaluator, wd: Path) extends ExternalModule { outer =>
               mode = m.releaseMode() match {
                 case ReleaseMode.Debug => BloopConfig.LinkerMode.Debug
                 case ReleaseMode.Release => BloopConfig.LinkerMode.Release
+                case ReleaseMode.ReleaseFast => BloopConfig.LinkerMode.Release
```
Baccata, line 219: Got a non-exhaustive pat-match warning, figured I'd fix it. Bloop config doesn't seem to have corresponding enum values though

PR #857, opened 2020-05-06
Add flatMap on Result

joan38:
This is needed for example [here](https://github.com/joan38/mill-scalafix/blob/master/mill-scalafix/src/com/goyeau/mill/scalafix/ScalafixModule.scala#L35-L51) where I [patched it how I can](https://github.com/joan38/mill-scalafix/blob/master/mill-scalafix/src/com/goyeau/mill/scalafix/ScalafixModule.scala#L22-L28)

### DIFF
--- a/main/api/src/mill/api/Result.scala
+++ b/main/api/src/mill/api/Result.scala
@@ -1,11 +1,15 @@
 package mill.api
 
+import scala.language.implicitConversions
+
 /**
  * The result of a task execution.
+ *
  * @tparam T The result type of the computed task.
  */
 sealed trait Result[+T] {
   def map[V](f: T => V): Result[V]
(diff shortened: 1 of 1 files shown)


joan38:
@lefou Can I get review on this?

lefou:
Sure, I already had a look but honestly have no idea if a `flatMap` introduce any issues conceptually? It looks good to me though.

@lihaoyi, WDYT?

joan38:
It's really to be able to chain them possibly in a for comprehension.
I don't see how to use this type if I can't chain them really. See [the usage example](https://github.com/joan38/mill-scalafix/blob/master/mill-scalafix/src/com/goyeau/mill/scalafix/ScalafixModule.scala#L35-L51).

lihaoyi:
No idea, seems handy

PR #858, opened 2020-05-06
Upgrade coursier to 2.0.0-RC6-15

### LINKED ISSUE #853: Non-deterministic ordering of jars from $ivy imports

jodersky:
The newer version of coursier has deterministic ordering (see coursier/coursier#1546, thanks for the pointer @smarter), which in turn fixes #853.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -41,7 +41,7 @@ object Deps {
   val acyclic = ivy"com.lihaoyi::acyclic:0.2.0"
   val ammonite = ivy"com.lihaoyi:::ammonite:2.1.1"
   val bloopConfig = ivy"ch.epfl.scala::bloop-config:1.4.0-RC1"
-  val coursier = ivy"io.get-coursier::coursier:2.0.0-RC5-3"
+  val coursier = ivy"io.get-coursier::coursier:2.0.0-RC6-15"
   val flywayCore = ivy"org.flywaydb:flyway-core:6.0.1"
   val graphvizJava = ivy"guru.nidi:graphviz-java:0.8.3"
   val ipcsocket = ivy"org.scala-sbt.ipcsocket:ipcsocket:1.0.0"


jodersky:
This supersedes part of #854.

I can't reproduce the test failures of the original PR, so I'm submitting this as-is to see if CI encounters errors again

jodersky:
Hmm, all the tests pass on my laptop (Debian 11, OpenJDK 11). This is very strange

jodersky:
I can reproduce the errors if I clear the coursier cache prior to running tests. I'm not sure what is causing the problem, but it appears with coursier 2.0.0-RC6-2 (2.0.0-RC6-1 still works).

jodersky:
It looks like the lockfile handling changed slightly in coursier. I'm still not sure what's causing the problem, but found a fix (see linked PR) which I submitted to coursier.

jodersky:
A new version of coursier came out with my proposed fix! I'm still waiting for a successful CI run, but if you look at the last one and second to last:

- https://travis-ci.org/github/lihaoyi/mill/builds/684591597
- https://travis-ci.org/github/lihaoyi/mill/builds/684459302

you can see that the failures are intermittent and unrelated to coursier. The "union" of those CI reports is all green :)

Pinging @lefou for another review

lefou:
It's really unfortunate that Travis-CI jobs fail for pull requests that are not based on latest master commit. It's only in mill repo, so must be releated to how we calculate our version number.

@jodersky Can you re-base again, and I try to not push any commits into master. ;-)

jodersky:
I'll wait until no more builds are queued up and try again

lefou:
Some Travis jobs succeeded now which failed before, so I'd say: Looks good to me! I'll merge.

Thanks for your patience.

jodersky:
Great, thank you! I hope that the other issues will be solved after rebootstrapping mill

PR #849, opened 2020-05-04
fix `Cannot find scala-library-2.13.0.jar ..` Exception when compile dotty project

ohze:
+ build.sc
```scala
import mill._, scalalib._

object foo extends Cross[FooModule]("0.24.0-RC1", "2.13.2")
class FooModule(val crossScalaVersion: String) extends CrossSbtModule
```
+ output
<details>

```scala
% cat .mill-version 
0.6.2-35-7d1144
% mill 'foo[0.24.0-RC1]'.compile
Compiling /Users/thanhbv/ohze/oss/mill-example/build.sc
[26/26] foo[0.24.0-RC1].compile 
[...]

PR #859, opened 2020-05-06
Update Dotty standard library version

anatoliykmetyuk:


anatoliykmetyuk:
@lihaoyi do you think the CI failure is related to the PR? Any ideas of why it might not work on different JVM versions?

scalalib/worker/src/ZincWorkerImpl.scala
```diff
@@ -345,7 +345,7 @@ class ZincWorkerImpl(compilerBridge: Either[
         loader = getCachedClassLoader(compilersSig, combinedCompilerJars),
         libraryJar = libraryJarNameGrep(
           compilerClasspath,
-          if (isDotty(scalaVersion)) "2.13.0" else scalaVersion
+          if (isDotty(scalaVersion)) "2.13.2" else scalaVersion
```
lefou, line 348: Why is this a hardcoded value at all?

lefou:
#849 fixed this issue in a smarter way.

lefou:
Thank you!

PR #861, opened 2020-05-07
Fix generator for ivy.xml when using extra publish artifacts 

atooni:
When extra artifacts shall be published, these must occur in the generated ivy.xml as well. 
This fix includes those in the xml and also provides a test case for verifying the publication section of the ivy.xml.


scalalib/src/publish/ExtraPublish.scala
```diff
@@ -0,0 +1,25 @@
...
+ *  @param ivyClassifier An Optional classifier to be used when publishing the file
+ *  @param ivyConfig The ivy config to be used
+ * @param ivyExt The extension that will be used publishing the file to the ivy repo
+ */
+case class ExtraPublish(
+  file: PathRef,
+  ivyType: String = "jar",
+  ivyClassifier: Option[String] = None,
+  ivyConfig: String = "compile",
+  ivyExt: String = "jar"
```
lefou, line 18: The classifier, extentsion are also relevant for Maven repositories, so we should remove the "ivy" prefix.

What means the default value "compile" for `ivyConfig`. Does that mean, coursier will always resolve the artifact when compiling? That's not what I would expect for an additional resource like e.g. a distribution zip file.

scalalib/src/publish/ExtraPublish.scala
```diff
@@ -0,0 +1,25 @@
+package mill.scalalib.publish
```
lefou, line 1: I'd like to keep this in package `mill.scalalib` if possible.

atooni: Thats not possible as it would introduce cyclic dependencies. Perhaps if the PublishModule as such is also moved to scalalib.publish it might work

scalalib/src/publish/ExtraPublish.scala
atooni: TBH. I have no idea. AFAIK the config can be defined by the user and "compile" seemed to be a reasonable default. We can use "runtime" instead, but then you have to remember to use the config when coding the dependenca in the mill build. At the end of the day you only reference the extra published artifact just when you want it resolved .

PR #872, opened 2020-05-09
Disable client-daemon mode in WSL

lihaoyi:
I'm gonna spend a bit of time trying to investigate why WSL causes client-daemon mode to hang, but if I can't figure out we can just disable it in the launcher


lihaoyi:
Since https://github.com/lihaoyi/mill/issues/874 seems to be non-trivial, I'll merge this as is for now. We can revert later on once WSL supports file locks properly

PR #881, opened 2020-05-12
Minor cleanups

joan38:


contrib/bintray/src/mill/contrib/bintray/BintrayPublisher.scala
```diff
@@ -78,7 +78,7 @@ class BintrayPublisher(owner: String,
   private def reportPublishResults(publishResults: Seq[(BintrayPackage, Artifact, Seq[requests.Response])]) = {
     val errors =
       for {
-        (pkg, artifact, responses) <- publishResults
+        (_, artifact, responses) <- publishResults
```
lefou, line 81: Although I can understand your motivation to communicate that it is unused, it makes reasoning about the code harder.
The IDE is able to still detect unused vars, but reading the code immediately makes clear, what the unused variable means. It helps also when changing that tuple later.

main/src/main/MainScopts.scala
```diff
@@ -25,7 +25,5 @@ object Tasks{
 class EvaluatorScopt[T]()
   extends scopt.Read[mill.eval.Evaluator]{
   def arity = 0
-  def reads = s => {
-    Evaluator.currentEvaluator.get.asInstanceOf[mill.eval.Evaluator]
-  }
+  def reads = _ => Evaluator.currentEvaluator.get
```
lefou, line 28: It would be consequent to also make the return code explicit.

main/test/src/eval/JavaCompileJarTests.scala
```diff
@@ -132,15 +132,14 @@ object JavaCompileJarTests extends TestSuite{
 
       // Create the Jar again, but this time, filter out the Foo files.
       def noFoos(s: String) = !s.contains("Foo")
-      val filterFunc = (p: os.Path, r: os.RelPath) => noFoos(r.last)
-      eval(filterJar(filterFunc))
+      eval(filterJar((_, r) => noFoos(r.last)))
```
lefou, line 135: less readable, IMHO

lihaoyi:
Honestly not a fan of most of these changes.

main/src/main/MainScopts.scala
joan38: If we want to make it explicit we should do `Evaluator.currentEvaluator.get: mill.eval.Evaluator`
Casting is bad as you shortcut the type checker.

joan38:
I understand you guys prefer leaving unused variables to provide more context on what is what as opposed to meaningless `_`s. I started to remove unused code because for a time to time contributor like me it's hard to track what code is used and what is unused but fair enough.

main/src/modules/Jvm.scala
```diff
@@ -80,15 +78,15 @@ object Jvm {
...
+                    workingDir: os.Path): Unit = {
     val process = spawnSubprocess(commandArgs, envArgs, workingDir)
     val shutdownHook = new Thread("subprocess-shutdown") {
       override def run(): Unit = {
         System.err.println("Host JVM shutdown. Forcefully destroying subprocess ...")
         process.destroy()
       }
     }
-    Runtime.getRuntime().addShutdownHook(shutdownHook)
+    Runtime.getRuntime.addShutdownHook(shutdownHook)
```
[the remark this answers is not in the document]
joan38: Tho this isn't an effectful function

joan38:
That's ok I'll close this down.
This PR was more about fixing some warts in the code rather than subjective formatting. That is why I reverted most of the unused symbols changes.

That said there is quirks in the code, for example `val voidFuture = listening.get()` as a last statement so that it complies with the return type `Unit` or usage of deprecated `TraversableOnce` or documentation fixes or `path.last.startsWith(".")` instead of `path.toIO.isHidden`... most for which the compiler is issuing so many warnings.

If Mill is aspiring to be a build tool to be used by the broader Scala community it needs to also open its codebase to the broader Scala community.

Thanks

PR #883, opened 2020-05-13
cleanup + fix backticks in readme

rethab:
the `import` statement actually needs backticks after `$ivy.` so I added them. however using them with each snipped is somewhat confusing since they are already formatted correctly by indenting and in the case of the import they are actually needed. hopefully this makes it clearer and copy-paste ready


contrib/bsp/readme.md
```diff
@@ -30,11 +30,11 @@ After the bsp support module would be published, it should be enough to:
 1) Install mill
 2) Add the following import statement in the build.sc of your project:
 
-        `import $ivy.com.lihaoyi::mill-contrib-bsp:$OFFICIAL_MILL_VERSION`
+        import $ivy.`com.lihaoyi::mill-contrib-bsp:$OFFICIAL_MILL_VERSION`
```
lefou, line 33:
```suggestion
        import $ivy.`com.lihaoyi::mill-contrib-bsp:$MILL_VERSION`
```

This will then work out-of-the-box.

rethab:
Totally, linking to the docs would be better. Or just not have a readme at all. I think I just instinctively looked for a RAEDME and would have looked at the docs instead so I think a README is not necessary.

lefou:
Agreed. Can you change your PR to delete the README?

rethab:
Done @lefou

lefou:
Thanks!

PR #887, opened 2020-05-15
Relative library paths in GenIdea, and a relative COURSIER_CACHE on WSL2

lihaoyi:
This should allow the GenIdea output XML files to be usable cross-platform, e.g. we could resolve them on a WSL-Ubuntu environment and load them into Intellij on the host Windows machine

Tested locally on OS-X, seems to work, and tested on WSL2 as well. IntelliJ seems to be able to load the config successfully, even across the Ubuntu/Windows boundary

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -594,7 +594,7 @@ def launcherScript(shellJvmArgs: Seq[String],
          |# Client-server mode doesn't seem to work on WSL, just disable it for now
          |# https://stackoverflow.com/a/43618657/871202
          |if grep -qEi "(Microsoft|WSL)" /proc/version > /dev/null 2> /dev/null ; then
-         |    COURSIER_CACHE=out/.coursier ${java("mill.MillMain")}
+         |    COURSIER_CACHE=.coursier ${java("mill.MillMain")}
          |else
          |    case "$$1" in
          |      -i | --interactive )
--- a/build.sc
+++ b/build.sc
@@ -594,7 +594,7 @@ def launcherScript(shellJvmArgs: Seq[String],
          |# Client-server mode doesn't seem to work on WSL, just disable it for now
          |# https://stackoverflow.com/a/43618657/871202
          |if grep -qEi "(Microsoft|WSL)" foo.txt &> /dev/null ; then
-         |    ${java("mill.MillMain")}
+         |    COURSIER_CACHE=out/.coursier ${java("mill.MillMain")}
          |else
          |    case "$$1" in
          |      -i | --interactive )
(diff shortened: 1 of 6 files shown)


build.sc
```diff
@@ -594,7 +594,7 @@ def launcherScript(shellJvmArgs: Seq[String],
          |# Client-server mode doesn't seem to work on WSL, just disable it for now
          |# https://stackoverflow.com/a/43618657/871202
          |if grep -qEi "(Microsoft|WSL)" /proc/version > /dev/null 2> /dev/null ; then
-         |    ${java("mill.MillMain")}
+         |    COURSIER_CACHE=out/.coursier ${java("mill.MillMain")}
```
lefou, line 597: So we use a local coursier cache under `out` on Windows but not under the other OSes. Why it that. Shouldn't we use a location that it less sensitive to `mill clean`?

scalalib/src/GenIdeaImpl.scala
```diff
@@ -499,15 +499,19 @@ case class GenIdeaImpl(evaluator: Evaluator,
     </module>
   }
   def libraryXmlTemplate(name: String, path: os.Path, sources: Option[os.Path], scalaCompilerClassPath: Loose.Agg[Path]): Elem = {
-    def url(path: os.Path): String = if (path.ext == "jar") "jar://" + path + "!/" else "file://" + path
+    def url(path: os.Path): String = {
+      if (path.ext == "jar") "jar://$PROJECT_DIR$/" + path.relativeTo(cwd) + "!/"
```
lefou, line 503: We are using `cwd` which is relative to the Java process directory. Although I think it should work as expected in most cases, it would be good if we can use some better source for the current project directory.

ajrnz:
Could you please check that you can still run projects from within IntelliJ with a relative path setup. I seem to remember trying to do this in the past and there was some gotcha - however I could be mistaken.

build.sc
lihaoyi: I use a local coursier cache under *Windows Subsystem for Linux*, now just Windows! I do this because under WSL, your code is running in a VM, and so the home folder that coursier normally puts stuff in is inaccessible from the host machine that IntelliJ is running on.

We could pick another location; I've considered just `.coursier/`. It definitely is something you don't want to clean accidentally due to the time taken to download

scalalib/src/GenIdeaImpl.scala
lihaoyi: This `cwd` is passed in from elsewhere, and should be the base path for the `build.sc` file if I understand it right

PR #677, opened 2019-08-18
A hacky fix for the default scalafmt issue, but it works.

markehammons:
This change fixes issue #676, though in a hacky way. An alternative fix in the case that you can read the mill binary is described [here](https://stackoverflow.com/questions/25032716/getting-filesystemnotfoundexception-from-zipfilesystemprovider-when-creating-a-p/25033217)

Instantiating the zipfs using the instructions here got rid of the error outlined by the stack trace in the original issue, but the filesystem was constantly complaining that the file jar:file:/home/mhammons/mill-release!/mill/scalalib/scalafmt/default.scalafmt.conf doesn't exist.

Issue #676, opened 2019-08-18
Scalafmt module is broken


markehammons:
When attempting to reformat my projects using `project.reformat` or `mill.scalalib.scalafmt.ScalafmtModule/reformat _.sources`, I will get the following target failure:

```
mill.scalalib.scalafmt.ScalafmtModule.reformatAll java.nio.file.FileSystemNotFoundException
    com.sun.nio.zipfs.ZipFileSystemProvider.getFileSystem(ZipFileSystemProvider.java:171)
    com.sun.nio.zipfs.ZipFileSystemProvider.getPath(ZipFileSystemProvider.java:157)
    java.nio.file.Paths.get(Paths.java:143)
    mill.scalalib.scalafmt.ScalafmtWorker.reformatAction(ScalafmtWorker.scala:59)
    mill.scalalib.scalafmt.ScalafmtWorker.reformat(ScalafmtWorker.scala:32)
    mill.scalalib.scalafmt.ScalafmtModule$.$anonfun$reformatAll$2(ScalafmtModule.scala:32)
    scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
```

markehammons:
I have a fix for the stack trace, but the default scalafmt file can't be reached. java's NIO seems to be having trouble reading the data embedded inside the mill launcher/package

jodersky:
Looks like the issue only appears when attempting to extract a bundles default config for scalafmt. A workaround for this is to create the config file excplicity. E.g. `touch .scalafmt.conf`

markehammons:
@jodersky this is what my referenced fix does.

mgenereu:
I can confirm this bug. The `touch` method worked. This used to work.

PR #892, opened 2020-05-17
Fix fail `scalareformat` task execution when config file not exists

LangInteger:
Learning mills reading [reformat code part](http://www.lihaoyi.com/mill/page/configuring-mill.html#reformatting-your-code), the `scalareformat` task fails as #676 said. I'm using mill v0.7.0, problem still exists.

Inspired by #677, Here comes this pull request.

### DIFF
--- a/scalalib/src/scalafmt/ScalafmtWorker.scala
+++ b/scalalib/src/scalafmt/ScalafmtWorker.scala
@@ -10,6 +10,10 @@ import org.scalafmt.interfaces.Scalafmt
 import scala.collection.mutable
 import mill.api.Result
 import java.nio.file.Files
+import scala.util.Try
+import java.nio.file.Path
+import scala.util.Failure
+import scala.util.Success
 
 object ScalafmtWorkerModule extends ExternalModule {
   def worker: Worker[ScalafmtWorker] = T.worker { new ScalafmtWorker() }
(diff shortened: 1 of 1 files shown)


scalalib/src/scalafmt/ScalafmtWorker.scala
```diff
@@ -65,12 +66,15 @@ private[scalafmt] class ScalafmtWorker {
...
-        if (os.exists(scalafmtConfig.path))
+        if (isScalaFmtConfigExists)
           scalafmtConfig.path.toNIO
-        else
-          JPaths.get(getClass.getResource("default.scalafmt.conf").toURI)
-
+        else {
+          val temp = Files.createTempFile("temp", ".scalafmt.conf").toUri
+          JPaths.get(temp) 
+        }
```
lefou, line 76: Can we still try to get the default file as resource, and only fall back to the temp file in case that failed?

LangInteger: Agree, but the solution still seems not so elegant. When mill is used , unless customized scala format config file is placed, code always goes to the `create tmp file` branch. The default config file is never used.

I tested it via deleting `.scalafmt.conf` and `default.scalafmt.conf` files, and executing test cases already existed. It workes fine.

scalalib/src/scalafmt/ScalafmtWorker.scala
```diff
@@ -65,12 +70,23 @@ private[scalafmt] class ScalafmtWorker {
...
+      }
 
+      val (configPath, tmpFileCreated) =
+        if (os.exists(scalafmtConfig.path))
+          (scalafmtConfig.path.toNIO, false)
+        else {
+          readDefaultScalafmtConfig() match {
+            case Success(defaultConfig) => (defaultConfig, false)
+            case Failure(exception) => {
+              ctx.log.debug(s"Read default scalafmt config fail, create a temp one")
```
joan38, line 84: The `s` doesn't seem used here?

lefou:
Thank you!

scalalib/src/scalafmt/ScalafmtWorker.scala
LangInteger: My bad, I'll fix it someday.

PR #902, opened 2020-05-23
No need for -i for BSP

joan38:


lefou:
Is this covered by any test?

joan38:
@lefou I have a test ready but we need `evaluator.outPath` instead of `os.pwd / "out"` from https://github.com/lihaoyi/mill/pull/897 merged and rebase on it first

joan38:
Arf I tried to implement a similar test as https://github.com/lihaoyi/mill/blob/master/scalalib/test/src/GenIdeaExtendedTests.scala#L16
But I have 2 issues:
- It's still going to general the `.bsp` at the top of the `mill` project instead of the test target directory.
- It's running with `-i` anyway.

Any better idea here?

lefou:
> Arf, I tried to implement a similar test as https://github.com/lihaoyi/mill/blob/master/scalalib/test/src/GenIdeaExtendedTests.scala#L16
> But I have 2 issues with this approach:
> 
>     * It's still going to general the `.bsp` at the top of the `mill` project instead of the test target directory.

Why is that? Do we use the process current working directory instead of some better source for the project directory?

> 
>     * It's running with `-i` anyway.

I guess, this one will be hard to test from inside of mill. At least not with an in-process test setup. We probably need to spawn a sub-process and check its generated outcome afterwards. But I leave it to you to say that this is to much for this PR.

But having some test suite for checking the generated BSP files similar to the GenIdea tests is definitively desirable. :-)

joan38:
We are using `evaluator.outPath` do you have a better option?

lefou:
I think `Evaluator.outPath` is cortect, but under test it should hold a different location to start with.

joan38:
Actually we don't need the `out` folder but the one above so `evaluator.rootModule.millSourcePath` made it.

I put an initial test that we should extend in the future.
We can go ahead with this PR for now.

Thanks

PR #787, opened 2020-03-06
Dottydoc fix

heksesang:
Fix for #785, adds support for dottydoc in `ScalaModule.docJar`.

PR #904, opened 2020-05-24
Dottydoc fix

jodersky:
This is a rebase of #787 on master with some added tests.

In the last commit I added one additional change however: the scaladoc tool now has distinct classpath and ivydeps from the compiler. In Scala <= 2.13 this is not necessary, as the doc tool is included in the compiler artifact, however in dotty this is no longer the case. Feel free to drop that last commit if you'd rather not have the split.

### DIFF
--- a/scalalib/src/ScalaModule.scala
+++ b/scalalib/src/ScalaModule.scala
@@ -160,7 +160,7 @@ trait ScalaModule extends JavaModule { outer =>
     os.makeDir.all(javadocDir)
 
     if (isDotty(scalaVersion())) {
-      os.copy(millSourcePath / dottydocSiteRoot(), javadocDir, replaceExisting = true)
+      os.copy(dottydocSiteRoot(), javadocDir, replaceExisting = true)
     }
 
     val files = allSourceFiles().map(_.path.toString)
@@ -205,7 +205,7 @@ trait ScalaModule extends JavaModule { outer =>
 
   def dottydocProjectName = T { artifactName() }
 
-  def dottydocSiteRoot = T { "docs" }
+  def dottydocSiteRoot = T { millSourcePath / 'docs }
 
   /**
     * Opens up a Scala console with your module and all dependencies present,
--- a/scalalib/src/ScalaModule.scala
+++ b/scalalib/src/ScalaModule.scala
@@ -159,14 +159,28 @@ trait ScalaModule extends JavaModule { outer =>
     val javadocDir = outDir / 'javadoc
     os.makeDir.all(javadocDir)
 
+    if (isDotty(scalaVersion())) {
+      os.copy(millSourcePath / dottydocSiteRoot(), javadocDir, replaceExisting = true)
+    }
+
     val files = allSourceFiles().map(_.path.toString)
 
+    val docToolOptions =
+      if (isDotty(scalaVersion()))
+        Seq(
+          "-project", dottydocProjectName(),
+          "-siteroot", javadocDir.toNIO.toString
+        )
+      else
+        Seq(
+          "-d", javadocDir.toNIO.toString
+        )
     val pluginOptions = scalaDocPluginClasspath().map(pluginPathRef => s"-Xplugin:${pluginPathRef.path}")
     val compileCp = compileClasspath().filter(_.path.ext != "pom").map(_.path)
     val options = Seq(
-      "-d", javadocDir.toNIO.toString,
       "-classpath", compileCp.mkString(java.io.File.pathSeparator)
     ) ++
+      docToolOptions ++
       pluginOptions ++
       scalaDocOptions()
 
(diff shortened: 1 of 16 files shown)


jodersky:
fyi @heksesang and @lefou

scalalib/src/ScalaModule.scala
```diff
@@ -175,16 +203,24 @@ trait ScalaModule extends JavaModule { outer =>
...
+            else javadocDir
+          Result.Success(createJar(Agg(inputPath))(outDir))
         case false => Result.Failure("docJar generation failed")
       }
     }
   }
 
+  def dottydocProjectName = T { artifactName() }
+
+  def dottydocSiteRoot = T { millSourcePath / 'docs }
```
jodersky, line 222: these changes are from the original commits, but I do wonder if we should expose dotty-specific settings in the scala module? So far these are the only settings specific to dotty here.

heksesang: Well, once Dotty turns into Scala 3, we'd have to unify the settings if we want to support a single `ScalaModule` trait. And it would be odd to split them into separate traits, when it's just the next version of Scala. So it makes sense to expose the Dotty settings, though it might not be necessary in this case.

We could default to `-project ${artifactName()}` in the `scalaDocOptions` if it's a Dotty build, and maybe rename the `dottydocSiteRoot` to something more general like `docsSource` or something else if someone know a better name. The `dottydocSiteRoot` just defines where it should copy docs from, it's technically not dotty-specific, it could be used for any other doc generation tool as well which uses template files that has to be copied into the javadoc folder.

jodersky: I agree that it would be strange to split the ScalaModule trait into two. And since root and project name are actually required by dottydoc, I think it's fine to keep them as top-level defs here.

lefou: I think, we can skip the `dottydocProjectName` and make it the default in `scalaDocOptions`. Also ScalaDoc supports setting a title, and we already support changing/customization by overriding `scalaDocOptions`.

I'm also for a rename of the `dottydocSiteRoot` target. What about `docSources` ? I'm not sure if scaladoc can include documents too, but I know that alternative JavaDoc processors like Asciidoctlet also support additional files, so it's a nice general feature.

jodersky:
@lefou, @heksesang, as discussed, I moved the settings into separate targets. Let me know what you think of the latest commit.

heksesang:
The changes look good to me. :)

scalalib/src/ScalaModule.scala
```diff
@@ -169,34 +172,34 @@ trait ScalaModule extends JavaModule { outer =>
     os.makeDir.all(javadocDir)
 
     if (isDotty(scalaVersion())) {
-      val root = dottydocSiteRoot().path
-      if (os.exists(root)) {
-        os.copy(root, javadocDir, replaceExisting = true)
-      } else {
-        os.makeDir.all(javadocDir)
+      if (docSources().length > 1) {
+        T.ctx.log.error(
```
lefou, line 176: We could actually handle this situation by copying the content of all dirs into a new temporary location and using this instead.

jodersky: I thought of that, but then I also thought that it may not be worth the added complexity. Ideally dottydoc would actually support specifying multiple inputs (and also make the output path configurable), then we wouldn't need to handle merging directories.

In case you do think it's worthwhile to merge many input dirs, then I'd suggest we try a symlink based approach, where for every source directoy we symlink all contents into the final source dir (latter overwriting the former).

jodersky: Wdyt of something like
```
if (isDotty(scalaVersion())) {
      for {
        ref <- docSources()
        path = ref.path
        if os.exists(path) && os.isDir(path)
        children = os.list(path)
        src <- children
      } {
        val dest = javadocDir / src.last
        if (os.exists(dest)) os.remove(dest)
        os.symlink(dest, src)
      }
    }
```
?

jodersky: added that in the latest commit

lefou: I'm not sure if symlinks will work in all cases, e.g. Windows with FAT filesystem, if that is still a thing. Also, we should probably avoid that step, if we only have one exsisting docSources.

lefou: Let's remove that complexity then. Ideally, we could use just `T.source`, but that removes the optional nature. We can print the error in case we found more than one _existing_ docSource dir and the user can still work around by creating some merge targets themselves.

jodersky: :disappointed: oh well, I'll do a walk and copy then

jodersky: Up to you, I can also just walk all children and copy them instead of symlinking.

lefou: Ok. If more than one dir _exists_, copy them into a merge directory. Otherwise just use the original docSources. A `T.log.debug`  message when using the synthetic merge dir would be nice.

jodersky: Unfortunately we need the copy even for single docSources, as dottydoc (as it currently is) will will pollute the input directory with its output directory _site.

jodersky: Well, strictly speaking, we don't *need* the copy. However, if we don't then we'll modify something outside of the tasks `dest` dir. Not sure if that's acceptable?

lefou: Fine. But then merging is the most natural behaviour anyway.

jodersky: Ok, sounds good! I'll go ahead and implement the merging

jodersky: I also added a test to check that merging of multiple doc directories works as expected

lefou:
Thank you very much!

PR #911, opened 2020-05-30
BSP: Deduplicate targets to compile

joan38:
Due to an Intellij workaround (shared modules) to support cross compiled code we are receiving multiple times the same target.


joan38:
I'm thinking the same and I tried to fix it in Idea but that seemed complicated.

The example you gave is not applying here because clean and compile are 2 different REST calls to the BSP server (Tho is each requests you can give which targets you want to compile or clean).

Thanks!

PR #915, opened 2020-06-03
Replace scala-library artifact with dotty-library

jodersky:
This change is a "symptomatic treatment" to fix bloop config generation, which is currently broken for dotty projects.

In particular, [this issue](https://github.com/scalameta/metals/issues/1768) describes the error observed, and also hints to the fix.

The gist of it seems to be that Dotty publishes an empty jar named "scala-library" (not sure why?), but the actual standard library is published in the artifact "dotty-library". While the former is empty and isn't strictly a problem when compiling, it does cause bloop to crash if included in the classpath.

~Note that I am not exactly sure what the root cause is, and I'm also not sure what the future plans are regarding the naming of artifacts. It would be great to get feedback from someone more knowledgeable in this area before accepting these changes.~ (thanks for the review, Guillaume!)

### DIFF
--- a/scalalib/src/Lib.scala
+++ b/scalalib/src/Lib.scala
@@ -89,9 +89,16 @@ object Lib{
       // in Scala <= 2.13, the scaladoc tool is included in the compiler
       scalaCompilerIvyDeps(scalaOrganization, scalaVersion)
 
-  def scalaRuntimeIvyDeps(scalaOrganization: String, scalaVersion: String) = Agg[Dep](
-    ivy"$scalaOrganization:scala-library:$scalaVersion".forceVersion()
-  )
+  def scalaRuntimeIvyDeps(scalaOrganization: String, scalaVersion: String) =
+    if (mill.scalalib.api.Util.isDotty(scalaVersion))
+      Agg(
+        // note that dotty-library has a binary version suffix, hence the :: is necessary here
+        ivy"$scalaOrganization::dotty-library:$scalaVersion".forceVersion()
+      )
+    else
+      Agg(
+        ivy"$scalaOrganization:scala-library:$scalaVersion".forceVersion()
+      )
 
   def listClassFiles(base: os.Path): Iterator[String] = {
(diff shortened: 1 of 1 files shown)


smarter:
The empty scala-library we publish depend on dotty-library (which itself depends on the real scala 2 scala-library) and is only here in case someone has something like `scalaOrganization.value % "scala-lang" % scalaVersion.value` in their build (e.g., the sbt-ensime plugin used to require something like this), it will probably go away at some point. And in the not-too-distant future we'll probably rename dotty-{library,compiler,...} to scala3-{library,compiler,...}. scala3-library will still depend on the scala 2 scala-library to preserve binary compatibility.

scalalib/src/Lib.scala
```diff
@@ -89,9 +89,16 @@ object Lib{
...
       scalaCompilerIvyDeps(scalaOrganization, scalaVersion)
 
-  def scalaRuntimeIvyDeps(scalaOrganization: String, scalaVersion: String) = Agg[Dep](
-    ivy"$scalaOrganization:scala-library:$scalaVersion".forceVersion()
-  )
+  def scalaRuntimeIvyDeps(scalaOrganization: String, scalaVersion: String) =
+    if (mill.scalalib.api.Util.isDotty(scalaVersion))
+      Agg(
+        // note that dotty-library has a binary version suffix, hence the :: is necessary here
+        ivy"$scalaOrganization::dotty-library:$scalaVersion".forceVersion()
```
jodersky, line 96: @smarter, thanks for your explanation. Do you also know why dotty-library uses a binary suffix (_0.24), whereas the scala 2 lib does not?

smarter: I decided to use a binary suffix for dotty-library so that it would look like any other scala jar. I suspect scala-library doesn't use a suffix just because this convention was invented a long time after scala itself

jodersky: great, thanks. Was just curious :)

PR #918, opened 2020-06-13
Adding url to generated developer setting in published pom.

atooni:
This is just to include the url set in the Developer case class in the pom that is generated when publishing to Sonatype or to the local maven repo.


lefou:
Looks good to me

PR #920, opened 2020-06-19
Allows the ScalafmtModule to format build sources

hwielenberg:
This allows that the MillScalafmtModule can be utilised to format the mill build sources (such as build.sc) as well.

Here is an example how a command can be created to trigger the formatting:
```
def buildSources: Sources = T.sources(os.pwd / "build.sc")

def fmtBuild(): Command[Unit] =
  T.command {
    MillScalafmtModule.reformatAll(Tasks(Seq(buildSources)))
  }
```


lefou:
Thank you!

PR #921, opened 2020-06-19
Fix Scala.js 1.x where x != 0

lolgab:


lefou:
Thank you very much. I think our growing mill community is appreciating Scala.js 1.1 support. Do we have a test case that ensures, our setup works for Scala.js 1.0 as well as 1.1?

lolgab:
Ready for another review round :)

lolgab:
The failure in Travis seems unrelated.

lolgab:
Thank you for merging! 🎉

PR #923, opened 2020-06-22
Wrap class names in single quotes in launcher script

jodersky:
Quotes classnames to that bash doesn't interpret any `$` contained within them (as can happen with dotty's top-level main functions).

### DIFF
--- a/main/src/modules/Jvm.scala
+++ b/main/src/modules/Jvm.scala
@@ -376,7 +376,7 @@ object Jvm {
                               shebang: Boolean = false) = {
     universalScript(
       shellCommands =
-        s"""exec java ${jvmArgs.mkString(" ")} $$JAVA_OPTS -cp "${shellClassPath.mkString(":")}" $mainClass "$$@"""",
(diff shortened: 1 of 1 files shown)


main/src/modules/Jvm.scala
```diff
@@ -376,7 +376,7 @@ object Jvm {
                               shebang: Boolean = false) = {
     universalScript(
       shellCommands =
-        s"""exec java ${jvmArgs.mkString(" ")} $$JAVA_OPTS -cp "${shellClassPath.mkString(":")}" $mainClass "$$@"""",
+        s"""exec java ${jvmArgs.mkString(" ")} $$JAVA_OPTS -cp "${shellClassPath.mkString(":")}" '$mainClass' "$$@"""",
```
lefou, line 379: Yeah, makes sense. But we the different quoting sign?

jodersky: Not sure I follow. It has to be single quotes here, as double quotes would still expand any dollar sign.

lefou: By bad. You're completely right.

Issue #820, opened 2020-04-18
AssertionError when using predefScript.sc


francisdb:
Installed mill with homebrew

```
> mill --version
Mill Build Tool version 0.6.1
Java version: 13.0.2, vendor: N/A, runtime: /usr/local/Cellar/openjdk/13.0.2+8_2/libexec/openjdk.jdk/Contents/Home
Default locale: en_BE, platform encoding: UTF-8
OS name: "Mac OS X", version: 10.15.3, arch: x86_64
```
Followed docs here
http://www.lihaoyi.com/mill/page/configuring-mill.html#global-configuration

And when I start mill I see this:
```
> mill api:compile
Compiling /Users/myuser/.mill/ammonite/predefScript.sc
java.lang.AssertionError: assertion failed
  scala.Predef$.assert(Predef.scala:208)
  mill.define.Segment$Label.<init>(Ctx.scala:14)
  mill.define.Segment$Label$.apply(Ctx.scala:13)
  mill.define.Segment$Label$.apply(Ctx.scala:13)
  scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:237)
  scala.collection.IndexedSeqOptimized.foreach(IndexedSeqOptimized.scala:36)
  scala.collection.IndexedSeqOptimized.foreach$(IndexedSeqOptimized.scala:33)
  scala.collection.mutable.WrappedArray.foreach(WrappedArray.scala:39)
  scala.collection.TraversableLike.map(TraversableLike.scala:237)
  scala.collection.TraversableLike.map$(TraversableLike.scala:230)
  scala.collection.AbstractTraversable.map(Traversable.scala:108)
  mill.define.Segments$.labels(Ctx.scala:57)
  ammonite.predef.predefScript$.<init>(predefScript.sc:54)
  ammonite.predef.predefScript$.<clinit>(predefScript.sc)
```

Assertion errors are not very user friendly, this could give some more information on what is failing and why

https://github.com/lihaoyi/mill/blob/85aa6bdb1bd1d08fb2739a4fd75dd76b0ab6652c/main/core/src/define/Ctx.scala#L14

hypatian:
I'm having this same problem, both with the homebrew version (which appears to just be the bootstrap script) and with the bootstrap script by hand. I added to the assert to see where the `.` is being found, and discovered that the complaint is about the `.` in `.mill`. I'll see if I can figure out how to make this stop failing.

lefou:
[...]

PR #925, opened 2020-06-25
Handle dots in segment names for foreign modules

hypatian:
It seems like `.mill/ammonite/predefScript.sc` stopped working in mill 0.5.5. This encodes the segment label for `.mill` (and only `.mill`) as `"DOT-mill"` in order to pass the assertion, and also adds a message to the assertion so that if anything like this happens in the future it will be slightly easier to spot what it's coming from.

This will also encode `.mill` in other locations as `DOT-mill`, not just those in `~/.mill`.

One alternative would be to special-case `.mill` => `DOT-mill` in `Segments.labels` instead.

Another possibility would be to make `.<blah>` => `DOT-blah` globally, or make any dots into `DOT` to work with arbitrary directory names containing dots.

Both of those options would be more general and easier to test, but I opted for the smallest most specific fix that would allow `predefScript.sc` to start working again.

### DIFF
--- a/main/core/src/define/Ctx.scala
+++ b/main/core/src/define/Ctx.scala
@@ -11,7 +11,7 @@ sealed trait Segment{
 }
 object Segment{
   case class Label(value: String) extends Segment{
-    assert(!value.contains('.'))
+    assert(!value.contains('.'), s"$value contained a .")
   }
   case class Cross(value: Seq[Any]) extends Segment
 }
--- a/main/src/main/MainRunner.scala
+++ b/main/src/main/MainRunner.scala
@@ -179,7 +179,12 @@ class MainRunner(val config: ammonite.main.Cli.Config,
         // Encoding the number of `/..`
         val ups = if (relative.ups > 0) Seq(s"up-${relative.ups}") else Seq()
         val segs = Seq("foreign-modules") ++ ups ++ relative.segments
-        val segsList = segs.map(pprint.Util.literalize(_)).mkString(", ")
+        val segsList = segs
+          .map({
+            case ".mill" => "DOT-mill"
+            case seg => seg
+          })
+          .map(pprint.Util.literalize(_)).mkString(", ")
         s"Some(mill.define.Segments.labels($segsList))"
       }
       else "None"


main/core/src/define/Ctx.scala
```diff
@@ -11,7 +11,7 @@ sealed trait Segment{
 }
 object Segment{
   case class Label(value: String) extends Segment{
-    assert(!value.contains('.'))
+    assert(!value.contains('.'), s"$value contained a .")
```
lefou, line 14:
```suggestion
    require(!value.contains('.'), s"A label must not contain a '.' (actual value: '$value'")
```

An assert is probably not the best choice here?

hypatian:
> I'm not quite sure, what you fix here? Or maybe I do, but I don't get the connection to a `predefScript.sc`. Can you describe what kind of code/setup you have in your predef script? Also, If we fix the encoding of the special dir `~/.mill` here but the real issue is the fact that "hidden" directories (in Unix-like OSes) will break foreign modules, we should definitely fix it in a more general way. Also, we should probably add a test for it.

The connection to `predefScript.sc` is specifically that it lives under `~/.mill/ammonite` on Unixes, which means that it's impossible to do global configuration using `predefScript.sc` on these OSes. Since the ticket in question (#820) has been open since April, I figured I'd see if I could do a quick fix and at least get things moving towards a solution.

A general solution for hidden directories would be great, but as I described in my comment I wanted to make the smallest-possible fix to get things working, since I'm contributing to this project for the first time. I figured that if a broader solution was desired, that could be worked out in discussion on the PR.

As for what's in my `predefScript.sc` specifically: I'm trying to use it to work around proxy configuration problems using mill on Mac OS, so that I can tell my colleagues "just put this configuration file in place for now and you won't have to worry about it", since I'm moving us to Mill for this project. That kind of global user configuration is very helpful for workarounds of local environment oddities.

hypatian:
Given this discussion: I'd be happy to add code to make foreign modules either support hidden directories beginning with `.` characters or support any directories containing `.` characters in any location, if people want to chime in on what's best.

My inclination would be to support all directories for foreign modules, so `.mill` would become `DOT-mill`, `foo.bar` would become `foo-DOT-bar`, and so on. I'm not sure what impact having these segment labels not exactly matching the names of the source directories would have. Obviously, it means that there's be a collision between a directory named `foo.bar` and one named `foo-DOT-bar`. But I'm not sure how that gets further exposed.

Still, it seems like that would be the correct solution so that it's possible to do things like: ```
import $file.`foo.bar`.baz
```
for arbitrary paths containing dots.

**Comment by hypatian** [comment]
I guess another question is "why is this assertion in there to begin with?"

It was added in [6c66864](https://github.com/lihaoyi/mill/commit/6c66864dd45cb1f12524f9645b2c9158e5e0b592) which is named "WIP getting tests passing again", but it's not really clear to me what situation it's guarding against.

Testing with removing the assertion demonstrates it working for foreign modules, anyway, even for files with dots in their names, and not just for directories with dots:
```
~/src/X • ~/mill-release -i
Loading...
@ import $file.`foo.bar`
Compiling /Users/prevost1/src/X/foo.bar.sc
OHOHOH
import $file.$

@ `foo.bar`.z
res1: Int = 27
```

So: why was this assertion added? What problem does it prevent, and can it simply be removed?

lefou:
@lihaoyi Your input is much appreciated here. I've never used these predef stuff and also, besides being a heavy mill user, have not much experience with Ammonite. I haven't looked at the source code to get a better understanding though. My first (naive) thought why a dot may not be allowed was mill's cmdline syntax, but as @hypatian pointed out, it worked before mill 0.5.5 without any special encoding needed.

@hypatian Thanks for your investigations so far. It appears to me that commit https://github.com/lihaoyi/mill/commit/6c66864dd45cb1f12524f9645b2c9158e5e0b592 was already part of mill 0.1.0. (But I only used my normal browser-based GithHub foo, it's pretty late here.)

hypatian:
You're absolutely right that the assertion seems to have been in place for far longer than the problem has existed. I suspect that 0.5.5 changed something about how `predef.sc` and `predefScript.sc` (because I've confirmed the same thing happens for both) are compiled and used. Even though I hope there'll be a very easy solution, I'll still investigate more in the interest of understanding the internals of Mill better, since I seem to be digging rather deep right now.

Thanks very much for your comments and for poking me to look at the overall problem and not just my specific issue, @lefou.

Issue #926, opened 2020-06-26
Make cmdline options for mill-server mode and repl more clear


lefou:
We see confusion about the `-i` option quite a lot in Gitter channel and also advise to use `-i` very often when things don't work as expected.

I propose to deprecate the current `-i`, `--interactive` option in favour of two new options:

* `--repl` - Starts a interactive mill session and will implicitly not use the mill server (same as old `--interactive`)
* `--server 0|false|off|no` or `--no-server` - Starts mill without using a server process
* `--server 1|true|on|yes` - Starts mill with a background server process

hypatian:
I'd suggest that while server-related switches are being considered, a switch to force the server to restart or to stop should be added.

There will continue to be things that are frozen at server start time or build init time (whether unintentionally or because there's simply no other choice at the moment.) Not needing to find and kill the old server by hand would be a pleasant improvement.

hypatian:
Looking at this a bit right now. From what I can see, the reason `-i` is so fundamental (has to be the first argument, etc.) comes from way back in fd8a2567ad32c11bcf8adbaca85bdba72bb4f935.

When `-i` isn't used, it starts `MillClientMain` which is Java code and starts faster and then spawns off a separate process for the server which loads up Scala. And then if you _do_ use `-i`, it starts `mill.MillMain` which brings in the Scala libraries and is a bit slower.

`MillClientMain` currently doesn't interpret its arguments at all--it just shoves them all over to the server for handling.

Because of all of this it might be easiest to keep this as simple as possible, so it can be done in the launcher script without too much trouble.  So I'm going to suggest:

* `--repl` - Starts a interactive mill session and will implicitly not use the mill server (same as old `--interactive` with no target), and complains if a target is given.
* `--no-server` - Starts mill without using a server process (same as old `--interactive` with a target), and complains if no target is given.
* `--interactive` - Same as the current behavior for compatibility, `-i` does this. Maybe have this produce a warning that its behavior will be the same as `--no-server` at some point in the future.

All of these arguments are very simple (presence or absence), and can be restricted to "must be the very first argument". That makes them easy to handle with the current launcher script.

hypatian:
I just tossed up #927 as a first pass at doing this. It does what I suggested in my last comment.

lefou:
Merged right now. Thank you very much!

PR #927, opened 2020-06-27
Added --repl and --no-server and changed behavior of -i/--interactive for #926

hypatian:
This adds `--repl` and `--no-server` as options, and makes `-i` with no targets produce a warning message.

* `--repl` starts a build REPL, but fails if a target is given
* `--no-server` fails unless targets are given
* `-i`/`--interactive` is the same as `--no-server`, but emits a warning and starts a build REPL if no targets are given

Giving any of these arguments as the first option on the command-line causes `mill.MillMain` to run (interactive mode). Giving none of them as the first argument causes `mill.client.MillClientMain` to run. It is an error to give more than one of these options.

The only possible undesirable feature is that It is now possible to give one of these options more than once, with the first occurrence being as the first argument and another occurrence later in the switches. The "these switches have to be first" error is now emitted based on whether interactive mode is active when one of these switches are present. It would be possible to change the code to disallow any repeats if that's unacceptable.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -597,7 +597,7 @@ def launcherScript(shellJvmArgs: Seq[String],
          |    COURSIER_CACHE=.coursier ${java("mill.MillMain")}
          |else
          |    case "$$1" in
-         |      -i | --interactive )
+         |      -i | --interactive | --repl | --no-server )
          |        ${java("mill.MillMain")}
          |        ;;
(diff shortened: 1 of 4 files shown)


lihaoyi:
i think this looks reasonable

lefou:
Thank you!

PR #929, opened 2020-06-29
ZincProblemPosition: fix infinite loops

nafg:
JavaOptionConverter.convertInt overflowed the stack, probably because there is no Optional#isEmpty


nafg:
Should fix https://github.com/lihaoyi/mill/issues/814

nafg:
Thanks! Does this mean it's now available as a development version?

lefou:
> Thanks! Does this mean it's now available as a development version?

Yes. You may wait until the Travis CI job finished successfully (sometime it doesn't) and give Maven Central some minutes to sync though.

PR #930, opened 2020-06-29
BSP: Mill build Scala lib dep and forward env vars to server start

joan38:
This fixes:
- Some redness with the .sc files because if the missing Scala lib as dependency.
- The BSP server boot command to forward the env vars. That's useful when you set env vars like `JAVA_OPTS` or `COURSIER_REPOSITORIES`.

Thanks


joan38:
@lefou @lihaoyi This is ready to review

PR #931, opened 2020-06-29
Use default coursier repos

joan38:
Inspired by [this Gitter message](https://gitter.im/lihaoyi/mill?at=5ef62e5d6c06cd1bf4576e67).
Thanks @hypatian

### DIFF
--- a/scalalib/src/CoursierModule.scala
+++ b/scalalib/src/CoursierModule.scala
@@ -1,6 +1,6 @@
 package mill.scalalib
 
-import coursier.{Dependency, Repository}
+import coursier.{Dependency, Repository, Resolve}
 import mill.{Agg, T}
 import mill.define.Task
 import mill.eval.PathRef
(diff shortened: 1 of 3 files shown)


joan38:
@lefou @lihaoyi This is ready to review

lefou:
This change is probably some compatibility breaker for client code (aka `build.sc` files). So we should either but next mill version to `0.8` or wait with a merge.

joan38:
I think it would break only if you are implementing the `ZincWorkerModule` trait.
That seems uncommon but if we want to respect the semantic versioning yes 0.8 is the right way.

DamianReeves:
Is this going to be merged soon?

lefou:
Yes and no. It will introduce a binary incompatibility which might hit all those users who applied some repository overrides to `ZincWorkerModule`. So I'd like to merge it just shortly before we cut a mill `0.8'.

hypatian:
Just wanted to drop a note about this: Now that I've moved to getting things working in our internal CI setup, I've run into a need for this. (The CI server's access to the proxy is even more restrictive and requires fetching all artifacts through a cache, fun times!)

I have another workaround in place, but I'll be looking forward to the release of 0.8!

lihaoyi:
@lefou I'm propagating an update to my open source suite of libraries, which would require version bumps in a bunch of Mill dependencies. Let's thus merge this and call the next release 0.8

PR #932, opened 2020-07-02
Generate sourcemaps on scalajs 1.

### LINKED ISSUE #884: No source maps for 1.0.0
Here's my build.sc.  "> mill client.js" works fine using 'val scalaJsVer = "0.6.32"', but when I change to 'val scalaJsVer = "1.0.0"', it fails to find out.js.map.

tom@geeksite.org

```
// build.sc

import java.io.File

import mill._
import mill.scalalib._
import mill.scalajslib._
import coursier.maven.MavenRepository
import mill.define.Target

val scalaVer = "2.12.10"
val scalaJsVer = "0.6.32" // source maps worked with this
//val scalaJsVer = "1.0.0"

val developing = os.exists(os.home/".CHEAT")

object client extends ScalaJSModule {
  override def scalaVersion = T { scalaVer }
  def scalaJSVersion = T { scalaJsVer }
  def millSourcePath = os.pwd/"sjs"
  override def ivyDeps = Agg(ivy"org.scala-js::scalajs-dom::0.9.8")

  def repositories = super.repositories ++ Seq(
    MavenRepository("http://zeblon.redangus.org:8081/artifactory/ext-release-local")
  )

  /** Link scala js files from where we created them to where we want them */
  private def link(task: String): Unit = {
    val buildDir = os.pwd / "out" / "client" / task / "dest"
    val permDir = os.pwd/"src"/"main"/"webapp"/"scripts"/"scalajs"
    if (os.exists(permDir)) os.remove.all(permDir)
    os.makeDir.all(permDir)
    Seq("js", "js.map").foreach(ext =>
      os.hardlink(permDir/s"Spreadsheet.$ext", buildDir/s"out.$ext"))
  }

  /** Build fast of full depending where we are */
  def js: Target[Any] = if (developing) T {
    fastOpt()
    link("fastOpt")
  } else T {
    fullOpt()
    link("fullOpt")
  }
}
```

vic:
Sourcemaps were only being generated when using
`scalajs 0.6` but not for `scalajs 1`.

This changeset makes sure that we test that both
`fastOpt` and `fullOpt` will generate correct sourcemaps on all
scalajs supported versions.

Closes #884 

Context from gitter channel: https://gitter.im/lihaoyi/mill?at=5efcf211405be935cdd4e4c6

### DIFF
--- a/scalajslib/test/src/HelloJSWorldTests.scala
+++ b/scalajslib/test/src/HelloJSWorldTests.scala
@@ -111,8 +111,13 @@ object HelloJSWorldTests extends TestSuite {
         case FastOpt => HelloJSWorld.helloJsWorld(scalaVersion, scalaJSVersion).fastOpt
       }
       val Right((result, evalCount)) = helloWorldEvaluator(task)
-      val output = ScalaJsUtils.runJS(result.path)
+      val jsFile = result.path
+      val output = ScalaJsUtils.runJS(jsFile)
       assert(output == "Hello Scala.js")
+      val sourceMap= jsFile / os.up / (jsFile.last + ".map")
+      assert(sourceMap.toIO.exists()) // sourceMap file was generated
+      assert(os.read(jsFile).contains("//# sourceMappingURL=out.js.map")) // jsFile references sourceMap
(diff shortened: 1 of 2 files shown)


vic:
@lefou, could you please review ?

lefou:
Thank you!

PR #933, opened 2020-07-03
Passing args to launcher

LMnet:
Hi. I'm trying to fix https://github.com/lihaoyi/mill/issues/577. The pull request is still in progress. I want to verify my solution before full implementation.

I propose to pass extra JVM arguments to the launcher through `.mill-args` file in the root of the project. sbt using the same approach with `.sbtopts` file.

At the current moment, only mill server on linux/macos could recognize `.mill-args` file. Support for windows and interactive mode is needed for full feature implementation.

What do you think about this solution?

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -55,10 +55,10 @@ object Deps {
   val jettyServer = ivy"org.eclipse.jetty:jetty-server:8.1.16.v20140903"
   val jettyWebsocket =  ivy"org.eclipse.jetty:jetty-websocket:8.1.16.v20140903"
   val jgraphtCore = ivy"org.jgrapht:jgrapht-core:1.3.0"
-  
+
   val jna = ivy"net.java.dev.jna:jna:5.0.0"
   val jnaPlatform = ivy"net.java.dev.jna:jna-platform:5.0.0"
-  
+
   val junitInterface = ivy"com.novocode:junit-interface:0.11"
   val lambdaTest = ivy"de.tototec:de.tobiasroeser.lambdatest:0.7.0"
   val osLib = ivy"com.lihaoyi::os-lib:0.7.1"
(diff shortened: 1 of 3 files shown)


lihaoyi:
Seems reasonable to me. @lefou what do you think?

lefou:
In general, I support the idea of that approach. But if we now magically support `-X` options in that file, we should also support them on the regular command line, shouldn't we? Also people will wonder why regular mill cmdline options in that file will be ignored, e.g. adding a `-j 0`. Or maybe it's just the name that triggers some assumptions in me? If this file is only about JVM options at all but not about mill options, we might call it better `.mill-jvm-opts` or something like this.

LMnet:
@lefou 
> But if we now magically support -X options in that file, we should also support them on the regular command line, shouldn't we?

I thought about it too. We may add a variable like `MILL_ARGS` or `MILL_JVM_OPTS` for that. And it could be used for passing JVM options to the mill launcher. But, I think it could be done in a separate issue, as an improvement of a current file approach.

> Also people will wonder why regular mill cmdline options in that file will be ignored, e.g. adding a -j 0. Or maybe it's just the name that triggers some assumptions in me? If this file is only about JVM options at all but not about mill options, we might call it better .mill-jvm-opts or something like this.

Currently, `.mill-args` support only JVM options. And the original issue was exactly about passing JVM options to the mill. I'm not happy with the current naming too, it's mostly temporary. I think `.mill-jvm-opts` is a lot better.

lefou:
Yeah, ho ahead with the more specific name. Leaving the `-X` support for another PR, if needed.

LMnet:
@lihaoyi @lefou pull request is ready for the review.

Passing JVM options are now possible on all modes (client-server and interactive) and all operating systems. Windows support was hard :)

LMnet:
Should I do something else to make it merged?

lefou:
@LMnet Could you merge or rebase to latest master to work around the failing travis tests. This would make me more confident about merging.

LMnet:
@lefou sure. I rebased my branch.

LMnet:
CI is still failing on some environments, but it looks like these errors are unrelated to my changes:
```
E: Unable to locate package libunwind8-dev
The command "curl https://raw.githubusercontent.com/scala-native/scala-native/master/scripts/travis_setup.sh | bash -" failed and exited with 100 during .
```

lefou:
I restarted the failing jobs.

LMnet:
Finally green!

PR #934, opened 2020-07-05
Fix ScalaJS dependencies versions

lolgab:
I noticed that the ScalaJS build doesn't work with ScalaJS 1.0.1.
This is because it derives the `scalajs-env-nodejs` and `scalajs-env-jsdom-nodejs` dependencies from the ScalaJS version instead of using fixed versions, since they are not tied to the ScalaJS version.


lefou:
Thanks!

lolgab:
Thank you @sjrd and @lefou for the reviews and for merging this! 🙏

PR #940, opened 2020-07-27
WIP: Play 2.8

imxieshuang:
Mill added support for play2.8

### DIFF
--- a/contrib/playlib/worker/2.8/src/RouteCompilerWorker.scala
+++ b/contrib/playlib/worker/2.8/src/RouteCompilerWorker.scala
@@ -1,0 +1,79 @@
+package mill
+package playlib
+package worker
+
+import java.io.File
+
+import mill.playlib.api.{RouteCompilerType, RouteCompilerWorkerApi}
+import play.routes.compiler
+import play.routes.compiler.RoutesCompiler.RoutesCompilerTask
+import play.routes.compiler.{InjectedRoutesGenerator, RoutesCompilationError, RoutesCompiler, RoutesGenerator}
+
+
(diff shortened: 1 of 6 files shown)


contrib/playlib/worker/2.8/src/RouteCompilerWorker.scala
```diff
@@ -0,0 +1,79 @@
...
+                       reverseRouter: Boolean,
+                       namespaceReverseRouter: Boolean,
+                       generatorType: RouteCompilerType,
+                       dest: java.io.File): String = {
+    generatorType match {
+      case RouteCompilerType.InjectedGenerator =>
+        val result = compileWithPlay(files.map(os.Path(_)), additionalImports, forwardsRouter, reverseRouter,
+          namespaceReverseRouter, os.Path(dest), InjectedRoutesGenerator)
+        asMillResult(result)
+      case RouteCompilerType.StaticGenerator =>
```
lefou, line 27: Does just returning a string here means, mill will not detect any error but continue it's work?

contrib/playlib/worker/2.8/src/RouteCompilerWorker.scala
```diff
@@ -0,0 +1,79 @@
...
+        "Static generator was deprecated in 2.6.0 then removed in 2.7.x, see https://www.playframework.com/documentation/2.7.x/Migration27#StaticRoutesGenerator-removed"
```
lefou, line 28: Does just returning a string here means, mill will not detect any error but continue it's work?

lefou:
Close due to inactivity.

PR #941, opened 2020-07-28
Exclude trees from ammonite dep.

ckipp01:
The reason for doing this is to ensure that the version of
trees that mill wants is indeeed used instead of the newer
version that is being pulled in by ammonite-interp. Before
this was using `forceVersion()`, which was added https://github.com/lihaoyi/mill/compare/438ebc...6b492a, but by doing this the forced
version doesn't end up in the POM meaning that when someone
bootstraps/uses mill via coursier, they are getting the newer
trees, which is causing issues with how the main class is detected.

You can see we are getting `4.3.20` here instead of the _forced_ `4.3.7`:

```
❯ cs resolve mill | grep trees
org.scalameta:trees_2.13:4.3.20:default
```

This is coming from ammonite-interp:

```
❯ cs resolve mill --what-depends-on org.scalameta:trees_2.13
  Result:
└─ org.scalameta:trees_2.13:4.3.20
   ├─ com.lihaoyi:ammonite-interp_2.13.2:2.2.0
   │  ├─ com.lihaoyi:ammonite-repl_2.13.2:2.2.0
   │  │  └─ com.lihaoyi:ammonite_2.13.2:2.2.0
   │  │     └─ com.lihaoyi:mill-main-core_2.13:0.8.0
   │  │        └─ com.lihaoyi:mill-main_2.13:0.8.0
   │  │           └─ com.lihaoyi:mill-scalalib_2.13:0.8.0
   │  │              ├─ com.lihaoyi:mill-dev_2.13:0.8.0
   │  │              ├─ com.lihaoyi:mill-scalajslib_2.13:0.8.0
   │  │              │  └─ com.lihaoyi:mill-dev_2.13:0.8.0
   │  │              └─ com.lihaoyi:mill-scalanativelib_2.13:0.8.0
   │  │                 └─ com.lihaoyi:mill-dev_2.13:0.8.0
   │  └─ com.lihaoyi:ammonite_2.13.2:2.2.0
   │     └─ com.lihaoyi:mill-main-core_2.13:0.8.0
   │        └─ com.lihaoyi:mill-main_2.13:0.8.0
   │           └─ com.lihaoyi:mill-scalalib_2.13:0.8.0
   │              ├─ com.lihaoyi:mill-dev_2.13:0.8.0
   │              ├─ com.lihaoyi:mill-scalajslib_2.13:0.8.0
   │              │  └─ com.lihaoyi:mill-dev_2.13:0.8.0
   │              └─ com.lihaoyi:mill-scalanativelib_2.13:0.8.0
   │                 └─ com.lihaoyi:mill-dev_2.13:0.8.0
   └─ com.lihaoyi:mill-main-core_2.13:0.8.0 org.scalameta:trees_2.13:4.3.7 -> 4.3.20
      └─ com.lihaoyi:mill-main_2.13:0.8.0
         └─ com.lihaoyi:mill-scalalib_2.13:0.8.0
            ├─ com.lihaoyi:mill-dev_2.13:0.8.0
            ├─ com.lihaoyi:mill-scalajslib_2.13:0.8.0
            │  └─ com.lihaoyi:mill-dev_2.13:0.8.0
            └─ com.lihaoyi:mill-scalanativelib_2.13:0.8.0
               └─ com.lihaoyi:mill-dev_2.13:0.8.0
```

After this change you can see that now the desired `4.3.7` is here:
```
❯ cs resolve mill:0.8.0-2-b4e258 --what-depends-on org.scalameta:trees_2.13
  Result:
└─ org.scalameta:trees_2.13:4.3.7
   └─ com.lihaoyi:mill-main-core_2.13:0.8.0-2-b4e258
      └─ com.lihaoyi:mill-main_2.13:0.8.0-2-b4e258
         └─ com.lihaoyi:mill-scalalib_2.13:0.8.0-2-b4e258
            ├─ com.lihaoyi:mill-dev_2.13:0.8.0-2-b4e258
            ├─ com.lihaoyi:mill-scalajslib_2.13:0.8.0-2-b4e258
            │  └─ com.lihaoyi:mill-dev_2.13:0.8.0-2-b4e258
            └─ com.lihaoyi:mill-scalanativelib_2.13:0.8.0-2-b4e258
               └─ com.lihaoyi:mill-dev_2.13:0.8.0-2-b4e258
```

You can find more info about all of this in detail here: https://github.com/coursier/apps/issues/56 thanks to @alexarchambault and his investigative skills.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -41,6 +41,12 @@ object Deps {
 
   val acyclic = ivy"com.lihaoyi::acyclic:0.2.0"
   val ammonite = ivy"com.lihaoyi:::ammonite:2.2.0"
+  // Exclude trees here to force the version of we have defined. We use this
+  // here instead of a `forceVersion()` on scalametaTrees since it's not
+  // respected in the POM causing issues for Coursier Mill users.
+  val ammoniteExcludingTrees = ammonite.exclude(
+    "org.scalameta" -> "trees_2.13"
+  )
   val scalametaTrees = ivy"org.scalameta::trees:4.3.7"
   val bloopConfig = ivy"ch.epfl.scala::bloop-config:1.4.0-RC1"
   val coursier = ivy"io.get-coursier::coursier:2.0.0-RC6-15"
@@ -161,8 +167,8 @@ object main extends MillModule {
     )
 
     def ivyDeps = Agg(
-      Deps.ammonite,
-      Deps.scalametaTrees.forceVersion(),
+      Deps.ammoniteExcludingTrees,
+      Deps.scalametaTrees,
       Deps.coursier,
       // Necessary so we can share the JNA classes throughout the build process
       Deps.jna,


lihaoyi:
> but by doing this the forced
version doesn't end up in the POM meaning that when someone
bootstraps/uses mill via coursier, they are getting the newer
trees, which is causing issues with how the main class is detected.

Does anyone know if POM files have a way of forcing dependencies? IIRC there was a `<dependencyManagement>` section that could be used to pin versions; if so, we should fix Mill to generate a version-pinning pom rather than doing an exclusion

ckipp01:
>  Does anyone know if POM files have a way of forcing dependencies? IIRC there was a <dependencyManagement> section that could be used to pin versions; if so, we should fix Mill to generate a version-pinning pom rather than doing an exclusion

But wouldn't the most common way to do this using  `<dependencyManagement>` just be to also use an exclusion?
```
<dependencyManagement>
	<dependency>
	  <groupId>com.lihaoyi</groupId>
	  <artifactId>mill-main-core_2.13</artifactId>
	  <version></version>
	  <exclusions>
	    <exclusion>
	      <artifactId>org.scalameta</artifactId>
	      <groupId>trees_2.13</groupId>
	    </exclusion>
	  </exclusions>
	<dependency>
</dependencyManagement>
```

alexarchambault:
> Does anyone know if POM files have a way of forcing dependencies? 

There aren't AFAIK. The dependency management of transitive dependencies cannot be used to force versions of dependencies they pulled. (I just tested that in [this project](https://github.com/alexarchambault/mvn-transitive-dep-mgmt) for confirmation).

If project A transitively depends on `foo:2.2`, but puts `foo:1.3` in its dependency management, `foo:1.3` will be pulled when building A from Maven. But if A gets published, and project B depends on A, B will pull `foo:2.2` via A.

I had thought about having coursier depart from how Maven does, and use dependency management as a way to force versions too. But for now, it just does the same as Maven.

ckipp01:
> Does anyone know if POM files have a way of forcing dependencies? IIRC there was a <dependencyManagement> section that could be used to pin versions; if so, we should fix Mill to generate a version-pinning pom rather than doing an exclusion

Since this doesn't seem to be possible. Is the exclusion alright?

lefou:
Ok, after reading this PR again I understand the context a bit better and think we want this PR merged. Essentially, it tries to work around some incompatible transitive dependencies by excluding and explicitly listing them.

Chris (@ckipp01), it would be very nice to have a code comment (explaining that fact) as part of the patch, to make future maintenance more easy.

ckipp01:
> Ok, after reading this PR again I understand the context a bit better and think we want this PR merged. Essentially, it tries to work around some incompatible transitive dependencies by excluding and explicitly listing them.
> 
> Chris (@ckipp01), it would be very nice to have a code comment (explaining that fact) as part of the patch, to make future maintenance more easy.

Sure, I've gone ahead and added a short description in to explain why the `exclude` instead of the `forceVersion`.

lefou:
Thanks!

PR #947, opened 2020-08-13
Ability to relocate/shade in assembly

joan38:
Resolves https://github.com/lihaoyi/mill/issues/355

This gives the ability to relocate/shade in assembly on a global level.
This is a good first step but we might want later to provide the ability to shade only for a given lib. This is currently not possible because at the time we get the class path we lost the information of the dependency tree.

Example of usage:
```scala
  override def assemblyRules =
    Assembly.defaultRules ++ Seq(Assembly.Rule.Relocate("shapeless.**", "com.netflix.data.playback.shade.shapless.@1"))
```

### DIFF
--- a/main/src/modules/Assembly.scala
+++ b/main/src/modules/Assembly.scala
@@ -3,10 +3,8 @@ package mill.modules
 import java.io.InputStream
 import java.util.jar.JarFile
 import java.util.regex.Pattern
-
-import geny.Generator
 import mill.Agg
-
+import os.Generator
 import scala.collection.JavaConverters._
 
 object Assembly {
@@ -38,7 +36,10 @@ object Assembly {
     case class ExcludePattern(pattern: Pattern) extends Rule
   }
 
-  def groupAssemblyEntries(inputPaths: Agg[os.Path], assemblyRules: Seq[Assembly.Rule]): Map[String, GroupedEntry] = {
+  def groupAssemblyEntries(
+    mappings: Generator[(String, UnopenedInputStream)],
+    assemblyRules: Seq[Assembly.Rule]
+  ): Map[String, GroupedEntry] = {
     val rulesMap = assemblyRules.collect {
       case r@Rule.Append(path, _) => path -> r
       case r@Rule.Exclude(path) => path -> r
@@ -52,23 +53,19 @@ object Assembly {
       case Rule.ExcludePattern(pattern) => pattern.asPredicate().test(_)
     }
 
-    classpathIterator(inputPaths).foldLeft(Map.empty[String, GroupedEntry]) {
-      case (entries, entry) =>
-        val mapping = entry.mapping
-
+    mappings.foldLeft(Map.empty[String, GroupedEntry]) {
+      case (entries, (mapping, entry)) =>
         rulesMap.get(mapping) match {
           case Some(_: Assembly.Rule.Exclude) =>
             entries
           case Some(a: Assembly.Rule.Append) =>
             val newEntry = entries.getOrElse(mapping, AppendEntry(Nil, a.separator)).append(entry)
             entries + (mapping -> newEntry)
-
           case _ if excludePatterns.exists(_(mapping)) =>
             entries
           case _ if appendPatterns.exists(_(mapping)) =>
             val newEntry = entries.getOrElse(mapping, AppendEntry.empty).append(entry)
             entries + (mapping -> newEntry)
-
           case _ if !entries.contains(mapping) =>
             entries + (mapping -> WriteOnceEntry(entry))
           case _ =>
@@ -77,52 +74,36 @@ object Assembly {
     }
   }
 
-  private def classpathIterator(inputPaths: Agg[os.Path]): Generator[AssemblyEntry] = {
-    Generator.from(inputPaths)
-      .filter(os.exists)
-      .flatMap {
-        p =>
-          if (os.isFile(p)) {
-            val jf = new JarFile(p.toIO)
-            Generator.from(
-              for(entry <- jf.entries().asScala if !entry.isDirectory)
-                yield JarFileEntry(entry.getName, () => jf.getInputStream(entry))
-            )
-          }
-          else {
-            os.walk.stream(p)
-              .filter(os.isFile)
-              .map(sub => PathEntry(sub.relativeTo(p).toString, sub))
-          }
+  def loadClasspath(
+    inputPaths: Agg[os.Path]
+  ): Generator[(String, UnopenedInputStream)] = {
+    Generator.from(inputPaths).filter(os.exists).flatMap { path =>
+      if (os.isFile(path)) {
+        val jarFile = new JarFile(path.toIO)
+        Generator.from(jarFile.entries().asScala.filterNot(_.isDirectory))
+          .map(entry => entry.getName -> (() => jarFile.getInputStream(entry)))
       }
+      else {
+        os.walk
+          .stream(path)
+          .filter(os.isFile)
+          .map(subPath => subPath.relativeTo(path).toString -> (() => os.read.inputStream(subPath)))
+      }
+    }
   }
-}
 
-private[modules] sealed trait GroupedEntry {
-  def append(entry: AssemblyEntry): GroupedEntry
-}
+  type UnopenedInputStream = () => InputStream
 
-private[modules] object AppendEntry {
-  val empty: AppendEntry = AppendEntry(Nil, Assembly.defaultSeparator)
-}
-
-private[modules] case class AppendEntry(entries: List[AssemblyEntry], separator: String) extends GroupedEntry {
-  def append(entry: AssemblyEntry): GroupedEntry = copy(entries = entry :: this.entries)
-}
-
-private[modules] case class WriteOnceEntry(entry: AssemblyEntry) extends GroupedEntry {
-  def append(entry: AssemblyEntry): GroupedEntry = this
-}
-
-private[this] sealed trait AssemblyEntry {
-  def mapping: String
-  def inputStream: InputStream
-}
-
-private[this] case class PathEntry(mapping: String, path: os.Path) extends AssemblyEntry {
-  def inputStream: InputStream = os.read.inputStream(path)
-}
-
-private[this] case class JarFileEntry(mapping: String, getIs: () => InputStream) extends AssemblyEntry {
-  def inputStream: InputStream = getIs()
+  private[modules] sealed trait GroupedEntry {
+    def append(entry: UnopenedInputStream): GroupedEntry
+  }
+  private[modules] object AppendEntry {
+    val empty: AppendEntry = AppendEntry(Nil, defaultSeparator)
+  }
+  private[modules] case class AppendEntry(inputStreams: Seq[UnopenedInputStream], separator: String) extends GroupedEntry {
+    def append(inputStream: UnopenedInputStream): GroupedEntry = copy(inputStreams = inputStreams :+ inputStream)
+  }
+  private[modules] case class WriteOnceEntry(inputStream: UnopenedInputStream) extends GroupedEntry {
+    def append(entry: UnopenedInputStream): GroupedEntry = this
+  }
 }
--- a/main/src/modules/Assembly.scala
+++ b/main/src/modules/Assembly.scala
@@ -1,11 +1,13 @@
 package mill.modules
 
-import java.io.InputStream
+import com.eed3si9n.jarjarabrams.{ShadePattern, Shader}
+import java.io.{ByteArrayInputStream, InputStream}
 import java.util.jar.JarFile
 import java.util.regex.Pattern
 import mill.Agg
 import os.Generator
 import scala.collection.JavaConverters._
+import scala.tools.nsc.io.Streamable
 
 object Assembly {
 
@@ -29,6 +31,8 @@ object Assembly {
     case class AppendPattern(pattern: Pattern) extends Rule
 
     case class Exclude(path: String) extends Rule
+
+    case class Relocate(from: String, to: String) extends Rule
 
     object ExcludePattern {
       def apply(pattern: String): ExcludePattern = ExcludePattern(Pattern.compile(pattern))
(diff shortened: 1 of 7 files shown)


alexarchambault:
@joan38 Did you look at [jarjar-abrams](https://github.com/eed3si9n/jarjar-abrams), instead of jar-relocator? It allows to shade / relocate classes, but it also adjusts the scala annotations accordingly, so that the shaded classes can be used from scala code too (if library A is shaded in library B, and scala users depend on B, they'll be able to reference the shaded classes of A thanks to that. This doesn't work with other shading / relocating libraries, because of the discrepancy after shading between the - shaded - class names and the - non-shaded - scala annotations).

joan38:
Hi @alexarchambault,

I saw it while trying to implement the same as sbt-assembly and it's usage [here](https://github.com/sbt/sbt-assembly/blob/master/src/main/scala/sbtassembly/Assembly.scala#L225) and [here](https://github.com/sbt/sbt-assembly/blob/master/src/main/scala/sbtassembly/Assembly.scala#L252) and got stuck on how to use the lib (missing documentation).
But after another look at it actually I could use this one. I will need to unzip jars first. And we could reuse the same shading rules in a different Target.

joan38:
Sound like https://github.com/pantsbuild/jarjar is archived and has no activity since 2018. Is it safe to use that?

alexarchambault:
> Sound like https://github.com/pantsbuild/jarjar is archived and has no activity since 2018. Is it safe to use that?

Someone will probably fork it at some point… It's originally a fork itself too.

alexarchambault:
> If I need some help can I ask you?

Sure!

alexarchambault:
As another example, https://github.com/coursier/sbt-shading uses pantsbuild/jarjar. I need to migrate it to jarjar-abrams.

joan38:
@alexarchambault I have done the switch to [jarjar-abrams](https://github.com/eed3si9n/jarjar-abrams) but I didn't want to unzip, write jars, modify/shade jars, recopy jars. It's too many operations on disk.

So I copy and modified the `Shader` object to add support for `InputStream` shading, what do you guys think, cc @eed3si9n? If that sounds good to you I can make a PR in [jarjar-abrams](https://github.com/eed3si9n/jarjar-abrams) and then we can remove the `Shaderr` object from here.

Also if it's easier for you, I'd be happy to schedule a 30min call with you guys, let me know.

eed3si9n:
@joan38 Feel free to send a PR to jarjar-abrams. I've split it out of sbt-assembly so everyone can use it.

joan38:
Alright folks we are on jarjar-abrams 0.2.0 that supports shading over `InputStream`s.
@lefou @lihaoyi Could you review?
My team is in need of this shading capability as we are using Spark with shapeless and there is a clash of versions.

joan38:
Following from the Gitter chat with @lihaoyi, this "looks reasonable at a glance" but would like someone else to also review.
Can I ask a review from you @alexarchambault ?

alexarchambault:
> Can I ask a review from you @alexarchambault ?

I never really reviewed mill code, but I can have a look. It would have been nice to minimize the diff somehow… It seems there's quite a few added / removed empty lines, renamed variables, re-formatting, that hinder reviewing.

joan38:
@alexarchambault I just push the change in 2 separate commits for easier review.

main/src/modules/Assembly.scala
```diff
@@ -70,25 +73,33 @@ object Assembly {
...
         if (os.isFile(path)) {
           val jarFile = new JarFile(path.toIO)
-          jarFile
+          val mappings = jarFile
             .entries()
             .asScala
             .filterNot(_.isDirectory)
-            .map(entry => entry.getName -> jarFile.getInputStream(entry))
+            .map(entry => jarFile.getInputStream(entry) -> entry.getName)
+          Shader.shadeInputStreams(shadeRules, mappings.toSeq, verbose = false)
```
joan38, line 91: Shading jar entries

main/src/modules/Assembly.scala
```diff
@@ -70,25 +73,33 @@ object Assembly {
...
         }
         else {
-          os
+          val pathsWithMappings = os
             .walk(path)
             .filter(os.isFile)
-            .map(subPath => subPath.relativeTo(path).toString -> os.read.inputStream(subPath))
+              .map(subPath => os.read.inputStream(subPath) -> subPath.relativeTo(path).toString)
+          Shader.shadeInputStreams(shadeRules, pathsWithMappings, verbose = false)
```
joan38, line 98: Shading class files (that are not jars)

main/src/modules/Assembly.scala
```diff
@@ -70,25 +73,33 @@ object Assembly {
     }
   }
 
-  private def classpathIterator(inputPaths: Agg[os.Path]): Agg[(String, InputStream)] =
+  private def classpathIterator(inputPaths: Agg[os.Path], assemblyRules: Seq[Assembly.Rule]): Agg[(String, InputStream)] = {
+    val shadeRules = assemblyRules.collect {
+      case Rule.Relocate(from, to) => ShadePattern.Rename(List(from -> to)).inAll
```
joan38, line 78: I prefer exposing our own API (`Rule.Relocate`) and mapping it to jarjar-abram's `ShadePattern.Rename` in case we are swapping the shading lib.

main/src/modules/Assembly.scala
```diff
@@ -29,6 +30,8 @@ object Assembly {
 
     case class Exclude(path: String) extends Rule
 
+    case class Relocate(from: String, to: String) extends Rule
```
joan38, line 33: Later we can add more config on this to apply the shading only to specific libs.

alexarchambault (commented):
I left some comments. It would have been nice to have a test for that too, although I'm not too familiar with how things are tested in this repo. I'll let @lefou or @lihaoyi decide whether some tests should be added.

main/src/modules/Assembly.scala
```diff
@@ -77,52 +70,39 @@ object Assembly {
     }
   }
 
-  private def classpathIterator(inputPaths: Agg[os.Path]): Generator[AssemblyEntry] = {
-    Generator.from(inputPaths)
+  private def classpathIterator(inputPaths: Agg[os.Path]): Agg[(String, InputStream)] =
```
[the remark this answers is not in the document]
joan38: I changed the first level `inputPaths` back to `Generator.from(inputPaths)` but the `os.walk` is not possible unless jarjar-abram supports `Generator` (it looks for a `Seq`).

main/src/modules/Assembly.scala
```diff
@@ -77,52 +70,39 @@ object Assembly {
...
-              .map(sub => PathEntry(sub.relativeTo(p).toString, sub))
-          }
+      .flatMap { path =>
+        if (os.isFile(path)) {
+          val jarFile = new JarFile(path.toIO)
+          jarFile
+            .entries()
+            .asScala
+            .filterNot(_.isDirectory)
+            .map(entry => entry.getName -> jarFile.getInputStream(entry))
```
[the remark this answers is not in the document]
joan38: jarjar-abram is looking for a `Seq[(String, InputStream)]`.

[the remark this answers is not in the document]
alexarchambault: Oh, I see. Maybe then, `Shader.shadeInputStreams` in jarjar-abrams could be changed to something like
```scala
def processByteCode(
    rules: Seq[ShadeRule],
    verbose: Boolean
): (String, Array[Byte]) => Option[(String, Array[Byte])]
```
This method could be used to process the `InputStream`s one by one.

main/src/modules/Assembly.scala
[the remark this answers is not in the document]
alexarchambault: I think with the signature in my comment below, we shouldn't need to have jarjar-abrams be aware of the collection here.

main/src/modules/Assembly.scala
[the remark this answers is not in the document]
joan38: I've done `(String, Array[Byte]) => Option[(String, Array[Byte])]` in https://github.com/eed3si9n/jarjar-abrams/pull/6
I monkey patched in this PR to see how it plays. What's your thought on this?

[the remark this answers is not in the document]
joan38: I pushed again with `Array[Byte]` let me know what you think.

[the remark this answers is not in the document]
joan38: After thinking about this, I feel like I prefer the `InputStream` over `Array[Byte]` because at least with an `InputStream` we are not loading all the jars in memory if there is no rules. And also in the future we can think of jarjar-abram or another one that could process the shading on the fly inside an `InputStream`. Maybe?

lefou:
I'm back from my bike trip and happy to review this. Just give a short bump when you think your PR is ready for it, e.g. all required changes in thirdparty libs are settled. Also we should include a test case that uses this class relocation feature and ensures relocated classes work as expected.

joan38:
@alexarchambault I pushed a new version with the change you suggested in jarjar-abram.
What do you think?

main/src/modules/Shader.scala
```diff
@@ -0,0 +1,93 @@
...
+        bytes = Files.readAllBytes(path)
+        _ = Files.delete(path)
+        (shadedBytes, shadedName) <- shader(bytes, name)
+        out = dir.resolve(shadedName)
+        _ = Files.createDirectories(out.getParent)
+        _ = Files.write(out, shadedBytes, StandardOpenOption.CREATE)
+      } yield ()
+    }
+
+  def bytcodeShader(
```
ches, line 29: Fly-by remark: little typo here, `s/byt/byte`

I'm assuming the monkey patch should get moved to jarjar-abrams proper before this PR is finalized.

joan38:
@lefou this is good to go.

lefou:
> Looks good to me. As we remove the whole append logic I'm wondering when and where it was used before. I just want to ensure we don't remove something that is still used. Also we should add some documentation about shading packages including an example.

@joan38 We have failing test cases which look related to the removed append logic. E.g.

```
X mill.scalalib.HelloWorldTests.assembly.assemblyRules.appendPatternMultiModule 9298ms 

  utest.AssertionError: #2: referenceContent.contains("Model Reference Config File"),

  referenceContent: String = ##############################

  # Core Reference Config File #

  ##############################

  bar.baz=hello

    mill.scalalib.HelloWorldTests$.$anonfun$tests$168(HelloWorldTests.scala:718)

    mill.scalalib.HelloWorldTests$.$anonfun$tests$168$adapted(HelloWorldTests.scala:709)

    mill.scalalib.HelloWorldTests$.workspaceTest(HelloWorldTests.scala:314)

    mill.scalalib.HelloWorldTests$.checkAppendMulti$1(HelloWorldTests.scala:320)

    mill.scalalib.HelloWorldTests$.$anonfun$tests$182(HelloWorldTests.scala:741)
```

joan38:
Ah I didn't see it. Now it's fixed.

lefou:
@joan38 I find it really hard reviewing your PRs, as you constantly rebase exising commit(s). Adding new commits would better allow tracking what changed and would speed up review quite a bit. In most cases we squash when merging anyways, so adding commits is not an issue.

joan38:
Thanks for the feedback @lefou, I'll push again separating and avoid amend commits in the futur.

main/src/modules/Jvm.scala
```diff
@@ -267,19 +262,12 @@ object Jvm {
   def createAssembly(inputPaths: Agg[os.Path],
                      manifest: JarManifest = JarManifest.Default,
                      prependShellScript: String = "",
-                     base: Option[os.Path] = None,
```
joan38, line 270: I had to remove this because we need to process both the project's classes and dependencies with the same shader instance.

main/src/modules/Jvm.scala
```diff
@@ -293,30 +281,18 @@ object Jvm {
     manifest.build.write(manifestOut)
     manifestOut.close()
 
-    Assembly.groupAssemblyEntries(inputPaths, assemblyRules).view
```
joan38, line 296: No need for `.view` because we execute only a foreach right after

main/src/modules/Jvm.scala
```diff
@@ -293,30 +281,18 @@ object Jvm {
...
-        case (mapping, AppendEntry(entries, separator)) =>
-          val path = zipFs.getPath(mapping).toAbsolutePath
-          val separated =
-            if (entries.isEmpty) Nil
-            else
-              entries.head +: entries.tail.flatMap { e =>
-                List(JarFileEntry(e.mapping, () => new ByteArrayInputStream(separator.getBytes)), e)
-              }
-          val concatenated = new SequenceInputStream(
-            Collections.enumeration(separated.map(_.inputStream).asJava))
```
joan38, line 307: This is handled in `groupAssemblyEntries` now

main/src/modules/Jvm.scala
```diff
@@ -335,16 +311,14 @@ object Jvm {
     PathRef(output)
   }
 
-  private def writeEntry(p: java.nio.file.Path, is: InputStream, append: Boolean): Unit = {
```
joan38, line 338: We don't need to append now since we are doing the `InputStreams` concat and we are not merging with the upstream jar.

joan38:
I pushed again and added comments.
But again if you want more details I'm happy to jump on a call to discuss if it's easier.

lefou (commented):
Looks good to me. Thanks for your extra comments.

We now no longer use the `upstreamAssembly` target, but keep it. Give me some time to think about it. Maybe we can deprecate and remove that target in the future. If I understood it correctly, it's purpose was optimization, which we've lost now. Otherwise, I'm going to merge.

joan38:
Alright after writing a reply to you comment @lefou I realized the importance of `upstreamAssembly`. The goal is to save on rebuilding the biggest part of the jar (which is the dependencies) over and over every time we change some code in our project. I benchmarked it quickly and it is actually significant on my project.

I actually managed to revert back this benefit keeping the shading.

joan38:
Build is green now @lefou if you want to take a look.

lefou:
> I actually managed to revert back this benefit keeping the shading. Have a look.

Not sure how to read that statement.

Just to clarify: In its current state this PR makes building assemblies significantly slower, right? For the sake of class relocation support.

joan38:
It used to be slower in the case of running assembly then modifying your code and rerunning assembly. Because we would reprocess all the dependencies again.

Now it's back to what it was before, and we have shading. So no compromise done.

lefou:
Ok, great! Thank you very much!

joan38:
Thanks @lefou, @alexarchambault and @eed3si9n !

PR #953, opened 2020-08-31
Omit the suffix in artifactName for cross modules

joan38:
Users of mill (including me) struggle with publishing artifacts as soon as they start cross compiling between multiple Scala versions.
By default it will publish `Artifact(org,myproject-2.12.10_2.12,0.0.1-SNAPSHOT)` where we expect something like `Artifact(org,myproject_2.12,0.0.1-SNAPSHOT)`.
Here is an example: https://gitter.im/lihaoyi/mill?at=5f404a975580fa092b15a821

The usual fix is to override the `artifactName` as follows:
```
def artifactName = "myproject"
```

I personally get caught every time I start to cross compile a project and believe that doesn't look reasonable to ask the user to course correct the artifact name.

This proposal removes the added segment part in `artifactName` for `CrossModuleBase`.

cc @davesmith00000

### DIFF
--- a/scalalib/src/MiscModule.scala
+++ b/scalalib/src/MiscModule.scala
@@ -16,6 +16,7 @@ trait CrossModuleBase extends ScalaModule {
   def scalaVersion = T{ crossScalaVersion }
 
   override def millSourcePath = super.millSourcePath / ammonite.ops.up
+  override def artifactName: T[String] = millModuleSegments.parts.init.mkString("-")
   implicit def crossSbtModuleResolver: Resolver[CrossModuleBase] = new Resolver[CrossModuleBase]{
     def resolve[V <: CrossModuleBase](c: Cross[V]): V = {
       crossScalaVersion.split('.')
--- a/scalalib/test/src/HelloWorldTests.scala
+++ b/scalalib/test/src/HelloWorldTests.scala
@@ -492,6 +492,13 @@ object HelloWorldTests extends TestSuite {
       }
     }
 
+    'artifactNameCross - {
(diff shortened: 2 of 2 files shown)


davesmith00000:
I have no idea if that will work, but I'm excited! Thanks for looking into this @joan38. :-)

joan38:
@lefou @lihaoyi Any thoughts?

lefou:
I'd like to see a test case covering (and documenting) the expected outcome.

lihaoyi:
I'm a bit reluctant to do this, due to silently changing the existing behavior. OTOH if the existing behavior is confusing enough, then perhaps it's worth breaking. @joan38 perhaps you could flesh out the description a bit so someone looking at this PR can understand the issue without any external context?

joan38:
@lihaoyi @lefou both done

lihaoyi:
I think this looks good to me

lefou (approved):
Looks good to me.

I just wanted to mention, that at first I struggled with the idea, because in cross-compiling scenarios outside of Scala modules the current behaviour (appending the cross-parameter to the artifact name) might be more appropriate. Later I realized, that `CrossModuleBase` is only used for `CrossScalaModule` and `CrossSbtModule`, so yes, this PR makes lots of sense, as it should not affect other modules created with `Cross[T]`.

lefou:
Pull request: https://github.com/lihaoyi/mill/pull/953

lefou:
Thanks, @joan38 !

joan38:
Make sense, thanks!

PR #958, opened 2020-09-04
Add toMap and toJson methods to BuildInfo

sbly:
JSON is done via `ujson-circe` as this is what's already being used in the project itself

### DIFF
--- a/contrib/buildinfo/src/BuildInfo.scala
+++ b/contrib/buildinfo/src/BuildInfo.scala
@@ -1,11 +1,14 @@
 package mill.contrib.buildinfo
 
 import mill.T
-import mill.api.Logger
-import mill.api.PathRef
-import mill.scalalib.ScalaModule
+import mill.Agg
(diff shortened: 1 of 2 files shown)


lefou (commented):
Mill uses upickle for JSON serialization. The circe version is probably only present for some contrib module.

lihaoyi:
I don't think we should hardcode a ujson dependency in the BuildInfo modules. If anyone wants to convert it to JSON, calling `upickle.default.write(BuildInfo.toMap)` is trivial, and adding a ujson dependency bloats jars and leads to classpath conflicts

lefou:
+1 for only adding the `toMap` method. But leaving out the `toJson`.
If you really need a out-of-the-box `toJson`, you could also pre-render it at generation time, as all values are constant anyway. This would add no extra dependencies.

sbly:
@lihaoyi @lefou Thanks for the reviews! I removed the `toJson` method as recommended.

lefou:
Thank you!

PR #961, opened 2020-09-09
Do not remove out.js, let linker overwrite existing file

lavrov:
Usually during development I run continuous compilation
```
$ ./mill -w webapp.fastOpt
```
and webpack dev server which watches js output for changes and refreshes the page if any.

Webpack dev server doesn't like when the file it is watching disappears. In such cases dev servers fails with an error.

```
ERROR in ../out/client/fastOpt/dest/out.js
Module build failed (from ./node_modules/react-hot-loader/webpack.js):
Error: ENOENT: no such file or directory, open '/workspace/webapp/out/client/fastOpt/dest/out.js'
```

### DIFF
--- a/scalajslib/src/ScalaJSModule.scala
+++ b/scalajslib/src/ScalaJSModule.scala
@@ -125,7 +125,6 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
     val outputPath = ctx.dest / "out.js"
 
     os.makeDir.all(ctx.dest)
-    os.remove.all(outputPath)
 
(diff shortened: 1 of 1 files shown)


lefou:
Does that really solve the issue? I doubt it. The targets which calls the link method (`fastOpt`, `fullOpt`) are not persistent and will clean the whole dest directory even before calling the link method.

lavrov:
Ok then, if that's the case this fix is worthless.

lefou:
You could work around by creating a new perstistent target, which replaces a copy of `out.js whenever a new version was build.

PR #967, opened 2020-09-30
Get the full class loader for BSP

joan38:
I don't understand why we were using `SpecialClassLoader` but it seems to miss some jars, like `os.Path` for example is red in my build but not with this change.


joan38:
Closing in favor of https://github.com/lihaoyi/mill/pull/969

PR #972, opened 2020-10-05
Proguard contrib module

sbly:
Create Proguard contrib module

This module will run `proguard` on the normal `assembly` output jar to shrink it. By default all steps - shrink, optimize, obfuscate, and preverify - are run though this can be configured. The default entrypoint is `finalMainClass`. Any additional options can be specified freely with `additionalOptions` (proguard has many).

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -75,6 +75,7 @@ object Deps {
   val zinc = ivy"org.scala-sbt::zinc:1.4.0-M1"
   val bsp = ivy"ch.epfl.scala:bsp4j:2.0.0-M4"
   val jarjarabrams = ivy"com.eed3si9n.jarjarabrams::jarjar-abrams-core:0.3.0"
+  val proguard = ivy"com.guardsquare:proguard-core:7.0.0"
 }
 
 trait MillPublishModule extends PublishModule{
@@ -453,6 +454,11 @@ object contrib extends MillModule {
       scalalib.backgroundwrapper.testArgs()
     }
    }
+
+  object proguard extends MillModule {
+    override def moduleDeps = Seq(scalalib)
+    def ivyDeps = Agg(Deps.proguard, Deps.osLib)
+  }
 
   object tut extends MillModule {
     override def compileModuleDeps = Seq(scalalib)
--- a/build.sc
+++ b/build.sc
@@ -75,7 +75,6 @@ object Deps {
   val zinc = ivy"org.scala-sbt::zinc:1.4.0-M1"
   val bsp = ivy"ch.epfl.scala:bsp4j:2.0.0-M4"
   val jarjarabrams = ivy"com.eed3si9n.jarjarabrams::jarjar-abrams-core:0.3.0"
-  val proguard = ivy"com.guardsquare:proguard-core:7.0.0"
 }
 
 trait MillPublishModule extends PublishModule{
@@ -456,8 +455,7 @@ object contrib extends MillModule {
    }
 
   object proguard extends MillModule {
-    override def moduleDeps = Seq(scalalib)
-    def ivyDeps = Agg(Deps.proguard, Deps.osLib)
+    override def compileModuleDeps = Seq(scalalib)
   }
 
   object tut extends MillModule {
--- a/contrib/proguard/src/Proguard.scala
+++ b/contrib/proguard/src/Proguard.scala
@@ -35,9 +35,10 @@ trait Proguard extends ScalaModule {
 
   def proguard: T[PathRef] = T {
     val outJar = PathRef(T.dest / "out.jar")
+    val java = javaHome().path / "bin" / "java"
 
     val cmd = os.proc(
-      "java",
+      java,
       "-cp",
       proguardClasspath().map(_.path).mkString(":"),
       "proguard.ProGuard",
@@ -45,7 +46,7 @@ trait Proguard extends ScalaModule {
       "-injars",
       inJar().path,
       "-outjars",
-      outJar().path,
+      outJar.path,
       "-libraryjars",
       libraryJars().map(_.path).mkString(":"),
       entryPoint(),
--- a/contrib/proguard/src/Proguard.scala
+++ b/contrib/proguard/src/Proguard.scala
@@ -12,6 +12,9 @@ import os.Path
 import os.PathChunk
 
 trait Proguard extends ScalaModule {
+
+  // This is actually the only available version on Maven currently.
+  // But we should keep it as a configuration in case a new version is released.
   def proguardVersion: T[String] = T { "7.0.0" }
 
   def shrink: T[Boolean] = T { true }
@@ -19,28 +22,34 @@ trait Proguard extends ScalaModule {
   def obfuscate: T[Boolean] = T { true }
   def preverify: T[Boolean] = T { true }
 
+  def javaHome: T[PathRef] = T {
+    PathRef(Path(System.getProperty("java.home")))
+  }
+
   def inJar: T[PathRef] = T { assembly() }
   def outJar: T[PathRef] = T { PathRef(T.dest / "out.jar") }
   def libraryJars: T[Seq[PathRef]] = T {
-    upstreamAssemblyClasspath().toSeq
+    val javaJars = os.list(javaHome().path / "lib", sort = false).filter(_.ext == "jar")
+    javaJars.toSeq.map(PathRef(_))
   }
 
   def proguard: T[PathRef] = T {
     val cmd = os.proc(
-        "java",
-        "-cp",
-        proguardClasspath().map(_.path).mkString(":"),
-        "proguard.ProGuard",
-        "-injars",
-        inJar().path,
-        "-outjars",
-        outJar().path,
-        "-libraryjars",
-        libraryJars().map(_.path),
-        steps(),
-        additionalOptions()
-      )
-    System.out.println("Running command: " + cmd.command.flatMap(_.value).mkString(" "))
+      "java",
+      "-cp",
+      proguardClasspath().map(_.path).mkString(":"),
+      "proguard.ProGuard",
+      steps(),
+      "-injars",
+      inJar().path,
+      "-outjars",
+      outJar().path,
+      "-libraryjars",
+      libraryJars().map(_.path).mkString(":"),
+      entryPoint(),
+      additionalOptions()
+    )
+    System.out.println(cmd.command.flatMap(_.value).mkString(" "))
     cmd.call(stdout = T.dest / "stdout.txt", stderr = T.dest / "stderr.txt")
 
     // the call above already throws an exception on a non-zero exit code,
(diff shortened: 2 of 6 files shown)


build.sc
```diff
@@ -454,6 +455,11 @@ object contrib extends MillModule {
     }
    }
 
+  object proguard extends MillModule {
+    override def moduleDeps = Seq(scalalib)
```
lefou, line 459: This should be a `compileModuleDeps`.

contrib/proguard/src/Proguard.scala
```diff
@@ -0,0 +1,82 @@
...
+  def optimize: T[Boolean] = T { true }
+  def obfuscate: T[Boolean] = T { true }
+  def preverify: T[Boolean] = T { true }
+
+  def javaHome: T[PathRef] = T {
+    PathRef(Path(System.getProperty("java.home")))
+  }
+
+  def inJar: T[PathRef] = T { assembly() }
+  def outJar: T[PathRef] = T { PathRef(T.dest / "out.jar") }
```
lefou, line 30: Why is this needed. Isn't it always the output of the `proguard` target?

contrib/proguard/src/Proguard.scala
```diff
@@ -0,0 +1,82 @@
...
+      libraryJars().map(_.path).mkString(":"),
+      entryPoint(),
+      additionalOptions()
+    )
+    System.out.println(cmd.command.flatMap(_.value).mkString(" "))
+    cmd.call(stdout = T.dest / "stdout.txt", stderr = T.dest / "stderr.txt")
+
+    // the call above already throws an exception on a non-zero exit code,
+    // so if we reached this point we've succeeded!
+    outJar()
```
lefou, line 57: I do not recommend writing to a path which is in the `dest` folder of another target. This has a high potential to breaking mills caching.

lefou (commented):
Thank you for your contribution. The module looks ok to me. I added some comments I'd like you to address.

Also to accept this new contrib module, we need some documentation under `docs/pages/9 - Contrib Modules.md` (modules in alphabetical order). Having some Scaladoc for all public targets would greatly help later users of that module too.

build.sc
```diff
@@ -75,6 +75,7 @@ object Deps {
   val zinc = ivy"org.scala-sbt::zinc:1.4.0-M1"
   val bsp = ivy"ch.epfl.scala:bsp4j:2.0.0-M4"
   val jarjarabrams = ivy"com.eed3si9n.jarjarabrams::jarjar-abrams-core:0.3.0"
+  val proguard = ivy"com.guardsquare:proguard-core:7.0.0"
```
lefou, line 78: Why is this dependency needed. Don't we resolve it in the `proguardClasspath` target at runtime?

build.sc
```diff
@@ -454,6 +455,11 @@ object contrib extends MillModule {
     }
    }
 
+  object proguard extends MillModule {
+    override def moduleDeps = Seq(scalalib)
+    def ivyDeps = Agg(Deps.proguard, Deps.osLib)
```
lefou, line 460: Why we depend on `Deps.proguard` at build time?

lefou (changes requested):
For consistence, if you use an library from an explicit Java version (via JAVA_HOME) you should also call the Java executable from this specific version.

contrib/proguard/src/Proguard.scala
sbly: Yeah you're right

contrib/proguard/src/Proguard.scala
sbly: OK, will fix.

build.sc
sbly: Whoops, this was from when I was starting and didn't realize what I was doing. I forgot to remove it

build.sc
sbly: Will remove

sbly:
@lefou Thanks for the review! I've addressed most of your feedback. The one thing I'm a little confused about is this:

> if you use an library from an explicit Java version (via JAVA_HOME) you should also call the Java executable from this specific version.

Does Mill use `JAVA_HOME` or does it get the standard library from elsewhere? I'm having trouble figuring out how the code is compiled in the first place

lefou:
> @lefou Thanks for the review! I've addressed most of your feedback. The one thing I'm a little confused about is this:
> 
> > if you use an library from an explicit Java version (via JAVA_HOME) you should also call the Java executable from this specific version.
> 
> Does Mill use `JAVA_HOME` or does it get the standard library from elsewhere? I'm having trouble figuring out how the code is compiled in the first place

Mill uses `JAVA_HOME` but I was referring to the call out to the `java` executable in the `proguard` target. It's not necessarily  the same as `$JAVA_HOME/bin/java` and therefore problematic.

contrib/proguard/src/Proguard.scala
```diff
@@ -0,0 +1,84 @@
...
+  def libraryJars: T[Seq[PathRef]] = T {
+    val javaJars = os.list(javaHome().path / "lib", sort = false).filter(_.ext == "jar")
+    javaJars.toSeq.map(PathRef(_))
+  }
+
+  def proguard: T[PathRef] = T {
+    val outJar = PathRef(T.dest / "out.jar")
+
+    val cmd = os.proc(
+      "java",
```
lefou, line 40: Better use the `java` executable found under `javaHome`, to make it consistent with the later used classpath.

sbly:
Thanks for the help. I switched to using the java binary under JAVA_HOME, and also added documentation.

lefou (commented):
Looks good to me. Do you plan to also add some ScalaDoc and/or tests anytime soon? If not I can merge now, but those additions would help greatly in comsumption and maintenance of this contrib module.

lefou (approved):

sbly:
@lefou I can add some ScalaDoc. I was testing using just the `scratch/`folder. I wasn't sure how to write automated tests for this but I can look at some examples and try to do it. We can hold off on merging for now

contrib/proguard/src/Proguard.scala
```diff
@@ -11,28 +11,60 @@ import os.proc
...
   def preverify: T[Boolean] = T { true }
 
+  /** The path to JAVA_HOME.
+   *
+   * This is used for both the `java` command binary,
+   * as well as the standard library jars.
+   * Defaults to the `java.home` system property. */
   def javaHome: T[PathRef] = T {
     PathRef(Path(System.getProperty("java.home")))
   }
```
lefou: We need to mark this target as input, otherwise mills caching would hide potential changes of the system property.

```suggestion
  def javaHome: T[PathRef] = T.input {
    PathRef(Path(System.getProperty("java.home")))
  }
```

contrib/proguard/src/Proguard.scala
```diff
@@ -11,28 +11,60 @@ import os.proc
...
     javaJars.toSeq.map(PathRef(_))
   }
 
+  /** Run the proguard task.
+   *
+   *  The full command will be printed when run.
+   *  The stdout and stderr of the command are written to the `dest/` folder.
+   *  The output jar is written to `dest/our.jar`. */
   def proguard: T[PathRef] = T {
     val outJar = PathRef(T.dest / "out.jar")
```
lefou, line 69: We should defer the creation of the `PathRef` until the end of the target, as it will eagerly calculate a checksum to detect changes.

```suggestion
    val outJar = T.dest / "out.jar"
```

contrib/proguard/src/Proguard.scala
```diff
@@ -60,26 +92,38 @@ trait Proguard extends ScalaModule {
     outJar
```
lefou, line 92: Following the change above...

```suggestion
    PathRef(outJar)
```

lefou (commented):
Scaladoc looks good. I found two other subtle improvements that are specific to mills caching.

contrib/proguard/src/Proguard.scala
sbly: OK, I'll make the change. But I'm wondering now if this should actually be `Sources`? What's the difference?

lefou: `Sources` and also `Source` are specializations of Input. Also Source(s) support the watch mode (`mill -w`), whereas Input is not able to trigger a re-build on it's own.

sbly:
@lefou Hi, I finally had some time to work on this again. I added some basic tests. I am ready to merge this now if that's OK with you

lefou:
Thank you @sbly !

lefou:
I just notices, that our test setup wasn't running the proguard tests. Running them locally gives me the following error:

```
[263/263] contrib.proguard.test.testCached 
-------------------------------- Running Tests --------------------------------
mill.contrib.proguard.ProguardTests: Using experimental parallel evaluator with 4 threads
mill.contrib.proguard.ProguardTests: [#1] [2/2] proguardClasspath 
+ mill.contrib.proguard.ProguardTests.proguard.should download proguard jars 1504ms  
mill.contrib.proguard.ProguardTests: Using experimental parallel evaluator with 4 threads
mill.contrib.proguard.ProguardTests: [#0] [40/52] libraryJars 
mill.contrib.proguard.ProguardTests: [#3] [9/52] mill.scalalib.ZincWorkerModule.classpath failed
mill.contrib.proguard.ProguardTests: [#2] [41/52] mill.scalalib.ZincWorkerModule.worker failed
mill.contrib.proguard.ProguardTests: [#2] [42/52] compile failed
mill.contrib.proguard.ProguardTests: [#2] [43/52] finalMainClassOpt failed
mill.contrib.proguard.ProguardTests: [#1] [44/52] localClasspath failed
mill.contrib.proguard.ProguardTests: [#3] [45/52] prependShellScript failed
mill.contrib.proguard.ProguardTests: [#0] [47/52] manifest failed
mill.contrib.proguard.ProguardTests: [#2] [46/52] finalMainClass failed
mill.contrib.proguard.ProguardTests: [#1] [48/52] upstreamAssembly failed
mill.contrib.proguard.ProguardTests: [#2] [49/52] entryPoint failed
mill.contrib.proguard.ProguardTests: [#1] [50/52] assembly failed
mill.contrib.proguard.ProguardTests: [#1] [51/52] inJar failed
mill.contrib.proguard.ProguardTests: [#1] [52/52] proguard failed
X mill.contrib.proguard.ProguardTests.proguard.create proguarded jar 1741ms 
  scala.MatchError: Left(Failure(
    com.lihaoyi:mill-scalalib-worker_2.13:0.9.3-20-ea5b88 
        not found: /home/lefou/.ivy2/local/com.lihaoyi/mill-scalalib-worker_2.13/0.9.3-20-ea5b88/ivys/ivy.xml
        not found: https://repo1.maven.org/maven2/com/lihaoyi/mill-scalalib-worker_2.13/0.9.3-20-ea5b88/mill-scalalib-worker_2.13-0.9.3-20-ea5b88.pom
  ,None)) (of class scala.util.Left)
    mill.contrib.proguard.ProguardTests$.$anonfun$tests$7(ProguardTests.scala:44)
    mill.contrib.proguard.ProguardTests$.$anonfun$tests$7$adapted(ProguardTests.scala:43)
    mill.contrib.proguard.ProguardTests$.workspaceTest(ProguardTests.scala:33)
    mill.contrib.proguard.ProguardTests$.$anonfun$tests$6(ProguardTests.scala:36)
1 targets failed
contrib.proguard.test.testCached mill.contrib.proguard.ProguardTests mill.contrib.proguard.ProguardTests.proguard.create proguarded jar
```

sbly:
Hi @lefou. Sorry about being super late getting to this, I don't regularly check Github when I'm not actively working on a PR.

So locally, I would always run `mill -i all __.publishLocal` before testing. If I didn't, I got the same error as you above. I thought this was expected and was just the way mill's tests were set up, so I didn't mention it. I just tried again and it still works.

If this is not the way tests are expected to work, I am happy to work out the error with you. I based my tests off of the existing tests, so there was some copy-pasting that may be the culprit.

lefou:
@sbly Thanks for making your assumptions clear. All mill tests should run as-is. I created a PR to fix the setup.

PR #975, opened 2020-10-15
Add example

ShalokShalom:
I was annoyed that the Readme had no example, so I posted that one hidden in the documentation here, since this is what gives this projects its face.
IMHO


lefou (approved):
Looks good to me. Thanks!

PR #981, opened 2020-10-26
Redirect zinc log output to stderr

jodersky:
The compiler output is a diagnostic message and should hence be printed to stderr.

This also allows users to redirect output when running applications directly from mill, without including any diagnostic information. E.g.

```
mill foo.run > out.txt
```

### DIFF
--- a/scalalib/worker/src/ZincWorkerImpl.scala
+++ b/scalalib/worker/src/ZincWorkerImpl.scala
@@ -384,7 +384,7 @@ class ZincWorkerImpl(compilerBridge: Either[
 
     val consoleAppender = ConsoleAppender(
       "ZincLogAppender",
-      ConsoleOut.printStreamOut(ctx.log.outputStream),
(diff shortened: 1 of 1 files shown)


jodersky:
It appears that Travis keeps losing track of commits. Let's try again...

jodersky:
It appears that `mill publishVersion` is failing. I don't know how that's related to my proposed changes.

lefou:
> It appears that `mill publishVersion` is failing. I don't know how that's related to my proposed changes.

It isn't. Somehow the way we set up Travis-CI and the way it runs pull requests is not compatible, if the pull request is not based on the latest commit of the main/master branch. What typically helps is rebaing, but I'm also working on a setup for Github Actions.

jodersky:
My first commit was indeed not rebased, but the latest try is just one commit ahead of master. Could it be that travis has some bad cache? Let me try rebasing (and amending, just to be sure) once again.

lefou (approved):
Makes sense, thanks.

PR #982, opened 2020-10-27
Allow test classes with constructor parameters

RobinJDCox:
This PR removes a filter from the test discovery that ignores classes that have parameters in their constructor. I think this is problematic as it is perfectly valid for tests to have constructor parameters depending on the test framework, and so this should be a decision made by the testing framework rather than the build tool.

Here's an example where I encountered this issue recently: https://github.com/disneystreaming/weaver-test/issues/113. After running these tests again with the changes made in this PR the example tests shown in that issue now pass.

I've attempted to verify that these changes are compatible with other test frameworks. With some fairly basic examples uTest and Scalatest both seem to continue to work as expected. I'd be happy to verify the changes against other test frameworks if you have any further suggestions.

Thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ output/
 .DS_STORE
 .idea_modules
 .idea
+.vscode/
 out/
 /.bloop/
 /.metals/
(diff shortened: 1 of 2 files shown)


lefou (approved):
Looks good to me.

lefou:
Thank you!

PR #991, opened 2020-11-02
travis: use some newer JDKs

lefou:


lefou:
We switched to Github Actions

PR #995, opened 2020-11-03
Bloop: Also use compileModuleDeps to calc transitive classpath

lefou:
Attempt to fix https://github.com/lihaoyi/mill/issues/990


ckipp01 (approved):
Thanks a bunch for taking a look at this @lefou, it fixes the issue!

PR #996, opened 2020-11-06
Adds ScoverageReport for multi-module projects

RobinJDCox:
This is a module we've been using for a couple of months now that's been working well, originally suggested by @lefou on Gitter [here](https://gitter.im/lihaoyi/mill?at=5f5f7d8df969413294e55466). It allows scoverage to be run across multiple modules and then aggregated into a single report. This is particularly useful if you need a single report for SonarQube or other analysis tools and so contributing it back here.

Thanks @lefou for the help!

### DIFF
--- a/contrib/scoverage/src/ScoverageReport.scala
+++ b/contrib/scoverage/src/ScoverageReport.scala
@@ -1,0 +1,106 @@
+package mill.contrib.scoverage
+
+import mill.contrib.scoverage.api.ScoverageReportWorkerApi.ReportType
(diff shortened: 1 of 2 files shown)


lefou:
@RobinJDCox Would you be so kind and add me as co-author to the first git commit? (https://github.blog/2018-01-29-commit-together-with-co-authors/#how-it-works)

RobinJDCox:
Of course! That should be done now

lefou (approved):
Looks good to me.

Issue #989, opened 2020-10-29
[BSP] - Incorrect `Range` in `Diagnostic`.


ckipp01:
Given the following code snippet:
```scala
package example

object Hello extends Greeting with App {
  println(greeting)
}

trait Greeting {
  lazy val greeting: String = "hello"
  val x: Int = "hi" 
               ^^^^ // I'd expect the diagnostic to show this
}
```
According to the [BSP spec](https://build-server-protocol.github.io/docs/specification.html#publish-diagnostics):

> where Diagnostic is defined as it is in LSP.

And in the [LSP spec](https://microsoft.github.io/language-server-protocol/specification#range), the `Range` is zero based. However, it seems that the `Range` coming back from mill is not and the character also seems to be the character on the page rather than the line. For example, the correct diagnostic for the above should look like:

```
[Trace - 05:57:32 PM] Received notification 'build/publishDiagnostics'
Params: {
  "textDocument": {
    "uri": "file:///Users/ckipp/Documents/scala-workspace/mill-tester/MillTester/src/example/Hello.scala"
  },
  "buildTarget": {
    "uri": "file:/Users/ckipp/Documents/scala-workspace/mill-tester/MillTester/?id\u003dMillTester"
  },
  "diagnostics": [
    {
      "range": {
        "start": {
          "line": 8,
          "character": 15
        },
        "end": {
          "line": 8,
          "character": 19
        }
      },
      "severity": 1,
      "code": "  val x: Int \u003d \"hi\"",
      "source": "bloop",
      "message": "type mismatch;\n found   : String(\"hi\")\n required: Int"
    }
  ],
  "reset": true
}
```
However, from Mill it looks like this:
```
[Trace - 06:08:28 PM] Received notification 'build/publishDiagnostics'
Params: {
  "textDocument": {
    "uri": "file:/Users/ckipp/Documents/scala-workspace/mill-tester/MillTester/src/example/Hello.scala"
  },
  "buildTarget": {
    "uri": "file:///Users/ckipp/Documents/scala-workspace/mill-tester/MillTester/"
  },
  "diagnostics": [
    {
      "range": {
        "start": {
          "line": 9,
          "character": 151
        },
        "end": {
          "line": 9,
          "character": 151
        }
      },
      "severity": 1,
      "code": "  val x: Int \u003d \"hi\"",
      "source": "compiler from mill",
      "message": "type mismatch;\n found   : String(\"hi\")\n required: Int"
    }
  ],
  "reset": true
}
[...]

PR #999, opened 2020-11-10
#989 Correct error line and column number

joan38:


bsp/src/mill/bsp/BspLoggedReporter.scala
```diff
@@ -91,24 +91,25 @@ class BspLoggedReporter(client: bsp.BuildClient,
...
   private[this] def getSingleDiagnostic(problem: Problem): Diagnostic = {
     val pos = problem.position
-    val i: Integer = pos.startLine.orElse(pos.line).getOrElse[Int](0)
-    println(i)
+    val line = pos.line.map(_ - 1) // Zinc's range starts at 1 whereas BSP 0
     val start = new bsp.Position(
-      pos.startLine.orElse(pos.line).getOrElse[Int](0),
-      pos.startOffset.orElse(pos.offset).getOrElse[Int](0)
+      pos.startLine.orElse(line).getOrElse[Int](0),
+      pos.startOffset.orElse(pos.pointer).getOrElse[Int](0)
```
joan38, line 97: Not sure it might be `spacePointer` here. I need to do some testing.

ckipp01:
Thanks for taking a look at this @joan38! Just tested it out, and much better!

Screenshot 2020-11-24 at 18 02 37

One other thing that I do notice though is that the diagnostic from Mill is as follows:
```
"diagnostics": [
    {
      "range": {
        "start": {
          "line": 8,
          "character": 15
        },
        "end": {
          "line": 8,
          "character": 15
        }
      },
      "severity": 1,
      "source": "compiler from mill",
      "message": "type mismatch;\n found   : String(\"hi\")\n required: Int"
    }
  ],
```
Whereas both with sbt and with Bloop it's:

```
"diagnostics": [
    {
      "range": {
        "start": {
          "line": 8,
          "character": 15
        },
        "end": {
          "line": 8,
          "character": 19
        }
      },
      "severity": 1,
      "source": "sbt",
      "message": "type mismatch;\n found   : String(\"hi\")\n required: Int"
    }
  ]
```

It still seems to work fine (I think there is some magic on the LSP client side) to maybe highlight the full symbol, but it'd be better to correctly capture the entire range and not return 15 for both start character and end character.

If it helps at all, here is the code sbt uses to get the range from the `Problem`.

https://github.com/sbt/sbt/blob/13b09bcd8fa0fde86aa4f4f8d7089fcb18e345c6/main/src/main/scala/sbt/internal/server/BuildServerReporter.scala#L133-L159

lefou (approved):
Looks good to me.

lefou:
Merging this as it makes BSP error reporting significantly better. I'll open another issue for the open issue with the range end position.

lefou:
Thanks you @joan38 for the fix and @ckipp01 for the feedback!

PR #1002, opened 2020-11-13
ScalaPB additional args

DanielBlanco:
@lefou can you please review, and let me know if I'm missing something?

The plan with this change is to allow additional args for scalaPBC, this means I can just do this in my code for zio-grpc to work:

```
object example extends ScalaPBModule {
  def scalaVersion = "2.12.6"
  def scalaPBVersion = "0.7.4"

   override def scalaPBAdditionalArgs = Seq(
        s"--zio_out=${T.dest.toIO.getCanonicalPath}"
   )
}
```

### DIFF
--- a/contrib/scalapblib/src/ScalaPBModule.scala
+++ b/contrib/scalapblib/src/ScalaPBModule.scala
@@ -3,7 +3,13 @@ package contrib.scalapblib
 
 import java.net.URI
 import java.nio.file.attribute.BasicFileAttributes
-import java.nio.file.{FileSystems, Files, Path, SimpleFileVisitor, StandardCopyOption}
+import java.nio.file.{
+  FileSystems,
+  Files,
+  Path,
+  SimpleFileVisitor,
+  StandardCopyOption
+}
 
 import coursier.MavenRepository
 import coursier.core.Version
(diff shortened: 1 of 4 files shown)


contrib/scalapblib/src/ScalaPBWorker.scala
```diff
@@ -38,28 +30,51 @@ class ScalaPBWorker {
...
-        os.walk(inputDir).filter(_.last.matches(".*.proto"))
-          .foreach { proto =>
-            compiler.compileScalaPB(proto.toIO, scalaPBOptions, dest.toIO, scalaPBIncludePath)
-          }
+  /** Builds the compilation arguments for scalaPBC:
+   *
+   *   - 1st seq to encapsulate the sources.
+   *   - 2nd seq for proto files.
+   *   - 3rd seq for the scalaPBC args.
+   */
```
lefou, line 38: Instead of "1st", "2nd", ... please used proper Scaladoc `@param` attributes to make clear which parameter has which purpose.

lefou (commented):
Your pull request looks good to me. Personally, I'd name the new target "additionalArgs" instead of "customArgs" as that better transports the fact that other args logic will still be preserved and it also matches what we have in other plugins, but that's only a suggestion and up to you.

But I'd like to have some more ScalaDoc, at least on the newly added and changed targets. These will help any user of the plugin who edits their buildfile in an IDE.

contrib/scalapblib/src/ScalaPBModule.scala
```diff
@@ -71,17 +85,22 @@ trait ScalaPBModule extends ScalaModule {
 
   def scalaPBIncludePath: T[Seq[PathRef]] = T.sources { Seq.empty[PathRef] }
 
+  def scalaPBAdditionalArgs: T[Seq[String]] = T { Seq.empty[String] }
```
lefou, line 88: Please add some Scaladoc.

contrib/scalapblib/src/ScalaPBModule.scala
```diff
@@ -110,14 +129,24 @@ trait ScalaPBModule extends ScalaModule {
     PathRef(dest)
   }
 
+  def compilationArgsScalaPB: T[Seq[Seq[Seq[String]]]] = T {
```
lefou, line 132: Please add some Scaladoc.

lefou (approved):

lefou:
Thank you, @DanielBlanco !

chikei:
Actually this PR might break ScalaPB module as ScalaPB module append `generatedSources` with `compileScalaPB` which was the scalapbc output path, but here `compilationArgsScalaPB` incorrectly write output path using `T.dest` from `compilationArgsScalaPB`.

DanielBlanco:
OK, I'll try the test and come up with a solution if I can. Thx @chikei.

PR #1003, opened 2020-11-15
Fix BSP,  avoid assumptions about Logger implementation

srdo:
Some use cases need to be able to emit exact strings to stdout, e.g. BSP.
For those cases, we need to be able to skip log decoration, e.g. colors and prefixes.

I don't know if this is the most elegant solution, but I think emitting logs without decoration is something third party tasks may also need to be able to do, so I added the capability to Logger. For Mill itself, the BSP server needs to be able to do it.

### DIFF
--- a/bsp/src/mill/bsp/BSP.scala
+++ b/bsp/src/mill/bsp/BSP.scala
@@ -64,7 +64,7 @@ object BSP extends ExternalModule {
     write(
       BspConfigJson(
         name = "mill-bsp",
-        argv = Seq(millPath, "-i", s"${BSP.getClass.getCanonicalName.split("[$]").head}/start"),
+        argv = Seq(millPath, "-i", "--disable-ticker", "--color", "false", s"${BSP.getClass.getCanonicalName.split("[$]").head}/start"),
         millVersion = Util.millProperty("MILL_VERSION").getOrElse(BuildInfo.millVersion),
         bspVersion = bspProtocolVersion,
         languages = languages
@@ -101,12 +101,13 @@ object BSP extends ExternalModule {
         BuildInfo.millVersion)
       val executor = Executors.newCachedThreadPool()
 
-      val stdin = System.in
-      val stdout = System.out
+      val logger = T.ctx.log
+      val in = logger.inStream
+      val out = logger.outputStream
       try {
         val launcher = new Launcher.Builder[BuildClient]()
-          .setOutput(stdout)
-          .setInput(stdin)
+          .setOutput(out)
+          .setInput(in)
           .setLocalService(millServer)
           .setRemoteInterface(classOf[BuildClient])
           .traceMessages(new PrintWriter(
--- a/main/core/src/eval/Evaluator.scala
+++ b/main/core/src/eval/Evaluator.scala
@@ -525,7 +525,7 @@ case class Evaluator(
 
   def resolveLogger(logPath: Option[os.Path], logger: Logger): Logger = logPath match{
     case None => logger
-    case Some(path) => MultiLogger(logger.colored, logger, new FileLogger(logger.colored, path, debugEnabled = true))
+    case Some(path) => MultiLogger(logger.colored, logger, new FileLogger(logger.colored, path, debugEnabled = true), logger.inStream)
   }
 
   /**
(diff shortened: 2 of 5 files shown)


srdo:
Something weird is going on with Travis.
> fatal: 67ced87575ef57e73062ecc500371cc0282504d8 is not a valid 'commit' object

srdo:
@lefou I've updated this with an alternative solution.

I like your suggestion of moving inStream to Command, and I think that's the right solution. For now I've added a package private method to Logger to let us tell if the Logger provides stdin, but I'd like to solve it by moving inStream to Command once I get a chance to look at this more.

Regarding creating a new isolated non-concurrent Command, while the idea seems appealing, the BSP spec states that the client controls the lifecycle of the server. This means that the Command running the BSP server may run indefinitely. Wouldn't that cause issues if e.g. my IDE keeps the BSP server running, and then I want to run some Mill commands from my terminal?

An issue I think needs addressing after this PR is that unlike with colors and the ticker, there is no way to disable the PrefixLogger's prefixes. I think the BSP server is started using a code path that happens to set an empty prefix, but that will probably break in the future. I'm torn on whether the easiest fix here is to add a new CLI parameter (3 parameters is a lot to have to set to get raw output), or to go back to the rawLogger method from the first commit.

lefou (commented):
As discussed on Gitter, I think a better (quick) solution would be to set the `inStream` of `MultiLogger` explicitly instead of assuming and checking for `PrintStream` (which is not a trait that can be easily added with `with`).

lefou (commented):
Some of the changes in this PR are probably worth to be kept though. E.g. the options of the mill invocation as well as the use of the logger provided streams instead of `System.{in,out}`

srdo:
> the use of the logger provided streams instead of System.{in,out}

I think that probably has to be done that way in general in tasks, otherwise it will be difficult to handle parallel task execution, given that System.in/out get set right before task execution.

bsp/src/mill/bsp/BSP.scala
```diff
@@ -64,7 +64,7 @@ object BSP extends ExternalModule {
     write(
       BspConfigJson(
         name = "mill-bsp",
-        argv = Seq(millPath, "-i", s"${BSP.getClass.getCanonicalName.split("[$]").head}/start"),
+        argv = Seq(millPath, "-i", "--disable-ticker", "--color", "false", s"${BSP.getClass.getCanonicalName.split("[$]").head}/start"),
```
lefou, line 67: The `--disable-ticker` is not necessarily needed to make BSP work, but removes valuable information about what is being run by mill-bsp. At least when I try with IntelliJ IDEA, I can see the ticker in the `Build: Sync` window, and it's nice to see what's going on.

srdo: Okay, but I think it works due to luck. We want BSP to be able to log "cleanly" without Mill adding onto the output. I think the place we would want to enable the ticker would be in the MillBspLogger here https://github.com/lihaoyi/mill/blob/7530cf2e0b723590eb9febf1a2192ae888a61bcd/bsp/src/mill/bsp/MillBspLogger.scala#L21. As I understand it, that's the part responsible for wrapping task output so it can be sent to the IDE for display.

lefou: Hm, the `ticker` is used as progress, but has no progress semantic in it's API, so it will always be pure luck to parse it in the `MillBspLogger`. After all, the logger API is available to every target, including user written ones.

Of course, we can keep it disabled, but where am I suspected to see the ticker/progress (which is send to the BSP client in `MillBspLogger`) in the client, esp. IDEA?

lefou (approved):
Looks good to me

bsp/src/mill/bsp/BSP.scala
srdo: I think the way it should work is that BSP.start should be given a logger with all decoration disabled. MillBspLogger wraps that logger and enables the ticker (this is not currently possible since the ticker is managed by the root PrintLogger, but this can be changed). 

This will mean that the messages written to T.ctx.log by BSP will be "raw" which is what we want. When BSP runs a task, it will give that task the MillBspLogger, so the logs generated by those tasks will have the ticker. Those logs are sent to the IDE as TaskProgress messages, which I believe will show up in the "Build output" tab in IntelliJ.

srdo:
Great. I'll just squash it.

lefou:
Thank you!

PR #1004, opened 2020-11-17
Support useECMAScript2015 option in ScalaJSModule

lolgab:
- Support ESModule ModuleKind
- Update Scala.js 0.6 to 0.6.33
- Update Scala.js 1 to 1.3.1
- Add useECMAScript215 to test matrix just for last version
- Update Scala versions it tests
- Update ScalaJsUtils to use node.js instead of nashorn
  The initial idea was to use the NodeJSEnv from Scala.js to run the
  tests. But this created a problem since adding a particular NodeJSEnv
  to the classpath broke the other version of Scala.js linking. For example
  using NodeJSEnv from Scala.js 1.3.1 made fastOpt in Scala.js 0.6 break.
  So at the end the function is just calling "node" using `os.proc`

### DIFF
--- a/scalajslib/worker/0.6/src/ScalaJSWorkerImpl.scala
+++ b/scalajslib/worker/0.6/src/ScalaJSWorkerImpl.scala
@@ -23,7 +23,7 @@ class ScalaJSWorkerImpl extends mill.scalajslib.api.ScalaJSWorkerApi {
            testBridgeInit: Boolean, // ignored in 0.6
            fullOpt: Boolean,
            moduleKind: ModuleKind,
-           useECMAScript2015: Boolean /* ignored in 0.6 */) = {
+           useECMAScript2015: Boolean) = {
 
     val semantics = fullOpt match {
         case true => Semantics.Defaults.optimized
--- a/scalajslib/worker/0.6/src/ScalaJSWorkerImpl.scala
+++ b/scalajslib/worker/0.6/src/ScalaJSWorkerImpl.scala
@@ -15,14 +15,15 @@ import org.scalajs.jsenv._
 import org.scalajs.testadapter.TestAdapter
 
 class ScalaJSWorkerImpl extends mill.scalajslib.api.ScalaJSWorkerApi {
-    
+
   def link(sources: Array[File],
            libraries: Array[File],
            dest: File,
            main: String,
            testBridgeInit: Boolean, // ignored in 0.6
            fullOpt: Boolean,
-           moduleKind: ModuleKind) = {
+           moduleKind: ModuleKind,
+           useECMAScript2015: Boolean /* ignored in 0.6 */) = {
 
     val semantics = fullOpt match {
         case true => Semantics.Defaults.optimized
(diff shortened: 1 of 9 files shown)


lefou:
Great! Thanks for this pull request.

>     * Update ScalaJsUtils to use node.js instead of nashorn
>       The initial idea was to use the NodeJSEnv from Scala.js to run the
>       tests. But this created a problem since adding a particular NodeJSEnv
>       to the classpath broke the other version of Scala.js linking. For example
>       using NodeJSEnv from Scala.js 1.3.1 made fastOpt in Scala.js 0.6 break.
>       So at the end the function is just calling "node" using `os.proc`

I like the idea of running the tests without requiring some external tool (node). I think we are not necessarily forced to use the same classpath for test for different Scala.js versions. But I do not have that much Scala.js tooling experience, so it's probably easier to say that than to implement it. Do you have by any chance the other setup (which failed) around, so that I could play with it?

lolgab:
Consider that Scala-js NodeJSEnv is depending from "node" too:
- https://github.com/scala-js/scala-js-js-envs/blob/f3a156c92e70b94b3af86140e92f6d680743d32d/nodejs-env%2Fsrc%2Fmain%2Fscala%2Forg%2Fscalajs%2Fjsenv%2Fnodejs%2FNodeJSEnv.scala#L59
- https://github.com/scala-js/scala-js-js-envs/blob/f3a156c92e70b94b3af86140e92f6d680743d32d/nodejs-env%2Fsrc%2Fmain%2Fscala%2Forg%2Fscalajs%2Fjsenv%2Fnodejs%2FNodeJSEnv.scala#L221
- https://github.com/scala-js/scala-js-js-envs/blob/f3a156c92e70b94b3af86140e92f6d680743d32d/js-envs%2Fsrc%2Fmain%2Fscala%2Forg%2Fscalajs%2Fjsenv%2FExternalJSRun.scala#L130

So if you use it or not you require `node` either way.
By the way, my `ScalaJsUtils.scala` looked like this:
```scala
package mill.scalajslib

import org.scalajs.jsenv.nodejs.NodeJSEnv
import org.scalajs.jsenv.Input
import org.scalajs.jsenv.test.kit.TestKit
import mill.scalajslib.api.ModuleKind
import scala.concurrent.duration._

object ScalaJsUtils {
  def runJSExpectingOut(path: os.Path, moduleKind: ModuleKind, expectedOut: String): Unit = {
    val input: org.scalajs.jsenv.Input = moduleKind match {
      case ModuleKind.NoModule => new Input.Script(path.toNIO)
      case ModuleKind.CommonJSModule => new Input.CommonJSModule(path.toNIO)
      case ModuleKind.ESModule => new Input.ESModule(path.toNIO)
    }
    val jsEnv = new NodeJSEnv()
    val testKit = new TestKit(jsEnv, 1.second)

    testKit.withComRun(Seq(input)){ run =>
      run.expectOut(expectedOut)
    }
  }
}
```

And was requiring `org.scala-js::scalajs-js-envs-test-kit:1.1.1` and `org.scala-js::scalajs-env-nodejs:1.1.1`

lolgab:
@lefou Could you try to rerun https://travis-ci.org/github/lihaoyi/mill/jobs/744146502 ?
It seems to fail for different reasons. Maybe it is flakiness.

scalajslib/worker/0.6/src/ScalaJSWorkerImpl.scala
```diff
@@ -15,14 +15,15 @@ import org.scalajs.jsenv._
...
+
   def link(sources: Array[File],
            libraries: Array[File],
            dest: File,
            main: String,
            testBridgeInit: Boolean, // ignored in 0.6
            fullOpt: Boolean,
-           moduleKind: ModuleKind) = {
+           moduleKind: ModuleKind,
+           useECMAScript2015: Boolean /* ignored in 0.6 */) = {
```
gzm0, line 26: This isn't ignored..

gzm0 (approved):
LG from Scala.js POV.

scalajslib/worker/0.6/src/ScalaJSWorkerImpl.scala
lolgab: So Scala.js 0.6.33 supports `useECMAScript2015` ?

lolgab: Yes, you're right! I'm using it 🙈

lolgab: Fixed ✅

lolgab:
Did a `git rebase -i` to reword the first commit.
Fixed the typo `useECMAScript215 => useECMAScript2015`, otherwise the first commit is untouched.

lefou:
Tavis job 2 is now always failing. Please ignore it for now.

lefou:
LGTM. I started another Travis Jobs just to be sure: https://travis-ci.org/github/lefou/mill/builds/744212036

lefou:
Thank you, @lolgab !

lolgab:
Thank you @lefou as usual for reviewing and merging 🙏😊

PR #1006, opened 2020-11-19
PathRef: detect permissions change

vdebergue:
This adds the ability for `PathRef` to detects changes on file permissions.

Previously, a change to a file permission was not detected when using `T.source` and the depending tasks were not executed.

### DIFF
--- a/main/api/src/mill/api/PathRef.scala
+++ b/main/api/src/mill/api/PathRef.scala
@@ -24,18 +24,25 @@ object PathRef {
    */
   def apply(path: os.Path, quick: Boolean = false): PathRef = {
     val sig = {
+      val isPosix = path.wrapped.getFileSystem.supportedFileAttributeViews().contains("posix")
       val digest = MessageDigest.getInstance("MD5")
       val digestOut = new DigestOutputStream(DummyOutputStream, digest)
+      def updateWithInt(value: Int): Unit = {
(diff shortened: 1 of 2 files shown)


lefou (commented):
Looks reasonable. We should have a test for it, though. Also, have you tried it on Windows?

vdebergue:
Happy to write a test about this. Do you know where it should be ?

I don't have a Windows available so I could not test it on this platform

lefou:
> Happy to write a test about this. Do you know where it should be ?

Best would be a test local to the `main.api` module (`main.api.test`), which currently doesn't exist, but is rather easy to add.

> I don't have a Windows available so I could not test it on this platform

That is ok. But we should at least make some manual tests on Windows before merging. I have seen some oddly permission behavior on Windows in the past.

vdebergue:
Added some tests for normal `sig` behaviour and with permissions change. I had some issue with `os.perms` (permissions were not correctly set or I used the api wrongly) so I used `java.nio` api instead

lefou (commented):
Test looks ok. But I think you forgot to include your changes to `build.sc`.

vdebergue:
Tests are run with `main.test` task so I didn't need to change anything in the `build.sc`

lefou:
You are right. Thank you for this pull request!

PR #1007, opened 2020-11-20
Cache Scala.js Linker to enable incremental linking

lolgab:
Originated from [this message on gitter.im](https://gitter.im/scala-native/scala-native?at=5fb274fd2a354407152b54f3).
This should solve the problem.

This PR caches all the linkers separately. For now there are 12 combinations. But adding new parameters could explode the possibilities even more.
I did this so you can jump between different settings of fastOpt/fullOpt, ModuleKind without losing the incremental linking.

### DIFF
--- a/scalajslib/worker/0.6/src/ScalaJSWorkerImpl.scala
+++ b/scalajslib/worker/0.6/src/ScalaJSWorkerImpl.scala
@@ -14,19 +14,20 @@ import org.scalajs.core.tools.logging.ScalaConsoleLogger
 import org.scalajs.jsenv._
 import org.scalajs.testadapter.TestAdapter
 
+import scala.collection.mutable
+import scala.ref.WeakReference
+
 class ScalaJSWorkerImpl extends mill.scalajslib.api.ScalaJSWorkerApi {
   private case class LinkerInput(fullOpt: Boolean, moduleKind: ModuleKind, useECMAScript2015: Boolean)
   private object ScalaJSLinker {
-    private var cache = Map.empty[LinkerInput, Linker]
-    def reuseOrCreate(input: LinkerInput): Linker =
-      cache.getOrElse(
-        input,
-        {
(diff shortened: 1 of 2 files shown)


lefou (commented):
Looks good to me.
Do you have any numbers or estimates, what the memory consumption of a cached linker is, and how much speed we gain by caching? My initial though was, that a module-local worker should be enough to cache the linker, but it might be wrong.

lolgab:
> Do you have any numbers or estimates, what the memory consumption of a cached linker is, and how much speed we gain by caching?

I don't. Let me run some tests and I'll let you know.
> My initial though was, that a module-local worker should be enough to cache the linker, but it might be wrong.

I don't know if that works because it is common to run fastOpt/fullOpt on the same module.

lolgab:
Didn't measure the speedup but it's apparent using it.
About memory consumption, a linker instance goes from ~60Mb (fastOpt) to ~120Mb(fullOpt) for a small/medium sized Scala.js application:
image

lefou:
Thanks for the numbers, @lolgab ! IMHO, this is already some significant memory, given, that we also keep cached Scala compilers and probably other tools around. Maybe, we can keep `WeakReference`s of the linker, so that the JVM has some options when in trouble. We already do this in `ZincWorkerImpl`.

lolgab:
Didn't know about the `WeakReference` trick. Nice!
Used it and changed the cache to a `mutable.Map` following `ZincWorkerImpl`.

lefou:
Thank you! I'm going to merge after the CI has finished. Travis-CI is currently disabled, so I'm going to start the CI job on my fork.

lolgab:
All the relevant jobs are green ✅ 
image
https://travis-ci.org/github/lefou/mill/builds/745613555

lolgab:
Thank you for the review as usual 🙂

PR #993, opened 2020-11-03
Github Workflow Actions setup

lefou:
Enabled Github Actions for mill repository. The publish job might not work yet, but I already included it to outline the setup. 

What's missing: 
* [x] Add the environment variables with required publishing credentials to the Github Workflow settings /cc @lihaoyi 
* [x] Review and adapt the publish scripts

PR #1008, opened 2020-11-23
WIP Convert CLI argument parsing to use latest version of ammonite+mainargs

lihaoyi:
This replaces the copy-paste logic handling both Mill's own CLI argument parsing as well as the CLI argument parsing of user-defined `T.command`s to instead rely on the https://github.com/lihaoyi/mainargs implementation. This is similar to the earlier PR that performed such a migration in the lihaoyi/Ammonite repo https://github.com/lihaoyi/Ammonite/pull/1126

Where the semantics of Mill's own CLI parsing and `T.command` CLI parsing differ, we pass in the corresponding flags to handle the difference, e.g. `allowPositional = false` for Mill's own parameters and `allowPositional = false` for `T.command`s.

Since the MainArgs library is standalone, we also cut the dependency on Scopt since it is now unnecessary. Our existing `scopt.Reads` are all replaced by `mainargs.TokensReader`

Things compile, haven't tested


lefou:
Looks good to me. I haven't used the new `mainargs` yet, so I can't say if it will fit all later needs, but given that it is already used in ammonite it is ok. (I did some local experiments with a single entry point main class, to avoid the logic in the start script, which also causes trouble for mill installations via coursier. But that solution needs to be done in Java and also requires us to parse the cmdline params in Java. There are very lightweight parsers in Java available as well, but it would means yet another change or some duplication.)

About the Github Actions setup: I already have PR #993 open using a test matrix and also working for scala native tests, which waits for your input regarding the publishing step.

lihaoyi:
Yeah I expect to merge this into master and try it out for a bit before cutting a stable release.

The mainargs code is pretty raw - took some surgery to extract it from ammonite, and then some re-organization to pretty it up and make it re-usable - but hopefully it's 99% there and we can manually suss out the last 1% before cutting a stable release of Mill + Ammonite

Sorry I didn't notice https://github.com/lihaoyi/mill/pull/993; I kind of hacked together the github actions config here from first principles. I'll probably merge this as is once i get things working, but feel free to replace it after if you think it could be done better somehow!

lefou:
@lihaoyi , looks like you have disabled Travis-CI for the whole mill repository. Can you split up this PR in case you want to stabilize the `mainargs` migration longer, so that we can merge the Github Actions part before? We currently have other PRs awaiting merges which no longer get CI builds.

lihaoyi:
@lefou I've turned travis back on

lihaoyi:
All tests pass, merging this. Hopefully the post-merge logic passes too...

PR #1013, opened 2020-11-26
CI: Added win build and try out some parallelism

lefou:


lefou:
I already integrated win builds.

PR #1019, opened 2020-11-30
Add "xsbti" as a shared prefix of the Scala compiler class loader

### LINKED ISSUE #1018: The ScalaInstance loader of Scala 3 is incompatible with the Zinc loader
The Scala 3 compiler depends directly on `org.scala-sbt:compiler-interface`, the Zinc interface.
Mill also depends on Zinc.

This is problematic because the `xsbti.*` classes can be loaded twice, the first time by the Mill class loader and the second time by the `ScalaInstance` class loader. It ends up with `LinkageException`.

So far the solution was to deconstruct the Zinc class loader and the underlying `ScalaInstance.loader` in the Scala 3 compiler bridge. But this is a very hacky and fragile solution ([see code here](https://github.com/lampepfl/dotty/blob/master/sbt-bridge/src/xsbt/CompilerClassLoader.java))

A better alternative would be to build a `scalaInstance.loader` in Mill ([current implementation here](https://github.com/lihaoyi/mill/blob/aa3240a52a72cfa5fe76c4d4c747f7f61eef2f1d/scalalib/worker/src/ZincWorkerImpl.scala#L309-L321)) that is compatible with the Zinc loader.

adpi2:
Fixes #1018

### DIFF
--- a/main/api/src/mill/api/ClassLoader.scala
+++ b/main/api/src/mill/api/ClassLoader.scala
@@ -11,6 +11,7 @@ object ClassLoader {
   def java9OrAbove = !System.getProperty("java.specification.version").startsWith("1.")
   def create(urls: Seq[URL],
              parent: java.lang.ClassLoader,
+             sharedLoader: java.lang.ClassLoader = getClass.getClassLoader,
              sharedPrefixes: Seq[String] = Seq())
             (implicit ctx: Ctx.Home): URLClassLoader = {
     new URLClassLoader(
@@ -19,7 +20,7 @@ object ClassLoader {
     ) {
       val allSharedPrefixes = sharedPrefixes :+ "com.sun.jna"
       override def findClass(name: String): Class[_] = {
-        if (allSharedPrefixes.exists(name.startsWith)) getClass.getClassLoader.loadClass(name)
+        if (allSharedPrefixes.exists(name.startsWith)) sharedLoader.loadClass(name)
         else super.findClass(name)
       }
     }
--- a/main/src/modules/Jvm.scala
+++ b/main/src/modules/Jvm.scala
@@ -180,8 +180,7 @@ object Jvm {
                   (implicit ctx: Ctx.Home): T = {
     val urls = classPath.map(_.toIO.toURI.toURL)
     val cl = if (classLoaderOverrideSbtTesting) {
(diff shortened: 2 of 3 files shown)


adpi2:
@lihaoyi Could you review this pull request?

Here is the context: I want to remove the messy [CompilerClassLoader](https://github.com/lampepfl/dotty/blob/master/sbt-bridge/src/xsbt/CompilerClassLoader.java) in the Scala 3 bridge with Zinc. But if I do so I would break all Mill projects in the `community-build`. So my plan is:
- fix the Scala 3 loader in Mill
- release Mill
- update the Scala 3 community build projects
- fix the Scala 3 loader in `sbt-dotty` (https://github.com/lampepfl/dotty/pull/10541)
- remove the `CompilerClassLoader`

adpi2:
This is ready for review! @lefou @lihaoyi could you have a look at it?

lefou (commented):
Sorry for my late review. My only concern about this was, that we probably can't handle incompatible versions of the xsbti. But if I now understand correctly, we only share it with the classloader of the zinc worker, not with mills main classloader. So, this looks reasonable to me.

Just one minor thing. I'd like to keep non-wildcard imports were possible.

adpi2:
> But if I now understand correctly, we only share it with the classloader of the zinc worker, not with mills main classloader.

Yes, exactly, we share the `xsbti` classloader with the Zinc worker which contains the zinc compiler bridge and the Scala compiler. This was not necessary before because the Scala 2 compiler does not depend on `org.scala-sbt:compiler-interface` (aka `xsbti`). But the dependency is inverted in Scala 3 and it is crucial that all the `xsbti` classes are loaded from the same loader.

adpi2:
I pushed a fix for the wildcard import.

I don't know why the CI is stuck but I don't think it's related to this specific PR.

lefou (approved):

lefou:
Thank you!

adpi2:
Thank you @lefou for merging this. Since I want to follow this in lampepfl/dotty#10541 I'd like to know what the schedule for the next release.

lefou:
You can already start to use it. Its version is `0.9.3-21-002361`. I think we will also tag a new release later this month, depending on how fast we can merge the open PRs.

Issue #685, opened 2019-09-02
ScalaPB module should use `T.worker` for it's worker cache


lefou:
Please don't open issues for questions, but ask in our gitter channel at https://gitter.im/lihaoyi/mill

lefou:
* Fixed by #1027

PR #1026, opened 2020-12-03
ScalaPB: use isolated classloader

chikei:
If you use ScalaPBC 0.10.9+, `compileScalaPB` will failed with error message like `--scala_out: java.lang.NoSuchMethodError: 'com.google.protobuf.ProtocolStringList scalapb.options.Scalapb$MessageOptions.getUnknownFieldsAnnotationsList()'`. This is due to mill has ScalaPB runtime as transitive dependency which shares some class with ScalaPBC and those classes have introduced new methods for ScalaPBC in 0.10.9, and current classloader usage will find these classes as old one.


lefou:
I think, we should keep the ScalaPB runtime in a worker (http://www.lihaoyi.com/mill/page/tasks.html#workers). See also #685.

PR #1027, opened 2020-12-03
ScalaPB: various fix and improves

chikei:
* fix destination path issue mentioned in https://github.com/lihaoyi/mill/pull/1002#issuecomment-737769495
* isolated classloader as #1026
* use T.worker

### DIFF
--- a/contrib/scalapblib/test/src/TutorialTests.scala
+++ b/contrib/scalapblib/test/src/TutorialTests.scala
@@ -1,5 +1,7 @@
 package mill.contrib.scalapblib
 
+import mill.T
+import mill.api.PathRef
 import mill.util.{TestEvaluator, TestUtil}
 import utest.framework.TestPath
 import utest.{TestSuite, Tests, assert, _}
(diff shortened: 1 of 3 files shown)


contrib/scalapblib/test/src/TutorialTests.scala
```diff
@@ -33,21 +34,19 @@ object TutorialTests extends TestSuite {
 
   object TutorialWithAdditionalArgs extends TutorialBase {
     object core extends TutorialModule {
-      override def scalaPBAdditionalArgs = Seq(
-        s"--zio_out=..."
-      )
+      override def scalaPBAdditionalArgs = T{Seq(
+        "--additional-test=..."
+      )}
```
lefou, line 39: What's the point of changing this?

lefou (commented):
Looks good to me. I added one question below.

Also, I just recognized that we have a commented-out test case for ScalaPB. Could you please check if this test works for you locally? I'm pretty sure we can handle non-abilities of the CI env better than commenting out. (Can't tell if there a equivalent to  scalatest's `assume` in utest). I could imagine that this test will work on Github Actions as-is.

chikei:
Ya, I've verified the `calledDirectly` test works locally before push.

contrib/scalapblib/test/src/TutorialTests.scala
chikei: I was written from the version before last scalapb pr and thinking "additional-test" might easier for other people to reasoning and not specific for zio, not something matters in term of code execution.

lefou:
Thank you!

PR #1029, opened 2020-12-03
BSP: Provide mill source files

joan38:
Not sure why the following does not bring in the dependencies anymore since `0.9.x`:
```
val classpath = Try(evaluator.rootModule.getClass.getClassLoader.asInstanceOf[SpecialClassLoader]).fold(
  _ => Seq.empty,
  _.allJars
)
```
So this is resolving the source files directly. Let me know if you have a better idea.

### DIFF
--- a/bsp/src/mill/bsp/ModuleUtils.scala
+++ b/bsp/src/mill/bsp/ModuleUtils.scala
@@ -2,6 +2,7 @@ package mill.bsp
 
 import ammonite.runtime.SpecialClassLoader
 import ch.epfl.scala.bsp4j._
+import coursier.Resolve
 import java.net.URL
 import mill._
 import mill.api.Result.Success
@@ -9,15 +10,14 @@ import mill.api.{PathRef, Strict}
 import mill.define._
 import mill.eval.{Evaluator, _}
 import mill.scalajslib.ScalaJSModule
+import mill.scalalib._
 import mill.scalalib.Lib.{depToDependency, resolveDependencies, scalaRuntimeIvyDeps}
 import mill.scalalib.api.Util
-import mill.scalalib.{JavaModule, ScalaModule, TestModule}
 import mill.scalanativelib._
 import mill.util.Ctx
 import os.{Path, exists}
 import scala.collection.JavaConverters._
 import scala.util.Try
-import coursier.core.Repository
 
 /**
  * Utilities for translating the mill build into
@@ -86,9 +86,8 @@ object ModuleUtils {
     val scalaOrganization = "org.scala-lang"
     val scalaLibDep = scalaRuntimeIvyDeps(scalaOrganization, BuildInfo.scalaVersion)
 
-    val repos = Evaluator.evalOrElse(evaluator, T.task {
-        T.traverse(modules)(_.repositoriesTask)()
-      }, Seq.empty[Seq[Repository]])
+    val repos = Evaluator
+      .evalOrElse(evaluator, T.traverse(modules)(_.repositoriesTask), Seq.empty)
       .flatten
       .distinct
 
@@ -112,15 +111,24 @@ object ModuleUtils {
     target
   }
 
-  def getMillBuildClasspath(evaluator: Evaluator, source: Boolean): Seq[String] = {
-    val all = Try(evaluator.rootModule.getClass.getClassLoader.asInstanceOf[SpecialClassLoader]).fold(
+  def getMillBuildClasspath(evaluator: Evaluator, sources: Boolean): Seq[String] = {
+    val classpath = Try(evaluator.rootModule.getClass.getClassLoader.asInstanceOf[SpecialClassLoader]).fold(
       _ => Seq.empty,
       _.allJars
     )
-    val filtered =
-      if (source) all.filter(url => isSourceJar(url))
+
+    val millJars = resolveDependencies(
+      Resolve.defaultRepositories,
+      depToDependency(_, BuildInfo.scalaVersion),
+      Versions.millEmbeddedDeps,
+      sources = sources
+    ).asSuccess.toSeq.flatMap(_.value).map(_.path.toNIO.toUri.toURL)
+
+    val all = classpath ++ millJars
+    val binarySource =
+      if (sources) all.filter(url => isSourceJar(url))
       else all.filter(url => !isSourceJar(url))
-    filtered.filter(url => exists(Path(url.getFile))).map(_.toURI.toString)
+    binarySource.filter(url => exists(Path(url.getFile))).map(_.toURI.toString)
   }
 
   /**
--- a/bsp/src/mill/bsp/ModuleUtils.scala
+++ b/bsp/src/mill/bsp/ModuleUtils.scala
@@ -120,7 +120,7 @@ object ModuleUtils {
     val millJars = resolveDependencies(
       Resolve.defaultRepositories,
       depToDependency(_, BuildInfo.scalaVersion),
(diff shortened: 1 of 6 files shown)


joan38:
@lefou this is ready for review

bsp/src/mill/bsp/ModuleUtils.scala
```diff
@@ -86,9 +86,8 @@ object ModuleUtils {
     val scalaOrganization = "org.scala-lang"
     val scalaLibDep = scalaRuntimeIvyDeps(scalaOrganization, BuildInfo.scalaVersion)
 
-    val repos = Evaluator.evalOrElse(evaluator, T.task {
-        T.traverse(modules)(_.repositoriesTask)()
-      }, Seq.empty[Seq[Repository]])
+    val repos = Evaluator
+      .evalOrElse(evaluator, T.task(T.traverse(modules)(_.repositoriesTask)()), Seq.empty)
```
lefou, line 90: The `T.task` is probably redundant. `T.traverse` already produces a task.

scalalib/src/Lib.scala
```diff
@@ -36,7 +36,7 @@ object Lib{
                                   depToDependency: Dep => coursier.Dependency,
                                   deps: TraversableOnce[Dep],
                                   mapDependencies: Option[Dependency => Dependency] = None,
-                                  ctx: Option[mill.util.Ctx.Log] = None) = {
+                                  ctx: Option[mill.util.Ctx.Log] = None): (Seq[Dependency], Resolution) = {
```
lefou, line 39: :+1: for explicit return type.

bsp/src/mill/bsp/ModuleUtils.scala
```diff
@@ -112,15 +111,30 @@ object ModuleUtils {
...
+    val millJars = resolveDependencies(
+      Resolve.defaultRepositories,
+      depToDependency(_, BuildInfo.scalaVersion),
+      Agg(
+        ivy"com.lihaoyi::mill-main:$millVersion",
+        ivy"com.lihaoyi::mill-scalalib:$millVersion",
+        ivy"com.lihaoyi::mill-scalajslib:$millVersion",
+        ivy"com.lihaoyi::mill-scalanativelib:$millVersion"
+      ),
+      sources = sources
```
lefou, line 130: I think these deps were never there before, at least all my experiments with mill-bsp resulted in missing mill API when editing `build.sc`. But, we should not bury such  a hardcoded list here. Instead, this is knowledge best located in the buildfile (`build.sc`). We could generate a Scala file there (`generateSources` or BuildInfo plugin) and use it here. We have this exactly hard-coded list also in `GenIdeaImpl`, and I think this is an opportunity to fix both places the right way.

lefou (commented):
I suspect the `SpecialClassLoader` never carried the mill API. But, maybe it's a good thing to be more verbose in case we don't get a `SpecialClassLoader`.

I made some comments below.

bsp/src/mill/bsp/ModuleUtils.scala
joan38: Excellent idea! Check out the new commit.

bsp/src/mill/bsp/ModuleUtils.scala
```diff
@@ -112,15 +111,24 @@ object ModuleUtils {
...
       _ => Seq.empty,
       _.allJars
     )
-    val filtered =
-      if (source) all.filter(url => isSourceJar(url))
+
+    val millJars = resolveDependencies(
+      Resolve.defaultRepositories,
+      depToDependency(_, BuildInfo.scalaVersion),
+      BuildInfo.millEmbeddedModules.map(module => ivy"com.lihaoyi::mill-$module:${BuildInfo.millVersion}"),
```
lefou, line 123: I think using full ivy artifact is the way to go (using `module.publishSelfDependency()`  when generating the `BuildInfo`). Less assumptions about the deps.

joan38: Check out the new commit.

build.sc
```diff
@@ -260,6 +261,8 @@ object scalalib extends MillModule {
         |  val ammonite = "${Deps.ammonite.dep.version}"
         |  /** Version of Zinc. */
         |  val zinc = "${Deps.zinc.dep.version}"
+        |  /** Dependency artifacts embedded in mill by default. */
+        |  val millEmbeddedDeps = ${artifacts.map(artifact => s"""ivy"${artifact.group}::${artifact.id}:${artifact.version}"""")}
```
lefou, line 265:
```suggestion
        |  val millEmbeddedDeps = ${artifacts.map(artifact => s""""${artifact.group}:${artifact.id}:${artifact.version}"""")}
```

Artitfact ID is already platform-resolved. Also I'd rather just use the string.

scalalib/src/GenIdeaImpl.scala
```diff
@@ -82,13 +82,11 @@ case class GenIdeaImpl(evaluator: Evaluator,
...
 
           val repos = moduleRepos.foldLeft(Set.empty[Repository])(_ ++ _) ++ Set(LocalRepositories.ivy2Local, Repositories.central)
-          val artifactNames = Seq("main-moduledefs", "main-api", "main-core", "scalalib", "scalajslib")
           val Result.Success(res) = scalalib.Lib.resolveDependencies(
             repos.toList,
             Lib.depToDependency(_, "2.13.2", ""),
-            for(name <- artifactNames)
-            yield ivy"com.lihaoyi::mill-$name:${sys.props("MILL_VERSION")}",
-            false,
+            Versions.millEmbeddedDeps,
```
lefou, line 88:
```suggestion
            Versions.millEmbeddedDeps.map(d => ivy"$d"),
```

bsp/src/mill/bsp/ModuleUtils.scala
```diff
@@ -112,15 +111,24 @@ object ModuleUtils {
...
       _ => Seq.empty,
       _.allJars
     )
-    val filtered =
-      if (source) all.filter(url => isSourceJar(url))
+
+    val millJars = resolveDependencies(
+      Resolve.defaultRepositories,
+      depToDependency(_, BuildInfo.scalaVersion),
+      Versions.millEmbeddedDeps,
```
lefou, line 123:
```suggestion
      Versions.millEmbeddedDeps.map(d => ivy"$d"),
```

lefou (commented):
@joan38 I'm sorry when it appears as nitpicking. I think keeping the generated file dependency free is a good thing, so I suggest to use a plain string to transport the dependency. I added  suggestions to show what I have in mind. Don't know if it compiles though.

If test work, this looks good to me! Looking forward to try that out.

lefou:
@joan38, since our move to GitHub Actions, there is no longer the need to rebase or force-pushg all PRs to latest master to let tests pass. We can squash when merging.

joan38:
Your suggestions a good @lefou, keep on nitpicking that's how we achieve perfection :)

Do you recommend moving this back to `BuildInfo` in core since we don't need access to `Dep` in scalalib?

I'm just used to amend force pushing but would incremental commits better help incremental reviewing?

lefou:
> Your suggestions are good @lefou, keep on nitpicking that's how we achieve perfection :)

You asked for it. ;-)

> Do you recommend moving this back to `BuildInfo` in core since we don't need access to `Dep` in scalalib?

Yes.

> I'm just used to amend force pushing but would incremental commits better help incremental reviewing?

Yeah, avoiding forced pushes would greatly help, esp. in the Github UI.

Issue #1005, opened 2020-11-18
[BSP] Improper handling of build log output


srdo:
Following up on this discussion https://github.com/lihaoyi/mill/pull/1003#discussion_r524061548 I did a little poking at the current MillBspLogger and IntelliJ to see what we actually emit to the IDE currently.

#### Non-JSONRPC in the output sent to the IDE
Here's a line we emit when the MillBspLogger ticker is called
```
Content-Length: 244

{"jsonrpc":"2.0","method":"build/taskProgress","params":{"taskId":{"id":"1822590398"},"eventTime":1605708854564,"message":"[1/1] mill.bsp.BSP.start \u003e [340/344] bsp.test.resolvedIvyDeps ","total":1,"progress":1,"unit":"mill.bsp.BSP.start"}}[1/1] mill.bsp.BSP.start > [340/344] bsp.test.resolvedIvyDeps 
```
Note the trailing repeat of the message. This is because MillBspLogger is emitting to stdout both via the BSP client library, and directly via the proxied logger. IntelliJ ignores this trailing part, so is not visible in the UI, but other BSP clients may behave differently, since as far as I can tell, the spec doesn't say how to handle this.

#### Repeated log messages
When we call MillBspLogger.info, the message shows up twice in IntelliJ. The output we send to the IDE is

```
Hell
Content-Length: 83

{"jsonrpc":"2.0","method":"build/showMessage","params":{"type":3,"message":"Hell"}}
```
In this case the ordering is different because we happen to log to the proxied logger first, and then send the response on the client. I'm uncertain if this is the reason both messages are displayed by IJ. Either way, it is still out of spec.

#### Unexpected build failure
When we call MillBspLogger.error, the build fails. IJ happens to consider the build failed if it is sent a build/showMessage at error level. I can't find anywhere in the spec that says showMessage should be interpreted as a build failure, so maybe IJ is in the wrong here.

srdo:
I think the way to look at this is that the BSP server should output only the JSONRPC messages the specification defines, and not emit arbitrary extra content into stdout. The MillBspLogger should probably not be a proxy logger, instead it should forward messages to the BSP client. This should happen for both the info/debug/error methods, and also happen for the streams exposed by MillBspLogger.

The build failure from error logging I think might be an IJ issue. I'm not sure that's how those messages should be interpreted.

lefou:
[...]

PR #1032, opened 2020-12-04
Clean up BSP logging output.

srdo:
Previously, the MillBspLogger proxied the evaluator logger, which meant all
output went twice to stdout, once through the proxy and once through the BSP client
messages.
    
MillBspLogger now does not proxy a logger, instead it only writes to the BSP
client. It will write whatever is emitted to its streams, or written
via the Logger methods on it.
    
The logger no longer uses the ticker formatting or colors offered by PrintLogger.
This decoration is only suitable for the terminal, since e.g. color codes and navigation codes
are unlikely to be supported in an editor's log output.
    
To debug this, I've added a log file that logs the raw json output from
BSP/start. I've also moved the BSP client logging to the BSP/start dest
directory, so it no longer goes in .bsp/mill.log.
    
ProxyLogger is no longer a case class. It is intended for subclassing,
so it not suitable as a case class.

### DIFF
--- a/bsp/src/mill/bsp/MillBspLogger.scala
+++ b/bsp/src/mill/bsp/MillBspLogger.scala
@@ -1,57 +1,86 @@
 package mill.bsp
 
+import ammonite.util.Colors
 import ch.epfl.scala.bsp4j._
-import mill.api.Logger
-import mill.util.{ColorLogger, ProxyLogger}
+import java.io.{ ByteArrayOutputStream, InputStream, PrintStream }
+import java.nio.charset.Charset
+import mill.bsp.MillBspLogger.tickerPattern
+import mill.util.ColorLogger
 
+object MillBspLogger {
+  /**
+   * Creates a BSP-specialized logger class which sends `task-progress`
+   * notifications ( upon the invocation of the `ticker` method ) and
+   * `show-message` notifications ( for each error or information
+   * being logged ).
+   *
+   * @param client the client to send notifications to, also the
+   *               client that initiated a request which triggered
+   *               a mill task evaluation
+   * @param taskId unique ID of the task being evaluated
+   * @param inputStream The input stream to read input from
+   */
+  def createBspLogger(client: BuildClient, taskId: Int, inputStream: InputStream): ColorLogger = {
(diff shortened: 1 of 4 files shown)


srdo:
This isn't quite ready, it will cause Mill to fail to build itself.

For some reason Intellij interprets ShowMessage at error level as a signal to fail the build. Since Mill occasionally emits to stderr, this causes build failures. I'll go ask them why this is, since the BSP spec does not say to interpret ShowMessage this way.

There's a minor usability issue around tickers. https://github.com/lihaoyi/mill/blob/3f8afc2aaa0947283955022dacebda80413a64ac/main/core/src/util/Loggers.scala#L134-L138

While we might want tickers enabled on BSP, we do not want the navigation symbols (e.g. `[1A`) in the logging output, since editors are likely to display these symbols to users. I'm unsure what the best solution is here, maybe a new config flag that disables these symbols?

srdo:
Issue raised with IJ here https://youtrack.jetbrains.com/issue/SCL-18538

srdo:
For now I have put in a workaround that makes messages written to the logger's error stream go out as ShowMessage as info level. 

It is not clear to me what the right thing is to do here.

srdo:
fixes #1005

bsp/src/mill/bsp/MillBspLogger.scala
```diff
@@ -1,57 +1,86 @@
...
     client.onBuildShowMessage(new ShowMessageParams(MessageType.ERROR, s))
   }
 
-  override def info(s: String): Unit = {
-    super.info(s)
-    client.onBuildShowMessage(new ShowMessageParams(MessageType.INFORMATION, s))
+  override def ticker(s: String): Unit = {
+    try {
+      client.onBuildShowMessage(new ShowMessageParams(MessageType.INFORMATION, s))
+      tickerPattern.findFirstMatchIn(s).foreach { ticker =>
```
srdo, line 60: This ticker implementation is broken in a slightly different way from how it was broken previously. See https://github.com/lihaoyi/mill/issues/1034. Fixing that will likely require changing this again.

lefou:
BSP Support has undergone a large refactoring. Closing this due to too many conflicts.

PR #1072, opened 2020-12-15
CI: Cache coursier download dir

lefou:


lefou:
Can't update the cache, so a proper setup would be quite more complex.

PR #1078, opened 2020-12-16
Adjust Docs to MainArgs Syntax for `--gpgArgs`

fleipold:
With the move to MainArgs multiple arguments in a list are no longer
passed in as a comma separated list.

This commit reflects this change in the documentation and also updates
the release script accordingly, so that it will  work with mill 0.9.3.

### DIFF
--- a/ci/release-maven.sh
+++ b/ci/release-maven.sh
@@ -15,7 +15,14 @@ rm -rf ~/.mill
 
 out/dev/assembly/dest/mill mill.scalalib.PublishModule/publishAll \
     --sonatypeCreds $SONATYPE_DEPLOY_USER:$SONATYPE_DEPLOY_PASSWORD \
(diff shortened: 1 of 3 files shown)


lihaoyi:
I think maybe the right thing to do here is to make `--gpgArgs` get passed in as a comma-separated list. To do that in Mainargs you can define a wrapper type e.g. `class CommaSeparated(values: Seq[String])` and define a custom argument parser (https://github.com/lihaoyi/mainargs#custom-argument-parsers). That would help avoid churn in the ecosystem of people needing to update all their CI scripts etc. for what is really an accidental/incidental change

fleipold:
Ah, I understood this to be an intended change (and of course the way I realised that there was a change, was a "broken" release script).
 
I found the new syntax cleaner, which is probably why you introduced it in MainArgs, but it is a bit tedious in this particular case where you pass in a lot of values, so as you suggest sticking with the old behaviour a good solution.

I am not sure I'll find the time to make the proper fix. Closing this PR for now.

PR #1082, opened 2020-12-17
Don't follow symlinks when checking permissions of files

jodersky:
There's no guarantee that the symlink is actually valid. Note that this is called in a loop that itself follows valid symlinks, so there should not be any files missed.


lefou:
Thanks for the fix. This is indeed new code and explains why it worked with older mill versions. Would it be possible to reconstruct a failing case and add it to `PathRefTests`?

jodersky:
Sure, I added some tests related to symlink edge cases

lefou (approved):
Looks good to me. Thank you!

PR #1089, opened 2020-12-25
Replace Win32 named pipe with simple ServerSocket

sake92:
This currently works fine, but spawns a new server on every invocation!?  
I'm not sure if this is a problem introduced by this change but I doubt it.  
Maybe I just don't understand the locks code in Mill well enough.. 😄 

Anyways, it runs a simple `ServerSocket` on a random port on localhost, and writes that port in a file.  
Client can then read the port and connect to the server.  

I also "tested" the security part of server, by running an Android app that scans the ports of my computer.  
It doesn't show the allocated port, so should be good.

### DIFF
--- a/main/client/src/mill/main/client/MillClientMain.java
+++ b/main/client/src/mill/main/client/MillClientMain.java
@@ -3,9 +3,10 @@ package mill.main.client;
 import org.scalasbt.ipcsocket.*;
 
 import java.io.*;
-import java.net.Socket;
-import java.net.URISyntaxException;
+import java.net.*;
 import java.nio.channels.FileChannel;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.*;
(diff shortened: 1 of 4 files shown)


lihaoyi:
One thing missing from this PR is a description of what it is actually trying to accomplish, an understanding of what the behavior and problems of the current system are, and an explanation of how this is the best solution to fix those problems  😛 

The client/server code could definitely use a lot of work, particularly on windows, and definitely has bugs. But it's a pretty tricky area of code involving multi-process concurrency and locking, so it demands more "measure twice cut once"-style changes rather than "move fast and break things"

sake92:
@lihaoyi yeah, that makes sense. Can you describe a bit the logic behind `MillServerMain` code?  
I mean just conceptually, what is its main logic trying to achieve?  
Then we could get to some conclusion and improvements maybe? 😄 

For example, it's not clear to me why in `while (running) {` it reopens the server socket?

lihaoyi:
The logic is trying to implement the following:

1. If there is no MillServerMain, or there are existing MillServerMains but they're all in use, we create a new one and connect to it
2. If there is an existing MillServerMain running that is not currently being used, we connect to it and use it
3. If there are n=5 existing MillServerMains, and all are in use, we fail with an error (just to stop someone from accidentally spawning 100s of background processes)
4. An individual MillServerMain terminates after n=5 minutes or inactivity, or if Ctrl-C is given while it is executing tasks
5. If a MillServerMain is given Ctrl-C while in the file-watching mode, it should not terminate but remain in the background to pick up further tasks

A lot of the complexity around locks etc. is to do this reliably in the presence of multiple Mill clients being spawned concurrently and in arbitrary orders, with the given commands taking arbitrary amounts of time to run

sake92:
#1230 tries to fix existing code instead of this.

PR #1094, opened 2021-01-04
Bump zinc to 1.4.4

adpi2:
Unblock https://github.com/lampepfl/dotty/issues/10816

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -309,6 +309,25 @@ object scalalib extends MillModule {
     def testArgs = T{Seq(
       "-DMILL_SCALA_WORKER=" + runClasspath().map(_.path).mkString(",")
     )}
+
+    override def generatedSources = T{
+      val dest = T.ctx.dest
+      val artifacts = T.traverse(dev.moduleDeps)(_.publishSelfDependency)()
+      os.write(dest / "Versions.scala",
+        s"""package mill.scalalib.worker
+           |
+           |/**
+           | * Dependency versions.
+           | * Generated from mill in build.sc.
+           | */
+           |object Versions {
+           |  /** Version of Zinc. */
+           |  val zinc = "${Deps.zinc.dep.version}"
+           |}
+           |
+           |""".stripMargin)
+      super.generatedSources() ++ Seq(PathRef(dest))
+    }
   }
 }
 
--- a/build.sc
+++ b/build.sc
@@ -78,7 +78,7 @@ object Deps {
   val sourcecode = ivy"com.lihaoyi::sourcecode:0.2.1"
   val upickle = ivy"com.lihaoyi::upickle:1.2.2"
   val utest = ivy"com.lihaoyi::utest:0.7.5"
-  val zinc = ivy"org.scala-sbt::zinc:1.4.0-M1"
+  val zinc = ivy"org.scala-sbt::zinc:1.4.4"
   val bsp = ivy"ch.epfl.scala:bsp4j:2.0.0-M13"
   val jarjarabrams = ivy"com.eed3si9n.jarjarabrams::jarjar-abrams-core:0.3.0"
 }
--- a/contrib/proguard/test/src/ProguardTests.scala
+++ b/contrib/proguard/test/src/ProguardTests.scala
@@ -19,7 +19,7 @@ object ProguardTests extends TestSuite {
     override def millSourcePath: os.Path =
       TestUtil.getSrcPathBase() / millOuterCtx.enclosing.split('.')
 
-    override def scalaVersion = "2.12.0"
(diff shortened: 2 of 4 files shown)


lefou:
Unfortunately, it's isn't as easy as just bumping it. zinc API changed after 1.4.0-M1 and introduced some virtual file abstraction. The change to mills code base isn't trivial, at least not if your're not familiar with zinc API.

adpi2:
This PR is ready for review.

It bumps the version of Zinc but it does not take advantage of the new Zinc features: VirtualFile and Stamps.

A few remarks:
- The zinc `compiler-interface` now depends on the Scala library. It cannot be loaded globally in the `ZincWorkerModule` anymore. So it is loaded once for each scala version.
- Zinc cannot reuse the compiler bridges compiled by older version of Mill. So I classified them by the version of Zinc. The compiler bridge is now stored in `out/mill/scalalib/ZincWorkerModule/worker/dest/zinc-<zincVersion>/<scalaVersion>`.
- The current version of Zinc is incompatible with 2.12.0 but not with 2.12.1.

lefou:
@adpi2 Thank you very much for this upgrade. The change looks good to me after a first glace, but I'd like to give it a second view. I have some questions, see below.

> This PR is ready for review.
> 
> It bumps the version of Zinc but it does not take advantage of the new Zinc features: VirtualFiles and Stamps.

I have no idea, what we miss because of this. Could you point out what benefits we could gain? Do you see any low hanging fruits?

> A few remarks:
> 
>   * The zinc `compiler-interface` now depends on the Scala library. It cannot be loaded globally in the `ZincWorkerModule` anymore. So it is loaded once for each scala version.

If I understand correctly, we still load and share the zinc worker globally for modules with the same Scala version, right?

>   * Zinc cannot reuse the compiler bridges compiled by older version of Mill. So I classified them by the version of Zinc. The compiler bridge is now stored in `out/mill/scalalib/ZincWorkerModule/worker/dest/zinc-<zincVersion>/<scalaVersion>`.
> 
>   * The current version of Zinc is incompatible with 2.12.0 but not with 2.12.1.

What's the impact of it? If it only means we cannot use 2.12.0 this might be ok, as long as we print a proper error message indicating the issue and the solution in case a module wants to use `2.12.0`. Are we still able to compile older Scala versions, e.g. 2.11 ?

>   * The compiler bridge now depends on its resources to load itself

Can you elaborate what this means?

adpi2:
> I have no idea, what we miss because of this. Could you point out what benefits we could gain? Do you see any low hanging fruits?

The purpose of `VirtualFile` and `Stamps` is to enable [remote caching](https://eed3si9n.com/cached-compilation-for-sbt). It makes possible to download a the compilation result, including the zinc analysis file, produced on one computer to benefit from the incremental compilation on another computer.

In terms of implementation, it requires configuring the `FileConverter` and the `StampReader` properly. But then it is only useful if there is an infrastructure to publish and retrieve the cache.

> If I understand correctly, we still load and share the zinc worker globally for modules with the same Scala version, right?

Yes. The zinc worker is a singleton and it creates only one compiler bridge for modules with the same Sala version.

> What's the impact of it? If it only means we cannot use 2.12.0 this might be ok, as long as we print a proper error message indicating the issue and the solution in case a module wants to use 2.12.0. Are we still able to compile older Scala versions, e.g. 2.11 ?

It's only incompatible with `2.12.0`. Scala `2.10` and `2.11` can still be used. I pushed a commit to improve the error message

adpi2:
> Can you elaborate what this means?

Mill downloads the source of the `compiler-bridge`, then compiles it and loads the produced class files. But since Zinc 1.4.0 the  `compiler-bridge` jar also contains some resource files that must be copied to the destination folder so that the Zinc class loader can find them. Here they are: https://github.com/sbt/zinc/tree/develop/internal/compiler-bridge/src/main/resources/META-INF/services

lefou:
> > I have no idea, what we miss because of this. Could you point out what benefits we could gain? Do you see any low hanging fruits?
> 
> The purpose of `VirtualFile` and `Stamps` is to enable [remote caching](https://eed3si9n.com/cached-compilation-for-sbt). It makes possible to download a the compilation result, including the zinc analysis file, produced on one computer to benefit from the incremental compilation on another computer.
> 
> In terms of implementation, it requires configuring the `FileConverter` and the `StampReader` properly. But then it is only useful if there is an infrastructure to publish and retrieve the cache.

Ok, so we can safely live without that for now. It will be relevant for #726 though.

lefou (approved):
This looks good to me.

lefou:
Thank you, Adrien!

PR #1108, opened 2021-01-12
Support Scala Native 0.4.0

lolgab:
- Drop support for Scala Native 0.3
- Remove `Release` ReleaseMode
- Make `nativeTarget` `Option[String]` like upstream
- Rewrote tests to JUnit. Since Scala Native ships with JUnit and Utest
  depends on Mill itself we can't use it to test Mill.
  JUnit is the official testing framework and stresses all the features
  that were tested with Utest and Scalatest
- Testing Scala Native for Scala 2.11, 2.12 and 2.13
- Move worker again into the `0.4` folder. It was moved outside using
  `def millSourcePath = super.millSourcePath / os.up` to reuse the code
  between worker `0.3` and `0.4`. Since now we're dropping `0.3` it
  doesn't make sense anymore to do this, and we can be prepared for a
  future 0.5


scalanativelib/test/src/HelloNativeWorldTests.scala
```diff
@@ -49,38 +48,33 @@ object HelloNativeWorldTests extends TestSuite {
...
-        def ivyDeps = Agg(ivy"com.lihaoyi::utest::0.7.5")
-        def testFrameworks = Seq("utest.runner.Framework")
+        override def sources = T.sources{ millSourcePath / "src" / "no-tests" }
+        def testFrameworks = Seq("com.novocode.junit.JUnitFramework")
+        override def ivyDeps = Agg(
+          ivy"org.scala-native::junit-runtime::$sNativeVersion"  
+        )
+        override def compileIvyDeps = Agg(  
+          ivy"org.scala-native:::junit-plugin:$sNativeVersion"  
+        )
```
lefou, line 77: This looks like a repetitiion of the JUnit-variant. Maybe, we can have it in a shared trait?

scalanativelib/api/src/GetFrameworkResult.java
```diff
@@ -0,0 +1,14 @@
+package mill.scalanativelib.api;
```
lefou, line 1: This source files should be located in a directory structure reflecting the package name. The Java Standard isn't as flexible as the Scala one. Beside the fact that it just works with zinc, it won't for pure Java compiler.

Same, for all other Java files.

lefou (commented):
Thanks for keeping Scala Native support up-to-date (and ahead)! I had a quick look at it. Looks good so far. I added some comments. And we still need to wait for 0.4.0 release to really test it.

scalanativelib/api/src/GetFrameworkResult.java
lolgab: @lefou You're right! Moved them ✅

scalanativelib/test/src/HelloNativeWorldTests.scala
lolgab: Extracted on a separate trait ✅

lolgab:
@lefou I tried in every way on earth to make the Scala Native JUnit framework to work on mill but I can't make it work.
I tried manually with utest and everything works fine.
If you agree I would change it back to `utest` releasing a version for Scala Native `0.4.0` from my branch to bootstrap the test. When Mill is released I can update utest and move the test to use the official one from `com.lihaoyi`.
What do you think?

lefou:
> @lefou I tried in every way on earth to make the Scala Native JUnit framework to work on mill but I can't make it work.
> I tried with utest (after publishing an updated version locally) and everything works fine.
> If you agree I would change the test back to `utest` releasing a version for Scala Native `0.4.0` from my fork to bootstrap the test. When Mill is released I can update utest and move the test to use the official one from `com.lihaoyi`.
> What do you think?

@lolgab Makes sense. But I have no idea how long until a newer utest is released. Did you requested your changes for merge?

Let me restart Github tests first. If everything is green, we can merge to get a new snapshot release.

sideeffffect:
Build is green :tada:

lolgab:
I opened a PR [here](https://github.com/lihaoyi/utest/pull/229) but it is red because it needs this mill PR to work 😄.
Circular dependencies 🤷‍♂️.
If you merge this I can start updating utest, change it back and updating all the other libs that make the Scala Native ecosystem.

lefou:
Thank you!

PR #1116, opened 2021-01-20
scala 2.13.4 bump

lefou:
Depends on #1115

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -13,7 +13,7 @@ import mill.modules.Jvm.createAssembly
 object Deps {
 
   // The Scala version to use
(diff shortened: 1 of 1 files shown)


lefou:
CI is constantly failing. I haven't found a fix, but have already invested some hours on it. Also asking on sbt Gitter room provided no help, until we provide a better repro than our CI. I can't reproduce it locally, so am closing it.

Here is one pointer: https://github.com/scala/scala/pull/9174, a change possibly related. I also tried to bump zinc to 1.5.0-M3, as this is already on Scala 2.13.4, but no

PR #1120, opened 2021-01-21
Update utest to 0.7.6 on Scala Native tests

lolgab:


lefou (commented):
Can we bump `Deps.utest` instead and use it here, too?

lolgab:
@lefou Can a test share values with the build?
Can you point to an example? I have no idea on how to do this.

lefou:
My bad. My initial though was to bump all usages of utest in mill and then I must have overlooked  that your diff doesn't touch the `build.sc`. I'm going to merge this PR.

lefou:
Also, scala-steward already opened a PR for utest upgrade. #1119

PR #1133, opened 2021-01-29
Update bloop to 1.4.6-33-1c6f6712

lolgab:
- Remove Scala Native deprecations

### DIFF
--- a/contrib/bloop/test/src/mill/contrib/bloop/BloopTests.scala
+++ b/contrib/bloop/test/src/mill/contrib/bloop/BloopTests.scala
@@ -52,7 +52,7 @@ object BloopTests extends TestSuite {
     }
 
     object scalanativeModule extends scalanativelib.ScalaNativeModule with testBloop.Module {
-      override def scalaVersion = "2.11.12"
+      override def scalaVersion = "2.13.4"
       override def scalaNativeVersion = "0.4.0"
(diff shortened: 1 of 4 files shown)


lefou (commented):
Looks good to me.

lolgab:
@lefou Rebased on top of current master.

contrib/bloop/test/src/mill/contrib/bloop/BloopTests.scala
```diff
@@ -25,7 +25,7 @@ object BloopTests extends TestSuite {
 
     object scalaModule extends scalalib.ScalaModule with testBloop.Module {
       def scalaVersion = "2.12.8"
-      val bloopVersion = "1.4.6"
+      val bloopVersion = "1.4.6-33-1c6f6712"
```
lefou, line 28: If the intention is, to use the same version for testing as for building, then we should probably think about generating some `Versions` file in `build.sc` and use it here, as we do in other modules.

lolgab: ✅  Done!

lefou: Nice. I'm fine to merge after CI succeedes. Your PRs are a pleasure.

PR #1137, opened 2021-01-30
Fix linker used concurrently

lolgab:
After #1007 got merged, it generated an annoying bug: if you use `-j number` to use multiple parallel workers, multiple workers would try to use the same Scala.js linker, failing with: `java.lang.IllegalStateException: Linker used concurrently`.
This uses separate linkers per generated files so that doing `__.test` or `__.fastOpt` can't give the same linker to two different projects ( they generate different js files )


lefou:
Thanks!

PR #1139, opened 2021-02-03
Fix binary version computation of Scala nightly versions

alexarchambault:
The Scala nighly versions look like `2.13.5-bin-f956b11` for example, and can be downloaded from [here](https://scala-ci.typesafe.com/artifactory/scala-integration). AFAIU, their binary version should look like `X.Y`, just like stable versions, so that the binary version of `2.13.5-bin-f956b11` would be `2.13`. This PR fixes that.

This would be useful for https://github.com/lihaoyi/Ammonite/pull/1150, where using Scala 2 nightlies would allow to bump the Scala 3 version (from `3.0.0-M1` to `3.0.0-M3`).

I'm not really sure if / where I should add tests for that.

### DIFF
--- a/scalalib/api/src/ZincWorkerApi.scala
+++ b/scalalib/api/src/ZincWorkerApi.scala
@@ -86,12 +86,14 @@ object Util {
   val DottyVersion = raw"""0\.(\d+)\.(\d+).*""".r
   val Scala3Version = raw"""3\.(\d+)\.(\d+)-(\w+).*""".r
   val DottyNightlyVersion = raw"""(0|3)\.(\d+)\.(\d+)-bin-(.*)-NIGHTLY""".r
+  val NightlyVersion = raw"""(\d+)\.(\d+)\.(\d+)-bin-[a-f0-9]*""".r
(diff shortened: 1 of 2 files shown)


lefou (commented):
Looks reasonable. I wonder if we should limit it to a major of `2`. I can't point you right now to the best place for a test. I think we cover this only partially in intergration tests, but I think a simple unit test for `def scalaBinaryVersion` should do it.

smarter:
Note that for Scala 3 the binary version should always be 3: https://github.com/sbt/librarymanagement/pull/374/files

alexarchambault:
@smarter Even for the `RC`s?

smarter:
For all RCs that will happen after 3.0.0 is out yes (so 3.0.1-RC1 is _3, but 3.0.0-RC1 is _3.0.0-RC1), because they won't break binary compatibility (but if they break tasty compatibility we'll bump the tasty version number).

lefou (commented):
Change looks good to me. We need to fix the tests, though.

adpi2:
I am currently trying to fix the tests in https://github.com/com-lihaoyi/mill/pull/1259

lefou:
Superseeded by #1259 (contains the same commits).

PR #1140, opened 2021-02-03
Add missing trait in custom ZincWorkerModule example

alexarchambault:
While following the instructions to use a custom `ZincWorkerModule`, things didn't compile until `CoursierModule` was added too. (It seems it's a self type of `ZincWorkerModule` now.) This PR updates the documentation accordingly.


lefou:
You're right about the self type. Thank you!

PR #1141, opened 2021-02-04
Avoid withOptimize(false) on Scala Native until fixed

lolgab:


lefou:
@lolgab Do you expect a fix for this soon, e.g. in the next release? We might want to override this only for the exactly known buggy version. Does it make sense to provide an extra target in `ScalaNativeModule` to make this configurable for all users?

lolgab:
The sbt plugin makes it configurable and defaults to `true` if the `SCALANATIVE_OPTIMIZE` env variable is not defined, I didn't make it configurable because I followed what [mill does with Scala.js](https://app.circleci.com/pipelines/bitbucket/qamine/qa-automation-tests/4457/workflows/b7bdb437-c82e-4c7f-9907-2a809bfe09d4), it hides under fastOpt/fullOpt instead of making it configurable..
I don't really know what to do 😃
Sometimes I think that mill should recreate a copy of the Scala.js/Scala Native configuration to be friendlier to the people coming from Sbt.
At the same time I like how mill is configured.

lolgab:
> Do you expect a fix for this soon, e.g. in the next release? We might want to override this only for the exactly known buggy version.

At the moment there is only a version of Scala Native working with mill: `0.4.0`.
This setting is the default on Sbt, even on the debug mode, which adds very little ( < 1sec ) to the build time.
It's not a big deal anyway.

lefou (approved):
Ok. My initial idea was to just add a simple guard to the case `if scalaNativeVersion == "0.4.0"` but now I see that we haven't access to the version here and would require a greater change to make the version visible.  Doable. but probably not what you had in mind with this PR.

I'm still suspicious, that we need to configure this (and can't) for a specific version/situation. Guess we may expose this as a configuration option sometimes.

PR #1171, opened 2021-02-15
add play 2.8 support

peregin:
Adding Play 2.8 support. Tested with https://github.com/peregin/velocorner.com/blob/master/build.sc, was migrating the build from `sbt` to `mill`.


contrib/playlib/src/mill/playlib/RouteCompilerWorkerApi.scala
```diff
@@ -20,7 +20,7 @@ private[playlib] class RouteCompilerWorker {
         ctx.log.debug("Loading classes from\n"+toolsClassPath.mkString("\n"))
         val cl = mill.api.ClassLoader.create(
           toolsClassPath,
-          null,
```
lefou, line 23: This won't work and is most likely the cause of the test failure.  `null` as parent classloader is needed to separate the potential different binary-incompatible Scala versions of mill and the route compiler.

lefou (changes requested):
I just glanced at the diff and found one major issue, see below.

lefou:
Closing due to inactivity.

PR #1190, opened 2021-03-02
Add support for Scaladoc 3

jodersky:
This adds support for docJar generation under Scala 3.0.0-RC1, which is currently broken[1].

I've tried to keep backwards compatibility as much as I could. However, there is one sort-of-breaking change with this: `docJar` now has a dependency on the `compile` task (because Scaladoc 3 uses tasty files, not source files anymore). This dependency prevents generating documentation for projects that would previously not compile but could still generate documentation.

[1]: See the [discussion on Gitter](https://gitter.im/lampepfl/dotty?at=603e29b5823b6654d27ad28b) if you're interested, but the gist of it is that Scaladoc 3 is really not compatible with dottydoc, yet dottydoc is no longer available for Scala >= 3.0.0-RC1.

### DIFF
--- a/scalalib/api/src/ZincWorkerApi.scala
+++ b/scalalib/api/src/ZincWorkerApi.scala
@@ -53,6 +53,7 @@ case class CompilationResult(analysisFile: os.Path, classes: PathRef)
 object Util {
   def isDotty(scalaVersion: String) = scalaVersion.startsWith("0.")
   def isScala3(scalaVersion: String) = scalaVersion.startsWith("3.")
+  def isScala3Milestone(scalaVersion: String) = scalaVersion.startsWith("3.0.0-M")
   def isDottyOrScala3(scalaVersion: String) = isDotty(scalaVersion) || isScala3(scalaVersion)
 
   // eg, grepJar(classPath, name = "scala-library", versionPrefix = "2.13.")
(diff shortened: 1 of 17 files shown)


jodersky:
FYI: the integration test failure of BetterFiles seems unrelated. It currently also fails on my machine under mill master (c500ca986ab79af3ce59ba65a093146672092307).

lefou:
I'm pretty sure I review and commented before, but I can't find these comment anymore. Strange. I'll try to review soon again. Thanks for the PR.

lefou:
> I'm pretty sure I review and commented before, but I can't find these comment anymore. Strange. I'll try to review soon again. Thanks for the PR.

I'm sorry, I mixed up PRs. There is already a PR for scaladoc 3 https://github.com/com-lihaoyi/mill/pull/1154, on which I commented.

jodersky:
oh, no! How did I miss this? Let me also take a look at the other one

lefou (approved):
Looks good to me.

lefou:
I'm in favor merging this PR despite the fact that we have another older PR #1154 . This one handles doc resources better and comes with additional tests.

Unrelated to merging this one, I think we should better document all these different `isDotty`, `isScala3`, `isScala3Milestone` util method to better understand why and when to use them.

Thanks for this PR!

hmf:
I would like to know if static site generation was tested. The API is generated correctly, However the static site files are just copied over to the destination directory.  I may have an error on my side but just checking. 
TIA

PR #1197, opened 2021-03-06
Bump scala version: 2.13.3 -> 2.13.4 and 2.12.10 -> 2.12.13

lefou:


lefou:
Build error related: https://github.com/com-lihaoyi/mill/pull/1116#issuecomment-769397327

lefou:
Blocked by https://github.com/sbt/zinc/issues/974

lolgab:
@lefou The blocker should be solved now. Mill is already on zinc `1.5.3` that has your fix.

lefou:
> @lefou The blocker should be solved now. Mill is already on zinc `1.5.3` that has your fix.

Yeah, I'm very aware of that. But unfortuntely just buming zinc wasn't enough to bring mill to newer scala versions. And I wasn't able to get to the root cause (or maybe I just don't remember right now). There is more work to do, before we can support newer scala versions.

lefou:
I think I already sync'ed the branch #1185 and it still failed. It's probably related to support for older ScalaJS versions.

lolgab:
> I think I already sync'ed the branch #1185 and it still failed. It's probably related to support for older ScalaJS versions.

The Scala Steward PR is wrong because [it is trying to update to Scala `2.13.5` in the tests too](https://github.com/com-lihaoyi/mill/pull/1185/files#diff-224cd86b0846cfb4a2ba6382a000165d98156c849b0f667b725f837b666d14b6R27) which doesn't work, since there is no Scala.js release for Scala `2.13.5` /  Scala.js `1.4.0`.
I tried to run the Scala.js tests locally from this branch and I saw no problems.

lefou:
@lolgab Thanks for reviewing and pointing that out. I guess I lost the overview with all these bumps ;-).

I synchronized this PR with latest main. :crossed_fingers:

PR #1199, opened 2021-03-07
Enable Scala.js Scala 3 utest test

lolgab:


scalajslib/test/src/HelloJSWorldTests.scala
```diff
@@ -53,8 +53,9 @@ object HelloJSWorldTests extends TestSuite {
       extends BuildModule(crossScalaVersion, sjsVersion0, sjsUseECMA2015) {
       object test extends super.Tests with TestModule.Utest {
         override def sources = T.sources{ millSourcePath / 'src / 'utest }
+        val utestVersion = if(isScala3(crossScalaVersion)) "0.7.7" else "0.7.5"
```
lefou, line 56: Why do we need a different version here? Or asked differently, why can't we use 0.7.7 for all Scala versions?

lolgab: Because of Scala.js. utest `0.7.5` is based on Scala.js `1.0`, utest `0.7.7` is based on Scala.js `1.4` which is the first version providing Scala 3 support and can't be used on any Scala.js version < `1.4`. We could use different Scala.js versions for Scala 2 and Scala 3 on utest but I don't know if Scala.js introduced any bug-fixes on the scalac plugin side in the newer versions..

lefou: Makes sense. Thanks for clarifying. This is also the reason why all utest bumps in mill fail in CI.

PR #1230, opened 2021-03-22
Windows client-server fixes

sake92:
Only one thing in `ClientServerTests` doesn't work currently.  
Maybe someone will find a fix in the future... 😄 

I tried many things, even reimplementing the whole named pipes thing with JNA...  
But it boils down to `read/write` throwing an exception instead of returning `-1`...  
Since the read is done infinitely in a separate thread, it doesn't know when the output from server is done.  
So when a connection breaks, we should just ignore it, and not `System.exit(1);` as currently.

### DIFF
--- a/main/src/main/MillServerMain.scala
+++ b/main/src/main/MillServerMain.scala
@@ -77,19 +77,17 @@ class Server[T](lockBase: String,
   val originalStdout = System.out
   def run() = {
     Server.tryLockBlock(locks.processLock){
-      
-      val (serverSocket, socketClose) = if (Util.isWindows) {
-        val socketName = Util.WIN32_PIPE_PREFIX + new File(lockBase).getName
-        (new Win32NamedPipeServerSocket(socketName), () => new Win32NamedPipeSocket(socketName).close())
-      } else {
-        val socketName = lockBase + "/io"
-        new File(socketName).delete()
-        (new UnixDomainServerSocket(socketName), () => new UnixDomainSocket(socketName).close())
-      }
-      
       var running = true
       while (running) {
         Server.lockBlock(locks.serverLock){
+          val (serverSocket, socketClose) = if (Util.isWindows) {
+            val socketName = Util.WIN32_PIPE_PREFIX + "mill." + new File(lockBase).getName
+            (new Win32NamedPipeServerSocket(socketName), () => new Win32NamedPipeSocket(socketName).close())
+          } else {
+            val socketName = lockBase + "/io"
+            new File(socketName).delete()
+            (new UnixDomainServerSocket(socketName), () => new UnixDomainSocket(socketName).close())
+          }
 
           val sockOpt = Server.interruptWith(
             "MillSocketTimeoutInterruptThread",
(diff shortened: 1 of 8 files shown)


main/src/main/MillServerMain.scala
```diff
@@ -183,33 +183,26 @@ class Server[T](lockBase: String,
...
-
-    if (Util.isWindows) {
-      // Closing Win32NamedPipeSocket can often take ~5s
-      // It seems OK to exit the client early and subsequently
-      // start up mill client again (perhaps closing the server
-      // socket helps speed up the process).
-      val t = new Thread(() => clientSocket.close())
-      t.setDaemon(true)
-      t.start()
-    } else clientSocket.close()
```
lefou, line 195: Where de we close the client socket now?

lefou (commented):
Looks good to me. I added a question below. Does this PR really fix the server mode on Window, so that you no longer need the `--no-server` settings?

main/src/main/MillServerMain.scala
sake92: Good catch, thx!  
I reverted it now. Although it doesn't do anything on windows: https://github.com/sbt/ipcsocket/blob/v1.0.1/src/main/java/org/scalasbt/ipcsocket/Win32NamedPipeSocket.java#L97

An aside: the ipcsocket library we use doesn't 100% follow the tutorial on MSDN docs.  
I tried to make an alternative implementation here https://github.com/sake92/Win32NamedPipe  
But then I realized it works even with ipcsocket as long as we ignore the exception when closing the socket in client...

sake92:
> Looks good to me. I added a question below. Does this PR really fix the server mode on Window, so that you no longer need the `--no-server` settings?

It works 99% of the time.  
It throws this exception when you run mill on another project:
```
Exception in thread "main" java.io.IOException: Could not create lock for \\.\pipe\mill.ghrxuj6QppDw=-1_lock, error 5
        at org.scalasbt.ipcsocket.Win32NamedPipeServerSocket.<init>(Win32NamedPipeServerSocket.java:104)
```
That's because there should only be one locked pipe.  
But we give it the same pipe name, based on JDK we're running...  
Not sure how to improve on this one?

But after that it works fine. 😄 
This PR is a big improvement on status quo IMHO.

sake92:
@lefou I think we can take inspiration from [Gradle deamon docs](https://docs.gradle.org/current/userguide/gradle_daemon.html#sec:why_is_there_more_than_one_daemon_process_on_my_machine).  
It spawns a new deamon per **JDK+(JVM properties)**.  
That seems reasonable because not all projects use same props, maybe even in tests you want different props?

Then, it doesn't seem logical to lock on `processLock/serverLock` because JDK/daemon should be global for all its clients, not per-project... (thus the exception above)

lefou:
Thanks for this PR. I'll merge it. Nice to have mill server support on Windows.

@sake92, regarding the daemon: The mill daemon does not take multiple clients at once. Also I think we don't abstract the current working directory away in all cases, esp. when spawning sub-processes. So for having a shared daemon there is some more work to do...

PR #1222, opened 2021-03-18
Prepared documentation site to host multiple versions and external components

lefou:
This pull request is my attempt to improve the following aspects of our current documentation site:

* Better maintainability (by switching to AsciiDoc/Asciidoctor, which e.g. supports variables to easily update all references to the current version at once)
* Easier contribution (each page has a direct edit-link in the upper right corder)
* Multi-Version documentation (by using Antora as site generator)
* Potential to embed more external documentation, e.g. for plugins (by using Antora as site generator)

I create a gh-pages branch in my personal fork, which you can use as preview.

https://lefou.github.io/mill

This PR isn't ready for merge, but for review.

Links: 
* https://asciidoctor.org/
* https://antora.org/

PR #1234, opened 2021-03-24
Fix com lihaoyi urls

sake92:
Fix:
- new com-lihaoyi org URLs
- http -> https
- duplicate h1 tag on pages, very bad for SEO


lefou:
FYI, most of these issue should already be fixed via PR #1222 .

PR #1240, opened 2021-03-25
Add `CrossScalaVersionRanges` trait to mix into `Cross` instances

lolgab:
With this feature new sources directories will be generated
based on the set of values in `Cross[T]` to allow easier management
of projects with multiple Scala versions.
In particular there will be a `segment+` and `segment-` for every
version that correctly matches the version `<=` and `>=` based on
the numbers making the semantic version scheme.

Given:
```scala
object cross extends Cross[MyCrossModule]("2.11.12", "2.12.13", "2.13.4", "3.0.0-RC1")
class MyCrossModule(val crossScalaVersion: String) extends CrossScalaModule
```

Before this PR:
```
show > [2/2] cross[2.12.13].sources 
[
    "ref:2f52d18f:/Users/lorenzo/scala/cross/src",
    "ref:c984eca8:/Users/lorenzo/scala/cross/src-2.12.13",
    "ref:c984eca8:/Users/lorenzo/scala/cross/src-2.12",
    "ref:c984eca8:/Users/lorenzo/scala/cross/src-2"
]
```

After this PR:
```
show > [2/2] cross[2.12.13].sources 
[
    "ref:2f52d18f:/Users/lorenzo/cross/src",
    "ref:c984eca8:/Users/lorenzo/cross/src-2.12.13",
    "ref:c984eca8:/Users/lorenzo/cross/src-2.12",
    "ref:c984eca8:/Users/lorenzo/cross/src-2",
    "ref:c984eca8:/Users/lorenzo/cross/src-2+",
    "ref:c984eca8:/Users/lorenzo/cross/src-2.12.13+",
    "ref:c984eca8:/Users/lorenzo/cross/src-2.12+",
    "ref:c984eca8:/Users/lorenzo/cross/src-2.11.12+",
    "ref:c984eca8:/Users/lorenzo/cross/src-2.11+",
    "ref:c984eca8:/Users/lorenzo/cross/src-2.13.4-",
    "ref:c984eca8:/Users/lorenzo/cross/src-2.13-",
    "ref:c984eca8:/Users/lorenzo/cross/src-2.12.13-",
    "ref:c984eca8:/Users/lorenzo/cross/src-3.0-",
    "ref:c984eca8:/Users/lorenzo/cross/src-3-"
]
```


scalalib/src/CrossSbtModule.scala
```diff
@@ -1,24 +1,29 @@
 package mill.scalalib
 
+import mill.api.PathRef
 import mill.T
 import mill.scalalib.{CrossModuleBase, SbtModule}
 
 trait CrossSbtModule extends SbtModule with CrossModuleBase { outer =>
 
   override def sources = T.sources {
+    val crossScalaVersions = millOuterCtx.crossInstances.map(_.asInstanceOf[CrossScalaModule].crossScalaVersion)
```
lolgab, line 10: Here I would love to maintain the type of the `crossInstances` without them being of type `AnyRef`, but I couldn't find a way to achieve that.

scalalib/src/CrossModuleBase.scala
```diff
@@ -6,9 +6,8 @@ import mill.define.Cross.Resolver
 import mill.eval.PathRef
 
 object CrossModuleBase {
-  def scalaVersionPaths(scalaVersion: String, f: String => os.Path) = {
```
lolgab, line 9: Should I maintain the old function as deprecated for backwards binary compatibility?

lefou (commented):
To be honest, the outcome of supported source directories looks a bit hard to grasp. How often do you have the need for such sophisticated setup? Maybe, we can just merge the parts that make such a setup possible (cross factory), but leave the implementation to the user who really needs it. We could instead document it as extended use case. WDYT?

lolgab:
> To be honest, the outcome of supported source directories looks a bit hard to grasp. How often do you have the need for such sophisticated setup? Maybe, we can just merge the parts that make such a setup possible (cross factory), but leave the implementation to the user who really needs it. We could instead document it as extended use case. WDYT?

For my particular use case it would be fine. I would still leave the utility method `versionSpecificSources` (maybe with a different name) so people could use it. Then if this thing is used by many people we could even reiterate and add it as default to `CrossScalaModule` as I did here.

Do you think there is a way to maintain the type of the `crossInstances` as `Seq[T]`  instead of `Seq[AnyRef]`? I would love to make it typed but I don't see how

lefou:
> Do you think there is a way to maintain the type of the `crossInstances` as `Seq[T]` instead of `Seq[AnyRef]`? I would love to make it typed but I don't see how

The `Ctx` class is used in a broader scope. I can't see how we could transport the type of a specific cross module into it.

scalalib/src/CrossModuleBase.scala
lefou: I think we also break binary compatibility by the other changes too, but am not sure if that is an issue with current external plugins. I tend to keep this PR open a bit longer and merge before next major bump, if that is OK with you?

lolgab: It's totally fine either way. This particular question is not a problem anymore since you suggested to leave the `CrossScalaModule` untouched and leave only the utility methods + infrastructure for it to work instead of changing the default behaviour.

lolgab:
Whenever you decide that you want to trigger a major release ping me so I'll rebase this PR and make it mergeable.

scalalib/src/CrossModuleBase.scala
lefou: Fine. So we wait for next major (0.10.0) release. I hope we get some MiMa integration into it, too.

That said, if there are others who think we should integrate your initial change to `CrossScalaModule`, please speak up!

lolgab:
Rebased and added an integration test with version specific code (2.12- and 2.13+ folders).
It uncovered a bug in the `versionsSpecificSources` algorithm which I then solved.
Now I want to find the perfect naming for the added functions and check in the community if other people think it's worth it to add to `CrossScalaModule` and `CrossSbtModule` as I first did.

lefou:
@lolgab How about introducing a new trait, which only add version range source dirs? Then users can explicitly and easily opt-in complex compatibility source trees, whereas simple sources won't clutter, at least not in terms of meta-model.

lolgab:
> @lolgab How about introducing a new trait, which only add version range source dirs? Then users can explicitly and easily opt-in complex compatibility source trees, whereas simple sources won't clutter, at least not in terms of meta-model.

This is interesting. I could introduce two helper functions one for ranged versions and another for the current `CrossScalaVersions` behaviour. And have another trait only for the ranged versions. It's a cool idea :) Thanks!

lolgab:
> @lolgab How about introducing a new trait, which only add version range source dirs? Then users can explicitly and easily opt-in complex compatibility source trees, whereas simple sources won't clutter, at least not in terms of meta-model.

Tried the approach you suggested by creating a trait that can be mixed into `CrossScalaModule` or `CrossSbtModule` to add the version range specific code.

scalalib/src/CrossModuleBase.scala
lefou: > Should I maintain the old function as deprecated for backwards binary compatibility?

Sorry, for my late change of mind, but your initial question directed me to the question of binary compatibility, but this `def` is probably also used in users code, so yes, now I think we should still keep it around and `@deprecate` it with a good hint.

scalalib/src/CrossScalaRanges.scala
```diff
@@ -0,0 +1,14 @@
+package mill.scalalib
+
+import mill.api.PathRef
+import mill.scalalib.api.Util
+import mill.T
+
+trait CrossScalaRanges extends CrossModuleBase {
```
lefou, line 7: I'd love to see some documentation here.

lefou (commented):
I reviewed your latest changes and left two comments in place. Overall I think this is good to go. I'm still a bit undecided if your change to `Cross.Factory` is relevant for a minor version bump, binary compatibility-wise. I tend to cut a new `0.9.x` before merging this PR, so that those who need to stay a while have the newest mill. After merge, I could cut another release, probably a `0.10.0` so that you can start using this new feature. WDYT?

lefou:
@lolgab The mill main branch is now open for binary incompatible stuff, so I'd like to merge this PR, too. Could you please resolve the conflicts?

lolgab (commented):
Missing to reintroduce `scalaVersionPaths` method as deprecated, the other issues should be already addressed.

lefou (commented):
PR looks good to me. Only the re-introduction of `CrossModuleBase.scalaVersionPaths` marked as deprecated is open. Do you think this can be provided with a reasonable behavior? If not, I could also think of some stub that fails but gives the user a good hint what she needs to change.

lolgab:
> PR looks good to me. Only the re-introduction of `CrossModuleBase.scalaVersionPaths` marked as deprecated is open. Do you think this can be provided with a reasonable behavior? If not, I could also think of some stub that fails but gives the user a good hint what she needs to change.

@lefou Introduced back the `CrossModuleBase.scalaVersionPaths` method with the same behaviour as before. To do that (and to allow people to reuse the same logic for different versions than the `scalaVersion` (like `scalaJSVersion` or `scalaNativeVersion`)) I added a new method, that I named (without much creativity) `versionNumbers` that has the logic used by `CrossScalaModule`.

If you can propose a better name for it and for the companion `versionRanges` method I'd appreciate that.. This PR drained all my naming creativity and I can't really find a good name for anything now :D

lefou (commented):
> If you can propose a better name for it and for the companion versionRanges method I'd appreciate that.. This PR drained all my naming creativity and I can't really find a good name for anything now :D

What Scala developers (me included) sometimes forget: If neither the name nor the type transport the full meaning, there is also Scaladoc - a nice tool to add non-formal meaning to a type or a `def`. This is especially anticipated, when the `def` is meant to be used by others for the same purpose. ;-) - But seriously, I think adding some docs to the `versionsNumbers` and `versionRanges` `def`s will make the `Util` object more useful for others.

lolgab:
Good point! I'm going to add some documentation to both methods.
Names bother me particularly because you can't change them without breaking users code.
You can always add some scaladoc to existing apis without breaking anything.

lefou:
What about `compatibleVersion[Number]s` or `matchingVersion[Number]s`? I suck at naming things. too.

lolgab:
> What about `compatibleVersion[Number]s` or `matchingVersion[Number]s`? I suck at naming things. too.

Renamed `versionNumbers` to `matchingVersions`.

lefou (approved):
Very good. I hope people find this little gem when they need it. 
Thank you for your persistence!

lolgab:
> Thank you for your persistence!

Same to you for the many review rounds 🙏

PR #1241, opened 2021-03-27
Some subediting.

aholland:
- Capitalised Mill in text paragraphs.
 - Put in comments pointing out where double-underscore is used because it is not very clear visually that they are two separate characters.
 - Avoided the use of code style followed by "s" for `plural`s, which did not render well.
 - Changed "&" to "and" in text paragraphs.
 - Minor grammar fixes e.g. for concord.
 Let me know if this is welcome/unwelcome. I don't mind doing it as I read through docs and could do more if wanted. Thanks.

### DIFF
--- a/docs/pages/1 - Intro to Mill.md
+++ b/docs/pages/1 - Intro to Mill.md
@@ -60,11 +60,11 @@ Mill also works on a sh environment on Windows (e.g.,
 [WSL](https://docs.microsoft.com/en-us/windows/wsl); to get started, follow the instructions in the [manual](#manual)
 section below. Note that:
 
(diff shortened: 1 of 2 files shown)


lefou:
I'd rather wait for merge of #1222, which is almost ready, only release automation is to be done.

aholland:
Oh, absolutely! Those changes look great. Let's see if any of these are still relevant afterwards. Probably not.

lefou:
I think so. I've only applied some conversions and replaced some URLs by variables. With a bit of luck we 'll still get a usable diff.

lefou:
Superseded by #1445.

@aholland I tried to apply most of your edits to the docs. Feel free to review. Thank you!

PR #1245, opened 2021-03-30
Pass relative paths to gpg

alexarchambault:
On GitHub actions, this makes Sonatype publishing work from Windows workers (after a minimal setup such as [this script](https://github.com/coursier/apps/blob/f1d2bf568bf466a98569a85c3f23c5f3a8eb5360/.github/scripts/gpg-setup.sh)).


lefou:
Do you have any idea why the absolute URL wasn't working? Do you have an error message?

PR #1253, opened 2021-04-01
Read default Scala Native env variables

lolgab:
- Make `nativeOptimize` configurable
- Env variables are proper `T.input`s and are evaluated every time

PR #1258, opened 2021-04-07
Fix Scala Native env variables values to match sbt

lolgab:
- https://github.com/com-lihaoyi/mill/pull/1253 used `valueOf` which
  given a Java `enum` it matches the name of the classes and not
  the string values that match the official Scala Native env variables
  values. This PR uses instead `.values` which returns all the values
  then matches based on the `String value` the enums define.

### DIFF
--- a/scalanativelib/src/ScalaNativeModule.scala
+++ b/scalanativelib/src/ScalaNativeModule.scala
@@ -87,7 +87,7 @@ trait ScalaNativeModule extends ScalaModule { outer =>
 
   def logLevel: Target[NativeLogLevel] = T{ NativeLogLevel.Info }
 
-  private def releaseModeInput = T.input(
+  protected def releaseModeInput = T.input(
     sys.env.get("SCALANATIVE_MODE").map(v =>
       ReleaseMode
         .values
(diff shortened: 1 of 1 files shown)


lolgab:
My last PR had a problem I didn't notice. Sorry for that. Good that it didn't make in a stable release.
It was matching `SCALANATIVE_MODE=ReleaseFast` instead of `SCALANATIVE_MODE=release-fast` as sbt does.
I'm not sure about the phrasing of the error message. If you have a better idea, let me know.

scalanativelib/src/ScalaNativeModule.scala
```diff
@@ -87,7 +87,14 @@ trait ScalaNativeModule extends ScalaModule { outer =>
 
   def logLevel: Target[NativeLogLevel] = T{ NativeLogLevel.Info }
 
-  private def releaseModeInput = T.input(sys.env.get("SCALANATIVE_MODE").map(ReleaseMode.valueOf))
+  private def releaseModeInput = T.input(
```
lefou, line 90: I think making this protected is ok, e.g. to re-use it. Same for next change.

lefou (commented):
Thanks for following up with improvements.

scalanativelib/src/ScalaNativeModule.scala
lolgab: Done ✅ 
Made all the inputs I added protected.

lefou (approved):
Looks good to me. Thanks!

PR #1282, opened 2021-04-18
Fix TestModule.Utest framework class

lolgab:


lefou (approved):
Looks good to me. Thanks!

PR #1291, opened 2021-04-24
let bloop detect foreign modules included via moduleDeps

ajrnz:
Metals with VS Code uses the Bloop contrib module to extract the project setup. However when it does it only sees modules directly included in the build file. If you have a foreign-module (internally or externally) then it won't be picked up by bloop.

This has been bugging me for a while as it makes it difficult to modify external libraries without having a separate compile / publish cycle.

The PR fixes this problem by fully enumerating the list of `Modules` (`JavaModue` subclasses) using both the list of modules present in the build file but also the transitive `moduleDeps`.

It also creates better names for the bloop `out` directory for each module so that there is no danger of a clash with a foreign-module / sub-module name

### DIFF
--- a/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
@@ -94,10 +94,6 @@ class BloopImpl(ev: () => Evaluator, wd: Path) extends ExternalModule { outer =>
         T.log.info(s"Wrote $path")
         name(jm) -> PathRef(path)
       }
-
-      def writeTransitiveConfig = T {
-        Task.traverse(jm.transitiveModuleDeps)(_.bloop.writeConfig)
-      }
     }
 
     def asBloop: Option[Module] = jm match {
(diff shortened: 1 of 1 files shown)


contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
```diff
@@ -106,14 +102,29 @@ class BloopImpl(ev: () => Evaluator, wd: Path) extends ExternalModule { outer =>
     }
   }
 
+  // Compute all transitive modules from build children and via moduleDeps
+  def transitiveModules(mod: define.Module, found: Seq[define.Module] = Seq.empty): Seq[define.Module] = {
```
lefou, line 106: Note: The explicit `found` parameter looks a bit like the intention was to make this `def` tail recursive, but it isn't (yet).

lefou (approved):
Looks good to me.

I think the test suite should cover a foreign module too. It's not an requirement, but would help to keep bloop module working, as we have no maintenance guarantees on contrib modules besides "tests should pass". ;-)

contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
ajrnz: Actually I wasn't aiming for tail recursion. I would hope no one would have enough modules to require it. I just wanted to avoid infinite recursion when two modules are declared co-dependent.

lefou (commented):
I think the current transitive module resolution still misses compile-time only module dependencies (defined via `compileModuleDeps`).

ajrnz:
Thanks for that. Do you want me to add `compileModuleDeps` to the dependency list? I've never used that one before but I suppose it makes sense to add it.

lefou:
> Thanks for that. Do you want me to add `compileModuleDeps` to the dependency list? I've never used that one before but I suppose it makes sense to add it.

Thanky for your contribution! Yeah, I think it would be nice to support compileModuleDeps, too. ;-)

PR #1301, opened 2021-04-29
strip whitespace from version file

### LINKED ISSUE #1294: VersionFileModule strip whitespace from versionFile?
https://github.com/com-lihaoyi/mill/blob/main/contrib/versionfile/src/mill/contrib/versionfile/VersionFileModule.scala#L10

`echo 0.1.0-SNAPSHOT > versionFile/version` or editor will add `\n` at end of file，`Version.of` match will fail due to `\n`

```
[1/1] show
[1/1] show > [2/2] versionFile.currentVersion
1 targets failed
show 1 targets failed
versionFile.currentVersion scala.MatchError: 0.1.0-SNAPSHOT
 (of class java.lang.String)
    mill.contrib.versionfile.Version$.of(Version.scala:51)
    mill.contrib.versionfile.VersionFileModule.$anonfun$currentVersion$2(VersionFileModule.scala:10)
    mill.define.Task$MappedDest.evaluate(Task.scala:391)
```

why not strip whitespace ?

camper42:
strip whitespace to avoid MatchError in `Version.of` when version file ends with newline character.

This fix #1294

### DIFF
--- a/contrib/versionfile/src/mill/contrib/versionfile/VersionFileModule.scala
+++ b/contrib/versionfile/src/mill/contrib/versionfile/VersionFileModule.scala
@@ -7,7 +7,7 @@ trait VersionFileModule extends Module {
   /** The file containing the current version. */
   def versionFile: define.Source = T.source(millSourcePath / "version")
   /** The current version. */
-  def currentVersion: T[Version] = T { Version.of(os.read(versionFile().path)) }
+  def currentVersion: T[Version] = T { Version.of(os.read(versionFile().path).strip) }
   /** The release version. */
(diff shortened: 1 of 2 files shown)


lefou (commented):
Adding a unit test would be nice.

camper42:
> Adding a unit test would be nice.

ok, I'll add a test to check if it work on version file has newline.

camper42:
`strip` method added since java 11, so java 8 tests failed, I'll use `trim` to replace.

I am not familiar with utest, so it takes a little time to properly add a test

lefou (commented):
Looks good to me. Thank you for this improvement.

PR #1302, opened 2021-04-29
Initialize mill project command

HSalaila:
Given the discussion on https://github.com/com-lihaoyi/mill/discussions/1184, this is an attempt to support skeleton mill project generation via --init flag. I have also changed the default behavior when no flags are given. It now shows the help screen.

There is a lot to improve like specifying templates like g8. Please let me know what I can do to further improve this functionality and some guidance would be appreciated

### DIFF
--- a/main/src/MillMain.scala
+++ b/main/src/MillMain.scala
@@ -11,6 +11,8 @@ import mill.api.DummyInputStream
 
 case class MillConfig(
   ammoniteCore: ammonite.main.Config.Core,
+  @arg(doc = """Initialize a Mill project in current directory""")
+  init: Flag,
   @arg(doc = """Run Mill in interactive mode and start a build REPL. In this mode, no
     mill server will be used. Must be the first argument.""")
   repl: Flag,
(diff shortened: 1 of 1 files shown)


lefou (commented):
I'd prefer to implement it as a command instead of a command line option. But I haven't checked if that is even possible when there is no `build.sc`. But that should be fixable.

lefou:
> I'd prefer to implement it as a command instead of a command line option. But I haven't checked if that is even possible when there is no `build.sc`. But that should be fixable.

I'm sorry, if this was not clean enough what I was meaning with "command".
Mill knows the special command tasks (https://com-lihaoyi.github.io/mill/mill/Tasks.html#_commands). Most built-in commands of mill are defined in `MainModule` (https://github.com/com-lihaoyi/mill/blob/main/main/src/main/MainModule.scala), e.g. the `show` command (https://com-lihaoyi.github.io/mill/mill/Intro_to_Mill.html#_show).

So, there should be no need to handle command line parameters. Just add parameters and scaladoc to your `def init(...): Command[Unit]` and you're done.

sake92:
@HSalaila could you try giter8 integration in latest snapshot? See #1874 for ref.  
E.g. `mill -i init sake92/mill-scala-hello.g8`

lefou:
Closing in favor of #1874

Issue #1284, opened 2021-04-18
Windows Mill: control characters being outputted in PowerShell & Command Prompt - cannot read easily.


ScalaWilliam:
Hi, I am running Mill on Windows and this is what I get when running it:
![image](https://user-images.githubusercontent.com/2464813/115154534-38d37f00-a073-11eb-91dc-4fa088e48747.png)

Expectation: either display colors without control characters, or display no colors, so that the output is readable.

Running in Linux it works just fine (WSL).

Thanks!

alexarchambault:
@ScalaWilliam Does it work when running mill via coursier? Like
```text
$ cs resolve mill:0.9.5 --scala 2.13.3 -- ..mill-args..
$ cs resolve mill-interactive:0.9.5 --scala 2.13.3 -- ..mill-args..
```

coursier enables ANSI output in the terminal, via [this library](https://github.com/alexarchambault/windows-ansi). If that fixes it, it could be interesting for mill to do the same.

lefou:
@ScalaWilliam As a workaround, you can disable colored output with `--color false`.

ScalaWilliam:
> @ScalaWilliam Does it work when running mill via coursier? Like
> 
> ```
> $ cs resolve mill:0.9.5 --scala 2.13.3 -- ..mill-args..
> $ cs resolve mill-interactive:0.9.5 --scala 2.13.3 -- ..mill-args..
> ```
> 
> coursier enables ANSI output in the terminal, via [this library](https://github.com/alexarchambault/windows-ansi). If that fixes it, it could be interesting for mill to do the same.

That didn't work - did you mean the below for example?
```
> cs.exe launch mill-interactive:0.9.5 -- bsp.test
```

In which case the colour bug was not there.
However, it would hang (I tried 0.9.6 as well) on this stage: `[5/59] mill.scalalib.ZincWorkerModule.classpath`. I tried it both on `mill` and on my own mini test project.



> @ScalaWilliam As a workaround, you can disable colored output with `--color false`.

@lefou this seems to not have worked.
```
> mill --color false foo.test
[4/59] foo.test.resources
←[1A←[2K←[9999D[8/59] foo.sources
←[1A←[2K←[9999D[13/59] foo.resources
←[1A←[2K←[9999D[32/59] foo.test.sources
...
```

lefou:
> @lefou this seems to not have worked.
> 
> ```
> > mill --color false foo.test
> [4/59] foo.test.resources
[...]

PR #1308, opened 2021-05-02
Enable ANSI output on Windows upon start-up

alexarchambault:
This fixes https://github.com/com-lihaoyi/mill/issues/1284 by enabling ["console virtual terminal sequences"](https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences) in the Windows console, which are basically ANSI escape codes. This is done using some JNI bindings of jline, via [windows-ansi](https://github.com/alexarchambault/windows-ansi).

### DIFF
--- a/main/src/MillMain.scala
+++ b/main/src/MillMain.scala
@@ -4,6 +4,7 @@ import java.io.{InputStream, PrintStream}
 import java.util.Locale
 
 import scala.jdk.CollectionConverters._
+import scala.util.Properties
 import io.github.retronym.java9rtexport.Export
 import mainargs.{Flag, Leftover, arg}
 import mill.eval.Evaluator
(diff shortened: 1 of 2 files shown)


main/src/MillMain.scala
```diff
@@ -71,6 +72,10 @@ case class MillConfig(
 object MillMain {
 
   def main(args: Array[String]): Unit = {
+
+    if (Properties.isWin && System.console() != null)
+      io.github.alexarchambault.windowsansi.WindowsAnsi.setup()
```
lefou, line 77: Where does this packages comes from. Assuming it is a transitive dependency of coursier, we should make it explicit in mill's dependencies, too.

Is it ok to call the `setup()` method multiple times? Is there some _cleanup_ needed?

alexarchambault: Sure, I can add it explicitly. It's brought by `io.get-coursier::coursier-cache`.

`setup()` can be called multiple times, yes. It'll just return `false` starting from the second call. The only requirement is to only call it on Windows (JNI will complain else), when `System.console() != null` (or it'll throw an `IOException`).

lefou (approved):
Looks good to me. Thank you!

PR #1314, opened 2021-05-04
Update Intro_to_Mill.adoc

tkawachi:
Fix permission denied to redirect curl output.

In the original code, curl is run with sudo, but the file writing is done with user privileges.


lefou:
Thank you!

PR #1324, opened 2021-05-17
Play 2.8 support

lolgab:
Rebase and fix of #1171


lefou (commented):
Look good to me so far. Thank you!

Are there any tests for Play 2.8? I don't remember if we use the same for all play versions or if we need specific ones for each version, in which case I miss any in this PR.

lolgab:
Ok let me check if we have tests for all the versions!

contrib/playlib/test/src/mill/playlib/PlayTestsCommon.scala
```diff
@@ -0,0 +1,25 @@
+package mill.playlib
+
+import mill.util.{TestEvaluator, TestUtil}
+import utest.framework.TestPath
+
+trait PlayTestSuite {
```
lefou, line 6: I'd like to have filename and content in sync. Please rename the file to `PlayTestSuite.scala`.

lefou (commented):
Thanks for overhauling the play module test suite. Looks reasonable. One file name change would be nice, but otherwise: Great job!

contrib/playlib/test/src/mill/playlib/PlayTestsCommon.scala
lolgab: Ops, this was originally `object PlayTestsCommon` 😄
I'm renaming it!

lolgab:
Renamed the file and updated the syntax of one test that I forgot to update before.

lolgab:
Can this be merged now @lefou?

lefou (approved):
Looks good to me. Thank you!

PR #1285, opened 2021-04-18
Fix #1283, #1021: Windows BSP: request failed: buildTarget/dependencySources

ScalaWilliam:
… - cannot import into IDE

There is a PR already (#1092) but I am wary of dealing with Strings directly. Here is my approach.

On Windows, File URLs follow an peculiar representation in Java

```
scala> java.nio.file.Paths.get(".").toAbsolutePath.toUri.toURL
java.net.URL = file:/C:/Users/Developer/mill/./
```

From this, we wish to get a `Path` back, and the way to do this is:

```
scala> java.nio.file.Paths.get((java.nio.file.Paths.get(".").toAbsolutePath.toUri.toURL).toURI)
java.nio.file.Path = C:\Users\Developer\mill\.
```

Unit testing for this is more challenging because the `WindowsFileSystem` instance is a `sun.nio.fs` package, rather than a standard package.

The solution here, compared to the previous code, is to reduce the number of conversions; the key loss happens when you do `(URL).getFile`.

```
scala> java.nio.file.Paths.get(".").toAbsolutePath.toUri.toURL
java.net.URL = file:/C:/Users/Developer/mill/./
scala> java.nio.file.Paths.get(".").toAbsolutePath.toUri.toURL.getFile
String = /C:/Users/Developer/mill/./
```

**Other notes:**

I enabled edits by maintainers, and as such please feel free to modify.

**scalafmt**: I was not sure how to apply Scalafmt from Mill as I am very new to this, and my IntelliJ didn't have the option for this imported build to do a scalafmt.

**Testing**: Like #1092, I think unit testing this is very challenging. Are there instructions as to how you could test this, or are you able to do a test of this PR?

I am very pressed for time at the moment, but wanted to make a contribution however my knowledge on this tool is still very limited.

I tried `bsp.test` task, and then import the produced file however this did not work as expected (neither did the original version of mill), where the `mill.json` that is produced looked like:
```
{"name":"mill-bsp","argv":["C:\\Path\\to\\jar1.jar;C:\\Path\\to\jar2.jar...", "-i", ...] }
```

And when importing, it would say that the whole argument is not a file that can be found (rather than the individual files?)

PR #1325, opened 2021-05-18
Fix bsp: getMillBuildClasspath return URI string

camper42:
getMillBuildClasspath is expect to return URI, otherwise IDEA bsp import will have
`java.lang.IllegalArgumentException: Missing scheme` error on MacOS and
`Java.net.URISyntaxException: Illegal character in opaque part at index 2: ...` on Windows


camper42:
problem I mentioned at https://github.com/com-lihaoyi/mill/pull/1285#issuecomment-842305505
seems caused by https://github.com/com-lihaoyi/mill/pull/1285/files#diff-a8dcb229dc6fe3ad91d4667a94174520d2d13cf522738fed08023105ac955d00R165

`Path.wrapped` return `java.nio.file.Path` and `toUri` is missing

@lefou @ScalaWilliam pls review

bsp/src/mill/bsp/ModuleUtils.scala
```diff
@@ -162,7 +162,7 @@ object ModuleUtils {
     val binarySource =
       if (sources) all.filter(url => isPathSourceJar(url))
       else all.filter(url => !isPathSourceJar(url))
-    binarySource.filter(path => exists(path)).map(_.wrapped.toString)
+    binarySource.filter(path => exists(path)).map(_.toNIO.toUri.toString)
```
lefou, line 165: I'm wondering why we don't use a better return type here. At least, we should add some Scaladoc to line 126 to make it clear, that the returned `String`s are meant to be URIs.

lefou (commented):
As it seems this code isn't covered by any tests, I'd like to have at least one written "It works for me" from someone who really tested this locally, for each platform.

camper42:
> As it seems this code isn't covered by any tests, I'd like to have at least one written "It works for me" from someone who really tested this locally, for each platform.

will try to add an ut for at least getMillBuildClasspath first, If it seems to take a long time to do, I'm happy to write "It works for me" to get this fix merged.

ScalaWilliam:
My apologies, I did not get a chance to test this.

camper42:
hmmm，getMillBuildClasspath on `TestEvaluator.static(TestModule).evaluator` return empty list
so test always pass....

camper42:
> getMillBuildClasspath on TestEvaluator.static(TestModule).evaluator return empty list

test is running on a version like `0.9.7-4-c85970-DIRTY55902d6c`, which can't access before `ci/publish-local.sh`, so:
* `classpath` in `getMillBuildClasspath` is empty due to `Try(evaluator.rootModule.getClass.getClassLoader.asInstanceOf[SpecialClassLoader])` failed
* `millJars` in `getMillBuildClasspath` is empty due to  `Resolution failed`

It's impossiable to test, choose to leave a "It works for me on MacOS IDEA" comment...

camper42:
also worked for my Windows IDEA

`test-windows (8, cmd /C %GITHUB_WORKSPACE%\ci\mill.bat -i -d -k "{main,scalajslib,bsp}.__.test")` failure can be ignored?
nothing change about `mill.scalajslib`

lefou (commented):
The CI failure is some flakiness in GitHub Actions, I reran the tests, which are now green.

I think the test issue(s) can be solved by propagating the freshly built jars into the test setup via `forkEnv` and/or `forkArgs`. We do this in other integration tests and it works. But I think it's too much to ask for in a bug fix PR and your "works for me" is better than what we have currently. We have lots of open building sites in our BSP story, and integration testing is a central part of it. 

Your technical change looks good to me, but the Scaladoc needs to be moved before the start of the `def` on line 126, not inside the `def`. Otherwise, we're good to merge.

fabianhjr:
Had same issue on Linux (NixOS), ( `java.lang.IllegalArgumentException: Missing scheme`)

Tested this PR by:

1. Checking out this repo
2. Checking out 0.9.7
3. Cherrypicking this change
4. Retagging 0.9.7 to include this change
5. hand modifying a local project  `.bsp/mill.json` to point to a symbolic link to the run script (`/home/fabian/.local/bin/mill-checkout -> /home/fabian/Development/Open/mill/out/dev/launcher/dest/run`)
6. Restarting IntelliJ IDEA

This did fix the error and I was able to import the project again; local retagging was to avoid attempting to fetch nonexistant versions of mill-contrib-{bloop, scoverage} (0.9.7-[n]-[local commit that isn't available elsewhere]

camper42:
@fabianhjr would you mind if I add "works for @fabianhjr on Linux(NixOS)" ?

fabianhjr:
Not at all, go ahead.

Many thanks for the diagnosis and fix of this issue! UwU

lefou:
Thanks you!

sequencer:
When can we get this patch in a new version like 0.9.8? Since 0.9.7 of mill+bsp+IDEA is totally broken since this issue.

lefou:
You can start to use it right away. Just pick the latest snapshot-build version from https://github.com/com-lihaoyi/mill/releases (e.g. `0.9.7-13-bf96a1`) and put it into `.mill-version` in your project top directory. See https://com-lihaoyi.github.io/mill/mill/Intro_to_Mill.html#_overriding_mill_versions

lefou:
I'd like to get some other fixes into main branch before the next tag.

sequencer:
Got it! Thank you!
I'll use `0.9.7-13-bf96a1` fix this temporarily, since I'm using mill package from Arch Linux.

sequencer:
opps, seems that package from Arch Linux doesn't honor `0.9.7-13-bf96a1`

PR #1330, opened 2021-05-20
play-contrib: Fixed classloading of worker module and cleanups

lefou:
Fixed use of mills utility classloader by settings the correct parent classloader.
Using the classloader of the module as parent makes the shared API modules (which need to be already loaded by the mill build) visible.
The previously used default was using the classloader which loaded the mill build (and the ammonite script), which misses the play API module (loaded via `$ivy` import).

Fixed dependency resolution of worker classpath.
The collected classpath of the worker module was limited to only use the worker module.
But this misses all transitive dependencies of the worker.
This is probably because of older limitiations in mill builds, which also included the mill API itself.
Now, mill uses `compileModuleDeps` for the mill API and we have clean and usable transitive module and worker classpaths.

### DIFF
--- a/contrib/playlib/src/mill/playlib/RouterModule.scala
+++ b/contrib/playlib/src/mill/playlib/RouterModule.scala
@@ -79,7 +79,6 @@ trait RouterModule extends ScalaModule with Version {
       workerKey,
       s"mill-contrib-playlib-worker-${playMinorVersion()}",
       repositoriesTask(),
-      resolveFilter = _.toString.contains("mill-contrib-playlib-worker"),
       artifactSuffix = playMinorVersion() match {
         case "2.6" => "_2.12"
         case _ => "_2.13"
--- a/contrib/playlib/src/mill/playlib/RouteCompilerWorkerApi.scala
+++ b/contrib/playlib/src/mill/playlib/RouteCompilerWorkerApi.scala
@@ -21,6 +21,7 @@ private[playlib] class RouteCompilerWorker {
         val cl = mill.api.ClassLoader.create(
           toolsClassPath,
           null,
+          sharedLoader = getClass().getClassLoader(),
           sharedPrefixes = Seq("mill.playlib.api.")
         )
         val bridge = cl


contrib/playlib/src/mill/playlib/RouteCompilerWorker.scala
```diff
@@ -1,15 +1,15 @@
...
 import mill.playlib.api.RouteCompilerType
 import mill.scalalib.api.CompilationResult
 
 private[playlib] class RouteCompilerWorker {
+
   private var routeCompilerInstanceCache = Option.empty[(Long, mill.playlib.api.RouteCompilerWorkerApi)]
 
-  private def bridge(toolsClasspath: Agg[os.Path])
+  protected def bridge(toolsClasspath: Agg[os.Path])
                     (implicit ctx: Ctx) = {
```
lolgab, line 13: The alignment went a bit off here.

contrib/playlib/src/mill/playlib/RouterModule.scala
```diff
@@ -72,22 +74,22 @@ trait RouterModule extends ScalaModule with Version {
...
-  private def playRouteCompilerWorkerClasspath = T {
+  def playRouteCompilerWorkerClasspath = T {
     val workerKey = "MILL_CONTRIB_PLAYLIB_ROUTECOMPILER_WORKER_" + playMinorVersion().replace(".", "_")
 
     mill.modules.Util.millProjectModule(
       workerKey,
       s"mill-contrib-playlib-worker-${playMinorVersion()}",
       repositoriesTask(),
-      resolveFilter = _.toString.contains("mill-contrib-playlib-worker"),
+//      resolveFilter = _.toString.contains("mill-contrib-playlib-worker"),
```
lolgab, line 84: Should you delete this line?

lefou: Or course. Fixed.

contrib/playlib/src/mill/playlib/RouteCompilerWorker.scala
lefou: Fixed.

lolgab:
As a person that doesn't know a lot about Mill internals (but a bit more than the average user), I struggled a lot to find which one of you changes actually fixed the classloader bug in the `playlib` module.
I now understood how dangerous is `getClass.getClassloader`.. It's output radically changes when you move it to a different class!
I think you should split your work here in multiple PRs to address the different issues. It serves as good documentation for people in the future that would face similar issues.
Trying to apply your changes one by one on current master, I found this as the minimal diff to fix the bug, without doing any refactoring:
```diff
diff --git a/contrib/playlib/src/mill/playlib/RouteCompilerWorkerApi.scala b/contrib/playlib/src/mill/playlib/RouteCompilerWorkerApi.scala
index 3b632a68..5b2cfe06 100644
--- a/contrib/playlib/src/mill/playlib/RouteCompilerWorkerApi.scala
+++ b/contrib/playlib/src/mill/playlib/RouteCompilerWorkerApi.scala
@@ -21,6 +21,7 @@ private[playlib] class RouteCompilerWorker {
         val cl = mill.api.ClassLoader.create(
           toolsClassPath,
           null,
+          sharedLoader = getClass().getClassLoader(),
           sharedPrefixes = Seq("mill.playlib.api.")
         )
         val bridge = cl
diff --git a/contrib/playlib/src/mill/playlib/RouterModule.scala b/contrib/playlib/src/mill/playlib/RouterModule.scala
index 4c0efd6b..4be78bf8 100644
--- a/contrib/playlib/src/mill/playlib/RouterModule.scala
+++ b/contrib/playlib/src/mill/playlib/RouterModule.scala
@@ -79,7 +79,6 @@ trait RouterModule extends ScalaModule with Version {
       workerKey,
       s"mill-contrib-playlib-worker-${playMinorVersion()}",
       repositoriesTask(),
-      resolveFilter = _.toString.contains("mill-contrib-playlib-worker"),
       artifactSuffix = playMinorVersion() match {
         case "2.6" => "_2.12"
         case _ => "_2.13"
```

Can we have a PR with just this diff to solve the bug first?

lefou:
Yeah, makes sense.

lefou:
I amended my commit and will move other improvments to a new PR.
@lolgab Would be nice if you could validate, this really fixes the issue.

lolgab:
I'm going to do it!
Can you refactor the the PR description too and move the other parts to the new PR?

lefou:
> Can you refactor the the PR description too and move the other parts to the new PR?

Already done. :-D

lolgab:
> Already done. :-D

Yeah, was just editing the comment :D You're a machine!

lolgab:
I confirm it works!
I'm asking myself if it's possible to make an integration test fail in cases like this. Why our tests pass in first place?

lolgab (approved):
LGTM! 👍

lefou:
> I'm asking myself if it's possible to make an integration test fail in cases like this. Why our tests pass in first place?

That's because we don't use mills (ammonites) loading mechanism but just run the tests in the test classloader. This is a limitation but speeds up mills integration tests a lot.

But it's one of the reasons I prefer mill plugins in their own repos and integration test them with https://github.com/lefou/mill-integrationtest plugin (shameless plug). That's slower but catches such kind of problems. Also, you can better test compatibility to different mill versions.

PR #1332, opened 2021-05-21
play-contrib: Various improvents

lefou:
Moved and renamed some classes for clarity.

Fixed visibility of various targets.

Consolidated declared packages.

Made the use of the `RouteCompileWorkerModule` (as ExternalModule) customizable.

Introduced new `logger` parameter in `mill.api.Classloader.create`.

Removed code duplication in Play-version specific workers.
But kept the setup prepared to easily un-share it for future Play versions.


main/api/src/mill/api/ClassLoader.scala
```diff
@@ -8,25 +8,43 @@ import io.github.retronym.java9rtexport.Export
...
+      urls = urls,
+      parent = parent,
+      sharedLoader = sharedLoader,
+      sharedPrefixes = sharedPrefixes,
+      logger = None)
+  }
+
   def create(urls: Seq[URL],
              parent: java.lang.ClassLoader,
              sharedLoader: java.lang.ClassLoader = getClass.getClassLoader,
```
lolgab, line 30: Since this default argument is the dangerous part, should this be deprecated too?
I expect the deprecated method have this default value so the old code use the deprecated method and get the deprecation notice. Since you're defaulting the logger to `None` I think the old code is using this method directly.
Can you explain what's going on here?

lefou: The changed method is source-compatible to the old one, because the new argument (the logger) has a default value. But it's not binary compatible, so we need an overload with the old signature. But in Scala only one overload can have default arguments, that's why the deprecated method no longer has default arguments, but the new one.

The default value for the `sharedLoader` is ok as long as you don't need to share classes not already visible to mill core. This is true for in most cases (e.g. sbt-test-interface, jna, scala{_,js,native} worker api), but not in case of playlib.

The motivation to change this method was solely the wish to log the shared classloading when using mill with `--debug` option.

lefou: But you're probable right, we should take the opportunity to remove the default for the classloader, as even if it works now, it may hinder us later, when we want to change packaging, embedding or JVM.

lolgab: Yeah, it is a big footgun and can't be checked by our tests which is even scarier. But removing it would change the behaviour of other modules using it, unless you expose a deprecated `classloader` here for the other modules to consume.
As usual, a decision important enough to be tackled in a separate PR; as you said the deprecation here is only about the logger.

lolgab (approved):
LGTM! 👍

Issue #971, opened 2020-10-05
Piggyback on ammonite's BSP implementation for build files 


Baccata:
The BSP implementation (which allows editors to talk to mill when the user edits module files) is becoming [first-class-citizen in mill](https://github.com/lihaoyi/mill/pull/969), it'd be interesting to piggyback on ammonite's BSP implementation for `build.sc`, to get all the autocompletion/navigation features metals users would like to have when editing mill build files. 

#### A few possibly relevant details 

Mill's entrypoint that ties it to ammonite is [there](https://github.com/lihaoyi/mill/blob/ef71368bbd4f71ba3f22ac3da51533bacda3b197/main/src/main/RunScript.scala#L26). This method outputs a "Evaluator", which is a construct that is aware of the build-graph and able to evaluate mill-tasks. It is is used [by the BSP implementation](https://github.com/lihaoyi/mill/blob/master/contrib/bsp/src/mill/contrib/bsp/MillBuildServer.scala#L21). The evaluator potentially be accompanied by the information required by ammonite's bsp, in order to enrich mill's bsp for build-files. 

One essential thing, specific to mill, that differs from ammonite is the [codeWrapper]( https://github.com/lihaoyi/mill/blob/c9f8badca307d2c271c5fafa677de648193ba3d7/main/src/main/MainRunner.scala#L162), which would have to be communicated to the ammonite's bsp somehow. The classpath is also obviously different. 

@alexarchambault 👋 , could I ask you for a brain-dump when you have time, on how such an integration would be possible ?

lefou:
This is likely superseded by 

* https://github.com/com-lihaoyi/mill/issues/1675
* https://github.com/com-lihaoyi/mill/issues/1972

PR #1349, opened 2021-05-28
fix sourceItem for mill-build

### LINKED ISSUE #1159: The BSP server returns wrong sources for mill-build target
The response for `buildTarget/sources` returns the following item for the `mill-build` target:
```
{
  "target": {
    "uri": "file:///Users/heksesang/Source/test-project/"
  },
  "sources": [
    {
      "uri": "file:/Users/heksesang/Source/test-project/src/",
      "kind": 2,
      "generated": false
    }
  ]
}
```

This means that `build.sc` is not correctly picked up as a source for that target (you have to copy it into `/src` to get IDE support, which is a tad tedious).

camper42:
due to hardcode `sourceItem(evaluator.rootModule.millSourcePath / "src", generated = false)`,
Metals not recognize build.sc as source file when user choose mill-bsp build server,
no auto-complete and hover tooltips available

This commit fix #1159

### DIFF
--- a/bsp/src/mill/bsp/MillBuildServer.scala
+++ b/bsp/src/mill/bsp/MillBuildServer.scala
@@ -87,7 +87,7 @@ class MillBuildServer(evaluator: Evaluator, bspVersion: String, serverVersion: S
           if (targetId == millBuildTargetId)
             new SourcesItem(
               targetId,
-              Seq(sourceItem(evaluator.rootModule.millSourcePath / "src", generated = false)).asJava // Intellij needs one
+              Seq(sourceItem(evaluator.rootModule.millSourcePath / "build.sc", generated = false)).asJava // Intellij needs one
             )
           else {
             val module = getModule(targetId, modules)


heksesang:
> This hardcodes the build file, which is ok, but will miss any included file.
> 
> E.g. in mill we have some additional files:
> 
> ```scala
> import $file.ci.shared
> import $file.ci.upload
> ```
> 
> Those will still be missing.

Won't those be imported by regular ammonite support?

lefou:
AFAIK, mill BSP server does not provide any "regular" ammonite support (yet, see #971). So, either you mean the native support of your editor, or some separately started ammonite BSP server. But neither will have any knowledge of the right classpath to be used.

heksesang:
I was thinking of the editor support. But I forgot about that point, that if the separate files depend on the classpath as well, that support won't be enough.

Does Ammonite have anything that lets us obtain a list of `$file`-references in a script?

lefou:
I don't know Ammonite well enough, but I think the answer is yet. Ammonite itself seems to provide BSP support, to the best would be to look into that implementation.

Back to this concrete PR, as `src` in in almost all cases not the right setting, but `build.sc` is probably too restricted. Have you tried to just set the source path of `mill-build` root project to `.` or `millSourcePath`?

camper42:
> Have you tried to just set the source path of mill-build root project to . or millSourcePath?

set source path to `.`  or `millSourcePath` may be inappropriate too

https://build-server-protocol.github.io/docs/specification.html#build-target-sources-request

```
  /** Either a text document or a directory. A directory entry must end with a forward
   * slash "/" and a directory entry implies that every nested text document within the
   * directory belongs to this source item.
   */
```

heksesang:
> > Have you tried to just set the source path of mill-build root project to . or millSourcePath?
> 
> set source path to `.` or `millSourcePath` may be inappropriate too
> 
> https://build-server-protocol.github.io/docs/specification.html#build-target-sources-request
> 
> ```
>   /** Either a text document or a directory. A directory entry must end with a forward
>    * slash "/" and a directory entry implies that every nested text document within the
>    * directory belongs to this source item.
>    */
> ```

The path will be translated to an absolute path, though? Which would end with a forward slash.

heksesang:
> I don't know Ammonite well enough, but I think the answer is yet. Ammonite itself seems to provide BSP support, so the best would be to look into that implementation.
> 
> Back to this concrete PR, as `src` is in almost all cases not the right setting, but `build.sc` is probably too restricted. Have you tried to just set the source path of `mill-build` root project to `.` or `millSourcePath`?

`src` would in no normal circumstances match the build files, so if you change it to `build.sc`, I think it is no less restricted than before in practical terms.

lefou (approved):
Thank you!

PR #1352, opened 2021-06-02
Prevent compilation during bloop config generation

Baccata:
Compilation of a module and its upstreams should be avoided at all cost when generating bloop configuration. 

https://github.com/com-lihaoyi/mill/pull/1086 adds a reference to a task that depends on `compile`. This PR prevents that from happening, and adds a test to ensure that compilation isn't triggered by `Bloop/install` 

@tgodzik 
@lolgab

### DIFF
--- a/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
@@ -157,7 +157,7 @@ class BloopImpl(ev: () => Evaluator, wd: Path) extends ExternalModule { outer =>
 
   def bloopConfig(module: JavaModule): Task[BloopConfig.File] = {
     import _root_.bloop.config.Config
-    def out(m: JavaModule) = { 
+    def out(m: JavaModule) = {
       val allSegs = m.millModuleShared.value.getOrElse(Segments()) ++ m.millModuleSegments
       bloopDir / "out" / allSegs.render
     }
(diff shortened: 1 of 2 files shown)


contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
```diff
@@ -195,81 +195,6 @@ class BloopImpl(ev: () => Evaluator, wd: Path) extends ExternalModule { outer =>
       case _ => T.task(None)
     }
 
-    ////////////////////////////////////////////////////////////////////////////
```
Baccata, line 198: moved below to prevent a forward-reference.

contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
```diff
@@ -405,6 +330,81 @@ class BloopImpl(ev: () => Evaluator, wd: Path) extends ExternalModule { outer =>
...
+              options = {
+                // See https://github.com/scalacenter/bloop/issues/1167
+                val forkArgs = module.forkArgs().toList
+                if (forkArgs.exists(_.startsWith("-Duser.dir="))) forkArgs
+                else s"-Duser.dir=$wd" :: forkArgs
+              }
+            ),
+            mainClass = module.mainClass(),
+            runtimeConfig = None,
+            classpath = Some(classpath().map(_.toNIO).toList),
```
Baccata, line 402: This is the only line that changed. 

It would be good to understand what bloop uses this for

tgodzik: Isn't that just classpath for compilation?

Baccata: Nope, the classpath for compilation is [there](https://github.com/com-lihaoyi/mill/blob/abe5220d8958973a9afdc058d96df387f8f02520/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala#L430)

Baccata: I mean, yeah, the value that I'm assigning to the config (in this PR) is the the same as the one that's used for compilation, but here I'm asking about the meaning of that particular configuration element.

tgodzik: Wait, that is in platform, let me check it.

tgodzik: So this was added as a separate configuration for running on JVM. More info in the commit https://github.com/scalacenter/bloop/commit/8898a93c0f88219a2edbe53b6f8d0765b58d20f7

Baccata: Right so according to the comment it could be left empty. Right ?

tgodzik: It will use the main classpath if it's missing. Looks like it.

contrib/bloop/test/src/mill/contrib/bloop/BloopTests.scala
```diff
@@ -80,6 +80,18 @@ object BloopTests extends TestSuite {
...
+      "no-compilation" - {
+        val workspaceOut = os.pwd / "target" / "workspace" / "mill" / "contrib" / "bloop" / "BloopTests" / "testEvaluator"
+        val scalaModuleCompile = workspaceOut / "scalaModule" / "compile"
+        val scalaModule2Compile = workspaceOut / "scalaModule2" / "compile"
+
+        // Ensuring that bloop config generation didn't trigger compilation
+        assert(os.exists(workspaceOut / "scalaModule"))
+        assert(!os.exists(workspaceOut / "scalaModule" / "compile"))
+        assert(os.exists(workspaceOut / "scalaModule2"))
+        assert(!os.exists(workspaceOut / "scalaModule2" / "compile"))
```
lefou, line 92: Are these `val`s unused? Should they be used in the assertions or removed?

Baccata: My bad, 2b60d12

lefou (approved):
Looks good to me. Thanks!

PR #1381, opened 2021-06-16
Update bsp4j to 2.0.0-M14

forty:
Updating in the hope to fix https://github.com/com-lihaoyi/Ammonite/issues/1175

I already updated guava in bsp4j https://github.com/build-server-protocol/build-server-protocol/pull/176. Apparently I will also need to do a PR in Ammonite https://github.com/com-lihaoyi/Ammonite/blob/master/build.sc#L76


forty:
Closing in favor of https://github.com/com-lihaoyi/mill/pull/1378 which I missed earlier :)

PR #1383, opened 2021-06-22
Small typo

kag0:
@lefou's catch actually


lefou (approved):
Nice. Thanks for fixing.

PR #1385, opened 2021-06-23
Allow custom sources for Javadoc and Scaladoc

### LINKED ISSUE #1384: Can't modify scaladoc sources.
## Introduction
Version: 0.9.8
Mill doesn't provide any way to set custom scaladoc sources, for example, when you want to create a single scaladoc of all of your modules.

## Solution
Add a new dedicated property and use it in `docJar` instead of `allSourcesFiles` directly

Iltotore:
This pull request allows the user to define custom sources for Javadoc/Scaladoc though `processedDocSources`:
- `JavaModule#docJar` now uses `processedDocSources` instead of `allSources`
- `ScalaModule#docJar` now uses `processedDocSources` instead of `os.walk(compile().classes.path).map(PathRef(_))`

Fixes #1384 .


scalalib/src/ScalaModule.scala
```diff
@@ -283,7 +304,7 @@ trait ScalaModule extends JavaModule { outer =>
 
       packageWithZinc(
         Seq("-d", javadocDir.toNIO.toString),
-        allSourceFiles().map(_.path.toString),
+        processedDocSources().map(_.path.toString),
```
lefou, line 307: The doesn't look right. `processedDocSources` seem to contain directories, whereas `allSourceFiles` were concrete source files.

lefou (commented):
I'm not that happy with the name `processedDocSources`, as I immediately asked myself what "processed" could mean. Why not just `docSources`? Some scaladoc should indicate the purpose of the new target.

Also, one motivation to customize the docSources could be to adapt the set of sources files, so it might be probably worth a though, if we could feed files instead of dirs.

It looks like some unrelated formatting changes went into this PR.

scalalib/src/ScalaModule.scala
Iltotore: I also thook about docSources but this method already exists and it isn't the same feature. According to scaladocs, `docSources` is for static (not processed by scaladoc) files.

What if we have, for some obscure reason, a tasty file and we just want to have it in our doc without processing it ?

Iltotore:
> It looks like some unrelated formatting changes went into this PR.
Yes, I just noticed my IDE reformatted the code. If needed, I can undo this reformat.

lefou:
> > It looks like some unrelated formatting changes went into this PR.
> 
> Yes, I just noticed my IDE reformatted the code. If needed, I can undo this reformat.

That would be nice. If you enable scalafmt in your IDE, it should automatically use the `.scalafmt.conf` from the repo.

> > I'm not that happy with the name processedDocSources, as I immediately asked myself what "processed" could mean. Why not just docSources?
> 
> I also thook about docSources but this method already exists and it isn't the same feature. According to scaladocs, `docSources` is for static (not processed by scaladoc) files.
> 
> What if we have, for some obscure reason, a tasty file and we just want to have it in our doc without processing it ?

I must have overlooked the existing but quite new `docSources`. Unfortunately the existing `docSources` has also misleading name, as it does not contain sources (to be process by the doc tool) but static resources to be bundled as-is. I guess we should rename to `docResources` and deprecate the old name for the long-term benefit. The whole `docSources` seems to be unsupported outside of Scala 3, so we might as well clear it up without deprecation cycles, which also allows us to re-use it's name now.

Besides, the "processed" prefix is misleading, as these sources are going to be processed by the doc tool.

scalalib/src/ScalaModule.scala
lefou: This doesn't address my comment.

Iltotore:
> That would be nice. If you enable scalafmt in your IDE, it should automatically use the .scalafmt.conf from the repo.
That's what I did. I will push the code formatted using scalafmt

>  I guess we should rename to docResources
Do you want me to commit this change then, use `docSources` instead of processed ?

scalalib/src/ScalaModule.scala
Iltotore: My edit didn't pass. I'll re-edit the message

lefou: No, `processDocSources` is meant to be overridden as configuration. We have to handle it here. Just walk all dirs returned from `processedDocSources` and use all source files.

lefou:
> > I guess we should rename to docResources
> 
> Do you want me to commit this change then, use `docSources` instead of processed ?

Yes, please!

To avoid conflicts with users who used that target to provide static resources, we should probable output a warning in case the outcome of `docSources` isn't empty but does not contain any scala or java source files. Something like: "Since mill after 0.9.8 docSources should only be used to provide source files. To provide static resources, please use docResources."

Iltotore (commented):
Earlier we talked about reformatting modified files using ScalaFMT but it indicated that they're already formatted correctly.

Iltotore:
If this change is approved, I will add the suggested warning.

lefou:
> Earlier we talked about reformatting modified files using ScalaFMT but it indicated that they're already formatted correctly.

Scalafmt isn't going to remove new lines which you introduced earlier.

scalalib/src/JavaModule.scala
```diff
@@ -352,13 +352,15 @@ trait JavaModule
    */
   def javadocOptions: T[Seq[String]] = T { Seq[String]() }
 
+  def docSources: T[Seq[PathRef]] = allSources
```
lefou, line 355: Please add some scaladoc indicating the purpose of this target and the expected return value. A reference to `docResources` would be nice, too.

Here, in `JavaModule`, you implement `docSources` it with _source directories_, whereas in `ScalaModule` you override with with a list of _source files_. Source directories is the better choice.

scalalib/src/ScalaModule.scala
```diff
@@ -185,6 +201,11 @@ trait ScalaModule extends JavaModule { outer =>
       )
   }
 
+  override def docSources: T[Seq[PathRef]] = T {
```
lefou, line 204: Here, you generate a `Seq` of source files instead of source directories.

scalalib/src/ScalaModule.scala
```diff
@@ -272,9 +293,9 @@ trait ScalaModule extends JavaModule { outer =>
           "-siteroot",
           combinedStaticDir.toNIO.toString
         ),
-        os.walk(compile().classes.path)
-          .filter(_.ext == "tasty")
-          .map(_.toString),
+        docSources()
```
lefou, line 296: Here we need to walk the source directories to filter the files. Same in the other branch.

lefou (changes requested):

scalalib/src/ScalaModule.scala
Iltotore: Fixed in https://github.com/com-lihaoyi/mill/pull/1385/commits/e7934292cf2b2bfc781d788e1e3a530d90dc7b46

scalalib/src/ScalaModule.scala
Iltotore: Fixed in https://github.com/com-lihaoyi/mill/pull/1385/commits/e7934292cf2b2bfc781d788e1e3a530d90dc7b46

scalalib/src/JavaModule.scala
Iltotore: The first point is fixed in https://github.com/com-lihaoyi/mill/pull/1385/commits/ec90f4843d799a8875dc646e87c681d7fd2b059c and https://github.com/com-lihaoyi/mill/pull/1385/commits/521a41575c444f011eb33fb8f6abb57206c352a0.

The second is fixed in https://github.com/com-lihaoyi/mill/pull/1385/commits/e7934292cf2b2bfc781d788e1e3a530d90dc7b46.

scalalib/src/JavaModule.scala
```diff
@@ -355,10 +355,19 @@ trait JavaModule
   /**
    * Extra directories to be processed by the API documentation tool.
    *
+   * Typically includes other tasty/source files to generate documentation from.
```
lefou, line 358:
```suggestion
   * Directories to be processed by the API documentation tool.
   *
   * Typically includes other tasty/source files to generate documentation from.
```

scalalib/src/ScalaModule.scala
```diff
@@ -283,7 +302,10 @@ trait ScalaModule extends JavaModule { outer =>
 
       packageWithZinc(
         Seq("-d", javadocDir.toNIO.toString),
-        allSourceFiles().map(_.path.toString),
+        docSources()
+          .map(_.path)
+          .filter(_.ext == "tasty")
```
lefou, line 307: The tasty file filter is wrong for Scala 2, instead it should select same files as `allSourceFiles` does.

Iltotore: My bad. I read "Scala 3" instead of "2" on the above comment.

Fixed in https://github.com/com-lihaoyi/mill/pull/1385/commits/c214aa7244d8d4d719c7ef4afaaf99c39bcabdd4

lefou (changes requested):

scalalib/src/ScalaModule.scala
```diff
@@ -283,7 +302,7 @@ trait ScalaModule extends JavaModule { outer =>
 
       packageWithZinc(
         Seq("-d", javadocDir.toNIO.toString),
-        allSourceFiles().map(_.path.toString),
+        docSources().map(_.path.toString),
```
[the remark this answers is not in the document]
Iltotore: I didn't use os-lib. Should I walk the Sources of `docSources` or use files instead of directories inside `docSources` ?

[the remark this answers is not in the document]
lefou: `packageWithZinc` expects a `Seq` of source files (as `String`). `allSourceFiles()` already provides individual sources files, whereas `docSources()` only provides the directories containing these. So you need to find all sources files and make them relative. Refer to the implementation of `allSourceFiles` for a hint.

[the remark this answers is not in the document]
Iltotore: So can't I use `allSourceFiles()` as default implementation of `docSources` in `ScalaModule` ?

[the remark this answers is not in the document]
lefou: No, because the semantic of the new `docSources` is (obviously) "directory" not "files".

Iltotore:
This last change works for Scala 3 and Scala 3 Milestone. Haven't tested with Scala 2 yet.

lefou:
There are still failing tests related to scaladoc.

Iltotore:
Actually, I can't directly test the project on my machine due to a bug:
```
-- Error: C:\Users\rapha\IdeaProjects\mill\target\workspace\mill\scalalib\DottyDocTests\eval\multiple\multidocs\compile\dest\classes\pkg\SomeClass.tasty:2:407 
2 |??u?6??p?V3☻?@???o?t☻?3???p?V2☻?‼↨ ?☺??☺???p?V3???u?3?3?q?p?3?u?Pu?Pu?3?v?\r☻?3????)3????)u?3???p?Vu?3?R3♥?♠↨???)3???p?V3?↨?????3?↨\♦♦↨????3?↨?????u?3?t♥?P3☻??p?V3?↨‼↨??3☻???p?V3☻?@??☺?????
??☺????~???????????????????????????????????????????????????????☺?????☺?????☻?}???????????????????????????????????????/** A doc comment. */▬?☻?/** A doc comment. */▬?☻?/** A doc comment. */▬?
  |

                          ^
  |                                                illegal character '\u0000'
```

Seems like an encoding error but also fails when specifying UTF-8 as encoding.

scalalib/src/ScalaModule.scala
```diff
@@ -185,6 +201,8 @@ trait ScalaModule extends JavaModule { outer =>
       )
   }
 
+  override def docSources: Sources = T.sources(compile().classes.path)
```
lefou, line 204: This seem not correct for early dotty / scala 3 versions (`isDotty(scalaVersion()) || isScala3Milestone(scalaVersion())`) which use dottydoc and consumes scala source files.
Only Scala 3 releases (`isScala3(scalaVersion())`) use scaladoc 3 tool, which consumes tasty files.

lefou (changes requested):

Iltotore:
The last commit should fix this issue. All scalalib tests passed in local.

lefou (commented):
Your PR looks good now. There are still some newly introduced but unwanted newlines. Can you please remove these, and we are good to go.

scalalib/src/ScalaModule.scala
```diff
@@ -185,6 +187,11 @@ trait ScalaModule extends JavaModule { outer =>
       )
   }
 
+  override def docSources: Sources = T.sources(
+    if(isDotty(scalaVersion()) || isScala3Milestone(scalaVersion())) allSources()
+    else Seq(PathRef(compile().classes.path))
+  )
```
lefou, line 193: I think, this is still not correct, e.g. for Scala 2. Best would be to explicitly guard the scala 3 case, as suggested previously.

```suggestion
  override def docSources: Sources = T.sources {
    // Scaladoc 3.0.0 is consuming tasty files
    if(isScala3(scalaVersion() && !isScala3Milestone(scalaVersion())) Seq((compile().classes))
    else allSources()
  }
```

lefou (changes requested):

scalalib/src/ScalaModule.scala
```diff
@@ -185,6 +187,12 @@ trait ScalaModule extends JavaModule { outer =>
       )
   }
 
+  override def docSources: Sources = T.sources {
+    // Scaladoc 3.0.0 is consuming tasty files
+    if(isScala3(scalaVersion() && !isScala3Milestone(scalaVersion())) Seq((compile().classes))
+    else allSources()
+  }
```
lefou, line 194: Fixing compile issue

```suggestion
  override def docSources: Sources = T.sources {
    // Scaladoc 3.0.0 is consuming tasty files
    if(isScala3(scalaVersion()) && !isScala3Milestone(scalaVersion())) Seq(compile().classes)
    else allSources()
  }
```

lefou (approved):
Looks good to me. Thank you!

PR #1393, opened 2021-07-02
Fix #1392: Avoid using ClosureCompiler with ESModule

### LINKED ISSUE #1392: ScalaJSModule fails to link with ESModule + fullOpt
Scala.js doesn't allow using the Closure compiler together with `ESModule` `ModuleKind`.
The Sbt plugin handles this [here](https://github.com/scala-js/scala-js/blob/bc3df88ba3bf383bb58ddee2d1ca5312b2f926cd/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala#L419) and mill should do the same, otherwise an exception is thrown during linking.
```
java.lang.IllegalArgumentException: requirement failed: Cannot use module kind ESModule with the Closure Compiler
```

lolgab:


lefou (approved):
Looks good to me! 

I need to find out, why the test suite isn't succeeding, which looks unrelated, though. Is this urgent? Do we also need a fix in 0.9.x branch?

lolgab:
> Is this urgent? Do we also need a fix in 0.9.x branch?

I don't think so, we had one user falling into this who temporarily switched to `NoModule`.

lefou:
Merged. Thank you!

PR #1421, opened 2021-07-23
Use Java 8 to build release artifacts

lefou:


ice1000:
What's the rationale?

lefou:
Before this PR, we published with Java 9. Although we explicitly select the target as Java 8, some binary incompatibilities sneaked in all of our current releases. So, I think it's safer to just use Java 8 for compiling the release versions.

See the output of this experimental integration of MiMa in the mill build:

```
09:59:20 ⌁ lefou:~/work/opensource/mill mima+* 130 ± mill -j 0 main.__.mimaReportBinaryIssues
Using experimental parallel evaluator with 16 threads
[#07] [281/306] main.core.compile 
[#11] Scanning 79 classfiles for binary compatibility in /home/lefou/work/opensource/mill/out/main/api/compile/dest/classes ...
[#03] [282/306] scalalib.api.localClasspath 
[#11]  * static method transferTo(java.io.OutputStream)Long in class mill.api.DummyInputStream does not have a correspondent in current version
   filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("mill.api.DummyInputStream.transferTo")
[#11]  * static method readNBytes(Array[Byte],Int,Int)Int in class mill.api.DummyInputStream does not have a correspondent in current version
   filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("mill.api.DummyInputStream.readNBytes")
[#11]  * static method readAllBytes()Array[Byte] in class mill.api.DummyInputStream does not have a correspondent in current version
   filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("mill.api.DummyInputStream.readAllBytes")
```

You can see that comparing the HEAD of `0.9.x` branch with the `0.9.9` tag, some methods only defined in Java 9 sneak into our API. This happens when I run (and compile) mill with Java 8.

ice1000:
Thanks for the explanation!

lolgab:
Glad that the mill-mima plugin is already creating some value 😍

PR #1427, opened 2021-07-27
Update coursier to 2.0.16+73-gddc6d9cc9

alexarchambault:
On Windows, this should fix some issues where trying to access the coursier cache blocks indefinitely.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -117,7 +117,7 @@ jobs:
 
       - uses: actions/setup-java@v1
         with:
-          java-version: 9
+          java-version: 8
 
       - run: ci/release-maven.sh
 
@@ -137,6 +137,6 @@ jobs:
 
       - uses: actions/setup-java@v1
         with:
-          java-version: 9
+          java-version: 8
 
       - run: ./mill -i uploadToGithub $REPO_ACCESS_TOKEN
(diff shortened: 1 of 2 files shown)


lefou:
Can you provide any link to an issue or pull request with more details?

alexarchambault:
@lefou That's mainly to benefit from https://github.com/coursier/coursier/pull/2025. On my personal Windows machine, recent versions of Mill get stuck when coursier tries to compute its cache directory. https://github.com/coursier/coursier/pull/2025 is meant to address that. I'm not sure when or why this issue happens (it's not been a problem on GitHub actions Windows workers for example).

lefou (approved):
Looks good to me. Thank you! But I need to find out, why some tests now fail on Windows. Unrelated to this pull request I think, but still it fails the build and publish pipeline. I understand the error, but not why it fails now but not before. Maybe, the GH images have changed.

lefou:
I don't know why only this PR fails but not others. But I have identified the path logic incosistency in this test compared to the GenIdeaImpl logic and create another PR to fix it: #1466 .

alexarchambault:
That might be because the coursier cache path calculation changed (no more `null` with this coursier version), which makes it land in a different drive in these Windows jobs.

PR #1448, opened 2021-08-29
Fix formatting and wording in intro docs

sake92:
Found some formatting issues with asciidoc, it considered `_` as a start of an italic word.  
It needs to be escaped in some cases, like this: `\_`

Also, I ran some bigger sentences through https://hemingwayapp.com/, and the text was a bit complicated, so I split some of them into 2 simpler ones...

### DIFF
--- a/docs/antora/modules/ROOT/pages/Intro_to_Mill.adoc
+++ b/docs/antora/modules/ROOT/pages/Intro_to_Mill.adoc
@@ -5,16 +5,17 @@
 http://www.lihaoyi.com/post/SowhatswrongwithSBT.html[Scared of SBT]? Melancholy over Maven? Grumbling about Gradle?
 Baffled by Bazel? Give Mill a try!
 
-Mill aims for simplicity by re-using concepts you are already
+Mill aims for simplicity by reusing concepts you are already
 http://www.lihaoyi.com/post/BuildToolsasPureFunctionalPrograms.html[familiar with], borrowing ideas from modern tools
-like https://bazel.build/[Bazel], to let you build your projects in a way that's simple, fast, and predictable.
+like https://bazel.build/[Bazel].   
+It lets you build your projects in a way that's simple, fast, and predictable.
 
-Mill has built in support for the https://www.scala-lang.org/[Scala]
+Mill has built-in support for the https://www.scala-lang.org/[Scala]
 programming language, and can serve as a replacement for
-http://www.scala-sbt.org/[SBT], but can also be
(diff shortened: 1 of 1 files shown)


docs/antora/modules/ROOT/pages/Intro_to_Mill.adoc
```diff
@@ -213,7 +214,7 @@ $ mill -i foo.console              # start a Scala console within your project (
 $ mill -i foo.repl                 # start an Ammonite REPL within your project (in interactive mode: "-i")
 ----
 
-You can run `mill resolve __` to see a full list of the different tasks that are available, `mill resolve foo._` to see
+You can run `mill resolve \__` to see a full list of the different tasks that are available, `mill resolve foo._` to see
```
lefou, line 217: I think the correct asciidoc approach is to use `+` symbols here.

```suggestion
You can run +mill resolve __+ to see a full list of the different tasks that are available, `mill resolve foo._` to see
```

docs/antora/modules/ROOT/pages/Intro_to_Mill.adoc
```diff
@@ -514,7 +515,7 @@ Inputs:
 it also displays its source location and a list of input tasks. This is very useful for debugging and interactively
 exploring the structure of your build from the command line.
 
-`inspect` also works with the same `_`/`__` wildcard/query syntaxes that
+`inspect` also works with the same `\_`/`__` wildcard/query syntaxes that
```
lefou, line 518:
```suggestion
`inspect` also works with the same +_+/+__+ wildcard/query syntaxes that
```

lefou (commented):
Thanks for your eagle eyes! I bikeshedded a bit. ;-)

docs/antora/modules/ROOT/pages/Intro_to_Mill.adoc
sake92: Fixed

docs/antora/modules/ROOT/pages/Intro_to_Mill.adoc
```diff
@@ -52,8 +53,8 @@ pkg install mill
 
 === Windows
 
-To get started, download Mill from:
-{mill-github-url}/releases/download/{mill-last-tag}/{mill-last-tag}-assembly, and save it as `mill.bat`.
+To get started, download Mill from
+{mill-github-url}/releases/download/{mill-last-tag}/{mill-last-tag}-assembly[here], and save it as `mill.bat`.
```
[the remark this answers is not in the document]
sake92: Fixed

lefou (approved):
Great, thank you!

PR #1451, opened 2021-09-02
Add more information in default manifests

lefou:


alexarchambault:
You may want to add even more attributes, such as those added in https://github.com/coursier/coursier/commit/be88c978e74fb930738778e4fbf277b85369284d. sbt was adding those in particular (and most coursier versions rely on some of them in the `cs launch` command, even though the upcoming version will not require them).

lefou:
> You may want to add even more attributes, such as those added in [coursier/coursier@be88c97](https://github.com/coursier/coursier/commit/be88c978e74fb930738778e4fbf277b85369284d). sbt was adding those in particular (and most coursier versions rely on some of them in the `cs launch` command, even though the upcoming version will not require them).

@alexarchambault  Exactly the feedback I was hoping for. Thanks!

lefou:
@alexarchambault  Do you have a pointer which attributes are included by default by sbt?

alexarchambault:
A quick search gave me [this](https://github.com/sbt/sbt/blob/fe5497938a8b2898dcbab49b60e0f2ef53b06c29/main-actions/src/main/scala/sbt/Package.scala#L177-L207), called from [here](https://github.com/sbt/sbt/blob/8586e19f62c60d2e6cc903ddc71df94f3fb2e012/main/src/main/scala/sbt/Defaults.scala#L1679-L1682).

For https://github.com/coursier/coursier/commit/be88c978e74fb930738778e4fbf277b85369284d, I just looked at past coursier JARs, that were created by sbt.

PR #1456, opened 2021-09-06
Add More Config Options to Docker Contrib Module

LaurenceWarne:
Hi!  This PR adds some more options (pretty much correspond 1-1 with Dockerfile instructions) that can be used in `DockerConfig`.  It's not totally comprehensive, but hopefully it catches the most common ones so there's less of a need to override `DockerConfig.build`.

I've added some tests which validate the new options by calling `docker build` - I could possibly replace these with tests against the generated `Dockerfile` strings if you think this is too much.

Thanks, I hope this is helpful.

### DIFF
--- a/contrib/docker/test/src/DockerModuleTest.scala
+++ b/contrib/docker/test/src/DockerModuleTest.scala
@@ -10,15 +10,21 @@ import utest.framework.TestPath
 
 object DockerModuleTest extends TestSuite {
 
+  private def testExecutable =
+    if (isInstalled("podman")) "podman"
+    else "docker"
+
   object Docker extends TestUtil.BaseModule with JavaModule with DockerModule {
 
     override def millSourcePath = TestUtil.getSrcPathStatic()
     override def artifactName = testArtifactName
 
-    object dockerDefault extends DockerConfig
+    object dockerDefault extends DockerConfig {
+      override def executable = testExecutable
+    }
 
     object dockerAll extends DockerConfig {
-      override def baseImage = "openjdk:11"
+      override def baseImage = "docker.io/openjdk:11"
       override def labels = Map("version" -> "1.0")
       override def exposedPorts = Seq(8080, 443)
       override def exposedUdpPorts = Seq(80)
@@ -29,6 +35,7 @@ object DockerModuleTest extends TestSuite {
         "useradd -ms /bin/bash user1"
       )
       override def user = "user1"
+      override def executable = testExecutable
     }
   }
 
@@ -36,6 +43,8 @@ object DockerModuleTest extends TestSuite {
 
   val testModuleSourcesPath: Path =
     os.pwd / "contrib" / "docker" / "test" / "resources" / "docker"
+
+  val multineRegex = "\\R+".r
 
   private def isInstalled(executable: String): Boolean = {
     val getPathCmd = if (scala.util.Properties.isWin) "where" else "which"
@@ -45,7 +54,7 @@ object DockerModuleTest extends TestSuite {
   private def workspaceTest(m: TestUtil.BaseModule)(t: TestEvaluator => Unit)(
       implicit tp: TestPath
   ): Unit = {
-    if (isInstalled("docker")) {
+    if (isInstalled(testExecutable) && !scala.util.Properties.isWin) {
       val eval = new TestEvaluator(m)
       os.remove.all(m.millSourcePath)
       os.remove.all(eval.outPath)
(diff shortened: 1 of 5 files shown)


contrib/docker/test/src/DockerModuleTest.scala
```diff
@@ -0,0 +1,69 @@
...
+import mill.api.PathRef
+import mill.scalalib.JavaModule
+import mill.util.{TestEvaluator, TestUtil}
+import os.Path
+import utest._
+import utest.framework.TestPath
+
+object DockerModuleTest extends TestSuite {
+
+  object Docker extends TestUtil.BaseModule with JavaModule with DockerModule {
```
LaurenceWarne, line 13: I've written these based on the other contrib module test suites, let me know if there's anything untoward here.

lefou (commented):
Thank you for your contribution! Your changes look good to me. About the tests: I personally find the call to docker a rather hard requirement to pass the tests. Additionally nowadays, using podman is quite common, which is a daemon-less alternative to docker. As these test won't pass on Windows either, it's probably a good idea to add a test which only checks the expected `Dockerfile` correctness, and make the actual docker call test conditional, e.g. only when a docker (or podman) executable can be found.

LaurenceWarne:
Thanks for taking a look!  I've added tests against the dockerfile contents in ee6e6017159a46a9532eaecfbd1c09bda9ae0d95 and disabled the tests which invoke the docker executable in the case where it's not found.

> docker (or podman)

Do you think we need to check for podman as well?  https://podman.io/ recommends that you `alias docker=podman`, so the tests would still be ran in that case?

contrib/docker/test/src/DockerModuleTest.scala
```diff
@@ -0,0 +1,114 @@
...
+      override def user = "user1"
+    }
+  }
+
+  val testArtifactName = "mill-docker-contrib-test"
+
+  val testModuleSourcesPath: Path =
+    os.pwd / "contrib" / "docker" / "test" / "resources" / "docker"
+
+  private def isInstalled(executable: String): Boolean = {
```
LaurenceWarne, line 40: Stolen from https://github.com/com-lihaoyi/os-lib/blob/790b81aba491047b73dd481e290abb36ee1a7de8/os/test/src-jvm/TestUtil.scala#L12 :slightly_smiling_face:

contrib/docker/test/src/DockerModuleTest.scala
```diff
@@ -0,0 +1,114 @@
...
+      implicit tp: TestPath
+  ): Unit = {
+    if (isInstalled("docker")) {
+      val eval = new TestEvaluator(m)
+      os.remove.all(m.millSourcePath)
+      os.remove.all(eval.outPath)
+      os.makeDir.all(m.millSourcePath / os.up)
+      os.copy(testModuleSourcesPath, m.millSourcePath)
+      t(eval)
+    } else assert(true)
```
lefou, line 55: I'd like to see `println()` indicating that docker tests are skipped here.

lefou (commented):
> Do you think we need to check for podman as well? https://podman.io/ recommends that you `alias docker=podman`, so the tests would still be ran in that case?

The alias only works in a shell, so either we explicitly check and implement support for podman, or not. I leave it to you. But if we support it, this is probably worth a dedicated target, which is then also user configurable (in case both tools are present).

LaurenceWarne:
> I leave it to you to decide. But if we support it, this is probably worth a dedicated target, which is then also user configurable (in case both tools are present).

It's no problem, I've added a dedicated target in 9af2defc227ee04c94846d0a6c36b22265b1c94e.  I've ran the test against podman, docker and an environment without either, and the tests pass :+1:

LaurenceWarne:
I see a couple of the windows builds failed, currently :eyes:

LaurenceWarne:
`image operating system "linux" cannot be used on this platform`. I've disabled building for windows in the most recent commit, another solution could be to match on the platform to set the `baseImage` in the tests (though I'm not sure which image could be used :smile:).  What do you think?

lefou:
Disabling docker live tests on Windows is perfectly fine, I think.

LaurenceWarne:
Apologies, it's been a bit of a pain debugging some of the failures on the windows platforms, I _think_ it should be good now though.

lefou:
@LaurenceWarne Don't worry, it's the same for me. That's one of the reasons we have CI on Windows.

lefou (approved):
Looks good to me. Thank you!

LaurenceWarne:
Thanks!

PR #1461, opened 2021-09-08
Use unique pipe for each project

### LINKED ISSUE #1447: Nondeterministic client-server locks
As discussed [here](https://github.com/com-lihaoyi/mill/pull/1089#issuecomment-751461785), the goal is to have 5 Mill servers in total that serve multiple clients.  
The issue is that Mill [client code](https://github.com/com-lihaoyi/mill/blob/main/main/client/src/mill/main/client/MillClientMain.java#L134) locks on a file *inside a project*, and not globally.  
That means if you run `mill foo.compile` in another folder, it will create a new server instance (different lock file) **with the same pipe name**.  
That causes the client to *nondeterministically* connect to one of the servers...
Not sure if I am missing some crucial details here but that's my understanding... 😄 

The solution could be to put these lock files inside user home folder maybe?

sake92:
Closes #1447 

Now the pipe name takes the *full absolute path* of the project.  
Previously 2 server instances (for different projects) would use *same pipe name* and that would cause an issue...

https://docs.microsoft.com/en-us/windows/win32/ipc/pipe-names
> The pipe name string specified by PipeName can include any character other than a backslash, including numbers and special characters. The entire pipe name string can be up to 256 characters long. Pipe names are not case-sensitive.

### DIFF
--- a/main/client/src/mill/main/client/MillClientMain.java
+++ b/main/client/src/mill/main/client/MillClientMain.java
@@ -211,10 +211,10 @@ public class MillClientMain {
 
         while (ioSocket == null && System.currentTimeMillis() - retryStart < 5000) {
             try {
-                String socketBaseName = "mill." + md5hex(new File(lockBase).getCanonicalPath());
+                String socketBaseName = "mill-" + md5hex(new File(lockBase).getCanonicalPath());
                 ioSocket = Util.isWindows?
                         new Win32NamedPipeSocket(Util.WIN32_PIPE_PREFIX + socketBaseName)
-                        : new UnixDomainSocket(socketBaseName + "/io");
+                        : new UnixDomainSocket(lockBase + "/" + socketBaseName + "-io");
             } catch (Throwable e){
                 socketThrowable = e;
                 Thread.sleep(1);
@@ -263,7 +263,8 @@ public class MillClientMain {
         return processLimit;
     }
 
-    private static String md5hex(String str) throws NoSuchAlgorithmException {
+    /** @return Hex encoded MD5 hash of input string. */
+    public static String md5hex(String str) throws NoSuchAlgorithmException {
         return hexArray(MessageDigest.getInstance("md5").digest(str.getBytes(StandardCharsets.UTF_8)));
     }
 
--- a/main/client/src/mill/main/client/MillClientMain.java
+++ b/main/client/src/mill/main/client/MillClientMain.java
@@ -12,8 +12,10 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.UnsupportedEncodingException;
+import java.math.BigInteger;
 import java.net.Socket;
 import java.net.URISyntaxException;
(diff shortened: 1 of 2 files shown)


main/src/main/MillServerMain.scala
```diff
@@ -262,4 +268,10 @@ object Server {
       interrupt = false
     }
   }
+
+  def md5hex(str: String): String =
+    hexArray(MessageDigest.getInstance("md5").digest(str.getBytes(StandardCharsets.UTF_8)))
+
+  private def hexArray(arr: Array[Byte]) =
+    String.format("%0" + (arr.length << 1) + "x", new BigInteger(1, arr))
```
lefou, line 276: We should not have two copies of these helper methods. Same methods in class `MillClientMain` are already accessible but currently not visible.

main/client/src/mill/main/client/MillClientMain.java
```diff
@@ -259,4 +262,12 @@ private static int getServerProcessesLimit(String jvmHomeEncoding) {
...
         return processLimit;
     }
+
+    private static String md5hex(String str) throws NoSuchAlgorithmException {
+        return hexArray(MessageDigest.getInstance("md5").digest(str.getBytes(StandardCharsets.UTF_8)));
+    }
+
+    private static String hexArray(byte[] arr) {
+        return String.format("%0" + (arr.length << 1) + "x", new BigInteger(1, arr));
+    }
```
lefou, line 272: Make these public and add a small javadoc, to reuse them from `MillServerMain`.

lefou (changes requested):
Nice observation and fix. Thank you! Comments, see below.

main/client/src/mill/main/client/MillClientMain.java
```diff
@@ -209,9 +211,10 @@ public static int run(String lockBase,
 
         while (ioSocket == null && System.currentTimeMillis() - retryStart < 5000) {
             try {
+                String socketBaseName = "mill." + md5hex(new File(lockBase).getCanonicalPath());
                 ioSocket = Util.isWindows?
-                        new Win32NamedPipeSocket(Util.WIN32_PIPE_PREFIX + "mill." + new File(lockBase).getName())
-                        : new UnixDomainSocket(lockBase + "/io");
+                        new Win32NamedPipeSocket(Util.WIN32_PIPE_PREFIX + socketBaseName)
+                        : new UnixDomainSocket(socketBaseName + "/io");
```
lefou, line 217: I think on non-Windows systems we still need to provide an proper path to `UnixDomainSocket`, were all parent directories exists. After your change, this is no longer the case.

sake92: Yep, figured after tests failed.. 😄  
Should be fixed now.

main/client/src/mill/main/client/MillClientMain.java
sake92: Done.

lefou (approved):
Looks good to me.

PR #1483, opened 2021-09-21
replace FileSystems with ZipInputStream for parallel execution

atty303:
The steps to reproduce the problem are as follows.

1. Parallel execution is enabled.
2. Targets that use `scalaPBUnpackProto` will be executed in parallel.
3. `FileSystemAlreadyExistsException` will be thrown.

This is because FileSystems does not allow multiple instances of the same URI to be created.
Therefore, I have changed the implementation to use ZipInputStream.

### DIFF
--- a/contrib/scalapblib/src/ScalaPBModule.scala
+++ b/contrib/scalapblib/src/ScalaPBModule.scala
@@ -94,18 +94,18 @@ trait ScalaPBModule extends ScalaModule {
     val dest = T.dest
     cp.iterator.foreach { ref =>
       Using(new ZipInputStream(ref.path.getInputStream)) { zip =>
-        var entry = Option(zip.getNextEntry)
-        while (entry.nonEmpty) {
-          val name = entry.get.getName
-          if (name.endsWith(".proto")) {
-            val protoDest = dest / os.SubPath(name)
-            Using(os.write.outputStream(protoDest, createFolders = true)) { o =>
-              IO.stream(zip, o)
-            }
+        while ({
+          Option(zip.getNextEntry) match {
+            case None => false
+            case Some(entry) =>
+              if (entry.getName.endsWith(".proto")) {
+                val protoDest = dest / os.SubPath(entry.getName)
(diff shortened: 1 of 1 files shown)


contrib/scalapblib/src/ScalaPBModule.scala
```diff
@@ -94,35 +93,27 @@ trait ScalaPBModule extends ScalaModule {
...
-                }
-                super.visitFile(f, a)
+    val buffer = new Array[Byte](8192)
+    cp.iterator.foreach { ref =>
+      Using(new ZipInputStream(ref.path.getInputStream)) { zip =>
+        var entry = Option(zip.getNextEntry)
+        while (entry.nonEmpty) {
+          val name = entry.get.getName
+          if (name.endsWith(".proto")) {
+            val protoDest = os.Path(os.RelPath(name), dest)
```
lefou, line 103: I'd suggest `os.Subpath` which ensures we won't step outside the parent dir.

```suggestion
            val protoDest = dest / os.SubPath(name)
```

contrib/scalapblib/src/ScalaPBModule.scala
```diff
@@ -94,35 +93,27 @@ trait ScalaPBModule extends ScalaModule {
...
+            Using(os.write.outputStream(protoDest, createFolders = true)) { o =>
+              var r = 0
+              while (r != -1) {
+                r = zip.read(buffer)
+                if (r != -1) o.write(buffer, 0, r)
               }
```
lefou, line 109: You may want to use `mill.api.IO.stream` here.

lefou (commented):
Looks good to me. See my suggestion to use more mill API. Is this covered by any test?

atty303:
Thanks for the review!
I have fixed the points you pointed out.

> Looks good to me. See my suggestion to use more mill API. Is this covered by any test?

I don't think there are any tests related to this. I have tested it working on a project at work.

lefou (approved):
LGTM. Thank you!

PR #1486, opened 2021-09-21
GenIdea: Improved Scala 3 Support

lefou:
Implement `language-level` property for Scala library in IntelliJ IDEA.


lefou:
@informarte I hopefully improved the situation this time.

scalalib/src/GenIdeaImpl.scala
```diff
@@ -308,8 +308,14 @@ case class GenIdeaImpl(
...
       .view
+      .mapValues(_.sorted)
       .mapValues(_.zipWithIndex)
-      .flatMap(y => y._2.map(x => x._1 -> s"${y._1} (${x._2})"))
+      .flatMap(y =>
+        y._2.map {
+          case (path, 0) => path -> y._1
+          case (path, idx) => path -> s"${y._1} (${idx})"
+        }
+      )
```
lefou, line 318: Leaving out the index for the first element fixed failing tests on WIndows, where for unknown reasons we always ended up with a `scala-library-2.13.6.jar (0)`.

scalalib/src/GenIdeaImpl.scala
```diff
@@ -443,27 +449,40 @@ case class GenIdeaImpl(
...
+            val languageLevel = name match {
+              case _ if compilerCp.iterator.isEmpty => None
+              case _ if name.startsWith("scala3-library_3-3.0.") => Some("Scala_3_0")
+              case _ if name.startsWith("scala-library-2.13.") => Some("Scala_2_13")
+              case _ if name.startsWith("scala-library-2.12.") => Some("Scala_2_12")
+              case _ if name.startsWith("scala-library-2.11.") => Some("Scala_2_11")
+              case _ if name.startsWith("scala-library-2.10.") => Some("Scala_2_10")
+              case _ if name.startsWith("scala-library-2.9.") => Some("Scala_2_9")
+              case _ if name.startsWith("dotty-version-0.27") => Some("Scala_0_27")
+              case _ => None
```
lefou, line 470: Decided to hardcode the mapping, as Scala 3 support in IJ is still in development and it's a bit unclear if we settle with the current mapping or not.

PR #1488, opened 2021-09-22
scalapblib: support scalajs by default

atty303:
ScalaJS can be supported if gRPC is not used.


lefou (approved):
Looks good to me. Thanks!

PR #1489, opened 2021-09-22
Fix #1487: Introduce `mandatoryIvyDeps` and add to pom

### LINKED ISSUE #1487: `scala-library` dependency is not added to published jars' `pom.xml`
The issue was noticed by @olafurpg in Geny here: https://github.com/com-lihaoyi/geny/issues/32

When publishing libraries with Sbt the scala-library dependency is declared in the `dependencies` section of the `pom.xml`.
With libraries published with Mill this doesn't happen.

The same problem also happens with ScalaJS libraries that doesn't have any dependencies in the pom when created with Mill, but  when compiled with Sbt they emit these dependencies:
```xml
<dependencies>
    <dependency>
        <groupId>org.scala-lang</groupId>
        <artifactId>scala-library</artifactId>
        <version>2.13.5</version>
    </dependency>
    <dependency>
        <groupId>org.scala-js</groupId>
        <artifactId>scalajs-library_2.13</artifactId>
        <version>1.7.0</version>
    </dependency>
    <dependency>
        <groupId>org.scala-js</groupId>
        <artifactId>scalajs-test-bridge_2.13</artifactId>
        <version>1.7.0</version>
        <scope>test</scope>
    </dependency>
</dependencies>
```

lolgab:
`PublishModule` extends JavaModule so doesn't know about
`scalaLibraryIvyDeps`. To make it insert the `scalaLibraryIvyDeps`
into the pom we introduce two new targets in `JavaModule`:
`mandatoryIvyDeps` which will contain libraries users unlikely will want
remove, like the content of `scalaLibraryIvyDeps`, and `alIvyDeps`,
which will be useful for consumers to get all the ivyDeps without missing
the mandatory ones.

### DIFF
--- a/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
@@ -299,11 +299,6 @@ class BloopImpl(ev: () => Evaluator, wd: Path) extends ExternalModule { outer =>
     //  Ivy dependencies + sources
     ////////////////////////////////////////////////////////////////////////////
 
-    val scalaLibraryIvyDeps = module match {
-      case x: ScalaModule => x.scalaLibraryIvyDeps
-      case _ => T.task { Loose.Agg.empty[Dep] }
-    }
-
     /**
      * Resolves artifacts using coursier and creates the corresponding
      * bloop config.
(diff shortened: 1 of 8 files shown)


lefou:
Oh, that was a bit too quick and unexpected for me. I also tackled this one, and as it seems slightly different. If you wanna take a look: https://github.com/com-lihaoyi/mill/compare/main...lefou:scala-library-in-ivy-deps?expand=1

The good thing is, that we can better discuss the differences.

lolgab:
No worries, I was going to ask you before, but then was bored and did it right away :-D

What I did here was to replicate what I did in https://github.com/com-lihaoyi/mill/pull/1428 with the `mandatoryScalacOptions`.
I noticed that we treat the scalaLibraryIvyDeps as a special thing, but in the end, we always concatenate them with the other ivyDeps. Making `transitiveIvyDeps` depend on `allIvyDeps` lets us avoid having to concatenate `scalaLibraryIvyDeps` everywhere, makes the code simpler, and avoids bugs when you forget to do so.
On the negative side, `allIvyDeps` is a sad name since there are a lot of different types of ivyDeps (like `runIvyDeps` and `compileIvyDeps`) and it doesn't concatenate them all.
I didn't deprecate scalaLibraryIvyDeps because I made mandatoryIvyDeps depend on it, and maybe some user needs them alone, but we can deprecate them if you think it's better.
I also fixed the Scala Native module and Scala.js and tested that scala-library is actually in the POM

lolgab:
Superseded by #1490

lefou:
This PR was merged as part of #1490 .

PR #1490, opened 2021-09-23
Refined mandatory ivy libs handling

### LINKED ISSUE #1487: `scala-library` dependency is not added to published jars' `pom.xml`

lefou:
Add `JavaModule.mandatoryIvyDeps` target to contain typical required ivy dependencies without the need to override and include `super.ivyDeps`.

This fixes #1487 and streamlines platform dependency handling for downstream modules like `ScalaModule`, `ScalaJSModule`, `ScalaNativeModule`. It's probably also useful for other modules which need to bring-in additional dependencies.

This PR includes and refines PR #1489.

### DIFF
--- a/scalajslib/src/ScalaJSModule.scala
+++ b/scalajslib/src/ScalaJSModule.scala
@@ -169,8 +169,10 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
       }
     }
   }
-  override def scalaLibraryIvyDeps = T {
-    super.scalaLibraryIvyDeps() ++ Seq(
+
+  /** Adds the Scala.js Library as mandatory dependency. */
+  override def mandatoryIvyDeps = T {
+    super.mandatoryIvyDeps() ++ Seq(
       ivy"org.scala-js::scalajs-library:${scalaJSVersion()}".withDottyCompat(scalaVersion())
     )
   }
--- a/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
@@ -370,8 +370,8 @@ class BloopImpl(ev: () => Evaluator, wd: Path) extends ExternalModule { outer =>
 
     val bloopResolution: Task[BloopConfig.Resolution] = T.task {
       val repos = module.repositoriesTask()
-      val allIvyDeps =
-        module.transitiveIvyDeps() ++ module.transitiveCompileIvyDeps()
+      // same as input of resolvedIvyDeps
+      val allIvyDeps = module.transitiveIvyDeps() ++ module.transitiveCompileIvyDeps()
       val coursierDeps =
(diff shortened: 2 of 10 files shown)


lolgab (approved):
I was uncertain too about the `allIvyDeps` target. Good move to remove it.
It's so nice when you remove things and fix bugs at the same time.
LGTM! 👍

scalajslib/src/ScalaJSModule.scala
```diff
@@ -169,8 +169,10 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
       }
     }
   }
-  override def scalaLibraryIvyDeps = T {
-    super.scalaLibraryIvyDeps() ++ Seq(
+
+  /** Adds the Scala.js Library as mandatory dependency. */
+  override def mandatoryIvyDeps = T {
+    super.mandatoryIvyDeps() ++ Seq(
```
lolgab, line 175: This introduces a behavioural change, since now `scalaLibraryIvyDeps` don't contain the scala-js library anymore. I wonder if it makes sense to introduce a breaking change and remove the `scalaLibraryIvyDeps`target all together.

What do you think?

lefou: At least the target now returns what it claims with its name. ;-) I was under the impression this target is only used internally, but you're right, it's public after all. We should probably document that change. We could also add an additonal `scalaJSLibraryIvyDeps` which is then re-used in `mandatoryIvyDeps`.

lolgab: > We should probably document that change.

Agree! 👍 

> We could also add an additional scalaJSLibraryIvyDeps which is then re-used in mandatoryIvyDeps.

My usual take, let's wait for someone asking for it. I'm more on removing or deprecating `scalaLibraryIvyDeps` than add more of them to make it harder to remove. I would not change anything for now so we can think about it.

lefou: I added a changelog entry and keep implementation as is.

PR #1494, opened 2021-09-24
Setup mima plugin to check for binary compatibility

lefou:
- Setup mima plugin to check for binary compatibility
- Enable binary compatibility checks in CI


lolgab:
Wow! The plugin I wrote being used in mill itself! And the exclude filters seem to be working too! Nice

PR #1496, opened 2021-09-25
Fix broken link in shell script.

Wafelack:
Shell script was trying to download a file from a non-existing release (`0.10.0`) instead of `0.10.0-M2`.


lefou (changes requested):
Unfortunately this change will break support for snapshot releases, which are also separated by a dash (`-`).

I already fixed this issue in `millw`, an alternative launcher for mill (see PR https://github.com/lefou/millw/pull/20). Feel free to use and adapt it. 

As a side note: I was arguing with myself if I should copy this change myself, but honestly I think instead, we should just use millw in mill. But as that's a different discussion and should not be discussed in this PR.

lefou:
Superseded by #1505 .

PR #1517, opened 2021-10-07
Update coursier version to 2.0.16-158-gbdc8669f9

lefou:
This fixes some property activation for Maven dependencies, which may cause resolution failures for newer JavaFX dependencies.

See https://github.com/coursier/coursier/pull/2176


rom1dep:
Seems to work fine against this dependency!

lefou:
> Seems to work fine against this dependency!

Great. Thanks for testing!

PR #1521, opened 2021-10-10
Support a target separator (`+`) in the argument parser

lefou:
This PR adds support to specify arbitrary targets and target arguments in one mill run.

__Motivation:__ Current mill only supports the selection of one or multiple targets followed by an argument list which is applied to all targets. Providing different arguments is currently not supported.

__Solution:__ We introduce a new separator `+` which separates the target arguments from the next target. To pass the `+` as argument to a target, masking with `\+` is supported.

Example:

```bash
mill __.compile + core.testCached + itest.test fastTest slowTests + __.publishLocal
```

This command will:
* run `compile` for all modules
* run `testCached` for module `core`
* run `itest.test` command with arguments ` fastTest` and `slowTests`
* run `publishLocal` for all modules

The limitation, that a command can only run once is still present and not changed by this PR.

I could imagine, that we later also provide a `++` separator, to specify targets that need to run after each other. This would also allow running the same command multiple times (with different args).

This also makes the `all` and `par` commands obsolete.

Review by @lihaoyi

### DIFF
--- a/main/core/src/mill/util/ParseArgs.scala
+++ b/main/core/src/mill/util/ParseArgs.scala
@@ -40,6 +40,8 @@ object ParseArgs {
       selectMode: SelectMode
   ): Either[String, Seq[TargetsWithParams]] = {
 
+    val MaskPattern = ("""\\+\Q""" + TargetSeparator + """\E""").r
+
     /**
      * Partition the arguments in groups using a separator.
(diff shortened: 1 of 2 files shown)


lihaoyi:
How thorough is the masking? e.g. if i want to pass a literal `\+`, should I do `\\+`?

PR #1526, opened 2021-10-12
Added --import command-line option

lefou:
Motivation: Some plugins provide "External Modules" which can be used by mill without modifying the `build.sc`.

With this PR applied, you can run targets from external modules from the command-line without touching the `build.sc`.

Example:

```
$ dev-mill --plugin ivy:de.tototec::de.tobiasroeser.mill.vcs.version_mill0.10.0-M2:0.1.2 show de.tobiasroeser.mill.vcs.version.VcsVersion/vcsState
[1/1] show 
[1/1] show > [1/1] de.tobiasroeser.mill.vcs.version.VcsVersion.vcsState 
{
    "currentRevision": "5f6538e3cb7191c0a8c778f9d47601967b744036",
    "lastTag": [
        "0.10.0-M3"
    ],
    "commitsSinceLastTag": 36,
    "dirtyHash": [

    ]
}
```

This PR also add support for empty versions in $ivy-imports, which will get expanded to the exact mill version. This makes consuming mill contrib plugins even more easy.

Example: Generating Bloop Config without modyfing the `build.sc`

```
mill --plugin ivy:com.lihaoyi::mill-contrib-bloop: mill.contrib.bloop.Bloop/install
```

### DIFF
--- a/docs/antora/modules/ROOT/pages/Contrib_Plugins.adoc
+++ b/docs/antora/modules/ROOT/pages/Contrib_Plugins.adoc
@@ -9,6 +9,7 @@ For details about including plugins in your `build.sc` read xref:Extending_Mill.
 --
 When using one of these contribution modules, it is important that the versions you load match your mill version.
 To facilitate this, Mill will automatically replace the `$MILL_VERSION` literal in your ivy imports with the correct value.
+You can also leave the version completely empty to default to the mill version (but don't forget to keep the trailing colon).
 
 For instance:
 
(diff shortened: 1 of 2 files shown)


lefou:
@lihaoyi I'm asking for a review or comment here. I'm myself a bit undecided. I think the `--plugin` options is great, but I'm not sure about its name, the syntax (especially the "ivy:" prefix to match meaning of `$ivy` so we can later support others too) and also the new feature of leaving out the version (affects also `import $ivy`) to substitute the current mill version. The motivation for the version was twofold: convenience in `build.sc` but also the fact it's hard to escape the used shell to provide a `$MILL_VERSION` in the command-line.

PR #1547, opened 2021-11-06
Initial work for scoverage 2.0

ckipp01:
[scoverage 2.x](https://github.com/scoverage/scalac-scoverage-plugin/releases/tag/v2.0.0-M2) makes a few changes to the structure of the scoverage project, including the artifacts, and also will allow for Scala 3 support once https://github.com/lampepfl/dotty/pull/13880 gets merged. I wasn't super familiar with how mill was using it, so this was just me learning a bit and updating some of the things I saw. However, it introduces a few questions that I'd love to get your thoughts on.

This is a breaking change that won't easily allow for users to use 1.x anymore. Are you ok with that? Once scoverage bumps to 2.x it will be pretty much impossible to support both 1.x and 2.x for integrations since the structure is quite different, even some of the package names, method signatures etc. In the sbt plugin once I officially bump I'll error out if someone manually tries to use 1.x in the 2.x series of the plugin. What are you thoughts on something like that in Mill?

I'll ask the other question in the parts of the code to make them clearer.

I'll leave this as a draft for now but will iterate on it since there is a few things we'll need to update if you're ok with the 2.x approach listed above.

- [ ] Add some messages to make it clear to the user that once on 2.x they need to use 2.x
- [ ] Add in the logic to determine if the user is using Scala 2 or 3. If 3, then just set the `-coverage` flag and the datadir
- [ ] Add some stuff to the tests to cover both the Scala 2 and Scala 3 usecase

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -107,7 +107,7 @@ object Deps {
   val scalafmtDynamic = ivy"org.scalameta::scalafmt-dynamic:3.0.6"
   val scalametaTrees = ivy"org.scalameta::trees:4.4.29"
   def scalaReflect(scalaVersion: String) = ivy"org.scala-lang:scala-reflect:${scalaVersion}"
-  def scalacScoveragePlugin = ivy"org.scoverage:::scalac-scoverage-plugin:1.4.9"
+  def scalacScoverageReporter = ivy"org.scoverage::scalac-scoverage-reporter:2.0.0-M2"
   val sourcecode = ivy"com.lihaoyi::sourcecode:0.2.7"
   val upickle = ivy"com.lihaoyi::upickle:1.4.2"
   val utest = ivy"com.lihaoyi::utest:0.7.10"
@@ -573,7 +573,7 @@ object contrib extends MillModule {
       override def compileIvyDeps = T{
         Agg(
           // compile-time only, need to provide the correct scoverage version runtime
-          Deps.scalacScoveragePlugin,
+          Deps.scalacScoverageReporter,
           // provided by mill runtime
           Deps.osLib
         )
--- a/contrib/scoverage/api/src/ScoverageReportWorkerApi.scala
+++ b/contrib/scoverage/api/src/ScoverageReportWorkerApi.scala
@@ -5,7 +5,12 @@ import mill.api.Ctx
 trait ScoverageReportWorkerApi {
   import ScoverageReportWorkerApi._
 
-  def report(reportType: ReportType, sources: Seq[os.Path], dataDirs: Seq[os.Path])(implicit
+  def report(
+      reportType: ReportType,
+      sources: Seq[os.Path],
+      dataDirs: Seq[os.Path],
+      sourceRoot: os.Path
+  )(implicit
       ctx: Ctx
   ): Unit
 }
--- a/contrib/scoverage/src/ScoverageModule.scala
+++ b/contrib/scoverage/src/ScoverageModule.scala
@@ -59,8 +59,12 @@ trait ScoverageModule extends ScalaModule { outer: ScalaModule =>
   def scoverageRuntimeDep: T[Dep] = T {
     ivy"org.scoverage::scalac-scoverage-runtime:${outer.scoverageVersion()}"
   }
-  def scoveragePluginDep: T[Dep] = T {
-    ivy"org.scoverage:::scalac-scoverage-plugin:${outer.scoverageVersion()}"
+  def scoveragePluginDep: T[Agg[Dep]] = T {
+    Agg(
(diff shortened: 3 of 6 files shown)


build.sc
```diff
@@ -573,7 +573,7 @@ object contrib extends MillModule {
       override def compileIvyDeps = T{
         Agg(
           // compile-time only, need to provide the correct scoverage version runtime
-          Deps.scalacScoveragePlugin,
+          Deps.scalacScoverageReporter,
```
ckipp01, line 576: I didn't fully understand this. Is this _only_ being used to pull the version? I switched this to the reporter just to test it and it worked as well, but probably won't leave it as the reporter.

lefou:
> This is a breaking change that won't easily allow for users to use 1.x anymore. Are you ok with that? Once scoverage bumps to 2.x it will be pretty much impossible to support both 1.x and 2.x for integrations since the structure is quite different, even some of the package names, method signatures etc. In the sbt plugin once I officially bump I'll error out if someone manually tries to use 1.x in the 2.x series of the plugin. What are you thoughts on something like that in Mill?

There is a major difference between the sbt plugin for Scoverage and this mill plugin. The sbt plugin is external to sbt and users can probably use the older versions supporting scoverage 1.x for a very long time. As our scoverage plugin is developed in tree, we in general only support compatibility to the exact same mill version. (This is different for plugins developed outside of the tree.)

So my question is, do we loose something with Scoverage 2.x? E.g. support for older Scala versions? If that should be the case (which I assume to be true), I'd say we should try a bit harder to keep support for Scoverage 1.x in the plugin. We already load scoverage jars in the worker, so it should be pretty easy to have some logic switch and use a different worker module depending on the actual used Scoverage version.

build.sc
lefou: The `contrib.scoverage.worker` is loaded by the `contrib.scoverage` module in a isolated classloader. The `contrib.scoverage.api` module is shared between both. So we can load the scoverage jars only when needed and isolated from other mill plugins. Technically, this allows us to use different scoverage versions in the same build. As Scoverage 2.x has an incompatible API, we need to create a new worker module working with Scoverage 2.x. We also do this in other plugin, e.g. `contrib.playlib`.

contrib/scoverage/api/src/ScoverageReportWorkerApi.scala
```diff
@@ -5,7 +5,12 @@ import mill.api.Ctx
 trait ScoverageReportWorkerApi {
   import ScoverageReportWorkerApi._
 
-  def report(reportType: ReportType, sources: Seq[os.Path], dataDirs: Seq[os.Path])(implicit
+  def report(
+      reportType: ReportType,
+      sources: Seq[os.Path],
+      dataDirs: Seq[os.Path],
+      sourceRoot: os.Path
+  )(implicit
```
lefou, line 13: As the worker API is shared between all worker implementations, we need to keep/support the `sourceRoot` parameter, in case we decide to keep the Scoverage 1.x worker.

ckipp01:
> So my question is, do we loose something with Scoverage 2.x? E.g. support for older Scala versions? 

Yea, this is exactly the case. In the 2.x series we'll no longer publish for 2.11.

lefou:
@ckipp01 Let me know if you need further guidance in creating a new worker module for Scoverage 2.x. Or just ping me in our Gitter channel.

ckipp01:
> @ckipp01 Let me know if you need further guidance in creating a new worker module for Scoverage 2.x. Or just ping me in our Gitter channel.

Sounds good. It'll probably be a bit before I get around to it. Once I get the green light on the direction in the pr to Dotty I'll probably swing back around to this, but I'll let you know.

kierendavies:
@ckipp01 Are you planning to get back to this? I may be interested in picking it up.

ckipp01:
> @ckipp01 Are you planning to get back to this? I may be interested in picking it up.

Hey @kierendavies! I was planning on it, but no idea if it'd be soon or not. Seeing that scoverage 2.0.0 is release, by all means pick this up if you'd like. Feel free to ping me to review the scoverage parts or anything and I'll be happy to give some insight. If you do pick it up, just let me know so we don't duplicate work.

ckipp01:
> Is anybody working on this? @ckipp01 or @kierendavies ?

Not me. I don't really have any plans in the short-term to get back to this.

lefou:
As there is nobody currently working on this and we decided to keep the old worker but also as there are merge conflicts in this PR, I decided to re-start this work in a new branch and pull request:

* #2010 

I heavily used this PR to identify the required changes. Thank you Chris (@ckipp01) for creating this PR!

Closing this PR as it is superseded by #2010.

PR #1569, opened 2021-11-15
Scala 2.13.7 update

lefou:
- Update Ammonite version to 2.4.1
- Update Scala version to 2.13.7 and added some MiMa filters
- Toolchain: Mill 0.9.10 and plugin updates
- No patch needed anymore


build.sc
```diff
@@ -1,12 +1,14 @@
 import $file.ci.shared
 import $file.ci.upload
 import $ivy.`org.scalaj::scalaj-http:2.4.2`
-import $ivy.`de.tototec::de.tobiasroeser.mill.vcs.version_mill0.9:0.1.2`
-import $ivy.`com.github.lolgab::mill-mima_mill0.9:0.0.5`
+import $ivy.`de.tototec::de.tobiasroeser.mill.vcs.version::0.1.2-4-dcde72`
+import $ivy.`com.github.lolgab::mill-mima::0.0.6`
```
lolgab, line 5: The problem seems to be caused by the interaction between mill-mima and the `::` syntax.
You can try to add `_mill0.9` manually and see that the binary compatibility step succeeds.
```suggestion
import $ivy.`com.github.lolgab::mill-mima_mill0.9:0.0.6`
```

lefou: As shown in the other issue, this isn't necessarily the root problem. My suspicion is it's something classloading related. MiMa gets loaded directly into the ammonite classloader (it's not using a worker or dedicated classloaded) and I've see multiple unexpected jars been pulled in by Ammonite via `$ivy- imports in the past, even though they were already loaded or loaded in different versions. And MiMa depends on Scala 2.13.6. It's very weird.

lolgab: If this is the problem, using a worker with a separate classloader that doesn't depend on mill would solve it.
I'll try to make this change to mill-mima.

lefou: I'm not sure this is the issue here. Actually, I do think that it's something different. I already commented in the Bug report. This PR is already merged with a workaround.

PR #1570, opened 2021-11-18
Add installation instructions for Gentoo Linux

Flowdalic:


lefou:
Yeah, feel free to open a pull request. 

Besides, I think any Gentoo user should have figured it out by herself, as it's regular procedure:

```
$ eix mill
* dev-java/mill-bin
     Verfügbare Versionen:   ~0.9.9
     Startseite:             https://com-lihaoyi.github.io/mill/
     Beschreibung:           A Java/Scala build tool
...
$ emerge mill-bin
....
```

Also, I think this one installs a fixed version (which is perfectly in the spirit of Gentoo, which I'm using too, by the way), whereas most developers prefer to have a dynamic version, which means: Use whatever mill version is specified in `.mill-version`. This will likely require to use a [script like this](https://github.com/com-lihaoyi/mill/blob/main/mill) or [lefou/millw](https://github.com/lefou/millw).

lefou:
Thanks!

PR #1571, opened 2021-11-19
Doc: Add a special case for `zsh`

Renkai:


lefou (commented):
Great, thank you!

docs/antora/modules/ROOT/pages/Cross_Builds.adoc
```diff
@@ -29,6 +29,15 @@ mill show foo[2.12].suffix
...
 
+If you were a `zsh` user, the command should be a little complex.
+
+[source,zsh]
+----
+mill show foo\[2.10\].suffix
+mill show 'foo[2.10].suffix'
+mill show "foo[2.10].suffix"
+----
+
```
lefou, line 40: I suggest to move this into an extra info box.

```suggestion
[NOTE]
--
Please be aware that some shells like `zsh` interpret square brackets differently, so quoting or masking might be needed.

[source,zsh]
----
mill show foo\[2.10\].suffix
mill show 'foo[2.10].suffix'
mill show "foo[2.10].suffix"
----
--
```

PR #1573, opened 2021-11-21
Experimental support for scalajs module splitting

pbuszka:
new module tasks
- fastLinkJS
- fullLinkJS

supported use case:
generate multiple modules to be consumed by external js bundler like webpack, vite or nodejs. Key use case is to be able to build multiple modules and dynamicImport for frontend apps load time optimization.

supported configuration options:
- ModuelKind ESModule only, closure compiler used in CommonJSModule does not support multiple modules and fails in fullLinkJS
- ModuleSplitStyle
- ModuleInitializers
- OutputPatterns

current test coverage:
ESModule, ModuleSplitStyle, OutputPatterns, ModuleInitializers actual usage
- scala version 2.13, 2.12, 2.11 
- scala 3 tests commented out
- scala js version 1.7.1, 14.0, 1.3.1

generation of 2 modules with a ModuleInitializer and top level js module export in each of the modules

unsupported mill functionality at the moment
- test (uses fastOpt by default)
- run (uses fastOpt by default)
- main method definition support
- docs

open questions / unknowns / todo:
- should a flag at module level  (i.e. useLegacyJSLinker: T[Boolean]) be introduced for test and run tasks to use fastLinkJS instead. For scalajs >= 1.3 both fastOpt and fastLinkJS should be usable
- how to suport main method run, running `node main.mjs` does not invoke ModuleInitializer of the main module (from my early tests)
- support CommonJsModule case and drop closure compiler usage when multiple modules are involved (and effectively delegate the opt problem downstream to JS ecosystem toooling)


pbuszka:
Mima checks fail with:
```Found 1 issue when checking against com.lihaoyi:mill-scalajslib-api_2.13:0.10.0-M4
 * abstract method linkJs(Array[java.io.File],Array[java.io.File],java.io.File,scala.Option,Boolean,Boolean,mill.scalajslib.api.ModuleKind,mill.scalajslib.api.ModuleSplitStyle,scala.collection.immutable.Seq,mill.scalajslib.api.OutputPatterns,Boolean)mill.api.Result in interface mill.scalajslib.api.ScalaJSWorkerApi is present only in current version
   filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("mill.scalajslib.api.ScalaJSWorkerApi.linkJs")```
   
Does it mean I cannot extend mill.scalajslib.api.ScalaJSWorkerApi trait with a new method ? or it is just for safety to avoid accidental change

**Comment by pbuszka** [comment]
Windows test fails probably on test CR/LF v/s CR on Linux when testing output of a test program

**Comment by pbuszka** [comment]
SmallestModules is inefficient when doing dev cycle. I followed the logic from fastOpt and delete output before start but in this case whole dir is deleted beforehand.

I don't know the working of the scala-js linker  but wonder maybe output dir should be deleted only if a new linker is created. On linker reuse leave the previous output intact and hope for the best :)

**Comment by pbuszka** [comment]
I tested it in my project and seems to work as expected except the SmallestModules inefficiency

@lolgab Could you have a look at the PR and questions in the comments

File: scalajslib/src/ScalaJSModule.scala [hunk 1]
```diff
@@ -99,6 +99,50 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
     )
   }
 
+  private def isLinkJSSupported = T {
+    scalaJSVersion() match {
+      case v if v.startsWith("0.") => false
+      case v if v < "1.3" => false
+      case _ => true
+    }
+  }
```
lolgab, line 108:
```suggestion
  private def isLinkJSSupported = T {
    scalaJSVersion() >= "1.3"
  }
```

Also try if `T.task` work, since you don't need to save this to a file.

scalajslib/test/src/HelloSplitJSWorldTests.scala
```diff
@@ -0,0 +1,362 @@
...
+      if !skipScalaJS(scalaJS)
+      if !(skipECMAScript2015 && useECMAScript2015)
+    } {
+      if (scala.startsWith("2.11.")) {
+        TestUtil.disableInJava9OrAbove(f(scala, scalaJS, useECMAScript2015))
+      } else {
+        f(scala, scalaJS, useECMAScript2015)
+      }
+    }
+  }
```
lolgab, line 360: There is a lot of copy paste in this test. Can you add a minimal test for fastLinkJS avoiding to copy paste too much?

scalajslib/worker/0.6/src/ScalaJSWorkerImpl.scala
```diff
@@ -86,6 +93,20 @@ class ScalaJSWorkerImpl extends mill.scalajslib.api.ScalaJSWorkerApi {
...
+      destDirectory: File,
+      main: Option[String],
+      testBridgeInit: Boolean,
+      fullOpt: Boolean,
+      moduleKind: ModuleKind,
+      moduleSplitStyle: ModuleSplitStyle,
+      moduleInitializers: Seq[ModuleInitializer],
+      outputPatterns: OutputPatterns,
+      useECMAScript2015: Boolean
+  ): Result[File] = ???
```
lolgab, line 108: Why don't you move the `mill.api.Result.Failure("fastLinkJS/fullLinkJS is not supported in Scala.js below 1.3")` here instread of checking for the version earlier and then implement this with `???`.
This way you don't need to define any `isLinkJSSupported` target.

scalajslib/worker/1/src/ScalaJSWorkerImpl.scala
```diff
@@ -80,11 +91,11 @@ class ScalaJSWorkerImpl extends mill.scalajslib.api.ScalaJSWorkerApi {
       .withSourceMapURI(java.net.URI.create(sourceMap.getFileName.toString))
     val logger = new ScalaConsoleLogger
     val mainInitializer = Option(main).map { cls =>
-      ModuleInitializer.mainMethodWithArgs(cls, "main")
+      ScalaJSModuleInitializer.mainMethodWithArgs(cls, "main")
```
lolgab, line 94: Since you're now supporting `ModuleInitializer` in the frontend, I think you should use the value set by the user even in `fastOpt`, not only in `fastLinkJS`.
What do you think?

scalajslib/worker/0.6/src/ScalaJSWorkerImpl.scala
lolgab: Also, if you want to check the version in the `1` linker, the linker version is availably in the scalajs linker lib by calling:
```scala
org.scalajs.ir.ScalaJSVersions.current
```
I just tested and it displays correctly the used version.

lolgab:
Overall it looks good to me.
I think you can try to factor out the duplication in tests and in the worker implementation (most of the things are similar between `fastOpt` and `fastLinkJS`).
You can access the ScalaJS version via `org.scalajs.ir.ScalaJSVersions.current`. Probably leveraging that will make your implementation simpler (as I wrote in the code review comments).

lolgab (changes requested):

scalajslib/src/ScalaJSModule.scala
lolgab: After checking that you can use `org.scalajs.ir.ScalaJSVersions.current` inside the worker, now I think you probably don't need to make this check here, so you can call the worker directly.

scalajslib/api/src/ScalaJSWorkerApi.scala
```diff
@@ -36,6 +51,33 @@ object ModuleKind {
...
+case class ModuleInitializer(className: String, mainMethod: String, moduleId: String)
+
+sealed trait OutputPatterns
+
+object OutputPatterns {
+  import upickle.default.{ReadWriter => RW, macroRW}
+  implicit def rwFromJsFile: RW[OutputPatternsFromJsFile] = macroRW
+  implicit def rw: RW[OutputPatterns] = macroRW
+
+  def fromJSFile(jsFile: String): OutputPatterns = OutputPatternsFromJsFile(jsFile)
```
lolgab, line 74: Since this is internal API, we can probably use the case class apply directly and avoid defining helpers.

scalajslib/api/src/ScalaJSWorkerApi.scala
```diff
@@ -36,6 +51,33 @@ object ModuleKind {
...
+}
+
+case object OutputPatternsDefaults extends OutputPatterns
+
+case class OutputPatternsFromJsFile(jsFile: String) extends OutputPatterns
```
lolgab, line 79: Also, in Mill there is a convention to put the values of a sealed trait in the companion object of the trait itself.
Can you do the same here?

pbuszka:
@lolgab I'm sorry for lack any of progress I got ill and was completely unable to do anything useful. Hope to get to it in a few days as I'm getting better now

pbuszka:
moved to https://github.com/com-lihaoyi/mill/pull/1642 as I messed up with a rebase in my repo and didn't know how to fix it without creating even more mess

Issue #860, opened 2020-05-06
Build fails when using scala ivydep in scalamodule with java moduledep


nrktkt:
Given

```scala
// build.sc
import mill._, scalalib._
object core extends JavaModule {
  object scalaapi extends ScalaModule {
    def scalaVersion = "2.13.1"
    def moduleDeps = Seq(core)
    def ivyDeps = Agg(ivy"org.typelevel::cats-core:2.1.1")
  }
  object javaapi extends JavaModule {
    def moduleDeps = Seq(core, scalaapi)
  }
}
```
attempting to compile will yield 
```
[42/57] core.scalaapi.compile 
[info] Compiling 6 Scala sources to /home/nfischer/Code/2Auth/out/core/scalaapi/compile/dest/classes ...
[info] Done compiling.
[55/57] core.javaapi.compileClasspath 
1 targets failed
core.javaapi.compileClasspath java.lang.AssertionError: assertion failed: Not a Java dependency: Dep(Dependency(org.typelevel:cats-core, 2.1.1, Configuration(default(compile)), Set(), Publication(, Type(), Extension(), Classifier()), false, true),Binary(false),false)
    scala.Predef$.assert(Predef.scala:223)
    mill.scalalib.Lib$.depToDependencyJava(Lib.scala:24)
    mill.scalalib.CoursierModule.$anonfun$resolveCoursierDependency$3(CoursierModule.scala:17)
    mill.scalalib.CoursierModule.$anonfun$resolveDeps$2(CoursierModule.scala:30)
    scala.collection.immutable.Stream.map(Stream.scala:418)
    mill.scalalib.Lib$.resolveDependencies(Lib.scala:65)
    mill.scalalib.CoursierModule.$anonfun$resolveDeps$1(CoursierModule.scala:27)
    mill.define.ApplyerGenerated.$anonfun$zipMap$2(ApplicativeGenerated.scala:7)
    mill.define.Task$MappedDest.evaluate(Task.scala:380)
```
on 0.6.2

lefou:
On a first glance, it looks like when resolving the ivy deps of `javaapi` we also resolve all transitive deps from other modules, which include also scala dependencies. But in the scope of the `JavaModule` we no longer know the scala version to use.

lefou:
2 workaround come to my mind:
* make the `javaapi` project a `ScalaModule` and adapt the `artifactSuffix` to be empty
[...]

PR #1574, opened 2021-11-21
Bind cross-versions to the context of the originating module

lefou:
`mill.scalalib.Dep` represents a generic dependency, which include potential cross-compilation properties. E.g. it knows whether it is specific to a `platform` but it does not specify the platform. And it knows whether it is specific to a Scala binary or full version, but it does not specify the exact Scala version. This mean we can define a `Dep` without knowing those context. 

In this PR I introduce a new type `mill.scalalib.BoundDep`, which is kind-of a `Dep` where all these potential open context bindings are bound to an exact context. There is no further cross-resolution necessary nor possible. By having a separate type, we always know with what kind of dependency we work. Also, we can be explicit about the "when" the binding happens. 

When mixing dependencies of multiple modules, we also mix contexts. Until this PR, we had no way to get it right, as `Dep`s could come from different contexts. Now, we can safely mix `BoundDep`s as they are already context-bound.

We could encode the bound-state in the `Dep`, but this would result in only a dynamic property. To ensure a resolved state, we always would need to run some runtime checks. Also, it could lead to situations, where we need a resolved dep, but have not the proper context to do the resolution. This is exactly what I implemented in the initial attempt, but it didn't worked out very well. With the separated types, it is always clear from the API, what kind of dependency we expect.

We could also just resolve to a `coursier.Dependency`, but that would loose the `force` flag. Also, we want to rather decouple our API from the coursier API, which seem easier when we avoid to hand over direct coursier API to the user.

I `@deprecated` the existing `CoursierModule.resolveCoursierDependency` and bases the new `CoursierModule.bindDependency` on this old task. This should ease the transition. It's up for revisit, when we decouple from coursier API altogether.

This is an attempt to fix https://github.com/com-lihaoyi/mill/issues/860.

### DIFF
--- a/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
@@ -378,8 +378,7 @@ class BloopImpl(ev: () => Evaluator, wd: os.Path) extends ExternalModule { outer
       val repos = module.repositoriesTask()
       // same as input of resolvedIvyDeps
       val allIvyDeps = module.transitiveIvyDeps() ++ module.transitiveCompileIvyDeps()
-      val coursierDeps =
-        allIvyDeps.map(module.resolveCoursierDependency()).toList
+      val coursierDeps = allIvyDeps.map(_.dep).toList
       BloopConfig.Resolution(artifacts(repos, coursierDeps))
     }
 
--- a/contrib/flyway/src/FlywayModule.scala
+++ b/contrib/flyway/src/FlywayModule.scala
@@ -29,7 +29,10 @@ trait FlywayModule extends JavaModule {
   def flywayDriverDeps: T[Agg[Dep]]
 
   def jdbcClasspath = T {
-    resolveDeps(flywayDriverDeps)()
+    resolveDeps(T.task {
+      val bind = bindDependency()
+      flywayDriverDeps().map(bind)
+    })()
   }
 
   private def strToOptPair[A](key: String, v: String) =
--- a/contrib/jmh/src/mill/contrib/jmh/JmhModule.scala
+++ b/contrib/jmh/src/mill/contrib/jmh/JmhModule.scala
@@ -94,8 +94,12 @@ trait JmhModule extends JavaModule {
       (sourcesDir, resourcesDir)
     }
 
-  def generatorDeps =
+  def generatorDeps = T {
     resolveDeps(
-      T { Agg(ivy"org.openjdk.jmh:jmh-generator-bytecode:${jmhGeneratorByteCodeVersion()}") }
-    )
+      T.task {
+        val bind = bindDependency()
+        Agg(ivy"org.openjdk.jmh:jmh-generator-bytecode:${jmhGeneratorByteCodeVersion()}").map(bind)
+      }
+    )()
+  }
 }
--- a/contrib/playlib/src/mill/playlib/RouterModule.scala
+++ b/contrib/playlib/src/mill/playlib/RouterModule.scala
@@ -45,7 +45,10 @@ trait RouterModule extends ScalaModule with Version {
   def generatorType: RouteCompilerType = RouteCompilerType.InjectedGenerator
 
   def routerClasspath: T[Agg[PathRef]] = T {
-    resolveDeps(T.task { Agg(ivy"com.typesafe.play::routes-compiler:${playVersion()}") })()
+    resolveDeps(T.task {
+      val bind = bindDependency()
+      Agg(ivy"com.typesafe.play::routes-compiler:${playVersion()}").map(bind)
+    })()
   }
 
   protected val routeCompilerWorker: RouteCompilerWorkerModule = RouteCompilerWorkerModule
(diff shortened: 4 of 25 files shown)


lolgab:
This is great! Didn't expect the solution to be so simple!
Congrats 👏
Now a test exercising this would be great. For example a Scala 2.13 module that depends on a Scala 3 project using upickle macros but running the macros in the 3 side and exposing a macro free api, so we check that upickle_3 is resolved instead of `_2.13`

lolgab:
This is great! Didn't expect the solution to be so simple!
Congrats 👏
Now a test exercising this would be great. For example a Scala 2.13 module that depends on a Scala 3 project using upickle macros but running the macros in the 3 side and exposing a macro free api, so we check that `upickle_3` is resolved instead of `_2.13`

lefou:
> Now a test exercising this would be great. For example a Scala 2.13 module that depends on a Scala 3 project using upickle macros but running the macros in the 3 side and exposing a macro free api, so we check that `upickle_3` is resolved instead of `_2.13`

Yeah, will definitely add a test once I find the time for it. If you already have some example code, I wouldn't decline it. ;-)

lefou:
I feel like having a knot in my mind, so I though I gather your input, @lihaoyi @lolgab .

I have the following test case lying around for some time.

```scala
object HelloScala3 extends ScalaModule {
  def scalaVersion = "3.0.2"
  override def ivyDeps = T { Agg(ivy"com.lihaoyi::upickle:1.4.0") }
}

object HelloScala213 extends ScalaModule {
  def scalaVersion = "2.13.7"
  override def moduleDeps = Seq(HelloScala3)
  override def ivyDeps = T { Agg(ivy"com.lihaoyi::sourcecode:0.2.7") }
  override def scalacOptions = T { Seq("-Ytasty-reader") }
}

object HelloJava extends JavaModule {
    override def moduleDeps = Seq(HelloScala213)
    override def ivyDeps = T { Agg(ivy"org.slf4j:slf4j-api:1.2.6") }
}
```

What exactly should the correct `ivyDepsTree` output look like for each of the modules? 

* `mill HelloScala3.ivyDepsTree` should be trivial
* `mill HelloScala213.ivyDepsTree` does need to translate the different scala versions accordingly
* `mill HelloJava.ivyDepsTree` should be able to resolve all deps without knowing which Scala version is used, right?

lolgab:
@lefou I read the test case and I have the same expectations as you. If module `A` depends on module `B` all dependencies of `A` become transitive dependencies of `B`.
In the example `HelloJava` should have:
- `com.lihaoyi:upickle_3:1.4.0`
- `org.scala-lang:scala3-library_3:3.0.2`
- `org.scala-lang:scala-library:2.13.7` (which evicts scala-library:2.13.6 required by `scala3-library_3:3.0.2`)
- `com.lihaoyi:sourcecode_3:0.2.7` ( + its transitive dependencies )
- `org.slf4j:slf4j-api:1.2.6`

lolgab:
Why do we need to a add a new type of `Dep`? Why can't `Dep` itself be what `BoundDep` is?

lefou:
> Why do we need to a add a new type of `Dep`? Why can't `Dep` itself be what `BoundDep` is?

`Dep` represents a generic dependency, which include potential cross-compilation properties. E.g. it knows whether it is specific to a `platform` but it does not specify the platform. And it knows whether it is specific to a Scala binary or full version, but it does not specify the exact Scala version. This mean we can define a `Dep` without knowing those context. 

`BoundDep` is a `Dep` where all these potential open context bindings are bound to an exact context. There is no further resolution necessary nor possible. 

By having a separate type, we always know with what kind of dependency we work. Also, we can be explicit about the "when" the binding happens. 

We could encode the bound-state in the `Dep`, but this would result in only a dynamic property. To ensure a resolved state, we always would need to run some runtime checks. Also, it could lead to situations, where we need a resolved dep, but have not the proper context to do the resolution. This is exactly what I implemented in the initial attempt. With the separated types, it's is always clear from the API, what kind of dependency we expect.

lefou:
I'd be glad to get some feedback on this pull request. @lolgab @lihaoyi @nrktkt ?

I think is a great step into more robust dependencies. Especially as as transitive inter-module dependencies are now more like transitive dependencies from outside the project. But I'm also sure, we will find some corner cases which we want to adapt further. 

Although it's a incompatible API change, I think having a dedicated type makes sense here. (see also the comments above) Also, a user is probably not affected by the changed API so often.

nrktkt (commented):
This seems like a reasonable way to go about it. I'm glad to have this capability going forward

michelou:
@lefou Why did you introduce the new package `mil.scalalib` for `BoundDep.scala` (instead of using existing `mill.scalalib`) ? Do you e.g. consider that feature as still experimental ?

lefou:
> @lefou Why did you introduce the new package `mil.scalalib` for `BoundDep.scala` (instead of using existing `mill.scalalib`) ? Do you e.g. consider that feature as still experimental ?

Thanks for pointing it out. This is a typo. The correct full qualified name is `mill.scalalib.BoundDep` as stated in the PR description.

lefou:
Fixed in https://github.com/com-lihaoyi/mill/pull/2214

PR #1587, opened 2021-11-24
Fix URL to file path conversion for GenIdea command

yakivy:
PR fixes artifacts resolving from private repos for GenIdea command.

**Problem description**:
Coursier caches private repos under the `.../user@organization/...` path, for unix systems it translates to `.../user%40organization/...` path, and in turn `java.net.URL` translates it to `.../user%2540organization/...`, so when GenIdea constructs `buildDepsPaths` with OS independent `java.net.URL#getFile` call, double escaping breaks pathes for private repos.

**Fix description**:
OS idependent URL to file path conversion `java.net.URL#getFile` was replaced with OS dependent `java.nio.file.Paths#get(java.net.URI)`


lefou (approved):
Great PR description. Thanks for this PR! LGTM.

Issue #1576, opened 2021-11-23
Improve offline-support


lefou:
Mill already has limited offline support. Each module implementing `OfflineSupportModule` has a `def prepareOffline(): Command[Unit]` command to pre-fetch as much as possible. This include also dependency which are rarely needed and in some cases even not properly configured yet (prominent example: Setups with rather new Scala version but default Ammonite version will probably fail when trying to download).

I suggest to enhance the command to:

```scala
def prepareOffline(hint: String*) Command[Unit]
```

By default, we should only fetch dependencies required for day to day work.

If `hint` includes something like `"all"` or "`full`", we could fetch everything. The idea is, to provide a mechanism to leave it to the implementing module what to fetch by default. This is of course untyped, but my hope it, that  is still easy enough to match for strings in the implementation. All supported hints should be added to the ScalaDoc of the command, so that a user can easily get an overview with `mill inspect foo.prepareOffine`.

lefou:
Fixed, but solution is slighly simpler with just a boolean `all` flag.

PR #1591, opened 2021-11-28
Let prepareOffline accept a `all` flag

lefou:
This should support more control in inheriting modules.

Also, fetch only essential dependencies in Java and Scala modules,
and keep the full fetch behind a `all` flag, which can be set as parameter.

This is a binary-incompatible change and I have no idea how to preserve compatibility.
Although I could keep the old signature and just forward to the newer,
Loading plugins that also inherit older `OfflineSupportModule`s seem to fail nevertheless.
Also I think Mill isn't checking for multiple commands with the same name, 
so I fear just because it works on my machine, it may still lead to some failures for others.
Only fix would be to invent a new name for the command (and deprecate the old).
As Mill 0.10.x is binary-incompatible anyways, I think this one should be ok.

Fix https://github.com/com-lihaoyi/mill/issues/1576

### DIFF
--- a/scalalib/src/JavaModule.scala
+++ b/scalalib/src/JavaModule.scala
@@ -772,13 +772,32 @@ trait JavaModule
 
   def forkWorkingDir: Target[Path] = T { os.pwd }
 
+  /**
+   * @param hint "all" - fetch all dependencies,
+   *             "source" | "sources" - fetch source dependencies
+   */
   @nowarn("msg=pure expression does nothing")
-  override def prepareOffline(): Command[Unit] = T.command {
-    super.prepareOffline()()
-    resolvedIvyDeps()
-    zincWorker.prepareOffline()()
-    resolvedRunIvyDeps()
-    ()
+  override def prepareOffline(hint: String*): Command[Unit] = {
+    val withSources = hint.contains("all") || hint.contains("source") || hint.contains("sources")
+    val tasks = Seq(
+      if (withSources) Seq(
+        resolveDeps(
+          T.task {
+            transitiveCompileIvyDeps() ++ transitiveIvyDeps()
+          },
+          sources = true
+        )
+      )
+      else Seq()
+    ).flatten
+    T.command {
+      super.prepareOffline(hint: _*)()
+      resolvedIvyDeps()
+      zincWorker.prepareOffline()()
+      resolvedRunIvyDeps()
+      T.sequence(tasks)()
+      ()
+    }
   }
 
   @internal
--- a/scalalib/src/JavaModule.scala
+++ b/scalalib/src/JavaModule.scala
@@ -2,7 +2,9 @@ package mill
 package scalalib
 
 import scala.annotation.nowarn
+
 import coursier.Repository
+import mainargs.Flag
 import mill.api.Loose.Agg
 import mill.api.{PathRef, Result, internal}
 import mill.define.{Command, Sources, Target, Task, TaskModule}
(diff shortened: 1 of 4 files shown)


lefou:
Any opinion about the solution and the incompatible change? @lolgab

scalalib/src/JavaModule.scala
```diff
@@ -772,13 +772,32 @@ trait JavaModule
...
+   *             "source" | "sources" - fetch source dependencies
+   */
   @nowarn("msg=pure expression does nothing")
-  override def prepareOffline(): Command[Unit] = T.command {
-    super.prepareOffline()()
-    resolvedIvyDeps()
-    zincWorker.prepareOffline()()
-    resolvedRunIvyDeps()
-    ()
+  override def prepareOffline(hint: String*): Command[Unit] = {
```
lihaoyi, line 780: This signature is kind of weird

- Do we need both `source` and `sources` as options? 
- Shouldn't passing in all three values at the same time e.g. `./mill foo.prepareOffline source sources all` throw an error?
- Should we use a flag `./mill foo.prepareOffline --all` `./mill foo.prepareOffline --sources`? 
- Or even just separate targets `./mill foo.prepareOffline` `./mill foo.prepareOfflineWithSources`?

lefou: > This signature is kind of weird
> 
>* Do we need both `source` and `sources` as options?

No, either `source` or `sources` is enough.

>* Shouldn't passing in all three values at the same time e.g. `./mill foo.prepareOffline source sources all` throw an error?

I think not. 

> * Presumably invalid strings like `./mill foo.prepareOffline invalid-flag` should be an error too

Imagine a project with multiple different modules. If we reject all hints we don't support, we can never let other modules support new hints. E.g. To run `mill __.prepareOffline foo`  where hint is special to `BarModule` but only some modules are implemented with `extends BarModule` and all other modules should do default offline preparations.

>* Should we use a flag `./mill foo.prepareOffline --all` `./mill foo.prepareOffline --sources`?

That would be clearly the better choice, but I don't see how to implement it in Mill.

>* Or even just separate targets `./mill foo.prepareOffline` `./mill foo.prepareOfflineWithSources`?

Yeah, this would be the natural way to make it right, but then we don't need the `OfflineSupportModule` at all.  It would result in many `prepareOffline*` targets and we don't have any support to call them efficiently. In the end, the user needs to learn many different targets, but not a single one.

lefou:
Maybe, a cleaner but less flexible approach would be to just add a boolean `all` parameter. So we have only a binary option: resolve some or resolve all.

scalalib/src/JavaModule.scala
lihaoyi: I think I see what you're trying to do with the hints. It's still rather odd though; they're kind of separate targets but not really, so there's no way to inspect them or discover them. `./mill resolve foo._` won't show you which hints are valid, and as you said invalid hints won't throw an exception. Overall doesn't seem like a good setup to me, apart from not really fitting into how any of the other mill modules/targets/commands work.

Here's one idea: what if we made each module have an `object prepareOffline extends Module{}`, in which we would put whatever `def all = T{...}` `def sources = T{...}` targets we would like? Then people would select what they want via `./mill foo.prepareOffline.all`, or even across multiple targets with multiple kinds of hint via `./mill __.prepareOffline.{all,sources}`, while still letting people do things like `./mill resolve foo.prepareOffline._` to see what's available.

If we're worried about the `object prepareOffline` making it hard to override the tasks, we could do something like 

```scala
val prepareOffline: PrepareOfflineModule = new PrepareOfflineModule{}
trait PrepareOfflineModule extends Module{...}
``` 

to allow subclasses to subclass `trait PrepareOfflineModule` and override `val prepareOffline` if necessary

lefou:
Then, it's up to the those modules with special needs to provide additional targets

lihaoyi:
Yeah I suppose it depends on how much flexibility we expect to need; either a small set of flags if we don't expect to need to add much more, or the `PrepareOfflineModule` approach I described above if we expect to have lots of different things we'd want to call under `foo.prepareOffline.xyz`.

scalalib/src/JavaModule.scala
lefou: Nice trick, but I honestly think, this is overkill. I tend to just introduce a binary `all` flag to the target. If we need more, we have to introduce new targets though, but that true for your suggestion as well.

lihaoyi (approved):

PR #1593, opened 2021-11-29
Don't add semanticdb plugin for Scala 3.

ckipp01:
For Scala 3 semanticdb is produced natively by the compiler behind the
`-Xsemanticdb` flag, so the plugin isn't necessary. This change ensures
the correct scalacOptions are set when using Scala 3 with
ScalaMetalsSupport.

### DIFF
--- a/scalalib/src/mill/scalalib/bsp/ScalaMetalsSupport.scala
+++ b/scalalib/src/mill/scalalib/bsp/ScalaMetalsSupport.scala
@@ -3,23 +3,46 @@ package mill.scalalib.bsp
 import mill.api.experimental
 import mill.{Agg, T}
 import mill.define.Target
+import mill.scalalib.api.ZincWorkerUtil.isScala3
 import mill.scalalib.{Dep, DepSyntax, ScalaModule}
+import mill.api.Result
 
 /*+ Enable some common settings required to properly support Metals Language Server (via BSP). */
 @experimental
 trait ScalaMetalsSupport extends ScalaModule {
 
-  /** The semanticDB version to use. It needs to support your configured Scala versions. */
-  def semanticDbVersion: T[String]
+  /** The semanticDB version to use. It needs to support your configured Scala 2 version. */
+  def semanticDbVersion: T[String] = T { "" }
+
   override def scalacPluginIvyDeps: Target[Agg[Dep]] = T {
-    super.scalacPluginIvyDeps() ++ Agg(
-      ivy"org.scalameta:::semanticdb-scalac:${semanticDbVersion()}"
-    )
+    if (!isScala3(scalaVersion()) && semanticDbVersion().isEmpty) {
+      val msg =
+        """|
+           |When using ScalaMetalsSupport with Scala 2 you must provide a semanticDbVersion
+           |
+           |def semanticDbVersion = ???
+           |""".stripMargin
+      Result.Failure(msg)
+    } else if (isScala3(scalaVersion())) {
+      Result.Success(super.scalacPluginIvyDeps())
+    } else {
+      Result.Success(
+        super.scalacPluginIvyDeps() ++ Agg(
+          ivy"org.scalameta:::semanticdb-scalac:${semanticDbVersion()}"
+        )
+      )
+    }
   }
 
   /** Adds some options and configures the semanticDB plugin. */
   override def mandatoryScalacOptions: Target[Seq[String]] = T {
-    super.mandatoryScalacOptions() ++ Seq("-Yrangepos", s"-P:semanticdb:sourceroot:${T.workspace}")
+    super.mandatoryScalacOptions() ++ {
+      if (isScala3(scalaVersion())) {
+        Seq("-Xsemanticdb")
+      } else {
+        Seq("-Yrangepos", s"-P:semanticdb:sourceroot:${T.workspace}")
+      }
+    }
   }
 
   /** Filters options unsupported by Metals. */


scalalib/src/mill/scalalib/bsp/ScalaMetalsSupport.scala
```diff
@@ -3,23 +3,37 @@ package mill.scalalib.bsp
...
+import mill.scalalib.api.Util.isScala3
 import mill.scalalib.{Dep, DepSyntax, ScalaModule}
 
 /*+ Enable some common settings required to properly support Metals Language Server (via BSP). */
 @experimental
 trait ScalaMetalsSupport extends ScalaModule {
 
   /** The semanticDB version to use. It needs to support your configured Scala versions. */
-  def semanticDbVersion: T[String]
+  def semanticDbVersion: T[String] = T { "4.4.30" }
```
ckipp01, line 14: So I'm a little unsure what to do here. Making the user fill this in makes sense _only_ for Scala 2 and I'm afraid that making it mandatory when doing nothing for Scala 3 will be confusing. Since most people will look at Scalameta to get the latest published version for Scala 3 only to realize it's not published since the compiler produces it natively in Scala 3.

sbt for example gives it a default, but the only negative part about this is that it needs to be available for the version of Scala you're using. So if 2.13.8 comes out tomorrow, someone quickly changes their version, then indeed they'll need to set it to the newest until mill is released with a newer default.

What are your thoughts on this?

scalalib/src/mill/scalalib/bsp/ScalaMetalsSupport.scala
```diff
@@ -3,23 +3,37 @@ package mill.scalalib.bsp
...
+      }
+    }
   }
 
   /** Adds some options and configures the semanticDB plugin. */
   override def mandatoryScalacOptions: Target[Seq[String]] = T {
-    super.mandatoryScalacOptions() ++ Seq("-Yrangepos", s"-P:semanticdb:sourceroot:${T.workspace}")
+    super.mandatoryScalacOptions() ++ {
+      if (isScala3(scalaVersion())) {
+        Seq("-Xsemanticdb")
```
ckipp01, line 32: There is also a `-semanticdb-target` that can be set for Scala 3 if we want more fine grained control over where the semanticdb is placed, but the default seems to work here.

scalalib/src/mill/scalalib/bsp/ScalaMetalsSupport.scala
lefou: I try to keep out default versions as much as possible. Beside the cons you already mentioned, they typically don't age well and result in future maintenance burden. Eventually Mill is stable enough so that it's not necessary to upgrade, but that means versions will stay old, in most cases in disadvantage of the user. Also, once a setting has a default, it's hard to remove it without breakage. My rule of thumb is to avoid default versions.

As this target isn't useful for Scala 3, we either make it optional by type and provide a default `None`, or we could provide an empty String and fail with an explicit error in case it was not overridden but we are on Scala 2.

scalalib/src/mill/scalalib/bsp/ScalaMetalsSupport.scala
lefou: Can't say much, as I'm not very familiar with semanticdb details, but as we configure it explicitly for Scala 2, we might also do it for Scala 3 to have more consistency. Without any recommendation from semanticdb-side, it's a 50-50 chance we choose the right strategy. :-p

scalalib/src/mill/scalalib/bsp/ScalaMetalsSupport.scala
ckipp01: Alright I went the `""` route.

scalalib/src/mill/scalalib/bsp/ScalaMetalsSupport.scala
ckipp01: Seeing that this is "ScalaMetalsSupport" and it works with Metals I think it should probably be good with the default. If we find that it does't work I'll come back and address it.

scalalib/src/mill/scalalib/bsp/ScalaMetalsSupport.scala
```diff
@@ -3,23 +3,44 @@ package mill.scalalib.bsp
 import mill.api.experimental
 import mill.{Agg, T}
 import mill.define.Target
+import mill.scalalib.api.Util.isScala3
```
lefou, line 6: This one is deprecated. Please use `mill.scalalib.api.ZincWorkerUtil` instead.

Maybe we should enable `-deprecation` in this module too, haha.

ckipp01: 👍🏼 fixed

scalalib/src/mill/scalalib/bsp/ScalaMetalsSupport.scala
```diff
@@ -3,23 +3,44 @@ package mill.scalalib.bsp
...
+           |def semanticDbVersion = ???
+           |""".stripMargin
+      Result.Failure(msg)
+    } else {
+      Result.Success(
+        super.scalacPluginIvyDeps() ++ Agg(
+          ivy"org.scalameta:::semanticdb-scalac:${semanticDbVersion()}"
+        )
+      )
+    }
```
lefou, line 32: Looks like we're missing the happy Scala 3 case now, which does not need to add any dep at all.

scalalib/src/mill/scalalib/bsp/ScalaMetalsSupport.scala
```diff
@@ -3,23 +3,44 @@ package mill.scalalib.bsp
...
 import mill.define.Target
+import mill.scalalib.api.ZincWorkerUtil.isScala3
 import mill.scalalib.{Dep, DepSyntax, ScalaModule}
+import mill.api.Result
 
 /*+ Enable some common settings required to properly support Metals Language Server (via BSP). */
 @experimental
 trait ScalaMetalsSupport extends ScalaModule {
 
   /** The semanticDB version to use. It needs to support your configured Scala versions. */
```
lefou, line 14:
```suggestion
  /** The semanticDB version to use. It needs to support your configured Scala 2 version. 
   *  Not needed for Scala 3.
   */
```

scalalib/src/mill/scalalib/bsp/ScalaMetalsSupport.scala
ckipp01: ah, whoops, fixed.

lefou:
Thank you!

PR #1594, opened 2021-11-29
Add note about "Switch Build Server" in Metals.

ckipp01:
Now that https://github.com/scalameta/metals/pull/3308 is merged Metals can actually detect (based on mill version) if the current workspace can use Mill as a BSP server even if there is no `.bsp/mill-bsp.json` created yet. If the users selects this, it will automatically create the `.bsp/mill-bsp.json` file and then connect to mill. This pr just clarifies this in the docs and mentions the "Switch Build Server" command they'll need to use.

Just to show an example, you can see here there is no bsp entry, but it's still shown as an option:


lefou (approved):
Looks good to me. Thank you!

Issue #1595, opened 2021-11-29
Avoid using `build/showMessage` for ZincWorker cache.


ckipp01:
When using mill as a BSP server you'll occasionally see a message saying `ZinkWorker: using cache size 1` right in your editor. I don't think this is a great message to surface to the user since it's not actionable and arguable some users will have no idea what it means. This is sent from mill via:

```json
[Trace - 02:21:27 PM] Received notification 'build/showMessage'
Params: {
  "type": 4,
  "message": "ZinkWorker: using cache size 1"
}
```
Looking through the code the only place I see this is:

https://github.com/com-lihaoyi/mill/blob/c3b64a54be184d9d41b219d84761479dd61f10af/scalalib/src/ZincWorkerModule.scala#L52

But it's a debug  message, so I'm confused how this is floating up.  If I understand it I see the logger in here:

https://github.com/com-lihaoyi/mill/blob/c3b64a54be184d9d41b219d84761479dd61f10af/bsp/src/mill/bsp/MillBspLogger.scala#L52-L55

This is getting captured here, but again I'm unsure how this is floating up as I don't have debug set. Also, do you think it might be better to use https://build-server-protocol.github.io/docs/specification.html#log-message instead of showMessage for some of the debug messages? I feel like https://build-server-protocol.github.io/docs/specification.html#show-message should be used sparingly and only when there is some type of actionable output to the user or informative information that they'd what to know. What do you think?

lefou:
The interesting thing is how this kind of message should be handled? BSP defined 4 kind of message types: `ERROR`, `WARNING`, `INFORMATION` and `LOG`. This suggests that `LOG` level is below `INFORMATION` which is typically used for debug messages. Although I agree that it is questionable whether we should log this dedicated message at all, we should pass debug logging to the user somehow. Who's responsibility is to enable/disable debug logging. Should the client do it and just not show the message or should the server never send the message?

lefou:
As Mill BSP support is still in heavy development we want to see all message including debug in the `.bsp/mill-bsp.stderr` file. (This files catches everthing that goes to STDERR, for convenience and to not confuse the BSP client.) That's why we currently default (and hardcode I think) to enable debug logging.

ckipp01:
> Who's responsibility is it to enable/disable debug logging. Should the client do it and just not show the message or should the server never send the message?

[...]

PR #1597, opened 2021-11-30
Reviewed ZincWorker

lefou:
* Clenaup
* Added return types
* Removed deprecated API calls
* try harder to use correct javac for compilation of Zinc bridge

Fix https://github.com/com-lihaoyi/mill/issues/1595


scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
```diff
@@ -413,6 +420,10 @@ class ZincWorkerImpl(
       _ => None
     )
     val loggerId = Thread.currentThread().getId.toString
+    // The following three calls to [[LogExchange]] are deprecated, but the
+    // suggested alternatives aren't public API, so we can't really do anything
+    // to avoid calling these deprecated API.
+    // See issue https://github.com/sbt/zinc/issues/1007
```
lefou, line 426: Issue was moved.

```suggestion
    // See issue https://github.com/sbt/sbt/issues/6734
```

Issue #554, opened 2019-02-20
Tasks overridden in multiple traits are not cached


jeantil:
I think this may be related to #86 

Consider the following build :
```scala
trait A extends JavaModule{
  def aSources = T.sources{millSourcePath / 'a}
  def aOutput = T{aSources()}
  override def generatedSources = T{ super.generatedSources() ++  aOutput() }
}

trait B extends JavaModule{
  def bSources = T.sources{millSourcePath / 'b}
  def bOutput = T{bSources()}

  override def generatedSources = T{ super.generatedSources() ++  bOutput() }
}
object core extends A with B{
  def ok = T{
    aOutput() ++ bOutput()
  }
}
```

running `mill core.ok` twice yields the following `mill-profile.json`
```json
[
    {
        "label": "core.aSources",
        "millis": 2,
        "cached": false
    },
    {
        "label": "core.aOutput",
        "millis": 0,
        "cached": true
    },
    {
        "label": "core.bSources",
        "millis": 2,
        "cached": false
    },
    {
        "label": "core.bOutput",
        "millis": 1,
        "cached": true
    },
    {
        "label": "core.ok",
        "millis": 0,
        "cached": true
    }
]
```
as expected both `T.source`'s are reevaluated but everything else is cached. 

However running `mill core.generatedSources` yields : 
```json
[
    {
        "label": "core.generatedSources.overriden.mill.scalalib.JavaModule.generatedSources",
        "millis": 0,
        "cached": true
    },
    {
        "label": "core.aSources",
        "millis": 2,
        "cached": false
    },
    {
        "label": "core.aOutput",
        "millis": 1,
        "cached": true
    },
    {
        "label": "core.generatedSources",
        "millis": 2,
        "cached": false
    },
    {
        "label": "core.bSources",
        "millis": 2,
        "cached": false
    },
    {
        "label": "core.bOutput",
        "millis": 0,
        "cached": true
    },
    {
        "label": "core.generatedSources",
        "millis": 2,
        "cached": false
    }
]
```
Here the `core.generatedSources` task appears twice, neither of which is cached 

I would expect:
- the first occurrence to be labeled `core.generatedSources.overriden.A.generatedSources`), and to be cached.
- the second occurrence to be the actual `core.generatedSources` and to also be cached

Inspecting the task `core.generatedSources` yields : 
```
/Users/jean/mill-release inspect core.generatedSources
[1/1] inspect
core.generatedSources(build.sc:14)
[...]

PR #1600, opened 2021-12-07
Improve overriden target detection to let it handle stackable-traits-style overrides

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/554

This PR swaps out our detection of overriden targets and commands: 

- Previously, we tried to count how many overriden methods there are upstream of a particular target, statically at its definition site. Any target that has the max number of overriden methods is considered the "final" one, while others are considered overriden. This fails for stackable traits, e.g. the example in the test suite below, as multiple traits can have the same number of static definiton-site overrides but when stacked together on a module one of them ends up winning

- Now, we simply do a reverse walk of the targets list after they have been topo sorted. Since the directionality of overrides is exactly opposite that of the topo sort, we can assume that the first time we see a target of a particular label in the reverse walk it's the "final" one, while any subsequent targets we see with the same label must be the overriden versions

The new mechanism is simpler, and possibly more correct. Hopefully CI would catch any issues...

Added a test case that fails on master and passes on this PR. Also verified manually that the issue reported above can be reproduced on master, and does not reproduce (i.e. it behaves correctly) on this PR.


lefou:
I love that we get rid of the `overrides` parameter! I always avoided to understand it fully, haha.

lihaoyi:
@lefou looks like the bootstrap tests are failing due to some external plugins. Not sure what your SOP for dealing with these is, but IMO for plugins Mill's own build depends on we should just keep them in the mill repo's contrib folder to simplify this sort of change

lefou:
Bootstrap tests are special and can break when we change some API of Mill. Although you see it fail because of plugins, this issue can also happen when we don't depend on external plugins, but instead use some API that has changed directly. (I remember we had that issue in the past, e.g. when we changed some publishing API.) Our CI now applies a patch file before it runs bootstrap tests, it's `ci/mill-bootstrap.patch` and currently empty. That way, we can customize the `build.sc` or other files before we bootstrap.

I don't think we should deny ourselves to use external plugins, just because we want to run some bootstrap tests. Instead, we can adapt the build (and ad-hoc disable plugins) to run the bootstrap tests.

lefou:
I can prepare a patch file and push it to your branch later

main/core/src/mill/eval/Evaluator.scala
```diff
@@ -94,6 +94,13 @@ case class Evaluator(
       testReporter: TestReporter = DummyTestReporter
   ): Evaluator.Results = {
     val (sortedGroups, transitive) = Evaluator.plan(rootModule, goals)
+    val terminalNames = sortedGroups
+      .keys()
+      .map{
+        case Left(x) => x
+        case Right(x) => (System.identityHashCode(x), x.task.ctx.lineNum, x.segments.render)
+      }
+      .toList
```
lefou, line 103: What happens here and where is `terminalNames` used?

main/core/src/mill/eval/Evaluator.scala
```diff
@@ -659,44 +666,27 @@ object Evaluator {
   def plan(rootModule: BaseModule, goals: Agg[Task[_]]) = {
```
lefou, line 666: Looks like we no longer need that rootModule parameter here?

main/core/src/mill/eval/Evaluator.scala
lihaoyi: It's not used, this is leftover debugging logic. Will remove

main/core/src/mill/eval/Evaluator.scala
lihaoyi: Good catch, seems like we used to use it to try and resolve the "final" implementation of a task to do override count comparisons. Now we are simply finding it via the topo sort, it's no longer necessary

lefou (approved):
LGTM!

Issue #598, opened 2019-04-18
Input tasks require JSON readable and writable output


kag0:
I'm not sure what the intended behavior is, but the there's a little inconsistency between the documentation and the code for [input tasks](http://www.lihaoyi.com/mill/page/tasks.html#inputs).
The [task cheat sheet](http://www.lihaoyi.com/mill/page/tasks.html#cheat-sheet) confirms that inputs do not need to be JSON readable or JSON writable, however [`T.input`](https://github.com/lihaoyi/mill/blob/master/main/core/src/define/Task.scala#L153) requires an implicit upickle `ReadWriter` for the result type. 

steps to reproduce:
```scala
import mill._, scalalib._
import java.time.Instant

object baz extends JavaModule {
  def test = T.input(Instant.now)
}
```
Confirm the build does not compile due to 
> could not find implicit value for parameter rw: upickle.default.ReadWriter[java.time.Instant]

lihaoyi:
Not sure how easy it would be to remove the `ReadWriter` dependency in `Input`, but one workaround could be to define a `ReadWriter` for `java.time.Instant`. Since it's in the standard library, perhaps even push it upstream to the uPickle repo

kag0:
The workaround I'm currently using is 
```scala
def w[T] = writer[Unit].comap[T](_ => ())
def r[T] = reader[Unit].map[T](_ => ???)
def rw[T]: ReadWriter[T] = ReadWriter.join(r, w)
def task = T.input(...)(rw, implicitly[Ctx])
```

For two reasons
1. This works for any type, and as far as I can tell the actual JSON value is unused. 
2. Some of the values I want to use are sensitive (db credentials for the flyway plugin) and I don't want them written to disk unnecessarily. 

This second point could probably be expanded into a separate feature request except that the documented behavior of an input task already does pretty much what I want.

PR #1601, opened 2021-12-07
Make input tasks only require a JSON writer, not reader

lihaoyi:
Resolves https://github.com/com-lihaoyi/mill/issues/598. Input tasks results are re-evaluated every time and thus never read from the disk, and so a `upickle.default.Reader` is not required. Nevertheless, we still require that a `upickle.default.Writer` in order to generate the `*.json` files used for `./mill show` and other debugging purposes.

Targeting https://github.com/com-lihaoyi/mill/pull/1600 to avoid merge conflicts

Let's see what CI says...

TBH not sure if this is worth doing, or whether we should just specify that `Input` tasks require a `ReadWriter` just to keep things consistent. We don't currently have any use case for de-serializing the input task JSON (e.g. even `mill show` takes the JSON from disk and spits it out verbatim without de-serializing) but that doesn't mean we won't find such use cases in future. But if such use cases do re-appear, we can always add that functionality back. `Command`s are write-only, so making `Input`s write-only would also not be unprecedented


main/core/src/mill/eval/Evaluator.scala
```diff
@@ -18,17 +18,7 @@ import scala.jdk.CollectionConverters._
...
-case class Labelled[T](task: NamedTask[T], segments: Segments) {
-  def format = task match {
-    case t: Target[T] => Some(t.readWrite.asInstanceOf[upickle.default.ReadWriter[T]])
-    case _ => None
-  }
-  def writer = task match {
-    case t: mill.define.Command[T] => Some(t.writer.asInstanceOf[upickle.default.Writer[T]])
-    case t: Target[T] => Some(t.readWrite.asInstanceOf[upickle.default.ReadWriter[T]])
-    case _ => None
-  }
```
lihaoyi, line 30: These pattern matches are kind of weird, and go against the un-sealed open nature of our `Task` class hierarchy. Thus I moved the logic into abstract methods on `NamedTask`

lefou: Until now I always though of our tasks as a sealed hierarchy. We handle most of their specialities when evaluating them and I find it hard to come up with scenarios which could be realized by a new type of task, which can be solely realized through the tasks implementation. So we should make them sealed.

lihaoyi:
@lefou I reverted the pattern-match/virtual-methods change for now. It's orthogonal to the main goal of this PR, and I'm not yet confident enough to decide either way so easiest to just leave as is for now

lefou (commented):
I really like this change, as it removes an unnecessary restriction. And this issue already popped up in the chats.

I still have some questions to the implementation touched in this PR but not directly to this change. So feel free to address it or not. The `Labelled` case class has a `writer` method which is used to write, but it has also a `format` method. If I'm right, this one is only used to read stuff? I don't get why it is called `format`? Maybe you can take the opportunity to a) add proper return types (to make clear if we are also interested in the inferred writer-aspect) and b) probably rename it to `read` (or enlighten me why not).

lihaoyi:
`writer` basically returns the `Option[upickle.default.Writer[T]]` for a particular `Task[T]`, while `format` returns the `Option[upickle.default.ReadWriter[T]]`.

The name `format` is a holdover from the play-json and other JSON libraries which called their `ReadWriter` types `*Format`.

We should clean it up, but I'd like to get this fix in first and we can clean up separately. As discussed, we should probably do something about the `Task[T]` hierarchy, which may make this particular issue moot.

lefou (approved):
LGTM!

PR #1618, opened 2021-12-13
Reenable acyclic compiler plugin

lefou:


lefou:
Would be nice to have acyclic reporting on warning level.

See https://github.com/com-lihaoyi/acyclic/pull/24

PR #1619, opened 2021-12-14
POC: Updates needed for upcomming zinc 1.6.x

lefou:
- Update zinc versionto 1.5.8
- Output a warning message when using Scala 2.12.4
- Change used scala version in test


lefou:
Converted to draft. This PR was incomplete, but no longer necessary for next bump to zinc 1.5.9, as they reverted their chnage which introduced the issues. But later, when zinc 1.6.x comes out, we may still need this tweak, hence I keep this PR open for later use.

PR #1621, opened 2021-12-15
Properly wire stdin in client-server mode and enable "Enter to Re-run" for mill -w

lihaoyi:
This gets rid of our custom `watchAndWait` logic, standardizing it with what's upstream in Ammonite and allowing Mill to use the "Enter to re-run" functionality that was introduced in Ammonite 2.5.0.

While this worked for `./mill -i` out of the box, some work was necessary to get it working for Mill's client-server mode. This fixes the wiring for the socket's stdin to ensure it gets plumbed through, and works around the broken `available()` implementation in the `UnixDomainSocketInputStream` by proxying it via an `InputPumper`.

Tested manually:

```
lihaoyi mill$ ./mill -i dev.run scratch -w foo.run
[90/647] de.tobiasroeser.mill.vcs.version.VcsVersion.vcsState
[647/647] dev.run
[44/44] foo.run
false
Watching for changes to 4 paths... (Enter to re-run, Ctrl-C to exit)

<ENTER>
[44/44] foo.run
false
Watching for changes to 4 paths... (Enter to re-run, Ctrl-C to exit)

<ENTER>
[44/44] foo.run
false
Watching for changes to 4 paths... (Enter to re-run, Ctrl-C to exit)

<ENTER>
[44/44] foo.run
false
Watching for changes to 4 paths... (Enter to re-run, Ctrl-C to exit)
<Ctrl-C>
Host JVM shutdown. Forcefully destroying subprocess ...
```

Notably, this change also allows you to run interactive `mill foo.run` programs without using `-i`, e.g. using this modified `scratch/src/Foo.scala`:

```
package foo
object Foo{

    println("enter any character: ")
    val c = System.in.read().toChar
    println("c: [" + c + "]")
  }
}
```

```bash
lihaoyi mill$ ./mill -i dev.run scratch -w foo.run
...
[44/44] foo.run
enter any character:
g<ENTER>
c: [g]
Watching for changes to 4 paths... (Enter to re-run, Ctrl-C to exit)
```

### DIFF
--- a/main/client/src/mill/main/client/InputPumper.java
+++ b/main/client/src/mill/main/client/InputPumper.java
@@ -2,7 +2,6 @@ package mill.main.client;
 
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.util.Arrays;
 
 public class InputPumper implements Runnable{
     private InputStream src;
--- a/main/client/src/mill/main/client/InputPumper.java
+++ b/main/client/src/mill/main/client/InputPumper.java
@@ -2,6 +2,7 @@ package mill.main.client;
 
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.Arrays;
 
 public class InputPumper implements Runnable{
     private InputStream src;
--- a/main/src/mill/main/MillServerMain.scala
+++ b/main/src/mill/main/MillServerMain.scala
@@ -68,7 +68,7 @@ object MillServerMain extends mill.main.MillServerMain[EvaluatorState] {
       args,
       stateCache,
       mainInteractive = mainInteractive,
-      DummyInputStream,
+      stdin,
       stdout,
       stderr,
       env,
@@ -78,6 +78,7 @@ object MillServerMain extends mill.main.MillServerMain[EvaluatorState] {
     )
   }
 }
+
 
 class Server[T](
     lockBase: String,
@@ -134,12 +135,26 @@ class Server[T](
     }.getOrElse(throw new Exception("PID already present"))
   }
 
+  def proxyInputStreamThroughPumper(in: InputStream) = {
+    val pipedInput = new PipedInputStream()
+    val pipedOutput = new PipedOutputStream()
+    pipedOutput.connect(pipedInput)
+    val pumper = new InputPumper(in, pipedOutput, false)
+    val pumperThread = new Thread(pumper)
+    pumperThread.start()
+    pipedInput
+  }
   def handleRun(clientSocket: Socket, initialSystemProperties: Map[String, String]) = {
 
     val currentOutErr = clientSocket.getOutputStream
     val stdout = new PrintStream(new ProxyOutputStream(currentOutErr, 1), true)
     val stderr = new PrintStream(new ProxyOutputStream(currentOutErr, -1), true)
-    val socketIn = clientSocket.getInputStream
+    // Proxy the input stream through a pair of Piped**putStream via a pumper,
+    // as the `UnixDomainSocketInputStream` we get directly from the socket does
+    // not properly implement `available(): Int` and thus messes up polling logic
+    // that relies on that method
+    val proxiedSocketInput = proxyInputStreamThroughPumper(clientSocket.getInputStream)
+
     val argStream = new FileInputStream(lockBase + "/run")
     val interactive = argStream.read() != 0
     val clientMillVersion = Util.readString(argStream)
@@ -168,7 +183,7 @@ class Server[T](
             args,
             sm.stateCache,
             interactive,
-            socketIn,
+            proxiedSocketInput,
             stdout,
             stderr,
             env.asScala.toMap,
--- a/main/src/mill/main/MillServerMain.scala
+++ b/main/src/mill/main/MillServerMain.scala
@@ -141,6 +141,7 @@ class Server[T](
     pipedOutput.connect(pipedInput)
     val pumper = new InputPumper(in, pipedOutput, false)
     val pumperThread = new Thread(pumper)
+    pumperThread.setDaemon(true)
     pumperThread.start()
     pipedInput
   }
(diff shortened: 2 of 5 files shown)


main/client/src/mill/main/client/InputPumper.java
```diff
@@ -2,6 +2,7 @@
 
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.Arrays;
```
lefou, line 5: Unused

main/src/mill/modules/Jvm.scala
```diff
@@ -191,7 +190,7 @@ object Jvm {
     // to System.in/System.out/System.err, so the subprocess's streams get sent
     // to the parent process's origin outputs even if we want to direct them
     // elsewhere
-    if (System.in.isInstanceOf[ByteArrayInputStream]) {
+    if (System.in.isInstanceOf[PipedInputStream]) {
```
lefou, line 193: I'm not hundred percent certain, that we are correct here. We probably want to keep `ByteArrayInputStream` as alternative proof for a "fake" or redirected input.

main/src/mill/main/MillServerMain.scala
```diff
@@ -134,12 +135,26 @@ class Server[T](
     }.getOrElse(throw new Exception("PID already present"))
   }
 
+  def proxyInputStreamThroughPumper(in: InputStream) = {
+    val pipedInput = new PipedInputStream()
+    val pipedOutput = new PipedOutputStream()
+    pipedOutput.connect(pipedInput)
+    val pumper = new InputPumper(in, pipedOutput, false)
+    val pumperThread = new Thread(pumper)
```
lefou, line 143: This thread should be marked as daemon via `pumperThread.setDaemon(true)`, I think.

lefou (commented):
Another great change! I love that it fixes the need to use `--interactive` for interactive commands.

main/src/mill/modules/Jvm.scala
lefou: I mean, we should keep both, `PipedInputStream` and `ByteArrayInputStream` (OR-ed) in the check.

scratch/foo/src/Foo.scala
```diff
@@ -2,6 +2,8 @@ package foo
 object Foo{
 
   def main(args: Array[String]): Unit = {
-    println(1 == "")
+    println("enter any character: ")
+    val c = System.in.read().toChar
+    println("c: [" + c + "]")
```
lefou, line 7: As we encourage users to use `scratch` to locally play around, we should not modify it upstream.

main/src/mill/main/MillServerMain.scala
lihaoyi: Possibly? Things seem to work fine without it tho, because in `mill.MillMain.main` we explicitly call `System.exit` at the end

main/src/mill/modules/Jvm.scala
lefou: Yeah, I understand this.
My reasoning was the following: We use the `Jvm` API in many places, and it is especially also used to handle processes and in-process Java tools in Mill plugins. As such, the `runSubprocess` method isn't limited to its use as internal Mill. And it should work as expected when an external plugin uses it to handle some sub processes, As using `ByteArrayInputStream` is widely used to control the input of a spawned process, we should support it.
But I will not insist in that standpoint. Your solution is more lean and solves exactly one issue.

main/src/mill/main/MillServerMain.scala
lefou: Yeah, I think if we agree its nature _is_ daemon, it's no problem to also mark it as such. It may make probably also our own lives easier, eventually.

Motivation:
I collected a long history of attempts to use apps as libraries. Relying on `System.exit` is the first and foremost blocker to do that. In my integrator-role I often searched for alternative entry points to tools without `System.exit`. But this always assumes, that the rest of the resources are well managed.

lihaoyi: Got it. Added the `setDaemon` call

lefou: Thanks!

main/src/mill/modules/Jvm.scala
lihaoyi: I understand that `runSubprocess` is exposed to plugins and custom tasks and stuff, but all of those run in the context of a Mill build JVM? In that case, there won't be any `ByteArrayOutputStream` anymore, just a `PipedInputStream`.

If you were sub-processing out to a separate JVM you launched with the Mill classpath, and _then_ calling `Jvm.subProcess`, then i could imagine wanting to check for a dummy stdin. But AFAIK we never do that: Mill plugins and custom tasks are always in the Mill's JVM when they call `Jvm.subProcess`. `Jvm.subProcess` is also in the Mill `main` module, so it's not like some helper library we expect to use elsewhere.

lefou: Nice explanation. Agreed.

lefou (approved):
LGTM!

lihaoyi:
Notably, even the interactive Scala console works without `-i` now

```scala
$ ./mill -i dev.run scratch foo.console
[90/647] de.tobiasroeser.mill.vcs.version.VcsVersion.vcsState
[647/647] dev.run
[39/39] foo.console
Welcome to Scala 2.13.2 (OpenJDK 64-Bit Server VM, Java 11.0.7).
Dec 15, 2021 8:59:31 PM org.jline.utils.Log logr
WARNING: Unable to create a system terminal, creating a dumb terminal (enable debug logging for more information)
1
scala> + 1
val res0: Int = 2

scala> println(123, 456)
println(123, 456)(123,456)
```

Not 100%, `Ctrl-D` doesn't work properly to exit, and I'm guessing Ammonite won't work smoothly due to the clever TTY stuff it does. Nevertheless, it's moving in the right direction to get it all working for good

Issue #1625, opened 2021-12-15
ScalaPBModule  does not pick transformers


borissmidt:
I tend to replace the protobuf types to joda.time/java.time types when i generate protobuf dependencies:
This is possible by setting a single protobuf transformer which will replace one type with another.
But when i use this in 1 package these options are not picked up and the case class still contains google.porotbuf.timestamp.

when i have a project with 2 proto files and an converter:
`/protobuf/my/package/scalapb-options.proto`
```
package my.package
option (scalapb.options) = {
  scope: PACKAGE
  import: "com.package.ProtobufConversions._"
  field_transformations : [
    {
      when : {
        type: TYPE_MESSAGE
        type_name: ".google.protobuf.Timestamp"
      }
      set : {[scalapb.field] {type : 'org.joda.time.base.AbstractInstant' }}
    },
    {
      when : {
        type: TYPE_MESSAGE
        type_name: ".google.protobuf.Duration"
      }
      set : {[scalapb.field] {type : 'org.joda.time.Duration' }}
    }
  ]
};
```

and 
`/protobuf/my/package/protobuf/`

```
syntax = "proto3";

package my.package.protobuf;

import "google/protobuf/timestamp.proto";

message Interval {
    google.protobuf.Timestamp start = 1;
    google.protobuf.Timestamp end = 2;
}
```

https://scalapb.github.io/docs/transformations/

lefou:
Fixed by #1626

borissmidt:
Sadly the files with scalapb options aren't picked up even if you pass it as a `--proto_path` to scalapb.

To make multi projects build work i had to add this to my modules: 
```
override def scalaPBIncludePath: T[Seq[PathRef]] = T {
    Seq(scalaPBUnpackProto(), PathRef(this.millSourcePath / "resources")) ++ Target.sequence(moduleDeps.collect{case m: ScalaPBModule =>  m.scalaPBIncludePath})().flatten
 }
```
So we might need to add this as well to `ScalaPBModule` otherwise imports of protofiles difined in other packages are resolved and you fail wiht 'import x/y/z.proto` not found.

borissmidt:
I've done some further research it seems that this is a problem in general with protoc that option files are not picked up because only imported files are added to the generation path.
So this has to be resolved on the build tool level, so we will have to work together with the scalapb maintainer to implement it in all build tools in a consistent way. I'll make a follow up PR if we can find a good solution.
[...]

PR #1626, opened 2021-12-15
#1625 fix transformer not picked up

borissmidt:
fixes related issue #1625 
It was easy to resolve the missing package context file by passing all files that need to be generated to protoc at once.
This is similar how sbt and gradle use protoc to compile the sources.

### DIFF
--- a/contrib/scalapblib/src/ScalaPBWorker.scala
+++ b/contrib/scalapblib/src/ScalaPBWorker.scala
@@ -36,7 +36,7 @@ class ScalaPBWorker {
               s"--scala_out=${opts}${generatedDirectory.getCanonicalPath}",
               s"--proto_path=${root.getCanonicalPath}"
             ) ++ sources.map(_.getCanonicalPath)
-            ctx.log.debug(args.mkString(" "))
+            ctx.log.debug(s"ScalaPBC args: ${args.mkString(" ")}")
             mainMethod.invoke(null, args.toArray)
           }
         }
@@ -81,9 +81,7 @@ class ScalaPBWorker {
       scalaPBOptions: String,
       dest: os.Path,
       scalaPBCExtraArgs: Seq[String]
-  )(
-      implicit ctx: mill.api.Ctx
-  ): mill.api.Result[PathRef] = {
+  )(implicit ctx: mill.api.Ctx): mill.api.Result[PathRef] = {
     val compiler = scalaPB(scalaPBClasspath)
 
     def compileScalaPBDir(inputDir: os.Path): Unit = {
--- a/contrib/scalapblib/src/ScalaPBWorker.scala
+++ b/contrib/scalapblib/src/ScalaPBWorker.scala
@@ -25,7 +25,8 @@ class ScalaPBWorker {
 
         val instance = new ScalaPBWorkerApi {
           override def compileScalaPB(
-              source: File,
+              root: File,
+              sources: Seq[File],
               scalaPBOptions: String,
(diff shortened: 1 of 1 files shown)


contrib/scalapblib/src/ScalaPBWorker.scala
```diff
@@ -25,17 +25,18 @@ class ScalaPBWorker {
...
+            ): Unit = {
             val opts = if (scalaPBOptions.isEmpty) "" else scalaPBOptions + ":"
             val args = otherArgs ++ Seq(
               s"--scala_out=${opts}${generatedDirectory.getCanonicalPath}",
-              s"--proto_path=${source.getParentFile.getCanonicalPath}",
-              source.getCanonicalPath
-            )
+              s"--proto_path=${root.getCanonicalPath}"
+            ) ++ sources.map(_.getCanonicalPath)
+            println(args)
```
lefou, line 39: Looks like a debugging left over. If not, please use an appropriate method or stream of `ctx.log`, e.g. `debug` for debug info.

lefou (commented):
Thanks for your contribution! Please avoid reformatting unchanged lines, but if, please use Scalafmt with our config.

contrib/scalapblib/src/ScalaPBWorker.scala
borissmidt: Good catch, I've made it a debug log i think it is useful to see the exact command parameters passed to the generator then you can run always run it manually to find out what is 'wrong'.

contrib/scalapblib/src/ScalaPBWorker.scala
```diff
@@ -25,17 +25,18 @@ class ScalaPBWorker {
...
           ): Unit = {
             val opts = if (scalaPBOptions.isEmpty) "" else scalaPBOptions + ":"
             val args = otherArgs ++ Seq(
               s"--scala_out=${opts}${generatedDirectory.getCanonicalPath}",
-              s"--proto_path=${source.getParentFile.getCanonicalPath}",
-              source.getCanonicalPath
-            )
+              s"--proto_path=${root.getCanonicalPath}"
+            ) ++ sources.map(_.getCanonicalPath)
+            ctx.log.debug(args.mkString(" "))
```
lefou, line 39: I would prefer to have a more meaningful message, e.g. 

```suggestion
            ctx.log.debug("ScalaPBC args: ${args.mkString(" ")}")
```

borissmidt: Good idea thank you for the suggestion  i have added it.

lefou (approved):
Thanks. Looks good now! :+1:

borissmidt:
I might come back with another PR because i have projects with multiple modules that depend on each and they all do scalaPB. 
So i'm not sure if this will handle 'transititive' scalapb options.

borissmidt:
Ah damn i just found 1 bug i have to add 
```if(!files.isEmpty) {
          compiler.compileScalaPB(inputDir.toIO, files, scalaPBOptions, dest.toIO, scalaPBCExtraArgs)
        }
        ```
        otherwise we get an error if there is no file

lefou:
@borissmidt Best is to add a test case to reproduce the bug and open a new PR with the test and it's fix.

borissmidt:
Yes indeed i just couldn't fix my initial problem that scalapb.options on package scope level of other projects and probably jars are not picked up.
So i think this is a bigger problem within scalapb.

PR #1629, opened 2021-12-16
Support version scheme in pom

lolgab:


scalalib/src/PublishModule.scala
```diff
@@ -18,6 +18,11 @@ trait PublishModule extends JavaModule { outer =>
   def pomSettings: T[PomSettings]
   def publishVersion: T[String]
 
+  /**
+   * Version scheme used for the model: Supported values are VersionScheme.EarlySemVer, VersionScheme.PVP, and VersionScheme.SemVerSpec
```
lolgab, line 22: Comment taken and adapted from Sbt. They don't state all the possible values, probably because these are the recommended ones.

julienrf: As far as I know, these are the three supported versioning schemes, currently.

scalalib/src/publish/VersionScheme.scala
```diff
@@ -0,0 +1,12 @@
+package mill.scalalib.publish
+
+sealed abstract class VersionScheme(val value: Option[String])
+
+object VersionScheme {
+  case object NoScheme extends VersionScheme(None)
+  case object Always extends VersionScheme(Some("always"))
```
julienrf, line 7: This one is not a proper version scheme, it may only be used by users of libraries as an escape hatch to silence an eviction error.

scalalib/src/PublishModule.scala
```diff
@@ -18,6 +18,11 @@ trait PublishModule extends JavaModule { outer =>
...
+   */
+  def versionScheme: T[VersionScheme] = VersionScheme.NoScheme
```
julienrf, line 24: What do you think of making it an `Option[VersionScheme]` instead? (and modeling `NoScheme` as `None`, and getting rid of the inner `Option` in `VersionScheme`?

(Not sure this is important!)

scalalib/src/publish/Pom.scala
```diff
@@ -30,7 +30,8 @@ object Pom {
       artifact: Artifact,
       dependencies: Agg[Dependency],
       name: String,
-      pomSettings: PomSettings
+      pomSettings: PomSettings,
+      versionScheme: VersionScheme
```
julienrf, line 34: This change breaks the binary interface of the method `apply`, is this OK?

scalalib/test/src/publish/PomTests.scala
```diff
@@ -184,15 +191,33 @@ object PomTests extends TestSuite {
...
+    }
+
+    test("pomVersionScheme") {
+      val earlySemVerScheme = VersionScheme.EarlySemVer
+      val pomWithVersionScheme = pomXml(artifact, deps, artifactId, settings, earlySemVerScheme)
+
+      assert(
+        (pomWithVersionScheme \ "properties").nonEmpty,
+        (pomWithVersionScheme \ "properties" \ "versionScheme.info").length == 1,
+        (pomWithVersionScheme \ "properties" \ "versionScheme.info").head == "early-semver"
```
julienrf, line 208:
```suggestion
        (pomWithVersionScheme \ "properties" \ "info.versionScheme").length == 1,
        (pomWithVersionScheme \ "properties" \ "info.versionScheme").head == "early-semver"
```

julienrf (commented):
That was quick, thank you!

lefou (commented):
Thanks for this contribution. Can you tell, if coursier actually supports this `info.versionScheme` property, so that we also can benefit from it, or is this only used by sbt to filter coursier version resolutions by impact?

This PR effectively just adds some property to the `pom.xml`, so I think the better implementation would be to add generic property support to `mill.scalallib.publish.Pom`, and then just derive the correct property from the selected version scheme. Also adding a `properties` or `pomProperties` setting to `PublishModule` would be nice, but can also be done later in a separate PR. I don't know, if the ivy file also support these properties, but if, we also support it.

scalalib/src/publish/Pom.scala
lefou: Here I would rather like to see a more generic `properties: Map[String, String]` parameter. The logic to calc the version scheme should happen on caller side.

In general, I'd say we want to keep a binary compatible overload, but as we are looking to get the next 0.10.x milestone out which isn't binary compatible anyways, we might get away with it. Maybe, we want provide a default value, then the change is source compatible.

julienrf:
> Thanks for this contribution. Can you tell, if coursier actually supports this `info.versionScheme` property, so that we also can benefit from it, or is this only used by sbt to filter coursier version resolutions by impact?

As far as I know, the logic is implemented in sbt/librarymanagement: https://github.com/sbt/librarymanagement/blob/develop/core/src/main/scala/sbt/librarymanagement/EvictionError.scala

Here is Coursier documentation about versions handling: https://get-coursier.io/docs/other-version-handling

I believe it would make sense to move that logic upstream to coursier. Maybe @alexarchambault or @eed3si9n have an opinion?

>  I think the better implementation would be to add generic property support to mill.scalallib.publish.Pom

I agree. `properties` are also used to indicate the URL of the API documentation (via the `info.apiURL` elemen), which can be used by some tools. There are maybe other usages.

lefou:
@julienrf Thanks for the pointers. I think properties are in general useful for interoperability with downstream consumers.

lefou:
This PR needs to be rebased on latest main branch. The `versionScheme` target should then feed the `publishProperties`. The default scheme should be `None` so that (if nothing is configured) publishing works as previously and no extra property will be added. Some Scaladoc on `publishProperties` should denote, that some logic is already present and calling `super` is required to keep it.

lolgab:
Rebased the PR and applied the suggested change.
Do think it's better to exposed the Option with versionScheme being a `Option[VersionScheme]` or a `VersionScheme` as it was before where `NoScheme` was a special scheme containing `None` as value?

Defining a `versionScheme` in the previous way way was slightly nicer:

```scala
override def versionScheme = VersionScheme.EarlySemver
```

Instead of:
```scala
override def versionScheme = Some(VersionScheme.EarlySemver)
```

I took the idea from `ScalaJSModule` that defines the ModuleKind as `NoModule`.
It was my first contribution to Mill in 2018 :) https://github.com/com-lihaoyi/mill/pull/212

Anyway, @lefou let me know which one you prefer.

lefou:
> Do think it's better to exposed the Option with versionScheme being a `Option[VersionScheme]` or a `VersionScheme` as it was before where `NoScheme` was a special scheme containing `None` as value?

I think the use of an `Option[VersionScheme]` better transports the fact that no version scheme needs to be applied at all and also no property will be created from it when absent. With a `VersionScheme.NoScheme` this may be less obvious. A `None` can be understood without looking at `VersionScheme` and without studying all the provided cases.

scalalib/test/src/HelloWorldTests.scala
```diff
@@ -179,6 +179,7 @@ object HelloWorldTests extends TestSuite {
         developers =
           Seq(Developer("lihaoyi", "Li Haoyi", "https://github.com/lihaoyi"))
       )
+      override def versionScheme = Some(VersionScheme.EarlySemVer)
```
lefou, line 182: Can you please add the link to the documentation of the version scheme concept. Either here or at `VersionScheme`.

scalalib/src/publish/VersionScheme.scala
```diff
@@ -0,0 +1,11 @@
...
+
+sealed abstract class VersionScheme(val value: String)
+
+object VersionScheme {
+  case object Always extends VersionScheme("always")
+  case object EarlySemVer extends VersionScheme("early-semver")
+  case object PVP extends VersionScheme("pvp")
+  case object SemVerSpec extends VersionScheme("semver-spec")
+  case object Strict extends VersionScheme("strict")
+}
```
lefou, line 11: I think it would be nice to have some more info what scheme means what, esp. what "PVP" stands for. A concrete link would be helpful, too.

scalalib/test/src/HelloWorldTests.scala
lefou: And a proper return type, please.

scalalib/src/publish/VersionScheme.scala
```diff
@@ -0,0 +1,11 @@
+package mill.scalalib.publish
+
+sealed abstract class VersionScheme(val value: String)
```
lefou, line 3: What do you think of a `def toProperty: (String, String) = "info.versionScheme -> value`? This would make the encoding very explicit.

scalalib/src/publish/VersionScheme.scala
lefou: This needs to be addressed.

scalalib/test/src/HelloWorldTests.scala
lefou: We should no longer add targets without Scaladoc (including a `@since`).

lolgab: This is the test. The definition has already scaladoc. I have to add the `@since` tho.

julienrf: I realize that the documentation about early-semver is a bit spread over multiple places. Relevant information is here:
https://docs.scala-lang.org/overviews/core/binary-compatibility-for-library-authors.html#recommended-versioning-scheme
https://www.scala-lang.org/blog/2021/02/16/preventing-version-conflicts-with-versionscheme.html
https://www.scala-sbt.org/1.x/docs/Publishing.html#Version+scheme

lefou: Sorry for the wrong place.

scalalib/src/publish/VersionScheme.scala
lolgab: Done ✅

lefou (approved):
LGTM. Thank you very much!

PR #1635, opened 2021-12-19
Fix/1625 Poc automatically pick up scalapb-options.proto filles

borissmidt:
This is a poc and i still need to add tests, (i need to have a look for that)
This scans all files for scalapb-options.proto and puts it on the generator path thus it picks up the package options from all dependencies and trasistivie projects.

### DIFF
--- a/contrib/scalapblib/src/ScalaPBModule.scala
+++ b/contrib/scalapblib/src/ScalaPBModule.scala
@@ -3,12 +3,13 @@ package contrib.scalapblib
 
 import coursier.MavenRepository
 import coursier.core.Version
-import mill.define.{Sources, Target}
+import mill.define.{Target, Sources}
 import mill.api.{PathRef, IO, Loose}
 import mill.scalalib.Lib.resolveDependencies
 import mill.scalalib._
 
-import java.util.zip.ZipInputStream
+import java.util.zip.{ZipInputStream, ZipFile}
+import scala.jdk.CollectionConverters.IteratorHasAsScala
 import scala.util.Using
 
 /** @see [[http://www.lihaoyi.com/mill/page/contrib-modules.html#scalapb ScalaPB Module]] */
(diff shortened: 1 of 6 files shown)


borissmidt:
https://github.com/thesamet/sbt-protoc/pull/286
Good news the maintainer of scalapb is open to the convention 'scalapb-options.proto' for package wide conventions.

borissmidt:
I'm not sure if this is the right place but i found another issue with the scalapb module:
`FileSystems.newFileSystem` is not threadsafe so cannot be used in parallel builds. 
we should use this instead:
```
  override def scalaPBUnpackProto: T[PathRef] = T {
    val cp = scalaPBProtoClasspath()
    val dest = T.dest
    cp.foreach { ref =>
      val jarFs = new ZipFile(ref.path.toIO.getCanonicalFile)
      try {
        jarFs.entries().asIterator().asScala.filter(_.getName.endsWith(".proto")).filterNot(_.isDirectory).foreach { entry =>
          val protoDest = dest.toNIO.resolve(entry.getName)
          Files.createDirectories(protoDest.getParent)
          Files.copy(
            jarFs.getInputStream(entry),
            protoDest,
            StandardCopyOption.REPLACE_EXISTING
          )
        }
      } finally jarFs.close()
    }
    PathRef(dest)
  }

```

I also couldn't make the scalapbModule work in combination with a mill.Cross because the sources aren't picked up. 

```scala
val scalaVersions = Seq("2.12","2.13")
trait CommonModule extends ScalaPBModule {

  def scalaVersion = crossVersion match {
    case "2.12" => "2.12.15"
    case "2.13" => "2.13.7"
    case "3" => "3.1.0"
  }
  ...
}
object `common-protofiles` extends mill.Cross[`common-protofiles`](scalaVersions:_*)
class `common-protofiles-module` extends CommonModule
```

lefou:
Closing due to inactivity and conflicts. Also, the feature might be implemented in the meantime. Have a look at:

* #2126

PR #1636, opened 2021-12-21
[WIP] make override folders follow the foo.bar output file convention

lihaoyi:
Follow up from https://github.com/com-lihaoyi/mill/pull/1588

### DIFF
--- a/main/core/src/mill/eval/Evaluator.scala
+++ b/main/core/src/mill/eval/Evaluator.scala
@@ -676,16 +676,16 @@ object Evaluator {
           Labelled(
             t,
             if (!overriden(t)) segments
-            else Segments(
-              segments.value.init ++
-              Seq(Segment.Label(segments.value.last.asInstanceOf[Segment.Label].value + ".override")) ++
-              t.ctx.enclosing.split("\\.|#| ").map(Segment.Label):_*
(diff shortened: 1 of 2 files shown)


lihaoyi:
Seems to be bumping up against the "Label may not contain a dot" assert. Might take some refactoring in order to resolve that in a non-hacky manner

lefou:
Haha, I did the same change but obviously forgot to push the PR. See #1637

lefou:
I guess this PR is rather a no brainer, it looks good. We should discuss though, why we have the `require` in `Label`. It was an even `assert` before. We have a history regarding this assert/require, see https://github.com/com-lihaoyi/mill/issues/820 and also some non-merged PRs. If it is only to protect from parsing in CLI, maybe we can just ignore it and live with the fact that those labels are probably never used from CLI anyways.

lefou:
@lihaoyi I just wanted to make clear, that it wasn't my intention to make you close your PR, when I pushed my version later. I just wanted to make it visible and start the discussion about the require statement, my interpretation of it and the fact, that all tests pass, once it is removed.

lihaoyi:
@lefou the two PRs are about the same, so happy to go with yours for now

PR #1637, opened 2021-12-21
Changed out path for overridden targets for consistency

### LINKED ISSUE #820: AssertionError when using predefScript.sc

lefou:
To be more consistent with the changes from PR
https://github.com/com-lihaoyi/mill/pull/1588
which re-layouted the log, meta, and dest path for targets.
This change touches the path of overriden target results.
It replaces the `../target/overriden/..` part with
`../target.overriden/..`.

There is a catch though, to make this work I had to remove one
`require` assertion, which guarded the presence of dots (`.`) in
labels. TBH, I previously stubled upon this require
and also added the FIXME comment to it, as it is unclear why it
even exists. Probably, because labels with dots can't be properly
parsed when given on cli. More details and hisory can be found in
issue https://github.com/com-lihaoyi/mill/issues/820.

Fix #820.

Review by @lihaoyi.

### DIFF
--- a/main/core/src/mill/eval/Evaluator.scala
+++ b/main/core/src/mill/eval/Evaluator.scala
@@ -660,24 +660,31 @@ object Evaluator {
     val transitive = Graph.transitiveTargets(goals)
     val topoSorted = Graph.topoSorted(transitive)
     val seen = collection.mutable.Set.empty[Segments]
-    val overriden = collection.mutable.Set.empty[Task[_]]
+    val overridden = collection.mutable.Set.empty[Task[_]]
     topoSorted.values.reverse.foreach {
       case x: NamedTask[_] =>
         if (!seen.contains(x.ctx.segments)) seen.add(x.ctx.segments)
-        else overriden.add(x)
+        else overridden.add(x)
       case _ => // donothing
     }
 
     val sortedGroups = Graph.groupAroundImportantTargets(topoSorted) {
       case t: NamedTask[Any] =>
         val segments = t.ctx.segments
-
-        val additional =
-          if (!overriden(t)) Nil
-          else
-            Seq(Segment.Label("overriden")) ++ t.ctx.enclosing.split("\\.|#| ").map(Segment.Label)
-
-        Right(Labelled(t, segments ++ additional))
+        Right(
+          Labelled(
+            t,
+            if (!overridden(t)) segments
+            else {
+              val Segment.Label(tName) = segments.value.last
+              Segments(
+                segments.value.init ++
+                  Seq(Segment.Label(tName + ".overridden")) ++
+                  t.ctx.enclosing.split("\\.|#| ").map(Segment.Label): _*
+              )
+            }
+          )
+        )
       case t if goals.contains(t) => Left(t)
     }
 
--- a/docs/antora/modules/ROOT/pages/Configuring_Mill.adoc
+++ b/docs/antora/modules/ROOT/pages/Configuring_Mill.adoc
@@ -237,7 +237,7 @@ mill foo.test
 ----
 
 By default, tests are run in a subprocess, and `forkArg` and `forkEnv` can be
-overriden to pass JVM flags &amp; environment variables. You can also use
+overridden to pass JVM flags &amp; environment variables. You can also use
 
 [source,bash]
 ----
--- a/docs/antora/modules/ROOT/pages/Modules.adoc
+++ b/docs/antora/modules/ROOT/pages/Modules.adoc
@@ -135,7 +135,7 @@ would have its `millSourcePath` be `./foo/lols/baz`.
 
 Note that `millSourcePath` is generally only used for a module's input source files.
 Output is always in the `out/` folder and cannot be changed, e.g. even with the
-overriden `millSourcePath` the output paths are still the default `./out/foo/bar` and
+overridden `millSourcePath` the output paths are still the default `./out/foo/bar` and
 `./out/foo/baz/qux` folders.
 
 == External Modules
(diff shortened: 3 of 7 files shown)


lihaoyi (approved):

lihaoyi:
It would take some refactoring to make the path/segment handling "clean", but I'm fine with removing the assert for now

lihaoyi:
The fundamental issue with that assert is that `Segment` currently (with that assert) models things which can be parsed from the CLI, whereas the `.override` things cannot be parsed from the CLI. Thus using `Segment` to model the `foo.override/` paths is not correct.

One slightly-less-hacky thing we could do is to add a third case to `Segment`, e.g. `Segment.Override` or something, to explicitly model this. Then we can ensure that the parser only returns `Segment.Label` or `Segment.Cross`, but other places we can synthesize `Segment.Override`s as well. That would also be a good direction to go in if we in future decided to allow people to build the overridden targets directly from the CLI, which AFAIK is currently not allowed.

There are probably other things we could do, but this was the solution I thought of off the top of my head.

main/core/src/mill/eval/Evaluator.scala
```diff
@@ -670,12 +670,16 @@ object Evaluator {
...
-          if (!overriden(t)) Nil
-          else
-            Seq(Segment.Label("overriden")) ++ t.ctx.enclosing.split("\\.|#| ").map(Segment.Label)
+        val segments0 = t.ctx.segments
+        val (segments, additional) =
+          if (!overriden(t)) (segments0, Nil)
+          else {
+            val Segment.Label(tName) = segments0.value.last
+            (
+              Segments(segments0.value.reverse.tail.reverse: _*),
```
lihaoyi, line 679: `.reverse.tail.reverse` is just `.init`

main/core/src/mill/eval/Evaluator.scala
```diff
@@ -670,12 +670,16 @@ object Evaluator {
...
+              Seq(Segment.Label(tName + ".overriden")) ++ t.ctx.enclosing.split("\\.|#| ").map(Segment.Label)
```
lihaoyi, line 680: Shall we rename this from `.overriden` to `.override`? or `.overridden` to at least spell it correctly :P

lefou: I renamed this one at least 3 times in the last two years, but then reverted to avoid nitpicking. Also, after looking into a dictionary, "overriden" is actually not incorrect. But now, as you see it the same, I'd glad to rename it. I think "overridden" is more correct than "override" as the folder is holding the state of "overridden" targets.

main/core/src/mill/eval/Evaluator.scala
lefou: Wow, I'm using Scala for more than 10 years; today I learned something new.

lihaoyi:
On second thoughts, the `.overriden` data can only apply once to an entire `Segments`, and thus should apply to `Segments` as a whole and not be a replacement for any individual `Segment`

lefou:
@lihaoyi Nice thing it, this PR actually fixed #820.

lefou:
I'd like to keep the mapping to `Segment` as-is for now. Parsing is at some degree unrelated, as the removed `require` was just an after-the-fact guard and there is still a unique mapping of parsables to segments. The only issue is, that it's not bidirectional, but we don't generate cli arguments from segments, do we? (If we do, e.g. when completing, we could filter them out.) We could come up with some quoting or escaping, if someone really needs that feature, which is rather unlikely.

lihaoyi:
@lefou we actually do generate a string from the `Segments` data structure in a few places via `Segments#render`, e.g. to populate the `mill-profile.json`.

I don't think we actually need it to be parseable - in the end it's just for debugging purposes - so I'm happy to merge this PR as is and figure out a way to clean things up later

PR #1638, opened 2021-12-22
Remove redundant elements from versionRanges

lolgab:
Having the extremes in versionRanges is useless.
If you have Scala `2.11`, `2.12` and `2.13`, `src-2.11+` is equivalent
to `src` and `src-2.13-` is equivalent also to `src`.
Also, having `src-2.12.12+` if `2.12.12` is the only `2.12` version is
redundant, since you can use `src-2.12+`.


lolgab:
I just found a bug. The folder `3-` is not in the results but it should be. Let me put this PR as draft again.

lefou (approved):
I think this is now some rather complex logic for such a simple task and it's effect isn't obvious from a quick glace. This means, users who use a `src-2.12+` source dir and drops support for 2.11 may be surprised that their sources are no longer found. Yes, we provide tools like `mill inspect` to find out what the resulting source dirs are. But this is probably not that obvious either. 

After all, this is a power feature which needs explicitly opted in via `CrossScalaVersionRanges`, so I'm going merge, but not without an queasy feeling.

lolgab:
Let's drop this for now.
It is not changing any api so in case it can be revisited later.

PR #1641, opened 2021-12-25
[WIP] Support Scala Native for Scala 3

lolgab:
- Bump Scala Native to 0.4.3-SNAPSHOT
- Return `Result.Failure` on wrong env variables
- Use right dependencies in Scala 3
- Use `repositoriesTask` to resolve Scala Native `bridgeFullClassPath`
  instead of hardcoded list of resolvers.
  This allows builds to use SNAPSHOT versions of Scala Native.
- Drop support for Scala Native 0.4.2-

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -66,10 +66,10 @@ object Deps {
   }
 
   object Scalanative_0_4 {
-    val scalanativeTools = ivy"org.scala-native::tools:0.4.3-SNAPSHOT"
-    val scalanativeUtil = ivy"org.scala-native::util:0.4.3-SNAPSHOT"
-    val scalanativeNir = ivy"org.scala-native::nir:0.4.3-SNAPSHOT"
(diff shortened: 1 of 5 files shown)


lefou:
Can you please clarify, why you drop support for 2.4.2 and use 0.4.3-SNAPSHOT? Is this the first version, that supports Scala 3?

lolgab:
They are not compatible between each other. If you try to build the `0.4.3-SNAPSHOT` with the tools lib from `0.4.2` it fails to link and the same thing happens if you use `0.4.3-SNAPSHOT` tools to build `0.4.2`.
The alternative would be to create another worker with different versions, but IMO it's not worth it since the 0.4.3 is backward compatible with 0.4.2 and so users are not stuck if they upgrade to 0.4.3

lolgab:
Closed in favor of https://github.com/com-lihaoyi/mill/pull/1673

PR #1646, opened 2021-12-29
Added support to add properties to generated Maven pom

lefou:
Also added a packaging attribute to PomSettings.

### DIFF
--- a/scalalib/src/PublishModule.scala
+++ b/scalalib/src/PublishModule.scala
@@ -41,7 +41,8 @@ trait PublishModule extends JavaModule { outer =>
   }
 
   def pom: Target[PathRef] = T {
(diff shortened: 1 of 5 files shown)


lolgab (approved):
LGTM! 👍
About the version scheme I was defining in the other PR, how does it play together with `publishProperties`?

lefou:
@lolgab I'm a bit unsure what the better approach for `versionScheme` is. We can have a dedicated `def versionScheme` as you did in #1629 and derive the `publishProperties` from it, or we can provide some kind of constants to be used in `publishProperties` directly. Something like:

```scala
object VersionScheme {
  val Strict = ("info.versionScheme", "strict")
  ...
}

def publishProperties = Map(VersionScheme.Strict)
```

I like the former approach more.

I still struggle to find information if and how these properties are supported in `ivy.xml`.

PR #1657, opened 2022-01-08
Bump zinc to 1.6.0 which uses log4j 2.17.0 already

lolgab:
- Remove log4j from ivyDeps
- Bump `scalaVersion` in tests from `2.12.4` to `2.12.5` since Zinc introduces a bug in `2.12.4`.
- Update Docs to use newer Scala Versions.


lefou:
Can we close this?

PR #1658, opened 2022-01-08
Negate test-optional return code in CI

lolgab:
Having a always red CI doens't help identify real issues.
This negates the output of the `test-optional` step, so it will be a problem when it will start to pass, and so it will not become optional anymore.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -60,7 +60,8 @@ jobs:
       matrix:
         java-version: [8, 11]
         buildcmd:
-          - ./mill -i -k __.mimaReportBinaryIssues
+          # Negate return code with `!` since these tests are supposed to fail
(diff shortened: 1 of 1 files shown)


lefou:
The (badly named) test-optional tests are no longer optional once we release Mill 0.10. Also, they are not supposed to fail, but maintaining a whitelist while changing the API is cumbersome, so we don't require it in Milestone-phase.

lolgab:
Yes, I know.
Can't we revert this PR after 0.10 is released? I prefer much more to have a green CI.

lefou:
I'm not sure I want to be forced to remember that I need to revert something before release.

The first failing commit since M5 means actually, we broke binary compat again, which is useful info for plugin usage. And a hint we should release another milestone.

PR #1663, opened 2022-01-11
Granular cache invalidation with multiple build files

lolgab:
## Motivation

Mill cache gets invalidated everytime the build compile output changes. 
This makes Mill do a lot of repeted work. Even if you don't change a task and it's cached result could be used, it's invalidated aggressively.
This happens even if the task is defined in an external `.sc` that wasn't changed nor its dependencies weren't changed.

## Implementation
Before this PR, Mill was adding the `classLoaderSig`'s hash to the hash of every task.
`classLoaderSig` is a map from classloader elements (can be either URIs for external jars or strings for local ammonite scripts class output) to the hash of the files.
Every time one element of this big map changed the task got invalidated.

Now the classLoaderSig is split in two parts. The external libraries are handled as before.
The scripts part, instead, is:
- compacted to merge together multiple classes of the same file
  for example:
  ```scala
  Map(
    "ammonite.$file.build" -> 1,
    "ammonite.$file.build$" -> 2,
    "ammonite.$file.build$aModule$" -> 3
  )
  ```
  
  is compacted to:
  ```scala
  Map("ammonite.$file.build" -> 6)
  ```
  using `groupMapReduce`.
- filtered using the Ammonite interp cache import graph.
  The import graph is first collected for every compiled script.
  Then using the already existing `Graph.transitiveNodes` (that was adapted to be generic instead of specific to Tasks),
  the scriptsClassLoaderSig is filtered to remove all elements that are not transitive imports of the task file.
- This new map's hash is then added to the task hash.
- Since this didn't make it to the merge window of `0.10.0` I converted it to be binary compatible. I made the `Evaluator` and  `EvaluatorState` case classes non `case` and wrote the `case class` generated methods manually as `@deprecated(since = "0.10.0")` 

### TODO

- [x] Handle `^` imports in scripts
- [x] Make test run with `fork = false`
- [x] Implement the test's external stdout handling for both cases (fork = `true` or `false`)
- [x] handle imports from higher paths than pwd ( `import $file.^.^.afile` )
- [x] correctly merge classloader sigs from multiple compilation of classes when imported in different scopes
       If two files import the same file, Ammonite compiles the dependee file twice to two different class names:
       if `build.sc` depends on `$file.inputA` and on `$file.e.inputE` and at the same time `$file.e.inputE` depends on `$file.^.inputA`, Ammonite generates `"ammonite.$file.a.inputA"` and `"ammonite.$file.e.$up.a.inputA"`. We need to merge the sigs from these two compilations so we're sure we invalidate correcly.
- [x] fix and test algorithm to convert graph adjacency list to tree structure.


integration/test/src/local/Tests.scala
```diff
@@ -10,4 +10,4 @@ object UpickleTests extends mill.integration.UpickleTests(fork = false)
 object PlayJsonTests extends mill.integration.PlayJsonTests(fork = false)
 object CaffeineTests extends mill.integration.CaffeineTests(fork = false)
 object DocAnnotationsTests extends mill.integration.DocAnnotationsTests(fork = false)
-object ScriptsInvalidationTests extends mill.integration.ScriptsInvalidationTests(fork = true)
+object ScriptsInvalidationTests extends mill.integration.ScriptsInvalidationTests(fork = false)
```
lefou, line 13: We typically also have an instance with `fork = true` in package `mill.integration.forked`.

lolgab: Now I understand why I was breaking the convention of the test being `fork = true` in the wrong place! :D
Done by the way :white

lolgab:
I don't understand if we are running all the integration tests in CI
Some tests appear in some specific script file like:
```
./mill integration.test "mill.integration.local.{AcyclicTests,AmmoniteTests,DocAnnotationsTests}"
```
but some other don't.

Am I missing something?

lolgab:
Probably a good way to make the integration tests less likely to be skipped my mistake is to group them in separate packages like:
```scala
package mill.integration.local

package group1 {
    object AcyclicTests extends mill.integration.AcyclicTests(fork = false)
    object AmmoniteTests extends mill.integration.AmmoniteTests(fork = false)
    object DocAnnotationsTests extends mill.integration.DocAnnotationsTests(fork = false)
}
package group2 {
    object BetterFilesTests extends mill.integration.BetterFilesTests(fork = false)
    object HygieneTests extends mill.integration.HygieneTests(fork = false)
    object LargeProjectTests extends mill.integration.LargeProjectTests(fork = false)
    object JawnTests extends mill.integration.JawnTests(fork = false)
    object UpickleTests extends mill.integration.UpickleTests(fork = false)
    object PlayJsonTests extends mill.integration.PlayJsonTests(fork = false)
    object CaffeineTests extends mill.integration.CaffeineTests(fork = false)
    object ScriptsInvalidationTests extends mill.integration.ScriptsInvalidationTests(fork = false)
    object ScriptsInvalidationForeignTests
        extends mill.integration.ScriptsInvalidationForeignTests(fork = false)
}
```

So we can run them with:
```
./mill integration.test "mill.integration.local.group1"
```

lefou:
> I don't understand if we are running all the integration tests in CI Some tests appear in some specific script file like:
> 
> ```
> ./mill integration.test "mill.integration.local.{AcyclicTests,AmmoniteTests,DocAnnotationsTests}"
> ```
> 
> but some other don't.
> 
> Am I missing something?

I guess the reason is that some don't run on Windows, some others stopped working with newer JDKs. These integration tests are rather old snapshots of existing projects. Either we drop these that no longer work or we try to upgrade them.

lolgab:
What about grouping them in categories like noWindows, noJVM9 etc. and have a default group where new integration tests get added and we're sure they run?

lefou:
> What about grouping them in categories like noWindows, noJVM9 etc. and have a default group where new integration tests get added and we're sure they run? We can then slowly fix tests and move them to the main group.

Yeah, that's fine with me, as long the groups are meaningful and not "group1".

lefou:
About the Evaluator API change. Is it possible to separate it for this PR? I find this PR already complex enough. Also, the deprecation annotations should also have a "message" field indicating the reason or better a non-deprecated alternative. And the since-annotations can't obviously contain a `0.10.0`, as this is already released.

lolgab:
> About the Evaluator API change. Is it possible to separate it for this PR? I find this PR already complex enough. Also, the deprecation annotations should also have a "message" field indicating the reason or better a non-deprecated alternative. And the since-annotations can't obviously contain a `0.10.0`, as this is already released.

Tackled in https://github.com/com-lihaoyi/mill/pull/1740
I will rebase this PR once the other one lands.

main/core/src/mill/eval/Evaluator.scala
```diff
@@ -637,7 +708,8 @@ class Evaluator @deprecated(message = "Use apply instead", since = "mill 0.10.1"
     workerCache,
     env,
     failFast,
-    threadCount
+    threadCount,
+    this.importTree
```
lefou, line 712: I think this copy method does not fulfil the purpose of bin compat, as it already contains a changed signature.

main/core/src/mill/internal/Utils.scala
```diff
@@ -0,0 +1,24 @@
+package mill.internal
+
+@mill.api.internal
+private[mill] object Utils {
```
lefou, line 4: We already have more than one Uitl(s) class. Can you come up with another name (or prefix) to make reading the code more non-ambiguous? E.g. `AmmoniteUtils`.

lefou (commented):
I started the review. I'm stuck on the change to the test suite runner for today. Maybe, you can motivate your change you did there? I mean the `stdout` stuff.

main/core/src/mill/eval/Evaluator.scala
lolgab: It doesn't. I changed only the implementation here to copy maintaining the old `importTree`

main/core/src/mill/internal/Utils.scala
lolgab: Done ✅

lolgab:
> Maybe, you can motivate your change you did there? I mean the `stdout` stuff.

I needed to test that the tasks were cached and invalidated appropriately so I put `println`s on tasks so by checking what gets printed and what doesn't I can test the task invalidation.
To test I need to read the stdout so I adapted the `ScriptTestSuite` to expose an `evalStdout` function alternative to `eval` that returns a tuple with the Boolean result and a `Seq[String]` with the lines printed to the stdout by the evaluated task.

main/src/mill/main/RunScript.scala
```diff
@@ -64,15 +65,41 @@ object RunScript {
...
+                    rootModule,
+                    rootModule.getClass.getClassLoader.asInstanceOf[
+                      SpecialClassLoader
+                    ].classpathSignature,
+                    mutable.Map.empty[Segments, (Int, Any)],
+                    interp.watchedValues.toSeq,
+                    systemProperties.keySet,
+                    importTree
+                  )
+                }
```
lefou, line 102: I'd be nice, if we could refactor the whole imported script file handling into a separate `def` later.

main/test/src/util/ScriptTestSuite.scala
```diff
@@ -52,20 +52,43 @@ abstract class ScriptTestSuite(fork: Boolean) extends TestSuite {
...
+      val output = Seq.newBuilder[String]
+      val processOutput = os.ProcessOutput.Readlines(output += _)
+
+      val result = evalFork(processOutput, s)
+      (result, output.result())
+    }
+  }
+  private def evalFork(stdout: os.ProcessOutput, s: Seq[String]): Boolean = {
+    try {
+      os.proc(os.home / "mill-release", "-i", s).call(
```
lefou, line 83: Unrelated to your PR, but we should get rid of this hardcoded `$HOME/mill-release` thing.

lefou (commented):
Looks good to merge. I hesitated about the change in the test setup, because it is already quite complex and understanding why and when some output stream is redirected and also used and whether it is test relevant feels like a disservice. But I understand that we want to test this somehow. Maybe we can document the test infrastructure. Or make the simple code paths more obvious. The fact that it took me so long to actually finish my review is a good indicator (to me) that this part of our code base isn't finished yet.

lefou:
Thank you, @lolgab !

main/src/mill/main/RunScript.scala
lolgab: Extracted in a `private def` in the file.

PR #1670, opened 2022-01-12
Fix Scala Native for versions `> 0.4.2`

lolgab:
Workers in Mill work (pun not intended) by compiling to a base version and then substitute the classloader with the dependencies from the project they are compiling.
Before this PR, the Scala Native worker dependency was a uberjar created with `assembly` so it was depending on the Scala Native version defined in Mill's `build.sc`.
Now worker contains only the Worker classes and all the other dependencies are injected by Mill using the `scalaNativeVersion` from the `ScalaNativeModule` it is compiling.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -704,7 +704,7 @@ object scalanativelib extends MillModule {
 
   override def testArgs = T {
     val mapping = Map(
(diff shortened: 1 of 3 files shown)


lefou (approved):
LGTM, Thanks!

PR #1671, opened 2022-01-12
Support ESFeatures in Scala.js

lolgab:
`ScalaJSModule` currently supports the `useECMAScript2015` settings.
Nowadays Scala.js offers support for multiple `ESVersion`.
In the official Sbt plugin they are set inside an object called `ESFeatures` that contains other settings.
This PR tries to mimic the Scala.js API.
It is a breaking change, since overriding the old `useECMAScript2015` doesn't have any effect anymore.

### How deprecate overriding `useECMAScript2015`
Since we can't derive `ESFeatures` from `useECMAScript2015`, I can't use the same deprecation mechanism as `testFrameworks`. What I did instead is to define a private boolean `var` in ScalaJSModule called `overriddenUseECMAScript2015` that is first set to `true`. It is then set to `false` by the default implementation of `useECMAScript2015` which calls back the new `esFeatures`. Before `fastOpt` and `fullOpt` I call a private `checkDeprecations` task, which calls `useECMAScript2015()` and then checks the `overriddenUseECMAScript2015`. If it was overridden then the value remains `true` and it fails the task with an error message; otherwise it finds `false` and continues with the build.

### DIFF
--- a/scalajslib/src/ScalaJSModule.scala
+++ b/scalajslib/src/ScalaJSModule.scala
@@ -74,8 +74,7 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
 
   def scalaJSToolsClasspath = T { scalaJSWorkerClasspath() ++ scalaJSLinkerClasspath() }
 
-  private def checkDeprecations(): Task[Unit] = T.task {
-    pprint.pprintln("Calling checkDeprecations.")
+  private def checkDeprecations: Task[Unit] = T.task {
     useECMAScript2015()
     if(overriddenUseECMAScript2015) {
       Result.Failure("Overriding `useECMAScript2015` is not supported anymore. Override `esFeatures` instead")
@@ -85,7 +84,7 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
   }
 
   def fastOpt = T {
-    // checkDeprecations()
+    checkDeprecations()
     link(
       worker = ScalaJSWorkerApi.scalaJSWorker(),
       toolsClasspath = scalaJSToolsClasspath(),
--- a/scalajslib/src/ScalaJSModule.scala
+++ b/scalajslib/src/ScalaJSModule.scala
@@ -204,7 +204,9 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
 
   def esFeatures: T[ESFeatures] = T {
     if (useECMAScript2015.ctx.enclosing != s"${classOf[ScalaJSModule].getName}#useECMAScript2015") {
-      Result.Failure("Overriding `useECMAScript2015` is not supported anymore. Override `esFeatures` instead")
+      Result.Failure(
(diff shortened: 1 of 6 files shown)


lefou:
@lolgab You may want to look into `JavaModule.bspCompileClassesPath`, where I've implemented a override detection for the `compile` target.

https://github.com/com-lihaoyi/mill/blob/f0358b923b0da3829e9aff636a59e245f0d400a0/scalalib/src/JavaModule.scala#L274

Once we known we have an overridden deprecated target, we can issue a warning message and also use the value to derive the new target, otherwise we can just go with our defaults.

scalajslib/src/ScalaJSModule.scala
```diff
@@ -197,8 +197,22 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
...
-  def useECMAScript2015: T[Boolean] = T { 
-    !scalaJSVersion().startsWith("0.")
+  @deprecated("Use esFeatures().esVersion instead", since = "mill after 0.10.0-M5")
+  def useECMAScript2015: T[Boolean] = T {
+    esFeatures().esVersion != ESVersion.ES5_1
+  }
+
+  def esFeatures: T[ESFeatures] = T {
+    if (useECMAScript2015.ctx.enclosing != s"${classOf[ScalaJSModule].getName}#useECMAScript2015") {
+      Result.Failure(
```
lefou, line 207: Is it really necessary to fail the build? Can't we just use either ES5_1 or ES2015 when `useECMAScript2015` was overridden? I think the big red warning might be annoying enough on the CLI, and in the editor the deprecation warning also indicates the outdated target. It is far more important to make transition to newest Mill as easy as possible. The less likely older Mill versions are used, the better.

scalajslib/src/ScalaJSModule.scala
```diff
@@ -197,8 +197,22 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
 
   def moduleKind: T[ModuleKind] = T { ModuleKind.NoModule }
 
-  def useECMAScript2015: T[Boolean] = T { 
-    !scalaJSVersion().startsWith("0.")
+  @deprecated("Use esFeatures().esVersion instead", since = "mill after 0.10.0-M5")
+  def useECMAScript2015: T[Boolean] = T {
+    esFeatures().esVersion != ESVersion.ES5_1
```
lefou, line 202: I would rather just keep the old default, so we can depend on it in `esFeatures`.

scalajslib/src/ScalaJSModule.scala
lolgab: You are right! I couldn't see how to make it work.
But actually the new defaults are all expressible using the old ones.

scalajslib/src/ScalaJSModule.scala
lolgab: Done ✅

lefou (approved):
LGTM, thanks!

PR #1673, opened 2022-01-12
Support Scala 3 in `ScalaNativeModule`

lolgab:
- Improve handling of env variables to now throw exceptions
- Remove outdated test for test modules with no test (it is not handled by Mill anymore but fixed upstream)
- Use correct artifacts for scalalib in Scala 3

### DIFF
--- a/scalanativelib/src/ScalaNativeModule.scala
+++ b/scalanativelib/src/ScalaNativeModule.scala
@@ -82,10 +82,15 @@ trait ScalaNativeModule extends ScalaModule { outer =>
     super.mandatoryIvyDeps() ++ nativeIvyDeps()
   }
 
+  // Using `resolveDeps` from `CoursierModule` incorrectly resolves
+  // scala-library 2.13 instead of the required 2.12
   def bridgeFullClassPath: T[Agg[os.Path]] = T {
-    val workerClasspath = scalaNativeWorkerClasspath()
-    val resolvedIvyDeps = resolveDeps(toolsIvyDeps)()
-    (workerClasspath ++ resolvedIvyDeps).map(_.path)
+    Lib.resolveDependencies(
+      repositoriesTask(),
+      resolveCoursierDependency().apply(_),
+      toolsIvyDeps(),
+      ctx = Some(T.log)
+    ).map(t => (scalaNativeWorkerClasspath() ++ t).map(_.path))
   }
 
   override def scalacPluginIvyDeps: T[Agg[Dep]] = T {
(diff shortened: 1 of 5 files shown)


scalanativelib/src/ScalaNativeModule.scala
```diff
@@ -77,7 +85,7 @@ trait ScalaNativeModule extends ScalaModule { outer =>
 
   def bridgeFullClassPath = T {
     Lib.resolveDependencies(
-      Seq(coursier.LocalRepositories.ivy2Local, MavenRepository("https://repo1.maven.org/maven2")),
+      repositories = repositoriesTask(),
       Lib.depToDependency(_, scalaVersion(), platformSuffix()),
       toolsIvyDeps(),
       ctx = Some(implicitly[mill.util.Ctx.Log])
```
lefou, line 91: This can be replaced by `resolveDeps(toolsIvyDeps())`.

scalanativelib/test/src/HelloNativeWorldTests.scala
```diff
@@ -22,12 +23,12 @@ object HelloNativeWorldTests extends TestSuite {
     override def mainClass = Some("hello.Main")
   }
 
-  val scala213 = "2.13.6"
+  val scala213 = "2.13.7"
```
lefou, line 26: We probably might want to derive this from what is defined in `build.sc` (for easier maintenance), either by generated build info classes or by system properties. An example can be found in `contrib.scoverage`.

lefou (commented):
Your PR looks reasonable to me, but I'm not very into Scala Native. I added some comments below.

scalanativelib/test/src/HelloNativeWorldTests.scala
lolgab: I noticed that I can't update this version since Scala Native 0.4.0 is not published for newer Scala versions.
I reverted the change.

scalanativelib/src/ScalaNativeModule.scala
lolgab: Done ✅

lolgab:
I took the opportunity to make some breaking changes.
I don't think the ScalaNativeModule api is something people rely on, so I changed some targets to have `Agg` instead of `Seq` consistently with other modules.
I also removed the `nativeIvyDep` that wasn't used anywhere and I don't think it can be useful to have.

scalanativelib/src/ScalaNativeModule.scala
lolgab: I had to revert the change because then it was resolving scala-library 2.13 instead of 2.12 (which is the correct one for tools).
My best guess is that it resolves the wrong one because of this https://github.com/com-lihaoyi/mill/issues/860
Added a comment to explain the situation.

lefou (approved):
Looks reasonable.

lefou:
Thank you!

PR #1678, opened 2022-01-14
Fixes the default args for publish

markehammons:
It seems the default argument for gpgArgs in publish cannot ever resolve to anything but empty. So, if the argument is an empty Seq, replace it with the default gpgArgs. 

This allows projects like mine to publish easily.

### DIFF
--- a/scalalib/src/PublishModule.scala
+++ b/scalalib/src/PublishModule.scala
@@ -155,6 +155,11 @@ trait PublishModule extends JavaModule { outer =>
   def publish(
       sonatypeCreds: String,
       signed: Boolean = true,
+      // mainargs wasn't handling a default value properly, 
+      // so we instead use the empty Seq as default.
+      // see https://github.com/com-lihaoyi/mill/pull/1678
+      // TODO: In mill 0.11, we may want to change to a String argument 
+      // which we can split at `,` symbols, as we do in `PublishModule.publishAll`.
       gpgArgs: Seq[String] = Seq.empty,
       release: Boolean = false,
       readTimeout: Int = 60000,
--- a/scalalib/src/PublishModule.scala
+++ b/scalalib/src/PublishModule.scala
@@ -155,7 +155,7 @@ trait PublishModule extends JavaModule { outer =>
   def publish(
       sonatypeCreds: String,
       signed: Boolean = true,
-      gpgArgs: Seq[String] = PublishModule.defaultGpgArgs,
+      gpgArgs: Seq[String] = Seq.empty,
       release: Boolean = false,
       readTimeout: Int = 60000,
       connectTimeout: Int = 5000,
@@ -168,7 +168,7 @@ trait PublishModule extends JavaModule { outer =>
       sonatypeSnapshotUri,
       sonatypeCreds,
       signed,
-      gpgArgs.flatMap(_.split("[,]")),
+      if (gpgArgs.isEmpty) PublishModule.defaultGpgArgs else gpgArgs,
       readTimeout,
       connectTimeout,
       T.log,


markehammons:
The motivation for this PR is that publish, as written, requires multiple `--gpgArgs` to work, as the default is never used. the publishAll command mentioned in the docs is harder for me to use than publish can be, thanks to the fact that publishAll uses a default sonatype url that is not useable for new sonatype publishers (all us newbies are being asked to publish to `https://s01.oss.sonatype.org/service/local`).

lefou (commented):
I still not really understand why the default argument(s) does not work (for you). Is it an issue with mainargs library? Or do we something not supported by mainargs?

I can see that we also have a `gpgArgs` argument in external `PublishModule.publishAll` target, which does work. It only differs in the type (`Seq[String]` instead `String`). 

Besides that I can't reconstruct your initial issue, I'd like to avoid your solution, as it hides the actual defaults from the (code) reader at first glance. It would be nice if we could take advantage of the clarity a pure Scala default argument provides.

lefou:
Btw, the `PublishModule.publishAll` does support a `--sonatypeUri`.

markehammons:
yes, publishAll has sonatype uri. My problem is I'm trying to use `./mill _.publish "credentials" --release false` and getting a gpg error: 

```
gpg: WARNING: no command supplied.  Trying to guess what you mean ...
gpg: no valid OpenPGP data found.
gpg: processing message failed: Unknown system error
1 targets failed
polymorphics.publish os.SubprocessException: CommandResult 2

    os.proc.call(ProcessOps.scala:85)
    mill.scalalib.publish.SonatypePublisher.gpgSigned(SonatypePublisher.scala:178)
    mill.scalalib.publish.SonatypePublisher.$anonfun$publishAll$4(SonatypePublisher.scala:44)
    scala.collection.immutable.List.map(List.scala:246)
    mill.scalalib.publish.SonatypePublisher.$anonfun$publishAll$2(SonatypePublisher.scala:43)
    scala.collection.Iterator$$anon$9.next(Iterator.scala:577)
    scala.collection.mutable.Growable.addAll(Growable.scala:62)
    scala.collection.mutable.Growable.addAll$(Growable.scala:57)
    scala.collection.mutable.ArrayBuilder.addAll(ArrayBuilder.scala:66)
    scala.collection.IterableOnceOps.toArray(IterableOnce.scala:1282)
    scala.collection.IterableOnceOps.toArray$(IterableOnce.scala:1276)
    scala.collection.AbstractIterable.toArray(Iterable.scala:926)
    scala.Array$.from(Array.scala:73)
    scala.collection.immutable.ArraySeq$.from(ArraySeq.scala:282)
    scala.collection.immutable.ArraySeq$.from(ArraySeq.scala:273)
    scala.collection.ClassTagIterableFactory$AnyIterableDelegate.from(Factory.scala:671)
    scala.collection.IterableOps$WithFilter.map(Iterable.scala:891)
    mill.scalalib.publish.SonatypePublisher.publishAll(SonatypePublisher.scala:34)
    mill.scalalib.publish.SonatypePublisher.publish(SonatypePublisher.scala:30)
    mill.scalalib.PublishModule.$anonfun$publish$2(PublishModule.scala:177)
```
to fix this I have to write `./mill _.publish "credentials" --release true --gpgArgs --batch --gpgArgs -a --gpgArgs -b  `, because the default gpgArgs on the `publish` task just doesn't work.

lefou:
Thanks to clarifying.

lefou (approved):
Looks good to me, thanks!

scalalib/src/PublishModule.scala
```diff
@@ -155,7 +155,7 @@ trait PublishModule extends JavaModule { outer =>
   def publish(
       sonatypeCreds: String,
       signed: Boolean = true,
-      gpgArgs: Seq[String] = PublishModule.defaultGpgArgs,
+      gpgArgs: Seq[String] = Seq.empty,
```
lefou, line 158:
```suggestion
      // TODO: mainargs seem to not does a default here, we may want to switch
      // to a String argument which we split at `,` symbols, as we do in `publishAll`.
      gpgArgs: Seq[String] = Seq.empty,
```

PR #1687, opened 2022-01-22
New sonatype URLs

nrktkt:
Looks like current hosts are now are legacy

### DIFF
--- a/scalalib/src/PublishModule.scala
+++ b/scalalib/src/PublishModule.scala
@@ -130,9 +130,9 @@ trait PublishModule extends JavaModule { outer =>
       ).map(PathRef(_))
   }
 
-  def sonatypeUri: String = "https://oss.sonatype.org/service/local"
+  def sonatypeUri: String = "https://s01.oss.sonatype.org/service/local"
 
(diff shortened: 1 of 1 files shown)


lefou:
Can you please provide a source for your findings?

nrktkt:
https://issues.sonatype.org/browse/OSSRH-76812

> Note: As of February 2021, all new projects began being provisioned on https://s01.oss.sonatype.org/. If your project is not provisioned on https://s01.oss.sonatype.org/, you will want to login to the legacy host https://oss.sonatype.org/.

https://central.sonatype.org/publish/publish-guide/#deployment

lefou:
It's a pity Sonatype hasn't some dispatching facade for it, but I'm not going to change these defaults. As these are just defaults, they can and should be overridden by the user. It's probably questionable that we have a default at all, but changing it does not make much sense, as It will break lots of workflows including Mills one, and it will still be the wrong URL for other users.

lefou:
We may rething, once the legacy host is shut down and _all_ user are migrated to the new one.

PR #1692, opened 2022-01-22
GenIdea: Use common content root and allow optional package prefixes

Fred4106:
Allows a build.sc like 
```scala
import mill._
import mill.scalalib._

object foo extends ScalaModule{
  def scalaVersion = "3.1.0"
  override def sources = T.sources(
    millSourcePath / "src",
    millSourcePath / "src2",
  )
  override def packagePrefix: Map[String, String] = Map(
    "src" -> "foo",
  )
}
```
to produce a module.xml like this
```xml
<module type="JAVA_MODULE" version="4">
    <component name="NewModuleRootManager">
        <output url="file://$MODULE_DIR$/../../out/foo/ideaCompileOutput.dest/classes"/>
        <exclude-output/>
        <content url="file://$MODULE_DIR$/../../foo">
            <sourceFolder url="file://$MODULE_DIR$/../../foo/src" isTestSource="false" packagePrefix="foo"/>
            <sourceFolder url="file://$MODULE_DIR$/../../foo/src2" isTestSource="false"/>
            <sourceFolder url="file://$MODULE_DIR$/../../foo/resources" type="java-resource"/>
        </content>
        <orderEntry type="inheritedJdk"/>
        <orderEntry type="library" name="scala-library-2.13.6.jar" level="project"/>
        <orderEntry type="library" name="scala3-library_3-3.1.0.jar" level="project"/>
    </component>
</module>
```
module settings

Compare this to the current implementation
```scala
import mill._
import mill.scalalib._

object foo extends ScalaModule{
  def scalaVersion = "3.1.0"
  override def sources = T.sources(
    millSourcePath / "src",
    millSourcePath / "src2",
  )
}
```
```xml
<module type="JAVA_MODULE" version="4">
    <component name="NewModuleRootManager">
        <output url="file://$MODULE_DIR$/../../out/foo/ideaCompileOutput.dest/classes"/>
        <exclude-output/>
        <content url="file://$MODULE_DIR$/../../foo/src">
            <sourceFolder url="file://$MODULE_DIR$/../../foo/src" isTestSource="false"/>
        </content>
        <content url="file://$MODULE_DIR$/../../foo/src2">
            <sourceFolder url="file://$MODULE_DIR$/../../foo/src2" isTestSource="false"/>
        </content>
        <content url="file://$MODULE_DIR$/../../foo/resources">
            <sourceFolder url="file://$MODULE_DIR$/../../foo/resources" type="java-resource"/>
        </content>
        <orderEntry type="inheritedJdk"/>
        <orderEntry type="library" name="scala-library-2.13.6.jar" level="project"/>
        <orderEntry type="library" name="scala3-library_3-3.1.0.jar" level="project"/>
    </component>
</module>
```
module settings

### DIFF
--- a/scalalib/src/GenIdeaImpl.scala
+++ b/scalalib/src/GenIdeaImpl.scala
@@ -3,11 +3,11 @@ package mill.scalalib
 import scala.collection.immutable
 import scala.util.Try
 import scala.xml.{Elem, MetaData, Node, NodeSeq, Null, UnprefixedAttribute}
-
 import ammonite.runtime.SpecialClassLoader
 import coursier.core.compatibility.xmlParseDom
 import coursier.maven.Pom
 import coursier.{LocalRepositories, Repositories, Repository}
+
 import java.nio.file.Paths
 import mill.Agg
 import mill.api.Ctx.{Home, Log}
@@ -18,6 +18,8 @@ import mill.modules.Util
 import mill.scalalib.GenIdeaModule.{IdeaConfigFile, JavaFacet}
 import mill.{BuildInfo, T, scalalib}
 import os.{Path, SubPath}
+
+import scala.util.chaining.scalaUtilChainingOps
 
 case class GenIdeaImpl(
     evaluator: Evaluator,
@@ -243,6 +245,7 @@ case class GenIdeaImpl(
 
     val allResolved: Seq[Path] =
       (resolvedModules.flatMap(_.classpath).map(_.value) ++ buildLibraryPaths ++ buildDepsPaths)
+        .filter(_.toIO.exists())
         .distinct
         .sorted
 
@@ -411,7 +414,7 @@ case class GenIdeaImpl(
         case WithSourcesResolved(path, _) =>
           Seq(pathToLibName(path))
         case OtherResolved(path) =>
-          Seq(pathToLibName(path))
+          pathToLibName.get(path).map(Seq(_)).getOrElse(Seq.empty)
       }
 
     def resolvedLibraries(resolved: Seq[os.Path]): Seq[ResolvedLibrary] =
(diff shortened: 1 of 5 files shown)


lefou:
I'm linking PR https://github.com/com-lihaoyi/mill/pull/873 as I think, this one introduced the explicit content roots. I don't remember to full detail, but I think the issue is, that this change resulted in IDEA handling shared source dirs much better. Unfortunately, there is not test case to validate this, as it is mostly runtime behavior we are dealing with.

Also, I think skipping empty entry (dirs) isn't a good idea, as this means we need to regenerate the IDEA project whenever we change some files, which can happen with each git checkout for example. Currently, we only need to regenerate once we made significant changes to the build structure or settings.

lefou:
@Fred4106 Do you intend to update this PR. I'd like to get some features of commit https://github.com/com-lihaoyi/mill/pull/1692/commits/14b69c9539074846207bc40b2456a5285d152700 in, namely Scala 3.1 and prefix suppport.

scalalib/src/GenIdeaImpl.scala
```diff
@@ -411,7 +414,7 @@ case class GenIdeaImpl(
         case WithSourcesResolved(path, _) =>
           Seq(pathToLibName(path))
         case OtherResolved(path) =>
-          Seq(pathToLibName(path))
+          pathToLibName.get(path).map(Seq(_)).getOrElse(Seq.empty)
```
lefou, line 417: Do you fix an issue here? If yes, which?

scalalib/src/GenIdeaModule.scala
```diff
@@ -17,6 +17,11 @@ trait GenIdeaModule extends Module {
    */
   def skipIdea: Boolean = false
 
+  /**
+   * Map containing a map of source folder names to package prefixes for Idea project file generation.
+   */
+  def packagePrefix: Map[String, String] = Map.empty[String, String]
```
lefou, line 23: Just using folder names isn't enough. Consider two `src` folders in different parent directories. Best is to use a `os.Path` as key.

Fred4106:
I'll be updating my or this weekend. Been busy with work and am by no means a scala expert. I'll see what I can get done though and see where it goes.

lefou:
Closing due to inactivity.

PR #1693, opened 2022-01-22
Make ScalaNativeModule use Scala 2.13 toolchain when available

lolgab:
# Motivation

- Consistency with `ScalaJSModule`
- Not resolving Scala 2.12 if not needed (imagining a Scala 2.13 only project) could make the build faster
- Not loading a new classloader from scratch and populating it with a brand new scala-library for Scala 2.12 could be faster (I have no data to prove this theory)

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -68,10 +68,10 @@ object Deps {
   }
 
   object Scalanative_0_4 {
-    val scalanativeTools = ivy"org.scala-native::tools:0.4.0"
-    val scalanativeUtil = ivy"org.scala-native::util:0.4.0"
-    val scalanativeNir = ivy"org.scala-native::nir:0.4.0"
-    val scalanativeTestRunner = ivy"org.scala-native::test-runner:0.4.0"
+    val scalanativeTools = ivy"org.scala-native::tools:0.4.2"
+    val scalanativeUtil = ivy"org.scala-native::util:0.4.2"
+    val scalanativeNir = ivy"org.scala-native::nir:0.4.2"
+    val scalanativeTestRunner = ivy"org.scala-native::test-runner:0.4.2"
   }
 
   val acyclic = ivy"com.lihaoyi::acyclic:0.2.1"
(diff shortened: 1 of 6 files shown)


lolgab:
Tested with Scala Native 0.4.0 and it fails with:
```
1 targets failed
module.bridgeFullClassPath 
Resolution failed for 2 modules:
  org.scala-native:test-runner_2.13:0.4.1 
        not found: /Users/lorenzo/.ivy2/local/org.scala-native/test-runner_2.13/0.4.1/ivys/ivy.xml
        not found: https://repo1.maven.org/maven2/org/scala-native/test-runner_2.13/0.4.1/test-runner_2.13-0.4.1.pom
  org.scala-native:tools_2.13:0.4.1 
        not found: /Users/lorenzo/.ivy2/local/org.scala-native/tools_2.13/0.4.1/ivys/ivy.xml
        not found: https://repo1.maven.org/maven2/org/scala-native/tools_2.13/0.4.1/tools_2.13-0.4.1.pom
```
Given the very low usage of Scala Native + Mill I'd say this error is good enough, because you just need to follow the maven link to see that the first `tools_2.13` and `test-runner_2.13` artifacts are of version `0.4.2`:
https://repo1.maven.org/maven2/org/scala-native/test-runner_2.13/

Also Scala Native 0.4.3 adds Scala 3 support so it is likely everyone will migrate to it very very soon.

lefou (commented):
As I'm not using Scala Native myself, I have no deep understanding how people are using it and how easy it is to upgrade. Are there any real issues for those still using Mill + SN 0.4.0 or 0.4.1 to just use 0.4.2 instead? Can they just bump and are good to go, or are there potential incompatibilities and required changes? Are they forced to also have all dependencies released for at least 0.4.2?

If there are no real hurdles for SN users to upgrade, than your PR is reasonable. Otherwise, we should consider introducing just a new second worker which is essentially what you provided here, and keep the old one for those forces to stay a bit longer with SN 0.4.{0,1}. I think the arguments to avoid loading another Scala version with comes at a time and memory sounds nice, but should not drive the decision, if there is also potential we loose some users.

And finally a note to myself: We should invide users to add theirself to some users list, so we have a better understanding how many users are using which features of mill. Unfortunately, as I can't access the Sonatype staging group, I can't see the download statistics for those artifacts.

lefou (approved):
This looks good to me. I appreciate the efforts you made to keep compatibility with SN 0.4.0.

lefou:
I merged via merge commit, to preserve the result of having only one Scala 2.13 based worker.

PR #1694, opened 2022-01-22
Skip Mima checks in worker modules

lolgab:
- Introduce MillInternalModule in build
- Make workers extend MillInternalModule

## Motivation

Workers are internal components that shouldn't give any binary compatibility guaranties.
This removes Mima from them so we are free to change them whenever we want.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -192,14 +192,14 @@ trait MillMimaConfig extends mima.Mima {
   )
 }
 
-trait MillApiModule
+trait MillInternalModule
     extends MillPublishModule
     with ScalaModule
-    with MillCoursierModule
-    with MillMimaConfig {
+    with MillCoursierModule {
   def scalaVersion = Deps.scalaVersion
   override def ammoniteVersion = Deps.ammonite.dep.version
 }
+trait MillApiModule extends MillInternalModule with MillMimaConfig
 
 trait MillModule extends MillApiModule { outer =>
   override def scalacPluginClasspath =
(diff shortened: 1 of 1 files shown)


lefou (commented):
What is this magic, that makes MiMa ignore them? Does MiMa guess based on the "internal" word?

Btw, we already have an `@mill.api.internal` annotation for exactly that case. I'd rather avoid to manifest tooling configuration via invasive types.

lolgab:
There is no magic, I'm just avoid adding the mima plugin in these modules.

The annotation is not enough. I'm the Scala Native PR I opened the mima check failed because the 2.13 artifact of the worker doesn't exist at all. No annotation can fix this.

lefou:
Sorry, I must have completely overlooked the fact that you introduced the trait in the build, not the code. This make sense, I agree. 

But let me just discuss MiMa again. I've stumbled about an issue before, that you can't have MiMa configured with an empty set of previous artifacts. I know, that this helps to avoid misconfiguration, but I think it's a common case, e.g. in case when one releases the first version of a new compatibility series. My workaround was to apply a patch to the build and remove the MiMa entirely, whereas it would be much nicer if we just could provide an empty set of previous versions or artifacts. 

Presumed, we could configure MiMa that way to disable it for workers modules, wouldn't this be the preferred solution?

lolgab:
Yes, that would be desirable and we probably want to add that. At the same time, having some internal modules loaded via reflection in a separate class path hanging

lefou (approved):
I think the reason why I resisted this change initially is because our build is already hard to understand, especially to reason about when to use which common trait when creating new modules. This can be probably improved by just documenting the purpose of each trait.

Your change makes sense. Thanks!

lolgab:
Great thank you!

Issue #1691, opened 2022-01-22
Support Scala 3 compiler plugins


lolgab:
Scala 3 plugins are different from Scala 2 plugins.
https://docs.scala-lang.org/scala3/reference/changed-features/compiler-plugins.html

Trying to update Scala Native to Scala 3 in Utest ( which uses a compiler plugin ) failed with this log:
[message.txt](https://github.com/com-lihaoyi/mill/files/7918658/message.txt)

It was suggested in the Scala Native discord that:
> It looks like -Xplugin:path points to something which is not a compiler plugin
In Scala2 there was a xml file describing plugin, in Scala 3 there is plugin.properties file

It seems that Mill adds `-Xplugin` to all the jars in the plugin classpath, while Scala 3 seems to require this to be passed only in the plugin jar.

lefou:
See this related issue #1690, with an extended discussion about plugin handling in MIll.

lolgab:
Thank you!
So this is another problem which is also happening in Scala 2!

lefou:
Yes, our current handling for scalac plugins is incorrect. I outlined two solutions here https://github.com/com-lihaoyi/mill/issues/1690#issuecomment-1019326117, and I would value your thoughs / expertise, @lolgab .

Quoting it here for completeness:
> Maybe, we are not supposed to resolve compiler plugins transitively? I could not find any information about this in the scala documentation. (We, @ckipp01 and me, also spoke about this before in context of scoverage plugin.) I think most plugins do not require additional dependencies but that's just a guess. The best transitive dependency mapping for compiler plugins would be to declare all API dependencies, which should be provided by the compiler itself (like scala library jar and compiler jar) as Maven provided scope, or in Mill speak as compile-only dependencies. Unfortunately we don't have control over this so I see two options to improve the situation:
> 
> 1. Never resolve compiler plugins transitively. This is essentially the same as providing each with `.exclude("*" -> "*")`. This will change behavior, as the resulting compiler classpath is likely smaller. This change might break compilations. It looks like this is what sbt does (from your given example above).
> 
> 2. Only generate `-Xplugin` options from the directly listed ivy deps but still provide all transitive dependencies as compiler classpath. This will only generate less options, which seems the right thing to do but still provides all declared transitive dependencies. This will not break currently succeeding compilations, but might still violate the expectation of the compiler, as we potentially provide to much (conflicting) dependencies.

lefou:
[...]

Issue #1690, opened 2022-01-22
Bloop export issues with the plugin classpath


ckipp01:
So I'm not really sure how I haven't noticed this before, but it just bit me for the fist time and I'm trying to figure out where a change should happen. However I noticed today when looking at [`scala-cli`](https://scala-cli.virtuslab.org/) and trying to do a bloop export with the build that there were some errors being thrown in the logs about not being able to find the `scalac-plugin.xml` file. This was then seemingly causing issues with indexing and causing navigation not to work in the project.

_Example of the error_
```
2022.01.22 15:54:11 ERROR /Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/modules/scala-xml_2.12/1.0.6/scala-xml_2.12-1.0.6.jar
java.nio.file.NoSuchFileException: /scalac-plugin.xml
```

I thought it was odd that Metals was looking for a `scalac-plugin.xml` for the `scala-xml` dep, but then I was able to figure out that the main issue was that the Bloop config that was being generated had everything in the `scalacPluginClasspath()` prefaced with `-Xplugin`. For example, here is what was generated for the cli module of `scala-cli` under the `scala` key:

```
"scala": {
    "organization": "org.scala-lang",
    "name": "scala-compiler",
    "version": "2.12.15",
    "options": [
        "-Xasync",
        "-Ywarn-unused",
        "-deprecation",
        "-Xplugin:/Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scalamacros/paradise_2.12.15/2.1.1/paradise_2.12.15-2.1.1.jar",
        "-Xplugin:/Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scalameta/semanticdb-scalac_2.12.15/4.4.31/semanticdb-scalac_2.12.15-4.4.31.jar",
        "-Xplugin:/Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-compiler/2.12.15/scala-compiler-2.12.15.jar",
        "-Xplugin:/Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.12.15/scala-library-2.12.15.jar",
[...]

PR #1705, opened 2022-01-23
Explicit handling of scalac plugins and related enablement options

lefou:
We no longer generate an enablement option for each transitively resolved plugin jar, but instead, only generated one `-Xplugin:` option for each explicitly given plugin. This will fix previously seen warnings for the scalac compiler.

I took the opportunity to clean things up on the go.
I removed the handling of `-Xplugin:` from the ZincWorker into the ScalaModule.
This fixes the issue that both the BSP server implementation and the Bloop Config exporter needed to handle this on their own.
Now, this is handled in the IMHO right place, is properly shared and still overrideable by the user if the situation calls for it.

I recognized, that we don't support scalac plugin dependencies from in-project modules at the moment, but this can be added later if needed. Previously, it was enough to add the `module.jar()` to the `scalacPluginClassapth`, now it is also required to add the ```s"-Xplugin:{module.jar().path}"``` manually. Not to hard, but a future `scalacPluginModuleDeps` target could make this easier. As this also hits Mills build when we bootstrap our own project, that's why I did that change to `build.sc`.

Fix https://github.com/com-lihaoyi/mill/issues/1690
Fix https://github.com/com-lihaoyi/mill/issues/1691

### DIFF
--- a/scalalib/src/ScalaModule.scala
+++ b/scalalib/src/ScalaModule.scala
@@ -3,13 +3,13 @@ package scalalib
 
 import scala.annotation.nowarn
 import mill.define.{Command, Sources, Target, Task}
-import mill.api.{DummyInputStream, Loose, PathRef, Result, internal}
+import mill.api.{DummyInputStream, PathRef, Result, internal}
 import mill.modules.Jvm
 import mill.modules.Jvm.createJar
 import Lib._
 import ch.epfl.scala.bsp4j.{BuildTargetDataKind, ScalaBuildTarget, ScalaPlatform}
 import mill.api.Loose.Agg
-import mill.eval.{Evaluator, EvaluatorPathsResolver}
+import mill.eval.EvaluatorPathsResolver
 import mill.scalalib.api.{CompilationResult, ZincWorkerUtil}
 import mill.scalalib.bsp.{BspBuildTarget, BspModule}
 
@@ -191,7 +191,7 @@ trait ScalaModule extends JavaModule { outer =>
   }
 
   // Keep in sync with [[bspCompileClassesInfo]]
-  override def compile: T[mill.scalalib.api.CompilationResult] = T.persistent {
+  override def compile: T[CompilationResult] = T.persistent {
     val sv = scalaVersion()
     if (sv == "2.12.4") T.log.error(
       """Attention: Zinc is known to not work properly for Scala version 2.12.4.
--- a/scalalib/src/ScalaModule.scala
+++ b/scalalib/src/ScalaModule.scala
@@ -2,9 +2,8 @@ package mill
 package scalalib
 
 import scala.annotation.nowarn
-
 import mill.define.{Command, Sources, Target, Task}
-import mill.api.{DummyInputStream, PathRef, Result, internal}
+import mill.api.{DummyInputStream, Loose, PathRef, Result, internal}
 import mill.modules.Jvm
 import mill.modules.Jvm.createJar
 import Lib._
@@ -13,8 +12,8 @@ import mill.api.Loose.Agg
 import mill.eval.{Evaluator, EvaluatorPathsResolver}
 import mill.scalalib.api.{CompilationResult, ZincWorkerUtil}
 import mill.scalalib.bsp.{BspBuildTarget, BspModule}
-import scala.jdk.CollectionConverters._
 
+import scala.jdk.CollectionConverters._
 import mainargs.Flag
 
 /**
@@ -86,7 +85,7 @@ trait ScalaModule extends JavaModule { outer =>
       publish.Artifact.fromDep(
         _: Dep,
         scalaVersion(),
-        mill.scalalib.api.Util.scalaBinaryVersion(scalaVersion()),
+        ZincWorkerUtil.scalaBinaryVersion(scalaVersion()),
         platformSuffix()
       )
     }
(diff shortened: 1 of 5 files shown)


scalalib/src/ScalaModule.scala
```diff
@@ -104,6 +104,14 @@ trait ScalaModule extends JavaModule { outer =>
    */
   protected def mandatoryScalacOptions = T { Seq.empty[String] }
 
+  /**
+   * Scalac options to active the compiler plugins.
+   */
+  protected def enablePluginScalacOption: Target[Seq[String]] = T {
```
lolgab, line 110: What do you think of making this `private` until we're sure this is the way to go? Otherwise we will need to maintain this for binary compatibility until `0.11`.
Also I don't get the naming. Should it be plural like `enablePluginScalacOptions` ?
What about something like `compilerPluginsScalacOptions` ?

lolgab (commented):
Left a comment about visibility and naming

scalalib/src/ScalaModule.scala
```diff
@@ -104,6 +104,14 @@ trait ScalaModule extends JavaModule { outer =>
...
+    val resolvedJars = resolveDeps(scalacPluginIvyDeps.map(_.map(_.exclude("*" -> "*"))))()
+    resolvedJars.iterator.toSeq.map(jar => s"-Xplugin:${jar}")
```
lolgab, line 112: `jar` is a `PathRef` here. Also if you delay the creation of the Seq it avoid allocating an extra `Seq`
```suggestion
    resolvedJars.iterator.map(jar => s"-Xplugin:${jar.path}").toSeq
```

lolgab (commented):
Github wants me to add a comment otherwise I can't send inline comments.

lolgab:
Also docJar should use the new target: https://github.com/com-lihaoyi/mill/blob/1af99f568340c6079bc9beba116a396c3067b317/scalalib/src/ScalaModule.scala#L236

scalalib/src/ScalaModule.scala
lefou: You're early to the party. I'm still editing, but also juggling with family time and waiting for CI. The plural form is already fixed. The private makes sense. The enable because this is not "all" plugin options but only those we infer from the ivy deps.

scalalib/src/ScalaModule.scala
lefou: Good catch.

scalalib/src/ScalaModule.scala
lolgab: > You're early to the party.

Definitely! It felt almost like pair programming :-)
I'll leave you until it is ready for review :)

lefou: haha. We can try pair programming some time tough.

scalalib/src/ScalaModule.scala
lefou: To my defense, I tried to use BSP in IJ... didn't work so well, obviously.

scalalib/src/ScalaModule.scala
```diff
@@ -86,35 +85,53 @@ trait ScalaModule extends JavaModule { outer =>
...
+   */
+  private def enablePluginScalacOptions: Target[Seq[String]] = T {
+    val resolvedJars = resolveDeps(scalacPluginIvyDeps.map(_.map(_.exclude("*" -> "*"))))()
+    resolvedJars.iterator.map(jar => s"-Xplugin:${jar.path}").toSeq
+  }
+
+  /**
+   * Scalac options to active the compiler plugins.
+   */
+  private def enableScalaDocPluginScalacOption: Target[Seq[String]] = T {
```
lolgab, line 117: This should be plural, right? It is a `Seq[String]`..

lolgab (commented):
You know you're doing something right, when to fix a bug you also simplify code. Now everything makes more sense and bloop config and other take directly the Mill output without any modification. Congrats!
I left a little comment about a name.
Great the new targets are private so we don't have to commit on the naming/existence.

scalalib/src/ScalaModule.scala
lefou: Yes. Fixed.

scalalib/src/ScalaModule.scala
lefou: I thought about the name that you suggested but discarded it because of my argument above.

lolgab (approved):

scalalib/src/ScalaModule.scala
```diff
@@ -2,9 +2,8 @@ package mill
 package scalalib
 
 import scala.annotation.nowarn
-
 import mill.define.{Command, Sources, Target, Task}
-import mill.api.{DummyInputStream, PathRef, Result, internal}
+import mill.api.{DummyInputStream, Loose, PathRef, Result, internal}
```
ckipp01, line 6: It looks like `Loose` isn't being used unless I'm missing it.

ckipp01 (approved):
Nice! From trying it out the errors you normally see in the Metals logs about the unfound scalac-plugin.xml files are gone:

_Not More 🚀 _
```
2022.01.22 15:54:11 ERROR /Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/modules/scala-xml_2.12/1.0.6/scala-xml_2.12-1.0.6.jar
java.nio.file.NoSuchFileException: /scalac-plugin.xml
```

And as I assumed because of the above, the Bloop export looks correct as well:
```
        "scala": {
            "organization": "org.scala-lang",
            "name": "scala-compiler",
            "version": "2.12.15",
            "options": [
                "-Xplugin:/Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scalamacros/paradise_2.12.15/2.1.1/paradise_2.12.15-2.1.1.jar",
                "-Ywarn-unused"
            ],
            "jars": [
                "/Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-compiler/2.12.15/scala-compiler-2.12.15.jar",
                "/Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-reflect/2.12.15/scala-reflect-2.12.15.jar",
                "/Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.12.15/scala-library-2.12.15.jar",
                "/Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/modules/scala-xml_2.12/1.0.6/scala-xml_2.12-1.0.6.jar"
            ]
        },
```

scalalib/src/ScalaModule.scala
lefou: Good catch. IJ always sneaks it in, when I let it add return types.

Issue #1684, opened 2022-01-18
Scala.js linking to support multiple modules


shishkin:
Hi! I'm running into an issue trying to compile two top-level exports `@JSExportTopLevel(moduleID = "foo", name = "handler")` and `@JSExportTopLevel(moduleID = "bar", name = "handler")`.

Here is the error I see:

```
1 targets failed
lambda.fastOpt The linker produced a result not supported by the pre v1.3.0 legacy API. Call the overload taking an OutputDirectory instead. Linking returned more than one public module. Full report:
Report(
  publicModules = [
    Module(
  moduleID      = foo,
  jsFileName    = foo.js,
  sourceMapName = Some(foo.js.map),
  moduleKind    = CommonJSModule,
),
Module(
  moduleID      = bar,
  jsFileName    = bar.js,
  sourceMapName = Some(bar.js.map),
  moduleKind    = CommonJSModule,
)
  ],
)
```

I've already set the CommonJS ModuleKind:

```scala
  def scalaVersion = "3.1.0"
  def scalaJSVersion = "1.8.0"
  def moduleKind = ModuleKind.CommonJSModule
  def esFeatures = ESFeatures.Defaults.withESVersion(ESVersion.ES2020)
```

This seems to come from https://github.com/scala-js/scala-js/blob/v1.7.1/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Linker.scala#L39-L57. While Mill worker assumes the output to be a single file: https://github.com/com-lihaoyi/mill/blob/0.10.0/scalajslib/worker/1/src/ScalaJSWorkerImpl.scala#L122-L125.

I could try to make the change that scala-js linker is proposing, but would need some guidance about the potential impact of the change. Thanks in advance for any help here.

lolgab:
Fixed in #1714.
Now `fastOpt` and `fullOpt` emit other modules as well.
Available since version `0.10.0-30-0f1a37`

shishkin:
Thanks @lolgab, this is awesome! Now compilation works. I'm still having an issue running tests when top-level export is present in the project:

```
[96/97] lambda.test.fastOptTest
Found multiple public modules but module support is disabled: [hello, out]
  exported to JavaScript with @JSExport
Linker: Compute reachability: 293622 us
Linker: 309311 us
1 targets failed
lambda.test.fastOptTest There were linking errors
```

(No additional logs about linking errors and `out/lambda/test/fastOptTest.dest` is empty)

[...]

PR #1714, opened 2022-01-28
Link Scala.js using Directory path when available

lolgab:
This doesn't change any API but makes the Scala.js emit top level
exports.

## Motivation

Thinking about APIs is hard, so I tried to make the process of support the new `faskLinkJS` a two step approach. For now I migrated only the implementation to the new api using `DirectoryOutputPath` (in Scala. js 1.3+ only) but maintaed the `out.js` output path. It doesn't fail anymore if you try to export a module from the Scala.js code. At the same time it is not optimal since the ausiliary files are not returned by the `fastOpt` task, so are not considered for the tasks cache purposes.


lefou (approved):
Looks good to me. Thank you!

lefou:
Does this fix https://github.com/com-lihaoyi/mill/issues/1684 ?

PR #1717, opened 2022-01-29
Fixed Version handling of short versions and made API public

lefou:
This is in preparation to use Version in various module to make version checks without reinventing regexes over and over.

### DIFF
--- a/scalalib/src/dependency/versions/Version.scala
+++ b/scalalib/src/dependency/versions/Version.scala
@@ -47,9 +47,9 @@ case class ValidVersion(
 ) extends Version {
   def major: Long = releasePart.headOption getOrElse 0
 
-  def minor: Long = releasePart.drop(1).headOption getOrElse 1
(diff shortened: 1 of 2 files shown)


scalalib/src/dependency/versions/Version.scala
```diff
@@ -47,9 +47,9 @@ case class ValidVersion(
 ) extends Version {
   def major: Long = releasePart.headOption getOrElse 0
 
-  def minor: Long = releasePart.drop(1).headOption getOrElse 1
+  def minor: Long = releasePart.drop(1).headOption getOrElse 0
 
-  def patch: Long = releasePart.drop(2).headOption getOrElse 1
+  def patch: Long = releasePart.drop(2).headOption getOrElse 0
```
lefou, line 52: Defaulting to `1` seems like a bug from the beginning. At least I can't come up with why this may be the right thing.

PR #1725, opened 2022-01-31
CI: Add macosx to the matrix

lefou:


lefou:
Bah, GH, you're drunk.

.github/workflows/actions.yml
```diff
@@ -10,6 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-latest,macos-11]
```
lefou, line 13:
```suggestion
        os: [ubuntu-latest,macOS-latest]
```

IdiosApps:
In a private repo I use
```
jobs:
  release:
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        include:
          - os: windows-latest
            artifactName: windows
          - os: macos-latest
            artifactName: macos
    steps:
    ...
```
`macos-latest` should work :)

IdiosApps:
The filename for this action is `actions.yml`. I'd suggest `build-and-release.yml` (as per the `name` at the top of the file), because you might add more actions later.

lefou:
> In a private repo I use
> 
> ```
> jobs:
>   release:
>     runs-on: ${{ matrix.os }}
>     strategy:
>       matrix:
>         include:
>           - os: windows-latest
>             artifactName: windows
>           - os: macos-latest
>             artifactName: macos
>     steps:
>     ...
> ```
> 
> `macos-latest` should work :)

Already tried that, didn't work.

IdiosApps:
I'

> > In a private repo I use
> > ```
> > jobs:
> >   release:
> >     runs-on: ${{ matrix.os }}
> >     strategy:
> >       matrix:
> >         include:
> >           - os: windows-latest
> >             artifactName: windows
> >           - os: macos-latest
> >             artifactName: macos
> >     steps:
> >     ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `macos-latest` should work :)
> 
> Already tried that, didn't work.

I'm trying to find my notes, but I think it's possible that even though the action has 
```
on:
  push:
  pull_request:
```
it is actually running the action.yml file that is on `main`  (which does not yet have this change, thus `macos-latest` won't run)

lefou:
I'm pretty sure it's using the changed `actions.yml`. But it fails for the complete `test` jobs. None is run, not even those with `ubuntu-latest`.

IdiosApps:
Yeah, I tried putting it to main on a fork and then doing a PR - and it isn't in the jobs list ([link](https://github.com/IdiosApps/mill/actions/runs/2474432130)).

I double checked my private repo GH Action works with macos-latest. It's OK:
image

Also trying [act](https://github.com/nektos/act) to test locally, but it doesn't properly support macos runners. I will have a play in the fork and let you know if I get an update.

lefou:
@IdiosApps I guess I found the issue. I have some explicitly included jobs. Turns out, I also need to explicitly specify a value for `matrix.os` for these jobs. Now, GHA starts all jobs.

IdiosApps:
> @IdiosApps I guess I found the issue. I have some explicitly included jobs. Turns out, I also need to explicitly specify a value for `matrix.os` for these jobs. Now, GHA starts all jobs.

Nice one! I just found out it's the `includes` section too - tore it down to barebones and added stuff back bit-by-bit.

PR #1729, opened 2022-02-03
Universal Main

lefou:
This PR is an attempt to improve the startup process of Mill. Currently, to run in different modes (client/server, standalone/no-server) we use different entry points which reside in different classes. Also, the client main is implemented in pure Java to improve the startup time of the client (by keep classloading at a minimum). To reflect the different entry points, we currently need a rather sophisticated start script. Although this PR does not simplify the process as such, it simplifies it's user side by supporting all modes through only one main class, the `mill.main.client.MillClientMain`. That way the good startup times in client mode remain.

Because we handle all modes in a single point, we can better handle server startup issues. E.g. we can now gracefully fall back to a in-process or sub-process runner when the server runner can not be properly started. This should improve the user experience on some Windows setups and newer M1 Macs a lot.

I didn't removed the complexity of the startup script for now, but simply ensured it always calls the client main.

The client main now ensures that custom JVM properties (`.mill-jvm-opts`) are properly applied in case we can't start the server and will, if necessary, start in a sub-process (with properly applied JVM options) instead of in-process.

The `.mill-version` is currently not handled (as before), but there is no reason that we could not detect a mismatch and even implement fetching and starting of another version in future. 

So, this PR should also improve the way Mill can be embedded into other tools or launcher, like coursier.

I have no idea how to test all this automatically, so I tried to test as much locally as possible, but any help in testing this, especially in edge cases like special environments, high load scenarios, many mill server instances (e.g. with different JVMs) would be really appreciated.


main/client/src/mill/main/client/IsolatedMillMainLoader.java
```diff
@@ -0,0 +1,80 @@
...
+            canLoad = Optional.of(result);
+            return result;
+        }
+    }
+
+    public static void runMain(String[] args) throws Exception {
+        LoadResult loadResult = load();
+        if (loadResult.millMainMethod.isPresent()) {
+            if (!MillEnv.millJvmOptsAlreadyApplied() && MillEnv.millJvmOptsFile().exists()) {
+//                System.err.println("Warning: Settings from file `" + propOptsFile + "` are currently ignored.");
```
lolgab, line 51: Is this commented line here on purpose?

lefou: No, it's a left over. I'll remove it.

lefou:
```suggestion
```

lolgab:
This is a great improvement for Mill. I tested on my biggest project and I didn't notice any problem. I'd say to merge this and let users test it before releasing `0.10.1`.

lolgab (approved):

justf0rfun:
The documentation still says that installing Mill via coursier is unsupported:

> Coursier (unsupported)
> 
> Installing mill via coursier or cs is currently not officially supported. There are various issues, especially with interactive mode.
https://com-lihaoyi.github.io/mill/mill/Intro_to_Mill.html#_coursier_unsupported

When this PR is merged is are the docs still right on this or should they be updated? If other blocking issues exist where are these issues filed? The only issue I found is [this](https://github.com/coursier/coursier/issues/1773) in coursier's repo which directs to this PR.

lefou:
> The documentation still says that installing Mill via coursier is unsupported:
> 
> > Coursier (unsupported)
> > Installing mill via coursier or cs is currently not officially supported. There are various issues, especially with interactive mode.
> 
> https://com-lihaoyi.github.io/mill/mill/Intro_to_Mill.html#_coursier_unsupported
> 
> When this PR is merged is are the docs still right on this or should they be updated? If other blocking issues exist where are these issues filed? The only issue I found is [this](https://github.com/coursier/coursier/issues/1773) in coursier's repo which directs to this PR.

This is a good question. I didn't follow recent changes in coursier and it's installable apps. If there is only one Mill app which is running the Mill client, than we are closer to remove that notice. It needs further review though, e.g. whether handling of `.mill-version` and `.mill-jvm-opts` works as expected. I doubt that, but can't say for sure. A minimal acceptable coursier-installed app should at least detect and warn in situations where the running version is different than requested by `.mill-version` or `$MILL_VERSION` and/or when it's not using the requested options.

lefou:
> > It needs further review though, e.g. whether handling of `.mill-version` and `.mill-jvm-opts` works as expected. I doubt that, but can't say for sure. A minimal acceptable coursier-installed app should at least detect and warn in situations where the running version is different than requested by `.mill-version` or `$MILL_VERSION` and/or when it's not using the requested options.
> 
> It doesn't for now. `cs launch mill:X.Y.Z` basically launches Mill version X.Y.Z. This doesn't read `.mill-version`. [coursier/apps#166](https://github.com/coursier/apps/pull/166) should make it read `.mill-jvm-opts`.

I think [coursier/apps#166](https://github.com/coursier/apps/pull/166) is not doing the right thing. The properties are not meant to be loaded by the Mill client JVM. 

Mill client automatically loads the properties file to start the server process. But in case we don't want a server (e.g. options `-i` or `--no-server`), the JVM needs to run with these options. After reviewing the startup process in main branch, Mill already detects cases, when the options where not applied, and just starts a new JVM. This can be slower and inefficient, but should be enough for coursier.

https://github.com/com-lihaoyi/mill/blob/bc2bcfe6bfabf78ad5d57719f79aa0858fb694d2/main/client/src/mill/main/client/MillClientMain.java#L64

https://github.com/com-lihaoyi/mill/blob/bc2bcfe6bfabf78ad5d57719f79aa0858fb694d2/main/client/src/mill/main/client/IsolatedMillMainLoader.java#L50

PR #1740, opened 2022-02-12
Make Evaluator and EvaluatorState normal classes

lolgab:
Extracted from https://github.com/com-lihaoyi/mill/pull/1663 to allow introducing new fields in the Evaluator and EvaluatorState classes without breaking the binary compatibility.

### DIFF
--- a/main/core/src/mill/eval/Evaluator.scala
+++ b/main/core/src/mill/eval/Evaluator.scala
@@ -33,31 +33,50 @@ case class Labelled[T](task: NamedTask[T], segments: Segments) {
 
 /**
  * Evaluate tasks.
- * @param home
- * @param outPath The output base path.
- * @param externalOutPath The output base path to use for external modules.
- * @param rootModule The projects root module.
- * @param baseLogger
- * @param classLoaderSig
- * @param workerCache Mutable worker cache.
- * @param env
- * @param failFast If `true` the first failing task will fail the evaluation.
- *                 If `false`, it tries to evaluate all tasks, running longer and reporting possibly more than one failure.
- * @param threadCount If a [[Some]] the explicit number of threads to use for parallel task evaluation,
- *                    or [[None]] to use n threads where n is the number of available logical processors.
  */
-case class Evaluator(
-    home: os.Path,
-    outPath: os.Path,
-    externalOutPath: os.Path,
-    rootModule: mill.define.BaseModule,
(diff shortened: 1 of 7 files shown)


main/core/src/mill/eval/Evaluator.scala
```diff
@@ -33,31 +33,50 @@ case class Labelled[T](task: NamedTask[T], segments: Segments) {
...
-    externalOutPath: os.Path,
-    rootModule: mill.define.BaseModule,
-    baseLogger: ColorLogger,
-    classLoaderSig: Seq[(Either[String, java.net.URL], Long)] = Evaluator.classLoaderSig,
-    workerCache: mutable.Map[Segments, (Int, Any)] = mutable.Map.empty,
-    env: Map[String, String] = Evaluator.defaultEnv,
-    failFast: Boolean = true,
-    threadCount: Option[Int] = Some(1)
-) {
+class Evaluator @deprecated(message = "Use apply instead", since = "0.10.1") (
```
lolgab, line 37: When a new field gets added to the class, this constructor needs to be moved to a separate `def this` and the main constructor needs to be made `private`

lefou (commented):
Looks reasonable to me. 

I'd prefer to include the word "mill" in the `@deprecated since` attribute, to improve reading of compiler deprecation messages. Note, that the compiler just spits out all deprecation warnings as they occur and as a user it can be hard to guess where these are coming from and what dependency could be the culprit. So I'd go with the convention to always include the library name. The "after x.y.z" might be a bit weird, but it's actually the most accurate message we can leave without the need for later maintenance edits.

Example:

```scala
@deprecated("...", "mill 0.10.1")
// or if the next version is still unclear
@deprecated("...", "mill after 0.10.0")

@since Mill after 0.10.0
```

lolgab:
If we merge this PR in `0.10.0` we know for sure that next version will be `0.10.1`. Isn't it correct to say deprecated since `mill 0.10.1` ?

lefou:
> If we merge this PR now (befor the release of `0.10.1`) we know for sure that next version will be `0.10.1`. Isn't it correct to say deprecated since `mill 0.10.1`?

Sure. Just add the "mill" then. I wanted to explain my whole reasoning, as I had situations before, e.g. in the late 0.9 and early 0.10 phase.

lefou (approved):
Thank you!

PR #1744, opened 2022-02-17
Add separator to AppendPattern assembly rule

yakivy:
PR adds separator to AppendPattern assembly rule.

Implementation and motivation are identical to `mill.modules.Assembly.Rule.Append#separator`

### DIFF
--- a/main/src/mill/modules/Assembly.scala
+++ b/main/src/mill/modules/Assembly.scala
@@ -26,18 +26,36 @@ object Assembly {
   object Rule {
     case class Append(path: String, separator: String = defaultSeparator) extends Rule
 
-    @deprecated("Use mill.modules.Assembly.Rule.AppendByPattern instead", since = "mill 0.10.1")
     object AppendPattern {
-      def apply(pattern: String): AppendPattern = AppendPattern(Pattern.compile(pattern))
-    }
-    @deprecated("Use mill.modules.Assembly.Rule.AppendByPattern instead", since = "mill 0.10.1")
-    case class AppendPattern(pattern: Pattern) extends Rule
+      def apply(pattern: Pattern): AppendPattern = new AppendPattern(pattern, defaultSeparator)
+      def apply(pattern: String): AppendPattern = apply(pattern, defaultSeparator)
+      def apply(pattern: String, separator: String): AppendPattern =
+        new AppendPattern(Pattern.compile(pattern), separator)
 
-    object AppendByPattern {
-      def apply(pattern: String, separator: String = defaultSeparator): AppendByPattern =
-        AppendByPattern(Pattern.compile(pattern), separator)
+      @deprecated(message = "Binary compatibility shim. Don't use it. To be removed", since = "mill 0.10.1")
+      def unapply(value: AppendPattern): Option[Pattern] = Some(value.pattern)
     }
-    case class AppendByPattern(pattern: Pattern, separator: String) extends Rule
+    class AppendPattern(val pattern: Pattern, val separator: String) extends Rule {
+      def this(pattern: Pattern) = this(pattern, defaultSeparator)
+
+      override def productPrefix: String = "AppendPattern"
+      override def productArity: Int = 2
+      override def productElement(n: Int): Any = n match {
+        case 0 => pattern
+        case 1 => separator
+        case _ => throw new IndexOutOfBoundsException(n.toString)
+      }
+      override def canEqual(that: Any): Boolean = that.isInstanceOf[AppendPattern]
+      override def hashCode(): Int = scala.runtime.ScalaRunTime._hashCode(this)
+      override def equals(obj: Any): Boolean = obj match {
+        case that: AppendPattern => this.pattern == that.pattern && this.separator == that.separator
+        case _ => false
+      }
+      override def toString: String = scala.runtime.ScalaRunTime._toString(this)
+
+      @deprecated(message = "Binary compatibility shim. Don't use it. To be removed", since = "mill 0.10.1")
+      def copy(pattern: Pattern = pattern): AppendPattern = new AppendPattern(pattern, separator)
+    }
 
     case class Exclude(path: String) extends Rule
 
@@ -59,14 +77,15 @@ object Assembly {
     }.toMap
 
     val matchPatterns = assemblyRules.collect {
-      case r @ Rule.AppendPattern(pattern) => pattern.asPredicate() -> r
-      case r @ Rule.AppendByPattern(pattern, _) => pattern.asPredicate() -> r
+      case r : Rule.AppendPattern => r.pattern.asPredicate() -> r
       case r @ Rule.ExcludePattern(pattern) => pattern.asPredicate() -> r
     }
 
     mappings.foldLeft(Map.empty[String, GroupedEntry]) {
       case (entries, (mapping, entry)) =>
-        rulesMap.get(mapping) orElse matchPatterns.find(_._1.test(mapping)).map(_._2) match {
+        def simpleRule = rulesMap.get(mapping)
+        def patternRule = matchPatterns.find(_._1.test(mapping)).map(_._2)
+        simpleRule orElse patternRule match {
           case Some(_: Rule.Exclude) =>
             entries
           case Some(a: Rule.Append) =>
@@ -74,10 +93,7 @@ object Assembly {
             entries + (mapping -> newEntry)
           case Some(_: Rule.ExcludePattern) =>
             entries
-          case Some(_: Rule.AppendPattern) =>
-            val newEntry = entries.getOrElse(mapping, AppendEntry.empty).append(entry)
-            entries + (mapping -> newEntry)
-          case Some(a: Rule.AppendByPattern) =>
+          case Some(a: Rule.AppendPattern) =>
             val newEntry = entries.getOrElse(mapping, AppendEntry(Nil, a.separator)).append(entry)
             entries + (mapping -> newEntry)
           case _ if !entries.contains(mapping) =>
--- a/main/src/mill/modules/Assembly.scala
+++ b/main/src/mill/modules/Assembly.scala
@@ -35,7 +35,8 @@ object Assembly {
       @deprecated(message = "Binary compatibility shim. Don't use it. To be removed", since = "mill 0.10.1")
       def unapply(value: AppendPattern): Option[Pattern] = Some(value.pattern)
     }
-    class AppendPattern(val pattern: Pattern, val separator: String) extends Rule {
+    class AppendPattern private(val pattern: Pattern, val separator: String) extends Rule {
+      @deprecated(message = "Binary compatibility shim. Don't use it. To be removed", since = "mill 0.10.1")
       def this(pattern: Pattern) = this(pattern, defaultSeparator)
 
       override def productPrefix: String = "AppendPattern"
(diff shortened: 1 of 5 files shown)


yakivy:
I can fix `apply` and `this` binary backward compatibility in trade of readability, but `copy` and `unapply` will still fail. Let me know if we need this.

lolgab:
Can you write some motivation for the change you're proposing?

yakivy:
> Can you write some motivation for the change you're proposing?

As mentioned in PR description, motivation is identical to `mill.modules.Assembly.Rule.Append#separator`: some files (in my case service files) doesn't have new line at the end, so they merged into single line.

main/src/mill/modules/Assembly.scala
```diff
@@ -27,9 +27,28 @@ object Assembly {
...
+      def apply(pattern: String, separator: String): AppendPattern =
+        new AppendPattern(Pattern.compile(pattern), separator)
+
+      @deprecated(message = "Binary compatibility shim. To be removed", since = "mill 0.10.1")
+      def unapply(value: AppendPattern): Option[Pattern] = Some(value.pattern)
+    }
+    class AppendPattern(val pattern: Pattern, val separator: String) extends Rule {
+      def this(pattern: Pattern) = this(pattern, defaultSeparator)
+
+      override def productArity: Int = 2
```
lolgab, line 41: This should be deprecated with the same message as below.

main/src/mill/modules/Assembly.scala
```diff
@@ -27,9 +27,28 @@ object Assembly {
...
+      override def productElement(n: Int): Any = n match {
```
lolgab, line 42: This should be deprecated with the same message as below.

main/src/mill/modules/Assembly.scala
```diff
@@ -27,9 +27,28 @@ object Assembly {
...
+        case 0 => pattern
+        case 1 => separator
+        case _ => throw new IndexOutOfBoundsException(n.toString)
+      }
+      override def canEqual(that: Any): Boolean = that.isInstanceOf[AppendPattern]
```
lolgab, line 47: This should be deprecated with the same message as below.

main/src/mill/modules/Assembly.scala
```diff
@@ -27,9 +27,28 @@ object Assembly {
...
+
```
lolgab, line 48: You need to implement also `toString`, `equals`, and `hashCode`

main/src/mill/modules/Assembly.scala
```diff
@@ -50,26 +69,23 @@ object Assembly {
...
+    val matchPatterns = assemblyRules.collect {
+      case r : Rule.AppendPattern => r.pattern.asPredicate() -> r
+      case r @ Rule.ExcludePattern(pattern) => pattern.asPredicate() -> r
     }
 
     mappings.foldLeft(Map.empty[String, GroupedEntry]) {
       case (entries, (mapping, entry)) =>
-        rulesMap.get(mapping) match {
-          case Some(_: Assembly.Rule.Exclude) =>
+        rulesMap.get(mapping) orElse matchPatterns.find(_._1.test(mapping)).map(_._2) match {
```
lolgab, line 79: You could split the two parts into two properly named variables and then do `orElse` separately.

lolgab (changes requested):
There are some missing things around binary compatibility.
I will leave the review of the assembly part to someone else (maybe @lefou ?) since I'm not familiar with that code.
Also, can you add a test that exercises the use case you have (when you need to pass a special separator)?

Thank you very much and sorry for the many requests.

main/src/mill/modules/Assembly.scala
```diff
@@ -27,9 +27,28 @@ object Assembly {
...
-      def apply(pattern: String): AppendPattern = AppendPattern(Pattern.compile(pattern))
+      def apply(pattern: Pattern): AppendPattern = new AppendPattern(pattern, defaultSeparator)
+      def apply(pattern: String): AppendPattern = apply(pattern, defaultSeparator)
+      def apply(pattern: String, separator: String): AppendPattern =
+        new AppendPattern(Pattern.compile(pattern), separator)
+
+      @deprecated(message = "Binary compatibility shim. To be removed", since = "mill 0.10.1")
+      def unapply(value: AppendPattern): Option[Pattern] = Some(value.pattern)
+    }
+    class AppendPattern(val pattern: Pattern, val separator: String) extends Rule {
```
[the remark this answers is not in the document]
yakivy: Solution with emulating case class binary compatibility with simple class become dirtier and dirtier, so I created separate `mill.modules.Assembly.Rule.AppendByPattern` class and deprecated `mill.modules.Assembly.Rule.AppendPattern`. Let me know if you don't like this idea.

main/src/mill/modules/Assembly.scala
yakivy: Want to notice that argument in `scala.Option#orElse` is "by-name", so extracting variables will degrade performance, or did you want me to extract functions?

yakivy:
I also found and fixed a bug: `mill.modules.Assembly.Rule.Append` ignores separator if merged files are in different assembly units.

main/src/mill/modules/Assembly.scala
lolgab: Then make the parameter a `def` so the result will be equivalent to a by-name parameter.

main/src/mill/modules/Assembly.scala
```diff
@@ -26,11 +26,19 @@ object Assembly {
...
     case class Append(path: String, separator: String = defaultSeparator) extends Rule
 
+    @deprecated("Use mill.modules.Assembly.Rule.AppendByPattern instead", since = "mill 0.10.1")
     object AppendPattern {
       def apply(pattern: String): AppendPattern = AppendPattern(Pattern.compile(pattern))
     }
+    @deprecated("Use mill.modules.Assembly.Rule.AppendByPattern instead", since = "mill 0.10.1")
     case class AppendPattern(pattern: Pattern) extends Rule
 
+    object AppendByPattern {
```
lolgab, line 36: By creating a new `case class` you're creating a new problem for the next person that will need another parameter.
Then we will have either 3 case classes or a deprecated `case class` and a class with all the ugly workaround you reverted.
I think Mill should not add any more `case class`es to the codebase unless there is a very very strong reason to do so.

yakivy: For me sounds like YAGNI violation, but you know best. Reverting revert...

main/src/mill/modules/Assembly.scala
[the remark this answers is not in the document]
yakivy: Why? Case classes don't have private constructors.

main/src/mill/modules/Assembly.scala
yakivy: Why? It's not a shim, valid and usable method, moreover rule extends product.

main/src/mill/modules/Assembly.scala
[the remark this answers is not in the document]
lolgab: This was a `case class` but doesn't need to be forever. The idea is to maintainSince `case class`es as bad for binary compatibility we want to get rid of them.
I'm trying to make the public API as small as possible and having two public APIs to define an object is against this principle (in this case `ApplyPattern(pattern, separator)` and `new ApplyPattern(pattern, separator)`).
We have to maintain the existing ones, but we don't want to create new ones unless it's necessary.

main/src/mill/modules/Assembly.scala
```diff
@@ -27,9 +27,35 @@ object Assembly {
...
+      def apply(pattern: Pattern): AppendPattern = new AppendPattern(pattern, defaultSeparator)
+      def apply(pattern: String): AppendPattern = apply(pattern, defaultSeparator)
+      def apply(pattern: String, separator: String): AppendPattern =
+        new AppendPattern(Pattern.compile(pattern), separator)
+
+      @deprecated(message = "Binary compatibility shim. Don't use it. To be removed", since = "mill 0.10.1")
+      def unapply(value: AppendPattern): Option[Pattern] = Some(value.pattern)
+    }
+    class AppendPattern(val pattern: Pattern, val separator: String) extends Rule {
+      def this(pattern: Pattern) = this(pattern, defaultSeparator)
```
lolgab, line 39: This constructor should be `deprecated` too.

main/src/mill/modules/Assembly.scala
lolgab: > moreover rule extends product.

Then you're right. We can deprecate the product things later for `Rule` and its subclasses.

main/src/mill/modules/Assembly.scala
```diff
@@ -50,26 +76,25 @@ object Assembly {
...
     }.toMap
 
-    val appendPatterns = assemblyRules.collect {
-      case Rule.AppendPattern(pattern) => pattern.asPredicate().test(_)
-    }
-
-    val excludePatterns = assemblyRules.collect {
-      case Rule.ExcludePattern(pattern) => pattern.asPredicate().test(_)
+    val matchPatterns = assemblyRules.collect {
+      case r : Rule.AppendPattern => r.pattern.asPredicate() -> r
```
lolgab, line 80: Can you scalafmt on the changed files?

scalalib/test/src/HelloWorldTests.scala
```diff
@@ -757,6 +765,28 @@ object HelloWorldTests extends TestSuite {
...
+            val Right((result, _)) = eval.apply(target)
+
+            Using.resource(new JarFile(result.path.toIO)) { jarFile =>
+              assert(jarEntries(jarFile).contains("without-new-line.conf"))
+
+              val fileContent = readFileFromJar(jarFile, "without-new-line.conf")
+
+              assert(
+                "without-new-line\\.first=first(?:\n|$)".r.findFirstIn(fileContent).isDefined,
+                "without-new-line\\.second=second(?:\n|$)".r.findFirstIn(fileContent).isDefined,
```
lolgab, line 785: Can you assert the exact content of the file instead?
Like:
```scala
val result  = readFileFromJar(jarFile, "without-new-line.conf")
val expected = "without-new-line.first=first\nwithout-new-line.second=second"
assert(result == expected)

```

lolgab (changes requested):
Some minor comments, otherwise it looks good to me.

scalalib/test/src/HelloWorldTests.scala
lolgab: Gotcha! Then what do you think about:
```scala
val result  = readFileFromJar(jarFile, "without-new-line.conf").split('\n').toSet
val expected = Set("without-new-line.first=first", "without-new-line.second=second")
assert(result == expected)
```
?

lolgab (approved):
LGTM! 👍
Thank you for your contribution @yakivy

lefou (approved):
I see the efforts you already put into this, so I guess it's good to merge. Thank you!

Besides, I don't like the bloat we add with this kind of patches. It's a maintainers nightmare. And will make future contributions much harder. We might evaluate some other approaches for powerful case class-like config objects which allow more binary compatibility, e.g. data-class or something like this.

lolgab:
@lefou The idea is to delete all the bloat of `case class` methods before `0.11` and just have a class with manual `apply`, `toString` and few more methods. It's the only way I know to evolve a codebase like the mill one. The Sbt team uses code-generation but it has its own tradeoffs.

PR #1748, opened 2022-02-21
Use Jib for the docker module

forty:
Hello,

I started this PoC a while ago mostly for learning reasons, and I just found it while cleaning up my computer. I'm pushing it here in case it may be useful.

This is replacing direct calls to the docker CLI by the use of  the [Jib](https://github.com/GoogleContainerTools/jib) library in the docker module. This avoid requiring the docker daemon (so avoid being root and/or DinD and/or --privileged in some CI that are using docker - for example Gitlab CI).

The build command outputs a tar file, which works nicely with mill conditional rebuild. The `push` command is still there to push to a registry, and I added a `load` command as a convenient way of docker loading the image in the local docker daemon.

I wish you a great week!


lefou:
Interesting. I think loosing the `run` target is a show stopper for some users. Also, the current version has a customizable `executable` target, so it is possible to use e.g. `podman` which also runs with user privileges and doesn't require a docker daemon.

On the other hand, it looks like your change removes the need to have some local tools installed, which is a great plus. I'd like to keep this PR a while open for comments, but won't merge it as-is. Maybe, someone makes it optional/switchable?

forty:
I love podman, but last time I checked it was not possible to run it in another container without some special privileges (see for example https://www.redhat.com/sysadmin/podman-inside-container) which is annoying if your CI runs in docker.

The traditional approach to solving that is too use things like kaniko (https://github.com/GoogleContainerTools/kaniko) which do support RUN commands but have their own gotchas, which would make it inconvenient to integrate to mill.

One possibility, if we want to keep both options (which I had initially choosen then changed mind) would to simply have a JibModule, completely unrelated to the DockerModule

forty:
(and yes, absolutely, this change removes the need to have any tool installed locally)

forty:
Ah and one last thing: Jib documentation does provide a "solution" for the lack of RUN support: they suggest to use a custom base image :D How this base image is built is "left as an exercise for the user" ;)

lefou:
I have no experience with Jib, but if gets the job of building a container done as good as docker, we could use Jib to build the image, but keep the existing `executable` and `run` tasks to run it in a platform depended way. Best of both worlds, so to say.

lefou:
Also, I'd prefer to load the Jib library/tools in a worker (in an isolated classloader) and maybe worker module, as it depends on lots of libraries and especially the google guava deps have a reputation to bring trouble into the classpath.

forty:
I can confirm, I reached out to you when I wrote this a while ago for this specific problem with guava ^^ https://github.com/com-lihaoyi/mill/pull/1378

After rebasing yesterday it looks like now the guava version is already good enough so I don't need to force it anymore, but it does look like potential for trouble indeed.

sideeffffect:
I think this is a relevant place for Mill to share the news

`jib-common` has been [factored out](https://github.com/sbt-jib/sbt-jib/pull/85) of sbt-jib and is available as an independent library -- that makes it available to be used in other build tools besides sbt, like Mill.
https://repo1.maven.org/maven2/de/gccc/sbt/jib-common_2.13/1.3.4/
https://github.com/sbt-jib/sbt-jib/releases/tag/v1.3.4

PR #1752, opened 2022-02-23
Update coursier to 2.1.0-M2

alexarchambault:
Compared to `2.0.16-200-ge888c6dea`, this version fixes issues with snapshot artifacts, and also adds new APIs (`ArchiveCache`, that can be useful from Mill tasks).

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -92,9 +92,8 @@ object Deps {
   )
   val asciidoctorj = ivy"org.asciidoctor:asciidoctorj:2.4.3"
   val bloopConfig = ivy"ch.epfl.scala::bloop-config:1.4.13"
-  val coursier = ivy"io.get-coursier::coursier:2.0.16-200-ge888c6dea"
+  val coursier = ivy"io.get-coursier::coursier:2.1.0-M2"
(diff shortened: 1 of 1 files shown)


build.sc
```diff
@@ -92,9 +92,8 @@ object Deps {
   )
   val asciidoctorj = ivy"org.asciidoctor:asciidoctorj:2.4.3"
   val bloopConfig = ivy"ch.epfl.scala::bloop-config:1.4.13"
-  val coursier = ivy"io.get-coursier::coursier:2.0.16-200-ge888c6dea"
+  val coursier = ivy"io.get-coursier::coursier:2.1.0-M2"
   val coursierReducedDeps = coursier.exclude(
-    "com.lihaoyi" -> "utest",
```
lefou, line 97: Is this no longer in the transitive dependency set, or is it mandatory now?

alexarchambault: It's not a dependency any more, it had ended up there by mistake…

lefou: Ok. Nice you fixed it.

lefou (approved):
LGTM, thanks!

PR #1753, opened 2022-02-23
Update coursier to 2.1.0-M5

alexarchambault:
Follow up of https://github.com/com-lihaoyi/mill/pull/1752. Just realized it should have been `2.1.0-M5` rather than `2.1.0-M2` 😅

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -92,10 +92,7 @@ object Deps {
   )
   val asciidoctorj = ivy"org.asciidoctor:asciidoctorj:2.4.3"
   val bloopConfig = ivy"ch.epfl.scala::bloop-config:1.4.13"
-  val coursier = ivy"io.get-coursier::coursier:2.1.0-M2"
-  val coursierReducedDeps = coursier.exclude(
-    "org.codehaus.plexus" -> "*"
-  )
+  val coursier = ivy"io.get-coursier::coursier:2.1.0-M5"
 
   val flywayCore = ivy"org.flywaydb:flyway-core:8.0.2"
   val graphvizJava = ivy"guru.nidi:graphviz-java:0.18.1"
@@ -278,7 +275,7 @@ object main extends MillModule {
     override def ivyDeps = Agg(
       Deps.ammoniteExcludingTrees,
       Deps.scalametaTrees,
-      Deps.coursierReducedDeps,
+      Deps.coursier,
       // Necessary so we can share the JNA classes throughout the build process
       Deps.jna,
       Deps.jnaPlatform,


build.sc
```diff
@@ -92,7 +92,7 @@ object Deps {
   )
   val asciidoctorj = ivy"org.asciidoctor:asciidoctorj:2.4.3"
   val bloopConfig = ivy"ch.epfl.scala::bloop-config:1.4.13"
-  val coursier = ivy"io.get-coursier::coursier:2.1.0-M2"
+  val coursier = ivy"io.get-coursier::coursier:2.1.0-M5"
   val coursierReducedDeps = coursier.exclude(
     "org.codehaus.plexus" -> "*"
```
alexarchambault, line 97: Why do you exclude this?

lefou: Good question. I guess we fixed/worked around some issues in CI? It was added the same time we also excluded the utest transitive dependency, so maybe, this can be removed too?

lefou: You may know better then me if this should be not excluded, so please change it if necessary.

alexarchambault: IIRC, one of those dependencies that was pulling… junit, but we now exclude it (from one of the coursier POMs). So it shouldn't be a problem.

Here are the dependencies that coursier `2.1.0-M5` pulls:
```text
$ cs resolve io.get-coursier::coursier:2.1.0-M5
com.chuusai:shapeless_2.13:2.3.3:default
com.github.alexarchambault:argonaut-shapeless_6.2_2.13:1.2.0:default
commons-io:commons-io:2.6:default
io.argonaut:argonaut_2.13:6.2.5:default
io.get-coursier:coursier-cache_2.13:2.1.0-M5:default
io.get-coursier:coursier-core_2.13:2.1.0-M5:default
io.get-coursier:coursier-util_2.13:2.1.0-M5:default
io.get-coursier:coursier_2.13:2.1.0-M5:default
io.get-coursier.jniutils:windows-jni-utils:0.3.2:default
io.github.alexarchambault:concurrent-reference-hash-map:1.1.0:default
io.github.alexarchambault.windows-ansi:windows-ansi:0.0.3:default
org.apache.commons:commons-compress:1.21:default
org.apache.xbean:xbean-reflect:3.7:default
org.codehaus.plexus:plexus-archiver:4.2.7:default
org.codehaus.plexus:plexus-classworlds:2.6.0:default
org.codehaus.plexus:plexus-container-default:2.1.1:default
org.codehaus.plexus:plexus-io:3.2.0:default
org.codehaus.plexus:plexus-utils:3.4.1:default
org.fusesource.jansi:jansi:1.18:default
org.iq80.snappy:snappy:0.4:default
org.scala-lang:scala-library:2.13.7:default
org.scala-lang:scala-reflect:2.13.6:default
org.scala-lang.modules:scala-collection-compat_2.13:2.6.0:default
org.scala-lang.modules:scala-xml_2.13:2.0.1:default
org.tukaani:xz:1.9:default
```

I don't think any of those is a problem.

lefou (approved):
LGTM, thanks!

PR #1755, opened 2022-02-24
feat: add framework name to the test classes BSP request

kpodsiad:
I've added `[skip ci]` because for now I only want to see what do you folks think about this PR.
This PR will stay as a draft until changes in `bsp4j` are merged, for now I've used locally published `bsp4j` version.

It is an implementation of the BSP proposal which is described [here](https://github.com/build-server-protocol/build-server-protocol/issues/296#issuecomment-1041525725). This proposal adds an optional framework field to the returned class.

There's already implementation for [bloop](https://github.com/scalacenter/bloop/pull/1695).

### DIFF
--- a/bsp/src/mill/bsp/MillBuildServer.scala
+++ b/bsp/src/mill/bsp/MillBuildServer.scala
@@ -13,6 +13,8 @@ import ch.epfl.scala.bsp4j.{
   CompileProvider,
   CompileResult,
   DebugProvider,
+  DebugSessionAddress,
+  DebugSessionParams,
   DependencyModule,
   DependencyModulesItem,
   DependencyModulesParams,
@@ -623,6 +625,12 @@ class MillBuildServer(
         }
 
       new CleanCacheResult(msg, cleaned)
+    }
+
+  override def debugSessionStart(debugParams: DebugSessionParams)
+      : CompletableFuture[DebugSessionAddress] =
+    completable(s"debugSessionStart ${debugParams}") { state =>
+      throw new NotImplementedError("debugSessionStart endpoint is not implemented")
     }
 
   def completableTasks[T: ClassTag, V](
--- a/build.sc
+++ b/build.sc
@@ -133,7 +133,7 @@ object Deps {
   val utest = ivy"com.lihaoyi::utest:0.7.11"
   val windowsAnsi = ivy"io.github.alexarchambault.windows-ansi:windows-ansi:0.0.3"
   val zinc = ivy"org.scala-sbt::zinc:1.6.1"
-  val bsp = ivy"ch.epfl.scala:bsp4j:2.0.0+60-b6a5df8b-SNAPSHOT"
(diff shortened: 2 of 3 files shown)


lefou (commented):
Looks reasonable to me. Please update as appropriate and remove the draft status when you think it is mergeable. Thank you!

build.sc
```diff
@@ -133,7 +133,7 @@ object Deps {
   val utest = ivy"com.lihaoyi::utest:0.7.11"
   val windowsAnsi = ivy"io.github.alexarchambault.windows-ansi:windows-ansi:0.0.3"
   val zinc = ivy"org.scala-sbt::zinc:1.6.1"
-  val bsp = ivy"ch.epfl.scala:bsp4j:2.0.0"
+  val bsp = ivy"ch.epfl.scala:bsp4j:2.0.0:+70-f6e47d42-SNAPSHOT"
```
lefou, line 136: Version doesn't look right. 

```suggestion
  val bsp = ivy"ch.epfl.scala:bsp4j:2.0.0+70-f6e47d42-SNAPSHOT"
```

lefou (commented):
Surprisingly, this is only for Scala projects and not for e.g. Java projects. Anyhow, I'm happy to merge, although this feature is probably not very relevant for Mill users, as we only support one test framework per module.

kpodsiad:
Resolution is failing because `ch.epfl.scala:bsp4j:2.0.0+70-f6e47d42-SNAPSHOT` is being resolved on the sonatype releases, but it's a snapshot. I'll wait until the official release is published.

bsp/src/mill/bsp/MillBuildServer.scala
```diff
@@ -625,6 +627,12 @@ class MillBuildServer(
       new CleanCacheResult(msg, cleaned)
     }
 
+  override def debugSessionStart(debugParams: DebugSessionParams)
+      : CompletableFuture[DebugSessionAddress] =
+    completable(s"debugSessionStart ${debugParams}") { state =>
+      throw new NotImplementedError("debugSessionStart endpoint is not implemented")
+    }
```
kpodsiad, line 634: There's a `DebugProvider` capability that is not set by the user so it's safe to throw an error here from the BSP client perspective.

lefou:
> @lefou I updated PR with the newly released bsp4j version.
> 
> > Surprisingly, this is only for Scala projects and not for e.g. Java projects.
> 
> Do you mean that method names of method and classes contain `ScalaTestClasses`?

I mean, the framework name is only relevant for Scala projects, as it is implemented in `ScalaBuildServer`. So pure Java projects, for which we also `sbt.testing`, don't profit from this test framework selection.

kpodsiad:
@lefou It's true that it's more general than Scala and there was a small discussion about it [here](https://github.com/build-server-protocol/build-server-protocol/pull/304#pullrequestreview-941147293).

lefou:
Thanks for the pointer. I think, this would be the right move, too.

PR #1759, opened 2022-03-02
Do not clutter user home directory with `mill-release` binary

lefou:
Instead, we now create that file locally under `target/mill-release`.
Also, it is no longer hard coded into the (forked) test suite, but instead
read from the `MILL_TEST_RELEASE` system environment variable.

### DIFF
--- a/ci/test-mill-bootstrap-0.sh
+++ b/ci/test-mill-bootstrap-0.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 
 set -eux
 
@@ -7,8 +7,8 @@ git stash -u
 git stash -a
 
 # First build
-./mill -i "{__.publishLocal,assembly}"
-cp out/assembly.dest/mill ~/mill-1
+./mill -i "__.publishLocal" + assembly
+cp out/assembly.dest/mill target/mill-1
 
 # Clean up
 git stash -u
@@ -24,8 +24,8 @@ echo "Build 2" > info.txt && git add info.txt && git commit -m "Add info.txt"
 ci/patch-mill-bootstrap.sh
 
 # Second build
-~/mill-1 -i "{__.publishLocal,assembly}"
-cp out/assembly.dest/mill ~/mill-2
+target/mill-1 -i "__.publishLocal" + assembly
+cp out/assembly.dest/mill target/mill-2
 
 # Clean up
 git stash -u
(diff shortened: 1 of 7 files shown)


ci/test-mill-bootstrap-0.sh
```diff
@@ -7,12 +7,15 @@ git stash -u
 git stash -a
 
 # First build
-./mill -i "{__.publishLocal,assembly}"
-cp out/assembly.dest/mill ~/mill-1
+./mill -i "__.publishLocal" + assembly
```
lolgab, line 10: Is this Mill syntax?

main/test/src/util/ScriptTestSuite.scala
```diff
@@ -79,8 +80,17 @@ abstract class ScriptTestSuite(fork: Boolean) extends TestSuite {
     }
   }
   private def evalFork(stdout: os.ProcessOutput, s: Seq[String]): Boolean = {
+    val millRelease = Option(System.getenv("MILL_TEST_RELEASE"))
+      .getOrElse(throw new NoSuchElementException(
+        s"System environment variable `MILL_TEST_RELEASE` not defined. It needs to point to the Mill binary to use for the test."
+      ))
```
lolgab, line 86:
```suggestion
    val millRelease = sys.env.getOrElse("MILL_TEST_RELEASE", throw new NoSuchElementException(
      "System environment variable `MILL_TEST_RELEASE` not defined. It needs to point to the Mill binary to use for the test."
    ))
```

ci/test-mill-bootstrap-0.sh
lefou: Yes

main/test/src/util/ScriptTestSuite.scala
lefou: `sys.env` retrieves the whole env into a Map and handles missing elements by a second case-insensitive lookup, etc. My motivation was to avoid doing unnecessary things here.

ci/test-mill-bootstrap-0.sh
lefou: * https://github.com/com-lihaoyi/mill/pull/1521

main/test/src/util/ScriptTestSuite.scala
lolgab: Makes sense. 👍 Thank you for writing down the motivation.

lolgab (approved):
LGTM! 👍

PR #1760, opened 2022-03-03
Support Coursier plugins via customizing FileCache used (with classloaders)

nitay:
Currently loading a Coursier plugin is not possible, because the `ClassLoaders` that are used by Mill's Coursier cannot be changed. I ran into this when trying to use [my plugin](https://github.com/ActionIQ/coursier-s3) which adds support for a [new protocol](https://get-coursier.io/docs/extra#extra-protocols). Loading via `$ivy.` or anything else doesn't work because Mill uses the `default` Coursier [cache](https://github.com/com-lihaoyi/mill/blob/246e92411c874ae8a33655dec7937af8c4041509/main/src/mill/modules/Jvm.scala#L571-L570)

This PR adds a hook to allow altering the `FileCache` used. With this, I can put the following in my build module:

```
  override def cacheCustomizer: Task[Option[FileCache[coursier.util.Task] => FileCache[coursier.util.Task]]] = T.task {
    Some( (fc: FileCache[coursier.util.Task]) => {
      fc.withClassLoaders(Seq(classOf[coursier.cache.protocol.S3Handler].getClassLoader)) }
    )
  }
```
Which then fixes the resolution so that when it looks for plugins it can find my class.

This PR borrows ideas from #775

If there's better way to achieve this lmk

### DIFF
--- a/main/src/mill/modules/Jvm.scala
+++ b/main/src/mill/modules/Jvm.scala
@@ -534,7 +534,7 @@ object Jvm {
       sources: Boolean = false,
       mapDependencies: Option[Dependency => Dependency] = None,
       customizer: Option[coursier.core.Resolution => coursier.core.Resolution] = None,
-      cacheCustomizer: Option[coursier.cache.FileCache[Task] => coursier.cache.FileCache[Task]] = None,
+      coursierCacheCustomizer: Option[coursier.cache.FileCache[Task] => coursier.cache.FileCache[Task]] = None,
       ctx: Option[mill.api.Ctx.Log] = None
   ): Result[Agg[PathRef]] = {
 
@@ -544,7 +544,7 @@ object Jvm {
       force,
       mapDependencies,
       customizer,
-      cacheCustomizer,
+      coursierCacheCustomizer,
       ctx
     )
     val errs = resolution.errors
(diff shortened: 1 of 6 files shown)


lefou (commented):
Thanks @nitay for this contribition! I think, this is another motivation for 

* https://github.com/com-lihaoyi/mill/discussions/1732

As this fixes a real world scenario, I think it is worth to have this feature in Mill now. Unfortunately, your change results in binary incompatible API changes. Before we merge it, a bit more work is needed. We need to provide the previous API, mark it `@deprecated` and delegate to the new one. Good `@deprecated` messages are key.

The term "cache" is pretty general and there is a chance a user, who sees the `cacheCustomizer` task, associates it with "Mill cache" or "target cache" or "download cache". I propose to rename it to `coursierCacheCustomizer`.

main/src/mill/modules/Jvm.scala
```diff
@@ -762,6 +769,52 @@ object Jvm {
...
+   ): Result[Agg[PathRef]] =
+    resolveDependencies(
+      repositories = repositories,
+      deps = deps,
+      force = force,
+      sources = sources,
+      mapDependencies = mapDependencies,
+      customizer = customizer,
+      coursierCacheCustomizer = None,
+      ctx = ctx
```
lefou, line 793: `coursierCacheCustomizer` needs to come after `ctx`, to keep the order of other overrides, which previously had default params.

lefou (commented):
Almost there. We also need to keep the order of the parameters of newly deprecated methods, which had default parameters, to keep to encoding of default params in scala working. Meaning: new params need to come last. See below.

scalalib/src/Lib.scala
```diff
@@ -29,6 +30,7 @@ object Lib {
       deps: IterableOnce[Dep],
       mapDependencies: Option[Dependency => Dependency] = None,
       customizer: Option[coursier.core.Resolution => coursier.core.Resolution] = None,
+      coursierCacheCustomizer: Option[coursier.cache.FileCache[Task] => coursier.cache.FileCache[Task]] = None,
       ctx: Option[Ctx.Log] = None
```
lefou, line 34: Same here, swap these two params to keep binary compatibility (for default params).

lefou (approved):
Looks good to me. Thank you!

nitay:
thx @lefou - how do I find a build with this that I can try using with `.mill-version`?

lefou:
@nitay Our release pipeline is currently broken. Once, we fixed it, I'll release a 0.10.1.

alexarchambault:
Coming back at this PR 2+ years later. I'm seeing that it's the one that added a call to `noCredentials` on `FileCache`. Was there a specific motivation for that? Or was it… something that you tried while debugging, and that made it anyway, say?

alexarchambault:
I would blame that for https://github.com/com-lihaoyi/mill/issues/3408

lefou:
@alexarchambault Can't tell. It probably slipped through my radar, since the term credentials isn't anywhere in this changeset.

PR #1764, opened 2022-03-04
Add support for -scala-output-version flag

prolativ:
WIP

### DIFF
--- a/scalalib/src/JavaModule.scala
+++ b/scalalib/src/JavaModule.scala
@@ -307,7 +307,7 @@ trait JavaModule
     transitiveLocalClasspath() ++
       resources() ++
       unmanagedClasspath() ++
-      resolvedIvyDeps()
+      resolvedCompilationIvyDeps()
   }
 
   /** Same as [[compileClasspath]], but does not trigger compilation targets, if possible. */
@@ -344,7 +344,7 @@ trait JavaModule
       lcp() ++
         resources() ++
         unmanagedClasspath() ++
-        resolvedIvyDeps()
+        resolvedCompilationIvyDeps()
     }
   }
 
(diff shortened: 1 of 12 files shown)


lefou:
It would be nice, if you could update the PR description with a short summary of what this PR does and probably a link to the respective feature in the Scala compiler. Can you give a rough estimate, when this feature is going to be consumable in scalac and hence this PR should get it? Maybe some ticket we can link to?

lefou:
* https://github.com/lampepfl/dotty/pull/14606

scalalib/src/JavaModule.scala
```diff
@@ -357,6 +357,11 @@ trait JavaModule
     })()
   }
 
+  /**
+   * Like [[resolvedIvyDeps]] but used only for compiling this module .
+   */
+  def resolvedCompilationIvyDeps: T[Agg[PathRef]] = resolvedIvyDeps
```
lefou, line 363: I'm not that happy with this name. First, there is a high change of reading it as `resolveCompileIvyDeps` which would be the dual of `compileIvyDeps` (which currently does not exists). Also, the scaladoc gives no clear idea, why this is needed exactly and what's the difference to `resolvedIvyDeps`.

scalalib/src/ScalaModule.scala
```diff
@@ -57,6 +58,8 @@ trait ScalaModule extends JavaModule { outer =>
    */
   def scalaVersion: T[String]
 
+  def scalaOutputVersion: T[String] = scalaVersion
```
lefou, line 61: Here we need some scaladoc explaining its purpose and also the distinction to `scalaVersion`.

lolgab:
Scala 3.1.2 was released and the companion Sbt PR was merged.
https://github.com/sbt/sbt/pull/6814

lolgab:
Since the compiler team decided to drop the `-scala-output-version` flag we don't need to support it anymore.

Issue #1763, opened 2022-03-04
mill show __.task produces invalid json


daddykotex:
Hey! Thanks for mill, its great.

I found some information on gitter but not exactly what I'm looking for so I'm opening this so it can serve as a documentation reference if I'm not the only one that tries to do this.

I have a build with multiple modules being published to a repository. I want to produce a list of all the artifact coordinates that are generated. So, naturally, I went with the following approach:

`mill show __.artifactMetdata`

`artifactMetadata` definition is : `def artifactMetadata: Target[Artifact]`.
`Artifact` looks like: `case class Artifact(group: String, id: String, version: String)`

And so the result is: 
```json
{
    "group": "com.acme",
    "id": "pkg1",
    "version": "1.0.0"
}
{
    "group": "com.acme",
    "id": "pkg2",
    "version": "1.0.0"
}
```

It's fine, that's exactly what I need, but you can't pipe that into another tool. IMO, there are a few solutions:

- use Newline Delimited JSON (that's what we have here), but do not prettify the JSON, so each object is one line
- wrap the whole thing in an Array it's a valid JSON
- implement a command to do that (suggested in gitter)

I have written this implementation, and it works, but I find it kinda ugly and I don't like that it's relying on on an [experimental feature](https://com-lihaoyi.github.io/mill/mill/Extending_Mill.html#_evaluator_commands_experimental):

```scala
import mill.scalalib._
import mill.eval.Evaluator
import mill.api.Result

import scala.util.control.NonFatal

def allPublishedArtifacts(ev: Evaluator): mill.define.Command[Seq[Artifact]] =
  T.command {
    try {
      val artMeta = ev.rootModule.millInternal.segmentsToModules.values
        .collect { case x: scalalib.PublishModule => x }
        .map { _.artifactMetadata }
        .toSeq

      Result.Success(Evaluator.evalOrThrow(ev)(artMeta))
    } catch {
      case NonFatal(e) =>
        Result.Exception(
          e,
          new Result.OuterStack(new java.lang.Exception().getStackTrace)
        )
    }
  }
```

Is there an alternative to `mill show __.artifactMetadata` that would produce an output directly usable via another tool?

Is there a better way to implement a command where I need to find all modules of a given type (`PublishModule`) and run a target (`artifactMetadata`) on that type?

Thanks

lefou:
I read your issue as two (unrelated) improvements requests:

1. make `mill show` always return valid output
2. Add support to select targets based on their (return) type

[...]

PR #1765, opened 2022-03-05
Improved `show` output to be valid JSON and added `showNamed`

lefou:
Currently, show invoked with more than one target returns invalid JSON, as it is only concatenating multiple JSON values.

Example:

```
$ mill show __.artifactMetdata
[1/1] show
{
    "group": "com.acme",
    "id": "pkg1",
    "version": "1.0.0"
}
{
    "group": "com.acme",
    "id": "pkg2",
    "version": "1.0.0"
}
```

This PR makes its output valid JSON by outputting a proper JSON array when there is more than one task result. 

Example: 

```
$ mill show __.artifactMetdata
[1/1] show
[
  {
    "group": "com.acme",
    "id": "pkg1",
    "version": "1.0.0"
  },
  {
    "group": "com.acme",
    "id": "pkg2",
    "version": "1.0.0"
  }
]
```

Also it introduces a new `showNamed` command, which puts each task result in a JSON dictionary. The keys are reflecting the task names. 

Example:

```
 mill showNamed __.artifactMetadata
[1/1] show
{
  "pkg1.artifactMetadata":
  {
      "group": "com.acme",
      "id": "pkg1",
      "version": "1.0.0"
  },
  "pkg2.artifctMetadata":
  {
      "group": "com.acme",
      "id": "pkg2",
      "version": "1.0.0"
  }
}
```

This makes the output of `show` and `showNamed` much more usable. For humans it's easier to recognize, which task outputted what, and for external tools, as they now can process the `mill show` output as-is.

Fix https://github.com/com-lihaoyi/mill/issues/1763

To implement this, I needed to enlarge the result values of some
evaluate methods to also contain the task name, so I added new methods.
These now also transport the actual result value in addition to the watched paths.
I was uncreative and just added a "1" to the existing methods names.
Any suggestions for better names are much appreciated.

I also took the opportunity to let `show` and `showNamed` return their result.

### DIFF
--- a/docs/antora/modules/ROOT/pages/Intro_to_Mill.adoc
+++ b/docs/antora/modules/ROOT/pages/Intro_to_Mill.adoc
@@ -660,6 +660,46 @@ $ mill show foo.compileClasspath
 `show` is also useful for interacting with Mill from external tools, since the JSON it outputs is structured and easily
 parsed and manipulated.
 
+When `show` is used with multiple targets, it's output will slightly change to a JSON array, containing all the results of the given targets.
+
+[source,bash]
+----
+$ mill show "foo.{sources,compileClasspath}"
+[1/1] show
+[2/11] foo.resources
+[
+  [
+    "ref:8befb7a8:/Users/lihaoyi/Dropbox/Github/test/foo/src"
+  ],
+  [
+    "ref:c984eca8:/Users/lihaoyi/Dropbox/Github/test/foo/resources",
+    ".../org/scala-lang/scala-library/2.13.1/scala-library-2.13.1.jar"
+  ]
+]
+----
+
+=== showNamed
+
+Same as `show`, but the output will always be structured in a JSON dictionary, with the task names as key and the task results as JSON values.
+
+[source,bash]
+----
+$ mill showNamed "foo.{sources,compileClasspath}"
+[1/1] show
+[2/11] foo.resources
+{
+  "foo.sources":
+  [
+    "ref:8befb7a8:/Users/lihaoyi/Dropbox/Github/test/foo/src"
+  ],
+  "foo.compileClasspath":
+  [
+    "ref:c984eca8:/Users/lihaoyi/Dropbox/Github/test/foo/resources",
+    ".../org/scala-lang/scala-library/2.13.1/scala-library-2.13.1.jar"
+  ]
+}
+----
+
 === path
 
 [source,bash]
--- a/docs/antora/modules/ROOT/pages/Intro_to_Mill.adoc
+++ b/docs/antora/modules/ROOT/pages/Intro_to_Mill.adoc
@@ -660,7 +660,7 @@ $ mill show foo.compileClasspath
 `show` is also useful for interacting with Mill from external tools, since the JSON it outputs is structured and easily
 parsed and manipulated.
 
-When `show` is used with multiple targets, it's output will slightly change to a JSON array, containing all the results of the given targets.
+When `show` is used with multiple targets, its output will slightly change to a JSON array, containing all the results of the given targets.
 
 [source,bash]
 ----
--- a/main/src/mill/main/MainModule.scala
+++ b/main/src/mill/main/MainModule.scala
@@ -267,8 +267,6 @@ trait MainModule extends mill.Module {
     ) { res: Seq[(Any, Option[(String, Value)])] =>
       val nameAndJson = res.flatMap(_._2)
       val output = ujson.Obj.from(nameAndJson)
-      // it would be nice to also have the task name in the result
-      // as described in https://github.com/com-lihaoyi/mill/issues/1763#issuecomment-1059329035
       T.log.outputStream.println(output.render(indent = 2))
     }
   }
--- a/main/src/mill/main/MainModule.scala
+++ b/main/src/mill/main/MainModule.scala
@@ -251,7 +251,37 @@ trait MainModule extends mill.Module {
    * Runs a given task and prints the JSON result to stdout. This is useful
    * to integrate Mill into external scripts and tooling.
    */
-  def show(evaluator: Evaluator, targets: String*) = T.command {
+  def show(evaluator: Evaluator, targets: String*): Command[Value] = T.command {
+    MainModule.evaluateTasks1(
+      evaluator.withBaseLogger(
+        // When using `show`, redirect all stdout of the evaluated tasks so the
+        // printed JSON is the only thing printed to stdout.
+        evaluator.baseLogger match {
+          case PrintLogger(c1, d, c2, c3, _, i, e, in, de, uc) =>
+            PrintLogger(c1, d, c2, c3, e, i, e, in, de, uc)
+          case l => l
+        }
+      ),
+      targets,
+      SelectMode.Separated
+    ) { res: Seq[(Any, Option[(String, ujson.Value)])] =>
+      val jsons = res.flatMap(_._2).map(_._2)
+      val output: ujson.Value =
+        if (jsons.size == 1) jsons.head
+        else { ujson.Arr.from(jsons) }
+      T.log.outputStream.println(output.render(indent = 2))
+      output
+    }.map { res =>
+      val Watched(Some(json), _) = res
+      json
+    }
+  }
+
+  /**
+   * Runs a given task and prints the results as JSON dictionary to stdout. This is useful
+   * to integrate Mill into external scripts and tooling.
+   */
+  def showNamed(evaluator: Evaluator, targets: String*): Command[Value] = T.command {
     MainModule.evaluateTasks1(
       evaluator.withBaseLogger(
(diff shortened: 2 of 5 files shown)


main/src/mill/main/MainModule.scala
```diff
@@ -248,10 +264,12 @@ trait MainModule extends mill.Module {
...
       SelectMode.Separated
-    ) { res =>
-      for (json <- res.flatMap(_._2)) {
-        T.log.outputStream.println(json.render(indent = 4))
-      }
+    ) { res: Seq[(Any, Option[(String, Value)])] =>
+      val nameAndJson = res.flatMap(_._2)
+      val output = ujson.Obj.from(nameAndJson)
+      // it would be nice to also have the task name in the result
+      // as described in https://github.com/com-lihaoyi/mill/issues/1763#issuecomment-1059329035
```
lefou, line 271:
```suggestion
```

main/src/mill/main/RunScript.scala
```diff
@@ -319,11 +320,38 @@ object RunScript {
...
+      }
+  }
+
   def evaluate(
       evaluator: Evaluator,
       targets: Agg[Task[Any]]
   ): (Seq[PathRef], Either[String, Seq[(Any, Option[ujson.Value])]]) = {
-    val evaluated = evaluator.evaluate(targets)
+    val res = evaluate1(evaluator, targets)
+    (res._1, res._2.map(_.map(p => (p._1, p._2.map(_._2)))))
```
lolgab, line 347: You can use pattern matching here:
```suggestion
    val (name, json) = evaluate1(evaluator, targets)
     (name, json.map(_.map(p => (name, json.map(_._2)))))
```

lolgab (dismissed):
Can you add a test that checks the newly generated JSON structure?

main/src/mill/main/RunScript.scala
lefou: The idea is nice, but your snippet don't look correct. The Left of `evaluate1` isn't the name.

lolgab: Yeah, the `suggestion` was just a way to show you what I meant :)

main/src/mill/main/RunScript.scala
```diff
@@ -343,10 +343,17 @@ object RunScript {
       evaluator: Evaluator,
       targets: Agg[Task[Any]]
   ): (Seq[PathRef], Either[String, Seq[(Any, Option[ujson.Value])]]) = {
-    val res = evaluate1(evaluator, targets)
-    (res._1, res._2.map(_.map(p => (p._1, p._2.map(_._2)))))
+    val (watched, results) = evaluate1(evaluator, targets)
+    // we drop the task name in the inner tuple
+    (watched, results.map(_.map(p => (p._1, p._2.map(_._2)))))
```
lolgab, line 348: You could also assign names in the `map`and avoid `p._1` and `p._2`

lihaoyi:
@lefou could you update the PR description to give an example before and after of what the `show` output looks like?

lefou:
> @lefou could you update the PR description to give an example before and after of what the `show` output looks like?

Done

main/src/mill/main/RunScript.scala
```diff
@@ -319,11 +320,45 @@ object RunScript {
       }
   }
 
+  def evaluateTasks1[T](
```
daddykotex, line 323: I don't like the name, but I dont have enough context with regards to `evalutateTasks` to propose a better one

main/test/src/main/MainModuleTests.scala
```diff
@@ -0,0 +1,50 @@
...
+import mill.api.Result
+import mill.{Agg, T}
+import mill.util.{TestEvaluator, TestUtil}
+import utest.{TestSuite, Tests, test}
+
+object MainModuleTests extends TestSuite {
+
+  object mainModule extends TestUtil.BaseModule with MainModule {
+    def hello = T { Seq("hello") }
+    def hello2 = T { Seq("hello2") }
```
daddykotex, line 12: minor: would make the returned value different from the target name so that we can differentiate them in the test

daddykotex (approved):
Great work, thanks for resolving this so quickly

lihaoyi:
I think this looks reasonable to me for the multi-target use case. One thing that worries me is that for running `show` on a single target, which is probably the majority of use cases, this adds some extra indirection for anyone trying to programmatically fish data out of the result JSON. e.g. if I have a Python script parsing the `show` output, now I need to remember the target name and plug it back into the JSON dictionary to get the actual data I want, or I need to use `json.values()[0]` to get the first result. Both are kind of clunky.

If we assume that `show` is most often called on a single target, with the multi-target use case being less common, could it be worth special-casing the single-target use case to skip the wrapping dictionary? Alternatively, we could leave `show` as is and introduce a new `showAll`/`showMulti`/`showNamed` command to satisfy the multi-target use case. Either alternative would preserve the conciseness of the common single-target use case while still providing a solution for those calling `show` on multiple targets at once

daddykotex:
in my use case, I have no need for the task(s) name in the json, as I said I can on target on multiple module, so I know what's coming out of the json

but if you can send multiple targets like `mill show t1 t2` then it might be interesting to disambiguate

personally, I would just run `mill show t1 > t1.json` and `mill show t2.json` if i wanted that behaviour

lefou:
> I think this looks reasonable to me for the multi-target use case. One thing that worries me is that for running `show` on a single target, which is probably the majority of use cases, this adds some extra indirection for anyone trying to programmatically fish data out of the result JSON. e.g. if I have a Python script parsing the `show` output, now I need to remember the target name and plug it back into the JSON dictionary to get the actual data I want, or I need to use `json.values()[0]` to get the first result. Both are kind of clunky.
> 
> If we assume that `show` is most often called on a single target, with the multi-target use case being less common, could it be worth special-casing the single-target use case to skip the wrapping dictionary? Alternatively, we could leave `show` as is and introduce a new `showAll`/`showMulti`/`showNamed` command to satisfy the multi-target use case. Either alternative would preserve the conciseness of the common single-target use case while still providing a solution for those calling `show` on multiple targets at once

Ok, yeah, looks like I hit a nerve here. (The thing with assumptions is, well, ... Here is some funny pic about someone using your product: https://twitter.com/WdeB/status/1482798979511373828, haha) 

I can imagine lots of different ways to address. Yet, I'd like to have a rather small solution, which is also backwards compatible, so here is a compromise:

1. `show` task behaves as before for single-entry results
2. `show` task will put the results in a JSON array when there is more than one result
3. a new `showNamed` task will always print (and return) a JSON dictionary. Task names are the keys, and results as JSON are the values

WDYT?

lefou:
If none of us already has "some python script" that does parse the output of `show`, I'd love to have some more stable `show` output though. Meaning: Always return an array. If you're interested in the first task result, just use the first element.

lihaoyi:
@lefou what you propose sounds good to me. I'd prefer single results appear alone, but I can see the arguments either way.

lefou:
In now implemented it and update the PR description. I will add it to the documentation later, too. Anyone who has a better suggestion to the new `def` names?

main/src/mill/main/MainModule.scala
```diff
@@ -56,6 +57,21 @@ object MainModule {
         Result.Success(Watched((), watched))
     }
   }
+
+  @internal
+  def evaluateTasks1[T](
```
lefou, line 62: What about `evaluateTasksNamed`?

main/src/mill/main/RunScript.scala
lefou: What about `evaluateTasksNamed`?

main/src/mill/main/RunScript.scala
```diff
@@ -319,11 +320,45 @@ object RunScript {
...
+    (watched, results.map(_.map(p => (p._1, p._2.map(_._2)))))
+  }
+
+  /**
+   *
+   * @param evaluator
+   * @param targets
+   * @return (watched-paths, Either[err-msg, Seq[(task-result, Option[(task-name, task-return-as-json)])]])
+   */
+  def evaluate1(
```
lefou, line 357: What about `evaluateNamed`?

lefou:
I'm done with editing, unless you have another objection.

daddykotex (approved):
Perfect for me

docs/antora/modules/ROOT/pages/Intro_to_Mill.adoc
```diff
@@ -660,6 +660,46 @@ $ mill show foo.compileClasspath
 `show` is also useful for interacting with Mill from external tools, since the JSON it outputs is structured and easily
 parsed and manipulated.
 
+When `show` is used with multiple targets, it's output will slightly change to a JSON array, containing all the results of the given targets.
```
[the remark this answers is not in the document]
lefou:
```suggestion
When `show` is used with multiple targets, its output will slightly change to a JSON array, containing all the results of the given targets.
```

PR #1785, opened 2022-03-16
Fixed deprecations, warnings and missing return types

lefou:
- Enabled -deprecation compiler option
- Fixed deprecations, warnings and missing return types


lefou:
@lolgab A review would be nice, thanks.

build.sc
```diff
@@ -188,6 +188,13 @@ trait MillMimaConfig extends mima.Mima {
     issueFilterByModule.getOrElse(this, Seq())
   }
   lazy val issueFilterByModule: Map[MillMimaConfig, Seq[ProblemFilter]] = Map(
+    main.core -> Seq(
+      // fixed from Some[T] to Option[T]
+      ProblemFilter.exclude[DirectMissingMethodProblem]("mill.define.Target.asTarget"),
+      ProblemFilter.exclude[DirectMissingMethodProblem]("mill.define.TargetImpl.asTarget"),
```
lolgab, line 194: This will make all plugins compiled with older Mill versions that use these methods crash at runtime.
I think we will need to delay this change for `0.11`

main/core/src/mill/define/Task.scala
```diff
@@ -38,16 +38,16 @@ abstract class Task[+T] extends Task.Ops[T] with Applyable[Task, T] with GraphNo
...
 
 trait NamedTask[+T] extends Task[T] {
   def ctx: mill.define.Ctx
-  def label = ctx.segment match { case Segment.Label(v) => v }
+  def label: String = ctx.segment match { case Segment.Label(v) => v }
   override def toString = ctx.segments.render
 }
 trait Target[+T] extends NamedTask[T] {
-  override def asTarget = Some(this)
+  override def asTarget: Option[Target[T]] = Some(this)
```
lolgab, line 50: Here you can change to `Some[Target[T]]` and put a comment to update once we are in binary breaking window

lolgab (changes requested):
I would avoid the binary incompatible changes and put a comment for future change

lolgab (approved):

lefou:
Thanks Lorenzo!

PR #1786, opened 2022-03-16
Let some workers implement AutoCloseable

lefou:
- Let some workers implement AutoCloseable
- Let VizWorker extends AutoCloseable

### DIFF
--- a/contrib/playlib/src/mill/playlib/RouteCompilerWorker.scala
+++ b/contrib/playlib/src/mill/playlib/RouteCompilerWorker.scala
@@ -5,7 +5,7 @@ import mill.playlib.api.{RouteCompilerType, RouteCompilerWorkerApi}
 import mill.scalalib.api.CompilationResult
 import mill.{Agg, T}
 
-private[playlib] class RouteCompilerWorker {
+private[playlib] class RouteCompilerWorker extends AutoCloseable {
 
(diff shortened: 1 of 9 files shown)


lolgab:
Very nice! Do you expect to have visible performance improvements from this?

lefou:
> Very nice! Do you expect to have visible performance improvements from this?

Not at all. But I care about resources (which tend to blow up in completely unrelated situations, and is hard to debug) and mill server processes can last very long, so it may make some difference. It's more the fact that we now support resource cleanup, and Module (worker) authors will look into existing code as a guide how to implement their stuff. If they see sme cleanup code, they hopefully learn that it is something worth thinking about.

lefou:
@lolgab Do you have an idea how we can implement the `VizWorker` cleanup in a binary compatible way? It's probably the only worker really worth to cleanup, as it starts (and never stops) a background thread. Or do you think we can make an exception and just ignore the binary breakage here? See my second commit as a first attempt.

lolgab:
The only idea I have is to define a new `vizWorker` and deprecate `worker`. In 0.11 we can delete worker and in 0.12 we can rename `vizWorker` back to `worker`.

lefou:
Ok, I will address the `VisualizeModule.worker` in a different PR.

PR #1801, opened 2022-03-28
Add moduleDeps to BSP evaluated modules

lolgab:
Not sure if this solves all my problems with foreign modules (I'm having troubles testing this since the only way I have at the moment is using Metals) but I think this is the right thing to do, since the bloop module does the exact same thing [here](https://github.com/com-lihaoyi/mill/blob/bb754974c8f118cb0425c46c48e65f050a1dc1fd/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala#L115) and my project works with Bloop


lefou:
Superseded by 

* #1802

Issue #1751, opened 2022-02-23
Support a way to run Scalafmt on the build files


lolgab:
Is there a way to lint the build files?
In my projects I usually define a target like:

```scala
def buildSources = T.sources(os.pwd / "build.sc")
```

And then I run Scalafmt with:
```bash
mill mill.scalalib.scalafmt.ScalafmtModule/reformatAll buildSources
```

It would be nice if there was a way to access the build sources without having to define them manually.

lefou:
Currently there is no such target, but I think it is a good idea to have it. It should probably also contain all included files, which we need also in other places like e.g. IDE/BSP support.

lefou:
I opened a RFC which might affect this issue.

* https://github.com/com-lihaoyi/mill/discussions/1857

lefou:
I opened two PR to support running targets on meta-builds (#2719) and to look up the `.scalafmt.conf` in the current working directory (#2720). After merging both, the following should work to format the `build.sc` and it's included files:

```
$ mill --meta-level 1 mill.scalalib.scalafmt.ScalafmtModule/reformatAll sources
Formatting 3 Scala sources
parsed config (v3.7.10): /home/lefou/work/opensource/mill/.scalafmt.conf
```

lolgab:
Thank you very much for this @lefou! 👏👏👏👏

PR #1802, opened 2022-03-28
Add transitive moduleDeps to BSP evaluated modules

### LINKED ISSUE #1800: Foreign Modules don't work with BSP
A build like:
```scala
// build.sc
import $file.foreign
import mill.scalalib._
object module extends ScalaModule {
  def scalaVersion = "2.13.8"
  def moduleDeps = Seq(foreign.foreign)
}

```

```scala
// foreign.sc
import mill._
import mill.scalalib._
object foreign extends ScalaModule {
  def scalaVersion = "2.13.8"
}
```

Makes Metals BSP import fail with:
```
Caught exception: java.lang.Exception: Failure during task evaluation: mill.define.Task$TraverseCtx@4bc7afb9 java.util.NoSuchElementException: key not found: foreign
    scala.collection.immutable.Map$Map2.apply(Map.scala:298)
    mill.bsp.MillBuildServer$$anonfun$2.applyOrElse(MillBuildServer.scala:279)
    mill.bsp.MillBuildServer$$anonfun$2.applyOrElse(MillBuildServer.scala:278)
    scala.collection.immutable.List.collect(List.scala:267)
    scala.collection.immutable.List.collect(List.scala:79)
    mill.bsp.MillBuildServer.$anonfun$workspaceBuildTargets$4(MillBuildServer.scala:278)
    mill.define.Task$TraverseCtx.evaluate(Task.scala:379)
java.lang.Exception: Failure during task evaluation: mill.define.Task$TraverseCtx@4bc7afb9 java.util.NoSuchElementException: key not found: foreign
    scala.collection.immutable.Map$Map2.apply(Map.scala:298)
    mill.bsp.MillBuildServer$$anonfun$2.applyOrElse(MillBuildServer.scala:279)
    mill.bsp.MillBuildServer$$anonfun$2.applyOrElse(MillBuildServer.scala:278)
    scala.collection.immutable.List.collect(List.scala:267)
    scala.collection.immutable.List.collect(List.scala:79)
    mill.bsp.MillBuildServer.$anonfun$workspaceBuildTargets$4(MillBuildServer.scala:278)
    mill.define.Task$TraverseCtx.evaluate(Task.scala:379)
	at mill.eval.Evaluator$.$anonfun$evalOrThrow$default$2$1(Evaluator.scala:998)
	at mill.eval.Evaluator$EvalOrThrow.apply(Evaluator.scala:987)
	at mill.bsp.MillBuildServer.targetTasks(MillBuildServer.scala:644)
	at mill.bsp.MillBuildServer.$anonfun$workspaceBuildTargets$1(MillBuildServer.scala:272)
	at mill.bsp.MillBuildServer.$anonfun$completable$1(MillBuildServer.scala:674)
	at mill.bsp.MillBuildServer.$anonfun$completable$1$adapted(MillBuildServer.scala:671)
	at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:484)
	at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1426)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)
workspaceBuildTargets took 882 msec
mar 27, 2022 9:53:14 PM org.eclipse.lsp4j.jsonrpc.RemoteEndpoint fallbackResponseError
GRAVE: Internal error: java.lang.Exception: Failure during task evaluation: mill.define.Task$TraverseCtx@4bc7afb9 java.util.NoSuchElementException: key not found: foreign
    scala.collection.immutable.Map$Map2.apply(Map.scala:298)
    mill.bsp.MillBuildServer$$anonfun$2.applyOrElse(MillBuildServer.scala:279)
    mill.bsp.MillBuildServer$$anonfun$2.applyOrElse(MillBuildServer.scala:278)
    scala.collection.immutable.List.collect(List.scala:267)
    scala.collection.immutable.List.collect(List.scala:79)
    mill.bsp.MillBuildServer.$anonfun$workspaceBuildTargets$4(MillBuildServer.scala:278)
    mill.define.Task$TraverseCtx.evaluate(Task.scala:379)
java.util.concurrent.CompletionException: java.lang.Exception: Failure during task evaluation: mill.define.Task$TraverseCtx@4bc7afb9 java.util.NoSuchElementException: key not found: foreign
    scala.collection.immutable.Map$Map2.apply(Map.scala:298)
    mill.bsp.MillBuildServer$$anonfun$2.applyOrElse(MillBuildServer.scala:279)
    mill.bsp.MillBuildServer$$anonfun$2.applyOrElse(MillBuildServer.scala:278)
    scala.collection.immutable.List.collect(List.scala:267)
    scala.collection.immutable.List.collect(List.scala:79)
    mill.bsp.MillBuildServer.$anonfun$workspaceBuildTargets$4(MillBuildServer.scala:278)
    mill.define.Task$TraverseCtx.evaluate(Task.scala:379)
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:331)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:346)
	at java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:704)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506)
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2088)
	at mill.bsp.MillBuildServer.$anonfun$completable$1(MillBuildServer.scala:683)
	at mill.bsp.MillBuildServer.$anonfun$completable$1$adapted(MillBuildServer.scala:671)
	at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:484)
	at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1426)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)
Caused by: java.lang.Exception: Failure during task evaluation: mill.define.Task$TraverseCtx@4bc7afb9 java.util.NoSuchElementException: key not found: foreign
    scala.collection.immutable.Map$Map2.apply(Map.scala:298)
    mill.bsp.MillBuildServer$$anonfun$2.applyOrElse(MillBuildServer.scala:279)
    mill.bsp.MillBuildServer$$anonfun$2.applyOrElse(MillBuildServer.scala:278)
    scala.collection.immutable.List.collect(List.scala:267)
    scala.collection.immutable.List.collect(List.scala:79)
    mill.bsp.MillBuildServer.$anonfun$workspaceBuildTargets$4(MillBuildServer.scala:278)
    mill.define.Task$TraverseCtx.evaluate(Task.scala:379)
	at mill.eval.Evaluator$.$anonfun$evalOrThrow$default$2$1(Evaluator.scala:998)
	at mill.eval.Evaluator$EvalOrThrow.apply(Evaluator.scala:987)
	at mill.bsp.MillBuildServer.targetTasks(MillBuildServer.scala:644)
	at mill.bsp.MillBuildServer.$anonfun$workspaceBuildTargets$1(MillBuildServer.scala:272)
	at mill.bsp.MillBuildServer.$anonfun$completable$1(MillBuildServer.scala:674)
	... 8 more

```

lolgab:
Fixes #1800 

Extracted the method used by `BloopImpl` to get the transitive modules from `rootModule` in a `internal` module and make BSP use it as well.

### DIFF
--- a/bsp/test/src/BspModulesTests.scala
+++ b/bsp/test/src/BspModulesTests.scala
@@ -25,7 +25,7 @@ object BspModulesTests extends ScriptTestSuite(false) {
           "HelloBsp.test",
           "foreign-modules.proj1.proj1",
           "foreign-modules.proj2.proj2",
-          // "proj3" // still not detected
+          // "foreign-modules.proj3.proj3" // still not detected
         ).sorted
         readModules ==> expectedModules
       }
--- a/bsp/test/src/BspModulesTests.scala
+++ b/bsp/test/src/BspModulesTests.scala
@@ -1,0 +1,34 @@
+import mill.bsp.BSP
+import mill.util.ScriptTestSuite
+import utest._
+
+object BspModulesTests extends ScriptTestSuite(false) {
+  override def workspaceSlug: String = "bsp-modules"
+  override def scriptSourcePath: os.Path = os.pwd / "bsp" / "test" / "resources" / workspaceSlug
+
+  def tests: Tests = Tests {
+    test("BSP module with foreign modules") {
+      test("can be installed") {
+        val workspacePath = initWorkspace()
+        eval("mill.bsp.BSP/install") ==> true
+        os.exists(workspacePath / ".bsp" / s"${BSP.serverName}.json") ==> true
+      }
+      test("ModuleUtils resolves all referenced transitive modules") {
+        val workspacePath = initWorkspace()
+        eval("validate") ==> true
+        val file = workspacePath / "out" / "validate.dest" / "transitive-modules.json"
+        os.exists(file) ==> true
+        val readModules = os.read.lines(file).sorted
+        val expectedModules = Seq(
+          "", // the root module has no segemnts at all
+          "HelloBsp",
+          "HelloBsp.test",
+          "foreign-modules.proj1.proj1",
+          "foreign-modules.proj2.proj2",
+          // "proj3" // still not detected
+        ).sorted
+        readModules ==> expectedModules
+      }
+    }
+  }
+}
--- a/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
@@ -4,7 +4,7 @@ import bloop.config.{Config => BloopConfig, Tag => BloopTag}
 import mill._
 import mill.define.{Module => MillModule, _}
 import mill.eval.Evaluator
-import mill.internal.ModuleUtils
+import mill.scalalib.internal.ModuleUtils
 import mill.scalajslib.ScalaJSModule
 import mill.scalajslib.api.{JsEnvConfig, ModuleKind}
 import mill.scalalib._
--- a/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
@@ -4,6 +4,7 @@ import bloop.config.{Config => BloopConfig, Tag => BloopTag}
 import mill._
 import mill.define.{Module => MillModule, _}
 import mill.eval.Evaluator
+import mill.internal.ModuleUtils
 import mill.scalajslib.ScalaJSModule
 import mill.scalajslib.api.{JsEnvConfig, ModuleKind}
 import mill.scalalib._
@@ -100,23 +101,12 @@ class BloopImpl(ev: () => Evaluator, wd: os.Path) extends ExternalModule { outer
   }
 
   // Compute all transitive modules from build children and via moduleDeps
+  @deprecated("Internal use only. To be removed", since = "mill 0.10.3")
   def transitiveModules(
       mod: define.Module,
       found: Seq[define.Module] = Seq.empty
   ): Seq[define.Module] = {
-    val skip = mod match {
-      case jm: JavaModule => skippable(jm)
-      case _ => false
-    }
-    if (found.contains(mod) || skip)
-      found
-    else {
-      val subMods = mod.millModuleDirectChildren ++ (mod match {
-        case jm: JavaModule => jm.moduleDeps
-        case other => Seq.empty
-      })
-      subMods.foldLeft(found ++ Seq(mod)) { (all, mod) => transitiveModules(mod, all) }
-    }
+    ModuleUtils.transitiveModules(mod, toSkip)
   }
 
   protected def computeModules: Seq[JavaModule] = {
@@ -128,8 +118,9 @@ class BloopImpl(ev: () => Evaluator, wd: os.Path) extends ExternalModule { outer
   }
 
   // class-based pattern matching against path-dependant types doesn't seem to work.
-  private def skippable(module: scalalib.JavaModule): Boolean =
-    if (module.isInstanceOf[outer.Module]) module.asInstanceOf[outer.Module].skipBloop
+  private def toSkip(module: MillModule): Boolean =
+    if (module.isInstanceOf[JavaModule] && module.isInstanceOf[outer.Module])
+      module.asInstanceOf[outer.Module].skipBloop
     else false
 
   /**
--- a/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
@@ -101,12 +101,12 @@ class BloopImpl(ev: () => Evaluator, wd: os.Path) extends ExternalModule { outer
   }
 
   // Compute all transitive modules from build children and via moduleDeps
-  @deprecated("Internal use only. To be removed", since = "mill 0.10.3")
+  @deprecated("Use mill.internal.ModuleUtils.transitiveModules instead", since = "mill 0.10.3")
   def transitiveModules(
       mod: define.Module,
       found: Seq[define.Module] = Seq.empty
   ): Seq[define.Module] = {
-    ModuleUtils.transitiveModules(mod, toSkip)
+    ModuleUtils.transitiveModules(mod, accept)
   }
 
   protected def computeModules: Seq[JavaModule] = {
@@ -118,10 +118,10 @@ class BloopImpl(ev: () => Evaluator, wd: os.Path) extends ExternalModule { outer
   }
 
   // class-based pattern matching against path-dependant types doesn't seem to work.
-  private def toSkip(module: MillModule): Boolean =
+  private def accept(module: MillModule): Boolean =
     if (module.isInstanceOf[JavaModule] && module.isInstanceOf[outer.Module])
-      module.asInstanceOf[outer.Module].skipBloop
-    else false
+      !module.asInstanceOf[outer.Module].skipBloop
+    else true
 
   /**
    * Computes sources files paths for the whole project. Cached in a way
--- a/scalalib/src/internal/ModuleUtils.scala
+++ b/scalalib/src/internal/ModuleUtils.scala
@@ -3,16 +3,18 @@ package mill.internal
 import mill.define.Module
 import mill.scalalib.JavaModule
 
+/**
+ * Compute all transitive modules from module children and via moduleDeps
+ */
 @mill.api.internal
 private[mill] object ModuleUtils {
-  // Compute all transitive modules from build children and via moduleDeps
-  def transitiveModules(module: Module, toSkip: Module => Boolean = _ => false): Seq[Module] = {
+  def transitiveModules(module: Module, accept: Module => Boolean = _ => true): Seq[Module] = {
     def loop(mod: Module, found: Seq[Module]): Seq[Module] = {
-      if (toSkip(mod) || found.contains(mod))
+      if (!accept(mod) || found.contains(mod))
         found
       else {
         val subMods = mod.millModuleDirectChildren ++ (mod match {
-          case jm: JavaModule => jm.moduleDeps
+          case jm: JavaModule => jm.moduleDeps ++ jm.compileModuleDeps
           case other => Seq.empty
         })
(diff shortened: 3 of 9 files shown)


lefou (commented):
* This change adds all those foreign modules to the BSP projects, which are referenced (in `moduleDeps`) by non-foreign modules. There is a slight chance, we still miss some modules. E.g. those, which aren't referenced at all and those whose targets are dependencies of other targets. The former modules are probably not an issue, as these also can't be reached from the Mill cli. One way to detect the latter is, to inspect all tasks/targets which is most probably an expensive operation.

* I'm not 100% familiar with the way Ammonite is handling all extra includes (which may represent foreign modules), but I think we should strive to collect enough knowledge about those modules in that early phase, and reuse here, for correctness. This doesn't need to happen in this PR though. In the more recent changes by @lolgab to the input files change detection (#1663), we should almost have touched the code which may also be aware of foreign modules, but I'm not sure of that. Also related, there is #1751, which demands some kind of access to the "source files" of a Mill builds.

* We should also include `compileModuleDeps` in the transitive module search.

scalalib/src/internal/ModuleUtils.scala
```diff
@@ -0,0 +1,24 @@
+package mill.internal
+
+import mill.define.Module
+import mill.scalalib.JavaModule
+
+@mill.api.internal
+private[mill] object ModuleUtils {
+  // Compute all transitive modules from build children and via moduleDeps
```
lefou, line 8: This should be a scaladoc comment.

scalalib/src/internal/ModuleUtils.scala
```diff
@@ -0,0 +1,24 @@
...
+  def transitiveModules(module: Module, toSkip: Module => Boolean = _ => false): Seq[Module] = {
```
lefou, line 9: I'd prefer a `accept` over a `toSkip`, which default to `true` to include a module. This will also better match semantics of `filter`.

scalalib/src/internal/ModuleUtils.scala
```diff
@@ -0,0 +1,24 @@
...
+    def loop(mod: Module, found: Seq[Module]): Seq[Module] = {
+      if (toSkip(mod) || found.contains(mod))
+        found
+      else {
+        val subMods = mod.millModuleDirectChildren ++ (mod match {
+          case jm: JavaModule => jm.moduleDeps
```
lefou, line 15:
```suggestion
          case jm: JavaModule => jm.moduleDeps ++ jm.compileModuleDeps
```

contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
```diff
@@ -100,23 +101,12 @@ class BloopImpl(ev: () => Evaluator, wd: os.Path) extends ExternalModule { outer
   }
 
   // Compute all transitive modules from build children and via moduleDeps
+  @deprecated("Internal use only. To be removed", since = "mill 0.10.3")
```
lefou, line 104:
```suggestion
  @deprecated("Use mill.internal.ModuleUtils.transitiveModules instead.", since = "mill 0.10.3")
```

lefou:
We definitely need the foreign module test case from issue #1800 in the test suite.

lefou:
> We definitely want the foreign module test case from issue #1800 in the test suite. If anybody want to contribute to some BSP test suite for Mill, your help would be very appreciated.

What we are really looking for is a BSP client test suite, which can simulate a server and runs test against arbitrary directories which represent the test cases. This tool is best located in BSP project, I think.

lolgab:
Thank you for the review @lefou, I will apply the suggestions.
I didn't update the tests because, as you also said, without a client and performing operations, the test passes before and after the changes, so it's pointless.
I tried writing a client but gave up.
I think this is an improvement on its own in the current state of things, and we can have the proper BSP tests as a separate Issue to tackle.
About foreign modules not work being first class ( yet :) ), I agree. There are a lot of rough edges, I'm trying them out on my personal projects so I can exercise the misbehaviours and fix them incrementally.
Another thing that breaks both Bloop and BSP is the fact that foreign modules identifiers don't contain any information about a module being foreign or about the file where it's stored.
So if you have two modules in two files with the same name, you break both Bloop and BSP imports.
In my vision Mill can become a build tool able to run big monorepos with minimal cache invalidation, and firstclass foreign modules are a very important part of this vision!

lefou:
> Thank you for the review @lefou, I will apply the suggestions. I didn't update the tests because, as you also said, without a client and performing operations, the test passes before and after the changes, so it's pointless.

Yeah. I realized that later, that's why I added the second comment and also asked (potential other readers) for help. 😉

> I tried writing a client but gave up.

Welcome to the club.

> I think this is an improvement on its own in the current state of things, and we can have the proper BSP tests as a separate Issue to tackle. About foreign modules not work being first class ( yet :) ), I agree. There are a lot of rough edges, I'm trying them out on my personal projects so I can exercise the misbehaviours and fix them incrementally. Another thing that breaks both Bloop and BSP is the fact that foreign modules identifiers don't contain any information about a module being foreign or about the file where it's stored. So if you have two modules in two files with the same name, you break both Bloop and BSP imports.

Mill is using a dedicated segment for foreign modules, I think. Maybe we can do the same in BSP?

> In my vision Mill can become a build tool able to run big monorepos with minimal cache invalidation, and firstclass foreign modules are a very important part of this vision!

Totally agreed.

lefou:
I think we can add a test for the transitive modules detection to `BspModuleTests`.

lefou:
See `mill.define.Ctx.foreign: Option[Segments]`.

lolgab:
> I think we can add a test for the transitive modules detection to `BspModuleTests`.

Since the transitive module detection happens inside the server I don't really know how to test it without a client.

> See `mill.define.Ctx.foreign: Option[Segments]`.

I will, but since it's a separate problem, I think it can be tackled separately.

lefou:
> > I think we can add a test for the transitive modules detection to `BspModuleTests`.
> 
> Since the transitive module detection happens inside the server I don't really know how to test it without a client.

I had some kind of Unit test for `ModuleUtils.transitiveModule` in mind. But since we want a mini Mill run, which read a `build.sc`, we probably want place it into `BspInstallTests` which is already a `ScriptTestSuite`.  We could call a `check` target which calls the unit under test and writes the result into a file. Then we can assert the content of that file. 

If that is too diffuse to you, I can try to implement and push my idea to your branch.

> > See `mill.define.Ctx.foreign: Option[Segments]`.
> 
> I will, but since it's a separate problem, I think it can be tackled separately.

Sure. I wanted to provide a pointer to what I meant by "dedicated segment".

lolgab:
> If that is too diffuse to you, I can try to implement and push my idea to your branch.

If you don't mind, that would be great.

bsp/test/src/BspModulesTests.scala
```diff
@@ -0,0 +1,34 @@
...
+        val file = workspacePath / "out" / "validate.dest" / "transitive-modules.json"
+        os.exists(file) ==> true
+        val readModules = os.read.lines(file).sorted
+        val expectedModules = Seq(
+          "", // the root module has no segemnts at all
+          "HelloBsp",
+          "HelloBsp.test",
+          "foreign-modules.proj1.proj1",
+          "foreign-modules.proj2.proj2",
+          // "proj3" // still not detected
```
lefou, line 28: I have no idea how to detect this. We could reflect all targets of known modules and inspect their inputs, but although we have their segments, there is no way to find the owning modules.

lefou:
> > If that is too diffuse to you, I can try to implement and push my idea to your branch.
> 
> If you don't mind, that would be great.

@lolgab Done.

bsp/test/src/BspModulesTests.scala
lolgab: No idea either..

lolgab:
Thank you for adding the tests 🙏
LGTM! 👍

bsp/test/src/BspModulesTests.scala
lefou: The expected project name was incorrect.

```suggestion
          // "foreign-modules.proj3.proj3" // still not detected
```

lolgab: Maybe a module could register itself in its constructor in some sort of global `identifier => module` Map.
Every module when it first gets instantiated registers itself in the Mill "runtime".
Could this be a thing?
Probably it can be generalized to tasks as well.

lefou: I'd rather avoid such global state. This will get in the way for any dynamically created module/target and also effectively impedes Mill usage as a library. It may even break our tests.

I think we should just find a way to add that missing information when processing the `import $file` lines. Then we can expose this info through a (synthetic) target in the main/base module or via API.

Issue #1792, opened 2022-03-22
Incorrect & non-performant implementation for `JavaModule.bspCompileClasspath`


abernardi597:
There's a bug in `JavaModule.bspCompileClasspath` that causes the build server to report modules' compile classpaths incorrectly.
```diff
@internal
def bspCompileClasspath(pathsResolver: Task[EvaluatorPathsResolver]): Task[Seq[PathRef]] = {
  def bspLocalClasspath(
      j: JavaModule,
      pathsResolver: Task[EvaluatorPathsResolver]
  ): Task[Seq[PathRef]] = {
-    val cl = bspCompileClassesPath(pathsResolver)
+    val cl = j.bspCompileClassesPath(pathsResolver)
    T.task {
-      resources() ++ Seq(cl())
+      j.resources() ++ Seq(cl())
    }
  }

  def bspTransitiveLocalClasspath(
      j: JavaModule,
      pathsResolver: Task[EvaluatorPathsResolver]
  ): Task[Seq[PathRef]] = {
    val res = T.traverse(j.moduleDeps ++ j.compileModuleDeps)(m =>
      T.task {
        bspLocalClasspath(m, pathsResolver)() ++ bspTransitiveLocalClasspath(m, pathsResolver)()
      }
    )
    T.task {
      val res2: Seq[PathRef] = res().flatten
      res2
    }
  }

  val lcp = bspTransitiveLocalClasspath(this, pathsResolver)
  T.task {
    lcp() ++
      resources() ++
      unmanagedClasspath() ++
      resolvedIvyDeps()
  }
}
```

Despite this fix, you'll note that a module can be encountered multiple times when it is a common dependency for any other dependencies.
For example, in the project structure below, `A` would traverse `D` twice: once while processing the direct dependencies of `B`, and once more while processing the direct dependencies of `C`

```mermaid
graph TD
A --> B
A --> C
B --> D
C --> D
```

This causes fairly abysmal performance with larger projects.

One solution would be to use DFS to collect a `Set[JavaModule]` of all the transitive dependencies so each is only expanded once.
Here is a mix-in trait based on this principle that should be correct and more performant:
```scala
trait BspFix { this: JavaModule =>
  override def bspCompileClasspath(pathsResolver: Task[EvaluatorPathsResolver]): Task[Seq[PathRef]] = {
    @scala.annotation.tailrec
[...]

PR #1794, opened 2022-03-24
Improve performance of BSP classpath retrieval

abernardi597:
Issue reported in #1792

PR #1803, opened 2022-03-28
Reworked some BSP module inspection tasks

lefou:
This should improve (fix) performance and OOM issues with large multi module projects where modules depend heavily on each other.

`bspCompileClassesPath` and `bspCompileClasspath` are now Targets instead of Tasks. This makes them cacheable and also easier to override in derived modules. I also changed the logic in some `transitiveXXX` targets from a recursive call to all dependencies to just calling the non-recursive target on a transitive modules list. This should mitigate OOM issue as reported here: https://github.com/com-lihaoyi/mill/issues/1792#issuecomment-1078815626

Fix https://github.com/com-lihaoyi/mill/issues/1792

Supersedes https://github.com/com-lihaoyi/mill/pull/1794

### DIFF
--- a/bsp/src/mill/bsp/MillJavaBuildServer.scala
+++ b/bsp/src/mill/bsp/MillJavaBuildServer.scala
@@ -23,15 +23,16 @@ trait MillJavaBuildServer extends JavaBuildServer { this: MillBuildServer =>
         agg = (items: Seq[JavacOptionsItem]) => new JavacOptionsResult(items.asJava)
       ) {
         case (id, m: JavaModule) =>
-          val pathResolver = T.task(evaluator.pathsResolver)
+          val pathResolver = evaluator.pathsResolver
           T.task {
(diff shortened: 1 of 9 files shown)


lefou:
Turns out, Mill's `compileModuleDeps` handling was a bit sloppy. The calclulation of the transitive module dependencies also included transitive compile-only dependencies, which should have never end up in. My change fixed that. As a result, I needed to correct our `build.sc` and add some dependencies, to make the self-bootstrap tests succeed.

lolgab:
@lefou There is a bug in `mill-mima` that ignores the `mimaExcludeAnnotations`. I reproduced it in a test but the implementation seems exactly the same as Sbt MiMa so I couldn't understand why it doesn't work.
I'm now trying to call MiMa from an insulated classloader to see if the problem is related to the classpath.
In the meantime, the only workaround is to manually add excludes as you did.

PR #1804, opened 2022-04-01
BSP: Indicate foreign modules by adding the foreign segments to the name

lefou:


scalalib/src/mill/scalalib/bsp/BspModule.scala
```diff
@@ -15,7 +15,7 @@ trait BspModule extends Module {
   /** Use to fill most fields of `BuildTarget`. */
   @internal
   def bspBuildTarget: BspBuildTarget = BspBuildTarget(
-    displayName = Some(millModuleSegments.render),
+    displayName = Some(s"${millOuterCtx.foreign.map(f => s"${f.render}.").mkString}${millModuleSegments.render}"),
```
lolgab, line 18: Do we use the same name in Bloop? Should it be calculated in a shared place and used from both?
I remember having had clashes in Bloop as well.

lefou: Looks like it's also adding a foreign prefix:

https://github.com/com-lihaoyi/mill/blob/c4bd33965579d718d5d854d6a9b92fa9fe3afe88/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala#L152

lefou: This was also used for the project file in Bloop, but not the project name, so I fixed it and also shared the implementation as suggested.

lolgab (approved):
Great changes! Thank you!

PR #1806, opened 2022-04-01
Bloop: Replace deprecated API call

lefou:
Follow up to #1802

### DIFF
--- a/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
@@ -112,7 +112,8 @@ class BloopImpl(ev: () => Evaluator, wd: os.Path) extends ExternalModule { outer
   protected def computeModules: Seq[JavaModule] = {
     val eval = ev()
     if (eval != null)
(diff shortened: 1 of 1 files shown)


contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
```diff
@@ -112,7 +112,8 @@ class BloopImpl(ev: () => Evaluator, wd: os.Path) extends ExternalModule { outer
   protected def computeModules: Seq[JavaModule] = {
     val eval = ev()
     if (eval != null)
-      transitiveModules(eval.rootModule).collect { case jm: JavaModule => jm }
+      ModuleUtils.transitiveModules(eval.rootModule, accept = _.isInstanceOf[JavaModule])
+        .collect { case jm: JavaModule => jm }
```
lefou, line 116: @lolgab I woulder why this is not the equivalent of the replaced call. The failed tests suggest there is a difference. Any idea?

lefou: Nevermind, I found the issue.

PR #1811, opened 2022-04-02
Don't include resources into compileClasspath.

### LINKED ISSUE #1807: Java/Scala: Improve resources handling
Mill currently only has one `resources` target. It's content will end up in the built jar and also in the classpath. But it is also evaluate before and feed into `compile` target. This has various implications:

* Each change to `resources` will trigger a recompilation
* Generated resources (after compile) are not supported

See also the following discussion on Gitter: https://gitter.im/lihaoyi/mill?at=6246909d9a09ab24f3d00982

> Victor Borja @vic 07:41
Hello, running compile on a ~400 scala source files module I've noticed something weird: turns out if you do mill inspect foo.compileClasspath it includes the output of foo.resources that is somewhat surprising to me, and noticed because if I change eg, foo/resources/application.conf then that will cause scalac to compile all ~400 sources again, just for changing a value that will be used at runtime. So in my project I fixed it by overriding the compileClasspath task and removing the same module resources from it. I believe if some project compilation depends on it's own resources, they should be explicitly added to compileClasspath instead of it being the default. The only way I imagine it useful for resources being on compileClasspath of the SAME module would be for some advanced macros to generate code from resources or something. On the other hand, editing a foo/resources/application.conf or whatever is a pretty much more common scenario and that would likely trigger compilation since one of the compile task input changes. I'd be happy to provide a pull-request if you people think this is something worth fixing.
>
>Tobias Roeser @lefou 08:39
>@vic In the past, I though about adding some "runtimeResources" multiple times. Something that ends up in the final JAR and/or the runtime classpath. The motivation was to add stuff that gets generated after the compilation. It's probably the same feature but motivated from the other side. As we probably going to need both: resources in the compile classpath and resources after, we may want to avoid breaking compatibility and just add a new runResources? WDYT?
>
>Andrew Richards @ajrnz 10:02
>Regarding the resources issue I'd vote for changing the default. I come across this issue when modifying application.conf, logback.xml, web content and probably more file types. Apart from the macros example above I can't think of any reason to have them on the compile path (counter examples welcome). In the past I've removed the resources from the compile path when doing for example web development where the .html, .css etc were in resources. Adding a runResources would require a new directory and a name for it and would be doing what I imagine most people want by default.

To improve the resource handling, I'm interested in your opinion. Which route do you prefer?

1. Break behavioral compatibility: No longer feed `resources` into `compile` and add a `compileResource` target which is feed into `compile` instead
2. Keep behavior and the `resources` target as is and introduce a new `runResources`

vic:
Instead we now have a compileResources that is part of compileClasspath.

Fixes #1807 

* #1826

### DIFF
--- a/scalalib/src/JavaModule.scala
+++ b/scalalib/src/JavaModule.scala
@@ -238,12 +238,15 @@ trait JavaModule
   def sources = T.sources { millSourcePath / "src" }
 
   /**
-   * The folders where the resource files for this module live
+   * The folders where the resource files for this module live.
+   * If you need resources to be seen by the compiler, use [[compileResources]].
    */
   def resources: Sources = T.sources { millSourcePath / "resources" }
 
   /**
-   * The folders where the compile time resource files for this module live
+   * The folders where the compile time resource files for this module live.
+   * If your resources files do not necessarily need to be seen by the compiler,
+   * you should use [[resources]] instead.
    */
   def compileResources: Sources = T.sources { millSourcePath / "compile-resources" }
 
--- a/scalalib/src/JavaModule.scala
+++ b/scalalib/src/JavaModule.scala
@@ -243,6 +243,11 @@ trait JavaModule
   def resources: Sources = T.sources { millSourcePath / "resources" }
 
   /**
+   * The folders where the compile time resource files for this module live
+   */
+  def compileResources: Sources = T.sources { millSourcePath / "compile-resources" }
+
+  /**
    * Folders containing source files that are generated rather than
    * hand-written; these files can be generated in this target itself,
    * or can refer to files generated from other targets
@@ -325,7 +330,7 @@ trait JavaModule
   // Keep in sync with [[bspCompileClasspath]]
   def compileClasspath: T[Agg[PathRef]] = T {
     transitiveLocalClasspath() ++
-      resources() ++
+      compileResources() ++
       unmanagedClasspath() ++
       resolvedIvyDeps()
   }
@@ -335,7 +340,7 @@ trait JavaModule
   @internal
   def bspCompileClasspath: Target[Agg[UnresolvedPath]] = T {
     bspTransitiveLocalClasspath() ++
-      (resources() ++ unmanagedClasspath() ++ resolvedIvyDeps())
+      (compileResources() ++ unmanagedClasspath() ++ resolvedIvyDeps())
         .map(p => UnresolvedPath.ResolvedPath(p.path))
   }
 
@@ -515,7 +520,7 @@ trait JavaModule
    */
   def sourceJar: Target[PathRef] = T {
     Jvm.createJar(
-      (allSources() ++ resources()).map(_.path).filter(os.exists),
+      (allSources() ++ resources() ++ compileResources()).map(_.path).filter(os.exists),
       manifest()
     )
   }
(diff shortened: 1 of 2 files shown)


lefou:
This looks ok so far, but this need to wait until the next major version (probably `0.11`), as it changes the dependencies between targets which is an issue for downstream plugins.

To get CI green, some test need to be adapted.

lefou:
> This looks ok so far, but this need to wait until the next major version (probably `0.11`), as it changes the dependencies between targets which is an issue for downstream plugins.
> 
> To get CI green, some test need to be adapted.

We have no clear rule when to start development on next major. Until now, we just bumped the version when needed. So that might be also the case for this PR. But as we now take compatibility serious, we should avoid too many major bumps in a short time to remove stress from downstream projects like external Mill plugins.

lefou:
This is the error in CI:

```
X mill.scalalib.bsp.BspModuleTests.bspCompileClasspath.single module.dependent module 362ms 
  utest.AssertionError: #2: relResults == expected,
  relResults: Seq[os.FilePath] = Vector(/home/runner/work/mill/mill/target/worksources/mill/scalalib/bsp/BspModuleTests/MultiBase/HelloBsp/compile-resources, /home/runner/work/mill/mill/target/worksources/mill/scalalib/bsp/BspModuleTests/MultiBase/HelloBsp2/resources, /home/runner/work/mill/mill/target/workspace/mill/scalalib/bsp/BspModuleTests/eval/bspCompileClasspath/single module/dependent module/HelloBsp/compile.dest/classes, logback-classic-1.1.10.jar, logback-core-1.1.10.jar, scala-library-2.13.8.jar, slf4j-api-1.7.34.jar)
  expected: Seq[os.FilePath] = List(/home/runner/work/mill/mill/target/worksources/mill/scalalib/bsp/BspModuleTests/MultiBase/HelloBsp/resources, /home/runner/work/mill/mill/target/worksources/mill/scalalib/bsp/BspModuleTests/MultiBase/HelloBsp2/resources, /home/runner/work/mill/mill/target/workspace/mill/scalalib/bsp/BspModuleTests/eval/bspCompileClasspath/single module/dependent module/HelloBsp/compile.dest/classes, logback-classic-1.1.10.jar, logback-core-1.1.10.jar, scala-library-2.13.8.jar, slf4j-api-1.7.34.jar)
    mill.scalalib.bsp.BspModuleTests$.$anonfun$tests$12(BspModuleTests.scala:81)
    mill.scalalib.bsp.BspModuleTests$.$anonfun$tests$12$adapted(BspModuleTests.scala:59)
    mill.scalalib.bsp.BspModuleTests$.workspaceTest(BspModuleTests.scala:37)
    mill.scalalib.bsp.BspModuleTests$.$anonfun$tests$11(BspModuleTests.scala:40)
```

You need to reflect the changed resource directory name in the `expected` result.

vic:
@lefou thanks will fix that. I saw a lot of ci failures and didn't still had enough time to find which of them were caused by my change. thanks for pointing out. any particular suite you recommend me to run ?

lefou:
As you touch `scalalib` it's a good idea to run `scalalib.test`, at least. It's perfectly ok to push add just review what the CI reports. Besides of some sporadic CI failures (which I typically review and restart from time to time), all test must pass.

lefou:
We finally merged it. Thanks @vic for this contribution and your patience!

PR #1815, opened 2022-04-05
Added `docJarUseArgsFile` target and default to `true` on Windows

### LINKED ISSUE #1814: docJar can easily overflow windows command length restrictions
When the combined length of all the source-file names and classpath entries gets too long, we get: `java.io.IOException: Cannot run program "javadoc" (in directory "D:\Tmp\AOI\out\Suite\javadoc.dest"): CreateProcess error=206, File Name or its extension is too long`

This seems to be because Windows has a limit to the length of command that is used to create a process. That length is right about 32k characters. There's a [Stack overflow post](https://stackoverflow.com/a/28452546/1922925) that has a nice, basic explanation as to the "why".

It does not take a particularly large project to trigger this behavior: The project that I'm working on has ~470 java files, and the *relative* paths for them (from `T.workspace`) add up to 27k characters. It does not take a very long base path to over-run the limit.

It does look like this has surfaced before, with no actual fix: #993 & #207

Options that I see:

- Work around by only building javadoc on non-windows systems. Not my first choice.
- `javadoc` can take *relative* filepaths. This might help a bit, especially with shallow package structures.
- `javadoc` can also take the input files in the form of a list written to a single file. This appears most flexible to me.
    - Note that the limit is for the whole command. A long `compileClasspath` would also trigger the issue, though it is less likely. Perhaps these should be passed in a file as well?

lefou:
* Fix #1814

### DIFF
--- a/main/src/mill/modules/Jvm.scala
+++ b/main/src/mill/modules/Jvm.scala
@@ -65,7 +65,7 @@ object Jvm {
   }
 
   /**
-   * Resolved a tool to a path under the currently used JDK (if known).
+   * Resolves a tool to a path under the currently used JDK (if known).
    */
   def jdkTool(toolName: String): String = {
     sys.props
--- a/main/src/mill/modules/Jvm.scala
+++ b/main/src/mill/modules/Jvm.scala
@@ -74,6 +74,16 @@ object Jvm {
       .filter(f => f.exists())
       .fold("java")(_.getAbsolutePath())
 
+  def javaDocExe: String =
+    sys.props
+      .get("java.home")
+      .map(h =>
+        if (isWin) new File(h, "bin\\javadoc.exe")
+        else new File(h, "bin/javadoc")
+      )
+      .filter(f => f.exists())
+      .fold("javadoc")(_.getAbsolutePath())
+
   /**
    * Runs a JVM subprocess with the given configuration and streams
    * it's stdout and stderr to the console.
--- a/main/src/mill/modules/Jvm.scala
+++ b/main/src/mill/modules/Jvm.scala
@@ -64,25 +64,22 @@ object Jvm {
     os.proc(commandArgs).call(cwd = workingDir1, env = envArgs)
   }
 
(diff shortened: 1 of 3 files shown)


Sailsman63:
Tested successfully on Windows 10 with Temurin 17.0.2+8

Test-case was a project large enough to trigger #1814

main/src/mill/modules/Jvm.scala
```diff
@@ -64,15 +64,22 @@ object Jvm {
     os.proc(commandArgs).call(cwd = workingDir1, env = envArgs)
   }
 
-  def javaExe: String =
+  /**
+   * Resolved a tool to a path under the currently used JDK (if known).
```
lefou, line 68:
```suggestion
   * Resolves a tool to a path under the currently used JDK (if known).
```

nightscape:
@lefou I just ran into the same `java.io.IOException: CreateProcess error=206, The filename or extension is too long` when trying to run `mill test.repl`.
I was wondering if the functionality introduced here should maybe moved to [Jvm#callSubprocess](https://github.com/lefou/mill/blob/17dea9c9a858abf61014f39f58959bfe0da15653/main/util/src/mill/util/Jvm.scala#L34), so that it applies to all spawned JVM processes?

lefou:
@nightscape If you use [Jvm#runSubprocess](https://github.com/lefou/mill/blob/17dea9c9a858abf61014f39f58959bfe0da15653/main/util/src/mill/util/Jvm.scala#L79), you will find the `useCpPassingJar`, which will help to cut the length of the otherwise long `-classpath` argument. 

Whether a tool supports reading it's arguments as a file and whether this file is denoted with a `@`-prefix (like in javadoc tool and like some cli parsers support, e.g. [CmdOption](https://github.com/ToToTec/CmdOption?tab=readme-ov-file#reading-arguments-from-file-prefix)), is typically specific and not easy to detect. I don't think it's easy to generically support that case.

PR #1825, opened 2022-04-07
Support transparently created tasks as command parameters

lefou:


lihaoyi (approved):
This looks really interesting. 

The code change is trivial and straightforward, so not much to say there. 

At first pass, this lets us call methods taking `Task[T]` directly from the CLI as `T.command`, which is a convenience we didn't have before. This might let us do without some trivial forwarders.

As a consequence, it seems we can call the `T.command` while making some changes to the parameters within a `T.task` block. This is a bit limited - you can't have code that depends on multiple Tasks in that block - but could still be useful. At a certain level of complexity, you would have to split out a helper method anyway.

Also, it seems that now `T.command`s are now kind of dynamic, as `T.task`s already are: the exact dependencies of the command may include stuff that was passed in. We could already have conditional dependencies in the status quo, e.g. by having code that depends only on the fixed input params before calling the `T.command` wrapper, so I don't think this will significantly loosen any invariants

lefou:
Yeah. More often ask question about overriding commands was in the context of `run`. Often, users want to hook something before the actual run occurs. Their naive attempts almost always results in execution after `super.run()()`, as it is a dependency. Now, they can just add their hook as arg-task to ensure, it runs before. Maybe not idiomatic or clean, it solves real world issues.

PR #1828, opened 2022-04-08
Add a delay to the client to catch all the server data

### LINKED ISSUE #1816: Mill output from server is truncated
There definitely seems to be some sort of miscommunication from the server to the client regarding termination. For example if you write a program:

```scala
object MillTest extends App {
  (1 to 100000).foreach{ i =>
    println(i)
  }
}
```

in a `mill-test` project and run

    mill mill-test.run > result.txt

the file `result.txt` is truncated. The same happens if you run it to the console. Running with `-i` does not have the problem.

[If you think this is unrelated I am happy to open it as its own issue]

_Originally posted by @ajrnz in https://github.com/com-lihaoyi/mill/issues/1711#issuecomment-1071354558_

### LINKED ISSUE #1817: FileToStreamTailer no longer serves a purpose
FileToStreamTailer appears to be obsolete in the new pumper scene introduced via @lihaoyi a few months back.

I'd like to remove it.

It appears broken anyway. The only time it writes to the stream is when it is not ignoring the head. Ie before it starts the process of tailing. Correct me if I'm mistaken.

I've been running with it disabled and it makes no difference.

I'd like to submit a PR to remove it and it's use - unless it is worth keeping (and fixing). Let me know.

ajrnz:
Fixes #1816 and closes #1817


lefou:
I test your branch, and didn't see the expected output "hello" anymore.

Here is my test `build.sc`

```scala
import mill._

def run() = T.command {
  os.proc("echo", "hello").call(stdout = os.Inherit)
}
```

Here is the output:

```
$ mill -i dev.run test-subprocess-output run
[807/807] dev.run 
[1/1] run 
```

And here the expected output with mill main

```
$ mill -i dev.run test-subprocess-output run
[807/807] dev.run 
[1/1] run 
hello

```

ajrnz:
Hmm, strange. I can reproduce. Let me take a look...

lefou (changes requested):
Output of subprocesses are no longer shown.

lefou:
@ajrnz Do you have a rough estimate when you can finish this? I'd like to cut a new release with the fixes we already have, and this one would be a nice one to also have. Maybe, just re-add the `FileToStreamTailer`. The still gain from the pumper delay.

ajrnz:
@lefou Sorry I haven't had time to look at this yet. 
I've added `FileToStreamTailer` for the moment. As you say, it is probably better with it. 
Feel free to merge it and I'll take a look at `FileToStreamTailer` when I get a chance.

lefou:
Great, thank you! Just right in time to be release as `0.10.3`.

PR #1843, opened 2022-04-18
make bloop aware of runtime dependencies

megri:
At the moment, mill generates bloop configurations without setting the runtime classpath of Java modules. This makes bloop fall back to `project.classpath` which corresponds with required compile-time classpath. It makes sense to put stuff like slf4j-adapters, postgres-drivers and such—dependencies that are only required during runtime—in the JVM-configuration.

The following discussion on the bloop issue tracker highlights the requirements and the accepted solution: https://github.com/scalacenter/bloop/issues/1233
https://github.com/scalacenter/bloop/pull/1244

This pull-request simply outputs the `runClasspath` target to `BloopConfig.Platform.Jvm#classpath`. I'm not a 100 % sure this is the correct approach, but it corresponds to what gets put on the classpath when having mill run a project (as far as I can tell, from looking at `mill inspect app.run`.

### DIFF
--- a/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
@@ -188,6 +188,27 @@ class BloopImpl(ev: () => Evaluator, wd: os.Path) extends ExternalModule { outer
         case None => T.task(None)
       }).map(_.getOrElse(Config.LinkerMode.Debug))
 
+    // //////////////////////////////////////////////////////////////////////////
+    //  Classpath
+    // //////////////////////////////////////////////////////////////////////////
+
+    val classpath = T.task {
+      val depModules = (module.compileModuleDeps ++ module.recursiveModuleDeps).distinct
+      // dep modules ++ ivy deps ++ unmanaged
+      depModules.map(classes) ++
+        module.resolvedIvyDeps().map(_.path) ++
+        module.unmanagedClasspath().map(_.path)
+    }
+
+    val runtimeClasspath = T.task {
+      // dep modules ++ ivy deps ++ unmanaged
+      module.recursiveModuleDeps.map(classes) ++
+        module.resolvedRunIvyDeps().map(_.path) ++
+        module.unmanagedClasspath().map(_.path)
+    }
+
+    val resources = T.task(module.resources().map(_.path.toNIO).toList)
+
     val platform: Task[BloopConfig.Platform] = module match {
       case m: ScalaJSModule =>
         T.task {
@@ -247,7 +268,7 @@ class BloopImpl(ev: () => Evaluator, wd: os.Path) extends ExternalModule { outer
             ),
             mainClass = module.mainClass(),
             runtimeConfig = None,
-            classpath = None,
+            classpath = Some(runtimeClasspath().map(_.toNIO).toList),
             resources = Some(module.resources().map(_.path.toNIO).toList)
           )
         }
@@ -356,23 +377,6 @@ class BloopImpl(ev: () => Evaluator, wd: os.Path) extends ExternalModule { outer
         allIvyDeps.map(module.resolveCoursierDependency()).toList
       BloopConfig.Resolution(artifacts(repos, coursierDeps))
     }
-
-    // //////////////////////////////////////////////////////////////////////////
-    //  Classpath
-    // //////////////////////////////////////////////////////////////////////////
-
-    val ivyDepsClasspath = module.resolvedIvyDeps.map(_.map(_.path).toSeq)
-
-    def transitiveClasspath(m: JavaModule): Task[Seq[os.Path]] = T.task {
-      (m.moduleDeps ++ m.compileModuleDeps).map(classes) ++
-        m.unmanagedClasspath().map(_.path) ++
-        T.traverse(m.moduleDeps ++ m.compileModuleDeps)(transitiveClasspath)().flatten
-    }
-
-    val classpath = T
-      .task(transitiveClasspath(module)() ++ ivyDepsClasspath())
-      .map(_.distinct)
-    val resources = T.task(module.resources().map(_.path.toNIO).toList)
 
     // //////////////////////////////////////////////////////////////////////////
     //  Tying up
--- a/contrib/bloop/test/src/mill/contrib/bloop/BloopTests.scala
+++ b/contrib/bloop/test/src/mill/contrib/bloop/BloopTests.scala
@@ -35,6 +35,14 @@ object BloopTests extends TestSuite {
         "-language:higherKinds"
       )
 
+      override def compileIvyDeps = Agg(
+        ivy"org.reactivestreams:reactive-streams:1.0.3"
+      )
+
+      override def runIvyDeps = Agg(
+        ivy"org.postgresql:postgresql:42.3.3"
+      )
+
       object test extends super.Tests with TestModule.Utest
     }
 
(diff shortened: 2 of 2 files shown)


megri:
I need to change my approach: the generated bloop config references classes in the mill out path. This had me very confused as changes to my code didn't register properly.

contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
```diff
@@ -188,6 +188,27 @@ class BloopImpl(ev: () => Evaluator, wd: os.Path) extends ExternalModule { outer
...
+    //  Classpath
+    // //////////////////////////////////////////////////////////////////////////
+
+    val ivyDepsClasspath = module.resolvedIvyDeps.map(_.map(_.path).toSeq)
+    val runIvyDepsClasspath = module.resolvedRunIvyDeps.map(_.map(_.path).toSeq)
+
+    def transitiveClasspath(m: JavaModule): Task[Seq[os.Path]] = T.task {
+      (m.moduleDeps ++ m.compileModuleDeps).map(classes) ++
+        m.unmanagedClasspath().map(_.path) ++
+        T.traverse(m.moduleDeps ++ m.compileModuleDeps)(transitiveClasspath)().flatten
```
lefou, line 201: I don't think adding `compileModuleDeps` transitively is the right thing here. Compile-time dependencies should only be relevant at compilation time and typically don't be part of the transitive dependencies.

lefou (commented):
I have two concerns:

First, I think the transitive dependency calculation isn't correct, because it includes also compile-time dependencies transitively, which is wrong. I also realize this issue was already present before your PR, but as you touched this code, I wanted to point it out. And at this point, I'd like to request a test case resembling a project with multiple modules and ensuring the classpaths and run-classpaths are correct, which brings me to the other point.

This is a contrib module, where we don't enforce the same strict quality standards. I guess this means, we can just merge as is, with missing tests and such. But as Bloop is used by many users (I guess), we should address my first concern.

lefou:
Oh, and thanks for creating this PR. Totally forget that, as I were so concentrated on the implementation.

contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
megri: > I have two concerns:
> 
> First, I think the transitive dependency calculation isn't correct, because it includes also compile-time dependencies transitively, which is wrong. I also realize this issue was already present before your PR, but as you touched this code, I wanted to point it out. And at this point, I'd like to request a test case resembling a project with multiple modules and ensuring the classpaths and run-classpaths are correct, which brings me to the other point.
> 
> This is a contrib module, where we don't enforce the same strict quality standards. I guess this means, we can just merge as is, with missing tests and such. But as Bloop is used by many users (I guess), we should address my first concern.

Thanks for getting back on the issue! Yeah, I was a bit suspicious of the quality of the class as well, but to be honest I don't know exactly what needs to go where.

What I do know is that the classpath of the jvm-specific bloop configuration needs to contain all compiled classes that the module depends on, as well as any runtime dependencies. So it's basically a superset of the project classpath if I understand things correctly. 

Are you saying it is correct that the classpath consist of module.ivyDeps ++ compiled moduleDeps classes but without their own corresponding classpath dependencies (and ++ module.runIvyDeps for the jvm-part of course)?

megri: If we trust the SBT implementation I could try to create a multi-project SBT configuration and have a look at the generated bloop-files from that, and try to align the output of the BloopImpl.

lefou: I can't say anything about the sbt implementation.

Compile-time dependencies should only be visible at compile time. Once compilation is done and the compiled classes are part of the classpath, compile-time dependencies should no longer be present. If you need the same dependency (or an alternative) at runtime, you need to add them to runIvyDeps.

There is already a test suite for Bloop (`mill.contrib.bloop.BloopTests`), so best would be to review or add to it, so that we can assert the wanted outcome. If it helps to cross-check against SBT, I'd say, go for it. Ideally, with the exact same test project. The more consistent our ecosystem is, the better.

lefou: > Are you saying it is correct that the classpath consist of module.ivyDeps ++ compiled moduleDeps classes but without their own corresponding classpath dependencies (and ++ module.runIvyDeps for the jvm-part of course)?

This is correct for compile-time! At runtime, we should leave the `compileModuleDeps` out.

lefou:
I wrote the classpaths down how I think they should be used by bloop. But please beware that I haven't tested them.

```scala
    val compileClasspath = T.task {
      val depModules = (module.compileModuleDeps ++ module.recursiveModuleDeps).distinct
      // dep modules ++ ivy deps ++ unmanaged
      depModules.map(classes) ++
        module.resolvedIvyDeps().map(_.path) ++
        module.unmanagedClasspath().map(_.path)
    }

    val runtimeClasspath = T.task {
      // dep modules ++ ivy deps ++ unmanaged
      module.recursiveModuleDeps.map(classes) ++
        module.resolvedRunIvyDeps().map(_.path) ++
        module.unmanagedClasspath().map(_.path)
    }
```

I don't know how bloop is handling resources, so I completely ignored that part.

megri:
> I wrote the classpaths down how I think they should be used by bloop. But please beware that I haven't tested them.
> 
> ```scala
>     val compileClasspath = T.task {
>       val depModules = (module.compileModuleDeps ++ module.recursiveModuleDeps).distinct
>       // dep modules ++ ivy deps ++ unmanaged
>       depModules.map(classes) ++
>         module.resolvedIvyDeps().map(_.path) ++
>         module.unmanagedClasspath().map(_.path)
>     }
> 
>     val runtimeClasspath = T.task {
>       // dep modules ++ ivy deps ++ unmanaged
>       module.recursiveModuleDeps.map(classes) ++
>         module.resolvedRunIvyDeps().map(_.path) ++
>         module.unmanagedClasspath().map(_.path)
>     }
> ```
> 
> I don't know how bloop is handling resources, so I completely ignored that part.

I tried recompiling mill with these changes and the generated bloop config seemingly works for my setup! That is, postgres driver and slf4j-adapter is made available only during runtime.

lefou:
Great. Do you intend to add a test case, too? 

If not, maybe you could share your `build.sc`, so I can try to make an realistic test case of it.

megri:
> Great. Do you intend to add a test case, too?
> 
> If not, maybe you could share your `build.sc`, so I can try to make an realistic test case of it.

I can have a look at it!

How do you imagine the test should look like? A simple build definition, running bloopimpl and then comparing the output to a predefined expected result, or something more involved?

lefou:
> I can have a look at it!
> 
> How do you imagine the test should look like? A simple build definition, running bloopimpl and then comparing the output to a predefined expected result, or something more involved?

Great. Yes, just comparing the generated output looks good to me. We do something similar in `GenIdeaTests`, where we also have some rudimentary support to ignore some parts of the generated files to work around variables parts. Don't know if that is necessary.

megri:
> > I can have a look at it!
> > How do you imagine the test should look like? A simple build definition, running bloopimpl and then comparing the output to a predefined expected result, or something more involved?
> 
> Great. Yes, just comparing the generated output looks good to me. We do something similar in `GenIdeaTests`, where we also have some rudimentary support to ignore some parts of the generated files to work around variables parts. Don't know if that is necessary.

I amended the existing tests in BloopTests.scala. I think I got it working, and also added a check for compile-time dependencies while at it.

I couldn't figure out how to get access to the resolved ivy deps so there's some extra in-place filtering done in the assertions to make sure that all deps except `compileIvyDeps` are present in the jvm platform classpath. I'm fine with cleaning this up if there's a better way of doing things!

megri:
The failing tests seem to be related to something else..

lefou:
> The failing tests seem to be related to something else..

Yeah, I'm already trying to fix them. This PR looks good to me now. Let's rebase/merge on main branch once I've pushed my test changes and if we are green, we can merge.

megri:
> > The failing tests seem to be related to something else..
> 
> Yeah, I'm already trying to fix them. This PR looks good to me now. Let's rebase/merge on main branch once I've pushed my test changes and if we are green, we can merge.

Sounds good!

lefou:
@megri Can you rebase this PR on latest main branch, please?

megri:
> 

Done!

lefou:
Thank you!

PR #1846, opened 2022-04-20
Put modules before libraries in Intellij iml files

yakivy:
PR fixes dependency priority for GenIdea command.

**Problem description:**
When multiple dependencies have a class with the same name, Mill takes first one from the classpath, as well as Intellij takes first one from the iml file. If some dependencies are modules, Mill put them higher in the classpath, so module classes always take precedence over libraries. But GenIdea put modules after the libraries and therefore Intellij resolves different than Mill class versions. Example of generated iml file:
```
...
<orderEntry type="inheritedJdk"/>
<orderEntry type="library" name="..." level="project"/>
...
<orderEntry type="module" module-name="..." exported=""/>
...
```

**Fix description:**
I swapped libraries and modules in the iml file, so result file looks like:
```
...
<orderEntry type="inheritedJdk"/>
<orderEntry type="module" module-name="..." exported=""/>
...
<orderEntry type="library" name="..." level="project"/>
...
```
To prove that this order is correct, I generated iml files for the same project with SBT and Mill BSP, both of them put modules before libraries.

### DIFF
--- a/scalalib/src/GenIdeaImpl.scala
+++ b/scalalib/src/GenIdeaImpl.scala
@@ -818,6 +818,14 @@ case class GenIdeaImpl(
         <orderEntry type="sourceFolder" forTests="false" />
 
         {
+      for (dep <- depNames.sorted)
+        yield dep.scope match {
+          case None => <orderEntry type="module" module-name={dep.value} exported="" />
+          case Some(scope) =>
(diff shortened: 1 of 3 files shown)


lolgab (approved):
LGTM! 👍

lefou (approved):
Thank you!

PR #1849, opened 2022-04-26
Added BSP JVM Extension support

atty303:
Add support for [BSP JVM Extension](https://build-server-protocol.github.io/docs/extensions/jvm.html) methods `buildTarget/jvmTestEnvironment` and `buildTarget/jvmRunEnvironment`. This will correctly set the classpath for execution and testing from IntelliJ IDEA.

Without this addition, IntelliJ IDEA determines the classpath when executing and testing in a BSP Imported project, but the order of precedence when multiple version Jars exist seems to be different than in Mill. Therefore, binary incompatibility error may occur at runtime.

With this addition, IntelliJ IDEA will add the classpath resolved by Mill to the beginning of the classpath, which is the same behavior as Mill.

* https://github.com/JetBrains/intellij-scala/blob/a3f19b5967c87d0e9f3ca49b2278ddf709dd798b/bsp/src/org/jetbrains/bsp/project/test/environment/BspJvmEnvironment.scala#L185-L198
* https://github.com/JetBrains/intellij-scala/blob/a3f19b5967c87d0e9f3ca49b2278ddf709dd798b/bsp/src/org/jetbrains/bsp/project/test/environment/BspJvmEnvironmentProgramPatcher.scala

It will also reflect `forkArgs`, `forkEnv`, and `forkWorkingDir` as defined in the Mill.

### DIFF
--- a/bsp/src/mill/bsp/BSP.scala
+++ b/bsp/src/mill/bsp/BSP.scala
@@ -132,7 +132,7 @@ object BSP extends ExternalModule {
         serverName = serverName,
         logStream = errStream,
         canReload = canReload
-      ) with MillJavaBuildServer with MillScalaBuildServer
+      ) with MillJvmBuildServer with MillJavaBuildServer with MillScalaBuildServer
 
     val executor = Executors.newCachedThreadPool()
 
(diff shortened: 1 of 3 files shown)


lefou (approved):
Nice addition. Looks good to me. Thank you!

PR #1851, opened 2022-04-27
Refactor ScalaJSApi to insulate worker API and User facing API

lolgab:
## Motivation
We want to support the latest Scala.js features without breaking the support for older versions.
In https://github.com/com-lihaoyi/mill/pull/1714 I migrated to the new Scala.js API which generated a wrong filename
(`main.js` instead of `out.js`) in case the build contained multiple modules.
It broke some people tests that were assuming a `out.js` file being created where the real file was a different one.

## Approach
Since workers are internal API, this PR takes a step further in insulating them (using the same approach adopted in mill-mima https://github.com/lolgab/mill-mima/pull/42/files) by duplicating its model classes in two sets:
  - the user facing APIs need to be stable and are serialized to json with upickle
  - the worker facing APIs need to not depend on mill internals and are unstable/internal.
  
Now the worker is responsible to convert between these two APIs.
The user facing API was moved into the main artifact, while the new internal worker-api was moved to a new artifact which is not checked by mima.

## New Features

This PR introduces two new targets that deprecate the main `fastOpt` and `fullOpt`:
`fastLinkJS` and `fullLinkJS`.
These two targets return a `Report` which is an adaptation of the Scala.js API [`Report`](https://www.scala-js.org/api/scalajs-linker-interface/1.10.0/org/scalajs/linker/interface/Report.html)

These two targets are polyfilled for Scala.js 0.6 - 1.2 by generating the new output from the old API output.
This is needed to make `test` work consistently on all versions since it's hard in Mill to run different targets depending on the value of a target like:
```scala
val task = if(scalaJSVersion() > "1.2") fastLinkJSTest else fastOptTest 
```

## Changes
- Introduce `Report` adapted from ScalaJS linker API
- Introduce `fastLinkJS` and `fastOptJS` which return `Report`
- Make test use `fastLinkJSTest`
- Make Scala.js 1.2- and 0.6 always link in legacy mode
- Pass `legacy` boolean parameter to Worker implementation
  to switch between Directory API and File (legacy) API
- deprecate old implementation


lolgab:
Rebased to fix conflicts after #1853 was merged.

build.sc
```diff
@@ -533,7 +533,7 @@ object scalalib extends MillModule {
 
 object scalajslib extends MillModule {
 
-  override def moduleDeps = Seq(scalalib, scalajslib.api)
+  override def moduleDeps = Seq(scalalib, scalajslib.`worker-api`)
```
lolgab, line 536: Moved the code of `api` inside `scalajslib` and created a new internal module called `worker-api`.

build.sc
```diff
@@ -575,13 +575,12 @@ object scalajslib extends MillModule {
 
   override def generatedSources: Target[Seq[PathRef]] = Seq(generatedBuildInfo())
 
-  object api extends MillApiModule {
-    override def moduleDeps = Seq(main.api)
```
lolgab, line 579: Having a completely separated worker module has the side effect that now the Scala.js worker doesn't depend on Mill classes **at all**! :)

scalajslib/src/mill/scalajslib/ScalaJSModule.scala
```diff
@@ -144,11 +178,12 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
       libraries,
       outputPath.toIO,
       mainClass,
+      legacy,
       testBridgeInit,
-      mode == FullOpt,
+      optimize,
```
lolgab, line 183: Since `FullOpt` and `FastOpt` were used only internally I deprecated them now.
We can delete them when we release Mill 0.11

scalajslib/src/mill/scalajslib/ScalaJSModule.scala
```diff
@@ -232,16 +267,32 @@ trait TestScalaJSModule extends ScalaJSModule with TestModule {
...
-      ScalaJSWorkerApi.scalaJSWorker(),
-      scalaJSToolsClasspath(),
-      scalaJSTestDeps() ++ runClasspath(),
-      None,
+    linkJs(
+      worker = ScalaJSWorkerExternalModule.scalaJSWorker(),
+      toolsClasspath = scalaJSToolsClasspath(),
+      runClasspath = scalaJSTestDeps() ++ runClasspath(),
+      mainClass = None,
+      legacy = true,
```
lolgab, line 277: I maintain the old behavior in the legacy methods.

scalajslib/src/mill/scalajslib/ScalaJSModule.scala
```diff
@@ -73,22 +75,31 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
...
-  private def linkTask(mode: OptimizeMode): Task[PathRef] = T.task {
-    link(
-      worker = ScalaJSWorkerApi.scalaJSWorker(),
+  def fastOpt: Target[PathRef] = T {
+    linkTask(optimize = false, legacy = true)().publicModules.head.jsFile
+  }
+
+  def fullOpt: Target[PathRef] = T {
+    linkTask(optimize = true, legacy = true)().publicModules.head.jsFile
+  }
```
lolgab, line 92: Didn't deprecate these two methods yet because I want to see if the new methods cover all people use cases.
I will deprecate them in a future PR.
I made `fastLinkJS` and `fullLinkJS` work with Scala.js 1.2- so we can safely delete them in Mill 0.11 without dropping support for Scala.js 0.6

scalajslib/src/mill/scalajslib/ScalaJSModule.scala
```diff
@@ -232,16 +267,32 @@ trait TestScalaJSModule extends ScalaJSModule with TestModule {
...
-      FastOpt,
-      moduleKind(),
-      esFeatures()
+      optimize = false,
+      moduleKind = moduleKind(),
+      esFeatures = esFeatures()
+    ).map { report => report.publicModules.head.jsFile }
+  }
+
+  def fastLinkJSTest: Target[Report] = T {
```
lolgab, line 285: This is the task tests will use from now on.

scalajslib/src/mill/scalajslib/ScalaJSWorkerApi.scala
```diff
@@ -0,0 +1,72 @@
...
+
+import java.io.File
+
+import mill.api.{Ctx, Result}
+import mill.define.Discover
+import mill.scalajslib.api._
+import mill.{Agg, T}
+
+@deprecated("Use mill.scalajslib.worker.ScalaJSWorker instead", since = "mill 0.10.4")
+class ScalaJSWorker extends AutoCloseable {
```
lolgab, line 11: Deprecated and implemented on behalf of the other `@internal private[scalajslib]` worker I created.
We can delete it on Mill 0.11

scalajslib/src/mill/scalajslib/ScalaJSModule.scala
```diff
@@ -73,22 +75,31 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
...
-      worker = ScalaJSWorkerApi.scalaJSWorker(),
+  def fastOpt: Target[PathRef] = T {
+    linkTask(optimize = false, legacy = true)().publicModules.head.jsFile
+  }
+
+  def fullOpt: Target[PathRef] = T {
+    linkTask(optimize = true, legacy = true)().publicModules.head.jsFile
+  }
+
+  private def linkTask(optimize: Boolean, legacy: Boolean): Task[Report] = T.task {
```
lefou, line 94: Can we have a more descriptive name than `legacy`? E.g. it this is indicating pre-1.2 versions, we might just name the parameter that way.

scalajslib/src/mill/scalajslib/ScalaJSWorkerApi.scala
```diff
@@ -0,0 +1,72 @@
...
+      toolsClasspath = toolsClasspath,
+      config = config,
+      frameworkName = frameworkName,
+      report = report
+    )
+  }
+
+  override def close(): Unit = newWorker.close()
+}
+@deprecated("Use mill.scalajslib.worker.ScalaJSExternalModule instead", since = "mill 0.10.4")
```
lefou, line 67: Could this be a typo?

```suggestion
@deprecated("Use mill.scalajslib.worker.ScalaJSWorkerExternalModule instead", since = "mill 0.10.4")
```

scalajslib/src/mill/scalajslib/ScalaJSWorkerApi.scala
```diff
@@ -0,0 +1,72 @@
...
+import java.io.File
+
+import mill.api.{Ctx, Result}
+import mill.define.Discover
+import mill.scalajslib.api._
+import mill.{Agg, T}
+
+@deprecated("Use mill.scalajslib.worker.ScalaJSWorker instead", since = "mill 0.10.4")
+class ScalaJSWorker extends AutoCloseable {
+  private[scalajslib] val newWorker = new worker.ScalaJSWorker()
```
lefou, line 12: This name is unfortunate.

scalajslib/src/mill/scalajslib/api/ScalaJSApi.scala
```diff
@@ -116,3 +94,44 @@ object JsEnvConfig {
       autoExit: Boolean
   ) extends JsEnvConfig
 }
+
+final class Report private (val publicModules: Iterable[Report.Module]) {
```
lefou, line 98: I know that all API case classes are in this file, but personally, I'd prefer to have a dedicated `Report.scala` file for it.

lefou (commented):
I'm a bit overwhelmed by this PR. I already added some comments, but need a second round to really understand which worker does what, and how the deprecated workers work. 🤪

scalajslib/src/mill/scalajslib/ScalaJSModule.scala
lolgab: Actually legacy means "Use the `PathOutputFile` linking mode instead of `PathOutputDirectory`" which is the one used by the `fastOptJS` (which is the command that Scala.js 1.3 deprecated).

So in Scala.js 1.2- it always links using `PathOutputFile` while in Scala.js 1.3+ the `legacy` flag decides if you want to use the output file or the directory.

scalajslib/src/mill/scalajslib/ScalaJSWorkerApi.scala
lolgab: Yes, it is a typo, thank you! :-)

scalajslib/src/mill/scalajslib/ScalaJSWorkerApi.scala
lolgab: Yeah, I don't like it either. I want to implement the deprecated worker using the methods of the new worker.
Do you have any suggestion?

scalajslib/src/mill/scalajslib/api/ScalaJSApi.scala
lolgab: Good idea! I'm going to move it to a separate file.

scalajslib/src/mill/scalajslib/ScalaJSWorkerApi.scala
lefou: I'll open a dedicated PR to your branch

lefou: * https://github.com/lolgab/mill/pull/1

scalajslib/src/mill/scalajslib/ScalaJSModule.scala
lefou: What about `returnPathIsFile`?

lolgab:
@lefou Merged your PR and tweaked the code a bit more so we don't need to fail in the old code calling the parameterless constructor to not put an extra burden on the users that are updating.

scalajslib/src/mill/scalajslib/ScalaJSModule.scala
lolgab: Since we polyfill the new API we return a `Report` even in legacy mode with the same structure.
I think `legacy` is a good enough name. Also it is used only in private apis so users don't need to know about it.

lefou: To me the name `legacy` is as good as `foo`. The switch clearly does something, otherwise we could just leave it out. Even when this is our internal API, we need to maintain it. If I have issues now, how much fun will have my successor later?

What about `forceOutJs`? We automatically use `out.js` for older Scala.js versions, but what we set this to `true`, we explicitly want to refer to a `out.js`.

scalajslib/src/mill/scalajslib/ScalaJSModule.scala
lefou: If we use `forceOutJs` instead of `legacy`, it would be more obvious what exactly we enable here.

lefou: If we use `forceOutJs` instead of `legacy`, it would be more obvious what exactly we enable here.

lefou (commented):
Looks good to me to merge. I still think, we do ourselves a favor, when we rename the `legacy` parameter to something like `forceOutJs`, to make it more clear (without additional documentation).

lolgab:
legacy is the same name Scala.js uses in its codebase and error messages: https://github.com/scala-js/scala-js/blob/0708917912938714d52be1426364f78a3d1fd269/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Linker.scala#L53
https://github.com/scala-js/scala-js/search?q=legacy
But if you really think we should rename feel free to commit directly on the branch.
It is ready to merge for me as well 🙂🎉

lefou:
> legacy is the same name Scala.js uses in its codebase and error messages: https://github.com/scala-js/scala-js/blob/0708917912938714d52be1426364f78a3d1fd269/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/Linker.scala#L53 https://github.com/scala-js/scala-js/search?q=legacy But if you really think we should rename feel free to commit directly on the branch. It is ready to merge for me as well 🙂🎉

Fair enough. I'll edit and merge, as I think, in the quoted link the word is used with the meaning: deprecated or antiquated.

This is my whole point. The meaning of "legacy" will change over time, and I'm pretty sure we will see other "legacy" changes in Scala.js in the future. That's why I'm nitpicking here.

Thank you for improving Scala.js support!

PR #1858, opened 2022-05-04
Add Junit5 TestModule

sake92:
As per https://github.com/com-lihaoyi/mill/discussions/1411 and https://junit.org/junit5/docs/current/user-guide/#launcher-api-discovery it seems like JUnit5 has its own test discovery mechanism so sbt's fingerprinting is not needed.  
It is indeed skipped in [sbt-jupiter-interface](https://github.com/sbt/sbt-jupiter-interface/blob/master/src/library/src/main/java/net/aichler/jupiter/api/JupiterTestFingerprint.java#L40-L41) so I needed to add a small hack to bypass Mill's internal filtering.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -1192,15 +1192,18 @@ object docs extends Module {
 
   /** Generates the mill documentation with Antora. */
   object antora extends Module {
+    private val npmExe = if (scala.util.Properties.isWin) "npm.cmd" else "npm"
+    private val antoraExe = if (scala.util.Properties.isWin) "antora.cmd" else "antora"
     def npmBase: T[os.Path] = T.persistent { T.dest }
     def prepareAntora(npmDir: os.Path) = {
       Jvm.runSubprocess(
         commandArgs = Seq(
-          "npm",
+          npmExe,
           "install",
           "@antora/cli",
           "@antora/site-generator-default",
-          "gitlab:antora/xref-validator"
+          "gitlab:antora/xref-validator",
+          "@antora/lunr-extension"
         ),
         envArgs = Map(),
         workingDir = npmDir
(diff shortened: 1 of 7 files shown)


lefou (commented):
Interesting. I see the value in it and thing special-casing could be ok here.

Out of pure interest, do you know how sbt is solving this issue? Wouldn't it be the best option to just extend sbt-testing to support alternative ways of detecting test cases?

lefou:
Does this PR by-pass the provided test framework completely, or just for some parts of test discovery. If we by-pass completely, we should add some (new) target which provides the discovered tests and override this to call the Junit5 / jupiter implementation. at least, this is my initial thought about it.

sake92:
> Interesting. I see the value in it and think, special-casing could be ok here.
> 
> Out of pure interest, do you know how sbt is solving this issue? Wouldn't it be the best option to just extend sbt-testing to support alternative ways of detecting test cases?

Not 100% sure, but it seems it bypasess sbt's resolving based on "TestFramework"  
https://github.com/sbt/sbt-jupiter-interface/blob/master/src/plugin/src/main/scala/net/aichler/jupiter/sbt/JupiterPlugin.scala#L79

sake92:
> Does this PR by-pass the provided test framework completely, or just for some parts of test discovery. If we by-pass completely, we should add some (new) target which provides the discovered tests and override this to call the Junit5 / jupiter implementation. at least, this is my initial thought about it.

It just bypasses the sbt fingerprinting mechanism AFAIK, because JUnit5 has its own way of finding test classes/methods.  
So instead of "manually" finding annotated classes and methods it delegates that job to JUnit.

sake92:
Dang, I screwed up my fork's main branch, I'll reopen a PR

PR #1859, opened 2022-05-04
Add search functionality based on @antora/lunr-extension

sake92:
CSS and JS are copied from node_modules, I followed instructions from here https://gitlab.com/antora/antora-lunr-extension  
The index itself is around 600+KB so it doesnt add too much to the site.

This is how it looks like from local:
image

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -1192,15 +1192,18 @@ object docs extends Module {
 
   /** Generates the mill documentation with Antora. */
   object antora extends Module {
+    private val npmExe = if (scala.util.Properties.isWin) "npm.cmd" else "npm"
+    private val antoraExe = if (scala.util.Properties.isWin) "antora.cmd" else "antora"
     def npmBase: T[os.Path] = T.persistent { T.dest }
     def prepareAntora(npmDir: os.Path) = {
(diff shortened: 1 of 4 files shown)


lefou:
Nice. Is that supposed to work with `mill docs.antora.localPages`? I see the search box, but typing anything in it does not reveal any result or result box.

sake92:
@lefou yes, that screenshot is from my local setup. Not sure why it doesnt work for you, any errors in browser console? Maybe try clean build again.

lefou:
Nothing in console. But I noticed there is no reference to any docsearch script in the page source. Probably an issue in the build, not the browser.

lefou:
Could it be my system is picking another version of antora which doesn't not include the `search-scripts.js` by default? I added it manually, and then it worked for me, too.

Shall I open a PR to your branch?

lefou:
Let's merge this as-is. If it not works out-of-the-box, as it does for you, fine. If not, I can push my fixes afterwards.

lefou:
Thank you!

sake92:
You're welcome! Seems it works fine on the live site. 🥳

PR #1862, opened 2022-05-05
Add dest to report and change jsFile to jsFileName

lolgab:
Report doesn't contain all linked js files but only the public
ones. Using that as Target return type breaks the cache invalidation.
This adds the dest folder PathRef as a safer cache invalidation input.
It changes jsFile to jsFileName as in official Scala.js API


lefou (approved):
Looks good to me. Thanks!

PR #1866, opened 2022-05-06
Bump uPickle to 2.0.0 together with Ammonite 2.5.4

lihaoyi:
Second attempt at landing https://github.com/com-lihaoyi/mill/pull/1854. Let's see if keeping the uPickle version consistent between Ammonite and Mill helps

I tried to repro the original problem using the locally-built `target/mill-release` binary, and am able to reproduce the problem (unwanted re-compilation of build.sc) without the Ammonite update, and am unable to reproduce the problem with the Ammonite update. I don't know what the problem is, but I can only conclude that the Ammonite update fixed it.

@lolgab could you help try this out see if it works properly for you too?

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -83,7 +83,7 @@ object Deps {
   }
 
   val acyclic = ivy"com.lihaoyi::acyclic:0.2.1"
-  val ammoniteVersion = "2.5.3-1-b7b4900a"
+  val ammoniteVersion = "2.5.4"
   val ammonite = ivy"com.lihaoyi:::ammonite:${ammoniteVersion}"
   val ammoniteTerminal = ivy"com.lihaoyi::ammonite-terminal:${ammoniteVersion}"
   // Exclude trees here to force the version of we have defined. We use this
--- a/build.sc
+++ b/build.sc
@@ -83,7 +83,7 @@ object Deps {
   }
 
   val acyclic = ivy"com.lihaoyi::acyclic:0.2.1"
-  val ammoniteVersion = "2.5.3"
+  val ammoniteVersion = "2.5.3-1-b7b4900a"
   val ammonite = ivy"com.lihaoyi:::ammonite:${ammoniteVersion}"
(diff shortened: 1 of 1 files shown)


lolgab:
Yes, I can try!
Still, I think it is a good idea to release Mill 0.10.4 before merging this, so we can safely release the bug fixes we have in master without risking any regression.

lefou:
> Yes, I can try! Still, I think it is a good idea to release Mill 0.10.4 before merging this, so we can safely release the bug fixes we have in master without risking any regression.

Mill 0.10.4 is already released.

lolgab:
@lihaoyi Yes updating uPickle and ammonite solves the problem for me as well.
To double-check I tried to downgrade ammonite only the problem appeared again.

lefou:
Nice. I'd like to wait a little bit longer so we can use a proper Ammonite release, tough. In the past, snapshot releases of Ammonite made it hard for me to reason about the changes and the steps of action needed to update.

lolgab:
We need to be sure that plugins compiled with Mill `0.10.0` still work as expected after this change.

lihaoyi:
@lolgab I believe they should. 

This is a format change, but it is backwards compatible and should not break, and regardless Mill doesn't care about compatibility of the format because we flush all caches when classpath changes. 

w.r.t. Binary compatibility, I believe it should be compatible unless you're writing custom `Visitor`s. I assume most Mill plugins (and users!) use it as a black box and do not work with the underlying `Visitor` API directly. So I think it should be safe

lolgab:
@lihaoyi I also believe this should work just fine, but let's manual-test some plugins to check if we don't get weird cache invalidations as we got with the build when there were two different versions of uPickle in the build.

lihaoyi:
@lolgab I don't have any plugins myself, if you have some in mind feel free to test them :)

lihaoyi:
@lolgab any updates here? If not I feel like we can merge this as is.

lolgab (approved):
@lihaoyi Sorry for the big late. I tried it on one project of mine and updating mill without updating any plugins doesn't create any cache invalidation problems.
LGTM! 👍

PR #1869, opened 2022-05-10
Enable ScalaJS optimizer by default

lolgab:
Introduce `scalaJSOptimizer: T[Boolean] = true` target
to make it configurable.
Set the default to `true` since it is what ScalaJS suggests and
uses in the official Sbt plugin


lefou (approved):
Looks good to me. Thank you!

PR #1874, opened 2022-05-15
Add 'init' command for creating giter8 projects

sake92:
This really simple solution works fine:

> PS C:\projects\sake\mill_examples\giter8> C:\projects\sake\mill\target\mill-release.bat -i new scala/scala-seed.g8
> No build file found, you should create build.sc to do something useful
> Compiling C:\projects\sake\mill_examples\giter8\build.sc
> [1/1] new
> Creating a new project...
> A minimal Scala project.
> 
> name [Scala Seed Project]: hello
> 
> Template applied in C:\projects\sake\mill_examples\giter8\hello

Generated:

> PS C:\projects\sake\mill_examples\giter8> tree /F
> Folder PATH listing for volume Windows
> Volume serial number is 6272-A159
> C:.
> ├───hello
> │   │   .gitignore
> │   │   build.sbt
> │   │
> │   ├───project
> │   │       build.properties
> │   │       Dependencies.scala
> │   │
> │   └───src
> │       ├───main
> │       │   └───scala
> │       │       └───example
> │       │               Hello.scala
> │       │
> │       └───test
> │           └───scala
> │               └───example
> │                       HelloSpec.scala
> │
> └───out
>         mill-profile.json
>         new.json
>         new.log

But there are some issues that we might improve:
- it creates a `out/` folder. Maybe we can ignore it for the time being, but it is definitely unwanted.
- it outputs "No build file found, you should create build.sc to do something useful"

### DIFF
--- a/scalalib/src/Giter8Module.scala
+++ b/scalalib/src/Giter8Module.scala
@@ -1,0 +1,37 @@
+package mill
+package scalalib
+
+import scala.util.control.NonFatal
+
+import mill.T
+import mill.define.{Command, Discover, ExternalModule}
+import mill.modules.Jvm
+import mill.define.Task
+
+object Giter8Module extends ExternalModule with CoursierModule {
+
+  override def resolveCoursierDependency: Task[Dep => coursier.Dependency] =
+    T.task { (dep: Dep) =>
+      val scalaVersion = scala.util.Properties.releaseVersion.getOrElse("2.13.8")
+      Lib.depToDependency(dep, scalaVersion)
+    }
+
+  def init(args: String*): Command[Unit] = T.command {
+    T.log.info("Creating a new project...")
+
+    val giter8Dependencies = resolveDeps(
+      T.task {
+        Agg(ivy"org.foundweekends.giter8::giter8:0.14.0")
+      }
+    )()
+
+    Jvm.runLocal(
+      "giter8.Giter8",
+      giter8Dependencies.map(_.path),
+      args
+    )
+  }
+
+  lazy val millDiscover = Discover[this.type]
+
+}
--- a/scalalib/src/giter8/Giter8Module.scala
+++ b/scalalib/src/giter8/Giter8Module.scala
@@ -8,7 +8,11 @@ import mill.scalalib._
 import mill.BuildInfo
 import mill.api.Loose
 
-object Giter8Module extends ExternalModule with CoursierModule {
+object Giter8Module extends ExternalModule with Giter8Module {
+  lazy val millDiscover = Discover[this.type]
+}
+
+trait Giter8Module extends CoursierModule {
 
   def init(args: String*): Command[Unit] = T.command {
     T.log.info("Creating a new project...")
@@ -25,7 +29,4 @@ object Giter8Module extends ExternalModule with CoursierModule {
       mainArgs = args
     )
   }
-
-  lazy val millDiscover = Discover[this.type]
-
 }
--- a/scalalib/src/giter8/Giter8Module.scala
+++ b/scalalib/src/giter8/Giter8Module.scala
@@ -19,10 +19,10 @@ object Giter8Module extends ExternalModule with CoursierModule {
       }
     )()
 
-    Jvm.runLocal(
+    Jvm.runSubprocess(
       "giter8.Giter8",
       giter8Dependencies.map(_.path),
-      args
+      mainArgs = args
     )
   }
 
--- a/scalalib/src/giter8/Giter8Module.scala
+++ b/scalalib/src/giter8/Giter8Module.scala
@@ -1,27 +1,21 @@
-package mill
-package scalalib
-
-import scala.util.control.NonFatal
+package mill.scalalib.giter8
 
 import mill.T
 import mill.define.{Command, Discover, ExternalModule}
 import mill.modules.Jvm
-import mill.define.Task
+import mill.scalalib.api.ZincWorkerUtil
+import mill.scalalib._
+import mill.BuildInfo
+import mill.api.Loose
 
 object Giter8Module extends ExternalModule with CoursierModule {
 
-  override def resolveCoursierDependency: Task[Dep => coursier.Dependency] =
-    T.task { (dep: Dep) =>
-      val scalaVersion = scala.util.Properties.releaseVersion.getOrElse("2.13.8")
-      Lib.depToDependency(dep, scalaVersion)
-    }
-
   def init(args: String*): Command[Unit] = T.command {
     T.log.info("Creating a new project...")
-
     val giter8Dependencies = resolveDeps(
       T.task {
-        Agg(ivy"org.foundweekends.giter8::giter8:0.14.0")
+        val scalaBinVersion = ZincWorkerUtil.scalaBinaryVersion(BuildInfo.scalaVersion)
+        Loose.Agg(ivy"org.foundweekends.giter8:giter8_${scalaBinVersion}:0.14.0")
       }
     )()
 
--- a/scalalib/src/giter8/Giter8Module.scala
+++ b/scalalib/src/giter8/Giter8Module.scala
@@ -26,7 +26,8 @@ trait Giter8Module extends CoursierModule {
     Jvm.runSubprocess(
       "giter8.Giter8",
       giter8Dependencies.map(_.path),
-      mainArgs = args
+      mainArgs = args,
+      workingDir = T.workspace
     )
   }
 }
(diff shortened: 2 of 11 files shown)


lefou:
I know that "new" comes from the same functionality in sbt, but just want to question this choice. I never used it, so I find it hard to imagine whether this is the best name or potential ambiguous. How about "gitter8" or "g8"?

The build-issue in CI are probably related to some transitive dependencies of the gitter8 library. It's IMHO a good idea to load this library in a isolated classloader. Ideally, we only resolve it with coursier when needed, so we don't need to bundle it.

sake92:
Maybe `init` would also work, similar to [Gradle](https://docs.gradle.org/current/userguide/build_init_plugin.html) and [npm](https://docs.npmjs.com/cli/v8/commands/npm-init).  
I don't mind either of those.

Can you point me to an implementation that uses isolated classloader?  
I guess I should use `Lib.resolveDependencies`, but that is in the `scalalib` (which depends on `main`) and the command is in the `main` module so I can't.. @lefou ?

scalalib/src/Giter8Module.scala
```diff
@@ -0,0 +1,37 @@
...
+      val scalaVersion = scala.util.Properties.releaseVersion.getOrElse("2.13.8")
+      Lib.depToDependency(dep, scalaVersion)
+    }
+
+  def init(args: String*): Command[Unit] = T.command {
+    T.log.info("Creating a new project...")
+
+    val giter8Dependencies = resolveDeps(
+      T.task {
+        Agg(ivy"org.foundweekends.giter8::giter8:0.14.0")
```
lefou, line 24: When we declare it as Java dependency, we don't need to override `resolveCoursierDependency`.

Also, using `mill.BuildInfo.scalaVersion` is IMHO the better source.

```suggestion
        val scalaBinVersion = ZincWorkerUtil.scalaBinaryVersion(BuildInfo.scalaVersion)
        Agg(ivy"org.foundweekends.giter8:giter8_${scalaBinVersion}:0.14.0")
```

scalalib/src/Giter8Module.scala
```diff
@@ -0,0 +1,37 @@
...
+import mill.modules.Jvm
+import mill.define.Task
+
+object Giter8Module extends ExternalModule with CoursierModule {
+
+  override def resolveCoursierDependency: Task[Dep => coursier.Dependency] =
+    T.task { (dep: Dep) =>
+      val scalaVersion = scala.util.Properties.releaseVersion.getOrElse("2.13.8")
+      Lib.depToDependency(dep, scalaVersion)
+    }
```
lefou, line 17: Should not be necessary, when declaring a Java dependency.

scalalib/src/Giter8Module.scala
```diff
@@ -0,0 +1,37 @@
+package mill
+package scalalib
```
lefou, line 2: I think we should create `Gitter8` in `mill.scalalib.gitter8`.

lefou:
Are there already usable g8 templates for Mill? Should we provide one ourselves? Should there be even a default, if no template is provided? At least, in case a user does not give any arguments or we detect a missing `build.sc`, we may provide a nice message which suggest which g8 template(s) they could use to start with.

About the misleading message abut the missing `build.sc`. I think we can fix this (in another PR) to just show up, when we would otherwise fail with an error (exit code != 0).

sake92:
@lefou I made a hello world example here, just to test it out: https://github.com/sake92/mill-scala-hello.g8 (you can copy it to com-lihaoyi ofc)  
I don't think there should be any default, user's should specify which template they want to use.  

It will be beneficial to Mill docs also, users won't need do download->extract a zip file, rather they can just type `mill -i init sake92/mill-scala-hello.g8` and initialize a new project.  
Another advantage is that you can parametrize these templates with: project name, folders structure or whatever you need.  

---
Side note:   
I thought about updating https://github.com/scala/scala-seed.g8 directly, but we could have potential issues:
- incompatible folders structure (it might be possible with Giter8 [conditionals](http://www.foundweekends.org/giter8/template.html#Conditionals) )
- not supporting some features that sbt does
- not liking the code examples
Maybe some giter8 scaffold projects in the wild can be updated though.

lefou (approved):
Thanks @sake92 for your changes and your input. LGTM!

I agree, we should incorporate this also into the documentation. Having an easy way to reproduce the examples with a single command might benefit the first Mill experience.

lefou:
Oh, do you think you can add a simple test case?

scalalib/test/src/giter8/Giter8Tests.scala
```diff
@@ -0,0 +1,33 @@
...
+  
+  object mainModule extends TestUtil.BaseModule with MainModule
+
+  override def tests: Tests = Tests {
+    test("init") {
+      val eval = new TestEvaluator(mainModule)
+      test("hello world") {
+        val outDir = os.temp.dir(prefix = "mill_giter8_test")
+        val giter8Args = Seq(
+          "sake92/mill-scala-hello.g8",
```
lefou, line 19: I think, for the test we should use a local test template. G8 should support that. My intend it to avoid dependencies to external (mutable) targets and reduce avoidable network traffic. Also, that way we can make the test case as simple or complex as we want, regardless of its usefulness for others.

scalalib/test/src/giter8/Giter8Tests.scala
```diff
@@ -0,0 +1,33 @@
...
+          s"-o=${outDir}",
+          "--name=hello", // skip user interaction
+          "--description=hello_desc" // need to pass all args
+        )
+        val evaluator = new TestEvaluator(mainModule)
+        val res =  evaluator.evaluator.evaluate(Agg(mainModule.init(evaluator.evaluator, giter8Args: _*)))
```
lefou, line 25: As this seems to not work, we should make the unit under test smaller. Using `Giter8Module` directly should work, as there is no indirection involved.

```suggestion
        val res =  evaluator.evaluator.evaluate(Agg(Giter8Module.init(giter8Args: _*)))
```

scalalib/src/giter8/Giter8Module.scala
```diff
@@ -0,0 +1,31 @@
...
+import mill.define.{Command, Discover, ExternalModule}
+import mill.modules.Jvm
+import mill.scalalib.api.ZincWorkerUtil
+import mill.scalalib._
+import mill.BuildInfo
+import mill.api.Loose
+
+object Giter8Module extends ExternalModule with CoursierModule {
+
+  def init(args: String*): Command[Unit] = T.command {
```
lefou, line 13: It would be nice, if we could return the final destination directory. But I'm not sure how easy it is to detect the correct location when the user is passing arbitrary options, so just take it as a suggestion, not a demand.

```suggestion
  def init(args: String*): Command[PathRef] = T.command {
```

scalalib/test/src/giter8/Giter8Tests.scala
sake92: Yep, makes sense.

scalalib/test/src/giter8/Giter8Tests.scala
sake92: Nice, it's easier to test it like this, thx!

scalalib/src/giter8/Giter8Module.scala
sake92: Not sure if parsing the output of command is sufficient? It outputs something like: `Template applied in C:\Users\sake_\AppData\Local\Temp\mill_giter8_test4305961942039948174` in one line.

lefou: Hm, I think it's probably not worth it. Too easy to break in a next release. If we really need this output, I'd rather try to run g8 as library in-jvm and try to get the final path for it's API.

sake92: Ok, I will make try to invoke its internal API, seems simple enough. 👍🏻

lefou: I'm even not sure such an API exists, and the change to run in-JVM isn't trivial either. I don't stop you, but I think it's not worth it until there is need for it.

sake92: Ok, so not that simple.. 😄  
Seems like they have these custom processors that we could hook in: https://github.com/foundweekends/giter8/blob/develop/app/src/main/scala/giter8.scala#L48-L58  
But this needs to be in a separate JAR for us to download it dynamically

sake92: Yeah, agreed. It's not like users will be confused or something. 99.99% of the time they will `init` without any args, in current folder.  
And it's not like we need to cache the outputs or use it in another task..

PR #1876, opened 2022-05-16
BSP: fix mill-build module interpretation in IntelliJ IDEA

atty303:
## Current behaviour
Snag_d9bcecc
Snag_d9cc8cc

* `sub` is interpreted as a package.
* `Source Folders` is `.`.

This creates a problem where IntelliJ may interpret `root.Foo` class as  `sub.foo.src.root.Foo` class.

## Changed behaviour
Snag_d9f2fd9
Snag_da04792

* `sub` is interpreted as a directory.
* `Source Folders` is `.bsp`.

Everything works well. `build.sc` can also be edited correctly.

If IntelliJ supports [per-file source configuration](https://youtrack.jetbrains.com/issue/SCL-19717), this change would not be necessary.

### DIFF
--- a/bsp/src/mill/bsp/MillBuildServer.scala
+++ b/bsp/src/mill/bsp/MillBuildServer.scala
@@ -89,6 +89,7 @@ class MillBuildServer(
   var clientInitialized = false
   var shutdownRequested = false
   var clientWantsSemanticDb = false
+  var clientIsIntelliJ = false
 
   class State(val evaluator: Evaluator) {
     private[this] object internal {
@@ -221,6 +222,11 @@ class MillBuildServer(
       capabilities.setJvmRunEnvironmentProvider(true)
       capabilities.setJvmTestEnvironmentProvider(true)
 
+      request.getDisplayName match {
+        case "IntelliJ-BSP" => clientIsIntelliJ = true
+        case _ => clientIsIntelliJ = false
+      }
+
       request.getData match {
         case d: JsonObject =>
           log.debug(s"extra data: ${d} of type ${d.getClass}")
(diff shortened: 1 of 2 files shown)


bsp/src/mill/bsp/MillBuildServer.scala
```diff
@@ -335,6 +341,15 @@ class MillBuildServer(
         targetIds = sourcesParams.getTargets.asScala.toSeq,
         agg = (items: Seq[SourcesItem]) => new SourcesResult(items.asJava)
       ) {
+        case (id, _: MillBuildTarget) if clientIsIntelliJ =>
```
atty303, line 344: This change may break Metals, so we are limiting it to IntelliJ.

bsp/src/mill/bsp/MillBuildServer.scala
```diff
@@ -335,6 +341,15 @@ class MillBuildServer(
...
+          T.task {
+            val sources = new SourcesItem(
+              id,
+              Seq(sourceItem(evaluator.rootModule.millSourcePath / ".bsp", false)).asJava
```
atty303, line 348: This can be any subdirectory, but we chose the `.bsp` directory, which definitely exists and is set to ignore.

lefou: Instead we could use some generated source directory under `out` which we define in `MillBuildTarget`. 

```scala
// in MillBuildTarget
/** Used in BSP IntelliJ, which can only work with directories */
def dummySources: Sources = T.sources(T.dest)
```

lefou (commented):
Thanks for this PR. Sounds reasonable.

bsp/src/mill/bsp/MillBuildServer.scala
atty303: Review applied.

Issue #1875, opened 2022-05-15
mill.scalalib.GenIdea/idea and mill.contrib.Bloop/install both fail for VirtusLab/scala-cli


mtk:
i'm not sure if this is a bug in mill or in scala-cli's build.sc.  but even if it is a scala-cli build.sc bug, the mill error messages don't provide any useful info.

a few weeks ago, both of these targets began to fail reliably.  i verified this from a brand new empty linux ID with no ~/.config nor ~/.local directory.  i did a "git clean -dfx" and removed any lingering java processes with the help of 'jps' before running either failing command.

i git bisect'ed the problem and found that the first bad commit was https://github.com/VirtusLab/scala-cli/commit/e11410595592ee1787512a515662f552ae586868 and the last good commit was https://github.com/VirtusLab/scala-cli/commit/ced5e0dbbb820fc17d4d30772a3c305573018cfd.

i don't know enough mill to know whether this change is problematic or not but it doesn't seem like it would be.

i'm attaching a log from the invocation of each command.

[fail.log](https://github.com/com-lihaoyi/mill/files/8695995/fail.log)

lefou:
Looks like the `integration.modulesPath` target is throwing an exception. Can you run it from cli and verify it's succeeding? Could it be, that it tries, for some reasons, to return a `PathRef` of the project directory, and while calculating the content hash, it fails because it can't open the unix socket file? 

I wonder whether `PathRef` should be more forgiving when it encounters a socket file?

lefou:
Yeah, I can reproduce with: 

```scala
import mill._

def socketfile = T{
  PathRef(T.workspace)
}
```

Output:

```
± mill socketfile
No mill version specified.
You should provide a version via '.mill-version' file or --mill-version option.
Retrieving latest mill version ...
Using mill version 0.10.4
Compiling /home/lefou/work/opensource/mill/reproduce-socketfile/build.sc
[1/1] socketfile 
1 targets failed
socketfile java.nio.file.FileSystemException: /home/lefou/work/opensource/mill/reproduce-socketfile/out/mill-worker-LthlJvSD+shQfuObr5wuSIlVd0o=-1/mill-ead6604ff439ef261d60771d013bedff-io: Kein passendes Gerät bzw. keine passende Adresse gefunden
    java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:100)
    java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
[...]

PR #1878, opened 2022-05-17
Handle FileSystemException when creating an input stream

lefou:
Fix https://github.com/com-lihaoyi/mill/issues/1875


main/api/src/mill/api/PathRef.scala
```diff
@@ -45,9 +46,21 @@ object PathRef {
...
               updateWithInt(value)
             } else if (jnio.Files.isReadable(path.toNIO)) {
-              val is = os.read.inputStream(path)
-              StreamSupport.stream(is, digestOut)
-              is.close()
+              val is =
+                try Some(os.read.inputStream(path))
+                catch {
+                  case _: jnio.FileSystemException =>
+                    // This is known to happen, when we try to digest a socket file.
```
lolgab, line 53: Maybe a Debug log of the exception would be nice here.
At the same time you probably don't have access to the logger here, right?

lefou: That was exactly my initial thought. At least I tried to minimize the try-catch to the creation of the input stream, so we don't swallow other errors while reading.

lolgab (approved):

PR #1887, opened 2022-05-26
Add basic docs for 'init' command

sake92:
Currently there is only [one template](https://github.com/sake92/mill-scala-hello.g8). 😄  
But I hope we will soon have more of them.

I made a PR for Play Framework seed https://github.com/playframework/play-scala-seed.g8/pull/164

### DIFF
--- a/docs/antora/modules/ROOT/pages/Intro_to_Mill.adoc
+++ b/docs/antora/modules/ROOT/pages/Intro_to_Mill.adoc
@@ -522,6 +522,25 @@ If you need to feed a `+` as argument to your target, you can mask it by precedi
 
 Mill comes with a few useful command-line utilities built into it:
 
+=== init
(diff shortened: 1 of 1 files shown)


lefou (commented):
Looks good to me. Before merging it, I'd like to create / make sure that the referenced template under com-lihaoyi organization exists. 

I need to learn a bit more about templates for this, as I still have open questions about your proposed template.

ckipp01:
>  I'd like to create / make sure that the referenced template under com-lihaoyi organization exists.

When you do this we can also probably sunset https://github.com/scalameta/mill-scala-seed.g8 which we created so that when users do a `new Scala project` in Metals they got a mill option. We can change this to include the one from here then.

lefou:
Thank you!

PR #1889, opened 2022-05-30
Fix lunr-extension version

sake92:
Seems like newer version of `lunr-extension` broke the search.  
The version is now fixed to `v1.0.0-alpha.6` (at least for now), which we know works properly.


lefou:
Thanks!

PR #1891, opened 2022-06-03
Use ClearableLinker to recover from exceptions in Scala.js

lolgab:


scalajslib/worker/1/src/ScalaJSWorkerImpl.scala
```diff
@@ -103,6 +103,7 @@ class ScalaJSWorkerImpl extends ScalaJSWorkerApi {
         .withSemantics(semantics)
         .withModuleKind(scalaJSModuleKind)
         .withESFeatures(scalaJSESFeatures)
+        .withBatchMode(true)
```
sjrd, line 106: Hum, you probably don't want this. This will force clearing every time, including upon success, discarding any incremental state.

lolgab: Then I misread the docs! I will remove it!

sjrd: Or the docs where not good enough ;)

Anyway, `ClearableLinker` automatically clears and renews if any exception occurs (including linking errors, but also other crashes). If you add batch mode, it *also* clears every time.

Batch mode can be useful in CIs, but shouldn't be used for iterative development.

sjrd (approved):

lolgab:
@lefou CI is green. I stopped the Push checks since I pushed to `origin` by mistake instead of pushing to `lolgab` and didn't want to waste CI resources. All the Pull Request checks passed.
It is ready to merge from my side.

lefou (approved):
Looks good to me. Thank you!

PR #1905, opened 2022-06-23
Fixed ordering-sensitive test

lefou:
We saw CI failures lately in this specific test.

As test cases are picked up from class files which order may depend on the used filesystem and other factors like creation time, inode number, ..., we better always try to avoid any ordering assumptions in tests.

### DIFF
--- a/scalalib/test/src/HelloJavaTests.scala
+++ b/scalalib/test/src/HelloJavaTests.scala
@@ -89,23 +89,16 @@ object HelloJavaTests extends TestSuite {
 
       val Right((v3, _)) = eval.apply(HelloJava.app.testJunit5.test())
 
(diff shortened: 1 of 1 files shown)


IdiosApps:
Hey @lefou , bit of thread necromany here but:

We saw this funny situation before: 
- for a few users, tests would be fine (e.g. on Linux)
- but, for other users' devices (and maybe Jenkins, on a different Linux build I guess) those same tests failed

We fixed it up by improving how we use shared resource in the test suite, or something like this.

Would there be any way to randomise the test order at some levels (tests in a file / files in modules)? 
Given that most tests shouldn't require a specific order to be run in, that could be nice to uncover flakiness earlier on.

PR #1916, opened 2022-07-04
Default to `platformed = true` and introduce `withPlatformed(...)`

lolgab:
## Motivation
Scala.js and Scala Native libraries need to be compiled and published separately from the Scala JVM.
The Scala.js and Scala Native libraries use a specific suffix to distinguish themselves from JVM counterparts, `_sjs1` for Scala.js, where `1` is the binary version of Scala.js, and `_native0.4` for Scala Native.
The official Sbt plugins defined a new syntax to define Scala.js (which then was adopted by Scala Native as well) dependencies:
```scala
"com.github.foo" %%% "bar" % "1.0.0"
```
that the Scala.js plugin converts to:
```scala
"com.github.foo" %% "bar_sjs1" % "1.0.0"
```
which gets converted then by Sbt to:
```scala
"com.github.foo" % "bar_sjs1_2.13" % "1.0.0"
```
given that the `scalaVersion` is `2.13.x`.

Mill adopted a similar convention by creating a second pair of colons:
```scala
ivy"com.github.foo::bar::1.0.0"
```
in a Scala.js module, is equivalent to:
```scala
ivy"com.github.foo::bar_sjs1:1.0.0"
```
and in a Scala Native module to:
```scala
ivy"com.github.foo::bar_native0.4:1.0.0"
```

### What's the problem with it?
The fact you need a special syntax for Scala.js projects is problematic because it can create bugs. If you load a testing library on Scala.js with a JVM artifact, the tests don't run because the Scala.js tooling is not able to find the Scala.js specific files in the dependency jar and assumes it is an empty dependency. Very often people (also very experienced people in Scala.js and Scala Native) get confused and waste time because of this.
Even a *website was created* to raise awareness on this issue: https://youforgotapercentagesignoracolon.com/

### Scala.js/Scala Native dependencies without the platform suffix (`_sjs1`, `_native0.4`)
There are rare cases, [like `scalajs-library`](https://search.maven.org/artifact/org.scala-js/scalajs-library_2.13/1.10.1/jar) from the Scala.js runtime, where a Scala.js library **doesn't have** the `sjs_1` suffix. Also, sometimes it is also useful to have Scala JVM compile-time dependency in a Scala.js project that is used only at compile time by a macro but doesn't end in the runtime classpath.

### Are we optimizing for the most common case?
Even if there are times when dependencies without platform suffix are needed, most of the times what you want is to add the platform suffix (`_sjs1`, `_native0.4`).

### Proposed solution
The proposed solution is to deprecate the `ivy"com.github.foo::bar::1.0.0"` syntax with the second double colon and add the platform suffix by default. To cover the rare cases when you don't want a platformed dependency, this introduces a new method in `Dep` called `withPlatformed` which allows you to remove the platform suffix with:
```scala
`ivy"com.github.foo::bar:1.0.0".withPlatformed(false)`
```
will generate `bar_2.13` as artifact name.

### Proposed behaviour
Input: `ivy"com.github.foo::bar:1.0.0"`

Output:

|        | JVM        | Scala.js        | Scala Native
| ------ | ---------- | --------        | ------------
| Before | `bar_2.13` | `bar_2.13`      | `bar_2.13`
| After  | `bar_2.13` | `bar_sjs1_2.13` | `bar_native0.4_2.13`

Input: `ivy"com.github.foo::bar::1.0.0"` (deprecated after this PR)

Output:

|        | JVM        | Scala.js        | Scala Native
| ------ | ---------- | --------        | ------------
| Before | `bar_2.13` | `bar_sjs1_2.13` | `bar_native0.4_2.13`
| After  | `bar_2.13` | `bar_sjs1_2.13` | `bar_native0.4_2.13`

Input: `ivy"com.github.foo::bar:1.0.0".withPlatformed(false)`

Output:

|       | JVM        | Scala.js   | Scala Native
| ----- | ---------- | ---------- | ------------
| After | `bar_2.13` | `bar_2.13` | `bar_2.13`


scalalib/src/Dep.scala
```diff
@@ -74,6 +74,12 @@ case class Dep(dep: coursier.Dependency, cross: CrossVersion, force: Boolean) {
       case _ =>
         this
     }
+  def withPlatformed(platformed: Boolean): Dep =
```
lolgab, line 77: Here I'm not sure if it's better this or
```suggestion
  def withPlatformed(value: Boolean): Dep =
```

scalalib/src/Dep.scala
```diff
@@ -93,9 +99,9 @@ object Dep {
     (module.split(':') match {
       case Array(a, b, c) => Dep(a, b, c, cross = empty(platformed = false))
```
lolgab, line 100: Didn't set `platformed = true` for java deps since I see them as "raw" it means what you write is what you get.
Also there is no such thing as a platformed java dependency.

lefou:
The idea is good and I fully support it, but I think, before we can this be an simple change, we need properly handle another open issue, transitive module dependencies. This is an open issue in Mill and

lolgab:
@lefou I don't know if the two problems are related.
First of all this change is only syntactical, to make the old `ivy"a::b:c"` become `ivy"a::b:c".withPlatformed(false)` and `ivy"a::b::c"` become `ivy"a::b:c"`.
Second I think that issue relates more to the Scala versioning of dependencies, which helps in Scala 2.13 / Scala 3 sandwiches.
In the case of `platformed`, the story is different and I think that the current behavior is the best we can get with the current model of using Scala `trait`s to share code in the build.
There is a common pattern currently used in Mill which is to define a common trait extending `ScalaModule` and then having different objects of different platforms with the platform-specific code.
Example:
```scala
trait Common extends ScalaModule {
  def ivyDeps = super.ivyDeps() ++ Seq(
    ivy"com.lihaoyi::geny::0.7.1"
  )
}

object js extends Common with ScalaJSModule {
  // ...
}
object jvm extends Common {
  // ...
}
object native extends Common with ScalaNativeModule {
  // ...
}
```

In the code snippet above, the `geny` defined in the `Common` `ivyDeps` is correctly resolved as `geny_3` in `jvm`, `geny_sjs1_3` in `js` and as `geny_native0.4_3`. If we resolve the platform suffix in the `Common` trait we brake the `js` and `native`.

lefou:
Although I said I support the idea, I'm not yet sure, I like this proposal. I'm still in a thinking process. I mean yes, we probably want to optimize for the common case, but we also need to make sure we keep it understandable. I don't like the fact, that the same dependency notation can mean different things in different contexts/modules. We already know, that by using two or three colons, we depend on some context of the module (Scala version, Platform version), but when we use a single colon, we are more or less exact. With your proposal/pr, we loosen this safety, which is IMHO not good.

Sure, the use of the additional colons is also problematic, but it's also the the status quo.

Please note, we use the platform suffix also for Mill plugin, e.g. when importing them into scripts. Once, we deprecate the second platform colon marker, we either need to always write `.withPlatformed(true)` when importing plugins or `withPlatformed(false)` when importing regular libs. There is no clear _more common_ case. 

An alternative could be the introduction of a new dependency declaration like `platformIvyDeps`, which does this conversion you proposed. But this has downsides too (e.g. we probabably also want the dual for `runIvyDeps`). I still find it hard to come up with a satisfying solution. 

Your change is rather fundamental, so I think we really should agree about the best option before making such change. Maybe, we need further discussions. Also, speaking with with other tool developers and user like sbt or Scala-CLI won't hurt.

lolgab:
I'm going through open PRs and re-reading old comments.

> I can safely detect Scala.js and Scala Native modules, but I can't detect pure Scala modules (targeting the JVM). Obviously, each Scala.js module is also a ScalaModule, but in some way it isn't. Maybe, we should introduce a `ScalaBaseModule` or a `ScalaJvmModule`?

It's not the best experience ever, but what I did to solve this particular issue was this:
```scala
module match {
  case _: ScalaJSModule => "js"
  case _: ScalaNativeModule => "native"
  case _: ScalaModule => "jvm"
}
```
It's ugly but it does the job, so we don't need to create a new `ScalaJvmModule` (which I don't like since JVM is the "default" platform for Scala and people using Mill for the first time don't need to know that a Js and a Native platform also exist. `ScalaBaseModule` could work, but I don't see this big need to justify the change.

lolgab:
I'm having second thoughts as well about this change.
What I would do instead is to standardize use:
```scala
// for java deps
ivy"groupId:artifactId:version"
// for scala deps
ivy"groupId::artifactId::version"
```
And leave the other syntaxes to power users.
`ivy"groupid::artifactid:version"` is "needed" only for some specific Scala.js dependencies (specifically when Scala.js uses a JVM artifact) in any other case a platformed dependency works.

Probably after you added `platformSuffix` to `JavaModule` it's also possible to do this for java:
```scala
ivy"groupId:artifactId::version"
```
but this doesn't please my eye 😃

lihaoyi:
I would like to restart discussion here and get it. So far, a lot of the points raised against this proposal are orthogonal and not related to this proposal:

1. How easy it is to distinguish Scala-JVM/JS/Native modules is not affected by which kind of ivyDep is the default

2. Mill plugins are all on Scala-JVM, and are also not going to be affected to what kind of ivyDep is the default for non-JVM Scala modules

This proposal is the right thing to do because:

1. The default thing should be the correct thing. In this case, the correct thing 99.999% of the time is to have a Scala ivyDep that is for the platform your module is

2. You want support for the 0.001% of the time you don't want the default: in this case we have `withPlatformed(false)` as an easy fallback if people don't want the default behavior

3. The change should be backwards compatible: in this case, we can trivially leave the old `foo::bar::qux` syntax in place, and make it equivalent to `foo::bar:qux`, which will work for 99.999% of existing builds

The use case of "I want to depend on Scala-JVM when running on Scala-JS or Scala-Native" is a niche use case and should have niche syntax. The original idea was for things like Scala-Async that were not cross-published for JS, but at this point even that library is cross-published (https://search.maven.org/search?q=a:scala-async_sjs1_2.13). I struggle to come up with a single current example where someone would want a Scala-JVM dependency on Scala-JS and Scala-Native to be the default. In contrast, every single `ivyDep` wants the dep to be of the same Scala platform as the current module.

Given that, this PR is clearly the right thing to do. Any further improvements we want to make to the `ScalaModule` trait hierarchy or other things can be discussed separately

lihaoyi:
@lefou @lolgab I have merged this into latest master and added a unit test to verify that the old syntax continues to work unchanged (for backwards compat)

lefou:
Ok, we definitely need to add this to our Mill 0.11 migration guide, so this info is present to anyone who is slightly interested in what's changed.

@lihaoyi Can you please send a PR with updates to the documentation, which is now outdated, esp. the library dependencies section? Ideally, we keep a `NOTE` block which explains the "old" `::` notation, so users are able to understand and handle found in `build.sc`s or on pages like scaladex.

lihaoyi:
@lefou I'll merge this first and submit a follow up PR with a tentative 0.11.0-M8 changelog and updates the documentation with the user-facing changes so far, including https://github.com/com-lihaoyi/mill/pull/2427. Having those in place first should make it easier to put together a migration guide for all these changes

lefou:
@lihaoyi In the context of imports into Mill build scripts: How do you think we should handle Mill plugins (which target a scala version and a Mill platform, e.g. `group::mill-foo::0.1.0`) vs. normal libraries (which target a scala version but are not Mill specific, e.g. `group::some-lib:1.0.1`). I'd rather have a consistent notation, as these could also be `ivyDeps` in a meta build, which should currently look like this (I'm not sure whether the `RootModule` already propertly sets the correct `platformSuffix`):

```scala
// meta-build

trait FooBuild extends RootModule {
  def ivyDeps = Agg(
    ivy"group::mill-foo::0.1.0",
    ivy"group:some-lib:1.0.1"
  )
}

```

scalalib/src/mill/scalalib/ScalaModule.scala
```diff
@@ -171,7 +171,8 @@ trait ScalaModule extends JavaModule { outer =>
     resolveDeps(
       T.task {
         val bind = bindDependency()
-        Lib.scalaDocIvyDeps(scalaOrganization(), scalaVersion()).map(bind)
+        Lib.scalaDocIvyDeps(scalaOrganization(), scalaVersion()).map(_.withPlatformed(false)).map(bind)
+
```
lolgab, line 175: There is an extra empty line here

lolgab: If `scalaDocIvyDeps` always require `.withPlatformed(false)` we can move that to the definition instead, what do you think? Is it possible?

scalanativelib/src/mill/scalanativelib/ScalaNativeModule.scala
```diff
@@ -114,7 +114,7 @@ trait ScalaNativeModule extends ScalaModule { outer =>
 
   override def scalacPluginIvyDeps: T[Agg[Dep]] = T {
     super.scalacPluginIvyDeps() ++ Agg(
-      ivy"org.scala-native:::nscplugin:${scalaNativeVersion()}"
+      ivy"org.scala-native:::nscplugin:${scalaNativeVersion()}".withPlatformed(false)
```
lolgab, line 117: Compiler plugins usually are JVM dependencies. Are there use cases for platformed dependencies with `:::`?
Are there Scala.js or Scala Native libraries that are published with the full Scala version?

lihaoyi:
@lefou don't mill plugins use `group::mill-foo:$MILL_VERSION`? I don't recall seeing the double colon syntax used in the context of mill plugins before

lolgab:
> @lefou don't mill plugins use `group::mill-foo:$MILL_VERSION`? I don't recall seeing the double colon syntax used in the context of mill plugins before

What you describe are the contrib plugins that are maintained in Mill's monorepo and released together with Mill.
External plugins are released with `_mill0.10` suffix for binary compatibility and @lefou used the `groupId::artifactId::version` syntax that gets lowered to `groupId:artifactId_mill0.10_2.13:version`. It's how all external plugins work these days.
This allows to easily backport changes to older Mill versions (by cross-compiling plugins).

lihaoyi:
We can kee the syntax for importing Mill plugins, or we can change it from `import $ivy` to `import $plugin` to distinguish them, with a corresponding`millPluginIvyDeps` target in the MillBuildRootModule.. I'd say we keep the syntax for now, and if we'd like we can change the magic import in a follow up PR if we want

lefou:
What about some new settings target to control whether "normal" deps should resolve to platformed by default? We could make it default to `true` in `ScalaJSModule` and `ScalaNativeModule`, but in `MillBuildModule` and alike, we probably want default to `false`.

With a reasonable default, we make all common use cases easy, but with the option to change that value (per module), we allow for all possible cases that we allow currently.

I know this is not failure proof (esp. when trying to help in the chats with wrong assumptions), but it would solve the current issue (make default case easy and less error-prone) and keep flexibility at the same time.

lefou:
Best location for application of that setting is probably `CoursierModule.bindDependency`, which is IMHO a better place than the various places where the conversion happens currently in this PR.

scalanativelib/src/mill/scalanativelib/ScalaNativeModule.scala
lihaoyi: I think this is rendered moot by moving the `withPlatformed` call into `bindDependencies`, so a user no longer needs to care about this at all when specifying their `ivy"..."` strings

scalalib/src/mill/scalalib/CoursierModule.scala
```diff
@@ -21,8 +21,12 @@ trait CoursierModule extends mill.Module {
    * Bind a dependency ([[Dep]]) to the actual module contetxt (e.g. the scala version and the platform suffix)
    * @return The [[BoundDep]]
    */
-  def bindDependency: Task[Dep => BoundDep] = T.task { dep: Dep =>
-    BoundDep((resolveCoursierDependency(): @nowarn).apply(dep), dep.force)
+  def bindDependency(platformed: Boolean): Task[Dep => BoundDep] = T.task { dep: Dep =>
```
lefou, line 24: Can you motivate the extra `platformed` flag? I was hoping, we can avoid to decide with every use of `bindDependency` whether we need to enable platformed. What are the use cases?

lihaoyi: @lefou the basic issue is that someone somewhere needs to decide which dependencies within e.g. a `ScalaJSModule` need to be platform-bound, and which don't:

1. Scaladoc, Scala-compiler, Scala-compiler-plugins *should not* be platform bound

2. `compileIvyDeps`, `runIvyDeps` *should* be platform bound

Approaches:

1. The status quo in master is that this distinction is done on a dep-by-dep basis, via `ivy"foo::bar:baz"` vs `ivy"foo::bar::baz"`. This has the downside it is confusing, and allows you to mix platformed/non-platformed dependencies in the same resolved classpath which is basically never what you want

2. My earlier update to this PR preserved that approach, except making `ivy"foo::bar:baz"` do what `ivy"foo::bar::baz"` does  and adding `ivy"foo::bar:baz".withPlatformed(false)` to do what  `ivy"foo::bar:baz"`  used to do. This is less confusing, but still has the weakness of being able to mix platformed and non-platformed dependencies

3. This update in response to your feedback moves the `withPlatformed` logic inside `bindDependency`, but we still need to tell `bindDependency` somehow whether these are platform-specific dependencies or not. Hence the extra flag. Since `bindDependency` is typically called once per classpath (whether compile, run, compiler, scaladoc, etc.), this has the added advantage that it becomes impossible to accidentally mix platformed and non-platformed dependencies

But the bottom line is we need to specify which dependencies are platformed and which are not, which is both necessary and not something that can be inferred from the other data we already have

lihaoyi: Another place we could potentially put the distinction is when we call `resolveDeps`, but AFAICT `resolveDeps` and `bindDependency` are ~always called together anyway, so maybe it's a moot point

lefou: Yeah, I think because there are various differing use site, we initially ended up with the explicit demarcation, which I still think is the best approach. Only the choosen `::` is a bit too insignificant and not easy to spot. 

About the distinction of `bindDependency` and `resolveDeps`, both have distinct use cases and should not be merged.

`bindDependency` is needed to bind a rather loose dependency declaration to the current module, such that it can be used in transitive contexts, e.g. where no Scala version is defined. Hence, its call scope is per-module in contrast to `resolveDeps`, which is called per-classpath.

`resolveDeps` is machinery to resolve and fetch some already bound dependencies (which were defined in potentially different modules). (Eventually but unrelated, we may want to split this even further into resolve and fetch, so we can better simulate lock file semantics and improve cache granularity.)

I think we should keep the platform-decision in the `Dep` class. If we need a third state (set, unset, don't care) we can model it there, e.g. with an `Option[Boolean]`. I think the `bindDependency` task should only depend on other tasks in the same module, not a flag.

lefou: The best place to default-bind the `platform` flag are probably the various `resolvedIvyDeps`, `resolvedRunIvyDeps`, `resolvedScalacPluginIvyDeps` tasks, which all call the `resolveDeps` task. We may need to introduce a dedicated `resolvedScaladocIvyDeps` to honor the fact that these may always need non-platformed dependencies (in the JS, Native cases, at least).

lefou (commented):
@lihaoyi @lolgab 
From this discussion I think, a picture of the "best solution" is still very vague and we should not rush it. We should postpone this to the next milestone and get Mill `0.11.0-M8` out the door to re-bootstrap our build and start to experiment with it in real projects. WDYT?

lihaoyi:
@lefou sure, happy to postpone this till after 0.11.0-M8

lolgab:
Two things that might influence the discussion:
- Sbt 2.0 is [designing a solution for this same problem now](https://eed3si9n.com/sbt-2.0-ideas). It's worth checking what they are doing, even though they have different constraints since the Scala Native and Scala.js plugins live in different repositories and are maintained by different organizations.
- This problem is particularly annoying with Scala.js and Scala Native because if you forget the second `::` it doesn't fail to resolve the artifact since most libs have JVM counterparts. With Mill plugins, it fails to resolve since there is a `mill-mima_mill0.10_2.13` but there isn't any `mill-mima_2.13`. It's still problematic but you have a failure instead of a success with silent empty test suites passing (in case of test frameworks).

lefou:
I think if we know in some special task, that we only want dependencies of a special kind, e.g. platform only or non-platform only, we should check early and just fail. With a nice error messages, users would be even ok with adding that extra colon. As @lolgab said, it's mostly about awareness and failure reporting.

lihaoyi:
I don't think I can accept the "we provide two syntax options, one is always wrong, and we know which one and provide an error message if you give the wrong syntax" approach. It just seems so silly, especially if we can just provide one syntax option and it "always" does the right thing.

It definitely seems like that is possible to automate all of this, even if we have to scatter some constants throughout the core Mill libraries. It's vastly better to have e.g. ~20 decision points in Mill library code rather than to having ~20,000 decision points (one for every single `ivy` expression)  scattered throughout hundreds of different user build.sc files

scalalib/src/mill/scalalib/CoursierModule.scala
lihaoyi: @lefou going through the code, it definitely seems like `bindDependency` is the right place to do this. After all, `bindDependency`'s doc says

> Bind a dependency ([[Dep]]) to the actual module contetxt (e.g. the scala version and the platform suffix)
>   * @return The [[BoundDep]]

With `BoundDep`

> Same as [[Dep]] but with already bound cross and platform settings.

Because `BoundDep` already has all artifact name suffixes already "flattened out" and appended to the artifact name, it is impossible to robustly take an existing `BoundDep` and add the platform suffix: we would have to e.g. take `scala-async_2.13`, split it into `scala-async` and `2.13`, and then splice in `scala-sjs1` in the middle to make it `scala-async_sjs1_2.13`. Maybe possible to do hackily for the suffices we use now, but would be very fragile e.g. if someone's platform suffix or artifact name has a `_` everything will break.

As is, the current state of this PR already basically performs the `platformed` decision everywhere close to we call `resolveDeps`, It's just that the usage of `resolveDeps` is a bit more heterogenous than the small number of tasks you listed above. We can refactor the call graph later if we want, but fundamentally it does seem that performing the `platformed=true` or `platformed=false` decision does belong inside `bindDependency` and nowhere else

lefou:
> I don't think I can accept the "we provide two syntax options, one is always wrong, and we know which one and provide an error message if you give the wrong syntax" approach. It just seems so silly, especially if we can just provide one syntax option and it "always" does the right thing.
> 
> It definitely seems like that is possible to automate all of this, even if we have to scatter some constants throughout the core Mill libraries. It's vastly better to have e.g. ~20 decision points in Mill library code rather than to having ~20,000 decision points (one for every single `ivy` expression) scattered throughout hundreds of different user build.sc files

I see it likewise. But I wanted to point out, we should move some decision points into the respective special modules rather that the generic implementations. Esp. as not all platform-specific modules behave equal (Native vs. Mill). 

We should start with asserting at those positions if something illegal is provided. 

If we can provide nice default behavior, we definitely should. But it must be understandable from the outside. And sometimes, less magic in favor of more explicitness is the right thing to do. A clear eror can be more that some magic implicit behavior hard to reason about. TBH, I never saw the need to have a JVM dependency in a native/js build, so I don't know whether we need that support at all (does that mean we can default to use platformed ivyDeps?, probably yes.). I know, in a Mill plugin module, we do have both (we should default to non-platformed ivyDeps).

lihaoyi:
OK, I think there isn't any more disagreement. The PR as written already satisfies everything brought up so far:

1. `MillBuildModule` has non-platformed deps as a default
2. Everyone else has platformed-deps as a default
3. Platform/no-platform decision is made as close to `resolveDeps` as possible given the current code infrastructure

I'm working on getting CI green, then we can probably merge this

lefou:
> OK, I think there isn't any more disagreement. The PR as written already satisfies everything brought up so far:
> 
>     1. `MillBuildModule` has non-platformed deps as a default
> 
>     2. Everyone else has platformed-deps as a default
> 
>     3. Platform/no-platform decision is made as close to `resolveDeps` as possible given the current code infrastructure
> 
> 
> I'm working on getting CI green, then we can probably merge this

I still strive towards a `bindDependency` without flags.

lihaoyi:
The basic logic is that scala compiler classpaths always have to be non-platformed, runtime code always has to be platformed, with special cases for special libraries.

This is actually exactly the same way the cross-scala-version suffixes like `_2.13` works: `scala-compiler` is non-platformed, most libraries are platformed, `scala-library` is a special library and non-platformed.

The vast majority of macros have to be platformed, unless the macro has zero runtime support. This used to be the case with Scala-Async, for example. But even that is now cross-published (as I have linked above).

That's about it. In the end, the goal of this is not to change any behavior - we just replicate the behavior we already expect people to follow, except automatically.

lefou:
We should probably drop the `resolveCoursierDependency` compatibility shim (which is called from bindDependency) at this point, as it makes understanding override strategies unnecessary hard.

lihaoyi:
> I still strive towards a bindDependency without flags.

There isn't such an option unless we remove `bindDependency` from the code path transforming `Deps` to `BoundDeps`. We're effectively already passing in flags to `bindDependency` anywhere, except the flag is called `::` and it's scattered throughout all sorts of user build files and targets

lefou:
`bindDependency` is essentially injection platform bindings when requested. 

Why scatter the platform decision to all `bindDependency` call sites, if it is only needed in some places and could be perfectly done by adapting the `Dep]` we pass to `bindDependency`?

To my understanding, we can derive the decision whether to platform or not, from the location of a dependency. If it comes from `ivyDeps` or `runIvyDeps`, then they need to be platformed.

lihaoyi:
> Why scatter the platform decision to all `bindDependency` call sites, if it is only needed in some places and could be perfectly done by adapting the `Dep` we pass to `bindDependency`?

That's exactly what this PR did in https://github.com/com-lihaoyi/mill/pull/1916/commits/6a1af1264134ddabda51cf8b4c0431f5f7e9a96b. You requested to move the logic into `bindDependency`, so I did. 

If you'd prefer the call to be made on the arguments that we pass to `bindDependency` instead of within `bindDependency` itself controlled by a flag, I can revert this PR to 6a1af1264134ddabda51cf8b4c0431f5f7e9a96b and proceed from there

lihaoyi:
Ok, will revert

PR #1931, opened 2022-07-08
Make docker builds able to COPY custom files in

joan38:
I've been monkey patching this for awhile in an internal plugin to be able to `COPY` custom files in the `Dockerfile` since we have `final def build`.
And I think today I found a potential permanent solution to this.

The fact that `docker build` does not allow absolute paths outside of the "context" is annoying since Mill is all about absolute path with `T.dest`. The compromise here is that the user can put any absolute path in the `dockerfile` but mill-docker will perform a `replace(os.pwd.toString, "")` and we'll execute the "context" in `os.pwd` which refers to the root of the project.

Thanks


contrib/docker/src/DockerModule.scala
```diff
@@ -136,11 +130,11 @@ trait DockerModule { outer: JavaModule =>
       val pullLatestBase = IterableShellable(if (pull) Some("--pull") else None)
 
       val result = os
-        .proc(executable(), "build", tagArgs, pullLatestBase, dest)
+        .proc(executable(), "build", tagArgs, pullLatestBase, "-f", dockerfilePath, os.pwd)
```
joan38, line 133: The "context" path is now `os.pwd` and we specify the Dockerfile path with `-f`

contrib/docker/src/DockerModule.scala
```diff
@@ -113,20 +112,15 @@ trait DockerModule { outer: JavaModule =>
...
 
     final def build = T {
-      val dest = T.dest
-
-      val asmPath = outer.assembly().path
-      os.copy(asmPath, dest / asmPath.last)
-
-      os.write(dest / "Dockerfile", dockerfile())
+      val dockerfilePath = T.dest / "Dockerfile"
+      os.write(dockerfilePath, dockerfile().replace(os.pwd.toString, ""))
```
joan38, line 123: We have to scale down all mentions of absolute paths to the project root.

contrib/docker/src/DockerModule.scala
```diff
@@ -113,20 +112,15 @@ trait DockerModule { outer: JavaModule =>
         if (user().isEmpty) "" else s"USER ${user()}"
       ).filter(_.nonEmpty).mkString(sys.props("line.separator"))
 
-      s"""
-         |FROM ${baseImage()}
+      s"""FROM ${baseImage()}
          |$lines
-         |COPY $jarName /$jarName
-         |ENTRYPOINT ["java", "-jar", "/$jarName"]""".stripMargin
+         |COPY ${assembly().path} /${assembly().path.last}
```
joan38, line 117: We can now put any absolute paths in the `dockerfile`.

contrib/docker/src/DockerModule.scala
lefou: `T.workspace` is a special accessor to the project root directory.

```suggestion
      os.write(dockerfilePath, dockerfile().replace(T.workspace.toString, ""))
```

contrib/docker/src/DockerModule.scala
lefou: Same here.

```suggestion
        .proc(executable(), "build", tagArgs, pullLatestBase, "-f", dockerfilePath, T.workspace)
```

lefou (commented):
Nice. Can you also add a note to the plugin documentation about the path conversion?

lefou:
> If you have a better idea without too much user confusion, let me know.

I can't tell whether this idea is better, but we could also introduce a `dockerResources` target or `dockerContext`  which could provide additional resources. This would result in extra copying operations, but OTOH, we had a more local and controlled context.

lefou:
> Not sure either because you have to make the `COPY`s relative to that `dockerResources`. The SBT Native Packager's docker plugin uses this `mappings` Map which is similar to this idea but gives also control on the path where to put those files.

Given, that you could implement a `dockerResources` target however you want (just return a dir or  compute, copy or download various files and return those), this is probably much in the spirit of `mapping`, but a bit more flexible.

I was assuming, that we need to merge multiple resource dirs before we run `docker build` on it, that's why I said "extra copying".

> Are the tests actually failed? Do I need to rerun something?

Yes. Multiple docker tests failed.

```
contrib.docker.test.test 5 tests failed: 
  mill.contrib.docker.DockerModuleTest mill.contrib.docker.DockerModuleTest.docker build.default options
  mill.contrib.docker.DockerModuleTest mill.contrib.docker.DockerModuleTest.docker build.all options
  mill.contrib.docker.DockerModuleTest mill.contrib.docker.DockerModuleTest.dockerfile contents.default options
  mill.contrib.docker.DockerModuleTest mill.contrib.docker.DockerModuleTest.dockerfile contents.all options
  mill.contrib.docker.DockerModuleTest mill.contrib.docker.DockerModuleTest.#utestAfterAll
```

For example:

```
STEP 10/12: USER user1
--> d7170a4c97b
STEP 11/12: COPY /home/runner/work/mill/mill/target/workspace/mill/contrib/docker/DockerModuleTest/eval/docker build/all options/assembly.dest/out.jar /out.jar
Error: error building at STEP "COPY /home/runner/work/mill/mill/target/workspace/mill/contrib/docker/DockerModuleTest/eval/docker build/all options/assembly.dest/out.jar /out.jar": checking on sources under "/home/runner/work/mill/mill/target/worksources/mill/contrib/docker/DockerModuleTest/Docker/millSourcePath": copier: stat: "/home/runner/work/mill/mill/target/workspace/mill/contrib/docker/DockerModuleTest/eval/docker": no such file or directory
```

See https://github.com/com-lihaoyi/mill/runs/7248330085?check_suite_focus=true

joan38:
I solved the issue with tests.
@lefou can you have a last look?

lefou:
> I solved the issue with tests. @lefou can you have a last look?

Sure, will probably take some time though. Also, your constant rebasing makes tracking of changes (and remembering where I left the review) really hard and frustrating.

lefou (commented):
Can you elaborate why you changed `TestEvaluator` and `TaskTests`? Those looks unrelated to a change in contrib.docker, at first sight.

main/test/src/util/TestEvaluator.scala
```diff
@@ -29,10 +29,9 @@ class TestEvaluator(
     threads: Option[Int] = Some(1),
     outStream: PrintStream = System.out,
     inStream: InputStream = DummyInputStream,
-    debugEnabled: Boolean = false,
-    extraPathEnd: Seq[String] = Seq.empty
+    debugEnabled: Boolean = false
 )(implicit fullName: sourcecode.FullName, tp: TestPath) {
-  val outPath = TestUtil.getOutPath() / extraPathEnd
```
joan38, line 35: @lefou I do not see any point having this `extraPathEnd`. I added `out` to simulate the `out` we usually have, followed by the test path instead of the cmd path, to avoid the test conflicts.

joan38: I actually rolled back the `extraPathEnd` since it's actually useful.
About using `module.millSourcePath` instead of `TestUtil.getOutPath()`, it's because we need to have the `outPath` inside the source path like it is in "production".

joan38:
Should be good now.

main/testkit/src/testkit/MillTestkit.scala
```diff
@@ -72,7 +72,7 @@ trait MillTestKit {
       debugEnabled: Boolean = false,
       extraPathEnd: Seq[String] = Seq.empty
   )(implicit fullName: sourcecode.FullName) {
-    val outPath = getOutPath(testPath) / extraPathEnd
+    val outPath = module.millSourcePath / "out" / testPath / extraPathEnd
```
lefou, line 75: Why you replaced the `getOutPath` call with a rather hard-coded path?

contrib/scoverage/test/src/HelloWorldTests.scala
```diff
@@ -115,7 +115,7 @@ trait HelloWorldTests extends utest.TestSuite {
 
             assert(
               result.path.toIO.getPath.replace("""\""", "/").endsWith(
-                "mill/target/workspace/mill/contrib/scoverage/HelloWorldTests/eval/HelloWorld/core/scoverage/data/core/scoverage/data.dest"
+                "mill/target/worksources/mill/contrib/scoverage/HelloWorldTests#HelloWorld/out/HelloWorld/core/scoverage/data/core/scoverage/data.dest"
```
lefou, line 118: This looks a bit fishy to me. IIRC, we use `worksources` only for sources (e.g. something to check out only once), but `workspace` to modify as needed. If we now modify `worksources`, this might invalidate various assumptions in other tests and also inpact re-running tests.

main/test/src/util/TestEvaluator.scala
lefou: In other places, we copied all required sources into the workspace before the test. That way, the tests are stay reproducible.

main/testkit/src/testkit/MillTestkit.scala
joan38: So the issue here is that in real life (not in tests) we have the `millSourcePath` pointing to the top project folder and then we have a subfolder `out`.
So in real life, the outPath (or workspace) is nested into the worksources.
Tests need to be the same and my change is making it closer to real life.

contrib/scoverage/test/src/HelloWorldTests.scala
joan38: See my comment above since those 2 are tightly coupled.

main/testkit/src/testkit/MillTestkit.scala
lefou: I'm not sold by this argumentation. There _is_ the option of overriding `def getOutPath` to customize it to your needs. If there is a indirection, there is a potential need for it. Also, in other tests we create "real life" test setups by copying the `worksources` into the `workspace`. to the contrary, in tests we need more control over the world under test, so we can actually separate preconditions, assumptions and results.

joan38: Outside of tests, this PR relies on doing a docker build in `T.workspace` assuming all files we are pointing to in the Dockerfile are in `T.workspace`. Now this does not work if out path is not a sub folder of `T.workspace`. Are we in agreement that out path needs to be a sub folder of `T.workspace`?

lefou: > Outside of tests, this PR relies on doing a docker build in `T.workspace` assuming all files we are pointing to in the Dockerfile are in `T.workspace`. Now this does not work if out path is not a sub folder of `T.workspace`. Are we in agreement that out path needs to be a sub folder of `T.workspace`?

I think, this is an assumption that may not hold `true` in all conditions. IMHO, we don't want this hard requirement. Also, I can't exactly say, how this behaves for foreign modules. (mill-docker  will probably fail because either `T.workspace` is no longer the dir containing `build.sc` or the `out` location changes.)

Maybe, using `T.workspace` is not the right thing then. You could introduce a `dockerRoot` target, which can be customized by the user. The default could be just `millSourcePath`.

After all, I think the merge of this PR was premature and changing core Mill to get a change in a contrib package through without final approval is an unwanted behavior. At least, I don't like it.

joan38: I'm reverting this and creating a new PR with that suggestion

PR #1938, opened 2022-07-11
Improve Scala.js module config docs

carlosedp:
Improve Scala.js module config docs adding test and dependency descriptions.

### DIFF
--- a/docs/antora/modules/ROOT/pages/Common_Project_Layouts.adoc
+++ b/docs/antora/modules/ROOT/pages/Common_Project_Layouts.adoc
@@ -94,8 +94,9 @@ import mill.scalajslib._
 object foo extends ScalaJSModule {
   def scalaVersion = "2.13.8"
   def scalaJSVersion = "1.10.1"
+  // Make the DOM available by using the jsdom library (as in Scala.js tutorial)
   def jsEnvConfig = T(scalajslib.api.JsEnvConfig.JsDom())
-  ivy"org.scala-js::scalajs-dom::2.2.0"
+  def ivyDeps = Agg(ivy"org.scala-js::scalajs-dom::2.2.0")
   
   object test extends Tests with TestModule.Utest {
     def ivyDeps = Agg(ivy"com.lihaoyi::utest::0.8.0")
--- a/docs/antora/modules/ROOT/pages/Common_Project_Layouts.adoc
+++ b/docs/antora/modules/ROOT/pages/Common_Project_Layouts.adoc
@@ -110,9 +110,9 @@ latter of which runs your code on Node.js, which must be pre-installed)
 generating the optimized Javascript file.
 
 ScalaJSModule requires declaring it's dependencies as third party
-libraries (note the two sets of double-colons `::` necessary between the library name and the version).
+libraries (don't forget to use double-colons `::` between the library name and the version for Scala.js dependencies).
 
-If you are following the [Scala.js Tutorial](http://www.scala-js.org/doc/tutorial/basic/), in the section where you test the code, it's mentioned that you need to make the DOM available to Node.js. Install the `jsdom` node library as instructed then add `def jsEnvConfig = T(JsEnvConfig.JsDom())` to your `build.sc` object. In Scala 3, you don't need the "scalajs-env-jsdom-nodejs" mentioned in the tutorial.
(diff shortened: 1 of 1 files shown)


docs/antora/modules/ROOT/pages/Common_Project_Layouts.adoc
```diff
@@ -92,8 +92,15 @@ import mill._
 import mill.scalajslib._
 
 object foo extends ScalaJSModule {
-  def scalaVersion = "2.12.4"
-  def scalaJSVersion = "0.6.22"
+  def scalaVersion = "2.13.8"
+  def scalaJSVersion = "1.10.1"
+  // Make the DOM available by using the jsdom library (as in Scala.js tutorial)
+  def jsEnvConfig = T(scalajslib.api.JsEnvConfig.JsDom())
+  def ivyDeps = Agg(ivy"org.scala-js::scalajs-dom::2.2.0")
```
lolgab, line 99: I would first `import mill.scalajslib.api._` at the beginning of the file and then write `def jsEnvConfig = T(JsEnvConfig.JsDom())`
What do you think?

At the same time, `jsdom` needs to be installed with npm and has different complications, so I don't know if it should be referenced in the first Scala.js example.

carlosedp: Humm, I agree on npm dependency but I thought about adding it since it's not documented anywhere making a newcomer life harder if starting with the Scala.js tutorial and using Mill...

What do you think? Remove it or do something like linking to the tutorial?

docs/antora/modules/ROOT/pages/Common_Project_Layouts.adoc
```diff
@@ -103,6 +109,11 @@ latter of which runs your code on Node.js, which must be pre-installed)
 `ScalaJSModule` also exposes the `foo.fastOpt` and `foo.fullOpt` tasks for
 generating the optimized Javascript file.
 
+ScalaJSModule requires declaring it's dependencies as third party
+libraries (note the two sets of double-colons `::` necessary between the library name and the version).
+
+If you are following the [Scala.js Tutorial](http://www.scala-js.org/doc/tutorial/basic/), in the section where you test the code, it's mentioned that you need to make the DOM available to Node.js. Install the `jsdom` node library as instructed then add `def jsEnvConfig = T(JsEnvConfig.JsDom())` to your `build.sc` object. In Scala 3, you don't need the "scalajs-env-jsdom-nodejs" mentioned in the tutorial.
```
lefou, line 115: This should be an Asciidoc link and https.

```suggestion
If you are following the https://www.scala-js.org/doc/tutorial/basic/[Scala.js Tutorial], in the section where you test the code, it's mentioned that you need to make the DOM available to Node.js. Install the `jsdom` node library as instructed then add `def jsEnvConfig = T(JsEnvConfig.JsDom())` to your `build.sc` object. In Scala 3, you don't need the "scalajs-env-jsdom-nodejs" mentioned in the tutorial.
```

docs/antora/modules/ROOT/pages/Common_Project_Layouts.adoc
```diff
@@ -103,6 +109,11 @@ latter of which runs your code on Node.js, which must be pre-installed)
 `ScalaJSModule` also exposes the `foo.fastOpt` and `foo.fullOpt` tasks for
 generating the optimized Javascript file.
 
+ScalaJSModule requires declaring it's dependencies as third party
+libraries (note the two sets of double-colons `::` necessary between the library name and the version).
```
lefou, line 113: Might look as nitpicking, but the wording  made me understand that double colons are needed for "thirdparty libraries" whereas they are needed for "Scala.js libraries" in general.

```suggestion
libraries (don't forget to use double-colons `::` between the library name and the version for Scala.js dependencies).
```

docs/antora/modules/ROOT/pages/Common_Project_Layouts.adoc
lefou: Can you add, why this is not needed in Scala 3? Is it already included or no longer necessary?

lefou (commented):
I like the additional information presented as text.

PR #1941, opened 2022-07-12
CI: skip building dependabot branches, we already build PRs

lefou:


IdiosApps:
If we skip the `test` job for dependabot on branch merge, should we also skip ` test-bin-compat` & `test-windows` for the same situation/reason?

.github/workflows/actions.yml
```diff
@@ -7,6 +7,9 @@ on:
 jobs:
 
   test:
+    # don't build dependabot branches on push, we already build them on pull request
+    if: github.event != 'push' || !startsWith(github.ref, 'refs/heads/dependabot')
+
```
IdiosApps, line 12:
```suggestion
    # We already run tests on Dependabot PRs, and we don't want to test again on push
    if: !(startsWith(github.ref, 'refs/heads/dependabot') && github.event == 'push')
```
I think the logic you tried first is a bit easier to understand. I changed the order of conditions to match a rewritten comment.
Either way is fine though. It would be nice if Github Actions had `if: condition skip`, so conditions could be inverted for simplicity 🤞

IdiosApps: > I made this change, because GitHub did not understand the first version.

Oh right, that's why there's no ✅ / ❌ next to the shorthash. Interesting 🤔

IdiosApps: @lefou https://github.com/github/docs/issues/3001

I confirmed we can do `${{!(conditions)}}`. It's not the prettiest but might make the logic easier to reason about, so could be worth it.
Here's my test example, which I saw print in an Action run (in a private repo)

```  
condition3_shouldEcho:
    if: ${{!(1==2)}}
    name: c3
    runs-on: ubuntu-latest
    steps:
      - run: echo c3
```

lefou: Interesting. Does it also works for `${{!(1==2 || 3==4)}}` ?

lefou: I tried to change it, but to be honest, it really doesn't look easier to read. I'd like to keep it as is.

IdiosApps: {{no worries}} 😁

PR #1947, opened 2022-07-14
Experimental dependency management, force transitive versions

lefou:


lefou (commented):
Although we don't propagate these bumps to downstream users, we immediately profit from this change as we no longer bundle to old or known to be vulnerable dependencies. Hence, I'll merge as is.

PR #1953, opened 2022-07-17
update: flyway-core 8.0.2 -> 8.5.13, requires NOTICE log level

IdiosApps:
This was one of the version updates scala steward has missed for a while.

I guess this is because scala steward couldn't add the required NOTICE log code change.


lefou (approved):
Looks good to me. Thanks!

PR #1959, opened 2022-07-18
Twirllib: new twirlScalaVersion target to support for newer versions

IdiosApps:
- https://github.com/playframework/twirl/releases 1.5.1 is latest
- updated docs to latest version


IdiosApps:
I need to update my fork with changes from main. Will do soon 🥵

contrib/twirllib/src/TwirlModule.scala
```diff
@@ -25,10 +25,10 @@ trait TwirlModule extends mill.Module {
         coursier.LocalRepositories.ivy2Local,
         MavenRepository("https://repo1.maven.org/maven2")
       ),
-      Lib.depToDependency(_, "2.12.5"),
+      Lib.depToDependency(_, "2.13.8"),
```
lefou, line 28: Is the twirl compiler now on Scala 2.13?

And even if so, we may want to keep support of older versions, at least in Mill 0.10.x

Alternatively, we may state some (non-)compatibility notes in the plugin documentation page and the release notes.

IdiosApps: 2.13 support since twirl 1.4.2 https://mvnrepository.com/artifact/com.typesafe.play/twirl-compiler

Does this line of code mean it's either 2.12 or 2.13 support? I don't know much about the whole "binary incompatibility" situation. Is it a lot more complex to support older versions too?

lefou: In this specific use case, it means, all Scala ivy deps (those with a double colon before the name) will resolve to the given Scala version. You can't mix them in the same classpath, so yes, it's either 2.12 or 2.13.

We can support both versions and we should, but it may not work with this hardcoded version here.

lefou (commented):
I think we should make sure, that we have tests for older and never twirl versions with Scala 2.13 and 2.12. I've no idea how widely this is used, though.

IdiosApps:
> I think we should make sure, that we have tests for older and never twirl versions with Scala 2.13 and 2.12. I've no idea how widely this is used, though.

That makes sense. I had never heard of twirl before looking into this. 
Like with the proguard updates PR, maybe it's best to leave these updates to people who need/want them - I have to direct energy elsewhere for a while.

IdiosApps:
@lefou thank you for finishing this one off :)

PR #1963, opened 2022-07-19
docs: Add example of overriding moduleKind

nightscape:


lefou:
Seems useful. Thank you!

PR #1966, opened 2022-07-20
Use SoftReference instead of WeakReference for caches

lolgab:
`SoftReference`s are [better suited to implement memory-sensitive caches](https://docs.oracle.com/javase/8/docs/api/java/lang/ref/SoftReference.html).

Suggested in a code review in bloop: https://github.com/scalacenter/bloop/pull/1761#discussion_r925512643


lefou:
Thank you!

PR #1968, opened 2022-07-22
Pass clientServer parameter to the IntegrationTestSuite constructor

hiloboy0119:
I've been having problems with my .mill-jvm-opts file not being applied while in client-server mode.  I was doing some testing and realized the integration tests were always using `--no-server`.

I found that the clientServer parameter was not being passed down.  This is a small one line PR to fix that.

Regarding my original issue, it doesn't seem to happen as of c81ac64ee91b8d85074ff373df1c01f8bd91b06e


lefou (approved):
Nice catch. Looks good to me. Thank you!

PR #1981, opened 2022-07-31
bloop: use the same naming scheme for adding to and accessing moduleSourceMap

ajrnz:
When the bloop module is computing the sources references for the project it uses the `moduleSourceMap` target. This puts paths into the map using `m.millModuleSegments.render` however when accessing the map to build the bloop configuration it uses `name(module)`. These differ for foreign modules resulting in the sources for foreign modules being omitted.


lefou (approved):
Looks good to me. Thank you!

PR #1988, opened 2022-08-12
feat: allow for a non-local Java Compiler to be used

### LINKED ISSUE #1983: [feature request] Add support for setting `javaHome` in `javaModule`
So a bit of back story first. I'm currently working with the [semanticdb-java](https://sourcegraph.github.io/scip-java/docs/manual-configuration.html) compiler plugin to enable Java sources to emit semanticdb. One of the issues I hit on is that on jdk 17 there are some extra flags that need to be passed in to the compiler to ensure that some things are available, like `com.sun`. Currently when you try to pass in the necessary flags for the plugin to work you'll see the following:

```
[info] compiling 1 Java source to /Users/ckipp/Documents/scala-workspace/minimal/out/javaModule/compile.dest/classes ...
[warn] Javac is running in 'local' mode. These flags have been removed:
[warn]  -J--add-exports, -Jjdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED, -J--add-exports, -Jjdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED, -J--add-exports, -Jjdk.compiler/com.sun.tools.javac.model=ALL-UNNAMED, -J--add-exports, -Jjdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED, -J--add-exports, -Jjdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
[error] An exception has occurred in the compiler (17.0.3). Please file a bug against the Java compiler via the Java bug reporting page (http://bugreport.java.com) after checking the Bug Database (http://bugs.java.com) for duplicates. Include your program, the following diagnostic, and the parameters passed to the Java compiler in your report. Thank you.
[error] java.lang.IllegalAccessError: class com.sourcegraph.semanticdb_javac.SemanticdbPlugin (in unnamed module @0x79fa2a2c) cannot access class com.sun.tools.javac.api.BasicJavacTask (in module jdk.compiler) because module jdk.compiler does not export com.sun.tools.javac.api to unnamed module @0x79fa2a2c
```

While I'm not super familiar with `LocalJavaCompiler` I see we hit on this in Metals [here](https://github.com/scalameta/metals/blob/main/sbt-metals/src/main/scala/scala/meta/metals/MetalsPlugin.scala#L54-L64) and it seemed that the solution was to make sure `javaHome` was set in sbt. By doing this when [`directOrFork`](https://github.com/sbt/zinc/blob/5804d50afd960330f34af35e4367fdb171204f5a/internal/zinc-compile-core/src/main/scala/sbt/internal/inc/javac/JavaCompiler.scala#L41-L67) eventually gets called in Zinc if the `javaHome` is not set then you see a call to `JavaCompiler.local.getOrElse(JavaCompiler.fork(None))` which causes us to get a local compiler that rejects `-J` flags.

I was curious how this was handled in Mill and it seems that in [`ZincWorkerImpl.scala`](https://github.com/com-lihaoyi/mill/blob/main/scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala#L66) where we call `compilers` we are passing `None` for the `javaHome` that will eventually call `directOrFork` meaning that we will always get a local fork.

Would it be possible to instead of always passing `None` to instead create a new `def javaHome: T[Option[PathRef]] = None` so that this can be overridden in specific scenarios like this? Then when `compileJava` or `compileMixed` is called there is another parameter, `javaHome` that also gets passed, which can then be used?

ckipp01:
Currently if you want to pass in some `javacOptions` that start with
`-J` the Java compiler will reject them since the local java compiler
won't accept `-J` flags. You can see this in an example on JDK 17 where
you need some `-J` flags to get the java semantidb compiler plugin to
work:

```scala
object javaModule extends JavaModule {

  def ivyDeps = Agg(ivy"com.sourcegraph:semanticdb-javac:0.8.2")

  def javacOptions = Seq(
    "-J--add-exports",
    "-Jjdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED",
    "-J--add-exports",
    "-Jjdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED",
    "-J--add-exports",
    "-Jjdk.compiler/com.sun.tools.javac.model=ALL-UNNAMED",
    "-J--add-exports",
    "-Jjdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED",
    "-J--add-exports",
    "-Jjdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED",
    s"-Xplugin:semanticdb -sourceroot:${T.workspace} -targetroot:${T.dest}"
  )

}
```

Which will result in:

```
❯ mill javaModule.compile
[19/19] javaModule.compile
[info] compiling 1 Java source to /Users/ckipp/Documents/scala-workspace/minimal/out/javaModule/compile.dest/classes ...
[warn] Javac is running in 'local' mode. These flags have been removed:
[warn]  -J--add-exports, -Jjdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED, -J--add-exports, -Jjdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED, -J--add-exports, -Jjdk.compiler/com.sun.tools.javac.model=ALL-UNNAMED, -J--add-exports, -Jjdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED, -J--add-exports, -Jjdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
[error] An exception has occurred in the compiler (17.0.4). Please file a bug against the Java compiler via the Java bug reporting page (http://bugreport.java.com) after checking the Bug Database (http://bugs.java.com) for duplicates. Include your program, the following diagnostic, and the parameters passed to the Java compiler in your report. Thank you.
[error] java.lang.IllegalAccessError: class com.sourcegraph.semanticdb_javac.SemanticdbPlugin (in unnamed module @0x4e59433c) cannot access class com.sun.tools.javac.api.BasicJavacTask (in module jdk.compiler) because module jdk.compiler does not export com.sun.tools.javac.api to unnamed module @0x4e59433c
...
```

This change looks at the `javacOptions` when creating the java comipler
and if detected makes sure you get a non-local compiler. If none are
detected, then the behavior basically is the same as it was before where
a local instance is used.

We also cache this to ensure it's not created over and over. However if
your `javacOption` change, then we do give a new instance.

Fixes #1983

### DIFF
--- a/scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
+++ b/scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
@@ -33,11 +33,12 @@ class ZincWorkerImpl(
 ) extends ZincWorkerApi with AutoCloseable {
   private val zincLogLevel = if (zincLogDebug) sbt.util.Level.Debug else sbt.util.Level.Info
   private[this] val ic = new sbt.internal.inc.IncrementalCompilerImpl()
-  private val javaOnlyCompilersCache = mutable.Map.empty[Seq[String], Compilers]
+  private val javaOnlyCompilersCache = mutable.Map.empty[Seq[String], SoftReference[Compilers]]
 
   def javaOnlyCompilers(javacOptions: Seq[String]) = {
-    javaOnlyCompilersCache.getOrElseUpdate(
-      javacOptions, {
+    javaOnlyCompilersCache.get(javacOptions) match {
+      case Some(SoftReference(compilers)) => compilers
+      case _ =>
         // Keep the classpath as written by the user
         val classpathOptions = ClasspathOptions.of(
           /*bootLibrary*/ false,
@@ -81,9 +82,10 @@ class ZincWorkerImpl(
           javac.JavaTools(javaCompiler, javaDoc)
         }
 
-        ic.compilers(javaTools, scalac)
-      }
-    )
+        val compilers = ic.compilers(javaTools, scalac)
+        javaOnlyCompilersCache.update(javacOptions, SoftReference(compilers))
+        compilers
+    }
   }
 
   val compilerBridgeLocks = collection.mutable.Map.empty[String, Object]
--- a/scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
+++ b/scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
@@ -4,8 +4,6 @@ import java.io.File
 import java.util.Optional
 import java.util.concurrent.ConcurrentHashMap
 
-import scala.ref.SoftReference
-import scala.util.Properties.isWin
 import mill.api.Loose.Agg
 import mill.api.{CompileProblemReporter, KeyedLockedCache, PathRef, internal}
 import mill.scalalib.api.{CompilationResult, ZincWorkerApi, ZincWorkerUtil => Util}
@@ -17,6 +15,9 @@ import xsbti.compile.{CompilerCache => _, FileAnalysisStore => _, ScalaInstance 
 import xsbti.{PathBasedFile, VirtualFile}
 
 import scala.annotation.nowarn
+import scala.collection.mutable
+import scala.ref.SoftReference
+import scala.util.Properties.isWin
 
 @internal
 class ZincWorkerImpl(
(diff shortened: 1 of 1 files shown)


lefou (commented):
This change looks reasonable. I think we should limit the newly introduced cache though, as it will grow indefinitely when the user frequently changes options. Either, we introduce a limit, which we probably also want to control via some settings, or we use a `SoftReference[Compilers]`.

ckipp01:
>  I think we should limit the newly introduced cache though, as it will grow indefinitely when the user frequently changes options

I was actually thinking that this would be super rare in the lifetime of this? Even if a user changes this 10 times, that's not a huge deal right? What would too large be in your opinion, and in what scenario do you think a user would be constantly updating these? I might be missing that use case in my mind.

lefou:
> > I think we should limit the newly introduced cache though, as it will grow indefinitely when the user frequently changes options
> 
> I was actually thinking that this would be super rare in the lifetime of this? Even if a user changes this 10 times, that's not a huge deal right? What would too large be in your opinion, and in what scenario do you think a user would be constantly updating these? I might be missing that use case in my mind.

If we don't expect the cache to be large, there is no reason to let it grow to exactly that. Even if this is not a critical spot in Mill, out of experience, sometimes, this will blow up. Imagine an unnoticed bug in `Compilers` which consumes a lot of memory, causing various other spurious issues in Mill.

Mill server is expected to run in the background and this can be very long. Also, BSP is starting an instance and keeps it for the full IDE session. If one plays around with some `javacOptions` which affect lots of modules, boom. I also to lazy to come up with a good default value, instead, I propose to just use a `SoftReference`, as we do in other places in Mill, e.g. `val classLoaderCache` in the same file. It's use is rather simple, but we prevent a whole class of issues.

ckipp01:
Alright, refactored to use a `SoftReference`. I've actually never used that before, so it was interesting reading up on it.

lefou (approved):
Looks good to me. Thank you!

PR #1989, opened 2022-08-14
docs: add some documentation about generating a doc jar

ckipp01:
I recently wanted to play around with the static site generator that
comes with Scaladoc for Scala 3, but the upstream documentation is very
sbt-focused and a bit confusing. I was also pretty unfamiliar with the
various doc-related tasks in Mill and since there wasn't any
documentation on it, I've included some basic documentation and also an
example of how to generate a full site with Scala 3.


lefou (approved):
Thanks for improving the documentation!

PR #1992, opened 2022-08-15
Fix filesystem-dependent sig in PathRef

### LINKED ISSUE #1808: PathRef for directories is filesystem-dependent and not reproducible
The hashCode of PathRef returns a value that depends on the file listing order of the file system.

https://github.com/com-lihaoyi/mill/blob/7d9eeac445f3153ef2c600534a40cd185d2ee9fc/main/api/src/mill/api/PathRef.scala#L38-L52

`os.walk` uses [DirectoryStream](https://docs.oracle.com/javase/8/docs/api/java/nio/file/DirectoryStream.html), which returns a list of files in an indefinite order depending on the file system.

> The elements returned by the iterator are in no specific order. Some file systems maintain special links to the directory itself and the directory's parent directory. Entries representing these links are not returned by the iterator.

As a concrete example, if an out directory is copied across machines in a Linux+XFS environment, the cache will be invalidated because the order in which files are stored changes. Note that ext4 hides this problem because it returns a stable file list.

I think mill should compute hashCode by sorting instead of filesystem-dependent order.

## File listing order in XFS

```
$ touch a
$ touch b
$ ls -f
.  ..  a  b

$ touch b
$ touch a
$ ls -f
.  ..  b  a
```

## File listing order in EXT4

```
$ touch a
$ touch b
$ ls -f
b  .  a  ..

$ touch b
$ touch a
$ ls -f
b  .  a  ..
```

atty303:
I think this is sufficient.

* Fix #1808

### DIFF
--- a/main/api/src/mill/api/PathRef.scala
+++ b/main/api/src/mill/api/PathRef.scala
@@ -36,7 +36,10 @@ object PathRef {
         digest.update(value.toByte)
       }
       if (os.exists(path)) {
-        for ((path, attrs) <- os.walk.attrs(path, includeTarget = true, followLinks = true)) {
+        for (
+          (path, attrs) <-
+            os.walk.attrs(path, includeTarget = true, followLinks = true).sortBy(_._1.toString)
+        ) {
           digest.update(path.toString.getBytes)
           if (!attrs.isDir) {
             if (isPosix) {


lefou (approved):
Looks good to me. Thank you!

PR #1994, opened 2022-08-16
feat: add in a pluginArtifact helper

ckipp01:
I've noticed myself (and copied from others) doing the same thing over
and over using the `scalaNativeBinaryVersion` help and then putting
together the artifact string ourselves, but this is easy to screw up if
you're not familiar with expected format. This just creates a helper
method that you can pass in the artifact name and Mill version and it
will put together the correct artifact name for you.

Note, I wasn't 100% sure where to put this. I sort of wanted to put it in
`module.Util` instead, but there are other similar things where I put it. Do
you think I should move this somewhere else?

### DIFF
--- a/scalalib/api/src/mill/scalalib/api/ZincWorkerUtil.scala
+++ b/scalalib/api/src/mill/scalalib/api/ZincWorkerUtil.scala
@@ -50,6 +50,8 @@ trait ZincWorkerUtil {
   val NightlyVersion = raw"""(\d+)\.(\d+)\.(\d+)-bin-[a-f0-9]*""".r
   val TypelevelVersion = raw"""(\d+)\.(\d+)\.(\d+)-bin-typelevel.*""".r
 
+  private val FullVersion = """^([0-9]+)\.([0-9]+)\.([0-9]+)(-.*)?$""".r
+
   def scalaBinaryVersion(scalaVersion: String) = scalaVersion match {
     case Scala3EarlyVersion(milestone) => s"3.0.0-$milestone"
     case Scala3Version(_, _) => "3"
@@ -61,12 +63,10 @@ trait ZincWorkerUtil {
     case _ => scalaVersion
   }
 
-  private val ScalaJSFullVersion = """^([0-9]+)\.([0-9]+)\.([0-9]+)(-.*)?$""".r
-
   def scalaJSBinaryVersion(scalaJSVersion: String) = scalaJSVersion match {
     case _ if scalaJSVersion.startsWith("0.6.") =>
       "0.6"
-    case ScalaJSFullVersion(major, minor, patch, suffix) =>
+    case FullVersion(major, minor, patch, suffix) =>
       if (suffix != null && minor == "0" && patch == "0")
         s"$major.$minor$suffix"
       else
@@ -76,14 +76,12 @@ trait ZincWorkerUtil {
   def scalaJSWorkerVersion(scalaJSVersion: String) = scalaJSVersion match {
     case _ if scalaJSVersion.startsWith("0.6.") =>
       "0.6"
-    case ScalaJSFullVersion(major, _, _, _) =>
+    case FullVersion(major, _, _, _) =>
       major
   }
 
-  private val ScalaNativeFullVersion = """^([0-9]+)\.([0-9]+)\.([0-9]+)(-.*)?$""".r
-
   def scalaNativeBinaryVersion(version: String) = version match {
-    case ScalaNativeFullVersion(major, minor, patch, suffix) =>
+    case FullVersion(major, minor, patch, suffix) =>
       if (suffix != null && patch == "0")
         version
       else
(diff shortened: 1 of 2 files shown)


scalalib/api/src/mill/scalalib/api/ZincWorkerUtil.scala
```diff
@@ -50,6 +50,8 @@ trait ZincWorkerUtil {
   val NightlyVersion = raw"""(\d+)\.(\d+)\.(\d+)-bin-[a-f0-9]*""".r
   val TypelevelVersion = raw"""(\d+)\.(\d+)\.(\d+)-bin-typelevel.*""".r
 
+  private val FullVersion = """^([0-9]+)\.([0-9]+)\.([0-9]+)(-.*)?$""".r
```
ckipp01, line 53: `ScalaJSFullVersion` and `ScalaNativeFullVersion` were exactly the same, and private, so I just combined them into this.

scalalib/api/src/mill/scalalib/api/ZincWorkerUtil.scala
```diff
@@ -148,6 +146,19 @@ trait ZincWorkerUtil {
...
+   *
+   * Ex. If your plugin name is example and you're using 0.10.0 this would create:
+   *   - example_mill0.10
+   *
+   * @param name the name of your plugin
+   * @param millVersion the Mill version you're targeting
+   * @return
+   */
+  def pluginArtifact(name: String, millVersion: String) =
+    s"${name}_mill${scalaBinaryVersion(millVersion)}"
```
lefou, line 161: I think the same makes it hard to guess, which part of the artifact(name/id) is the result of this `def`. What about just having a `def` that only calculated the platform suffix?

`scalaBinaryVersion` is specific for Scala versions, e.g. versions starting with `3` are handled differently than those starting with `2`. Here we should keep logic specific to Mill plugins, e.g. when starting with `0`, we use the major+minor. But for major `1`, we probably want to switch to a more stable platform which only includes the major.

ckipp01: > I think the name makes it hard to guess, which part of the artifact(name/id) is the result of this def. What about just having a def that only calculated the platform suffix?

I'm not fully sure I understand what you mean? So you're saying that the `def` here should only calculate `_mill0.10` part? So then the user would just use this with their name locally to put together the name? If so, I still think it's nicer to have a method that does it all, so then the user doesn't have to think about it at all, just pass in their name and the mill version, and boom, done.

> scalaBinaryVersion is specific for Scala versions, e.g. versions starting with 3 are handled differently than those starting with 2. Here we should keep logic specific to Mill plugins, e.g. when starting with 0, we use the major+minor. But for major 1, we probably want to switch to a more stable platform which only includes the major.

I can change this back and instead create a `millBinaryVersion` maybe? And this is only used for this purpose? wdyt?

lefou: > > I think the name makes it hard to guess, which part of the artifact(name/id) is the result of this def. What about just having a def that only calculated the platform suffix?
> 
> I'm not fully sure I understand what you mean? So you're saying that the `def` here should only calculate `_mill0.10` part? So then the user would just use this with their name locally to put together the name? If so, I still think it's nicer to have a method that does it all, so then the user doesn't have to think about it at all, just pass in their name and the mill version, and boom, done.

Sorry for being not clear enough. I think a `def pluginArtifact` leaves too much doubt what it is returning to the user. The pure name ( `foo`), or the full artifactId (`foo_mill0.10_2.13.8`) or some partial full name (`foo_mill_0.10`). Maybe the name is too vague? I wonder where this method is to be used?

Shouldn't we instead just set the `JavaModule#platformSuffix`? And always render it as part of the `artifactSuffix`, which is currently done identically in `ScalaJSModule` and `ScalaNativeModule` ?

We could even went ahead and introduce a `MillPluginModule`, that also contains the logic of inferring the correct platform suffix based on the Mill version (or some enum-like selection).


> > scalaBinaryVersion is specific for Scala versions, e.g. versions starting with 3 are handled differently than those starting with 2. Here we should keep logic specific to Mill plugins, e.g. when starting with 0, we use the major+minor. But for major 1, we probably want to switch to a more stable platform which only includes the major.
> 
> I can change this back and instead create a `millBinaryVersion` maybe? And this is only used for this purpose? wdyt?

Yeah, a `def millBinaryVersion` is what I had in mind.

ckipp01: Ahhhh yes, good idea. I like the idea of a `MillPluginModule`. I'll close this and pursue that. I didn't even think about the `JavaModule#platformSuffix` which also make sense.

lefou:
I liked the cleanup and the consolidation to `FullVersion` though.

PR #2000, opened 2022-08-23
docs: add mill-scip to external plugins

ckipp01:


lefou (commented):
It's a bit unclear, how to use this plugin, as the added documentation only speaks about thirdparty tools like `scip-java` and `cs`.

Is there also a way to run this plugin via `mill`? If not, you may want to add some info, that it is some kind of integration tooling, but not directly usable by Mill users.

ckipp01:
> It's a bit unclear, how to use this plugin, as the added documentation only speaks about thirdparty tools like `scip-java` and `cs`.
> 
> Is there also a way to run this plugin via `mill`? If not, you may want to add some info, that it is some kind of integration tooling, but not directly usable by Mill users.

Sure I added the instructions to also use it directly via `mill`. I didn't include it originally because I assumed that most user will use this in CI and doing:

```
scip-java index
```
Will get you what you want (using Mill under the hood) without needing to worry about the version or the syntax since it takes care of it all for you. Even locally it's easier to use this way, but I get your point. I added the section in on manual usage.

lefou (commented):
Thank you! Your additional notes make it much clearer, e.g. the point of this Mill plugin is to enable `scip-java` to function by extracting some data. I think, it's helpful for the user to be able to see the bigger picture.

lefou (approved):

PR #2006, opened 2022-08-24
BSP: use build/logMessage for logging

### LINKED ISSUE #2005: `build/showMessage` that shouldn't be a `build/showMessage` when using Mill BSP
Using Mill as a BSP server (version 0.10.7) I am often getting this message shown to me when using Metals. I don't believe this is a message that should be shown to the user. I see this is marked as `debug` in https://github.com/com-lihaoyi/mill/blob/main/scalalib/src/mill/scalalib/SemanticDbJavaModule.scala#L87, so I wouldn't expect this to be a `build/showMessage` from the build server.

```
[Trace - 08:32:57 pm] Received notification 'build/showMessage'
Params: {
  "type": 4,
  "message": "effective scalac options: List(-Xplugin:/Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scalameta/semanticdb-scalac_2.13.8/4.5.9/semanticdb-scalac_2.13.8-4.5.9.jar, -encoding, utf8, -Xsource:3, -explaintypes, -feature, -language:existentials, -language:experimental.macros, -language:higherKinds, -language:implicitConversions, -unchecked, -Xcheckinit, -Xlint:adapted-args, -Xlint:constant, -Xlint:delayedinit-select, -Xlint:deprecation, -Xlint:doc-detached, -Xlint:inaccessible, -Xlint:infer-any, -Xlint:missing-interpolator, -Xlint:nullary-unit, -Xlint:option-implicit, -Xlint:package-object-classes, -Xlint:poly-implicit-overload, -Xlint:private-shadow, -Xlint:stars-align, -Xlint:type-parameter-shadow, -Wunused:nowarn, -Wdead-code, -Wextra-implicit, -Wnumeric-widen, -Xlint:implicit-recursion, -Wunused:implicits, -Wunused:explicits, -Wunused:imports, -Wunused:locals, -Wunused:params, -Wunused:patvars, -Wunused:privates, -Wvalue-discard, -Vimplicits, -Vtype-diffs, -Xplugin:/Users/ckipp/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scalameta/semanticdb-scalac_2.13.8/4.5.11/semanticdb-scalac_2.13.8-4.5.11.jar, -Yrangepos, -P:semanticdb:sourceroot:/Users/ckipp/Documents/scala-workspace/mill-scip, -Ystop-after:semanticdb-typer)"
}
```

ckipp01:
This pr makes 2 changes. First, debug messages shouldn't go through
`build/showMessage`. Taken from the BSP [docs](https://build-server-protocol.github.io/docs/specification.html#show-message):

> The show message notification is sent from a server to a client to ask
> the client to display a particular message in the user interface.

From my experience these are reserved for helpful messages that the user
should know about. So I changed this to `build/logMessage` which seems
more appropriate:

> The log message notification is sent from the server to the client to
> ask the client to log a particular message.

I also then just went ahead and removed this because even if we keep it
as log message, this will get logged all the time, and I don't think
it's that useful to log unless you think it should be. It's just every
time new semanticdb is being produced this is getting logged, which is
almost constant during development.

Closes #2005

### DIFF
--- a/bsp/src/mill/bsp/MillBspLogger.scala
+++ b/bsp/src/mill/bsp/MillBspLogger.scala
@@ -53,7 +53,7 @@ class MillBspLogger(client: BuildClient, taskId: Int, logger: Logger)
   override def debug(s: String): Unit = {
     super.debug(s)
     if (debugEnabled) {
-      client.onBuildShowMessage(new ShowMessageParams(MessageType.LOG, s))
+      client.onBuildLogMessage(new LogMessageParams(MessageType.LOG, s))
     }
   }
 


scalalib/src/mill/scalalib/SemanticDbJavaModule.scala
```diff
@@ -84,7 +84,6 @@ trait SemanticDbJavaModule extends JavaModule { hostModule =>
             }
         )
           .filterNot(_ == "-Xfatal-warnings")
-        T.log.debug(s"effective scalac options: ${scalacOptions}")
```
lefou, line 87: The idea is, that a user only sees this when option `--debug` is enabled. I agree, that we probably used the wrong mechanism to transport this to a BSP client, but we should not be forced to avoid debug logs in favor of a nice IDE experience. Why is the BSP client showing debug messages without the user explicitly asking for it? Isn't there any other option?

ckipp01: With the way the logger was set up before we were doing exactly that, telling the build client **show this to the user**. By switching it to `onBuildLogMessage` we avoid this but these are still sent to the client and the client will log this out every time. I looked to see if there was a `MessageType.DEBUG` which would be perfect here, but there really isn't that. So you end up in a situation where if you would look in the client log you'll just see this spammed in there since it's coming as an info message. This _sort_ of seems like a missing feature of BSP.

lefou: Ok. I already have a stashed patch which enabled `--debug` in the BSP server, when it was enabled at`BSP/install`-time. So, if we only send this to the client when `--debug` is enabled (`Logger.debugEnabled` is `true`), we should be fine. WDYT?

ckipp01: Yea I think that should be fine, and useful.

lefou: Ok. I already have a stashed patch which enabled `--debug` in the BSP server, when it was enabled at`BSP/install`-time. So, if we only send this to the client when `--debug` is enabled (`Logger.debugEnabled` is `true`), we should be fine. WDYT?

It's basically this: github.com/com-lihaoyi/mill/pull/1977/commits/3328d99513416a9535e739a4d809d2078f967fdb

bsp/src/mill/bsp/MillBspLogger.scala
```diff
@@ -52,7 +52,7 @@ class MillBspLogger(client: BuildClient, taskId: Int, logger: Logger)
 
   override def debug(s: String): Unit = {
     super.debug(s)
-    client.onBuildShowMessage(new ShowMessageParams(MessageType.LOG, s))
+    client.onBuildLogMessage(new LogMessageParams(MessageType.LOG, s))
```
lefou, line 55: According to the other comment (https://github.com/com-lihaoyi/mill/pull/2006/files#r954194830), we should put this behind a guard.

```suggestion
    if(debugEnabled) {
      client.onBuildLogMessage(new LogMessageParams(MessageType.LOG, s))
    }
```

scalalib/src/mill/scalalib/SemanticDbJavaModule.scala
lefou: * Created PR #2007

ckipp01: Ah ok great. Should we merge that other pr, and then I can rebase and updaet?

bsp/src/mill/bsp/MillBspLogger.scala
lefou: I moved this guard in the other PR.

scalalib/src/mill/scalalib/SemanticDbJavaModule.scala
lefou: The other PR is now independent of this one. Just keep the debug log message.

lefou (approved):
Looks good to me. Thank you!

PR #2008, opened 2022-08-25
Contrib: Gitlab plugin

aheiska:
This contrib plugin provides support for Gitlab maven package registry publishing / dependencies and gitlab CI/CD. Gitlab does not support http basic authentication, instead a correct http header / token type must be inserted.

Code is inspired code of artifactory and bintray plugins but does not take credentials from command line. Instead it searches token from environment according to configuration (with hopefully sane defaults). By default CI_JOB_TOKEN environment variable is read which provides no-configuration support for gitlab CI/CD.

This works for me (self hosted gitlab 15.2.2-ee). I'm pretty sure though there is some use case that falls through the cracks. I added quite powerful customization mechanism that allows coding over possible limitations of this implementation. README.md should explain usage quite well and code is not too big to digest.

I'm especially interested if this follow the mill way of doing things and if not, what should be changed to make it such.


lefou (commented):

I can't really say anything to the implementation. I neither use Gitlab, not are there any tests, but it sounds like you are using it already and have tested it that way in your environment.

I like the rich documentation, but want to request, to add it to the documentation site under `docs/antora/modules/ROOT/pages/`. You need to convert it to Asciidoc(tor). There are various tools like IntelliJ, which can do this conversion automatically. Have a look how the other plugins you've already mentioned are documented. If you have questions, don't hesitate to ask.

build.sc
```diff
@@ -869,6 +869,10 @@ object contrib extends MillModule {
   object bintray extends MillModule {
     override def compileModuleDeps = Seq(scalalib)
   }
+  
+  object gitlab extends MillModule {
```
lefou, line 873:
```suggestion
  object gitlab extends MillInternalModule {
```

This should fix the MiMa check issues

aheiska:
Thanks. I'll fix the documentatio tomorrow.

aheiska:
Moved and converted the documentation.

aheiska:
There is still one test failure. Not really sure what is the problem with that.

lefou:
I restarted the test. It's probably just some CI hickup.

Can you also add a link to the plugin page to `docs/antora/modules/ROOT/nav.adoc`

aheiska:
Added.

aheiska:
Just noticed that change to `MillInternalModule` dropped `MillAutotestSetup`. Made the module explicitly inherit it.

lefou:
> Just noticed that change to `MillInternalModule` dropped `MillAutotestSetup`. Made the module explicitly inherit it.

Yeah, when I suggested that cahnge, there were no tests, so it looked good enough.

contrib/gitlab/src/mill/contrib/gitlab/GitlabAuthHeaders.scala
```diff
@@ -0,0 +1,11 @@
+package mill.contrib.gitlab
+
+case class GitlabAuthHeaders(headers: Seq[(String, String)])
+
+object GitlabAuthHeaders {
+  def apply(header: String, value: String): GitlabAuthHeaders = GitlabAuthHeaders(Seq(header -> value))
+
+  def personalHeader(token: String): GitlabAuthHeaders = GitlabAuthHeaders("Private-Token", token)
+  def deployHeader(token: String): GitlabAuthHeaders   = GitlabAuthHeaders("Deploy-Token", token)
+  def jobHeader(token: String): GitlabAuthHeaders      = GitlabAuthHeaders("Job-Token", token)
```
lefou, line 10: I would recommend, to stay with the name close the the intended result. E.g. 

```scala
def privateToken(token: String): GitlibAuthHeaders = GitlabAuthHeadsers("Private-Token", token)
```

or 

```scala
def `Private-Token`(token: String): GitlibAuthHeaders = GitlabAuthHeadsers("Private-Token", token)
```

contrib/gitlab/src/mill/contrib/gitlab/GitlabHttpApi.scala
```diff
@@ -0,0 +1,32 @@
+package mill.contrib.gitlab
+
+import scala.concurrent.duration._
+
+object GitlabUploader {
+  type Upload = (String, Array[Byte]) => requests.Response
```
lefou, line 6: It would be nice to explicitly add the `requests` dependency to the modules `ivyDeps`. It's currently only transitively available, which may change and break thinks.

lefou (commented):
Nice, this is almost good to go. I added two small comments, otherwise this looks good to me.

contrib/gitlab/src/mill/contrib/gitlab/GitlabPackageRepository.scala
```diff
@@ -0,0 +1,34 @@
...
+
+import mill.scalalib.publish.Artifact
+
+sealed trait GitlabPackageRepository {
+  def url(): String
+}
+
+// Could also support project name (https://docs.gitlab.com/ee/user/packages/maven_repository/index.html#project-level-maven-endpoint)
+// though only ID can be used for publishing.
+case class ProjectRepository(baseUrl: String, projectId: Int) extends GitlabPackageRepository() {
```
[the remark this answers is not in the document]
aheiska: Hmm. I think at some point GitlabPackageRepository was an abstract case class and had a parameter, so leftovers. Will remove.

contrib/gitlab/src/mill/contrib/gitlab/GitlabAuthHeaders.scala
aheiska: Gitlab documentation refers these as personal tokens (https://docs.gitlab.com/ee/user/packages/maven_repository/#authenticate-with-a-personal-access-token-in-maven). I'm a bit dubious if this should be personalToken or privateToken. I do agree, that *Token is a better name.

aheiska: Come to think of it. This is quite "internal" (though not any way hidden). I'll change to privateToken but keep the name `Personal` in `GitlabTokenLookup`.

contrib/gitlab/src/mill/contrib/gitlab/GitlabHttpApi.scala
aheiska: Will add 👍 . This depends from from `os` also. Should it be added also?

aheiska: Added all. Added to artifactory, bintray and codearticat also since they depend from requests also.

lefou (approved):
Looks good to merge.

lefou:
Thank you!

aheiska:
Thank you :). Happy to contribute.

PR #2024, opened 2022-09-05
Fix issue with lazy val on Scala Native + Scala 3.2

lolgab:
Scala Native depends on the scala3-library `3.1.3`.
To work on Scala 3.2 you need to explicitly depend on scala3-library
for `3.2.0`.
In Scala.js it is a single artifact generated in the dotty repository,
so it's platformed. In Scala Native, however, we need to depend on the
jvm artifact, which is used by the compiler, so we patch
`scalaLibraryIvyDeps` to be `platformed = false`

### DIFF
--- a/scalanativelib/src/ScalaNativeModule.scala
+++ b/scalanativelib/src/ScalaNativeModule.scala
@@ -8,7 +8,7 @@ import mill.api.{internal, Result}
 import mill.define.{Target, Task}
 import mill.modules.Jvm
 import mill.scalalib.api.Util.{isScala3, scalaBinaryVersion}
-import mill.scalalib.{Dep, DepSyntax, Lib, SbtModule, ScalaModule, TestModule}
+import mill.scalalib.{CrossVersion, Dep, DepSyntax, Lib, SbtModule, ScalaModule, TestModule}
 import mill.testrunner.TestRunner
 import mill.scalanativelib.api._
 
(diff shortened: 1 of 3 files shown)


scalanativelib/src/ScalaNativeModule.scala
```diff
@@ -84,7 +84,13 @@ trait ScalaNativeModule extends ScalaModule { outer =>
...
 
   override def scalaLibraryIvyDeps = T {
-    if (isScala3(scalaVersion())) Agg.empty[Dep] else super.scalaLibraryIvyDeps()
+    super.scalaLibraryIvyDeps().map(dep =>
+      dep.copy(cross = dep.cross match {
+        case c: CrossVersion.Constant => c.copy(platformed = false)
+        case c: CrossVersion.Binary => c.copy(platformed = false)
+        case c: CrossVersion.Full => c.copy(platformed = false)
+      })
+    )
```
lefou, line 93: Looks like the match on the concrete `CrossType` is superfluous, as we always set `platformed = false`, for each of the possible sealed trait instances.

lolgab: Yes but `copy` is not available on the `trait`. I have some API changes in the PR I opened to remove the second `::` in deps but I didn't want to move the discussion here.
For now we can live with this I guess and then refactor in case the other PR gets merged

lefou: Oh right, sorry.

lefou:
Looks good to me, so far.

PR #2042, opened 2022-09-19
Customize the native binary output file name

carlosedp:
This allows customizing the output file name from `out` to anything.

Previously it would require:

```scala
  def nativeLink = T {
    os.Path(scalaNativeWorker().nativeLink(nativeConfig(), (T.dest / "binaryname").toIO))
  }
```

With this PR, it would be: `def nativeBinaryName   = "binaryname"`

### DIFF
--- a/scalanativelib/src/ScalaNativeModule.scala
+++ b/scalanativelib/src/ScalaNativeModule.scala
@@ -215,7 +215,7 @@ trait ScalaNativeModule extends ScalaModule { outer =>
   }
 
   // Defines the native binary name
-  def nativeBinaryName: String = { "out" }
+  def nativeBinaryName: String = { this.toString }
 
   // Generates native binary
   def nativeLink = T {
--- a/scalanativelib/src/ScalaNativeModule.scala
+++ b/scalanativelib/src/ScalaNativeModule.scala
@@ -214,9 +214,12 @@ trait ScalaNativeModule extends ScalaModule { outer =>
     )
   }
 
+  // Defines the native binary name
+  def nativeBinaryName: String = { "out" }
+
   // Generates native binary
   def nativeLink = T {
-    os.Path(scalaNativeWorker().nativeLink(nativeConfig(), (T.dest / "out").toIO))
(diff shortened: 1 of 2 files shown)


carlosedp:
Now it's passing :)

lefou:
Can you elaborate whether this is pure convenience or there is a real need for this? I'm not so deep in Scala Native workflows, so I also cc'ing @lolgab here.

carlosedp:
I could also do:

```scala
def nativeBinaryName: String = { this.toString }
``` 

So the output binary would be named as the object being generated (also could be set by the user if needed). I think it's even more useful while keeping build.sc cleaner.

scalanativelib/src/ScalaNativeModule.scala
```diff
@@ -214,9 +214,12 @@ trait ScalaNativeModule extends ScalaModule { outer =>
     )
   }
 
+  // Defines the native binary name
+  def nativeBinaryName: String = { this.toString }
```
lefou, line 218: Regardless of the feature itself, this is IMHO a bad default. We have more meaningful options like `artifactName`.

lefou (commented):
TBH, I'm not sold by this feature. I think it leads to API bloat. Esp. as for consistence, this would also be wanted in other modules, e.g. `JavaModule.jar`.

I'm open to a good motivation though. There are alternative ways to achieve that. Typically not much longer that overriding `nativeBinaryName`. E.g. a user could simply override the target and rename/copy the result of `super.<target>` or anything else required to customize the out file.

scalanativelib/src/ScalaNativeModule.scala
lefou: Besides, the choosen fixed default ("out") isn't arbitrary, but allows other tools to easily consume files without the need to deeply inspect the build.

carlosedp: Ah sure, I could fix that.

carlosedp:
I don't want to bloat the API in any way. The motivation is to give flexibility to the user without bloating `build.sc` itself (I remember as a user looking at both build.sc and build.sbt and thinking "why the hell are they so complex!").
No problem canceling this PR if it would cause maintenance burden in the future.

PR #2059, opened 2022-10-07
Show gpg output in server/client in publish (fix #387)

### LINKED ISSUE #387: The publish task should display an error message when the user does not have a gpg secret key
### Current output

```
$ mill hamster[2.11.11].jvm.publish --gpgPassphrase passphrase --release false --sonatypeCreds "user:password"
```
```
[125/125] hamster[2.11.11].jvm.publish
1 targets failed
hamster[2.11.11].jvm.publish ammonite.ops.InteractiveShelloutException
    ammonite.ops.Shellout$.executeInteractive(Shellout.scala:56)
    ammonite.ops.Shellout$.$anonfun$$percent$1(Shellout.scala:14)
    ammonite.ops.Shellout$.$anonfun$$percent$1$adapted(Shellout.scala:14)
    mill.scalalib.publish.SonatypePublisher.poorMansSign(SonatypePublisher.scala:144)
    mill.scalalib.publish.SonatypePublisher.$anonfun$publishAll$4(SonatypePublisher.scala:34)
    scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:234)
    scala.collection.Iterator.foreach(Iterator.scala:944)
    scala.collection.Iterator.foreach$(Iterator.scala:944)
```

**NOTE:** When using the interactive mode with the `-i` flag the error message is display:

```
gpg: no default secret key: No secret key
gpg: signing failed: No secret key
```

### Expected output

```
$ mill hamster[2.11.11].jvm.publish --gpgPassphrase passphrase --release false --sonatypeCreds "user:password"
```
```
[125/125] hamster[2.11.11].jvm.publish
1 targets failed
gpg: no default secret key: No secret key
gpg: signing failed: No secret key
hamster[2.11.11].jvm.publish ammonite.ops.InteractiveShelloutException
    ammonite.ops.Shellout$.executeInteractive(Shellout.scala:56)
    ammonite.ops.Shellout$.$anonfun$$percent$1(Shellout.scala:14)
    ammonite.ops.Shellout$.$anonfun$$percent$1$adapted(Shellout.scala:14)
    mill.scalalib.publish.SonatypePublisher.poorMansSign(SonatypePublisher.scala:144)
    mill.scalalib.publish.SonatypePublisher.$anonfun$publishAll$4(SonatypePublisher.scala:34)
    scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:234)
    scala.collection.Iterator.foreach(Iterator.scala:944)
    scala.collection.Iterator.foreach$(Iterator.scala:944)
```

We could go one step further and capture the error message from the `gpg`command line to help the user.

```
[125/125] hamster[2.11.11].jvm.publish
1 targets failed
gpg: no default secret key: No secret key
gpg: signing failed: No secret key
You need to set up a GPG key to sign artifacts. Generate a new key, using the following command: "gpg --gen-key".
```

I think the user will also need to publish the key:
```
$ gpg --keyserver hkp://ipv4.pool.sks-keyservers.net:80 --send-keys <key-id>
```

lolgab:
Avoid using `os.Inherit` which is known to break the output in server/client mode.

Fixes #387
Checked manually and the errors now appear with and without `-i`.

### DIFF
--- a/scalalib/src/publish/SonatypePublisher.scala
+++ b/scalalib/src/publish/SonatypePublisher.scala
@@ -16,9 +16,37 @@ class SonatypePublisher(
     readTimeout: Int,
     connectTimeout: Int,
     log: Logger,
+    workspace: os.Path,
+    env: Map[String, String],
     awaitTimeout: Int,
-    stagingRelease: Boolean = true
+    stagingRelease: Boolean
 ) {
+  @deprecated("Use other constructor instead", since = "mill 0.10.8")
+  def this(
+      uri: String,
+      snapshotUri: String,
+      credentials: String,
+      signed: Boolean,
+      gpgArgs: Seq[String],
+      readTimeout: Int,
+      connectTimeout: Int,
+      log: Logger,
+      awaitTimeout: Int,
+      stagingRelease: Boolean = true
+  ) = this(
+    uri = uri,
+    snapshotUri = snapshotUri,
+    credentials = credentials,
+    signed = signed,
+    gpgArgs = gpgArgs,
+    readTimeout = readTimeout,
+    connectTimeout = connectTimeout,
+    log = log,
+    workspace = os.pwd,
+    env = sys.env,
+    awaitTimeout = awaitTimeout,
+    stagingRelease = stagingRelease
+  )
 
   private val api = new SonatypeHttpApi(
     uri,
@@ -175,7 +203,7 @@ class SonatypePublisher(
     val fileName = file.toString
     val command = "gpg" +: args :+ fileName
 
-    Jvm.runSubprocess(command, sys.env, workingDir = null)
+    Jvm.runSubprocess(command, env, workspace)
     os.Path(fileName + ".asc")
   }
 
--- a/scalalib/src/publish/SonatypePublisher.scala
+++ b/scalalib/src/publish/SonatypePublisher.scala
@@ -175,7 +175,7 @@ class SonatypePublisher(
     val fileName = file.toString
     val command = "gpg" +: args :+ fileName
 
-    Jvm.runSubprocess(command, Map.empty[String, String], workingDir = null)
+    Jvm.runSubprocess(command, sys.env, workingDir = null)
     os.Path(fileName + ".asc")
   }
 
--- a/scalalib/src/publish/SonatypePublisher.scala
+++ b/scalalib/src/publish/SonatypePublisher.scala
@@ -4,6 +4,7 @@ import java.math.BigInteger
 import java.security.MessageDigest
 
 import mill.api.Logger
+import mill.modules.Jvm
 import os.Shellable
 
 class SonatypePublisher(
@@ -174,8 +175,7 @@ class SonatypePublisher(
     val fileName = file.toString
     val command = "gpg" +: args :+ fileName
 
-    os.proc(command.map(v => v: Shellable))
-      .call(stdin = os.Inherit, stdout = os.Inherit, stderr = os.Inherit)
+    Jvm.runSubprocess(command, Map.empty[String, String], workingDir = null)
     os.Path(fileName + ".asc")
   }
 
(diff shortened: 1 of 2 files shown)


scalalib/src/publish/SonatypePublisher.scala
```diff
@@ -174,8 +175,7 @@ class SonatypePublisher(
     val fileName = file.toString
     val command = "gpg" +: args :+ fileName
 
-    os.proc(command.map(v => v: Shellable))
-      .call(stdin = os.Inherit, stdout = os.Inherit, stderr = os.Inherit)
+    Jvm.runSubprocess(command, Map.empty[String, String], workingDir = null)
```
lefou, line 178: Why are you setting `workingDir` explicitly to `null`? (The current implementation is handling it as `os.pwd`.) Besides being ugly and non-idiomatic, it may not always be the right location. I think using [`T.workspace`](https://com-lihaoyi.github.io/mill/mill/Task_Context_API.html#_mill_api_ctx_workspace) is the better option here.

lefou (commented):
Great catch. This change makes sense. One thing though...

scalalib/src/publish/SonatypePublisher.scala
lolgab: Are you sure the current implementation is handling as `os.pwd`?
My understanding is that we are calling `os.proc()` which by default uses `null` as `cwd`
https://github.com/com-lihaoyi/os-lib/blob/main/os/src-jvm/ProcessOps.scala#L53

To better match the current behavior I need to pass `sys.env` as the env instead of an empty `Map`.

lefou: Yeah, I think so.

https://github.com/com-lihaoyi/os-lib/blob/6b31f388799743b1410110e474a47ac372aa4be4/os/src-jvm/ProcessOps.scala#L120

But it's an undocumented (I think) implementation detail, so I'd prefer to make it explicit.

lefou: Also, the current env should come via `T.env`, as sys.env might be stale in server-mode, and this whole PR is about improving server-mode, right?

lolgab: Yes, you are right!
When `cwd` is `null` it fallbacks to `os.pwd`:
```scala
@ os.proc("pwd").call(cwd = null).out.text
res1: String = """/Users/lorenzo
"""

@ os.proc("pwd").call(cwd = os.pwd).out.text
res2: String = """/Users/lorenzo
"""
```

lolgab: Now, `SonatypePublisher` doesn't know `Ctx` and can't access `T.workspace` or `T.env`. Should I deprecate the APIs and define new ones that receive `implicit ctx: Ctx`? It's a bit of work though

lolgab: Since the class was already receiving a `Logger` explicitly via `T.log`, I added another two parameters, `workspace` and `env` and I'm passing `T.workspace` and `T.env`. In the deprecated constructor, I'm using `sys.env` and `os.pwd` instead.

lefou: The only issue I see with this approach is, that there is a chance the created instance is used when the context is no longer valid. Should be no issue for `Ctx.workspace` or `Ctx.log` (although misleading) , but `Ctx.dest` of `Ctx.env` could be already invalid. 

As a compromise, we can keep it this way for now, but schedule a rework of the Publisher API to the next major release. I already have a tracker ticket for that:

* https://github.com/com-lihaoyi/mill/discussions/1826

lolgab: I understand the issue.
Currently, we create a fresh instance of `SonatypePublisher` every time so it's not an issue internally. But we can change the API when we are in the next breaking window (0.11.x)

lefou (approved):
Looks good to me. Thank you!

PR #2068, opened 2022-10-17
Switched to external mill-moduledefs dependency

### LINKED ISSUE #2035: Split out scalac compiler plugins
We have essentially two compiler plugins in Mill which we currently build as one:

* we don't require `override` on overridden `def`s and `val`s and special handle `mill.moduledefs.Cacher` instances
* we encode the ScalaDoc of targets as annotation `mill.moduledefs.Scaladoc`, se we can later extract and re-use them in `mill inspect`

Those are currently packaged in one ordinary Scala artifact `mill-moduledefs` based on Scala binary version.
Instead, they should be packaged for the full Scala version, to avoid conflicts.

Also, because of their packaging, they are currently not really usable outside of Mill, which makes developing external Mill plugins a bit more inconvenient, as e.g. the scaladoc encoding would be useful for external plugins too.

Contrary to the fact, that these modules change quite seldom, they need to be (re-)released whenever new Scala versions come out.

I suggest, to make them standalone project with their own release cycle and well defined compatibility scheme, so we can use them in Mill accordingly.

lefou:
This is the second step of externalizing the `moduledefs` module.

The first step was to migrate the current `moduledefs` code into an external project [`mill-moduledefs`](https://github.com/com-lihaoyi/mill-moduledefs). 

This second step now removes the `moduledefs` local module and refactors the build setup to use the external one. In the external project, we split the shared code and the scalac plugin code into two modules, to respect the different use cases and technical requirements.

* The new `mill-moduledefs` aritfact contains the traits and classes, which need to be on the classpath
* The `scalac-mill-moduledefs-plugin` artifact contains the scalac plugin, which is distributed for the full compiler version and is only used at compile-time

The overall goal is, to develop and maintain both projects (Mill and mill-moduledefs) independently. It especially enables proper distribution of the compiler plugin for each full version of the Scala compiler. In addition, we can post-release the now independent mill-moduledefs project for newer Scala versions without bumping the version (if needed), so externally developed Mill plugin projects can safely use newer Scala versions without risking compiler issues.

* Fixes #2035

Review by @lihaoyi


build.sc
```diff
@@ -131,6 +131,9 @@ object Deps {
   val lambdaTest = ivy"de.tototec:de.tobiasroeser.lambdatest:0.7.1"
   val log4j2Core = ivy"org.apache.logging.log4j:log4j-core:2.18.0"
   val osLib = ivy"com.lihaoyi::os-lib:0.8.1"
+  val millModuledefsVersion = "0.10.9-alpha-1"
```
lolgab, line 134: Why are we using an alpha version?

lefou: It's until we settle on this setup. There will be a stable release (`0.10.9`) once we do. 

Currently (before this PR), Mill also contains a `mill-moduledefs` artifact, and we need to distinguish it from the other one. Hopefully, once we tag Mill 0.10.9, we can also release it as `0.10.9` without any collisions. 

Also, Maven-compatible tools treat `0.10.9-alpha-1` as _before_ `0.10.9`, and the meaning of `alpha` in software development will make it semantically clear. Although in Mill, we used to treat a `0.10.9-1` as _after_ `0.10.9`.

lolgab (approved):

PR #2070, opened 2022-10-18
Remove jetty dependencies from Scala.js worker

lolgab:


lefou (approved):
Looks good to me. I don't have deep enough insights to judge, whether this is correct, but tests don't fail and all those ditched dependencies are rather old. According to an issue search in scala-js, PhantomJS is no longer recommended for at least some years and by removing these old Jetty libs, we also close multiple security vulnerabilities. Thank you, Lorenzo!

PR #2080, opened 2022-10-23
Add actual Mill cli as sysprop and use it in BSP config

lefou:
Uses the Java system property `mill.main.cli` internally.
Also accepts a `MILL_MAIN_CLI` environment variable to override the cli name (if needed).

When BSP connection files (JSON) are generated, the value of the sysprop is used as application, resulting in a BSP server setup, that will respect config changes e.g. to `.mill-version`.

Read more about the motivation in discussion:

* https://github.com/com-lihaoyi/mill/discussions/2079

### DIFF
--- a/mill
+++ b/mill
@@ -48,4 +48,4 @@ fi
 unset MILL_DOWNLOAD_PATH
 unset MILL_VERSION
 
-exec $MILL_EXEC_PATH "$@"
+exec $MILL_EXEC_PATH -D "mill.main.cli=${0}" "$@"
--- a/mill
+++ b/mill
@@ -45,7 +45,11 @@ if [ ! -s "$MILL_EXEC_PATH" ] ; then
   unset MILL_DOWNLOAD_URL
 fi
 
+if [ -z "$MILL_MAIN_CLI" ] ; then
+  MILL_MAIN_CLI="${0}"
+fi
+
 unset MILL_DOWNLOAD_PATH
 unset MILL_VERSION
 
-exec $MILL_EXEC_PATH -D "mill.main.cli=${0}" "$@"
+exec $MILL_EXEC_PATH -D "mill.main.cli=${MILL_MAIN_CLI}" "$@"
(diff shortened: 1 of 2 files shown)


mill
```diff
@@ -48,4 +48,4 @@ fi
 unset MILL_DOWNLOAD_PATH
 unset MILL_VERSION
 
-exec $MILL_EXEC_PATH "$@"
+exec $MILL_EXEC_PATH -D "mill.main.cli=${0}" "$@"
```
ckipp01, line 51: Thanks for working on this @lefou! So quick question just to make sure I understand. In the situation where a Metals user just lets Metals generate the BSP config  (and let's say there is no millw in the repo), then Metals will use the millw it has in it's resources to do so. When we use the updated one, what is the actual value of `mill.main.cli` here then? I'm not sure I fully understand.

lefou: Good question. I also have a PR to [`millw`](https://github.com/lefou/millw) which is used by Metals. So, it's rather a question what the actual value of the shell variable `${0}` will look like, which depends on how `millw` it is handled by Metals. It may be a problem, if the script is invoked from an temporary (non-permatent) location.

* https://github.com/lefou/millw/pull/41

ckipp01: Let me use that pr from `millw` in Metals and just try it out. I'll report back.

lefou: Just as an idea, if there is need for a change: Metals could place the `millw` script inside `.metals` so it stays permanent for subsequent runs.

ckipp01: So this ended up being way harder to test than I thought due to how we do stuff in Metals. All that to say, we will however end up using `millw` inside of a temp directory that we create. I'm looking into now if we can instead just do as you mention and copy it into `.metals/millw` and just use it there.

lefou:
I don't think we risk much, by merging this.

ckipp01:
Yes, don't let me be a blocker here. I do have a draft pr locally that changes the Metals behavior and copies the `millw` into `.metals/millw`, and I think that will help. I'll polish that up soon and get that into Metals.

lefou:
@ckipp01 We could check for the existence of some other env var, to override the value, if needed.

lefou:
Or use an env var in the first place. If we only set when undefined, metals could even provide it's own preferred (stable) location to the script.

ckipp01:
btw opened https://github.com/scalameta/metals/pull/4586

PR #182, opened 2018-03-01
WIP: Play json build

rockjam:
This is work in progress, will add more commits. 

### What works now
- [X] Scalariform support
	- [X] simple reformat on compile
	- [X] check that everything is formatted, or fail otherwise
- [X] MiMa support
	- [X] check binary issues and fail build if there are some
- [X] build for both scala JVM and scala.js
- [X] compile/tests both for JVM and JS
- [x] compile/tests on scala 2.10, 2.11, 2.12.
- [x] automatically set headers
- [x] run jmh tests

### What else needs to be done(important first):
- [x] release task
- [ ] Omnidoc support

PR #2100, opened 2022-10-28
Add JMH contrib module

LaurenceWarne:
Hi, I was looking around for something providing JMH Mill integration along the lines of sbt-jmh, and I found a module in this PR: https://github.com/com-lihaoyi/mill/pull/182 (file link is here: https://github.com/com-lihaoyi/mill/pull/182/files#diff-191dd5ddc345255594500610c630c72beac8c28eacd593db2d55604b41c33efa).

I copied the script locally, and it works great OOTB, it appears I'm not the only person that's done this: https://github.com/jodersky/ustats/blob/master/jmh.sc :sweat_smile:, so I thought it might be useful if this was added as a contrib module.

This PR just adds the script (with minor modifications) as a module under contrib, along with tests and documentation (let me know if anything more is needed for defining a new contrib module).

Anyway, let me know what you think, thanks!

Also cc @rockjam, I know it was a long time ago, but would you have any problem with this? (I've added you as a co-auth)

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -890,7 +890,7 @@ object contrib extends MillModule {
     )
   }
 
-  object jmh extends MillModule {
+  object jmh extends MillInternalModule with MillAutoTestSetup with WithMillCompiler {
     override def compileModuleDeps = Seq(scalalib)
     override def testArgs = T {
       Seq(
--- a/build.sc
+++ b/build.sc
@@ -890,6 +890,15 @@ object contrib extends MillModule {
     )
   }
 
+  object jmh extends MillModule {
+    override def compileModuleDeps = Seq(scalalib)
+    override def testArgs = T {
+      Seq(
+        "-DMILL_SCALA_LIB=" + scalalib.runClasspath().map(_.path).mkString(",")
+      ) ++ scalalib.worker.testArgs()
+    }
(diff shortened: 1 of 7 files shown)


build.sc
```diff
@@ -890,6 +890,15 @@ object contrib extends MillModule {
     )
   }
 
+  object jmh extends MillModule {
```
lefou, line 893:
```suggestion
  object jmh extends MillInternalModule with AutoTestSetup with WithMillCompiler {
```

This might help to avoid the failing MiMa checks.

lefou (commented):
Currently, compilation of the new module fails. Can you check?

LaurenceWarne:
> Thank you for your contribution!
> 
> Currently, compilation of the new module fails. Can you check?

Ah sorry, compilation for java 8 should be fixed now.

LaurenceWarne:
Hmm compilation appears to be failing for windows: https://github.com/com-lihaoyi/mill/actions/runs/3392608569/jobs/5639002918#step:5:4897 :eyes:

lefou (approved):
Looks good to me. Thank you!

LaurenceWarne:
Thanks!

PR #2103, opened 2022-11-03
Gitlab maven repo error message

aheiska:
Print error message when token lookup in creating gitlab maven repository fails. Minor documentation improvements.

### DIFF
--- a/contrib/gitlab/src/mill/contrib/gitlab/GitlabPublishModule.scala
+++ b/contrib/gitlab/src/mill/contrib/gitlab/GitlabPublishModule.scala
@@ -101,8 +101,18 @@ trait GitlabMavenRepository {
   def repository: GitlabPackageRepository // For package discovery
 
   def mavenRepo(implicit context: Ctx): MavenRepository = {
-    val gitlabAuth = tokenLookup.resolveGitlabToken(context.log, context.env, sys.props.toMap).get
-    val auth = Authentication(gitlabAuth.headers)
-    MavenRepository(repository.url(), Some(auth))
+    val maybeRepository = tokenLookup
+      .resolveGitlabToken(context.log, context.env, sys.props.toMap)
+      .map(gitlabAuth => Authentication(gitlabAuth.headers))
+      .map(auth => MavenRepository(repository.url(), Some(auth)))
+
+    maybeRepository.getOrElse {
+      val indent = "      "
+      val tokenSearchPlaces = tokenLookup.tokenSearchOrder.mkString("", s",\n$indent", "\n")
+      val error =
+        s"Could not find Gitlab authentication for $repository, \n  Searched from:\n$indent$tokenSearchPlaces"
+      context.log.error(error)
+      throw new RuntimeException()
+    }
   }
 }
--- a/docs/antora/modules/ROOT/pages/Plugin_Gitlab.adoc
+++ b/docs/antora/modules/ROOT/pages/Plugin_Gitlab.adoc
@@ -153,6 +153,12 @@ Making mill to fetch package from gitlab package repository is simple:
 
 [source,scala]
 ----
+import mill._, scalalib._, mill.scalalib.publish._
+import coursier.MavenRepository
+import coursier.core.Authentication
+import $ivy.`com.lihaoyi::mill-contrib-gitlab:`
+import mill.contrib.gitlab._
+
 // DON'T DO THIS
 def repositoriesTask = T.task {
   super.repositoriesTask() ++ Seq(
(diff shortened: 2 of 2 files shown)


contrib/gitlab/src/mill/contrib/gitlab/GitlabPublishModule.scala
```diff
@@ -101,8 +101,18 @@ trait GitlabMavenRepository {
...
+      .map(gitlabAuth => Authentication(gitlabAuth.headers))
+      .map(auth => MavenRepository(repository.url(), Some(auth)))
+
+    maybeRepository.getOrElse {
+      val indent = "      "
+      val tokenSearchPlaces = tokenLookup.tokenSearchOrder.mkString("", s",\n$indent", "\n")
+      val error =
+        s"Could not find Gitlab authentication for $repository, \n  Searched from:\n$indent$tokenSearchPlaces"
+      context.log.error(error)
+      throw new RuntimeException()
```
lefou, line 115: Why not also add the error message to the exception?

aheiska: By default its a bit long and multiple lines and gets printed twice (first the log, then the exception). So when using from command line i though it was a bit spammy. 

I actually initially placed it only as exception parameter, but changed it to use context.log. Either way is fine for me, just thought that log would be bit easier to read from console.

lefou: TBH, I really hate exceptions without any message/context. Exception with stacktraces are already ugly, and beauty should be the leading criteria for leaving them out.  There may also exist situations, where the exception pops up, but the log message isn't visible.

If it's rather the one or the other, I vote for the exception message. The fact, that Mill is currently presenting the full stack trace instead of some nicer message may change.

aheiska: Yeah me to. Would prefer to return some type that is capable of presenting error state, but this is rather close to user and that would cause at least extra .get in users build.sc.

lefou: Instead, you could make it a task. Tasks have a proper result type. Also, you automatically get access to the correct context. 

```scala
def mavenRepo: Task[MavenRepository] = T.task {
  ...
}
```

Either just throw the exception or return a `mill.api.Result.Failure`.

https://com-lihaoyi.github.io/mill/mill/Tasks.html#_anonymous_tasks

aheiska: Just realized there are some logical duplicates. I'll do a bit larger cleanup. Hopefully on Monday (a bit busy weekend ahead).

lefou: @aheiska Gently ping.

aheiska: Sorry for being quiet. Suddenly got a lot more work to do, so open source halted for a while. I have a some implementation that is converted to Tasks somewhat ready. From what I can recall I got a compilation error and then the other issues started pressing. Branch is not on this computer, I'll push tomorrow. It wont compile (probably something minor), but I'd like to have feedback on solution.

aheiska:
Pushed a wip. It still has compilation error. Trying to get back to it as soon as possible.

aheiska:
This took like ages, but finally had time to put this together. Sorry for extreme delay. This ended up breaking backwards compatibility, but I assume there are not many users currently. Naming is no more coherent and hopefully more clear.

Added some tests. Improved documentation and now this provides nicer errors if token lookup fails.

lefou (commented):
Thanks for bringing this to the finish line. Breaking compatibility is currently no issue, as we are in a breaking window anyway.

One question. Why do you use a function `os.Path => os.Path` in `WorkspaceFile`? Wouldn't it be sufficient to use a `os.RelPath` or even `os.SubPath` and resolve it to an absolute path when it is actually used?

aheiska:
Thanks for review. I think function is there because of my poor knowledge of os-lib api (read: probably did not look at all..). RelPath or SubPath if surely the way to go. I'll fix this on monday when I get to computer where I have dev env set up.

lefou (approved):
Looks good to me. Thanks for your contribution and persistence!

aheiska:
Thank you for review and patience :). Hopefully this makes mill more useful also for others.

Issue #803, opened 2020-03-27
Protected commands fail at runtime with cryptic error


heksesang:
```scala
trait VersionFileModule extends Module {

  /** The file containing the current version. */
  def versionFile: define.Source = T.source(millSourcePath / "version")
  /** The current version. */
  def currentVersion: T[Version] = T { Version.of(os.read(versionFile().path)) }
  /** The release version. */
  def releaseVersion = T { currentVersion().asRelease }
  /** The next snapshot version. */
  def nextVersion(bump: String) = T.command { currentVersion().asSnapshot.bump(bump) }

  /** Writes the release version to file. */
  def setReleaseVersion = T {
    setVersionTask(releaseVersion)()
  }

  /** Writes the release version to file. */
  def setNextVersion(bump: String) = T.command {
    setVersionTask(nextVersion(bump))()
  }

  /** Writes the given version to file. */
  def setVersion(version: Version) = T.command {
    setVersionTask(T.task { version })()
  }

  protected def setVersionTask(version: T[Version]) = T.command {
    T.ctx.log.info(generateCommitMessage(version()))
    writeVersionToFile(versionFile(), version())
  }

  def writeVersionToFile(versionFile: mill.eval.PathRef, version: Version) =
    os.write.over(
      versionFile.path,
      version.toString
    )

  def generateCommitMessage(version: Version) =
    version match {
      case release: Version.Release => s"Setting release version to $version"
      case snapshot: Version.Snapshot => s"Setting next version to $version"
    }
}
```

If you try to run `mill versionFile.setVersion --version 0.3.0`, the above code will crash on runtime with a cryptic error:
```
Exception in thread "main" java.util.NoSuchElementException: None.get
        at scala.None$.get(Option.scala:366)
        at scala.None$.get(Option.scala:364)
        at mill.eval.Evaluator$$anonfun$3.applyOrElse(Evaluator.scala:468)
        at mill.eval.Evaluator$$anonfun$3.applyOrElse(Evaluator.scala:447)
        at scala.PartialFunction$Lifted.apply(PartialFunction.scala:228)
        at scala.PartialFunction$Lifted.apply(PartialFunction.scala:224)
        at mill.define.Graph$.$anonfun$groupAroundImportantTargets$1(Graph.scala:20)
        at mill.api.AggWrapper$Agg$Mutable.$anonfun$flatMap$1(AggWrapper.scala:70)
[...]

PR #2108, opened 2022-11-08
Use unique `out` path for private named tasks - fixes #2107

### LINKED ISSUE #2107: private targets name clashes overwrite outputs
## Problem
`private` target json files in the `out` directory use the name of the method as key. But when using traits you can have two different private targets that have the same name.

### Reproduction

Given this `build.sc` file:
```scala
import mill._
import mill.scalalib._

trait M1 extends Module {
  private def foo = T { "foo-m1" }
  def bar = T { foo() }
}

trait M2 extends Module {
  private def foo = T { "foo-m2" }
  def baz = T { foo() }
}

object mod extends M1 with M2
```

```
> mill show mod.bar
"foo-m1"
> mill show mod.baz
"foo-m1"
```

As you can see they both print `"foo-m1"` since the same `foo.json` is used for the two private targets in the two traits. This creates a name collision and Mill thinks that the target was already evaluated with `"foo-m1"` when in reality it is using the output from another trait.

lefou:
This change ensures, we always use the `super`-`out` path when we try to store metadata and scratch files for private named tasks. As private tasks can't be seen from other traits/classes in the same module hierarchy, our previous logic to detect potential collisions of the metadata location fails. Hence we now keep the `isPrivate` property in the named target.

I also added a test case that reproduce the initial issue but now succeeds.

* Fixes: https://github.com/com-lihaoyi/mill/issues/2107

This is a binary compatibility breaking change.

By applying the same approach and tracking also `public`-ness of a target, there is potential to fix also other issues like

* https://github.com/com-lihaoyi/mill/issues/803

This PR should be merged after 

* #2262

### DIFF
--- a/ci/mill-bootstrap.patch
+++ b/ci/mill-bootstrap.patch
@@ -1,0 +1,119 @@
+diff --git a/build.sc b/build.sc
+index be462bb7c..50eb14412 100644
+--- a/build.sc
++++ b/build.sc
+@@ -2,22 +2,10 @@
+ import $file.ci.shared
+ import $file.ci.upload
+ import $ivy.`org.scalaj::scalaj-http:2.4.2`
+-import $ivy.`de.tototec::de.tobiasroeser.mill.vcs.version_mill0.10:0.3.0`
+-import $ivy.`com.github.lolgab::mill-mima_mill0.10:0.0.13`
+ import $ivy.`net.sourceforge.htmlcleaner:htmlcleaner:2.25`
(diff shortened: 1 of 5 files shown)


lefou:
As this is a binary breaking change, the bootstrap tests currently fail. I need to prepare and apply some bootstrap patch, so we don't use external incompatible plugins.

lefou:
I added the bootstrap patch, all tests succeed now.

PR #2110, opened 2022-11-08
BSP: implement buildTarget/OutputPaths request

lefou:
Also bumped bsp4j to 2.1.0-M3


lefou:
Waiting for a BSP client that supports this feature, to try it out, as we don't have a test suite that can validate it.

Issue #1910, opened 2022-07-01
Checksum failure on Github CI while fetching dependencies 


lefou:
Sometimes, Mill can't fetch some dependencies. I'm not able to reproduce locally, but it can be seen occasionally on Github CI, e.g. in this run: https://github.com/com-lihaoyi/mill/runs/6181553905?check_suite_focus=true

The error message looks like this:

```
1 targets failed
scalalib.scalaDocClasspath java.nio.file.NoSuchFileException: /home/runner/.cache/coursier/v1/https/repo1.maven.org/maven2/net/java/dev/jna/jna/5.9.0/.jna-5.9.0.jar__sha1.computed1274112506791367117.tmp -> /home/runner/.cache/coursier/v1/https/repo1.maven.org/maven2/net/java/dev/jna/jna/5.9.0/.jna-5.9.0.jar__sha1.computed
```

An retry typically fixes the issue.

Although the issue can be seen now and then on Github CI, the dependency on which it actually fails is pretty random.

<details>
<summary>Expand for a more complete log...</summary>

```
2022-04-26T18:42:35.2215843Z + ./mill -i -j 0 installLocal --binFile target/mill-1
2022-04-26T18:42:36.3942103Z Preparing Java 11.0.15 runtime; this may take a minute or two ...
2022-04-26T18:42:44.4974024Z Compiling /home/runner/work/mill/mill/ci/shared.sc
2022-04-26T18:42:49.3204603Z Compiling /home/runner/work/mill/mill/ci/upload.sc
2022-04-26T18:42:50.5143058Z Compiling /home/runner/work/mill/mill/build.sc
2022-04-26T18:43:08.5515122Z [#1] [2/2629] bsp.scalaVersion 
...
2022-04-26T18:43:12.0530854Z [1A[2K[9999D[#0] [1133/2629] bsp.scalaDocClasspath 
2022-04-26T18:43:12.0538027Z [1A[2K[9999D[#1] [1134/2629] scalalib.scalaDocClasspath 
2022-04-26T18:43:12.0625080Z [1A[2K[9999D[#0] [1133/2629] bsp.scalaDocClasspath | Downloading [1/1] artifacts (~0/0 bytes)
2022-04-26T18:43:12.0654554Z [1A[2K[9999D[#0] [1133/2629] bsp.scalaDocClasspath | Downloading [2/2] artifacts (~0/0 bytes)
[...]

Issue #1028, opened 2020-12-03
randomly failing dependency downloads. concurrent download


axaluss:
I'm seeing randomly failing dependency downloads since upgrade to 0.9.3 

```
.resolvedIvyDeps Failed to load source dependencies
   concurrent download: https://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-databind/2.2.3/jackson-databind-2.2.3.jar
```
another run
```
Failed to load source dependencies
   concurrent download: https://repo1.maven.org/maven2/io/netty/netty-all/4.0.23.Final/netty-all-4.0.23.Final.jar
```
another run
```
resolvedRunIvyDeps Failed to load source dependencies
   concurrent download: https://repo1.maven.org/maven2/org/apache/kafka/kafka_2.10/0.10.1.0/kafka_2.10-0.10.1.0.jar
   concurrent download: https://repo1.maven.org/maven2/org/rocksdb/rocksdbjni/4.9.0/rocksdbjni-4.9.0.jar
```
By chance eventually it resolves.

lefou:
There are some concurrency issues with current coursier. E.g. https://github.com/coursier/coursier/issues/1815#issuecomment-736738585 or https://github.com/coursier/coursier/issues/1818. Question is, if we hit one of the existing or even a new one.

axaluss:
in my build setup also only scala 2.11 builds were affected. 2.12 was fine

lefou:
This issue is even more close to ours: https://github.com/coursier/coursier/issues/2022

lefou:
Unfortunately, this isn't fixed by #1510.

lefou:
Here is a reproducer:

```scala
import mill._, scalalib._

object foo extends Cross[Foo](1.to(20).map(_.toString): _*)
class Foo(nr: String) extends JavaModule {
  def ivyDeps = Agg(ivy"org.slf4j:slf4j-api:1.7.25")
}
```

```bash
$ rm -r ~/.cache/coursier/v1/https/repo1.maven.org/maven2/org/slf4j/slf4j-api/1.7.25/ ; mill -i clean ; mill -i -j 16 -d __.resolvedIvyDeps
No mill version specified.
You should provide a version via '.mill-version' file or --mill-version option.
Using mill version 0.10.1
Compiling /home/lefou/work/opensource/mill/repro-parallel-fetch/build.sc
[1/1] clean 
No mill version specified.
[...]

PR #2112, opened 2022-11-09
Catch even more coursier concurrency glitches, some consolidations

lefou:
This is an extension to our workarounds to coursier concurrency issues.

We handle some issues, already detected by courier and reported as a result, as well as one which is not detected by coursier, so we try-catch it and try to match based on the exception type and the message.

Some links:
* https://github.com/com-lihaoyi/mill/issues/1028
* https://github.com/com-lihaoyi/mill/issues/1910
* https://github.com/coursier/coursier/issues/2022
* https://github.com/coursier/coursier/issues/1815

As it is hard to reproduce such kind or issues, I created a GitHub Actions workflow, which consistently failed without this very PR in step "1. Run Mill with settings that should stress coursier". After I applied this PR, all steps passed. That's not a 100-percent guarantee, but it means, we're probably right on track.

```yaml
name: Stress test coursier downloads to reproduce concurrency issues

on:
  push:
  pull_request:

jobs:

  test-coursier:
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v3
        with:
          fetch-depth: 0

      - uses: actions/setup-java@v3
        with:
          java-version: 11
          distribution: temurin

      - name: "Install local Mill into target and ivy cache"
        run: ./mill -i installLocal

      - name: "1. Cleanup courier cache"
        run: rm -r $HOME/.cache/coursier

      - name: "1. Run Mill with settings that should stress coursier"
        run: target/mill-release -i -j 20 -d __.prepareOffline

      - name: "2. Cleanup courier cache"
        run: rm -r $HOME/.cache/coursier

      - name: "2. Run Mill with settings that should stress coursier"
        run: target/mill-release -i -j 20 -d __.prepareOffline

```

### DIFF
--- a/main/src/mill/modules/Jvm.scala
+++ b/main/src/mill/modules/Jvm.scala
@@ -12,7 +12,7 @@ import java.io.{
 }
 import java.lang.reflect.Modifier
 import java.net.URI
-import java.nio.file.{FileSystems, Files, StandardOpenOption}
+import java.nio.file.{FileSystems, Files, NoSuchFileException, StandardOpenOption}
 import java.nio.file.attribute.PosixFilePermission
 import java.util.jar.{Attributes, JarEntry, JarFile, JarOutputStream, Manifest}
 import coursier.{Dependency, Repository, Resolution}
@@ -27,7 +27,7 @@ import mill.modules.Assembly.{AppendEntry, WriteOnceEntry}
 import scala.collection.mutable
 import scala.util.Properties.isWin
 import scala.jdk.CollectionConverters._
-import scala.util.Using
+import scala.util.{Failure, Success, Try, Using}
 import mill.BuildInfo
 import os.SubProcess
 import upickle.default.{ReadWriter => RW}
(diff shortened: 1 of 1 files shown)


alexarchambault:
@lefou Is the repository you ran the workflow above public?

alexarchambault:
@lefou Thanks, I can reproduce locally the "concurrent download" errors quite consistently.

These are triggered by downloads happening in the same JVM. I'm wondering about one thing: do the actual in-the-wild issues also tend happen in that configuration (parallel downloads in the same JVM), or do they tend to happen when several JVMs try to download stuff?

Even though both cases should work (or be fixed for it), same JVM and different JVM concurrency isn't handled the same way under-the-hood here.

lefou:
Most issues I've seen originate from running concurrent jobs in the same JVM. Mostly on CI, as these often start with empty or outdated caches. Especially the checksum issues also occurred without any concurrency (outside of coursier) in the same JVM. That's what this PR is trying the fix for Mill. E.g. in Mill, varying CI job needed a restart occasionally.

PR #2114, opened 2022-11-09
Try to safely encode foreign module paths which can't made relative

lefou:


lefou:
This PR lacks a test case. Contributions welcome!

lihaoyi:
Should be rendered moot since we only allow `package.mill` files in subfolders now

PR #2126, opened 2022-11-16
scalapblib- support specifying individual source files

oridag:
Currently the ScalalPB plugins only allows specifying whole directories as source. This change allows specifying individual files as inputs.
In addition, ScalaPBC is now invoked once for all input paths instead of separate invocation for each input path. This is important because it allows setting package scoped options to take affect across all input paths.

### DIFF
--- a/contrib/scalapblib/src/ScalaPBWorker.scala
+++ b/contrib/scalapblib/src/ScalaPBWorker.scala
@@ -25,7 +25,7 @@ class ScalaPBWorker extends AutoCloseable {
 
         val instance = new ScalaPBWorkerApi {
           override def compileScalaPB(
-              root: File,
+              roots: Seq[File],
               sources: Seq[File],
               scalaPBOptions: String,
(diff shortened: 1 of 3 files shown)


contrib/scalapblib/src/ScalaPBWorker.scala
```diff
@@ -108,7 +106,7 @@ class ScalaPBWorker extends AutoCloseable {
 
 trait ScalaPBWorkerApi {
   def compileScalaPB(
-      root: File,
```
lefou, line 111: I suggest to keep an overloaded (maybe deprecated) method with the old signature (it can simply forward to the new implementation though), to keep binary compatibility and avoid breaking someones workflow. You never know, who is already using or integrating this module.

lefou (commented):
Looks reasonable to me. I added one remark inline.

lefou (approved):
Looks good to me. Thank you!

PR #2127, opened 2022-11-17
Safe Linux home install.

wiltonlazary:
It's is a good pratice to install softwares on the user "HOME" directory.


lefou (approved):
Looks reasonable. Thank you!

Issue #2130, opened 2022-11-19
Foreign modules in `T.workspace` are not stored in `out/foreign-modules/`


lolgab:
If you have to `.sc` files in the root, the other files modules are not stored in `out/foreign-modules`, colliding when the modules have the same name.
Example:
```scala
// build.sc
import mill._
import mill.scalalib._
import $file.other

object foo extends JavaModule {
  def moduleDeps = Seq(other.foo)
}
```

```scala
// other.sc
import mill._
import mill.scalalib._
import $file.other

object foo extends JavaModule {
  // EDIT: the following line needs to be commented out
  def moduleDeps = Seq(other.foo)
}
```

lefou:
You find me puzzled. Since when is it possible to declare Mill modules in objects in non-`build.sc` files? Maybe, I'm remembering it wrong, but I'm pretty sure it didn't work some major versions ago. How do you invoke these targets? 

So, after the first shock, back to the point. I think you posted the wrong content for `other.sc`, as it is including and referring to itself creating an infinite loop and resulting in a `StackOverflowError`. 

The right solution depends somewhat on whether we stick with Ammonite or go an alternative route, which is based on a new `MillBuildModule` (see https://github.com/com-lihaoyi/mill/pull/1972). In the latter case, having two `object foo` in the same scope might be a compile error and we don't have to deal with it in Mills task naming and caching.

If we want to stick with Ammonite (or just want a fix as soon as possible), we need to encode the name of the include file name. We currently only do this, if its location is different from `T.workspace`: https://github.com/com-lihaoyi/mill/blob/2a28295eb9f3fc4f33e2afa62b3daeb6c697864e/main/src/mill/main/MainRunner.scala#L192

lolgab:
> How do you invoke these targets?

I don't. I use other files to put modules that contain generated code or something I don't have to interact with, so the `build.sc` is cleaner, takes less time to compile, and also I can take advantage of the precise build cache invalidation I implemented in Mill `0.10.1`.

> The right solution depends somewhat on whether we stick with Ammonite or go an alternative route, which is based on a new MillBuildModule (see https://github.com/com-lihaoyi/mill/pull/1972). In the latter case, having two object foo in the same scope might be a compile error and we don't have to deal with it in Mills task naming and caching.

[...]

PR #2132, opened 2022-11-21
Always give included files a unique foreign module path

lefou:
## What is changed

With this change, we always give included files a unique foreign module path. This makes the dest path more consistent and robust.

Unfortunately, this slightly changes the encoding of foreign modules names. This is no issue when writing `.sc` files, but it will be visible in the progress output and in BSP client and IntelliJ Idea. E.g. the name of an included file in directory `proj1` containing a module `proj1`, will now have the module name "foreign-modules.proj1.build.proj1", whereas previously it was "foreign-modules.proj1.proj1".

* Fix https://github.com/com-lihaoyi/mill/issues/2130

## Reproduction of the new fixed issue (reported in #2130)

If you import `.sc` files from the same directory, the cache location is shared with `build.sc` , resulting in colliding dest paths when modules have the same name.

Example:
```scala
// build.sc
import mill._
import $file.other

def foo = T {
  println(s"build.foo: ${T.dest}")
  "build.foo"
}

def bar = T{
  other.foo()
}
```

```scala
// other.sc
import mill._

def foo = T {
  println(s"other.foo: ${T.dest}")
  "other.foo"
}
```

Now the cache location is re-used and overridden by different targets.

```
$ mill show bar
[1/1] show 
[1/1] show > [1/2] foo 
other.foo: /tmp/mill-1/out/foo.dest
"other.foo"

$mill show foo
[1/1] show 
[1/1] show > [1/1] foo 
build.foo: /tmp/mill-1/out/foo.dest
"build.foo"
```

### DIFF
--- a/main/test/resources/examples/foreign/project/build.sc
+++ b/main/test/resources/examples/foreign/project/build.sc
@@ -65,25 +65,28 @@ def checkProjectDests = T {
 
 def checkInnerDests = T {
   val foreignOut: os.Path = millSourcePath / "out" / "foreign-modules"
-  assertPathsEqual(inner.build.sub.selfDest(), foreignOut / "inner" / "sub")
-  assertPathsEqual(inner.build.sub.sub.selfDest(), foreignOut / "inner" / "sub" / "sub")
+  assertPathsEqual(inner.build.sub.selfDest(), foreignOut / "inner" / "build" / "sub")
+  assertPathsEqual(inner.build.sub.sub.selfDest(), foreignOut / "inner" / "build" / "sub" / "sub")
 }
 
 def checkOuterDests = T {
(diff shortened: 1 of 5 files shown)


main/test/resources/examples/foreign/project/build.sc
```diff
@@ -56,6 +57,13 @@ def checkOuterInnerPaths = T {
   )
 }
 
+def checkOtherPaths = T {
```
lolgab, line 60: Is this used? I can't find any usage in this PR

lefou: Good catch. This slipped through in my last selective commit.

lolgab (approved):
You could run scalafmt on these files if you want, otherwise LGTM

PR #2141, opened 2022-11-24
Support incremental compilation in Scala Native 0.4.9

lolgab:
Scala Native 0.4.9 added support for this feature: https://github.com/scala-native/scala-native/releases/tag/v0.4.9

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -104,10 +104,10 @@ object Deps {
   }
 
   object Scalanative_0_4 {
-    val scalanativeTools = ivy"org.scala-native::tools:0.4.8"
-    val scalanativeUtil = ivy"org.scala-native::util:0.4.8"
-    val scalanativeNir = ivy"org.scala-native::nir:0.4.8"
-    val scalanativeTestRunner = ivy"org.scala-native::test-runner:0.4.8"
+    val scalanativeTools = ivy"org.scala-native::tools:0.4.9"
+    val scalanativeUtil = ivy"org.scala-native::util:0.4.9"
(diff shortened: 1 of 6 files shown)


lefou (commented):
Nice addition. Would be good to seen it working in some test, tough.

lolgab:
> Nice addition. Would be good to seen it working in some test, tough.

I thought about that. Since it doesn't change the binary behaviour, it's hard to assert it's working without assumptions that will probably break on the next version (since it's an experimental feature). Also Scala Native tests are expensive. If you think we need it I can add it. I just manual tested it in scratch.

lefou:
> > Nice addition. Would be good to seen it working in some test, tough.
> 
> I thought about that. Since it doesn't change the binary behaviour, it's hard to assert it's working without assumptions that will probably break on the next version (since it's an experimental feature). Also Scala Native tests are expensive. If you think we need it I can add it. I just manual tested it in scratch.

I see. As I'm not familiar with Scala Native API, I can't tell whether it is easy, possible ,or useless, but maybe we can add some unit test just comparing the expected with the actual used config?

lolgab:
@lefou Added a test that exercises the incremental compilation feature in Scala Native (the `scala.ll` file is created only if you set `nativeIncrementalCompilation = true` otherwise the test fails.

lefou (approved):
Looks good to me. Thanks for adding the extra test.

I wonder, whether we want to print a warning message, when a feature is enabled but ignored, because of an unsupported version.

Issue #2144, opened 2022-11-25
Make JsEnv extensible


nafg:
In Mill, the JsEnv selection is based on a sealed trait with only 3 choices: NodeJs, JsDom, and Phantom (which is an obsolete project). But there are many cases when one would want to use a different strategy. As far as I can tell, currently the only way to accomplish that now is to override `testTask` and do _everything_ from scratch, because all of the scalajs APIs are private.

One use case for this is described in https://github.com/scala-js/scala-js-env-jsdom-nodejs/issues/44, where @sjrd suggests using https://github.com/exoego/scala-js-env-jsdom-nodejs. Hopefully that project will continue to be maintained, but in any case it's out of reach from Mill, as far as I can tell.

I think it would need a major redesign of the APIs. IIUC (mentioned in another issue) the reason for the current design is to enable one project to use multiple versions of Scala.js, which is a great feature that even SBT doesn't support. I don't know if it's possible to have both at the same time (user-supplied JsEnv and multiple scalajs versions in one project).

But I don't see any other way to use jsdom to test code that uses modules.

sjrd:
> IIUC (mentioned in another issue) the reason for the current design is to enable one project to use multiple versions of Scala.js, which is a great feature that even SBT doesn't support. I don't know if it's possible to have both at the same time (user-supplied JsEnv and multiple scalajs versions in one project).

Since Scala.js 1.1.1, the `JSEnv` APIs come from separate artifacts with their own, backwards binary compatible versioning. So Mill could always use the latest version of `scalajs-js-envs` without compromising its ability to link with multiple versions of Scala.js.

lefou:
I think some internal used code is private to avoid the maintenance costs of preserving binary compatibility, which we take seriously in Mill. But it's not set in stone, so if you come up with a good use case and a proper suggestion which API you need to change and how, we can change / improve it.

You need to be a bit more specific though, as your current description isn't actionable for me. 

Also, I admit that Scala.js is not my field of expertise, so you better be verbose when describing the needed changes. I would prefer some draft pull request, so we can focus on concrete code and the impact of changes in terms of tests.

That being said, I'm very open to support more use cases and looking forward to it.

lolgab:
@sjrd I tried in the past to use the Scala.js API directly, but when changing version of Scala.js it was breaking the classpath. Maybe if we compiled with Scala.js `1.0.0` but then we couldn't use the newer features. If we compile against the latest versions then I remember that the older versions were breaking.
[...]

Issue #1036, opened 2020-12-06
Support `forkWorkingDir` in `ScalaJSModule` (`run`, `test`, ...)


lolgab:
Working with multiple node.js subprojects, it is needed to have different installation directories that carry their own `node_modules` folder.
Currently it is not possible to change set the pwd for a subproject, so tests always run by using the root project directory as pwd.
This is probably something that can be useful on a more general level than the `ScalaJSModule` since it can be useful for Scala and Java projects too.

lefou:
In general, mill supports changing the working directory for forked tests and in general forked runs. This is controlled via the `JavaModule.forkWorkingDir` target. But it looks like (as you have noted yourself on gitter channel) that this is not used in `ScalaJSModule` to run tests.

nafg:
I can send a PR if I'm told how to do it

lefou:
I think, you need to add a working dir parameter to `ScalaJSWorker.run` method and plumb it into the right tools invocations in the implementation. Once, that is done, all usages of the worker in `ScalaJSModule` need to use the `forkWorkingDir` target to feed that parameter. This is the `run` target, and maybe others.

nafg:
I started writing a longer comment but tldr this requires a breaking change to scalajs-js-envs.

But I just realized that it may not be necessary for my use case. The original problem was how to get Node to use a generated `node_modules` somewhere under `out/`. And I had tried using the `NODE_PATH` environment variable and it didn't seem to work, and AFAICT nobody on the internet uses it this way, and not being able to set the working directory seems like an arbitrary limitation that I assumed would easy to fix, so I started thinking the solution was in that direction.

But I realized that I was probably doing it wrong, setting it to the directory containing `node_modules`. I think instead it should point to the actual `node_modules`, as implied in https://nodejs.org/api/modules.html#loading-from-the-global-folders.

So maybe I can get ScalaJS tests to run without bundling that way.

lefou:
> I started writing a longer comment but tldr this requires a breaking change to scalajs-js-envs.

Can you elaborate on this? Why needs it a change in an upstream dependency? This issue is about defining the working (or start) directory of the js/node process. Everything else seems unrelated to this issue.

[...]

PR #2147, opened 2022-11-25
Support exoego fork of JsDom JsEnv

lolgab:
My first approach to solve #2144 was to create a `Custom` `JsEnvConfig` which would take a `className` and then the implementation would instantiate that class with java reflection. It worked fine but didn't support parameters or builders used by the JsEnv to create itself.
Since there aren't many custom `JsEnv`s out there (the only one I know is [exoego](https://github.com/exoego/scala-js-env-jsdom-nodejs)'s fork of jsdom-nodejs, I'm proposing of supporting it officially in Mill.
This PR also changes the worker classpath to only download and load the jsEnv dependencies that are needed. This skips downloading artifacts and loading classes for jsEnvs that aren't used, like this new one for people not using it, or the deprecated phantom jsEnv.
 
Testing this properly means installing `jsdom` which can only be done in the root because of https://github.com/com-lihaoyi/mill/issues/1036 (to my understanding). I tested in manually in scratch but I needed to have a global `package.json`

### DIFF
--- a/scalajslib/src/mill/scalajslib/ScalaJSModule.scala
+++ b/scalajslib/src/mill/scalajslib/ScalaJSModule.scala
@@ -5,7 +5,7 @@ import ch.epfl.scala.bsp4j.{BuildTargetDataKind, ScalaBuildTarget, ScalaPlatform
 import mill.api.{Loose, PathRef, Result, internal}
 import mill.scalalib.api.ZincWorkerUtil
 import mill.scalalib.Lib.resolveDependencies
-import mill.scalalib.{DepSyntax, Lib, TestModule}
+import mill.scalalib.{Dep, DepSyntax, Lib, TestModule}
 import mill.testrunner.TestRunner
 import mill.define.{Command, Target, Task}
 import mill.scalajslib.{ScalaJSWorker => DeprecatedScalaJSWorker}
@@ -50,6 +50,21 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
     )
   }
 
+  def scalaJSJsEnvIvyDeps: Target[Agg[Dep]] = T {
+    val dep = jsEnvConfig() match {
+      case _: JsEnvConfig.NodeJs =>
+        ivy"${ScalaJSBuildInfo.Deps.scalajsEnvNodejs}"
+      case _: JsEnvConfig.JsDom =>
+        ivy"${ScalaJSBuildInfo.Deps.scalajsEnvJsdomNodejs}"
+      case _: JsEnvConfig.ExoegoJsDomNodeJs =>
+        ivy"${ScalaJSBuildInfo.Deps.scalajsEnvExoegoJsdomNodejs}"
+      case _: JsEnvConfig.Phantom =>
+        ivy"${ScalaJSBuildInfo.Deps.scalajsEnvPhantomJs}"
+    }
+
+    Agg(dep)
+  }
+
   def scalaJSLinkerClasspath: T[Loose.Agg[PathRef]] = T {
     val commonDeps = Seq(
       ivy"org.scala-js::scalajs-sbt-test-adapter:${scalaJSVersion()}"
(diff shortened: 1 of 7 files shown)


nafg:
Thank so much!

I think @sjrd mentioned that there's a jsenv for using selenium. And in some ticket it was implied that users writing their own jsenv was a normal use case.

lolgab:
> Thank so much!
> 
> I think @sjrd mentioned that there's a jsenv for using selenium. And in some ticket it was implied that users writing their own jsenv was a normal use case.

Supporting custom jsEnvs is harder and the only solution I have at the moment is clunky, which makes me prefer to support them in the core repository.
We can re-discuss this again when we are in bin-compat breaking window and we dropped support for Scala.js 0.6.
Also if something is supported doesn't mean that there are actual users writing their own private `JsEnv`s implementations and are also using Mill. When these users appear, we can talk and understand what to do. Adding a partial feature that uses reflection and is ugly for no-users is a no for me.

lolgab:
I've just noticed the [scala-js-env-selenium](https://github.com/scala-js/scala-js-env-selenium) which I can add to this PR as well.

scalajslib/src/mill/scalajslib/ScalaJSModule.scala
```diff
@@ -50,6 +50,21 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
     )
   }
 
+  def scalaJSJsEnvIvyDeps: Target[Agg[Dep]] = T {
```
lolgab, line 53: I'm very happy with this change which will speed up downloading dependencies in Scala.js builds, since now Mill downloads only the dependency you need instead of downloading all of them :)

lefou (commented):
Nice one. I like that change too.

Would it be possible to add an additional `Custom` config, which requires the user to define the dependency as well as the executable and the parameters? That way, we don't need to change whenever a user wants to play with a new `JsEnv`. It might be not as safe, tested and convenient as the supported types, but also don't stand in the way.

lolgab:
> Nice one. I like that change too.
> 
> Would it be possible to add an additional `Custom` config, which requires the user to define the dependency as well as the executable and the parameters? That way, we don't need to change whenever a user wants to play with a new `JsEnv`. It might be not as safe, tested and convenient as the supported types, but also don't stand in the way.

I tried but the problem is that `executable` and `parameters` is not something generic that comes from `JsEnv` but something specific to the node jsenv..
If you [see the selenium one](https://github.com/scala-js/scala-js-env-selenium/blob/main/seleniumJSEnv/src/main/scala/org/scalajs/jsenv/selenium/SeleniumJSEnv.scala#L49) (which is the only official one we don't support yet), it has a completely different set of parameters.

lefou:
> > Nice one. I like that change too.
> > Would it be possible to add an additional `Custom` config, which requires the user to define the dependency as well as the executable and the parameters? That way, we don't need to change whenever a user wants to play with a new `JsEnv`. It might be not as safe, tested and convenient as the supported types, but also don't stand in the way.
> 
> I tried but the problem is that `executable` and `args` is not something generic that comes from `JsEnv` but something specific to the node jsenv.. If you [see the selenium one](https://github.com/scala-js/scala-js-env-selenium/blob/main/seleniumJSEnv/src/main/scala/org/scalajs/jsenv/selenium/SeleniumJSEnv.scala#L49) (which is the only official one we don't support yet), it has a completely different set of parameters.

This means we need to provide some factory function which is actually creating the `JsEnv` instance. But as this function needs access to the implementation via the current classpath, we can't provide it in the `build.sc`. Is this understanding correct?

Here is an idea. Maybe, we can delegate to some factory on the freshly build classpath. E.g. the user implements an factory (`() => JsEnv` or `(someContextProvidedByMill) => JsEnv` ) which is returning a customized instance of `JsEnv` and we just run it to get the instance. As this factory is in the sources of the (test) module, the compile classpath already contains the specific `JsEnv`.

lefou (approved):
Looks good to me. I'll merge as is. Thank you!

Potential generic support for other `JsEnv`s can be discussed and implemented independently.

PR #2148, opened 2022-11-26
Add missing backticks

kubukoz:
Currently, the sentence is rendered wrong - I think the extra pair of ticks will make it work like the "Targets" in the previous sentence.

image


lefou (approved):
Thank you!

PR #2154, opened 2022-11-29
docs: Fix favicon

nafg:
Currently [the docs](https://com-lihaoyi.github.io/mill/mill/Intro_to_Mill.html) have no favicon. This fixes that to use , which was probably the original intention


nafg:
Updated description

lefou (approved):
Nice. Thank you!

PR #2164, opened 2022-12-03
Bloop: Fixed re-generation of bloop config dir

lefou:
The `bloop.writeConfig` was a cached target, but it generated files outside of `T.dest`. Instead, it should be a command.

To preserve source and binary compatibilty, I created a new command `writeConfigFile`, and let the `writeConfig` target forward the result of `writeConfigFile`.


adpi2:
Thanks for taking care of that

adpi2 (approved):
LGTM

One question though: Would Mill remove old config files for modules that no longer exists?

lefou:
> LGTM
> 
> One question though: Would Mill remove old config files for modules that no longer exists?

Yes. It's simply deleting all other `*.json` files it finds in the bloop config dir.

PR #2180, opened 2022-12-09
Pass more settings from `ScalaJSModule` to its `Tests`

lolgab:
By default I expect that tests run with the same Scala.js settings as the parent.


lefou (approved):
Looks good to me. Thank you!

PR #2182, opened 2022-12-09
Pass more settings from `ScalaJSModule` to its `Tests` (Backport of #2180) 

lolgab:
* Backport of:  https://github.com/com-lihaoyi/mill/pull/2180

### DIFF
--- a/scalajslib/src/mill/scalajslib/ScalaJSModule.scala
+++ b/scalajslib/src/mill/scalajslib/ScalaJSModule.scala
@@ -34,6 +34,11 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
     override def scalaVersion = outer.scalaVersion()
     override def scalaJSVersion = outer.scalaJSVersion()
     override def moduleDeps = Seq(outer)
+    override def moduleKind = outer.moduleKind()
+    override def moduleSplitStyle = outer.moduleSplitStyle()
+    override def esFeatures = outer.esFeatures()
+    override def jsEnvConfig = outer.jsEnvConfig()
+    override def scalaJSOptimizer = outer.scalaJSOptimizer()
   }
 
   def scalaJSBinaryVersion = T { ZincWorkerUtil.scalaJSBinaryVersion(scalaJSVersion()) }


lolgab:
I think we should merge this kind of PRs with rebase strategy. Right @lefou ?
I ran:
```
git cherry-pick 89ffd1ce6da5199cfa1016b9f1adf67b81683f66
```

lefou:
> I think we should merge this kind of PRs with rebase strategy. Right @lefou ? I ran:
> 
> ```
> git cherry-pick 89ffd1ce6da5199cfa1016b9f1adf67b81683f66
> ```

I'm not quite sure what you mean with that. I'm going to merge this PR as any other to the 0.10.x branch.

lolgab:
I meant that if you merge with rebase we will have a nicer history in the 0.10.x branch since the commit comes from an already merged PR. if we rebase we have the same exact commit in both branches, which makes it easier to understand the differences between the two branches.

lefou:
> I meant that if you merge with rebase we will have a nicer history in the 0.10.x branch since the commit comes from an already merged PR. if we rebase we have the same exact commit in both branches, which makes it easier to understand the differences between the two branches.

The only change to have the exact same commit (with same commit hash) is to rewrite history, which we obviously won't do. It's also impractical. I merged #2180 with "Squash and merge" which produces a nice linear history. If I do the same with this PR, we have also a single commit with almost identical name, but in the comment, there is the additional reference to the original pull request.

lolgab:
> Is that good enough?

It probably is, I was wondering if merging with rebase strategy would be better, since the two branch would share the same exact commit. But it's a detail. Do what you think it's best 👌

lefou:
> > Is that good enough?
> 
> It probably is, I was wondering if merging with rebase strategy would be better, since the two branch would share the same exact commit. But it's a detail. Do what you think it's best ok_hand

I start to see what you probably mean. The thing is, once you made that commit with `git cherry-pick` it is already a new commit, since it has a different parent. It contains the same change, but it's no longer the same commit (Other hash and time).

A rebase will never work. The only chance to keep an identical commit is a merge with merge-commit.

PR #2183, opened 2022-12-09
Use only classpath paths strings to cache Scala.js worker

lolgab:
This makes the `bridge` method take 15ms instead of 600ms It is safe since the filenames contain the Scala.js version, so if you switch version the hashCode will be different as expected


scalajslib/src/mill/scalajslib/worker/ScalaJSWorker.scala
```diff
@@ -12,14 +12,15 @@ import mill.{Agg, T}
 private[scalajslib] class ScalaJSWorker extends AutoCloseable {
   private var scalaJSWorkerInstanceCache = Option.empty[(Long, workerApi.ScalaJSWorkerApi)]
 
-  private def bridge(toolsClasspath: Agg[os.Path])(implicit ctx: Ctx.Home) = {
-    val classloaderSig =
-      toolsClasspath.map(p => p.toString().hashCode + os.mtime(p)).sum
+  private def bridge(toolsClasspath: Agg[mill.PathRef])(implicit ctx: Ctx.Home) = {
+    // toolsClasspath is stable and made of external libraries which have the versions
+    // in the file names. So we use the paths string to hash since it's ~50x faster
+    val classloaderSig = toolsClasspath.iterator.map(p => p.path.hashCode).sum
```
lefou, line 18: I think using the `hashCode` of the `PathRef` is a better option. Using only the name is not guaranteed to be stable. E.g. you could dynamically download to a same local path `T.dest / "out.jar"` and would miss any change when this jar changes later.

lolgab:
I naively thought that calculating the hashcode of `PathRef`s was more expensive than of `Path`s but the price of `PathRef`s is on creation. Once you have it, you don't need to perform any file system operation to calculate the hashCode, which is great. Also the hashCode of a `Agg` is exactly `.map(_.hashCode()).sum` so we can use that directly.

lefou (commented):
This looks good to me. After you update the PR title and description, I think we are good to go.

lolgab:
@lefou Done! Thank you for the review! I'm so happy about these performance improvements. Mill is noticeably faster!

lefou:
> @lefou Done! Thank you for the review! I'm so happy about these performance improvements. Mill is noticeably faster!

Do you have any numbers?

lefou (approved):
Nice change! Less code, better performance.

PR #2185, opened 2022-12-09
Use classpath PathRefs hashCode as cache key for Zinc worker

lolgab:
The previous implmentation was receiving `os.Path`s so it needed to access the filesystem to know if the files did change.
Now we propagate `PathRef`s so we simply use their `hashCode` which changes if the files did change.
Here some numbers collected by compiling a Scala file 176 times and writing down the time calculating `compilersSig` took.
|               | Before  |After  |
|:--------------|:--------|:------|
|    Average    | 1364 ms | 12 ms |
| Std deviation | 6533.41 | 91.35 |

### DIFF
--- a/scalalib/api/src/mill/scalalib/api/ZincWorkerApi.scala
+++ b/scalalib/api/src/mill/scalalib/api/ZincWorkerApi.scala
@@ -1,6 +1,6 @@
 package mill.scalalib.api
 
-import mill.api.CompileProblemReporter
+import mill.api.{CompileProblemReporter, PathRef}
 import mill.api.Loose.Agg
 
 object ZincWorkerApi {
@@ -26,8 +26,8 @@ trait ZincWorkerApi {
       scalaVersion: String,
       scalaOrganization: String,
       scalacOptions: Seq[String],
-      compilerClasspath: Agg[os.Path],
-      scalacPluginClasspath: Agg[os.Path],
+      compilerClasspath: Agg[PathRef],
+      scalacPluginClasspath: Agg[PathRef],
       reporter: Option[CompileProblemReporter]
   )(implicit ctx: ZincWorkerApi.Ctx): mill.api.Result[CompilationResult]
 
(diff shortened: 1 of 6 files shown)


lolgab:
This removes a constant of 180ms from every compilation that happens in Mill. With this, the compilation is really fast!

lefou:
> Since the compiler jars have the Scala version in the name, it is enough The `compilersSig` takes now ~20ms vs ~200ms to build in my machine Since it runs on every recompilation, it makes the compiler noticeably faster

First, my feeling about this is not good. The whole correctness is based on assumptions we don't check or enforce. You need a counter example? Yeah, probably seldom, but you could use a locally built compiler jar, which might be always in the same place and lack any indicator of change in it's name. Whenever you rebuild it, you probably also should update the bridge. Or you build a custom compile in another Mill target and return it as `T.dest / "out.jar"`. The path stays stable whereas the content may not. 

As a consequence, basing the whole correctness on an weak assumption for "only" some milliseconds isn't worth it. 

But, as we already on this issue, I have another idea to offer. We already know much more about those compiler classpath that the file names, the `PathRef`. The `PathRef` is exactly built for such kind of situation: detect changes. Calculating the `PathRef` of the compiler classpath is expensive (if not used with `quick`), so we don't want to produce it here. But if we look closer, it was already calculated further up in the call stack, but later dropped. If we change the worker API to accept a `PathRef` of the compiler classpath, we have a perfect cache key for free. I'm pretty sure that we already work with `PathRef`s  in all places, where we use the compiler API, but we drop them later to only extract the `path` and pass that to the current API.

lolgab:
I implemented the approach you suggested and now we also perform much fewer conversions.

lefou (approved):
Looks good to me. Thank you!

lefou (commented):
Before merging, can you please update the PR title and description? Also, are the numbers for the speed improvements roughly correct, now that you changed the implementation a bit?

lolgab:
Updated the description and collected numbers in a better way :)

PR #2186, opened 2022-12-10
Update testng to 7.7.0

lefou:


lefou:
TestNG 7.7.0 does not work with Java 8.

PR #2189, opened 2022-12-10
Use PathRef hashCode for workers caches

lolgab:
Porting the approach used for the Zinc and Scala.js workers to all the workers used in the mill codebase.


lolgab:
I first pushed on the `com-lihaoyi` remote by mistake, that's why there are a bunch of cancelled jobs.

lefou (approved):
Looks good to me. I didn't know we were using `os.mtime` that much. Nice to see it removed. Thank you!

PR #2193, opened 2022-12-12
Update Scala Native API and worker

lolgab:
Drop support for Scala Native 0.4.0 and 0.4.1


lefou:
I'm still struggeling with the `api`/`worker-api` and `worker` naming. First, it's not really a typical API, rather some subset SPI to share some classes to define a worker implementation.

In some of my plugin projects I switched to `worker` (for the former worker `api`) and `worker.impl` or `worker.<better-name-for-the-concrete-worker-impl>`, and find it nicer. 

WDYT?

lolgab:
Can I do that in a separate PR where I use the same convention for all the workers? I just copied `scalajslib` here.

lefou (commented):
Looks good to me. I'm not that much a fan of multiple classes in the same source file. Also the limited scope of the worker api trait is IMHO not necessary and may stand in the way for some individual project customizations. Alternatively we could annotate it `@internal`. But do as you think is best, it's your maintenance job. We probably should review the published version policy for worker modules before we make the next stable release.

Issue #2015, opened 2022-09-02
Consider deprecating or at least avoiding credentials as command line arguments


Flowdalic:
Mill current has at least some Sonatype related command line options that take credentials as arguments. For example from the `publish` task

```
  --sonatypeCreds   String (format: "username:password")
```

Users even potentially provide their gpg passphrase via command line

```
--gpgArgs --passphrase=<gpgPassword>
```

Command line arguments are not considerd private in some (most?) operating systems security model. So other users on the same system could easily learn those credentials. Instead, secrets should be obtained via environment variables or (protected) files.

lefou:
Yeah, that makes sense. I immediately see two actionable tasks:

* Add support for reading these credentials from environment variable via `T.env`.
* Add some note to the documentation (Pages, ScalaDoc) meantioning the security risks and the alternatives via env variables

Flowdalic:
Yes, that would be a good start. +1

Ideally the sonatype credentials are read from a file under `~/.config/`. At least that is how most of my gradle-based projects are setup: IIRC gradle "sources" a file under `~/.config` that potentially contains `sonatypeUsername` and `sonatypePassword`. Then the project's build script can check if those properties are set, and consume them if so.
And instead of providing the gpg passphrase, the gpg-agent is used. It may be worth to look how gradle did it.

lefou:
Using `gpg-agent` should work out of the box.

PR #2203, opened 2022-12-18
Accessing repository credentials via T.env

seroperson:
Supported all implemented repositories: sonatype, bintray, artifactory. Now it is possible not to pass credentials directly and they will be retrieved from approptiate environment variables (`SONATYPE_USERNAME`, `SONATYPE_PASSWORD`, `BINTRAY_USERNAME`, `BINTRAY_PASSWORD`, `ARTIFACTORY_USERNAME`, `ARTIFACTORY_PASSWORD`).

Fixes https://github.com/com-lihaoyi/mill/issues/2015.

I'm not sure everything done correctly so point me if I did something wrong.


docs/antora/modules/ROOT/pages/Common_Project_Layouts.adoc
```diff
@@ -336,6 +336,13 @@ This uploads them to `oss.sonatype.org` where you can log-in and stage/release
 them manually. You can also pass in the `--release true` flag to perform the
 staging/release automatically:
 
+[NOTE]
+--
+Sonatype credentials must have format `username:password` and must be passed as
+second argument or by named argument `--sonatypeCreds` or by specifying environment
+values `SONATYPE_USERNAME` and `SONATYPE_PASSWORD`.
+--
+
```
lefou, line 345: In context of #2015 (security), we should probably suggest to prefer the environment variables, and only use the parameter as a fall back.

lefou (commented):
This looks reasonable to me. Having the `checkSonatypeCreds` method covered by some test would be nice.

docs/antora/modules/ROOT/pages/Common_Project_Layouts.adoc
seroperson: Changed this note a little bit. I don't sure if we should mark this argument as deprecated or just note that environment variables are preferred?

seroperson:
> This looks reasonable to me. Having the `checkSonatypeCreds` method covered by some test would be nice.

Covered in `mill.scalalib.HelloWorldTests.publish`

contrib/artifactory/src/mill/contrib/artifactory/ArtifactoryPublishModule.scala
```diff
@@ -51,6 +73,23 @@ object ArtifactoryPublishModule extends ExternalModule {
...
+      }).getOrElse(
+        throw new IllegalArgumentException(
+          "Consider using ARTIFACTORY_USERNAME/ARTIFACTORY_PASSWORD environment variables or passing `credentials` argument"
+        )
+      )
+    } else {
+      credentials
+    }
+  }
+
```
lefou, line 92: If you make this `def` return a `Task[String]` with `T.task { }`, you'd get a proper way to return a `Result.Failure` which can then nicely bubble up the task graph and reported by Mill. That way, you also don't need to declare the implicit `ctx`, as the task already has it.

docs/antora/modules/ROOT/pages/Common_Project_Layouts.adoc
lefou: I think just the note is OK. Deprecating it would imply, that we are going to remove it someday, but that's not our intention.

lefou (commented):
Your changes look good, also the test. I think we could change the `checkXXXCreds` methods to anonymous tasks, which would make the error handling more explicit.

lefou (approved):
Looks good to me. Great, that you also added some ScalaDoc to existing code. Thank you for this contribution!

PR #2207, opened 2022-12-21
Use `millModuleDirectChildren` consistently everywhere

lolgab:
Make `super.millModuleDirectChildren` available using a private lazy val + a public def.
After this PR it's possible to override `millModuleDirectChildren` and expect consistent behavior across all Mill's functionality.
What I'm especially interested in, is a way to dynamically disable modules based on build variables (like the `crossScalaVersion`). It seems to do the trick but not all the codebase is using `millModuleDirectChidren` as the source of truth. `Cross`, for example, has a copy of its implementation since `lazy val`s don't support calling `super.` on them.

### DIFF
--- a/main/core/src/mill/define/Cross.scala
+++ b/main/core/src/mill/define/Cross.scala
@@ -46,12 +46,11 @@ object Cross {
  *   ...
  * }
  */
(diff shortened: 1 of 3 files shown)


lefou (commented):
This looks good to me. I'd like to wait with merging it until #2208 is in merged, so we have MiMa checks again, which is a nice indicator when we need to publish the next development milestone.

PR #2215, opened 2022-12-26
Support `def scalaJSSourceMap: T[Boolean]`

lolgab:


lefou (commented):
Do we have some test that checks for properly generated sourcemap file?

lolgab:
> Do we have some test that checks for properly generated sourcemap file?

Added a test that checks the sourceMap to be actually disabled

scalajslib/src/mill/scalajslib/ScalaJSModule.scala
```diff
@@ -232,6 +235,8 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
 
   def scalaJSOptimizer: Target[Boolean] = T { true }
 
+  def scalaJSSourceMap: Target[Boolean] = T { true }
```
lefou, line 238: Can we add some docs here?

lefou (approved):
Looks good to me. Thank you, Lorenzo!

PR #2218, opened 2022-12-27
Drop Scala.js 0.6 support and removed some deprecated API

lefou:


lefou:
@lolgab Can you help out with the failing test?

scalajslib/test/src/MultiModuleTests.scala
```diff
@@ -13,8 +13,8 @@ object MultiModuleTests extends TestSuite {
 
   object MultiModule extends TestUtil.BaseModule {
     trait BaseModule extends ScalaJSModule {
-      def scalaVersion = "2.13.4" // the last 2.13 with JS 0.6 support
-      def scalaJSVersion = sys.props.getOrElse("TEST_SCALAJS_0_6_VERSION", ???)
+      def scalaVersion = sys.props.getOrElse("TEST_SCALA_2_13_VERSION", ???)
+      def scalaJSVersion = sys.props.getOrElse("TEST_SCALAJS_VERSION", ???)
```
lefou, line 17: The `MultiModuleTests.run` is the only failing test. Somehow, the expected output (line 76) is not captured since the switch to Scala.js 1. 

@lolgab Any idea? Maybe, we could just write a file and assert that instead? But I don't know the required API to do it.

lolgab: Maybe we can do what we do [here](https://github.com/com-lihaoyi/mill/blob/6c1b8dda2960bff4d8df6389e3092493ec6ca5d0/scalajslib/test/src/HelloJSWorldTests.scala#L274).
That file needs also a change since that code is unreachable now.

lolgab (approved):
LGTM!

PR #2227, opened 2023-01-02
SemanticDB data generation for Java modules

lefou:
This adds support for SemanticDB data generation for Java modules and Mixed Java-/Scala-Modules.

* Fix https://github.com/com-lihaoyi/mill/issues/2226

Requires the following features:

* https://github.com/com-lihaoyi/mill/pull/2231

<details>
<summary>Old PR comment: ...</summary>

For the sake for experimenting, I added a test project under `scratch-hello-mixed`.

```
± mill --version
Mill Build Tool version 0.10.10
Java version: 11.0.15, vendor: Eclipse Adoptium, runtime: /opt/openjdk-bin-11.0.15_p10
Default locale: de_DE, platform encoding: UTF-8
OS name: "Linux", version: 5.15.69-gentoo-x86_64, arch: amd64

± mill -i dev.run scratch-hello-mixed -i showNamed java.semanticDbData
[69/763] de.tobiasroeser.mill.vcs.version.VcsVersion.vcsState 
[763/763] dev.run 
[1/1] showNamed 
[1/1] showNamed > [23/23] java.semanticDbData 
[info] compiling 1 Java source to /home/lefou/work/opensource/mill/scratch-hello-mixed/out/java/semanticDbData.dest/classes ...
[error] /home/lefou/work/opensource/mill/scratch-hello-mixed/java/src/example/HelloJava.java:1:1: 
[error] ## Exception when compiling 1 sources to /home/lefou/work/opensource/mill/scratch-hello-mixed/out/java/semanticDbData.dest/classes
[error] java.lang.RuntimeException: java.lang.RuntimeException: boom
[error] jdk.compiler/com.sun.tools.javac.api.JavacTaskImpl.handleExceptions(JavacTaskImpl.java:163)
[error] jdk.compiler/com.sun.tools.javac.api.JavacTaskImpl.doCall(JavacTaskImpl.java:100)
[error] jdk.compiler/com.sun.tools.javac.api.JavacTaskImpl.call(JavacTaskImpl.java:94)
[error] sbt.internal.inc.javac.LocalJavaCompiler.run(LocalJava.scala:345)
[error] sbt.internal.inc.javac.AnalyzingJavaCompiler.$anonfun$compile$12(AnalyzingJavaCompiler.scala:172)
[error] scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
[...]

PR #2231, opened 2023-01-04
Support non-local Java compiler use in mixed Java/Scala compilation

lefou:
We need this feature to support SemanticDB generation (#2227) via a Java compiler plugin in mixed Scala/Java projects.

We already introduced non-local Java compilers in #1988. This change now also allows the use of a non-local Java compiler in mixed Java/Scala modules, if any `-J` option is given.

As compilers are cached, I added the `hashCode` of the relevant runtime `javacOptions` to the cache key.

I took the opportunity to remove non-relevant options from the existing Java compiler cache, to reduce the amount of cached compiler instances.

/CC @ckipp01 who implemented #1988.

### DIFF
--- a/scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
+++ b/scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
@@ -3,9 +3,8 @@ package mill.scalalib.worker
 import java.io.File
 import java.util.Optional
 import java.util.concurrent.ConcurrentHashMap
-
 import mill.api.Loose.Agg
-import mill.api.{CompileProblemReporter, KeyedLockedCache, PathRef, internal}
+import mill.api.{CompileProblemReporter, KeyedLockedCache, PathRef, Result, internal}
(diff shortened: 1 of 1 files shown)


lefou:
Currently, we maintain two different caches, one for Scala/Mixed usage, and one for Java-only usage. But technically, the same Java compiler can be used in both setups. 

Thus we could avoid holding additional Javac instances in the Scala-compilers-cache, by retrieving the instance from the javac-only-cache first.

ckipp01 (approved):
O nice! It doesn't take much at all to support this. This is great!

lefou:
> Currently, we maintain two different caches, one for Scala/Mixed usage, and one for Java-only usage. But technically, the same Java compiler can be used in both setups.
> 
> Thus we could avoid holding additional Javac instances in the Scala-compilers-cache, by retrieving the instance from the javac-only-cache first.

I think we should work on this separately, as it probably involves further refactoring. We currently cache `Compilers` which is a pair of Scala and Java compile tools, whereas we want to re-use the same `JavaTools` in multiple `Compilers`.

PR #2232, opened 2023-01-04
Add a documentation link to the coursier resolution failure output

### LINKED ISSUE #1401: Improve error message when coursier resolution failed
Improve error message when coursier resolution failed.

It should contain some rudimentary explanation of the `ivy".."` syntax as well as a link to some documentation.

It is probably worth, to create a dedicated doc page explaining all the facetes and pitfalls with using ivy dependencies for scala, scala.js, scala-native, excludes, platform specific OS resolution (JavaFX), ...

lbunschoten:
I've created a small PR which adds a link to the CLI output whenever dependency resolution fails. I am happy to change the text if needed. Any feedback is welcome.

An example of the output:
example

Small note:
I am not really an active contributor in the open source community (yet). I am trying it out, to see if I'll enjoy it. I've decided to just get started with this easy task. Again, any feedback is welcome!

Fixes #1401

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -413,6 +413,8 @@ object main extends MillModule {
            |  val millEmbeddedDeps = ${artifacts.map(artifact =>
             s""""${artifact.group}:${artifact.id}:${artifact.version}""""
           )}
+           |  /** Mill documentation url. */
+           |  val millDocUrl = "${Settings.docUrl}"
            |}
       """.stripMargin.trim
 
--- a/main/src/mill/modules/Jvm.scala
+++ b/main/src/mill/modules/Jvm.scala
@@ -548,7 +548,7 @@ object Jvm {
             |--------------------------------------------
             |
             |For additional information on library dependencies, see the docs at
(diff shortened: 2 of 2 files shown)


lefou (commented):
Looks good. To reduce future maintenance, we probably want to re-use the already configured Documentation URL from https://github.com/com-lihaoyi/mill/blob/e8525bf16c97f98ac4d15f0171af25b7baaebb3a/build.sc#L36 and add it to https://github.com/com-lihaoyi/mill/blob/e8525bf16c97f98ac4d15f0171af25b7baaebb3a/build.sc#L495-L504. Then we can access it via `mill.BuildInfo.millDocumentationUrl`.

lbunschoten:
Thanks for the feedback. I had already checked if I could reuse that docUrl variable, but didn't think to look at that generated bit of code.

I've updated the PR according to your comments.

lefou (approved):
Looks good to me. Thank you!

PR #2233, opened 2023-01-04
Add support for scalaJSOutputPatterns

### LINKED ISSUE #1721: Support Scala.js linker config `OutputPatterns`
In order to be able to run `ESModule`s on Node without tinkering with `package.json`, linker output needs to have `.mjs` extension instead of `.js`. Scala.js linker supports this via [`OutputPatterns`](https://www.scala-js.org/doc/project/module.html#es-modules-and-nodejs). Mill `ScalaJSModule` doesn't expose that config in its API.

Also, would it make sense to expose Scala.js linker API directly instead of the current facade API?

lolgab:
Fixes #1721

### DIFF
--- a/scalajslib/src/mill/scalajslib/api/ScalaJSApi.scala
+++ b/scalajslib/src/mill/scalajslib/api/ScalaJSApi.scala
@@ -109,11 +109,11 @@ object JsEnvConfig {
 }
 
 class OutputPatterns private (
-    private[scalajslib] val jsFile: String,
-    private[scalajslib] val sourceMapFile: String,
-    private[scalajslib] val moduleName: String,
-    private[scalajslib] val jsFileURI: String,
-    private[scalajslib] val sourceMapURI: String
+  val jsFile: String,
+  val sourceMapFile: String,
+  val moduleName: String,
+  val jsFileURI: String,
+  val sourceMapURI: String
 ) {
 
   /** Pattern for the JS file name (the file containing the module's code). */
--- a/scalajslib/src/mill/scalajslib/api/ScalaJSApi.scala
+++ b/scalajslib/src/mill/scalajslib/api/ScalaJSApi.scala
@@ -107,3 +107,93 @@ object JsEnvConfig {
       autoExit: Boolean
   ) extends JsEnvConfig
 }
+
+class OutputPatterns private (
+    private[scalajslib] val jsFile: String,
+    private[scalajslib] val sourceMapFile: String,
+    private[scalajslib] val moduleName: String,
+    private[scalajslib] val jsFileURI: String,
+    private[scalajslib] val sourceMapURI: String
+) {
+
+  /** Pattern for the JS file name (the file containing the module's code). */
+  def withJSFile(jsFile: String): OutputPatterns =
+    copy(jsFile = jsFile)
+
+  /** Pattern for the file name of the source map file of the JS file. */
+  def withSourceMapFile(sourceMapFile: String): OutputPatterns =
+    copy(sourceMapFile = sourceMapFile)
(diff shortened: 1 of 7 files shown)


scalajslib/src/mill/scalajslib/api/ScalaJSApi.scala
```diff
@@ -107,3 +107,93 @@ object JsEnvConfig {
       autoExit: Boolean
   ) extends JsEnvConfig
 }
+
+class OutputPatterns private (
+    private[scalajslib] val jsFile: String,
+    private[scalajslib] val sourceMapFile: String,
+    private[scalajslib] val moduleName: String,
+    private[scalajslib] val jsFileURI: String,
+    private[scalajslib] val sourceMapURI: String
```
lefou, line 116: Why are these private?

scalajslib/worker/0.6/src/ScalaJSWorkerImpl.scala
```diff
@@ -83,7 +83,8 @@ class ScalaJSWorkerImpl extends ScalaJSWorkerApi {
       sourceMap: Boolean, // ignored in 0.6
       moduleKind: ModuleKind,
       esFeatures: ESFeatures,
-      moduleSplitStyle: ModuleSplitStyle // ignored in 0.6
+      moduleSplitStyle: ModuleSplitStyle, // ignored in 0.6
+      outputPatterns: OutputPatterns // ignored in 0.6
```
lefou, line 87: It makes sense to focus on dropping 0.6 support at this point. See #2218.

lefou (commented):
Looks good. Only one question about visibility.

scalajslib/src/mill/scalajslib/api/ScalaJSApi.scala
lefou: As the `apply` and the `withX` methods are public, there is no reason to keep these `val`s non-public.

scalajslib/worker/0.6/src/ScalaJSWorkerImpl.scala
lolgab: Yes, this is meant to go very soon, so we don't want to invest time on it :)

scalajslib/src/mill/scalajslib/api/ScalaJSApi.scala
lolgab: Inspired by Scala.js. This configuration class is meant to be a "write-only", there is no usage on the reading side, that's why I think they make the values package-private [here](https://github.com/scala-js/scala-js/blob/1ec64710273cec8ba41da4aa4c94dfa97d3fb872/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/OutputPatterns.scala#L31).

lolgab: The `apply` is private as well

lefou: I just noticed the `apply` is private, too. Still, I think, if we can create it with some method e.g. `withJSFile` accepting a `String`, than there is sooner or later someone who will want to programmatically retrieve that `String`, e.g. to manipulate it slightly. 

If there is no other good reason (as we don't protect use from extra API maintenance here), we should not make it harder that necessary and make it public.

lolgab: Usually, I prefer to follow whatever the Scala.js plugin does, but what you said makes sense.
Made them public ✅

lefou (approved):

PR #2245, opened 2023-01-09
Refactored BSP module and use isolated classloading for implementations

lefou:
I also excluded bsp4j from transitive Ammonite dependencies, as we don't use Ammonites BSP support and it brought older versions into the tree. 

I removed any use of bsp4j API in scalalib, which was only there previously by accident, as I overlooked it's presence through Ammonites transitive dependencies. `scalalib` now does not depend on bsp4j. I added more case classes (equivalents to the bsp4j classes) if the BSP support required it.

The `bsp` module is now only a thin module, which provides the entry point to BSP and also manages the classloading of the implementation, which is now contained in `bsp.worker` module.

I also minimized the use of reflection in `MillMain` to start the BSP server, which should reduce the potential for errors which can only be catched in (manual) tests.


main/src/mill/MillMain.scala
```diff
@@ -265,31 +265,19 @@ object MillMain {
...
-                        System.err,
-                        MillMain.initialSystemStreams.in,
-                        os.pwd / ".bsp",
-                        java.lang.Boolean.TRUE,
-                        Some(bspServerHandle)
-                      ): _*
+                    // We cannot link this class directly, as it would give us a circular dependency
+                    val bspClass = MillMain.this.getClass.getClassLoader.loadClass("mill.bsp.BspServerStarterImpl")
+                    val method = bspClass.getMethod("get")
+                    val serverStarter = method.invoke(null).asInstanceOf[BspServerStarter]
```
lefou, line 271: Move reflection logic into `BspServerStarter` companion, paired with more info why it is done that way.

bsp/src/mill/bsp/Constants.scala
```diff
@@ -0,0 +1,9 @@
+package mill.bsp
+
+object Constants {
```
lefou, line 3: Mark internal or make it private.

bsp/worker/src/mill/bsp/worker/BspTestReporter.scala
```diff
@@ -1,28 +1,10 @@
...
-  TaskDataKind,
-  TaskFinishParams,
-  TaskId,
-  TaskStartParams,
-  TestFinish,
-  TestReport,
-  TestStart,
-  TestStatus
-}
+import ch.epfl.scala.bsp4j._
```
lefou, line 3: Avoid the mass-import

bsp/worker/src/mill/bsp/worker/BspTestReporter.scala
```diff
@@ -1,28 +1,10 @@
...
 import mill.api.TestReporter
-import sbt.testing.{
-  Event,
-  NestedSuiteSelector,
-  NestedTestSelector,
-  SuiteSelector,
-  TestSelector,
-  TestWildcardSelector
-}
+import sbt.testing._
```
lefou, line 5: Avoid the mass-import

bsp/worker/src/mill/bsp/worker/BspWorkerImpl.scala
```diff
@@ -0,0 +1,163 @@
...
+import java.util.concurrent.Executors
+import scala.concurrent.duration.Duration
+import scala.concurrent.{Await, CancellationException, Promise}
+import scala.util.chaining.scalaUtilChainingOps
+
+@internal
+class BspWorkerImpl() extends BspWorker {
+
+  // TODO: get the classpath of the worker here and persist it for reuse at start time
+  // creates a Json with the BSP connection details
```
lefou, line 22: Done. Remove that TODO!

bsp/worker/src/mill/bsp/worker/MillJvmBuildServer.scala
```diff
@@ -1,22 +1,16 @@
...
-import ch.epfl.scala.bsp4j.{
-  BuildTargetIdentifier,
-  JvmBuildServer,
-  JvmEnvironmentItem,
-  JvmRunEnvironmentParams,
-  JvmRunEnvironmentResult,
-  JvmTestEnvironmentParams,
-  JvmTestEnvironmentResult
-}
+import ch.epfl.scala.bsp4j.{BuildTargetIdentifier, JvmBuildServer, JvmEnvironmentItem, JvmRunEnvironmentParams, JvmRunEnvironmentResult, JvmTestEnvironmentParams, JvmTestEnvironmentResult}
```
lefou, line 3: Format?

bsp/worker/src/mill/bsp/worker/MillScalaBuildServer.scala
```diff
@@ -1,33 +1,23 @@
...
-  ScalaMainClassesParams,
-  ScalaMainClassesResult,
-  ScalaTestClassesItem,
-  ScalaTestClassesParams,
-  ScalaTestClassesResult,
-  ScalacOptionsItem,
-  ScalacOptionsParams,
-  ScalacOptionsResult
-}
+import ch.epfl.scala.bsp4j.{ScalaBuildServer, ScalaMainClass, ScalaMainClassesItem, ScalaMainClassesParams, ScalaMainClassesResult, ScalaTestClassesItem, ScalaTestClassesParams, ScalaTestClassesResult, ScalacOptionsItem, ScalacOptionsParams, ScalacOptionsResult}
```
lefou, line 3: Format?

build.sc
```diff
@@ -924,28 +926,65 @@ object scalanativelib extends MillModule {
...
-  override def compileModuleDeps = Seq(scalalib, scalajslib, scalanativelib, testrunner)
-  override def ivyDeps = Agg(
-    Deps.bsp,
-    Deps.sbtTestInterface
-  )
+  override def compileModuleDeps = Seq(scalalib)
+  override def testModuleDeps: Seq[JavaModule] = super.testModuleDeps ++ compileModuleDeps
+
   def generatedBuildInfo: T[Seq[PathRef]] = T {
+    val workerDep = worker.publishSelfDependency()
```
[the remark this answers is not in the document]
ckipp01: This is only done this way because we needed in in our `BuildInfo` right?

scalalib/src/ScalaModule.scala
```diff
@@ -516,13 +514,14 @@ trait ScalaModule extends JavaModule { outer =>
...
   override def bspBuildTargetData: Task[Option[(String, AnyRef)]] = T.task {
     Some((
-      BuildTargetDataKind.SCALA,
-      new ScalaBuildTarget(
-        scalaOrganization(),
-        scalaVersion(),
-        ZincWorkerUtil.scalaBinaryVersion(scalaVersion()),
-        ScalaPlatform.JVM,
-        scalaCompilerClasspath().map(_.path.toNIO.toUri.toString).iterator.toSeq.asJava
+      "scala",
```
ckipp01, line 517:
```suggestion
      ScalaBuildTarget.dataKind,
```

bsp/src/mill/bsp/BspWorker.scala
```diff
@@ -0,0 +1,80 @@
...
+    worker match {
+      case Some(x) => Result.Success(x)
+      case None =>
+        // load extra classpath entries from file
+        val cpFile = millCtx.workspace / Constants.bspDir / s"${Constants.serverName}.resources"
+        if (!os.exists(cpFile)) return Result.Failure(
+          "You need to run `mill mill.bsp.BSP/install` before you can use the BSP server"
+        )
+
+        // TODO: if outdated, we could regenerate the resource file and re-load the worker
```
ckipp01, line 49: This would be nice, detect the Mill version somehow and warn the user if there is a mismatch.

ckipp01: Ahhh, I see we actually do this down below, so is this a leftover comment?

ckipp01 (commented):
Nice, from a pure code point of view this looks good! I like the separation it introduces although I'm not super familiar with workers yet.

I'll try to use this locally a bit today to see if I hit on anything from a usability standpoint.

lefou:
> Nice, from a pure code point of view this looks good! I like the separation it introduces although I'm not super familiar with workers yet.

"Worker" is a somewhat overloaded term. I don't use a `T.worker` here. I'd like to, but as we start the BSP as early as possible (and therefore can't wait for Mill to resolve and initialize), we need to do it without Mill's task support.

> I'll try to use this locally a bit today to see if I hit on anything from a usability standpoint.

That would be nice. We still have no proper automated test setup that can guarantee a workable server. I typically just delete all `.bsp` and `.metals` dirs and fire up a vscode to check whether Mill BSP somehow ends up with a reasonable modules list in vscode. As I don't use it daily, if find it hard to idicators for a working/non-working integration. It would be nice, if we had some server-independent test suite, which only requires a BSP connection file and allows to assert on the outcome.

build.sc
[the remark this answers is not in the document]
lefou: Yeah. As we want to load (and resolve) it at runtime and don't want to bundle it in the assembly, we need to know the correct dependency coordinates.

bsp/src/mill/bsp/BspWorker.scala
lefou: No, it's still valid. We detect that we are outdated, but at this point we can't do anything about it, except logging or failing. Failing is too rude here, as we would force the user to understand and handle it. Logging is only helpful to understand the issue when we get access to the logs.

At this stage (firing up BSP server without having a completely initialized Evaluator) we can't trigger a re-build of the BSP connection file, as we can't use the evaluator and the task graph.

ckipp01:
>  if we had some server-independent test suite, which only requires a BSP connection file and allows to assert on the outcome.

Sort of an aside, but we talked about this a bit at the Center. It's on our radar and something we'd like to explore and potentially work on to better ensure the BSP ecosystem as a whole is more tested and less error-prone. I'll loop you in when I have more information there.

bsp/src/mill/bsp/BspWorker.scala
lefou: One way to handle a detected outdated worker could be to:

* remember that fact in the BSP service
* Wait for an Evaluator
* Trigger a rebuild of the connection file
* Send an event to the client, that we need to re-start or (if possible) are going to restart the server

ckipp01:
Just a head up, using this locally for a bit, everything seems to work expected when using Metals.

PR #2272, opened 2023-01-20
WIP: track public-ness of named tasks

lefou:
WIP. Track the public-ness of named tasks and restrict cli resolver to public tasks.

* Fixes https://github.com/com-lihaoyi/mill/issues/803

This breaks binary compatibility with older releases.


lefou:
Not sure, whether #803 is still an issue, so this PR is on hold.

Issue #2243, opened 2023-01-09
GenIdeaImpl#sbtLibraryNameFromPom generate illegal file name for Windows 10.


yankun1992:
The method GenIdeaImpl#sbtLibraryNameFromPom return value is 
```scala
s"SBT: ${pom.module.organization.value}:$artifactWithScalaVersion:${pom.version}:jar"
```
But on Windows 10, it report error when writing xml file, the output is:
```text
...
Writing libraries/concurrent_reference_hash_map_1_1_0_jar.xml ...
Writing libraries/java_diff_utils_4_12_jar.xml ...
Writing libraries/jni_loader_3_0_1_0_jar.xml ...
Writing libraries/SBT: io_github_otavia_projects:mill_rust_mill0_10_2_13:0_1_0:jar.xml ...
1 targets failed
mill.scalalib.GenIdea.idea java.nio.file.InvalidPathException: Illegal char <:> at index 13: libraries/SBT: io_github_otavia_projects:mill_rust_mill0_10_2_13:0_1_0:jar.xml
    sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
    sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
    sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
    sun.nio.fs.WindowsPath.parse(WindowsPath.java:94)
    sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:255)
    sun.nio.fs.AbstractPath.resolve(AbstractPath.java:53)
    os.Path.$div(Path.scala:456)
    mill.scalalib.GenIdeaImpl.$anonfun$run$3(GenIdeaImpl.scala:53)
    mill.scalalib.GenIdeaImpl.$anonfun$run$3$adapted(GenIdeaImpl.scala:51)
    scala.collection.IterableOnceOps.foreach(IterableOnce.scala:575)
    scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:573)
    scala.collection.AbstractIterable.foreach(Iterable.scala:933)
    scala.collection.IterableOps$WithFilter.foreach(Iterable.scala:903)
    mill.scalalib.GenIdeaImpl.run(GenIdeaImpl.scala:51)
```

PR #2273, opened 2023-01-20
fix generation of library paths containing `:` for Windows

unkarjedy:
Fixes: #2243

### DIFF
--- a/scalalib/src/GenIdeaImpl.scala
+++ b/scalalib/src/GenIdeaImpl.scala
@@ -394,8 +394,17 @@ case class GenIdeaImpl(
       }
     }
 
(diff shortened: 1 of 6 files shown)


lefou (commented):
This looks ok to me, but it is not covered by any test. Have you tested that it works on Windows manually? Even better, we cover this scenario in `GenIdeaTests` or `GenIdeaExtendedTests`.

unkarjedy:
> Have you tested that it works on Windows manually

Yes, I have tested and it works fine.

Added tests

lefou (approved):
Looks good to me. I especially like that you added some background info to the IntelliJ IDEA workaround, we have in place, in addition to the tests. Thanks you!

PR #2274, opened 2023-01-21
Make ScalaJS and Native Tests modules extend ScalaModule

lolgab:
Some targets like `scalacOptions` were not inherited by `ScalaJSModule`'s `Tests` trait and by `ScalaNativeModule`'s `Tests` trait This fixes that by making it follow the same convention `ScalaModule` follows

### DIFF
--- a/scalajslib/src/mill/scalajslib/ScalaJSModule.scala
+++ b/scalajslib/src/mill/scalajslib/ScalaJSModule.scala
@@ -18,12 +18,10 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
 
   def scalaJSVersion: T[String]
 
-  trait Tests extends TestScalaJSModule {
-    override def zincWorker = outer.zincWorker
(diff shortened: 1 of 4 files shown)


scalajslib/src/mill/scalajslib/ScalaJSModule.scala
```diff
@@ -254,7 +252,7 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
 
 }
 
-trait TestScalaJSModule extends ScalaJSModule with TestModule {
+trait TestScalaJSModule extends TestModule with ScalaJSModule {
```
lefou, line 255: Is this change of order relevant? I think it shouldn't.

lolgab: Reverted ✅

lefou (approved):
Looks good to me.

PR #2288, opened 2023-01-26
SemanticDB for Java improvements

lefou:
This change updates semanticdb-java to 0.8.10.

As this version properly added support for the `-build-tool:mill` option, we enable it, but only, if we really use this version (or newer) at runtime.

The thing is: Metals is sending us its preferred versions for semanticdb-scalac and semanticdb-java. There is a high change, that Mill was build with another version. (When I introduced the semanticdb-java support, Metals came with a much outdated version, resulting in  build failures due to these unexpected low version during development.) Before, Mill used the version from the context (requested by Metals), even if the built-time version was newer. Now, we compare the built-time version with the context version and use whatever is newer. This should be the safest option, but there is still a chance of producing too new semanticdb data, which is not understood by the consumer (Metals). I don't know, how to better check for compatibility though.

I added a new `mill.util.Version` class, which is meant to help with comparing versions.

### DIFF
--- a/main/util/src/mill/util/Version.scala
+++ b/main/util/src/mill/util/Version.scala
@@ -2,6 +2,7 @@ package mill.util
 
 import mill.api.experimental
 
+@experimental
 class Version private (
     val major: Int,
     val minor: Option[Int],
@@ -46,6 +47,7 @@ final class IgnoreQualifierVersion(val underlying: Version) extends AnyVal {
     underlying.isNewerThan(other.underlying)(Version.IgnoreQualifierOrdering)
 }
 
+@experimental
 object Version {
 
   /**
--- a/main/util/src/mill/util/Version.scala
+++ b/main/util/src/mill/util/Version.scala
@@ -1,0 +1,161 @@
+package mill.util
+
+import mill.api.experimental
+
+class Version private (
+    val major: Int,
+    val minor: Option[Int],
+    val micro: Option[Int],
+    val qualifierWithSep: Option[(String, String)]
+) { outer =>
+
+  override def toString(): String = Seq(
+    Some(major),
+    minor.map("." + _),
+    micro.map("." + _),
+    qualifierWithSep.map(q => q._2 + q._1)
+  ).flatten.mkString
+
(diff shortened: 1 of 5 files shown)


main/util/src/mill/util/Version.scala
```diff
@@ -0,0 +1,161 @@
+package mill.util
+
+import mill.api.experimental
+
+class Version private (
```
ckipp01, line 5: I wonder how many versions of something like this are floating around the ecosystem. For example we have one in Metals as well https://github.com/scalameta/metals/blob/main/mtags/src/main/scala/scala/meta/internal/semver/SemVer.scala 😆.

main/util/src/mill/util/Version.scala
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  /**
+   * Missing minor or micro versions are equal to zero,
+   * a qualifier is lower that no qualifier,
+   * qualifiers are sorted alphabetically,
+   * stable ordering.
+   *
+   * TODO: Review ordering wrt Maven 3
+   * TODO: also consider a coursier ordering
```
ckipp01, line 99: I _think_ coursier may just have built-in versioning things we could piggy back on. Or even just use it if it's exposed.

ckipp01 (commented):
I don't _think_ that choosing the newest should really be a problem, but the issue of Metals not having an up-to-date version listed has been fixed. We didn't realize Steward wasn't sending in updates for it, and it's since been fixed. So in reality, if you wanted to be 100% safe and know that the version will work with metals, you should be able to rely on that version now as we will be regularly updating it.

main/util/src/mill/util/Version.scala
lefou: Yeah, the issue is, that such simple yet non-trivial helper neither deserves its own library nor its worth to pull in some larger existing library into the classpath.

lefou:
Thank you @ckipp01 for your feedback. I'm a bit eager to get this merged to improve the user experience. We can improve the version support later.

PR #2291, opened 2023-01-27
BSP: Improve support for `mill-build` module and the `build.sc`

lefou:
This change improves the editing experience for `build.sc` files in Metals. 

We now rudimentary detect ammonite includes (`$file` and `$ivy`).

We somewhat break the BSP protocol now, as we return `sources` as paths to single files instead of directories. It is known, that IDEA does not handle it well, but Metals seems to have no problem with it. We don't report directories as the `build.sc` is located in the project top-level directory, and it would contain all other project files too, which is definitely not what we want. 

Also I refined the module dependencies: `compileIvyDeps` now returns Mill-provided fixed dependencies, `ivyDeps` returns all extra resources included via Ammonite `$ivy` imports, and `scalacPluginIvyDeps` now includes Mill own compiler plugins which among other things make the `override` keyword optional.

I also restructured some source files in `mill.scalalib.bsp` package.

Fixed the overridden-check for `compiledClassesAndSemanticDbFiles`. This should significantly speed things up for larger projects.


ckipp01 (approved):
👍🏼 nice improvements!

PR #2294, opened 2023-01-30
Fetch more tooling deps in `prepareOffine`

lefou:
`ScalaJSModule` and `ScalaNativeModule` did not fetched their worker and tooling classpaths previously.

### DIFF
--- a/scalajslib/src/mill/scalajslib/ScalaJSModule.scala
+++ b/scalajslib/src/mill/scalajslib/ScalaJSModule.scala
@@ -1,6 +1,7 @@
 package mill
 package scalajslib
 
+import mainargs.Flag
 import mill.api.{Loose, PathRef, Result, internal}
 import mill.scalalib.api.ZincWorkerUtil
 import mill.scalalib.Lib.resolveDependencies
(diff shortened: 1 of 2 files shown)


scalajslib/src/mill/scalajslib/ScalaJSModule.scala
```diff
@@ -235,6 +236,17 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
   /** Name patterns for output. */
   def scalaJSOutputPatterns: Target[OutputPatterns] = T { OutputPatterns.Defaults }
 
+  override def prepareOffline(all: Flag): Command[Unit] = {
+    val tasks =
+      if (all.value) Seq(scalaJSToolsClasspath)
```
lolgab, line 241: `scalaJSToolsClasspath` are not very big and necessary to do anything useful with ScalaJS (except plain compilation), so maybe they could be downloaded even without requiring the `all` flag.
What do you think?

lefou: Yeah, makes sense. But with the same argument, we should always fetch Zinc dependencies in ScalaModule. Current scaladoc says:

| @param all If `true` , fetches also sources, Ammonite and compiler dependencies.

PR #2297, opened 2023-01-31
Support reporting of all compilation problems

lefou:
In it's incremental nature, Zinc does only report problems for re-compiled files,
which leads to situations, where you only see a glimpse of the overall compiler warnings.

Now, we use the Zinc internal analysis to recover all found problems. We also record all already shown problems and simply log the not shown but still valid problems after the compilation finished.

This behavior can be configured by the new target `JavaModule.zincReportCachedProblems: T[Boolean]`. It defaults to `false` to keep backward-compatible output.

### DIFF
--- a/scalalib/src/mill/scalalib/SemanticDbJavaModule.scala
+++ b/scalalib/src/mill/scalalib/SemanticDbJavaModule.scala
@@ -199,7 +199,7 @@ trait SemanticDbJavaModule extends CoursierModule { hostModule: JavaModule =>
               compilerClasspath = m.scalaCompilerClasspath(),
               scalacPluginClasspath = semanticDbPluginClasspath(),
               reporter = T.reporter.apply(hashCode),
-              reportOldProblems = zincReportOldProblems()
+              reportCachedProblems = zincReportCachedProblems()
             )
             .map(r => copySemanticdbFiles(r.classes.path, T.workspace, T.dest / "data"))
         }
@@ -219,7 +219,7 @@ trait SemanticDbJavaModule extends CoursierModule { hostModule: JavaModule =>
               compileClasspath = compileClasspathTask(m)().map(_.path),
               javacOptions = javacOpts,
               reporter = T.reporter.apply(m.hashCode()),
-              reportOldProblems = zincReportOldProblems()
+              reportCachedProblems = zincReportCachedProblems()
             ).map(r => copySemanticdbFiles(r.classes.path, T.workspace, T.dest / "data"))
         }
     }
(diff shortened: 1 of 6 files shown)


lefou:
The output in the first draft isn't correct yet. Also, I think we probably can also access complete problems via the internal compiler analysis, which need to be handled in the worker classpath though. I think I will change the design again.

lefou:
Yeah, `ZincWorkerImpl.discoverMainClasses` also uses the zinc analysis, and the `SourceInfo` interface seems to contain problems for all source files, including those not recompiled.

lefou:
I changed the design and it seems to work reasonable now. Not sure about the name and the default value of configuration target `zincReportOldProblems`.

scalalib/src/mill/scalalib/SemanticDbJavaModule.scala
```diff
@@ -198,7 +198,8 @@ trait SemanticDbJavaModule extends CoursierModule { hostModule: JavaModule =>
               scalacOptions = scalacOptions,
               compilerClasspath = m.scalaCompilerClasspath(),
               scalacPluginClasspath = semanticDbPluginClasspath(),
-              reporter = T.reporter.apply(hashCode)
+              reporter = T.reporter.apply(hashCode),
+              reportOldProblems = false
```
lefou, line 202: My initial though was, that semanticdb data is used via Metals, and I think BSP clients are persisting problems anyway. Problems are send as events and not as a result of a compilation, so the client need to manage them anyway.

But OTOH, when semanticdb data is needed for other use, e.g. scalafix, warnings might be still of interest. WDYT?

lolgab: I don't know if Metals can correctly handle duplicated problems.
Maybe @tgodzik can answer this?
About Scalafix, having the warnings reported again would be good.
But this decision can also be taken in a second moment. We can merge this as `false` and then experiment in a second PR to see the implications for BSP.

tgodzik: We will show all the diagnostics that we are sent until we get `reset` in PublishDiagnosticsParams , since we don't know if problems were solved until we are told so explicitely.

lefou: @tgodzik Thanks Tomasz! That means, in case of incremental compilation, diagnostics where until now either incomplete (if there was a `reset`) or may even contain outdated or duplicate problems (if there was no `reset`). I think, always respecting the setting of `zincReportOldProblems` is probably the cleanest option.

lefou:
This PR is now complete. I update the description.

PR #2305, opened 2023-02-06
Update upickle to 3.0.0-M2 and Ammonite to 3.0.0-M0-1-beb96bfb

lefou:
This is a binary incompatible change.


lefou:
We need a compatible Ammonite version first.

PR #2308, opened 2023-02-09
Changed encoding of inner `Tests` modules

lefou:
## TL;DR

I now switched back to name all derivations `Tests` again, but also keep a local type alias with an unique name.
As these types are unique names, it's easy to reference each derivation in the hierearchy exactly, by still keeping the convenience of a simple `object test extends Tests`.

```scala

  // using the convenient `Tests` trait
  object test extends Tests
  // or the more explicit unique name
  object test2 extends ScalaModuleTests

  // this allows explicit reference to super trait
  object test3 extends JavaModuleTests {
    // ...
  }

}
```

## Details

In very older Mill versions, all inner test modules had the same `Tests` class name.
This made recognizing as user and referencing concrete super test modules rather hard.
Therefore, we started to give each derived instance an unique name (e.g. `ScalaModuleTests`) and created a new local `Tests` trait also deriving this unique trait.
Although convenient, it led to compiler warning, that a derived `Tests` trait shadows its super `Tests` trait.
In Scala 3, this will even be a error, hence I tried to find another encoding.

Making just the `Tests` trait an type alias to the uniquely named trait is not possible, as the compiler will error out with an `incompatible type in overriding` message.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -294,7 +294,8 @@ trait MillScalaModule extends ScalaModule with MillCoursierModule { outer =>
     if (this == main) Seq(main)
     else Seq(this, main.test)
 
-  trait MillScalaModuleTests extends ScalaModuleTests with MillCoursierModule
+  type MillScalaModuleTests = Tests
+  trait Tests extends ScalaModuleTests with MillCoursierModule
       with WithMillCompiler {
     override def forkArgs = T {
       Seq(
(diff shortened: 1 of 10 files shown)


lefou:
As pointed out by @lolgab on Gitter, this does not compile with Scala 3. 

I think there is no clean way to provide specialized but equally named `Tests` traits, and we should deprecate them in favor of the uniquely named ones. The pattern is to use the modules trait name plus the `Tests` suffix. E.g `JavaModule` --> `JavaModuleTests`, `ScalaModule` --> `ScalaModuleTests`, ...

PR #2324, opened 2023-02-18
`ScalaJSModule`: Deprecate `fastOpt` and `fullOpt`

lefou:


lefou:
I have no idea why this simple change is failing the build. Do you have an idea, @lolgab ?

lefou:
> Not sure, I'm going to try to run those tests locally later.

@lolgab Did you have any success? I've seen the same issue in PR #2339 and tried locally to bisect it, but all older tags (0.10.11..0.10.5) failed, so it is probably related to some untracked dependencies. As it only affects ScalaJS tests, my suspicion is a changed node.js version.

PR #2325, opened 2023-02-18
Move `JarManifest` to `mill.api`

lefou:
Kept deprecated forwarders in `mill.scalalib.Jvm` to maintain source compatibility.

### DIFF
--- a/main/api/src/mill/api/JarOps.scala
+++ b/main/api/src/mill/api/JarOps.scala
@@ -11,7 +11,7 @@ trait JarOps {
 
   /**
    * Create a JAR file with default inflation level.
-   * d
+   *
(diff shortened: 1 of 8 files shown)


main/api/src/mill/api/JarOps.scala
```diff
@@ -28,13 +28,46 @@ trait JarOps {
...
+    inputPaths = inputPaths,
+    manifest = manifest.build,
+    fileFilter = fileFilter,
+    includeDirs = includeDirs,
+    timestamp = timestamp
+  )
 
+  /**
+   * Create a JAR file with default inflation level.
+   * d
```
lefou, line 46:
```suggestion
```

Issue #2157, opened 2022-11-29
Re-download JARs missing from Coursier cache


kierendavies:
If I clear my Coursier cache (`rm -r ~/.cache/coursier/`) then recompile a module, I get errors like
```
main.compile java.nio.file.NoSuchFileException: /home/kieren/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala3-compiler_3/3.2.2-RC1/scala3-compiler_3-3.2.2-RC1.jar
```

It would be nice if something detected the missing JARs and re-downloaded them.

It would also simplify some of my CI pipelines. Then I could propagate the entire `out` directory between jobs, and not worry about what's in the Coursier cache of the host running each job.

lefou:
Thanks for reporting this. This is a tricky one. I think Mill's evaluator and cache manager isn't prepared for such scenario. I though about such cases in the past, and reasoned, that some kind of `--pedantic` mode might be nice, which always check for valid return `PathRef`s. But I never saw it reported, so forgot about it.

The thing is, that we currently don't detect in-valid cached targets based on the existence of their outcome. As none of the inputs has changed, we simple miss the opportunity to refresh the targets cached value. I think the initial assumption was, that things we control inside `T.dest` doesn't need to costly checked on each run. But cached dependencies resemble paths outside of `T.dest`, so we should provided an (optional) way to check whether the cached value is still valid, before using it as target value.

I have no clear picture yet of a nice API to do that. Some `T.checked` comes to mind. Or an additional `Result` type.

lefou:
Here is a first draft, how we could implement it:

* Annotate the target or persistent target with `@validated`
* The `T` or `T.persistent` macro needs to detect the `@validate` and creates the internal `TargetImpl` with a new `validate = true` attribute
* The macro needs to make sure, the result type can be validated, e.g. `PathRef` or `Seq[PathRef]` (same as we support in `Source` and `Sources`).
* At evaluation time, when we found an already cached value, we need to check if the target has the `validate` attribute set to `true` and need to validate that the path references are valid (path needs to exists and the sig needs to be equal; best is to rely on the same logic as when creating a `PathRef`, so a new `def PathRef.validate(): Boolean` makes sense.)
* If the result is valid, continue with the cached value, otherwise, drop the cached value and re-evaluate the target.

This above algorithm should be implementable and work for the OP use case:

```scala
@validated def resolvedIvyDeps: Target[Seq[PathRef]] = T { ??? }
[...]

PR #2328, opened 2023-02-19
Support automatic `PathRef` validation for cached targets

lefou:
We now check the validity of `PathRef`s upon upickle reading
and throw an exception if it is invalid. This also works, if the
`PathRef` is deeply nested in the result. A `PathRef` is valid,
when it produced the same `sig` on the current filesystem.

Failing deserialization is already handled in Evaluator and
the cached value is automatically dropped. In such a case we also
write a debug level log entry.

A `PathRef` can have three revalidation states, never, once and always.
Once means, that the revalidation state is changed to never when it
is successfully read from json.

* Fix https://github.com/com-lihaoyi/mill/issues/2157

Why is this needed? Normally, `PathRef`s referring to files under `T.dest` do not need extra validation, as the target specific directory is considered as Mill managed storage. But if a `PathRef`s refers to a file or directory outside the Mill cache directory, it can become out of sync. The most prominent example is `JavaModule.resolvedIvyDeps`, which resolves and downloads the `ivyDeps` into an external storage location (`~/.cache/coursier`). If a user clears that cache storage, the cached result is stale, yet Mill does not recognize it. By marking these `PathRef`s to re-validate, the evaluator will check that the external cache storage is as expected and the returned `PathRef`s are valid. Otherwise, it will be re-resolved.

### DIFF
--- a/contrib/twirllib/src/mill/twirllib/TwirlWorker.scala
+++ b/contrib/twirllib/src/mill/twirllib/TwirlWorker.scala
@@ -200,6 +200,5 @@ trait TwirlWorkerApi {
 }
 
 object TwirlWorkerApi {
-
-  def twirlWorker = new TwirlWorker()
+  val twirlWorker = new TwirlWorker()
 }
--- a/contrib/twirllib/test/src/mill/twirllib/HelloWorldTests.scala
+++ b/contrib/twirllib/test/src/mill/twirllib/HelloWorldTests.scala
@@ -41,9 +41,10 @@ trait HelloWorldTests extends TestSuite {
 
   def workspaceTest[T](
       m: TestUtil.BaseModule,
-      resourcePathSuffix: String
+      resourcePathSuffix: String,
+      debug: Boolean = false
   )(t: TestEvaluator => T)(implicit tp: TestPath): T = {
-    val eval = new TestEvaluator(m)
+    val eval = new TestEvaluator(m, debugEnabled = debug)
     os.remove.all(m.millSourcePath)
     os.remove.all(eval.outPath)
     os.makeDir.all(m.millSourcePath / os.up)
@@ -92,7 +93,7 @@ trait HelloWorldTests extends TestSuite {
         )
       }
     }
-    "compileTwirl" - workspaceTest(HelloWorld, "hello-world") { eval =>
+    "compileTwirl" - workspaceTest(HelloWorld, "hello-world", debug = true) { eval =>
       val res = eval.apply(HelloWorld.core.compileTwirl)
       assert(res.isRight)
       val Right((result, evalCount)) = res
--- a/main/api/src/mill/api/PathRef.scala
+++ b/main/api/src/mill/api/PathRef.scala
@@ -33,13 +33,13 @@ case class PathRef private (
   def withRevalidateOnce: PathRef = copy(revalidate = PathRef.Revalidate.Once)
 
   override def toString: String =
-    getClass().getSimpleName() + "(path=" + path + ",quick=" + quick + ",sig=" + sig + ")"
+    getClass().getSimpleName() + "(path=" + path + ",quick=" + quick + ",sig=" + sig + ",revalidated=" + revalidate +")"
 }
 
 object PathRef {
 
   class PathRefValidationException(val pathRef: PathRef)
-      extends RuntimeException(s"Invalid path signature detected: ${pathRef.path}")
+      extends RuntimeException(s"Invalid path signature detected: ${pathRef}")
 
   sealed trait Revalidate
   object Revalidate {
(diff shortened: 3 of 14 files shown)


lefou:
Hm, validate currently enforces existing paths, as the tests didn't work without it. We probably don't encode non-existence in the `PathRef.sig`. We should ensure, that a non-existing path and a missing path have different sigs.

lefou:
> Hm, validate currently enforces existing paths, as the tests didn't work without it. We probably don't encode non-existence in the `PathRef.sig`. We should ensure, that a non-existing path and an empty path have different sigs.

I removed the existence test from validate. I changed the test cases so that the validated paths also contains actual file(s). This should be in line with how we handle empty directories vs. non-existing directories elsewhere.

lihaoyi:
Not sure that an annotation and doing a type-test on the return value of the target is the right thing to do here. It seems pretty fragile and inextensible, e.g. if someone returns tuple of `(PathRef, PathRef)` or some other data structure.

How about the following: 

1. We replace the `@validated` annotation and instead make those targets return a `PathRef(...,  validated = true)`
2. `PathRef(...,  validated = true)` is ~identical to a normal `PathRef`, but will perform validation on instantiation. If validation fails, it will `throw new PathRefValidationException(...)`
3. Inside `Evaluator`, we catch the `PathRefValidationException` when trying to deserialize the cached data into in-memory structures, and fall back to re-computing the value instead.

Notes:
1. My proposed approach has the advantage that we can handle a failed validation nested arbitrarily deep within arbitrary data structures.
2. There isn't really any other hook we can apply to every data type returned by a `Target` other than instantiation, so the only way we can perform any checks is by putting them in the data type constructor
3. The alternative would be to replace `upickle.default.ReadWriter` with our own custom typeclass that can perform "deep" checks inside the data structure in a statically-type-safe way. That seems a pretty invasive change and probably not worth it for such a small edge case
4. Throwing exceptions during uPickle deserialization isn't unheard of: the core parser relies heavily on exceptions deep inside the parser/visitor logic together with try/catch blocks to report errors, so while a bit gross it's no worse than the status quo

@lefou WDYT?

lefou:
> Not sure that an annotation and doing a type-test on the return value of the target is the right thing to do here. It seems pretty fragile and inextensible, e.g. if someone returns tuple of `(PathRef, PathRef)` or some other data structure.
> 
> How about the following:
> 
>     1. We replace the `@validated` annotation and instead make those targets return a `PathRef(...,  validated = true)`
> 
>     2. `PathRef(...,  validated = true)` is ~identical to a normal `PathRef`, but will perform validation on instantiation. If validation fails, it will `throw new PathRefValidationException(...)`
> 
>     3. Inside `Evaluator`, we catch the `PathRefValidationException` when trying to deserialize the cached data into in-memory structures, and fall back to re-computing the value instead.
> 
> 
> Notes:
> 
>     1. My proposed approach has the advantage that we can handle a failed validation nested arbitrarily deep within arbitrary data structures.
> 
>     2. There isn't really any other hook we can apply to every data type returned by a `Target` other than instantiation, so the only way we can perform any checks is by putting them in the data type constructor
> 
>     3. An alternative would be to replace `upickle.default.ReadWriter` with our own custom typeclass that can perform "deep" checks inside the data structure in a statically-type-safe way. That seems a pretty invasive change and probably not worth it for such a small edge case
> 
>     4. Another alternative is to keep the `@validated` annotation use java reflection to walk the reference tree of the object returned by a target. That is more robust than the original `@validated` proposal, and there is precedence e.g. java reflection is how we get the mill module tree. But IMO this would be worse than `PathRef(validated = true)` since it still allows the annotation to be used incorrectly on targets that do not include `PathRef`s
> 
>     5. Throwing exceptions during uPickle deserialization isn't unheard of: the core parser relies heavily on exceptions deep inside the parser/visitor logic together with try/catch blocks to report errors, so while a bit gross it's no worse than the status quo
> 
> 
> @lefou WDYT?

Yeah, an annotation is sub-optimal. I like your proposal as it removes many of the limitations. Thank you! I see one potential issue due to the side effects of re-using that `validated`-enabled `PathRef` later on, e.g. in a downstream target. It's something that could be handled by creating a new copy of it, but it's a bit of suprising effect. (I dropped the idea to use a wrapper type which would make it explicit but to cumbersome earlier.) Do you see any issues, by introducing a `validateOnce` flag, which is dropped after we instantiated the `PathRef`? It would require us to drop the `case` of the class and implement our own `hashCode`/`equals` though.

lihaoyi:
I think passing the pathref around is fine. There will always be a TOCTTOU race condition here no matter what we do, so we just need to do a best effort validation. "perform validation when pathref is first read from cache" feels best-effort enough for me, since that is guaranteed to happen at least once every time `./mill` is invoked

lefou:
@lihaoyi Based on your input, I reworked the `PathRef` validation completely.

We now check the validity of `PathRef`s upon upickle reading
and throw an exception if it is invalid. Beside a much simpler
implementation, this has also the benefit, that we can detect
deeply nested invalid `PathRef`s.

Failing deserialization is already handled in Evaluator and
the cached value is automatically dropped.

A `PathRef` can have three revalidation states, never, once and always.
Once means, that the revalidation state is changed to never when it
is successfully read from json.

PR #2083, opened 2022-10-24
Don't loose `moduleDeps` from super traits in `PublishModule`

lolgab:
Instead of overriding `moduleDeps` to be `PublishModule`s we iterate over the existing `moduleDeps` and we cast to `PublishModule`s if they are of that type, otherwise we throw an exception with an error message to fix the problem.

* Fixes #2082 

<details>
<summary>

### Outdated docs
</summary>

**The first version of this PR was using a `NoStackTrace` exception. Now the following doesn't apply anymore** 

This is with the `NoStackTrace` custom exception:

```scala
lorenzo@MacBook-Pro-di-Lorenzo ~/s/snunit (bump-deps) [1]> ../mill/out/dev/launcher.dest/run --no-server show snunit-tapir[3.2.0].native.publishLocal
Compiling /Users/lorenzo/scala/snunit/versions.sc
Compiling /Users/lorenzo/scala/snunit/unitd.sc
Compiling /Users/lorenzo/scala/snunit/build.sc
[1/1] show 
1 targets failed
show mill.scalalib.PublishModule$Error: PublishModule moduleDeps need to be also PublishModules. snunit[3.2.0].native is not a PublishModule.
```

Since the Exception is create in the `PublishModule` class it's easy to locate where the error comes from.

This is with a normal `Exception`
```scala
lorenzo@MacBook-Pro-di-Lorenzo ~/s/snunit (bump-deps) [1]> ../mill/out/dev/launcher.dest/run --no-server show snunit-tapir[3.2.0].native.publishLocal
[1/1] show 
1 targets failed
show java.lang.Exception: PublishModule moduleDeps need to be also PublishModules. snunit[3.2.0].native is not a PublishModule.
        at mill.scalalib.PublishModule.$anonfun$moduleDeps$1(PublishModule.scala:19)
        at scala.collection.immutable.List.map(List.scala:246)
        at scala.collection.immutable.List.map(List.scala:79)
        at mill.scalalib.PublishModule.moduleDeps(PublishModule.scala:16)
        at mill.scalalib.PublishModule.moduleDeps$(PublishModule.scala:16)
        at ammonite.$file.build$SNUnitTapirModule$native$.moduleDeps(build.sc:148)
        at mill.scalalib.JavaModule.recursiveModuleDeps(JavaModule.scala:155)
        at mill.scalalib.JavaModule.recursiveModuleDeps$(JavaModule.scala:154)
[...]

PR #2339, opened 2023-03-03
Call overridden task from `ScalaModule.mapDependencies` (Backport of #2338)

lefou:
Otherwise, we loose settings mixed in before `ScalaModule`.

Original Pull Request: https://github.com/com-lihaoyi/mill/pull/2338


lefou:
This seems to break binary compatibility, but maybe it's ok to merge anyway, as the issue looks similar to the one we accepted in PR https://github.com/com-lihaoyi/mill/pull/2083.

lefou:
Closing as it is not binary compatible.

PR #2347, opened 2023-03-07
Do not add `Cross` matrix items to `millSourcePath`

lefou:
- Avoid changing millSourcePath in `Cross`
- Revert SingleModule deprecation
- Keep cross module in millSourcePath, but remove cross items


main/core/src/mill/define/Cross.scala
```diff
@@ -24,6 +24,7 @@ object Cross {
           override def millOuterCtx = ctx0.copy(
             crossInstances = vs.map(v => new $tpe(..$argTupleValues))
           )
+          def millSourcePath: os.Path = millOuterCtx.millSourcePath
```
lolgab, line 27: Why is it needed?

lefou: The overridden impl is

```scala
def millSourcePath: os.Path = millOuterCtx.millSourcePath / millOuterCtx.segment.pathSegments
```

As we don't want the cross matrix parameter end up in the path, we need to remove them.

lefou: The issue with your initial solution is, that you removed the segment of the cross module but kept the segments of the cross instances. I changed it to keep the segment of the cross module, but remove the segments of the cross instances.

Unfortunately, this isn't a proper solution. Although it does the right thing, it breaks the ability to customize the `millSourcePath` in a cross module, as mix-in our override _after_ the users customization, we now loose any customized `millSourcePath`.

I think, I need to introduce another context value with always returns the desired segments, so we can use it in the default implementation of `Module.millSourcePath`.

lefou: Change default `millSourcePath` to drop cross segments in 7c0edcc5f183886f886684064cb9712280d5835d. This might work...

integration/local/resources/cross/build.sc
```diff
@@ -0,0 +1,44 @@
...
+//
+//object foo extends Module {
+//  object levelCross extends Cross[LevelCross](("a", "1"), ("b", "2"))
+//  class LevelCross(p1: String, p2: String) extends Module {
+//    def param1 = T { p1 }
+//    def param2 = T { p2 }
+//    def path = T { PathRef(millSourcePath) }
+//  }
+//
+//}
```
lolgab, line 44: Should we remove this commented-out code?

lefou: Yes

main/core/src/mill/define/Module.scala
```diff
@@ -28,7 +28,10 @@ class Module(implicit outerCtx0: mill.define.Ctx) extends mill.moduledefs.Cacher
   private lazy val millModuleDirectChildrenImpl: Seq[Module] =
     millInternal.reflectNestedObjects[Module].toSeq
   def millOuterCtx: Ctx = outerCtx0
-  def millSourcePath: os.Path = millOuterCtx.millSourcePath / millOuterCtx.segment.pathSegments
+  def millSourcePath: os.Path = millOuterCtx.millSourcePath / (millOuterCtx.segment match {
+    case Segment.Label(s) => Seq(s)
+    case Segment.Cross(_) => Seq.empty[String] // drop cross segments
+  })
```
lolgab, line 34: That's the cleanest solution! Congrats!

integration/local/resources/cross/build.sc
lefou: Done

lolgab (approved):

Issue #2341, opened 2023-03-03
Logical loops in moduleDeps


tballard:
It appears that logical loops in moduleDeps are rewarded with a hearty stack trace as of 10.9. An error message with a  list of the modules in the loop(s) would be the right thing, I suppose.

lefou:
Can you please provide a reproduction?

tballard:
Sure:

import mill.scalalib._
import mill.define.Task

object a extends ScalaModule() {
   override def moduleDeps = Seq(a)

   override def scalaVersion = "2.13.7"
}

PR #2349, opened 2023-03-08
Cycle detection in `moduleDeps` and `compileModuleDeps`

lefou:
The problem: We currently don't detect cycles in transitive `moduleDeps` and `compileModuleDeps`. This is results in a `StackOverflowError`.

To solve this, we first need to detect cycles. This is the newly added `Util.recursive` method. Instead of `moduleDeps` we use the `moduleDepsChecked` in some places to ensure we checked for cycles at least once.

Unfortunately, we can't easily report the detected cycle at evaluation time, as `moduleDeps` needs to be evaluated before we can create a task graph (via `T.traverse`), hence we need an alternative way to propagate the exception. The current approach throws an `BuildScriptException` which itself derives from the also new `MillException`. 

In `MillMain`, our central entry point for Mill, we catch `MillException` and report the error to the user.

* Fix https://github.com/com-lihaoyi/mill/issues/2341


lihaoyi:
I don't think we have any choice but to do it lazily. Scala `object`s are already lazily instantiated, so any error will get thrown during the `resolveTasks` stage, as we collect the necessary tasks, but before evaluation actually starts. Throwing and catching seems like the only way

lihaoyi (approved):
Looks good to me I think

lefou:
Hm, some tests run way too long and got cancelled. Probably unrelated, restarting them again.

lefou:
Superseded by #2790

PR #2354, opened 2023-03-12
Update Ammonite to 3.0.0-M0-5 and uPickle to 3.0.0

lihaoyi:


lefou:
Waiting for coursier, which also depends on these dependencies: 

* https://github.com/coursier/coursier/pull/2710
* https://github.com/coursier/coursier/pull/2709

alexarchambault:
@lefou These are not "external" dependencies of coursier (they don't appear in `cs resolve io.get-coursier:coursier_2.13:2.1.0-RC6`). upickle is only used in the tests, and fastparse is shaded. So there's no need to wait for these PRs.

lefou (approved):

PR #2365, opened 2023-03-17
Backport fix for concurrency issue with test event reporting

ollyw:
Hi, it would be really useful to backport this fix. It looks like Mill 0.11.x will be while before release and in the meantime our CI on 0.10.x is failing intermittently, sometimes multiple times a day.


lefou:
Makes sense. I will open another PR #2366 which cherry-picks the original commit.

About the failing tests. Some Scala.js tests in `0.10.x` branch fail, and I have no idea why. Maybe a change in the Github environment like different node version? @lolgab wanted to give it a look.

lefou:
* Superseded by #2366

ollyw:
Thanks for the backport :bow:

PR #2371, opened 2023-03-19
Improve binary compatibility in case classes

lefou:
- Make some case classes more bin-compat friendly
- More case classes...


lefou:
Superseded by #2379

PR #2376, opened 2023-03-21
Remove Ammonite as a dependency, handle script running and bootstrapping ourselves

lihaoyi:
`./mill -i -w dev.run example/example-1 foo.run` passes, printing `Hello World` (as of [ebf0cc8](https://github.com/com-lihaoyi/mill/pull/2376/commits/ebf0cc8ea226ab691ce553a3f9b08394664f2b63))

The basic idea is that instead of relying on Ammonite to evaluate the `build.sc` and return us a `mill.define.BaseModule` value, we instead

1. First instantiate an in-memory `MillBootstrapModule`, which is a special `ScalaModule` configured to read `build.sc` and transform it into `Build.scala` before compiling it with the necessary classpath

2. Use a `mill.eval.Evaluator` to call `runClasspath` on the `MillBootstrapModule`, which gives us a list of files and folders that are necessary to use it

3. Put the files and folders into a `java.net.URLClassloader` to load it into the JVM, and use java reflection to fish out the final `mill.define.BaseModule` runtime value

These are all largely things that were already happening before, just inside the guts of the Ammonite script runner, and mixed in with a bunch of other Ammonite REPL stuff we don't want, and via parallel code paths (e.g. `ammonite.compiler.Compiler` vs `mill.scalalib.worker.ZincWorker`). By pulling them out, we can re-use the bulk of `ScalaModule` logic we already have, and just need to do the minimal amount of classloader plumbing to get us what we want

In order to let us depend on `ScalaModule` within `MillMain`, we need to split out `MillMain` from the rest of `main/` to avoid a circular dependency with `scalalib/`. For now I dumped them in a `entrypoint/` module that depends on both `main/` and `scalalib/`.

Very much WIP

Known broken tests:

- `mill.integration.local.GenIdeaExtendedTests`
- `mill.integration.local.BspInstallDebugTests`


lihaoyi:
Superseded by https://github.com/com-lihaoyi/mill/pull/2377

PR #2378, opened 2023-03-23
CI: cancel older running jobs for a pull requests

lefou:


lihaoyi:
Oh great, that means I can stop doing this manually haha

Is there some way to stop GH actions from double-running jobs? Somehow when I open a PR, it runs the same job once for the PR and once for the commit that was pushed to a branch. That even seems to apply for commits pushed to my own fork `github.com/lihaoyi/mill-1`

PR #2389, opened 2023-03-27
Use scala-parser-combinators:1.1.2 on scala 2, 2.1.0 otherwise

### LINKED ISSUE #2388: `twirlIvyDeps` don't align with twirl dependencies on scala 3
The default [`twirlIvyDeps`](https://github.com/com-lihaoyi/mill/blob/main/contrib/twirllib/src/mill/twirllib/TwirlModule.scala#L30-L35) refer to `scala-parser-combinators:1.1.2` but in the latest twirl 1.6 milestone/RC releases, [they've updated to use `scala-parser-combinators:2.1.0` when on scala 3](https://github.com/playframework/twirl/blob/main/build.sbt#L17-L22).

This causes a resolution error that can't be fixed just by adding the correct `scala-parser-combinators` version to `ivyDeps`. I was able to resolve this by overriding `twirlIvyDeps` in my `build.sc`

mrdziuban:
Fixes #2388

### DIFF
--- a/contrib/twirllib/src/mill/twirllib/TwirlModule.scala
+++ b/contrib/twirllib/src/mill/twirllib/TwirlModule.scala
@@ -24,13 +24,18 @@ trait TwirlModule extends mill.Module { twirlModule =>
     millSourcePath / "views"
   }
 
+  private def scalaParserCombinatorsVersion: T[String] = twirlScalaVersion.map {
+    case v if v.startsWith("2.") => "1.1.2"
+    case _ => "2.1.0"
+  }
+
   /**
    * @since Mill after 0.10.5
    */
   def twirlIvyDeps: T[Agg[Dep]] = T {
     Agg(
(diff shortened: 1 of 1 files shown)


contrib/twirllib/src/mill/twirllib/TwirlModule.scala
```diff
@@ -24,13 +24,18 @@ trait TwirlModule extends mill.Module { twirlModule =>
     millSourcePath / "views"
   }
 
+  private def scalaParserCombinatorsVersion: T[String] = twirlScalaVersion.map {
```
lefou, line 27:
```suggestion
  /** Replicate the logic from twirl build, 
      see: https://github.com/playframework/twirl/blob/bdac9bb9470a7533a44b40c37fb3064737418768/build.sbt#L17-L22
  */
  private def scalaParserCombinatorsVersion: T[String] = twirlScalaVersion.map {
```

lefou (approved):
Looks good to me. Thank you!

PR #2391, opened 2023-03-28
Add nativeDump to ScalaNativeModule

lolgab:
This is useful for debugging Scala Native applications


lefou (approved):

lefou:
Just loud thinking, as `ScalaNativeModule` grows, we should start to create some kind of module reference in the documentation.

PR #2392, opened 2023-03-28
Reworked `PathRef.revalidate` handling based on `DynamicVariable`

lefou:
This uses `scala.util.DynamicVariable` instead of `InheritableThreadLocal` to avoid loosing some state in case we have some nested evaluation going on, e.g. in evaluator command.

I added mostly duplicate code to `Evaluator` `sequentialEvaluate` and `parallelEvaluate` instead of only one in method `evaluate`, as both are public entry points to start evaluation.

### DIFF
--- a/main/api/src/mill/api/PathRef.scala
+++ b/main/api/src/mill/api/PathRef.scala
@@ -2,9 +2,11 @@ package mill.api
 
 import java.nio.{file => jnio}
 import java.security.{DigestOutputStream, MessageDigest}
-import scala.util.chaining.scalaUtilChainingOps
+import java.util.concurrent.ConcurrentHashMap
 import scala.util.Using
 import upickle.default.{ReadWriter => RW}
+
+import scala.runtime.ScalaRunTime
 
 /**
  * A wrapper around `os.Path` that calculates it's hashcode based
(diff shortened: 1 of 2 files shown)


lihaoyi (approved):
Looks good I think

main/api/src/mill/api/PathRef.scala
```diff
@@ -37,6 +39,37 @@ case class PathRef private (
...
+     * Revalidates the given [[PathRef]], if required.
+     * It will only revalidate a [[PathRef]] if it's value for [[PathRef.revalidate]] says so and also considers the previously revalidated paths.
+     * @throws PathRefValidationException If a the [[PathRef]] needs revalidation which fails
+     */
+    def revalidateIfNeededOrThrow(pathRef: PathRef): Unit = {
+      def mapKey(pr: PathRef): Int = (pr.path, pr.quick, pr.sig).hashCode()
+      pathRef.revalidate match {
+        case Revalidate.Never => // ok
+        case Revalidate.Once if map.contains(mapKey(pathRef)) => // ok
+        case _ =>
```
lihaoyi, line 59: Should this line be `case Revalidate.Always`? That would make it more clear that this is a fixed set of options, and there isn't some unbounded number of cases that the `case _` is meant to catch

lefou: Yeah, it's currently `case Revalidate.Once | Revalidate.Always`. But I can also make it explicit, so we get a incomplete match warning if we extend the cases.

PR #2418, opened 2023-04-05
Remove `mill._` magic import from builds

lolgab:
Follow up to discussion in https://github.com/com-lihaoyi/mill/discussions/2381


runner/src/mill/runner/MillBuildRootModule.scala
```diff
@@ -208,9 +208,7 @@ object MillBuildRootModule {
         s"_root_.mill.main.RootModule.Foreign(Some(_root_.mill.define.Segments.labels($segsList)))"
       }
 
-    s"""
-       |package ${pkg.mkString(".")}
-       |import _root_.mill._
+    s"""package ${pkg.mkString(".")}
        |import mill.runner.MillBuildRootModule
```
lefou, line 212: While we're here, we should fix the hygiene and add a `_root_` to the of `MillBuildRootModule`.

lolgab:
`MillBuildRootModule` was also imported via a magic import.
Should we require users to `import mill.runner._` or should we add a type alias in the `package object mill`?

lihaoyi:
Requiring a `mill.runner._` import is probably not what we want, from a user perspective

The reason we need a separate import is because the `mill._` import is in `main`, but `mill.runner.MillBuildRootModule` is in `runner`

I think we don't need to fix that right now. It's definitely a bit awkward, but it's not causing any user-facing issues, so we can take our time and fix it next binary incompatible release

runner/src/mill/runner/MillBuildRootModule.scala
```diff
@@ -275,6 +275,8 @@ object MillBuildRootModule {
     val miscInfoName = s"MiscInfo_$name"
 
     s"""package ${pkg.map(backtickWrap).mkString(".")}
+       |
+       |import mill.runner.MillBuildRootModule
```
lefou, line 279:
```suggestion
       |import _root_.mill.runner.MillBuildRootModule
```

lefou (approved):

Issue #2419, opened 2023-04-05
Peculiar cache behavior for external modules


lefou:
In the process of fixing regressions in the BSP module (after the `remove-ammonite`) I noticed some strange behavior which is probably caused by some stale caches.

When I run a newer Mill version, the `bspWorkerLibs` target of the external `mill.bsp.BSP` module uses a stale cache value. I think, previously, it was re-computed. The default value of that target is computed inside of that target. As the external module is part of the classpath, a change was  tracked by the classpath sig and resulted in an invalidation (I think). It looks like somehow we no longer track a change in the classpath.

lihaoyi:
Seems like it could be related to https://github.com/com-lihaoyi/mill/blob/df86a0757b40ef1ad454e9fcea1f62c739213f58/runner/src/mill/runner/MillBuildBootstrap.scala#L132-L140

It seems to me that that logic relies on at least one meta-build to be present in `mill-build/build.sc`, in order to propagate the Mill executable classpath to the user-level evaluator in order for it to invalidate stuff. The Mill executable classpath is added to the `unmanagedClasspath` of all levels of meta-builds, so >1 meta build would work fine. However, with 0 meta builds, the `getOrElse(0)` would return `0`, ignoring the Mill executable classpath entirely.

Maybe instead of `getOrElse(0)`, it should be `getOrElse(millBootClasspath.map(PathRef(_).sig)` or similar?

lefou:
Yeah, that seem to fix the issue for me. Any idea, how to test this?

I opened PR #2421 to address it.

lihaoyi:
Not sure about testing TBH. This misbehavior only happens when the Mill version changes, and we don't really have any precedence for using different Mill versions in our tests. All our tests use Mill build from HEAD

lefou:
It's probably ok to merge the linked PR as is, as it appears to fix my concrete case. 

Maybe, we can come up with some shell or Scala scripting doing the necessary version bumps and building two different mill versions, in the spirit of our bootstrap tests. 

NB: We also have issues with handling a Mill version change in a running server, which we could approach this time with some tests. I fixed most of them manually in the past, but it was a time consuming process, and it partially regressed when we switched our rpc library to junixsocket.

PR #2420, opened 2023-04-05
Fix BSP regressions and improved resource file handling

lefou:
Fix some BSP regression and some improvements

* [x] Restore proper streams handling; currently `mill --bsp` is dumping all output on `STDERR` which is very noisy
* [x] Better handling of the resource file
* [x] Fixed buildinfo parser for changed embedded dependencies format

### DIFF
--- a/bsp/src/mill/bsp/BSP.scala
+++ b/bsp/src/mill/bsp/BSP.scala
@@ -19,25 +19,12 @@ object BSP extends ExternalModule with CoursierModule with BspServerStarter {
 
   private[this] val millServerHandle = Promise[BspServerHandle]()
 
-  private def bspWorkerIvyDeps: T[Agg[Dep]] = T {
-    Agg(Dep.parse(BuildInfo.millBspWorkerDep))
-  }
-
-  private def bspWorkerLibs: T[Agg[PathRef]] = {
-    sys.props.getOrElse("MILL_UNDER_TEST", "0") match {
-      case "1" => T {
-          mill.modules.Util.millProjectModule(
-            "MILL_BSP_WORKER",
-            "mill-bsp-worker",
-            repositoriesTask()
-          )
-        }
-      case _ => T {
-          resolveDeps(T.task {
-            bspWorkerIvyDeps().map(bindDependency())
-          })()
-        }
-    }
+  private def bspWorkerLibs: T[Agg[PathRef]] = T {
+    mill.modules.Util.millProjectModule(
+      "MILL_BSP_WORKER",
+      "mill-bsp-worker",
+      repositoriesTask()
+    )
   }
 
   /**
--- a/bsp/src/mill/bsp/BSP.scala
+++ b/bsp/src/mill/bsp/BSP.scala
@@ -77,7 +77,6 @@ object BSP extends ExternalModule with CoursierModule with BspServerStarter {
       canReload: Boolean,
       serverHandle: Option[Promise[BspServerHandle]] = None
   ): BspServerResult = {
-
     val ctx = new Ctx.Workspace with Ctx.Home with Ctx.Log {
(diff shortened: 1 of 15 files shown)


lefou:
While testing this locally, I hit #2419

bsp/src/mill/bsp/BSP.scala
```diff
@@ -23,12 +23,21 @@ object BSP extends ExternalModule with CoursierModule with BspServerStarter {
...
   }
 
-  private def bspWorkerLibs: T[Agg[PathRef]] = T {
-    mill.modules.Util.millProjectModule(
-      "MILL_BSP_WORKER",
-      "mill-bsp-worker",
-      repositoriesTask()
-    )
+  private def bspWorkerLibs: T[Agg[PathRef]] = {
+    sys.props.getOrElse("MILL_UNDER_TEST", "0") match {
```
lihaoyi, line 27: What's this change for? AFAICT, it doesn't modify the security characteristics at all: the existing `millProjectModule` already uses `sys.props` to control things, so anyone who has control over the `MILL_BSP_WORKER` property would also have control over the `MILL_UNDER_TEST` property

main/util/src/mill/util/SystemStreams.scala
```diff
@@ -2,4 +2,9 @@ package mill.util
 
 import java.io.{InputStream, PrintStream}
 
-class SystemStreams(val out: PrintStream, val err: PrintStream, val in: InputStream)
+class SystemStreams(
+    val out: PrintStream,
+    val err: PrintStream,
+    val in: InputStream,
+    val bspLog: Option[PrintStream] = None
```
lihaoyi, line 9: I'm not sure this is the right place to put this, since it is really only used by the BSP logic and not the rest of the code in the repo. Furthermore, it kind of breaks the encapsulation here, which is that `SystemStreams` is meant to be the `stdout`/`stdin`/`stderr` streams that are provided by the operating system and not include any application-specific values

bsp/src/mill/bsp/BSP.scala
lefou: Oh, this seems to be left over. I searched for the issues which later turned out to come from #2419. I'll revert it.

main/util/src/mill/util/SystemStreams.scala
lefou: > I'm not sure this is the right place to put this, since it is really only used by the BSP logic and not the rest of the code in the repo. 

Yeah, makes sense. Let me find a better solution.

> Furthermore, it kind of breaks the encapsulation here, which is that `SystemStreams` is meant to be the `stdout`/`stdin`/`stderr` streams that are provided by the operating system and not include any application-specific values

Here I disagree. Initially, `SystemStreams` was created to hold the OS provided streams, but after your refactoring, it's rather a class to pass the three typical streams around. The point in the BSP mode is, that we very early redirect and run the rest of Mill with these customizations. If we want `SystemStreams` to hold the original OS values, we can't use this class further down to pass customized streams around.

lihaoyi: Oh I don't mean that it's only meant to pass the oringal OS streams around, more that it's meant to pass some abstraction that looks like them and expose the same interface. So they can get substituted or redirected, but they're still meant to have 3 streams. So any code that receives it can use those three streams as if it's the OS-provided ones, even if they got replaced somewhere along the way

lefou: Understood. Let we work out some alternative for `bspLog`.

lefou:
@lihaoyi I addressed your review comments and am currently in the process of rebasing/merging to latest main, which is a pita.

lihaoyi (approved):
Not that familiar with the BSP code, but I think this looks ok

PR #2425, opened 2023-04-07
Move BuildInfo from sources into resources, defer use of resources during compilation

lihaoyi:
Instead of compiling the buildinfo values into the source code, we only compile the buildinfo keys into the source code and store the buildinfo values in the resources, to be loaded at runtime. This greatly reduces the amount of time spent compiling things when buildinfo changes. In this repo, some rough tests indicate that it makes the second `-i installLocal` after a one-line change in the `LICENSE` file drop from 50s to 14s of runtime

`mill-profile.json` Before:

[mill-profile.txt](https://github.com/com-lihaoyi/mill/files/11177836/mill-profile.txt)

`mill-profile.json` After:

[mill-profile.txt](https://github.com/com-lihaoyi/mill/files/11177798/mill-profile.txt)

Notes:

1. To really benefit from this, I had to make `compileClasspath` depend only on upstream module's `compileClasspath()`/`compile().classes`, so it doesn't end up depending on their runtime `resources()` and forcing unnecessary recompiles when the buildinfo resources change. This is probably a change we want to do anyway, given the direction set by https://github.com/com-lihaoyi/mill/pull/1811

2. I consolidated the BuildInfo implementation with `contrib/buildinfo/`. 

3. Extended `contrib/buildinfo/` to support Java modules, which we dogfood in `main.client`.

4. The old behavior of statically compiling the values into the binary is still available under `def buildInfoStaticCompiled = true`, for anyone who needs it. That includes anyone using `BuildInfo` in Scala.js, which doesn't support JVM resources by default

5. For now, I copy-pasted the implementation into the root `build.sc` file, and updated `mill-bootstrap.patch` to remove it and make use of the `mill-contrib` version. When we re-bootstrap Mill on latest main, we can remove the copy-pasta

6. `contrib/buildinfo/` should be mostly source compatible with the previous implementation, except that I made `buildInfoPackageName` a required field to encourage the best practice of not putting code in the empty package


lefou (commented):
Interesting approach. It swaps compiletime work with runtime work. Not really my taste.

I think this approach complicates things. We now do some work at runtime which can fail. To be performant, we also need to cache things a bit, to avoid loading the resources again and again. Also, in the current draft, we loose Scaladoc, but that could be changed, and in general the flexibility to provide more than just key-value pair.

What I like about this approach. It is still better than using system properties for things, which are not meant to be changed.

lefou:
What I wonder. How great is the impact of using generated code that changes with every git change? As we use zinc, which tracks binary changes and our generated classes does not change its API (only the values, which are not part of the ABI), downstream compilation targets should not really be affected. Only other targets like `jar` or `assembly`.

lihaoyi:
Yeah being limited to primitive key-values is a bit annoying, but I don't think it's the end of the world. We can always add back support for more complex data structures with the appropriate code-gen. But there's only one place in `com-lihaoyi/mill` which uses a `Seq[String]` rather than a plain `String`, and `contrib/buildinfo/` doesn't support data structures at all, so I think we're ok with dropping that functionality for now

contrib/buildinfo/src/mill/contrib/buildinfo/BuildInfo.scala
```diff
@@ -1,58 +1,201 @@
...
+              isScala: Boolean,
+              buildInfoPackageName: String,
+              buildInfoObjectName: String): String = {
+    val bindingsCode = buildInfoMembers
+      .toSeq
+      .sortBy(_._1)
+      .map {
+        case (k, v) =>
+          if (isScala) s"""val $k = this.readMillBuildInfo("$k")"""
+          else s"""public static java.lang.String $k = readMillBuildInfo("$k");"""
```
lefou, line 141: This needs to be a method, not a field, otherwise the compiler is allowed to inline the value at the call side, which we need to avoid.

lefou (commented):
Currently, you put each property in a single resource file. Have you considered using `java.util.Properties` for loading from a single file? This would also reduce complexity.

lihaoyi:
> Currently, you put each property in a single resource file. Have you considered using `java.util.Properties` for loading from a single file? This would also reduce complexity.

> Currently, you put each property in a single resource file. Have you considered using `java.util.Properties` for loading from a single file? This would also reduce complexity.

I considered it, but some things I liked about separate files:

1. You don't need to worry about data formats, escaping newlines, separators, etc. With one-file-per-entry, the file contains the literal string, that's it

2. If multiple people define buildinfo keys in the same package having them in separate files means that they automatically get unioned together when classpaths are combined. This applies to whether they're in separate folders, separate jars, combined into a single jar, etc. all without needing any special configuration (e.g. `Assembly.Rule.Append`) to make it happen. (This should be mostly avoided by forcing people to provide a packagename, but nonetheless...),

3. General buildinfo keys are pretty few, like O(10), certainly smaller than the number of classfiles that are in the jar anyway. So having them in separate files won't be too bad w.r.t. perf

contrib/buildinfo/src/mill/contrib/buildinfo/BuildInfo.scala
lihaoyi: > This needs to be a method, not a field, otherwise the compiler is allowed to inline the value at the call side, which we need to avoid.

I don't think that happens unless the RHS of the `final` field is a literal constant or (constant-folded). So in this case, a method call loading a JVM resource, it won't get inlined

lefou:
> > Currently, you put each property in a single resource file. Have you considered using `java.util.Properties` for loading from a single file? This would also reduce complexity.
> 
> I considered it, but some things I liked about separate files:
> 
>     1. You don't need to worry about data formats, escaping newlines, separators, etc. With one-file-per-entry, the file contains the literal string, that's it

This is already handled by the implementation of `Properties`, which is a simple String-based Dictionary.

>     2. If multiple people define buildinfo keys in the same package having them in separate files means that they automatically get unioned together when classpaths are combined. This applies to whether they're in separate folders, separate jars, combined into a single jar, etc. all without needing any special configuration (e.g. `Assembly.Rule.Append`) to make it happen. (This should be mostly avoided by forcing people to provide a packagename, but nonetheless...),

Split packages are evil! I don't think going the extra round to make them work a bit better is helpful.

Also, this is unrelated to the use of `Properties`, as we most likely will choose the file name equal to the generated class name, e.g. `my/package/BuildInfo.scala` and `my/package.BuildInfo.properties`.

>     3. General buildinfo keys are pretty few, like O(10), certainly smaller than the number of classfiles that are in the jar anyway. So having them in separate files won't be too bad w.r.t. perf

This still means we have `n - 1` files more compared to a single `.properties` file.

Why I'm arguing? `Properties.{load,store}` method are proven and exist in all Java versions, whereas our own code is new, needs to be tested and might still contain bugs. The concept rather unique and requires learning. Also seeing and recognizing some random `.buildinfo` file in a jar is far from obvious for someone inspecting the jar.

lefou (commented):
I'd rather have an additional field for a JavaDoc comment per key, so that the generated code can be as helpful as it is in the current implementation.

lihaoyi:
@lefou I swapped the implementation for `java.util.Properties`, and extended it to allow comments.

Note that the API had to change to allow comments; `Map("key" -> "value")` is no longer sufficient, so now it's `Seq(BuildInfo.Value("key", "value")) or `Seq(BuildInfo.Value("key", "value", "comment"))`. It's a bit verbose, but I don't think there's any way around it, since fundamentally we do want more data than just `(String, String)` for each buildinfo entry

lefou (commented):
The actual state looks really good. I still need to make up my mind about the `transitiveCompileClasspath` thing. This does mean, that resources are no longer part of it. It might be ok, but means consuming a module as module dependency (resources excluded) differs from consuming it as an ivy dependency (resources included). This has most likely an impact on the mill-osgi plugin and the OSGi setup I work on. I probably need to experiment a bit to get a better feeling.

contrib/buildinfo/src/mill/contrib/buildinfo/BuildInfo.scala
```diff
@@ -1,58 +1,200 @@
...
+   */
+  def buildInfoStaticCompiled: Boolean = false
+
+  /**
+   * A mapping of key-value pairs to pass from the Build script to the
+   * application code at runtime.
+   */
+  def buildInfoMembers: T[Seq[BuildInfo.Value]]
+
+  def resources =
```
lefou, line 30: The file generation should be a dedicated target. This helps when someone needs to override the `resources` target but wants to keep some overridden functionality, or if the file is needed elsewhere.

lefou (commented):
Some remark. You could also use `Class.getResourceAsStream` API, which will read a file in the same package when the path does not start with a `/`. Hence, we can avoid the logic to calculate that ourselves.

I think, it is a good idea to just use the `BuildInfo.objectName` (class name) for the properties file, with the `.properties` extension.
That would make the relation of the files more obvious. It also scales better, if a user needs to generate more that one build info file (disregarding whether the plugin currently supports that).

contrib/buildinfo/src/mill/contrib/buildinfo/BuildInfo.scala
lefou: In the `buildInfoStaticCompiled = true` case, we need to make it a method, as the constants are candidates for inlining (https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.28), but that means we also need to make it a method in the other case, otherwise the generated code is not interchangeable.

contrib/buildinfo/src/mill/contrib/buildinfo/BuildInfo.scala
```diff
@@ -1,58 +1,200 @@
...
+   * rather than the default behavior of storing them as a JVM resource. Needed
+   * to use BuildInfo on Scala.js which does not support JVM resources
+   */
+  def buildInfoStaticCompiled: Boolean = false
+
+  /**
+   * A mapping of key-value pairs to pass from the Build script to the
+   * application code at runtime.
+   */
+  def buildInfoMembers: T[Seq[BuildInfo.Value]]
```
lefou, line 28: We should provide a default of `Seq.empty` here. This better works with stacked traits, as it allows the use of `super`-call. We already handle the empty Seq.

lihaoyi: I don't quite understand this point. How does providing a default allow traits to stack better than otherwise?

lihaoyi:
> I think, it is a good idea to just use the `BuildInfo.objectName` (class name) for the properties file, with the `.properties` extension.
> That would make the relation of the files more obvious.

how about `objectName.buildinfo.properties`? That would better follow filesystem naming coventions of putting the variable part first and the "type" of the file as the extensiin

contrib/buildinfo/src/mill/contrib/buildinfo/BuildInfo.scala
lefou: IIRC, if there is no default, you can't call `super.buildInfoMembers`, as super is abstract.

lefou:
> > I think, it is a good idea to just use the `BuildInfo.objectName` (class name) for the properties file, with the `.properties` extension.
> > That would make the relation of the files more obvious.
> 
> how about `objectName.buildinfo.properties`? That would better follow filesystem naming coventions of putting the variable part first and the "type" of the file as the extensiin

Yeah, why not. I had just `objectName.properties` in mind, but I can live with both.

lihaoyi:
> The actual state looks really good. I still need to make up my mind about the `transitiveCompileClasspath` thing. This does mean, that resources are no longer part of it. It might be ok, but means consuming a module as module dependency (resources excluded) differs from consuming it as an ivy dependency (resources included). This has most likely an impact on the mill-osgi plugin and the OSGi setup I work on. I probably need to experiment a bit to get a better feeling.

Yeah it's a bit subtle. I'm basically moving towards the Bazel convention of separating compile and run, for the same purpose (delaying run dependency generation until the very end for performance/caching). Even in the small com-lihaoyi/mill repo, the time savings for some manual workflows are enormous.

Ivy doesn't have this concept built in, and thus these distinctions get lost when publishing. But I'd argue it's ok, since when publishing you expect downstream users to change the version rarely, and so the performance consideration of improved caching for run target updates doesn't really apply.

The only worry is that maybe having additional resources available at compile time will change behavior somehow vs having no resources, causing compile behavior to diverge when using a local vs remote dependency. I think that is ok for two reasons

1. Compiler plugins and macros very very very rarely use classpath resources, so 99.9% won't notice this change at all

2. If the behavior diverging is problematic, the simple fix is to override `compileResources` to point at `resources`. The you're back to the behavior before these changes. Anyone who somehow requires the old behavior (maybe they have some special compiler plugin configured via resource files) can opt-in to this, paying the performance cost themselves while leaving everyone else to benefit from this speedup

contrib/buildinfo/src/mill/contrib/buildinfo/BuildInfo.scala
lihaoyi: done

lihaoyi:
> Some remark. You could also use `Class.getResourceAsStream` API, which will read a file in the same package when the path does not start with a `/`. Hence, we can avoid the logic to calculate that ourselves.
> 
> I think, it is a good idea to just use the `BuildInfo.objectName` (class name) for the properties file, with the `.properties` extension. That would make the relation of the files more obvious. It also scales better, if a user needs to generate more that one build info file (disregarding whether the plugin currently supports that).

done

contrib/buildinfo/src/mill/contrib/buildinfo/BuildInfo.scala
lihaoyi: done

lihaoyi:
One thing of note is that I had to add `./mill contrib.buildinfo.publishLocal` to `ci/patch-mill-bootstrap.sh`, now renamed to `ci/prepare-mill-bootstrap.sh`, since Mill's own `build.sc` now for the first time is using a contrib plugin. 

I think that's probably a good thing that we've started dogfooding our own plugins, even if it takes a bit more setup to bootstrap.

lihaoyi:
all the feedback so far has been addressed and CI is green

lefou:
> One thing of note is that I had to add `./mill contrib.buildinfo.publishLocal` to `ci/patch-mill-bootstrap.sh`, now renamed to `ci/prepare-mill-bootstrap.sh`, since Mill's own `build.sc` now for the first time is using a contrib plugin. `millProjectModule` isn't wired up to support `contrib` plugins, which I think might take a bit more thought to do nicely.
> 
> I think that's probably a good thing that we've started dogfooding our own plugins, even if it takes a bit more setup to bootstrap.

One thought, not a request: Isn't it now possible to directly include the `contrib/buildinfo/src/mill/contrib/buildinfo/BuildInfo.scala` into `build.sc` via `import $file`?

lihaoyi:
It isn't, because `import $file` assumes a script, which means `.sc` extension and disallows package declarations. That's unchangeable while we're still using Ammonite

Once we tag a milestone and rebootstrap, we'll be able to configure the `mill-build/build.sc` to add the `.scala` to the sources to be compiled without the script code wrapper. So in future we should be able to avoid this copy-paste-patch messiness and just directly share the `.scala` file between application code and the root `build.sc`

lefou:
I was referring to the bootstrap tests, where we need to publishLocal the plugin, but instead could load it directly as source file.

lefou:
In another project we kept `.scala` symlinks to `.sc` files, to use plugin code in Mill directly and to publish it as plugin.

lihaoyi:
Oh, for the bootstrap tests, we might be able to. I don't see how `import $file` would help there, for the same reasons. My understanding is that we'd need to wire them up the same way we do `millProjectModule`s

lihaoyi:
How does the symlinks work with the package declaration st the top of the file? My understanding is that it's a blocker, since it's required to have one for `.scala` files and required *not* to have one for `.sc` files

lefou (approved):
Some files needs formatting. Otherwise, this looks good to go. And I'm happy about the outcome. The buildinfo plugin is now in a usable state (Java support and Java/ScalaDoc). Nice work!

lihaoyi:
Is this how you're meant to auto-format the files? It's the command you gave inhttps://github.com/com-lihaoyi/os-lib/pull/160, but seems to give me in an error

```
lihaoyi mill$ ./mill __.reformat
Cannot resolve __.reformat. Try `mill resolve _` to see what's available.

```

lefou:
https://com-lihaoyi.github.io/mill/mill/Configuration.html#_reformatting_your_code

lolgab:
This broke my Scala.js build.
Can we set the default so `ScalaJSModule`s and `ScalaNativeModule`s always gets `def buildInfoStaticCompiled = true`?

PR #2428, opened 2023-04-10
Cleanup stream handling in PrintLogger to try and avoid over-written lines

lihaoyi:
Right now, if we add a `println` inside `evaluateGroupCached` for debugging purposes:

```
diff --git a/main/core/src/mill/eval/Evaluator.scala b/main/core/src/mill/eval/Evaluator.scala
index 6805efd41..1fe1e5b02 100644
--- a/main/core/src/mill/eval/Evaluator.scala
+++ b/main/core/src/mill/eval/Evaluator.scala
@@ -306,7 +306,7 @@ class Evaluator private (
       testReporter: TestReporter,
       logger: ColorLogger
   ): Evaluated = {
-
+    println("evaluateGroupCached")
     val externalInputsHash = scala.util.hashing.MurmurHash3.orderedHash(
       group.items.flatMap(_.inputs).filter(!group.contains(_))
         .flatMap(results(_).asSuccess.map(_.value._2))
```

It gets over-written by the ticker in the terminal and not reliably shown:

```
$ rm -rf example/basic/1-hello-world/out && ./mill -i dev.run example/basic/1-hello-world -i run --text hello
evaluateGroupCached
[35/48] finalMainClassOpt
[36/48] finalMainClass
[37/48] resources
[38/48] localClasspath
[39/48] transitiveLocalClasspath
[40/48] runIvyDeps
[41/48] resolvedRunIvyDeps
[42/48] upstreamAssemblyClasspath
[43/48] runClasspath
[44/48] forkArgs
[45/48] forkEnv
[46/48] forkWorkingDir
[47/48] runUseArgsFile
[48/48] run
<h1>hello</h1>
```

With this PR, the `println` is reliably shown

```
$ rm -rf example/basic/1-hello-world/out && ./mill -i dev.run example/basic/1-hello-world -i run --text hello
evaluateGroupCached
[35/48] finalMainClassOpt
evaluateGroupCached
[36/48] finalMainClass
evaluateGroupCached
[37/48] resources
evaluateGroupCached
[38/48] localClasspath
evaluateGroupCached
[39/48] transitiveLocalClasspath
evaluateGroupCached
[40/48] runIvyDeps
evaluateGroupCached
[41/48] resolvedRunIvyDeps
evaluateGroupCached
[42/48] upstreamAssemblyClasspath
evaluateGroupCached
[43/48] runClasspath
evaluateGroupCached
[44/48] forkArgs
evaluateGroupCached
[45/48] forkEnv
evaluateGroupCached
[46/48] forkWorkingDir
evaluateGroupCached
[47/48] runUseArgsFile
evaluateGroupCached
[48/48] run
<h1>hello</h1>
```

The basic problem is that the `PrintLogger` uses its own internal `printState` variable to keep track of what's been printed so far, and whether it needs to put the next progress ticker entry on a new line, but anything printed outside the evaluation of a target ends up bypassing this logic and not properly recorded. Thus any debugging `println`s you put inside Mill are prone to being over-written by the ticker as your build progresses. Furthermore, anyone who ends up copying or creating new `PrintLogger`s will have a different inconsistent `printState`, which also can cause lines to be incorrectly overwritten.

This PR lifts the ticker `printState` variable out of the `printLogger` and instantiates it in `MillMain.scala`, and immediately overrides all System streams to make use of wrappers that correctly update this `printLoggerState` as they print things. This ensures that all `println`s throughout the Mill codebase will correctly interact with the ticker, and your debug `println`s you scatter throughout Mill won't be incorrectly over-written in the terminal

I also took the opportunity for some cleanups:

1. Broke out a `Util.withStreams` helper, that we can use in various parts of the codebase to perform our stdout/stderr/stdin redirection in a consistent way

2. Broke up `Loggers.scala` into a bunch of separate files, since it had grown huge and unwieldy

3. Refactored `Logger` to use `SystemStreams` rather than passing around the three streams separately

4. Removed `PrintLogger.infoStream`, which was not serving any real purpose

The specific issue that is fixed here isn't easily covered by automated tests, since it requires adding `println`s to the Mill codebase, which is not something we can easily do in integration tests. The parts that are easy to test (e.g. `println`s in target definitions) aren't currently broken, so I didn't see the need to add tests for that. For now I just tested this PR manually.


lihaoyi:
Should be ready for review, I expect CI should be green once it's done

main/api/test/src/mill/util/VersionTests.scala
lefou, line 1: Is this an accidental copy? I think `Version` is currently in `mill.util.Version` which belongs to `mill.main` module.

lihaoyi: Oh, I think this was leftover from an aborted refactor. I'll remove it

integration/src/mill/integration/IntegrationTestSuite.scala
```diff
@@ -43,65 +44,67 @@ abstract class IntegrationTestSuite extends TestSuite {
...
+            logger = logger
+          ).evaluate()
+        }
+      )
 
-    (isSuccess, newRunnerState)
+      runnerState = newRunnerState
+      (isSuccess, newRunnerState)
+    }
+    println("-runnerStdout")
```
lihaoyi, line 86: I need to remove this as well

lihaoyi:
Updated according to feedback

PR #2435, opened 2023-04-14
enforce stdout/stderr contents in `.local` example/integration tests

lihaoyi:
This should allow us to do debugging or iteration on example/integration in `.local` mode, which is a lot faster than `.fork` or `.server`

Even after https://github.com/com-lihaoyi/mill/pull/2425 landed, the `.local` integration tests are still much faster to run than `.fork` and `.server`, with publishing taking ~20s down from ~60s before which is still a very annoying wait. Better if we can avoid that, using `.local` in the common development workflow and leaving `.fork` and `.server` for CI to catch rare edge-case bugs

We basically needed to be a bit more robust in spawning subprocesses, allowing the `InputPumper` machinery to work for any redirected streams rather than being hardcoded to work with the `PipedInputStream` that `MillServerMain` uses.

I had to fix an issue in `ScalaJSModule#run` to make `example.web[3-scalajs-module].local.test` pass with this additional enforcement. It appears we were not properly pumping the JSEnv subprocess stdout/stderr. This would cause the outputs to disappear when run repeatedly in client-server mode, since the output would be inherited by first MillClient process that spawned the server, and not subsequent clients that simply connect to it

### DIFF
--- a/main/api/src/mill/api/SystemStreams.scala
+++ b/main/api/src/mill/api/SystemStreams.scala
@@ -34,7 +34,7 @@ object SystemStreams{
    * want to print stuff while the system streams override are messed up
    */
 
-  def originalErr = original.err
+  def originalErr: PrintStream = original.err
 
   def withStreams[T](systemStreams: SystemStreams)(t: => T): T = {
     val out = System.out
--- a/main/api/src/mill/api/SystemStreams.scala
+++ b/main/api/src/mill/api/SystemStreams.scala
@@ -9,10 +9,34 @@ class SystemStreams(
 )
 
 object SystemStreams{
-  val original = new SystemStreams(System.out, System.err, System.in)
+
+  private val original = new SystemStreams(System.out, System.err, System.in)
+
+  /**
+   * Used to check whether the system streams are all "original", i,e. they
+   * have not been overriden. Used for code paths that need to work differently
+   * if they have been overriden (e.g. handling subprocess stdout/stderr)
+   *
+   * Assumes that the application only uses [[withStreams]] to override
+   * stdout/stderr/stdin.
+   */
+  def isOriginal() = {
+    (System.out eq original.out) &&
+    (System.err eq original.err) &&
+    (System.in eq original.in) &&
+    (Console.out eq original.out) &&
+    (Console.err eq original.err) &&
+    (Console.in eq original.in)
+  }
+
+  /**
+   * The original non-override stderr, used for debugging purposes e.g. if you
+   * want to print stuff while the system streams override are messed up
+   */
+
+  def originalErr = original.err
 
   def withStreams[T](systemStreams: SystemStreams)(t: => T): T = {
-
     val out = System.out
     val in = System.in
     val err = System.err
--- a/main/api/src/mill/api/SystemStreams.scala
+++ b/main/api/src/mill/api/SystemStreams.scala
@@ -7,3 +7,30 @@ class SystemStreams(
     val err: PrintStream,
     val in: InputStream
 )
+
+object SystemStreams{
+  val original = new SystemStreams(System.out, System.err, System.in)
+
+  def withStreams[T](systemStreams: SystemStreams)(t: => T): T = {
+
+    val out = System.out
+    val in = System.in
+    val err = System.err
+    try {
+      System.setIn(systemStreams.in)
+      System.setErr(systemStreams.err)
+      System.setOut(systemStreams.out)
+      Console.withIn(systemStreams.in) {
+        Console.withOut(systemStreams.out) {
+          Console.withErr(systemStreams.err) {
+            t
+          }
+        }
+      }
+    } finally {
+      System.setErr(err)
+      System.setOut(out)
+      System.setIn(in)
+    }
+  }
+}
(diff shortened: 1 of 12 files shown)


lihaoyi:
@lefou this should be ready to review i think, CI looks good

scalajslib/worker/1/src/mill/scalajslib/worker/ScalaJSWorkerImpl.scala
```diff
@@ -260,7 +260,27 @@ class ScalaJSWorkerImpl extends ScalaJSWorkerApi {
   def run(config: JsEnvConfig, report: Report): Unit = {
     val env = jsEnv(config)
     val input = jsEnvInput(report)
-    val runConfig = RunConfig().withLogger(new ScalaConsoleLogger)
+    val runConfig0 = RunConfig().withLogger(new ScalaConsoleLogger)
+    val runConfig =
+      if (System.in == mill.api.SystemStreams.original.in) runConfig0
+      else runConfig0
```
lolgab, line 266: Do you need `ScalaConsoleLogger` also in this case? Did you try without it?

main/api/src/mill/api/SystemStreams.scala
```diff
@@ -7,3 +7,30 @@ class SystemStreams(
     val err: PrintStream,
     val in: InputStream
 )
+
+object SystemStreams{
+  val original = new SystemStreams(System.out, System.err, System.in)
```
lefou, line 12: I don't think this works reliable, as objects are only instantiated when their class is loaded/accessed (I think). It works for us, as we use `SystemStream` from `MillMain`, which is "early enough", but there is no guarantee this still works, if we change `MillMain` later.

Either we set this from outside, or we add some notes about these potential issue.

lefou (commented):
I like the fix but I'm not completely sold by exposing the original streams. There are probably reasons to swap them and with this change, we leak the original streams. If some code is using them, we may see unexpected behavior/error. Could we instead just store a hash or provide a `isOriginalInStream(stream)`?

lihaoyi:
@lefou I actually wanted to expose the original streams intentionally for debugging purposes. 

Sometimes when I want to print stuff and the stream redirects are screwed up, I have to do something like `os.proc("cat", os.temp("foo\n")).call(stdout = os.Inherit)` to get my debugging lines to the console. Exposing the original streams would let me just do `mill.api.SystemStreams.original.out.println`, which while a bit long is definitely easier to remember than jumping through hoops with `os.proc`

main/api/src/mill/api/SystemStreams.scala
lihaoyi: So my idea was that this is guaranteed to work on the condition that we only use `SystemStreams.withStreams` to do the redirects. I updated all the existing code to do so, and is probably a best practice anyway in order to keep the logic consistent throughout the codebase.

Now I can't guarantee that someone in future won't add code that redirects stuff manually using `System.setOut` or `Console.withOut`, but as long as we stick with the convention of using `SystemStreams.withStreams`, then this should work robustly

lefou:
> @lefou I actually wanted to expose the original streams intentionally for debugging purposes.
> 
> Sometimes when I want to print stuff and the stream redirects are screwed up, I have to do something like `os.proc("cat", os.temp("foo\n")).call(stdout = os.Inherit)` to get my debugging lines to the console. Exposing the original streams would let me just do `mill.api.SystemStreams.original.out.println`, which while a bit long is definitely easier to remember than jumping through hoops with `os.proc`

One reason we added the top-level stream redirection is the way our BSP server is implemented. Stdin and stdout are reserved for the protocol layer and anything printed there would break the protocol. I do see the value when debugging stuff, but the risk to break something is real. Maybe we can only leak the original stderr channel, which is more unlikely to be used for "protocol" stuff?

lihaoyi:
@lefou sure, I can do that. Will update the PR

lihaoyi:
Made the discussed changes, hopefully CI is happy

main/api/src/mill/api/SystemStreams.scala
```diff
@@ -7,3 +7,54 @@ class SystemStreams(
...
+
+  /**
+   * Used to check whether the system streams are all "original", i,e. they
+   * have not been overriden. Used for code paths that need to work differently
+   * if they have been overriden (e.g. handling subprocess stdout/stderr)
+   *
+   * Assumes that the application only uses [[withStreams]] to override
+   * stdout/stderr/stdin.
+   */
+  def isOriginal() = {
```
lefou, line 23:
```suggestion
  def isOriginal(): Boolean = {
```

main/api/src/mill/api/SystemStreams.scala
```diff
@@ -7,3 +7,54 @@ class SystemStreams(
...
+    (Console.err eq original.err) &&
+    (Console.in eq original.in)
+  }
+
+  /**
+   * The original non-override stderr, used for debugging purposes e.g. if you
+   * want to print stuff while the system streams override are messed up
+   */
+
+  def originalErr = original.err
```
lefou, line 37:
```suggestion
  def originalErr: PrintStream = original.err
```

lefou (approved):
Looks good to me. I'd like to have explicit return types in public API.

PR #2448, opened 2023-04-19
[WIP] Consolidate Docs with Examples

lihaoyi:


lefou:
I like the approach, but I doubt antora can build that setup for older versions. Currently, antora can checkout older version of the Mill repo and build what's defined in `antora.yml`. I don't think, we can start a Mill process for each version to generate additional files.

lihaoyi:
I think we can make it work:

1. Antora seems to be pretty flexible in how it handles older versions (https://docs.antora.org/antora/latest/content-source-versioning-methods/). 

2. All old versions are already built: they were built when the version was tagged and released. There's probably some way we can reuse either the materialized docsite folder or the final HTML to avoid doing redundant work.

I'm still getting used to Antora, but this is definitely the right direction to take the docs in. In the process of working on this PR I've fixed tons of typos and mistakes in the docs that could never be caught before because they were never executable. 

Having tested-correct documentation would be a big step forward in both internal maintainability and end-user experience: reducing duplication between the integration tests and the docs and giving us confidence that the doc examples actually work, while giving users known-working examples that they can actually download and execute whether to learn about Mill or to bootstrap their projects

lefou:
Yeah, I really like the direction. Maybe, we can push the "generated sources" (out/docs/source) to some doc-branch to be used by antora? That way, it would also be easier to maintain documentation of older versions, too.

@lihaoyi Do you mind if push some small changes I saw while reviewing it, as these can't be easily done as suggestion? Some typos and asciidoc things. Also a more generic way to iterate over contrib modules.

lihaoyi:
@lefou go for it. It's still pretty rough, will take a few more days to get it into properly reviewable state I think

lihaoyi:
@lefou btw I thought this was hilarious. I must have spent like half an hour trying to figure this out myself, AI chatbot got the answer in 30 seconds 😛 

Screenshot 2023-04-19 at 6 45 31 PM

build.sc
```diff
@@ -1631,6 +1633,134 @@ object dev extends MillModule {
...
+        options ++ scaladocFiles.map(_.toString)
+      ) match {
+        case true => Result.Success(PathRef(T.dest))
+        case false => Result.Failure("doc generation failed")
+      }
+
+      docReturn
+    }
+
+    def push() = T.command {
```
lefou, line 1707: Instead of manual pushing to some site, we could publish a synthetic artifact and upload the scaladoc with the javadoc classifier to Sonatype. There are already lots of nice online viewers like https://javadoc.io/doc/com.lihaoyi/mill-scalalib_2.13/latest/index.html

lihaoyi: The `def push()` stuff was copy-pasta from the original source, but I was thinking we could just upload the scaladoc HTML to github pages together with the Antora docsite, since we already have to deploy that stuff anyway

lefou (commented):
@lihaoyi, FYI, as I notices that you moved some Java examples in less prominent places: In previous discussions with @lolgab, we agreed, that we should also position Mill as a build tool for Java. We have lots of features to offer for Java folks, and there are Java devs ready to use a script/code based build tool (e.g. Gradle). It would be nice to keep/have some prominent Java examples.

lihaoyi:
If we want we can move `basic/11-java-modules` example to the front page, but before we do that I think we need at least one dedicated follow-on page for non-Scala folks to go to. Otherwise people would come to the front page, see "Java", go to "configuring your project" and immediately bounce off after being confronted with Scala-versions/Scala-compiler-plugins/Scala-dependencies/Scala-test-frameworks/Scala.js/SBT-compatibility/etc., not to mention all the Scala code we put front-and-center in the code snippets.

I agree that I'd like to try and get Java people to use Mill. I think we can make putting together Java learning materials follow-on work, after this current overhaul lands

lefou (commented):
We consciously choose to have a separate page for plugin usage for various reasons:

* User typically sooner or later search for "plugin" in the TOC
* We need a landing page for all things related to plugin, e.g. we link to it from the contrib and thirdparty plugin lists, but I expect links from various external plugins to target that page
* We wanted to extend that page to add some plugin writers guide with best practices
* Although "extending Mill" is technically correct, it's probably not a catching enough title for users that just want to understand how plugins work; maybe it's enough to reserve a section under extending plugins and either link to the separate page or just embed/include (= duplicate?) it's content.

lihaoyi:
Got it. I merged it because I thought it was a bit short to be a standalone page, but I can split it out again

.github/workflows/actions.yml
```diff
@@ -60,15 +60,15 @@ jobs:
           - java-version: 17
             buildcmd: ci/test-mill-dev.sh
           - java-version: 17
-            buildcmd: ./mill -i -k "example.__.test"
+            buildcmd: ./mill -i -k "example.__.local.test"
```
lihaoyi, line 63: The examples folder has grown pretty big, and they're usually pretty straightforward. Doesn't make sense to run each one three times in local/fork/server configs

.github/workflows/run-mill-action.yml
```diff
@@ -36,8 +36,8 @@ jobs:
 
       - name: Prepare git config
         run: |
-          git config --add user.name "Mill GithHub Actions"
-          git config --add user.email "mill-ci@localhost"
+          git config --global user.name "Mill GithHub Actions"
+          git config --global user.email "mill-ci@localhost"
```
lihaoyi, line 40: Some of the example tests use git in a newly created repo in a sub-folder, so need the name and email to be set globally

build.sc
```diff
@@ -1559,191 +1650,301 @@ object dev extends MillModule {
...
 
+
+/** Generates the mill documentation with Antora. */
 object docs extends Module {
+  // This module isn't really a ScalaModule, but we use it to generate
+  // consolidated documentation using the Scaladoc tool.
+  object site extends UnidocModule {
+    def scalaVersion = Deps.scalaVersion
 
-  /** Generates the mill documentation with Antora. */
```
lihaoyi, line 1564: I eliminated one layer of nesting; there wasn't anything interesting inside `docs` outside `docs/antora/` anyway

build.sc
```diff
@@ -1559,191 +1650,301 @@ object dev extends MillModule {
...
-          s"    mill-last-tag: '${millLastTag()}'"
-        case l => l
+  }
+
+  /**
+   * Mix this in to any [[ScalaModule]] to provide a [[unidocSite]] target that
+   * can be used to build a unified scaladoc site for this module and all of
+   * its transitive dependencies
+   */
+  trait UnidocModule extends ScalaModule {
```
lihaoyi, line 1674: This gets deleted in `ci/patch-mill-bootstrap.sh`, and instead uses the version published in `mill.scalalib`

build.sc
```diff
@@ -1559,191 +1650,301 @@ object dev extends MillModule {
...
+    s"""site:
+       |  title: Mill
+       |  url: ${if (authorMode) s"${T.dest}/site/index.html" else Settings.docUrl}
+       |  start_page: mill::Intro_to_Mill.adoc
+       |
+       |content:
+       |  sources:
+       |    - url: ${if (authorMode) baseDir else Settings.projectUrl}
+       |      branches: []
+       |      tags: ["0.10.12", "0.9.12"]
```
lihaoyi, line 1810: I just set this to publish the latest of each minor version. This is what SBT does. Publishing docs for every point release is overkill IMO

build.sc
```diff
@@ -1559,191 +1650,301 @@ object dev extends MillModule {
...
-          "--attribute",
-          "page-pagination"
-        )
+    os.write.over(dest / "antora.yml", lines.mkString("\n"))
+    PathRef(dest)
+  }
+  def githubPagesPlaybookText(authorMode: Boolean): Task[String] = T.task {
+    s"""site:
+       |  title: Mill
+       |  url: ${if (authorMode) s"${T.dest}/site/index.html" else Settings.docUrl}
```
lihaoyi, line 1803: This makes the home navigation links in the top left work even when looking at the local version of the site

build.sc
```diff
@@ -1559,191 +1650,301 @@ object dev extends MillModule {
...
+       |    url: https://gitlab.com/antora/antora-ui-default/-/jobs/artifacts/master/raw/build/ui-bundle.zip?job=bundle-stable
+       |    snapshot: true
+       |  supplemental_files: ${supplementalFiles().path.toString()}
+       |
+       |asciidoc:
+       |  attributes:
+       |    mill-github-url: ${Settings.projectUrl}
+       |    mill-doc-url: ${if (authorMode) s"file://${T.dest}/site" else Settings.docUrl}
+       |    mill-download-url: ${if (authorMode) s"file://${exampleZips().head.path / os.up}" else "https://github.com/com-lihaoyi/mill/releases/latest"}
+       |    mill-example-url: ${if (authorMode) s"file://${T.workspace}" else "https://github.com/com-lihaoyi/mill/blob/main/"}
```
lihaoyi, line 1828: Tested these URLs to work for the local version of the doc-site. Hopefully they work when published

main/api/src/mill/api/AggWrapper.scala
```diff
@@ -7,7 +7,7 @@ import scala.language.implicitConversions
 object Strict extends AggWrapper(true)
 object Loose extends AggWrapper(false)
 
-sealed class AggWrapper(strictUniqueness: Boolean) {
+private[mill] sealed class AggWrapper(strictUniqueness: Boolean) {
```
lihaoyi, line 10: Private so it doesn't show up in scaladoc

main/api/src/mill/api/Ctx.scala
```diff
@@ -82,8 +127,7 @@ class Ctx(
     with Ctx.Workspace {
 
   def dest: Path = dest0()
-  def length: Int = args.length
-  def apply[T](index: Int): T = {
```
lihaoyi, line 86: These weren't really used anywhere, and were just thin shims over `args` anyway, so I removed them

main/api/src/mill/java9rtexport/Copy.java
```diff
@@ -24,18 +24,17 @@ LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
 NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
-package io.github.retronym.java9rtexport;
+package mill.java9rtexport;
```
lihaoyi, line 27: Moved this into `mill` so I could make them package private

main/core/src/mill/define/Compat.scala
```diff
@@ -1,12 +0,0 @@
-package mill.define
-
-import scala.reflect.macros.blackbox.Context
-
-object Compat {
-  def copyAnnotatedType(c: Context)(
```
lihaoyi, line 6: This was unused

main/core/src/mill/define/Task.scala
```diff
@@ -43,48 +46,54 @@ abstract class Task[+T] extends Task.Ops[T] with Applyable[Task, T] {
...
 
     def filter(f: T => Boolean) = this
     def withFilter(f: T => Boolean) = this
-    def zip[V](other: Task[V]) = new Task.Zipped(this, other)
+    def zip[V](other: Task[V]): Task[(T, V)] = new Task.Zipped(this, other)
 
   }
 
-  class Sequence[+T](inputs0: Seq[Task[T]]) extends Task[Seq[T]] {
+  private[define] class Sequence[+T](inputs0: Seq[Task[T]]) extends Task[Seq[T]] {
```
lihaoyi, line 57: Privatized all these things so they don't show up in the scaladoc

main/core/src/mill/define/Task.scala
```diff
@@ -106,327 +115,486 @@ trait NamedTask[+T] extends Task[T] {
...
+   */
+  def traverseCtx[I, R](xs: Seq[Task[I]])(f: (IndexedSeq[I], mill.api.Ctx) => Result[R])
+      : Task[R] = {
+    new Task.TraverseCtx[I, R](xs, f)
   }
 
-  def sources(values: Result[Seq[PathRef]])(implicit ctx: mill.define.Ctx): Target[Seq[PathRef]] =
-    macro sourcesImpl2
+  object Internal {
+    private def isPrivateTargetOption(c: Context): c.Expr[Option[Boolean]] = {
```
lihaoyi, line 343: Shoved all the macro impls into `object Internal` so they don't turn up in Scaladoc. They can't be made private due to macro limitations

main/src/mill/main/MainModule.scala
```diff
@@ -180,7 +182,7 @@ trait MainModule extends mill.Module {
       def rec(t: Task[_]): Seq[Segments] = {
         if (seen(t)) Nil // do nothing
         else t match {
-          case t: TargetImpl[_] if evaluator.rootModule.millInternal.targets.contains(t) =>
+          case t: mill.define.Target[_] if evaluator.rootModule.millInternal.targets.contains(t) =>
```
lihaoyi, line 185: This was a bug in https://github.com/com-lihaoyi/mill/pull/2402 that was surfaced when I ported doc snippets to example tests

main/src/mill/main/MainModule.scala
```diff
@@ -204,7 +206,8 @@ trait MainModule extends mill.Module {
         Iterator(
           ctx.applyPrefixColor(t.toString).toString,
           "(",
-          t.ctx.fileName.split('/').last,
+          // handle both Windows or Unix separators
+          t.ctx.fileName.split('/').last.split('\\').last,
```
lihaoyi, line 210: Windows bug surfaced by example tests

runner/src/mill/runner/MillCliConfig.scala
```diff
@@ -1,4 +1,4 @@
-package mill
+package mill.runner
```
lihaoyi, line 1: Moved this to `runner` since that's where it belongs, and to simplify the scaladoc

scalalib/src/mill/scalalib/CrossScalaModule.scala
```diff
@@ -12,7 +12,7 @@ import mill.T
  * - src-2.11
  * - src-2.12.3
  */
-trait CrossScalaModule extends ScalaModule with CrossModuleBase { outer =>
```
lihaoyi, line 15: This was unused

scalalib/src/mill/scalalib/UnidocModule.scala
```diff
@@ -0,0 +1,56 @@
+package mill.scalalib
+import mill._
+
+/**
+ * Mix this in to any [[ScalaModule]] to provide a [[unidocSite]] target that
+ * can be used to build a unified scaladoc site for this module and all of
+ * its transitive dependencies
+ */
+trait UnidocModule extends ScalaModule {
```
lihaoyi, line 9: This is new. We need it for the Mill scaladoc. Other folks probably need it too. It might not be perfect, but I figure let's get something in and people can improve it over time

lihaoyi:
I think this is a natural stopping point for this PR. I didn't managed to review all the pages (see the checklist), and the example/`mill init` integration isn't there yet, and there are a lot of other improvements to make. But this PR is already pretty huge, and stacking on further improvements will make them un-manageable

build.sc
lefou: We should reuse already defined variables where possible, e.g. `Settings.projectUrl` instead of `https://github.com/com-lihaoyi/mill` and others.

main/api/src/mill/java9rtexport/Copy.java
lefou: Great!

scalalib/src/mill/scalalib/UnidocModule.scala
lefou: Very nice. I wonder whether we should put it in another package? I think, we should. The scalalib package contains already lots of various modules. Also, I think we eventually want to provide something like this for `JavaModule`s. A separate package makes grouping that stuff together easier.

lefou (commented):
This is really good work. Thanks for doing it! The documentation is definitely in a better shape after we merge this. I'd like to inspect some smaller aspects around the page generation process and will merge afterwards, ok?

lihaoyi:
Sure, take your time to go over it

lihaoyi:
@lefou I tweaked the HTML a bit, removed a lot of the extraneous UI components to make it less noisy and distracting

Before
Screenshot 2023-04-22 at 7 16 06 PM


After
Screenshot 2023-04-22 at 7 12 27 PM

build.sc
```diff
@@ -1559,191 +1650,301 @@ object dev extends MillModule {
...
+       |  title: Mill
+       |  url: ${if (authorMode) s"${T.dest}/site/index.html" else Settings.docUrl}
+       |  start_page: mill::Intro_to_Mill.adoc
+       |
+       |content:
+       |  sources:
+       |    - url: ${if (authorMode) baseDir else Settings.projectUrl}
+       |      branches: []
+       |      tags: ["0.10.12", "0.9.12"]
+       |      start_path: docs/antora
```
lefou, line 1784: I'd expect some issue once we include tag `0.11.0-M8` which already moved `docs/antora` -> `docs`.

The workaround could be to switch from `tags` to `branches`, then create doc-branches from these current tags and in these branches, move the location to what we expect.

lihaoyi: Yes. We could also leave this content-source unchanged as `tags`, and add a new content type for the `branches` going forward.

lefou: Yeah, I just re-discovered that we can use as many content sources for distinct tags/branch sets, so this should be not an issue. But I guess, we need to add all those generated/copied files and add them to the branches, so that antora can work it out. The alternative would be to switch to a folder-based approach, where we fetch all tags/branches ourselves and copy/generate the missing files.

lefou (commented):
This PR looks good to me. The edit links are currently broken, but I haven't found a quick solution. Also, some some of the copied/generated files it will be hard to provide a edit link. I've not found a way so far, to provide an edit URL per file, only per source set. https://docs.antora.org/antora/latest/playbook/content-edit-url/

Would be nice to find a solution in a follow up PR or disable the edit links (for now).

lihaoyi:
We can probably get the edit links working the same way we do scaladoc source links: walking over the HTML and regex-replacing the broken links with the links we want :p

I'm happy to leave them out for now, they're a nice to have but not strictly necessary. The github repo has all the sources and searching for a snippet of text on the page is usually enough to bring you exactly to where you want to go

PR #2451, opened 2023-04-23
Fix spurious `T.command` definitions must have 1 parameter list error

lihaoyi:
This used to happen when a target had a compile error, causing it's return type to be `Nothing`. This PR explicitly ignores that case

Tested manually. For some reason, the `compileError` macro in uTest isn't able to pick up this error, so I can't catch it and assert


lefou (commented):
Nice fix.

@lihaoyi Can you please fill the milestone field when closing prs/issues.

Issue #1948, opened 2022-07-14
JavaModule's `run` command arguments should be a `Task`


daddykotex:
Currently, we can't define a `def a: Task[String] = ???` and use that to call the `run` command from the `JavaModule`. See the minimal repo:

```scala
import mill._
import mill.scalalib._

object proj extends JavaModule {

/*

/private/tmp/def/build.sc:10: Target#apply() call cannot use `value args` defined within the T{...} block
    run(args)()
        ^
/private/tmp/def/build.sc:8: `T.sources` definitions must have 0 parameter lists
  def aCommand() = T.command {
      ^
Compilation Failed
*/
  def myArgs: T[String] = T { "arg" }

  def aCommand() = T.command {
    val args = myArgs()
    run(args)()
  }
}

```

As per a [conversation](https://gitter.im/lihaoyi/mill?at=62d01d0b9f73251a2c2e6d8b) on `gitter` with @lefou , a straightforward fix for that is to turn `def run(args: String*): Command[Unit]` into `def run(args: Task[Seq[String]]): Command[Unit]`

lefou:
* See also https://github.com/com-lihaoyi/mill/discussions/1857

lihaoyi:
I tried a prototype in https://github.com/com-lihaoyi/mill/pull/2452. Thinking about this, it appears there are a few options here:

1. `def run(args: Task[String]*)`: That should work already I think, but it limits the number way arguments can be passed to `run` since the number of arguments is fixed up front

2. `def run(args: Task[String*])` is not allowed by the Scala type system

3. `def run(args: Task[mainargs.Leftover[String]])` would require some patches to the `mainargs` library, because currently we hard-code support for `mainargs.Leftover[T]`. We'd need to un-hard-code it somehow to make the `Task[Leftover[String]]` param signature work, and then maybe provide an implicit constructor so people can pass in `run(Seq("foo"))` rather than `run(mainargs.Leftover(Seq("foo")))` at the callsite

4. Leave `def run(args: String*)` and just make it forward to an anonymous task `def run(args: Task[Seq[String]]) = T.task{...}`. This is probably the easiest thing to do to solve the immediate issue without major surgery, but doesn't generalize to other `T.command`s unless we repeat the boilerplate of `T.command`/`T.task` pairing for each one

It feels like either (3) or (4) are the best options here, depending on how much we want to generalize this v.s. how difficult it is to make the upstream `mainargs` changes necessary to support (3)

[...]

PR #2452, opened 2023-04-23
Make `run` command arguments Tasks

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/1948. Should probably be reviewed concurrently with https://github.com/com-lihaoyi/mainargs/pull/62 which it depends on

# Major Changes

1. Pull in https://github.com/com-lihaoyi/mainargs/pull/62, which allows MainArg's "aggregate leftover arguments" logic to apply to custom types and not be hardcoded to `mainargs.Leftover`

2. Update MainArgs integration code in this repo to accomadate the changes

3. Replace the `def run(args: String*)` with `def run(args: Task[Args] = T.task(Args()))`.

# Minor Changes

1. I moved the `import mill.main.TokenReaders._` into the `Discover` macro, so we can stop manually importing it everywhere.

2. I moved all the aliases from `import mill.define._` to `import mill._`. `Task[_]` in particular is a type I believe we should start encouraging people to use more often - e.g. annotating all command parameter types as `Task[_]`s - and so it should be in the default `import mill._` without needing a separate import.

3. Added a direct dependency to `com.lihaoyi::pprint`, since the implicit dependency in `mainargs` was removed in 0.5.0 due to being unnecessary

# Testing

1. All existing tests pass, at least those I've run locally so far (`./mill -i -j 3 example.__.local.test`, can't use CI until the mainargs PR is published). This includes unit tests that call `run` programmatically and integration/example tests that call `run` via the CLI.

4. Added a new example test `5-module-run-task`, that demonstrates and explains how to use a `ScalaModule`'s `run` method to implement a `task`, passing in input parameters from upstream tasks and using the output (in this case generated files) in downstream tasks

# Notes

1. There is still some boilerplate around `run`, both defining it `args: Task[Args] = T.task(Args())` and calling it programmatically, `run(T.task(Args(...)))`. For now I haven't figured out how to reduce this, as we already "used up" our implicit conversion budget with the `T.*{...}` macros. Probably can be done, although it won't be quite as low-boilerplate as the original `args: String*` syntax. Maybe in a follow-up PR

2. The example test `5-module-run-task` is something I've wanted to be able to write for years now. Nice to finally be able to do it without having to piece together a `Jvm.runSubprocess()` call myself!

3. `5-module-run-task` is still a bit clunkier than I was hoping. Having to `def barWorkingDir` beforehand to pass it to both the `bar.run` and the `def sources` override is weird and unintuitive. Maybe we can change the `def run` return type to make it easier to return the `T.dest` folder or other metadata out of the body of  `def run` so it can be used by downstream tasks? e.g. making `def run` return `Command[PathRef]` rather than `Command[Unit]`

5. IMO we should recommend that all CLI arguments to `T.command`s come in the form of `Task[T]`s, unless there's concrete reasons otherwise (e.g. we want to dynamically change `T.command` implementations based on the input). That would maximise the ability to re-use commands in the body of of tasks, which is something people probably want


lefou:
* Related: https://github.com/com-lihaoyi/mill/discussions/1857

lihaoyi:
https://github.com/com-lihaoyi/mainargs/pull/62 has the mainargs necessary to make (3) work. It would be a binary compatibility breakage for Mainargs, but IMO it's probably worth doing sooner or later, as apart from supporting this feature it would also clean up a lot of cruft within the MainArgs codebase

lefou (commented):
Is there a reason to use `mainargs.Leftover` instead of `Seq[String]`?

lihaoyi:
@lefou this was just the shortest path to verify it working end-to-end; with that mainargs PR de-hardcoding things it should be possible to swap out the type for whatever we want

We probably shouldn't go with `Seq[String]`, because that already has meaning in mainargs as a repeatable `--foo bar` param, and re-using it for varargs would be confusing/conflicting. But we can make our own `mill.Args(value: Seq[String])` with implicit constructors to remove boilerplate

lihaoyi:
@lefou this should be ready to review, looks like CI is green

main/define/src/mill/define/Discover.scala
```diff
@@ -114,7 +114,7 @@ object Discover {
       }
 
       c.Expr[Discover[T]](
-        q"_root_.mill.define.Discover(_root_.scala.collection.immutable.Map(..$mapping))"
+        q"import mill.main.TokenReaders._; _root_.mill.define.Discover(_root_.scala.collection.immutable.Map(..$mapping))"
```
lefou, line 117:
```suggestion
        q"import _root_.mill.main.TokenReaders._; _root_.mill.define.Discover(_root_.scala.collection.immutable.Map(..$mapping))"
```

lefou (approved):

PR #2463, opened 2023-04-25
Rebootstrap Mill on 0.11.0-M9

lihaoyi:


lihaoyi:
Rebootstrapping fails due to https://github.com/com-lihaoyi/mill/pull/2471, looks like we'll be needing a 0.11.0-M9 to re-bootstrap successfully

lihaoyi:
@lefou do you know what the `ivyDepsTree` failure is about? I've never seen that one before.

lefou:
> @lefou do you know what the `ivyDepsTree` failure is about? I've never seen that one before.

Me too. Looks a bit like a internal hiccup, maybe in coursier. I can't reproduce locally, so re-running may solve it.

lihaoyi:
Looks like it's caused by a bugfix in mainargs (https://github.com/com-lihaoyi/mainargs/pull/66). Previously, we were incorrectly passing the `true` to the `--withRuntime` flag. Now, it gets parsed as a separate positional argument, end up in `whatDependsOn` where it's treated as an invalid module. Removing the spurious `true` should fix it

```diff
-              ./mill -i -k __.ivyDepsTree --withRuntime true
+              ./mill -i -k __.ivyDepsTree --withRuntime
```

Issue #2468, opened 2023-04-26
Integration tests fail on Windows with Mill 0.11.0-M8


lefou:
My integration tests for mill-vcs-version plugin with Mill 0.11.0-M8 fail on Windows.

https://github.com/lefou/mill-vcs-version/actions/runs/4801930870/jobs/8544774098?pr=102

Here is the output:

```
Output of failed test: 01-simple
---------- Output for Invocation: TestInvocationResult(Targets(targets=List(-d, verify1),expectedExitCode=0,env=Map(),noServer=true),Failed,Vector(),Vector(),Some(D:\a\mill-vcs-version\mill-vcs-version\out\itest\0.11.0-M8\testCached.dest\01-simple\out-3280110244997307355.log))

D:\a\mill-vcs-version\mill-vcs-version\out\itest\0.11.0-M8\testCached.dest\01-simple>set JAVA_OPTS="-Divy.home=D:\a\mill-vcs-version\mill-vcs-version\out\itest\0.11.0-M8\testCached.dest\ivyRepo" 

D:\a\mill-vcs-version\mill-vcs-version\out\itest\0.11.0-M8\testCached.dest\01-simple>"C:\Users\runneradmin\.cache\mill\download\0.11.0-M8.bat" -i --color false -d verify1 
Exception in thread "main" java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at mill.main.client.IsolatedMillMainLoader.runMain(IsolatedMillMainLoader.java:58)
	at mill.main.client.MillClientMain.main(MillClientMain.java:64)
Caused by: java.lang.IllegalArgumentException: requirement failed: \home\runner\work\mill\mill\scalalib\src\mill\scalalib\JavaModule.scala is not an absolute path
	at scala.Predef$.require(Predef.scala:337)
	at os.Path.<init>(Path.scala:453)
	at os.Path$.apply(Path.scala:405)
	at mill.eval.Evaluator.$anonfun$evaluateGroupCached$8(Evaluator.scala:322)
	at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:575)
[...]

PR #2471, opened 2023-04-26
Filter out non-script paths to avoid blow up due to invalid path format

lihaoyi:
Should fix https://github.com/com-lihaoyi/mill/issues/2468 I think


lefou (approved):
Looks good enough to try it out.

lefou:
This fixes the Windows issues. Thanks!

Issue #2474, opened 2023-04-27
Mill special syntaxes to import plugins and contrib modules doesn't work anymore


lolgab:
```scala
import $ivy.`com.github.lolgab::mill-crossplatform::0.2.1`
```

should desugar to:
```scala
import $ivy.`com.github.lolgab::mill-crossplatform_mill0.11.0-M8:0.2.1`
```

but, instead, desugars to:
```scala
import $ivy.`com.github.lolgab::mill-crossplatform:0.2.1`
```

Moreover:
```scala
import $ivy.`com.lihaoyi::mill-contrib-bloop:`
```

doesn't desugar anymore to:
```scala
import $ivy.`com.lihaoyi::mill-contrib-bloop:$MILL_VERSION`
```

lefou:
Partial duplicate of #2457,but this once covers also the platform, hence I'll close the other one as duplicate.

PR #2377, opened 2023-03-23
Remove Ammonite as a dependency, handle script running and bootstrapping ourselves

lihaoyi:
The basic idea is that instead of relying on Ammonite to evaluate the `build.sc` and return us a `mill.define.BaseModule` value, we instead

1. First instantiate an in-memory `MillBuildRootModule`, which is a special `ScalaModule` configured to read `build.sc` and transform it into `Build.scala` before compiling it with the necessary classpath

2. Use a `mill.eval.Evaluator` to call `runClasspath` on the `MillBootstrapModule`, which gives us a list of files and folders that are necessary to use it

3. Put the files and folders into a `java.net.URLClassloader` to load it into the JVM, and use java reflection to fish out the final `mill.define.BaseModule` runtime value

These are all largely things that were already happening before, just inside the guts of the Ammonite script runner, and mixed in with a bunch of other Ammonite REPL stuff we don't want, and via parallel code paths (e.g. `ammonite.compiler.Compiler` vs `mill.scalalib.worker.ZincWorker`). By pulling them out, we can re-use the bulk of `ScalaModule` logic we already have, and just need to do the minimal amount of classloader plumbing to get us what we want

## Major Code Changes

1. In order to let us depend on `ScalaModule` within `MillMain`, we need to split out `MillMain` from the rest of `main/` to avoid a circular dependency with `scalalib/`. For now I dumped them in a `runner/` module that depends on both `main/` and `scalalib/`, along with similar classes like `BspContext.scala`, `GraphUtils.scala`, etc.

2. Vendored/adapted a bunch of Ammonite code: `Parsers.scala`, `Classpath.scala`, `Colors.scala`, `class AnsiNav`, `Watchable.scala`, `LineNumberPlugin.scala`, `Util.scala`

3. The old `MillMain`/`MainRunner`/`RunScript` logic has been reorganized into `MillMain`/`MillBuildBootstrap`/`MillBuildModule`:

    a. `MillBuildModule` is a hard-coded `build.sc` file. This module knows how to parse the `build.sc`, walk its import `$file`/`$ivy`, and generate the relevant `generatedSources`/`ivyDeps` for it to be compiled as a normal `ScalaModule` with the appropriate compiler-plugins/scala-version/etc.

    b. `MillBuildBootstrap` performs the main bootstrapping logic: creating a `MillBootModule`, evaluating its `runClasspath`, loading the classfiles into a classloader, grabbing the resultant `BaseModule` from that classloader and then evaluating the targets that the user asked for

[...]

PR #2476, opened 2023-04-28
Re-introduce Mill special handling of ivydeps, overhaul test classpath management to allow testing

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/2474

These were overlooked in https://github.com/com-lihaoyi/mill/pull/2377 and deleted, and were not covered by any tests. But it's straightforward to put them back.

# Major Changes

1. Put back `MillIvy.scala` and call it in `MillBuildRootModule.scala`

# Testing 

1. I added an example test `example/misc/6-contrib-import` to both exercise the code as well as serve as an example we can include in our docs

2. In order to allow usage of contrib module in `.local` integration and example tests, I moved the handling of Mill test classpath overrides from `Util.millProjectModule` to `CoursierSupport#resolveDependencies`. This lets us be more override contrib modules dependency resolution, even though they don't have a neat single location for us to call our `millProjectModule` helper.

3. I refactored `millProjectModule` to not need a `key`, so we just compute the key based on the dependency name, keeping them consistent and removing an unnecessary degree of freedom

4. The local-testing-classpath-overrides were moved from using JVM system properties to instead use classpath resources: we look for overrides in `mill/local-test-overrides/*`. This should remove any security worries: 
    1. Before you only needed to modify the JVM props or `JAVA_OPTS` environment variable, and could replace the code of a Mill module to any local filesystem path
    2. Now, you need access to modify the Mill classpath to trigger the test overrides, and at that point you already have access to modifying the Mill classfiles being executed anyway

While it probably was not strictly necessary to clean up the test classpath overrides logic as part of this PR, the status quo in master was a pile of hacks, and I didn't feel like adding another hack to get `.local` testability of contrib libraries working. With these changes, contrib libraries work the same as the existing Mill test classpath overrides, and things are cleaned up so much that the net lines of code for this PR is negative

There's still more cleanup to do in `build.sc`, but those can come in follow up PRs


lihaoyi:
This should also be reviewable I think

lefou (approved):
Looks good to me.

PR #2478, opened 2023-04-28
Fix crash aggregating failures

lolgab:
The current code is crashing since it's trying to insert twice the same failure when it's linked to multiple tasks.
My solution was to wrap the Iterator into a `Loose.Agg` so we remove the duplicates before adding to the `MultiBiMap`, which uses `Strict.Agg`.

Before:
immagine

After:
immagine

### DIFF
--- a/main/core/src/mill/eval/Evaluator.scala
+++ b/main/core/src/mill/eval/Evaluator.scala
@@ -182,7 +182,11 @@ class Evaluator private (
     for ((k, vs) <- sortedGroups.items()) {
       failing.addAll(
         k,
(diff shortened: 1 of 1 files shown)


lolgab:
Side question, why `MultiBiMap` uses `mill.api.Strict.Agg` instead of `mill.api.Loose.Agg`?

lefou (commented):
Great. (I just saw the same error message with stack trace while experimenting with an implementation for `import $repo`.)

Don't know about the Strict vs. Loose in MultiBiMap. Looks like Loose is enough, but there could be a slight risk to loose data that way, as Loose is just dropping a new element whereas Strict is failing in that case.

lefou (approved):

PR #2480, opened 2023-04-29
Support `import $repo` magic import

### LINKED ISSUE #2477: Restore support for `import $repo`
This was removed as part of #2377, but is IMHO an essential feature already used by our users.

lefou:
Re-added support for "magic" `import $repo`, which was supported in all Mill version before 0.11.0-M8, but got (accidentally) removed as we no longer base on Ammonite.

This PR restores the `$repo` import on a par with previous versions. Documentation needs to be reviewed and we should also look after support of repositories which need authentication.

Fix #2477


main/src/mill/modules/CoursierSupport.scala
```diff
@@ -306,4 +307,32 @@ object CoursierSupport {
     }
   }
 
+  def repoFromString(str: String, origin: String): Option[Seq[Repository]] = {
+    val spaceSep = "\\s+".r
+
+    val l =
```
lihaoyi, line 313: Can we have a better name than `l` here? Maybe `repositoryStringsToParse` or something? AFAICT it's basically the half-parsed repository names that we split before passing to Coursier.

Also, what's the reason we split them differently depending on `spaceSep.findFirstIn(str).isEmpty`? I assume there's a good reason, but probably should go in a comment somewhere

main/src/mill/modules/CoursierSupport.scala
```diff
@@ -306,4 +307,32 @@ object CoursierSupport {
...
+          .split(str)
+          .toSeq
+          .filter(_.nonEmpty)
+
+    RepositoryParser.repositories(l).either match {
+      case Left(errs) =>
+        System.err.println(
+          s"Ignoring $origin, error parsing repositories from it:" + System.lineSeparator() +
+            errs.map("  " + _ + System.lineSeparator()).mkString
+        )
```
lihaoyi, line 330: Can we fail loudly if repository parsing fails, rather than logging and continuing? I think we can make `parseBuildFiles` return a `mill.eval.Result`, and then it'll be nicely shown to the user when the `parseBuildFiles` target fails without an ugly stack trace

example/misc/7-repo-import/build.sc
```diff
@@ -0,0 +1,33 @@
...
+import mill._, mill.define._, mill.scalalib._
+import coursier.Repository
+
+//object foo extends RootModule {
+def verify(): Command[Unit] = T.command {
+  //  val repos: Seq[Repository] = build.asInstanceOf[ScalaModule].repositoriesTask()
+  os.write(T.dest / "paths.txt", s"${T.workspace} || ${T.dest}")
+  val model = os.read(T.workspace / "out" / "mill-build" / "parseBuildFiles.json")
+  assert(model.contains("""file:///tmp/testrepo""""), "Model was unexpected: " + model)
+}
```
lihaoyi, line 11: The goal of the example tests is to show a semi-realistic workflow that is relevant to the user, while also serving as a test case. 

I think the easiest setup here would be to just manually commit a minimal `file://` repository to the example folder, using the `example/scalamodule/7-unmanaged-jars/lib/nanojson-1.8.jar` file we're already using to keep thing small, but under a different jar name that doesn't exist on maven central. Then we can `import $repo` that local file repo and `import $ivy` our custom jar name and demonstrate usage of `nanojson` inside our `build.sc`

main/src/mill/modules/CoursierSupport.scala
lefou: This code is copy-paste as-is from Ammonite and I just wanted the import working. Feel free to change it as you want.

example/misc/7-repo-import/build.sc
lefou: Two reasons why I didn't use some other file. I don't know how I can publish some file before the example build.sc is run. And also, I don't know how to refer to a project-local repository as file URL from the `import $repo` statement in the example `build.sc`.

main/src/mill/modules/CoursierSupport.scala
lefou: Good idea. Feel free to change it, I will be some hours away from my laptop.

example/misc/7-repo-import/build.sc
lihaoyi: Maybe lets move it to the `integration/feature/` folder then? Those are the tests which are meant to assert things but aren't presentable enough for a user to read or include verbatim in the docsite

lefou: Ok

lefou: Done

main/src/mill/modules/CoursierSupport.scala
lefou: Done

lihaoyi (approved):

main/src/mill/modules/CoursierSupport.scala
lefou: I changed it to return a Result to propagate the parse error.

Issue #2479, opened 2023-04-29
Document build setup / capabilities / syntax for `example/build.sc` 


lefou:
Mill can automatically build project under `example/` directory, run them as integration tests with specific build command, and generate documentation from it. This is currently not documented.

lefou:
Systax of example `build.sc` is documented here: https://github.com/com-lihaoyi/mill/blob/87d5b926999633562ae01e4c6c0cb800f5f9ad36/example/src/mill/integration/ExampleTestSuite.scala#L6-L42

lihaoyi:
I'll write something up in the readme

PR #2483, opened 2023-04-30
Consolidate manual testing, automated testing, CI and release configurations

lihaoyi:
# Configuration Matrix Changes

| Config | Automated Testing | Manual Testing | Manual Testing CI |
|:-----|:----|:-----|:----|
| **BEFORE THIS PR** |
| In-Process unit tests, `TestEvaluator` | `scalalib.test`, `contrib.buildinfo.test`, etc. |  | |
| In-Process `MillBuildBootstrap` | `example.__.local`, `integration.__.local` | | |
| Sub-Process w/o packaging/publishing |  | `dev.run` |  |
| Sub-Process w/ packaging w/o publishing |  | `dev.assembly` | `test-mill-dev.sh`  |
| Sub-Process w/ publishing| `example.__.{fork,server}`, `integration.__.{fork,server}` |  | |
| Bootstrapping | | `installLocal` | `test-mill-release.sh`, `test-mill-bootstrap-{0,1}.sh` |
| **AFTER THIS PR** |
| In-Process | `scalalib.test`, `contrib.buildinfo.test`, etc. | | |
| Sub-Process w/o packaging/publishing| `example.__.local`, `integration.__.local` | `dev.run` | `test-mill-dev.sh` |
| Sub-Process w/ packaging/publishing| `example.__.{fork,server}`, `integration.__.{fork,server}` | `dev.assembly` | `test-mill-release.sh` |
| Bootstrapping |  | `installLocal` | `test-mill-bootstrap.sh` |

Essentially, we collapse the 6 different manual/automated/CI-testing setups down to just 4, ensuring that every automated test workflow can be debugged manually and ensuring that every example can be run.

## Automated Testing Changes 

1. This PR makes `.local` tests launch in a subprocess using the `dev.launcher`, rather than running in-memory via 

## Manual Testing Changes

1. Replaced `dev.assembly` with `assembly`. They did more or less the same thing, and were similar enough it didn't make sense to maintain two copies of it. Might as well use the real "production" assembly for the integration-testing use cases that we were previously using `dev.assembly` for

## CI Changes

1. Move `test-mill-dev` to test the `dev.run` manual testing workflow, which was previously un-tested. I just run it on `example/scalabuilds/10-scala-realistic` to perform some basic operations to verify that they are successful

2. `test-mill-release` now tests the `dev.assembly` manual testing workflow. It also just runs a few basic operations on `example/scalabuilds/10-scala-realistic` as a smoketest

2. Combined `test-mill-bootstrap-0` and `test-mill-bootstrap-1` into `test-mill-bootstrap`; they never seemed to catch things that `test-mill-release` didn't, and were previously very slow. Now we do 1 level of bootstrapping instead of 2, and run a much smaller/faster selection of tests:
    1. The same `example/scalabuilds/10-scala-realistic` tests that `test-mill-dev` runs
    4. `__.compile` to make sure everything builds
    5. `main.__.test` to make sure the fast unit tests pass
    6. `scalalib.__.test` to make sure the tests actually building Scala with `TestEvaluator` and using our local classpath overrides are working
    7. `example.basic[1-simple-scala].server.test` to make sure our example/integration tests work

# Notes

1. While we do lose some coverage, any bugs that the removed-tests would have caught would would get caught during the re-bootstrapping process anyway, so if we re-bootstrap at least once per release, that would catch everything else without paying the heavy cost of running all the expensive tests every single PR.

2. This change reduces the number of degrees of freedom we have in selecting how to test things, which does give us less flexibility. However, I expect that the current status quo is already un-manageable, and thus the extra freedom just gives us confusion rather than flexibility. 

3. In exchange, this PR helps ensure that the various automated-manual-CI testing workflows line up nicely, and reduces both the potential divergence between those workflows and the maintenance burden by allowing us to share implementations and configurations for each one. e.g. we can be more sure that everything we integration/example test with `.local` can also be run via `dev.run`, whereas previously there was some cases that couldn't

5. Moving `.local` tests from in-process to sub-process doesn't seem to have a significant impact on performance, with `example.basic[1-simple-scala].local.test` taking ~14s on my laptop to run in either case. It also lets us remove a bunch of copy-pasty code from `IntegrationTestSuite.scala`, reducing the maintenance burden

8. It's not immediately clear to me what is the difference between `assembly` and `dev.assembly`. For now I just kept the `assembly` target and got rid of the other one, to keep the published artifact un-changed, even though the `assembly` target is significantly more complicated. We can figure out if `assembly` can be simplified in a follow up PR

9. The `local`/`fork`/`server` matrix isn't actually complete; there are two axes `{publish-package,not-publish-package}X{server,no-server}`, and the current names do not have a good indication of which one is which. Cleaning this up can come in a follow-up

10. I included a rewrite of `readme.adoc`, so this should fix https://github.com/com-lihaoyi/mill/issues/2479

11. A cursory benchmark summing up the CI times indicates that this cuts down CI from ~700minutes to ~500minutes. We can probably cut further, but this should be a good speedup for now to avoid long waits for PR validation


lihaoyi:
CI is mostly green, should hopefully be green after this run

lefou:
> * While we do lose some coverage, any bugs that the removed-tests would have caught would would get caught during the re-bootstrapping process anyway, so if we re-bootstrap at least once per release, that would catch everything else without paying the heavy cost of running all the expensive tests every single PR.

I think the assumption to re-bootstrap the Mill build on each new Mill release is, especially when doing milestones, not ideal. Maybe, we can just run some more complete nighty builds, so nobody needs to wait for them, but we will get a daily report, whether be broke something less obvious.

> * It's not immediately clear to me what is the difference between `assembly` and `dev.assembly`. For now I just kept the `assembly` target and got rid of the other one, to keep the published artifact un-changed, even though the `assembly` target is significantly more complicated. We can figure out if `assembly` can be simplified in a follow up PR

I think `assembly` is reusing the `dev.assembly` artifact and enhances it with some "executable" properties, e.g. it prepends the JAR with a mixed executable Windows Batch and Shell script. I think there is no value to distribute the `dev.assembly` but instead we should distribute `assembly`, as it is identical to out released assembly on GitHub, but not all users can download from GitHub easily. Just using Maven Central which can be proxied and cached seems to be a huge win.

Currently, `assembly` is handmade and not part of a `JavaModule`, so publishing it was not directly possible.

lihaoyi:
Do you know of any examples where the old bootstrap tests caught things that the other tests didn't? I'm not aware of any. I think if we find some new common failure modes we can grow the CI matrix to compensate, but I don't think it makes sense running e.g. the double-bootstrapped scalajs tests just for the heck of it.

With this PR, `test-mill-{dev,release,bootstrap}` act as smoketests for the three main manual development workflows that we support, rather than some arbitrary selection of very slow tests.

This PR already keeps `assembly`, i just move it to the `dev.assembly` path to keep things neat, but the logic is all the old `assembly` logic

lefou:
> Do you know of any examples where the old bootstrap tests caught things that the other tests didn't? I'm not aware of any. 

Me neither. I'm all-in in removing the second bootstrap level. But I read your comment so, that you wanted to re-bootstrap Mill before each release to make sure we can do it. I don't think that is necessary and I also don't want to be forced to use the latest experimental version of Mill on Mill itself "in production". I wanted to say that, if we come up with some good test setup that has the chance to catch even more things but it runs longer, we could just run some nightly tests.

> I think if we find some new common failure modes we can grow the CI matrix to compensate, but I don't think it makes sense running e.g. the double-bootstrapped scalajs tests just for the heck of it.

I agree.

> This PR already keeps `assembly`, i just move it to the `dev.assembly` path to keep things neat, but the logic is all the old `assembly` logic

Ok, great. We should also publish it (I think I disabled it in the past, because the old `dev.assembly` was pretty useless for others, but wasted a lot of space) and communicate it as an alternative to GitHub Releases.

lefou (approved):

lihaoyi:
Great. Yes, I'd like to publish the assembly to maven central, and make downloading from maven central the default. Having 1 SPOF in Maven Central is better than having 2 SPOFs in Maven Central & Github, w.r.t. firewall rules and security concerns & other things. I'll do that in a follow up

Issue #909, opened 2020-05-27
`show` combined with `--watch` does not watch files if evaluation fails


lihaoyi:
Seems like we're not properly propagating the watched files out of `def show` in the case of failure

lefou:
My suspicion is, that this is true for all evaluator command.

Issue #838, opened 2020-04-29
`--watch` does not work for `T.input`


lefou:
See also related issue #832 and it's fix 2b3fb0578233de00feea5f34a09a6736e0e456ce.

lefou:
Created a PR #839 which tackles the issue for most common types: `PathRef`, `Seq[PathRef]`, `os.Path` and `Seq[os.Path]`.

lefou:
A possible solution looks like this (copied from #839): We want to plumb the lambda that T.input captures and change the rest of the RunScript infrastructure that does watching/etc. to know how to handle either PathRefs or lambdas, instead of being hard-coded to PathRefs as it is now. We could even re-use the ammonite.interp.Watchable type, since that models exactly this and is used for exactly the same use case

PR #2489, opened 2023-05-01
[WIP] Add ability to watch `T.input`s and `interp.watchValue`s

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/838 and also fixes https://github.com/com-lihaoyi/mill/issues/909

We add the ability to re-run `T.input{...}` blocks during the polling loop, just like we re-stat the `os.Paths` reported by `T.source` and `T.sources`. This involves some complexity, since `T.input{...}` blocks are not just a single lambda but instead a wrapper for a part of the `Task` graph that feeds into it. We capture the body of `evaluateGroup` in a lambda so we can call it later

# Major Changes

1. Modify `Evaluator.Results#results` from a `Map[Task, Result]` to `Map[Task, TaskResult]`, where `TaskResult[T]` is a wrapper of `Result[T]` that also contains a `recalcOpt: Option[() => Result[T]]` callback

2. Wrap the bulk of `Evaluator#evaluateGroup` in a function, that we evaluate once to populate `TaskResult#result` and store in a callback to populate `TaskResult#recalcOpt`

3. Update `RunScript.evaluateNamed` to handle `InputImpl`s, in addition to the current handling of `SourceImpl`s and `SourcesImpl`s, and translate the `recalcOpt` callback into a `Watchable.Value`

4. The `Watchable.Value` is then propagated to `MillBuildBootstrap.evaluateWithWatches`, which ends up feeding it into `Watching.scala` where it is processed the same way as the `Watchable.Value`s we get from `interp.watchValue`

5. Overhauled the implementation of `show`; rather than returning a `Watched` wrapper, we now call back to `interp.evalWatch0` to register the things we need to watch. I also consolidated the common code in `show` and `showNamed` into one shared helper.

6. I wrapped most of the `Any`s or `_`s in `Evaluator` with a new `case class Val(value: Any)` wrapper. This should considerably increase type safety: e.g. you can no longer pass an `Option[Val]` where a `Val` was expected, unless you manually wrap it in `Val(...)` which is hard to do accidentally. This makes it much easier to ensure the various data structures inside `Evaluator` are passed around correctly and not mis-used

# Minor Changes

1. Cleaned up `RunScripts` considerably, now there's only two methods left and no copy-pasty code

2. Removed some `T.input`s from `MillBuildRootModule`: `def unmanagedClasspath` can be a `T.sources`, and `def parseBuildFiles` is now a normal target that depends on `allBuildFiles` which is a `T.sources`. That should improve the `--watch` status message in the terminal and reduce the amount of work that needs to be done every 100ms while polling for changes

# Testing 

Tested manually with the following `build.sc`; verified with `-w bar` that `bar` re-runs every 2s, and the script re-evaluates every 10s

```scala
import mill._

interp.watchValue(System.currentTimeMillis() / 10000)

println("Setting up build.sc")
def foo = T.input{ System.currentTimeMillis() / 2000 }

def bar = T{ foo() + " seconds since the epoch" }
```

Also added `integration/feature/watch-source-input/` suite that starts up a `--watch` in a background thread, makes changes to the files on disk, and ensures the `build.sc` and tasks within it re-evaluate as expected.

# Notes

1. I'm not sure if the new way `show` registers watches is better than the old one. The old way with `Watched` can be made to work even in failure cases if we pass it as the optional second parameter to `Result.Failure`

2. The `Val` change is not strictly necessary, and is pretty invasive. But it was very hard to make the necessary changes to `Evaluator` without it due to bugs passing the wrong `Any`s around. So I think it's a reasonable time to introduce `Val` to support future work inside `Evaluator` with increased compiler support and type safety

3. The `recalc` logic would be a lot simpler if `input`s couldn't have upstream tasks, since we could just re-run the block rather than re-evaluating the task group. I had some thoughts about restricting `T.input`s to allow this, but couldn't convince myself that was the right thing to do, so left the `T.input` semantics in place and just made it work by capturing a callback that can evaluate the entire task group


main/src/mill/main/VisualizeModule.scala
```diff
@@ -43,7 +43,7 @@ trait VisualizeModule extends mill.define.TaskModule {
     )
     val visualizeThread = new java.lang.Thread(() =>
       while (true) {
-        val res = Result.create {
+        val res = Result.Success{
```
lefou, line 46: This version does not catch exceptions and converts them into `Failing`. Is this on purpose?

main/api/src/mill/api/Result.scala
```diff
@@ -27,11 +27,14 @@ object Result {
    * @param value The value computed by the task.
    * @tparam T The result type of the computed task.
    */
-  case class Success[+T](value: T) extends Result[T] {
-    def map[V](f: T => V): Success[V] = Result.Success(f(value))
+  case class Success[+T](value: T, recalc: () => T) extends Result[T] {
```
lefou, line 30: Can you elaborate on the purpose of `recalc` and add some API doc?

I only have a fuzzy idea how I should fill the `recalc` arg of a newly created `Success`.

lefou (commented):
Happy to see the `Watched` removed. But I don't get the `recalc` thing yet.

scalanativelib/test/src/mill/scalanativelib/HelloNativeWorldTests.scala
```diff
@@ -38,8 +38,8 @@ object HelloNativeWorldTests extends TestSuite {
         extends Cross.ToSegments[ReleaseMode](v => List(v.toString))
 
     val matrix = for {
-      scala <- Seq("3.2.1", "3.1.3", scala213, "2.12.13", "2.11.12")
```
lefou, line 41: Is this version no longer supported?

scalajslib/test/src/mill/scalajslib/HelloJSWorldTests.scala
```diff
@@ -28,8 +28,8 @@ object HelloJSWorldTests extends TestSuite {
   }
 
   object HelloJSWorld extends TestUtil.BaseModule {
-    val scalaVersions = Seq("2.13.3", "3.0.0-RC1", "2.12.12", "2.11.12")
-    val scalaJSVersions = Seq("1.8.0", "1.3.1", "1.0.1")
```
lefou, line 32: Are these versions no longer supported? AFAIK, those partly changed/extended their API, and those tests make sure we can still build against them.

lihaoyi: I thought we could kill them, but I'll revert this and leave it for another PR

lihaoyi:
@lefou this should be ready for another round of review I think, CI is finally green

lefou (approved):
Making Evaluator even more complex will probably hit use later (at least me, as complex code tends to see less (necessary) changes), but I also think that `T.input` and in general a more robust watch implementation is a very nice thing.

I was hoping, that we somehow could track watchable values directly from Evaluator and keep it in some context, as the current infrastructure you build up in `MainModule` to support `show` and `showNamed` still looks a bit fragile and specific. I wonder whether each evaluator target needs to deal with this kind of handling at some point?

I think, this change is OK to merge.

lihaoyi:
Yeah `Evaluator` has definitely grown a bit messy. I can see if I can refactor it a bit in a follow up

PR #2490, opened 2023-05-02
Put BSP timing tests behind a retry, as these fail sometimes on Windows

lefou:


lihaoyi (approved):
I've noticed windows CI performance is realy variable. The `integration.feature.__.local` tests take anywhere from 30-90min to run, huge amount of variation

PR #2494, opened 2023-05-03
[WIP] Reviewed and updates some target types

lefou:
- Reviewed return types of major modules
- Changed GenIdeaModule targets from `Command` to `Task` and updated `IdeaConfigFile`


runner/src/mill/runner/MillBuildRootModule.scala
```diff
@@ -179,7 +179,7 @@ object MillBuildRootModule {
       baseModuleInfo.discover.asInstanceOf[Discover[this.type]]
   }
 
-  case class Info(
+  private case class Info(
```
lefou, line 182:
```suggestion
  case class Info(
```

PR #2498, opened 2023-05-04
Cleanup eval

lihaoyi:


main/eval/src/mill/eval/Evaluator.scala
```diff
@@ -750,225 +45,64 @@ object Evaluator {
...
-        case r if r.failing.items().nonEmpty =>
-          throw exceptionFactory(r)
-        case r => r.values.map(_.value).asInstanceOf[Seq[T]]
-      }
-  }
+  // This needs to be a ThreadLocal because we need to pass it into the body of
+  // the TargetScopt#read call, which does not accept additional parameters.
+  // Until we migrate our CLI parsing off of Scopt (so we can pass the BaseModule
+  // in directly) we are forced to pass it in via a ThreadLocal
+  val currentEvaluator = new ThreadLocal[mill.eval.Evaluator]
```
lefou, line 83: How about making this a `scala.util.DynamicVariable`, which uses a `ThreadLocal` internally, but handles re-setting and cleanup more safely through the use of `withValue`?

lihaoyi: done

lefou (approved):
Looks good to me.

I also had thought about unifying parallel and sequential evaluation before, but wanted to avoid to add the overhead of all these Futures to the sequential evaluation. Do you have any idea how much overhead this produces? Probably nothing to worry about, I guess.

lihaoyi:
Yeah the `Future` overhead should be negligible, especially with the `RunNow` execution context that doesn't indirect through a thread pool. IIRC last time I profiled the Mill evaluator most of the time was spent in disk IO reading/writing `meta.json` cache files, which we create a lot of

lefou (commented):
IIUC, we now also produce a `out/mill-par-profile.json` in the sequential case, which is a nice thing. 

Maybe, we can drop the `out/mill-profile.json` now. The "cached" attribute is also in the `mill-par-profile.json`. But it's easier to read without a tool.

PR #2505, opened 2023-05-06
Tweak CI to improve UI display

lihaoyi:
Now I can actually see which windows job is running what without it being truncated

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -100,31 +100,31 @@ jobs:
         java-version: [8, 17]
         buildcmd:
           # Run a stripped down build matrix on Windows, to avoid taking too long
-          - cmd /C %GITHUB_WORKSPACE%\ci\mill.bat -i -d -k "{__.publishLocal,dev.assembly,__.compile}"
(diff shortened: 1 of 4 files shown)


.github/workflows/actions.yml
```diff
@@ -93,43 +93,43 @@ jobs:
...
+          - "{__.publishLocal,dev.assembly,__.compile}"
+          - "{main,scalalib,scalajslib,bsp}.__.test"
 
         include:
           # Limit some tests to Java 17 to limit CI times
           - java-version: 17
             # just run a subset of examples/ on Windows, because for some reason running
             # the whole suite can take hours on windows v.s. half an hour on linux
-            buildcmd: cmd /C %GITHUB_WORKSPACE%\ci\mill.bat -i -d -k "example.{scalabuilds,cross,tasks}.__.local.test"
+            buildcmd: "example.{scalabuilds,cross,tasks}.__.local.test"
```
lefou, line 111: We should rename `buildcmd` to `millargs`

.github/workflows/actions.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           - ./mill -i -k "integration.feature.__.server.test"
           - ./mill -i -k "integration.feature.__.fork.test"
 
-          # contrib tests
+          # contrib s
```
lefou, line 45: ?

PR #2511, opened 2023-05-07
More refactoring for Resolve logic

lihaoyi:
# Major Changes

1. Break out `main/resolve/` from `main/`, depending on `main/define/` but *not* depending on `main/eval/`. 
    1. Resolving tasks should not depend on `Evaluator`, and the core logic of `Evaluator` should not depend on task resolution (even if individual tasks might require it). 
    2. Breaking it up into a separate modules enforces that, and helps simplify each module so the complex logic within each one can be reasoned about independently.

2. Minimize the amount of actual instantiations we perform in `ResolveCore`
    1. We can postpone instantiating `Target`s and `Command`s until resolution is complete, removing a whole bunch of messy `mainargs` logic for the already-messy recursive resolution code. A lot of it ends up in `Resolve.scala`, which is somewhat more straightforward and better able to absorb the messiness
    2. We only need to instantiate `Module`s when trying to resolve tasks within a `Cross` module, since that requires us to instantiate the `Cross` object to see it's cross keys. Most resolution can take place purely by reflecting on `java.lang.Class[_]`s without needing the modules to actually be instantiated
    3. `ResolveCore` now only returns the `segments: Segments` for the things it resolves, leaving instantiation to `Resolve.Tasks`
    4. This significantly increases the laziness of `ResolveCore`, letting `mill resolve` work faster and work even when the modules cannot instantiate, unless there is a `Cross` module we need to instantiate

4. Refactored `ResolveNonEmpty` to `ResolveNotFoundHandler`, to only be called by `Resolve` if `ResolveCore` returns a `ResolveCore.NotFound`. That helps avoid deep call hierarchies, and allows us to ensure that `ResolveCore` and `ResolveNotFoundHandler` independent of each other (rather than one wrapping the other)

5. Extracted `mill.define.Module.Internal` into a separate `mill.define.Reflect` object. This helps clean up `Module.scala`. Ideally I wanted to move `Reflect` to `mill.resolve`, but it's difficult to do that while still providing nice helper methods on `mill.define.Module` objects e.g. `millModuleDirectChildren` while allowing them to be overriden etc.

6. `ResolveTasks` is now `Resolve.Tasks`, `ResolveSegments` is now `Resolve.Segments`, `ResolveMetadata` is removed since `Resolve.Segments` does more or less the same thing anyway

7. Make `Cross#items` a `List[Item]`, so we can publicize it while continuing to evolve it in a binary-compatible manner

# Testing

1. Added more asserts to the existing `ResolveTests.scala` unit tests to check that `ResolveSegments` can still succeed even when `ResolveTasks` fails due to module initialization errors

2. Added additional test cases to check the behavior of cross-module failures in more scenarios: when `_` and `__` is used, when the cross module itself fails to initialize (previously we only checked for children failing to initialize), and when the cross module's parent module fails to initialize

3. Added a few test cases to `integration/failure/module-init-error/` exercise the "resolve works even when module cannot initialize" behavior end-to-end

4. Existing test cases in `ResolveTests.scala` and downstream tests that exercise task resolution all pass


lefou (approved):
This looks good to me. I think, we should provide some `@deprecated` forwarders for the classes moved to `mill.resolve` in the `mill.main` package object. This should improve migration convenience and cross compilation.

lefou:
@lihaoyi We have one failing test in main and it seems related to this PR. Could you please look at it?

```
X mill.resolve.ResolveTests.overridenModule.0 5ms 
  utest.AssertionError: resolvedTasks.map(_.map(_.toString).toSet[String]) ==
  resolvedTasks: Either[String,List[mill.define.NamedTask[Any]]] = Left(Cannot resolve sub.inner.subTarget. Try `mill resolve sub.inner._` to see what's available.)
  expected: scala.util.Either[String,scala.collection.immutable.Set[mill.define.NamedTask[Any]]] = Right(Set(sub.BaseInnerModule.subTarget))
    mill.resolve.ResolveTests$Checker.checkSeq(ResolveTests.scala:24)
    mill.resolve.ResolveTests$Checker.apply(ResolveTests.scala:14)
    mill.resolve.ResolveTests$.$anonfun$tests$257(ResolveTests.scala:754)
```

lihaoyi:
will take a look

lihaoyi:
CI is green now, but I do not understand what caused the breakage, nor whay fixed it, so we'll have to see if the issue reappears. If it turns out the test is just flaky then we may be forced to disable it

lihaoyi:
CI is green now, but I do not understand what caused the breakage, nor whay fixed it, so we'll have to see if the issue reappears. If it turns out the test is just flaky then we may be forced to disable it

PR #2512, opened 2023-05-09
Try to fix windows CI

lihaoyi:


lihaoyi:
CI is green, but it was green on the original PR before merging too. Going to try merging this to see if the problem goes away

PR #2513, opened 2023-05-10
Kill mill.modules package

lihaoyi:
Ticks off https://github.com/com-lihaoyi/mill/discussions/1826#discussioncomment-5358469

`mill.modules` is a very confusing name that I came up with very early in Mill's development. It misleadingly has nothing to do with `mill.Module`, and is just a grab bag of helpers. This PR moves them to the JVM packages which are most related to their functionality

We basically move the following files:

1. `mill.modules.Assembly` -> `mill.scalalib.Assembly` 
2. `mill.modules.Util` -> `mill.util.Util`
3. `mill.modules.Jvm` -> `mill.util.Jvm`
4. `mill.modules.CoursierSupport` -> `mill.util.CoursierSupport`

There is probably further cleanup and re-organization we can do here. Currently, `Util` and `Jvm` are pretty tightly coupled together, with `VisualizeModule` living in `main` stopping us from moving them all to `scalalib`. An alternative is to move `Assembly` to `mill.util`, so at least `createJar` and `createAssembly` can live in the same package rather than being split across `mill.{util,scalalib}`.

But for now, just breaking up `mill.modules` should already be a step forward, and further cleanups can come in follow up PRs


lefou (commented):
Looks good to me, but we need some `@deprecated` type aliases and forwarder objects to keep source compatibility for a while. Migration and cross compilation is otherwise a pita and especially, when not the whole ecosystem is migrated, it's less attractive to try something out if there is no easy going back.

lihaoyi:
Added backwards compat forwarders. Needed to put them in `scalalib/` even though they're in `mill.main`, due to `Assembly` being moved there

scalalib/src/mill/main/package.scala
```diff
@@ -0,0 +1,15 @@
...
+  @deprecated("use mill.util.Jvm")
+  val Jvm = mill.util.Jvm
+
+  @deprecated("use mill.util.Util")
+  val Util = mill.util.Util
+
+  @deprecated("use mill.util.CoursierSupport")
+  val CoursierSupport = mill.util.CoursierSupport
+
+  @deprecated("use mill.scalalib.Assembly")
```
lefou:
```suggestion
  @deprecated("Moved to mill.util.Jvm", "Mill 0.11.0-M9")
  val Jvm = mill.util.Jvm

  @deprecated("Moved to mill.util.Util", "Mill 0.11.0-M9")
  val Util = mill.util.Util

  @deprecated("Moved to mill.util.CoursierSupport", "Mill 0.11.0-M9")
  type CoursierSupport = mill.util.CoursierSupport
  @deprecated("Moved to mill.util.CoursierSupport", "Mill 0.11.0-M9")
  val CoursierSupport = mill.util.CoursierSupport

  @deprecated("Moved to mill.scalalib.Assembly", "Mill 0.11.0-M9")
```

lefou (approved):

PR #2515, opened 2023-05-12
Update zinc to 1.8.1

lefou:
Includes a fix for infinite compilation issues and a security fix. We should also consider a backport to Mill 0.10.


lefou:
@lihaoyi You you help out and explain what's missing? My guess is, that we need to bump our precompiled compiler brigdes due to the version bump of the compiler. What are the required steps?

lihaoyi:
@lefou I think you just need to update this line to use the current Zinc version, and it should work

https://github.com/com-lihaoyi/mill/blob/156c7a4ce0397336065beff918fe7c66b24e7e14/scalalib/test/src/mill/scalalib/HelloWorldTests.scala#L545

lefou:
@lihaoyi Thank you! I changed the tests to not use hard-coded zinc versions.

PR #2517, opened 2023-05-12
add play 2.9 support

PeuTit:
Early support for PLay 2.9 using play `2.9.0-M4` with Scala `2.13` & `3`. 
We are also using ScalaTestPlus `6.0.0-M3`.


lefou:
Thank you for this PR. Looks like Play 9 does neither support Java 8 nor Scala 3.2. Is that correct? You need to update the test setup.

PeuTit:
Thanks for your response. I've updated the Scala version used for the test to `3.3.0-RC6` but I'm unsure on were to update the Java version used. Any insight would be appreciated.

lefou:
Superseded by #2772

PR #2519, opened 2023-05-15
Enable warnings in build.sc, overhaul `Agg` to avoid triggering spurious deprecation warnings

lihaoyi:
Seems like all the `IterableOnce` method implementations were deprecated at some point, and you need to pull in `IterableOps` to have non-deprecated versions

If we want to start making use of `@deprecated` in 0.11.0, we need to make sure people see the warnings, and that they're not drowned in a sea of noise


lefou (approved):
Looks good to me.

Issue #2198, opened 2022-12-15
`inspect` should render a paraemetr usage help for commands


lefou:
The `inspect` command already prints the scaladoc. Whan commands have parameters documented with `@param`, we already see them too, but often parameters are not documented. Also, if we use dedicated classes that used the delegation support on `mainargs`, we don't see the documentation.

As Mill is automatically supporting parameters with `mainargs` and `mainargs` can render nice usage displays, we should include such usage output to the output of `inspect`.

lefou:
* Related to #503

PR #2522, opened 2023-05-16
Print out CLI help message when inspecting commands

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/2198

There's a bit of duplication if a user both adds `@mainargs.arg(doc = "...")` scala annotation as well as a `@param foo` scaladoc annotation, but that's an inherent issue with mainargs and not something we aim to fix here.

We render the mainargs CLI help message the same way it's done in mainargs. There isn't a nice helper that renders the whole thing at once, so i have to piece it together from various helpers, but it's not too complex and works out ok.

Tested via additions to `integration.feature[docannotations].local.test`, and manually

```bash
$ ./mill -i dev.run example/basic/1-simple-scala -i inspect run # command with 1 arg
run(JavaModule.scala:720)
    Runs this modules code in a subprocess and waits for it to finish

    args <str>...

Inputs:
    finalMainClass
    runClasspath
    forkArgs
    forkEnv
    forkWorkingDir
    runUseArgsFile

$ ./mill -i dev.run example/basic/1-simple-scala -i inspect compile # not a command
compile(ScalaModule.scala:211)
    Compiles the current module to generate compiled classfiles/bytecode.

    When you override this, you probably also want to override [[bspCompileClassesPath]].

Inputs:
    scalaVersion
    upstreamCompileOutput
    allSourceFiles
    compileClasspath
    javacOptions
    scalaOrganization
    allScalacOptions
    scalaCompilerClasspath
    scalacPluginClasspath
    zincReportCachedProblems

$ ./mill -i dev.run example/basic/1-simple-scala -i inspect console # command with no args
console(ScalaModule.scala:369)
    Opens up a Scala console with your module and all dependencies present,
    for you to test and operate your code interactively.

Inputs:
    scalaVersion
    runClasspath
    scalaCompilerClasspath
    forkArgs
    forkEnv
    forkWorkingDir

$ ./mill -i dev.run example/basic/1-simple-scala -i inspect ivyDepsTree # command with lots of args
ivyDepsTree(JavaModule.scala:688)
    Command to print the transitive dependency tree to STDOUT.

    --inverse              Invert the tree representation, so that the root is on the bottom val
                           inverse (will be forced when used with whatDependsOn)
    --whatDependsOn <str>  Possible list of modules (org:artifact) to target in the tree in order to
                           see where a dependency stems from.
    --withCompile          Include the compile-time only dependencies (`compileIvyDeps`, provided
                           scope) into the tree.
    --withRuntime          Include the runtime dependencies (`runIvyDeps`, runtime scope) into the
                           tree.

Inputs:
    transitiveIvyDeps
    scalaVersion
    scalaVersion
    scalaOrganization
    scalaVersion
```

I also removed the `numOverrides` field on `mill.define.Discover`, since it's unused since https://github.com/com-lihaoyi/mill/pull/1600 landed a while back

### DIFF
--- a/integration/feature/docannotations/test/src/DocAnnotationsTests.scala
+++ b/integration/feature/docannotations/test/src/DocAnnotationsTests.scala
@@ -3,28 +3,103 @@ package mill.integration
 import utest._
 
 object DocAnnotationsTests extends IntegrationTestSuite {
+  def globMatches(glob: String, input: String) = {
+    StringContext.glob(glob.stripMargin.split("\\.\\.\\."), input).isDefined
+  }
   val tests = Tests {
     initWorkspace()
     test("test") - {
       val res = eval("inspect", "core.test.ivyDeps")
       assert(res == true)
+
       val inheritedIvyDeps = ujson.read(meta("inspect"))("value").str
       assert(
-        inheritedIvyDeps.contains("core.test.ivyDeps"),
-        inheritedIvyDeps.contains("Overriden ivyDeps Docs!!!"),
-        inheritedIvyDeps.contains("Any ivy dependencies you want to add to this Module")
+        globMatches(
+          """core.test.ivyDeps(build.sc:...)
+            |    Overriden ivyDeps Docs!!!
+            |
+            |    Any ivy dependencies you want to add to this Module, in the format
+            |    ivy"org::name:version" for Scala dependencies or ivy"org:name:version"
+            |    for Java dependencies
+            |
+            |Inputs:
+            |""".stripMargin,
+          inheritedIvyDeps
+        )
       )
 
-      assert(eval("inspect", "core.task"))
-      val task = ujson.read(meta("inspect"))("value").str
+      assert(eval("inspect", "core.target"))
+      val target = ujson.read(meta("inspect"))("value").str
+      pprint.log(target)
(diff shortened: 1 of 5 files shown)


integration/feature/docannotations/test/src/DocAnnotationsTests.scala
```diff
@@ -3,28 +3,102 @@ package mill.integration
...
       val res = eval("inspect", "core.test.ivyDeps")
       assert(res == true)
+
       val inheritedIvyDeps = ujson.read(meta("inspect"))("value").str
       assert(
-        inheritedIvyDeps.contains("core.test.ivyDeps"),
-        inheritedIvyDeps.contains("Overriden ivyDeps Docs!!!"),
-        inheritedIvyDeps.contains("Any ivy dependencies you want to add to this Module")
+        globMatches(
+          """core.test.ivyDeps(build.sc:...)
```
lihaoyi, line 18: These tests are a bit fragile, but it's probably worth having at some _some_ test where we assert on the overall shape of the inspect output. These docs and task inputs don't change very often anyway, so I expect the maintenance burden of keeping the tests up to date should be tolerable

lefou (approved):

Issue #2461, opened 2023-04-25
Make meta builds (`mill-build/build.sc`) opt-in and configurable


lefou:
As a follow up on PR #2377, we need to come up with some way to configure the `mill-build/build.sc` aka "_meta build_".

Meta builds need to be disabled, unless the `build.sc` explicitly enables it.

Meta build location should not be hardcoded but configured when enabled.

```scala
// Enable meta-build and specify the location
import $meta.`mill-build`
```

lefou:
@lihaoyi wrote a [comment to issue #2516](https://github.com/com-lihaoyi/mill/issues/2516#issuecomment-1545910923), which belongs to this discussion:

> Previously we had discussed having the whole `mill-build` folder be under `import $meta.mill-build`, where a user could choose what the folder name was. Elsewhere we had also discussed being able to rename `build.sc` via a flag, e.g. `--build-file-name mill-build` to avoid conflicts.
> 
> What if we combined the two ideas, such that a single flag `--build-file-name` controls both the meta-build path as well as the meta-build folder? e.g. it would default to `build.sc`/`build/`, but a user who has a conflict in `build/` could do `--build-file-name mill-build` to rename both together to `mill-build.sc`/`mill-build/`. In this way, the `build-file-name.sc` file is the source of truth while the `build-file-name/` folder is basically the associated extension point if the `.sc` format becomes too limiting.
> 
> This would help ensure that the names of both file and folder remain in-sync, removing a degree of freedom a `import $meta.foo` provides that is unnecessary, while remaining explicitly customizable and ensuring the relationship between the two is clear to the user. Conceptually it's like having a customizable `Shared Primary Key` relationship v.s. having an `Foreign Key` column in a database: the schema enforces the invariants you want by-construction, rather than starting with an unnecessarily flexible schema (`import $meta.foo` with a different folder name, multiple `import $meta.foo`s...) and having to lock it down after-the-face via assertions.

I think we should not bind the build file name to the directory name of it's potential meta-build. These are two complete different things.

My most important mantra is: __A `build.sc` is completely self-sufficient.__ It's the single source of truth. (I envision to even move all external settings like the `.mill-version` into some magic-import or using directives.)

By adding a `import $meta` we can opt-in a meta-build.

A meta-build is not necessarily tied to a single `build.sc`. E.g. multiple project can select the same meta-build via `import $meta.^.company-defaults`.

[...]

PR #2527, opened 2023-05-18
Make mill-build folder opt in via `import $mill.build`

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/2461

For now it remains hard-coded as `mill-build/`, for simplicity. Given the `mill-` prefix, it's unlikely to collide with user-defined folders. Making it configurable would open another can of worms where the `import $mill.foo` can now collide with an `object foo`, which is very unintuitive.

For now, we just parse the script files one additional time in `MillBuildBootstrap.scala`. This is a bit wasteful, but is probably fast enough for now, and we're already pretty sloppy parsing everything twice in `MillBuildRootModule#scriptSources` and `MillBuildRootModule#parseBuildFiles`, so parsing things three times isn't the end of the world. We can look into optimizing it in future if necessary

### DIFF
--- a/example/misc/4-mill-build-folder/build.sc
+++ b/example/misc/4-mill-build-folder/build.sc
@@ -21,7 +21,7 @@ object foo extends RootModule with ScalaModule {
 // file and it's `import $file` and `$ivy` are a shorthand syntax for defining
 // a Mill `ScalaModule`, with sources and `ivyDeps` and so on, which is
 // compiled and executed to perform your build. This module lives in
-// `mill-build/`.
+// `mill-build/`, and can be enabled via the `import $mill.build` statement above.
 
 /** See Also: mill-build/build.sc */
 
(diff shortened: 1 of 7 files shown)


lihaoyi:
Seems the integration tests need to be tweaked abit since we're no longer unconditionally going through all levels of meta-builds.

lefou (commented):
I think the `$mill.build` is a bit unfortunate. Although the `$mill.build` somehow corresponds to `mill-build`, once we want to support a custom name, it will no longer match as nice. Just because it already contains two segments.

I like the originally proposed `$meta` better, as it can support a custom name when given, e.g. 

```scala
import $meta.`my-mill-build`
```

but also use the default name when used without any extra segment like

```scala
import $meta // short form of import $meta.`mill-build`.
```

lefou (approved):
Looks good to me!

PR #2528, opened 2023-05-19
Swap over to mill-build.com domain name

lihaoyi:


lihaoyi:
Currently the page is down due to redirect loops, once the docsite is rebuilt on this commit it should start working again

PR #2534, opened 2023-05-20
Publicize PlatformScalaModule#platformScalaSuffix

lihaoyi:


lefou (commented):
Explicit return types would be nice.

PR #2537, opened 2023-05-20
Move `mill.BuildInfo` to`mill.main.BuildInfo`

lihaoyi:
We currently encourage `import mill._`, which causes problems when people want to also `import mill.contrib.buildinfo.BuildInfo`. There's a workaround to `import mill.{BuildInfo => MillBuildInfo, _}`, which we do in several places in our own codebase, but it's a pain in the neck and an unnecessary hurdle for users to cross. The majority of users would not need to access `mill.BuildInfo` at all, so putting it in the default namespace just adds clutter.

The `mill.BuildInfo` is used throughout `mill.scalalib` and `mill.scalajslib` and so on, but is not used in `main/`, so I put it in as downstream a module as possible which turns out to be `mill.main.BuildInfo`

### DIFF
--- a/bsp/src/mill/bsp/BSP.scala
+++ b/bsp/src/mill/bsp/BSP.scala
@@ -1,8 +1,9 @@
 package mill.bsp
 
 import mill.api.{Ctx, PathRef}
-import mill.{Agg, T, BuildInfo => MillBuildInfo}
+import mill.{Agg, T}
 import mill.define.{Command, Discover, ExternalModule}
+import mill.main.BuildInfo
 import mill.eval.Evaluator
 import mill.util.Util.millProjectModule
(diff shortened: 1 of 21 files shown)


lefou (commented):
Change looks ok-ish. I think having these constant defined earlier in the dependency chain is a plus, but I also see the actual collision. But now, that we identified the root issue, a plugin and it's product (a generated class) sharing the same name, we should try to fix it there. Or at least try to. Could we rename the plugin to e.g. `GenBuildInfo` if we want stick with the generated `BuildInfo` name?

lihaoyi:
Sounds good. I'll merge this first, and open a follow up to possibly rename `BuildInfo`

Issue #2533, opened 2023-05-20
`runBackground` broken in Mill `0.11.0-M9`


lolgab:
The latest Mill snapshot 0.11.0-M9 converted `mill-scalalib-backgroundwrapper` from a `JavaModule` to a `ScalaModule`.

Running `.runBackground` I get:
```
mill.scalalib.ZincWorkerModule.backgroundWrapperClasspath 
1 targets failed
mill.scalalib.ZincWorkerModule.backgroundWrapperClasspath 
Resolution failed for 1 modules:
--------------------------------------------
  com.lihaoyi:mill-scalalib-backgroundwrapper:0.11.0-M9 
        not found: /Users/lorenzo/.ivy2/local/com.lihaoyi/mill-scalalib-backgroundwrapper/0.11.0-M9/ivys/ivy.xml
        not found: https://repo1.maven.org/maven2/com/lihaoyi/mill-scalalib-backgroundwrapper/0.11.0-M9/mill-scalalib-backgroundwrapper-0.11.0-M9.pom
```
Since it can't find anymore the java artifact without the `_2.13` suffix.

lihaoyi:
Should be caused by https://github.com/com-lihaoyi/mill/pull/2476

PR #2538, opened 2023-05-21
Make `mill.backgroundwrapper` a `JavaModule` again

lihaoyi:
Should fix https://github.com/com-lihaoyi/mill/issues/2533, which was broken by https://github.com/com-lihaoyi/mill/pull/2476

Just requires shuffling `def testDeps` and related definitions around to make them available outside `MillScalaModule`. They moved to `MillCoursierModule`, which in order to satisfy their dependencies had to be converted to `MillJavaModule`. The overall logic of the build is unchanged

Tighted up the test artifact resolution logic for `.local` example/integration tests. Previously, they would ignore the `artifactSuffix`, which is why tests exercising `runBackground` like `example.web[1-todo-webapp].local` passed despite the bug. With the more strict test artifact resolution, this bug gets surfaced properly on `main`, and is fixed by this PR

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -284,7 +284,14 @@ trait MillPublishModule extends PublishModule {
   override def javacOptions = Seq("-source", "1.8", "-target", "1.8", "-encoding", "UTF-8")
 }
 
-trait MillCoursierModule extends CoursierModule {
+trait MillJavaModule extends JavaModule {
+
+  // Test setup
+//  def artifactName: T[String]
+  def testDep = T { (s"com.lihaoyi-${artifactName()}", testDepPaths().map(_.path).mkString("\n")) }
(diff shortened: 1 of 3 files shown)


lolgab (approved):

build.sc
```diff
@@ -284,7 +284,14 @@ trait MillPublishModule extends PublishModule {
   override def javacOptions = Seq("-source", "1.8", "-target", "1.8", "-encoding", "UTF-8")
 }
 
-trait MillCoursierModule extends CoursierModule {
+trait MillJavaModule extends JavaModule {
+
+  // Test setup
+//  def artifactName: T[String]
```
lefou, line 290: This commented code can be removed

PR #2543, opened 2023-05-24
Re-enable Acyclic enforcement and fix existing violations

lihaoyi:
Turning on enforcement meant breaking up a number of the existing dependency cycles in the codebase:

1. We had to break up `ModuleUtils` to have a separate `JavaModuleUtils`, with dependencies `JavaModuleUtils` -> `JavaModule` -> ModuleUtils`

2. We have to avoid using `mill.*` shorthand inside `mill.main`, so all the `mill.T.*` had to be replaced by `mill.define.Target.*`

3. `BoundDep.scala` was merged with `Dep.scala`, since neither files are very large, and they seem closely related enough that it didn't seem trying to break them apart properly.

4. Broke up the circular dependency between `JavaModule <-> SemanticDbJavaModule <-> ScalaModule`, so now it's `ScalaModule` -> `JavaModule` -> `SemanticDbJavaModule`, with `SemanticDbJavaModule` declaring abstract targets it expects `JavaModule` to fulfil and also exposing some of its targets as `protected` for `ScalaModule` to override

5. Broke up the circular dependency between `JavaModule <-> TestModule <-> ScalaModule`, which is now `ScalaModule -> JavaModule -> TestModule`. I do this by having `TestModule` depend on `TestModule.JavaModuleBase` and `TestModule.Specs2` depend on `TestModule.ScalaModuleBase` traits that only include the targets we need to override, which `JavaModule` and `ScalaModule` then inherit. This avoids `TestModule` needing to depend on the bulk of `JavaModule` or `ScalaModule`

Overall, enforcing acyclicity forces you to depend on explicitly declared narrow interfaces rather than implicitly depended on fat implementations, properly inject dependencies via parameters/abstract-members, and ensure "code that depends on everything" is split up from "code that everything depends on". Overall, this is a great way for enforcing "high level" code quality, which is very different from the "low level" code quality that autoformatters and other linters help with


scalalib/src/mill/scalalib/TestModule.scala
```diff
@@ -1,13 +1,22 @@
...
-import mill.define.{Command, Task, TaskModule}
-import mill.api.{Ctx, Result}
+import mill.define.{Command, ModuleRef, Task, TaskModule}
+import mill.api.{Ctx, PathRef, Result}
 import mill.util.Jvm
 import mill.scalalib.bsp.{BspBuildTarget, BspModule}
 import mill.testrunner.TestRunner
 
-trait TestModule extends JavaModule with TaskModule {
+trait TestModule extends TaskModule with TestModule.JavaModuleBase {
```
lefou, line 10: Why was this necessary? Now, that `TestModule` is no lomger a complete `JavaModule`, my `mill-jacoco` plugin broke.

lefou: Turns out just mix-in `JavaModule` to my `JacacoTestModule` fixed the compilation in my case.

Issue #2545, opened 2023-05-24
`UnsupportedOperationException` at `Server.handleRun`


SpriteOvO:
I'm using `mill` v0.10.12, and I encountered this error several days.

```
java.lang.UnsupportedOperationException
        at java.base/java.lang.Thread.stop(Thread.java:1705)
        at mill.main.Server.handleRun(MillServerMain.scala:209)
        at mill.main.Server.$anonfun$run$2(MillServerMain.scala:117)
        at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
        at mill.main.Server$.lockBlock(MillServerMain.scala:227)
        at mill.main.Server.$anonfun$run$1(MillServerMain.scala:98)
        at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
        at mill.main.Server$.tryLockBlock(MillServerMain.scala:235)
        at mill.main.Server.run(MillServerMain.scala:95)
        at mill.main.MillServerMain$.main(MillServerMain.scala:53)
        at mill.main.MillServerMain.main(MillServerMain.scala)
```

The error is always thrown after a build is completed, the build was successful. Passing `--no-server` to `mill` will suppress the error, but obviously the server will be disabled.

It seems to be thrown from

https://github.com/com-lihaoyi/mill/blob/cfbafb806351c3e1664de4e2001d3d1ddda045da/main/src/mill/main/MillServerMain.scala#L209

Any idea or workaround would be appreciated.

PR #2546, opened 2023-05-24
Support Java 20 (Backport of #2501)

lefou:
Ignore the `UnsupportedOperationException` of `Thread.stop`. This is a
quick fix to de-block users from using Java 20.

* Original Pull request: https://github.com/com-lihaoyi/mill/pull/2501

* Fix: https://github.com/com-lihaoyi/mill/issues/2545


lefou:
Test suite won't succeed on Java 20, but the backported fix seems to work.

PR #2549, opened 2023-05-25
Better "Get started" for single module examples

KarinaTyulebaeva:
Hi! I faced the problem while starting using the mill with basic examples. 
When you build `basic/1-simple-scala` the output shows an error:
 
```bash
1-simple-scala $ mill foo.compile
1-simple-scala $ mill foo.run

1 targets failed
foo.finalMainClass No main class specified or found
```

This happens because `basic/1-simple-scala/build.sc` looks like:

```scala
object foo extends RootModule with ScalaModule {
  def scalaVersion = "2.13.8"
  def ivyDeps = Agg(
    ivy"com.lihaoyi::scalatags:0.8.2",
    ivy"com.lihaoyi::mainargs:0.4.0"
  )

  object test extends Tests {
    def ivyDeps = Agg(ivy"com.lihaoyi::utest:0.7.11")
  }
}
```

The root module is called `foo` but there is no such directory. It occurs in almost all examples with single module. So, this refactoring should fix the problem and make the first experience better


KarinaTyulebaeva:
Got it, thanks! Btw, running tests doesn't find `RootModule`. I suppose it happens cause the mill version which is used for tests launching is different from current repo version. Is it possible to set it to the current version so that it is convenient to debug? Maybe you have contribution guide?

lefou:
The examples are meant to be used with the Mill version represented by the sources in the repository. We're currently stabilizing the `main` branch and a new Mill 0.11 release is close ahead. 

We have some contributor information in the `readme.adoc`. If you have concrete questions, please open a question in the discussions or ask in our chat.

Issue #408, opened 2018-08-13
Configurable server timeout


seieewkx:
`mill` is the **fastest** Scala compiler I find. I have a special requirement for Scala compilation.
The code need to be compiled is very small, around 100 line of code. I need it compile **as fast as possible**.
If the `mill` server(daemon) is alive, The compilation needs 2.x seconds. And after 5mins, the server(daemon) dead, it needs 16 seconds. 
I want the compilation always be **as fast as possible**. By keeping the server(daemon) alive. Thanks!
 
https://github.com/lihaoyi/mill/blob/22e19245e05a43b645c8a65cc9b1266df3a85579/main/src/mill/main/MillServerMain.scala#L43

lihaoyi:
should be easy, send a PR

lolgab:
This should be fixed by #2550.
You can configure the timeout by setting the `MILL_SERVER_TIMEOUT_MILLIS` environment variable.
Can we close this issue @lefou?

lefou:
@lolgab, yeah. Can be closed. We should add it to the documentation though.

PR #2550, opened 2023-05-25
issue-408 - server timeout setting configurable through envs

Ivoyaa:
Trying to implement [issue-408](https://github.com/com-lihaoyi/mill/issues/408). 
Not exactly sure if it is a correct way to achieve it, so would appreciate any feedback!


lefou:

PR #2554, opened 2023-05-27
Deprecate SelectMode.Single and backport evalOrThrow API

lolgab:


lefou (commented):
There is some value in those deprecations, but not all users will recognize them, as Mill up to 0.10.12 is not showing any deprecations for the `build.sc` out of the box. Maybe we can/should change that for 0.10.13. 

Also, I would leave Mill 0.11 out of the descriptions. It doesn't add any clue.

lolgab:
> Also, I would leave Mill 0.11 out of the descriptions. It doesn't add any clue.

The idea was that this is not some kind of deprecation that you can leave there forever, but you need to fix now if you want to migrate.

lefou:
> > Also, I would leave Mill 0.11 out of the descriptions. It doesn't add any clue.
> 
> The idea was that this is not some kind of deprecation that you can leave there forever, but you need to fix now if you want to migrate.

I think we more or less removed all deprecated API in 0.11.

lolgab:
Superseded by https://github.com/com-lihaoyi/mill/pull/2570

PR #2556, opened 2023-05-27
Tighten up compatibility promises and make them explicit

lihaoyi:
1. List out the exact modules we want to provide bincompat guarantees for
2. Privatize a lot more stuff


main/src/mill/main/TokenReaders.scala
```diff
@@ -23,19 +23,20 @@ object Tasks {
   }
 }
 
-class EvaluatorTokenReader[T]() extends mainargs.TokensReader.Constant[mill.eval.Evaluator] {
```
lefou, line 26: Those token readers where needed to be accessible in the past, e.g. in some plugins that provided external modules with commands.

lihaoyi: But do they need to be accessible directly? I'd expect people to use them via `import mill.main.TokenReaders._`, and use them abstracted behind a `mainargs.TokensReader[Evaluator]` interface, rather than `new mill.main.EvaluatorTokenReader` and using the concrete type directly

lefou: Yeah, that might work. I overlooked that the implicit `def`s are accessible.

PR #2560, opened 2023-05-30
Distribute Mill Assembly via Maven Central 

lihaoyi:
Creates a stub `dist` module whose only purpose is to take the `dev.assembly` and put it on maven central, which the updated bootstrap script will download.

This should help reduce the number of external failure points in Mill bootstrap script; rather than having hard dependency on both Github and Maven Central, this PR reduces it to only a hard dependency on Maven Central.  Maven Central is also by default a write-only package store. That means that short of sending them a lawyer letter, the packages published there are immutable (https://central.sonatype.org/faq/can-i-change-a-component). This is in contrast to Github release assets which are read/write. 

With Github, a bug in our CI auto-upload system could easily override/delete/corrupt already-published assemblies, causing a widespread outage to anyone depending on the bootstrap script (e.g. in CI) until someone fixes it. It is also vulnerable to Github outages, which are not unheard of. Distribution via Maven Central removes both Github outages and over-writing-existing-artifacts as possible failure modes. While Sonatype can and does have outages, that would already cause a failure in Mill's bootstrapping, so it would be no worse than it already is today

The validity of the various URLs and bootstrap scripts etc. have been tested manually, but end-to-end testing will need to be done post-merge

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -1622,18 +1622,18 @@ def millBootstrap = T.sources(T.workspace / "mill")
 
 def bootstrapLauncher = T {
   val outputPath = T.ctx.dest / "mill"
-  val millBootstrapGrepPrefix = "DEFAULT_MILL_VERSION="
-  val millDownloadUrlPrefix = "MILL_DOWNLOAD_URL="
+  val millBootstrapGrepPrefix = "(\n *DEFAULT_MILL_VERSION=)"
+  val millDownloadUrlPrefix = "(\n *MILL_DOWNLOAD_URL=)"
   os.write(
     outputPath,
     os.read(millBootstrap().head.path)
       .replaceAll(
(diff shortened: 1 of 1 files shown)


build.sc
```diff
@@ -1254,7 +1259,7 @@ object dev extends MillPublishScalaModule {
   )
 
   lazy val allPublishModules = build.millInternal.modules.collect {
-    case m: PublishModule if m ne this => m
+    case m: PublishModule if (m ne this) && (m ne dist) => m
```
lefou, line 1262: Can we change the name of that target or add some comment to clarify it's purpose?

lefou (commented):
Looks good to me. One nitpick below.

lihaoyi:
Tested this via the unstable version `0.11.0-M10-9-db6d7d`, seems to work

Issue #716, opened 2019-10-16
Tests fail due to class loading issue - GraalVM cannot be used.


hmf:
I have forked the a [Maven project](https://github.com/hmf/graal-js-jdk11-maven-demo) that demonstrates the use of the GraalVM. It also includes a set of tests that confirm that the GraalVM JS engine is faster than the Nashorn engine. These tests work correctly. 

I set-up the same project using Mill. The example executes correctly - Nashorn is nearly twice as slow as Graal.js. However the tests fail due to class loading. In order for this to work we need to pass parameters to the JVM and add the libraries to the classpath. Due to issue #690 I created both a [JUnit version](https://github.com/hmf/graal-js-jdk11-maven-demo/tree/mill_maven) and a [uTest version](https://github.com/hmf/graal-js-jdk11-maven-demo/tree/mill_utest). The former uses the same directory structure as the original project (intended for contribution), the latter uses the Mill standard directory structure.   

I tried to track the issue that leads me to the [test call](https://github.com/lihaoyi/mill/blob/ed8ea584c67f1ab83c488441779356a36d8fa3bb/scalalib/src/JavaModule.scala#L570). which then executes a [subprocess](https://github.com/lihaoyi/mill/blob/ca1def4d6f5c829701e04f86dacb720bb57f036f/main/src/modules/Jvm.scala#L54) call that ends in the call to [TestRunner](https://github.com/lihaoyi/mill/blob/9ba4cb69331386dfde9bac69dc2d5b22401face3/scalalib/src/TestRunner.scala). I see that this last execution takes the classpath passed to it as an argument and sets it according. However when I print the classpath (see [this example](https://github.com/hmf/graal-js-jdk11-maven-demo/blob/mill_maven/src/main/java/com/mycompany/app/App.java#L180)), it does not list the required libraries. Nevertheless I have confirmed that the classpath argument is passed on correctly.

Because the problem manifests itself in both test frameworks I suspect the problem is in the `TestRunner` and not for example in the [JUnitRunner](https://github.com/sbt/junit-interface/blob/20183572cb15cc302199b48824bd1812a5d4941e/src/main/java/com/novocode/junit/JUnitRunner.java). 

I have also seen issues #302 and #370 that seem to be related. These class loading issues seem to be tricky and at this point I do not know what other steps can be taken to diagnose and solve the problem. 

Any suggestions are welcome.

PR #2561, opened 2023-05-31
Update thirdparty tests, make subprocess testrunner run user code in root classloader

lihaoyi:
1. Move `thirdparty` from `integration/` to `example/`, and simplify them down to `acyclic`, `fansi`, and `jimfs`. We can add more complex builds later if we want, but for now just having some realistic builds that are not bitrotted and old is already an improvement over what we have now.

2. Flip the sub-process `TestRunner` classloaders on its head: rather than runner `TestRunner.scala` in the boot classloader and user code in a sub-classloader, we run user code in the boot classloader and `TestRunner.scala` in a sub classloader. This is necessary to make JimFS work, and would probably fix https://github.com/com-lihaoyi/mill/issues/716 and related issues. 

3. Broke up and cleaned up `TestRunner.scala`:
    1. Separated `TestRunner` and `TestRunnerMain0`, the two entrypoints, from `TestRunnerUtils`
    2. Replaced all the complicated logic for converting logic to/from CLI flags with JSON written to a file by `TestModule` and read from the file by `TestRunner`
    3. Replaced `ClosableIterator` with `geny.Generator`, which also provides guaranteed closing after iteration

4. Marked most of `testrunner` as `@internal`. Most of these could be `private[mill]`, but I think we don't have a clear idea of how this stuff is meant to be used yet, and it's plausible that other people may want to interact with the testrunner in their own custom test suites. I think `@internal` is better suited for now until we either (a) know better what a stable API looks like or (b) decide that this stuff should not be used externally.

We have much more control over `TestRunner.scala` than we do over user code, so we can make sure that `TestRunner.scala` works when running in a semi-isolated classloader, whereas for user code we can offer no such guarantees. We can't do anything for `.testLocal` that runs in-process, but at least for `.test` we can try to offer them a clean classloader environment.

The classloader for user code is not 100% clean: we still have a single class `mill.testrunner.entrypoint.TestRunnerMain` added within it. That's probably OK, since no matter what we do w.r.t. shading and stuff we still need somewhere to put our own main method to bootstrap the `mill.testrunner` classloader.

The testrunner changes are covered by existing unit and integration/example tests, and the new "user code runs in clean classloader" logic is covered by `example.thirdparty[jimfs].local`.


testrunner/src/mill/testrunner/TestRunner.scala
```diff
@@ -140,28 +141,6 @@ object TestRunner {
...
-      classpath,
-      arguments,
-      sysProps,
-      outputPath,
-      colored,
-      testCp,
-      homeStr,
-      globSelectors
-    )
-
```
lefou, line 164: We made this one private and some other tricks to keep the option to evolve this class in a binary compatible way. What's the intention of changing that?

lihaoyi: This is still WIP, but I want to try to adjust the code path so we don't need to use a `case class` here. That would make maintaining binary compatibility much simpler

lihaoyi:
Should be ready for review I think

PR #2562, opened 2023-06-01
Enable `buildInfoStaticCompiled` on ScalaJSModule and ScalaNativeModule

lolgab:
Status quo fails linking with a cryptic error message where previous version was working out of the box.
This sets the default to the static behavior on Scala.js and Scala Native so users don't need to care.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -730,7 +730,6 @@ object contrib extends Module {
 
   object buildinfo extends ContribModule {
     def moduleDeps = Seq(scalalib, scalajslib, scalanativelib)
-    def testModuleDeps = super.testModuleDeps 
   }
 
   object proguard extends ContribModule {
--- a/build.sc
+++ b/build.sc
@@ -729,7 +729,7 @@ object contrib extends Module {
   }
 
   object buildinfo extends ContribModule {
-    def compileModuleDeps = Seq(scalalib)
+    def compileModuleDeps = Seq(scalalib, scalajslib, scalanativelib)
     def testModuleDeps = super.testModuleDeps ++ Seq(scalalib)
   }
 
--- a/build.sc
+++ b/build.sc
@@ -729,8 +729,8 @@ object contrib extends Module {
   }
 
   object buildinfo extends ContribModule {
-    def compileModuleDeps = Seq(scalalib, scalajslib, scalanativelib)
-    def testModuleDeps = super.testModuleDeps ++ Seq(scalalib, scalajslib, scalanativelib)
+    def moduleDeps = Seq(scalalib, scalajslib, scalanativelib)
+    def testModuleDeps = super.testModuleDeps 
   }
 
   object proguard extends ContribModule {
--- a/build.sc
+++ b/build.sc
@@ -730,7 +730,7 @@ object contrib extends Module {
 
   object buildinfo extends ContribModule {
     def compileModuleDeps = Seq(scalalib, scalajslib, scalanativelib)
-    def testModuleDeps = super.testModuleDeps ++ Seq(scalalib)
+    def testModuleDeps = super.testModuleDeps ++ Seq(scalalib, scalajslib, scalanativelib)
   }
 
   object proguard extends ContribModule {
--- a/contrib/buildinfo/src/mill/contrib/buildinfo/BuildInfo.scala
+++ b/contrib/buildinfo/src/mill/contrib/buildinfo/BuildInfo.scala
@@ -2,6 +2,8 @@ package mill.contrib.buildinfo
 
 import mill.{PathRef, T}
 import mill.scalalib.{JavaModule, ScalaModule}
+import mill.scalanativelib.ScalaNativeModule
+import mill.scalajslib.ScalaJSModule
 
 trait BuildInfo extends JavaModule {
 
@@ -20,7 +22,11 @@ trait BuildInfo extends JavaModule {
    * rather than the default behavior of storing them as a JVM resource. Needed
    * to use BuildInfo on Scala.js which does not support JVM resources
    */
-  def buildInfoStaticCompiled: Boolean = false
+  def buildInfoStaticCompiled: Boolean = this match {
+    case _: ScalaJSModule => true
+    case _: ScalaNativeModule => true
+    case _ => false
+  }
 
   /**
    * A mapping of key-value pairs to pass from the Build script to the


lihaoyi (approved):

lefou (commented):
IMHO, we should make static compilation the general default, on all platforms. After all, this feature is some kind of optimization of the developer experience and not without runtime costs.

lihaoyi:
The issue with making static compilation the default is that with common practice - setting versions based on git dirty sha, including the version as a buildinfo field, and using buildinfo in some upstream util module - it ends up invalidating all compilation caches everywhere if you touch any file in the repo.

It's really problematic, and made working with the com-lihaoyi/mill repo very annoying for me. I don't think we should make that the common-case default for users running Scala on the JVM

lefou:
I see it different, but it is configurable after all. I think, we should add the implications (pros, cons) to the docs of the setting, as these are far from obvious to a user.

IMHO, a project should test the same artifacts it delivers, and using a fine grained classpath in (integration) tests but a jar in production always has some potential to fail due to the mismatch. E.g. in one of my client projects OSGi is used, which has enriched metadata in the jar manifest and it is key to use the full jars downstream when testing. This has an impact to the developer experience, but instead of compromising accuracy, we just changed the way we format  dirty git hashes in version numbers (we keep the "dirty", but not the hash). And as another side note: I still have to fix `mill-osgi` plugin for Mill 0.11.0-M10 because of that initial change. Apparently I need to override a lot more targets now, to restore the old behavior.

lolgab:
Another possibility is to ship it without a default and let users decide. It's not that nice fox UX but allows users to take informed decisions.

lefou:
> Another possibility is to ship it without a default and let users decide. It's not that nice fox UX but allows users to take informed decisions.

I think a sane default is better than non, in this case. I'd assume, most projects don't include a dirty git hash (which changes with every file change) in their versions, and are not affected at all. Those projects will profit from a static config default. Projects that have some highly dynamic content in their build info like Mill, might want to flip that setting. My own projects also have dynamic build info content, yet I still prefer the safer setting, as it can't be the cause of any trouble. The safer settings also makes the implementation simpler, as we don't need to match based on the module type, but which was needed to implement this PR. Also, there are already a lot of settings to override when using the BuildInfo plugin. That was the main reason I typically just wrote the generated files by hand instead of using the BuildInfo plugin.

lihaoyi:
I think this is fine to merge as is. 

The added complexity isn't too bad. In future, we could even bundle up some JSON with the Scala.js Javascript to support out-of-band buildinfo bundling there, and come up with some solution for Scala-Native: not sure what, but definitely possible. People have been bundling data with their executables since forever.

I think "version string includes dirty without the SHA" isn't a good compromise. The SHA is useful. If I'm running `publishLocal`, or my own internal Maven server, or distributing jars, I want the SHA. Even for quick hacks with un-committed code, the SHA helps ensure I can distinguish distinct builds of the code, avoiding confusion about artifacts that look identical but are not. Furthermore, it does not solve the problem: I don't see why running `git commit` should require me to re-compile all my code. Why should we punish users for "commit early, commit often"?

It's true that compiling things into bytecode is simpler, but it's too simple: given the common-usage-pattern of `mill-vcs-version` we demonstrate throughout the com-lihaoyi projects, and the common case of `BuildInfo`  being used to propagate the version, it ends up totally disabling a major feature of Mill.

People with advanced use cases would already be equipped with the skills for dealing with libraries requiring auxiliary non-classfile classpath entries, so it won't come as too big a surprise: SPI files, conf files, manifests, etc.. They would also have larger projects that would suffer most from the lack of incrementality, which might not matter that much to you, but does matter to many. 

Let's keep the flag for the people who prefer it, and this PR flipping the default for JS/Native is fine since we don't have an alternative for now. But for the bulk of people with JVM code with uncomplicated classpaths, or with complicated classpaths but the skills to manage them, leaving the flag off by default is the right thing to do.

build.sc
```diff
@@ -729,8 +729,8 @@ object contrib extends Module {
   }
 
   object buildinfo extends ContribModule {
-    def compileModuleDeps = Seq(scalalib, scalajslib, scalanativelib)
-    def testModuleDeps = super.testModuleDeps ++ Seq(scalalib, scalajslib, scalanativelib)
+    def moduleDeps = Seq(scalalib, scalajslib, scalanativelib)
```
lefou, line 732: This means, if somebody includes this module into his project, it will also transitively include the Mill dependendencies. And if the versions do not match exactly, we end up with two different versions in the classpath.

lihaoyi: But as a contrib module, wouldn't the versions always line up? Since contrib modules are co-versioned with Mill

lefou: Yeah, if used as we suggest. But I've seen also other constructs in the wild. And sometimes, these are even valid. We have no guard against it. The error are typically some strange compiler error messages and aren't helpful.

lihaoyi: Got it. I'll open a follow up to move back to `compileIvyDeps`

lolgab: > And if the versions do not match exactly, we end up with two different versions in the classpath.

Doesn't the version get evicted to the newer one?

lefou: Now, that we control the whole build process of `build.sc` scripts with Mill, we could probably enhance our dependency management. E.g. we could apply some `forceVersion` to all Mill dependencies and such things.

lihaoyi: Yeah that works too. Or just skip them entirely during resolution (is that possible???) since they'll all be bundled anyway

lihaoyi: > Doesn't the version get evicted to the newer one?

It wouldn't, because the version bundled in the Mill assembly isnt resolved through ivy/maven and thus can't be evicted nor can it evict the other

lefou: > > Doesn't the version get evicted to the newer one?
> 
> It wouldn't, because the version bundled in the Mill assembly isnt resolved through ivy/maven and thus can't be evicted nor can it evict the other

Also, in case we include a newer transitive Mill, it could still fail.

Maybe, we should try to resolve them from coursier and reduce the size of our bundle further. If we only come with our runner and coursier, we could load everything else via coursier on demand.

lihaoyi: Opened a PR to go back to the earlier behavior via https://github.com/com-lihaoyi/mill/pull/2562

The non-transitive nature of `compileModuleDeps` is definitely unfortunate; it means that anyone who depends on the `buildInfo` plugin needs to manually depend on `scalajslib` and `scalanativelib` as well. This isn't a problem for user code, but can be an annoyance for plugins.

I think the status quo is probably fine for 0.11.0. After that, we can consider other options, e.g. making use of `exclude` to allow normal dependencies while automatically removing the Mill jars from dependency resolution.

Shrinking the bundle is possible. With the current bootstrapping mechanism, we would need to keep at least `scalalib`, since that's required for `MillBootstrapRootModule`. With an alternative bootstrapping mechanism, we could cut it further, though it's unclear whether that's worth the complexity: for small size, the current bootstrap script already wins. We can leave that debate for later

PR #2568, opened 2023-06-04
Make `contrib.buildinfo` use `compileModuleDeps` rather than `moduleDeps`

lihaoyi:
Follow up from https://github.com/com-lihaoyi/mill/pull/2562, to avoid potentially having duplicate Mill classfiles on the classpath


lolgab:
If we want to avoid the dependency to `scalajslib` and `scalanativelib` we can maybe use reflection or structural types to check if the module is a `ScalaJSModule`, for example by checking if the module has `scalaJSVersion: Target[String]` available or `scalaNativeVersion: Target[String]`.

PR #2569, opened 2023-06-05
Make buildinfo not depend on scalajslib and scalanativelib

lolgab:
This uses structural types and catching `NoSuchMethodException` to check if a module is a `ScalaJSModule` or a `ScalaNativeModule`.
This way, users of `buildinfo` don't have to depend no `scalajslib` and `scalanativelib` mitigating the transitive dependencies problem.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -701,8 +701,7 @@ object contrib extends Module {
 
     // So we can test with buildinfo in the classpath
     def testModuleDeps =
-      super.testModuleDeps ++
-      Seq(scalalib, scalajslib, scalanativelib, contrib.buildinfo)
+      super.testModuleDeps ++ Seq(scalalib, contrib.buildinfo)
 
     // Worker for Scoverage 1.x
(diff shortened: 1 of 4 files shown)


lefou (commented):
I'm a bit reluctant to this change. This contrib module is always bound to the exact in-repo Mill version, and that version currently always includes the `scalaljslib` and `scalanativelib` module in the classpath. The structural types on the other hand, although unlikely,  may match unrelated modules, resulting in unexpected or unwanted behavior. IMHO, either we match on the exact class names in the type hierarchy, as we want to detect these exact classes, or we don't check at all. We can revisit, once we change the default Mill classpath.

lihaoyi:
Yeah, I think trying to solve this using structural types isn't worth the complexity. A bit of awkwardness around transitive plugin dependencies is ok for now, especially since it's the status quo. We can take our time to resolve it properly later, via coursier exclude configuration or something else

PR #2572, opened 2023-06-06
Renamings for conciseness: Scala*ModuleTests -> Scala*Tests, InnerCrossModule -> CrossValue

lihaoyi:
Last minute idea I had before we release 0.11.0 final

The current names are pretty verbose, especially for things that people would use often, and especially compared to the old `Tests` name. This cuts down the verbosity considerably.

WDYT @lefou @lolgab? If you don't like this I can skip it and release 0.11.0 as is, but I thought it's worth bringing up before the names become locked in for a while

### DIFF
--- a/ci/mill-bootstrap.patch
+++ b/ci/mill-bootstrap.patch
@@ -1,0 +1,13 @@
+diff --git a/build.sc b/build.sc
+index 8a61eaa089..464d12d012 100644
+--- a/build.sc
++++ b/build.sc
+@@ -315,7 +315,7 @@ trait MillScalaModule extends ScalaModule with MillJavaModule { outer =>
+ 
+   /** Default tests module. */
+   lazy val test: MillScalaModuleTests = new MillScalaModuleTests {}
+-  trait MillScalaModuleTests extends ScalaModuleTests with MillBaseTestsModule {
++  trait MillScalaModuleTests extends ScalaTests with MillBaseTestsModule {
+     def runClasspath = super.runClasspath() ++ writeLocalTestOverrides()
+     def forkArgs = super.forkArgs() ++ outer.testArgs()
+     def moduleDeps = outer.testModuleDeps
--- a/example/thirdparty/fansi/build.sc
+++ b/example/thirdparty/fansi/build.sc
@@ -20,7 +20,7 @@ trait FansiModule extends PublishModule with CrossScalaModule with PlatformScala
 
   def ivyDeps = Agg(ivy"com.lihaoyi::sourcecode::0.3.0")
 
-  trait FansiTestModule extends ScalaTests with TestModule.Utest {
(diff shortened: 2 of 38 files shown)


lolgab:
I like it! 👍

lefou:
My acceptance criteria for renaming the test traits is to provide some (deprecated) compatibility API, and to follow some pattern strictly (which this PR already does, I think), e.g. outer module name minus "Module" plus "Tests".

lefou (changes requested):
We need at least some compatibility shims for the test traits, as the existing names are already used for a while.

lihaoyi:
@lefou Great! I can add type aliases for migration compatibility, but if you're otherwise tolerant of this change I'll get it in before 0.11.0 final

scalanativelib/src/mill/scalanativelib/ScalaNativeModule.scala
```diff
@@ -31,6 +31,8 @@ trait ScalaNativeModule extends ScalaModule { outer =>
   def scalaNativeVersion: T[String]
   override def platformSuffix = s"_native${scalaNativeBinaryVersion()}"
 
+  @deprecated("use ScalaNativeTests", "0.11.0")
```
lolgab, line 34:
```suggestion
  @deprecated("use ScalaNativeTests instead.", since = "mill 0.11.0")
```

scalajslib/src/mill/scalajslib/ScalaJSModule.scala
```diff
@@ -19,6 +19,8 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
 
   def scalaJSVersion: T[String]
 
+  @deprecated("use ScalaJSTests", "0.11.0")
```
lolgab, line 22:
```suggestion
  @deprecated("use ScalaJSTests instead.", since = "mill 0.11.0")
```

ci/mill-bootstrap.patch
lefou, line 1: Technically, this patch should not be necessary.

example/thirdparty/fansi/build.sc
```diff
@@ -20,27 +20,27 @@ trait FansiModule extends PublishModule with CrossScalaModule with PlatformScala
 
   def ivyDeps = Agg(ivy"com.lihaoyi::sourcecode::0.3.0")
 
-  trait FansiTestModule extends ScalaModuleTests with TestModule.Utest {
+  trait FansiTestModule extends ScalaTests with TestModule.Utest {
```
lefou, line 23: To establish the pattern, we might want to follow it in the examples.

```suggestion
  trait FansiTests extends ScalaTests with TestModule.Utest {
```

lefou (approved):
Looks good to me!

Issue #2579, opened 2023-06-07
`[error] class type required but {some traits} found _root_.mill.define.Discover[build]`


lefou:
## Issue

With Mill `0.11.0` and `0.11.0-M11` I see the following error message when using the scoverage contrib module:

```
[error] /tmp/out/mill-build/generateScriptSources.dest/millbuild/build.sc:14:32: class type required but mill.define.Module.BaseClass with ScoverageModule.this.ScoverageData found
[error]     _root_.mill.define.Discover[build]
[error]                                ^
```

## Real World Projects

* Initial discovered in https://github.com/lefou/mill-vcs-version/pull/121
* Also in https://github.com/lefou/mill-integrationtest/pull/164

## Minimized Reproduction

```scala
import $ivy.`com.lihaoyi::mill-contrib-scoverage:`
import mill._
import mill.contrib.scoverage.ScoverageModule
import mill.scalalib._

object core extends Cross[CoreCross]("2.13.11")
trait CoreCross extends CrossScalaModule with ScoverageModule {
  override def scoverageVersion = "2.0.10"
}
```

```
$ mill __.compile
Using mill version 0.11.0-M11
[build.sc] [41/49] compile 
[info] compiling 1 Scala source to /home/lefou/work/opensource/mill-vcs-version/repro-discover-issue/out/mill-build/compile.dest/classes ...
[error] /home/lefou/work/opensource/mill-vcs-version/repro-discover-issue/out/mill-build/generateScriptSources.dest/millbuild/build.sc:14:32: class type required but mill.define.Module.BaseClass with ScoverageModule.this.ScoverageData found
[error]     _root_.mill.define.Discover[build]
[error]                                ^
[error] one error found
1 targets failed
compile Compilation failed
```

```scala
$ cat out/mill-build/generateScriptSources.dest/millbuild/build.sc 

package millbuild
import _root_.mill._
import mill.runner.MillBuildRootModule
object MiscInfo_build{
  implicit val millBuildRootModuleInfo: _root_.mill.runner.MillBuildRootModule.Info = _root_.mill.runner.MillBuildRootModule.Info(
    Vector("/home/lefou/.cache/mill/download/0.11.0-M11", "/home/lefou/work/opensource/mill-vcs-version/repro-discover-issue/out/mill-launcher/0.11.0-M11.jar").map(_root_.os.Path(_)),
    _root_.os.Path("/home/lefou/work/opensource/mill-vcs-version/repro-discover-issue"),
[...]

Issue #2582, opened 2023-06-07
Exception in thread "MillServerActionRunner" `java.lang.ExceptionInInitializerError`


lefou:
Mill 0.11.0

Project: https://github.com/lefou/mill-vcs-version/commit/d09eb3d9e5acd573c59ca08a6fd7d5b351118293

```
$ mill -j 0 __.compile + __.testCached
[build.sc] [#00] [22/49] cliImports 
Exception in thread "MillServerActionRunner" java.lang.ExceptionInInitializerError
        at java.base/jdk.internal.misc.Unsafe.ensureClassInitialized0(Native Method)
        at java.base/jdk.internal.misc.Unsafe.ensureClassInitialized(Unsafe.java:1042)
        at java.base/jdk.internal.reflect.UnsafeFieldAccessorFactory.newFieldAccessor(UnsafeFieldAccessorFactory.java:43)
        at java.base/jdk.internal.reflect.ReflectionFactory.newFieldAccessor(ReflectionFactory.java:186)
        at java.base/java.lang.reflect.Field.acquireFieldAccessor(Field.java:1105)
        at java.base/java.lang.reflect.Field.getFieldAccessor(Field.java:1086)
        at java.base/java.lang.reflect.Field.get(Field.java:418)
        at mill.runner.MillBuildBootstrap$.getRootModule(MillBuildBootstrap.scala:347)
        at mill.runner.MillBuildBootstrap.evaluateRec(MillBuildBootstrap.scala:106)
        at mill.runner.MillBuildBootstrap.evaluate(MillBuildBootstrap.scala:45)
        at mill.runner.MillMain$.$anonfun$main0$5(MillMain.scala:222)
        at mill.runner.Watching$.watchLoop(Watching.scala:27)
        at mill.runner.MillMain$.$anonfun$main0$1(MillMain.scala:210)
        at scala.util.DynamicVariable.withValue(DynamicVariable.scala:59)
        at scala.Console$.withErr(Console.scala:193)
        at mill.api.SystemStreams$.$anonfun$withStreams$2(SystemStreams.scala:56)
        at scala.util.DynamicVariable.withValue(DynamicVariable.scala:59)
        at scala.Console$.withOut(Console.scala:164)
        at mill.api.SystemStreams$.$anonfun$withStreams$1(SystemStreams.scala:55)
        at scala.util.DynamicVariable.withValue(DynamicVariable.scala:59)
        at scala.Console$.withIn(Console.scala:227)
[...]

PR #2583, opened 2023-06-07
Fix type annotation for inner scoverage module

lefou:
* [x] Fix https://github.com/com-lihaoyi/mill/issues/2579

I also changed from `lazy val` to `def` to make overridding and `super.scoverageData`-calls easier.

### DIFF
--- a/integration/feature/scoverage/repo/build.sc
+++ b/integration/feature/scoverage/repo/build.sc
@@ -1,0 +1,226 @@
+// mill plugins
+import $ivy.`de.tototec::de.tobiasroeser.mill.vcs.version::0.4.0`
+import $ivy.`de.tototec::de.tobiasroeser.mill.integrationtest::0.7.1`
+import $ivy.`com.lihaoyi::mill-contrib-scoverage:`
+import $ivy.`com.github.lolgab::mill-mima::0.0.23`
+
+// imports
+import mill._
+import mill.contrib.scoverage.ScoverageModule
+import mill.define.{Command, Sources, Target, Task, TaskModule}
+import mill.scalalib._
+import mill.scalalib.api.ZincWorkerUtil
+import mill.scalalib.publish._
+import de.tobiasroeser.mill.integrationtest._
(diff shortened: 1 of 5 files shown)


lihaoyi (approved):

lefou:
I added a reproduction for issues
* https://github.com/com-lihaoyi/mill/issues/2579
* https://github.com/com-lihaoyi/mill/issues/2582

integration/feature/scoverage/repo/build.sc
```diff
@@ -0,0 +1,174 @@
+// Reproduction of issue https://github.com/com-lihaoyi/mill/issues/2579
```
lihaoyi, line 1: Can we minimize this? The external plugins will cause headaches in future if we let them get scattered throughout the codebase, but even ignoring those this test suite  is way larger than what I would expect is needed to reproduce the two issues.

lefou: Sure, this is WIP, to find the issue soon.

integration/feature/scoverage/repo/build.sc
```diff
@@ -0,0 +1,28 @@
...
+// mill plugins
+import $ivy.`com.lihaoyi::mill-contrib-scoverage:`
+import $ivy.`com.github.lolgab::mill-mima::0.0.23`
+
+// imports
+import mill._
+import mill.contrib.scoverage.ScoverageModule
+import mill.scalalib._
+
+val baseDir = build.millSourcePath
```
lefou, line 13: @lihaoyi Once, I remove this line, the test succeeds.

lefou:
I'm going to split this PR up to only fix the scoverage issue. I'll open a new PR for the reproducer of the second issue.

integration/feature/scoverage/repo/build.sc
lefou: I wonder whether we should add another section "repro" to explicitly keep reproducers for known issues.

PR #2584, opened 2023-06-08
Refined fix for `ScoverageModule` inner module

lefou:
This is a follow-up on #2583.

Using a ModuleRef make accesss to the inner module for the CLI impossible.
I added that case to the integration test and changed to use a `val`.


lihaoyi:
Yeah `ModuleRef` was added somewhat intentionally as a workaround to allow you to refer to a modulenwithout making it visible from the CLI; e.g. you don't want everyone's `zincWorker` to be spamming your mill resolve __. Going back to `def` for modules you do want to be visible sounds reasonable

Issue #2602, opened 2023-06-12
Resolution of test targets


ex0ns:
I think this is related to 0.11.0 as well, with a simple `build.sc`:

```scala
import mill._
import scalalib._


object core extends ScalaModule {
  def scalaVersion = "2.13.11"

  object test extends ScalaTests with TestModule.ScalaTest {

  }
}
```

The resolution of the tests targets is as follow:

```
mill resolve __.test
[1/1] resolve 
core.test
core.test.test
```

This is an issue when running the tests (mostly in parallel ? I can't confirm for sequential), for instance using `mill -j 0 __.test` because it runs all the tests twice and can has some unexpected race condition (when using zio-test for instance, the test often crashes because of a missing report file), furthermore the output of the test command is (without any surprise) duplicated and hence hard to follow/read as multiple execution of the same tests are happening at the same time.

A workaround has been to change my build process to call `mill -j 0 __.test.test` instead as it only resolve the inner target.
I think that we should take care of de-duplicating targets because running the build, this could improve the performances and reliability of mill.

For reference here is the output of the inspect command for this build:

```
core.test.test(build.sc:9)
    Discovers and runs the module's tests in a subprocess, reporting the
    results to the console.
    @see [[testCached]]

    args <str>...

Inputs:
    core.test.testUseArgsFile
    core.test.forkArgs
    core.test.forkArgs
    core.test.testFramework
    core.test.runClasspath
    core.test.compile
    core.test.runClasspath
    core.test.forkEnv
    core.test.forkWorkingDir

core.test.test(build.sc:9)
    Discovers and runs the module's tests in a subprocess, reporting the
    results to the console.
    @see [[testCached]]

    args <str>...

Inputs:
    core.test.testUseArgsFile
    core.test.forkArgs
    core.test.forkArgs
    core.test.testFramework
    core.test.runClasspath
    core.test.compile
    core.test.runClasspath
    core.test.forkEnv
    core.test.forkWorkingDir
```

Here we can see that both target are indeed referencing the same build (same line, same inputs, same doc).

lefou:
So the with duplicated tests is with Mill 0.10.12 or 0.11.0?

_Can you please ask the second topic as a question in the discussions? Otherwise we loose track of the original issue and nobody will find the potential interesting answer to the other question._

ex0ns:
I didn't test it with 0.10.12, maybe it's doing the same, I don't know, I noticed that when upgrading project to mill 0.11.0

[...]

PR #2604, opened 2023-06-13
Properly distinct-ify commands when resolved multiple times by query

lihaoyi:
Should fix https://github.com/com-lihaoyi/mill/issues/2602. The previous `.distinct` doesn't work because `T.command`s take arguments, and their `Command` instances cannot be cached the same way that `T{}` and other tasks can be. Also, the fact that it was in `RunScript` meant that code resolving tasks without running them didn't benefit, e.g. when you run `inspect __.test`, so I moved it into `Resolve` with separate implementations for `Resolve.Task` and `Resolve.Segments`

Tested manually, both `./mill -i dev.run example/basic/1-simple-scala -i inspect __.test` and `./mill -i dev.run example/basic/1-simple-scala -i __.test` no longer contain duplicates. Also covered by new unit tests to ensure that various kinds of potential duplication (via wildcards, brace-expansion, `+`) all are properly deduplicated

### DIFF
--- a/main/resolve/src/mill/resolve/Resolve.scala
+++ b/main/resolve/src/mill/resolve/Resolve.scala
@@ -28,7 +28,7 @@ object Resolve {
       Right(resolved.map(_.segments))
     }
 
-    private[mill] def deduplicate(items: List[Segments]) = items.distinct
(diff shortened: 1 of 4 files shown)


lefou (approved):
Looks good to me. Thanks for the quick fix!

PR #2605, opened 2023-06-13
Fix source links in API documentation

ysthakur:
The API documentation has links to local files (e.g. [mill.define.Args](https://mill-build.com/api/latest/mill/define/Args.html) points to `file:///home/runner/work/mill/mill/main/define/src/mill/define/Args.scala`). This PR is for fixing that.

There's already code to replace every occurrence of `file://path/to/project/root` with the right URL, but it's for `.scala` files so I didn't touch it. I figured `-sourcepath` was a safer way to do the same thing.


lefou (approved):
Looks good, I guess. Thank you!

Issue #2601, opened 2023-06-12
Better error message for empty selector


ysthakur:
If `mill` is run without arguments, the error message says `Selector cannot be empty`. A user may not understand that this means they need to give it arguments. It'd be nice if the error message were changed to something like ``Selector cannot be empty. Try `mill resolve _` to see what's available.`` (similar to the message you get if you run `mill nonexistentthing`). It could also print the help message instead, as some other applications do.

PR #2610, opened 2023-06-14
Make empty selector message more helpful

ysthakur:
Regarding #2601. Currently, if no arguments are given to mill, it says `Selector cannot be empty`. It's not immediately clear what that's supposed to mean, so I figured it might be more helpful to print the help message instead in that case (some other tools do that). In addition to this, I modified the empty selector message to say ``Selector cannot be empty. Try `mill resolve _` to see what's available.`` (the empty selector message still shows up if you do `./mill --repl`).

### DIFF
--- a/docs/modules/ROOT/pages/Installation_IDE_Support.adoc
+++ b/docs/modules/ROOT/pages/Installation_IDE_Support.adoc
@@ -123,7 +123,7 @@ https://gitforwindows.org[Git-Bash],
 https://docs.microsoft.com/en-us/windows/wsl[WSL]); to get started, follow the instructions in the <<_manual>>
 section. Note that:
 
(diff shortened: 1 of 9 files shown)


lefou (commented):
I think showing a typical `--help` test without request isn't an expected user experience. If there is an issue, we should tell it the user. There is a large gap between some usage hint and an error message with details.

lefou:
While reviewing this PR, I recognized that we left some traces of --repl suport in the code base. I opened #2611 to fix that.

lefou:
> Does `` Selector cannot be empty. Try `mill resolve _` to see what's available. `` look okay? 

I think so.

> Maybe instead of saying "selector," it could say you need a target? (since it's not immediately clear what a selector is)

How about "Target selector cannot ..."

lefou (approved):
Thank you!

PR #2611, opened 2023-06-15
Cleanup leftover --repl flag handling, fail when --repl is given

lefou:
In Mill `0.11.0-M8` we change Mill internals to no longer use Ammonite. Since then, we no longer support the `--repl` mode, but forgot to cleanup the CLI configuration.


lefou:
@lihaoyi We probably also want to remove the `--predef` and `--no-default-predef` options?

lihaoyi (approved):
Yeah `--predef` and `--no-default-predef` are both irrelevant now

PR #2613, opened 2023-06-16
Fix proper override hierarchy of `JavaModule.ivyDeps`

lefou:
Since `JavaModule` now inherits the `ivyDeps` target from `TestModule.JavaModuleBase`, we no longer should default implement it with the empty `Agg`, instead we need to preserve what's potentially already in it. I simply removed the re-definition and moved the ScalaDoc up.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -530,11 +530,10 @@ object scalalib extends MillStableScalaModule {
   def testIvyDeps = super.testIvyDeps() ++ Agg(Deps.scalaCheck)
   def testTransitiveDeps = super.testTransitiveDeps() ++ Seq(worker.testDep())
 
-  override def mimaBackwardIssueFilters: T[Map[String, Seq[ProblemFilter]]] = Map(
-    "0.11.0" -> Seq(
-      ProblemFilter.exclude[ReversedMissingMethodProblem](
-        "mill.scalalib.JavaModule.mill$scalalib$JavaModule$$super$ivyDeps"
-      )
+  override def mimaBinaryIssueFilters: T[Seq[ProblemFilter]] = Seq(
+    // See https://github.com/com-lihaoyi/mill/pull/2613
+    ProblemFilter.exclude[ReversedMissingMethodProblem](
+      "mill.scalalib.JavaModule.mill$scalalib$JavaModule$$super$ivyDeps"
     )
   )
 


lefou:
This PR results in one MiMa issue: 

```
Found 1 issue when checking against com.lihaoyi:mill-scalalib_2.13:0.11.0
 * abstract synthetic method mill$scalalib$JavaModule$$super$ivyDeps()mill.define.Target in interface mill.scalalib.JavaModule is present only in current version
   filter with: ProblemFilter.exclude[ReversedMissingMethodProblem]("mill.scalalib.JavaModule.mill$scalalib$JavaModule$$super$ivyDeps")
1 targets failed
scalalib.mimaReportBinaryIssues Failed binary compatibility check! Found 1 potential problems

```

I'm not totally sure whether this is acceptable or an issue for our bin-compat story. To be honest, I have no idea, when this should be an issue at all. The only code calling `super.ivyDeps` which get's resolved to `JavaModuleBase.ivyDeps` is the new implementation, which is no issue. Since we forward the value of `JavaModuleBase.ivyDeps`, it should also be no issue for derived modules. WDYT?

lihaoyi:
TBH I have no idea. 

1. Normally a new abstract method is a clear bincompat breakage, because subclasses compiled against the old version of the class would not implement all the methods necessary for the new version of the class. 

2. In this case, the abstract method is synthetic, so it shouldn't be called by user code. 

3. But presumably it must be called by *some* code somewhere, otherwise the compiler wouldn't need to add it! 

4. So this method is important for some code path somewhere, and subclasses compiled against the old class not providing an implementation would be a problem for that code path.

Maybe it's worth asking on the #scala-contributors discord to see if any of the compiler folks know? @lolgab do you have any idea? This question goes into the details of the Scala compiler's desugaring into JVM classes, interfaces and forwarders, which is beyond my area of expertise

lefou:
After consulting [#scala-contributors](https://app.element.io/#/room/#scala-contributors:matrix.org/$AMm0iAYbQWc_zm788IbP51_jiRrUTZ9KMyZ-kWGvCrQ) (How I hate, that #Scala chat rooms are on a closed platform), I came to the conclusion that it is not safe to merge this. I'll close this (to keep it as-is) and create a new PR with some `TODO` comment for later changes.

Issue #2595, opened 2023-06-10
Exception when running tests using ScalaTest


ex0ns:
When upgrading to mill 0.11.0 I got the following stack trace when trying to run my tests using scalatest

```
ex0ns@build:~/tmp$ mill  __.test
[82/83] core.test.test.super.mill.scalalib.TestModule.test 
java.lang.NoClassDefFoundError: sbt/testing/EventHandler
java.lang.NoClassDefFoundError: sbt/testing/EventHandler
        at mill.testrunner.TestRunnerMain0$.main0(TestRunnerMain0.scala:29)
        at mill.testrunner.TestRunnerMain0.main0(TestRunnerMain0.scala)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:568)
        at mill.testrunner.entrypoint.TestRunnerMain.main(TestRunnerMain.java:41)
Caused by: java.lang.ClassNotFoundException: sbt.testing.EventHandler
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:520)
        at mill.testrunner.entrypoint.TestRunnerMain$1.findClass(TestRunnerMain.java:30)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:587)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:520)
        ... 7 more
```

The minimized `build.sc`:
```scala
import mill._
import mill.scalalib._

object library {

  object Version {
    val scalaTest  = "3.2.16"
    val sbtTesting = "1.0"
  }

  val scalaTest  = ivy"org.scalatest::scalatest::${Version.scalaTest}"
  val sbtTesting = ivy"org.scala-sbt:test-interface:${Version.sbtTesting}"
}

object core extends ScalaModule {
  def scalaVersion = "2.13.11"

[...]

PR #2617, opened 2023-06-17
Convert entrypoint.runIvyDeps to ivyDeps

lefou:
As runtime dependencies are not transitive in Mill, we need to declare the sbt test-interface as normal dependency.

* Fix: #2595

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -514,7 +514,7 @@ object testrunner extends MillPublishScalaModule {
   def moduleDeps = Seq(scalalib.api, main.util, entrypoint)
 
   object entrypoint extends MillPublishJavaModule {
-    override def runIvyDeps = Agg(Deps.sbtTestInterface)
+    override def ivyDeps = Agg(Deps.sbtTestInterface)
   }
 }
 


build.sc
```diff
@@ -514,7 +514,7 @@ object testrunner extends MillPublishScalaModule {
   def moduleDeps = Seq(scalalib.api, main.util, entrypoint)
 
   object entrypoint extends MillPublishJavaModule {
-    override def runIvyDeps = Agg(Deps.sbtTestInterface)
+    override def ivyDeps = Agg(Deps.sbtTestInterface)
```
lolgab, line 517:
```suggestion
    override def ivyDeps = super.ivyDeps() ++ Agg(Deps.sbtTestInterface)
```

lefou: Is there something we miss here. I think the `entrypoint` module is pretty dependency less, except the runtime dependency on `test-interface`.

lihaoyi: yeah there are no supers here to worry about. And this is probably the right thing: we *want* it to be dependency-free except for what's listed, rather than adding onto whatever may be there before

lihaoyi (approved):

PR #2623, opened 2023-06-20
Support default crossSegment selection with `cross[]` notation

lefou:
* Implements idea https://github.com/com-lihaoyi/mill/discussions/2588

With this change, one can select the default cross segment by leaving the cross selection empty.

```
$ mill cross[].compile
```

The default selection is the first provided cross value for each cross level.

To override the default cross segment, one can override the `defaultCrossSegments` method.

### DIFF
--- a/main/define/src/mill/define/Cross.scala
+++ b/main/define/src/mill/define/Cross.scala
@@ -340,6 +340,12 @@ class Cross[M <: Cross.Module[_]](factories: Cross.Factory[M]*)(implicit
     .mapValues(_.value)
 
   /**
+   * The default cross segments to use, when no cross value is specified.
+   * Defaults to the first cross value per cross level.
+   */
+  def defaultCrossSegments: Seq[String] = items.head.crossSegments
+
+  /**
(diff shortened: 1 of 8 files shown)


main/define/src/mill/define/Cross.scala
```diff
@@ -128,6 +128,7 @@ object Cross {
   )
 
   object Factory {
+    import scala.language.implicitConversions
```
lolgab, line 131: Why do we need implicit conversions?

lefou: This is to silence a compiler warning already present before this change.

lihaoyi:
Looks reasonable to me. I don't like the way `Cross` is configured by both passing parameters and overriding methods, but that's not fixable for now due to compatibility guarantees and will have to wait for 0.12.x

lefou:
In most cases, users will simply put their defaults in the first position. Overriding `defaultCrossSegments` is more likely, if at all, when cross values are provided dynamically.

Issue #2634, opened 2023-06-26
`MillBuildRootModule` should preset the `platformSuffix` to `_mill0.11`


lefou:
It's a typical use case to add plugin dependencies to a meta-build. To make this more convenient, we can use the platform notation for dependencies, which automatically handles the platform suffix.

Currently, the `platformSuffix` is unset and needs to be set manually, to allow platform notations.

For example:

```
// mill-build/build.sc
import mill._
import scalalib._

object millbuild extends MillBuildRootModule {
  def platformSuffix ="_mill0.11"
  def ivyDeps = Agg(
    // calculate the version from git version
    ivy"de.tototec::de.tobiasroeser.mill.vcs.version::0.4.0"
  )
}
```

PR #2635, opened 2023-06-26
Define reasonable `platformSuffix` in `MillBuildPlatformModule`

lefou:
It's a typical use case to add plugin dependencies to a meta-build. To make this more convenient, we set the `platformSuffix` to `_mill<binary-platform>`.

* Fix https://github.com/com-lihaoyi/mill/issues/2634

### DIFF
--- a/runner/src/mill/runner/MillBuildRootModule.scala
+++ b/runner/src/mill/runner/MillBuildRootModule.scala
@@ -11,6 +11,8 @@ import mill.scalalib.api.Versions
 import os.{Path, rel}
 import pprint.Util.literalize
 import FileImportGraph.backtickWrap
+import mill.main.BuildInfo
+
 import scala.util.Try
 
 /**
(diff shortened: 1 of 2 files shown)


lihaoyi:
This looks good to me. If we're relying on `platformSuffix`, should we also remove the special-case logic in https://github.com/com-lihaoyi/mill/blob/8e6684f71364114c2e6670188c71f0309f5632b4/runner/src/mill/runner/MillIvy.scala#L10-L16 that duplicates the handling of the second `::`?

lefou:
> This looks good to me. If we're relying on `platformSuffix`, should we also remove the special-case logic in
> 
> https://github.com/com-lihaoyi/mill/blob/8e6684f71364114c2e6670188c71f0309f5632b4/runner/src/mill/runner/MillIvy.scala#L10-L16
> that duplicates the handling of the second `::`?

I cleaned that up. We still need to pre-processor to handle special place holders and the empty version.

PR #2648, opened 2023-07-04
Refactored root modules finder

lefou:
Moved root module finder logic into dedicated `RootModuleFinder` object.

Changed `PrefixLogger` to accept a `Logger` instead of a `ColorLogger`.

Also added a `quiet` flag to `MainModule.resolve`. This is useful, if we need the return value, but do not want the command to print the resolved outcome.


lefou:
This PR is obsolete.

PR #2653, opened 2023-07-04
deps: update semanticdb to 4.8.1

ckipp01:
This just bumps the default version of semanticdb from 4.7.8 to 4.8.1. I
realized when using mill-scalafix that I needed to manually do this
since the new Mill version is still using an older semanticdb version
that doesn't support for example 2.13.11.


lefou (approved):
Thanks!

Issue #2650, opened 2023-07-04
assembly error


romain-gilles-ultra:
Hi mills,
I get an error when I try to run the output of the assembly command.
I created this repository to explain the issue: https://github.com/romain-gilles-ultra/mill-assembly-issue

In the big lines
If I add:
- `ivy"org.apache.spark::spark-core:3.4.0"`
- `ivy"dev.zio::zio:2.0.15"` or `ivy"org.typelevel::cats-core:2.9.0"`

Then when I try to run the `out.jar` I get an error.

Here the readme
# mill-assembly-issue
Mill 0.11.1 assembly issue demo

### Java
```
> java --version                                                                                                                         
openjdk 11.0.19 2023-04-18
OpenJDK Runtime Environment Temurin-11.0.19+7 (build 11.0.19+7)
OpenJDK 64-Bit Server VM Temurin-11.0.19+7 (build 11.0.19+7, mixed mode)

```

### Scala
```
> scala -version
Scala code runner version 2.13.11 -- Copyright 2002-2023, LAMP/EPFL and Lightbend, Inc.

```

### Mill
```
> ./mill version
[1/1] version 
0.11.0-19-d4e35f
```

## Description

Try to create an assembly of a project to run it with `java -jar foo.jar` but fail
I reproduce a bug I get in my project with a simple project and fiew but not small dependency.

The issue is when you add at the same time:
- `ivy"org.apache.spark::spark-core:3.4.0"`
- `ivy"dev.zio::zio:2.0.15"` or `ivy"org.typelevel::cats-core:2.9.0"`

When you run the following commands:
```
> ./mill foo.assembly
[build.sc] [41/49] compile 
[info] compiling 1 Scala source to xxx ...
[info] done compiling
[47/47] foo.assembly 
> java -jar out/foo/assembly.dest/out.jar --text tutu
Error: An unexpected error occurred while trying to open file out/foo/assembly.dest/out.jar
```

If you try the following it will work:
```
> ./mill foo.run --text tutu
[48/48] foo.run 
<h1>tutu</h1>
```

If you try the following to understand you get this:
```
java -cp out/foo/assembly.dest/out.jar ultra.Main
Error: Could not find or load main class ultra.Main
Caused by: java.lang.ClassNotFoundException: ultra.Main
```

While the class is in the jar
```
> unzip -l out/foo/assembly.dest/out.jar | grep ultra
error: End-of-centdir-64 signature not where expected (prepended bytes?)
[...]

Issue #528, opened 2019-01-18
JavaModule.assembly produces invalid or corrupt jar file (was: publishLocal as altenative to jars : which are corrupt often and at random)


siddhartha-gadgil:
I am getting a lot of corrupt jars, even though I have cleared/checked everything in sight several times. This happens to fat jars in sbt too, but there we have the alternative (for use in Jupyter, for example) of publishLocal. Can _mill_ provide such an option?

lihaoyi:
Mill already supports `.publishLocal`

lefou:
@siddhartha-gadgil So, what exactly is your issue? What is corrupt? And how did this happen?

siddhartha-gadgil:
Firstly, I have set up _publishLocal_ and this is working fine for the case
I needed now (Jupyter Notebooks with Almond). In my work desktop, but not
my home laptop, the output of assembly was corrupted for some modules. This
meant that executable binaries would crash with "mainClass ... not found"
and if I loaded in ammonite using `import $cp.myjar` subsequent commands
would not find the contents. I found online that I should try `jar tf`, and
this indeed confirmed cooruption.

Presumably some source/doc jar of a dependency is corrupted, causing the
upstream corruption. In fact it would be best for me to have a thinFat jar
for running, excluding the sources and docs (especially of dependencies).
Is that already an option?

lefou:
To be honest, I have no clue what you wanna tell us. It would be helpful to be more concrete. E.g. the exact mill cmdline you used when mill crashed. The exact error message.If possible, the `build.sc`...

Looks like you should better ask on https://gitter.im/lihaoyi/mill.

siddhartha-gadgil:
I did not mean that _mill_ crashes, but that the _executable jar_ generated by `mill myproject.assembly` crashes, and in a random and hard to diagnose (at least for me) way. The reason is almost certainly some cached corrupt jar, and probably a source/documentation jar.

* My main request was to `publishLocal`, as an alternative, which @lihaoyi pointed out is already part of mill.
* A feature that would be nice is a _minimal fat-jar_, i.e. excluding source and documentation dependencies, so reducing the chance of corrupt jars fouling this up.
* Even better would be better diagnostics while building jars, though this may not be a mill issue.

lihaoyi:
[...]

PR #2655, opened 2023-07-05
Analyze assembly error - Reproduction of #528 and #2650

lefou:
Reprodcution of issues
* https://github.com/com-lihaoyi/mill/issues/528
* https://github.com/com-lihaoyi/mill/issues/2650

### DIFF
--- a/integration/feature/assembly/repo/build.sc
+++ b/integration/feature/assembly/repo/build.sc
@@ -1,0 +1,21 @@
+import mill._, scalalib._
+
+trait Setup extends ScalaModule {
+  def scalaVersion = "2.13.11"
+  def ivyDeps = Agg(
+    ivy"com.lihaoyi::scalatags:0.8.2",
+    ivy"com.lihaoyi::mainargs:0.4.0",
+    ivy"org.apache.avro:avro:1.11.1",
+    ivy"dev.zio::zio:2.0.15",
+    ivy"org.typelevel::cats-core:2.9.0",
+    ivy"org.apache.spark::spark-core:3.4.0",
+    ivy"dev.zio::zio-metrics-connectors:2.0.8",
+    ivy"dev.zio::zio-http:3.0.0-RC2"
+  )
+}
+
+object foo extends Setup {
+  override def prependShellScript: T[String] = ""
+}
+
+object bar extends Setup
--- a/integration/feature/assembly/repo/foo/src/Main.scala
+++ b/integration/feature/assembly/repo/foo/src/Main.scala
@@ -1,0 +1,17 @@
+package ultra
+
+import scalatags.Text.all._
+import mainargs.{main, ParserForMethods}
+
+object Main {
+  def generateHtml(text: String) = {
+    h1(text).toString
+  }
+
+  @main
+  def main(text: String) = {
+    println(generateHtml(text))
+  }
+
+  def main(args: Array[String]): Unit = ParserForMethods(this).runOrExit(args)
+}
--- a/integration/feature/assembly/test/src/AssemblyTests.scala
+++ b/integration/feature/assembly/test/src/AssemblyTests.scala
@@ -1,0 +1,41 @@
+package mill.integration.local
+
+import mill.integration.IntegrationTestSuite
+import mill.util.Jvm
+import utest._
+
+// Ensure the assembly is runnable, even if we have assembled lots of dependencies into it
+// Reproduction of issues:
+// - https://github.com/com-lihaoyi/mill/issues/528
+// - https://github.com/com-lihaoyi/mill/issues/2650
+
+object AssemblyTests extends IntegrationTestSuite {
+  def tests: Tests = Tests {
+    test("Assembly") {
+      test("without-prependShellScript") {
+        val workspacePath = initWorkspace()
+        assert(eval("foo.assembly"))
+        val assemblyFile = workspacePath / "out" / "foo" / "assembly.dest" / "out.jar"
+        assert(os.exists(assemblyFile))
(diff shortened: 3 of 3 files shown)


lefou:
It looks to me, the main class can't be found if the jar reaches a specific size and the main class is located at the end of the jar.

I made some tests with my alpha-grade `mill-spring-boot` plugin, which also provided support to executable jars. The difference is, that not all jars are unpacked but included as is and a dedicated `main` is used. I added this failing test case and it can be started. My assumption is, that the specific spring-boot main class is located at the start of the JAR vs. in our case we append the main class at the end.

lihaoyi:
The JDK java code in that stack trace says

```java
                if (i >= total) {
                    // This will only happen if the zip file has an incorrect
                    // ENDTOT field, which usually means it contains more than
                    // 65535 entries.
                    initCEN(countCENHeaders(cen, limit), zc);
                    return;
                }
```

lihaoyi:
Another data point: taking the problematic jar, `unzip`ing it and re-`zip`ing it generates a non-problematic zip file, without a prepended shell script, that can be read without issue via `jar tf`, But `cat`ing it to append to it a launcher shell script again creates a final `jar` file that cannot be read by `jar tf` (same error as above) but can be read by `zip`.

This somewhat confirms that it isn't our implementation of create-a-zip-file that is problematic: a native unix `zip`-created archive, prepended with a shell script, fails to read via `jar tf` (which uses `java.util.zip.ZipFile`) but can be successfully read via `unzip -l`.

That means the problem has to be the interaction between the JVM zip reading code and the the prepended shell script, since the problem remains even after swapping out the zip _writing_ code, but the problem goes away if we remove the prepended shell script or we use a non-JVM program to read the prepended jar file

lefou:
I made almost the same tests a while ago. I think it worth to note that it's not the sheer file size, that makes the JRE zip implementation fail, as an assembly jar built with mill-spring-boot and the exact same dependencies can be started. The difference is, that the spring boot assembly contains a lot less zip entries, as it is not unpacking all the dependencies, but rather just embeds the original jars. So, it's either the entry count (=catalog size) or the position (index) of the main class in that catalog. An interesting test would be to add the main class first (don't append to the upstream assembly jar). Though, I don't think it will fix it, from looking at the JDK code. Maybe, we can count the zip entries ourself and print a warning if we detect, that it will be too large to be executable. What I also wonder, we can't be the first who failed into that trap.

lihaoyi:
Another data point, I tried this with SBT-1.8.2 and SBT-Assembly 2.2.1, and it has the same failure mode: `jar tf` fails if the assembly jar was created with `assemblyPrependShellScript := Some(defaultUniversalScript(shebang = true))`, and `jar tf` succeeds if the assembly jar was created without that config.

lihaoyi:
Another data point is that `zip` (note: not `unzip`!) provides different errors when trying to load the two prepended jars, but does not give any errors trying to process the non-prepended jars

```
lihaoyi mill$ zip /Users/lihaoyi/Github/mill/integration/feature/assembly/repo/out/exe/large/assembly.dest/out.jar
	zip warning: Zip64 EOCDR not found where expected - compensating
	zip warning: (try -A to adjust offsets)
	zip warning: expected 75516 entries but found 75512

zip error: Zip file structure invalid (/Users/lihaoyi/Github/mill/integration/feature/assembly/repo/out/exe/large/assembly.dest/out.jar)
lihaoyi mill$ zip /Users/lihaoyi/Github/mill/integration/feature/assembly/repo/out/exe/small/assembly.dest/out.jar
	zip warning: unexpected signature on disk 0 at 10221122

	zip warning: archive not in correct format: /Users/lihaoyi/Github/mill/integration/feature/assembly/repo/out/exe/small/assembly.dest/out.jar
	zip warning: (try -F to attempt recovery)

zip error: Zip file structure invalid (/Users/lihaoyi/Github/mill/integration/feature/assembly/repo/out/exe/small/assembly.dest/out.jar)
lihaoyi mill$ zip /Users/lihaoyi/Github/mill/integration/feature/assembly/repo/out/noExe/large/assembly.dest/out.jar

zip error: Nothing to do! (/Users/lihaoyi/Github/mill/integration/feature/assembly/repo/out/noExe/large/assembly.dest/out.jar)
lihaoyi mill$ zip /Users/lihaoyi/Github/mill/integration/feature/assembly/repo/out/noExe/small/assembly.dest/out.jar

zip error: Nothing to do! (/Users/lihaoyi/Github/mill/integration/feature/assembly/repo/out/noExe/small/assembly.dest/out.jar)
```

lihaoyi:
Final bit of debugging: it appears that the misbehavior of `jar tf` on a prepended jar starts when the number of entries in the jar file hits exactly 65535 zip entries. 

A prepended jar with 65534 entries can be `jar tf`ed without issue, while one with 65535 entries fails with the `java.util.zip.ZipException: invalid CEN header (bad signature)` error.  The size of the entries in the jar file doesn't seem to matter. Zeroing them out does not change the number of entries necessary to hit the problem.

It seems that there must be some change in the handling of zip files when the number of entries reaches 65535.

lefou:
Thanks for this great analysis, @lihaoyi. It would be nice, if we could evaluate the entry count and warn the user, if it is too large. We could continue, or fail or produce a jar without prepend script in case of too many entries. Don't know which option is the best though. All are unpredictable. Or we make the behavior configurable and default to fail, which feels like the best option to me.

lefou:
I opened PR https://github.com/com-lihaoyi/mill/pull/3140, which detects faulty assembly JARs, bases on [Haoyi's findings](https://github.com/com-lihaoyi/mill/pull/2655#issuecomment-1672468284) and fails the build with an actionable error message.

PR #2661, opened 2023-07-15
Avoid intermediate mapping in ZincWorkerImpl

lolgab:
Profiled Mill compiling Mill itself and 0.07% of the time was spent in mapping the result of `compileMixed0` to `CompilationResult`.
This changes the `compileInternal` method to already use the right types so less conversions are needed.


lihaoyi (approved):

lefou (approved):
The change looks good. But I don't think we save anything here, as the measured wrapping "overhead" comes most likely from the `PathRef` construction, which we now do in another place.

lolgab:
> The change looks good. But I don't think we save anything here, as the measured wrapping "overhead" comes most likely from the `PathRef` construction, which we now do in another place.

The gain is very small but doesn't make the code less readable so totally worth it for me.
profile

PR #2662, opened 2023-07-17
Fix `init` command and support running Mill without existing `build.sc`

### LINKED ISSUE #2660: Lost ability to install Mill for GHA when build.sc is not at the project root
Hi,

I have a project that builds with a mix of Mill and sbt (for historical reasons), this means that neither sbt nor Mill live at the repo's root level. This worked fine we tried to upgrade to Mill 0.11.1, but now doesn't due to a recent change:

https://github.com/com-lihaoyi/mill/blob/1e5871a8a0740193b339030b420331dc90bca091/runner/src/mill/runner/MillBuildBootstrap.scala#L71

_(To answer to the question: "No, I'm not, but I will be." :smile:)_

The [setup-mill](https://github.com/jodersky/setup-mill) action can no longer install Mill for us, producing this error:

image

I realise that...

1. It will be suggested we use the bootstrap script thingy.
2. I can probably fix this by placing a dummy `build.sc` at the root.
3. This is an edge case, most people don't do this.

But I thought I'd raise the issue in case it was indicative of a wider problem. Feel free to close the issue based on any or all of the above reasons.

lefou:
I first added an integration test for the init command to reproduce the issues and make sure the feature now works.

Instead of failing in case there is no `build.sc`, we now check the targets against a whitelist of tasks (`init` and `version`) and continue with an empty setup.

We also accept a missing `build.sc` in case additional imports are given, which could mean, the imports contain Mill plugins which itself provide commands that could work without a `build.sc`.

Only when we discover an error afterwards, we add a message about the missing `build.sc` to the underlying error message. That means, we don't see the spurious "no build.sc" message when everything works as expected.

* Fix #2660

### DIFF
--- a/bsp/worker/src/mill/bsp/worker/State.scala
+++ b/bsp/worker/src/mill/bsp/worker/State.scala
@@ -39,7 +39,8 @@ private class State(projectRoot: os.Path, baseLogger: ColorLogger, debug: String
       threadCount = None,
       targetsAndParams = Seq("resolve", "_"),
       prevRunnerState = mill.runner.RunnerState.empty,
-      logger = baseLogger
+      logger = baseLogger,
+      needBuildSc = true
     ).evaluate()
 
     val rootModules0 = evaluated.result.frames
--- a/integration/feature/init/test/src/MillInitTests.scala
+++ b/integration/feature/init/test/src/MillInitTests.scala
@@ -1,0 +1,16 @@
+package mill.integration
+package local
+import mill.bsp.Constants
+import utest._
+
+object MillInitTests extends IntegrationTestSuite {
+
+  def tests: Tests = Tests {
+    test("Mill init works") {
+      val workspacePath = initWorkspace()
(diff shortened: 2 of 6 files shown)


lefou:
@lihaoyi Before continuing it would be nice to have your feedback on this issue. Beside initialization of new projects an maybe the `version` command, I don't see why Mill should run without a top-level `build.sc`. This has often the potential to leave `out` directories in unwanted locations. OTOH, we have that `init` feature, which can't run without this ability.

We could make it an built-in option though. We already have `--version` to avoid evaluating `MainModule.version`.

lihaoyi:
Yeah this goes back to the "how do evaluator commands work" discussion. I'd be fine with either magic whitelisting the `init` command or baking it in as `--init`. Since we don't have an overarching principle yet, I expect things would be messy regardless of what we do, and I also expect we'll clean it up later in some new version so I don't have strong opinion on how it's implemented now

lefou:
I added a whitelist of tasks, so we only continue without a `build.sc`, if we want to execute a known well-behaving task.

PR #2691, opened 2023-08-11
Re-add Mill version marker into universal script

Flowdalic:
Mill used to invoke the JVM with -DMILL_VERSION=… as part of its universal script. This was used by millw (the Mill wrapper) to efficiently detect the version of a system-wide mill installation [1].

However, lately this changed and the universal script did not longer include the -DMILL_VERSION "marker". This re-introduces the marker so that millw can detect the version of a system-wide mill installation again.

1: https://github.com/lefou/millw/blob/e62923f3182861d2d1cc0273828f63e5d2ab6763/millw#L108-L139

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -1128,7 +1128,8 @@ def launcherScript(
     shellJvmArgs: Seq[String],
     cmdJvmArgs: Seq[String],
     shellClassPath: Agg[String],
-    cmdClassPath: Agg[String]
+    cmdClassPath: Agg[String],
+    millVersion: String,
 ) = {
 
   val millMainClass = "mill.main.client.MillClientMain"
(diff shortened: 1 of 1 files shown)


lefou (changes requested):
The removal of the `MILL_VERSION` variable fro the Mill launcher was already planed when @Flowdalic implemented the mentioned feature in `millw`. [[1]](https://github.com/lefou/millw/pull/29#pullrequestreview-873932182) I'm against re-adding it, especially in the proposed way as a comment, which seems to have no meaning, but obviously is supposed to have one.

The use of `MILL_VERSION` was removed for good reasons, so we should not rely on it. It's an option to request a specific version from the launcher, not more and not less.

We should not fix marginal and untested features of `millw` here but improve `millw` instead.

Flowdalic:
> The use of `MILL_VERSION` was removed for good reasons, so we should not rely on it. It's an option to request a specific version from the launcher, not more and not less.

I think we do not re-add it, at least not in the sense as it was used, as you describe, before.

> We should not fix marginal and untested features of `millw` here but improve `millw` instead.

More than happy to improve `millw`. :)

But how should `millw` detect the version of the system's mill installation if the file does not contain a marker with the version? The nice thing about the mill version being part of the universal script, and if its only in a comment, is that you can very efficiently and fast extract the version from it. As opposed to, for example, running `mill --version`.

Please help me understand what you request should be done instead.

lefou:
Options to improve this where discussed on Gitter.

Issue #2643, opened 2023-06-29
`CompletionException` in BSP server for `buildTarget/jvmRunEnvironment` request 


lefou:
@tgodzik wrote:
> I checked with 0.11.1 and the same exact thing seems to be happening:

```
[Trace - 04:37:26 PM] Received response 'buildTarget/jvmRunEnvironment - (16)' in 364ms
Result: null
Error: {
  "code": -32603,
  "message": "Internal error.",
[...]

PR #2692, opened 2023-08-14
Use original `Evaluator`s to preserve their association with modules in BSP to avoid using the wrong evaluator

lihaoyi:
Should fix https://github.com/com-lihaoyi/mill/issues/2643.

The basic problem is previously BSP always used the `frame(0)` evaluator passed in to `updateEvaluator`, which is meant specifically for evaluating user tasks in the root `./build.sc`. However, BSP also querying of modules and tasks both in `build.sc` as well as meta-builds in `mill-build/build.sc` and nested meta-builds. This ends up evaluating the meta-builds in the same `out/` folder as the main build, rather than in the `out/mill-build/` folders they should be evaluated in.

This PR avoids using the `Evaluator` given in `updateEvaluator`, and instead preserves the `Evaluator`s we already construct as part of `MillBuildBootstrap` in `mill.bsp.worker.State`, and passes them as a `Seq[Evaluator]` from the enclosing `MillBuildBootstrap` code all the way to the Mill-BSP code so we can associate each `Module` with its respective `Ealuator`. We then use those associated `Evaluator`s to perform the actual evaluation throughout the `Mill{Scala,Java,Jvm,}BuildServer` implementations. There's some additional re-shaping, so that in case a single BSP API call wants to query modules/tasks from multiple evaluators at once we properly group the tasks/modules together according to their respective evaluators for evaluation.

By passing the `Seq[Evaluator]` from the enclosing `MillBuildBootstrap`, this also lets us remove the now-redundant call to `MillBuildBootstrap` buried within `State.scala`. We already have all the necessary data structures from the enclosing `MillBuildBootstrap` and its evaluators, so we don't need to re-run the bootstrap process and re-construct all of it from scratch just for BSP

I haven't managed to reproduce the original error, so confirming the fix will have to wait until this lands in master and people can try it out. But I have manually tested using VSCode to load and jump around `example/basic/1-simple-scala/` and `example/scalabuilds/10-scala-realistic/`, and it seems all the references and modules work and are set up correctly. So at least we have some confidence  that this doesn't cause any regressions.

This should also supersede https://github.com/com-lihaoyi/mill/pull/2648, since we now no longer need to re-run the `MillBuildBootstrap`/`RootModuleFinder` logic at all since we just take the existing list of `Evaluator`s from the enclosing bootstrap process


bsp/worker/src/mill/bsp/worker/State.scala
```diff
@@ -1,74 +1,35 @@
...
-      imports = Nil,
-      env = Map.empty,
-      threadCount = None,
-      targetsAndParams = Seq("resolve", "_"),
-      prevRunnerState = mill.runner.RunnerState.empty,
-      logger = baseLogger,
-      disableCallgraphInvalidation = disableCallgraphInvalidation,
-      needBuildSc = true
-    ).evaluate()
+  lazy val rootModules = evaluators.map(_.rootModule)
```
lefou, line 31: Type needed

main/eval/src/mill/eval/Evaluator.scala
```diff
@@ -66,6 +66,7 @@ object Evaluator {
   // Until we migrate our CLI parsing off of Scopt (so we can pass the BaseModule
   // in directly) we are forced to pass it in via a ThreadLocal
   val currentEvaluator = new DynamicVariable[mill.eval.Evaluator](null)
+  val allEvaluators = new DynamicVariable[Seq[mill.eval.Evaluator]](null)
```
lefou, line 69: Can we base `currentEvalutor` on the value of `allEvaluators` and make it a `def`?

runner/src/mill/runner/RunnerState.scala
```diff
@@ -8,6 +8,8 @@ import mill.api.JsonFormatters._
 import mill.eval.Evaluator
 import mill.main.RootModule
 
+import scala.util.DynamicVariable
+
```
lefou, line 12: Is this unused?

lefou (commented):
This looks good to me. I added some small nits.

main/eval/src/mill/eval/Evaluator.scala
lihaoyi: We can. Not sure if we should; on one hand making it a def would be less redundant state, on the other hand the current setup is more direct. TBH since it doesn't make a difference now either way, I'm inclined to leave it unchanged for the time being

main/src/mill/main/TokenReaders.scala
```diff
@@ -38,6 +42,9 @@ object TokenReaders {
   implicit def millEvaluatorTokenReader[T]: mainargs.TokensReader[Evaluator] =
     new mill.main.EvaluatorTokenReader[T]()
 
+  implicit def millAllEvaluatorsTokenReader[T]: mainargs.TokensReader[Seq[Evaluator]] =
+    new mill.main.AllEvaluatorsTokenReader[T]()
+
```
lefou, line 47: Using a `Seq[Evaluator]` to express the wish to get all evaluators is a bit vague. A Seq can hold almost everything, also nothing. Should we introduce a dedicated type to hold all evaluators instead?

main/eval/src/mill/eval/Evaluator.scala
lefou: That's fine, too.

main/src/mill/main/TokenReaders.scala
lihaoyi: Sure, I'll wrap it in `Evaluator.AllBootstrapEvaluators(value: Seq[Evaluator])`

Issue #2658, opened 2023-07-13
Suport running tasks on a bootstrap module (`--root`)


lefou:
With the new bootstrap module concept, it would be nice to explicitly run tasks on the bootstrap module. This makes it easier to inspect a bootstrap module, but also improves the maintenance tasks (like finding dependency updates) on them.

PR #2719, opened 2023-08-29
Allow Mill CLI to select the meta-build frame it operates on

lefou:
Add a new CLI option `--meta-level` accepting an `Int`. Default is `0` and means the root project, `1` is the parent meta-build, if defined, or the built-in bootstrap module, and so on.

This is a first draft, to get more familiar with the recursive but mutable nature of our meta-build support.
Don't hesitate to point out shortcomings.

* Fixes https://github.com/com-lihaoyi/mill/issues/2658

Review by @lihaoyi

**Example: Find version updates in the meta build**

Here is some example output (applied to the mill repo):

``` 
$ mill --meta-level 1 mill.scalalib.Dependency/showUpdates
[1657/1657] dev.run 
Found 3 dependency update for 
  net.sourceforge.htmlcleaner:htmlcleaner : 2.25 -> 2.26 -> 2.27 -> 2.28 -> 2.29
  com.lihaoyi:mill-contrib-buildinfo_2.13 : 0.11.2-6-261437 -> 0.11.2
  com.github.lolgab:mill-mima_mill0.11_2.13 : 0.0.23 -> 0.0.24
```

**Meta information about the build**

I also added a new external module `mill.runner.MillBuild` to get some meta-information about the project, for now, the meta-module count or frame count.

Here on a project with one meta-build:

```
$ mill show mill.runner.MillBuild/levelCount
3
```

Pull request: https://github.com/com-lihaoyi/mill/pull/2719

### DIFF
--- a/runner/src/mill/runner/MillCliConfig.scala
+++ b/runner/src/mill/runner/MillCliConfig.scala
@@ -117,7 +117,11 @@ class MillCliConfig private (
            code changes, and instead fall back to the previous coarse-grained implementation
            relying on the script `import $file` graph"""
     )
-    val disableCallgraphInvalidation: Flag
+    val disableCallgraphInvalidation: Flag,
+    @arg(
+      doc = """Experimental: Select a meta-build to run the given targets."""
+    )
+    val frame: Option[Int]
 ) {
   override def toString: String = Seq(
     "home" -> home,
@@ -139,7 +143,8 @@ class MillCliConfig private (
     "silent" -> silent,
     "leftoverArgs" -> leftoverArgs,
     "color" -> color,
-    "disableCallgraphInvalidation" -> disableCallgraphInvalidation
+    "disableCallgraphInvalidation" -> disableCallgraphInvalidation,
+    "frame" -> frame
   ).map(p => s"${p._1}=${p._2}").mkString(getClass().getSimpleName + "(", ",", ")")
 }
 
@@ -148,8 +153,8 @@ object MillCliConfig {
    * mainargs requires us to keep this apply method in sync with the private ctr of the class.
    * mainargs is designed to work with case classes,
    * but case classes can't be evolved in a binary compatible fashion.
-   * mainargs parses the class ctr for itss internal model,
-   * but used the companion's apply to actually create an instance of the config class,
+   * mainargs parses the class ctr for its internal model,
+   * but uses the companion's apply to actually create an instance of the config class,
    * hence we need both in sync.
    */
   def apply(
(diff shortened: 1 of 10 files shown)


lihaoyi:
The approach looks reasonable. Maybe one quibble is with the naming `--frame`. "Frame" isn't a term we use anywhere else in the user-facing documentation. Maybe we can call it `--meta-build-level` or `--mill-build-level` or something similar to that, since "Meta Build" and `mill-build/` are both things that the user would likely encounter in the course of working with Mill?

lefou:
What about `--meta` or `--meta-level`?

lihaoyi:
Sure that works. But I'd err on the side of "more verbose" since we don't expect this flag to be commonly used

lefou:
@lolgab Yeah, I also want sometimes to be able to run something on all levels. This PR is a first step, to be able to request to run targets on any specific level, which was not possible before. Once, we have that working, we can add more features.

From a high level view, what you want should be rather simple to implement, as we just need to aggregate all runs of all levels. But devil is in the details. Currently, we have different evaluators for each level, and the logic to aggregate results is also in the evaluator. Also parallelization logic is in the evaluator. So, some more work and refactoring is necessary. Whereas explicitly selecting one level is a rather easy goal to achieve.

I'm even not sure if it is possible at all to run targets from all levels at the same time, as evaluation of meta-build can affect the effective module and target relations on dependent levels. So maybe, we will never be able to parallelize evaluation of different levels.

runner/src/mill/runner/MillCliConfig.scala
```diff
@@ -117,7 +117,15 @@ class MillCliConfig private (
...
-    val disableCallgraphInvalidation: Flag
+    val disableCallgraphInvalidation: Flag,
+    @arg(
+      name = "meta-level",
+      doc =
+        """Experimental: Select a meta-build level to run the given targets.
+           Level 0 is the normal project, level 1 the first meta-build, and so on.
+           The last level is the built-in synthetic meta-build which Mill uses to bootstrap the project."""
+    )
+    val metaLevel: Option[Int]
```
lolgab, line 128: Is there a reason why we don't use default values here? `val metaLevel: Int = 0` which should be supported by mainargs?

lefou: Defaults are provided in the `apply` method. 

```scala
frame: Option[Int] = None
```

Using `Option[Int]` in contrast to `Int` is more explicit whether the user provided a value or not.

lihaoyi (approved):

PR #2724, opened 2023-09-07
Add support for multi-release Jar files

Flowdalic:
If the user configures javacOptions to include Seq("--release", "17"), for example, then multi-release Jar files will be opened and processed accordingly.


lefou (commented):
I think dropping Java 8 support for such an rarely used feature is not an option.

As discussed in the [chat](https://app.element.io/#/room/#mill:matrix.am/$_fQeTIbBV0NZfqXZu05FL-miYGJtw6BpnLQcdvBD0DY), adding a `Multi-Release: true` header to the manifest of the assembly jar is enough to let the JVM recognize the incorporated multi-release jars.

Flowdalic:
Right, I am not even convinced that the approach of this PR is sensible. You probably  want the Multi-Release classes even if you build with `--release`, in case there are variants for releases newer than specified and you execution JVM supports this newer release.

Issue #2732, opened 2023-09-11
Mill doesn't evict older versions of transitive dependencies from `moduleDeps`


lolgab:
Reproduction:

```scala
// build.sc
import mill._
import mill.scalalib._

object bar extends ScalaModule {
  def scalaVersion = "2.13.12"
  def moduleDeps = Seq(foo)
  def ivyDeps = super.ivyDeps() ++ Agg(
    ivy"org.typelevel::cats-core::2.9.0",
  )
}
object foo extends ScalaModule {
  def scalaVersion = "2.13.12"
  def ivyDeps = super.ivyDeps() ++ Agg(
    ivy"io.circe::circe-core::0.14.0"
  )
}
```

classpath with Mill `0.11.2` (note the two artifacts of `cats-core_2.13`):
```json
> mill show bar.compileClasspath
[
  "ref:v0:c984eca8:/Users/lorenzo/scala/repro/foo/compile-resources",
  "qref:v1:19c63377:/Users/lorenzo/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/io/circe/circe-core_2.13/0.14.0/circe-core_2.13-0.14.0.jar",
  "qref:v1:0004e741:/Users/lorenzo/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.13.12/scala-library-2.13.12.jar",
  "qref:v1:03c52e9e:/Users/lorenzo/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/io/circe/circe-numbers_2.13/0.14.0/circe-numbers_2.13-0.14.0.jar",
  "qref:v1:59fac2bc:/Users/lorenzo/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/typelevel/cats-core_2.13/2.6.1/cats-core_2.13-2.6.1.jar",
  "qref:v1:b856e496:/Users/lorenzo/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/typelevel/cats-kernel_2.13/2.6.1/cats-kernel_2.13-2.6.1.jar",
  "qref:v1:cfb647ca:/Users/lorenzo/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/typelevel/simulacrum-scalafix-annotations_2.13/0.5.4/simulacrum-scalafix-annotations_2.13-0.5.4.jar",
[...]

PR #2735, opened 2023-09-11
Fix conflicting dependencies between upstream `JavaModule`s

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/2732

We introduce a new `localCompileClasspath`, which differs from `compileClasspath` in that it leaves out the `transitiveCompileClasspath` from upstream modules and `resolvedIvyDeps` from third party dependencies.

Also tried to do some cleanup of `JavaModule`, refactoring out a `transitiveModuleCompileModuleDeps` helper to DRY things up, re-arranging things so all the `fooModuleDeps` helpers are all together.

Tested via a new unit test `mill.scalalib.HelloWorldTests.multiModuleClasspaths`, which fails on master and passes on this PR. This test case asserts the overall "shape" of the classpaths, which should help rule out an entire class of mis-configuration w.r.t. how the classpaths are wired up

### DIFF
--- a/scalalib/src/mill/scalalib/JavaModule.scala
+++ b/scalalib/src/mill/scalalib/JavaModule.scala
@@ -413,7 +413,7 @@ trait JavaModule
    * assembly, but without this module's contribution
    */
   def upstreamAssemblyClasspath: T[Agg[PathRef]] = T {
-    resolvedRunIvyDeps() ++ transitiveLocalClasspath() ++ unmanagedClasspath()
+    resolvedRunIvyDeps() ++ transitiveLocalClasspath()
   }
 
   def resolvedRunIvyDeps: T[Agg[PathRef]] = T {
--- a/scalalib/src/mill/scalalib/JavaModule.scala
+++ b/scalalib/src/mill/scalalib/JavaModule.scala
@@ -427,7 +427,7 @@ trait JavaModule
    * necessary to run this module's code after compilation
    */
   def runClasspath: T[Seq[PathRef]] = T {
-    resolvedRunIvyDeps() ++ transitiveLocalClasspath() ++ localClasspath()
+    resolvedRunIvyDeps().toSeq ++ transitiveLocalClasspath() ++ localClasspath()
   }
 
   /**
--- a/scalalib/src/mill/scalalib/JavaModule.scala
+++ b/scalalib/src/mill/scalalib/JavaModule.scala
@@ -369,7 +369,7 @@ trait JavaModule
   @internal
   def bspLocalClasspath: T[Agg[UnresolvedPath]] = T {
     (compileResources() ++ resources()).map(p => UnresolvedPath.ResolvedPath(p.path)) ++
-    Agg(bspCompileClassesPath())
+      Agg(bspCompileClassesPath())
   }
 
   /**
@@ -402,7 +402,7 @@ trait JavaModule
    * Resolved dependencies based on [[transitiveIvyDeps]] and [[transitiveCompileIvyDeps]].
    */
   def resolvedIvyDeps: T[Agg[PathRef]] = T {
-    resolveDeps(T.task {transitiveCompileIvyDeps() ++ transitiveIvyDeps()})()
+    resolveDeps(T.task { transitiveCompileIvyDeps() ++ transitiveIvyDeps() })()
   }
 
   /**
@@ -414,7 +414,7 @@ trait JavaModule
   }
 
   def resolvedRunIvyDeps: T[Agg[PathRef]] = T {
-    resolveDeps(T.task {runIvyDeps().map(bindDependency()) ++ transitiveIvyDeps()})()
+    resolveDeps(T.task { runIvyDeps().map(bindDependency()) ++ transitiveIvyDeps() })()
   }
 
   /**
--- a/scalalib/src/mill/scalalib/JavaModule.scala
+++ b/scalalib/src/mill/scalalib/JavaModule.scala
@@ -154,9 +154,8 @@ trait JavaModule
   /** The compile-only transitive ivy dependencies of this module and all it's upstream compile-only modules. */
   def transitiveCompileIvyDeps: T[Agg[BoundDep]] = T {
     // We never include compile-only dependencies transitively, but we must include normal transitive dependencies!
-    compileIvyDeps().map(bindDependency()) ++ T
-      .traverse(compileModuleDeps)(_.transitiveIvyDeps)()
-      .flatten
+    compileIvyDeps().map(bindDependency()) ++
+      T.traverse(compileModuleDeps)(_.transitiveIvyDeps)().flatten
   }
 
   /**
@@ -380,7 +379,7 @@ trait JavaModule
    */
   // Keep in sync with [[bspCompileClasspath]]
   def compileClasspath: T[Agg[PathRef]] = T {
-    transitiveCompileClasspath() ++ localCompileClasspath() ++ resolvedIvyDeps()
+    resolvedIvyDeps() ++ transitiveCompileClasspath() ++ localCompileClasspath()
   }
 
   /**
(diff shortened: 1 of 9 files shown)


scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -424,8 +427,7 @@ trait JavaModule
    * necessary to run this module's code after compilation
    */
   def runClasspath: T[Seq[PathRef]] = T {
-    localClasspath() ++
-      upstreamAssemblyClasspath()
+    resolvedRunIvyDeps().toSeq ++ transitiveLocalClasspath() ++ localClasspath()
```
lihaoyi, line 430: We remove `upstreamAssemblyClasspath` so we can remove the dependency on `unmanagedClasspath`, since it now also comes from `localClasspath`, and we want to avoid duplicate classpath entries

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -355,11 +355,11 @@ trait JavaModule
...
 
   /**
-   * The output classfiles/resources from this module, excluding upstream
-   * modules and third-party dependencies
+   * The *output* classfiles/resources from this module, used for execution,
+   * excluding upstream modules and third-party dependencies
    */
   def localClasspath: T[Seq[PathRef]] = T {
-    compileResources() ++ resources() ++ Agg(compile().classes)
+    localCompileClasspath().toSeq ++ resources() ++ Agg(compile().classes)
```
lihaoyi, line 362: This change adds `unmanagedClasspath` to `localClasspath`. This is probably the right thing to do, since it means it'll probably be aggregated from upstream modules into `upstreamAssemblyClasspath`, where it wouldn't before

scalalib/test/src/mill/scalalib/HelloWorldTests.scala
```diff
@@ -1341,5 +1395,196 @@ object HelloWorldTests extends TestSuite {
...
+            //
+            "MultiModuleClasspaths/ModCompile/qux/compile-resources",
+            "MultiModuleClasspaths/ModCompile/qux/unmanaged",
+            "MultiModuleClasspaths/ModCompile/qux/resources",
+            "multiModuleClasspaths/modCompile/ModCompile/qux/compile.dest/classes"
+          ),
+          expectedCompileClasspath = List(
+            "com/lihaoyi/geny_2.13/0.4.0/geny_2.13-0.4.0.jar",
+            // We aggregate upstream module `compileIvyDeps` during compilation
+            "com/lihaoyi/sourcecode_2.13/0.2.2/sourcecode_2.13-0.2.2.jar",
```
lihaoyi, line 1523: I'm not sure if this is correct: should `compileIvyDeps` be transitive at all? But it is the current behavior

lefou: `compileIvyDeps` should not be transitive. They only are for local compilation. Downstream consumers need to provide them themselves if needed.

lihaoyi: It turns out that I was mistaken, and this was a normal `ivyDeps` but in a `compileModuleDeps`.

@lefou it seems you left a comment in here stating that this is necessary. Do you remember the reasoning?

https://github.com/com-lihaoyi/mill/pull/955/files#diff-85ad27993fcc36e23f993baae04e41cb1bfa900502fa0622197d964abca08ddcR103

I'm finding that without propagating `ivyDeps` from the `compileModuleDeps`, a whole bunch of Mill's build stops compiling because e.g. 

1. `A` ends up compiling against module `B` with ivy dependency `C` on it's classpath
2. But if `B` has a method that returns a type from `C`, then `A` will cause the compiler to crash since it needs to have `C` on the classpath to look up and compile against

It seems to me that this should apply equally for `ivyDeps` and `compileIvyDeps`. Currently, a module's `compileClasspath` only picks up `ivyDeps` from it's `compileModuleDeps`, but does not pick up it's `compileIvyDeps`. This has likely not been a problem so far only because `compileIvyDeps` is so uncommonly used.

Given that `ivyDeps` and `compileIvyDeps` have the same problem, should they be inherited by `compileClasspath` in the same way?

lefou: Ha, I asked myself the same question, when I tried to fix this issue. Unfortunately I don't remember exactly. The timing for me to deep dive into this issue is a bit unlucky. I hope I can later review and think about it. (I also pushed my WIP from last night, just to see the CI results.)

I think we definitely want the handle `compileIvyDeps` and `compileModuleDeps` equally, as the latter is converted to the former once published, and Mill behave the same, disregarding the compile-dependency comes from the local build or from a coursier dependency.

lihaoyi: sure, this isn't super urgent, so I'll leave it open until you have a chance to think about it

lihaoyi: I think feel like the best thing to do here is to have all compile dependencies and run dependencies be transitive, but limited to their respective classpaths.

Otherwise, there'll be no end to problems where ModuleA compiling against an upstream ModuleB fails due to ModuleB compiling against ModuleC and exposing it in a type signature. Similarly, for run dependencies, it makes sense that an application that needs e.g. log4j at runtime will continue to need log4j at runtime if used as a library for another application.

It's possible to work around these issues in user land, but it would basically always come down to cherry picking arbitrary dependencies from your upstream modules until things work, or manually traversing the module graph and collecting everything. The former is fragile and the latter is something that it feels like we should automate.

lefou: There are various out-side givens to incorporate. E.g. we consume dependencies via coursier, so we can't change how transitivity is handled by it, hence, we are forced to adapt our inter-module dependencies handling likewise.

Also, Maven-like `provided` dependencies are useful for compile-time-only APIs, like macros (like `acyclic`?), which should not leak transitively. Or to ensure you to against some older baseline APIs. You don't want to force those old versions transitively, but you want to force your local compilation to use that old version to ensure, you don't accidentally use newer stuff.

lihaoyi: Good point. We can't change the maven pom scopes, and those have no way to express "transitive compile time dependency". That means that no matter what we do with `moduleDeps`, there's no way to have a transitive compile time dependency published to maven central

I guess in that case, it sounds like we follow the current status quo, where we pick up the transitive `*Deps` of `compile*Deps`, but not the transitive `compile*Deps`.

lihaoyi:
I'm going to merge this as is for now: with `compileIvyDeps` and `compileModuleDeps` not transitive, but `ivyDeps` and `moduleDeps` being transitive, even if depended on from a `compileModuleDeps`, though in that case only from the `compileClasspath`. 

That should mostly match how `provided` works in SBT: adding a `provided` dependency in SBT pulls in all its transitive non-provided dependencies as well. One difference is that this applies to the `fullClasspath`, because SBT doesn't have a distinction between compile and run classpaths. This should also mostly match the current behavior, at least before the regression described in https://github.com/com-lihaoyi/mill/issues/2732

lefou (commented):
This looks reasonable to me. It's a bit hard to review in the browser, due to the various other (unrelated) refactings. 

I found one potential discrepancy between `transitiveCompileClasspath` and `bspTransitiveCompileClasspath`. To former now used `localCompileClasspath` whereas the latter still uses `bspCompileClasspath` which should be in sync with `compileClasspath`.
But if I look closer, `bspCompileClasspath` is also using `localCompileClasspath`. It could be right, but hard to tell in the browser.

Also, this PR changes the order of dependencies and puts all elements from `ivy` sources before others. This is a behavioral change. Was this intentional?

lefou:
Just for completeness, I'd like to mention, there is also PR #2736, which also fixes issue #2732 and is more minimal. Just in case we later find this approach too invasive in behavior.

lihaoyi:
> Also, this PR changes the order of dependencies and puts all elements from ivy sources before others. This is a behavioral change. Was this intentional?

Somewhat. The status quo was broken, so of course we'd need to change things, so I while changing things I thought we might as well put things in a consistent order. With this PR, the order is always ivy deps, then upstream modules, then the current module. IIRC the previous orderings weren't as consistent between the various `*ClassPaths`

lihaoyi:
Classpath ordering is definitely a risk of breakage, it happens when a user has multiple different files at the same path on the classpath in different jars or classfile folders.

But I don't think that's something we should guarantee too strongly. Even listing things on the filesystem can vary in order, so if a user deploys jars in a folder and does `java -cp *` classpath ordering can break things if deployed on different machines or even different folders on the same machine (we've seen bugs like this before at work). So in some ways code that depends on classpath ordering is already broken

PR #2738, opened 2023-09-14
Update zinc to 1.9.5

lefou:


lefou:
Changed to draft state, as sbt 1.9.5 (which also includes zinc 1.9.4) has compile issues due to PR https://github.com/sbt/zinc/pull/1244

lefou:
I update zinc to 1.9.5 to avoid 1.9.4.

PR #2739, opened 2023-09-14
Fix the scala-library dependency for (generic) platform modules

lefou:
In Scala 3 modules, that also define a platform, the scala library is not correctly resolved.
The current implementation exists merely as a convenience for JS and Native modules. `ScalaNativeModule` already handles a correct scala library by overriding `scalaLibraryIvyDeps`. This pull request overrides the same in `ScalaJSModule` and removed the faulty platfrom dependency from `ScalaModule`.

E.g. imagine, someone tries to build a Mill plugin with Scala 3 (which is currently not supported) and sets the platform to `_mill0.11`, then you would run into the following issue:

```scala
object root extends ScalaModule {
  override def scalaVersion = "3.3.0"
  override def platformSuffix = "_mill0.11"
  // ..
}
```

```
>> mill ivyDepsTree
1 targets failed
resolvedIvyDeps
  org.scala-lang:scala3-library_mill0.11_3:3.3.0
	not found: /Users/me/ivy2/local/org.scala-lang/scala3-library_mill0.11_3/3.3.0/ivys/ivy.xml
	not found: https://repo1.maven.org/maven2/org/scala-lang/scala3-library_mill0.11_3/3.3.0/scala3-library_mill0.11_3-3.3.0.pom
```

* See also https://github.com/com-lihaoyi/mill/pull/1187#discussion_r1326540338

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -364,14 +364,16 @@ trait MillPublishScalaModule extends MillScalaModule with MillPublishJavaModule
 /** Publishable module which contains strictly handled API. */
 trait MillStableScalaModule extends MillPublishScalaModule with Mima {
   import com.github.lolgab.mill.mima._
-  // MIMA doesn't properly ignore things which are nested inside other private things
-  // so we have to put explicit ignores here (https://github.com/lightbend/mima/issues/771)
   override def mimaBinaryIssueFilters: T[Seq[ProblemFilter]] = Seq(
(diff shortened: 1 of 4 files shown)


lolgab (approved):
Thank you for the fix! ❤️

lolgab (approved):

PR #2752, opened 2023-09-18
Fix sneaky binary incompatibility in Discover.scala

lihaoyi:
fixes #2749

The basic problem is that the signature of `Discover.apply()` changed, but it only changed in the argument's generic type parameters, so it was nominally the same type after erasure, and accepted the `Map[Class[_], Seq]` even though it was expecting a `Map[Class[_], (Seq, Seq)]`, but later on in the code it would blow up with a `ClassCastException`

This causes problems with Mill plugins containing external modules, which have their own `Discover[T]` macro pre-expanded and would not get re-compiled with the new 0.11.3 version of Mill.

This PR works around the problem by adding a forwarder `Discover.apply(value: Map[Class, Seq])` that does the right thing and expands the `Seq` into a `(Seq, Seq)`, while adding a dummy paramter to `case class Discover` to avoid conflicts. A similar forwarder `Discover.<init>(value: Map[Class, Seq])` is needed to preserve binary compatibility (I think?)

Tested manually via `./mill -i dev.run example/basic/1-simple-scala --import ivy:io.chris-kipp::mill-github-dependency-graph::0.2.6 io.kipp.mill.github.dependency.graph.Graph/generate`, which fails before this PR and passes after

PR #2755, opened 2023-09-18
Implement `runBackgroundInherit` / `runMainBackgroundInherit`

swaldman:
For what it's worth, here's a sketch of one possible way to address where `runBackground` process output should go.

I'm not at all wedded to it! Mostly, since I'm coming at you with asks, I want to do my part to help. For example, @lolgab's suggestion of overridable

```scala
def runBackgroundStdout: T[ProcessOutput] = os.Pipe
def runBackgroundStderr: T[ProcessOutput] = os.Pipe
```

strikes me as perhaps more elegant and flexible than duplicating `runBackground` commands, which I've done here.

In general, this pull request

1. factors out the common implementations of `JavaModule.runBackground` and `JavaModule.runMainBackground` into a helper method.

2. creates variants of `Jvm.spawnSubprocess` and `Jvm.runSubprocess` that permit specifying `out` and `err` via an optional `Tuple2` of `ProcessOutput` (and reimplements the existing methods, which offer a `Boolean` flag `background`, in terms of these methods)

3. modifies the common `doRunBackground` implementation to accept that optional tuple of `ProcessOutput`

4. modifies `JavaModule.runBackground` and `JavaModule.runMainBackground` to call the new methods, **using `T.dest/stdout.log` and `T.dest/stderr.log`** as outputs

5. adds two new commands `JavaModule.runBackgroundInherit` and `JavaModule.runMainBackgroundInherit` which sets both `out` and `err` to `os.Inherit`

Please note the bit in bold. In general, everything old is preserved. No method signatures are removed or changed, there are only additions and reimplementations in terms of those additions. But, following the discussion on discord, I did make one change to the current behavior, dropping the default logs in `runBackground.dest` / `runMainBackground.dest` rather than in the project root directory.

I've had to bring in https://github.com/com-lihaoyi/mill/pull/2752 so that I could give this a try.  It seems like it works!

I apologize for the amateurishness. I feel like I'm groping my way around in `mill`. It was very helpful to learn from @lihaoyi on the discord that `./mill -i installLocal` generates a binary one can play with in `target/`

Thank you all for your work. I love this tool.

### DIFF
--- a/main/api/src/mill/api/JsonFormatters.scala
+++ b/main/api/src/mill/api/JsonFormatters.scala
@@ -17,6 +17,12 @@ trait JsonFormatters {
       os.Path(_)
     )
 
+  implicit val filePathReadWrite: RW[os.FilePath] = upickle.default.readwriter[String]
+    .bimap[os.FilePath](
+      _.toString,
+      os.FilePath(_)
+    )
+
   implicit val regexReadWrite: RW[Regex] = upickle.default.readwriter[String]
     .bimap[Regex](
       _.pattern.toString,
--- a/main/api/src/mill/api/JsonFormatters.scala
+++ b/main/api/src/mill/api/JsonFormatters.scala
@@ -17,6 +17,18 @@ trait JsonFormatters {
       os.Path(_)
     )
 
+  implicit val relPathReadWrite: RW[os.RelPath] = upickle.default.readwriter[String]
+    .bimap[os.RelPath](
+      _.toString,
+      os.RelPath(_)
+    )
+
+  implicit val subPathReadWrite: RW[os.SubPath] = upickle.default.readwriter[String]
+    .bimap[os.SubPath](
+      _.toString,
+      os.SubPath(_)
+    )
+
   implicit val filePathReadWrite: RW[os.FilePath] = upickle.default.readwriter[String]
     .bimap[os.FilePath](
       _.toString,
--- a/main/util/src/mill/util/Jvm.scala
+++ b/main/util/src/mill/util/Jvm.scala
@@ -3,7 +3,7 @@ package mill.util
 import mill.api.Loose.Agg
 import mill.api._
 import mill.main.client.InputPumper
-import os.{ProcessOutput,SubProcess}
+import os.{ProcessOutput, SubProcess}
 
 import java.io._
 import java.lang.reflect.Modifier
@@ -56,8 +56,8 @@ object Jvm extends CoursierSupport {
 
   def javaExe: String = jdkTool("java")
 
-  def defaultBackgroundOutputs( outputDir : os.Path ) : Option[(ProcessOutput,ProcessOutput)]
-    = Some( (outputDir / "stdout.log", outputDir / "stderr.log") )
+  def defaultBackgroundOutputs(outputDir: os.Path): Option[(ProcessOutput, ProcessOutput)] =
+    Some((outputDir / "stdout.log", outputDir / "stderr.log"))
 
   /**
    * Runs a JVM subprocess with the given configuration and streams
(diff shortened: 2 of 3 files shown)


lihaoyi:
@swaldman thanks for the PR!

I think the suggestion of making it a flag on the `JavaModule`, rather than duplicating the `runBackground` commands, is the right thing to do. We might not even need `def runBackgroundStdout: T[ProcessOutput]`, just a simple `def runBackgroundLogToDisk: Boolean` should be enough I think. If anyone wants a more advanced use case, they can always call the helper methods themselves, so we don't need to complicate the public API for the common simple use case.

swaldman:
@lihaoyi I agree that sounds better! Proliferating a menagerie of `runBackgroundXXX` tasks seems inelegant.

I've pushed a new attempt, which...

1. Eliminates `runBackgroundInherit` / `runMainBackgroundInherit`.
2. Changes the visibility of the run-background-task-implementing helper method `doRunBackground(...)` from `private` to `protected`, so that modules can use that helper method to roll their own, using any choice of `os.ProcessOutput`.
3. Implements, per your suggestion, overridable `runBackgroundLogPath` tasks that configure the built-in `runBackground` tasks. 

    However, I've modified your suggestion in two ways:
    1. Instead of a single `runBackgroundLogPath` task, I've bifurcated into 
        * `runBackgroundLogOutPath` 
        * `runBackgroundLogErrPath`
        
        on the theory that sending output and error to different places is not such an unusual case, 
        so maybe should be supported.
    2. Instead of returning `T[Option[os.Path]]`, they return `T[Option[os.FilePath]]`

The reason for that last change was an annoyance defining paths within `runBackgroundLogOutPath` / `runBackgroundLogErrPath`.

A common intention might be to log these streams to files in e.g. `out/foo/runBackground.dest`. But the usual source of an `os.Path` within `runBackgroundLogOutPath` -- `T.dest` -- would place files into `out/foo/runBackgroundLogOutPath.dest`, which seems less desirable.

So the current implementation
1. sends a stream to `os.Inherit` by default -- usually the `mill` console -- when the returned log path is `None`.
2. sends the stream to whatever absolute path is specified, if `Some[os.Path]` is returned
4. if `Some[os.RelPath]` or `Some[os.SubPath]` is returned, those are interpreted 
    _relative to the destination director of the `runBackground` or `runMainBackground` task._

So if `Some(os.sub/"out.txt")` is returned by `runBackgroundLogOutPath`, and `runBackground` is executed, stdout ends up in `out/foo/runBackground.dest/out.txt`. This is probably where you want it, rather than `Some(T.dest/"out.txt")`.

This implementation does everything one might want, I think, but it's a judgment call whether it's worth its complexity. Will users understand this well enough to use it, or will it just not be a big deal to return `Some(T.dest/"out.txt")` and find stdout in `out/foo/runBackgroundLogOutPath.dest/out.txt`?

Implementing this required adding to `JsonFormatters` codecs for `os.RelPath`, `os.SubPath`, and `os.FilePath`.
I think it's all okay, the codecs are all mutually compatible in the sense that a thing pickled as an `os.Path` then
unpickled as an `os.FilePath` will result in the expected `os.Path`, etc. (`os.RelPath` might sometimes canonicalize
to `os.SubPath`, but only where `os.FilePath` or `os.SubPath` would be acceptable.)

Nevertheless it's maybe more uncertainty than my very niche concern over `runBackground` logging merits.

We could certainly fallback to returning `os.Path`, and do without the new `ReadWriter` implicits.

Alternatively, perhaps most parsimoniously, we could forget about the overridable configuration methods entirely, choose whatever as the default behavior for `runBackground` / `runMainBackground`, and just leave the `doRunBackground(...)` helper method exposed for use by submodules. People with a preference for where the streams go can then pretty trivially just implement their own run background tasks that behave however they want.

Sorry to go on for so long!

As always, the intent is to contribute back, not create headaches. I'm not offended if this is all not worth the cognitive load. Thanks again for everything!

swaldman:
This PR has grown IMHO too complicated and unwieldy. Replaced by https://github.com/com-lihaoyi/mill/pull/2792

PR #2759, opened 2023-09-19
Update Scala to 2.13.12

lefou:


lefou:
Ah, scoverage reporting needs to run in the Mill classpath, hence it must be resolved with the Scala version used to run Mill.

PR #2768, opened 2023-09-19
Fix some hardcoded scala versions

lefou:
The Scala version used in `MillBuildBootstrapModule` to compile the `build.sc` was hardcoded to `2.13.10`. Instead, we want to always use the same version we used to build that Mill version.

As an side effect, Mill is now compatible with new Java 21, due to the use of a newer Scala version.


lefou:
Fix #2769

PR #2772, opened 2023-09-20
Support Play Framework 3.0 and 2.9

lolgab:
- Update Play 2.8 to 2.8.20
- Add default `twirlVersion` for play based on `playVersion`
- Add default `twirlScalaVersion` based on `twirlVersion`

Pull request: https://github.com/com-lihaoyi/mill/pull/2772


lefou:
Nice! Does this PR supersedes #2517? A test using Scala 3, which is supported since Play 2.9, would be nice, but if you don't have the bandwidth to add it, we can add it later.

lefou (approved):
Looks good to me. Thank you!

PR #2773, opened 2023-09-20
Add and implement new `--onlydeps` option

lefou:


main/api/src/mill/api/MillException.scala
```diff
@@ -0,0 +1,7 @@
+package mill.api
+
+/**
+ * This exception is specifically handling in [[mill.runner.MillMain]]. You can use it, if you need to exit Mill with a nice error message.
```
lefou, line 4:
```suggestion
 * This exception is specifically handled in [[mill.runner.MillMain]] and [[mill.runner.MillServerMain]]. You can use it, if you need to exit Mill with a nice error message.
```

sequencer:
I looked around and saw this PR. 
I wonder if it would be possible to provide a tool from mill to expose all ivy as json as the lock file. Later we can use this lockfile to download it w/ coursier to prepare the build artifacts for no network building. 
A dirty hack has been achieved in https://github.com/chipsalliance/t1/blob/master/nix/pkgs/mill-builder.nix
I wonder if there are some clean way to use. E.g. this PR.

lefou:
> I wonder if it would be possible to provide a tool from mill to expose all ivy as json as the lock file. 

It certainly is possible, but it looks unrelated to this PR. `--onlydeps` in this context means: "build only the target dependencies".

@sequencer Please open a dedicated discussion for your topic, thanks.

PR #2782, opened 2023-09-23
Add note about gpg --pinentry-mode=loopback argument to documentation

swaldman:
A few years ago how to handle `gpg`'s changing requirements were extensively discussed here, and very reasonably resolved. See https://github.com/com-lihaoyi/mill/pull/753

Nevertheless, having just upgraded from a very ancient `gpg`, I hit this issue and it took a small storm of googling to figure it out. Maybe it's worth adding a note to the documentation?


lefou (approved):
Looks good to me. Thank you!

PR #2785, opened 2023-09-26
Use `quick = true` when calculating PathRef for Mill executable

lolgab:
The Mill executable is a static file that is downloaded and doesn't change. Changing the `PathRef` calculation to `quick = true` makes it invalidate the cache everytime the mill executable changes.
Before it took ~160ms, after it takes ~100μs ~100 times faster. This code runs twice on every Mill command line task execution.
This also stores the `PathRef`s in memory so they don't need to be computed on every command invocation

Pull request: https://github.com/com-lihaoyi/mill/pull/2785


lefou (approved):
Looks reasonable. Thanks!

PR #2792, opened 2023-09-28
Simple control over runBackground stdout/stderr streams

swaldman:
Replaces https://github.com/com-lihaoyi/mill/pull/2755, which has grown complicated and unwieldy for the very limited benefit it might provide.

Allow stdout/stderr outputs for the built-in `runBackground` tasks to be controlled by an overridable

```scala
  def runBackgroundLogToConsole: Boolean = true
```

If true, output goes to `os.Inherit`, usually the mill console. If false, output goes to `runBackground.dest/stdout.log` (+ `stderr.log`).

People who prefer elsewheres for these streams can implement their own `runBackground` tasks in terms of `doRunBackground(...)`. (Previously this was difficult, because the `mill.util.Jvm` run/spawn methods did not permit specifying stream destinations, and because the `backgroundSetup(...)` method is private.)

Pull request: https://github.com/com-lihaoyi/mill/pull/2792

### DIFF
--- a/main/util/src/mill/util/Jvm.scala
+++ b/main/util/src/mill/util/Jvm.scala
@@ -3,7 +3,7 @@ package mill.util
 import mill.api.Loose.Agg
 import mill.api._
 import mill.main.client.InputPumper
-import os.SubProcess
+import os.{ProcessOutput, SubProcess}
 
 import java.io._
 import java.lang.reflect.Modifier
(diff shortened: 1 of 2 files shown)


lihaoyi (approved):
This looks fine to me I think. Can you run the autoformatter `./mill mill.scalalib.scalafmt.ScalafmtModule/reformatAll __.sources` and we can merge it

swaldman:
@lihaoyi Gack! Sorry about that... Done.

lefou (approved):
Thank you!

PR #2800, opened 2023-09-30
protobuf module: fix truncation issue when unpacking

jodersky:
It is possible that unpacked files from the classpath share the same name but not the same content. In case an existing unpacked file is replaced by a shorter file, it needs to be properly truncated lest the result be garbage.

A warning messages is logged when a proto file gets overwritten.

Pull request: https://github.com/com-lihaoyi/mill/pull/2800

### DIFF
--- a/contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBModule.scala
+++ b/contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBModule.scala
@@ -104,7 +104,7 @@ trait ScalaPBModule extends ScalaModule {
             case Some(entry) =>
               if (entry.getName.endsWith(".proto")) {
                 val protoDest = dest / os.SubPath(entry.getName)
-                Using(os.write.outputStream(protoDest, createFolders = true))(IO.stream(zip, _))
+                Using(os.write.over.outputStream(protoDest, createFolders = true))(IO.stream(zip, _))
               }
               zip.closeEntry()
               true
--- a/contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBModule.scala
+++ b/contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBModule.scala
@@ -104,10 +104,9 @@ trait ScalaPBModule extends ScalaModule {
             case Some(entry) =>
               if (entry.getName.endsWith(".proto")) {
                 val protoDest = dest / os.SubPath(entry.getName)
-                Using(os.write.over.outputStream(protoDest, createFolders = true))(IO.stream(
-                  zip,
-                  _
-                ))
+                Using.resource(os.write.over.outputStream(protoDest, createFolders = true)) { os =>
+                  IO.stream(zip, os)
+                }```
               }
               zip.closeEntry()
               true
--- a/contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBModule.scala
+++ b/contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBModule.scala
@@ -104,7 +104,10 @@ trait ScalaPBModule extends ScalaModule {
             case Some(entry) =>
               if (entry.getName.endsWith(".proto")) {
                 val protoDest = dest / os.SubPath(entry.getName)
-                Using(os.write.over.outputStream(protoDest, createFolders = true))(IO.stream(zip, _))
+                Using(os.write.over.outputStream(protoDest, createFolders = true))(IO.stream(
+                  zip,
+                  _
+                ))
               }
               zip.closeEntry()
               true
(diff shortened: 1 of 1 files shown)


contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBModule.scala
```diff
@@ -104,7 +104,10 @@ trait ScalaPBModule extends ScalaModule {
             case Some(entry) =>
               if (entry.getName.endsWith(".proto")) {
                 val protoDest = dest / os.SubPath(entry.getName)
-                Using(os.write.outputStream(protoDest, createFolders = true))(IO.stream(zip, _))
+                Using(os.write.over.outputStream(protoDest, createFolders = true))(IO.stream(
+                  zip,
+                  _
+                ))
```
lefou, line 110: Two issues here: first the formatting was ugly ;) , second, `Using.apply` returns a `Try` which may hide any issues, hence I added the `resource` which will let bubble up any exception that may occur.

```suggestion
                Using.resource(os.write.over.outputStream(protoDest, createFolders = true)) { os =>
                  IO.stream(zip, os)
                }```

Comment (jodersky) [reply contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBModule.scala:110]: My original commit basically only adds a `.over` after the `os..write` call (https://github.com/com-lihaoyi/mill/pull/2800/commits/77be6cd0f8f4267039b094cae820af6a3f051e40). I ran scalafmt and blindly committed after seeing CI fail, but I agree that the format is ugly.

File: contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBModule.scala [hunk 2]
```diff
@@ -104,7 +104,9 @@ trait ScalaPBModule extends ScalaModule {
             case Some(entry) =>
               if (entry.getName.endsWith(".proto")) {
                 val protoDest = dest / os.SubPath(entry.getName)
-                Using(os.write.outputStream(protoDest, createFolders = true))(IO.stream(zip, _))
+                Using.resource(os.write.over.outputStream(protoDest, createFolders = true)) { os =>
+                  IO.stream(zip, os)
+                }```
```
lefou, line 109:
```suggestion
                }
```

lefou (commented):
Is this even the correct behavior? Having two files with the same path in the classpath is always a call for trouble, as it may fail when the classpath entries have different order. Is this a runtime issue or a compile time issue you try to fix here? It would be nice to have a test case for that, so we make sure, we don't loose it with the next refactoring.

jodersky:
I agree that having files with the same path on the classpath is most often a sign of trouble, but I would argue that the fix here is correct, because it keeps the same behavior of the current implementation but prevents a certain failure. I don't think that dealing with classpath ordering should be the responsibility if this utility function.

The way I came across this error was when cross-building scala JVM and JS versions of a library that needed some common protocol buffer files. Through some transitive dependencies, it turned out that a specific file of the same name was unpacked twice, but once from a version made for Scala JS and once for a version made for Scala JVM. Unfortunately the files did not have the same content, which led to an unreadable final protobuf file because of said bug. Now, the best way to deal with this would be to inspect the dependency chain and submit patches to upstream libraries that cause this inclusion, but again I think that this way should not exclude relying on implicit ordering.

lefou:
Yeah, that sounds reasonable. Yet, once we identified a potential source of trouble, it would be nice to make it easier for the developer who runs into it next. Your fix ensures that, disregarding which file we choose, we at least keep it in a valid state. What do you think of an additional check if the file already exists, and print a warning in addition to overwriting the existing file? Easy enough for us, non-disruptive DX as we don't fail, and some hint which helps in case the result isn't the expected one.

contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBModule.scala
```diff
@@ -104,7 +104,9 @@ trait ScalaPBModule extends ScalaModule {
             case Some(entry) =>
               if (entry.getName.endsWith(".proto")) {
                 val protoDest = dest / os.SubPath(entry.getName)
-                Using(os.write.outputStream(protoDest, createFolders = true))(IO.stream(zip, _))
+                Using.resource(os.write.over.outputStream(protoDest, createFolders = true)) { os =>
```
lefou, line 107:
```suggestion
                if (os.exists(protoDest))
                  T.log.error(s"Warning: Overwriting ${dest} / ${os.SubPath(entry.getName)} ...")
                Using.resource(os.write.over.outputStream(protoDest, createFolders = true)) { os =>
```

lefou:
Thank you!

jodersky:
Thanks for the fix! Your rationale makes sense, I agree that a warning here is useful

PR #2803, opened 2023-09-30
[BSP] Use consistent file URI format

lolgab:
Previously we were emitting different URI formats if the sourceFile info was missing or not.
Not sure if this is a problem, but `targetId.getUri` is returning files as `file:///foo/bar/file.scala` while our `java.io.File` `toURI` returns `file:/foo/bar/file.scala`
Changing it to `.toPath.toUri` for consistency

### DIFF
--- a/bsp/worker/src/mill/bsp/worker/BspCompileProblemReporter.scala
+++ b/bsp/worker/src/mill/bsp/worker/BspCompileProblemReporter.scala
@@ -69,9 +69,9 @@ private class BspCompileProblemReporter(
     val diagnostic = getSingleDiagnostic(problem)
     val sourceFile = problem.position.sourceFile
     val textDocument = new TextDocumentIdentifier(
-      sourceFile.getOrElse(None) match {
+      sourceFile match {
         case None => targetId.getUri
-        case f: File => f.toURI.toString
+        case Some(f) => f.toPath.toUri.toString
       }
     )
     val params = new bsp.PublishDiagnosticsParams(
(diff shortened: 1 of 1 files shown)


lefou (commented):
So, we now always have `file:///` for local file URIs?

I know there were other issues (Exceptions) when converting files to URIs, but I can't remember right now. Using `toPath` looks ok, but we should definitely leave a comment in the code indicating it's purpose.

bsp/worker/src/mill/bsp/worker/BspCompileProblemReporter.scala
```diff
@@ -69,9 +69,9 @@ private class BspCompileProblemReporter(
     val diagnostic = getSingleDiagnostic(problem)
     val sourceFile = problem.position.sourceFile
     val textDocument = new TextDocumentIdentifier(
-      sourceFile.getOrElse(None) match {
+      sourceFile match {
         case None => targetId.getUri
-        case f: File => f.toURI.toString
+        case Some(f) => f.toPath.toUri.toString
```
lefou, line 74:
```suggestion
        case Some(f) => 
          // The extra step invoking `toPath` results in a nicer URI starting with `file:///`
          f.toPath.toUri.toString
```

lefou (approved):

bsp/worker/src/mill/bsp/worker/BspCompileProblemReporter.scala
```diff
@@ -69,9 +69,11 @@ private class BspCompileProblemReporter(
     val diagnostic = getSingleDiagnostic(problem)
     val sourceFile = problem.position.sourceFile
     val textDocument = new TextDocumentIdentifier(
-      sourceFile.getOrElse(None) match {
+      sourceFile match {
         case None => targetId.getUri
-        case f: File => f.toURI.toString
+        case Some(f) => 
+          // The extra step invoking `toPath` results in a nicer URI starting with `file:///`
+          f.toPath.toUri.toString
```
lefou, line 76:
```suggestion
        case Some(f) =>
          // The extra step invoking `toPath` results in a nicer URI starting with `file:///`
          f.toPath.toUri.toString
```

lolgab:
Thank you for making the changes for me!

Issue #2426, opened 2023-04-07
BSP: mill-bsp doesn't publish `build/publishDiagnostics` when there's no diagnostics


tanishiking:
- mill v0.10.12
- Java 17
- Metals 0.11.11
- Scala 3.2.2

## Problem

mill-bsp doesn't publish `build/publishDiagnostics` when there's no diagnostics.
As a result, previously published diagnostic won't be cleaned up when error is fixed.

https://user-images.githubusercontent.com/9353584/230577077-377f65db-407c-4a26-a6eb-bdadfe2d62f3.mov

## Expect

mill-bsp should publish `build/publishDagnostics` when error is fixed (bloop and sbt-bsp does)

```json
[Trace - 05:56:44 PM] Received notification 'build/publishDiagnostics'
Params: {
  "textDocument": {
    "uri": "file:///Users/tanishiking/ghq/github.com/tanishiking/metals-playground/HelloWorld/src/main/scala/hello/Foo.scala"
  },
  "buildTarget": {
    "uri": "file:/Users/tanishiking/ghq/github.com/tanishiking/metals-playground/HelloWorld/?id\u003dHelloWorld"
  },
  "diagnostics": [],
  "reset": true
}
```

KaranAhlawat:
I was trying to work on a solution for this today, and came across some very strange results.

In my own simple project, if I use Mill BSP, the empty diagnostics array is not published.
However, when I was working in the `mill` repository itself, the diagnostic *was* being published


```txt
[Trace - 08:28:10 pm] Received notification 'build/publishDiagnostics'
Params: {
  "textDocument": {
    "uri": "file:/home/karan/Secondary/dev/mill-contrib/bsp/worker/src/mill/bsp/worker/BspCompileProblemReporter.scala"
  },
  "buildTarget": {
    "uri": "file:///home/karan/Secondary/dev/mill-contrib/bsp/worker"
  },
  "diagnostics": [
    {
      "range": {
        "start": {
          "line": 56,
          "character": 12
        },
        "end": {
          "line": 56,
          "character": 24
        }
      },
      "severity": 1,
      "source": "mill",
      "message": "value toI is not a member of String"
    }
  ],
  "reset": true
}


[Trace - 08:28:10 pm] Received notification 'build/taskFinish'
Params: {
  "taskId": {
    "id": "332966380"
  },
  "eventTime": 1694876290825,
  "message": "Compiled bsp.worker",
--
[...]

PR #2810, opened 2023-10-02
fix: Errors not cleaning up in BSP for Scala 3

lolgab:
The Scala 3 compiler is not calling `startUnit` as the Scala 2 compiler. I checked what sbt is doing and it sends a build/publishDiagnostics for all the compiled files at the end of compilation. So I'm doing the same here.
I removed the `CompileProgress` and I'm calling fileVisited for all input files instead.
I tried to not touch the `CompileProblemReporter` trait for now since it's public API and I don't want to deal with the binary compatibility issues. Moreover, our current `fileVisited` API seems more than enough.

Fixes https://github.com/com-lihaoyi/mill/issues/2426

Pull request: https://github.com/com-lihaoyi/mill/pull/2810

### DIFF
--- a/scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
+++ b/scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
@@ -492,28 +492,6 @@ class ZincWorkerImpl(
       .map(path => converter.toVirtualFile(path.toNIO))
       .toArray
 
-    // We want to listed to all compiled files
-    val compileProgress = reporter.map { reporter =>
-      new CompileProgress {
-        val seen = new ConcurrentHashMap[String, java.lang.Boolean]()
-        override def startUnit(phase: String, unitPath: String): Unit = {
-          // ctx.log.info(s"compiler starts unit `${unitPath}` (phase: ${phase})")
-          try {
-            seen.putIfAbsent(unitPath, java.lang.Boolean.TRUE) match {
-              case null =>
-                // first visit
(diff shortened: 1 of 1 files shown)


lefou (approved):
Looks good to me.

scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
```diff
@@ -492,28 +492,6 @@ class ZincWorkerImpl(
       .map(path => converter.toVirtualFile(path.toNIO))
       .toArray
 
-    // We want to listed to all compiled files
-    val compileProgress = reporter.map { reporter =>
```
tgodzik, line 496: The issue is most likely here https://github.com/lampepfl/dotty/issues/13082

Wouldn't  CompileProgress be useful for reporting % of the compilation or does mill handle it differently?

lolgab: We are not reporting % of the compilation in Mill yet. I'm creating an issue so we don't forget about it.

tgodzik: Thanks! Anyway, this needs the fix for Scala 3, but otherwise we can probably copy the logic from Bloop

lolgab: Issue created here: #2816

PR #2812, opened 2023-10-02
Revert "Update bsp4j to 2.1.0-M6 and adapt to breaking API changes (#2794)"

lefou:
This reverts commit 1d2ba276d4aafef3f2d3b3ba10a1c229bfc42459.


lefou:
Not needed, instead, we need to bump to 2.1.0-M7.

See https://github.com/build-server-protocol/build-server-protocol/issues/619#issuecomment-1743632213

PR #2814, opened 2023-10-02
Cleanup BSP server implementation

lolgab:
- Avoid BSP server integration when generating the semanticdb data
- Make BSPCompileProblemReporter thread unsafe since can be called from a single thread (the Scala compiler is single threaded, now only one compilation has BSP and every module gets its own instance)
- Start sending partial events to BSP using `reset = false` when the compiler reports a problem. Before we were sending all the issues every time, by using the `reset = true` flag. Now we send one problem at a time using `reset = false` and at the end of the compilation we send them all using `reset = true` (which happens in `fileVisited`)


lefou (commented):
Is `T.reporter` now useful for anything? Should we deprecate it?

lefou (approved):
Looks good to me. Thanks!

PR #2822, opened 2023-10-05
Allow publishing bridge for a single version

lolgab:
Releasing a new version of the bridge just for adding a Scala version is wasteful.
We can publish the bridge just for the new version (in our case `2.13.12`) without tagging a `0.0.2`.
This PR adds a `bridge_versions` input to the `workflow_dispatch` which can be `all` for the current behavior or a comma separated list of Scala versions. This way we can release the current bridge for Scala `2.13.12` and then add `2.13.12` to the list of supported bridges in a later PR.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -185,13 +185,6 @@ def millBinPlatform: T[String] = T {
 
 def baseDir = build.millSourcePath
 
-// We limit the number of compiler bridges to compile and publish for local
-// development and testing, because otherwise it takes forever to compile all
-// of them. Compiler bridges not in this set will get downloaded and compiled
-// on the fly anyway. For publishing, we publish everything.
-val buildAllCompilerBridges = interp.watchValue(sys.env.contains("MILL_BUILD_COMPILER_BRIDGES"))
-val bridgeVersion = "0.0.1"
-
 val bridgeScalaVersions = Seq(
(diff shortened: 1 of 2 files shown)


build.sc
```diff
@@ -223,7 +216,17 @@ val bridgeScalaVersions = Seq(
...
-val buildBridgeScalaVersions = if (!buildAllCompilerBridges) Seq() else bridgeScalaVersions
+// We limit the number of compiler bridges to compile and publish for local
+// development and testing, because otherwise it takes forever to compile all
+// of them. Compiler bridges not in this set will get downloaded and compiled
+// on the fly anyway. For publishing, we publish everything or a specific version
+// if given.
+val compilerBridgeScalaVersions = interp.watchValue(sys.env.get("MILL_COMPILER_BRIDGE_VERSIONS")) match {
+  case None => Seq.empty[String]
+  case Some("all") => bridgeScalaVersions
+  case Some(versions) => versions.split(',').toSeq
```
lefou, line 227: I'd add a `.map(_.trim)`, just to be sure.

lefou (commented):
Good idea. I wasn't aware, that we can request input when starting a `workflow_dispatch` job.

lefou (approved):

PR #2823, opened 2023-10-05
Avoid recalculating `PathRef`s for `allSourceFiles`

lolgab:
Calculating `PathRef`s is expensive, so reducing them to the minimum improves performances.
This PR adds an internal task called `allSourceFilesTask` that is calculated everytime and returns `Seq[os.Path]` instead of `Seq[PathRef]`. The invalidation still works since `sources`, which is the input of `allSourceFilesTask` returns the `PathRef`s of the sources directories

### DIFF
--- a/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
+++ b/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
@@ -311,7 +311,7 @@ private class MillBuildServer(
       val tasksEvaluators = state.bspModulesById.iterator.collect {
         case (id, (m: JavaModule, ev)) =>
           T.task {
-            val src = m.allSourceFiles()
+            val src = m.allSourceFilesTask()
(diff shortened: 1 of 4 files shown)


lolgab:
Doesn't work since breaks the ability to override `allSourceFiles`

lihaoyi:
How expensive is calling `allSourceFiles` really? Source files are normally pretty small, so hashing them twice is wasteful but shouldn't be too expensive. The timing should be visible in your `mill-profile.json`

lolgab:
> How expensive is calling `allSourceFiles` really? Source files are normally pretty small, so hashing them twice is wasteful but shouldn't be too expensive. The timing should be visible in your `mill-profile.json`

Not a big deal, it's a couple of milliseconds per module. I just want Mill to be optimal, so I'm looking for any opportunity to gain milliseconds here and there :)

PR #2827, opened 2023-10-05
Microsecond precision in chrome profile

lolgab:
Chrome profiles support microsecond precision, so we can use `System.nanoTime` instead of `System.currentTimeMillis` to have better profiles

### DIFF
--- a/main/eval/src/mill/eval/EvaluatorCore.scala
+++ b/main/eval/src/mill/eval/EvaluatorCore.scala
@@ -97,7 +97,7 @@ private[mill] trait EvaluatorCore extends GroupEvaluator {
             .flatMap(_.iterator.flatMap(_.newResults))
             .toMap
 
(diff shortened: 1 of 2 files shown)


lefou (commented):
API-wise, I'd rather keep the start/end params in the helper functions and avoid the need to calculate the duration on the client side, but it looks like you are convinced we can make Mill twice as fast. haha.

lolgab:
@lefou You got me :-D
Since they are private functions I'd go for the micro-optimized way of calculating it only once, but if you prefer as before I can use `endTime`. As you prefer :)

lefou (approved):
LGTM.

lefou:
> @lefou You got me :-D Since they are private functions I'd go for the micro-optimized way of calculating it only once, but if you prefer as before I can use `endTime`. As you prefer :)

I'm eager to analyze the benchmarks. 😜

PR #2829, opened 2023-10-06
docs: proguard - more detailed example

IdiosApps:
Pretty much all these new entries will be needed to use this plugin successfully


lefou (approved):
Thank you!

Issue #2808, opened 2023-10-02
Sporadic failures with `-w`/`--watch`


lihaoyi:
Started seeing this when working on a new project. Pretty trivial single-module Scala build. Mill 0.11.x. Not sure why I've never seen it before elsewhere, but it seems pretty frequent (happens several times a day). Using Intellij BSP.

Not sure what the `~` is at the end of `/Users/lihaoyi/Github/usql/test/src/MainTests.scala~`. Maybe some kind of temporary file IntelliJ/IntelliJ-BSP is creating and deleting quickly, cause a TOCTTOU race condition?

```
Watching for changes to 11 paths and 4 other values... (Enter to re-run, Ctrl-C to exit)
Exception in thread "main" java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at mill.main.client.IsolatedMillMainLoader.runMain(IsolatedMillMainLoader.java:58)
	at mill.main.client.MillClientMain.main(MillClientMain.java:64)
Caused by: java.nio.file.NoSuchFileException: /Users/lihaoyi/Github/usql/test/src/MainTests.scala~
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
	at java.base/sun.nio.fs.UnixFileAttributeViews$Posix.readAttributes(UnixFileAttributeViews.java:257)
	at java.base/sun.nio.fs.UnixFileAttributeViews$Posix.readAttributes(UnixFileAttributeViews.java:168)
	at java.base/sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:148)
	at java.base/java.nio.file.Files.readAttributes(Files.java:1851)
	at java.base/java.nio.file.Files.getPosixFilePermissions(Files.java:2125)
	at os.perms$.apply(PermsOps.scala:13)
[...]

PR #2832, opened 2023-10-08
Try to make `PathRef` robust against concurrent filesystem modifications

lihaoyi:
Attempts to fix https://github.com/com-lihaoyi/mill/issues/2808

No real tests; the failure is timing-dependent and hard to reproduce. But hopefully just wrapping everything up in a big try-catch will fix it

Pull request: https://github.com/com-lihaoyi/mill/pull/2832


lefou (approved):
Sounds reasonable. I have the urge to log it at debug level somewhere, but as we don't have a task context, I have no clear idea how.

PR #2836, opened 2023-10-09
Check for collisions in Cross instances, ban forward slash in Cross segments, support os.SubPath in Cross modules

### LINKED ISSUE #2835: Mill cannot handle forward slash in crossValue
I am using mill version `0.11.5`.

My main goal is handling a nested hierarchy of files with [Dynamic Cross Modules](https://mill-build.com/mill/Cross_Builds.html#_dynamic_cross_modules), but I am running into issues where attempting to use relative paths as cross values causes mill to error out.

Consider the following:
```scala
import mill._
object foo extends Cross[Foo]("a", "b/c")
trait Foo extends Cross.Module[String] {
  def value = T { crossValue }
}
```
Mill will happily build these cross modules:
```
$ mill resolve foo._
foo[a]
foo[b/c]
```
But when you try to do Tasks that cache to disk, I think Mill is hitting an internal error due to the `/`:
```
$ mill foo[_].value
Exception in thread "MillServerActionRunner" os.PathError$InvalidSegment: [b/c] is not a valid path segment. [/] is not a valid character to appear in a path segment. If you want to parse an absolute or relative path that may have multiple segments, e.g. path-strings coming from external sources use the Path(...) or RelPath(...) constructor calls to convert them. 
        at os.BasePath$.fail$1(Path.scala:125)
        at os.BasePath$.checkSegment(Path.scala:131)
        at os.PathChunk$StringPathChunk.<init>(Path.scala:25)
        at os.PathChunk$.StringPathChunk(Path.scala:24)
        at mill.eval.EvaluatorPaths$.$anonfun$resolveDestPaths$3(EvaluatorPaths.scala:31)
        at scala.collection.immutable.List.map(List.scala:250)
        at scala.collection.immutable.List.map(List.scala:79)
        at os.PathChunk$SeqPathChunk.<init>(Path.scala:46)
        at os.PathChunk$.SeqPathChunk(Path.scala:43)
        at mill.eval.EvaluatorPaths$.resolveDestPaths(EvaluatorPaths.scala:31)
        at mill.eval.GroupEvaluator.evaluateGroupCached(GroupEvaluator.scala:162)
        at mill.eval.GroupEvaluator.evaluateGroupCached$(GroupEvaluator.scala:44)
        at mill.eval.EvaluatorImpl.evaluateGroupCached(EvaluatorImpl.scala:15)
        at mill.eval.EvaluatorCore.$anonfun$evaluate0$2(EvaluatorCore.scala:116)
        at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:467)
        at mill.eval.ExecutionContexts$RunNow$.execute(ExecutionContexts.scala:13)
        at scala.concurrent.impl.Promise$Transformation.submitWithValue(Promise.scala:429)
        at scala.concurrent.impl.Promise$DefaultPromise.submitWithValue(Promise.scala:338)
        at scala.concurrent.impl.Promise$DefaultPromise.dispatchOrAddCallbacks(Promise.scala:312)
        at scala.concurrent.impl.Promise$DefaultPromise.map(Promise.scala:182)
        at mill.eval.EvaluatorCore.$anonfun$evaluate0$1(EvaluatorCore.scala:92)
        at mill.eval.EvaluatorCore.$anonfun$evaluate0$1$adapted(EvaluatorCore.scala:90)
        at scala.collection.immutable.Vector.foreach(Vector.scala:2124)
        at mill.eval.EvaluatorCore.evaluate0(EvaluatorCore.scala:90)
        at mill.eval.EvaluatorCore.$anonfun$evaluate$1(EvaluatorCore.scala:43)
        at scala.util.DynamicVariable.withValue(DynamicVariable.scala:59)
        at mill.eval.EvaluatorCore.evaluate(EvaluatorCore.scala:34)
        at mill.eval.EvaluatorCore.evaluate$(EvaluatorCore.scala:26)
        at mill.eval.EvaluatorImpl.evaluate(EvaluatorImpl.scala:15)
        at mill.main.RunScript$.evaluateNamed(RunScript.scala:38)
        at mill.main.RunScript$.$anonfun$evaluateTasksNamed$2(RunScript.scala:26)
        at scala.util.Either.map(Either.scala:382)
        at mill.main.RunScript$.evaluateTasksNamed(RunScript.scala:26)
        at mill.runner.MillBuildBootstrap$.evaluateWithWatches(MillBuildBootstrap.scala:399)
        at mill.runner.MillBuildBootstrap.$anonfun$processFinalTargets$3(MillBuildBootstrap.scala:308)
        at scala.util.DynamicVariable.withValue(DynamicVariable.scala:59)
        at mill.runner.MillBuildBootstrap.processFinalTargets(MillBuildBootstrap.scala:308)
        at mill.runner.MillBuildBootstrap.evaluateRec(MillBuildBootstrap.scala:196)
        at mill.runner.MillBuildBootstrap.evaluate(MillBuildBootstrap.scala:48)
        at mill.runner.MillMain$.$anonfun$main0$6(MillMain.scala:234)
        at mill.runner.Watching$.watchLoop(Watching.scala:27)
        at mill.runner.MillMain$.$anonfun$main0$1(MillMain.scala:219)
        at scala.util.DynamicVariable.withValue(DynamicVariable.scala:59)
        at scala.Console$.withErr(Console.scala:193)
        at mill.api.SystemStreams$.$anonfun$withStreams$2(SystemStreams.scala:62)
        at scala.util.DynamicVariable.withValue(DynamicVariable.scala:59)
        at scala.Console$.withOut(Console.scala:164)
        at mill.api.SystemStreams$.$anonfun$withStreams$1(SystemStreams.scala:61)
        at scala.util.DynamicVariable.withValue(DynamicVariable.scala:59)
        at scala.Console$.withIn(Console.scala:227)
        at mill.api.SystemStreams$.withStreams(SystemStreams.scala:60)
        at mill.runner.MillMain$.main0(MillMain.scala:101)
        at mill.runner.MillServerMain$.main0(MillServerMain.scala:83)
        at mill.runner.MillServerMain$.main0(MillServerMain.scala:35)
        at mill.runner.Server.$anonfun$handleRun$1(MillServerMain.scala:187)
        at java.base/java.lang.Thread.run(Thread.java:833)
```

If I had to guess, I would guess that https://github.com/com-lihaoyi/mill/blob/1cd494b64a5749b976da8074c4c1ae9c396577d4/main/eval/src/mill/eval/EvaluatorPaths.scala#L31 should be converting the `Seq[String]` to a `os.RelPath`.

jackkoenig:
Fixes #2835

This PR does 3 things (and I'm happy to split it into smaller PRs since each commit is separable):
1. Checks for collisions in cross instances
2. Bans '/' in cross segments (my original approach is now disallowed, and the user is encouraged to use os.SubPath)
3. Adds support for os.SubPath in Cross instances

My implementations for (1) and (2) could be better, I don't love throwing exceptions, but I wanted to get something functional out for review to get pointers on how better to report errors to the user here. It would be nice to print the actual line rather than just printing the filename and line.

For (3), I have not yet added a test because I figure it would make sense to include in the examples, probably in a section immediately after https://mill-build.com/mill/Cross_Builds.html#_dynamic_cross_modules (ie. before `Use Case: Static Blog`), but I wanted @lefou's blessing before starting that work.

### DIFF
--- a/integration/failure/dest-directory-collision/repo/build.sc
+++ b/integration/failure/dest-directory-collision/repo/build.sc
@@ -1,0 +1,4 @@
+import mill._
+
+object foo extends Cross[Foo](Seq((Seq("a", "b"), Seq("c")), (Seq("a"), Seq("b", "c"))))
+trait Foo extends Cross.Module2[Seq[String], Seq[String]]
--- a/integration/failure/dest-directory-collision/test/src/DestDirectoryCollision.scala
+++ b/integration/failure/dest-directory-collision/test/src/DestDirectoryCollision.scala
@@ -1,0 +1,19 @@
+package mill.integration
+
+import utest._
+
+object DestDirectoryCollisionTests extends IntegrationTestSuite {
+  val tests = Tests {
+    val workspaceRoot = initWorkspace()
+
+    test("success") {
+      val res = evalStdout("resolve", "foo._")
+      assert(!res.isSuccess)
+      assert(
+        res.err.contains(
+          "have colliding destination path segments"
+        )
+      )
+    }
+  }
+}
--- a/integration/failure/forward-slash-in-path-segment/repo/build.sc
+++ b/integration/failure/forward-slash-in-path-segment/repo/build.sc
@@ -1,0 +1,4 @@
+import mill._
+
+object foo extends Cross[Foo]("a/b")
+trait Foo extends Cross.Module[String]
--- a/integration/failure/forward-slash-in-path-segment/test/src/ForwardSlashInPathSegment.scala
+++ b/integration/failure/forward-slash-in-path-segment/test/src/ForwardSlashInPathSegment.scala
@@ -1,0 +1,19 @@
+package mill.integration
+
+import utest._
+
+object ForwardSlashInPathSegmentTests extends IntegrationTestSuite {
+  val tests = Tests {
+    val workspaceRoot = initWorkspace()
+
+    test("success") {
+      val res = evalStdout("resolve", "foo._")
+      assert(!res.isSuccess)
+      assert(
+        res.err.contains(
+          "contains '/' which is not allowed. Try using os.SubPath"
+        )
+      )
+    }
+  }
+}
--- a/main/define/src/mill/define/Cross.scala
+++ b/main/define/src/mill/define/Cross.scala
@@ -111,6 +111,7 @@ object Cross {
     implicit object ShortToPathSegment extends ToSegments[Short](v => List(v.toString))
     implicit object ByteToPathSegment extends ToSegments[Byte](v => List(v.toString))
     implicit object BooleanToPathSegment extends ToSegments[Boolean](v => List(v.toString))
+    implicit object SubPathToPathSegment extends ToSegments[os.SubPath](v => v.segments.toList)
     implicit def SeqToPathSegment[T: ToSegments]: ToSegments[Seq[T]] = new ToSegments[Seq[T]](
       _.flatMap(implicitly[ToSegments[T]].convert).toList
     )
--- a/main/define/src/mill/define/Cross.scala
+++ b/main/define/src/mill/define/Cross.scala
@@ -286,28 +286,39 @@ class Cross[M <: Cross.Module[_]](factories: Cross.Factory[M]*)(implicit
     def cls: Class[_]
   }
 
-  val items: List[Item] = for {
-    factory <- factories.toList
-    (crossSegments0, (crossValues0, (cls0, make))) <-
-      factory.crossSegmentsList.zip(factory.crossValuesListLists.zip(factory.makeList))
-  } yield {
-    val relPath = ctx.segment.pathSegments
-    val module0 = new Lazy(() =>
-      make(
-        ctx
-          .withSegments(ctx.segments ++ Seq(ctx.segment))
-          .withMillSourcePath(ctx.millSourcePath / relPath)
-          .withSegment(Segment.Cross(crossSegments0))
-          .withCrossValues(factories.flatMap(_.crossValuesRaw))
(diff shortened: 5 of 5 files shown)


lefou:
I https://github.com/com-lihaoyi/mill/issues/2835#issuecomment-1754794042 I wrote:

> I don't think allowing a cross-value to open up multiple sub-paths is the right solution. We already map `n` cross-values to `n` sub-directories. Now, when we also allow the value itself to contain sub-directories, we can't no longer guarantee unique `T.dest` directories.
> 
> Here is my counter example, in which both cross module instances whould use the same `T.dest` path for their targets.
> 
> ```scala
> import mill._
> object foo extends Cross[Foo](("a", "b/c"), ("a/b", "c"))
> trait Foo extends Cross.Module2[String, String] {
>   ...
> }
> ```
> 
> Instead, we could encode the special `/` character in the path. But that would create less nice cache directory names.

jackkoenig:
@lefou I ended up taking a slightly different approach, rather than using `os.SubPath` when processing segments, I banned `/` in segments and instead encourage the user to use `os.SubPath` themselves. See the 1st PR comment up top where I've described the new approach in detail.

lefou:
I'm not convinced. Why we want to force a path on things that aren't paths, like a version matrix. As we agreed that we want cross module parameters to be able to take more that one segment, why should we only allow that for paths, but not other values? Instead we should encourage to model a build using domain types when needed. This includes avoiding generic data types like strings and paths in favor to more specific types, hence the type class to still be able to map to a segmented path.

IMHO, the essential thing is to make sure, there are no collisions, so we have safe cache locations and unambiguous CLI arguments.

lefou:
Superseded by #2984

PR #2848, opened 2023-10-25
Test Arguments

cheeseng:
Added testArgs to TestModule to allow specifying test arguments as part of the build.

### DIFF
--- a/scalalib/src/mill/scalalib/TestModule.scala
+++ b/scalalib/src/mill/scalalib/TestModule.scala
@@ -25,13 +25,21 @@ trait TestModule extends TaskModule with TestModule.JavaModuleBase {
   def testFramework: T[String]
 
   /**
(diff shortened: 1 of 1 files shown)


lefou (commented):
Is the logic to either use the `testArgs` or the explicitly given args already used in other build tools? I'm asking, to get a feeling for how familar users might be with this concept. Currently, tests may fail behave differently once the user specified extra args, due to the fact, that the default args are no longer applied. `testDefaultArgs` might be more clear name then.

I wonder whether `testCachedArgs` should default to use the value of the new `testArgs`. Or if we could join both or even just use `testCachedArgs` in both cases.

lefou:
Switched to draft due to pending feedback.

lihaoyi:
Closing due to inactivity

PR #2865, opened 2023-11-08
ExampleTestSuite: Also support `mill` command in addition to `./mill`

lefou:
Replace `./mill` with `mill` in most examples, as this is easier to read. I kept the `./mill` in the first simple scala project, which also describes that there is a `./mill` script in each example archive.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -1025,7 +1025,10 @@ object example extends MillScalaModule {
     def compile = example.compile()
     def forkEnv = super.forkEnv() ++ Map("MILL_EXAMPLE_PARSED" -> upickle.default.write(parsed()))
 
-    def parsed = T {
+    /**
+     * Parses a `build.sc` for specific comments and return the split-by-type content
+     */
+    def parsed: T[Seq[(String, String)]] = T {
       val states = collection.mutable.Buffer("scala")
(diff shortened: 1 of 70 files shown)


lihaoyi:
Should we support both? I feel like supporting 1 should be enough, and would be good to enforce some degree of consistency throughout the docsite which gets built from these test suites. If we want to use `mill`, we should cut over everything to `mill`

lihaoyi:
I wanted to go with `./mill` to remove the need for installation from the example projects, as all the zip downloads come with a `./mill` script built in, so in theory anyone just needs a JVM installed and they can download the zip and follow the instructions and be ready to go

lefou:
Yeah, I this I got your initial motivation for `./mill`. Yet, I think it reads better with just `mill`. It also encourages to just install mill on the system and avoid dependency on some local script. It gives a newcommer a bit less hack-ish feeling. I for myself always prefer a system-installed tool over some repo-local script.

I think keeping both variants supported is not really adding inconsistency. As we explain the user that we have that script in example `1-simple-scala`, I thought we should keep at the `./mill` at least in this example, and it will still work as an executable test.

lihaoyi:
sounds good to me, go for it

PR #2874, opened 2023-11-14
Add check for right Tests traits in ScalaJS and Native

lolgab:
## Motivation

People get confused when they do things like:
```scala
import mill._
import mill.scalalib._
import mill.scalajslib._

object root extends ScalaJSModule {
  def scalaVersion: T[String] = "3.3.1"
  def scalaJSVersion: T[String] = "1.14.0"
  object test extends ScalaTests with TestModule.Utest
}
```

which doesn't work since we need to extend `ScalaJSTests` instead of `ScalaTests`.
Now we crash with an exception:
```
[build.sc] [49/53] compile 
[info] done compiling
[build.sc] [53/53] methodCodeHashSignatures 
mill.api.MillException: root is a `ScalaJSModule`. root.test needs to extend `ScalaJSTests`.
    mill.scalalib.ScalaModule$ScalaTests.$init$(ScalaModule.scala:37)
    millbuild.build$root$test$.<init>(build.sc:8)
    millbuild.build$root$.test$lzycompute$1(build.sc:30)
    millbuild.build$root$.test(build.sc:30)
    java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    java.base/java.lang.reflect.Method.invoke(Method.java:566)
    mill.resolve.ResolveCore$.$anonfun$resolveDirectChildren0$10(ResolveCore.scala:272)
```

Pull Request: https://github.com/com-lihaoyi/mill/pull/2874

### DIFF
--- a/scalalib/src/mill/scalalib/ScalaModule.scala
+++ b/scalalib/src/mill/scalalib/ScalaModule.scala
@@ -20,6 +20,19 @@ trait ScalaModule extends JavaModule with TestModule.ScalaModuleBase { outer =>
   type ScalaModuleTests = ScalaTests
 
   trait ScalaTests extends JavaModuleTests with ScalaModule {
+    require(
+      !Class.forName("mill.scalajslib.ScalaJSModule").isInstance(outer) || Class.forName(
+        "mill.scalajslib.ScalaJSModule$ScalaJSTests"
+      ).isInstance(this),
+      s"$outer is a `ScalaJSModule`. $this needs to extend `ScalaJSTests` instead of `ScalaTests`"
+    )
+    require(
+      !Class.forName("mill.scalanativelib.ScalaNativeModule").isInstance(outer) || Class.forName(
+        "mill.scalajslib.ScalaNativeModule$ScalaNativeTests"
+      ).isInstance(this),
+      s"$outer is a `ScalaNativeModule`. $this needs to extend `ScalaNativeTests` instead of `ScalaTests`"
+    )
+
     override def scalaOrganization: Target[String] = outer.scalaOrganization()
     override def scalaVersion: Target[String] = outer.scalaVersion()
     override def scalacPluginIvyDeps: Target[Agg[Dep]] = outer.scalacPluginIvyDeps()
--- a/scalalib/src/mill/scalalib/ScalaModule.scala
+++ b/scalalib/src/mill/scalalib/ScalaModule.scala
@@ -37,7 +37,7 @@ trait ScalaModule extends JavaModule with TestModule.ScalaModuleBase { outer =>
           "mill.scalanativelib.ScalaNativeModule$ScalaNativeTests"
         ).isInstance(this)
       ) throw new Exception(
-        s"$outer is a `ScalaNativeModule`. $this needs to extend `ScalaNativeTests` instead of `ScalaTests`"
+        s"$outer is a `ScalaNativeModule`. $this needs to extend `ScalaNativeTests`."
       )
     } catch {
       case _: ClassNotFoundException => // if we can't find the classes, we certainly are not in a ScalaNativeModule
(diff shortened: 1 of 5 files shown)


scalalib/src/mill/scalalib/ScalaModule.scala
```diff
@@ -20,6 +20,29 @@ trait ScalaModule extends JavaModule with TestModule.ScalaModuleBase { outer =>
   type ScalaModuleTests = ScalaTests
 
   trait ScalaTests extends JavaModuleTests with ScalaModule {
+    try {
+      if (
+        Class.forName("mill.scalajslib.ScalaJSModule").isInstance(outer) && !Class.forName(
+          "mill.scalajslib.ScalaJSModule$ScalaJSTests"
+        ).isInstance(this)
+      ) throw new Exception(
```
lefou, line 28: You can throw a `MillException` or even a `BuildScriptException` here, which will result in a nicer error message in the end.

scalalib/src/mill/scalalib/ScalaModule.scala
```diff
@@ -20,6 +20,29 @@ trait ScalaModule extends JavaModule with TestModule.ScalaModuleBase { outer =>
...
+        s"$outer is a `ScalaJSModule`. $this needs to extend `ScalaJSTests` instead of `ScalaTests`"
```
lefou, line 29: As a `ScalaJSTests` is by inheritance effectively also a `ScalaTests`, I'd avoid any extra confusion through the additional wording here. Just adding the `ScalaJSTests` is enough to resolve the problem in most cases. It is also fine to mix-in both traits, e.g. to configure other aspects.

```suggestion
        s"$outer is a `ScalaJSModule`. $this needs to extend `ScalaJSTests`."
```

scalalib/src/mill/scalalib/ScalaModule.scala
```diff
@@ -20,6 +20,29 @@ trait ScalaModule extends JavaModule with TestModule.ScalaModuleBase { outer =>
...
+      )
+    } catch {
+      case _: ClassNotFoundException => // if we can't find the classes, we certainly are not in a ScalaJSModule
+    }
+    try {
+      if (
+        Class.forName("mill.scalanativelib.ScalaNativeModule").isInstance(outer) && !Class.forName(
+          "mill.scalanativelib.ScalaNativeModule$ScalaNativeTests"
+        ).isInstance(this)
+      ) throw new Exception(
```
lefou, line 39: Should be a `MillException`.

scalalib/src/mill/scalalib/ScalaModule.scala
```diff
@@ -20,6 +20,29 @@ trait ScalaModule extends JavaModule with TestModule.ScalaModuleBase { outer =>
...
+        s"$outer is a `ScalaNativeModule`. $this needs to extend `ScalaNativeTests` instead of `ScalaTests`"
```
lefou, line 40:
```suggestion
        s"$outer is a `ScalaNativeModule`. $this needs to extend `ScalaNativeTests`."
```

lefou (commented):
A great addition to fix some regularly seen user problems. See my remarks below.

lefou (approved):
Looks good to me.

lefou:
@lolgab Just to be sure we don't overlook something. Are there definitely no valid use cases to have a `ScalaTests` module inside a `ScalaJSModule`? If we are not certain, we at least need some way to not fail the build.

lolgab:
> @lolgab Just to be sure we don't overlook something. Are there definitely no valid use cases to have a `ScalaTests` module inside a `ScalaJSModule`? If we are not certain, we at least need some way to not fail the build.

Sorry, I had already merged the pull request.
Anyway, I don't think there are use cases for it.

PR #2882, opened 2023-11-18
Fix #2881 Propagate environment from `ctx` to subprocesses

### LINKED ISSUE #2881: Stale server environment is propagated when running processes with `run`
## Reproduction

``` scala
// build.sc
import mill._
import mill.scalalib._

object root extends RootModule with ScalaModule {
  def scalaVersion = "3.3.1"
}
```

```scala
// src/main.scala
@main
def run =
  println(s"[app] MY_ENV=${sys.env.get("MY_ENV")}")
```

Run the following command:
```
rm -rf out; MY_ENV=hello mill run; mill run
```

It outputs:
```
rm -rf out; MY_ENV=hello mill run; mill run
[build.sc] [48/52] compile 
[info] done compiling
[35/48] compile 
[info] compiling 1 Scala source to /Users/lorenzo/scala/repro/out/compile.dest/classes ...
[info] done compiling
[48/48] run 
[app] MY_ENV=Some(hello)
[48/48] run 
[app] MY_ENV=Some(hello)
```

## Expectation
The second time it should output `[app] MY_ENV=None`

lolgab:
Fixes #2881

## Dependencies

- [x] os-lib fix for `propagateEnv` https://github.com/com-lihaoyi/os-lib/pull/238

### DIFF
--- a/main/util/src/mill/util/Jvm.scala
+++ b/main/util/src/mill/util/Jvm.scala
@@ -230,7 +230,8 @@ object Jvm extends CoursierSupport {
         env = envArgs,
         stdin = if (backgroundOutputs.isEmpty) os.Pipe else "",
         stdout = backgroundOutputs.map(_._1).getOrElse(os.Pipe),
-        stderr = backgroundOutputs.map(_._2).getOrElse(os.Pipe)
+        stderr = backgroundOutputs.map(_._2).getOrElse(os.Pipe),
+        propagateEnv = false
       )
 
       val sources = Seq(
@@ -255,7 +256,8 @@ object Jvm extends CoursierSupport {
         env = envArgs,
         stdin = if (backgroundOutputs.isEmpty) os.Inherit else "",
         stdout = backgroundOutputs.map(_._1).getOrElse(os.Inherit),
-        stderr = backgroundOutputs.map(_._2).getOrElse(os.Inherit)
+        stderr = backgroundOutputs.map(_._2).getOrElse(os.Inherit),
+        propagateEnv = false
       )
     }
   }
--- a/main/util/src/mill/util/Jvm.scala
+++ b/main/util/src/mill/util/Jvm.scala
@@ -146,21 +146,46 @@ object Jvm extends CoursierSupport {
 
     ctx.log.debug(s"Run subprocess with args: ${args.map(a => s"'${a}'").mkString(" ")}")
 
+    val envWithPropagated = ctx.env ++ envArgs
+
     if (backgroundOutputs.nonEmpty)
-      spawnSubprocessWithBackgroundOutputs(args, envArgs, workingDir, backgroundOutputs)
+      spawnSubprocessWithBackgroundOutputs(
+        args,
+        envWithPropagated,
+        workingDir,
+        backgroundOutputs,
+        propagateEnv = false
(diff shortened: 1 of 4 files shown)


main/util/src/mill/util/Jvm.scala
```diff
@@ -230,7 +230,8 @@ object Jvm extends CoursierSupport {
         env = envArgs,
         stdin = if (backgroundOutputs.isEmpty) os.Pipe else "",
         stdout = backgroundOutputs.map(_._1).getOrElse(os.Pipe),
-        stderr = backgroundOutputs.map(_._2).getOrElse(os.Pipe)
+        stderr = backgroundOutputs.map(_._2).getOrElse(os.Pipe),
+        propagateEnv = false
```
lefou, line 234: I'm not convinced this is correct. Yes, in order to avoid using stale env vars from the server, we need to set `propagaEnv` to `false`. But I think we should then always use the env vars send to us from the client as a base line. And if some env vars are given by the parameters, we should add them too.

This logic is of course not easily applyable here, as we don't have the access to the target context here.

lefou: The test case should use a env var typically used by the OS which is now unset.

lolgab: I did not know that envArgs are meant to be additional environment variables and not all the environment to be passed to the process.
I'm not sure how to achieve this to be honest..
We could add a `propagateEnv: Boolean = true` parameter here and set it to false when we know that we don't want to propagate the environment.
We could also add a private global `AtomicReference` holding the current env and set it when we read the environment from client and read it here.
Do you have other ideas?

lefou: TBH, we never specified whether a `run` with an empty `forkEnv` should use the OS env or the empty env. But the status quo seems to exactly do this. But in `run` we have access to a task context, so we could also accept an implicit one in the `Jvm` API.

Without passing it explicitly/implicitly, holding a static (global) map might be the alternative.

lolgab: > TBH, we never specified whether a `run` with an empty `forkEnv` should use the OS env or the empty env. But the status quo seems to exactly do this. But in `run` we have access to a task context, so we could also accept an implicit one in the `Jvm` API.

We could change this but, in case, I would prefer to do it in Mill 0.12 breaking window.

> Without passing it explicitly/implicitly, holding a static (global) map might be the alternative.

This seems the easiest solution as it doesn't need to introduce new APIs / deprecate APIs. Also, the environment is stable enough since you can't change it at runtime on the JVM. So the variable would change not that often (at most once per CLI execution).

lolgab: I ended up adding some private methods that receive also the `propagateEnv` argument so we can pass the env we have already in an `implicit ctx: Ctx` in `Jvm`. I don't want to make these overloads public since in 0.12 it's better to add `(implicit ctx: Ctx)` to the methods instead.

lolgab:
Now I don't have idea how to test this. Do we have integration tests where the client is used to call `run` on a module?

lolgab:
Superseded by https://github.com/com-lihaoyi/mill/pull/3437

Issue #2844, opened 2023-10-20
Project configuration fails after migrating to 0.11.2+


MaciejG604:
After migrating from 0.11.1 to any higher version, a project, that was working fine, fails on certain commands with an exception.

It happens in this repo: https://github.com/VirtusLab/scala-js-cli e.g. after doing ./mill -i 'tests.test'

Stack trace after setting the version to `0.11.2`:
```
Exception in thread "main" java.lang.reflect.InvocationTargetException
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:568)
        at mill.main.client.IsolatedMillMainLoader.runMain(IsolatedMillMainLoader.java:58)
        at mill.main.client.MillClientMain.main(MillClientMain.java:64)
Caused by: java.util.NoSuchElementException: head of empty list
        at scala.collection.immutable.Nil$.head(List.scala:662)
        at scala.collection.immutable.Nil$.head(List.scala:661)
        at mill.eval.GroupEvaluator$$anonfun$1.applyOrElse(GroupEvaluator.scala:103)
        at mill.eval.GroupEvaluator$$anonfun$1.applyOrElse(GroupEvaluator.scala:88)
        at scala.collection.Iterator$$anon$7.hasNext(Iterator.scala:525)
        at scala.collection.Iterator$$anon$10.hasNext(Iterator.scala:601)
        at scala.collection.IterableOnceOps.foldLeft(IterableOnce.scala:675)
        at scala.collection.IterableOnceOps.foldLeft$(IterableOnce.scala:670)
        at scala.collection.AbstractIterator.foldLeft(Iterator.scala:1300)
        at scala.collection.IterableOnceOps.sum(IterableOnce.scala:945)
        at scala.collection.IterableOnceOps.sum$(IterableOnce.scala:943)
        at scala.collection.AbstractIterator.sum(Iterator.scala:1300)
        at mill.eval.GroupEvaluator.evaluateGroupCached(GroupEvaluator.scala:136)
        at mill.eval.GroupEvaluator.evaluateGroupCached$(GroupEvaluator.scala:43)
[...]

PR #2883, opened 2023-11-19
Fix discovery of targets whose names get mangled due to their pseudo-private nature

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/2844

Adds an integration test to cover this edge case, a few other permutations of it that I could come up with, and other `private`-related behavior. I'm not sure if I managed to catch all different ways these methods can be mangled, but if I missed any we can discover/add them later. This new test fails if I remove any part of the code change in `GroupEvaluator.scala`

I'm actually not sure if this should be an integration test or a unit test. 

* Technically this code path should be exercised in unit tests as well I think, and I think this failure mode should be triggerable. 
* But it might be affected by exactly how the `def`s are wrapped in nested `object`s/`trait`s/`class`es, which is something that unit tests do not do realistically. 
* Also, a unit test cannot exercise the failure code path when you try and specify a `private` target from the command, since it will simply not compile (being `private` and all), and AFAICT this is somewhere we're missing coverage (unrelated to this specific failure mode)

So I went with integration test.

Pull request: https://github.com/com-lihaoyi/mill/pull/2883


lefou (approved):
Looks good to me!

PR #2884, opened 2023-11-19
Make `mill show` skip `-j` prefixes to ensure machine readability

lihaoyi:
This requires us add another method `Logger#rawOutputStream`, which the various implementations forward without decorating with prefixes or other things. 

We cannot just use `SystemStreams.original` because when running Mill in client/server mode we still need to use the `OutputStream` that gets forwarded from server to client, as the `SystemStreams.original.out` just ends up forwarded to `out/mill-worker-hash-1/stdout` files that aren't surfaced to users

Tested manually with `./mill -i dev.run tmp -i -j 2 show foo` on an example build in the `tmp/` folder. Previously it would show the output of `show` with a prefix e.g. `[#0] "bar"`, after this PR it shows the output without any prefixes `"bar"`

`Logger#rawOutputStream` has to be a non-abstract method for binary compatibility. This makes it a bit error prone since you can easily forget to override it where necessary. So I first made it abstract, made sure everyone overrides it, and then added the default case.

Pull request: https://github.com/com-lihaoyi/mill/pull/2884

### DIFF
--- a/bsp/src/mill/bsp/BspContext.scala
+++ b/bsp/src/mill/bsp/BspContext.scala
@@ -56,7 +56,10 @@ private[mill] class BspContext(
       override def error(s: String): Unit = streams.err.println(s)
       override def ticker(s: String): Unit = streams.err.println(s)
       override def debug(s: String): Unit = streams.err.println(s)
+
       override def debugEnabled: Boolean = true
+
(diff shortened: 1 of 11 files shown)


lefou (approved):
Looks good to me!

PR #2885, opened 2023-11-19
Report when we could not detect the declaring parent of a target

lefou:
This is related to #2844.

We report the target name in case we weren't able to detect it's declaring parent class.

Tested manually with a Mill version before the merge of #2883.

```scala
// /tmp/mill-2844/build.sc
import mill._

object mod extends Module {

  object foo extends Module {
    def bar = T {
      baz()
    }
  }

  private def baz = T { "baz" }
}
```

```
> mill -i dev.run /tmp/mill-2844 -i mod.foo.bar
[1691/1691] dev.run 
[info] compiling 1 Scala source to /tmp/mill-2844/out/mill-build/compile.dest/classes ...
[info] done compiling
Could not detect the parent class of target mod.baz. Please report this at https://github.com/com-lihaoyi/mill/issues/new/choose. As a workaround, you can run Mill with `--disable-callgraph-invalidation` option.
```

Pull request: https://github.com/com-lihaoyi/mill/pull/2885

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -1419,15 +1419,17 @@ object dev extends MillPublishScalaModule {
       case wd0 +: rest =>
         val wd = os.Path(wd0, T.workspace)
         os.makeDir.all(wd)
-        try Jvm.runSubprocess(
+        try {
+          Jvm.runSubprocess(
             Seq(launcher().path.toString) ++ rest,
             forkEnv(),
             workingDir = wd
           )
-        catch {
-          case e: Throwable => () /*ignore to avoid confusing stacktrace and error messages*/
+          mill.api.Result.Success(())
+        } catch {
+          case e: Throwable =>
+            mill.api.Result.Failure(s"dev.run failed with an exception. ${e.getMessage()}")
         }
-        mill.api.Result.Success(())
     }
   }
 }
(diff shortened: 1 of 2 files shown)


build.sc
```diff
@@ -1417,15 +1424,17 @@ object dev extends MillPublishScalaModule {
...
         os.makeDir.all(wd)
-        try Jvm.runSubprocess(
+        try {
+          Jvm.runSubprocess(
             Seq(launcher().path.toString) ++ rest,
             forkEnv(),
             workingDir = wd
           )
-        catch {
-          case e: Throwable => () /*ignore to avoid confusing stacktrace and error messages*/
```
lefou, line 1426: This hides any failures as `dev.run` will succeed.

main/eval/src/mill/eval/GroupEvaluator.scala
```diff
@@ -116,15 +123,15 @@ private[mill] trait GroupEvaluator {
                 ctx.enclosingModule match {
                   case null => None
                   case m: mill.define.Module => Some((m, m.millOuterCtx))
-                  case unknown => sys.error(s"Unknown ctx: $unknown")
+                  case unknown => throw new MillException(s"Unknown ctx of target ${namedTask}: $unknown")
```
lefou, line 126: Unrelated change adding more context to the error message.

PR #2886, opened 2023-11-21
Apply coursier default config files (repositories and mirrors config)

### LINKED ISSUE #2312: can not config repository mirror via ~/.config/coursier/mirror.properties
Repository mirror for coursier can be config in ~/.config/coursier/mirror.properties(in Linux), but this seems not apply to mill.

ref1: https://get-coursier.io/blog/2019/07/05/1.1.0-M14#mirrors
ref2: https://scalameta.org/metals/docs/troubleshooting/proxy/#mirrors

jilen:
Fix #2312 

Copied from 
https://github.com/coursier/coursier/blob/main/modules/coursier/shared/src/main/scala/coursier/Resolve.scala#L61

It is possible to use `Resolve().finalRepositories.future()`  to avoid the implementation detail.

But that requires `Await` and `ExecutionContext`,  I'm not sure whether worth doing it.

### DIFF
--- a/scalalib/src/mill/scalalib/CoursierModule.scala
+++ b/scalalib/src/mill/scalalib/CoursierModule.scala
@@ -8,6 +8,8 @@ import mill.define.Task
 import mill.api.PathRef
 
 import scala.annotation.nowarn
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
 
 /**
  * This module provides the capability to resolve (transitive) dependencies from (remote) repositories.
@@ -60,7 +62,12 @@ trait CoursierModule extends mill.Module {
    * The repositories used to resolved dependencies with [[resolveDeps()]].
    */
   def repositoriesTask: Task[Seq[Repository]] = T.task {
-    Resolve.defaultRepositories
+    import scala.concurrent.ExecutionContext.Implicits.global
+    val repos = Await.result(
+      Resolve().finalRepositories.future(),
+      Duration.Inf
+    )
+    repos
   }
 
   /**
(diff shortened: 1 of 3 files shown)


lefou:
I'd favor just calling the configuration logic from coursier. Copied code will get out of sync eventually. Await is standard Scala, so it should be fine to use it. It's easy to identify as boilerplate when reviewing the code.

jilen:
@lefou  Now it is implemented using `finalRepositories` .

lefou (commented):
Yeah, that's better.
We can import the execution context more local, e.g. just in the task.
Beside that, it looks good to me.

How did you test this? Would it be possible to add a small integration test under `integration.feature` or `example.misc`?

jilen:
> Yeah, that's better. We can import the execution context more local, e.g. just in the task. Beside that, it looks good to me.
> 
> How did you test this? Would it be possible to add a small integration test under `integration.feature` or `example.misc`?

I test it by remove  `repo1.maven.org`  in the `coursier` cache directory.  And it won't appear again.
I'll try to add some test

lefou:
In an integration test, we could use a different empty temporary cache location and could either assert, that Mill fails as expected, or fills that repo as expected.

jilen:
@lefou  I've add a test suite by setting the `courser.mirrors` system property.

lefou:
Thank you for this contribution, @jilen !

lefou:
@jilen It would be really cool, if you could review our documentation regarding repository configuration. We should mention all potential ways to customize it. IIUC, we now also support some coursiers-specific config files.

PR #2899, opened 2023-11-27
Update zinc from 1.9.6 to 1.10.0, use improved analysis store

lefou:
This PR updates Zinc to 1.10.0. We make use of a new `ConsistentFileAnalysisStore` which is used by zinc to store internal analysis data used for incremental compilation. The new store is faster and more reproducible.


jilen:
A new `ConsistentFileAnalysisStore` introduced to improve serialization performance. 
It will become default in next sbt version.
Maybe mill can switch to it.

lefou:
https://github.com/sbt/zinc/pull/1326

lefou:
> A new `ConsistentFileAnalysisStore` introduced to improve serialization performance. It will become default in next sbt version. Maybe mill can switch to it.

Done.

PR #2900, opened 2023-11-30
Added per-dependency format to dependency update finder

lefou:
Previously, we always presented all possible updates per modules.
Since, many Mill projects declare dependencies in a shared location
and use the same versions in all modules, this list can be rather long
and redundant.

This change adds a new `--format` option to select the format,
when searching for dependency updates. The old format is now named
`PerModule` and is the default (for compatibility).

Example:

```
> mill --meta-level 1 mill.scalalib.Dependency/showUpdates
Found 1 dependency update for root module
  io.github.classgraph:classgraph : 4.8.164 -> 4.8.165
```

I also added a new `PerDependency` format, which presents the dependency
with all it's possible versions and all affected modules.

Example:

```
> mill --meta-level 1 mill.scalalib.Dependency/showUpdates --format PerDependency
io.github.classgraph:classgraph : 4.8.164 -> 4.8.165 in
  root module
```

### DIFF
--- a/scalalib/src/mill/scalalib/Dependency.scala
+++ b/scalalib/src/mill/scalalib/Dependency.scala
@@ -3,7 +3,7 @@ package mill.scalalib
 import mill.T
 import mill.define.{Command, Discover, ExternalModule}
 import mill.eval.Evaluator
-import mill.scalalib.dependency.DependencyUpdatesImpl
+import mill.scalalib.dependency.{DependencyUpdatesImpl, Format}
 import mill.scalalib.dependency.updates.ModuleDependenciesUpdates
 
 object Dependency extends ExternalModule {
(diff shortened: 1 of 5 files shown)


scalalib/src/mill/scalalib/dependency/updates/DependencyUpdates.scala
lefou, line 1: Moved this class into it's own file

scalalib/src/mill/scalalib/Dependency.scala
```diff
@@ -24,9 +24,17 @@ object Dependency extends ExternalModule {
...
 
   /** Show possible dependency updates. */
-  def showUpdates(ev: Evaluator, allowPreRelease: Boolean = false): Command[Unit] = T.command {
-    DependencyUpdatesImpl.showAllUpdates(updates(ev, allowPreRelease)())
+  def showUpdates(
+      ev: Evaluator,
+      allowPreRelease: Boolean = false,
+      format: Format = Format.PerModule
+  ): Command[Unit] = T.command {
+    DependencyUpdatesImpl.showAllUpdates(updates(ev, allowPreRelease)(), format)
```
lefou, line 32: By adding this command before the other (now deprecated) overload, Mill resolves it correctly from CLI. I'm not sure, if that is guaranteed or arbitrary. Maybe, we should detect overloads and apply some heuristics (`@deprecated`) or explicit annotation to select the right one or fail.

PR #2931, opened 2023-12-22
fix: ensure the BSP clean works as expected

### LINKED ISSUE #2915: `buildTarget/cleanCache` throws an error
When using Mill as a BSP server I noticed that I'm getting an internal error when I try to do a clean compile. For example in the logs I see

```
[Trace - 02:09:53 pm] Sending request 'buildTarget/cleanCache - (24)'
Params: {
  "targets": [
    {
      "uri": "file:///Users/ckipp/Documents/scala-workspace/mill-plugins/mill-giter8/plugin/0.10"
    },
    {
      "uri": "file:///Users/ckipp/Documents/scala-workspace/mill-plugins/mill-giter8/plugin/0.11"
    },
    {
      "uri": "file:///Users/ckipp/Documents/scala-workspace/mill-plugins/mill-giter8/mill-build"
    }
  ]
}

...

[Trace - 02:09:53 pm] Received response 'buildTarget/cleanCache - (24)' in 4ms
Result: null
Error: {
  "code": -32603,
  "message": "Internal error.",
  "data": "java.util.concurrent.CompletionException: scala.MatchError:  (of class millbuild.build$)\n\tat java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:332)\n\tat java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:347)\n\tat java.base/java.util.concurrent.CompletableFuture$UniAccept.tryFire(CompletableFuture.java:708)\n\tat java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)\n\tat java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2162)\n\tat mill.bsp.worker.MillBuildServer.$anonfun$completable$1(MillBuildServer.scala:708)\n\tat mill.bsp.worker.MillBuildServer.$anonfun$completable$1$adapted(MillBuildServer.scala:696)\n\tat scala.concurrent.impl.Promise$Transformation.run(Promise.scala:484)\n\tat java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1395)\n\tat java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:373)\n\tat java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1182)\n\tat java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1655)\n\tat java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1622)\n\tat java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:165)\nCaused by: scala.MatchError:  (of class millbuild.build$)\n\tat mill.bsp.worker.MillBuildServer.$anonfun$buildTargetCleanCache$2(MillBuildServer.scala:595)\n\tat scala.collection.immutable.List.map(List.scala:246)\n\tat scala.collection.immutable.List.map(List.scala:79)\n\tat mill.bsp.worker.MillBuildServer.$anonfun$buildTargetCleanCache$1(MillBuildServer.scala:595)\n\tat mill.bsp.worker.MillBuildServer.$anonfun$completable$1(MillBuildServer.scala:699)\n\t... 8 more\n"
}
```

This then causes the compilation to be cancelled in the clean compile command.

### To reproduce

1. Make a simple hello world project
2. Open with Metals and make sure everything is imported while using Mill as your build server
3. Create a `.metals/bsp.trace.json` file
4. Restart the build server
5. Trigger a Metals Clean Compile command
6. See error above in logs

ckipp01:
Looking at the last change that was made to this code in https://github.com/com-lihaoyi/mill/commit/65fbd53108fb155858094d242a16a67ef7d067f4 it's not clear to me why this line was added where we map everything to `BspModule` and then filtering the build targets on that which where in the request. My assumption is that anything that the build client is requesting to clean _should_ be a valid identifier. If not, then that originally came from Mill, so I'm unsure why we have this. Before this change it was causing issues since not everything in `state.rootModules` was a `BspModule` so you'd get a match exception.

To expand a bit the build targets here in an example `mill-test` (hello world) project are:

```
BuildTargetIdentifier [
  uri = "file:///Users/<me>/Documents/scala-workspace/mill-test/milltest"
],BuildTargetIdentifier [
  uri = "file:///Users/<me>/Documents/scala-workspace/mill-test/mill-build"
]
```

But when looking into the `state.rootModules` one of them is

```
/Users/<me>/Documents/scala-workspace/mill-test
```

This was the problematic one. I'm not sure if this is just a mapping issue or not, but the change I made instead just takes the build target identifies given from the client, does a look up in the `state.bspModulesById` mapping, and goes with it. This seems to work and simplifies this unless this is introducing an issue that I don't see.

closes #2915

### DIFF
--- a/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
+++ b/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
@@ -592,9 +592,8 @@ private class MillBuildServer(
   override def buildTargetCleanCache(cleanCacheParams: CleanCacheParams)
       : CompletableFuture[CleanCacheResult] =
     completable(s"buildTargetCleanCache ${cleanCacheParams}") { state =>
-      val targetIds = state.rootModules.map { case b: BspModule => state.bspIdByModule(b) }
       val (msg, cleaned) =
-        cleanCacheParams.getTargets.asScala.filter(targetIds.contains).foldLeft((
+        cleanCacheParams.getTargets.asScala.foldLeft((
           "",
           true
         )) {


lefou:
This is a tricky one. Also, it's most likely underspecified. The issue is, that due to the newly introduced meta-build, all modules of a project depend implicitly on that meta-build. If you clean the meta-build, Mill will need to re-run it before it can understand any new actions. That is probably the reason, why the introduction of the meta-build also excluded it from the clean target. The predecessor of the meta-build was the Ammonite script build, which wasn't included in the `clean` run.

In difference to the Mill CLI, where all actions belong to the same build level, in BSP we make the levels transparent, so that all modules and meta-modules can be edited, disregarding of their meta-level. It might be safe to clean all BSP build targets (including the meta-builds), but some review needs to happen whether we properly re-run the meta-build from the next BSP request. In any case, cleaning the meta-build will definitely make the next BSP action take very long. Question is, if that is what the user wants?

ckipp01:
> This is a tricky one. Also, it's most likely underspecified. The issue is, that due to the newly introduced meta-build, all modules of a project depend implicitly on that meta-build. If you clean the meta-build, Mill will need to re-run it before it can understand any new actions. 

Ahh, ok interesting. Can you clarify what is meant by

> Mill will need to re-run before it can understand any new actions

In the BSP sense of things, what does this practically mean? Since in this case the user would be requesting a clean/compile, so is there anything else that would really be needed or hindered?

> In any case, cleaning the meta-build will definitely make the next BSP action take very long. Question is, if that is what the user wants?

Good question. I'm honestly not really sure. I guess in terms of correctness the BSP server _should_ listen to the request and clean everything that is requested to clean, but like you said, I don't know if a user would really ever want that. I did just check what happens with sbt when using it as your build server and I see the same thing, a request for the build sent in and it is indeed cleaned. Either way, I'm not sure. This was just my attempt to avoid the crash that I was seeing since I kept getting in states that were sort of borked and wanted to see if a full clean/compile would help.

lefou:
> Ahh, ok interesting. Can you clarify what is meant by
> 
> > Mill will need to re-run before it can understand any new actions
> 
> In the BSP sense of things, what does this practically mean? Since in this case the user would be requesting a clean/compile, so is there anything else that would really be needed or hindered?

In the BSP sense, I guess the only impact is that any next request whether already cached or not (even a subsequent `clean` request) will need to re-build the meta-module to determine the up-to-date state. Is it always "clean all" in Metals? Or can you select specific build targets, to avoid the overhead? In that case, we could think about some extra properties in the build target, indicating the meta-build level, which might be useful for sbt too. With that, Metals could by-default deselect meta-builds from clean, for example.

I think Mill already takes care of meta-build re-creation. The proof might be, that this PR just works for you. So we could just merge it as-is.

bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
```diff
@@ -592,9 +592,8 @@ private class MillBuildServer(
   override def buildTargetCleanCache(cleanCacheParams: CleanCacheParams)
       : CompletableFuture[CleanCacheResult] =
     completable(s"buildTargetCleanCache ${cleanCacheParams}") { state =>
-      val targetIds = state.rootModules.map { case b: BspModule => state.bspIdByModule(b) }
       val (msg, cleaned) =
-        cleanCacheParams.getTargets.asScala.filter(targetIds.contains).foldLeft((
```
lefou, line 597: Instead or removing this filter completely, we could log, that we clean the meta-build which will result in a slow subsequent build request.

PR #2932, opened 2023-12-22
fix: set `project.git = true` in scalafmt settings

ckipp01:
This ensures that if someone is using `scalafmt` via the CLI that it
correctly knows to ignore Scala files that are also ignored via git.
Without this locally when I used `scalafmt` from the command line I got
a ton of undesired changes.


lefou (approved):
Makes sense.

PR #2950, opened 2024-01-06
Enable colored output for Scala 2

### LINKED ISSUE #2947: Colors in Scala 2 compile errors are not rendered
For example, when using the `-Vimplicits` option, the colors in the chain of implicits that the compiler prints are not rendered properly. Here's a repo that reproduces the issue: https://github.com/mrdziuban/mill-compile-error-colors

image

lolgab:
Fixes #2947

Pull Request: https://github.com/com-lihaoyi/mill/pull/2950

### DIFF
--- a/scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
+++ b/scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
@@ -534,7 +534,10 @@ class ZincWorkerImpl(
       stampReader = Stamps.timeWrapBinaryStamps(converter)
     )
 
+    val scalaColorProp = "scala.color"
+    val previousScalaColor = sys.props.get(scalaColorProp).getOrElse("auto")
     try {
+      sys.props(scalaColorProp) = if (ctx.log.colored) "true" else "false"
       val newResult = ic.compile(
         in = inputs,
         logger = logger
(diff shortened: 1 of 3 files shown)


scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
```diff
@@ -534,7 +534,10 @@ class ZincWorkerImpl(
       stampReader = Stamps.timeWrapBinaryStamps(converter)
     )
 
+    val scalaColorProp = "scala.color"
+    val previousScalaColor = sys.props.get(scalaColorProp).getOrElse("auto")
```
lolgab, line 538: Inspired by sbt: https://github.com/sbt/sbt/blob/cfe0fc8e2ee3c551389fc24d836b04764e9a2e2b/main-actions/src/main/scala/sbt/Console.scala#L68

lefou: This means, we will later "reset" it to a new value "auto" instead of the original value, in case it was not defined. I'd rather reset to the original, which means we need to unset when it was unset before.

lolgab: I agree with you.

lefou (approved):
Looks good to me.

PR #2952, opened 2024-01-06
Fix #2946: Avoid evaluating `T.input`s twice

### LINKED ISSUE #2946: `T.input`s are evaluated twice
## Reproduction

```scala
import mill._

def foo = T.input{ println("bla") }
```

## Current behavior

```
mill foo
[1/1] foo 
bla
bla
```

## Expected behavior

It should print `bla` once:

```
mill foo
[1/1] foo 
bla
```

lolgab:
Fixes #2946

In `evaluateNamed`, after evaluating `Task`s we are matching for `ImputImpl[_]` instances to construct a `Watchable`.
A `Watchable` instance contains a `hashCode` for the current result and a callback to evaluate it again and get the next result.
Instead of using the current result which was contained in `TaskResult`, we were already calling the `recalc` callback, which executes again the `T.input`s and performs their side effects twice.
Now we are returning the hashCode for the already computed result, so we don't execute the `recalc` callback right away.

Pull Request: https://github.com/com-lihaoyi/mill/pull/2952

### DIFF
--- a/main/src/mill/main/RunScript.scala
+++ b/main/src/mill/main/RunScript.scala
@@ -46,7 +46,7 @@ object RunScript {
           Seq(Watchable.Path(p))
         case (t: InputImpl[_], TaskResult(result, recalc)) =>
           val pretty = t.ctx0.fileName + ":" + t.ctx0.lineNum
-          Seq(Watchable.Value(() => recalc().hashCode(), result.hashCode, pretty))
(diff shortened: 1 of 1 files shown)


lihaoyi (approved):
Looks reasonable I think. @lolgab can you update the PR description with your understanding of the problem and your proposed solution before merging it?

Issue #2958, opened 2024-01-08
Repro for Duplicated item inserted into OrderedSet


Quafadas:
@lefou  Aha! I believe I have a semi-minimal reproduction for this on a public repository. 

1. `git clone https://github.com/Quafadas/vecxt.git`
2. To prove that main works (you don't need to do this)
   *  `mill -j 1 vecxt.jvm.prepareOffline`
3. `git checkout build_error`
4.  `mill clean`
5.  `mill -j 1 vecxt.jvm.prepareOffline`

The first time I do this, I get 

```
simon@Simons-Mac-mini vecxt % mill clean
[1/1] clean 
simon@Simons-Mac-mini vecxt % mill -j 1 vecxt.jvm.prepareOffline
[23/23] vecxt.jvm.prepareOffline 
An unexpected error occurred
Exception in thread "MillServerActionRunner" java.lang.Exception: Duplicated item inserted into OrderedSet: mill.define.Task$TraverseCtx@630efd4
        at mill.api.AggWrapper$Agg$Mutable.append(AggWrapper.scala:99)
        at mill.api.AggWrapper$Agg$Mutable$.$anonfun$from$1(AggWrapper.scala:60)
        at scala.collection.IterableOnceOps.foreach(IterableOnce.scala:576)
        at scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:574)
        at scala.collection.AbstractIterator.foreach(Iterator.scala:1300)
        at mill.api.AggWrapper$Agg$Mutable$.from(AggWrapper.scala:60)
        at mill.api.AggWrapper$Agg$.from(AggWrapper.scala:56)
        at mill.eval.EvaluatorCore.evaluate0(EvaluatorCore.scala:170)
        at mill.eval.EvaluatorCore.$anonfun$evaluate$1(EvaluatorCore.scala:43)
        at scala.util.DynamicVariable.withValue(DynamicVariable.scala:59)
        at mill.eval.EvaluatorCore.evaluate(EvaluatorCore.scala:34)
        at mill.eval.EvaluatorCore.evaluate$(EvaluatorCore.scala:26)
        at mill.eval.EvaluatorImpl.evaluate(EvaluatorImpl.scala:15)
        at mill.main.RunScript$.evaluateNamed(RunScript.scala:38)
        at mill.main.RunScript$.$anonfun$evaluateTasksNamed$2(RunScript.scala:26)
        at scala.util.Either.map(Either.scala:382)
[...]

PR #2959, opened 2024-01-08
Fix issue #2958 - deduplicate (anonymous) tasks in result

lefou:
Anonymous tasks can be defined as `val` or `def`. This is in contrast to named targets, which always need to be defined as `def`.

Also, anonymous tasks end up in the terminal group of the target that's using them.

If the same anonymous task is used in multiple targets, it also ends up multiple time in the task results. When a `Evaluator.Results` is created, this clashes, due to the use of a `Strict.Agg` for `Evalutor.Results.evaluated`.

This change deduplicated the results before constructing the `Evaluator.Results`.

Fix: https://github.com/com-lihaoyi/mill/issues/2958

### DIFF
--- a/main/eval/src/mill/eval/EvaluatorCore.scala
+++ b/main/eval/src/mill/eval/EvaluatorCore.scala
@@ -170,7 +170,7 @@ private[mill] trait EvaluatorCore extends GroupEvaluator {
     EvaluatorCore.Results(
       goals.indexed.map(results(_).map(_._1).result),
       // result of flatMap may contain non-distinct entries,
-      // se we manually clean it up before converting to a `Strict.Agg`
+      // so we manually clean it up before converting to a `Strict.Agg`
       // see https://github.com/com-lihaoyi/mill/issues/2958
       Strict.Agg.from(
         finishedOptsMap.values.flatMap(_.toSeq.flatMap(_.newEvaluated)).iterator.distinct
--- a/main/eval/src/mill/eval/EvaluatorCore.scala
+++ b/main/eval/src/mill/eval/EvaluatorCore.scala
@@ -169,7 +169,11 @@ private[mill] trait EvaluatorCore extends GroupEvaluator {
 
     EvaluatorCore.Results(
       goals.indexed.map(results(_).map(_._1).result),
-      finishedOptsMap.map(_._2).flatMap(_.toSeq.flatMap(_.newEvaluated)),
+      Strict.Agg.from(
+        Loose.Agg.from(
+          finishedOptsMap.values.flatMap(_.toSeq.flatMap(_.newEvaluated))
+        )
+      ),
       transitive,
       getFailing(sortedGroups, results),
       results.map { case (k, v) => (k, v.map(_._1)) }
--- a/main/eval/src/mill/eval/EvaluatorCore.scala
+++ b/main/eval/src/mill/eval/EvaluatorCore.scala
@@ -169,10 +169,11 @@ private[mill] trait EvaluatorCore extends GroupEvaluator {
 
     EvaluatorCore.Results(
       goals.indexed.map(results(_).map(_._1).result),
+      // result of flatMap may contain non-distinct entries,
+      // se we manually clean it up before converting to a `Strict.Agg`
+      // see https://github.com/com-lihaoyi/mill/issues/2958
       Strict.Agg.from(
-        Loose.Agg.from(
-          finishedOptsMap.values.flatMap(_.toSeq.flatMap(_.newEvaluated))
-        )
+        finishedOptsMap.values.flatMap(_.toSeq.flatMap(_.newEvaluated)).iterator.distinct
       ),
       transitive,
       getFailing(sortedGroups, results),
--- a/main/eval/test/src/mill/eval/TaskTests.scala
+++ b/main/eval/test/src/mill/eval/TaskTests.scala
@@ -124,7 +124,7 @@ trait TaskTests extends TestSuite {
       val task1 = T.task { "task1" }
       def task2 = T { task1() }
       def task3 = T { task1() }
-      def com1() = T.command {
+      def command() = T.command {
         val t2 = task2()
         val t3 = task3()
         s"${t2},${t3}"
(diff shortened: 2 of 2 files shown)


main/eval/src/mill/eval/EvaluatorCore.scala
```diff
@@ -169,7 +169,11 @@ private[mill] trait EvaluatorCore extends GroupEvaluator {
 
     EvaluatorCore.Results(
       goals.indexed.map(results(_).map(_._1).result),
-      finishedOptsMap.map(_._2).flatMap(_.toSeq.flatMap(_.newEvaluated)),
+      Strict.Agg.from(
+        Loose.Agg.from(
+          finishedOptsMap.values.flatMap(_.toSeq.flatMap(_.newEvaluated))
+        )
+      ),
       transitive,
```
lihaoyi, line 177: So as far as I can tell, `evaluated: Agg[Task[_]]` is only used in the test suite, and `transitive: Agg[Task[_]]` is not used at all. So this should be fine from a runtime behavior perspective. And anyway it's `private[mill]` so we change things without worrying about compatibility.

As for what the correct thing to do here is, Strict.Agg.from(Loose.Agg.from(` can just be a single `.distinct` call I think. Then maybe just a comment and a simple test case illustrating the edge case and then we should be good to go

lefou:
My new test fails on Windows, but I have no idea why.

```
[#1] [#1] [#3] [#00] mill.eval.SeqTaskTests: [1/3] repro2958.task2 
[#1] [#1] [#3] [#00] mill.eval.SeqTaskTests: [2/3] repro2958.task3 
[#1] [#1] [#3] [#00] mill.eval.SeqTaskTests: [3/3] repro2958.com1 
X mill.eval.SeqTaskTests.duplicateTaskInResult-issue2958 4ms 
  java.nio.file.NoSuchFileException: D:\a\mill\mill\target\workspace\mill\eval\SeqTaskTests\check\duplicateTaskInResult-issue2958\SeqTaskTests$\repro2958\com1.json
    sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:79)
    sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:97)
    sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:102)
    sun.nio.fs.WindowsFileSystemProvider.newByteChannel(WindowsFileSystemProvider.java:230)
    java.nio.file.Files.newByteChannel(Files.java:361)
    os.write$.write(ReadWriteOps.scala:64)
    os.write$over$.apply(ReadWriteOps.scala:142)
    mill.eval.GroupEvaluator.$anonfun$handleTaskResult$2(GroupEvaluator.scala:406)
    mill.eval.GroupEvaluator.$anonfun$handleTaskResult$2$adapted(GroupEvaluator.scala:405)
    scala.Option.foreach(Option.scala:437)
    mill.eval.GroupEvaluator.handleTaskResult(GroupEvaluator.scala:405)
    mill.eval.GroupEvaluator.evaluateGroupCached(GroupEvaluator.scala:226)
    mill.eval.GroupEvaluator.evaluateGroupCached$(GroupEvaluator.scala:45)
    mill.eval.EvaluatorImpl.evaluateGroupCached(EvaluatorImpl.scala:15)
    mill.eval.EvaluatorCore.$anonfun$evaluate0$2(EvaluatorCore.scala:116)
    scala.concurrent.impl.Promise$Transformation.run(Promise.scala:467)
    mill.eval.ExecutionContexts$RunNow$.execute(ExecutionContexts.scala:13)
    scala.concurrent.impl.Promise$Transformation.submitWithValue(Promise.scala:429)
    scala.concurrent.impl.Promise$DefaultPromise.submitWithValue(Promise.scala:338)
    scala.concurrent.impl.Promise$DefaultPromise.dispatchOrAddCallbacks(Promise.scala:312)
    scala.concurrent.impl.Promise$DefaultPromise.map(Promise.scala:182)
    mill.eval.EvaluatorCore.$anonfun$evaluate0$1(EvaluatorCore.scala:92)
    mill.eval.EvaluatorCore.$anonfun$evaluate0$1$adapted(EvaluatorCore.scala:90)
    scala.collection.immutable.Vector.foreach(Vector.scala:2124)
    mill.eval.EvaluatorCore.evaluate0(EvaluatorCore.scala:90)
    mill.eval.EvaluatorCore.$anonfun$evaluate$1(EvaluatorCore.scala:43)
    mill.eval.EvaluatorCore.evaluate(EvaluatorCore.scala:34)
    mill.eval.EvaluatorCore.evaluate$(EvaluatorCore.scala:26)
    mill.eval.EvaluatorImpl.evaluate(EvaluatorImpl.scala:15)
    mill.testkit.MillTestKit$TestEvaluator.apply(MillTestkit.scala:128)
    mill.testkit.MillTestKit$TestEvaluator.apply(MillTestkit.scala:121)
    mill.eval.TaskTests.$anonfun$tests$34(TaskTests.scala:257)
    mill.eval.TaskTests.$anonfun$tests$34$adapted(TaskTests.scala:256)
    mill.eval.SeqTaskTests$.withEnv(TaskTests.scala:271)
    mill.eval.TaskTests.$anonfun$tests$33(TaskTests.scala:137)
```

lihaoyi:
@lefou is it because your target starts with the word `com` which is prohibited on the windows filesystem?

lefou:
> @lefou is it because your target starts with the word `com` which is prohibited on the windows filesystem?

Thanks for the hint. In fact, it's the exact name "com1" that is reserved, although I wonder why the file name extension isn't counted as part of the name.

There are other reserved names, which we should try to detect and handle somehow in Mill, to avoid misleading error messages: `CON, PRN, AUX, NUL, COM0, COM1, COM2, COM3, COM4, COM5, COM6, COM7, COM8, COM9, COM¹, COM², COM³, LPT0, LPT1, LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, LPT9, LPT¹, LPT², and LPT³`

* Source: https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#naming-conventions

Issue #2923, opened 2023-12-14
Mill cache/metadata path for cross segments need to escape colons (`:`) on Windows


lefou:
This is the error on Windows CI for PR https://github.com/com-lihaoyi/mill/pull/2911

(https://github.com/com-lihaoyi/mill/actions/runs/7201802915/job/19619640309)

```
An unexpected error occurred
Exception in thread "main" java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at mill.main.client.IsolatedMillMainLoader.runMain(IsolatedMillMainLoader.java:58)
	at mill.main.client.MillClientMain.main(MillClientMain.java:78)
Caused by: java.nio.file.InvalidPathException: Illegal char <:> at index 26: dummy/com.google.code.gson:gson:2.10.1/scalaVersion
	at sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
	at sun.nio.fs.WindowsPath.parse(WindowsPath.java:94)
	at sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:255)
	at sun.nio.fs.AbstractPath.resolve(AbstractPath.java:53)
	at os.Path.$div(Path.scala:501)
	at mill.eval.EvaluatorPaths$.resolveDestPaths(EvaluatorPaths.scala:31)
	at mill.eval.GroupEvaluator.evaluateGroupCached(GroupEvaluator.scala:177)
	at mill.eval.GroupEvaluator.evaluateGroupCached$(GroupEvaluator.scala:45)
	at mill.eval.EvaluatorImpl.evaluateGroupCached(EvaluatorImpl.scala:15)
	at mill.eval.EvaluatorCore.$anonfun$evaluate0$2(EvaluatorCore.scala:116)
	at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:467)
[...]

PR #2965, opened 2024-01-10
Mask colons in segment paths

lefou:
Again a Windows thing, but the evaluator paths should not contain colons. We mask them with `$colon`.

Fix https://github.com/com-lihaoyi/mill/pull/2923


lihaoyi:
Since we're escaping things with `$`, should we also mask `$` with `$$` to avoid potential conflicts?

lefou:
> Since we're escaping things with `$`, should we also mask `$` with `$$` to avoid potential conflicts?

Can you a bit more specific. What do we mask with `$` and which conflicts do you expect with this concrete `$colon`?

lefou:
@lihaoyi Did you mean the name mangling for private inner classes in Scala? Or something else?

I think the masking I introduce here doesn't need to be bijective. It's only from segment path to file path. So, the question is, will two different segment paths project to the same file path?

lihaoyi:
@lefou oh I wasn't thinking about Scala compiler name mangling, I was more thinking about potential collisions caused by Mill's own mangling. Like if we have two tasks, with names `a:b` and `a$colonb`, they would collide unless we escaped the `$` such that `a$colonb` becomes `a$$colonb`

This isn't an edge case likely to cause any issues in practice, but it's definitely a thing

PR #2967, opened 2024-01-11
Fix outdated ArchLinux package URL

FantasqueX:
ArchLinux community repository has been merged into extra repository, according to https://archlinux.org/news/git-migration-announcement/. Current mill package URL doesn't work.


lefou (commented):
Thanks for updating it. The package version (0.11.0-m0-1) itself looks pretty outdated to me.

Issue #2598, opened 2023-06-12
Provide a way to extend the Mill BSP server


lefou:
The actual use case is a planned Kotlin extension for the BSP protocol, but Mill's Kotlin support is maintained outside the git tree.

* https://github.com/build-server-protocol/build-server-protocol/issues/520
* https://github.com/lefou/mill-kotlin

lihaoyi:
Probably going to call this a wontfix since kotlin seems to work pretty well now

PR #2970, opened 2024-01-14
POC: Make BSP modular and support dynamic extensions

lefou:
Introduce a `BspExtension` trait, which holds a class name and it's potential ivy dependencies.

Split up the various BSP server extensions and instantiate them independently. They are now concrete classes with either a default contructor or one accepting the new `MillBuildServerBase` parameter.

The `BspModule` now has a new `bspExtensions` method, which declares what BSP extension should be used to support the module. The extension is loaded via reflection. `JavaModule` and `ScalaModule` now explicitly declare their extensions. That means, if no `ScalaModule` is part of the project, we also should not load the extension.

* Fix https://github.com/com-lihaoyi/mill/issues/2598

### DIFF
--- a/bsp/src/mill/bsp/BSP.scala
+++ b/bsp/src/mill/bsp/BSP.scala
@@ -2,18 +2,44 @@ package mill.bsp
 
 import mill.api.{Ctx, PathRef}
 import mill.{Agg, T}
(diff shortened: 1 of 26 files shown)


lefou:
* Related discussion: https://github.com/build-server-protocol/build-server-protocol/discussions/651

PR #2971, opened 2024-01-14
Update coursier from 2.1.8 to 2.1.9

lefou:


lefou:
Ups, coursier 2.1.9 is not yet released.

PR #2972, opened 2024-01-14
Update bsp4j from 2.1.0-M7 to 2.1.0

lefou:
Should contain no new changes, but finally a stable version.


lefou:
Looks like the release isn't properly published.

https://github.com/build-server-protocol/build-server-protocol/issues/644

PR #2978, opened 2024-01-17
Don't leak runtime cli imports into generated build scripts

lefou:
Since the cli option `--import` is supposed to affect only the runtime
classpath, there is no need to persist it's value in the generated scripts.

Instead, we hold it in an `DynamicVariable` at script evaluation time.

Pull request: https://github.com/com-lihaoyi/mill/pull/2978

### DIFF
--- a/runner/src/mill/runner/CliImports.scala
+++ b/runner/src/mill/runner/CliImports.scala
@@ -1,0 +1,8 @@
+package mill.runner
+
+import scala.util.DynamicVariable
+
+/**
+ * Hold additional runtime dependencies given via the `--import` cli option.
+ */
+private[runner] object CliImports extends DynamicVariable[Seq[String]](Seq.empty)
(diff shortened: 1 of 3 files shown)


runner/src/mill/runner/CliImports.scala
```diff
@@ -0,0 +1,8 @@
+package mill.runner
+
+import scala.util.DynamicVariable
+
+/**
+ * Hold additional runtime dependencies given via the `--import` cli option.
+ */
+private[runner] object CliImports extends DynamicVariable[Seq[String]](Seq.empty)
```
lihaoyi, line 8: Huh I never knew `extends DynamicVariable` was a thing, but I guess it works

lihaoyi (approved):

runner/src/mill/runner/CliImports.scala
lefou: Yeah, I first had it as a `val` in the companion object of `MillBuildBootstrap`, but `acyclic` wasn't happy with that. So I moved it and tried to avoid all that nesting, without using a `val` in a new `package object` and ended up with this compact thing.

Issue #2926, opened 2023-12-15
BSP: Invalid `textDocument` field in `build/publishDiagnostics` notification


kasiaMarek:
Mill build server sends `build/publishDiagnostics` notification with invalid `textDocument` field, when publishing information about deprecation. It should contain an uri of a text document but is the build target uri instead. This message should most probably just be send via `build/logMessage`.
```
[Trace - 02:45:10 pm] Received notification 'build/publishDiagnostics'
Params: {
  "textDocument": {
    "uri": "file:///Users/kmarek/Desktop/Developer/Tests/scala3-mill-example-project/examples"
  },
  "buildTarget": {
    "uri": "file:///Users/kmarek/Desktop/Developer/Tests/scala3-mill-example-project/examples"
  },
  "diagnostics": [
    {
      "range": {
        "start": {
          "line": 0,
          "character": 0
        },
        "end": {
          "line": 0,
          "character": 0
        }
      },
      "severity": 2,
      "code": "-1",
      "source": "mill",
      "message": "there were 3 deprecation warnings; re-run with -deprecation for details"
    }
  ],
  "reset": false
}
```
mill-version: 0.11.6-24-0770d3

lefou:
Thank you @kasiaMarek for opening this issue. I agree with you. Since the compiler message in question does not belong to a single source file, it's not actionable (in the sense of a published diagnostics), hence we should instead just push a `logMessage`.

I opened PR #2979 to address this.

PR #2979, opened 2024-01-18
BSP: Send `logMessage` instead of diagnostics when `textDocument` is unknown

lefou:
Since some compiler messages do not belong to a single source file, they are not actionable (in the sense of a BSP `build/publishDiagnostics`). In such situation (no text document), we send `build/logMessage` instead.

Fix https://github.com/com-lihaoyi/mill/issues/2926


ckipp01 (approved):
Yup, makes sense 👍🏼

Issue #2818, opened 2023-10-04
BSP: tasks seem to be executed duplicately and concurrently


pieter-bos:
Setup:

```scala
import mill._
import scalalib._
import os._

object tasks extends JavaModule {
    def directory = T.sources {
        Thread.sleep(1000)
        os.write(T.dest / (Thread.currentThread.getId.toString + "-" + System.nanoTime), "hello")
        T.dest
    }

    def resources = T { 
        directory()
    }

    def sources = T { 
        directory()
    }

    def allStuff = T { 
        resources()
        sources()
        ()
    }
}
```

As far as I understand mill's concurrency model: different tasks may be executed in parallel and their order is not guaranteed (other than implied by task dependencies). However, one specific task is always run with a fresh `T.dest` and only once at a time. This seems to be the case if I run e.g. `mill -j 8 tasks.allStuff` here: `directory` is evaluated only once (assumed from run time) and from a fresh `T.dest` (afterwards there is only one file there).

Via BSP it seems this is not the case. I tried importing this build configuration from IntelliJ, and can also recreate this with a [small python script](https://gist.github.com/pieter-bos/7552c0375f1205e631032ddb2367c494). The idea is I request `buildTarget/sources` and `buildTarget/resources` without waiting for a response for the first - as IntelliJ also seems to do. Afterwards I can see that `directory` was run twice in parallel within the same `T.dest`:

```
$ ls out/tasks/directory.dest/
47-13089034942802  49-13089037254006
```

It seems to me that this should never be possible, though I apologize if I'm misunderstanding something :)

pieter-bos:
All the above was tested on `0.11.4`. I notice there's significant changes to BSP stuff in `0.11.5`, but the issue does still reproduce in `0.11.5`.

lefou:
I'd assume, that if the `source` and the `resource` request run in parallel, and as there is no previous cached result, both also need to evaluate the `directory` target (for the first time). Mill has no mechanism to detect and avoid concurrent execution of the same task. This is rarele needed, as the task dependency tree only runs each dependency once, even in parallel mode. But in case you run two Mill processes in parallel or BSP runs multiple requests in parallel, this can become an issue.

lefou:
I opened discussion https://github.com/com-lihaoyi/mill/discussions/2820 to address this

lefou:
I opened PR #2980 with a fix.

PR #2980, opened 2024-01-18
Synchronize evaluateGroupCached to avoid concurrent access to cache

lefou:
With this change, we synchronize evaluations of the same terminal task. This isn't necessarily needed for normal Mill executions, but in a BSP context, where multiple requests where handled concurrently in the same Mill instance, evaluating the same task concurrently  can happen.

We don't synchronize multiple Mill-instances (e.g. run in two shells) or multiple evaluator-instances (which should have different `out`-dirs anyway).

Fix: https://github.com/com-lihaoyi/mill/issues/2818

Pull request: https://github.com/com-lihaoyi/mill/pull/2980

### DIFF
--- a/main/eval/src/mill/eval/KeyLock.scala
+++ b/main/eval/src/mill/eval/KeyLock.scala
@@ -1,0 +1,21 @@
+package mill.eval
+
+private class KeyLock[K]() {
+  private[this] val lockKeys = new java.util.HashSet[K]()
+
+  def lock(key: K): AutoCloseable = {
+    lockKeys.synchronized {
+      while (!lockKeys.add(key)) {
+        lockKeys.wait();
+      }
+    }
+    () => unlock(key)
+  }
+
+  private def unlock(key: K): Unit = {
+    lockKeys.synchronized {
+      lockKeys.remove(key)
+      lockKeys.notifyAll()
+    }
+  }
+}
--- a/main/eval/src/mill/eval/KeyLock.scala
+++ b/main/eval/src/mill/eval/KeyLock.scala
@@ -3,7 +3,7 @@ package mill.eval
 private class KeyLock[K]() {
   private[this] val lockKeys = new java.util.HashSet[K]()
 
-  def lock(key: K,  onCollision: Option[() => Unit] = None): AutoCloseable = {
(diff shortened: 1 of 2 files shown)


lihaoyi (approved):

main/eval/src/mill/eval/KeyLock.scala
```diff
@@ -0,0 +1,26 @@
+package mill.eval
+
+private class KeyLock[K]() {
+  private[this] val lockKeys = new java.util.HashSet[K]()
+
+  def lock(key: K, onCollision: Option[() => Unit] = None): AutoCloseable = {
+    lockKeys.synchronized {
+      var shown = false
+      while (!lockKeys.add(key)) {
```
lihaoyi, line 9: Does this need to be a `while`? I thought we're inside `lockKeys.synchronized`, there should be no need to "try again" since nobody else else going to be able to mutate `lockKeys`

lihaoyi:
Using a file lock to allow mutex between different Mill processes would definitely also be nice. e.g. It's not uncommon for IntelliJ to be running a Mill process and for my shell to be running a Mill process. But intra-process mutex is already an improvement to the status quo so we can go ahead with this first

main/eval/src/mill/eval/KeyLock.scala
lefou: I think yes. At least, the Javadoc of the `Object.wait` method suggests to do so:

> As in the one argument version, interrupts and spurious wakeups are possible, and this method should always be used in a loop:
>
> ```java
> synchronized (obj) {
>     while (<condition does not hold>)
>         obj.wait();
>     ... // Perform action appropriate to condition
> }
> ```

lefou:
> Using a file lock to allow mutex between different Mill processes would definitely also be nice. e.g. It's not uncommon for IntelliJ to be running a Mill process and for my shell to be running a Mill process. But intra-process mutex is already an improvement to the status quo so we can go ahead with this first

I think providing a file-based lock strategy is the next logical step, but as it still needs to be performed on a task-granualarity, it might come with some performance implications. Maybe, we want it to be configurable. Or just use it, if we detect, that there is another Mill running.

PR #2988, opened 2024-01-23
Prepare the Scala.js integration for SIP-51.

sjrd:
When using Scala.js 1.15.0+ together with Scala 2, we now add an explicit dependency to the `scalajs-scalalib` artifact. That artifact comes with Scala.js 1.15.0+, and is double-versioned with the Scala version and the Scala.js version.

As SIP-51 gets implemented and future versions of the Scala library may break forward binary compatibility, the new `scalajs-scalalib` will be able to bring in new APIs back to existing Scala.js versions.

This commit mimics the changes done to sbt-scalajs in https://github.com/scala-js/scala-js/pull/4913/files#diff-adc7ad642b8f89aab3d2a81ac25b23e345bbea758b1c8bbea46cc9d1f351834c

### DIFF
--- a/scalajslib/src/mill/scalajslib/ScalaJSModule.scala
+++ b/scalajslib/src/mill/scalajslib/ScalaJSModule.scala
@@ -218,9 +218,25 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
 
   /** Adds the Scala.js Library as mandatory dependency. */
   override def mandatoryIvyDeps = T {
-    super.mandatoryIvyDeps() ++ Seq(
-      ivy"org.scala-js::scalajs-library:${scalaJSVersion()}".withDottyCompat(scalaVersion())
-    )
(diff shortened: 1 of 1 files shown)


lefou:
SIP-51: https://docs.scala-lang.org/sips/drop-stdlib-forwards-bin-compat.html

lefou:
Thank you for opening this pull request, @sjrd . It looks reasonable.

What's the next step? Should we merge it right away or wait for some state change / announcement of SIP-51?

sjrd:
It can be merged right away. sbt-scalajs 1.15.0 already behaves this way.

It is already beneficial irrespective of SIP-51 (though with much smaller consequences): if there is a bugfix in a newer version of the standard library, users of the new Scala version will get the bug fix, instead of having to wait for a newer version of Scala.js that has itself upgraded to the newer Scala.

PR #2990, opened 2024-01-26
Bump MainArgs to 0.6.1

lihaoyi:
Also updated the bash launchers' special casing of `-i` to allow for it to be part of a combined token, as long as the combined token starts with `-i`

With the `installLocalCache`ed version of Mill, I can now run:

```bash
$ ./mill -ikw contrib.__.compile
```

and the flags `-i`, `-k`, and `-w` will all take effect independently

The auto-kebab-case name mapping applies to Mill's own CLI args, the args of user-defined `T.command`s, but _not_ the names of the target/command themselves. e.g. you can call;

- `./mill --noServer core.ivyDepsTree --withCompile --withRuntime` or
- `./mill --no-server core.ivyDepsTree --with-compile --with-runtime`

but you cannot call 
- `./mill --no-server core.ivy-deps-tree --with-compile --with-runtime`

It should be easy to add name-mapping for the target selector as well, or I could also disable the name-mapping for user-define command arguments if we decide that's not the right thing to do. I'm not sure TBH what the right thing to do here is; bash naming conventions is fundamentally incompatible with Scala naming conventions, so regardless what Mill decides there will be some inconsistencies between:

- Bash naming conventions
- Mill's own CLI flag naming conventions
- Mill CLI task naming conventions
- Mill CLI command argument naming conventions
- Mill Scala object/method naming conventions
- Mill Scala method argument naming conventions
- Scala object/method naming conventions
- Scala method argument naming conventions

I guess it just depend where we want to draw the line and tolerate the inconsistency. I feel like maybe something like `./mill --no-server core.ivy-deps-tree --with-compile --with-runtime` would be the ideal outcome, where everything in Bash-land is consistent and everything in Scala-land is consistent.

Pull request: https://github.com/com-lihaoyi/mill/pull/2990

### DIFF
--- a/main/resolve/src/mill/resolve/Resolve.scala
+++ b/main/resolve/src/mill/resolve/Resolve.scala
@@ -146,7 +146,8 @@ object Resolve {
       flattenedArgSigsWithDefaults,
       allowPositional = true,
       allowRepeats = false,
-      allowLeftover = ep.argSigs0.exists(_.reader.isLeftover)
+      allowLeftover = ep.argSigs0.exists(_.reader.isLeftover),
+      nameMapper = mainargs.Util.kebabCaseNameMapper
     ).flatMap { (grouped: TokenGrouping[_]) =>
       val mainData = ep.asInstanceOf[MainData[Any, Any]]
       val mainDataWithDefaults = mainData
--- a/main/resolve/src/mill/resolve/Resolve.scala
+++ b/main/resolve/src/mill/resolve/Resolve.scala
@@ -148,7 +148,7 @@ object Resolve {
       allowRepeats = false,
       allowLeftover = ep.argSigs0.exists(_.reader.isLeftover)
     ).flatMap { (grouped: TokenGrouping[_]) =>
-      val mainData = ep.asInstanceOf[MainData[_, Any]]
+      val mainData = ep.asInstanceOf[MainData[Any, Any]]
       val mainDataWithDefaults = mainData
         .copy(argSigs0 = mainData.argSigs0.map(withNullDefault))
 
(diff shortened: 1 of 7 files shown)


main/resolve/src/mill/resolve/Resolve.scala
```diff
@@ -148,7 +148,7 @@ object Resolve {
       allowRepeats = false,
       allowLeftover = ep.argSigs0.exists(_.reader.isLeftover)
     ).flatMap { (grouped: TokenGrouping[_]) =>
-      val mainData = ep.asInstanceOf[MainData[_, Any]]
+      val mainData = ep.asInstanceOf[MainData[Any, Any]]
```
lihaoyi, line 151: Not sure why this is necessary, something about type inference must have changed in 0.6.0 to make it source incompatible (even though it's binary compatible)

main/test/src/mill/main/MainModuleTests.scala
```diff
@@ -161,7 +161,7 @@ object MainModuleTests extends TestSuite {
       test("command") {
         val Left(Result.Failure(failureMsg, _)) = evaluator.evalTokens("show", "helloCommand")
         assert(
-          failureMsg.contains("Expected Signature: helloCommand"),
+          failureMsg.contains("Expected Signature: hello-command"),
```
lihaoyi, line 164: Note that this change to the error message does not mean the original `helloCommand` syntax stops working; the test (below) continues to use `helloCommand`. 

With mainargs 0.6.0, both `helloCommand` and `hello-command` are accepted, but we put the dashed version `hello-command` in the help message by default (unless a different nameMapper is given)

lefou (approved):

PR #3008, opened 2024-02-10
[POC] Use Unroll plugin for binary compatibility

lihaoyi:


contrib/scoverage/api/src/mill/contrib/scoverage/api/ScoverageReportWorkerApi.scala
```diff
@@ -1,35 +1,19 @@
...
-  ): Unit = {
-    report(reportType, sources, dataDirs, ctx.workspace)
-  }
-
   def report(
       reportType: ReportType,
       sources: Seq[os.Path],
       dataDirs: Seq[os.Path],
-      sourceRoot: os.Path
+      @Unroll sourceRoot: os.Path = null
```
lefou, line 13: Can you elaborate how the use of the unroll plugin changes the code? Will API users who freshly compile against the binary code, always use the new API? Are the backward-compatible stubs only for binary compat, or are they still public non-deprecated API?

build.sc
```diff
@@ -364,11 +364,13 @@ trait MillScalaModule extends ScalaModule with MillJavaModule with ScalafixModul
   def scalacPluginIvyDeps =
     super.scalacPluginIvyDeps() ++
       Agg(Deps.acyclic) ++
+      Agg(ivy"com.lihaoyi::unroll-plugin:0.1.9") ++
```
lefou, line 367: We should move this into `Deps.unroll` or `Deps.scalac.unroll`.

contrib/scoverage/api/src/mill/contrib/scoverage/api/ScoverageReportWorkerApi.scala
lihaoyi: @lefou it generates bincompat forwarder overloads. The plugin repo has some description and links to further discussion. I hope to get this into the Scala language via a SIP

https://github.com/com-lihaoyi/unroll

For now, the generated forwarders aren't hidden. It doesn't really matter, since it's just a forwarder, calling it does the same thing as calling the "real" definition. There's some discussion on how to hide them, but nothing really decided upon

lefou: Understood. Would be nice if there is an option to mark the forwarders as deprecated, since then, it's in the hand of the API user to hack around them and produce bytecode that may be more safe, if needed. But it's not needed here, I think.

Issue #3011, opened 2024-02-14
Make BSP import more robust


nafg:
(Previously posted at https://discord.com/channels/632150470000902164/940067748103487558/1207105622366228480)

With a project imported from Mill into IntelliJ via BSP, I regularly have scenarios like the following:

1. I update dependencies and sync (reimport the project)
2. It fails with something cryptic like this: ![image](https://github.com/com-lihaoyi/mill/assets/98384/11735e81-ea95-4318-8db6-852ac823bbbc)
3. Now the entire project is red, so the IDE is very unheplful
4. I check mill-bsp.stderr and see something like
```
buildTargetScalacOptions caught exception: java.lang.Exception: Failure during task evaluation: app_common.test.compile Compilation failed
java.lang.Exception: Failure during task evaluation: app_common.test.compile Compilation failed
    at mill.eval.Evaluator.$anonfun$evalOrThrow$default$1$1(Evaluator.scala:42)
    at mill.eval.EvaluatorImpl$EvalOrThrow.apply(EvaluatorImpl.scala:67)
```
In short, unless there are no compile errors anywhere when I sync the build it breaks the whole IDE with no easy to find reason

This does not happen with SBT BSP import.

Suggestions:
A. If importing fails, it shouldn't make things worse in the IDE than before
B. A module failing to compile is an an exception, it should be a case the BSP code thinks about
C. Ideally it would be able to give IDE whatever information it can
D. Errors shouldn't be hidden somewhere inside a ginormous log file, they should be returned to the IDE

nafg:
Another thing is that often after reimporting -Xsource:3 syntax like `*` imports are red until I do Repair IDE and get past Reopen Project (I confess, I'm not sure if I tried going straight to manually closing and reopening the project).

lihaoyi:
I've hit this too. We should not need to compile everything to generate the IDE project

I vaguely remember fixing a bug like this before, but maybe I'm mistaken or it regressed

lihaoyi:
In my case I suspect it's because I'm doing unusual things, e.g. compiling compiler plugins as part of the build, or depending on output `.jar` targets as part of `unmanagedClasspath`. Doing that, compiling stuff when generating the IDE project is unavoidable

@nafg is there any chance you are doing something similar?

lefou:
Would be good to have a reproducer for that.

lihaoyi:
[...]

PR #3022, opened 2024-02-19
Make BSP task processing more resilient

lefou:
Since BSP clients can handle incomplete result lists (e.g. they request a task on 5 build targets, but only receive the results for 3 of them), we just don't include the failed targets in the final request result. This is also what sbt and Bloop does to handle failures.

In addition, we send a log message to the server, so the failure doesn't go unnoticed.

Fix: https://github.com/com-lihaoyi/mill/issues/3011

Pull request: https://github.com/com-lihaoyi/mill/pull/3022


nafg:
How do I test it?

PR #3048, opened 2024-02-23
commons-compress: 1.25.0 → 1.26.0

fabianhjr:
Direct dependency of plexus-archiver via coursier

Already updated upstream pending a release: https://github.com/codehaus-plexus/plexus-archiver/commit/6077fd7385965710ed8dd5918acf6c2079497e9d

https://mvnrepository.com/artifact/org.apache.commons/commons-compress/1.25.0 https://mvnrepository.com/artifact/org.apache.commons/commons-compress/1.26.0

Addresses:

- CVE-2024-25710

PR #3049, opened 2024-02-24
Force transitive dependency commons-compress to 1.26.0 or newer

lefou:
This is to fix security issue CVE-2024-25710.

Supersedes https://github.com/com-lihaoyi/mill/pull/3048


fabianhjr:
ty <3

PR #3051, opened 2024-02-25
Fix some warnings

nightscape:
I'm not sure if some of those warnings were intentional.
If so, please let me know and I'll take them out of the PR.

### DIFF
--- a/main/codesig/src/JvmModel.scala
+++ b/main/codesig/src/JvmModel.scala
@@ -3,6 +3,7 @@ import scala.collection.mutable.LinkedHashMap
 import upickle.default.{ReadWriter, readwriter, stringKeyRW}
 
 import scala.annotation.switch
+import scala.collection.immutable.ArraySeq
 
 // This file contains typed data structures representing the types and values
 // found in the JVM bytecode: various kinds of types, method signatures, method
@@ -223,7 +224,10 @@ object JvmModel {
         args.addOne(JType.read(s.substring(index, split + 1)))
         index = split + 1
       }
-      new Desc(args.result(), JType.read(s.substring(closeParenIndex + 1)))
+      new Desc(
+        ArraySeq.unsafeWrapArray(args.result()),
+        JType.read(s.substring(closeParenIndex + 1))
+      )
     }
 
     implicit val ordering: Ordering[Desc] = Ordering.by(_.pretty)
--- a/main/codesig/test/cases/callgraph/realistic/3-par-merge-sort/src/Hello.scala
+++ b/main/codesig/test/cases/callgraph/realistic/3-par-merge-sort/src/Hello.scala
@@ -3,11 +3,12 @@ package hello
 // Taken from https://github.com/handsonscala/handsonscala/blob/ebc0367144513fc181281a024f8071a6153be424/examples/13.7%20-%20ParallelMergeSort/MergeSort.sc
 import mainargs.main
 
-object Main{
+object Main {
 
   import scala.concurrent._, duration.Duration.Inf, java.util.concurrent.Executors
 
-  implicit val ec = ExecutionContext.fromExecutorService(Executors.newFixedThreadPool(8))
+  implicit val ec: ExecutionContextExecutorService =
+    ExecutionContext.fromExecutorService(Executors.newFixedThreadPool(8))
 
   def mergeSortParallel[T: Ordering](items: IndexedSeq[T]): IndexedSeq[T] = {
     Await.result(mergeSortParallel0(items), Inf)
@@ -89,4 +90,4 @@ object Main{
         "hello.Main$#mergeSortSequential(scala.collection.immutable.IndexedSeq,scala.math.Ordering)scala.collection.immutable.IndexedSeq"
     ]
 }
-*/
+ */
--- a/main/codesig/test/cases/callgraph/realistic/3-par-merge-sort/src/Hello.scala
+++ b/main/codesig/test/cases/callgraph/realistic/3-par-merge-sort/src/Hello.scala
@@ -90,4 +90,4 @@ object Main {
         "hello.Main$#mergeSortSequential(scala.collection.immutable.IndexedSeq,scala.math.Ordering)scala.collection.immutable.IndexedSeq"
     ]
 }
- */
+*/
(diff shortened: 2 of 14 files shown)


contrib/scoverage/api/src/mill/contrib/scoverage/api/ScoverageReportWorkerApi.scala
```diff
@@ -28,7 +28,7 @@ trait ScoverageReportWorkerApi {
     ctx.log.error(
       "Binary compatibility stub may cause infinite loops with StackOverflowError. You need to implement: def report(ReportType, Seq[Path], Seq[Path], os.Path): Unit"
     )
-    report(reportType, sources, dataDirs)
+    report(reportType, sources, dataDirs, ctx.workspace)
```
lefou, line 31: This doesn't look right to me. At least in the browser. Your change make the explicit code comment incorrect. If I read this correctly, this report method is meant to be implemented by downstream classes, but to preserve binary compatibility, we can't make it abstract. Hence, we need to call the versions which is guaranteed to be implemented (which is the deprecated one) to avoid any kind of runtime failure.

main/codesig/src/JvmModel.scala
```diff
@@ -223,7 +223,10 @@ object JvmModel {
         args.addOne(JType.read(s.substring(index, split + 1)))
         index = split + 1
       }
-      new Desc(args.result(), JType.read(s.substring(closeParenIndex + 1)))
+      new Desc(
+        collection.immutable.ArraySeq.unsafeWrapArray(args.result()),
```
lefou, line 227: If this is `scala.collection.immutable.ArraySeq`, then I think it's reasonable to just import it at the top of the file.

scalajslib/test/src/mill/scalajslib/FullOptESModuleTests.scala
```diff
@@ -29,7 +29,7 @@ object FullOptESModuleTests extends TestSuite {
 
     test("fullOpt with ESModule moduleKind") {
       val result =
-        fullOptESModuleModuleEvaluator(FullOptESModuleModule.fullOptESModuleModule.fullOpt)
+        fullOptESModuleModuleEvaluator(FullOptESModuleModule.fullOptESModuleModule.fullLinkJS)
```
lefou, line 32: Hard to tell, whether this was intentional. But if we are going to test `fullLinkJS` instead of the deprecated `fullOpt`, we should also rename the test case and the involved `val`s and `object`s. They currently all contain the `fullOpt`.

scalajslib/test/src/mill/scalajslib/HelloJSWorldTests.scala
```diff
@@ -124,7 +124,7 @@ object HelloJSWorldTests extends TestSuite {
       val module = HelloJSWorld.helloJsWorld(scalaVersion, scalaJSVersion)
       val jsFile =
         if (legacy) {
-          val task = if (optimize) module.fullOpt else module.fastOpt
+          val task = if (optimize) module.fullLinkJS else module.fastLinkJS
```
lefou, line 127: This seems to be wrong. I think the `legacy: Boolean` switch is exactly there to test the `fullOpt` and `fastOpt` targets.

scalajslib/test/src/mill/scalajslib/MultiModuleTests.scala
```diff
@@ -39,7 +39,7 @@ object MultiModuleTests extends TestSuite {
     prepareWorkspace()
 
     def checkOpt(optimize: Boolean) = {
-      val task = if (optimize) MultiModule.client.fullOpt else MultiModule.client.fastOpt
+      val task = if (optimize) MultiModule.client.fullLinkJS else MultiModule.client.fastLinkJS
```
lefou, line 42: Doesn't look right. The test cases, that call `checkOpt` are named `fastOpt` and `fullOpt`.

lefou (changes requested):
Thank you for cleaning the code base! I commented some lines. Some changes seem to be not right. If a test case says it's testing some target, then we should ensure it's calling that target, even if it is deprecated.

nightscape:
I addressed your comments and rebased on the latest `main`.
Let's see what the tests say 😉

nightscape:
@lefou I guess the tests failing on some platforms are unrelated to my changes?
If so, this would be good to merge from my side 😃

lefou:
I think this test error is not unrelated.

```
 X mill.codesig.CallGraphTests.realistic.3-par-merge-sort 38ms 
  ujson.ParseException: expected whitespace or eof got "*" at index 2817
    ujson.CharParser.die(CharParser.scala:98)
    ujson.CharParser.parse(CharParser.scala:77)
    ujson.StringParser$.transform(StringParser.scala:28)
    ujson.StringParser$.transform(StringParser.scala:27)
    ujson.Readable$fromTransformer.transform(Readable.scala:19)
    upickle.Api.$anonfun$read$1(Api.scala:42)
    upickle.core.TraceVisitor$.withTrace(TraceVisitor.scala:18)
    upickle.Api.read(Api.scala:42)
    upickle.Api.read$(Api.scala:41)
    upickle.default$.read(Api.scala:238)
    mill.codesig.CallGraphTests$.$anonfun$parseJson$8(CallGraphTests.scala:191)
    scala.Option.map(Option.scala:242)
    mill.codesig.CallGraphTests$.parseJson(CallGraphTests.scala:191)
    mill.codesig.CallGraphTests$.testDirectCallGraph(CallGraphTests.scala:112)
    mill.codesig.CallGraphTests$.testExpectedCallGraph(CallGraphTests.scala:97)
    mill.codesig.CallGraphTests$.$anonfun$tests$59(CallGraphTests.scala:13)
```

main/codesig/test/cases/callgraph/realistic/3-par-merge-sort/src/Hello.scala
```diff
@@ -89,4 +90,4 @@ object Main{
         "hello.Main$#mergeSortSequential(scala.collection.immutable.IndexedSeq,scala.math.Ordering)scala.collection.immutable.IndexedSeq"
     ]
 }
-*/
+ */
```
lefou, line 93: This inserted whitespace could be the issue of the test failure.

lefou:
```suggestion
*/
```

nightscape: Ah, looks like the IntelliJ code formatter has automatically formatted the file...
Your change seems to have fixed it 👍

lefou (approved):
Looks good to me.

PR #3064, opened 2024-03-05
Restructure `TestModule`, add `RunModule` 

lefou:
This idea of this change is to restructure the `TestModule` to decouple it from the concept of compilation. After this change, only a `runClasspath` and a `testClasspath` is needed to run tests. This makes adding test modules for the sake of using a different test framework a breeze.

See the following example: the `test2` module is an additional test framework on the same classes of `test`.

```scala
import mill._
import mill.scalalib._
import mill.scalalib.api.CompilationResult

object foo extends RootModule with ScalaModule {
  def scalaVersion = "2.13.11"
  def ivyDeps = Agg(
    ivy"com.lihaoyi::scalatags:0.12.0",
    ivy"com.lihaoyi::mainargs:0.6.2"
  )

  object test extends ScalaTests {
    def ivyDeps = Agg(
      ivy"com.lihaoyi::utest:0.7.11",
      ivy"org.scalatest::scalatest-freespec:3.2.18"
    )
  }
  object test2 extends TestModule with TestModule.ScalaTest {
    override def compile: T[CompilationResult] = ???
    override def runClasspath: T[Seq[PathRef]] = foo.test.runClasspath()
    override def testClasspath = foo.test.testClasspath()
  }
}
```

Please note the `compile` target is a legacy to our binary-compatibility promise. The target is not used directly in `TestModule`.

This pull request additionally contains the following changes:

* Introduces a new `RunModule` and moved some `run`-related task previously in `TestModule` up.
* Extend `RunModule` in `JavaModule` to share run-releated targets and resolve super-hierarchy
* Introduces a `WithZincWorker` as a shared base trait to resolve super-hierarchies for using and overriding a common worker.

I plan to move more run-releated target from `JavaModule` to `RunModule` in  a subsequent PR. (See https://github.com/com-lihaoyi/mill/pull/3090)

See also the following discussion:
* https://github.com/com-lihaoyi/mill/discussions/3076


lefou:
@lihaoyi , @lolgab , do you have any concerns about this pull request?

lihaoyi:
Looks good to me. I've needed this before: apart from the testFramework use case discussed earlier, I've needed to share the same compiled code but testing it with different flags or environment variables. It's possible to do today with a bunch of `override` hacks, but having `TestModule` be separated out is definitely a lot cleaner

lefou:
> Looks good to me. I've needed this before: apart from the testFramework use case discussed earlier, I've needed to share the same compiled code but testing it with different flags or environment variables. It's possible to do today with a bunch of `override` hacks, but having `TestModule` be separated out is definitely a lot cleaner

That's probably even better covered by my other PR which is in the making:
* https://github.com/com-lihaoyi/mill/pull/3090

It's exactly for the case, when you want run with different flags, envs, or main classes or classpaths.

chikei:
in case someone else trying out similar setup, you also need `override def bspBuildTarget = super.bspBuildTarget.copy(canCompile = false, canRun = false, languageIds = Seq.empty)` in the additional test module to prevent `BuildTarget{Scalac, Javac}Options` bsp requests which mill will eventually calls the dummy compile method and cause unimplemented exception.

lefou:
> in case someone else trying out similar setup, in IJ you also need `override def bspBuildTarget = super.bspBuildTarget.copy(canCompile = false, canRun = false, languageIds = Seq.empty)` in the additional test module to prevent `BuildTarget{Scalac, Javac}Options` bsp requests which mill will eventually calls the dummy compile method and cause unimplemented exception.

Nice catch, @chikei ! That was an oversight.

chikei:
Honestly I am not really sure if tweaking `canCompile` and `canRun` matters, IJ's [code](https://github.com/JetBrains/intellij-scala/blob/idea241.x/bsp/src/org/jetbrains/bsp/project/importing/BspProjectResolver.scala#L348) seems only care language id, but IJ's bsp import UI flow is so annoy I just don't bother to reduce.

As for `canCompile = true`, that's due to I used some other common trait which extends SbtModuleTests instead of the minimal `TestModule with TestModule.ScalaTest` showed here.

lefou:
> Honestly I am not really sure if tweaking `canCompile` and `canRun` matters, IJ's [code](https://github.com/JetBrains/intellij-scala/blob/idea241.x/bsp/src/org/jetbrains/bsp/project/importing/BspProjectResolver.scala#L348) seems only care language id, but IJ's bsp import UI flow is so annoy I just don't bother to reduce.

We currently don't check the `canCompile` before running the compilation in each module, we just communicate it to the client. I think the client is supposed to not request compilation for that module at all.

> As for `canCompile = true`, that's due to I used some other common trait which extends SbtModuleTests instead of the minimal `TestModule with TestModule.ScalaTest` showed here.

Ok. So I think if you stay with the simple example from this PR, `canCompile` should already be set to `false`. If that fails in IJ too, it's and issue worth to be reported there.

If you extends from an real language module like `SbtModuleTests`, which is set up to compile files, you either need to tweak the BSP settings when you customize the `compile` target or you need to implement the `compile` in a more defensive way. The refactoring of this PR is exactly the solution to this issue.

chikei:
So I tried defining additional test without an real language module,  [this line in mill](https://github.com/com-lihaoyi/mill/blob/main/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala#L215) will throw a MatchError as the additional test module does not inherent `JavaModule`.

lefou:
> So I tried defining additional test without an real language module, [this line in mill](https://github.com/com-lihaoyi/mill/blob/main/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala#L215) will throw a MatchError as the additional test module does not inherent `JavaModule`.

This is in fact a bug. I opened #3110 to track it.

chikei:
So with PR #3111, I can make IJ import without error, but build project will fail as mill return `statusCode = ERROR`.

Skimming through IJ's BSP code, my understanding is that IJ ignores `BuildTargetCapabilities` properties returned from `WorkspaceBuildTargets` call, therefore build project in IJ results a `BuildTargetCompile` call with all `BuildTarget` from`WorkspaceBuildTargets`, and mill [reports failure](https://github.com/com-lihaoyi/mill/blob/main/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala#L436) for non JavaModule.

lefou:
@chikei Thanks for the feedback.

PR #3066, opened 2024-03-08
Add 2.13.13 and 2.12.19 to supported bridges list

lolgab:
The bridges are already on Maven Central.

Pull Request: https://github.com/com-lihaoyi/mill/pull/3066

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -252,6 +252,7 @@ val bridgeScalaVersions = Seq(
   "2.12.16",
   "2.12.17",
   "2.12.18",
+  "2.12.19",
   /*"2.13.0", "2.13.1", "2.13.2",*/ "2.13.3",
   "2.13.4",
   "2.13.5",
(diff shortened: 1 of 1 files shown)


lefou (commented):
They aren't needed anywhere except when trying to publish them with the `all` value in `MILL_COMPILER_BRIDGE_VERSIONS` env var, right?

It would be good, if we could set this env var in some CI job, to cover any issues with bridge modules.

lolgab:
No, they are used by mill to decide if try to get the precompiled bridge via coursier, or to compile the bridge in place. With current mill if you use Scala `2.13.13` you see the `Compiling compiler  bridge` log.
Ideally IMO, we would try anyways, and then in case we get a not found from coursier mill should fallback to compile the bridge in place. Otherwise you lose the precompiled bridge if you don't update Mill to the latest snapshot when a new version of Scala is released.

lolgab:
No, they are used by mill to decide if try to get the precompiled bridge via coursier, or to compile the bridge in place. With current mill if you use Scala `2.13.13` you see the `Compiling compiler  bridge` log.
Ideally IMO, we would try anyways, and then in case we get a not found from coursier mill should fallback to compile the bridge in place. Otherwise you lose the precompiled bridge if you don't update Mill to the latest snapshot when a new version of Scala is released.

lefou:
That means, no older Mill version is using the bridges. I also think we should always try to download them.

lefou (approved):

Issue #3068, opened 2024-03-08
Tasty files are not deleted


lolgab:
When deleting a class in Scala `3.4.0`, the depending files don't stop seeing the deleted class.
Checking the `out/compile.dest/classes` directory inside I noticed that the class files are deleted while the tasty files are still there.

### Reproduction
```scala
// build.sc
import mill._
import scalalib._

object root extends RootModule with ScalaModule {
  def scalaVersion = "3.4.0"
}
```

```scala
// src/file1.scala
def foo = 1
```

```scala
// src/file2.scala
def bar = foo
```

Run:
```
mill -w compile
```

Delete or comment `file1.scala`

The compilation succeeds.

Switching to Scala `3.3.3` the compilation fails as it should be.

lefou:
Sounds like an upstream issue in zinc to me.

lolgab:
Sbt works fine though.

lefou:
That's surprising, since Mill isn't handling any kind of class file removal based on source fies. All it does is ensuring the classes directory exists when invoking zinc API with a target path (dir for class files or JAR file) and zinc is managing it's state. Also, the same code path in Mill is used for compilation with Scala 3.3.3 and 3.4.0, so it should behave equally with both.

lefou:
> Sbt works fine though.

@lolgab Which sbt version, which zinc version?

lefou:
It could be related to the fact, that we are including the current compilation output directory in the classpath we feed to the compiler. Don't know whether this is supposed to be the correct invocation, but it worked for a long time.

lolgab:
Wow, that was a very good catch!
Removing `++ Some(classesDir)` fixes the problem 😲 
👏

lefou:
Would be interesting to check how others invoke zinc.

lefou:
It's worth to mention that class file removal worked as expected. It's only the `tasty` file(s), which don't get deleted since scalac 2.4.0.

Compilation with `file1.scala`:

```
scratch-3068-class-removal/out/compile.dest/classes/
scratch-3068-class-removal/out/compile.dest/classes/file2$package.class
scratch-3068-class-removal/out/compile.dest/classes/file1$package.tasty
scratch-3068-class-removal/out/compile.dest/classes/file1$package$.class
scratch-3068-class-removal/out/compile.dest/classes/file1$package.class
scratch-3068-class-removal/out/compile.dest/classes/file2$package$.class
[...]

PR #3071, opened 2024-03-09
Don't include the compilation output in the compile classpath

lefou:
Fix https://github.com/com-lihaoyi/mill/issues/3068

But we need to make sure this is the right thing to do. We include the classes dir in the classpath since very early.


lefou:
This change break a lot of tests. Also, PR https://github.com/com-lihaoyi/mill/pull/3072 which fixes the left-over tasty files issue, provides a test case that shows that compilation fails properly when tasty-files are not present. So I'm closing this as not correct and needed.

PR #3072, opened 2024-03-09
Pass auxiliary class files to zinc so they are deleted together

lolgab:
Fixes https://github.com/com-lihaoyi/mill/issues/3068
`*.tasty`, `*.nir` and `*.sjsir` files were not cleaned up by zinc when the companion class files were deleted.
This passes the right `AuxiliaryClassFiles` to zinc.

Pull Request: https://github.com/com-lihaoyi/mill/pull/3072


scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
```diff
@@ -496,6 +504,25 @@ class ZincWorkerImpl(
...
       .toArray
 
+    val incOptions = {
+      val platformSpecificFiles: Array[AuxiliaryClassFiles] = platformSuffix match {
+        case s"_sjs$_" => Array(ScalaJSFiles.instance())
+        case s"_native$_" => Array(ScalaNativeFiles.instance())
+        case _ => Array.empty[AuxiliaryClassFiles]
+      }
+
+      val scalaVersionSpecificFiles: Array[AuxiliaryClassFiles] =
```
lefou, line 514: Are auxiliary class files expected to exist? Can't we just always set them for tasty? Is there some API documentation for this?

lefou (commented):
I think the tests should also assert the success or failure of the compile task, since we have a strong expectation, that the second run should fail.

scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
lefou: I even think we should not hard-code this setting here, but source it from the actual modules. I'm thinking about semanticdb files for examples, but theoretically, every compiler plugin could provide extra files which should be cleaned up in some way.

lolgab: > I'm curious, can't we just always set them for tasty?

We probably could. I think it was designed this way to avoid unnecessary work when we are certain a specific file can't exist.
Maybe @adpi2 can answer this, since I believe he designed the `AuxiliaryClassFiles` abstraction.

lolgab:
> I think the tests should also assert the success or failure of the compile task, since we have a strong expectation, that the second run should fail.

To not make other tests more complex since they used a single file, I added a separate test that writes another source file in place and asserts that the compilation first succeeds and then fails after deleting the `foo.scala` file.

scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
lefou: > I even think we should not hard-code this setting here, but source it from the actual modules.

This will also change the API. So, to avoid a bunch of new `@deprecated` overloads, we probably want replace the new `platformSuffix` parameter with some to configure the auxiliary files. (These long parameter lists scream for a config class, but we can do that later.)

lolgab: I tried a new approach by adding `protected def auxiliaryClassFileExtensions: T[Seq[String]]` to `ScalaModule` and overriding it in `ScalaJSModule` and `ScalaNativeModule`.

scalalib/src/mill/scalalib/ScalaModule.scala
```diff
@@ -549,6 +549,14 @@ trait ScalaModule extends JavaModule with TestModule.ScalaModuleBase { outer =>
     else ZincWorkerUtil.scalaBinaryVersion(scalaVersion())
   }
 
+  /**
+   * Files extensions that need to be managed by Zinc together with class files
+   */
+  protected def auxiliaryClassFileExtensions: T[Seq[String]] = T {
```
lefou, line 555: I think we should declare it in `JavaModule`, since we also use Zinc for Java-compilation. Although I currently can't tell if there is a need, wince we use zinc, we in fact run incremental Java compilation, and in it's nature we need a way to control auxiliary files removal.

scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
lefou: Great. I like that there is now less knowledge about specific Scala toolchains in zinc worker. Having it explicitly in the respective modules make them more discoverable, easier to remember when reviewing or refactoring and also more modular,since you don't need to think about Native when implementing JS.

lefou: About the `protected` keyword: From a bin-compat standpoint, it's not a difference. With `public` targets, we gain the ability to easily inspect a module from cli and from other modules. We're not so much designing a Scala API here but rather a templating a build module. In this context I think encapsulation it less a property, to the contrary. E.g. if we want to transition a build from `ModuleA` to `ModuleB`, it isn't helpful if part of the settings forwarding to the underlying used tools are not readable.

scalalib/src/mill/scalalib/ScalaModule.scala
lolgab: Done ✅

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -1006,6 +1006,11 @@ trait JavaModule
 
   def forkWorkingDir: T[Path] = T { T.workspace }
 
+  /**
+   * Files extensions that need to be managed by Zinc together with class files
```
lefou, line 1010: This means essentially, zinc will remove them (but not create to rename/move them) when a equally named class files is about to be removed, right?

```suggestion
   * Files extensions that need to be managed by Zinc together with class files.
   * This means, if zinc needs to remove a class file, it will also remove files
   * which match the class file basename and a listed file extensions.
```

scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
lolgab: Made it public ✅

scalalib/api/src/mill/scalalib/api/ZincWorkerApi.scala
```diff
@@ -105,7 +108,40 @@ trait ZincWorkerApi {
       scalacPluginClasspath = scalacPluginClasspath,
       reporter = reporter,
       reportCachedProblems = reportCachedProblems,
-      incrementalCompilation = true
+      incrementalCompilation = true,
+      auxiliaryClassFileExtensions = Seq.empty[String]
+    )
+
+  /** Compile a mixed Scala/Java or Scala-only project */
+  @deprecated("Use override with `platformSuffix` parameter", "Mill 0.11.8")
```
lefou, line 116:
```suggestion
  @deprecated("Use override with `auxiliaryClassFileExtensions` parameter", "Mill 0.11.8")
```

lefou (commented):
Looks good to me. Just two small doc improvements.

scalalib/api/src/mill/scalalib/api/ZincWorkerApi.scala
lefou: @lolgab Looks like this got lost duering your last force-push.

scalalib/src/mill/scalalib/JavaModule.scala
lefou: @lolgab Looks like this got lost during your last force-push.

scalalib/api/src/mill/scalalib/api/ZincWorkerApi.scala
lolgab: Yeah, messed up with the branches. Pushed it again.

scalalib/src/mill/scalalib/JavaModule.scala
lolgab: Yeah, messed up with the branches. Pushed it again.

lefou (approved):
Looks good to me. Thank you!

PR #3078, opened 2024-03-11
Use semanticdb files directly in ScalaModule

lolgab:
A copy mechanism was introduced in https://github.com/com-lihaoyi/mill/pull/2227 because of a bug in semanticdb-java
But we don't need to use it in ScalaModule as well if the Scala semanticdb works correctly.


lefou:
There are test failures.

Also, I'd like to get PR https://github.com/com-lihaoyi/mill/pull/3080 in, which brings additional test for incremental semanticDB generation.

lefou (commented):
The copy operation isn't the only functionality of `copySemanticDbFiles`. It also ensures, we return a path with only`semanticdb` files in it. Your change returns the complete compiler output directory, which may contain `class` files and other auxiliary files.

PR #3080, opened 2024-03-11
Added incremental tests for semanticDbData

lefou:
- Added incremantal test for JavaModule.semanticDbData
- Added incremental test for ScalaModule.semanticDbData

### DIFF
--- a/scalalib/test/src/mill/scalalib/HelloJavaTests.scala
+++ b/scalalib/test/src/mill/scalalib/HelloJavaTests.scala
@@ -55,24 +55,84 @@ object HelloJavaTests extends TestSuite {
       )
     }
     "semanticDbData" - {
-      val eval = init()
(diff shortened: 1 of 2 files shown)


lefou:
Looks like in all test cases zinc is never compiling incrementally. More work is needed.

lefou:
Here are some issues I identified with our current approach to generate semanticdb data in a Scala 2.13 module:

* When we stop scalac after phase `semanticdb-typer`, zinc is not able to incrementally compile the next round, therefore we currently always generate all semanticdb files in each run
* If we don't stop scalac after phase `semanticdb-typer`, we can generate semanticdb data incrementally, but we also unnecessarily compile all scala files to class files
* Incremental compilation doesn't seem to clean semanticdb files for removed sources. So, we need to manually remove those that belong to non-existing source files

PR #3084, opened 2024-03-14
Prepare Mill for Scala SIP-51 (unfreeze scala-library)

lefou:
- Update zinc to 1.10.0-M3
- Use newer coursier `Resolve` API to detect conflicting dependencies


lihaoyi:
Closing due to inactivity

PR #3094, opened 2024-03-19
Support Scala.js minify via `scalaJSMinify: Target[String]`

lolgab:
`minify` is a new feature released today by Scala.js https://www.scala-js.org/news/2024/03/19/announcing-scalajs-1.16.0

Pull Request: https://github.com/com-lihaoyi/mill/pull/3094

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -70,7 +70,7 @@ object Deps {
   val testScala33Version = "3.3.1"
 
   object Scalajs_1 {
-    val scalaJsVersion = "1.15.0"
+    val scalaJsVersion = "1.16.0"
     val scalajsEnvJsdomNodejs = ivy"org.scala-js::scalajs-env-jsdom-nodejs:1.1.0"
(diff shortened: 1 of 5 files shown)


build.sc
```diff
@@ -70,7 +70,7 @@ object Deps {
   val testScala33Version = "3.3.1"
 
   object Scalajs_1 {
-    val scalaJsVersion = "1.15.0"
+    val scalaJsVersion = "1.16.0"
```
lefou, line 73: This version is not forward binary compatible with older releases, so users can't use lower versions downstream, which might be an issue for libraries. `1.15.0` was forward compatible down to `1.13.0`. Can you tell how important it is for user to use or stay compatible to older versions? Should we cross-compile our workers to support older versions?

lolgab: Our worker can link correctly from Scala.js `1.0.0` to future `1` versions.
We check the Scala.js version in the linker to avoid using features that don't exist in older versions.

lefou (approved):

lefou:
Looks like it's time for a dedicated documentation page, describing all the options we have when configuring a Scala.js module. It should include some background information what the compiler and what the linker does.

PR #3099, opened 2024-03-22
TestModule: Support generation of JUnit-compatible xml report

atty303:
fixes #2700

image

I am verifying with MUnit.

PR #3103, opened 2024-03-27
Add support for JUnit legacy XML reports

romain-gilles-ultra:
Some CI environments provide Unit test reports but for that, they need a specific output format.

This PR adds a general usage module to export TestResult into JUnit legacy XML test suite reports.


lefou:
@romain-gilles-ultra I just merged #3099, which provides almost the same report (I guess) out of the box. Can you please comment, if this PR is now obsolete or still valid?

PR #3122, opened 2024-04-16
Scoverage improve classpath override to the minimal

romain-gilles-ultra:
The previous version the `ScoverageTests.runClasspath` was more modified and reorganized.
In this version, only the outer complied classes folder is replaced (in place) by the outer.scoverage complied classes folder.
An additional runtime dependency is also added: `org.scoverage::scalac-scoverage-runtime`.
All duplicated entries in the classpath are also removed.

### DIFF
--- a/contrib/scoverage/src/mill/contrib/scoverage/ScoverageModule.scala
+++ b/contrib/scoverage/src/mill/contrib/scoverage/ScoverageModule.scala
@@ -232,19 +232,18 @@ trait ScoverageModule extends ScalaModule { outer: ScalaModule =>
   trait ScoverageTests extends ScalaTests {
 
     /**
-     * Alter classfiles and resources from upstream modules and dependencies
-     * by removing the ones from outer.localRunClasspath() and replacing them
-     * with outer.scoverage.localRunClasspath()
+     * Alter classpath from upstream modules by replacing in-place outer module
+     * classes folder by the outer.scoverage classes folder and adding the
+     * scoverage runtime dependency.
      */
     override def runClasspath: T[Seq[PathRef]] = T {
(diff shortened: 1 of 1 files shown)


lefou (approved):
Look good to me.

romain-gilles-ultra:
It sounds like the failing test has nothing to do with this PR :(

contrib/scoverage/src/mill/contrib/scoverage/ScoverageModule.scala
```diff
@@ -232,19 +232,18 @@ trait ScoverageModule extends ScalaModule { outer: ScalaModule =>
...
-        outerLocalRunClasspath
-      ) ++
-        outer.scoverage.localRunClasspath() ++
-        resolveDeps(T.task {
-          outer.scoverageRuntimeDeps().map(bindDependency())
-        })()
+      val outerClassesPath = outer.compile().classes
+      val outerScoverageClassesPath = outer.scoverage.compile().classes
+      (super.runClasspath().flatMap { path =>
+        if (outerClassesPath == path) Seq(outerScoverageClassesPath) else Seq(path)
```
lolgab, line 243: Since both sequences have exactly one element, you can use `map` here instead of `flatMap`, so you don't create an extra `Seq`

romain-gilles-ultra: Good point I will fix that thanks

romain-gilles-ultra: It sounds like @lefou already did it ...
Sorry I'm in vacations so I only check my PR now :(

PR #3124, opened 2024-04-17
Use docker image hash to determine if build should rerun

nrktkt:
previously the docker build would re-run if the user had `pullBaseImage` set to true. now it always checks the remote repository for a new version of the base image, but only re-runs the build if the base image hash has changed


lefou (approved):
Looks good to me. Thank you!

lefou:
@nrktkt Can you please reformat your changes?

PR #3125, opened 2024-04-19
Retry example integration tests that didn't finish after 5 minutes

lefou:
The idea is, that our example integration tests should finish after a short period of time. But sometimes then hang in CI, so we simply abort and retry them automatically, instaed of manually.

The hardcoded timeout of 5 minutes is just a guess. I want to see the CI results. Maybe, we can make it configurable per test suite.

### DIFF
--- a/example/src/mill/integration/ExampleTestSuite.scala
+++ b/example/src/mill/integration/ExampleTestSuite.scala
@@ -72,20 +72,18 @@ object ExampleTestSuite extends IntegrationTestSuite {
     }
 
     test("exampleUsage") {
-      try {
-        def body = {
-          val parsed = upickle.default.read[Seq[(String, String)]](sys.env("MILL_EXAMPLE_PARSED"))
-          val usageComment = parsed.collect { case ("example", txt) => txt }.mkString("\n\n")
-          val commandBlocks = ("\n" + usageComment.trim).split("\n> ").filter(_.nonEmpty)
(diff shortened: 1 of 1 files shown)


lolgab:
The problem with this approach is that the `Future` is not cancelled, so the other runs will run concurrently with the first.

lefou:
> The problem with this approach is that the `Future` is not cancelled, so the other runs will run concurrently with the first.

How about using a Java future, which is cancellable?

lolgab:
I'm not very familiar with it, but if it's cancelable it is better, yes. Otherwise you need to run a `Thread` and handle it manually.

example/src/mill/integration/ExampleTestSuite.scala
```diff
@@ -46,15 +49,40 @@ object ExampleTestSuite extends IntegrationTestSuite {
...
+      }
+
+    }
+
     test("exampleUsage") {
       try {
-        val parsed = upickle.default.read[Seq[(String, String)]](sys.env("MILL_EXAMPLE_PARSED"))
-        val usageComment = parsed.collect { case ("example", txt) => txt }.mkString("\n\n")
-        val commandBlocks = ("\n" + usageComment.trim).split("\n> ").filter(_.nonEmpty)
+        def body = {
```
lolgab, line 76: Why don't you write the function inline like:
```scala
retryOnTimeout(3) {
 ..... 
}

```

lefou: just some missing cleanup, I wanted to restructure it further, now done.

lolgab (approved):

PR #3126, opened 2024-04-23
Remove some warnings

romain-gilles-ultra:
While I was working on the `ScoverageModule` I got some warnings from IDEA.
I fix some of the warnings, unwarp some forwarding calls, and make type declarations more consistent with the parent trait.

### DIFF
--- a/contrib/scoverage/src/mill/contrib/scoverage/ScoverageModule.scala
+++ b/contrib/scoverage/src/mill/contrib/scoverage/ScoverageModule.scala
@@ -79,13 +79,13 @@ trait ScoverageModule extends ScalaModule { outer: ScalaModule =>
     } else {
       if (isScoverage2()) {
         Agg(
-          ivy"org.scoverage:::scalac-scoverage-plugin:${sv}",
-          ivy"org.scoverage::scalac-scoverage-domain:${sv}",
-          ivy"org.scoverage::scalac-scoverage-serializer:${sv}",
-          ivy"org.scoverage::scalac-scoverage-reporter:${sv}"
+          ivy"org.scoverage:::scalac-scoverage-plugin:$sv",
+          ivy"org.scoverage::scalac-scoverage-domain:$sv",
+          ivy"org.scoverage::scalac-scoverage-serializer:$sv",
+          ivy"org.scoverage::scalac-scoverage-reporter:$sv"
         )
       } else {
-        Agg(ivy"org.scoverage:::scalac-scoverage-plugin:${sv}")
+        Agg(ivy"org.scoverage:::scalac-scoverage-plugin:$sv")
       }
     }
   }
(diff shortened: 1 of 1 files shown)


contrib/scoverage/src/mill/contrib/scoverage/ScoverageModule.scala
```diff
@@ -79,13 +79,13 @@ trait ScoverageModule extends ScalaModule { outer: ScalaModule =>
...
       if (isScoverage2()) {
         Agg(
-          ivy"org.scoverage:::scalac-scoverage-plugin:${sv}",
-          ivy"org.scoverage::scalac-scoverage-domain:${sv}",
-          ivy"org.scoverage::scalac-scoverage-serializer:${sv}",
-          ivy"org.scoverage::scalac-scoverage-reporter:${sv}"
+          ivy"org.scoverage:::scalac-scoverage-plugin:$sv",
+          ivy"org.scoverage::scalac-scoverage-domain:$sv",
+          ivy"org.scoverage::scalac-scoverage-serializer:$sv",
+          ivy"org.scoverage::scalac-scoverage-reporter:$sv"
```
lefou, line 85: Although this might be personal, I'd prefer to always use the `${..}` version.

contrib/scoverage/src/mill/contrib/scoverage/ScoverageModule.scala
```diff
@@ -211,9 +205,9 @@ trait ScoverageModule extends ScalaModule { outer: ScalaModule =>
       T {
         val extras =
           if (isScala3()) {
-            Seq(s"-coverage-out:${data().path.toIO.getPath()}")
+            Seq(s"-coverage-out:${data().path.toIO.getPath}")
```
lefou, line 208: I'd prefer to call methods as they are defined, and Java API methods always have exactly one parameter list. I don't understand why correctness is treated here. In Scala, methods can have none or multiple parameter lists, so we should always be extra careful to use the exact version.

contrib/scoverage/src/mill/contrib/scoverage/ScoverageModule.scala
```diff
@@ -189,18 +183,18 @@ trait ScoverageModule extends ScalaModule { outer: ScalaModule =>
     }
 
     override def compileResources: T[Seq[PathRef]] = outer.compileResources
-    override def generatedSources: Target[Seq[PathRef]] = T { outer.generatedSources() }
-    override def allSources: Target[Seq[PathRef]] = T { outer.allSources() }
+    override def generatedSources: T[Seq[PathRef]] = outer.generatedSources
```
lefou, line 186: I must admit I added those wrappers on purpose in the past, in the believe they add more meaningful context to the targets, which might be useful when debugging. But I never verified, whether this is needed or useful. I see your point with this cleanup, as after your change we produce less byte code and safe some cycles at runtime.

contrib/scoverage/src/mill/contrib/scoverage/ScoverageModule.scala
romain-gilles-ultra: I don't know but IDEA complains about it... I can rollback it

PR #3130, opened 2024-04-24
feat: adding sonatype central support

Andrapyre:
### Description
This PR adds support for the new Sonatype Central publishing api, documented [here](https://central.sonatype.org/publish/publish-portal-api/) and [here](https://central.sonatype.com/api-doc). To use, ensure that your environment variables contain the appropriate credentials and configure your module like so in `build.sc`:

```scala
import $ivy.`com.lihaoyi::mill-contrib-sonatypecentral:`
import mill.contrib.sonatypecentral.SonatypeCentralPublishModule

object mymodule extends SonatypeCentralPublishModule {
  ...
}
```

### New Dependencies
This PR introduces the following new dependency:
```scala
ivy"com.lumidion::sonatype-central-client-requests:0.2.0" 
```


Andrapyre:
@lihaoyi , could you take a look at this? A few open questions:

1. What's the best way to make these changes accessible to the user? By altering the current `PublishModule` or by creating a new `SonatypeCentralPublishModule` or something like it?
2. There are a number of changes to the `PublishModule` affecting binary compat. Basically, I refactored as many publish options as possible into two case classes (one for legacy publishing and one for central publishing) that the user can configure, instead of having the user pass them directly into the `publish` function. Is breaking binary compat acceptable here, since mill still hasn't hit version `1.0.0`?
3. The current implementation outputs publishable jar files to `./out/central-publish`. Let me know if that should change (I don't think the older Sonatype implementation outputted any jar files, unless I'm mistaken) or if the directory should be configurable.

scalalib/src/mill/scalalib/publish/SonatypeCentralPublisher.scala
```diff
@@ -0,0 +1,84 @@
...
+    val (_, releases) = mappings.partition(_._1.isSnapshot)
+
+    val releaseGroups = releases.groupBy(_._1.group)
+    val wd = os.pwd / "out" / "publish-central"
+    os.makeDir.all(wd)
+
+    for ((_, groupReleases) <- releaseGroups) {
+      groupReleases.foreach { case (artifact, data) =>
+        val jarFile =
+          (wd / s"${artifact.group}-${artifact.id}-${artifact.version}.jar").toIO
```
nightscape, line 45: This is just for the temporary artifact that gets uploaded to Sonatype, right?
Afaik, one could also use a `.zip` file extension, which would make it clear that this is not an ordinary JAR, but an export of a Maven package hierarchy.

build.sc
```diff
@@ -560,7 +561,8 @@ object main extends MillStableScalaModule with BuildInfo {
     Deps.windowsAnsi,
     Deps.mainargs,
     Deps.coursierInterface,
-    Deps.requests
+    Deps.requests,
+    Deps.sonatypeCentralClient
```
nightscape, line 565: Nitpick: Maybe add a trailing comma here, so the next time someone needs to add a line, it doesn't change an unrelated line (like `Deps.requests` here).

lefou: That would not be in line with our current formatting rules

lefou (commented):
Thank you for this contribution. We have a lot of other publishers, that currently reside as contrib plugins. Since your PR a) is rather new, b) touches a lot of heavily used and depend-on functionality which is c) rather proven and tested to work, I'd suggest to start this addition as a contrib module. So users can start to adopt this new API without compromizing existing setups.

That is also the best way to provide an alternative `PublishModule` `v2`, starting with the refactorings you mentioned applied to it. I haven't reviewed this PR in detail, but since you mentioned that you introduced cases classes, which themselves aren't ideal to maintain binary compatibility, I suggest to go with a contrib module for now.

If that's not what you want, I'd like to have some smaller change set, preserved binary compatibility, some docs describing how the new mechanism works and how it differs to the existing logic and some test coverage of the newly introduced functionality, if possible.

Please also keep in mind, that the current publish logic in `PublishModule` is also used for remote repositorries other than Sonatype OSS / Maven Central, but for many local / company wide repositories like Nexus.

build.sc
Andrapyre: This was removed anyway in the process of moving everything to a contrib module - resolving.

Andrapyre:
I've updated with requested changes and documentation. Before I go on to add tests, could you review and approve the current implementation?

Andrapyre:
@lefou and @nightscape , just making sure this doesn't get lost.

nightscape:
@lefou how would the separate `PublishModule` work with e.g. mill-ci-release which [`extends PublishModule`](https://github.com/ckipp01/mill-ci-release/blob/27bcbf7a119e9a691d0cac021ba1c9a142c78d62/plugin/src/io/kipp/mill/ci/release/CiReleaseModule.scala#L26) and provides additional functionality on top?

Andrapyre:
@nightscape , `mill-ci-release` would need a new module that extends `SonatypeCentralPublishModule`. The two publishing processes (standard and Sonatype Central) are different enough that @lefou is probably right that these should live in separate modules.

lefou:
I can't say much about `mill-ci-release`, which I don't use. I think publishing with Mill built-in support is already straight forward, given that you have a properly setup pgp/gpg key. If `mill-ci-release` needs to be aware of any additional target not already provided by `PublishModule` (which you could mix-in), then it most likely also needs extra customizations.

lefou (dismissed):
[As written before](https://github.com/com-lihaoyi/mill/pull/3130#pullrequestreview-2019692076):

> Thank you for this contribution. We have a lot of other publishers, that currently reside as contrib plugins. Since your PR a) is rather new, b) touches a lot of heavily used and depend-on functionality which is c) rather proven and tested to work, I'd suggest to start this addition as a contrib module. So users can start to adopt this new API without compromizing existing setups.
> 
> That is also the best way to provide an alternative `PublishModule` `v2`, starting with the refactorings you mentioned applied to it. I haven't reviewed this PR in detail, but since you mentioned that you introduced cases classes, which themselves aren't ideal to maintain binary compatibility, I suggest to go with a contrib module for now.
> 
> If that's not what you want, I'd like to have some smaller change set, preserved binary compatibility, some docs describing how the new mechanism works and how it differs to the existing logic and some test coverage of the newly introduced functionality, if possible.
> 
> Please also keep in mind, that the current publish logic in `PublishModule` is also used for remote repositorries other than Sonatype OSS / Maven Central, but for many local / company wide repositories like Nexus.

Andrapyre:
@lefou , this PR is ready to be reviewed. I have made all requested changes (e.g. this is now a contrib module, no changes to bincompat, removed case classes, etc). Could you explain why this PR was changed to draft with a reference to your previous comment?

lefou:
@Andrapyre I somehow missed, that you already changed to PR. Sorry for that.

scalalib/src/mill/scalalib/publish/SonatypeHelpers.scala
```diff
@@ -0,0 +1,75 @@
+package mill.scalalib.publish
+
+import mill.util.Jvm
+
+import java.math.BigInteger
+import java.security.MessageDigest
+
+trait SonatypeHelpers {
```
lefou, line 8: Looks like there are only helper `def`s inside which don't use any state and don't seem to be customized via inhertance, so we could as well make it an `object` and avoid the `extends` on the use site.

scalalib/src/mill/scalalib/PublishModule.scala
```diff
@@ -306,22 +314,44 @@ object PublishModule extends ExternalModule {
...
-        username <- T.env.get("SONATYPE_USERNAME")
-        password <- T.env.get("SONATYPE_PASSWORD")
-      } yield {
-        Result.Success(s"$username:$password")
-      }).getOrElse(
-        Result.Failure(
-          "Consider using SONATYPE_USERNAME/SONATYPE_PASSWORD environment variables or passing `sonatypeCreds` argument"
-        )
-      )
+  private[mill] def getFinalGpgArgs(initialGpgArgs: String): Seq[String] = {
```
lefou, line 317: If this is to be used from derived classes, `protected` is better than `private[mill]`. If it's API useful outside of this module, just make it public.

contrib/sonatypecentral/src/mill/contrib/sonatypecentral/SonatypeCentralPublishModule.scala
```diff
@@ -0,0 +1,143 @@
...
+trait SonatypeCentralPublishModule extends PublishModule {
+  def gpgArgs: T[String] = T { defaultStringGpgArgs }
+
+  def connectTimeout: T[Int] = T { defaultConnectTimeout }
+
+  def readTimeout: T[Int] = T { defaultReadTimeout }
+
+  def awaitTimeout: T[Int] = T { defaultAwaitTimeout }
+
+  def shouldRelease: T[Boolean] = T { true }
```
lefou, line 32: I think it would by good to prefix these target names with something unique to your module, e.g. `sonatype` or `sonatypeCentral`. This makes reading larger build files much easier.

contrib/sonatypecentral/src/mill/contrib/sonatypecentral/SonatypeCentralPublishModule.scala
```diff
@@ -0,0 +1,143 @@
...
+    }
+}
+
+object SonatypeCentralPublishModule extends ExternalModule {
+
+  private val defaultCredentials = ""
+  private val defaultReadTimeout = 60000
+  private val defaultConnectTimeout = 5000
+  private val defaultAwaitTimeout = 120 * 1000
+  private val defaultShouldRelease = true
```
lefou, line 64: Defaults should be `public` so API users can access them.

scalalib/src/mill/scalalib/PublishModule.scala
```diff
@@ -252,8 +256,12 @@ trait PublishModule extends JavaModule { outer =>
 
 object PublishModule extends ExternalModule {
   val defaultGpgArgs: Seq[String] = Seq("--batch", "--yes", "-a", "-b")
+  val defaultStringGpgArgs: String = defaultGpgArgs.mkString(",")
 
-  case class PublishData(meta: Artifact, payload: Seq[(PathRef, String)])
+  case class PublishData(meta: Artifact, payload: Seq[(PathRef, String)]) {
+    def toDataWithConcretePath: (Seq[(Path, String)], Artifact) =
+      (payload.map { case (p, f) => (p.path, f) }, meta)
+  }
```
lefou, line 264: I'd like some scaladoc here and maybe, you can flip the tuple so that `Artifact` comes first, so it's more clear it's just the unpacked `Path` version of the initial case class values.

lefou (commented):
Can you update the PR description to reflect the current state.

scalalib/src/mill/scalalib/PublishModule.scala
Andrapyre: Unfortunately, this signature ```(Seq[(Path, String)], Artifact)``` is the one required by all of the publishAll signatures and refactoring it is starting to get out-of-scope for this PR. I agree that flipping it feels a little random here, but that isn't something that I introduced. This is just a helper function that centralizes what was already being done. I've added a scaladoc and made it `private[mill]`so that it can be refactored later.

scalalib/src/mill/scalalib/PublishModule.scala
```diff
@@ -275,7 +283,7 @@ object PublishModule extends ExternalModule {
       publishArtifacts: mill.main.Tasks[PublishModule.PublishData],
       sonatypeCreds: String = "",
       signed: Boolean = true,
-      gpgArgs: String = defaultGpgArgs.mkString(","),
+      gpgArgs: String = defaultStringGpgArgs,
```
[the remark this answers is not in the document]
Andrapyre: I tried to standardize anything that was being used in multiple times and this is being used in three different places. I've made defaultStringGpgArgs `private [mill]`, so any refactoring won't break bincompat later on.

scalalib/src/mill/scalalib/publish/SonatypeHelpers.scala
Andrapyre: Done.

contrib/sonatypecentral/src/mill/contrib/sonatypecentral/SonatypeCentralPublishModule.scala
Andrapyre: Prefixed all with `sonatypeCentral`.

contrib/sonatypecentral/src/mill/contrib/sonatypecentral/SonatypeCentralPublishModule.scala
Andrapyre: Done.

Andrapyre:
> Can you please update the PR description to reflect the current state?

Done.

scalalib/src/mill/scalalib/PublishModule.scala
Andrapyre: This is used in non-derived classes in other modules  (e.g. in `SonatypeCentralPublishModule.publishAll`), but I don't think that it makes sense to expose it to outside users, as it's just for internal parsing of arguments and only relevant to the Sonatype/Nexus world. I've left it as `private [mill]`, especially to give later contributors the freedom to refactor it without breaking bincompat.

scalalib/src/mill/scalalib/PublishModule.scala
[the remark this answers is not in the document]
lefou: The issue is, that users who want to override this methods might want to re-use the default arguments. Since these are now pre-formatted and private, they no longer can't. Whereas previously, they could just re-use the `Seq`, maybe filter some args and add others and are done. Theoretically, they still could use `defaultGpgArgs`, but there is nowhere any indication of that option, not in the docs nor in the visible (to them) code.

Hence, I think keeping `defaultGpgArgs` is the better choice. Even if that means, we need to format them (`.mkString(","`) in multiple places. But the API is more clear and obvious defaults are reusable. And this is not a time-critical code path.

```suggestion
      gpgArgs: String = defaultGpgArgs.mkString(","),
```

scalalib/src/mill/scalalib/PublishModule.scala
lefou: We can leave it for now.

lefou (commented):
We're almost there. One nit, otherwise this looks good to me.

scalalib/src/mill/scalalib/PublishModule.scala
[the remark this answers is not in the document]
Andrapyre: Done.

lefou (approved):

lbialy:
Out of curiosity @Andrapyre - was Sonatype Central able to handle snapshots at the time of this PR?

PR #3140, opened 2024-05-03
Detect assemblies with too many entries to fail shell script prepending

lefou:
This is an attempt to fix the issue of invalid assembly files which occurs when the following two conditions are met:
* The JAR file has more than `65535` ZIP entries
* The JAR has a prepended shell script

The issue was reported and analyzed in the following tickets:
* https://github.com/com-lihaoyi/mill/issues/528
* https://github.com/com-lihaoyi/mill/issues/2650

This issue also hits other build tools, but it seems Mill is the only tools which automatically enables shell script prepending by default.

Since there is no real fix available, we simply try to detect the issue after the fact and fail the assemble task with a actionable error message.

To make the fix binary compatible, I had to deprecated the `upstreamAssembly` target in favor to the new `upstreamAssembly2` target, which returns also the added ZIP entry count. Since this can be a behavioral change when users have overridden the `upstreamAssembly` target, I also added some warning messages with will detect this at runtime and provide actionable help.


scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -583,13 +599,35 @@ trait JavaModule
...
-      Some(upstreamAssembly().path),
+      prependScript,
+      Some(upstream.pathRef.path),
       assemblyRules
     )
+    // See https://github.com/com-lihaoyi/mill/pull/2655#issuecomment-1672468284
+    val problematicEntryCount = 65535
+    if (
+      prependScript.isDefined &&
+      (upstream.addedEntries + created.addedEntries) > problematicEntryCount
```
lihaoyi, line 617: I wonder if it's really worth keeping track of the number of entries incrementally? v.s. creating the assembly first, scanning it, and then failing if it has too many entries. If the performance of scanning the assembly is acceptable (milliseconds to tens of milliseconds?) then that would save us a bunch of book-keeping passing around `addedEntries` values, and a bunch of churn in replacing `upstreamAssembly` with `upstreamAssembly2`

integration/feature/assembly/repo/build.sc
```diff
@@ -0,0 +1,34 @@
...
+  }
+  object large extends Setup with ExtraDeps {
+    override def prependShellScript: T[String] = ""
+  }
+}
+
+object exe extends Module {
+  object small extends Setup
+  object large extends Setup with ExtraDeps
+}
```
lihaoyi, line 34: Does this need to be an integration test? I feel like having it be a `scalalib` unit test should be sufficient. Integration tests are slow and flaky so better to avoid them unless truly necessary (e.g. testing launcher logic, bootstrap logic, script compilation, etc)

scalalib/src/mill/scalalib/JavaModule.scala
lefou: It probably depends. When we scan the jar right after it's written, it should be in the file system cache of any reasonable OS and scanning should be fast. On the other side, just remembering the added count is quasi for free, esp. when we assume, that the upstream-assembly is the larger portion of the assembly and keeps stable.

I think I want to experiment what the result looks like and how it performs.

integration/feature/assembly/repo/build.sc
lefou: No. Since we now identified the issue, it can be a unit test that just make sure it properly works.

lefou: Changed.

scalalib/src/mill/scalalib/JavaModule.scala
lefou: The issues with scanning afterwards:
* Scanning and counting entries is slow. Scanning the jar of test case `noExe.large` took almost 3 seconds, although we just wrote it
* Java's `ZipInputStream` and `JarInputStream` aren't able to find any `ZipEntry` in a prependend jar making scanning after packaging non-trivial
* Shelling out to some OS-installed zip tools doesn't seem right

lihaoyi (approved):
lgtm

PR #3143, opened 2024-05-04
Add multi platform support to contrib.docker through docker buildx

GeorgOfenbeck:
A very minor change to the contrib.docker plugin (and its docs).

Allows to build for other architectures - a use case that becomes more and more useful with ARM architectures becoming more relevant.

**Relevant docs:**
[docker buildx (Buildkit)](https://docs.docker.com/reference/cli/docker/buildx/build/#platform)

### DIFF
--- a/contrib/docker/readme.adoc
+++ b/contrib/docker/readme.adoc
@@ -64,7 +64,11 @@ object docker extends DockerConfig {
   def user = "new-user"
   // Optionally override the docker executable to use something else
   def executable = "podman"
+  // If the executable is docker (which is also the default), you can optionally also pass a platform parameter
+  // docker buildx is then used to build multi-platform images
+  def platform = "linux/arm64"
 }
 ----
+
 
 Run mill in interactive mode to see the docker client output, like `mill -i foo.docker.build`.
--- a/contrib/docker/src/mill/contrib/docker/DockerModule.scala
+++ b/contrib/docker/src/mill/contrib/docker/DockerModule.scala
@@ -85,12 +85,12 @@ trait DockerModule { outer: JavaModule =>
     def user: T[String] = ""
 
     /**
-      * Optional platform parameter, if set uses buildkit to build for specified platform.
-      * 
-      * See also the Docker docs on
-      * [[https://docs.docker.com/reference/cli/docker/buildx/build/#platform]]
-      * for more information.
-      */
+     * Optional platform parameter, if set uses buildkit to build for specified platform.
+     *
+     * See also the Docker docs on
+     * [[https://docs.docker.com/reference/cli/docker/buildx/build/#platform]]
+     * for more information.
+     */
     def platform: T[String] = ""
 
     /**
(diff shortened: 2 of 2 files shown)


lefou (approved):
Looks good to me. Thank you!

yankay:
It's fascinating to see the multi-arch images supported by Mill.

There are two comments on the PR:

1.
From the https://docs.docker.com/reference/cli/docker/buildx/build/#platform , we can build multi-arch images with:
``
docker buildx build --platform=linux/arm64 .
docker buildx build --platform=linux/amd64,linux/arm64,linux/arm/v7 .
docker buildx build --platform=darwin .
```

The `--platform` is a string array, not a string.


2. The image built by `docker buildx` should be pushed with `docker buildx --push` and can not use `docker push` later.

**Comment by GeorgOfenbeck** [comment]
For 1. I ignored the multi-arch image option a bit as currently that is not enabled by default on docker:
see: [https://docs.docker.com/build/building/multi-platform/](https://docs.docker.com/build/building/multi-platform/)

"By default, Docker can build for only one platform at a time."

One can of course make it work as described in the article.
For the option I felt that going for the single target platform is the better choice as by default otherwise one gets the error
"ERROR: Multi-platform build is not supported for the docker driver.
Switch to a different driver, or turn on the containerd image store, and try again.
Learn more at https://docs.docker.com/go/build-multi-platform/" 
Which might misslead people on first glance that other platforms are not working 

For those that know what they are doing, they could always pass it as a single string. Changing to an Array is of course an option, but one should definetly point out that issue then in the docu.


For 2.
https://github.com/com-lihaoyi/mill/blob/1792ac4442dfcff4ff726ba54e08b2eb580eb86b/contrib/docker/src/mill/contrib/docker/DockerModule.scala#L190 doesnt use either build nor buildx - therefore i dont think it should be an issue. 
calling "_.docker.push" works as expected for me. Are you referring to that push?

**Comment by yankay** [comment]
> For 1. I ignored the multi-arch image option a bit as currently that is not enabled by default on docker: see: https://docs.docker.com/build/building/multi-platform/
> 
> "By default, Docker can build for only one platform at a time."
> 
> One can of course make it work as described in the article. For the option I felt that going for the single target platform is the better choice as by default otherwise one gets the error "ERROR: Multi-platform build is not supported for the docker driver. Switch to a different driver, or turn on the containerd image store, and try again. Learn more at https://docs.docker.com/go/build-multi-platform/" Which might misslead people on first glance that other platforms are not working
> 
> For those that know what they are doing, they could always pass it as a single string. Changing to an Array is of course an option, but one should definetly point out that issue then in the docu.
> 
> For 2.
> 
> https://github.com/com-lihaoyi/mill/blob/1792ac4442dfcff4ff726ba54e08b2eb580eb86b/contrib/docker/src/mill/contrib/docker/DockerModule.scala#L190
> 
> doesnt use either build nor buildx - therefore i dont think it should be an issue.
> calling "_.docker.push" works as expected for me. Are you referring to that push?

HI @GeorgOfenbeck 

1. 
Follow the Blog: https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

The buildx is usually used to build Multi-Platform Image.
```
# building an image for two platforms
docker buildx build --platform=linux/amd64,linux/arm64 .
```
The command can work correctly in most environments when the docker build is installed. :-) For example Ubuntu x86.


2.
If the user using `mill docker.push` with `platform`, it may not work correctly. 
The command of build and push cannot split: https://github.com/docker/buildx/issues/1152

For multi-arch image
```
docker buildx build --platform=linux/amd64,linux/arm64 --push .
```
can work correctly

GeorgOfenbeck:
2. can definitely be done in separate phases - the push code was not introduced in this PR! Also see docu here: https://docs.docker.com/reference/cli/docker/image/push/

It can easily be verified locally by running docker push <tag> for any architecture image available locally.

For 1. as written - one can definitely go that route - but as the rest of the plugin also "just" covers the most common cases (e.g. not being able to add any files etc)  i opted for the simpler approach that should always work

PR #3177, opened 2024-05-22
Update utest form 0.8.2 to 0.8.3

lefou:


lefou:
Looks like utest 0.8.3 dropped support for Scala Native 0.4. Why can't there be a higher version bump?

PR #3180, opened 2024-05-22
Make scalafmt version dynamic in ScalafmtModule

pieter-bos:
I see that the `scalafmt` version has been stuck at `3.7.15` for a bit, since java 8 support was dropped past that version. I'd like to use some bugfixes from later versions, so how about we make the version dynamic in `ScalafmtModule`? Scalafmt offers a nice binary-compatible interface through `scalafmt-interfaces`, so mill itself can stick with that. Some considerations:

* The BuildInfo entry does not make much sense now, since in general the interfaces version is unrelated to any used scalafmt version.
* I'm not a fan of setting an arbitrary old default version for `scalafmtVersion`, but I guess this is required for compatibility (?)

Thanks!

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -165,7 +165,7 @@ object Deps {
   val scalaCheck = ivy"org.scalacheck::scalacheck:1.17.0"
   def scalaCompiler(scalaVersion: String) = ivy"org.scala-lang:scala-compiler:${scalaVersion}"
   // last scalafmt release supporting Java 8 is 3.7.15
-  val scalafmtDynamic = ivy"org.scalameta::scalafmt-dynamic:3.7.15" // scala-steward:off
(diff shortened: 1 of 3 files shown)


lefou:
> I'd like to use some bugfixes from later versions, so how about we make the version dynamic in `ScalafmtModule`?

`scalafmt-dynamic-3.7.15` is dynamically using the scalafmt version specified in `.scalafmt.conf`. So you should able to use whatever newer version you want. That's the point of using `scalafmt-dynamic` with a config file over a configuration inside `build.sc`.

pieter-bos:
I see! I tried this earlier and it had no effect then, but it's working now, so I must have missed something. Sorry for the PR noise!

lefou:
No problem and thanks for your contribution! 

If you find that our documentation could be improved, feel free to open another PR!

PR #3184, opened 2024-05-24
Add unit tests to the JUnit xml report generation

romain-gilles-ultra:
Add unit tests to the JUnit XML report generation.

### DIFF
--- a/scalalib/src/mill/scalalib/TestModule.scala
+++ b/scalalib/src/mill/scalalib/TestModule.scala
@@ -385,12 +385,12 @@ object TestModule {
 
       <testsuite name={fqn}
                  tests={testResults.length.toString}
-                 failures={testResults.count(_.status == Status.Failure.toString).toString}
-                 errors={testResults.count(_.status == Status.Error.toString).toString}
+                 failures={testResults.count(_.status == FailureStatus).toString}
+                 errors={testResults.count(_.status == ErrorStatus).toString}
                  skipped={
         testResults.count(testResult => SkippedStatus.contains(testResult.status)).toString
       }
-                 time={(testResults.map(_.duration).sum / 1000.0).toString}
+                 time={durationAsString(testResults.map(_.duration).sum)}
                  timestamp={formatTimestamp(timestamp)}>
         {properties}
         {cases}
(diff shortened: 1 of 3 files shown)


scalalib/src/mill/scalalib/TestModule.scala
```diff
@@ -386,9 +387,11 @@ object TestModule {
                  tests={testResults.length.toString}
                  failures={testResults.count(_.status == Status.Failure.toString).toString}
                  errors={testResults.count(_.status == Status.Error.toString).toString}
-                 skipped={testResults.count(_.status == Status.Skipped.toString).toString}
+                 skipped={
+        testResults.count(testResult => SkippedStatus.contains(testResult.status)).toString
```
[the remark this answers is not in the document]
romain-gilles-ultra: Hi @lefou 
You are right I should always use the `Enum.name()` function which is the right function to get the string representation of the enum name. I will fix it. BTW I introduced constants to for those name it's my mistake to not use them here ;)

scalalib/src/mill/scalalib/TestModule.scala
```diff
@@ -206,6 +206,10 @@ trait TestModule
 
 object TestModule {
   private val FailedTestReportCount = 5
+  private val ErrorStatus = Status.Error.name()
+  private val FailureStatus = Status.Failure.name()
+  private val SkippedStatus =
```
lefou, line 211:
```suggestion
  private val SkippedStates =
```

scalalib/src/mill/scalalib/TestModule.scala
```diff
@@ -423,17 +431,17 @@ object TestModule {
         )
     ).getOrElse("")
     e.status match {
-      case Error if (e.exceptionMsg.isDefined && e.exceptionName.isDefined) =>
+      case ErrorStatus if (e.exceptionMsg.isDefined && e.exceptionName.isDefined) =>
```
lefou, line 434: That's an interesting change. The old type must have been some kind of `Status` or `enum`, the new type is `String`. So either the old or the new version doesn't work or I'm missing something.

lefou: Nevermind. Just realized, the renamed the `val`s.

lefou (approved):
Looks good to me. Thank you!

Issue #3075, opened 2024-03-11
Implement SIP-51 (unfreezing the Scala library)


lrytz:
There is an [open PR for sbt](https://github.com/sbt/sbt/pull/7480) to implement [SIP-51 (unfreezing the Scala library)](https://docs.scala-lang.org/sips/drop-stdlib-forwards-bin-compat.html). The topic was discussed at a recent Scala meeting and will be brought up in the next SIP meeting (Friday Mar 15); everyone seems to be supporting it.

Example:

```
➜ m cat build.sc
import mill._, scalalib._
object proj extends ScalaModule {
  def scalaVersion = "2.13.8"
  def ivyDeps = Agg(
    ivy"com.softwaremill.sttp.client3::core:3.8.3", // depends on scala-library 2.13.10, ws 1.3.10
    ivy"com.softwaremill.sttp.shared::ws:1.2.7",
  )
}
➜ m mill show proj.runClasspath
[
  "qref:v1:868554b6:/.../maven2/com/softwaremill/sttp/client3/core_2.13/3.8.3/core_2.13-3.8.3.jar",
  "qref:v1:f3ba6af6:/.../maven2/com/softwaremill/sttp/shared/ws_2.13/1.3.10/ws_2.13-1.3.10.jar",
  "qref:v1:438104da:/.../maven2/org/scala-lang/scala-library/2.13.8/scala-library-2.13.8.jar",
  ...
]
```

`ws` is upgraded to 1.3.10, but `scala-library` remains at 2.13.8.

### Constraints for Scala 2.13

In the example shown above, sbt willl fail the build and require the user to upgrade `scalaVersion` to 2.13.10. The reason is interactions between the compiler's compile-time and run-time classpath.

On a high level, the compiler is invoked as `java -cp RUNTIME_CLASSPATH scala.tool.nsc.Main -cp COMPILETIME_CLASSPATH Source.scala`. Both classpaths contain a scala-library.
  - The scala-library on the runtime classpath needs to match the compiler version (also on the runtime classpath) exactly because compiler releases are built with the inliner enabled.
  - The scala-library on the runtime classpath cannot be newer because
    - When running the REPL (`mill console`), REPL lines compiled against a newer library could fail to execute on the older library
    - Macros from the classpath compiled against a newer library could fail to expand on the older library

The implementation for Scala 2.13 in sbt is therefore
  - unpin scala-library in dependency resolution
  - fail the build if the Scala library is newer than the build's `scalaVersion`
[...]

PR #3185, opened 2024-05-25
Use newer resolve API from coursier

lefou:
As part of experimenting the solution space for feature request https://github.com/com-lihaoyi/mill/issues/3075, I found that some resolution validation errors are only correctly detected, when the newer coursier API is used. Therefore I extracted some commits of PR https://github.com/com-lihaoyi/mill/pull/3084 into this standalone PR and fixed any issues on the way.

Changes:
* Use newer coursier `Resolve` API to really detect conflicting dependencies
* Introduce a new `CoursierModule.enforceSameDependencyVersions`, which allows to sync versions for artifacts belonging to the same release, e.g. `scala-library` and `scala-reflect`.

### DIFF
--- a/main/util/src/mill/util/CoursierSupport.scala
+++ b/main/util/src/mill/util/CoursierSupport.scala
@@ -1,14 +1,15 @@
 package mill.util
 
-import coursier.cache.{ArtifactError, Cache, FileCache}
-import coursier.core.{Dependency, Module, Repository, Resolution}
+import coursier.cache.{ArtifactError, FileCache}
+import coursier.core._
+import coursier.error.ResolutionError
 import coursier.params.ResolutionParams
+import coursier.params.rule.{RuleResolution, SameVersion}
 import coursier.parse.RepositoryParser
-import coursier.util.{Gather, Task}
-import coursier.{Artifacts, Fetch, Resolve}
-import mill.api.{Ctx, PathRef, Result}
+import coursier.util.{Gather, ModuleMatcher, Task}
+import coursier.{Organization, Resolve}
 import mill.api.Loose.Agg
-import mill.api.PathRef.Revalidate
+import mill.api.{Ctx, PathRef, Result}
 
 import java.io.File
 import java.nio.file.NoSuchFileException
@@ -19,8 +20,6 @@ import scala.util.{Failure, Success, Try}
 
 trait CoursierSupport {
   import CoursierSupport._
-
-  private val useNewCoursierApi = true
 
   private val CoursierRetryCount = 5
   private val CoursierRetryWait = 100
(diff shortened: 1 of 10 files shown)


main/util/src/mill/util/CoursierSupport.scala
```diff
@@ -208,60 +227,126 @@ trait CoursierSupport {
...
+        ctx = ctx,
+        errorMsgExtractor = (r: Resolution) => r.errors.flatMap(_._2)
+      )(resolve.run)
 
-    val resolution =
-      retry(ctx = ctx, errorMsgExtractor = (r: Resolution) => r.errors.flatMap(_._2)) {
-        () => start.process.run(fetch).unsafeRun()
-      }
+      val resolution = resolution0
+        .withMapDependencies(mapDependencies)
```
lefou, line 314: I think this `mapDependencies` is applied too late. It must be applied before resolution instead. But I can't find any other way in the new `Resolve` API to pass a map function. @alexarchambault Do you have an idea?

main/util/src/mill/util/CoursierSupport.scala
```diff
@@ -208,60 +227,126 @@ trait CoursierSupport {
...
+      .map(d => mapDependencies.fold(d)(_.apply(d)))
       .map { d => d.module -> d.version }
       .toMap
 
-    val start0 = Resolution()
-      .withRootDependencies(
-        deps.iterator.map(mapDependencies.getOrElse(identity[Dependency](_))).toSeq
-      )
-      .withForceVersions(forceVersions)
-      .withMapDependencies(mapDependencies)
```
lefou, line 235: For this `withMapDependencies` I couldn't find an equivalent in the new API.

alexarchambault:
FYI I'm giving a try at this too, I should open a PR for it soon

lefou:
* https://github.com/com-lihaoyi/mill/pull/3513

main/util/src/mill/util/CoursierSupport.scala
lefou: Related?

* https://github.com/coursier/coursier/pull/3072

main/util/src/mill/util/CoursierSupport.scala
lefou: Related?

* https://github.com/coursier/coursier/pull/3072

main/util/src/mill/util/CoursierSupport.scala
alexarchambault: I hadn't realized how deeply ingrained in Mill `mapDependencies` was (it's basically exposed to Mill users). On the [sbt side](https://github.com/coursier/sbt-coursier), it's only used internally, so these uses were converted to more explicit parameters in [`ResolutionParams`](https://github.com/coursier/coursier/blob/36341ddc7701bd38c3f8ac0e0e89ac116a2e2921/modules/coursier/shared/src/main/scala/coursier/params/ResolutionParams.scala) (`typelevel`, `doForceScalaVersion`, `doOverrideFullSuffix`, `selectedScalaVersion`).

https://github.com/coursier/coursier/pull/3072 adds back the possibility for API users to transform dependencies this way, yes.

lihaoyi:
Closing in favor of https://github.com/com-lihaoyi/mill/pull/3513

lefou:
We probably want to add the `enforceSameDependencyVersions` target to `ScalaModule` in a separate PR.

PR #3195, opened 2024-05-30
[BSP] Support buildTarget/wrappedSources for build scripts

lolgab:
scala-cli supports a bespoke extension to the the build server protocol.
This extension allows the support of Scala scripts. Scala scripts
are Scala code which is then wrapped at build time by some Scala
code to make them valid Scala code.
scala-cli-bsp's buildTarget/wrappedSources allows build tools to
give the following information:
- the scala script file
- the generated scala file generated from the script
- the string that is added on the top of the file
- the string that is added on the bottom of the file

This PR adds Virtuslab's scala-cli-bsp library to the bsp worker module,
and implements the handler for the buildTarget/wrappedSources call.
To make the information available, it changes the return type of
`generateScriptSources` in `MillBuildRootModule` to return, together
with the `PathRef` of the destination folder, the four elements
described above, so it can be used by BSP.

### DIFF
--- a/bsp/src/mill/bsp/Constants.scala
+++ b/bsp/src/mill/bsp/Constants.scala
@@ -6,6 +6,6 @@ private[mill] object Constants {
   val bspProtocolVersion = BuildInfo.bsp4jVersion
   val bspWorkerImplClass = "mill.bsp.worker.BspWorkerImpl"
   val bspWorkerBuildInfoClass = "mill.bsp.worker.BuildInfo"
-  val languages: Seq[String] = Seq("scala", "java")
+  val languages: Seq[String] = Seq("scala", "java", "scala-sc")
   val serverName = "mill-bsp"
 }
--- a/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
+++ b/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
@@ -135,7 +135,7 @@ private class MillBuildServer(
 
       // TODO: scan BspModules and infer their capabilities
 
-      val supportedLangs = Seq("java", "scala").asJava
+      val supportedLangs = Seq("java", "scala", "scala-sc").asJava
       val capabilities = new BuildServerCapabilities
 
       capabilities.setBuildTargetChangedProvider(false)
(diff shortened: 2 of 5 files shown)


lefou:
That BSP extension looks new to me, which surprises me, since I try to follow the BSP development closely. I think making a bespoken BSP extension without a proper specification the status quo (by supporting it in scala-cli and a major BSP client in the Scala world) is a worst case scenario and will render the whole BSP idea to nonsense. We should strive for getting this extension into the BSP spec. Therefore, please provide a link to the BSP issue, where this extension is discussed. If there is none, we should create one and link all existing implementations and discussions. Further discussion how the extension should behave should continue there (e.g. not is some discord channels). as it looks like, the extension already developed, so all that is needed is to formalize it and properly support it with a capability flag.

lolgab:
@lefou Sure, I don't have anything against making the extension an official part of BSP.
My focus is first to make sure this works.
I believe it wasn't added as part of the protocol yet because it is an experimental feature and needed validation. Having two different build tools supporting it is validation that it is general enough to be added to the protocol.

lefou:
Yeah, but there is not a single trace in the BSP project, not even a discussion or a wiki. What a waste of developer resources (you never know who else fiddles with this stuff) and also miss of valuable input (Should we put this behind a capability? Yes. Should we define a new supported langauge for that? No.)

lolgab:
Started working on adding the message to the official BSP protocol: https://github.com/build-server-protocol/build-server-protocol/pull/673

lefou:
> Yeah, but there is not a single trace in the BSP project, not even a discussion or a wiki. What a waste of developer resources (you never know who else fiddles with this stuff) and also miss of valuable input (Should we put this behind a capability? Yes. Should we define a new supported langauge for that? No.)

Thank you! ❤️

lefou:
> Started working on adding the message to the official BSP protocol: [build-server-protocol/build-server-protocol#673](https://github.com/build-server-protocol/build-server-protocol/pull/673)

Thank you! ❤️

lihaoyi:
Closing due to inactivity

lefou:
Sorry @lolgab for me being naive and thinking the BSP project is the way to go. (Meanwhile, I think it's a corporate project serving the interest of some vendor. Decisions seem to happen somewhere else.)

I think we might support this feature nonetheless, as it might also fix the issue we have in Mill with our generated script sources. IDEA and probably Metals often jumps to the derived sources instead to the originals. If we can fix that for at least Metals users, this would be great.

PR #3196, opened 2024-06-03
Fix Scala.js toolchain logs in server-client mode

lolgab:
For some reason logging with `scala.Console` doesn't work with the Scala.js linker in server-client mode, maybe it's related to some classloader issue. This works around the problem by using `System.out` instead, which works correctly.

Pull Request: https://github.com/com-lihaoyi/mill/pull/3196

### DIFF
--- a/scalajslib/worker/1/src/mill/scalajslib/worker/ScalaJSWorkerImpl.scala
+++ b/scalajslib/worker/1/src/mill/scalajslib/worker/ScalaJSWorkerImpl.scala
@@ -17,7 +17,7 @@ import org.scalajs.linker.interface.{
   ModuleSplitStyle => _,
   _
 }
-import org.scalajs.logging.ScalaConsoleLogger
+import org.scalajs.logging.{Level, Logger}
 import org.scalajs.jsenv.{Input, JSEnv, RunConfig}
 import org.scalajs.testing.adapter.TestAdapter
 import org.scalajs.testing.adapter.{TestAdapterInitializer => TAI}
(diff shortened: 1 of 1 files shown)


scalajslib/worker/1/src/mill/scalajslib/worker/ScalaJSWorkerImpl.scala
```diff
@@ -158,6 +158,15 @@ class ScalaJSWorkerImpl extends ScalaJSWorkerApi {
       (linker, irFileCacheCache)
     }
   }
+  private val logger = new Logger {
+    def log(level: Level, message: => String): Unit = level match {
+      case Level.Warn | Level.Error => System.err.println(message)
+      case Level.Info | Level.Debug => System.out.println(message)
```
lefou, line 164: Should go all log messages to the same channel, STDERR? If a user wants to use STDOUT, she would use `println` instead of logging, I'd assume.

lolgab: This is what `ConsoleScalaLogger` from Scala.js does.
That would deviate from Sbt but it would mean that [the `"Starting process: node"`](https://github.com/scala-js/scala-js-js-envs/blob/d0683564ef64c86731e1037a30d5bec13c95b5fd/js-envs/src/main/scala/org/scalajs/jsenv/ExternalJSRun.scala#L146) will be printed to `stderr` and not to `stdout` anymore, which is nice.

I don't know what's best to be honest.

lefou: I think the message you mentioned is predestinated to end up on STDERR. It's just tooling side-output, not the output of the actual process to run.

lefou (approved):

PR #3201, opened 2024-06-06
[WIP] Experiment with jpackage to create executable binaries

lefou:
_Still in experimenting phase_

Created a new module `JpackageModule`, which wraps around the `jpackage` tool, that's part of newer JDK releases and is used to produced native executables and distributable packages.

Although, there are some configuration points, in most cases, it should be as easy usable as `assembly`.

```
> mill show foo.jpackage
[1/1] show > [93/93] foo.jpackageAppImage 
".../out/foo/jpackageAppImage.dest/out/foo/bin/foo"

> file out/foo/jpackageAppImage.dest/image/foo/bin/foo
out/foo/jpackageAppImage.dest/out/foo/bin/foo: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.18, not stripped
```

### DIFF
--- a/example/scalamodule/10-assembly-config/build.sc
+++ b/example/scalamodule/10-assembly-config/build.sc
@@ -1,7 +1,27 @@
 import mill._, scalalib._
 import mill.scalalib.Assembly._
 
-object foo extends ScalaModule {
+trait Jpackage extends JavaModule {
+  def jpackageName: T[String] = artifactName
+  def jpackage: T[PathRef] = T {
+    val localCp = localClasspath().map(_.path)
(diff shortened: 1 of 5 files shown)


lefou:
Based on https://dotnet.social/@fornever/112560707798240443, I tried to implement a generic `jpackage` target that can produce self-contained executables for any `JavaModule` (or `ScalaModule`). This is a WIP PR just to see what works.

lefou:
Merged as part of https://github.com/com-lihaoyi/mill/pull/4038

PR #3202, opened 2024-06-10
Add `usePipelining: T[Boolean]` to `ScalaModule`

lolgab:
Scala 3.5.0-RC1 added support for compilation pipelining

### DIFF
--- a/scalalib/api/src/mill/scalalib/api/ZincWorkerApi.scala
+++ b/scalalib/api/src/mill/scalalib/api/ZincWorkerApi.scala
@@ -49,7 +49,25 @@ trait ZincWorkerApi {
       incrementalCompilation = true
     )
 
+  def compileMixed(
+      upstreamCompileOutput: Seq[CompilationResult],
+      sources: Agg[os.Path],
+      compileClasspath: Agg[os.Path],
+      javacOptions: Seq[String],
+      scalaVersion: String,
+      scalaOrganization: String,
+      scalacOptions: Seq[String],
+      compilerClasspath: Agg[PathRef],
+      scalacPluginClasspath: Agg[PathRef],
+      reporter: Option[CompileProblemReporter],
+      reportCachedProblems: Boolean,
+      incrementalCompilation: Boolean,
+      auxiliaryClassFileExtensions: Seq[String],
+      usePipeling: Boolean
+  )(implicit ctx: ZincWorkerApi.Ctx): mill.api.Result[CompilationResult] = ???
+
   /** Compile a mixed Scala/Java or Scala-only project */
+  @deprecated("Use override with `usePipelining` parameter", "Mill 0.11.8")
   def compileMixed(
       upstreamCompileOutput: Seq[CompilationResult],
       sources: Agg[os.Path],
@@ -78,7 +96,8 @@ trait ZincWorkerApi {
       reporter = reporter,
       reportCachedProblems = reportCachedProblems,
       incrementalCompilation = incrementalCompilation,
-      auxiliaryClassFileExtensions = auxiliaryClassFileExtensions
+      auxiliaryClassFileExtensions = auxiliaryClassFileExtensions,
+      usePipeling = false
     )
 
   /** Compile a mixed Scala/Java or Scala-only project */
@@ -109,7 +128,8 @@ trait ZincWorkerApi {
       reporter = reporter,
       reportCachedProblems = reportCachedProblems,
       incrementalCompilation = true,
-      auxiliaryClassFileExtensions = Seq.empty[String]
+      auxiliaryClassFileExtensions = Seq.empty[String],
+      usePipeling = false
     )
 
   /** Compile a mixed Scala/Java or Scala-only project */
@@ -141,7 +161,8 @@ trait ZincWorkerApi {
       reporter = reporter,
       reportCachedProblems = reportCachedProblems,
       incrementalCompilation = incrementalCompilation,
-      auxiliaryClassFileExtensions = Seq.empty[String]
+      auxiliaryClassFileExtensions = Seq.empty[String],
+      usePipeling = false
     )
 
   /**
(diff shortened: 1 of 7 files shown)


lefou (commented):
Are there any (known) downsides to use pipelining = `true`? If not, could we make it the default?
What happens, if pipelining is enabled for older Scala versions?
Do you have any link to (official) documentation of the pipelining feature and how it works?

The "pipelining" feature I know about (for some years) is based on starting downstream module compilation before upstream modules are compiled completely. It's essentially a more fine-grained dependency resolution based on compiler phases. But I fail to see how that can work in Mill with the current design and this PR.

lihaoyi:
IIRC pipelining can slow things down in some cases, and it's still pretty experimental so probably won't be ready to be made the default for a while yet

scalalib/api/src/mill/scalalib/api/ZincWorkerApi.scala
```diff
@@ -49,7 +49,25 @@ trait ZincWorkerApi {
...
+      scalaOrganization: String,
+      scalacOptions: Seq[String],
+      compilerClasspath: Agg[PathRef],
+      scalacPluginClasspath: Agg[PathRef],
+      reporter: Option[CompileProblemReporter],
+      reportCachedProblems: Boolean,
+      incrementalCompilation: Boolean,
+      auxiliaryClassFileExtensions: Seq[String],
+      usePipeling: Boolean
+  )(implicit ctx: ZincWorkerApi.Ctx): mill.api.Result[CompilationResult] = ???
```
lefou, line 67: It's probably time to use a parameter class instead of adding new parameters and deprecated `def`s. That way, we only need to extends the API in one place. We could also move validation into that class, if needed.

lihaoyi:
I also wonder whether the config change in this PR is enough. Does the build tool need to do anything to integrate with it? Or is it all handled transparently by zinc?

lefou:
> I also wonder whether the config change in this PR is enough. Does the build tool need to do anything to integrate with it? Or is it all handled transparently by zinc? Can we demonstrate some speedup experimentally to show that it's actually working?

First we should link to the correct "pipelining" feature. If it's the in deed one I think it is, then this change won't be enough. It would only start to take effect, if we start compilation of multiple modules quasi-parallel, so zinc can decide when enough upstream compilation was done to start some downstream compilation. But maybe, there are multiple developments under the "pipelining" concept/term, hence my questions.

lolgab:
> > I also wonder whether the config change in this PR is enough. Does the build tool need to do anything to integrate with it? Or is it all handled transparently by zinc? Can we demonstrate some speedup experimentally to show that it's actually working?
> 
> First we should link to the correct "pipelining" feature. If it's in deed the one I think it is, then this change won't be enough. It would only start to take effect, if we start compilation of multiple modules quasi-parallel, so zinc can decide when enough upstream compilation was done to start some downstream compilation. But maybe, there are multiple developments under the "pipelining" concept/term, hence my questions.

I started implementing this PR in "auto-pilot" without thinking too much, but you are totally right! For this to work correctly, zinc needs to be able to start compiling the downstream modules before the upstream ones end, which of course doesn't happen for how Mill works.
This makes me think that Mill should have a different task graph if `usePipelining` is enabled or not, but that means that should be a `Boolean` instead of `Target[Boolean]`, which will break the cache. Maybe this requires running an evaluation to know what to run.
What I know for sure is that I greatly underestimated the effort and changes required for pipelining to work.
Maybe @bishabosha can give some input here, since he implemented pipelining in the Scala 3 compiler.
It seems that [this code in Sbt](https://github.com/sbt/sbt/blob/586e0a752cd5d0f0335deaa910c4355e0e0a0e56/main/src/main/scala/sbt/internal/ClasspathImpl.scala#L36-L50) decides the flow to take via a conditional task.

Maybe Mill should support some kind of conditional tasks?

lefou:
> Maybe Mill should support some kind of conditional tasks?

We would loose a lot of other properties if we go that route, since modifying the task graph while it is already running would require a monadic approach. I too sometimes wished we had this feature, but in most cases I found a satisfying other solution. Being able to make some decisions before returning the final task is tremendously helpful. 

Since compilation requires a lot of prerequisites, which should be managed by Mill, I don't think we should by-pass any Mill task semantics to enable pipelining. There is no point in starting a compilation which has still other unresolved dependencies. 

What I could imaging is to split the compilation model into a `zincCompilation: T.worker` of of the known `compile: T`. The worker encapsulation the running zinc process and the `compile` only holding the result once compilation has finished. Downstream compilations could only depend on the other workers instead of the final result. That way, we could manage multiple compilations with or without pipelinening via the worker, but keep all other tasks and machinery as before.

lihaoyi:
> That way, we could manage multiple compilations with or without pipelinening via the worker, but keep all other tasks and machinery as before.

That sounds like it could work. The `T.worker` basically can encapsulate arbitrary in-memory state and logic. 

I could imagine having two targets `def compile` and `def compileOutline` (or whatever), where most use cases still use `compile`, but downstream pipelined compiles can depend on `compileOutline`, and both `compile` and `compileOutline` are just stubs wrapping the `T.worker`'s compilation logic that happens on another thread and provides results to the two targets as they become available.

There is one small quirk since we'd be blocking threads in `compile` or `compileOutline` waiting for the logic inside `T.worker` to complete. Given the O(10s) of threads used in a typical build that's totally fine from a performance perspective and concurrency.

lefou:
> Given the O(10s) of threads used in a typical build that's totally fine from a performance perspective and concurrency.

Yeah. A build with is configured to only use one worker thread, shouldn't be able to profit from pipelined compilation, which sounds correct. 

We could develop pipelined compilation in a `@experimental` separate trait, so we don't risk anything while figuring out the best approach, but can roll it out early to testers.

lefou (changes requested):
This isn't working as-is.

PR #3208, opened 2024-06-14
Support `build.sc` in Bloop

lolgab:
After Mill `0.11` `mill-contrib-bloop` stopped supporting the build module.
Now `BloopImpl` gets a sequence of `Evaluator`s and gets all modules for all the evaluators in `Evaluator.allBootstrapEvaluators`, so it supports the build module again.

Pull Request: https://github.com/com-lihaoyi/mill/pull/3208

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -1447,6 +1447,7 @@ object dev extends MillPublishScalaModule {
   def testTransitiveDeps = super.testTransitiveDeps() ++ Seq(
     runner.linenumbers.testDep(),
     scalalib.backgroundwrapper.testDep(),
+    contrib.bloop.testDep(),
     contrib.buildinfo.testDep(),
     contrib.scoverage.testDep(),
     contrib.scoverage.worker2.testDep(),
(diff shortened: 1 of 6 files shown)


lefou (commented):
I'd like to seem some test coverage for a meta-build support.

lefou (commented):
The newly added test isn't really using a meta-build.

lolgab:
You are right! I was confusing the normal mill build with the meta-build.
After Mill 0.11 `mill-contrib-bloop` stopped supporting the build module.
This gets all modules for all the evaluators, so it supports the build module again.
Maybe I should change the PR description since what I did is not specific to meta builds.
What do you think?

lefou:
> You are right! I was confusing the normal mill build with the meta-build. After Mill 0.11 `mill-contrib-bloop` stopped supporting the build module. This gets all modules for all the evaluators, so it supports the build module again. Maybe I should change the PR description since what I did is not specific to meta builds. What do you think?

Yeah, makes sense. If you can move this message into a PR description, we should be good to merge it. 

I assume full support for meta-builds should be around the corner, just by the fact that the bootstrap build works. What's needed for confidence is a test case with a meta-build. But this can be another PR.

lefou (approved):

PR #3211, opened 2024-06-15
Fix classpath regression in Bloop export #1918

### LINKED ISSUE #1918: Classpath regression in Bloop export
There seems to be a regression caused by #1843 where now when you do a Bloop export for the Mill codebase itself, bloop is unable to build the contrib modules. This was reported before in #990 and fixed in the past.

### How to reproduce

1. Clone the mill code base
2. `./mill --import ivy:com.lihaoyi::mill-contrib-bloop: mill.contrib.bloop.Bloop/install` on the mill codebase
3. `bloop compile contrib.bloop`
4. Notice a bunch of compilation errors

You can then switch Mill back to 0.10.3, repeat the above and see that everything works fine.

### Difference in Bloop files

Between 0.10.3 and 0.10.4 you can see that the bloop export is different. Here are the missing classpath entries that the `contrib.bloop` target expects, but they aren't there:

```
<$PWD>/.bloop/out/scalalib/classes
<$PWD>/.bloop/out/scalajslib/classes
<$PWD>/.bloop/out/scalanativelib/classes
<$PWD>/.bloop/out/main/classes
<$PWD>/.bloop/out/scalalib.api/classes
<$PWD>/.bloop/out/testrunner/classes
<$PWD>/.bloop/out/main.core/classes
<$PWD>/.bloop/out/main.client/classes
<$PWD>/.bloop/out/main.moduledefs/classes
<$PWD>/.bloop/out/main.api/classes
<$PWD>/.bloop/out/main.util/classes
<$PWD>/.bloop/out/scalajslib.worker-api/classes
<$PWD>/.bloop/out/scalanativelib.api/classes
```

### Potential fix

Taken from https://github.com/com-lihaoyi/mill/issues/990#issuecomment-1174722396

> I think a proper implementation for the classpath and runtimeClasspath to be used in Bloop should not hand-roll the same logic found in JavaModule and ScalaModule but instead it should re-use the bspCompileClasspath target, which is already exactly doing this (and is meant to be an maintained abstraction to be used by BSP). It returns an UnresolvedPath, which is either a local path (which can be re-used in Bloop, e.g. a resolved depenency) or a path relative to another target. E.g. the compiler output of a dependencies is some classes/ sub path inside the dest dir of that compile target. When generating the Bloop config, we need to map these unresolved paths to the real compile output paths Bloop used for these modules. All information should be already present in the BloopImpl to implement this mapping. But it's not a trivial change, so I won't do it right now.

lolgab:
Fixes #1918.

Pull Request: https://github.com/com-lihaoyi/mill/pull/3211

### DIFF
--- a/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
@@ -201,21 +201,25 @@ class BloopImpl(ev: () => Evaluator, wd: os.Path) extends ExternalModule { outer
     // //////////////////////////////////////////////////////////////////////////
 
     val classpath = T.task {
-      val depModules = (module.compileModuleDepsChecked ++ module.recursiveModuleDeps).distinct
-      // dep modules ++ ivy deps ++ unmanaged
-      depModules.map(classes) ++
-        module.resolvedIvyDeps().map(_.path) ++
-        module.unmanagedClasspath().map(_.path)
+      val transitiveCompileClasspath = T.traverse(module.transitiveModuleCompileModuleDeps)(m =>
+        T.task { m.localCompileClasspath().map(_.path) ++ Agg(classes(m)) }
+      )().flatten
+
+      module.resolvedIvyDeps().map(_.path) ++
+        transitiveCompileClasspath ++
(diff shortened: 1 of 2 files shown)


lefou (approved):
Looks reasonable.

Issue #3207, opened 2024-06-14
Automatically exclude Mill transitive depedencies from build classpath


lolgab:
Plugins should add Mill dependencies as `provided`, but sometimes they don't and Mill breaks when you update Mill to a newer version which adds new fields to traits.
Mill could solve the problem by automatically add exclusion rules for `mill-scalalib` and its other artifacts when adding dependencies to the build via `import $ivy.`.

lolgab:
This was already fixed in https://github.com/com-lihaoyi/mill/pull/3189.
I just triggered a new snapshot release for it.

lolgab:
I tried with the latest snapshot but it doesn't seem to be solved.
Reproduction:
```scala
// build.sc
import $ivy.`com.disneystreaming.smithy4s::smithy4s-mill-codegen-plugin::0.18.22`

import mill._
import mill.scalalib._

object root extends RootModule with ScalaModule {
  def scalaVersion = "3.4.2"
}
```

.mill-version
```
0.11.7-118-56323c
```

Errors with:
```
java.lang.AbstractMethodError: Receiver class millbuild.build$root$ does not define or inherit an implementation of the resolved method 'abstract mill.define.ModuleRef mill$scalalib$JavaModule$$super$zincWorker()' of interface mill.scalalib.JavaModule.
    mill.scalalib.JavaModule.zincWorker(JavaModule.scala:38)
    mill.scalalib.JavaModule.zincWorker$(JavaModule.scala:38)
    millbuild.build$root$.zincWorker(build.sc:27)
    mill.scalalib.ScalaModule.$anonfun$compile$1(ScalaModule.scala:287)
    scala.collection.mutable.HashMap.getOrElseUpdate(HashMap.scala:469)
    mill.moduledefs.Cacher.cachedTarget(Cacher.scala:10)
    mill.moduledefs.Cacher.cachedTarget$(Cacher.scala:9)
    mill.define.Module$BaseClass.cachedTarget(Module.scala:58)
    mill.scalalib.ScalaModule.compile(ScalaModule.scala:280)
    mill.scalalib.ScalaModule.compile$(ScalaModule.scala:280)
    millbuild.build$root$.compile(build.sc:27)
    java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #3221, opened 2024-06-20
Improve Mill plugin classpath exclusions

lefou:
Also refer to transitive modules coordinates in the `BuildInfo.millEmbededDeps` data field.

Added an integration test to make sure it's effectively excluding transitive Mill dependencies.

Fix https://github.com/com-lihaoyi/mill/issues/3207

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -590,7 +590,7 @@ object main extends MillStableScalaModule with BuildInfo {
       "millEmbeddedDeps",
       (
         T.traverse(
-          dev.transitiveModuleDeps.collect { case m: PublishModule => m }
+          dev.recursiveModuleDeps.collect { case m: PublishModule => m }
         )(
           _.publishSelfDependency
         )()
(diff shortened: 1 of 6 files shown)


lolgab:
There is some commented out code. Is it supposed to be there?

lefou:
> There is some commented out code. Is it supposed to be there?

Kind of. It's the rest of the full transitive dependency chain, which I preserved/snapshot-ed that way. I need to make up my mind, if we need to handle/check them too in the test, or not.

lefou:
> > There is some commented out code. Is it supposed to be there?
> 
> Kind of. It's the rest of the full transitive dependency chain, which I preserved/snapshot-ed that way. I need to make up my mind, if we need to handle/check them too in the test, or not.

I removed them for now.

lolgab:
You can also merge main or rebase

lolgab (approved):

PR #3249, opened 2024-07-04
Improve web build documentation

arturaz:


lihaoyi:
Probably going to call this a wontfix. We can't really afford to duplicate all the minutae of the class hierarchy and task graph in the docs. Anyone who wants to see what it does can jump to definition in IntelliJ

arturaz:
The problem is that docs do not state that this trait is only available in certain scope, which is super confusing when you get a compilation error.

arturaz:
And you can't jump to definition because the symbol is unresolved.

Issue #3132, opened 2024-04-25
`docker.build` should invalidate results when environment changes


lolgab:
Docker images are built when the environment changes, since Docker depends on some environment variables.

Docker documentation [here](https://docs.docker.com/engine/reference/commandline/cli/#environment-variables).

Mill should invalidate the docker `build` target when the environment changes.

lefou:
Since the `DockerModule` already defines a `envVars` target, I'd argue that the current implementation assumes they are the only used env vars at play. This is probably not correct, since os-lib's `os.proc` is silently using all default env vars of the Java process if the `propagateEnv` parameter isn't set to `false`. So a fix should probably set it. If a user wants to depend on all actual env vars, she already can override `envVars` with `T.input { sys.props }`.

PR #3257, opened 2024-07-06
 Add `dockerEnv` target to customize environment passed to `docker` command

lolgab:
Fixes https://github.com/com-lihaoyi/mill/issues/3132

Pull Request: https://github.com/com-lihaoyi/mill/pull/3257


lefou (commented):
Do I understand correctly, that this `env` is about the invocation of the `docker` command? Since we now have two targets containing some `env`s in the docker module, we should clarify how they differentiate.

lolgab:
I already added some scaladoc to the new command. They also have links to two different pages in docker documentation

lefou (approved):
Looks good to me.

The `dockerEnv` name is a bit misleading, but mostly because docker has a build-time and a runtime. This module is about building docker images but is called just "Docker". I think it's consistent enough for now.

Issue #2985, opened 2024-01-23
Attempt to update Upickle results in linkage error


kubukoz:
Hi!

When trying to update the version of Upickle being used in the app, I'm running into a linkage error: a `java.lang.NoSuchMethodError`.

The problem appears to be caused by the Mill launcher putting Mill's own jar dependencies on the classpath first, which results in the newer version (mine) being evicted.

To reproduce:

1. `mill-build/build.sc`
```scala
import mill._

import scalalib._

object millbuild extends MillBuildRootModule {

  override def ivyDeps = T {
    // Mill 0.11.6 pulls in 3.1.3 here
    List(ivy"com.lihaoyi::upickle:3.1.4")
  }

  override def runIvyDeps = T {
    List(ivy"com.lihaoyi::upickle:3.1.4")
  }

}
```

2. `build.sc`
```scala
import mill._

import upickle.default._
import $meta._

object demo extends mill.Module {

  final case class Demo(
  )

  object Demo {
    implicit val readWriter: ReadWriter[Demo] = macroRW
  }

  def compile() = T.command {
    println(read[Demo]("""{"extra": 42}"""))

    ()
  }

}
```

3. Setup `millw`
4. `./mill-version`: `0.11.6`
5. `./millw --no-server demo.compile`

Full reproduction: https://github.com/kubukoz/demos/tree/mill-upickle-bump-bincompat

Stack trace:

```
1 targets failed
demo.compile java.lang.NoSuchMethodError: 'boolean upickle.default$.allowUnknownKeys()'
    millbuild.build$demo$Demo$$anon$1$$anon$2.visitKeyValue(build.sc:12)
    ujson.CharParser.visitStringKey(CharParser.scala:650)
    ujson.CharParser.parseStringKey(CharParser.scala:626)
    ujson.CharParser.parseNested(CharParser.scala:390)
    ujson.CharParser.parseTopLevel0(CharParser.scala:339)
    ujson.CharParser.parseTopLevel(CharParser.scala:323)
    ujson.CharParser.parse(CharParser.scala:72)
    ujson.StringParser$.transform(StringParser.scala:28)
    ujson.StringParser$.transform(StringParser.scala:27)
    ujson.Readable$fromTransformer.transform(Readable.scala:16)
    upickle.Api.$anonfun$read$1(Api.scala:38)
    upickle.core.TraceVisitor$.withTrace(TraceVisitor.scala:18)
    upickle.Api.read(Api.scala:38)
[...]

Issue #2703, opened 2023-08-24
Rework handling of `mill-launcher` in `MillBuildModule`


lefou:
Currently, we use the Mill launcher jar to compile the Mill project. To do that, we add it to the `unmanagedClasspath` of the `MillBUildModule`. Technically, this is correct. But in comparison with dependencies defined via `ivyDeps` we loose some semantics we'd like to keep. For example, when editing the build sources files, we don't have properly attached sources, as there is no metadata to use to get the sources jar on demand.

Instead we could define the dependency under `ivyDeps`. There are some issues to that approach, too:
* We need to download the jar (if not present), although we already have it locally as `mill-launcher`
* Download can fail when offline, which is a no-go, as we always want be able to build a Mill build without internet access
* When a locally build development version is used, a published jar with the same version may not exist

To overcome all of those issues, we could intercept the `ivyDeps` resolution of the binary `mill-launcher` and return the already present jar.

lefou:
As an alternative, we could embed the source into the `mill-launcher`. This will make the jar file larger, but will always provide the correct sources files for lookup.

lihaoyi:
We should consider removing `mill-launcher` from the classpath and having a smaller explicitly-defined classpath. We don't want everything that Mill uses internally to be on the classpath and accessible from user `build.sc` files. 

Things like OS-Lib, uPickle, Requests-Scala, etc. probably need to be there, but we don't want random things like `org.ow2.asm` or `com.kohlschutter.junixsocket` to end up there for users to use unless they explicitly import it

lihaoyi:
Now that we have Coursier bundled in the Mill client, both native and JVM versions, we could conceivably avoid using an assembly-jar at all and instead use the Mill client to resolve the jars necessary for the Mill server. 

However, there then comes the problem that the jars necessary to run the Mill server can only be known after running the meta-build, and each layer of the meta-build may require different sets of jars at different versions.

Each level of the meta build is already in an isolated classloader, but shares the Evaluator and other Mill classes. To resolve the issue above, we would need to move the Evaluator for each level of the meta-build into the corresponding isolated classloader, with a much thinner enclosing classpath that just orchestrates these isolated classloaders and interacts with them via the bare minimum set of classes and interfaces.

lihaoyi:
Aiming to get this done before 0.13.0

PR #3260, opened 2024-07-13
Make build.sc dependency resolution independent of Mill assembly

lihaoyi:
Should fix https://github.com/com-lihaoyi/mill/issues/2985 and https://github.com/com-lihaoyi/mill/issues/2703, at least as a first pass. 

- I did not change the actual classpath contents to try and minimize it; that can come in a follow up PR
- I did not try to cache or bundle the dependencies locally; that can also come in a follow up

At least with this change, all the Mill `build.sc` dependencies go through the same code path (Coursier resolution) and behave similarly (e.g. you should be able to update upickle via `import $ivy`. Further improvements to hygiene or performance can come in follow ups (e.g. resolving dependencies bundled in the assembly)

Notes:

- Changed `MillBuildRootModule#ivyDeps` to `mandatoryIvyDeps` to prevent accidental overriding (not sure why this didn't cause problems before??? Maybe people just didn't override `ivyDeps` much in the meta-build)

- Removed `resolveDepsExclusions`, since now we are no longer hard-coding a dependency on the Mill assembly, we also no longer need to treat these dependencies specially and they can be upgraded or evicted as the user sees fit

- Removed `Lib.resolveMillBuildDeps` and its use sites, since now the build dependencies go through the same code path as any other `ivyDeps`

Tested the `mill.idea.GenIdea/idea` workflow manually in `example/basic/1-simple-scala` and `example/misc/4-mill-build-folder` manually, the intellij jump-to-definition in `build.sc` and `mill-build/build.sc` seem to work

### DIFF
--- a/scalalib/src/mill/scalalib/Lib.scala
+++ b/scalalib/src/mill/scalalib/Lib.scala
@@ -150,7 +150,6 @@ object Lib {
     Util.millProperty("MILL_BUILD_LIBRARIES") match {
       case Some(found) => found.split(',').map(os.Path(_)).distinct.toList
       case None =>
-        millAssemblyEmbeddedDeps
         val Result.Success(res) = scalalib.Lib.resolveDependencies(
           repositories = repos.toList,
           deps = millAssemblyEmbeddedDeps,
--- a/scalalib/src/mill/scalalib/Lib.scala
+++ b/scalalib/src/mill/scalalib/Lib.scala
@@ -4,8 +4,6 @@ package scalalib
 import coursier.util.Task
 import coursier.{Dependency, Repository, Resolution}
 import mill.api.{Ctx, Loose, PathRef, Result}
-import mill.main.BuildInfo
-import mill.util.Util
 import mill.scalalib.api.ZincWorkerUtil
 
 object Lib {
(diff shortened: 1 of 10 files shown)


scalalib/src/mill/scalalib/Lib.scala
```diff
@@ -150,7 +150,6 @@ object Lib {
     Util.millProperty("MILL_BUILD_LIBRARIES") match {
       case Some(found) => found.split(',').map(os.Path(_)).distinct.toList
       case None =>
-        millAssemblyEmbeddedDeps
```
lihaoyi, line 153: This line is dead code, but more broadly not sure if this `def resolveMillBuildDeps` method is still correct or whether it or its use sites need to be updated

lefou (commented):
I think there should be a test case covering the (accidential) downgrade of a Mill dependency (e.g. we run Mill `0.11.8` but there is a explicit `mill-scalalib-0.11.6` or a `os-lib-0.9.3`). It should assert, there is no older version present.

lihaoyi:
Added a test, which fails to demonstrate the current code changes are not enough

The issue as of now is that even without the mill jar added to the classpath as part of `MillBuildRootModule`, it still ends up on the classpath as part of the classloader hierarchy when we spawn sub-classloaders to evaluate each `build.sc` file. 

Removing that connection to make each spawned classloader independent of the parent will require some more work, because as of now the sub-classloaders and parent classloader share a `RootModule` interface along with everything `RootModule` depends on. This will need to be narrowed down significantly to make it possible to make the spawned classloaders independent, and thus the `build.sc` classpath independent of the Mill binary

lefou:
Maybe, we can implement this specific test as an server integration test, which should already run in a spawned, thus isolated, Java process?

lihaoyi:
In this case its not the added test that is a problem, but the festure itself. It seems like it would be a bunch more work to actually isolate the build.sc classpath

lihaoyi:
The issue with using this PR as is is that because the dependencies may have different versions, your IDE may pull the wrong sources (you'd get the resolved version sources rsther than the pinned version sources)

As I imagined it, if we fully supported resolving Mill's own dependencies via coursier rather than using the bundled versions, people `import $ivy` stuff may end up with arbitrarily different versions of stuff on the classpath vs what they have now depending on how dependency resolution goes

lefou:
All classloaders I know repect the order of the entries, so we could try to put our "provided" dependencies first on the classpath. Another option would be to use version ranges or forced versions like `[0.10.2]` in the `ivyDeps`. Unfortunately, the old coursier API we're using doesn't properly report mismatches, and my attempts to replace it with the newer API have other issues (see https://github.com/com-lihaoyi/mill/pull/3185).

lihaoyi:
The issue here is that the build.sc classloader *has* to share classes with the Mill binary classloader, because they interact via Scala class instances. They *have* to be using the same classes

So although setting `parent = null` to fully isolate them is trivial, we would then also need to completely rewrite the interactions to be through thin interfaces, rather than the fat Module/RootModule/Task classes we have today. 

It's possible, but would take more work, and is not as simple as just changing the classpath on the MillBuildRootModule

lefou:
If you use forced versions (Maven term, I think, like `[0.11.8]`) or ranges (`[0.11.8,0.11.8]`) to pin a version, coursier will fail, if transitive selected versions don't match that range.

Documentation: https://get-coursier.io/docs/other-version-selection

Example: https://github.com/com-lihaoyi/mill/issues/3075#issuecomment-1998525943

But it does not work in our current implemntation, due to Mill using the "low level" coursier API.

In PR https://github.com/com-lihaoyi/mill/pull/3185 I therefore replaced that API with the "high level" API, but I struggle to find a proper way to forward mapped dependencies to that API.

PR #3270, opened 2024-07-18
Fix resolution of children of `override lazy val` modules

lihaoyi:
Filtering out `abstract` methods in `Reflect` seems unnecessary, since compilation checks that every method is implemented already


lefou (approved):
Looks good to me.

PR #3275, opened 2024-07-19
Automatically pump `os.proc` streams when `SystemStreams` are redirected

lihaoyi:
Depends on https://github.com/com-lihaoyi/os-lib/pull/283

This moves the subprocess stream handling logic out of `Jvm.spawnSubprocess` and makes it apply to all `os.proc` invocations, greatly reducing the room for error. With this, `Jvm.spawnSubprocess` becomes a very thin wrapper around `os.proc.spawn`. We also rely directly on OS-Lib's own pumper threads to pump to our destination, rather than having them pump into in-memory buffers and then spawning our own pumper threads to pump from those buffers to the destination

I spent some time looking into how to do the stdout/err handling at the process level, but couldn't find any reasonable mechanism to do so that allows us to preserve the ordering of the stdout/stderr. This is the original motivation to squishing it into one stream via `ProxyOutputStream`/`ProxyStreamPumper` and is important because otherwise you find e.g. stack traces out of order with printlns, which makes debugging very difficult. Might be possible using some socket/fifo/pipe cleverness, but not as part of this PR

Added an integration test to assert on the subtleties of stdout, stderr, and their inherited alternatives. This PR is required for the test to pass

### DIFF
--- a/main/client/src/mill/main/client/InputPumper.java
+++ b/main/client/src/mill/main/client/InputPumper.java
@@ -2,29 +2,34 @@ package mill.main.client;
 
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.function.Supplier;
 
 public class InputPumper implements Runnable{
-    private InputStream src;
-    private OutputStream dest;
+    private Supplier<InputStream> src0;
+    private Supplier<OutputStream> dest0;
+
     private Boolean checkAvailable;
     private java.util.function.BooleanSupplier runningCheck;
-    public InputPumper(InputStream src,
-                       OutputStream dest,
+    public InputPumper(Supplier<InputStream> src,
+                       Supplier<OutputStream> dest,
                        Boolean checkAvailable){
         this(src, dest, checkAvailable, () -> true);
     }
-    public InputPumper(InputStream src,
-                       OutputStream dest,
+    public InputPumper(Supplier<InputStream> src,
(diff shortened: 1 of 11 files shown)


main/client/src/mill/main/client/InputPumper.java
```diff
@@ -2,29 +2,34 @@
...
-    private OutputStream dest;
+    private Supplier<InputStream> src0;
+    private Supplier<OutputStream> dest0;
+
     private Boolean checkAvailable;
     private java.util.function.BooleanSupplier runningCheck;
-    public InputPumper(InputStream src,
-                       OutputStream dest,
+    public InputPumper(Supplier<InputStream> src,
+                       Supplier<OutputStream> dest,
```
lihaoyi, line 14: Making these lazily-initialized is necessary to avoid infinite recursion when initializing the pumper <-> stream <-> subprocess data structure, which is circular

lihaoyi:
I think this should be ready for review @lefou

lefou (commented):
Looks good to me. Due to lack of test coverage, we should test this at least once in a BSP session.

lihaoyi:
Tested the BSP integration via intellij on example/scalabuilds/9-realistic. Seems to load successfully with some `println`s and `os.proc` calls added to the necessary tasks

PR #3284, opened 2024-07-22
Fix ci/mill.bat typo

sake92:


lefou (approved):

lefou (commented):
Thank you!

I think we should just update to the latest version `0.4.12` of `millw.bat`, where is already fixed.

But it's interesting, that this PR now also shows some build failured in the assembly test. Could be, that we previously always used the mill binary from Github, and your fix now results in the binary from Maven Central. Maybe they differ?

sake92:
Yeah, not sure to be honest.  
Havent been following mill dev for quite a while.. :/  
I will take a look at the failing tests, if that other PR also has failures.

Issue #3283, opened 2024-07-22
Windows CI seems to be broken on newer versions of Mill (500USD)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 500USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

Ever since https://github.com/com-lihaoyi/mill/commit/126f57476ff9b7ab3f797e41e8880ace81b85362, the various Windows CI jobs seem to be finishing in ~1min, rather than the 10-30min they used to. Something's probably wrong and they're not actually testing anything. We should figure out what, fix it, and ensure Windows CI is able to pass on the current latest version of Mill built from HEAD

lefou:
This PR only fixed the Windows batch script but doesn't raise the Mill version, yet it shows the same test failure:

* https://github.com/com-lihaoyi/mill/pull/3284 

We should compare the two binary sources Github releases and Maven Central as well as the batch script `mill.bat` which is a copy of https://github.com/lefou/millw.

lefou:
* [ ] #3290

sake92:
@lihaoyi thanks! I've sent you an email.

PR #3285, opened 2024-07-22
Update mill.bat script including some fixes

lefou:


lefou:
Supersedes #3284

lihaoyi (approved):

lefou:
It's interesting, that this PR now also shows some build failures in the assembly test. Looks exactly as issue #3283, but with a different reproduction:

* https://github.com/com-lihaoyi/mill/issues/3283

Could be, that we previously always used the mill binary from Github, due to a bug in the script, and the newer version now uses binaries from Maven Central. Maybe they differ?

lefou:
So, I think we discovered that there are more issues in Windows CI then just the assembly test when we now have a fix for. Unfortunately, I already merged this, so we have a failing CI pipeline in `main`.

lihaoyi:
No worries, we can just ignore the windows CI jobs for now and merge regardless

PR #3287, opened 2024-07-23
Fix AssemblyTests on Windows

sake92:
Depends on #3285  

Windows has an env var called `PATHEXT` whish is usually something like this: `.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC`  
Basically a list of extensions considered as executable.  
So the file in Mill's case needs to be a `.bat` file.


lefou:
Don't know why GH closed this one. @sake92 Can you please reopen it?

PR #3288, opened 2024-07-23
Fix AssemblyTests on Windows

lefou:
Depends on #3285 

Windows has an env var called `PATHEXT` whish is usually something like this: `.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC`  
Basically a list of extensions considered as executable.  
So the file in Mill's case needs to be a `.bat` file.

Issue analyzed and fixed by @sake92 

Fix https://github.com/com-lihaoyi/mill/issues/3283


lefou (commented):
@sake92 Thank you for the fix! Can you please format it and also bump the Mill version to `0.11.9` so we see it succeeds? Thank you

sake92:
@lefou where do I bump the mill version, in `.config\mill-version` ?

lefou:
> @lefou where do I bump the mill version, in `.config\mill-version` ?

Yes.

sake92:
Seems like scalalib tests are green now.  
https://github.com/com-lihaoyi/mill/actions/runs/10060348007

lefou:
Looks like there are some new unrelated test failures for Windows now. Since we now know, the assembly tests are fixed by this PR, we could remove the Mill bump,  and merge this PR (and close the bounty, I guess).

.config/mill-version
```diff
@@ -1 +1 @@
-0.11.7-17-17dea9
+0.11.9
```
lefou, line 1: Try to avoid the possibly unrelated test errors.

```suggestion
0.11.8
```

sake92:
> Looks like there are some new unrelated test failures for Windows now. Since we now know, the assembly tests are fixed by this PR, we could remove the Mill bump, and merge this PR (and close the bounty, I guess).

I guess it would be fair to make the CI green first, before closing the bounty. 😁

lefou:
> > Looks like there are some new unrelated test failures for Windows now. Since we now know, the assembly tests are fixed by this PR, we could remove the Mill bump, and merge this PR (and close the bounty, I guess).
> 
> I guess it would be fair to make the CI green first, before closing the bounty. 😁

Yeah. That was the idea by not bumping Mill to cause unrelated issues.

lefou:
I think the other test failure are valid but unrelated. They probably went under the radar since our Windows CI wasn't fully functional before #3285 .

sake92:
@lefou we can merge this one?  
It is easier to fix them one by one I think.

PR #3294, opened 2024-07-26
Overhaul IDE and Installation page

lihaoyi:
* Added screenshots to the getting started steps for IntelliJ and VSCode

* Put IDE support first, on the assumption that most people will be using an example project with a bootstrap script already set up

* Put global installation last, since it's both very verbose and also unlikely to be used (since we encourage bootstrap scripts)

* Removed a bunch of the command line bloop/bsp instructions, since this is now done transparently by the IDE and wont be of interest to users

* Removed some of the blurb about `millw`. Having two bootstrap scripts cannot avoid being confusing, we should merge them so there is only one choice going forward.


lefou:
> * Added screenshots to the getting started steps for IntelliJ and VSCode

Nice

> * Put IDE support first, on the assumption that most people will be using an example project with a bootstrap script already set up

This assumption don't really match my personal taste or how I digest/consume products.

> * Removed some of the blurb about `millw`. Having two bootstrap scripts cannot avoid being confusing, we should merge them so there is only one choice going forward.

One property of `millw` is, it provides support for __all__ Mill versions and all operating systems. It also has some convenience features like finding/using the latest released Mill version automatically. In previous discussions we had about this topic, you were against keeping support for older versions in the in-repo script and wanted a rather lean script. That at least is my personal takeaway. 

So, having `millw` follow its own versioning makes sense. It's definitely not the most user friendly setup.

Ideally, we only have one script, and keep the latest version of it as a copy in root of the Mill repo. That follows the scheme we propose as best practice in our documentation.

PR #3296, opened 2024-07-29
Use latest versions of SBT junit test frameworks

lihaoyi:


lefou (commented):
I guess we should forward them via `BuildInfo`, so we can detect updates with Mill.

lihaoyi:
Possibly, though in this case I think it wouldn't have been enough since the new artifacts are under different organization ids than before

lefou (approved):

PR #3321, opened 2024-07-30
Add example using JNI

lihaoyi:


sequencer:
FWIW here is our panama example:
https://github.com/chipsalliance/chisel/blob/8923c56111b70db7691e374e6f0cb8caab3ed9ca/common.sc#L103

example/javamodule/12-jni/build.sc
```diff
@@ -0,0 +1,59 @@
+import mill._, javalib._
+
+object foo extends RootModule with JavaModule {
+  // Additional source folder to put C sources
+  def nativeSources = T.sources(millSourcePath / "native-src")
+
+  // Auto-generate JNI `.h` files from Java classes using Javac
+  def nativeHeaders = T {
+    os.proc("javac", "-h", T.dest, "-d", T.dest.toString, allSourceFiles().map(_.path)).call()
```
lefou, line 9: We have `Jvm.jdkTool`, which tries to look up the tool in the currently used JRE/JDK.

lihaoyi: Thanks! I'll use that

PR #3323, opened 2024-07-30
Format Scala files in codesig test cases

lefou:


lihaoyi:
The tests are failing because the test framework scrapes the `/* */` comment to find the expected assertion data. Should be easy to tweak the scraping (Should be https://github.com/com-lihaoyi/mill/blob/a16c97dd645f6c152ca8c13153b7b95232c70ca4/main/codesig/test/src/CallGraphTests.scala#L175-L183) to accomadate the adjusted format.

PR #3326, opened 2024-07-31
[WIP] Add a Netty example `build.sc` and write about it

lihaoyi:


docs/modules/ROOT/pages/Java_Case_Study.adoc
```diff
@@ -0,0 +1,430 @@
...
+
+The Mill build for Netty is much more performant than the default Maven build. This applies to
+most workflows:
+
+=== Sequential Clean Compile
+
+```bash
+$ time ./mvnw -DskipTests  -Dcheckstyle.skip -Denforcer.skip=true clean install
+
+$ time ./mill clean __.compile
```
lefou, line 41: I think these are not fair comparisions. We either also run `jar` in Mill or we don't `install` in Maven. 

```suggestion
$ time ./mvnw -DskipTests  -Dcheckstyle.skip -Denforcer.skip=true clean compile

$ time ./mill clean __.compile
```

Running just compile sometimes work in a Maven reactor build, sometimes not. Need to check.

Alternatively, we need to run `jar` in Mill.

```suggestion
$ time ./mvnw -DskipTests  -Dcheckstyle.skip -Denforcer.skip=true clean install

$ time ./mill clean __.jar
```

lefou (commented):
Great case study.

docs/modules/ROOT/pages/Java_Case_Study.adoc
lefou, line 1: Should be named `Java_Case_Study_Netty.adoc`

docs/modules/ROOT/pages/Java_Case_Study.adoc
lihaoyi: The thing is, in Mill the common workflow is running `compile`, while in Maven the equivalent workflow is running `install`. AFAICT, this is because Maven uses `~/.m2/local` for passing dependencies between subprojects, while Mill doesn't.

Part of the reason Mill is faster is because it does less work, but I don't think that makes the comparison unfair. In the end, nobody runs `compile` in Maven or `jar` in Mill. `mvn install` and `mill compile` are equivalent user workflows

docs/modules/ROOT/pages/Java_Case_Study.adoc
```diff
@@ -0,0 +1,430 @@
...
+  def generatedSources0 = T{
+    val shell = new groovy.lang.GroovyShell()
+    val context = new java.util.HashMap[String, Object]
+
+    context.put("collection.template.dir", "common/src/main/templates")
+    context.put("collection.template.test.dir", "common/src/test/templates")
+    context.put("collection.src.dir", (T.dest / "src").toString)
+    context.put("collection.testsrc.dir", (T.dest / "testsrc").toString)
+
+    shell.setProperty("properties", context)
```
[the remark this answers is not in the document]
lihaoyi: I'd like to leave this as is, since we're interoping with Java libraries, and this example is meant to be readable to Java programmers. Scala is more idiomatic Mill, but staying in `java.util.HashMap`-land will be much more familiar and less scary

docs/modules/ROOT/pages/Java_Case_Study_Netty.adoc
```diff
@@ -16,46 +16,334 @@ To do this, we have written a Mill `build.sc` file for the Netty project. This c
 with Mill to build and test the various submodules of the Netty project without needing to
 change any other files in the repository:
 
-- ???[Netty `build.sc` file]
+- https://github.com/com-lihaoyi/mill/blob/main/example/thirdparty/netty/build.sc[Netty `build.sc` file]
```
lefou:
```suggestion
- {mill-github-url}/blob/main/example/thirdparty/netty/build.sc[Netty `build.sc` file]
```

docs/modules/ROOT/pages/Java_Case_Study.adoc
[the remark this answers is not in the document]
lefou: I almost expected this answer as I had the same thought. ;)

PR #3330, opened 2024-08-01
Add overloads to try and simplify resolveDeps workflows

lihaoyi:
Still a bit awkward, but less than before. 

Semantics should be exactly the same, just moved a bunch of boilerplate into helper overloads.

Pull request: https://github.com/com-lihaoyi/mill/pull/3330


lefou (commented):
The issue with these overloads is, that they hide the fact dependencies need to bind to a context. API users are encouraged to collect dependencies from various contexts (e.g. different modules). This led to various issues in the past, where modules with different contexts (Scala version, platform, ...) where resolved with the wrong (local) context.

lihaoyi:
Sure, I think `resolveLocalDep` sounds great.

lihaoyi:
TBH looking at all the callsites, resolving to "this module" by default definitely seems like the right thing to do. I do think that for the bulk of users, resolving "this module's" dependencies is what they want. And it's true that occasionally people will need to aggregate across multiple modules and feed it through the same context, but that seems like a pretty advanced use case that doesn't happen that often

lihaoyi:
I have some ideas of how to make this API even better, will update the PR

scalalib/src/mill/scalalib/CoursierModule.scala
```diff
@@ -108,3 +122,38 @@ trait CoursierModule extends mill.Module {
...
+                                                  sources: Boolean = false) = {
+      Lib.resolveDependencies(
+        repositories = repositories,
+        deps = deps.map(implicitly[CoursierModule.Resolvable[T]].bind(_, bind)),
+        sources = sources,
+        mapDependencies = mapDependencies,
+        customizer = customizer,
+        coursierCacheCustomizer = coursierCacheCustomizer,
+        ctx = ctx
+      ).asSuccess.get.value
```
lihaoyi, line 146: This needs to be adjusted to throw a proper exception

lihaoyi:
@lefou took another pass at it. I think the code looks a lot nicer now; no more confusing nested `T.task` things, and it looks like a normal method call chain. 

One downside is the normal method call now needs to throw an exception if it fails, but that's what normal method calls do, and it shouldn't be too hard to wire up try/throw/catch to report a proper error

scalalib/src/mill/scalalib/CoursierModule.scala
```diff
@@ -108,3 +122,38 @@ trait CoursierModule extends mill.Module {
...
+  class Resolver(repositories: Seq[Repository],
+                 bind: Dep => BoundDep,
+                 mapDependencies: Option[Dependency => Dependency] = None,
+                 customizer: Option[coursier.core.Resolution => coursier.core.Resolution] = None,
+                 ctx: Option[mill.api.Ctx.Log] = None,
+                 coursierCacheCustomizer: Option[
+                   coursier.cache.FileCache[coursier.util.Task] => coursier.cache.FileCache[coursier.util.Task]
+                 ] = None) {
+
+    def resolveDeps[T: CoursierModule.Resolvable](deps: IterableOnce[T],
```
lihaoyi, line 136: Using a typeclass here is a bit annoying, but it's necessary to work around the "multiple overloads cant have default params" problem that would otherwise make the user-facing API a lot worse

lefou:
> One downside is the normal method call now needs to throw an exception if it fails, but that's what normal method calls do, and it shouldn't be too hard to wire up try/throw/catch to report a proper error

Couldn't we provide a version returning a `Task`, so we just add another `()` and get error handling for free.

lefou:
> > One downside is the normal method call now needs to throw an exception if it fails, but that's what normal method calls do, and it shouldn't be too hard to wire up try/throw/catch to report a proper error
> 
> Couldn't we provide a version returning a `Task`, so we just add another `()` and get error handling for free?

Nevermind. The issue is that the input-parameter(s) are dynamic.

lefou:
I had some idea in mind for quite a while, that we split the coursier resolution into two steps, resolution and fetching. Lots of meta-operations like finding updates, inspecting licenses, ... only require the resolution without the fetched artifacts. The resolution could also be locked and persisted for later reuse in a more reproducible setup. Since this affects `resolveDeps` as well, we might think it through as s whole for the next breaking Mill. We lived so long with `resolveDeps`, it's not time critical to change it now.

lihaoyi:
Making `defaultResolver().resolveDeps(...)()` return a task that you call with additional `()` doesn't really work, because then you cannot pass stuff to the `...` that comes from within the enclosing `T{...}` block. To make that work, we'd need to make `resolveDeps` itself take a `T.task{...}`, then we're back to where we started in messiness.

This is pretty fundamental to Mill's multi-stage evaluation model here. `defaultResolver().resolveDeps(...)` is actually less flexible than the old `resolveDeps(T.task{...})()`, because we are moving the decision of what gets passed to `resolveDeps` from resolve-time to evaluation-time. In theory, with the old API you could make decisions affecting the task graph structure based on resolve-time values, while the new API the task graph structure is fixed.

Given all the callsites within the Mill codebase work just fine with the more restricted callsites, I think it's probably the right move. I'd like to get it in sooner rather than waiting for a complete re-think of the API, since it's a backwards compatible addition that fixes a pretty warty part of the Mill experience, and it doesn't block us from making more drastic cleanups later

lihaoyi:
Updated a bunch of the examples to take advantage of this, which makes them a lot simpler

lefou (approved):

PR #3334, opened 2024-08-04
Trim down CI matrix further

lihaoyi:
- Stop running Scala.js, Native, and contrib tests across both JVM versions
- Cut down windows tests to avoid running the same tests against both JVM versions; rather run each set of tests once but pick the version arbitrarily so both JVM versions get some coverage
- We don't need `{__.publishLocal,dev.assembly,__.compile}` on windows because the `integration._.{local,fork,server}` tests should exercise those code paths
- Cut down job timeout from 90min to 60min. Most jobs finish in <25min right now, so 60min should be plenty of time and give faster feedback if jobs get stuck

For the various test suites, every test is still run on both Java8 and Java 17, and on Windows or Linux: e.g. "Java8/Windows + Java17/Linux" or "Java17/Windows + Java8/Linux". Not as good as full coverage of all spots in the matrix, but should hopefully be enough to catch most bugs

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -61,6 +61,9 @@ jobs:
           # Run unit and module tests on both oldest and newest Java versions
           - '"{main,scalalib,testrunner,bsp}.__.test"'
         include:
+          # For most tests, run them arbitrarily on Java 8 or Java 17 on Linux, and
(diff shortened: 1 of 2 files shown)


lefou:
@lihaoyi Can you please also remove the no longer existing CI jobs from the required status checks in our `main` branch protection rule on Github settings.

Issue #3344, opened 2024-08-07
Error in IntelliJ about Multiple 'scala3-library*.jar' files when using scala 3.x and scalajs


malcolmredheron:
When I use GenIdea in a mill/scalajs project and then build in intellij, it works fine if build.sc uses scala 2.13.14 but for every version of scala 3 that I have tried IJ gives me errors of the form `scala: Multiple 'scala3-library*.jar' files (scala3-library_3-3.3.1.jar, scala3-library_sjs1_3-3.3.1.jar) in Scala compiler classpath in Scala SDK scala-SDK-3.3.1`

https://github.com/malcolmredheron/mill-scalajs-experiment contains code and exact steps that demonstrate the problem. This happens for me with mill 0.11.8, 0.11.10, 0.11.11 and 0.11.11-1-066ea7, which is the most recent commit that I could find a build for

lefou:
Thank you for this report. We introduced the "scala-SDK" entry in Mill `0.11.8`. It looks like some handling for ScalaJS is missing.

lefou:
I just checked your example project. 

```
> mill --version
Mill Build Tool version 0.11.11
Java version: 17.0.11, vendor: Eclipse Adoptium, runtime: /opt/openjdk-bin-17.0.11_p9
Default locale: de_DE, platform encoding: UTF-8
OS name: "Linux", version: 6.1.90-gentoo-x86_64, arch: amd64

> mill mill.idea.GenIdea/idea
[1/1] mill.idea.GenIdea.idea 
Analyzing modules ...
[build.sc] [1/1] mill.idea.GenIdea.idea > [7/9] sources.super.mill.scalalib.JavaModule.sources 
Writing 111 IDEA project files to /tmp/mill-scalajs-experiment/.idea ...

> mill show __.scalaCompilerClasspath
[1/1] show 
[
  "qref:v1:878e902e:/home/lefou/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala3-compiler_3/3.3.1/scala3-compiler_3-3.3.1.jar",
  "qref:v1:7d4aa14c:/home/lefou/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala3-library_sjs1_3/3.3.1/scala3-library_sjs1_3-3.3.1.jar",
[...]

PR #3345, opened 2024-08-07
GenIdea: Make Scala SDK specific for JS / Native

lefou:
* Related to https://github.com/com-lihaoyi/mill/issues/3344


lefou:
I'd like to add a test case, but for that, I need a proper Scala JS example config file for IntelliJ.

lefou:
Merging this to disanbiguate `scala-SDK` from `scala-js-SDK`. We can improve IntelliJ JS support further once we have more insights.

PR #3346, opened 2024-08-07
Fix the warning not being printed when -i/--interactive/--no-server/--bsp are not the first argument

### LINKED ISSUE #2869: `-i` and `--no-server` gets ignored with no error when passed as the second option (200USD Bounty)
---

**From the maintainer Li Haoyi**: I'm putting a 200USD bounty on this issue, payable by bank transfer on a merged PR fixing this. `-i` should raise an error when not passed as the first argument, because it can only take effect when it is first, and passing it as a later argument is a easy footgun that we should give a good error message to help people avoid

---

```bash
./mill --help
# (...)
# -i --interactive                  Run Mill in interactive mode, suitable for opening REPLs and
#                                     taking user input. This implies --no-server and no mill server
#                                     will be used. Must be the first argument.
# (...)
# --no-server                       Run Mill in single-process mode. In this mode, no Mill server
#                                     will be started or used. Must be the first argument.
```

Both `-i` and `--no-server` options require to be the first argument, but no error seems to be produced if they're passed elsewhere.

```bash
▶ ./mill --disable-callgraph-invalidation --no-server __.fix
# (standard build logs, no errors or warnings, --no-server gets ignored)
▶ ./mill --disable-callgraph-invalidation -i __.fix
# (standard build logs, no errors or warnings, -i gets ignored)
```

This caused my team a major headache when `--disable-callgraph-invalidation` has been added to the Scala CLI mill script as the first argument (attempted workaround for #2844), with `-i` sneakily being ignored everywhere as a result, with no failures.

I'd at the very least expect a warning, and preferably just an error.

mkwpz:
The bug stems from the fact that a `PipedInputStream` is passed to `MillMain.main0` where `DummyInputStream` used to be. This PR fixes the bug by using a `PipedInputStream` instead of a `DummyInputStream` on the check that leads to the warning being printed.
Fixes #2869.

### DIFF
--- a/runner/src/mill/runner/MillMain.scala
+++ b/runner/src/mill/runner/MillMain.scala
@@ -1,11 +1,11 @@
 package mill.runner
 
-import java.io.{FileOutputStream, PrintStream}
+import java.io.{FileOutputStream, PrintStream, PipedInputStream}
 import java.util.Locale
 import scala.jdk.CollectionConverters._
 import scala.util.Properties
 import mill.java9rtexport.Export
-import mill.api.{DummyInputStream, MillException, internal, SystemStreams}
+import mill.api.{MillException, internal, SystemStreams}
 import mill.bsp.{BspContext, BspServerResult}
 import mill.main.BuildInfo
 import mill.util.PrintLogger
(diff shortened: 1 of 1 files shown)


lihaoyi:
Thanks @mkwpz! Email me your bank transfer details to haoyi.sg@gmail.com and I'll close out the bounty

mkwpz:
@lihaoyi I've sent you an email.

PR #3352, opened 2024-08-08
CI: Use explicit binary platform for external plugins when bootstrapping

lefou:


lihaoyi:
Rather than changing it at the import-site, should we hardcode it inside Mill such that Mill 0.12.x both resolves and publishes plugins with a `_mill0.11` suffix? That would make the upgrade mostly transparent to folks trying to use 0.12, which would be very nice

lefou:
I had the same though, but dropped it for now, as it might come as a surprise when the version diverges from the binary platform. We are very early in development, maybe a bit too early to decide, but I'd like to have a functional CI.

lihaoyi (approved):
sounds good to me

lihaoyi:
I opened https://github.com/com-lihaoyi/mill/pull/3353 if we want to go that route

lefou:
@lihaoyi Turns out, this is not the only issue that makes CI red. Do you have an idea about the other failures?

lihaoyi:
no, but I can investigate

lihaoyi:
https://github.com/com-lihaoyi/mill/pull/3353 seems to fixes more tests than this PR, but still one job is red

lihaoyi:
I suspect the failures on the other PR are flakes due to `java.nio.file.AccessDeniedException`, going to retry it

PR #3353, opened 2024-08-08
Hardcode millBinPlatform to 0.11

lihaoyi:
Since 0.12.x is meant to be bincompatible with 0.11.x


lihaoyi:
@lefou gonna merge this first since its green.

Issue #3354, opened 2024-08-08
IDEA 2024.2 changed to non-transitive module dependencies


lefou:
Instead of just generating the direct module dependencies, we need to unroll the transitive module dependencies.

PR #3355, opened 2024-08-08
GenIdea: unroll module dependencies

lefou:
Since IDEA 2024.2, module dependencies in the `.iml` files are no longer interpreted transitively.
Therefore, we need to unroll all transitive module dependencies when generating the project files.

Fix https://github.com/com-lihaoyi/mill/issues/3354


lefou:
Although we fixed this issue as it was a confirmed change in IDEA, it seems it was an unplanned regression.

See issue https://youtrack.jetbrains.com/issue/SCL-22935

It should be fixed in IDEA 2024.2.567 (Nightly), so we need to check later, whether we want/need to revert this change.

Issue #3338, opened 2024-08-05
[RFC] Rename `T{...}` and `T[_]` to `task{...}` and `Task[_]`, standardize "Task"/"Target" terminology around "Task"


lihaoyi:
`T{...}` and `T[_]` syntaxes are kind of weird. They are concise, but I don't think the conciseness they bring pays for the weirdness. And the weirdness does get noticed, e.g. [recently on Reddit](https://www.reddit.com/r/programming/comments/1ea400h/comment/lekoohw/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button). I notice it too. Part of the reason it's there is that `Target[_]` and `Target{...}` is too verbose, but maybe that's fixable

This proposal would be to leave the `T` aliases in place but deprecated, add a new `task` alias for the expressions:

* `T{...}` -> `task{...}`
* `T.command{...}` -> `task.command{...}`
* `T.task{...}` -> `task.anon{...}`
* `T.dest` -> `task.dest`, ...)
* `T[_]` -> `Task[_]` 

`Task[_]` and `task{...}` are maybe short enough (compared to `Target`) that we can do without the `T` alias

This would make the codebase a lot more approachable for folks not used to weird Scala DSLs

There's some messiness around how the `Task`/`Target` hierarchy is arranged right now that may require a breaking change to make this work well.e.g. `Target` may need to become `Task`, if we want to use the `Task` terminology everywhere instead of `T`. But then `Task` would need to be renamed to avoid confusion (`BaseTask`?). This would basically be cleaning up some of the messiness that we failed to totally resolve earlier in https://github.com/com-lihaoyi/mill/pull/2402, as the status quo today the terminology of whether we should call things *tasks* or *targets* is still a bit muddled.

e.g. we may need to go from today's:

- `Task`
    - `Task.Sequence`, `Task.TraverseCtx`, `Task.Mapped`, `Task.Zipped`, `T.task`, etc.
    - `NamedTask`
        - `Command`
        - `Worker`
        - `Target`
            -  `TargetImpl`
                - `PersistentImpl`
            - `InputImpl`
                - `SourceImpl`
                - `SourcesImpl`

To something like

- `BaseTask`
    - `Task.Sequence`, `Task.TraverseCtx`, `Task.Mapped`, `Task.Zipped`, `T.task`, etc.
    - `NamedBaseTask`
        - `Command`
        - `Worker`
        - `Task`
            -  `TargetImpl`
[...]

PR #3356, opened 2024-08-08
Make `T` an alias for `Task` instead of `Target`, move `T.*` operations to `Task.*`

lihaoyi:
This PR attempts to fix https://github.com/com-lihaoyi/mill/issues/3338:

1. `T` is now an alias for `Task` instead of `Target`. We expect to use `foo: Task[V]`, `Task { ... }`, `Task.dest`, etc. APIs going forward, but all existing APIs remain on the `T` aliases for backwards source compatibility

2. `Task.apply` or `Task { ... }` continues to return a `Target` instance, but now is typed as returning the supertype `Task`. This allows people to write `def foo: Task[T] = Task { ... }` and have it do the right thing, rather than the otherwise necessary `def foo: Target[T] = Task { ... }` which is pretty awkward

3. In order to support using `Task[T]` everywhere as types,  I flattened out the type hierarchy further: now, most APIs work directly with `Task` instead of `Target` or `NamedTask`
    1. `Task` now supports `.ctx`, which throws an `UnsupportedOperationException` by default. `NamedTask` overrides it with an implementation.
    2. Rather than relying on methods returning `Target[_]` for resolution, we now rely on methods returning `Task[_]` with zero argument lists. As the number of argument lists is not available through Java reflect (we can only see the total number of arguments), I rely on information collected by the `Discover[T]` macro.

4. I made `Input[_]`, `Source[_]`, `Target[_]`, etc. type aliases of `Task[_]`. This avoids issues where people might accidentally declare `def foo: Target[V]` and find themselves unable to override it later as type `def foo: Task[V] = Task { ... }`. This is similar to how https://github.com/com-lihaoyi/mill/pull/2402 earlier made `Input` `Source` etc. type aliases of `Target[_]`, and serves a similar purpose.

5. Distinguishing of cached targets (zero argument lists) with uncached other things (1 or more argument lists) is done via the information we already collect in the `Discover` macro. For safety, I made the `def millDiscover` abstract, because it cannot really be implemented in abstract superclasses or traits (due to is macro implementation) so it really needs to be implemented in every concrete `object`. This should avoid a common issue where we forget to override it in the test suite and command resolution wouldn't work.

The flattening out of the `Task[_]` type hierarchy is a compromise. `def ctx: Ctx = throws` on the `Task` trait is gross. We lost the ability to distinguish between various subtypes of `Task` based on method return types, and instead have to go through other ad-hoc mechanisms like number-of-argument-lists. But in exchange for a bit of sloppiness in the type signatures, users can now use `Task{ ... }` to construct their most common tasks (targets), `: Task[V]` to annotate their types, and `Task.sources` and `Task.dest` for most common task APIs

This is a binary incompatible change and will have to go into 0.13.0. I tried to leave the existing `T`/`Target`/`Input`/etc. aliases in place, so hopefully it's mostly source compatible (though probably not 100%)

This PR is a bit messy, but important files to review include:

- `main/src/mill/package.scala`, which re-writes the various `type` and `val` aliases
- `main/define/src/mill/define/Task.scala`, which moves the various `T.*` operations from `object Target` onto a shared `trait TaskCompanion` that can be shared with `object Task`, and makes them all return `Task[_]` instead of a more specific subtype
- `main/define/src/mill/define/Discover.scala`, which tweaks the class-task-name-discovery logic to make it look for `Task` instead of `Target`, and make it precise enough to work as the primary resolution mechanism (previously there were some bugs, that didn't matter when it was just used for error message reporting, but weren't acceptable now that it is becoming the primary way we find tasks in a module)
- `main/resolve/src/mill/resolve/ResolveCore.scala`, which converts the task resolution logic to use the information already collected by the updated `Discover` macro
- `main/define/src/mill/define/BaseModuleTree.scala`, which pre-computes a bunch of values used throughout the task resolution logic
- `main/resolve/src/mill/resolve/Resolve.scala` has more changes to use `BaseModuleTree` pre-computed values


main/define/test/src/mill/define/ApplicativeTests.scala
```diff
@@ -28,7 +28,7 @@ object ApplicativeTests extends TestSuite {
       value
     }
   }
-  @compileTimeOnly("Target.ctx() can only be used with a T{...} block")
+  @compileTimeOnly("Target.ctx() can only be used with a Task {...} block")
```
lefou, line 31:
```suggestion
  @compileTimeOnly("Target.ctx() can only be used within a Task {...} block")
```

lihaoyi: This doesn't work yet; need to make changes in the mill-moduledefs repo. TBH I feel like just in-sourcing it again, since even when in-sourced people can depend on the artifact separately if necessary, and it makes doing these changes in one PR a lot more convenient

lefou: Since it's a compiler module, it needs releases for every Scala version we want to use. That's the main reason we out-sourced it. We post-release for every new Scala 3.13 version.

It might be ok for a development version to use a in-sourced patched version, but I think the goal should be to push any changes back to the external project before we reach a stable version.

lihaoyi: How about we in-source it to the Mill repo but still keep the publishing separate? That would mean it's treated like the `bridge` modules, and would keep everything in one place while still preserving the per-scala-version releases and separate publishing

lefou: That will work too, _if_ it's well documented. We do these things only occasionally, there is no muscle memory for the process, so we need to good step-by-step documentation targeting the developer/publisher.

lefou (commented):
Why is all the mill-moduledefs meta info no longer needed.It does still have a dedicated life cycle, so we should manage it via the `build.sc` and generated `BuildInfo`.

lihaoyi:
I think I've come up with a way go avoid modifying the compiler plugims by moving the info tonthe discover macro. That will let us avoid having to republish them, and also avoid failure modes of people forgetting to include the plugins in their published mill plugins. Will see if i can make it work

lihaoyi:
Mostly superseded by other PRs

Issue #3152, opened 2024-05-14
Migrate `build.sc` to Scala 3


lihaoyi:
The goal of this ticket is to move Mill from Scala 2.13 to Scala 3.

Most of Mill's code should be easily portable with minimal changes (e.g. syntax tweaks), but there are three major areas which may require more work:

1. The `import $ivy` parser is currently using FastParse's ScalaParse, and would need to be moved over to use a Scala 3 compatible parser similar to what [Ammonite](https://github.com/com-lihaoyi/Ammonite) does

2. The [compiler plugin for auto-override](https://github.com/com-lihaoyi/mill-moduledefs/blob/236b00a205ef841bfc4bf64b66c9648b71d60870/moduledefs/plugin/src/AutoOverridePlugin.scala) would need to be migrated over 

3. The various macros like [Caller](https://github.com/com-lihaoyi/mill/blob/25d975dda4f68c27835f4ee4ec491199a06b3a25/main/define/src/mill/define/Caller.scala#L9), [Task](https://github.com/com-lihaoyi/mill/blob/25d975dda4f68c27835f4ee4ec491199a06b3a25/main/define/src/mill/define/Task.scala#L237), and [Applicative](https://github.com/com-lihaoyi/mill/blob/25d975dda4f68c27835f4ee4ec491199a06b3a25/main/define/src/mill/define/Applicative.scala#L9) would need to be ported over

4. The [CodeSig](https://github.com/com-lihaoyi/mill/tree/25d975dda4f68c27835f4ee4ec491199a06b3a25/main/codesig) analyzer used for fine grained invalidation on `build.sc` code changes may need to be ported over if Scala 3 generates very different bytecode (not sure, seems unlikely?)

5. Any older dependencies that only support Scala 2 may need to be updated or replaced

The acceptance criteria is A PR that passes all existing tests on Scala 3, and either modify some existing tests to exercise Scala 3 language features, or add new tests to do so. We may not merge the PR immediately, but having the PR open with all tests passing should be enough for us to integrate the Scala 3 support and at our leisure. This would be a breaking change and would need a version bump, with associated downstream inconvenience

lihaoyi:
Some prior art, an attempt to move the `Applicative` macro to Scala 3: https://github.com/Iltotore/mill-scala-3

Quafadas:
I have a personal interest / curiosity in learning about mills internals, and support here could be a good way to do it. I'm realistic that I would fail alone through constraints on both time and skill :-).

[...]

Issue #3062, opened 2024-03-03
Support an alternative to `$`-imports


lefou:
We inherited the various `$`-imports like `import $ivy` or `import $file` from Ammonite. It turns out, the Scala compiler isn't very fond of us using the `$` symbol in object and class names and newer versions starting from `3.4.0` enforce this more strictly. 

See also:
* https://github.com/lampepfl/dotty/issues/19702
* https://github.com/com-lihaoyi/Ammonite/pull/1394

I think a nice alternative to these magic imports could be something similar as the _using directives_  as known from Scala CLI.

From a compiler perspective, they are just comments, so we don't need to mangle scripts just for the sake of replacing directives with something regular. Something we needed to do before with the `$`-imports.

My preferred usage would look like this:

```scala
//> mill version 0.11.7
//> mill dep g:a:v
```

But since my proposal to make them tool/context specific wasn't accepted, the "standard" way to support tool specific key-value pair would be to encode the tool/context in the key. It would look like this:

```scala
//> using milll.version 0.11.7
//> using mill.dep g:a:v
```

The pros for following this standard would be, that IDEs, formatter and syntax highlighters might handle them better. (Which is already the case at the time of writing this. GitHub renders my first example just grey, but the second nicely colored.)

lefou:
FTR, here is my mentioned proposal: https://github.com/scalacenter/tooling-working-group/discussions/68#discussioncomment-6945562

@lefou wrote on Sept 8, 2003:
> I've been playing around with using directive for a while, mostly in the context of Mill build scripts, and also had many discussions including those on the first Tooling Summit this year.
> 
> I suggest, to split the "using directives" (as known in it's current form) into **three** parts: the comment marker, the context marker and the directive, and discus each part, it's purpose and it's potential for standardization separately.
> 
>     1. **The comment marker `//>`**. This is undoubtedly agreed, that is nice. It's a comment, so it's transparent to the compiler. It's easy to recognize and survives most code formatting attempts. This is a nice candidate for a "standard".
> 
>     2. **The context marker**, currently the "using" keyword. There was lots of discussion about the "using" keyword, it's name and also the fact whether it's needed or not. I'd like to associate the context or the intepreting tools with it. Although a more generice "scala-cli" whould be more clear, "using" is now associated with Scala CLI and will do it. For other tools, there should be a different marker like "mill" or "ammonite"/"amm". That way, tool-specific configuration can co-exist and a script can be compatible to multiple tools/runners.
> 
[...]

PR #3369, opened 2024-08-13
build.mill files compiled by Scala 3

bishabosha:
This is work in progress to fix https://github.com/com-lihaoyi/mill/issues/3152

Numerous fixes were also needed to com-lihaoyi/mainargs, com-lihaoyi/sourcecode, and com-lihaoyi/mill-moduledefs

With the current state, only 1 example/integration tests is still failing:
- `integration.feature[plugin-classpath].local`

known TODOs:
- [x] Discover macro
- [x] Applicative macro
- [x] Caller macro
- [x] Cross.Factory macro
- [x] EnclosingClass macro
- [x] Task macros
- [x] Cacher macro
- [x] Moduledefs compiler plugin (override inferrence)
- [x] All core Mill modules compile with Scala 3.5.0
- [x] Fix Zinc reporter patch linenumbers of build scripts
- [x] Check that bytecode analyzers work with Scala 3
- [x] cleanup library dependency conflicts
- [x] Support new Scala 3 syntax in build.sc files
- [x] ~Fix BSP reporter linenumbers for build scripts~ (Zinc reporter forwards to bsp)
- [ ] Cleanup compiler warnings for outdated syntax

known incompatibilities:
- [ ] can't use `ExplicitResultTypes` scalafix rule - need to upgrade mill-scalafix
- [x] ~`Cross.scala` uses the new quoted type syntax which scalafmt crashes on, (and version is frozen) so skip the file~ upgraded Scalafmt so not skipped anymore
- [ ] skipping Mima currently due to 1000s of errors (perhaps we should generate filters?)
- [ ] filtered one flaky test from `example.thirdparty[mockito]`
- [ ] filtered out `integration.feature[plugin-classpath]` due to third party plugin dep


lihaoyi:
Thanks @bishabosha ! Lets get your mainargs changes landed, that way we can cut a release and you can test your WIP in CI

lefou:
+1 for switching over to _using directives_!

* https://github.com/com-lihaoyi/mill/issues/3062

bishabosha:
@lihaoyi see https://github.com/com-lihaoyi/mainargs/pull/148

bishabosha:
in https://github.com/com-lihaoyi/mill/actions/runs/10418081276/job/28853542122?pr=3369 you can see integration tests like `example.tasks[6-workers].local.test` passing

scalalib/src/mill/scalalib/CoursierModule.scala
```diff
@@ -137,6 +138,15 @@ object CoursierModule {
...
     ): Agg[PathRef] = {
+      // TODO: `getOrThrow` swallows error messages when Failure is returned.
+      // so I introduced resolveDeps0 where it makes sense to propagate the error message
+      resolveDeps0(deps, sources).getOrThrow
+    }
+
+    def resolveDeps0[T: CoursierModule.Resolvable](
+        deps: IterableOnce[T],
+        sources: Boolean = false
+    ): Result[Agg[PathRef]] = {
```
lolgab, line 149: This was fixed by adding special handling for `Result.Exception(Result.Failure())` when we show the error messages here: https://github.com/com-lihaoyi/mill/pull/3391
If you rebase on top of the current `main` branch you should not need to add `resolveDeps0`

bishabosha: oh cool!

bishabosha:
Rebased and included the new mill-moduledefs 0.11.0-M1

bishabosha:
just recording here that the mockito third party test can be flaky (specifically `StubbingLookupListenerCallbackTest .add_listeners_concurrently_sanity_check`):

https://github.com/mockito/mockito/blob/1e337ed703ce2b78ce9ea40ff8b8c645f8219cf2/src/test/java/org/mockitousage/debugging/StubbingLookupListenerCallbackTest.java#L177C17-L177C56

bishabosha:
Only scala 3 syntax to go

lihaoyi:
@bishabosha all tests are green???

bishabosha:
@lihaoyi I still haven't introduced the scala 3 syntax support, but everything previously there passes. I have explicitly skipped certain tests in `TEMP:` commit messages - i.e. scalafix, mima, the one flaky mockito test, and the smithy4s plugin

bishabosha:
@lihaoyi green CI with scala 3 syntax 

Perhaps I will do a round of cleanup for syntax warnings ?

main/define/test/src/mill/define/BasePathTests.scala
```diff
@@ -54,15 +65,6 @@ object BasePathTests extends TestSuite {
...
     test("overridden") {
-      object overriddenBasePath extends TestBaseModule {
-        override def millSourcePath = os.pwd / "overriddenBasePathRootValue"
-        object nested extends Module {
-          override def millSourcePath = super.millSourcePath / "overriddenBasePathNested"
-          object nested extends Module {
-            override def millSourcePath = super.millSourcePath / "overriddenBasePathDoubleNested"
-          }
-        }
-      }
```
lihaoyi, line 65: Was this move necessary? I wonder if it's incidental, or if it indicates a bug somewhere in utest/mill/scala3 that we should log somewhere

scalajslib/test/src/mill/scalajslib/EsModuleRemapTests.scala
```diff
@@ -25,7 +25,10 @@ object EsModuleRemapTests extends TestSuite {
       ESModuleImportMapping.Prefix("@stdlib/linspace", remapTo)
     )
 
-    override lazy val millDiscover = Discover[this.type]
+    override lazy val millDiscover = {
+      import mill.main.TokenReaders.given
```
lihaoyi, line 29: Can this be moved into the `Discover` macro body? The Scala 2 version does `import mill.api.JsonFormatters._;` as part of the macro quasiquote, I assume something similar should be possible in Scala3?

scalajslib/test/src/mill/scalajslib/HelloJSWorldTests.scala
```diff
@@ -50,15 +50,15 @@ object HelloJSWorldTests extends TestSuite {
       )
 
       object `test-utest` extends ScalaJSTests with TestModule.Utest {
-        override def sources = Task.Sources { millSourcePath / "src/utest" }
+        override def sources = Task.Sources { this.millSourcePath / "src/utest" }
```
lihaoyi, line 53: Is the `this.` here necessary?

scalalib/src/mill/scalalib/Assembly.scala
```diff
@@ -13,13 +13,21 @@ import java.util.Collections
...
 
+  private object Streamable {
+    def bytes(is: InputStream): Array[Byte] = {
+      val out = new java.io.ByteArrayOutputStream
+      IO.stream(is, out)
+      out.close()
+      out.toByteArray
+    }
+  }
+
```
lihaoyi, line 30: We're dropping Java 8 in Mill 0.12.0 and above, so I think this can be replaced with `is.readAllBytes()`

lihaoyi:
Just looked through the whole PR. Looks great overall. Some thoughts on next steps:

**Timing**

* We want to land this PR as late as possible, as it breaks binary compatibility, and will require us forking off a 0.13.0 branch from 0 .12.0. 
* @bishabosha is tentatively with us until end oct latest, and we're planning on releasing 0.12.0 in early/mid oct
* Thus we should aim to land this PR in late oct, after the Mill 0.12.0 release and before @bishabosha's time with us comes to and end

**Rebase/Merging**

* There will likely be some more merge/rebase conflicts landing in the leadup to 0.12.0: https://github.com/com-lihaoyi/mill/pull/3617, https://github.com/com-lihaoyi/mill/pull/3478, possibly others. 
* I suspect would make most sense to do one big merge/rebase at the end rather than trying to keep up individually

**Remaining Major Issues**

* The source-code-generation of mirrors for uPickle seems like the biggest hack that remains in this PR. My understanding is that it is because uPickle in Scala 3 does not work well for non-case classes. I see two paths forward here:
    * Automate this mirror generation as part of the uPickle macro. It seems pretty mechanical code. Even if we can't get it down to zero lines, if we could get it down to 1-2 lines, that would already be a big win
    * Alternatively, for the types we own (and even for some types we don't) we could convert them all to `case class`es and bank on https://github.com/scala/improvement-proposals/pull/78 landing soon. In the interim, we could use the existing compiler plugin to bridge the gap until it lands in a Scala 3 release.

scalajslib/test/src/mill/scalajslib/HelloJSWorldTests.scala
bishabosha: this causes an ambiguity error if it isn't there

scalajslib/test/src/mill/scalajslib/EsModuleRemapTests.scala
bishabosha: no, imports aren't visible to implicit search inside a macro, this is the same problem that was blocking the design of scalasql

bishabosha: I will try again maybe if if move where the import is?

lihaoyi: I think a `import mill.given` at the top of every file would look better, even if it's the same number of lines. For the version in `CodeGen.scala` nobody sees that code so the ugly version is fine.

Is there an open issue for the `imports aren't visible to implicit search inside a macro` thing? If so we should link it somewhere in a comment

bishabosha: I did that in the later commits, ill be happy to make that change if the macro itself can't be fixed

bishabosha:
> We should run through this PR, identify all such 2.13-compatible changes, and land them separately before the breaking Scala 3 changes

you can include even more with `-Xsource:3`

lefou:
> > We should run through this PR, identify all such 2.13-compatible changes, and land them separately before the breaking Scala 3 changes
> 
> you can include even more with `-Xsource:3`

I think `-Xsource:3` breaks bin-compat of the produced jars. I tried this before (https://github.com/com-lihaoyi/mill/pull/2557).

bishabosha:
@lihaoyi  fix in https://github.com/com-lihaoyi/mill/pull/3369/commits/d7ae5aeddfd9b2bef906496c557a5eda11610b13 depends on releasing a new `mill-moduledefs` - see https://github.com/com-lihaoyi/mill-moduledefs/pull/18

lihaoyi:
@bishabosha I noticed that without the line number plugin, runtime line number reporting no longer properly subtracts the lines of from the generated code:

```
lihaoyi mill$ git rev-parse head
d7ae5aeddfd9b2bef906496c557a5eda11610b13
lihaoyi mill$ git diff
diff --git a/example/javalib/basic/1-simple/build.mill b/example/javalib/basic/1-simple/build.mill
index 3f4e27a8b3..8d870ab632 100644
--- a/example/javalib/basic/1-simple/build.mill
+++ b/example/javalib/basic/1-simple/build.mill
@@ -3,6 +3,7 @@ package build
 import mill._, javalib._

 object `package` extends RootModule with JavaModule {
+  ???
   def ivyDeps = Agg(
     ivy"net.sourceforge.argparse4j:argparse4j:0.9.0",
     ivy"org.apache.commons:commons-text:1.12.0"
lihaoyi mill$ ./mill -i dist.run example/javalib/basic/1-simple -i compile
[2517/2517] dist.run
[61/61] ================================================== compile ======================================================
An unexpected error occurred
Exception in thread "main" scala.NotImplementedError: an implementation is missing
	at scala.Predef$.$qmark$qmark$qmark(Predef.scala:344)
	at build_.package_.<init>(build.mill:23)
	at build_.package_$.<init>(build.mill:40)
	at build_.package_$.<clinit>(build.mill:40)
```

What do you think we should do here? It seems we have two options: either we fix it in a Scala 3 compiler plugin, or we fix it by mangling the bytecode with ASM. Either seems like it should work

bishabosha:
> runtime line number reporting no longer properly subtracts the lines of from the generated code

as in the stack trace :/ I hadn't considered that.

When I first tried I did with a plugin and it seemed like positions are less easily mutable than scala 2, but someone taking a fresh look might find its not so complex

bishabosha:
@lihaoyi I remembered the original reason I did the zinc reporter changes was due to the fact that Plugins must run after typer, so any type/parse errors would have incorrect numbers.

If its simple enough to update positions via plugin then still we can do it for the purpose of stack traces at runtime.

bishabosha:
doing the new linenumbers now

bishabosha:
@lihaoyi in commit https://github.com/bishabosha/mill/commit/84eb0891ed4a285f41d8431aae5a5da6d8afb2bf in another branch, I successfully remap the positions of trees - but it seems the stack trace offset is still incorrect?

bishabosha:
@lihaoyi seemed to have got the trick with a two-phase encoding scheme in https://github.com/bishabosha/mill/commit/64446481156710b7861fff5283c4fbcc06f1cd5e

```scala
package build
import $meta._
import mill.{Task, Command, RootModule, Cross}, Task.Anon
import $packages._
import $file.foo.Box
import $file.foo.{given Box[Int]}

def inner: Int = ??? // not implemented

object `package` extends RootModule:

  given Cross.ToSegments[DayValue](d => List(d.toString))

  given mainargs.TokensReader.Simple[DayValue] with
    def shortName = "day"
    def read(strs: Seq[String]) =
      try
        Right(DayValue.withName(strs.head))
      catch
        case _: Exception => Left("not a day")

  type DayValue = DayValue.Value
  object DayValue extends Enumeration:
    val Monday, Tuesday, Wednesday, Thursday, Friday, Saturday, Sunday = Value

  object day extends Cross[DayModule](DayValue.values.toSeq)

  def anyDay(myDay: DayValue): Command[Unit] = Task.Command:
    println(s"Today is $myDay")
  end anyDay

  trait DayModule extends Cross.Module[DayValue]:
    def myDay: DayValue = crossValue

    def day(): Command[Unit] = Task.Command:
      println(s"Today is $myDay")
    end day
  end DayModule

  def someTopLevelCommand(): Command[Unit] = Task.Command:
    println(s"Hello, world! ${summon[Box[Int]]} $inner")
  end someTopLevelCommand

end `package`
```

```text
millb -i someTopLevelCommand
[mill-build/build.mill-57/61] compile
[mill-build/build.mill-57] [info] compiling 1 Scala source to /Users/jamie/Workspace/mill-repo/integration/feature/scala-3-syntax/resources/out/mill-build/mill-build/compile.dest/classes ...
[mill-build/build.mill-57] [info] done compiling
[build.mill-57/61] compile
[build.mill-57] [info] compiling 3 Scala sources to /Users/jamie/Workspace/mill-repo/integration/feature/scala-3-syntax/resources/out/mill-build/compile.dest/classes ...
[build.mill-57] [info] -- /Users/jamie/Workspace/mill-repo/integration/feature/scala-3-syntax/resources/build.mill:8:18
[build.mill-57] [info] 8 │def inner: Int = ??? // not implemented
[build.mill-57] [info]   │                 ^^^
[build.mill-57] [info]   │moved position to [/Users/jamie/Workspace/mill-repo/integration/feature/scala-3-syntax/resources/build.mill,[319..322]](/Users/jamie/Workspace/mill-repo/integration/feature/scala-3-syntax/resources/out/mill-build/generateScriptSources.dest/build_/build.mill%319L8)
[build.mill-57] [info] -- /Users/jamie/Workspace/mill-repo/integration/feature/scala-3-syntax/resources/build.mill:8:18
[build.mill-57] [info] 8 │def inner: Int = ??? // not implemented
[build.mill-57] [info]   │                 ^^^
[build.mill-57] [info]   │Adjusted tree is now [build.mill:14,[319..322]]
[build.mill-57] [info] done compiling
[1/1] =========================================== someTopLevelCommand ========================================= 8s
1 tasks failed
someTopLevelCommand scala.NotImplementedError: an implementation is missing
    scala.Predef$.$qmark$qmark$qmark(Predef.scala:344)
    build_.build$package$.inner(build.mill:8)
    build_.package_.someTopLevelCommand$$anonfun$1$$anonfun$1(build.mill:42)
    build_.package_.someTopLevelCommand$$anonfun$1$$anonfun$adapted$1(build.mill:42)
```

lihaoyi:
> Edit: on further inspection, the dotty backend phase is hardcoded to compute line numbers by looking at the offset of the position in the source file of the current compilation unit, not the source file of the position - so updating positions will not fix this
>
> one very big hack could be to adjust the position so it is whatever the position of the start line of the current file is, but then reporter would be broken :)

@bishabosha I think that sounds reasonable, backend-specific limitations dealt with via backend-specific hacks. 

One concern would be the source positions in non-backend outputs: things like `sourcecode.Line` calls, TASTY or SemanticDB. Could you try doing a `println(sourcecode.Line())` and see what that does with your hack?

bishabosha:
> One concern would be the source positions in non-backend outputs: things like sourcecode.Line calls, TASTY or SemanticDB. Could you try doing a println(sourcecode.Line()) and see what that does with your hack?

All of these things are generated at type checking, so they will be the incorrect version

bishabosha:
@lihaoyi `sourcecode.Line()` macro is implemented as 
```scala
def lineImpl(using Quotes): Expr[sourcecode.Line] = {
    val line = quotes.reflect.Position.ofMacroExpansion.startLine + 1
    '{new sourcecode.Line(${Expr(line)})}
  }
```
there's no way to intercept that

lihaoyi:
@bishabosha got it. Could you open a discussion thread on https://github.com/scala/scala3 with your findings so far? If it turns out to be impossible to implement line-number/position fixing 100% correctly with Scala 3, then we don't have much of a choice, but we should make sure we have a discussion about it so that the upstream folks have a chance to add support for this use case in future

bishabosha:
So an ASM rewrite could be done the same-time/after `methodCodeHashSignatures`?

this would still mean semanticdb, tasty will be wrong

bishabosha:
My other solution: patch ASAP in compiler the ability to trust positions in the code

bishabosha:
so the compiler backend fix was merged

bishabosha:
@lihaoyi it seems like Scala 3 really doesn't like the task invocation syntax `()` in a lot of cases, perhaps it would be worth an  alternative like `.eval` ?

lihaoyi:
@bishabosha am I correct to say that single `()` style syntax is fine, and the problem is mostly with double `()()` style syntax? I think that's something I want to find a way around anyway, so even if this turns out to be a wontfix upstream we probably want to find another syntax locally, but the single `()` for the common cases should probably remain

bishabosha:
Yeah I don't think there has been a problem with a single apply

bishabosha:
@lihaoyi I'm also not sure why there is a difference in the `integration.invalidation[codesig-subfolder]` test - (any change to a single build file compiles all build files) - where is the logic that decides this based on codesig? or perhaps its a difference in how Scala 3 emits Zinc API

lihaoyi:
@bishabosha for `integration.invalidation[codesig-subfolder]`, let's just skip that for the moment. I expect that to change when https://github.com/sbt/zinc/pull/1462 lands, which is likely related to the differences in compilation issues. The "who is recompiled" is 100% on zinc, and will change when that linked PR lands

Not sure about the `test-docs` stuff. Guessing it's some trivial bug in the doc-generation pipeline as part of the scala3 changes. I'd say it's not super critical for you to look into it, since it's something anyone should be able to debug later if necessary

lihaoyi:
>  let's just skip that for the moment. 

Or just comment out the "number of files being recompiled" asserts, the "task hashes are properly invalidated" asserts are still useful (you didn't mention them failing so I assume they still work?)

bishabosha:
@lihaoyi is there any updates to the plan for merging this post 0.12.0?

lihaoyi:
@bishabosha let's hold the PR open for now, going forward I can take over responsibilities for keeping the PR up to date so you don't need to maintain it forever. 

I don't know exactly how much development we'll see in 0.12.x yet, so not sure when the best time to break bincompat in `main` will be, but I expect there'll be some short-term follow up from the 0.12.0 release.

1. Keeping the PR open and occasionally merging `main` into it would be isomorphic to merging it, cutting a 0.12.x release branch, sending PRs to the release branch and then doing batch-merges into `main`. 

2. Either of those options would probably be less tedious than having to backport every PR if 0.12.x is still under active development

3. But keeping the PR open and updated is *probably* going to be less confusing than cutting/targeting 0.12.x and batch-merging into `main`

So for now I think it makes the most sense to keep the PR open and periodically merge `main` into it until we're ready to break bincompat. I can do the work of keeping it up to date with `main` going forward

lihaoyi:
Seems that OS-Lib 0.11.4-M2 is causing an unreasonable compile time slowdown in Scala 3 so sticking with 0.11.3 for now

lihaoyi:
merged in latest main and got most tests passing; there are still some stragglers but nothing blocking.

The main thing still missing is proper support for corrected line-numbers and file-names. This kinda-sorta works in certain cases and not others, e.g. `sourcecode.Line` and `sourcecode.File` gives the wrong values still

bishabosha:
@lihaoyi maybe you can mine some stuff from https://github.com/bishabosha/mill/tree/scala3-build-sc-reintroduce-linenumbers and https://github.com/bishabosha/mill/tree/scala3-build-sc-reintroduce-linenumbers-with-dotc-patch

lihaoyi:
Seems like some of the new `./mill -w 'integration.invalidation[codesig-subfolder].fork.test' mill.integration.CodeSigSubfolderTests.simple` tests are failing on Scala 3. Probably just need to adjust some of the codesig heuristics we added in https://github.com/com-lihaoyi/mill/pull/4116 to be compatible with Scala 3

lihaoyi:
Seems like some lazy val initialization code in the `<clinit>` static initializer changes when you rename lazy vals in Scala 3

build.mill
```diff
@@ -50,7 +50,7 @@ object Deps {
   // The Scala version to use
   // When updating, run "Publish Bridges" Github Actions for the new version
   // and then add to it `bridgeScalaVersions`
-  val scalaVersion = "3.5.2"
+  val scalaVersion = "3.6.2"
```
bishabosha, line 53: you can also try the new clause interleaving in Task.scala

lihaoyi:
Seems like the Scala 3.6.2 upgrade is causing new issues, may be worth punting on that for now and just getting onto 3.5.x first

lihaoyi:
Somehow it is this line that is causing the bootstrap compile to hang

```diff
--- a/main/codesig/src/ReachabilityAnalysis.scala
+++ b/main/codesig/src/ReachabilityAnalysis.scala
@@ -91,7 +91,7 @@ class CallGraphAnalysis(
     case None => ujson.Obj()
   }
 
-  logger.mandatoryLog(spanningInvalidationTree)
+  //logger.mandatoryLog(spanningInvalidationTree)
 }
 
```

lihaoyi:
Looks like bootstrapping on Scala 3.6.2 works again after optimizing `def breadthFirst` a bit. Not sure why it wasn't causing problems before, and why it started causing problems now, but optimizing the queue management in the BFS seems to unblock most of the CI jobs

bishabosha:
Whats the main blocker now?

lihaoyi:
@bishabosha no blockers, I'm just clearing as much non-breaking fixes for 0.12.6 before we merge this and break bincompat for 0.13.0. I expect to bring this up to date and merge it some time this or next week

build.mill
```diff
@@ -50,7 +50,7 @@ object Deps {
   // The Scala version to use
   // When updating, run "Publish Bridges" Github Actions for the new version
   // and then add to it `bridgeScalaVersions`
-  val scalaVersion = "2.13.15"
+  val scalaVersion = "3.6.2"
```
He-Pin, line 53: So mill itself will depend on the latest Scala 3 version? I think this is cool.

.gitignore
```diff
@@ -5,6 +5,7 @@ output/
 .idea_modules
 .idea
 .vscode/
+.zed/
```
lefou, line 8: Shouldn't this be in `.git/info/exclude`, since it's user specific and not project specific?

PR #3381, opened 2024-08-18
Update guava from 33.2.1-jre to 33.3.0-jre

lefou:
- Update guava to 33.3.0-jre
- Revert some formatting
- Set test env locale to "C"


lefou:
CI always fails in `ClientServerTests`, but it looks like `main` isn't building clean, too.

* https://github.com/com-lihaoyi/mill/issues/3387

PR #3386, opened 2024-08-19
Override Failure.toString implementation

jilen:
Avoid the very confusing failure message like
```
1 targets failed
ws.resolvedIvyDeps mill.api.Result$Failure
    mill.api.Result$Failure.map(Result.scala:75)
    mill.api.Result$Failure.map(Result.scala:74)
    mill.scalalib.Lib$.resolveDependencies(Lib.scala:77)
    mill.scalalib.CoursierModule$Resolver.resolveDeps(CoursierModule.scala:146)
    mill.scalalib.JavaModule.$anonfun$resolvedIvyDeps$3(JavaModule.scala:535)
```

### DIFF
--- a/main/api/src/mill/api/Result.scala
+++ b/main/api/src/mill/api/Result.scala
@@ -75,6 +75,7 @@ object Result {
     def map[V](f: T => V): Failure[V] = Result.Failure(msg, value.map(f(_)))
     def flatMap[V](f: T => Result[V]): Failure[V] =
       Failure(msg, value.flatMap(f(_).asSuccess.map(_.value)))
+    override def toString(): String = msg
   }
 
   /**


lolgab:
I don't understand what happened between Mill `0.11.10` and `0.11.11` which broke the error messages.

lihaoyi:
Probably was broken by https://github.com/com-lihaoyi/mill/pull/3330

lolgab:
`getOrThrow` is a lossy conversion in case of `Failure`. Isn't it better to return `Result[Agg[PathRef]]` in `resolveDeps`?

lihaoyi:
Possibly; throwing exceptions is quick and dirty, but the current loss of error message is just a bug AFAICT and isn't fundamental to the question of exception vs result. I thought I propagated the error message in that PR, but clearly i missed something.

We could add an overload to return a `Result`, but I think that's orthogonal to fixing the issue at hand

lolgab:
The bug should be fixed in #3391.
Nevertheless, we can improve the `toString` implementation so that in the rare case that someone calls `.toString`, the result has meaning. It can also be `"Result.Failure($msg, $value)"`

jilen:
> I think the provided patch won't improve the error message presented in the initial comment. In contrast to the merged PR #3391. We should keep the default `toString` of the `case class`.

The `toString` already overrided by `Failing` trait. Implement `toString` actually ensure the msg not lost at every corner case.

PR #3389, opened 2024-08-19
Fix deprecation warning coming from Mill itself

lolgab:


lihaoyi:
Should already be fixed by https://github.com/com-lihaoyi/mill/pull/3379

lolgab:
Yes, I didn't notice the PR.

PR #3390, opened 2024-08-19
Introduce resolveDepsResult which returns failures correctly

lolgab:


lihaoyi:
We should be able to fix it by properly handling the thrown exception message without changing the API and rewriting all the callsites I think. Unless that proves to be impossible, I think that's definitely the better way to go

lolgab:
So the idea is that when we get an exception during an evaluation, if it's of type `Result.Failure` it behaves as if `Result.Failure` was returned by a `Task` ?

lolgab:
On it :)

lolgab:
Implemented in #3391

Issue #3395, opened 2024-08-19
Don't share working dir for parallel running `--no-server` instances


lefou:
https://github.com/com-lihaoyi/mill/pull/3367 introduced a sandbox as a working diretcory for Mill instances.

> For `-i`/`--no-server`, we still always spawn a sub-process for the Mill build logic (as of #3363), and thus are able to control it's PWD. The sandbox for that lives in `out/mill-no-server/`, since it doesn't have a `out/mill-worker-*/` folder to use.

We might need to add something unique to that path, e.g. the process ID, since it's possible to start many `mill --no-server` processes in parallel.

_Originally posted by @lefou in https://github.com/com-lihaoyi/mill/issues/3367#issuecomment-2293955273_

lefou:
Since we need to directory before we created the process, we can't use process ID. We could just use a UUID or a shortened hash of it, since we never need to reused that sandbox directory.

PR #3396, opened 2024-08-19
Add a unique integer to the sandbox dir of `mill --no-server`

lefou:
Fix https://github.com/com-lihaoyi/mill/issues/3395

### DIFF
--- a/runner/client/src/mill/runner/client/MillProcessLauncher.java
+++ b/runner/client/src/mill/runner/client/MillProcessLauncher.java
@@ -1,30 +1,32 @@
 package mill.runner.client;
 
 import static mill.main.client.OutFiles.*;
+
 import java.io.File;
 import java.io.FileInputStream;
-import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.*;
+
 import mill.main.client.Util;
 import mill.main.client.ServerFiles;
-import mill.main.client.ServerCouldNotBeStarted;
 import mill.main.client.EnvVars;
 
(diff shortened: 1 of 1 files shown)


lihaoyi:
If we're going to create a new folder for every run, we should do a best effort to clean it up at the end in a `finally` block. Won't be able to catch every scenario, but at least in the happy path we won't leak folders on disk

lefou:
Yeah, makes sense. I can add it. Should we prepare for the scenario where we want to keep the sandbox dir, e.g. for later inspection in case of a non-zero exit code?

lihaoyi:
How about we clean up asynchronously? e.g. 

1. Each process takes a lock on the sandbox,
2. Every subsequent process lists the existing sandboxes and probe the lock
3. If the lock is available, it means the process has exited, and we can delete the folder

That way for the common workflow of running Mill from the command line, we don't delete stuff immediately, but only upon the next command.

Probably something similar can be done for the long lived `mill-worker-*` folders as well: if the folder exists but both the clientLock and processLock are available, it means the process has exited and can be cleaned up

lefou:
Might work. But is a bit more work than I anticipated with this quick fix. Feel free to take over this PR or merge right away and open another one for the probe-based approach.

lihaoyi:
Let's merge it as is then

lihaoyi (approved):

lefou:
@lihaoyi I added basic cleanup and opened a new issue for the asynchronous cleanup idea.

Issue #3371, opened 2024-08-15
Warning: @deprecated("Use JavaTests instead", since = "Mill 0.11.10")


james-s-w-clark:
I just upgraded to mill 0.11.11, and see a warning for `trait JavaModuleTests in trait JavaModule is deprecated` even though our project isn't using that deprecated trait (`JavaModuleTests`).

I also saw this warning in a recent issue (https://github.com/com-lihaoyi/mill/issues/3359#issuecomment-2282284235), so I figured the warning may come from mill itself. 

I think it's coming from the first line here:
```
  type JavaTests = JavaModuleTests
  @deprecated("Use JavaTests instead", since = "Mill 0.11.10")
  trait JavaModuleTests extends JavaModule with TestModule {
```
https://github.com/com-lihaoyi/mill/blob/main/scalalib/src/mill/scalalib/JavaModule.scala#L39-L41

It can be reproduced on `main` with e.g. `./mill -i dev.run anything`, even if you update all `JavaModuleTests ` to `JavaTests`.

I had a try with some @nowarns on the non-deprecated type in the repo, and removing references to the deprecated JavaModuleTests, but didn't manage to fix it. It looks like it could help though (quick check in Scastie), but no luck trying it in the actual repo:

![image](https://github.com/user-attachments/assets/7e3d1369-bf6a-4164-9b3d-32bf39b2fa11)

![image](https://github.com/user-attachments/assets/23f01197-f126-4874-b9fc-b2eedefcf591)

(Tangent: somehow our -Xfatal-warnings aren't triggered by the warning, so that's good - but we might just have something misconfigured :sweat_smile: )

tculshaw:
Same here. I reverted to 0.11.9 for now.

PR #3393, opened 2024-08-19
Print pretty error message when Result.Failure is thrown (Backport #3391)

lolgab:
The new `.getOrThrow` API allows to short-circuit the result of a `Result`. This prints a correct error message when `Result.Failure` is thrown in a method returning `T`, instead of returned via a method returning `Result[T]`.

Backport of pull request: https://github.com/com-lihaoyi/mill/pull/3391

Pull Request: https://github.com/com-lihaoyi/mill/pull/3393

PR #3400, opened 2024-08-20
Avoid spurious deprecation warnings in generated code (Backport #3379)

lolgab:
Fixes https://github.com/com-lihaoyi/mill/issues/3371

The direct problem is that we are generating code with resolved type aliases, resulting in the `JavaTests` aliases being resolved to deprecated `JavaModuleTests` traits. However, in general generated code should probably not display warnings at all, so better to just slap `@nowarn` on the whole chunk rather than trying to fix the immediate alias resolution problem

Pull Request: https://github.com/com-lihaoyi/mill/pull/3400

### DIFF
--- a/runner/src/mill/runner/MillBuildRootModule.scala
+++ b/runner/src/mill/runner/MillBuildRootModule.scala
@@ -364,6 +364,7 @@ object MillBuildRootModule {
        |
        |import _root_.mill.runner.MillBuildRootModule
        |
+       |@scala.annotation.nowarn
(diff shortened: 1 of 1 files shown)


lihaoyi (approved):

lolgab:
After this PR I think we can release Mill 0.11.12 since the two annoying UX regression are fixed in this PR and in #3393.

lihaoyi:
Yes feel free to tag it

PR #3407, opened 2024-08-22
Remove Reference to Scala Specific Modules in Java Section

gamlerhart:
The target audience of this documentation section are Java developers. So, they probably do not care about the Scala/SBT. 

Further, it isn't actively used in this section of the docs


lihaoyi:
Thanks!

PR #3426, opened 2024-08-27
Overhaul build file management with new `build.mill`/`package.mill` format

lihaoyi:
This PR overhauls how build files are handled, especially in subfolders. The goal is to come up with an approach that is (1) scalable to large builds and (2) works intuitively with how people expect things to work and (3) plays well with IDEs (4) converges with the Scala language to reduce the special casing we have to do.

There's a bunch of hackiness in the implementation, but the end result is that when I load the updated `10-multi-file-build` project into IntelliJ, navigation around the multi-file project works seamlessly (after adding the `.mill` file association):

Screenshot 2024-08-31 at 4 35 16 PM

VSCode works as well, though it only understands `.scala` extensions and will likely need to be patched to work with `.mill`

Screenshot 2024-08-31 at 4 49 46 PM

This means that to begin with, all the IDE tooling around Mill just works, with the IDE being blissfully unaware of the nasty code transformations and other things Mill does. From that base, we can slowly look into removing boilerplate in an IDE/language-compliant way, and removing the backend hackiness while keeping the user experience mostly unchanged

Major user-facing changes:

1. `build.sc` is now `build.mill`, sub-folders can define modules in `package.mill` files. Helper libraries live in files like `foo/bar.mill`
    * Long term Mill needs our own extension because `.sc` is overloaded to work with Ammonite/Scala-CLI scripts. That means IDEs like IntelliJ treat them specially, in ways that are incompatible with Mill (e.g. IDEs don't understand importing between script files). `.sc` is still supported for migration/compatibility reasons. Editors don't currently associate `.mill` with Mill Scala code, but it's a few clicks for the end user and a trivial PR to send to IntelliJ and Metals to support it
    * `.scala` as an extension could possibly work from a technical level, but there will still be user-facing confusion if we want Mill to be able to target non-Scala developers, and possibly technical confusion for tools that can't differentiate between Mill build files and application source files

2. All `.mill` files need a `package` declaration at the top, e.g. `build.mill` needs `package build`, `foo/bar/qux.sc` needs `package build.foo.bar`. I left it optional for the root `build.mill` for migration purposes
    * This is necessary for full IDE support, as editors do not understand the "things in subfolders are automatically in packages" thing that Mill and Ammonite did in the past. We had hacks around `import $file`, but they never were super robust, and `import $file` is itself a hack we've discussed getting rid of
    * We can look into making package declarations optional again in future, but for now this gives us working IDE support with minimal effort, converging with vanilla Scala, at the cost of a little boilerplate per-file. 

3. No more `import $file`: any files with `.mill` (or `.sc`) extensions adjacent to a `build.mill` or `package.mill` are treated as Mill files
    * This simplifies file discovery considerably, v.s. the previous approach that required multiple-passes of parsing and traversing all imported scripts as a graph. It also aligns things more with how Scala and other JVM projects work, where files are typically picked up by folder regardless of imports

Major internal changes:

1. We move the handling of `RootModule` from runtime reflection/classpath-scanning/custom-resolve-logic to code-generation logic that "unpacks" any `object module`, `object build`, or `object foo extends RootModule` into the enclosing wrapper class
    * We can do this as a compiler plugin, which would help preserve line numbers and avoid showing generated code, but it also makes things terribly hard to debug when things go wrong
    * This allows us to ensure `RootModule`s are handled uniformly from Scala code and from the command line, where previously Scala code would need to write `bar.qux.module.mytask` suffix while the CLI would just write `bar.qux.mytask`
    * By making the unpacking name-based, rather than type-based, this removes the "what should we name the root module?" degree of freedom that really shouldn't be necessary and ensures everyone names them consistently

2. Replace the `package object`s previously generated for root modules with normal `object`s named `build` or `module` (reflecting the file names)
    * `package object`s are on their way out in Scala 3 https://docs.scala-lang.org/scala3/reference/dropped-features/package-objects.html. No concrete timeline, but good to stop using them. They also are a common source of bugs around compilation, incremental compilation, etc. since even their present-day implementation isn't terribly robust
    * This also removes the need to refer to `build` as `build.package` since it is now a normal object

3. Generate `final def` aliases in the generated `object`s in order to make code references and module discovery work
    * These aliases allow both Scala-code references to modules without needing this `.build` or `.module` suffix, as well as allow the `Resolve.scala` logic to work
    * We can thus eliminate a whole bunch of gnarly code plumbing multiple-root-modules all over the Mill codebase and gnarly classpath-scanning code to try and identify the sub-folder root modules and wire them up. 
    * These aliases should more reliably generate conflicts than overlapping `package object`/`object`s, which seemed pretty flaky

Limitations:

* Partially migrating subfolders out of a parent `build.sc` into a child `module.sc` no longer works. The aliases we generate for subfolder `module.sc` files always conflicts at compile time with any locally-defined modules

* References to helper files e.g. `foo/bar.sc` with `def qux` would get referenced via `build_.foo.bar.qux` is kind of awkward. I expect this to go away once we get Scala 3 support, and we can use `export` clauses to allow referencing them via `build.foo.qux`, which is more in line with how Scala code normally works where different files in the same folder are all part of the same combined namespace.

* To work around `package object` weirdness e.g. needing to reference things via `foo.bar.package`, we build our own parallel object hierarchy via codegen and provide that to the user. This is awkward, but it looks similar enough people shouldn't notice it, and we can look into somehow fixing the `package object` issues upstream in scala/scala3 in future

* The `package build` prefix is kind of verbose, but it's necessary so we have some way of reliably referencing other files by their fully qualified names. e.g. a file in `util/package.mill` can be referenced by `build.util.*`, but cannot be referenced by `util.*` because `import mill._` pulls in `mill.util` which shadows it.


lihaoyi:
This is probably ready for review

lihaoyi:
After some playing around, I think going with `object`s named `package` rather than `module` seems like the right thing to do here. These are already well-supported by both IntelliJ and VSCode, and minimize the amount of work that would need to be done to make subfolder build.sc files work. 

In fact, if we're willing to add a `package` declaration to the top of each file, IDE support works out of the box today if we name the root module `object package` rather than `object module`: both IntelliJ and VSCode are able to jump to definition to the root module of each folder when its named `object package`, as well as to the contents of each module. In contrast, using `object module`, we would need to teach Intellij how to jump around ourselves, which I expect to be a lot more work to implement and maintain

lefou:
So, one we start to use packages, how can we access objects from the `default` (empty) package? Isn't this a JVM restriction? Don't we need some package for the root build object too?

lihaoyi:
@lefou I updated the PR description with the current approach and results

lefou:
This is a massive change. Unfortunately, my time to review and answer is short, so here are just some thoughts. There is a not so slight chance that I just follow my human urge to resist change. I hope not. I'm just discussion things for now.

## File extension

Since build files don't aim for dedicated editors but instead try to be normal scala code, we should also use a normal Scala file type. (Is a `pom.xml` called `pom.maven`?) We should consider making it `build.mill.sc` or `mill.sc`. We could also try to use a using directive indicating it's a mill file.

## No longer explicit imports

I don't like the fact that we now randomly pick up all `.sc` (`.mill`) files. Mill was always on of the good tools with a single place to inspect to get the whole picture. By default there were no additional files to consider. Unless you find an explicit `$file`-import, then look into that file too, or a `$meta`-import, that also consider a meta-build. Or an (not yet implemented) `$submodules`-import, ok, then also look into subdirectories. After your change, a user, a human or a tool needs to scan the whole directory and it's subdirectories to get the whole picture.

I don't think the arguement, that the scala compiler is doint it the same counts. Sourcecode is typically organised in dedicated directories. But build tools typically don't dictate what's in a project directory alone. It might follow conventions, but should be flexible (e.g. to adapt to the needs of different tools or managers). The directory containing the build files for Mill also have other content, meant for different tools.

We easily could limit the existing file import support to the `build.sc` and `module.sc` files, to avoid deep import traversals to detect all imported scripts. In the end, all `$`-imports are just magic shortcuts for the meta-build. Our meta-build is effectively supporting directory and single-files as it's sources.

## IDE Support

I also don't seen the "hackines" in existing IDE support from Mills side. Or if I understand it your wrong, we should not optimize for the quirks of IDEs. The build project itself is a dedicated BSP build target listing all it's sources (the main build and the module files). If that doesn't work, that's because of a IDEs not fully supporting BSP in the sense, that it collides with dysfunctionalities in the IDE. E.g. associating a file in a subdirectory to one build targets, but another file next to it to another build target. But from a BSP perspective it should work.

## Java Packages

I do see why the use of packages make things more natural to the language. I also see the technical reasons why we can't use the default package for the root project. But maybe we should use a more special root package? Something like ` _mill_`. This is also to avoid collistions with imported libraries. Can't we let the user choose the root package name so she is able to avoid collisions? We could just pick the package of the root project and append it when we specify targets on the cli.

We could also assume, that a root build without a package does not support sub-packages and therefore we have a intuitive way to enable/disable submodules.

## Finally

I struggle with the holistic approach we take here. We enforce all supported files in a project to follow the rules of a single tool. Other build tools like `make` allow me to pick a dedicated buildfile. `Makefile` is just a convenient shortcut. That allows me to experiment by copy and editing and the run `make -f MakefileTest`. I always hoped I can extends Mill to support the same. But when we no longer control the files we use but instead just pick up what we find, we more or less absent from that goal.

lihaoyi:
Yeah this is a huge change. Nevertheless, I think it's one that is really necessary. It's definitely a big change to dump on you, but do give it a chance! 

Apart from the philosophical differences, the migration cost should be reasonable: `.sc` files are still supported, `import $file.foo.bar` becomes `import build_.foo.bar`, and we could even add back `import $file` as an alias for `import build_` as a backwards compatibility shim if necessary. The [mill-bootstrap.patch](https://github.com/com-lihaoyi/mill/pull/3426/files#diff-176852af620fe105e2003a5b304a86df01becb669ea666be5c4090617de73e3a) changes show we need 5 lines changed for Mill's 2000-line build config, which isn't nothing but is also not a huge burden.

A lot of these changes are to try and improve Mill's "IDE support", which IMO is worth a lot. Previously when everything was in one file, even half-broken IDE support was ok since the IDE could resolve most things locally in the same file. But if we want to properly support multi-file builds, we'd basically be going from "decent IDE support" to "totally broken IDE support", which isn't a good value proposition at all. Especially as we go outside the Scala community, "totally broken IDE support" is a non-starter. But with this PR, just need a 1-line PR to IntelliJ and VSCode to add the file-extension-association, and we're done.

I agree that we shouldn't be limited by the quirks of IDEs, but I think we need to be pragmatic in terms of timing and ordering. Mill's IDE support has not significantly improved in the 7 years the project has been around, so I don't think we can rely on "IDE catching up" to any changes we make: even BSP is really just another way to receive the same semi-broken support we used to have via `GenIdea`. I also don't have the resources to go patch IntelliJ and VSCode separately to teach it about all of Mill's own quirks.

Many of Mill's quirks also go against how every IDE out there works: e.g. the idea that _this_ `foo.sc` file is a Mill file, but _that_ `bar.sc` file is a Scala-CLI script is something that no IDE will easily support, as every IDE works by associating file types with files in folders matching name/extension regexes. No IDE does import graph analysis to figure out which file belongs to which file type. Of course it's always theoretically possible to implement whatever you want in an IDE, but by being so unusual Mill makes it especially hard for IDE maintainers to provide a decent experience, which concretely means receiving Mill-specific IDE support later rather than sooner.

In terms of naming, `pom.xml` is AFAIK the exception. There are also a few multi-tool formats like `pyproject.toml`, and things like `project.clj`. But most build files are named after their tool: `Makefile`, `Dockerfile`, `Jenkinsfile`, `build.gradle`, `BUILD.bazel`, `BUCK`, `pants.toml`, `Cargo.toml`, `build.sbt`, `webpack.config.js`, `vite.config.js`, `build.boot`. Basically every build tool out there is `<project-name><suffix>`, `<project-name>.<extension>` or `build.<project-name>`

lihaoyi:
I added in the `build_`/`$file` migration compatibility measures mentioned above, along with making `package` declarations optional for the old `.sc` file extension. Now, `ci/mill-bootstrap.patch` just adds a single empty `ci/package.sc` file to allow `ci/upload.sc` and `ci/shared.sc` to be discovered

```diff
diff --git a/ci/package.sc b/ci/package.sc
index e69de29bb2..6a5b3e8b8e 100644
--- a/ci/package.sc
+++ b/ci/package.sc
@@ -0,0 +1 @@
+
```

lihaoyi:
Did some more updates for migration compatibility. As of this comment, the older `.sc` file behavior is largely preserved, including `import $file`-based discovery and semantics, and so `ci/patch-mill-bootstrap.sh` is empty. The new `.mill` files have the new requirements, including the need for `package` declarations and `package.mill` files.

The new compile-time unpacking of `RootModule`s and the need for the root module to be named `package` applies to both `.mill` and `.sc` files.

lihaoyi:
Going to merge this and proceed with the dogfooding, rebootstrapping, and cutting of RC1. There's definitely issues we'll discover, and further things we need to do. But that can happen during the RC period

nafg:
> .scala as an extension could possibly work from a technical level, but there will still be user-facing confusion if we want Mill to be able to target non-Scala developers, and possibly technical confusion for tools that can't differentiate between Mill build files and application source files

Have you considered something like `.mill.scala`?

This way most tools (e.g. `bat`) will continue working, but I think that solves the concern in the quote?

lihaoyi:
I considered `.mill.scala` and `.mill.sc`, but decided to go with `.mill`

I think net-net, the cost of adding a new file extension is not that large. Sure, initially you get zero support and everything sucks, but most editors allow you to associate a file extension manually, and it's usually a trivial PR to each of the various editors or environments (e.g. Github linguist) to associate the new extension. 

The major thing that I think we need to be more careful about is the _semantics_. Teaching an editor a a file extension is trivial; teaching an editor custom file-type identification logic or scoping/identifier-resolution logic (which is what would be needed to properly support the old `import $file`-based scripts) is much more difficult. There's a reason that after 7 years, IntelliJ still doesn't know whether something is an Mill build, Ammonite script, Scala-CLI script, or IntellIJ worksheet.

Overall, a new file extension seems big, but I bet in the next week or so we'll be able to send PRs to most of the major ones, and that's that. By the time 0.12.0-RC1 is even cut it should already be a non-issue. Whether we use `.mill` or `.mill.sc` or `.mill.scala` then becomes immaterial

lefou:
> There's a reason that after 7 years, IntelliJ still doesn't know whether something is an Mill build, Ammonite script, Scala-CLI script, or IntellIJ worksheet.

I guess that's because `build.sc` files were already supported as Ammonite script very early. And even Ammonite re-used `.sc` from workbooks before.

You make it sound easy, but making a new file extension recognized is hard, esp. if there are already other file extensions for the same type. (I know many editors prefer to guide the user to use _one_ standard instead of many, e.g `.adoc` vs `.asciidoc`.) 

While `*.sc` files are scripts paired with some semantics, `*.scala` files are never standalone, they need a version and a classpath to have any support in IDEs. And any IDE will need to have this support for Mill anyways, otherwise it's just syntax highlighting, which would work reasonably better with `.mill.scala`.

runner/src/mill/runner/CodeGen.scala
```diff
@@ -0,0 +1,297 @@
...
+    def applyTo(s: String, replacement: String): String =
+      s.patch(start, replacement.padTo(end - start, ' '), end - start)
+  }
+
+  private case class ObjectData(obj: Snippet, name: Snippet, parent: Snippet)
+
+  // Use Fastparse's Instrument API to identify top-level `object`s during a parse
+  // and fish out the start/end indices and text for parts of the code that we need
+  // to mangle and replace
+  private class ObjectDataInstrument(scriptCode: String) extends fastparse.internal.Instrument {
```
lihaoyi, line 243: CC @bishabosha this parsing logic may need to be ported to the Scala 3 port. Not sure how the Scala 3 `import $ivy` parser works, but I assume you have some kind of AST, which may make things a lot easier than the visitor-style parsing I have to do here using ScalaParse

bishabosha: Thanks, this is probably next for me

lefou:
> I think net-net, the cost of adding a new file extension is not that large. Sure, initially you get zero support and everything sucks, but most editors allow you to associate a file extension manually, and it's usually a trivial PR to each of the various editors or environments (e.g. Github linguist) to associate the new extension.

I think this equation is false. What I know, avoiding the `.mill` extension on our side is rather cheap. Compared to the count of editors on the planet. And even if you get your PRs accepted eventually in five years to all these editors you think are relevant, you often can't use the latest version or need sometimes work with even another editor or a formatter or tools/libs like editorconfig. We even don't have an idea what editors are used, e.g. I often use a simple CLI editor like `nano` to make bumps in maintained projects. Some editors I use are implemented in languages I'm not proficient in. I don't think opening a PR or whatever the workflow needed for it is easy at all. Let alone the social factor. How to convince an arbitrary editor maintainer who has to maintain hundrets of languages, that she needs to add yet another extention for Scala, just for a little niche build tools she never has heard about? How to convince anyone, if we even don't agree in our team. You're choice seems pragmatic, but dealing with it for others is everything but easy. We loose what we already had.

PR #3431, opened 2024-08-29
Make Mill commands require named CLI parameters by default

lihaoyi:
This aligns the semantics with MainArgs library defaults. Previously we grandfathered in the "positional by default" behavior because Mill pre-dates mainargs, with original behavior inherited from Ammonite. But most CLIs do not allow keyword args to be naively replaced by positional args, and MainArgs thus chose to have such behavior as opt-in. Mill should follow suite

Added a flag `--allow-positional-command-args` to fall back to the old behavior, and covered this behavior with example tests, docs, and updated the changelog.

This is a binary-compatible but semantically-incompatible change, so it should go into 0.12.0. It's a pretty painful breaking change, but it's definitely the right thing to do long term, so better to get it over with. People can also add `--allow-positional-command-args` to their `./mill` file if they want to preserve the old behavior in perpetuity

### DIFF
--- a/example/depth/tasks/2-primary-tasks/build.sc
+++ b/example/depth/tasks/2-primary-tasks/build.sc
@@ -52,6 +52,13 @@ def lineCount: T[Int] = T {
 // on other targets are defined using `foo()` to extract the value from them.
 // Apart from the `foo()` calls, the `T {...}` block contains arbitrary code that
 // does some work and returns a result.
+//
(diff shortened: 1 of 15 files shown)


lefou:
Looks like a good candidate for persistent local configuration.

* https://github.com/com-lihaoyi/mill/discussions/1226

lefou (approved):

PR #3432, opened 2024-08-29
Fix testing as Java application

AnthonyGrod:
Hello, I'm currently working on integrating Mill over the BSP plugin for IntelliJ (https://github.com/JetBrains/hirschgarten - now the BSP plugin code is merged with Bazel-related plugins). 

I have stumbled upon an issue, where the `Test <module.name> as Java application` action did not work with a testing suite. It only performed a Query JVM environment task and tests were not run (later on it turned out that it was caused by the `zincWorker` in `MillJvmBuildServer#jvmRunTestEnvironment`, which could not find the correct test main class). 

After digging into firstly the BSP plugin's code and later on Mill's code, I discovered that in the `MillJvmBuildServer#jvmRunTestEnvironment` there is testing logic lacking. It did not manage to fetch the correct test-specific main class, classpath and main arguments necessary for the testing as a local Java application. 

I figured out that in order to make it work, I needed to leverage the `TestModule`'s logic - I reproduced the `MillJvmBuildServer#testTask` logic of creating the java process' arguments and used it in a new `TestModule#getTestEnvironmentVars` function, which is then used in the `MillJvmBuildServer#jvmRunTestEnvironment` in case of a `TestModule with JavaModule` module (note: the code reproduction I've done might not be the most graceful solution - please let me know if I can do it better). Then the test-specific main class, classpath and main arguments I mentioned are assigned to adequate fields of the function's result - `JvmEnvironmentItem`, which is finally used on the BSP plugin side, which launches the actual local Java application.

I tested the solution with an example Mill project (generated with the g8 template). With BSP plugin installed, I imported the project with the new importing via BSP plugin option (it's still WIP and not yet available in official Scala plugin distribution). Then I checked that indeed the `Test <module.name> as Java application` gutter triggers the tests and their result is displayed in the terminal.

If you need any help with e.g. setting up the BSP plugin, I'm happy to help


lihaoyi:
@AnthonyGrod can you revert all non-essential changes in the PR so it is easier to review exactly what you are trying to do

AnthonyGrod:
@lihaoyi sorry, done.

lihaoyi:
Thanks @AnthonyGrod !

lihaoyi:
I think this looks reasonable, would be good if @lefou could take a look before I merge it

lihaoyi:
@AnthonyGrod can you update the PR description to include how you tested this? Manual testing is fine, but we should at least write down how it was done in case someone needs to reproduce the test in future. After that I can merge this

AnthonyGrod:
@lihaoyi I have updated the description.

lihaoyi:
Thanks!

PR #3436, opened 2024-08-29
Use Zinc-independent main class discovery everywhere

lefou:
Instead we use the compilation-task independent discovery of main classes.

### DIFF
--- a/bsp/worker/src/mill/bsp/worker/MillJvmBuildServer.scala
+++ b/bsp/worker/src/mill/bsp/worker/MillJvmBuildServer.scala
@@ -69,7 +69,15 @@ private trait MillJvmBuildServer extends JvmBuildServer { this: MillBuildServer 
             state,
             id,
             _: JavaModule,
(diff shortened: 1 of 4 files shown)


lefou:
This might also fix the issue described in https://github.com/com-lihaoyi/mill/pull/3432, since it uses a main-class discovery independent of the used compiler. @AnthonyGrod could you please verify?

AnthonyGrod:
@lefou I've also tried this approach earlier but it does not work unfortunately. Discovering the main test class (which `zincWorker` fails to do - the main test class is always `mill.testrunner.entrypoint.TestRunnerMain` and I think the actual testing suite that needs to be run is somehow resolved with main test class arguments) itself is not the biggest issue - it's discovering main test class arguments and it's corresponding classpath for the local Java application launching. I don't think it's possible to resolve all of it without the `TestModule` logic.

lefou:
I see. I think the BSP spec is misleading here, as we aren't interested in any local main class but in the test runner itself. And the actual test runner isn't in the local classpath (compiler output) but the provided dependencies. Although the spec mentions, that this endpoint is also supposed to launch a custom main class.

PR #3437, opened 2024-08-30

Pass `env` parameter to `os.SubProcess.env` in `MillMain`

lolgab:
Fixes https://github.com/com-lihaoyi/mill/issues/2881 since now correct env is used all the time.

Also updates os-lib to `0.10.5`

## Reproduction

``` scala
// build.sc
import mill._
import mill.scalalib._

object root extends RootModule with ScalaModule {
  def scalaVersion = "3.3.1"
}
```

```scala
// src/main.scala
@main
def run =
  println(s"[app] MY_ENV=${sys.env.get("MY_ENV")}")
```

Run the following command:
```
rm -rf out; MY_ENV=hello mill run; mill run
```

Before this PR it outputs:
```
rm -rf out; MY_ENV=hello mill run; mill run
[build.sc] [48/52] compile 
[info] done compiling
[35/48] compile 
[info] compiling 1 Scala source to /Users/lorenzo/scala/repro/out/compile.dest/classes ...
[info] done compiling
[48/48] run 
[app] MY_ENV=Some(hello)
[48/48] run 
[app] MY_ENV=Some(hello)
```

After this PR it outputs:
```
rm -rf out; MY_ENV=hello mill run; mill run
[build.sc] [48/52] compile 
[info] done compiling
[35/48] compile 
[info] compiling 1 Scala source to /Users/lorenzo/scala/repro/out/compile.dest/classes ...
[info] done compiling
[48/48] run 
[app] MY_ENV=Some(hello)
[48/48] run 
[app] MY_ENV=None
```

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -142,7 +142,7 @@ object Deps {
   val commonsIO = ivy"commons-io:commons-io:2.16.1"
   val lambdaTest = ivy"de.tototec:de.tobiasroeser.lambdatest:0.8.0"
   val log4j2Core = ivy"org.apache.logging.log4j:log4j-core:2.23.1"
-  val osLib = ivy"com.lihaoyi::os-lib:0.10.4"
+  val osLib = ivy"com.lihaoyi::os-lib:0.10.4-11-380c77"
   val pprint = ivy"com.lihaoyi::pprint:0.9.0"
   val mainargs = ivy"com.lihaoyi::mainargs:0.7.1"
   val millModuledefsVersion = "0.10.9"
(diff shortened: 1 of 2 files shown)


lihaoyi (approved):
Looks good to me. @lolgab could you update the PR description with how this was tested? Manual testing is fine, just need to write down how you did it in case we need to reproduce the tests later

thierry-st:
Hi here, and congrats for this amazing tool! I was at Scalar conf and @lihaoyi convinced me that I should give it a try :)

This is just to let you know that this (manual) test is currently broken on 0.13.0-M2 (but was still working on 0.13.0-M1).

thierry-st:
As an aside, a workaround on 0.13.0-M2 is to do a `./mill shutdown` when changing env variables values for them to be taken into account.

A related (at least in spirit) behaviour is that the `forkEnv` task using env variables as in 
```
  override def forkEnv = T {
    super.forkEnv() ++ Map("MY_ENV_COPY" -> T.ctx().env("MY_ENV"))
  }
```

does not get invalidated when updating env variables. This is so on (at least) 0.12 and 0.13 mill versions. For this, the workaround is to force re-evaluation by cleaning the task (so `./mill clean forkEnv`).

PR #3442, opened 2024-08-31
Set `forkEnv` to an empty Map

lolgab:
Since os-lib propagates the system environment by default when it runs subprocesses, setting `forkEnv` to the system environment, is redundant, but also confusing, since removing a key from the Map doesn't stop mill to pass the environment variable to the run program. This sets a new default, which is a semantic breaking change, so it should go in Mill 0.12.0, which is a binary compatible but behavior incompatible release.


lihaoyi (approved):

lihaoyi:
What user-visible effect does this have?

lolgab:
> What user-visible effect does this have?

It doesn't make `run` much different, since it's a `T.command` so it runs again every time.
I can see a change in behavior if a user used it as a `T.input` to get the environment, because now it's a fixed `T {}` of an empty map, but that's not the "correct" use of this target. It is still safe to change it in `0.12.0` but it doesn't seem visible in the "normal" use case.

lihaoyi:
Sounds good!

PR #3458, opened 2024-09-04
Remove unnecessary subfolder wrappers for integration tests

lihaoyi:
Since https://github.com/com-lihaoyi/mill/pull/3347 made all test suites run with `pwd` in the test's `.dest` folder by default, we are already guaranteed that different concurrent test suites will not collide, so we do not need the `out/blah/blah` segments to avoid conflicts

### DIFF
--- a/testkit/src/mill/testkit/IntegrationTesterBase.scala
+++ b/testkit/src/mill/testkit/IntegrationTesterBase.scala
@@ -2,18 +2,15 @@ package mill.testkit
 
 trait IntegrationTesterBase {
   def workspaceSourcePath: os.Path
-  private val workspacePathBase = os.pwd / "out" / "integration-tester-workdir"
-  os.makeDir.all(workspacePathBase)
 
   /**
    * The working directory of the integration test suite, which is the root of the
    * Mill build being tested. Contains the `build.mill` file, any application code, and
    * the `out/` folder containing the build output
    *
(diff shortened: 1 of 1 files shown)


lefou:
Does that mean, we can't no longer inspect the out files of multiple (failed) tests?

lihaoyi:
Yes. But previously they were all in `out/<random-number>` folders, so trying to find the output of a particular test was impossible anyway from my experience. We can definitely try to add more logic to assign each test a unique folder

lefou:
I never experienced those random number paths, this was probably after your test suite refactoring. I though, we still had long sub-paths resembling the full test name.

lihaoyi:
yeah the test suite refactor moved a bunch of hardcoded paths to os.temp. Made the code a lot cleaner, but at a cost of filesystem tidyness. Can probably add them back depending on sourcecode.FullName and utest.framework.TestPath to get the tidyness without the boilerplate

lefou:
That sounds reasonable. Both generic sources (`FullName` as well as `TestPath`) already provide meaningful names. Paired with a random number, there should only one or a low count of test output directories to inspect, when in need. In `LambdaTest`, my Java library to write functional tests on top of JUnit/TestNG, I have a autoclosable test path provider, that can clean up automatically on success, but keeps the temp content when failed.

Issue #3447, opened 2024-09-03
Support ErrorProne for Java (1000USD Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 1000USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

ErrorProne is a common Java linter, and we should support it as a `contrib/` module to better support Java projects. We should support all common configuration options, and provide it as an `ErrorProneModule` to allow usage on `JavaModule`s and `ScalaModule`s (which can contain Java sources). This needs to come with a reasonable set of unit tests, docs, and at least one example test that can be included in the documentation

May be able to adapt something from https://github.com/google/error-prone/issues/291

lefou:
Implemented in https://github.com/com-lihaoyi/mill/pull/3460

PR #3460, opened 2024-09-04
Add ErrorProne plugin

lefou:
This adds a new `ErrorProne` plugin to the `contrib` section of Mill.

It provides the `ErrorProneModule` `trait`, which can be mixed into `JavaModule`s. Derived modules like `ScalaModule` are also supported, as long as they support the `javacOptions` task to configure the Java compiler.

We use the [Error Prone](https://errorprone.info/index) project as a compiler plugin. Although this is a simple javac compiler plugin, it's usage isn't trivial. Fetching the correct classpath and supplying options can be tricky, since they need to be provided in a special form. Also using with JVMs starting from version 16 requires special options handling due to the new more restrictive module classpath. This plugin is handling all of the known issues and make its use easy for the user.

Additional configuration options are supported via the `errorProneOptions` target. 

Here is a usage example:

```scala
package build
import mill._, javalib._
import mill.contrib.errorprone._

import $ivy.`com.lihaoyi::mill-contrib-errorprone:`

object `package` extends RootModule with JavaModule with ErrorProneModule {
  def errorProneOptions = Seq("-XepAllErrorsAsWarnings")
}
```

Tasks:
* [x] Implement `ErrorProneModule` and tests
* [x] Support extra options
* [x] Manually test `ErrorProneModule` in private projects
* [x] Finish plugin readme
* [x] Add example and include in documentation
* [x] Document current quirks (Java 17 module classpath, test sub-module behavior)
* [x] (Open follow-up issues)

Fix https://github.com/com-lihaoyi/mill/issues/3447

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -731,6 +731,7 @@ object dist0 extends MillPublishJavaModule {
     build.contrib.jmh.testDep(),
     build.contrib.playlib.testDep(),
     build.contrib.playlib.worker("2.8").testDep(),
+    build.contrib.errorprone.testDep(),
     build.bsp.worker.testDep(),
     build.testkit.testDep()
   )
--- a/build.mill
+++ b/build.mill
@@ -191,9 +191,10 @@ object Deps {
   val sonatypeCentralClient = ivy"com.lumidion::sonatype-central-client-requests:0.3.0"
 
   object RuntimeDeps {
-    val sbtTestInterface = ivy"com.github.sbt:junit-interface:0.13.2"
+    val errorProneCore = ivy"com.google.errorprone:error_prone_core:2.31.0"
     val jupiterInterface = ivy"com.github.sbt.junit:jupiter-interface:0.11.4"
-    def all = Seq(sbtTestInterface, jupiterInterface)
+    val sbtTestInterface = ivy"com.github.sbt:junit-interface:0.13.2"
+    def all = Seq(errorProneCore, jupiterInterface, sbtTestInterface)
   }
 
   /** Used to manage transitive versions. */
--- a/contrib/errorprone/readme.adoc
+++ b/contrib/errorprone/readme.adoc
@@ -1,20 +1,2 @@
 = Mill ErrorProne Plugin
 
-== Caveats / JVM Options
-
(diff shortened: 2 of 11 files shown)


lefou:
I discovered a downside of our current inner test sub-module configuration. Since the inner `JavaTests` trait automatically initializes it's compiler setup from its outer module (e.g. `def javacOptions = T{ outer.javacOptions() }`), compiler plugins like `error-prone` are automatically applied to all test sources. It's rather hard to later disable it in the actual modules, esp. if other `trait`s contribute to the configuration too.

lefou:
This is ready for review. I can't get the assertions in the example project right. @lihaoyi Any idea?

lihaoyi:
I think this looks good.

One question on `-XepAllErrorsAsWarnings`: does it cause compilation to fail? From the name I would have expected it to log the warnings without failing the compile, but your example test indicates the compilation indeed fails

lihaoyi (approved):

lefou:
> I think this looks good.
> 
> One question on `-XepAllErrorsAsWarnings`: does it cause compilation to fail? 

No, it would faile without this options, so it's a nice demontration that options work as expected.

> From the name I would have expected it to log the warnings without failing the compile, but your example test indicates the compilation indeed fails

I can't get the example to work. Have a closer look at the enclosing comment. Either I don't expect any output (which make the example usage missing the point) or I don't use that option, making the example less useful.

lihaoyi:
got it. Lemme give a shot at fixing the example and see if I can get it working

lihaoyi:
Pushed a fix for the example. Turns out `/src/ShortSet.java` should have been `/src/example/ShortSet.java` haha.

The error messages for example tests do suck though. I've lived with them for a while now, but really should do something to make them easier to read

lefou:
> Pushed a fix for the example. Turns out `/src/ShortSet.java` should have been `/src/example/ShortSet.java` haha.

Thanks. So simple yet hard to catch.

lihaoyi:
@lefou I think this looks good to me. If you're happy with it, i can merge it and pay out the bounty using the same details I used before

PR #3467, opened 2024-09-05
Parallelize windows CI

lihaoyi:
* Remove `-j1` parallelism limit on windows, enable parallelism for windows tests by changing `.test` to `.testCached`

* Clean up coursier retry logic and add more cases

* Two of our example tests were conflicting on port `8088`, moved one of them to `8089`


lihaoyi:
Not 100% stable but stable enough, will merge this and keep working on stability in main

PR #3468, opened 2024-09-06
Properly close Scala compiler classloaders in server and unit tests

lihaoyi:
Previously we were just leaking the `URLClassLoaders`, which would give `CodeHeap 'non-profiled nmethods' is full. Compiler has been disabled.` in `scalajslib.test` and sometimes in actual usage for a long lived server. This PR ensures we `ZincWorkerImpl.close` actually calls `URLClassLoader.close`, and also replaced `val eval = UnitTester` in our test suite with `UnitTester.scoped{ eval =>` so we can ensure closing of the `UnitTester` and `URLClassLoader` after each test completes

With this change, `scalajslib.test` no longer prints the above error. Hopefully it makes it stop happening in production as well. `scalajslib.test` also drops from 7.5min to 6.5min (and drops further to 5min when I removed Scala.js 1.3.1 from the test matrix)

### DIFF
--- a/scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
+++ b/scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
@@ -643,9 +643,10 @@ class ZincWorkerImpl(
   override def close(): Unit = {
     val closeableClassloaders = classloaderCache
       .flatMap(_._2.get)
-      .collect{case v: AutoCloseable => v}
+      .collect { case v: AutoCloseable => v }
 
-    val urlClassLoaders = classloaderCache.flatMap(_._2.get).collect{case t: java.net.URLClassLoader => t}
+    val urlClassLoaders =
+      classloaderCache.flatMap(_._2.get).collect { case t: java.net.URLClassLoader => t }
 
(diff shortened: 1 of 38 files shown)


scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
```diff
@@ -641,6 +641,19 @@ class ZincWorkerImpl(
...
+    val closeableClassloaders = classloaderCache
+      .flatMap(_._2.get)
+      .collect { case v: AutoCloseable => v }
+
+    val urlClassLoaders =
+      classloaderCache.flatMap(_._2.get).collect { case t: java.net.URLClassLoader => t }
+
+    // Make sure we at least pick up all the URLClassLoaders, since we know those are
+    // AutoCloseable, although there may be other AutoCloseable classloaders as well
+    assert(urlClassLoaders.toSet[AutoCloseable].subsetOf(closeableClassloaders.toSet))
```
lefou, line 653: This `assert` could fail at runtime, potentially breaking a workflow that could work before. Since we're just detecting after the fact, a log/error message would be more appropriate.

Issue #3332, opened 2024-08-03
Add Java Checkstyle Plugin (1000USD Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 1000USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

Checkstyle is a common Java linter, and we should support it as a `contrib/` module to better support Java projects. We should support all common configuration options, and provide it as an `CheckstyleModule` to allow usage on `JavaModule`s and `ScalaModule`s (which can contain Java sources). This needs to come with a reasonable set of unit tests, docs, and at least one example test that can be included in the documentation

Can possibly be adapted from https://github.com/etsy/sbt-checkstyle-plugin

brbrown25:
I’m going to try and look at this one this week

ajaychandran:
@lihaoyi 
A usecase like a CI build would require the build to fail if Checkstyle errors were found.
I will add a task that checks the error count. Is there a naming convention for such tasks?

PR #3472, opened 2024-09-06
Add Checkstyle plugin

m50d:
(Commits not clean as I'm expecting to squash)

Fix: #3332


m50d:
Build failures look like test infra flakiness/races (not finding the just-published module for the integration test, and a concurrent file opening error on Windows) rather than real issues?

lihaoyi:
Will take a look. Our test suite is flaky, I've been working on it but not 100% stabilized yet

lihaoyi:
Manually retries those jobs

contrib/checkstyle/src/com/etsy/sbt/checkstyle/CheckstyleXSLTSettings.scala
```diff
@@ -0,0 +1,10 @@
+package com.etsy.sbt.checkstyle
+
+import java.io.File
+
+/**
+ * Configuration for a single XSLT transformation of the checkstyle output file
+ *
+ * @author Andrew Johnson
+ */
+case class CheckstyleXSLTSettings(xslt: File, output: File)
```
lihaoyi, line 10: Do we actually need an `output: File` for each XSLT transformation? If we're always applying the same transformations in the same sequence on the input checkstyle config between using it, the choice of output file name seems irrelevant and seems like something Mill can manage

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
```diff
@@ -0,0 +1,184 @@
...
+  /** The `checkstyle` version to use. Defaults to `10.18.1`. */
+  def checkstyleVersion: T[String] = T.input { "10.18.1" }
+
+  /** The location of the `checkstyle` configuration file */
+  def checkstyleConfigLocation: T[CheckstyleConfigLocation] = T.input {
+    CheckstyleConfigLocation.File((T.workspace / "checkstyle-config.xml").toString())
+  }
+
+  /** An optional set of XSLT transformations to be applied to the checkstyle output */
+  def checkstyleXsltTransformations: T[Option[Set[CheckstyleXSLTSettings]]] = T.input { None }
```
lihaoyi, line 37: This type signature here looks a bit off. Do we need it to be wrapped in `Option`, when it could be a `Set.empty`? If transforms are to be run sequentially, and the ordering matters, should it be a `Seq` instead?

contrib/checkstyle/src/com/etsy/sbt/checkstyle/CheckstyleConfigLocation.scala
```diff
@@ -0,0 +1,43 @@
...
+import upickle.default._
+
+/**
+ * Represents a Checkstyle XML configuration located locally, on the class path or remotely at a URL
+ *
+ * @author Joseph Earl
+ */
+sealed trait CheckstyleConfigLocation {
+  def read(resources: Seq[Path]): String
+}
```
lihaoyi, line 15: Could this trait hierarchy be simplified? `Classpath` seems strange, since it's loading stuff from the Mill classloader, which won't contain any of the application classpath. 

`URL` seems redundant, since if we ask for a file, Mill can convert that into a URL by doing `override def foo = T{ requests.get("...") }`, so I don't think we need to have a special case for it in the configuration ADT.

Given that, it might be best to just make the config a `PathRef`, and led anyone who wants to fetch that file from unusual locations just do an `override` the task to call `requests.get` or `os.read(os.resource / ...)` themselves

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
```diff
@@ -0,0 +1,184 @@
...
+import com.etsy.sbt.checkstyle.{
+  CheckstyleConfigLocation,
+  CheckstyleSeverityLevel,
+  CheckstyleXSLTSettings
+}
+import mill.api.Logger
+import mill.{Agg, T}
+import mill.scalalib.{Dep, DepSyntax, JavaModule}
+import mill.util.Jvm
+import net.sf.saxon.s9api.Processor
```
lihaoyi, line 13: I'm curious why we need this external dependency for `Processor`? What is it doing that we cannot do via `scala.xml` or `javax.xml`?

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
```diff
@@ -0,0 +1,184 @@
...
+  def checkstyleClasspath = T {
+    defaultResolver().resolveDeps(checkstyleDeps())
+  }
+
+  /**
+   * Runs checkstyle
+   */
+  def checkstyle = T {
+    val outputLocation = T.dest / "checkstyle-report.xml"
+    val targetFolder = T.dest.toIO
```
lihaoyi, line 63: I think we can avoid calling `.toIO` here; `os.Path`'s `toString` should be the same and we can use `os.makeDir.all`

example/scalalib/module/15-contrib-checkstyle/build.mill
```diff
@@ -0,0 +1,22 @@
...
+}
+
+/** Usage
+
+> ./mill checkstyle # Run checkstyle to produce a report, optionally failing the build if violations are severe enough
+Checkstyle warning found in ...src/InputFileName1.java:2: Top-level class MyAnnotation1 has to reside in its own source file.
+Checkstyle warning found in ...src/InputFileName1.java:13: Top-level class Enum1 has to reside in its own source file.
+Checkstyle warning found in ...src/InputFileName1.java:26: Top-level class TestRequireThisEnum has to reside in its own source file.
+3 issue(s) found in Checkstyle report: ...out/checkstyle.dest/checkstyle-report.xml
+error: ...Checkstyle issues found
```
lihaoyi, line 20: Let's flesh this out a bit more:

- Assert on a handful of different checkstyle errors, not just `Top-level class has to reside in its own source file`
- Let's exercise the process of fixing the errors: we can run `sed` to mangle the files (see other example tests), and run `checkstyle` again and verify that it no longer errors

example/scalalib/module/15-contrib-checkstyle/build.mill
```diff
@@ -0,0 +1,22 @@
+package build
+import mill._, scalalib._
```
lihaoyi, line 2: Let's put this in `example/javalib` since it's more targeted at Java than Scala

contrib/checkstyle/src/com/etsy/sbt/checkstyle/CheckstyleConfigLocation.scala
m50d: `Classpath` uses the application's dependencies (if you comment out the dependency in the `checkstyleClasspathConfig` test it will fail). I think that's an important use case (use a config that's published in some organisational jar). Happy to cut `URL` if you still think that's worthwhile given that we do need the ADT?

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
m50d: Just took the impl from the SBT plugin, I don't have a particular preference myself.

contrib/checkstyle/src/com/etsy/sbt/checkstyle/CheckstyleXSLTSettings.scala
m50d: I think the idea is that each transformation has its own output file (e.g. one transformation for a human-readable HTML page and one for some kind of internal XML report). To be honest it's not a feature I've used, just trying to retain parity with the SBT plugin.

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
m50d: Agreed that there's no point in the `Option` wrapping, will fix that.

I think this is a pretty niche feature so probably the default should be no such file.

example/scalalib/module/15-contrib-checkstyle/build.mill
m50d: I agree with fixing and checking it passes, have extended the test to cover that part.

What sort of different errors do we want to cover? I don't see a lot of value in duplicating the whole checkstyle checksuite, we're not really exercising the plugin code more by having more than a couple of errors (as long as we can pick up more than 1 error reported by checkstyle, I can't see any reasonable way for the plugin to break on a different error), but happy to add any you think are important.

lihaoyi:
@m50d I'm going to go with https://github.com/com-lihaoyi/mill/pull/3516 as the implementation, but I'll pay out 30% of the bounty for the effort you put into this. Email me your bank transfer detail and I'll wire it over. Thanks for working through this with me, and this code has definitely been useful in helping me understand the solution and design space even though we may not end up merging it

m50d:
Thank you - glad to have contributed something and agreed that that's ultimately the better approach.

PR #3473, opened 2024-09-06
Added Checkstyle plugin

### LINKED ISSUE #3332: Add Java Checkstyle Plugin (1000USD Bounty)

ajaychandran:
Resolves #3332.

The implementation differs significantly from the SBT counterpart but I feel this approach is idiomatic.

The basic functionality is ready but there is some more work (mostly documentation) to de done.
Could someone take a look and provide feedback?


contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
```diff
@@ -0,0 +1,97 @@
...
+    val args = Seq.newBuilder[String]
+      .++=(opts)
+      .+=("-c")
+      .+=(conf.path.toString())
+      .+=("-f")
+      .+=(fmt.toString())
+      .+=("-o")
+      .+=(out.toString())
+      .++=(srcs)
+      .result()
```
lihaoyi, line 29: Can we just use `Seq(...)` here?

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
```diff
@@ -0,0 +1,97 @@
...
+    )
+  }
+
+  def checkstyleConfig: T[PathRef] = T.source {
+    checkstyleDir().path / "config.xml"
+  }
+
+  def checkstyleDir: T[PathRef] = T.source {
+    millSourcePath / "checkstyle"
+  }
```
lihaoyi, line 74: This shouldn't be a `.source` unless we expect to find other random files within the folder that we want to account for as part of its hash signature. Otherwise `checkstyleConfig` and `checkstyleXsltDir` are already `.source`s so they already are able to account for changes in their respective files

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleXsltModule.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  override final def checkstyleFormat: T[String] = T {
+    "xml"
+  }
+
+  def checkstyleXslt: T[Set[(PathRef, Seq[String])]] = T {
+    val xslt = checkstyleXsltDir().path
+    T.log.info("scanning for transforms ...")
+    T.log.info(s"  $xslt")
+    os.walk(xslt, maxDepth = 1)
```
lihaoyi, line 38: `os.walk maxDepth = 1` seems like it can be `os.list`

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
```diff
@@ -0,0 +1,183 @@
...
+
+    val classpath = checkstyleClasspath().map(_.path)
+    val options = checkstyleOptions()
+    val config = checkstyleConfig()
+    val format = checkstyleFormat()
+    val report = dst / checkstyleReport()
+    val inputs = sources().map(_.path.toString())
+    val _throw = checkstyleThrow()
+    val transforms = checkstyleTransforms()
+    val transformer = checkstyleTransformer(format)
```
lihaoyi, line 27: These should be inlined and do not need to be lifted to the top of the block like this

contrib/checkstyle/test/src/mill/contrib/checkstyle/CheckstyleModuleTest.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  case class cs(format: String, version: String = ver) {
+
+    def -(src: String) = intercept[RuntimeException] {
+      passed(src)
+    }
+
+    def +(src: String) = assert {
+      passed(src)
+    }
```
lihaoyi, line 45: Let's get rid of these helpers and just inline the `intercept`s and `assert`s

main/util/src/mill/util/Jvm.scala
```diff
@@ -39,6 +39,30 @@ object Jvm extends CoursierSupport {
     os.proc(commandArgs).call(cwd = workingDir1, env = envArgs)
   }
 
+  /**
+   * A version of [[Jvm.callSubprocess]] that does not raise an exception on a non-zero exit.
```
lihaoyi, line 43: Can we add a `check: Boolean = true` parameter to the normal `Jvm.callSubprocess`? We can add a forwarder with the older signature to preserve binary compatibility

contrib/checkstyle/test/src/mill/contrib/checkstyle/CheckstyleModuleTest.scala
```diff
@@ -0,0 +1,67 @@
...
+
+    def passed(src: String) = {
```
lihaoyi, line 47: The name `passed` is kind of weird. Can we change it? Something like `check` or `test`. Also, what's `src`?

contrib/checkstyle/test/src/mill/contrib/checkstyle/CheckstyleModuleTest.scala
```diff
@@ -0,0 +1,67 @@
...
+
+import mill.scalalib.JavaModule
+import mill.testkit.UnitTester
+import mill.testkit.TestBaseModule
+import utest._
+
+object CheckstyleModuleTest extends TestSuite {
+
+  val res = os.Path(sys.env("MILL_TEST_RESOURCE_FOLDER"))
+  val ver = "10.18.1"
```
lihaoyi, line 12: Let's call these `resources` and `version`

contrib/checkstyle/test/src/mill/contrib/checkstyle/CheckstyleModuleTest.scala
```diff
@@ -0,0 +1,67 @@
...
+      val Right(outputs) = eval(mod.checkstyle)
+      val Right(report) = eval(mod.checkstyleReport)
+      val Right(reports) = eval(mod.checkstyleTransforms)
+
+      val reported = outputs.value.exists(_.path.endsWith(os.rel / report.value))
+      val transformed = reports.value.forall {
+        case (_, rel) => outputs.value.exists(_.path.endsWith(os.rel / rel))
+      }
+
+      reported & transformed
```
lihaoyi, line 64: Where are the actual error messages here? I expect to see some assertions that the checkstyle module is reporting the lint warnings/errors that we expect it to report given the code being checked

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
```diff
@@ -0,0 +1,183 @@
...
+    )
+  }
+
+  def checkstyleConfig: T[PathRef] = T.source {
+    checkstyleDir().path / "config.xml"
+  }
+
+  def checkstyleDir: T[PathRef] = T {
+    PathRef(millSourcePath / "checkstyle")
+  }
```
lihaoyi, line 84: This shouldn't be a `PathRef`, since the downstream `T.source`s will wrap things in `PathRef` for us

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
```diff
@@ -0,0 +1,183 @@
...
+
+  def checkstyleFormat: T[String] = T {
+    "xml"
+  }
+
+  def checkstyleOptions: T[Seq[String]] = T {
+    if (isScala) Seq("-x", ".*[\\.]scala") else Seq.empty
```
lihaoyi, line 91: Should we just hardcode this `-x` clause at the point in which we call `Jvm.callSubprocess`? Even for `JavaModule`s there's no hard in adding the exclusion, and that will ensure people who override it without calling `super` don't end up accidentally overriding it and causing issues

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
```diff
@@ -0,0 +1,183 @@
...
+package contrib.checkstyle
+
+import mill.api.{Loose, PathRef}
+import mill.scalalib.{DepSyntax, JavaModule, ScalaModule}
+import mill.util.Jvm
+
+import javax.xml.transform.TransformerFactory
+import javax.xml.transform.stream.{StreamResult, StreamSource}
+
+trait CheckstyleModule extends JavaModule {
```
lihaoyi, line 11: This `trait` and all the `def`s inside need some Scaladoc. Doesn't need to be anything fancy, just 1-2 lines summarizing what it is and what it's for would be enough to help future maintainers and users find their way around the codebase

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
```diff
@@ -0,0 +1,183 @@
...
+
+  def checkstyleOptions: T[Seq[String]] = T {
+    if (isScala) Seq("-x", ".*[\\.]scala") else Seq.empty
+  }
+
+  def checkstyleReport: T[String] = T {
+    val fmt = checkstyleFormat()
+    val ext = if (fmt == "plain") "txt" else fmt
+    s"report.$ext"
+  }
```
lihaoyi, line 98: I feel we don't need to make this configurable and can just hardcode it in `def checkstyle`. If anyone wants to rename it, they can override the target and copy the files to new paths

main/util/src/mill/util/Jvm.scala
ajaychandran: How about renaming the `streamOut` parameter? It's not used in the implementation.

lihaoyi:
The `CheckstyleXsltModule` mentioned in the description seems to be missing in the code changes. Should it be there? Or  was it folded into `CheckstyleModule`?

main/util/src/mill/util/Jvm.scala
lihaoyi: Let's go with adding another parameter. Re-using existing flags is a bad idea in general; people have lost billions of dollars due to that kind of mistake haha https://blog.statsig.com/how-to-lose-half-a-billion-dollars-with-bad-feature-flags-ccebb26adeec

ajaychandran:
@lihaoyi Yes, the modules were unified.

lihaoyi:
Got it

Could you also add an example test? Something similar to https://github.com/com-lihaoyi/mill/pull/3460/files, so we have something working end-to-end we can include in the documentation

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
```diff
@@ -0,0 +1,183 @@
...
+  def checkstyleConfig: T[PathRef] = T.source {
+    checkstyleDir().path / "config.xml"
+  }
+
+  def checkstyleDir: T[PathRef] = T {
+    PathRef(millSourcePath / "checkstyle")
+  }
+
+  def checkstyleFormat: T[String] = T {
+    "xml"
```
lihaoyi, line 87: What are the valid values for this? That should go into the scaladoc

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
```diff
@@ -0,0 +1,183 @@
...
+            .iterator
+            .filter(os.isFile)
+            .map(transform => PathRef(transform) -> Seq(s"${transform.baseName}.${ext.baseName}"))
+        }
+        .toSet
+    } else Set.empty[Transform]
+  }
+
+  def checkstyleTransformer(format: String): Option[Transformer] =
+    Option.when(format == "xml")(Transformer.Xml)
```
lihaoyi, line 146: I don't think this needs to be a public method since it's only used inside `checkstyle`

contrib/checkstyle/test/src/mill/contrib/checkstyle/CheckstyleModuleTest.scala
```diff
@@ -0,0 +1,67 @@
...
+  val ver = "10.18.1"
+
+  def tests = Tests {
+    val src = "sbt-checkstyle"
+
+    test("checkstyle") {
+      test("plain") {
+        cs("plain", "6.2") - src
+        cs("plain", "6.3") + src
+        cs("plain") + src
```
lihaoyi, line 21: I have no idea how to read your short names here. Please make them more verbose and do not use unnecessary operators

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
ajaychandran: Leaving it `public` allows users to plugin their own.
For example, the `sarif` output is a JSON file. A user may chose to run some custom transformation.

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
ajaychandran: Turns out this isn't required at all. The exclusion pattern should be for `java` files.

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
lihaoyi: Rather than having `def checkstyleTransformer(format: String): Option[Transformer]` be an overrideable function that is then used in `def checkstyle` to transform some files, what if we instead break up `def checkstyle` into two separate tasks? 

1. `def checkstyleRaw` that takes generates the initial untransformed files
2. `def checkstyle` that takes the untransformed files and transforms them

In general it's better to have overrideable `T` targets than overrideable helper methods in Mill, because targets can be inspected via the command line or on disk whereas helper methods are opaque

ajaychandran:
@lihaoyi Please review.

Also, how can one detect the root directory of a monorepo? I am thinking a use-case where the Checkstyle configuration is shared.

lihaoyi:
@ajaychandran `T.workspace` inside of tasks will give you the `build.sc` project root

lihaoyi:
or `build.millSourcePath` outside of a task

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
```diff
@@ -0,0 +1,195 @@
...
+    T.log.info("generating checkstyle report ...")
+    T.log.debug(s"running checkstyle with $args")
+
+    val errors = Jvm.callSubprocess(
+      mainClass = "com.puppycrawl.tools.checkstyle.Main",
+      classPath = checkstyleClasspath().map(_.path),
+      mainArgs = args,
+      workingDir = dest,
+      check = false
+    ).exitCode
```
lihaoyi, line 48: This is a bit of an odd API having the exit code be the number of errors, given it's what's available upstream that's fine. But what happens if there are more than 255 errors in detected? Does it cap out, or wrap around, or something else? Whichever behavior it is should document it

contrib/checkstyle/test/resources/compatible-scala/src/com/etsy/sbt/TestClass.scala
```diff
@@ -0,0 +1 @@
+// no content since file should be ignored
```
lihaoyi, line 1: To make this test case a bit more robust, how about putting some garbage syntax/compile/etc. errors in this file? Right now since the file is empty, whether it is ignored or not by checkstyle doesn't affect the test. If the file compiles errors and checkstyle still passes, that tells us that the logic ignoring `.scala` files is working correctly

contrib/checkstyle/test/src/mill/contrib/checkstyle/CheckstyleModuleTest.scala
```diff
@@ -0,0 +1,154 @@
...
+    test("checkstyle") {
+
+      test("format") {
+
+        test("plain") {
+          assert(
+            cs("plain").checkJava(resources / "compatible-java"),
+            cs("plain").checkJava(
+              resources / "sbt-checkstyle",
+              "Utility classes should not have a public or default constructor"
```
lihaoyi, line 25: Is it possible to add a few more things for checkstyle to check for, and assert that they get found? Just to exercise a few more code paths and make sure it works for more than just a single error

contrib/checkstyle/test/src/mill/contrib/checkstyle/CheckstyleModuleTest.scala
```diff
@@ -0,0 +1,154 @@
...
+            // instead of exiting Checkstyle generates a report with a cryptic error
+            cs("xml", "6.2").checkJava(resources / "compatible-java", "File not found"),
+            cs("xml", "6.3").checkJava(resources / "compatible-java")
+          )
+        }
+      }
+    }
+  }
+
+  case class cs(format: String, version: String = "10.18.1", options: Seq[String] = Nil) {
```
lihaoyi, line 97: Let's name this `Tester` or something to be a bit less cryptic

ajaychandran:
@lihaoyi 
Is there a way to detect the root folder in a project? I am thinking a use-case where the Checkstyle configuration needs to be shared across modules.

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
```diff
@@ -0,0 +1,195 @@
...
+   * Additional arguments for Checkstyle.
+   *
+   * @see [[https://checkstyle.org/cmdline.html#Command_line_usage CLI options]]
+   */
+  def checkstyleOptions: T[Seq[String]] = T {
+    Seq.empty[String]
+  }
+
+  /**
+   * Returns a [[Transformer transformer]] for `format`, if supported.
```
lihaoyi, line 119: A bit more verbosity here would be useful. What is a `transformer`? What does it transform? What would I as a Java programmer who's not an expert in checkstyle care about this?

lihaoyi: Also, what's the difference between this and `def checkstyleTransform` below?

contrib/checkstyle/src/mill/contrib/checkstyle/Checkstyle.scala
```diff
@@ -0,0 +1,18 @@
...
+import mill.api.PathRef
+
+/**
+ * [[CheckstyleModule]] output.
+ *
+ * @param errors number of errors found
+ * @param report Checkstyle report
+ * @param transformations transformations applied on the Checkstyle report
+ */
+case class Checkstyle(errors: Int, report: PathRef, transformations: Seq[PathRef])
```
lihaoyi, line 12: Maybe call this `CheckstyleOutput`? Just to be a bit more explicit that it's the output of checkstyle, and not the interface or implementation or something else

ajaychandran:
@lihaoyi 
I botched a merge. Should I resubmit this in a new PR?

lihaoyi:
@ajaychandran whatever you prefer. You can also make a new branch locally force push it to the existing PR branch on github

contrib/checkstyle/test/src/mill/contrib/checkstyle/CheckstyleModuleTest.scala
ajaychandran: How about `Checker`?

lihaoyi: Sure

ajaychandran:
One item is pending.

Below is an excerpt from the plugin documentation. Could some one help with the question in the comment?

```scala
import mill._
import mill.contrib.checkstyle.CheckstyleModule
import mill.scalalib._

object foo extends Module {

  object bar extends MyModule
  object baz extends Module {
    object fizz extends MyModule
    object buzz extends MyModule
  }

  trait MyModule extends JavaModule with CheckstyleModule {

    def checkstyleDir = {
      // TODO how to detect project root directory from any nested module?
      val root: os.Path = ???
      api.PathRef(root / "checkstyle")
    }
  }
}
```

contrib/checkstyle/test/resources/compatible-scala/src/com/etsy/sbt/TestClass.scala
ajaychandran: The logic in Checkstyle already filters non Java files.

I added proper content though. The module name does say compatible.

contrib/checkstyle/test/src/mill/contrib/checkstyle/CheckstyleModuleTest.scala
ajaychandran: I switched to a more idiomatic approach.

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
ajaychandran: Type aliases were extracted out to classes.

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
ajaychandran: I think somewhere along the way the `byte` value for -1 is being transformed into an unsigned `int`.
Will update the comment.

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
ajaychandran: I prefer not to. If Checkstyle adds a new output format (like `sarif` was added in `8.43`), the plugin will work and the documentation would become stale.

If this change is a must, then its probably better to define the values as an `Enumeration`.

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
ajaychandran: This was the original design. I blame my poor naming convention (`checkstyleTransforms` and `checkstyleTransformer`) for the confusion.

The new `getCheckstyleTransformations` function identifies `CheckstyleTransformations` given a `definitionDir` (`checkstyle` configuration directory) and `outputDir` (`checkstyle` task destination directory).

If this is changed to a task, is it possible to access `checkstyle` task destination directory?

ajaychandran: Actually, the new task can use its own destination directory.  I will make the change.

contrib/checkstyle/test/resources/compatible-scala/src/com/etsy/sbt/TestClass.scala
lihaoyi: I know the logic already filters non-java files, I mean to update this `TestClass.scala` to help assert that it filters non-java files, such that if the filtering logic is buggy the test will fail due to finding non-valid-java-code in `TestClass.scala`

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
lihaoyi: I think this is not applicable now that i've gone through your code again

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
lihaoyi: At least the known values should go in the Scaladoc. It doesn't matter if the scaladoc isn't perfect, as whatever people end up googling isn't perfect either. Just list the ones you know of, any caveats your aware of, and we can flesh it out in future if people hit issues

contrib/checkstyle/src/mill/contrib/checkstyle/CheckstyleModule.scala
lihaoyi: Ok. Yes it should be able to just copy the necessary files into its own directory and return PathRefs to them from there. That way we can skip the entire helper method thing entirely and have the logic live in the target implementations

ajaychandran:
@lihaoyi
A usecase like a CI build would require the build to fail if Checkstyle errors were found.
I will add a task that checks the error count. Is there a naming convention for such tasks?

lihaoyi:
@ajaychandran we don't have a convention unfortunately; the current names we have used are

```
./mill -i mill.scalalib.scalafmt.ScalafmtModule/checkFormatAll __.sources + __.mimaReportBinaryIssues + __.fix --check
```

`checkCheckStyle` would be a bit weird. I'd say come up with a name that makes sense yourself and we can normalize all these things later

ajaychandran:
I am changing `checkstyle` to a command with `check` argument.

ajaychandran:
@lihaoyi Please review.

Notable changes:
- `checkstyle` is now a command that accepts `--check` and `--stdout` flags. It also accepts a list of files/folders to be processed.
- The exit code byte conversion was removed.

The only pending issue is that the `--stdout` argument is not working.
The output report does not show up on the console. The program only prints the number of violations.

ajaychandran:
@lihaoyi 
I think the original design with 2 plugins was cleaner.
If you have no objections, I will close this PR and submit a new one.

lihaoyi:
go for it

PR #3479, opened 2024-09-07
Use OS-Lib's `pwdDynamicFunction` to allow sandboxing of `os.pwd`, write sandboxing doc page, introduce `RunModule#runner`

lihaoyi:
Also did some misc cleanup: `synchronizedEval` is now unnecessary since https://github.com/com-lihaoyi/mill/pull/3243 added a coarse grained lock around BSP, prettied up the `#threadId` prefixes, remove empty `.dest/` folders for tidyness 

`RunModule#runner` continues the pattern estabilished with `CoursierModule#defaultResolver`: a convenient way to bundle up the various tasks related to running a modules code in a way that lets downstream callers pass things to `run` while being able to override the default configs (e.g. `forkEnv`, `forkArgs`, etc.) and without the hassle of wrapping things in `T.task`s. If this pattern works out, it would serve as a great alternative to the "pass tasks as arguments to command" pattern we've used in the past. This makes the `11-module-run-task` example much nicer

Also more CI optimizations


main/eval/src/mill/eval/GroupEvaluator.scala
```diff
@@ -342,6 +315,20 @@ private[mill] trait GroupEvaluator {
...
+      this.synchronized {
+        for {
+          p <- usedDest
+          if os.list.stream(p).headOption.isEmpty
+          // workers can be used after they return their value and
+          // may need to continue working with their `T.dest`,
+          if !isWorker
+        } os.remove(p)
+      }
+
```
lefou, line 331: I think this could break tasks, that export their `dest` dir in on purpose. E.g. I use this sometimes to refer to working directories used by other task. Also the `GenIdea` uses empty `dest` dirs as unique `output` dirs when configuring the Java/Scala compiler.

https://github.com/com-lihaoyi/mill/blob/6e35f6e56145da74f11252c1ded78f08b8a9964c/scalalib/src/mill/scalalib/GenIdeaModule.scala#L34-L36

.github/workflows/run-mill-action.yml
```diff
@@ -39,8 +39,6 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
```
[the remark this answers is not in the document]
lihaoyi: We dont need it jn every job, we just need it i  the jobs that do publishing, which arent affected by this change

main/eval/src/mill/eval/GroupEvaluator.scala
lihaoyi: Yeah i noticed that. The use cases i hit was in 11-module-run-task, and in how we used to save a folder for running Mill's own integrstion tests. Both were hacks and i found solutions. We always tell people not to write to upstream dest folders after all.

Could there be another way of marking those tasks that need the folder? e.g. putting a marker file in the folder like git requires (https://www.theserverside.com/), or defining another variant of `Task` that is just a scratchspace that preserves the folder for usage

lihaoyi: another alternstive is to create a folder inside .dest and export that instead, just to tell Mill that you have something in dest tou want to keep

lihaoyi: decided to revert this for now, can discuss it in a follow up

PR #3486, opened 2024-09-08
Swap gitter references to Discord

lihaoyi:
Gitter seems totally borked these days and most discussion is in discord, this makes it official


gamlerhart:
Does this link work if you are not already in the channel? If I use it, it get Discord, but not a Mill channel.

I'm not familiar enough with Discord to know why. Maybe another link it required to 'join' the channel?

PR #3503, opened 2024-09-10
Add `JavaModule.managedJavacOptions`

lefou:
This will hold options contributed by traits like `ErrorProneModule`.

Those are not propagated to the inner test trait intentionally, since those options are typically configured by other means, e.g. dedicated targets.


scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -160,6 +161,11 @@ trait JavaModule
    */
   def javacOptions: T[Seq[String]] = T { Seq.empty[String] }
 
+  /**
+   * Additional options for the java compiled derived from other module settings.
```
lefou, line 165:
```suggestion
   * Additional options for the java compiler derived from other module settings.
```

lihaoyi:
I think the idea sounds reasonable. Why is it called `managed` instead of `mandatory`? Should we standardize on one name?

In the past the convention was just that `mandatory*` were things that traits would set and end users wouldn't. Are we adding a new convention that they're also not propagated? That seems fine, but is something to be aware of (and probably should be written down somewhere)

lefou:
I though about `mandatory`, but went with a different name as the main goal was to have a different propagation behavior then the preexisting `mandatoryIvyDeps` and `mandatoryScalacOptions`.

Both could be changed to no propagate (their value should still compute to the same value, which is a good idicator), but unfortunately, this would be a binary incompatible change and has to wait for Mill `2.13`. 

For the sake of consistency, I can change to `mandatoryJavacOptions`, although the name is even less specific.

lefou:
Changed to `mandatoryJavacOptions`.

lihaoyi (approved):
yeah the name could be improved, maybe in 0.13.0 or something. Until then we'll just have to compensate by writing really good scaladoc so people who encounter it can read what it is for

lolgab:
What if we exposed a task in `JavaModule` that decides how dependencies are propagated to the `Tests` submodules?
It could be something like:

```scala
def javacOptionsTestsPropagation: Task[Seq[String] => Seq[String]] = T.task { identity }
```

This way your plugin can override it to filter out the option that you don't want to propagate.

lefou:
I think filtering options isn't my favorite go-to solution, since it's always a bit error-prone. Since we use stackable traits which can't be remove we also should take care to not add unnecessary options, which can't be easily removed.

lefou:
I took the opportunity to no longer propagate `mandatoryScalacOptions` to see if all tests pass. They should. I kept the override to just can `super`, so that binaries expecting this `def` in `ScalaTests` should work without issues.

lefou:
I checked, we're not propagating `mandatoryIvyDeps` to tests, so we now no longer propagate any `mandatory*` target to tests. Nice and consistent.

Issue #3465, opened 2024-09-05
Mill parallelism is not stable on windows


lihaoyi:
Seems to constantly hit problems with Coursier. For now I turned off parallelism in our windows CI, but this results in windows CI being really slow, since we're not making use of the 4 cores the Github Actions runners have. Really the root cause should really be fixed

Examples

https://github.com/com-lihaoyi/mill/actions/runs/10709481836/job/29694216401
```
X mill.integration.MillPluginClasspathTest.exclusions 12487ms 
  utest.AssertionError: assert(res1.isSuccess)
  res1: mill.testkit.IntegrationTester.EvalResult = EvalResult(false,,1 targets failed
  resolvedIvyDeps java.io.FileNotFoundException: C:\Users\runneradmin\AppData\Local\Coursier\cache\v1\https\repo1.maven.org\maven2\com\disneystreaming\smithy\smithytranslate-proto_2.13\0.5.3\smithytranslate-proto_2.13-0.5.3.pom (The process cannot access the file because it is being used by another process))
    utest.asserts.Asserts$.assertImpl(Asserts.scala:30)
    mill.integration.MillPluginClasspathTest$.$anonfun$tests$2(MillPluginClasspathTest.scala:36)
```

https://github.com/com-lihaoyi/mill/actions/runs/10729224027/job/29755559797?pr=3467

```
main.resolve.resolvedIvyDeps java.nio.file.AccessDeniedException: C:\Users\runneradmin\AppData\Local\Coursier\cache\v1\https\repo1.maven.org\maven2\com\lihaoyi\mainargs_2.13\0.7.4\.mainargs_2.13-0.7.4.jar__sha1.computed11218541499758406365.tmp -> C:\Users\runneradmin\AppData\Local\Coursier\cache\v1\https\repo1.maven.org\maven2\com\lihaoyi\mainargs_2.13\0.7.4\.mainargs_2.13-0.7.4.jar__sha1.computed
Error: Process completed with exit code 1.
```

https://github.com/com-lihaoyi/mill/actions/runs/10729224027/job/29755559201?pr=3467

```
    com.fasterxml.jackson:jackson-base:2.11.4 
  	not found: C:\Users\runneradmin\.ivy2\local\com.fasterxml.jackson\jackson-base\2.11.4\ivys\ivy.xml
  	not found: C:\Users\runneradmin\AppData\Local\Coursier\cache\v1\.structure.lock (The process cannot access the file because it is being used by another process)
```

[...]

PR #3505, opened 2024-09-10
Update coursier to 2.1.11, disable coursier retries

alexarchambault:
Fixes https://github.com/com-lihaoyi/mill/issues/3465


alexarchambault:
About disabling the retries: there's still a chance some cases aren't handled by https://github.com/coursier/coursier/pull/3066 (the retry in case of `download error` is pretty large in particular) or that errors aren't caught at the right place in it. But I should be around in the near future, so I can fix things swiftly in coursier if we see such errors again.

lihaoyi:
Looks good to me. We'll have to see how stable it is post merge, but we can fix any further flakiness as we encounter it

PR #3511, opened 2024-09-10
Format using scalafmt command rather than ScalafmtModule/checkFormatAll task

alexarchambault:
This adds a `scalafmt` command to `ScalafmtModule`, that can be run like
```text
$ ./mill mill.scalalib.scalafmt.ScalafmtModule/scalafmt
```
(can be done from the Mill repo once it's re-bootstrapped with these changes)

This command fetches and runs the scalafmt CLI as-is. It accepts arguments, and just passes them to the scalafmt CLI.

This also makes scalafmt ignores more files in the Mill repo, so that running `scalafmt` from the root of the repo succeeds, with https://github.com/com-lihaoyi/mill/pull/4691 and https://github.com/com-lihaoyi/mill/pull/4693.


lihaoyi:
I think it's reasonable, and arguably better than the fine grained approach of doing it vis `__.sources`, but we need to make sure it works and thete's escape hatches to disable when it doesn't (e.g. some kind of .fmtignore file?)

lihaoyi:
In principle I think a coarse grained formatting has many advantages over the fine grained approach of going through the build tool to list sources. In practice I think it still can be done, just needs to be done carefully so we can handle the inevitable edge cases (template files, magic comments, generated code, output folders, test data, ...)

lihaoyi:
Maybe something we can do eventually but probably not an immediate priority for the 0.12.0/0.13.0 releases coming up

lefou:
I'd expect some failing tests. Those, that test the `scalafmt` integration.

.github/workflows/run-mill-action.yml
```diff
@@ -53,10 +53,10 @@ jobs:
 
       - uses: coursier/cache-action@v6
 
-      - uses: actions/setup-java@v4
+      - uses: coursier/setup-action@v1
         with:
-          java-version: ${{ inputs.java-version }}
-          distribution: temurin
+          apps: scalafmt
+          jvm: temurin:${{ inputs.java-version }}
```
lihaoyi, line 59: I'd like to avoid adding a dependency on a global `scalafmt` binary. Adding it in CI will necessitate people installing it locally to match the CI behavior, doubles the number of things that need to be installed (previously it was only 1: the JVM).

We already have Scalafmt on the Mill classpath, so instead of installing it globally we could define a top-level command in `build.mill` that uses Scalafmt programmatically (either inprocess or subprocess) on the files in `T.workspace`. That would achieve the same thing but maintain Mill's minimal unmanaged dependency footprint

.scalafmt.conf
```diff
@@ -22,3 +22,11 @@ project.git = true
 
 runner.dialect = scala213
 
+project.excludeFilters = [
+   # can't get fileOverride to work to specify scala3 as dialect for these
+  "scalalib/test/resources/hello-dotty/*"
+  # these have scaladoc comments with a particular shape that we don't want to reformat
+  "example/depth/large/*"
```
lihaoyi, line 29: The only reason we don't need to ignore the rest of `example/` is because Scalafmt doesn't pick up the `.mill` files. If Scalafmt learns how to do so, or if we rename them to `.mill.scala`, we'll need to go back and ignore the rest of `example/` as well

.github/workflows/run-mill-action.yml
lihaoyi: Alternatively, if we want to bypass mill entirely (e.g. for performance), we could have a `ci/scalafmt.sh` script to download the ScalaFmt graal native executable appropriate for your OS, cache it somewhere, and run that

alexarchambault: That's what I did in the last push

alexarchambault:
Converted to draft, so that we don't merge by mistake before https://github.com/com-lihaoyi/mill/pull/3526

lefou:
Can't we add a command to `ScalaFmtModule` which accepts free arguments and runs from the workspace root as an drop-in replacement for `scalafmt`? Convenient and we could avoid the extra `ci/scalafmt.sh` script.

alexarchambault:
@lefou I did that for https://github.com/com-lihaoyi/mill/pull/3511#discussion_r1752975390 (and I personally use the [native scalafmt binaries](https://github.com/VirtusLab/scalafmt-native-image) that the script pulls), but we can probably add both, maybe. The native scalafmt is quite convenient, as it's much faster.

lefou:
> @lefou I did that for [#3511 (comment)](https://github.com/com-lihaoyi/mill/pull/3511#discussion_r1752975390) (and I personally use the [native scalafmt binaries](https://github.com/VirtusLab/scalafmt-native-image) that the script pulls), but we can add both, maybe. The native scalafmt is quite convenient, as it's much faster (more for local use than on the CI, actually).

I fail to see that you did that, but it's probably squashed away (which is btw not so nice for reviewing PRs). 

I thought, Haoyi was commenting on the `coursier/setup-action` pulling `scalafmt` by using `coursier` which is essentially one of many ways to install another binary on the machine. What I'm proposing is, that we use Mill to act as an drop-in replacement for `scalafmt` in case the user wants it. Mill already has the classpath anyways, so it's rather easy to forward all params to the right `main` method. From a user perspective, it does not require a new locally installed tool beside the already configured Mill build.

lihaoyi:
Yes I don't want an additional dependency on the coursier install step. We currently bootstrap Mill pretty narrowly using only an installed JVM, and I would like to keep it that way.

This discussion also needs to broadened into what kind of autoformatting workflows we recommend to Mill users. As of now in master, running ScalaFmt is done via:

```
./mill mill.scalalib.ScalafmtModule/
```

Unless there's some requirements unique to Mill's own codebase, whatever autoformatting workflow we do ourselves should also be the best practice we recommend to users. I don't think Mill is that special w.r.t. autoformatting, so we should try to keep the dev/user workflows the same here. And users should not need to install separate binaries in order to use Mill

alexarchambault:
That makes sense. I guess a command to blindly reformat the workspace could be added in `ScalafmtModule `.

.scalafmt.conf
```diff
@@ -24,7 +24,13 @@ runner.dialect = scala3
...
   excludePaths = [
     "glob:**/example/scalalib/linting/1-scalafmt/src/Foo.scala",
-    "glob:**/mill/out/**"
+    "glob:**/scalalib/test/resources/checkstyle/**",
+    "glob:**/mill/out/**",
+    "glob:**/scalalib/test/resources/giter8/hello.g8/src/main/g8/build.mill",
+    "glob:**/scalalib/test/resources/javalib/palantirformat/**",
+    "glob:**/integration/failure/parse-error/**",
+    "glob:**/testkit/test/resources/example-test-example-project/build.mill",
+    "glob:**/scalalib/test/resources/scalafmt/core/src/**"
```
lefou, line 33: I think we should leave comments why these are excludes, to guide the maintainer. I think [`scalafmt.conf` is HOCON](https://scalameta.org/scalafmt/docs/configuration.html), not JSON.

lefou (approved):

Issue #3408, opened 2024-08-22
Please utilize COURSIER_CREDENTIALS environment variable for all dependency retrievals


johnsonjii:
When running Mill in an environment that can access Maven Central artifacts without authentication, `build.sc` scripts that use `import $ivy` statements and `ivyDeps` work correctly.

When running Mill in an environment that has to authenticate with the artifact repository, `build.sc` scripts fail to retrieve the requested artifacts. This is true for `import $ivy` statements, `ivyDeps`, and other artifacts that Mill is trying to fetch for itself.

In both kinds of environment, plain Ammonite scripts work correctly.

Given the following:

- Ammonite REPL & Script-Runner, 3.0.0-M2
- The Mill bootstrapping wrapper scrpit

.mill-version
```
0.11.12
```

build.sc
```
import mill._, scalalib._
import $ivy.`commons-io:commons-io:2.16.1`

object main extends JavaModule {
  def ivyDeps = Agg(ivy"commons-io:commons-io:2.16.1")
}
```

plain.sc
```
import $ivy.`commons-io:commons-io:2.16.1`
println("Done")
```

Note: Between each of the various experiments below, I throw away the Mill `out` directory and the Coursier cache directory to simulate an initial environment.

### In an unrestricted environment:

```
$ export COURSIER_REPOSITORIES="ivy2Local|central"
```

Mill behaves correctly...
```
$ ./mill main.prepareOffline
[13/13] main.prepareOffline
```

and Ammonite behaves correctly...
```
$ amm plain.sc
https://repo1.maven.org/maven2/commons-io/commons-io/2.16.1/commons-io-2.16.1.pom
  100.0% [##########] 19.2 KiB (16.8 KiB / s)
https://repo1.maven.org/maven2/org/apache/commons/commons-parent/69/commons-parent-69.pom
  100.0% [##########] 75.2 KiB (238.0 KiB / s)
https://repo1.maven.org/maven2/org/apache/apache/31/apache-31.pom
  100.0% [##########] 23.0 KiB (117.2 KiB / s)
https://repo1.maven.org/maven2/org/junit/junit-bom/5.10.2/junit-bom-5.10.2.pom
  100.0% [##########] 5.5 KiB (28.6 KiB / s)
https://repo1.maven.org/maven2/commons-io/commons-io/2.16.1/commons-io-2.16.1-sources.jar
  100.0% [##########] 515.3 KiB (429.1 KiB / s)
[...]

PR #3513, opened 2024-09-10
Use coursier 2.x API

alexarchambault:


alexarchambault:
(CI error looks legit)

lihaoyi:
I think this looks reasonable to me, though I'm not an expert in these APIs so I'll have to trust you are using them correctly cc @lefou

alexarchambault:
Fixes https://github.com/com-lihaoyi/mill/issues/3408 (`Don't clear coursier FileCache credentials` commit)

main/util/src/mill/util/CoursierSupport.scala
```diff
@@ -63,136 +76,132 @@ trait CoursierSupport {
...
-        val errorDetails = errors.map(e => s"${System.lineSeparator()}  ${e.describe}").mkString
-        Result.Failure(
-          s"Failed to load ${if (sources) "source " else ""}dependencies" + errorDetails
-        )
       }
     }
   }
 
+  @deprecated(
+    "Prefer resolveDependenciesMetadataSafe instead, that handles errors better",
```
lefou, line 116:
```suggestion
    "Prefer resolveDependenciesMetadataSafe instead, that handles errors better",
    "Prefer resolveDependenciesMetadataSafe instead, which returns a Result instead of throwing exceptions",
```

lefou (commented):
Looks good to me. Two nits below.

lefou (approved):

Issue #3451, opened 2024-09-03
First Party Support for the Kotlin language (2000USD)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 2000USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

https://github.com/lefou/mill-kotlin/ exists, but it is a standalone project without many examples, and not enough for a newbie to pick up and begin using Mill with Kotlin. This ticket is to do the following:

1. In-source https://github.com/lefou/mill-kotlin/ into the Mill repo as a `kotlinlib` submodule
2. Write Kotlin versions of the `example/javalib/{basic,builds,module,testing}` example integration tests, where reasonable
3. Flesh out whatever missing functionality is needed to do the above (I'm sure we'll hit things!)

As this ticket is large, I am splitting the bounty into two parts:

1. 1000USD to get the `mill-kotlin` in-sourced and the `example/javalib/basic` examples converted to Kotlin and included on the docsite
2. 1000USD to get the remaining examples in  `example/javalib/{builds,module,testing,publishing}` converted to Kotlin and included on the docsite

The end result is that we have a suite of tested example documentation demonstrating how to use the Mill build tool to build a large variety of Kotlin projects mirroring the example documentation for Java and Scala projects.

Some code may be adaptable from https://github.com/rife2/bld-kotlin. ChatGPT (or claude or whatever) would probably also be a great help translating the examples; I used it heavily when setting up the Java example tests, but in the end the human still needs to debug things and polish it up so it's presentable for pedagogical purposes

githubbin765:
If I claim this, can others also claim or start work on the bounty?

lihaoyi:
@githubbin765 As mentioned in https://github.com/orgs/com-lihaoyi/discussions/5, bounties remain open until completed. 

In this case I (just) broke up the bounty into two phases so it's not such a big all-or-nothing ticket, and I can break it up further if you think it would be useful. 

But fundamentally there is no locking of bounties, so others can pick it up as well, and first PR wins (at my discretion)

githubbin765:
Have you considered changing this to where developers can claim the bounty since non claiming potentially means multiple people working on bounties without any guaranteed compensation?

lihaoyi:
I have. It's ultimately a tradeoff. Multiple people
racing on each bounty has not been a problem so far, but i understand it may change as the scope of the biunty grows. Maybe I'll try a locking policy next round

[...]

PR #3514, opened 2024-09-10
Add Kotlin integration

0xnm:
This PR solves Part 1 from https://github.com/com-lihaoyi/mill/issues/3451:

* It in-sources https://github.com/lefou/mill-kotlin with minimal changes
* It adds basic samples

I still need to finalize it (especially automated test for example `4-builtin-commands` fails right now), but most of the work is done.

Things I noticed in the original implementation (https://github.com/lefou/mill-kotlin):

* Kotlin stdlib is added by overriding `ivyDeps`, meaning that if the user of `KotlinModule` wants to add its own deps and forget to call `super.ivyDeps() ++ <their deps>` and just does `def ivyDeps = <their deps>`, then no stdlib will be in the classpath during the compilation and it will fail with cryptic error. Such approach is thus error-prone.
* No Kotlin Reflection library is added automatically to the compile classpath, meaning that if code to compile contains any Kotlin Reflection classes, compilation will fail (because `-nostdlib` [flag](https://kotlinlang.org/docs/compiler-reference.html#nostdlib) is used for compilation)

@lihaoyi Feel free to give your feedback while this is in a Draft state if you want.


example/kotlinlib/basic/1-simple/build.mill
```diff
@@ -0,0 +1,101 @@
...
+  def ivyDeps = super.ivyDeps() ++ Agg(
+    ivy"net.sourceforge.argparse4j:argparse4j:0.9.0",
+    ivy"org.apache.commons:commons-text:1.12.0"
+  )
+
+  object test extends KotlinModuleTests with TestModule.Junit4{
+    def ivyDeps = super.ivyDeps() ++ Agg(
+      ivy"com.google.guava:guava:33.3.0-jre"
+    )
+  }
```
lihaoyi, line 20: For the purposes of these examples, let's swap over to Kotlin libs where possible:

* https://github.com/ajalt/clikt seems to be a popular argument parser and can replace `argparse4j`
* https://github.com/Kotlin/kotlinx.html can replace `commons-text` for what we're doing (we use Scalatags in the Scala versions)
* https://github.com/kotest/kotest seems like the winner for Kotlin-specific test frameworks
* If we use `kotest` via `ivyDeps`, we can drop the `guava` thing (whose main use in the Java example is to show test third party dependencies, and which we do not have in the Scala example)

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -0,0 +1,211 @@
...
+
+  /**
+   * The dependencies of this module.
+   * Defaults to add the kotlin-stdlib dependency matching the [[kotlinVersion]].
+   */
+  override def ivyDeps: T[Agg[Dep]] = T {
+    super.ivyDeps() ++ Agg(
+      ivy"org.jetbrains.kotlin:kotlin-stdlib:${kotlinVersion()}"
+    )
+  }
```
lihaoyi, line 51: Let's move this to `mandatoryIvyDeps`? That's where we put most things that are included by `trait`s and not intended to be removed by users. That would also let us remove the `super.ivyDeps()` from the examples, and bring them closer to the Java/Scala examples

lihaoyi:
I think not supporting kotlin-reflect is fine for this initial PR for Part 1 of the bounty. Part 2 of the bounty covers a lot more scenarios, so we can look at figuring out supporting kotlin-reflect and other fanciness (compiler plugins?) then, as the counterpart to `example/javalib/module/6-annotation-processors` and `example/scalalib/module/6-scala-compiler-plugins`

example/kotlinlib/basic/1-simple/build.mill
0xnm: I would say that plenty of tests in Kotlin-first codebases are using JUnit framework, but for the sake of the examples/illustrations/sanity check `kotest` can be used indeed.

0xnm:
@lihaoyi Code-wise this PR is done, I think. I addressed your suggestions. The only remaining thing is to add Kotlin samples to the generated docs.

lihaoyi:
The code looks good I think. The example tests just need to be tweaked I think to match the stdout format of the kotest library. You can run them locally via `./mill -i 'example.kotlinlib.basic[1-simple].local'` and adjust them accordingly, and once they're green I'll merge this and wire the example tests into the docsite myself

lefou:
> * It in-sources https://github.com/lefou/mill-kotlin with minimal changes

Don't forget to address copyright requirements. `mill-kotlin` is not MIT licensed.

lefou (commented):
I'd expect some refactorings, since the original codebase was cross-built against multiple Mill versions. Some compromises had to be made to accomplish that, which are no longer necessary here. E.g. the `KotlinModulePlatform` is superfluous and should be merged with `KotlinModule` to keep things simple and understandable.

Beside the requirements of the original Apache License 2.0, I'd expect at least some traces of history in the documentation and/or Scaladocs. Not doing so not only pisses the original contributors of the code off (which is mostly me), but it also doesn't guide the user and developer in understanding the code and which relationship it has to the pre-existing `mill-kotlin` project. I don't think, it's acceptable for an Open Source Software project, to dishonor others people work.

0xnm:
@lefou Yes, sure, it was on my to-do list before marking PR as final, after code changes are complete.

I added the copyright now in https://github.com/com-lihaoyi/mill/pull/3514/commits/85c1f30650682a68e2bd609eca49566e93b7b020. However, since https://github.com/lefou/mill-kotlin has no explicit copyright statement neither in the source files, nor in the [license file](https://github.com/lefou/mill-kotlin/blob/main/LICENSE#L190), I had to come up with the following (since I traced back the first commit to 2020 and there are 2 contributors):

```
/*
 * Copyright 2020-Present Original lefou/mill-kotlin repository contributors.
 */
```

Feel free to suggest another one if needed.

Regarding the refactoring: I merged `KotlinModulePlatform` in `KotlinModule` in https://github.com/com-lihaoyi/mill/pull/3514/commits/3385f948245ebd30e7c0f16e63e6f1ee41e5e2d1, but I would refrain from the further changes, because I think it is better to keep the focus of this PR on the integration part and keep the original files as-is for that. Refactoring can be done in the further PRs if needed.

0xnm:
@lefou I fixed Kotlin tests, but there is one job (`'integration.{feature,failure}[_].fork.testCached`) which failed and I guess it is not related to this PR. Probably there is a flaky test there, but I have no rights to re-run the job (without pushing empty commit).

lihaoyi:
I re-ran the CI job, looks good. I'll go ahead and merge it and add a landing page for it, and will include a reference to the original repo as @lefou requested. @0xnm you can email me your bank transfer details and I'll pay out the first half of the bounty

lefou:
What I actually meant to request is to respect the original license, which is Apache License 2. Although it allows you to do almost everything, it also demands to point it out and to include the license text somewhere in the project, if you use or derive the work. 

The fact that the actual code you included is from me is rather secondary here. My point is about free and open source work in general. It's only free if you comply to the license, whether you like it or not. And since this is missing in this pull request, I had rather not accepted this PR as-is.

Issue #3510, opened 2024-09-10
Document how to break out of Mill Sandboxing


lihaoyi:
T.workspace, build.millSourcePath, mill.api.WorkspaceRoot.workspaceRoot, and we proably need to add something similar for tests (maybe`MILL_WORKSPACE_ROOT`?)

PR #3518, opened 2024-09-12
Add docs for how to break out of Mill sandbox folder

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/3510

I decided to try and standardize on a `MILL_WORKSPACE_ROOT` environment variable that is present in both build files and tests, to try and provide a seamless experience to our users. 

`mill.api.WorkspaceRoot.workspaceRoot` and `T.workspace` are still there but perhaps we can get rid of them in 0.13.0 in favor of using `MILL_WORKSPACE_ROOT` everywhere. After all, accessing the workspace root should be inconvenient to discourage people from doing it unnecessarily, so providing short concise helper methods to do so may even be a net negative

### DIFF
--- a/docs/modules/ROOT/pages/Mill_Sandboxing.adoc
+++ b/docs/modules/ROOT/pages/Mill_Sandboxing.adoc
@@ -8,6 +8,10 @@ include::example/depth/sandbox/1-task.adoc[]
 
 include::example/depth/sandbox/2-test.adoc[]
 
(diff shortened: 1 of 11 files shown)


lefou:
I don't think we should get rid of `T.workspace`. Accessing the workspace from a task is typically a required action when integration additional tools into Mill. Falling back to string-ly typed typo-prone or undefined environment variables is a step back.

PR #3521, opened 2024-09-12
Move `.mill.sc` extension to `.mill.scala`

lihaoyi:
`.scala` has a lot better support for things like `package` declarations that `.sc` does not, and so will serve as a better alternative to `.mill` until IDE support for the new extension catches up


mill-build/src/ExampleParser.scala
```diff
@@ -6,9 +6,15 @@ object ExampleParser {
     val states = collection.mutable.Buffer("scala")
     val chunks = collection.mutable.Buffer(collection.mutable.Buffer.empty[String])
 
-    val rootBuildFileNames = Seq("build.sc", "build.mill", "build.mill.sc")
-    val buildFile = rootBuildFileNames.map(testRepoRoot / _).find(os.exists)
-    for (line <- os.read.lines(buildFile.get)) {
+    val rootBuildFileNames = Seq("build.sc", "build.mill", "build.mill.scala")
+    val buildFile = rootBuildFileNames.map(testRepoRoot / _)
+      .find(os.exists)
```
lefou, line 11: I think we hould report multiple detected root files? If we have a strong precedence, just reporting a warning should be enough. E.g. to allow for sym-linking `build.mill` to `build.mill.scala` without build failures.

lihaoyi: Added a logged warning

PR #3523, opened 2024-09-12
Try and simplify publishing setup

lihaoyi:
This PR attempts to in-source most of the convenience provided by https://github.com/ckipp01/mill-ci-release into Mill proper. The boilerplate necessary to call `publishAll` on mac laptops or linux github actions environments was trivial and straightforward, so we just try to automate it:

1. Prefix all environment variables with `MILL_`: `MILL_SONATYPE_USERNAME`, `MILL_SONATYPE_PASSWORD`, `MILL_PGP_SECRET_BASE64`, `MILL_PGP_PASSPHRASE` to try and add some rudimentary ownership and namespacing

2. Auto-detect and import `MILL_PGP_SECRET_BASE64`, and auto-detect `MILL_PGP_PASSPHRASE` and include it in the default GPG arguments. `MILL_SONATYPE_USERNAME` and `MILL_SONATYPE_PASSWORD` were already auto-detected

3. Flesh out the `defaultGpgArgs` to allow the optional passphrase to be passed in, and include all the flags that Mill uses in its own CI publishing job

4. Set the defaults for `readTimeout`, `awaitTimeout`, `stagingRelease`, `publishArtifacts` to match the common convention

5. Set `release = true` by default

Updated the `ci/release-maven.sh` script to publish using a bootstrap version of Mill, so once this is merged I can immediately test it without needing to go through a rebootstrapping process. The current behavior when passing in flags explicitly should be unchanged, this just acts a lot more aggressively at setting defaults


lihaoyi:
@ckipp01 as the author of mill-ci-release would love to have your review on this if possible

ckipp01 (commented):
Overall I'm really happy to see this being inlined! I think small Quality of Life improvements like this go a long way! Here is a couple things that I'm not 100% sure this PR covers but I'd recommend making sure they are either covered already or should be considered. I found a lot of newcomers get bit by this and Sonaytpe just makes it confusing:

- Make sure there is an easy way to set the sontaype host uri. I've found having small helper like [this](https://github.com/ckipp01/mill-ci-release/blob/ce4bbae2b5a4fb25a6b6d696487604adae15a098/plugin/src/io/kipp/mill/ci/release/CiReleaseModule.scala#L42-L48) quite useful..
- Warn if necessary things for Sonatype aren't configured to avoid confusion. Sontaype will just fail without clear messaging as to why. So I've found including checks for things like license and developers which are necessary really helpful. [Here](https://github.com/ckipp01/mill-ci-release/blob/ce4bbae2b5a4fb25a6b6d696487604adae15a098/plugin/src/io/kipp/mill/ci/release/CiReleaseModule.scala#L99-L106) is how I did it. This would be nice to include.

lefou (commented):
I never used `mill-ci-release` but I think having better and sensible defaults is nice.

If this is a directly code in-sourcing, please address the copyright, since `mill-ci-release` has a compatible but different license.

lihaoyi:
Thanks @ckipp01! Will take that into account. @lefou this isn't a literal in-sourcing, most of the config defaults are taken from Mill's own build

lefou:
Checking validation rules should not be hardcoded into `publish` IMHO. Each remote reporsitory can implement different rules, local publishing typically has non, so the best would be to have some new convenient `check*` target with some preset for Sonatype OSS Repositoy Hosting. We could create a predefined map from repo URL to Ruleset to configure which checks we should run.

PR #3525, opened 2024-09-12
Include `T.workspace` in the docs

lihaoyi:


lefou (approved):
Other cases where I needed `T.workspace`: determine workspace-relative paths for sources to produce portable results, e.g. I18n message catalogs, test reports, ...

Issue #3144, opened 2024-05-06
Support alternative/configurable `out` directory


lefou:
Use cases:
* Build read-only projects
* Preserve multple out-directories when switching branches
* Use a different file system for `out`

PR #3530, opened 2024-09-12
Allow users to use a custom output dir via an env var

alexarchambault:
This allows users to change the Mill output directory (by default, `out` under the project root) to a directory of their choice, via the `MILL_OUTPUT_DIR` environment variable.

Fixes https://github.com/com-lihaoyi/mill/issues/3144

### DIFF
--- a/docs/modules/ROOT/pages/Out_Dir.adoc
+++ b/docs/modules/ROOT/pages/Out_Dir.adoc
@@ -111,3 +111,7 @@ This is very useful if Mill is being unexpectedly slow, and you want to find out
 
 `mill-server/*`::
  Each Mill server instance needs to keep some temporary files in one of these directories. Deleting it will also terminate the associated server instance, if it is still running.
+
+== Using another location than the `out/` directory
+
+include::example/depth/out-dir/1-custom-out.adoc[]
--- a/example/depth/out-dir/1-custom-out/build.mill
+++ b/example/depth/out-dir/1-custom-out/build.mill
@@ -1,0 +1,28 @@
+// The default location for Mill's output directory is `out/` under the project workspace.
+// A task `printDest` of a module `foo` will have a default scratch space folder
+// `out/foo/printDest.dest/`:
+package build
+
+import mill._
+
+object foo extends Module {
+  def printDest = Task {
+    println(T.dest)
+  }
+}
+
+/** Usage
+> ./mill foo.printDest
+...
+.../out/foo/printDest.dest
+*/
+
+// If you'd rather use another location than `out/`, that lives
+// in a faster or a writable filesystem for example, you can change the output directory
+// via the `MILL_OUTPUT_DIR` environment variable.
+
+/** Usage
+> MILL_OUTPUT_DIR=build-stuff/working-dir ./mill foo.printDest
+...
+.../build-stuff/working-dir/foo/printDest.dest
+*/
--- a/example/package.mill
+++ b/example/package.mill
@@ -59,6 +59,7 @@ object `package` extends RootModule with Module {
     object modules extends Cross[ExampleCrossModule](build.listIn(millSourcePath / "modules"))
     object cross extends Cross[ExampleCrossModule](build.listIn(millSourcePath / "cross"))
     object large extends Cross[ExampleCrossModule](build.listIn(millSourcePath / "large"))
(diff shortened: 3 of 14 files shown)


alexarchambault:
Fixes https://github.com/com-lihaoyi/mill/issues/3144

lihaoyi:
How about we use an env var rather than a flag? That way we dont get to parse it and can keep the server files in the same place (relatively) as they are today. I'd rather not split them into a separate folder if it can be avoided

lihaoyi:
env var isnt perfect either from a UX perspective, but given this is a bit of a niche case I think its a better solution than reorganizing the file layout for this feature due to implementation concerns. 

lets go with an envvar for now and at some point we may have a graal native or scalanative client and can have it be smart enough to use flags

lefou:
As for the cli args, if at all, I'd suggest to just go with `--output`. It's a bit too early to settle on a short option. Short options are scarce resources. Otherwise, I think using a env var is ok for the start, too.

Here are some more motivations for alternative out dirs:

* read-only project dirs
* project file systems unsuitable to keep run-state of a process
* `out` dirs in RAM filesystems

As a consequence, I think we should be able to split the two uses of `out`, Mill target cache vs Mill worker state. I'd really like to be able to keep all Mill worker state in one single location. e.g. something in `$XDG_RUNTIME_DIR`. Most code should be already prepared for this, since we made it independent of `os.pwd`.

lihaoyi:
Let's go with the env var for now and keep the `out/` folder intact, at least as part of this ticket. I think that "all generated files in the `out/` folder" is a very nice property to have, even though I had different ideas initially (e.g. our integration test fixtures used to separate them build-generated-files and misc-housekeeping-files), and even though today it's not strictly true (we dump some stuff in `~/.cache` as well).

I do agree that generating an `./out` folder in every dir that Mill starts in isn't great, and we may still want to break it up later, but we don't need to start that right now. e.g. I could imagine that with graal/scala-native allowing a thicker client, and some kind of overhaul of the evaluator command system, the problem may solve itself in a nicer way. But that will take more time to research and experiment

alexarchambault:
I went for the `MILL_OUTPUT` environment variable for now

alexarchambault:
I'd go for something like `MILL_OUTPUT`, `MILL_OUTPUT_DIRECTORY`, or `MILL_OUTPUT_FOLDER`. Env var ending in `PATH` evoke Unix's `PATH`, `LD_LIBRARY_PATH`, or Java's `CLASSPATH`, which all accept a `:` (or `;`) separated list of directories, which we don't here.

lefou:
Then my favorite is `MILL_OUTPUT_DIR`, `DIRECTORY` is too long and `FOLDER` is a Windows term.

lihaoyi:
I think this looks great. As a last thing before landing, let's convert the integration test into an example test in `example/depth/` (not sure what subfolder, maybe a new one?) and include it as part of a dedicated section on the `Out_Dir.adoc` page of our doc-site with a write-up.

You may need to hack `ExampleTester` to allow stuff like `MILL_OUTPUT_DIR=foo ./mill blah` to work properly, but that's probably fine, and it would ensure that the documentation and tests stay in sync and that the feature is discoverable to our users

alexarchambault:
> As a last thing before landing, let's convert the integration test into an example test in `example/depth/` (not sure what subfolder, maybe a new one?) and include it as part of a dedicated section on the `Out_Dir.adoc` page of our doc-site with a write-up.

Just added an example under `example/depth/out-dir/1-custom-out`, and added it to `Out_Dir.adoc`. I kept the integration tests, as more things can be tested from them (out dir outside of the workspace for example), and their checks are more precise than the grepping done in the examples.

alexarchambault:
> You may need to hack `ExampleTester` to allow stuff like `MILL_OUTPUT_DIR=foo ./mill blah` to work properly, but that's probably fine, and it would ensure that the documentation and tests stay in sync and that the feature is discoverable to our users

That ended up not being needed, as these commands are run through `bash -c`, and specifying env vars this way is valid bash.

lihaoyi (approved):
Looks great, @alexarchambault just need to update the PR description to reflect the current state of the PR and we can merge it

alexarchambault:
(Last push is just a rebase on the latest `main`)

PR #3531, opened 2024-09-12
Added plugin for formatting Java sources using Palantir

### LINKED ISSUE #3448: Support Palantir-Java-Format for Java (500USD Bounty)
---

---

Palantir-Java-Format is a common Java autoformatter, and we should support it as a `contrib/` module to better support Java projects. We should support all common configuration options to allow usage on `JavaModule`s and `ScalaModule`s (which can contain Java sources). It should broadly work the same way ScalaFmt works for Scala (https://mill-build.org/mill/0.11.12/Scala_Module_Config.html#_reformatting_your_code) This needs to come with a reasonable set of unit tests, docs, and at least one example test that can be included in the documentation.

May be able to adapt something from https://github.com/sbt/sbt-java-formatter

ajaychandran:
Added `contrib.palantir.JavafmtModule` for formatting Java sources using [Palantir](https://github.com/palantir/palantir-java-format/).
The plugin also supports command line arguments for
- checking for formatting errors with `--check` flag
- formatting specific files or folders

Resolves #3448.


ajaychandran:
@lihaoyi 
Is the naming convention fine?
Are there any additional features that must be implemented?

contrib/palantir/src/mill/contrib/palantir/JavafmtModule.scala
```diff
@@ -0,0 +1,49 @@
+package mill
+package contrib.palantir
+
+import mill.util.Jvm
+
+trait JavafmtModule extends PalantirModule {
```
lihaoyi, line 6: Let's fold this and `PalantirModule` into a combined `JavaFormatModule`, since there isn't a term `javafmt` in common usage and just calling it `Palantir` is a bit too unspecific

contrib/palantir/src/mill/contrib/palantir/JavafmtModule.scala
```diff
@@ -0,0 +1,49 @@
...
+    T.log.info("formatting java sources ...")
+
+    Jvm.runSubprocess(
+      mainClass = PalantirModule.mainClass,
+      classPath = palantirClasspath().map(_.path),
+      jvmArgs = PalantirModule.jvmArgs,
+      mainArgs = PalantirModule.mainArgs(sources().map(_.path), palantirOptions()),
+      workingDir = T.dest
+    )
+  }
```
lihaoyi, line 48: This should be moved into the `JavaFormatModule` companion object and made to work as an external module task that can take a selector of `__.sources` to format. See how `object ScalafmtModule` does it

contrib/package.mill
```diff
@@ -240,4 +240,9 @@ object `package` extends RootModule {
     def compileModuleDeps = Seq(build.scalalib)
     def testModuleDeps = super.testModuleDeps ++ Seq(build.scalalib)
   }
+
+  object palantir extends ContribModule {
```
lihaoyi, line 244: Let's call it `palantirfmt`

lihaoyi:
Looks good overall, just need a few tweaks. Need a short `readme.adoc` in the contrib module folder, and an example test in `example/javalib/linting/` next to checkstyle and errorprone to demonstrate usage.

contrib/palantir/src/mill/contrib/palantir/JavafmtModule.scala
ajaychandran: IIUC, the results of the task will be cached (subsequent formats are triggered only if upstream settings change). So this would be useful at the module level.
Maybe rename it to `javafmtCached`?

I will add a `formatAll` task for the extenal module.

lihaoyi: In this case we do not want formats to be cached targets, because cached targets are meant to be "pure", whereas these format tasks mutate the source code. So better to leave them as commands rather than targets

ajaychandran:
One improvement would be to accept Palantir Java Format options (leftover arguments that start with `-`) from the command line.
This would expose the full power of the [CLI](https://github.com/palantir/palantir-java-format/blob/16e9ff6fdfef392e0f70089acf879bb468d7efd3/palantir-java-format/src/main/java/com/palantir/javaformat/java/CommandLineOptionsParser.java#L64) (even format a range within a source file).
```sh
# need to figure out the syntax for range formatting

# command to only fix imports (sort, remove unused)
mill __.javafmt --fix-imports-only src/a/B.java
```
I could not find any documentation for the options. Should examples be included in the plugin `readme.adoc`?

@lihaoyi WDYT?

example/javalib/linting/3-palantirjavaformat/src/A.java
```diff
@@ -0,0 +1,84 @@
+package com.google.googlejavaformat.java.test;
```
lihaoyi, line 1: For the example, let's make it small (e.g. ~5 lines), and include it in the example page via a `/** See Also: src/A.java */` comment. That would help make the example more complete and understandable to readers, while the unit tests cover the large input scenarios for us.

example/javalib/linting/3-palantirjavaformat/build.mill
```diff
@@ -0,0 +1,23 @@
...
+> ./mill javafmt --check                # check Java source files for formatting errors
+...: checking format in java sources ...
+.../src/A.java
+...: palantirjavaformat aborted due to format error(s) (or invalid plugin settings/palantirjavaformat options)
+
+> ./mill javafmt src/A.java             # format Java source file
+
+> ./mill javafmt                        # alternatively, format all Java source files
+
+*/
```
lihaoyi, line 21: We should also show off the `./mill mill.contrib.palantirjavaformat.PalantirJavaFormat/` workflow, and after formatting is run we should run `check` again and ensure it does not error

contrib/palantirjavaformat/src/mill/contrib/palantirjavaformat/PalantirJavaFormatModule.scala
```diff
@@ -0,0 +1,187 @@
...
+  )
+
+  /**
+   * Palantir Java Format version. Defaults to `2.50.0`.
+   */
+  def palantirjavaformatVersion: T[String] = T {
+    "2.50.0"
+  }
+}
+object PalantirJavaFormatModule extends ExternalModule with PalantirJavaFormatModule {
```
lihaoyi, line 86: Does this need to inherit from the full `PalantirJavaFormatModule`? `PalantirJavaFormatModule` pulls in `JavaModule`, and I don't think this external module needs to be able to compile java sources and package it etc... Maybe we can extract a more lightweight `trait` that contains the shared stuff for both `trait PalantirJavaFormatModule` and `object PalantirJavaFormatModule`, whereas only `trait PalantirJavaFormatModule` inherits from `JavaModule`

Also, we should define a `def defaultCommandName = "formatAll"`, so people can call this via `mill.contrib.palantirjavaformat.PalantirJavaFormatModule/` without needing to pass in the `formatAll` suffix explicitly

contrib/palantirjavaformat/src/mill/contrib/palantirjavaformat/PalantirJavaFormatModule.scala
```diff
@@ -0,0 +1,187 @@
+package mill
+package contrib.palantirjavaformat
```
lihaoyi, line 2: I edited a comment earlier to say this, but let's standardize the naming convention of `palantirformat`/`PalantirFormat`. I know it doesn't quite match the `palantir-java-format` project name, but using the full name over and over is getting verbose enough I think we should shorten it here just for ease of use

contrib/palantirjavaformat/src/mill/contrib/palantirjavaformat/PalantirJavaFormatModule.scala
```diff
@@ -0,0 +1,187 @@
...
+  /**
+   * Formats Java source files.
+   *
+   * @param check if an exception should be raised when formatting errors are found
+   *              - when `true`, files are not formatted
+   * @param sources list of file or folder path(s) to be processed
+   *                - path must be relative to [[millSourcePath]]
+   *                - when empty, all [[sources]] are processed
+   */
+  def javafmt(
```
lihaoyi, line 26: For consistency with the rest of Mill plugin traits, we should name this `def palantirformat`, so it matches the name of the trait that's providing it (unless it's a "shared" task that many different traits would use)

lihaoyi:
Left some comments

The tests are failing because we need to wire up `contrib.palantirformat` into our `dist` so it will get properly published during integration testing. Should just involve adding it to the list here along with the other integration-tested contrib modules https://github.com/com-lihaoyi/mill/blob/2dc343fac9da8234d66ec522b68f03095f3e5d4b/build.mill#L727

example/javalib/linting/3-palantirjavaformat/build.mill
```diff
@@ -0,0 +1,23 @@
...
+
+object `package` extends RootModule with PalantirJavaFormatModule {
+}
+
+/** Usage
+
+> ./mill javafmt --check                # check Java source files for formatting errors
+...: checking format in java sources ...
+.../src/A.java
+...: palantirjavaformat aborted due to format error(s) (or invalid plugin settings/palantirjavaformat options)
```
lihaoyi, line 15: If this returns an exit code != 0, you need to prefix the first line of output with `error: ` to tell the example testing framework to expect the failure

contrib/palantirjavaformat/src/mill/contrib/palantirjavaformat/PalantirJavaFormatModule.scala
ajaychandran: @lihaoyi 
How about `palantirjavafmt`? Get a lot more clarity with the addition of a single character.

lihaoyi: Let's go with `palantirformat`, googling `palantir format` brings up only this one project, so there shouldn't be any ambiguity for now

lihaoyi: One consideration is that I may want to move this directly into `javalib`, just like how `ScalaFmtModule` is in `scalalib` and `KLintModule` is in `kotlinlib`. That way it'll be `mill.javalib.palantirformat` which should be unambiguous enough

ajaychandran: Where would the documentation go in this case? `docs/modules/ROOT/pages/Java_Module_Config.adoc`?

example/javalib/linting/3-palantirjavaformat/src/A.java
ajaychandran: Updated. Please review.

example/javalib/linting/3-palantirjavaformat/build.mill
ajaychandran: Updated. Please review.

ajaychandran: Had to remove `check` examples.
```sh
 Incomplete argument --check <bool> is missing a corresponding value
  Expected Signature: palantirformat
    --check <bool>
    sources <str>...
  

  ,)
```

example/javalib/linting/3-palantirformat/build.mill
```diff
@@ -0,0 +1,18 @@
...
+}
+
+/** Usage
+
+> ./mill palantirformat                                                 # format all Java source files
+...formatting java sources ...
+
+> ./mill mill.javalib.palantirformat.PalantirFormatModule/ __.sources   # format all Java source files in existing modules
+...formatting java sources ...
+*/
```
lihaoyi, line 16: We should show `--check` failing before calling `palantirformat`, and then passing after

example/javalib/linting/3-palantirformat/build.mill
```diff
@@ -0,0 +1,18 @@
...
+
+/** See Also: src/A.java */
```
lihaoyi, line 18: Let's move this above the `Usage` block, since it shows the raw source code before the usage commands are run

lihaoyi:
I think this looks great, two small nits in the example and I think we can merge this and close out the bounty

example/javalib/linting/3-palantirformat/build.mill
ajaychandran: Using `--check` results in a failure.
```sh
./mill example.javalib.linting[3-palantirformat].local
```
```
 Incomplete argument --check <bool> is missing a corresponding value
  Expected Signature: palantirformat
    --check <bool>
    sources <str>...
  

  ,)
```

lihaoyi: You need to make it `check: mainargs.Flag` for it to not require a boolean after

lihaoyi:
Thanks @ajaychandran! will transfer the bounty using the same details as earlier

Issue #2929, opened 2023-12-21
`inspect` should show some useful info about modules (500USD Bounty)


lefou:
---

**From the maintainer Li Haoyi**: I'm putting a 500USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

Currently, `mill inspect` only works for targets, but not for modules. For module, it just outputs an error message "inspect Cannot find default task to evaluate for module foo".

Instead, it should provide some useful information:

* Scaladoc, source location (filename:linenumber)
* the inherited / mixed-in modules
* the available tasks
* an optionally defined default task
* potentially known sub-modules or cross instances

Shri333:
@lefou can I take a stab at this? are there any expected challenges/hurdles, or should the task be relatively straightforward?

lihaoyi:
@Shri333 go for it. It should be relatively straightforward. It will probably involve changes to the mill-moduledefs plugin to save the scaladoc for `object`s and `lazy val`s to Java annotations, and then updates to `inspect` so when it sees a module it uses Java reflect to fetch the corresponding info to display

Shri333:
Does mill not work with JDK 21? I had issues with `./mill installLocal` and had to downgrade to JDK 17 for it to finish successfully.

Shri333:
I am trying to build mill's source (main). It failed with JDK 21 but worked with JDK 17 (`./mill installLocal`). I am also not getting any syntax highlighting for .mill files and BSP appears to be inactive in IntelliJ (I built idea files with `./mill -j 0 mill.idea.GenIdea/idea`). I also tested with VSCode + Metals and syntax highlighting is inactive there as well. Is there anything I have to do to get the syntax highlighting and build configurations to appear (for debugging)?

lihaoyi:
@Shri333 for IDE support you can set up a Scala <-> `*.mill` file association via `Settings / Editor / File Types / Scala` in Intellij. That should let IntelliJ's BSP support work and do the right thing. I'm working on getting it set up by default in IntelliJ but for now you can configure it manually.

Mill is currently only built and tested using Java 17, so not surprising it doesn't build cleanly on newer versions. We'll get to upgrading our CI at some point

Shri333:
Is there a particular debug configuration that you use in IDEA?

lihaoyi:
@Shri333 typically I run tests from the terminal and use println debugging. I'm not actually that familiar with usage of the IntelliJ debugger haha

[...]

PR #3532, opened 2024-09-12
Issue #2929: implement inspect for modules

Shri333:
Implement inspect for modules:
- Name of module with file/line info
- Scaladoc on module
- Inherited modules
- Module dependencies
- Default task
- Implemented tasks

### DIFF
--- a/integration/feature/docannotations/resources/build.mill
+++ b/integration/feature/docannotations/resources/build.mill
@@ -20,6 +20,10 @@ trait JUnitTests extends TestModule.Junit4 {
   }
 }
 
+trait foo
+
+object basic extends Module with foo
+
 /**
  * The Core Module Docz!
  */
@@ -36,6 +40,20 @@ object core extends JavaModule {
   /**
    * Core Target Docz!
    */
+  def target = Task {
+    import collection.JavaConverters._
+    println(this.getClass.getClassLoader.getResources("scalac-plugin.xml").asScala.toList)
+    "Hello!"
+  }
+}
+
+object core2 extends JavaModule
+
+object MyJavaTaskModule extends JavaModule {
+  override def moduleDeps: Seq[JavaModule] = Seq(core, core2)
+  def lineCount = T {
+    allSourceFiles().map(f => os.read.lines(f.path).size).sum
+  }
   def target = Task {
     import collection.JavaConverters._
     println(this.getClass.getClassLoader.getResources("scalac-plugin.xml").asScala.toList)
--- a/integration/feature/docannotations/src/DocAnnotationsTests.scala
+++ b/integration/feature/docannotations/src/DocAnnotationsTests.scala
@@ -1,7 +1,6 @@
 package mill.integration
 
 import mill.testkit.UtestIntegrationTestSuite
-
 import utest._
 
 object DocAnnotationsTests extends UtestIntegrationTestSuite {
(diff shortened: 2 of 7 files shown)


lihaoyi:
@Shri333 let's move the discussion for this PR here rather than the original issue

I played around a bit myself and it does seem problematic to get the Scala declared methods via java reflection. Rather than using scala reflection at runtime, we can instead use scala reflection in the `Discover.scala` macro where we already do this kind of thing. 

`Discover.scala` currently looks for every class that is part of our module hierarchy, and generates a tuple of two lists: 

1. A list of mainargs entrypoints, used to support `Task.Command`s, which we need statically generated so we can perform typeclass resolution on the parameter names
2. A list of all method names, inherited or not, used to perform "did you mean foo?" error reporting

It should be possible to extend that tuple to include a third list for each class: a list of method names that are directly declared by this particular `object`. You would then be able to use that information in `inspect` to return the list of directly declared methods.

This approach is a bit roundabout compared to using scala-reflect at runtime, but it should work, and help us preserve the "no scala reflect at runtime" invariant we've maintained so far

Shri333:
> @Shri333 let's move the discussion for this PR here rather than the original issue
> 
> I played around a bit myself and it does seem problematic to get the Scala declared methods via java reflection. Rather than using scala reflection at runtime, we can instead use scala reflection in the `Discover.scala` macro where we already do this kind of thing.
> 
> `Discover.scala` currently looks for every class that is part of our module hierarchy, and generates a tuple of two lists:
> 
>     1. A list of mainargs entrypoints, used to support `Task.Command`s, which we need statically generated so we can perform typeclass resolution on the parameter names
> 
>     2. A list of all method names, inherited or not, used to perform "did you mean foo?" error reporting
> 
> 
> It should be possible to extend that tuple to include a third list for each class: a list of method names that are directly declared by this particular `object`. You would then be able to use that information in `inspect` to return the list of directly declared methods.
> 
> This approach is a bit roundabout compared to using scala-reflect at runtime, but it should work, and help us preserve the "no scala reflect at runtime" invariant we've maintained so far

@lihaoyi I was able to implement some basic changes to Discover, but for some reason, I am not able to get the value in `inspect` (it prints an empty Map when I print `discover.value`). It prints the mapping in the apply function though. https://github.com/com-lihaoyi/mill/pull/3532/commits/fb20fa98c5be0ba381033bf8e76fd921475458cd

Shri333:
@lihaoyi I pushed some changes (https://github.com/com-lihaoyi/mill/pull/3532/commits/fb20fa98c5be0ba381033bf8e76fd921475458cd) to get all declared methods in `Discover.scala` and add it to the tuple. When I print out the mapping in `Discover.scala`, it prints the correct values. However, when I access it in `MainModule.scala` with `discover.value`, the `Map` is empty. Do you know why this could be occurring?

lihaoyi:
@Shri333 when you say the `Map` is empty, are you referring to the entire `Discover#value` field? That needs to get populated by Mill for normal functionality to work, so it shouldn't be empty in the `main` branch. 


You can try cherry-picking your minimal changes onto `main` to see whether it's still empty there. If it is empty after the cherry pick, it means your minimal changes are breaking something, so you'll need to bisect your minimal changes further to find out who is breaking it. If it's not empty after the cherry pick, then you'll need to bisect your other changes to figure out what is causing it to be empty

Shri333:
> @Shri333 when you say the `Map` is empty, are you referring to the entire `Discover#value` field? That needs to get populated by Mill for normal functionality to work, so it shouldn't be empty in the `main` branch.
> 
> You can try cherry-picking your minimal changes onto `main` to see whether it's still empty there. If it is empty after the cherry pick, it means your minimal changes are breaking something, so you'll need to bisect your minimal changes further to find out who is breaking it. If it's not empty after the cherry pick, then you'll need to bisect your other changes to figure out what is causing it to be empty

@lihaoyi I was able to fix the issue. Turns out I was supposed to use `evaluator.rootModule.millDiscover` rather than `Discover[t.module.type]`.

Shri333:
Also, I noticed that I can't print out information during tests that fail. Is there a way to view stdout/stderr for tests that fail? I am running `./mill main.resolve.test`

Shri333:
@lihaoyi if I am not mistaken, resolve itself seems to be working incorrectly. If I have the following `build.mill`:
```scala
package build
import mill._, javalib._

trait TypeA extends Module {
  def foo = Task { "foo" }
}
trait TypeB extends Module {
  def bar = Task { "bar" }
}
trait TypeC extends Module {
  def baz = Task { "baz" }
}
trait TypeAB extends TypeA with TypeB

object typeA extends TypeA
object typeB extends TypeB
object typeC extends TypeC {
  object typeA extends TypeA
}
object typeAB extends TypeAB
```
Running `mill resolve _:TypeA._` produces the following output:
```
typeA.foo
typeAB.bar
typeAB.foo
```
But, running `mill resolve __:TypeA._` produces the following output:
```
clean
init
inspect
path
plan
resolve
show
showNamed
shutdown
typeA
typeA.foo
typeAB
typeAB.bar
typeAB.foo
typeB
typeC
typeC.typeA.foo
version
visualize
visualizePlan
```
Is this expected? This was also the weird behavior I was seeing in the resolve tests. I know that `__` matches any # of segments while `_` matches one segment, but I was confused about this output.

lihaoyi:
Thanks @Shri333 ! I'm traveling now but I will try to find time to review this PR as soon as possible

lihaoyi:
Went through the code and tried it out locally. This looks great!

Screenshot 2024-09-28 at 12 25 08 PM

Nice work @Shri333, and thanks for bearing with me and with the Mill codebase the last 2-3 weeks! Email me your bank transfer details and I will close out the bounty

PR #3533, opened 2024-09-13
Break out publishing docs into its own documentation page

lihaoyi:
This should make it a lot more discoverable, provide a place to document things more thoroughly, and link to other non-sonatype publishing workflows e.g. bintray or artifactory. 

I tried to flesh it out a bit with more instructions and pointers at sonatype's GPG docs, handy CLI commands, and some github actions boilerplate. Can probably be further improved, but it's a start

CC @lefou


docs/modules/ROOT/partials/Publishing_Footer.adoc
```diff
@@ -0,0 +1,150 @@
...
+
+# The passphrase associated with your PGP key
+export MILL_PGP_PASSPHRASE=...
+----
+
+=== Publishing
+
+You can publish all eligible modules in your Mill project as follows:
+
+[source,scala]
```
lefou, line 49:
```suggestion
[source,bash]
```

Issue #3541, opened 2024-09-13
Discord Link not inviting / going to a Mill channel.


lefou:
### Discussed in https://github.com/com-lihaoyi/mill/discussions/3507

<div type='discussions-op-text'>

<sup>Originally posted by **gamlerhart** September 10, 2024</sup>
In the Docs the chat link was updated to a Discord link. However, that link opens up Discord for me and doesn't add me / invite me to any channel. Is it the correct link to  onboard people.

The link is: https://discord.com/channels/632150470000902164/940067748103487558</div>

PR #3542, opened 2024-09-13
Fix discord link to be inviting 

lefou:
This replaces a direct link to the discord channel with an invite to that channel. The invite does not time out. It has the advantage to easily let new users to register to discord.

Fix https://github.com/com-lihaoyi/mill/issues/3541


lihaoyi (approved):

GeorgOfenbeck:
@lefou The link seems to be (again?) broken.
image

lefou:
I explicitly created an invite that "never expires".

Besides, these issues make perfectly clear, why closed platforms where you need to register and login to _read_, are no good infrastructure for Open Source communities.

GeorgOfenbeck:
🤷‍♂️just happened to look over the recent changes and clicked on the link, hence i thought I should report it. Happy to verify if it again later - dont think i can help otherwise.

lefou:
Thanks for reporting, @GeorgOfenbeck. We need to figure out how to create invites that last longer. Updating the link manually every two weeks isn't sustainable.

Issue #3482, opened 2024-09-07
Allow fractional jobs to automatically scale by number of CPUs (200USD Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 200USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

so `-j 0.5` on a 10 CPU machine would be the same as `-j 5`, on a 4 CPU machine the same as `-j 2`, etc. `-j 1.0` would be the current default, and `-j 2.0` would mean twice as many jobs as CPUs (e.g. if they're mostly IO bound).

I've wanted `-j 0.5` myself to use in Mill's own CI, because Mill tests are very heavy and use more than one thread per task, and it would be nice if the same `-j` config would just do the right thing on different machines with different amounts of available compute

lefou:
Nice idea.

wb14123:
Shall we use another flag for this? It's confusing that `-j 1.0` is not equal to `-j 1`.

lihaoyi:
how about `-j 0.5n` or `-j 0.5*HOST_CPUS` (what baze uses)?

lihaoyi:
Bazel supports basic arithmetic and allows things like `HOST_CPUS-1` as well

wb14123:
Yeah I think we can take what Brazel uses, specifically in this [doc](https://bazel.build/docs/user-manual#local-resources):

> --local_{ram,cpu}_resources resources or resource expression
These options specify the amount of local resources (RAM in MB and number of CPU logical cores) that Bazel can take into consideration when scheduling build and test activities to run locally. They take an integer, or a keyword (HOST_RAM or HOST_CPUS) optionally followed by [-|*float] (for example, --local_cpu_resources=2, --local_ram_resources=HOST_RAM*.5, --local_cpu_resources=HOST_CPUS-1). The flags are independent; one or both may be set. By default, Bazel estimates the amount of RAM and number of CPU cores directly from the local system's configuration.

It only allows keyword at the beginning and only supports `*` and `-`, which makes parsing not too hard.

lihaoyi:
Yeah don't even need a real parser, pattern matching on `case s"HOST_CPUS*$n" =>` would be enough

lefou:
In Maven it is `-T n` for using `n` threads and `-T nC` for using `n` threads per CPU core, which I find nice. 

On a 16 core cpu, these three command would all mean the same:

```
-n 0
-n 16
-n 1C
```

[...]

PR #3554, opened 2024-09-15
Add fractional and subtraction for jobs flag

wb14123:
Resolve https://github.com/com-lihaoyi/mill/issues/3482

Allow jobs param to have following formats:

1. An integer indicates the parallel level.
2. A float N followed by character "C" indicates uses (N * all available cores). e.g. "0.5C" uses half of the available cores.
3. "C-" followed by an integer N indicates uses (all available cores - N). e.g. "C-1" leaves 1 core and uses all the other cores.


runner/src/mill/runner/MillMain.scala
```diff
@@ -275,6 +275,28 @@ object MillMain {
...
+          .getOrElse(Failure(err("Failed to find a int number after \"C-\".")))
+          .map(cores - _)
+      case Some(n) => n.toIntOption.map(Success(_))
+          .getOrElse(Failure(err("Failed to find a int number")))
+    }).flatMap {
+      case x if x < 0 => Failure(err("Calculated cores to use is a negative number."))
+      case 0 => Success(cores) // use all cores if 0
+      case x => Success(x)
+    }
+  }
```
lihaoyi, line 298: Few nits here:

1. Can we extract this into a helper function that we can unit test with various combinations of `detail: String` and `cores: Int`? That will give us some confidence at-a-glance that the code is doing what we expect without tedious manual testing

2. I don't think the current treatment of "0 = use all cores" makes sense for computed zeros. If someone runs `-j0.4C` on a 4 core machine, they probably don't intend to use all 4 cores. Let's treat literal 0 separately from computed numbers, and have computed numbers have a lower limit of 1 core (so `0` or `-1` becomes `1`)

3. Lets return an `Either[String, Int]` since that's all the information you need later on at the callsite

wb14123:
@lihaoyi addressed comments about the code. Where should I put the unit tests?

lihaoyi:
runner/test/src/mill/runner/ should work i think, run via mill runner.test

wb14123:
@lihaoyi Added unit tests. Seems the auto lint check failed on previous commit again. Is there any command that I can auto format code or show lint errors?

lihaoyi:
@wb14123 `./mill -i mill.scalalib.scalafmt.ScalafmtModule/reformatAll __.sources` does the formatting

wb14123:
Hmm I formatted the files. Not sure why it failed again. Does the command above also format the test files?

lihaoyi:
@wb14123 it should. Could you try merging in latest `main` and running it again? Maybe there's some formatting inconsistencies introduced by the merge

runner/test/src/mill/runner/MillMainTests.scala
```diff
@@ -0,0 +1,95 @@
...
+        assert(MillMain.parseThreadCount(Some("0.54C"), 10) == Right(5))
+        assert(MillMain.parseThreadCount(Some("0.59C"), 10) == Right(5))
+        assert(MillMain.parseThreadCount(Some(".5C"), 10) == Right(5))
+        assert(MillMain.parseThreadCount(Some("1.0C"), 10) == Right(10))
+        assert(MillMain.parseThreadCount(Some("1.5C"), 10) == Right(15))
+
+        assertParseErr(
+          MillMain.parseThreadCount(Some("0.09C"), 10),
+          "Calculated cores to use should be a positive number"
+        )
```
lihaoyi, line 48: For these computed cases (this one and `C-10` and `C-11` below), let's always just round up to `1` core minimum. That's probably what users want, so no point erroring out when we can just do it for them

lihaoyi:
Left one more comment, after that I think we're good to merge

PR #3565, opened 2024-09-17
Use `Task` instead of `T` when constructing targets and other tasks

lihaoyi:
Essentially a minimized binary-compatible version of https://github.com/com-lihaoyi/mill/pull/3356, renaming all the factory methods for various tasks to follow more standard naming conventions:

* `T {...}` ->  `Task {...}`
* `T.command {...}` ->  `Task.Command {...}`
* `T.input {...}` ->  `Task.Input {...}`
* `T.source {...}` ->  `Task.Source {...}`
* `T.sources {...}` ->  `Task.Sources {...}`
* `T.persistent {...}` ->  `Task.Persistent {...}`
* `T.task {...}` ->  `Task.Anon {...}`

The type `T[_]` remains an alias for `Target[_]`, and `Task{ ... }` returns a `T[_]`, to maintain binary compatibility. Not quite ideal but can probably be hand-waved away until Mill 0.13.0 when we are allowed to break binary compatibility. 

All the `T.*` operations have been duplicated to `Task.*` by sharing them via a `trait TargetBase`, except the factory methods which were copied over and upper-cased while the old version deprecated. I have updated all the code and examples to use `Task` instead of `T` where relevant. The only exceptions are the `implicit def apply`s which needed to be manually copied without the `implicit` (otherwise the multiple implicits cause ambiguity).

This gets us most of the user-facing benefits of https://github.com/com-lihaoyi/mill/pull/3356 without the bin-compat breakage: users no longer see an odd `T { ... }` syntax in the docs and in their build files, and now see `Task { ... }` which should be much more familiar. Although it does not allow us to do the type-hierarchy cleanups that the other PR provides, it's still worth doing so we can get it in in 0.12.0

The old `T { ... }` and `T.*` syntaxes should continue to work, and are exercised via the bootstrap tests as they continue to be used in Mill's own build. This PR should be source compatible to avoid migration pains, and given the prevalence of `T` everywhere we probably should just support it forever

### DIFF
--- a/bsp/src/mill/bsp/BSP.scala
+++ b/bsp/src/mill/bsp/BSP.scala
@@ -30,7 +30,7 @@ object BSP extends ExternalModule with CoursierModule {
    * reason, the message and stacktrace of the exception will be
    * printed to stdout.
    */
-  def install(jobs: Int = 1): Command[(PathRef, ujson.Value)] = T.command {
+  def install(jobs: Int = 1): Command[(PathRef, ujson.Value)] = Task.Command {
     // we create a file containing the additional jars to load
     val libUrls = bspWorkerLibs().map(_.path.toNIO.toUri.toURL).iterator.toSeq
     val cpFile =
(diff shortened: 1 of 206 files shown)


lihaoyi:
@lefou I also went and made the factory method names uppercase, e.g. `Task.Command` or `Task.Sources`, as discussed in https://github.com/com-lihaoyi/mill/issues/3338, leaving the old method names present but deprecated.

I didn't change the type names, because a bunch of types are already just thin aliases to `Target` already (`Input`, `Persistent`, `Source`, `Sources`), and I expect that once https://github.com/com-lihaoyi/mill/pull/3356 lands to flatten out the rest they'll basically all be aliases to `Task` and the issue of task type name v.s. task factory name will go away

lefou:
Maybe, it's enough to have the lower case creator methods in Target (as-is) and the new upper case creator methods in the new place in `Task`.

lihaoyi:
@lefou yep that's what I ended up doing

PR #3575, opened 2024-09-18
Fix BSP target capabilities

AnthonyGrod:
Context: I'm working on integrating Mill over the new BSP plugin (I created a pull request with more detail a few weeks ago here: https://github.com/com-lihaoyi/mill/pull/3432)

I encountered an issue where there were to many options of actions that could be performed on a certain testing module. Namely, there was a redundant run capability. On the first screenshot there is the option menu before my changes and on the second after them.

The visual aspect of this fix is nice but not the most important one. Without this fix, the testing via BSP does not work at all - the BSP plugin, even after triggering the `Test` button, launches `Run` action instead of the `Test` action if the module's `canRun` capability is set to true.

image

image


lefou:
I'd suspect some other underlying issue. Since `TestModule`s are also `RunModule`s, it's legit to run a main class without any test framework. Therefore I doubt simply disabling the run capability is not the correct solution, maybe a workaround.

Issue #3546, opened 2024-09-15
Improve Mill command line status reporting UI to display concurrent tasks


lihaoyi:
Now that parallelism is enabled by default, the previous "ticker" mechanism for reporting progress is no longer sufficient to tell the user what the build is doing.

SBT, Gradle, and Bazel all have nice terminal UIs with a multi-line status bar showing all tasks in progress at any point in time, although Maven doesn't have one. Mill should have one as well

Gradle

```
...logs...
<===========--> 86% EXECUTING [36s]
> :kotlinTest:test > 0 tests completed
> :inlineTest:test > 54 tests completed
> :kotlinReleaseCoroutinesTest:test > Executing test org.mockito.kotlin.SuspendTest
> :osgi-test:test > 0 tests completed
> :checkstyleMain
> :programmatic-test:test > 1 test completed
> :module-test:test > 21 tests completed, 17 skipped
> :kotlinReleaseCoroutinesTest:test > 0 tests completed
> :memory-test:test > 1 test completed
> :compileTestJava
> :memory-test:test > Executing test org...ShouldNotStarveMemoryOnLargeStackTraceInvocationsTest
> :subclass:test > 0 tests completed
> :inlineTest:test > Executing test org.mockitoinline.StressTest
> :osgi-test:test > Executing test org.mockito.osgitest.OsgiTest
> :module-test:test > Executing test org.mockito.moduletest.ModuleHandlingTest
> :programmatic-test:test > Executing test org.mockito.ProgrammaticMockMakerTest
> :kotlinTest:test > Executing test org.mockito.kotlin.FunctionTest
> IDLE
> :subclass:test > Executing test org.mockitosubclass.PluginTest
```

SBT

```
...logs...

  | => gatling-http-java / update 0s
  | => gatling-http-client / Compile / compileIncremental 0s
  | => gatling-jdbc-java / update 0s
  | => gatling-redis-java / update 1s
  | => gatling-test-framework / scalaCompilerBridgeBinaryJar 1s
  | => docSamples / scalaCompilerBridgeBinaryJar 3s
  | => gatling-commons / Compile / compileIncremental 3s
  | => gatling-jsonpath / Compile / compileIncremental 3s
  | => gatling-quicklens / Compile / compileIncremental 3s
```

Issue #602, opened 2019-05-08
IntelliJ generator for Scala native lacking ill.scalanativelib dependency


mkotsbak:
Importing the project into Intellij, the following packages are not found:

```
import mill.scalanativelib._
import mill.scalanativelib.api._
```

andres-pipicello-olx:
Could you please provide a minimal working example? I would like to help but I am not versed in the Scala native world.

lefou:
Looks like the marked duplicates of this issue were closed with #616 .

PR #3577, opened 2024-09-18
Multi-line status prompt ticker

lihaoyi:
fixes https://github.com/com-lihaoyi/mill/issues/3546

This PR introduces a new `MultilinePromptLogger` that allows the status of multiple in-progress threads to be reported and viewed during a Mill evaluation:

https://github.com/user-attachments/assets/32d2a7e4-6db9-4dcc-81ed-a775e307d1c2

`MultilinePromptLogger` is basically a minimal translation of the current `ticker` API to work with a multi-line prompt. There's probably a lot of other interesting improvements we can make now that we are multi-line, but this is a decent start that lets people know what their parallel build is doing. The UI works correctly both during scrolling and not-scrolling, and uses the same minimal ANSI codes that the existing ticker uses, so hopefully we won't hit any more edge cases that we aren't already hitting today

From an implementation perspective, `MultilinePromptLogger` is mostly a drop in replacement for the old `PrintLogger`. 

User-facing notes:

1. It prints multiple `ticker` lines at vertically at the bottom of the terminal, and has the logs printed above

2. It requires that you use `.endTicker()` after every `.ticker("...")` call, so we can know where we should clear the ticker status (previously they always just got overriden by the next `.ticker` call)

3. We need to introduce a `withPaused{...}` block so that when you're running REPLs and stuff the prompt is not shown

4. We only can support logs which end with newlines. This means things like interactive progress bars and other ANSI magic won't work in the scrollback. This is a limitation of the current implementation that is hard to fix without going for more advanced techniques, but should be enough for the vast majority of logs and is a similar limitation as a lot of other tools.

5. Console dimensions are propagated from the mill client to the mill server via a `terminfo` file specific to each server. This is used to size the prompt so it takes up the whole width and about 1/3 of the height. This happens every 100ms, so there may be some delay, but at least it means the prompt will right-size itself on next render. The regular cadence also means it shouldn't be a performance bottleneck

6. For non-interactive runs which lack a terminal, prompt WxH defaults to 120x50, and we tweak the rendering: we no longer render blank lines to preserve prompt height stability, we render less frequently and only if the statuses change, and we add a footer so the bottom of the prompt is clearly marked.

7. `--ticker` now defaults to `true` even for non-interactive runs, since the periodic display of prompt (I set to 60s intervals, if prompt changes) is useful e.g. for debugging stuck background jobs or CI runs.

Implementation Notes

1. The logger needs to extend `AutoCloseable`, since it uses a background thread to keep the prompt UI up to date. This renders every 100ms (arbitrary) to try and balance between prompt readability and latency. We have additional delays built in to hiding a status line and then finally removing it, to try and preserve the height so it doesn't bounce up and down too much as the set of running tasks changes

2. We re-use the `ProxyStream` classes to combine the stderr/stdout before re-splitting them. This allows us to perform some buffering, while simultaneously maintaining ordering of writes to each stream, while also allowing us to detect quiescence so we can only bother writing out a prompt when everything else is done and it won't get immediately over-written. `ProxyStream.Pumper` needed some tweaks to add hooks and remove client-specific assumptions

3. I experimented with having the ticker logic live in the Mill client rather than server, which would make more sense, but we need the server to have the ability to enable/disable the ticker logic to run `console` and similar interactive tasks, and so it has to live in the server

The old ticker remains available at `--disable-prompt`. Further improvements can come after 0.12.0.


lolgab:
Using `ivy"org.jline:jline-terminal:3.26.3"` you can do:

```scala
org.jline.terminal.TerminalBuilder.terminal().getWidth()
```
to get the terminal width.

lefou:
This looks like a good start. Very good. 

Here are some observed nits.

When Mill ends, the prompt overdraws some lines previously used for progress reporting. 

Example (` ± ` is my prompt):

```
11:08:40 θ85° lefou:~/work/tototec/xxx/2017-xxxxxx/xxxxxx refactor-build(+701/-5)+* 2s ± dev-mill __.compile======== 1s
Mill version 0.12.0-RC2-51-8ab791 is different than configured for this directory!
Configured version is 0.12.0-RC2 (/home/lefou/work/tototec/xxx/2017-xxxxxxx/xxxxxxx/.mill-version)
11:08:43 θ81° lefou:~/work/tototec/xxx/2017-xxxxxx/xxxxxx refactor-build(+701/-5)+* 2s ± =========================== 1s
#01 module.dynamicEntity.test.sources 
#05 cloud.vaadin.frontend.iba.DEV.sources 
#02 thirdparty.springweb.resources 
#08 thirdparty.mssqljdbc.resources 
#04 app.contractexporter.test.sources 
#03 cloud.vaadin.frontend.PwC.STAGE.sources 
#07 app.contractmngApp.eoeWinDemo.sources 
#06 app.rcpBundle.standalone.test.sources 
#14 app.rcpBundle.resources 
#09 app.mono.resources.super.xxxxxxbuild.xxxxxxModule 
#13 app.dbschema.test.sources 
#10 module.kernel.resources 
#11 framework.etlApi.resources 
#12 cloud.asb.base.test.sources 
#16 cloud.asb.mockserver.test.sources 
#15 cloud.asb.base.itest.sources
```

Maybe we can remember the space used for progress and keep it free?

The lines used for progress reporting isn't stable. This seems ok, but when scrolling with lots of output, the progress is unreadable. The progress bounces up and down from the bottom, also unchanged progress lines. This is probably because not all thread always run a job, so the overall progress line count changes frequently.

We don't limit the lines, resulting in unusable output when job counts rise. e.g. when starting with `-j 5C`, the progress fills the full screen height and more.

lihaoyi:
Thanks for the pointer @lolgab ! that looks very easy and is hopefully more battle-tested than anything I could jury rig from first principles

@lefou yeah the Mill-end-overwrite-prompt thing is something I noticed as well. It's not intentional: I want to leave the title bar visible at the end since it conveniently shows how long the evaluation took, which is a common point of interest, and may show other useful information in future. Just need to debug what's going wrong

The jumping-up-and-down thing I have also noticed. I didn't want to go with fixed height prompt because when I've used Gradle, often the prompt is mostly empty with 9 blank lines and 1 actually doing stuff. I'm going to try to add some more debouncing logic to adding/removing entries from the prompt, that way it can scale up and down over the course of a build without being too distracting

lefou:
I think with a lot of worker threads we should explore in the future if we can provide a multi-column output, where we cut line count by a factor at the price of shortened output.

lefou:
Maybe we can fix the height for a given slot of time, e.g. 5 seconds. It can grow and shrink as needed, but should still feel stable enough due to the grace period.

alexarchambault:
Code such as [this one](https://github.com/com-lihaoyi/Ammonite/blob/08c3afb3816a9c40ab96d3b8b8129e368d07d739/terminal/src/main/scala/ammonite/terminal/ConsoleDim.scala#L23) allows to be kept updated when the terminal dimensions change, as the jline call to get them can be costly (so you don't want to call it on every update of the prompt)

lihaoyi:
One issue I hit with `org.jline.terminal.TerminalBuilder.terminal` is that it only works in `-i` mode where we directly forward the client's stdin/out/err to the background process. In server mode, JLine is not able to properly determine terminal dimensions on the server. We may need to move that logic to the client and have some protocol for sending the dimensions to the server

lihaoyi:
@lefou @lolgab @alexarchambault This is probably in a good enough state to take another pass at review.

lefou:
Great. I first did some builds with the new ticker and it behaves nicely. So nice, that I started to think about minor cosmetic things like e.g. an empty line above the progress, so it gets easier to read the "real" output. It seems pretty robust. In once run out of ten, my shell prompt was overridden (under-ridden) with ticker output, but in most of the other runs it was clear. `^C` leaves the prompt in the ticker output too, but that is probably expected.

API wise, I think the whole ticker thing should be something internal to the Evaluator and should not be exposed in the `Logger` API. That means random tasks can't add ticker messages which sound ok to me. Since we already have the `Logger.ticker(String)` API, which we need to keep for compatibility anyways, we could re-purpose to provide some "ticker suffix", e.g. something to add the the ticker line of the current task in case it is printed.

lefou:
And here the bikeshedding continues. I always have some difficulties with reading the tasks output for a parallel mill run. It would be nice if we could print some status line whenever we switch the output from one task to another, indicating the current outputting task name.
E.g. like when running `tail -f` on multiple files, it will insert a `[<filename>]` whenever the output changes it's source.

lihaoyi:
> API wise, I think the whole ticker thing should be something internal to the Evaluator and should not be exposed in the Logger API. That means random tasks can't add ticker messages which sound ok to me. Since we already have the Logger.ticker(String) API, which we need to keep for compatibility anyways, we could re-purpose to provide some "ticker suffix", e.g. something to add the the default ticker line of the current task in case it is printed.

Yeah I imagine we can keep `.ticker` and maybe expose it together with `.endTicker` (or combined as `.withTicker`???) that would let tasks add suffixes or temporarily override the ticker for periods of time. Like coursier resolution would be `foo.resolveDeps / downloading artifact blah blah blah`

> And here the bikeshedding continues. I always have some difficulties with reading the tasks output for a parallel mill run. It would be nice if we could print some status line whenever we switch the output from one task to another, indicating the current outputting task name.

Bikeshedding is great, this is our chance to do so haha.

The issue with printing status lines whenever we switch output is that we often switch outputs very frequently, e.g. in the `example/scalalib/web/6-cross-version-platform-publishing` I've been using, the output switches almost every line. See the `#NN` prefixes in the log below:

```
#03 [info] compiling 1 Scala source to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/mill-build/compile.dest/classes ...
#03 [info] done compiling
#04 [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/bar/js/compile.dest/classes ...
#02 [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/bar/jvm/compile.dest/classes ...
#05 [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/3.3.3/bar/jvm/compile.dest/classes ...
#01 [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/3.3.3/bar/js/compile.dest/classes ...
#02 [info] done compiling
#03 [info] compiling 1 Scala source to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/bar/jvm/test/compile.dest/classes ...
#02 [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/qux/jvm/compile.dest/classes ...
#04 [info] done compiling
#09 [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/qux/js/compile.dest/classes ...
#04 [info] compiling 1 Scala source to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/bar/js/test/compile.dest/classes ...
#05 [info] done compiling
#10 [info] compiling 1 Scala source to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/3.3.3/bar/jvm/test/compile.dest/classes ...
#07 [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/3.3.3/qux/jvm/compile.dest/classes ...
#03 [info] done compiling
#01 [info] done compiling
#02 [info] done compiling
#01 [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/3.3.3/qux/js/compile.dest/classes ...
#02 [info] compiling 1 Scala source to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/3.3.3/bar/js/test/compile.dest/classes ...
#05 [info] compiling 1 Scala source to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/qux/jvm/test/compile.dest/classes ...
#09 [info] done compiling
#08 [info] compiling 1 Scala source to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/qux/js/test/compile.dest/classes ...
#04 [info] done compiling
#05 [info] done compiling
#08 [info] done compiling
#07 [info] done compiling
#09 [info] compiling 1 Scala source to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/3.3.3/qux/jvm/test/compile.dest/classes ...
#10 [info] done compiling
#01 [info] done compiling
#06 [info] compiling 1 Scala source to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/3.3.3/qux/js/test/compile.dest/classes ...
#02 [info] done compiling
#09 [info] done compiling
#06 [info] done compiling
```

`#NN` is enough to tell you the logs are from different threads/tasks, but not enough to tell you what thread/task each log line is from, and of course because `NN` is the thread ID the same number can refer to different tasks. We could expand it to `#NNNN` were `NNNN` is the index of the task in the plan, so they don't get re-used, but then you still need somewhere to look up what each task index is. Maybe we could provide a helper?

I don't know if any build tool has a good story here.  IIRC:

* Bazel and Gradle just hide all the logs by defaults
* Maven and SBT just dump all the logs to stdout without worrying about interleaving

One idea I had is that instead of showing the target selector in the ticker, we could instead show the path to the `.log` file. That would be something close enough that you could immediately see what target is running, but also conveniently copy-pasteable into `tail -f` so you could see the logs for any particular task.

Beyond that, it's probably possible to do some clever TUI thing where you can select the thread you want below and it'll show the logs above, but that's probably more complexity than I'm willing to bear at this point in time

lihaoyi:
> In one run out of ten, my shell prompt was overridden (under-ridden) with ticker output, but in most of the other runs it was clear. ^C leaves the prompt in the ticker output too, but that is probably expected.

I'm still chasing down some concurrency issues that I suspect remain in the logging subsystem. Previously about two days to chase down the fact that `Piped{Input,Output}Stream`s cannot be used across multiple threads, so I replaced them with our own `PipeStreams.*`. Hopefully another few days and another few bugs fixed will be enough to add another few 9s to the level of robustness

lefou:
The `#NN` worker id is pretty useless once you understand that you have the output of multiple threads interleaved and threads are reused.

By printing a unique (per run) task ID `#NNNN`, it could become much easier to capture the output of a dedicated single task. Either via piping and `grep`ing or by some search capability like my console (`konsole` under Linux/KDE) provides. If we print a headline whenever a new tasks (that has not printed something before) is about to output something, we can easily catch the task name as the first line starting with the task ID `#NNNN`.

Your example from above could looks like this mock:

```
#03 mill-build.compile
#03 [info] compiling 1 Scala source to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/mill-build/compile.dest/classes ...
#03 [info] done compiling

#04 foo[2.13.14].bar.js.compile
#04 [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/bar/js/compile.dest/classes ...

#202 foo[2.13.14].bar.jvm.compile
#202 [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/bar/jvm/compile.dest/classes ...

#305 foo[3.3.3].bar.jvm.compile
#305 [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/3.3.3/bar/jvm/compile.dest/classes ...

#401 foo[3.3.3].bar.js.compile
#401 [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/3.3.3/bar/js/compile.dest/classes ...

#202 [info] done compiling

#503 foo[2.13.14].bar.jvm.compile
#503 [info] compiling 1 Scala source to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/bar/jvm/test/compile.dest/classes ...

#602 foo[2.13.14].qux.jvm.compile
#602 [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/qux/jvm/compile.dest/classes ...

#04 [info] done compiling

#709 foo[2.13.14].qux.js.compile
#709 [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/qux/js/compile.dest/classes ...

#804 foo[2.13.14].bar.js.test.compile
#804 [info] compiling 1 Scala source to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/bar/js/test/compile.dest/classes ...

#305 [info] done compiling
...
```

Bonus points if the full log is retained in a files, so we could easily `cat <log> | grep '^#602 ' to get the full single task output.

lihaoyi:
Your suggestion for `#NNNN` prefixes combined with a `#NNNN <task-name>` header sounds pretty promising

lefou:
I revised my mock-up to merge current task run ID and overall progress. 

```
[3/1000] mill-build.compile
[3] [info] compiling 1 Scala source to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/mill-build/compile.dest/classes ...
[3] [info] done compiling
[4/1000] foo[2.13.14].bar.js.compile
[4] [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/bar/js/compile.dest/classes ...
[202/1000] foo[2.13.14].bar.jvm.compile
[202] [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/bar/jvm/compile.dest/classes ...
[305/1000] foo[3.3.3].bar.jvm.compile
[305] [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/3.3.3/bar/jvm/compile.dest/classes ...
[401/1000] foo[3.3.3].bar.js.compile
[401] [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/3.3.3/bar/js/compile.dest/classes ...
[202] [info] done compiling
[503/1000] foo[2.13.14].bar.jvm.compile
[503] [info] compiling 1 Scala source to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/bar/jvm/test/compile.dest/classes ...
[602/1000] foo[2.13.14].qux.jvm.compile
[602] [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/qux/jvm/compile.dest/classes ...
[4] [info] done compiling
[709/1000] foo[2.13.14].qux.js.compile
[709] [info] compiling 2 Scala sources to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/qux/js/compile.dest/classes ...
[804/1000] foo[2.13.14].bar.js.test.compile
[804] [info] compiling 1 Scala source to /Users/lihaoyi/Github/mill/example/scalalib/web/6-cross-version-platform-publishing/out/foo/2.13.14/bar/js/test/compile.dest/classes ...
[305] [info] done compiling
...
```

Also, I think we should order the status of currently running jobs in the order of their task run ID, so we see the oldest and therefore slowest jobs the most. This causes some awareness of the heavy tasks, which might trigger unconscious thought process towards improvement.

lihaoyi:
@lefou sounds good. This will take some tweaks to the Logger API to lazily print the task name message on the first log line, but shouldn't be too difficult

lefou:
> ```
> [3/1000] mill-build.compile
> [3] [info] compiling 1 Scala source to /Users/lihaoyi/Github/mill/example/scalalib/web/6-

Hmm, it seems to be less `grep` friendly.

lihaoyi:
I think we can just display the total number of tasks in the header bar. No need to repeat it for every task that gets run

lihaoyi:
@lefou here is what I have now, again using `example/scalalib/web/6-cross-version-platform-publishing`. 

Screenshot 2024-09-23 at 9 28 26 PM

lihaoyi:
One issue with the screenshot above is that the task name "title"s look too much like normal log lines. We might need some way of distinguishing them, maybe i can color them blue like the task IDs on the left?

lihaoyi:
Here's what it looks like when they're colored blue. Pretty nice to look at in a terminal and should be understandable, but we still need some other fallback delimited for scenarios without ANSI color highlighting (e.g. log files)

Screenshot 2024-09-23 at 9 33 42 PM

lefou:
I think this already comes close to our old ticker, and not many users complained. If we re-add the overall count to get some better progress feeling, the line should look visually different enough even without color, and I get by feature too.

```
[NN] / [MMMM] <taskName-in-blue>
```

lihaoyi:
@lefou added the ` / [MMMM]` thing, I think it looks reasonable

Screenshot 2024-09-24 at 7 05 33 AM

lihaoyi:
I think `[3/1000]` can work, we just need to left pad it to `[0003/1000]`. That would make `0003` greppable as a unique string, since we wouldn't need to worry about it being the prefix or suffix of another longer string

lefou:
The new ticker is already great. I have two minor wishes, probably rather personal taste.

In my current console, I see up to 20 jobs listed. This list is a bit too long to read in that short time span it is shown. Maybe, we can limit it to some reasonable size, e.g. `8`. If we don't reach consensus on "reasonable", we could just read it from an env variable and power users can customize.

The other thing is, that I find it a bit hard to follow the logs. My eye is constantly swiping to the lower part. Maybe we can experiment with inserting an empty line between the logs and the progress ruler? I can do this myself if you could point me to the code location.

lihaoyi:
@lefou the rendering logic lives in `MultiLinePromptLoggerUtil.scala`, in particular `def renderPrompt` and `def renderHeader`. I managed to make it unit testable in `MultilinePromptLoggerUtilTests.scala`, so you should be able to iterate and experiment without having to go through the hassle of creating a `dist.launcher`, and when you're done you can leave a test case to ensure your scenario behaves as expected. The current heuristic gives the prompt the full terminal width and 1/3 of the terminal height.

From your comments, it sounds like you are using Mill with a much taller terminal than I typically do, which is ~30x120 quarter-screen on my macbook pro, with 10 cores so at most 10 concurrent tasks. I don't mind adding some customizations for bigger terminals, e.g. a blank line before the header line, as long as we set a threshold so it only appears when the terminal is sufficiently tall so it doesn't waste space on smaller terminals

lihaoyi:
Also there are some rudimentary integration tests for the whole `MultilinePromptLogger` in `MultilinePromptLoggerUtilTests.scala`, including exercising all the method calls and emulating terminal handling of the relevant ANSI escape codes.

lefou (approved):

lihaoyi:
One remaining issue here is that ad-hoc usages of `.ticker` in build code now misbehaves, e.g. the coursier `Downloading` indicators below which do not disappear:
Screenshot 2024-09-25 at 8 44 58 PM

Probably need to tweak the API and semantics of `.ticker` a bit to make this work properly, will take a look tomorrow

Issue #3576, opened 2024-09-18
Jmh with -jvmArgs --add-modules=jdk.incubator.vector


Quafadas:
I'm trying to benchmark the start of my attempts to play with project Panama's `VectorAPI`. 

I've successfully run JMH like this;

`mill benchmark.runJmh -jvmArgs --add-modules=jdk.incubator.vector -rf json vecxt.benchmark.DgemmBenchmark`

Which works If I'm depending on a  _library_ which uses the VectorAPI. However, it seems that if I try to use it directly myself, Jmh becomes unhappy. 

```
Processing 3 classes from /Users/simon/Code/vecxt/out/benchmark/compile.dest/classes with "reflection" generator
Writing out Java source to /Users/simon/Code/vecxt/out/benchmark/generateBenchmarkSources.dest/jmh_sources and resources to /Users/simon/Code/vecxt/out/benchmark/generateBenchmarkSources.dest/jmh_resources
Annotation generator had thrown the exception.
java.lang.NoClassDefFoundError: jdk/incubator/vector/Vector
        at java.base/java.lang.Class.getDeclaredMethods0(Native Method)
        at java.base/java.lang.Class.privateGetDeclaredMethods(Class.java:3601)
        at java.base/java.lang.Class.getDeclaredMethods(Class.java:2686)
        at org.openjdk.jmh.generators.reflection.RFClassInfo.getMethods(RFClassInfo.java:99)
        at org.openjdk.jmh.generators.core.BenchmarkGenerator.buildAnnotatedSet(BenchmarkGenerator.java:213)
        at org.openjdk.jmh.generators.core.BenchmarkGenerator.generate(BenchmarkGenerator.java:77)
        at org.openjdk.jmh.generators.bytecode.JmhBytecodeGenerator.main(JmhBytecodeGenerator.java:100)
Caused by: java.lang.ClassNotFoundException: jdk.incubator.vector.Vector
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:525)
        ... 7 more


1 targets failed
benchmark.generateBenchmarkSources java.lang.Exception: Interactive Subprocess Failed (exit code 1)
    mill.util.Jvm$.runSubprocess(Jvm.scala:188)
[...]

PR #3578, opened 2024-09-18
Forward javacOptions to JMH

Quafadas:
I believe this to be  a solution for #3576 . However, I'm unclear how I could effectively test it. Feedback welcomed.

### DIFF
--- a/contrib/jmh/src/mill/contrib/jmh/JmhModule.scala
+++ b/contrib/jmh/src/mill/contrib/jmh/JmhModule.scala
@@ -72,7 +72,7 @@ trait JmhModule extends JavaModule {
   def generateBenchmarkSources =
     Task {
       val dest = T.ctx().dest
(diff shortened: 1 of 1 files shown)


lihaoyi:
@Quafadas I think manual testing is fine for this. Presumably you are able to reproduce the original reported issue, and can use `installLocalCache` to see if your PR fixes it?

Quafadas:
The original issue I could reproduce easily. In the issue I left a comment which was a workaround, where I copy pasted overrode the method in my local build, and that fixed the issue. 

I’m afraid I don’t understand `installLocalCache`?

lihaoyi:
Oh I meant that you could use it for manual testing: it generates a local `mill` binary with all thee stuff `publishLocaled` for you to try out. But if the override works in your local build, I think that's enough confidence for me

Issue #3276, opened 2024-07-19
Mill clean should stop/close workers


lefou:
### Discussed in https://github.com/com-lihaoyi/mill/discussions/3272

<div type='discussions-op-text'>

<sup>Originally posted by **roman-mibex-2** July 18, 2024</sup>
TLDR: `mill clean` should stop workers / or discard them. 

I have a mill build with some webpack workers, which stay up. I sometimes had a puzzling state 
where these workers seem complete out of sync. I've noticed that a `./mill clean` breaks them.

Here is a small demostration, without webpack ;)

build.sc:
```scala
import mill._
import os.Path

object `mill-demo` extends RootModule  {
  def resources = T.source(millSourcePath/"barely-changing.txt")
  def source = T.source(millSourcePath/"frequently-changing.js")

  def webpackWorker = T.worker{
    // When the worker starts, we copy stuff from other targets
    // These targets rarely change, and we want to keep them 'hot' in the worker
    os.copy(resources().path,T.dest/"resource.txt",replaceExisting = true)
    new MyWebpackWorker(T.dest)
  }

  def webpack = T{
    webpackWorker().compile(source().path)
  }

  class MyWebpackWorker(myWorkingDir:Path){
    println("Worker is starting up")
      def compile(latest:Path)={
        os.copy(latest,myWorkingDir/"code.js", replaceExisting = true)
        val state = PathRef(myWorkingDir)
        println(s"State of working dir is ${state}")
        state
      }
  }
}
```

Demostration of the issue:
```
$ date > barely-changing.txt
$ date > frequently-changing.js

# Initial build, all fine
$ ./mill show webpack
[1/1] show > [2/4] webpackWorker 
Worker is starting up, worker:1930565333
[1/1] show > [4/4] webpack 
worker: 1930565333
"ref:v0:51849a16:/home/roman/dev/private-dev/mill-demo/out/webpackWorker.dest"

# I change the source code, all works as expected
$ date > frequently-changing.js
$ ./mill show webpack
[1/1] show > [4/4] webpack
worker: 1930565333
"ref:v0:800319cb:/home/roman/dev/private-dev/mill-demo/out/webpackWorker.dest"

# In the rare occasion the resources change, the worker is restarted, works as expected
$ ./mill show webpack
[1/1] show > [2/4] webpackWorker 
Worker is starting up, worker:2000339927
[1/1] show > [4/4] webpack 
worker: 2000339927
[...]

PR #3579, opened 2024-09-18
Clean workers too from the clean command

alexarchambault:
This PR ensures the `clean` command works for worker tasks, and allows users to dispose of workers by removing their corresponding metadata file on disk (`out/foo/theWorker.json` for worker `foo.theWorker`) - in that case, the worker instance is dropped upon first access to it.

Fixes https://github.com/com-lihaoyi/mill/issues/3276

### DIFF
--- a/main/eval/src/mill/eval/Evaluator.scala
+++ b/main/eval/src/mill/eval/Evaluator.scala
@@ -21,7 +21,14 @@ trait Evaluator {
   def outPath: os.Path
   def externalOutPath: os.Path
   def pathsResolver: EvaluatorPathsResolver
+  // TODO In 0.13.0, workerCache should have the type of mutableWorkerCache,
+  // while the latter should be removed
   def workerCache: collection.Map[Segments, (Int, Val)]
+  private[mill] final def mutableWorkerCache: collection.mutable.Map[Segments, (Int, Val)] =
+    workerCache match {
+      case mut: collection.mutable.Map[Segments, (Int, Val)] => mut
+      case _ => sys.error("Evaluator#workerCache must be a mutable map")
+    }
   def disableCallgraphInvalidation: Boolean = false
 
   @deprecated(
--- a/main/resolve/src/mill/resolve/ResolveCore.scala
+++ b/main/resolve/src/mill/resolve/ResolveCore.scala
@@ -29,7 +29,7 @@ private object ResolveCore {
 
   object Resolved {
     case class Module(segments: Segments, cls: Class[_]) extends Resolved
-    case class Target(segments: Segments) extends Resolved
+    case class NamedTask(segments: Segments) extends Resolved
     case class Command(segments: Segments) extends Resolved
   }
 
@@ -327,7 +327,7 @@ private object ResolveCore {
         .map(
           _.map {
             case (Resolved.Module(s, cls), _) => Resolved.Module(segments ++ s, cls)
-            case (Resolved.Target(s), _) => Resolved.Target(segments ++ s)
+            case (Resolved.NamedTask(s), _) => Resolved.NamedTask(segments ++ s)
             case (Resolved.Command(s), _) => Resolved.Command(segments ++ s)
           }
             .toSet
@@ -376,10 +376,10 @@ private object ResolveCore {
       }
     }
 
-    val targets = Reflect
-      .reflect(cls, classOf[Target[_]], namePred, noParams = true)
+    val namedTasks = Reflect
+      .reflect(cls, classOf[NamedTask[_]], namePred, noParams = true)
       .map { m =>
-        Resolved.Target(Segments.labels(decode(m.getName))) ->
+        Resolved.NamedTask(Segments.labels(decode(m.getName))) ->
           None
       }
 
@@ -388,7 +388,7 @@ private object ResolveCore {
       .map(m => decode(m.getName))
       .map { name => Resolved.Command(Segments.labels(name)) -> None }
 
-    modulesOrErr.map(_ ++ targets ++ commands)
+    modulesOrErr.map(_ ++ namedTasks ++ commands)
   }
 
   def notFoundResult(
--- a/main/src/mill/main/MainModule.scala
+++ b/main/src/mill/main/MainModule.scala
@@ -2,7 +2,7 @@ package mill.main
 
 import java.util.concurrent.LinkedBlockingQueue
 import mill.define.{BaseModule0, Command, NamedTask, Segments, Target, Task}
-import mill.api.{Ctx, Logger, PathRef, Result}
+import mill.api.{Ctx, Logger, PathRef, Result, Val}
 import mill.eval.{Evaluator, EvaluatorPaths, Terminal}
 import mill.resolve.{Resolve, SelectMode}
 import mill.resolve.SelectMode.Separated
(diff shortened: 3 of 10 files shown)


alexarchambault:
Still a draft, as I'm unsure about some reflection-related changes in `ResolveCore` (testing them in an isolated fashion in https://github.com/com-lihaoyi/mill/pull/3580).

The type of `Evaluator.workerCache` could be left unchanged too, with some refactoring.

main/eval/src/mill/eval/Evaluator.scala
```diff
@@ -21,7 +21,7 @@ trait Evaluator {
   def outPath: os.Path
   def externalOutPath: os.Path
   def pathsResolver: EvaluatorPathsResolver
-  def workerCache: collection.Map[Segments, (Int, Val)]
+  def workerCache: collection.mutable.Map[Segments, (Int, Val)]
```
lihaoyi, line 24: If this counts as a change due to bincompat, we can create a new method with a new name that's `private[mill]` until 0.13.0 comes and we can change the type of the old method

main/resolve/src/mill/resolve/ResolveCore.scala
```diff
@@ -376,8 +376,8 @@ private object ResolveCore {
       }
     }
 
-    val targets = Reflect
-      .reflect(cls, classOf[Target[_]], namePred, noParams = true)
```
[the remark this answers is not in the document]
alexarchambault: That's the case, I checked that just before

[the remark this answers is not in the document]
alexarchambault: https://github.com/com-lihaoyi/mill/pull/3580 (whose changes are also in this PR) makes that explicit

alexarchambault:
(Seems the CI error is a flake - can't reproduce it locally)

lihaoyi:
kicked it

main/eval/src/mill/eval/Evaluator.scala
```diff
@@ -21,7 +21,14 @@ trait Evaluator {
   def outPath: os.Path
   def externalOutPath: os.Path
   def pathsResolver: EvaluatorPathsResolver
+  // TODO In 0.13.0, workerCache should have the type of mutableWorkerCache,
+  // while the latter should be removed
   def workerCache: collection.Map[Segments, (Int, Val)]
+  private[mill] final def mutableWorkerCache: collection.mutable.Map[Segments, (Int, Val)] =
+    workerCache match {
+      case mut: collection.mutable.Map[Segments, (Int, Val)] => mut
+      case _ => sys.error("should not happen")
```
lefou, line 30:
```suggestion
      case _ => sys.error("should not happen")
```

This is not actionable. We don't control how the Evaluator is instantiated, so this can happen very well. The message should indicate the concrete issue: a worker must be of type `mutable.Map`.

main/eval/src/mill/eval/Evaluator.scala
```diff
@@ -21,7 +21,14 @@ trait Evaluator {
   def outPath: os.Path
   def externalOutPath: os.Path
   def pathsResolver: EvaluatorPathsResolver
+  // TODO In 0.13.0, workerCache should have the type of mutableWorkerCache,
+  // while the latter should be removed
   def workerCache: collection.Map[Segments, (Int, Val)]
```
lefou, line 26: Althouh this must stay for compat, we already can deprecate it in favor of its replacement. We try to deprecate API we intend to remove, to provide a smoother migration path.

alexarchambault: I was imagining a simple change to `def workerCache: collection.mutable.Map[Segments, (Int, Val)]` in `0.13`, and a removal of `mutableWorkerCache`, that should only be used internally. It's a really small change.

Are there know users of the `Evaluator` trait, that don't rely on the Mill internal implementation? Implementing that seems more of an internal thing to me… (and for users of `Evaluator`, given that `collection.mutable.Map <: collection.Map`, the change should be source-compatible).

main/eval/src/mill/eval/Evaluator.scala
alexarchambault: Sure

main/eval/src/mill/eval/Evaluator.scala
lihaoyi: For now the policy is that the things that are exposed we treat as stable. We don't actually know if anyone uses it, but if we expect something to be internal-only we'll mark them as `private[mill]` and MIMA will let us do as we like

main/test/src/mill/main/MainModuleTests.scala
```diff
@@ -27,6 +29,16 @@ object MainModuleTests extends TestSuite {
...
     }
     def helloCommand(x: Int, y: Task[String]) = Task.Command { (x, y(), hello()) }
+
+    /**
+     * The hello worker
+     */
+    def helloWorker = Task.Worker {
+      // non-serializable, but everything should work fine nonetheless
+      new AutoCloseable {
+        def close() = ()
```
lihaoyi, line 39: We should make sure `.close()` ends up being called, because otherwise you end up leaking workers/classloaders/filehandles which can be pretty subtle and hard to debug. Some hacky global registry of "things we closed" would suffice for test suite purposes

main/src/mill/main/MainModule.scala
```diff
@@ -348,12 +348,25 @@ trait MainModule extends BaseModule0 {
...
+          .keysIterator
+          .filter(workerSegments => allSegments.exists(workerSegments.startsWith))
+          .toVector
+        val closeableCleanedWorkers = workersToRemove
+          .flatMap(workerSegments => evaluator.mutableWorkerCache.remove(workerSegments))
+          .collect {
+            case (_, Val(closeable: AutoCloseable)) =>
+              closeable
+          }
+        closeableCleanedWorkers.foreach(_.close())
```
lihaoyi, line 368: This is kind of roundabout, probably would be cleaner to write as a straight loop

```scala
for{
  workerSegments <- evaluator.workerCache.keysIterator.toList
  if allSegments.exists(workerSegments.startsWith))
}{
  evaluator.mutableWorkerCache.remove(workerSegments) match{
    case (_, Val(closeable: AutoCloseable)) => closeable.close()
    case _ => // do nothing
  }
}
```

main/eval/src/mill/eval/Evaluator.scala
alexarchambault: The current state of the PR doesn't break bin compat. But the question is, should this (minor and mostly source-compatible) change go through a deprecation cycle, or should we do it in one go in `0.13`? I'd be ok with a change in one go, for this particular minor change

main/test/src/mill/main/MainModuleTests.scala
alexarchambault: The `clean-worker` tests in that file do just that

main/src/mill/main/MainModule.scala
alexarchambault: Sure

lihaoyi (approved):
Last things before merging, apart from those already mentioned in https://github.com/com-lihaoyi/mill/pull/3580:


1. Can we make `rm -rf` clean the workers as well? If we leave a `module-name/worker-name.json` file on disk as I suggested in the other PR, we could compare it to what's in memory, and an `rm -r out/module-name` would invalidate the worker on next run even without a `mill clean`. Although not strictly necessary since people "should" be using `clean` instead of `rm -rf`, keeping both approaches working would help maintain the symmetry between the `out` folder and the module/task hierarchy

2. Update PR description

PR #3580, opened 2024-09-18
Show workers in resolve / show / inspect too

alexarchambault:
This PR generalizes some `Target`-related stuff in `Resolve` to all `NamedTask`s with no arguments (targets and workers), and makes sure `resolve` / `show` / `inspect` work fine with workers. In particular, running `show` on a worker now prints something like
```json
{
  "worker": "foo.theWorker",
  "toString": "build_.package_$foo$$anon$1@4acbdb44",
  "inputsHash": -2075822978
}
```

### DIFF
--- a/main/test/src/mill/main/MainModuleTests.scala
+++ b/main/test/src/mill/main/MainModuleTests.scala
@@ -27,6 +27,18 @@ object MainModuleTests extends TestSuite {
       Map("1" -> "hello", "2" -> "world")
     }
     def helloCommand(x: Int, y: Task[String]) = Task.Command { (x, y(), hello()) }
+
+    /**
+     * The hello worker
+     */
+    def helloWorker = Task.Worker {
+      // non-JSON-serializable, but everything should work fine nonetheless
+      new AutoCloseable {
+        def close() = ()
+        override def toString =
+          "theHelloWorker"
+      }
+    }
     override lazy val millDiscover: Discover = Discover[this.type]
   }
 
@@ -90,6 +102,17 @@ object MainModuleTests extends TestSuite {
         assert(
           res.startsWith("helloCommand("),
           res.contains("MainModuleTests.scala:"),
+          res.contains("hello")
+        )
+      }
+      test("worker") - UnitTester(mainModule, null).scoped { eval =>
+        val Right(result) = eval.apply("inspect", "helloWorker")
+
+        val Seq(res: String) = result.value
+        assert(
+          res.startsWith("helloWorker("),
+          res.contains("MainModuleTests.scala:"),
+          res.contains("The hello worker"),
           res.contains("hello")
         )
       }
@@ -173,6 +196,14 @@ object MainModuleTests extends TestSuite {
         val Seq(res) = result.value
         assert(res == ujson.Arr(1337, "lol", ujson.Arr("hello", "world")))
       }
+
+      test("worker") {
+        val Right(result) = evaluator.apply("show", "helloWorker")
+        val Seq(res: ujson.Obj) = result.value
+        assert(res("toString").str == "theHelloWorker")
+        assert(res("worker").str == "helloWorker")
+        assert(res("inputsHash").numOpt.isDefined)
+      }
     }
 
     test("showNamed") {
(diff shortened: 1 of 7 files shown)


alexarchambault:
(CI failure seems to be a flake, so the changes here seem fine)

alexarchambault:
Forgot about bin compat, I think…

lefou:
What's the motivation for this change? Workers aren't of any direct use from the cli. OTOH, to clean them up, the need to be discoverable. Is it that?

alexarchambault:
Yes, that's for https://github.com/com-lihaoyi/mill/pull/3579

alexarchambault:
I still find it handy to be able to create them from the command-line. I remember having been surprised that wasn't possible when I was discovering Mill, a few years ago.

main/test/src/mill/main/MainModuleTests.scala
```diff
@@ -27,6 +27,12 @@ object MainModuleTests extends TestSuite {
       Map("1" -> "hello", "2" -> "world")
     }
     def helloCommand(x: Int, y: Task[String]) = Task.Command { (x, y(), hello()) }
+    def helloWorker = Task.Worker {
+      // non-serializable, but everything should work fine nonetheless
```
lefou, line 31: What the story behind this comment? We never serialize workers in Mill.

main/test/src/mill/main/MainModuleTests.scala
```diff
@@ -93,6 +99,16 @@ object MainModuleTests extends TestSuite {
           res.contains("hello")
         )
       }
+      test("worker") - UnitTester(mainModule, null).scoped { eval =>
+        val Right(result) = eval.apply("inspect", "helloWorker")
+
+        val Seq(res: String) = result.value
+        assert(
```
lefou, line 106: Would be good to also cover the scaladoc of the worker.

main/test/src/mill/main/MainModuleTests.scala
alexarchambault: The `show` command tries to serialize them nonetheless, with the changes here. The `if t.writerOpt.nonEmpty` in `RunScript.scala` works around that.

main/test/src/mill/main/MainModuleTests.scala
alexarchambault: Sure, why not. I had copied the tests for commands, that don't test the scaladoc.

main/test/src/mill/main/MainModuleTests.scala
lefou:
```suggestion
      // non-json-serializable, but everything should work fine nonetheless
```

lihaoyi:
I'm fine with this change in principle, as making `resolve`/`show`/`inspect` more powerful seems like a good thing in general. e.g. https://github.com/com-lihaoyi/mill/pull/3532 is working to make those tasks work on `Module`s, which although they are different from tasks and are not serializable at all, would be very useful for a user trying to poke at a build from the command line.

For `Worker`s, some things could work, some things can't. Let's try and enumerate them:

- `inspect` could show the file name, line number, and Scaladoc same as other tasks
- `show` cannot show JSON since workers are not JSON serializable. Maybe we can just show some stub JSON string like `"path.to.worker:${worker.toString}"`?
- `resolve` listing workers seems fine
- Running workers from the command line seems like it should work. Notably it would not result in a long lived object with `--no-server`, whereas `runBackground` would, so it's unclear how useful it would be.

I think for the purposes of this PR I'm happy to move forward, but we should flesh out the integration tests a bit both for correctness as well as clarity in what we expect the behavior to be

1. We can add a worker to `integration/feature/docannotations` to exercise `inspect` and make sure that works
2. We can add an integration test somewhere for `resolve` to make sure that works
3. Running workers from the CLI and using `show` seems like something we can skip for now with an error

alexarchambault:
There are already tests for `resolve` and `inspect` (and `show` too), but unit ones, in `main/test/src/mill/main/MainModuleTests.scala`.

alexarchambault:
> * Running workers from the command line seems like it should work. Notably it would not result in a long lived object with `--no-server`, whereas `runBackground` would, so it's unclear how useful it would be.

I guess that's fine. That's the behavior I'd expect as a user. Ruuning it with `--no-server` basically checks that the worker can be instantiated fine.

alexarchambault:
> * `show` cannot show JSON since workers are not JSON serializable. Maybe we can just show some stub JSON string like `"path.to.worker:${worker.toString}"`? Or we could error out and say "cannot show value of worker"

In the current state of this PR, we get an empty JSON object

lihaoyi (approved):
I think this looks great. Three things before merging:

1. Let's flesh out the `show` output so it's a bit more informative than `{}`; once we start showing workers in `resolve`, people will start trying to poke at them via `inspect` and `show`. `inspect` already works as your tests demonstrate, but `show` is also a common thing to use so we should make it informative. How about something like this
```scala
{
  "worker": "<worker-selector>",
  "toString": "<worker.toString>",
  "inputHash": 123456789
}
```

2. Should we leave a `<worker-name>.json` file on disk containing the snippet above? That would make the `show` output and the "dig through files on disk" output more consistent, so anyone digging through the build using `cat` or `less` instead of `mill show` would be able to inspect workers as well

3. Fill out the PR description summarizing the current state of the PR

alexarchambault:
@lihaoyi Your 3 points should be addressed

PR #3585, opened 2024-09-20
Add kotlinlib `example/module` examples

0xnm:
This PR adds more stuff for #3451, namely `example/kotlinlib/module` examples.

Few notes about things being done:

* `1-compilation-execution-flags` - there is not so many flags which can be passed to the Kotlin Compiler itself. I simply used [-Werror](https://kotlinlang.org/docs/compiler-reference.html#werror).
* `6-annotation-processors` - I used `kotlinx.serialization` as proposed by @lefou. It seems Kotlin Compiler doesn't automatically pick compilation plugins from the classpath, so I had to pass path to it explicitly using `-Xplugin` option.
* `7-dokkajar` - obviously `javadoc` cannot be used to generate docs for Kotlin classes, so I added support for [Dokka](https://github.com/Kotlin/dokka) to `KotlinModule` and added a forwarding of `docJar` calls to `dokkaJar`.
* `13-jni` - there is no possibility to generate header files from Kotlin files (unless an ugly trick is used to take compiled `.class` files, decompile them back, clean-up and pass to `javac` to generate headers), so I dropped header generation.


example/kotlinlib/module/6-annotation-processors/build.mill
```diff
@@ -0,0 +1,44 @@
...
+  def processors = T {
+    defaultResolver().resolveDeps(
+      Agg(
+        ivy"org.jetbrains.kotlin:kotlin-serialization-compiler-plugin:1.9.24"
+      )
+    )
+  }
+
+  def kotlincOptions = super.kotlincOptions() ++ Seq(
+    s"-Xplugin=${processors().map(_.path).head}"
```
lihaoyi, line 27: `processors().head.path` is a simpler way of expressing what we want here

example/kotlinlib/module/6-annotation-processors/build.mill
```diff
@@ -0,0 +1,44 @@
...
+  object test extends KotlinModuleTests with TestModule.Junit5 {
+    def ivyDeps = super.ivyDeps() ++ Agg(
+      ivy"io.kotest:kotest-runner-junit5-jvm:5.9.1"
+    )
+  }
+}
+
+/**
+ * Usage
+ *
```
lihaoyi, line 39: The `Usage` comment has to follow the format of the others, this one was probably auto-reformatted which would cause the example tester to fail

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -134,6 +135,83 @@ trait KotlinModule extends JavaModule { outer =>
     ()
   }
 
+  override def docJar: T[PathRef] = T[PathRef] {
+    T.log.info("docJar task shouldn't be used for Kotlin modules, using dokkaJar instead")
```
lihaoyi, line 139: I wonder if we really need this warning? e.g. for both Java and Scala 2.x and Scala 3.x they all use `docJar`, despite being wildly different backend implementations. Can we just let KotlineModule users use `docJar` as well?

lihaoyi:
I think this looks pretty good. Left one more comment, and lemme try to get tests green before merging

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
0xnm: Sure, I added Dokka backend for `docJar` directly, without a dedicated `dokkaJar` task (although the example is still called `7-dokkajar` to avoid copying lines from `scalalib`).

lihaoyi:
Seems the mockito tests are flaky, will merge regardless and deal with the flaky test after

lihaoyi:
Thanks @0xnm! will transfer the bounty using the same details as earlier

PR #3589, opened 2024-09-22
Add Kotlin publishing and testing examples

0xnm:
Another addition for #3451: it adds `publishing` and `testing` examples for Kotlin.

The only thing that needs to be mentioned is test filtering in Kotest: it doesn't have `package.class.method` format to do the filtering, instead it is using `specs` + `classes` parameters (see [here](https://kotest.io/docs/framework/conditional/conditional-tests-with-gradle.html)).

Also `mockito-kotlin` requires JVM bytecode target to be 11.


lihaoyi:
Thanks @0xnm, looks great. I'll close out the bounty and possibly open up some further ones for the next steps in the kotlin support effort

PR #3595, opened 2024-09-23
add -sourceroot flag when generating semanticdb with scala 3

bishabosha:
while developing Scala 3 port, when using metals as the BSP server then semanticdb files are generated in the same folder as the source file.

With this change, and setting the bootstrapped mill version for BSP server, metals no longer puts semanticdb in the same directory, but instead in `out/../semanticdbData.dest/classes/META-INF/semanticdb/...`


lefou (approved):
Looks good to me. I remember vaguely, that when implementing it initially, someone said the `sourceroot` is supposed to determined automatically. But I like the explicit setups.

Issue #3519, opened 2024-09-12
Add a global lock around the `out/` folder to allow a single `mill` command to run at a time


lihaoyi:
Follow up to https://github.com/com-lihaoyi/mill/issues/3397

Mill does not have a proper concurrency model, so running multiple mill commands in parallel can result in all sorts of crashes and misbehaviors. This is especially true because IDEs tend to run Mill on their own when trying to import the project. 

Here are some issues I suspect are related:

* https://github.com/com-lihaoyi/mill/issues/1711
* https://github.com/com-lihaoyi/mill/issues/2484
* https://github.com/com-lihaoyi/mill/issues/3273
* https://github.com/com-lihaoyi/mill/issues/3454

https://github.com/com-lihaoyi/mill/pull/3243 was added to address parallel evaluations within the same process via BSP, but parallel evaluations across multiple processes remains a problem.

As a first pass, we should put all Mill evaluations between a global filesystem lock. This is the approach Bazel takes, and while not perfect, it works acceptably even for very large codebases and very large engineering organizations. We can add an optional `--no-global-lock` flag to disable this on an opt-in basis, but given the ongoing issues we need to go for correctness first by default, which means a global lock. The opt-out flag should be enough to handle edge cases where people really need the parallelism.

Implementing https://github.com/com-lihaoyi/mill/issues/3144 and moving the IDE work to a separate folder `.bsp/out` will help mitigate the most common concurrency issue of someone running 1 terminal and 1 IDE, at the cost of a bit of wasted computation (hopefully not too much)

We can figure out a finer-grained concurrency story later to try and optimize this, but I expect that will take a lot more time with no guarantee of success, so we should go with the global lock first

lihaoyi:
cc @alexarchambault

alexarchambault:
Fixed by https://github.com/com-lihaoyi/mill/pull/3599 I think

lihaoyi:
Let's merge this one too then we can close this https://github.com/com-lihaoyi/mill/pull/3683

PR #3599, opened 2024-09-23
Acquire a lock on the out dir in order to run tasks / commands

alexarchambault:
First cut at https://github.com/com-lihaoyi/mill/issues/3519

There's no tests yet, but it works fine in manual tests.

### DIFF
--- a/testkit/src/mill/testkit/IntegrationTester.scala
+++ b/testkit/src/mill/testkit/IntegrationTester.scala
@@ -5,11 +5,6 @@ import mill.eval.Evaluator
 import mill.resolve.SelectMode
 import ujson.Value
 
-import java.util.concurrent.atomic.AtomicInteger
-
-import scala.concurrent.{Future, Promise}
-import scala.util.Try
-
 /**
  * Helper meant for executing Mill integration tests, which runs Mill in a subprocess
  * against a folder with a `build.mill` and project files. Provides APIs such as [[eval]]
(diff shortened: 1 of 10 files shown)


alexarchambault:
This results in things like

https://github.com/user-attachments/assets/8254ec15-c536-41b7-8f7c-c21acc967d9d

alexarchambault:
Note that the lock is only for write operations, not those that just read things. In my experience, issues happen mainly when workers intend to run tasks and write things at the same time. But a shared lock (allowing several workers to hold it while they read things) could be added too.

alexarchambault:
(This includes https://github.com/com-lihaoyi/mill/pull/3530)

lihaoyi:
For a first pass, let's place the lock around the entire `MillMain.main0` workflow. It might be more coarse grained than necessary, but I think it's better to go for correctness first and we can figure out the finer-grained concurrency story later.

As an escape hatch, we should add a flag `--no-build-lock` or something so people who depend on concurrent Mill workflows can continue to do so (at their own risk, but not any more risk than they have today)

Lastly, let's make the `Waiting for lock...` message a bit more informative. Someone hitting `Waiting for lock...` probably has Mill running with `-w` in some other terminal they forgot about, or their IDE is running Mill on their behalf, and they would want to be told what is going on so they know what to do (use their other terminal, cancel Mill in their other terminal, wait for their IDE to finish indexing or cancel it). We could store the Mill command line args in an `out/mill-current-command.txt` file or something and display it as part of the message e.g. `Waiting for running Mill command ./mill -w __.compile to finish...` so they have some idea of what is going on and what they can do about it

alexarchambault:
Isn't a lock around the whole of `MillMain.main0` too much? That would basically ensure only one Mill process is running at a time on a project, IIUC. It would definitely break many of my workflows (I have apps and tests poking with the terminal in many of my projects, so I tend to use `-i` very often, and often run several Mill commands at the same time in different terminals - plus a BSP server). A lock on `MillMain.main0` would be quite harmful in those scenarios.

alexarchambault:
The last push makes the lock slightly broader compared to the previous state of this PR. The lock was around two things: `EvaluatorCore.evaluate0`, and the writing of the `mill-runner-state.json` file (look for the `logger.waitForLock` calls in the PR), it's now only around `MillBuildBootstrap.evaluate`.

Note that BSP is currently broken when the lock is enabled: the BSP server runs via the `mill.bsp.BSP/startSession` command, that itself runs other tasks / commands, and the lock isn't reentrant, so that the latter fails.

lihaoyi:
If I understand correctly, putting the lock around `MillBuildBootstrap#evaluate` rather than `MillMain.main0` would allow the lock to be released when `-w` is watching for files but not actually executing anything. I think that sounds great.

For BSP support, how about we disable the lock for BSP entirely, with the expectation that once https://github.com/com-lihaoyi/mill/pull/3530 lands we can move the BSP code path to use `.bsp/out/` as its `out/` folder which should eliminate conflicts between BSP and non-BSP commands. That would end up with manual CLI commands single-threaded due to locking on `out/`, and BSP commands single-threaded due to running isolated in `.bsp/out/`, at the cost of some duplication of work (hopefully not too much, since typically BSP does not need to compile the actual Scala code in the common case)

alexarchambault:
Rebased on top of #3608, to show that this doesn't break the BSP server (and commands asking for an `Evaluator` or a `Evaluator.AllBootstrapEvaluators` more generally)

alexarchambault:
> hopefully not too much, since typically BSP does not need to compile the actual Scala code in the common case

It should actually compile Scala code in most cases, AFAIK. In Metals at least. Compilation is needed to get the semanticdb files of each source, that Metals uses to provide code navigation in particular.

lihaoyi:
@alexarchambault got it. For the current status of the PR, I assume you ended up disabling the BSP server as discussed?

alexarchambault:
> For the current status of the PR, I assume you ended up disabling the BSP server as discussed?

There used to be a workaround so that the BSP server can use the lock too, but it breaks bin compat. I moved to hack to https://github.com/com-lihaoyi/mill/pull/3683 for later.

testkit/src/mill/testkit/IntegrationTester.scala
```diff
@@ -91,6 +96,49 @@ object IntegrationTester {
...
+        cwd: os.Path = workspacePath,
+        stdin: os.ProcessInput = os.Pipe,
+        stdout: os.ProcessOutput = os.Pipe,
+        stderr: os.ProcessOutput = os.Pipe,
+        mergeErrIntoOut: Boolean = false,
+        timeout: Long = -1,
+        check: Boolean = false,
+        propagateEnv: Boolean = true,
+        timeoutGracePeriod: Long = 100
+    ): Future[IntegrationTester.EvalResult] = {
```
lihaoyi, line 112: Rather than duplicating `eval` and having to duplicate the entire method signature and maintain it going forward, can we instead spawn a `ThreadPoolExecutor` and use `Future{eval()}` calls? The performance difference is immaterial since these are small test cases, and it would simplify the code considerably

alexarchambault: 👍  I was hesitating to do something along those lines

lihaoyi:
This looks good to me, @alexarchambault if you're done with I'll merge

alexarchambault:
> This looks good to me, @alexarchambault if you're done with I'll merge

I'm done, yes

PR #3600, opened 2024-09-24
Make Evaluator Closeable

alexarchambault:
Cleaning-up the worker cache upon close

### DIFF
--- a/main/eval/src/mill/eval/Evaluator.scala
+++ b/main/eval/src/mill/eval/Evaluator.scala
@@ -6,6 +6,8 @@ import mill.define.{BaseModule, Segments, Task}
 import mill.eval.Evaluator.{Results, formatFailing}
 import mill.util.{ColorLogger, MultiBiMap}
 
+import java.io.Closeable
+
 import scala.annotation.nowarn
 import scala.jdk.CollectionConverters._
 import scala.reflect.ClassTag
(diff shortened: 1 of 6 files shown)


alexarchambault:
I decided to open a PR for that for https://github.com/com-lihaoyi/mill/pull/3579 (https://github.com/com-lihaoyi/mill/pull/3579#discussion_r1772373344 in particular), but I'm now realizing this might not be needed for it after all. Don't know what you think of that, I think it's a good idea to clean up things when the `Evaluator` isn't needed anymore.

alexarchambault:
(This makes `Evaluator` extends `AutoCloseable`, so this breaks bin compat…)

lihaoyi:
I'm not sure this is the right thing to do. Right now in production, the `workerCache` is this long lived thing that is shared across `main0` invocations, and I think can out-live the individual evaluator that created it to be used in the next Mill command's `evaluator`. If that's the case, then having the `workerCache` be closed together with the evaluator at end of life would be wrong.

That code around sharing stuff across `main0` commands is pretty gnarly, but that's what I last remember it being like

alexarchambault:
It seems this works at the end (CI is green), as in the watch loop, the previous `RunnerState` isn't closed, so that underlying `Evaluator`s aren't either, and `workerCache`s are kept intact and can keep being used. But I don't think I did that on purpose 😅

lihaoyi:
Yeah that's why it's green, and then the next question is more what should the logic be here? I don't actually know, but although the current logic of injecting the `workerCache` into the `Evaluator` is pretty weird and roundabout, I'm inclined to leave it for now. I suspect that refactoring things enough to do things "properly" will require breaking bincompat, so we should wait until after 0.12.0 has landed and we're working towards 0.13.x and have the flexibility to make more sweeping changes without being restricted by bincompat requirements

PR #3602, opened 2024-09-25
Remove unused env var in examples forkEnv

alexarchambault:
Stumbled upon it when diving in the examples internals. It seems to me that this env var is unused (can't find it anywhere else by grepping the whole codebase).


lihaoyi (approved):
looks good, may be a relic of an older version of the example test suites before https://github.com/com-lihaoyi/mill/pull/3398 extracted them out into a self contained class

PR #3608, opened 2024-09-26
Add some BSP integration tests

alexarchambault:
This adds some BSP integration tests, ensuring the Mill BSP server answers correctly basic BSP requests. BSP responses are serialized on disk and committed, and we compare the requests we get to those on disk. These fixtures on disk can be automatically generated and updated by changing a variable (currently, `BspServerUtil.updateFixtures`), so that it's quite convenient to generate them, and update them and get a nice diff of the changes at the same time.


alexarchambault:
Not sure the CI error is related to my changes

lihaoyi:
It's the flakiness that should be fixed by https://github.com/com-lihaoyi/mill/pull/3642

alexarchambault:
Don't fear the apparent large diff, it's a lot of test fixtures

build.mill
```diff
@@ -417,6 +417,7 @@ trait MillBaseTestsModule extends TestModule {
       s"-DMILL_SCALA_2_13_VERSION=${Deps.scalaVersion}",
       s"-DMILL_SCALA_2_12_VERSION=${Deps.workerScalaVersion212}",
       s"-DTEST_SCALA_2_13_VERSION=${Deps.testScala213Version}",
+      s"-DMILL_TEST_SCALA_2_13_VERSION=${Deps.testScala213Version}",
```
lefou, line 420: This is already defined as `TEST_SCALA_2_13_VERSION`.

build.mill
```diff
@@ -775,7 +776,30 @@ object dist extends MillPublishJavaModule {
...
+      val classpath = runClasspath().map(_.path.toString)
+      launcherScript(
+        jvmArgs,
+        jvmArgs,
+        classpath,
+        Agg(pathingJar().path.toString) // TODO not working yet on Windows! see #791
+      )
+    }
+
+    os.write(outputPath, prependShellScript)
```
lefou, line 802: What's the motivation to make this local?

lefou (commented):
Great to have some test capabilities. I left some comments below.

bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
```diff
@@ -773,3 +775,14 @@ private class MillBuildServer(
     debug("onRunReadStdin is current unsupported")
   }
 }
+
+private object MillBuildServer {
+  private implicit class SeqGroupList[A](private val seq: Seq[A]) extends AnyVal {
+    def groupList[K, B](key: A => K)(f: A => B): Seq[(K, Seq[B])] = {
```
lihaoyi, line 781: Let's move this into a normal static method rather than an extension method. We don't use extension methods much in com-lihaoyi code unless they're super frequently used (e.g. FastParse's `~`), and `groupList` is used in only two places so definitely doesn't make the cut.

GIven that it's only used in two places, and seems pretty niche and specialized to these two callsites, we can probably also hard-code the `(_._2)(_._1)` thing as well

integration/ide/bsp-server/src/BspServerTests.scala
```diff
@@ -0,0 +1,164 @@
...
+              "org.scala-lang",
+              "scala-compiler",
+              scalaVersion
+            ))
+            .fetchResult()
+          scalaFetchRes.getDependencies.asScala
+            .filter(dep => dep.getModule.getOrganization != "org.scala-lang")
+            .map { dep =>
+              def basePath(version: String) =
+                s"${dep.getModule.getOrganization.split('.').mkString("/")}/${dep.getModule.getName}/$version/${dep.getModule.getName}-$version"
```
lihaoyi, line 40: This interpolated string is way too long and complicated, let's break it up into some local variables that we combined into a string at the end

testkit/src/mill/testkit/BspServerUtil.scala
```diff
@@ -0,0 +1,184 @@
...
+    def onBuildTaskFinish(params: b.TaskFinishParams): Unit = ()
+    def onBuildTaskProgress(params: b.TaskProgressParams): Unit = ()
+    def onBuildTaskStart(params: b.TaskStartParams): Unit = ()
+    def onRunPrintStderr(params: b.PrintParams): Unit = ()
+    def onRunPrintStdout(params: b.PrintParams): Unit = ()
+  }
+
+  lazy val dummyBuildClient: b.BuildClient = new DummyBuildClient {}
+
+  val gson: Gson = new GsonBuilder().setPrettyPrinting().create()
```
lihaoyi, line 36: I assume we use Gson here because that's what BSP uses?

testkit/src/mill/testkit/BspServerUtil.scala
```diff
@@ -0,0 +1,184 @@
...
+    }
+  )
+
+  trait MillBuildServer extends b.BuildServer with b.JvmBuildServer
+      with b.JavaBuildServer with b.ScalaBuildServer
+
+  def withBspServer[T](
+      workspacePath: os.Path,
+      millTestSuiteEnv: Map[String, String],
+      client: b.BuildClient = dummyBuildClient
```
lihaoyi, line 107: This parameter never seems to be passed; should we just hardcode it as `dummyBuildClient`?

testkit/src/mill/testkit/BspServerUtil.scala
```diff
@@ -0,0 +1,184 @@
...
+      proc.join(2000L)
+    }
+  }
+
+  lazy val millWorkspace: os.Path = {
+    val value = Option(System.getenv("MILL_PROJECT_ROOT")).getOrElse(???)
+    os.Path(value)
+  }
+
+  def replaceAllValues(
```
lihaoyi, line 168: This name is kind of generic and hard to figure out. How about `normalizePathsForTesting`?

testkit/src/mill/testkit/BspServerUtil.scala
```diff
@@ -0,0 +1,184 @@
...
+import mill.bsp.Constants
+import org.eclipse.{lsp4j => l}
+
+import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.{ExecutorService, Executors, ThreadFactory}
+
+import scala.jdk.CollectionConverters._
+import scala.reflect.ClassTag
+
+object BspServerUtil {
```
lihaoyi, line 16: Let's call this `BspServerTestUtil` so it is clear that this contains test fixtures and not application code helpers

testkit/src/mill/testkit/BspServerUtil.scala
```diff
@@ -0,0 +1,184 @@
...
+
+  val updateFixtures = false
+
+  lazy val bsp4jVersion: String = sys.props.getOrElse("BSP4J_VERSION", ???)
```
lihaoyi, line 20: We should just use a `val` or `def` here unless there is some explicit reason we want it to be lazy (e.g. due to side effecting initialization and/or performance overhead)

testkit/src/mill/testkit/BspServerUtil.scala
```diff
@@ -0,0 +1,184 @@
...
+    def onBuildShowMessage(params: b.ShowMessageParams): Unit = ()
+    def onBuildTargetDidChange(params: b.DidChangeBuildTarget): Unit = ()
+    def onBuildTaskFinish(params: b.TaskFinishParams): Unit = ()
+    def onBuildTaskProgress(params: b.TaskProgressParams): Unit = ()
+    def onBuildTaskStart(params: b.TaskStartParams): Unit = ()
+    def onRunPrintStderr(params: b.PrintParams): Unit = ()
+    def onRunPrintStdout(params: b.PrintParams): Unit = ()
+  }
+
+  lazy val dummyBuildClient: b.BuildClient = new DummyBuildClient {}
```
lihaoyi, line 34: The convention we've used elsewhere in the codebase is `object DummyBuildClient extends DummyBuildClient`

testkit/src/mill/testkit/BspServerUtil.scala
```diff
@@ -0,0 +1,184 @@
...
+    def doReplaceAll(input: String, inverse: Boolean = false): String =
+      replaceAll.foldLeft(input) {
+        case (input0, (from0, to0)) =>
+          val (from, to) = if (inverse) (to0, from0) else (from0, to0)
+          input0.replace(from, to)
+      }
+
+    val exists = os.exists(fixturePath)
+    val readOpt = Option.when(exists) {
+      val bytes = os.read.bytes(fixturePath)
```
lihaoyi, line 52: We can just `os.read` rather than `os.read.bytes`, then we don't need to `new String` later

testkit/src/mill/testkit/BspServerUtil.scala
```diff
@@ -0,0 +1,184 @@
...
+  ): Unit = {
+
+    def doReplaceAll(input: String, inverse: Boolean = false): String =
+      replaceAll.foldLeft(input) {
+        case (input0, (from0, to0)) =>
+          val (from, to) = if (inverse) (to0, from0) else (from0, to0)
+          input0.replace(from, to)
+      }
+
+    val exists = os.exists(fixturePath)
```
lihaoyi, line 50: I assume `exists -== false` only when generating fixtures for the first time; this should go in a comment or something because it's not obvious

integration/ide/bsp-server/src/BspServerTests.scala
```diff
@@ -0,0 +1,164 @@
...
+        val buildTargetOutputPaths = buildServer
+          .buildTargetOutputPaths(new b.OutputPathsParams(targetIds))
+          .get()
+        compareWithGsonFixture(
+          buildTargetOutputPaths,
+          fixturePath / "build-targets-output-paths.json",
+          replaceAll = replaceAll
+        )
+
+        val buildTargetJvmRunEnvironments = buildServer
```
lihaoyi, line 117: Let's inline all these `val`s. Right now we have to manually keep track of them and make sure we pass each one to the right `compareWithGsonFixture` call, whereas if inlined then the 1:1 relationship is clear by construction.

lihaoyi:
Left some comments, overall looks reasonable I think

testkit/src/mill/testkit/BspServerUtil.scala
```diff
@@ -0,0 +1,184 @@
...
+    val readOpt = Option.when(exists) {
```
lihaoyi, line 51: This name needs to be better. What did we read? Maybe `fixtureValueOpt`?

testkit/src/mill/testkit/BspServerUtil.scala
```diff
@@ -0,0 +1,184 @@
...
+    else if (updateFixtures) {
+      // Fixture on disk might need to be updated anyway, if replaceAll changed
+      // and new strings should be replaced
+      val jsonStr = doReplaceAll(
+        gson.toJson(
+          value,
+          implicitly[ClassTag[T]].runtimeClass
+        )
+      )
+      val onDiskJsonStr = os.read(fixturePath)
```
lihaoyi, line 82: How about `actualJsonStr` and `expectedJsonStr`?

testkit/src/mill/testkit/BspServerUtil.scala
```diff
@@ -0,0 +1,184 @@
...
+  trait MillBuildServer extends b.BuildServer with b.JvmBuildServer
+      with b.JavaBuildServer with b.ScalaBuildServer
+
+  def withBspServer[T](
+      workspacePath: os.Path,
+      millTestSuiteEnv: Map[String, String],
+      client: b.BuildClient = dummyBuildClient
+  )(f: (MillBuildServer, b.InitializeBuildResult) => T): T = {
+
+    val jsonFile = workspacePath / Constants.bspDir / s"${Constants.serverName}.json"
```
lihaoyi, line 110: `jsonFile` needs a better name. `serverMetadataFile`?

lihaoyi:
Using the term `fixture` seems a weird here. When I think of "test fixtures", what comes to mind is the helper methods and utils code that helps set up a tear down the test, whereas in this case you are using it to refer to the actual expected test data. Can we find a different name for these things, perhaps calling them `expectedResponses` or something?

bsp/worker/src/mill/bsp/worker/SyntheticRootBspBuildTargetData.scala
```diff
@@ -15,11 +14,11 @@ import ch.epfl.scala.bsp4j.BuildTarget
  */
 class SyntheticRootBspBuildTargetData(topLevelProjectRoot: os.Path) {
   val id: BuildTargetIdentifier = new BuildTargetIdentifier(
-    Utils.sanitizeUri(topLevelProjectRoot / s"synth-build-target-${UUID.randomUUID()}")
+    Utils.sanitizeUri(topLevelProjectRoot / "mill-synthetic-root-target")
```
lihaoyi, line 17: We're removing the UUID here. Do we know that it is not necessary?

testkit/src/mill/testkit/BspServerUtil.scala
```diff
@@ -0,0 +1,184 @@
...
+          new b.BuildClientCapabilities(List("java", "scala").asJava)
+        )
+      ).get()
+
+      f(buildServer, initRes)
+    } finally {
+      proc.stdin.close()
+      proc.stdout.close()
+
+      proc.join(2000L)
```
lihaoyi, line 159: 2s seems a bit of an odd timeout; it's short enough that it could conceivably cause flakiness in test under CPU/memory/GC pressure, but long enough it won't during development when tests are typically run alone. If we just want to make sure it doesn't hang forever, maybe a longer timeout like 10s or 30s? That should be long enough it will only trigger if something is really-truly stuck.

bsp/worker/src/mill/bsp/worker/State.scala
```diff
@@ -7,13 +7,13 @@ import mill.define.Module
 import mill.eval.Evaluator
 
 private class State(workspaceDir: os.Path, evaluators: Seq[Evaluator], debug: String => Unit) {
-  lazy val bspModulesById: Map[BuildTargetIdentifier, (BspModule, Evaluator)] = {
+  lazy val bspModulesIdList: Seq[(BuildTargetIdentifier, (BspModule, Evaluator))] = {
```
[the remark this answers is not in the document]
alexarchambault: Yes. Without this, the order of build targets in responses to `workspaceBuildTargets` would sometimes change over time.

build.mill
alexarchambault: Right, that's a debug leftover

bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
alexarchambault: Sure

testkit/src/mill/testkit/BspServerUtil.scala
alexarchambault: Yes. All bsp4j classes get serialized / deserialized well out-of-the-box with Gson.

alexarchambault: And bsp4j itself pulls gson in the class path.

testkit/src/mill/testkit/BspServerUtil.scala
alexarchambault: It can. Having it could be useful for more advanced tests, but the parameter can always be added back at that time.

testkit/src/mill/testkit/BspServerUtil.scala
alexarchambault: It's not only paths, it's paths and Java version. What about `normalizeLocalValuesForTesting`?

testkit/src/mill/testkit/BspServerUtil.scala
alexarchambault: Sure

testkit/src/mill/testkit/BspServerUtil.scala
alexarchambault: Sure (although bin compat might be harder to preserve with the `object`)

testkit/src/mill/testkit/BspServerUtil.scala
alexarchambault: Or `expectedValueOpt`, so that we avoid the term "fixture"?

alexarchambault:
> Using the term `fixture` seems a weird here. When I think of "test fixtures", what comes to mind is the helper methods and utils code that helps set up a tear down the test, whereas in this case you are using it to refer to the actual expected test data. Can we find a different name for these things, perhaps calling them `expectedResponses` or something?

In [snapshot4s](https://github.com/siriusxm/snapshot4s), they call those… "snapshots"

bsp/worker/src/mill/bsp/worker/SyntheticRootBspBuildTargetData.scala
alexarchambault: My guess is that this avoid conflicts with an actual module users created. But a collision seems quite unlikely.

build.mill
alexarchambault: `prependShellScript` was a task only used the `launcher` task. I needed the output of `prependShellScript` to hard-code (in a `milll.main.cli` Java property) the path of the launcher generated by `launcher`. Somehow, I needed `launcher` to pass an argument to `prependShellScript` (an argument involving `T.dest`), so I initially inlined `prependShellScript`. The new push makes it stop being a task, and makes it a standard method accepting arguments instead.

testkit/src/mill/testkit/BspServerUtil.scala
alexarchambault: Renamed it to `bspMetadataFile` in the latest push

testkit/src/mill/testkit/BspServerTestUtil.scala
```diff
@@ -0,0 +1,186 @@
+package mill.testkit
```
lihaoyi, line 1: `testkit` is meant for things that we publish for the rest of the world to use. Given that this stuff is for testing our own BSP implementation, it should probably just live in that module's test suite.

lihaoyi (approved):
Left one more comment but otherwise looks reasonable

integration/ide/bsp-server/src/BspServerTests.scala
```diff
@@ -0,0 +1,158 @@
...
+import scala.jdk.CollectionConverters._
+
+object BspServerTests extends UtestIntegrationTestSuite {
+  def snapshotsPath: os.Path =
+    super.workspaceSourcePath / "snapshots"
+  override protected def workspaceSourcePath: os.Path =
+    super.workspaceSourcePath / "project"
+
+  def tests: Tests = Tests {
+    test("BSP install") - integrationTest { tester =>
```
lihaoyi, line 19: We should rename this to `snapshotTests` or something

alexarchambault: Or `requests`? This tests that the main BSP requests work fine. `project structure` might work too, this tests the "descriptive" BSP requests, that expose the project structure somehow.

lihaoyi: `requestSnapshotTests`? I think being a bit verbose here is fine, since the suite is definitely doing something unusual

testkit/src/mill/testkit/BspServerTestUtil.scala
alexarchambault: Sure. I was thinking this might be useful in other tests (multi-module ones maybe), but we can always move that elsewhere later if needed.

integration/ide/bsp-server/src/BspServerTests.scala
alexarchambault: That would be `requestSnapshots` in the former case, right? No need of the `Test` suffix

testkit/package.mill
```diff
@@ -3,13 +3,18 @@ package build.testkit
 import mill._
 
 object `package` extends RootModule with build.MillPublishScalaModule {
-  def moduleDeps = Seq(build.main.eval, build.main.util, build.main)
-  def ivyDeps = Agg(build.Deps.TestDeps.utest)
+  def moduleDeps = Seq(build.bsp, build.main.eval, build.main.util, build.main)
+  def ivyDeps = Agg(
+    build.Deps.bsp4j,
+    build.Deps.TestDeps.utest
+  )
```
lefou, line 10: Why got `testkit` new dependencies? I can't see how this is needed.

integration/ide/bsp-server/src/BspServerTests.scala
alexarchambault: Renamed them the test to `requestSnapshots`

bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
```diff
@@ -773,3 +772,12 @@ private class MillBuildServer(
     debug("onRunReadStdin is current unsupported")
   }
 }
+
+private object MillBuildServer {
+  private def groupList[A, K, B](seq: Seq[A])(key: A => K)(f: A => B): Seq[(K, Seq[B])] = {
```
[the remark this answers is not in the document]
alexarchambault: Should be fixed in the latest push

testkit/package.mill
alexarchambault: Should be fixed now

lefou:
@alexarchambault It would really help if you could just push changes by new commits. If it's so important to you to rebase, please do it in the end, after we reviewed the pull request and approved it. We will most likely squash then anyways, so what's the point? 

And finally, if anyone really want's to `--force` a `git push`, please use `--force-with-lease` instead. Otherwise you ruin your co-workers `git` experience, and I think also GitHub review UI suffers a bit. I sometimes can't finish my started review, because of the fact the commits no longer exists. That's not fun.

alexarchambault:
> It would really help if you could just push changes by new commits. If it's so important to you to rebase

I wasn't aware of force pushes being a hindrance to started reviews. (But wouldn't waiting for the CI to be green make that unlikely to happen?)

Will try to avoid force push in the future (by pushing more "fixup" commits I guess). Rebasing helps me make sense of what I do. (In my own projects, I don't mind force pushes that allow to have a clean git history, quite the opposite. I find it much easier to review the final result via a clean git history than reviewing "diffs-of-diffs", that things such as [this](https://github.com/com-lihaoyi/mill/compare/2def6ea4179a71d397304fec39e3bfcd63469c51..324b5f01ff6099caf038e2ca14d7e863089902c1) are, especially for non trivial PRs)

I've always been using `--force-with-lease` though.

lihaoyi:
Once this is green we can merge it and rebase all the other downstream PRs for easier reviewing, I expect those will be easy to review/merge as well

alexarchambault:
(If the CI is green, feel free to review, and possibly un-draft and merge)

build.mill
```diff
@@ -795,7 +795,15 @@ object dist extends MillPublishJavaModule {
...
-    os.write(outputPath, prependShellScript())
+    val prependShellScript0 = prependShellScript(
+      T.dest / "prependShellScript",
+      forkArgs(),
+      runClasspath().map(_.path),
+      pathingJar().path,
+      outputPath
+    )
+
+    os.write(outputPath, prependShellScript0)
```
lihaoyi, line 806: @alexarchambault what are these changes about? They don't seem related to BSP integration tests and their purpose doesn't seem obvious at a glance

integration/package.mill
```diff
@@ -71,9 +71,17 @@ object `package` extends RootModule {
...
       extends Cross[IntegrationCrossModule](build.listIn(millSourcePath / "invalidation"))
-  object ide extends Cross[IntegrationCrossModule](build.listIn(millSourcePath / "ide"))
+  object ide extends Cross[Ide](build.listIn(millSourcePath / "ide"))
   trait IntegrationCrossModule extends build.MillScalaModule with IntegrationTestModule {
     override def moduleDeps = super[IntegrationTestModule].moduleDeps
+    def forkEnv = super.forkEnv() ++ Seq(
+      "MILL_PROJECT_ROOT" -> T.workspace.toString
+    )
+  }
+  trait Ide extends IntegrationCrossModule {
```
lihaoyi, line 81: Let's call this `IdeIntegrationCrossModule` just to be a bit more verbose and follow the existing convention

build.mill
alexarchambault: That's needed for [that step](https://github.com/com-lihaoyi/mill/blob/8b063c48d76837870af188ad48500bb739668f2b/bsp/src/mill/bsp/BSP.scala#L79-L83) of `mill.bsp.BSP/install` to work. This ensures that `mill.main.cli` has a valid value (path to the launcher), so that spawning a server from `.bsp/mill-bsp.json` works. Without this, `.bsp/mill-bsp.json` has `java.class.path` as executable, which doesn't work in the test (as that value is a `:`-separated list of JARs and directories, and not the launcher - the "pathing JARs" isn't there, only its list of JARs / directories)

integration/package.mill
alexarchambault: Sure

build.mill
lihaoyi: If we want to change where `mill.main.cli` points to, could we override it by passing in the `$MILL_MAIN_CLI` CLI variable? That would be something we can do directly in the integration test without needing to patch the launcher script

lihaoyi: That would let us keep `prependShellScript` as a `Task`, which would reduce the diff and keep things more easily inspectable

alexarchambault: Sure, the latest push does that

testkit/src/mill/testkit/IntegrationTester.scala
```diff
@@ -91,7 +91,11 @@ object IntegrationTester {
       )
     }
 
-    def millTestSuiteEnv: Map[String, String] = Map("MILL_TEST_SUITE" -> this.getClass().toString())
+    def millTestSuiteEnv: Map[String, String] =
+      Map("MILL_TEST_SUITE" -> this.getClass().toString()) ++
+        sys.props.get("TEST_SCALA_2_13_VERSION")
+          .map("TEST_SCALA_2_13_VERSION" -> _)
+          .toSeq
```
lihaoyi, line 98: Rather than patching this here, can we duplicate `TEST_SCALA_2_13_VERSION` as a `forkEnv` in `integration/package.mill`? This seems like a Mill-internal concern and shouldn't need to be published for external users of `mill-testkit`

alexarchambault: Should be fixed

lihaoyi: looks good, once it's green let's land this

PR #3614, opened 2024-09-27
Use coursier thread pool to run coursier-related task

alexarchambault:
Using the default global `ExecutionContext`, also used by the BSP server, can lead to deadlocks, as coursier doesn't mark blocking things with `blocking`. (It deadlocks locally for me, during Mill import via BSP in IntelliJ, with 4 threads being spawned and awaiting in the default `ExecutionContext`.)


alexarchambault:
(Never use the global `ExecutionContext`… 😬)

lihaoyi (approved):

Issue #3566, opened 2024-09-17
Rename `def test` to `def testUncached` to allow `__.test` to be parallel?


lihaoyi:
Otherwise we cannot set `def defaultCommandName` to `testCached`, because people who run `__.test` will pick up both the `foo.test.testUncached` default command name and also the `foo.test.test` test command.

TBH not quite sure what the right thing to do here is

This is a binary incompatible change and will need to go into 0.13.0

lolgab:
The biggest problem with this change is that now to pass arguments to tests you need to add a `.testUncached` at the end and know the difference between `testCached` and `testUncached`.
Since tests are very often not pure (flaky, etc.) I think they should not be cached by default.
If I know the whole story, the problem with the current `test` command is that commands are now run sequentially.
Have you thought about adding some flag to commands to make them parallel?
It could be an annotation, or a special `T.parallelCommand` method.

lefou:
I don't think we should explicitly request parallelism via the `T` API. Instead, we should demarcate commands that need serial or interactive execution if we think this is required.

The run commands last and sequentially should be configurable via a CLI option or some other settings. It's already configurable in the evaluator.

lefou:
We closed this, without renaming `test` to `testUncached`, since with #3617 we essentially run the `test` command in parallel to other targets now.

PR #3617, opened 2024-09-27
Introduce `Task.Command(exclusive = true)` and convert `Task.Persistent` to `Task(persistent = true)`

lihaoyi:
fixes https://github.com/com-lihaoyi/mill/issues/3566. Mostly a straightforward implementation of what was discussed.

We use `Task.Command(exclusive = true)` to make `console`/`repl`/`clean`/etc. run serially, all other commands should run in parallel. That includes most `test` commands. The name `exclusive` is taken from the Bazel action/target tag with the same meaning: that the tagged action runs alone with no other actions in parallel

`Task.Persistent` was changed to `Task(persistent = true)` for consistency, and also for other reasons: the new syntax is more composable, e.g. a user can more easily choose whether they want `persistent = true` or `persistent = false` based on a computed `Boolean` value, and Mill can in future extend it such that we can have `Task(exclusive = true, persistent = true)`, `Task.Command(exclusive = true, persistent = true)`, or other such combinations perhaps with even more flags (e.g. Bazel has a pretty long list https://bazel.build/reference/be/common-definitions#common.tags). 

To make overload resolution work correctly, I make the first parameter of the `(exclusive = true)` or `(persistent = true)` parameter list `dummy: NamedParameterOnlyDummy.type = NamedParameterOnlyDummy`. This ensures that there is no normal value the user could pass in positionally that would select that overload of `def Command` or `def apply`, and the only way to select it is by passing in `exclusive = true` or `persistent = true` as a named parameter

Tested manually by adding `println`s and making sure that `leafSerialCommands` no longer contains test tasks when I run tests in `example/scalalib/basic/1-simple`.

### DIFF
--- a/main/define/src/mill/define/NamedParameterOnlyDummy.scala
+++ b/main/define/src/mill/define/NamedParameterOnlyDummy.scala
@@ -3,4 +3,4 @@ package mill.define
 /**
  * Dummy class used to mark parameters that come after it as named only parameters
  */
-object NamedParameterOnlyDummy
+class NamedParameterOnlyDummy private[mill] ()
--- a/main/define/src/mill/define/NamedParameterOnlyDummy.scala
+++ b/main/define/src/mill/define/NamedParameterOnlyDummy.scala
@@ -1,0 +1,6 @@
+package mill.define
+
+/**
+ * Dummy class used to mark parameters that come after it as named only parameters
+ */
+object NamedParameterOnlyDummy
--- a/main/define/src/mill/define/Task.scala
+++ b/main/define/src/mill/define/Task.scala
@@ -123,6 +123,15 @@ object Task extends TaskBase {
   ): Command[T] = macro Target.Internal.commandImpl[T]
 
   /**
+   * [[ParallelCommand]] is a variant of [[Command]] that is safe to run in parallel
+   */
+  def ParallelCommand[T](t: Result[T])(implicit
+      w: W[T],
+      ctx: mill.define.Ctx,
+      cls: EnclosingClass
+  ): Command[T] = macro Target.Internal.parallelCommandImpl[T]
+
+  /**
    * [[Worker]] is a [[NamedTask]] that lives entirely in-memory, defined using
    * `Task.Worker{...}`. The value returned by `Task.Worker{...}` is long-lived,
    * persisting as long as the Mill process is kept alive (e.g. via `--watch`,
@@ -521,6 +530,27 @@ object Target extends TaskBase {
       )
     }
 
+    def parallelCommandImpl[T: c.WeakTypeTag](c: Context)(t: c.Expr[T])(
+        w: c.Expr[W[T]],
+        ctx: c.Expr[mill.define.Ctx],
+        cls: c.Expr[EnclosingClass]
+    ): c.Expr[Command[T]] = {
+      import c.universe._
+
+      val taskIsPrivate = isPrivateTargetOption(c)
+
+      reify(
+        new Command[T](
+          Applicative.impl[Task, T, mill.api.Ctx](c)(t).splice,
+          ctx.splice,
+          w.splice,
+          cls.splice.value,
+          taskIsPrivate.splice,
+          parallelizable = true
+        )
+      )
+    }
+
     def workerImpl1[T: c.WeakTypeTag](c: Context)(t: c.Expr[Task[T]])(ctx: c.Expr[mill.define.Ctx])
         : c.Expr[Worker[T]] = {
       import c.universe._
@@ -693,8 +723,16 @@ class Command[+T](
     val ctx0: mill.define.Ctx,
     val writer: W[_],
     val cls: Class[_],
-    val isPrivate: Option[Boolean]
+    val isPrivate: Option[Boolean],
+    val parallelizable: Boolean
 ) extends NamedTask[T] {
+  def this(
+      t: Task[T],
+      ctx0: mill.define.Ctx,
+      writer: W[_],
+      cls: Class[_],
+      isPrivate: Option[Boolean]
+  ) = this(t, ctx0, writer, cls, isPrivate, false)
   override def asCommand: Some[Command[T]] = Some(this)
   // FIXME: deprecated return type: Change to Option
   override def writerOpt: Some[W[_]] = Some(writer)
--- a/main/define/src/mill/define/Task.scala
+++ b/main/define/src/mill/define/Task.scala
@@ -579,7 +579,7 @@ object Target extends TaskBase {
           w.splice,
           cls.splice.value,
           taskIsPrivate.splice,
-          serial = c.prefix.splice.asInstanceOf[Task.CommandFactory].exclusive
+          exclusive = c.prefix.splice.asInstanceOf[Task.CommandFactory].exclusive
         )
       )
     }
@@ -762,7 +762,7 @@ class Command[+T](
     val writer: W[_],
     val cls: Class[_],
     val isPrivate: Option[Boolean],
-    val serial: Boolean
+    val exclusive: Boolean
 ) extends NamedTask[T] {
   def this(
       t: Task[T],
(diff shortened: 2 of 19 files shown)


lefou:
I'd prefer to explicitly denote commands that are not parallel capable, e.g. `SerialCommand` or `InteractiveCommand`. Since in Mill all other targets are suspected to run in parallel, having a `ParallelCommand` is irritating since it's pointing out the perceived normal ("beeing parallel"). Instead, we should explicitly point out the deviation from the rule.

Even in older Mill versions, we run all commands in parallel if the `-j` is different from `1`.

lefou:
It would be nice, if we could avoid the introduction of yet another target factory. Maybe, we could make it configurable via  a parameter, e.g. `Task.Command(parallel = false) { ...}`. Don't know if the compiler can handle to overload and the implicit conversion at the same time, but it would be nice. Or we could make it `Task.Command.serial { ... }`.

lihaoyi:
@lefou I think `Task.Command(serial = true)` could be made to work, I originally went with a Task.ParallelCommand factory method because that's how we've handled things like T.persistent in the past (rather than `T(persistent = true)`). But maybe we can cut over to a flag-based approach since we're overhauling the syntax for 0.12.0 anyway

lefou:
> @lefou I think `Task.Command(serial = true)` could be made to work, I originally went with a Task.ParallelCommand factory method because that's how we've handled things like T.persistent in the past (rather than `T(persistent = true)`). But maybe we can cut over to a flag-based approach since we're overhauling the syntax for 0.12.0 anyway. And it will scale better as more flags appear since you can mix and match the flags for any particular callsite

If we're doing this, we should review all target factories like `Source`, `Sources` and `Input`. But we should discuss this separately.

lihaoyi:
@lefou I've updated the PR, the implementation to make it work is a bit hacky but seems to work and compiles/passes-tests

lihaoyi:
I think `Source`, `Sources`, `Input`, etc. don't need support for flags yet. The only reason we need flags is if we have some behavioral change that does not change the type signature, as any change that affects the type signature has to be a new method. So `persistent`, `exclusive`, etc. can be flags, but changing from `Source` to `Sources` or `Task` to `Task.Anon` has to be a different method with a different signature

In theory we may find it useful to have `T.Command(exclusive = true, persistent = true)` or `T(exclusive = true, persistent = true)` and other such combinations, but I'll leave that for future follow ups to implement

main/define/src/mill/define/Task.scala
```diff
@@ -693,8 +761,16 @@ class Command[+T](
     val ctx0: mill.define.Ctx,
     val writer: W[_],
     val cls: Class[_],
-    val isPrivate: Option[Boolean]
+    val isPrivate: Option[Boolean],
+    val serial: Boolean
```
lefou, line 765:
```suggestion
    val exclusive: Boolean
```

main/define/src/mill/define/Task.scala
```diff
@@ -521,6 +563,27 @@ object Target extends TaskBase {
...
+      val taskIsPrivate = isPrivateTargetOption(c)
+
+      reify(
+        new Command[T](
+          Applicative.impl[Task, T, mill.api.Ctx](c)(t).splice,
+          ctx.splice,
+          w.splice,
+          cls.splice.value,
+          taskIsPrivate.splice,
+          serial = c.prefix.splice.asInstanceOf[Task.CommandFactory].exclusive
```
lefou, line 582:
```suggestion
          exclusive = c.prefix.splice.asInstanceOf[Task.CommandFactory].exclusive
```

lefou:
I'd like to hear what @lolgab thinks about this new notation.

main/define/src/mill/define/NamedParameterOnlyDummy.scala
```diff
@@ -0,0 +1,6 @@
+package mill.define
+
+/**
+ * Dummy class used to mark parameters that come after it as named only parameters
+ */
+object NamedParameterOnlyDummy
```
bishabosha, line 6: perhaps a SIP idea: an annotation that forces callsites to use named parameters?

bishabosha: or some new syntax at the declaration anyway

main/define/src/mill/define/Task.scala
```diff
@@ -122,6 +106,18 @@ object Task extends TaskBase {
       cls: EnclosingClass
   ): Command[T] = macro Target.Internal.commandImpl[T]
 
+  def Command(
+               t: NamedParameterOnlyDummy.type = NamedParameterOnlyDummy,
+               exclusive: Boolean = false
+  ): CommandFactory = new CommandFactory(exclusive)
+  class CommandFactory private[mill] (val exclusive: Boolean) extends TaskBase.TraverseCtxHolder {
```
bishabosha, line 113: why shouldn't this be a curried function?

lihaoyi: IIRC Scala 2 macros had an issue where they didnt allow you to call their parameters with explicit names. Actually I'm not sure if that still applies in 2.13.15 or 3.5.0, but it did at some point. Lemme try again to verify

bishabosha: works in scala 3.5 anyway:
```scala
scala> def mcr[T: Type](fn: Expr[Either[String, T]], exclusive: Expr[Boolean])(using Quotes): Expr[Either[String, T]] = {
     |   import scala.concurrent.Future
     |   import scala.concurrent.ExecutionContext.Implicits.global
     |   import scala.concurrent.Await
     |   import scala.concurrent.duration.Duration
     |
     |   val exc = exclusive.valueOrAbort
     |   if exc then '{ println("EXCLUSIVE"); $fn }
     |   else '{ val fut = Future({Thread.sleep(1000); $fn}); Await.result(fut, Duration.Inf) }
     | }
def mcr
  [T]
    (fn: scala.quoted.Expr[Either[String, T]], exclusive:
      scala.quoted.Expr[Boolean])
      (implicit evidence$1: scala.quoted.Type[T], x$3: scala.quoted.Quotes):
        scala.quoted.Expr[Either[String, T]]

scala> {
     | inline def task[T](inline op: Either[String, T]) = ${ mcr[T]('op, 'false) }
     | inline def task[T](ignore: Int = -1, inline exclusive: Boolean = false)(inline op: Either[String, T]) = ${ mcr[T]('op, 'exclusive) }
     | }
def task[T](op: Either[String, T]): Either[String, T]
def task
  [T](ignore: Int, exclusive: Boolean)(op: Either[String, T]): Either[String, T]
```

lihaoyi: Yeah looks like it's still a limitation in 2.13.15

```scala
[14186] [error] /Users/lihaoyi/Github/mill/main/src/mill/main/MainModule.scala:403:17: macro applications do not support named and/or default arguments
[14186] [error]     Task.Command(exclusive = true) {
[14186] [error]                 ^
```

So likely we'll go with the boilerplate-y `CommandFactory` so this can land in 0.12.0, and in 0.13.0 we can convert to normal curried functions as one of the benefits of being on Scala 3.x

main/define/src/mill/define/NamedParameterOnlyDummy.scala
lihaoyi: Yeah that's a good idea!

lefou (approved):

main/eval/src/mill/eval/EvaluatorCore.scala
```diff
@@ -180,8 +180,13 @@ private[mill] trait EvaluatorCore extends GroupEvaluator {
...
 
     val tasksTransitive = tasksTransitive0.toSet
-    val (tasks, leafCommands) = terminals0.partition {
-      case Terminal.Labelled(t, _) if tasksTransitive.contains(t) => true
+    val (tasks, leafSerialCommands) = terminals0.partition {
+      case Terminal.Labelled(t, _) =>
+        if (tasksTransitive.contains(t)) true
+        else t match {
+          case t: Command[_] => !t.exclusive
+          case _ => false
```
lefou, line 188: Shouldn't this be a `true` or a `!serialCommandExec`? 

Maybe, `serialCommandExec` isn't need anymore, since we now have fine control over exclusive execution.

PR #3620, opened 2024-09-27
Add Ktfmt support

### LINKED ISSUE #3612: Add KtFmt support (500USD Bounty)
---

---

This ticket is to add support for Kotlin https://github.com/facebook/ktfmt, in the spirit of our existing support for `mill.javalib.PalantirFormatModule` and `mill.scalalib.scalafmt.ScalaFmtModule`. We already have KtLint support which does support autoformatting, but KtFmt seems pretty popular so we should support it too.

Like our support for `PalantirFormat` or `ScalaFmt`, 

* KtFmt support should come in both a trait that can be mixed into `KotlinModule`s, as well as a standalone external module that can be used on any existing `KotlinModule`s unchanged. 
* Should include unit tests and an example test in `example/kotlinlib/linting/` that can be included in the documentation
* Should expose any relevant configuration on KtFmt that the user may want to override

0xnm:
Fixes #3612.

This PR is based on the logic done in #3531, so it looks quite similar.

### DIFF
--- a/kotlinlib/src/mill/kotlinlib/contrib/ktfmt/KtfmtModule.scala
+++ b/kotlinlib/src/mill/kotlinlib/contrib/ktfmt/KtfmtModule.scala
@@ -1,0 +1,139 @@
+package mill
+package kotlinlib.contrib.ktfmt
+
+import mainargs.Leftover
+import mill.api.{Loose, PathRef}
+import mill.define.{Discover, ExternalModule}
+import mill.kotlinlib.{DepSyntax, KotlinModule}
+import mill.main.Tasks
+import mill.util.Jvm
+
+trait KtfmtBaseModule extends KotlinModule {
+
+  /**
+   * Classpath for running Ktfmt.
+   */
+  def ktfmtClasspath: T[Loose.Agg[PathRef]] = T {
+    defaultResolver().resolveDeps(
+      Agg(ivy"com.facebook:ktfmt:${ktfmtVersion()}")
+    )
+  }
+
+  /**
+   * Ktfmt version.
+   */
+  def ktfmtVersion: T[String] = T {
+    "0.52"
+  }
+
+  /**
+   * Additional arguments for Ktfmt. Check
+   * [[https://github.com/facebook/ktfmt/blob/main/core/src/main/java/com/facebook/ktfmt/cli/ParsedArgs.kt#L51 available options]].
+   */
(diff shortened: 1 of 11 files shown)


lihaoyi:
Thanks @0xnm ! I'll transfer you the bounty using your earlier details

kotlinlib/src/mill/kotlinlib/contrib/ktfmt/KtfmtModule.scala
```diff
@@ -0,0 +1,139 @@
...
+    defaultResolver().resolveDeps(
+      Agg(ivy"com.facebook:ktfmt:${ktfmtVersion()}")
+    )
+  }
+
+  /**
+   * Ktfmt version.
+   */
+  def ktfmtVersion: T[String] = T {
+    "0.52"
```
lefou, line 26: If we want to provide a default at all, we should at least source it from a `BuildInfo` generated at build/release time. That way, we can automate bumping this dependency as part of Mill's version management.

kotlinlib/src/mill/kotlinlib/contrib/ktfmt/KtfmtModule.scala
```diff
@@ -0,0 +1,139 @@
...
+      _sources,
+      ktfmtClasspath(),
+      ktfmtOptions()
+    )
+  }
+}
+
+object KtfmtModule extends ExternalModule with KtfmtBaseModule with TaskModule {
+
+  def kotlinVersion = "1.9.24"
```
lefou, line 71: I don't like this hardcoded version here. We should source it from some generated `BuildInfo`.

kotlinlib/test/src/mill/kotlinlib/contrib/ktfmt/KtfmtModuleTests.scala
```diff
@@ -0,0 +1,150 @@
...
+  def afterFormat(
+      moduleRoot: os.Path,
+      style: String = "kotlin",
+      format: Boolean = true,
+      removeUnusedImports: Boolean = true,
+      sources: Seq[String] = Seq.empty
+  ): Seq[os.Path] = {
+
+    object module extends TestBaseModule with KotlinModule with KtfmtModule {
+      override def kotlinVersion: T[String] = "1.9.24"
```
lefou, line 97: The version used here should be defined at the top of the test file. Same, for the other occurences.

Issue #3611, opened 2024-09-26
Add Kotlin Web (Ktor & KotlinJS) examples (1650USD Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 1650USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

The goal of this ticket is to implement support for `example/kotlinlib/web` example tests for using Kotlin for web development, mirroring the `example/javalib/web` and `example/scalalib/web`, and support for KotlinJS in the spirit of `scalajslib/`. I would want the following example tests, broken down into smaller groups each with its own bounty:

650USD: 

* [x] `example/kotlinlib/web/1-hello-ktor`
    * A hello-world example using Kotlin and KTor to set up a web server serving a single `"<h1>Hello, World!</h1>"` web page, in the spirit of `example/javalib/web/2-hello-spring-boot`
* [x] `example/kotlinlib/web/2-todo-ktor`
    * An implementation of the common TodoMVC demo application, including HTML templating, JS/CSS resources, database access and unit/integration/`runBackground+curl` tests, in the spirit of `example/javalib/web/3-todo-spring-boot`

650USD: 

* [x] `example/kotlinlib/web/3-kotlinjs-module` (325USD)
    * How to define a simple KotlinJS module that can be compiled to Javascript, testing, and run using `node`, in the spirit of `example/scalalib/web/3-scalajs-module`
* [x] `example/kotlinlib/web/4-webapp-kotlinjs` (325USD)
    * How to wire up a KotlinJS module with a Kotlin/Ktor server to serve the generated Javascript, in the spirit of `example/scalalib/web/4-webapp-scalajs`

700USD: 
* [x] `example/kotlinlib/web/5-webapp-kotlinjs-shared` (350USD)
    * How to implement code sharing between a JVM KTor server and a KotlinJS javascript module, such that common code and utilities can be used by both. In the spirit of `example/scalalib/web/5-webapp-scalajs-shared`, 
* [ ] ~`example/kotlinlib/web/6-cross-platform-publishing` (350USD)~
    * ~How to define a library with both Kotlin-JVM and Kotlin-JS versions and publishing it to maven central. In the spirit of `example/scalalib/web/6-cross-version-platform-publishing`~
    * Moving this to a separate bounty

[...]

PR #3621, opened 2024-09-27
Add Kotlin Ktor Hello world example

javimartinez:
Part of #3611

Working in the TodoMVC demo application. I'll raise a follow up PR for that.


lihaoyi:
Looks good @javimartinez, looking forward to the TodoMVC application!

PR #3622, opened 2024-09-28
use Scalafmt latest version (won't build with Java 8)

bishabosha:
Scalafmt needs to be updated to support the latest scala 3 syntax.

according to the now-deleted comment, this version was frozen because of supporting building with JDK 8, but if 0.12.0 will drop JDK 8 support then why not update


lihaoyi (approved):
Let's do it. Looks like you just need to re-run the autoformatter to bring the code up to date to the latest scalafmt version

Issue #3610, opened 2024-09-26
Add Kotlinx-Kover support (500USD Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 500USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

The goal of this ticket is to add `example/kotlinlib/linting/3-kotlinx-kover` example tests for using https://github.com/Kotlin/kotlinx-kover to instrument a `KotlinModule`'s tests with code coverage tracking, in the spirit of `example/scalalib/linting/2-contrib-scoverage` or https://github.com/lefou/mill-jacoco. 

It should implement a `trait KotlinxKoverModule` in `kotlinlib/` that can be mixed in to any `KotlinModule` and provide `KotlinxKoverTests` that enable code coverage support, unit tests, and an example test folder that can be included in the documentation demonstrating its usage

PR #3624, opened 2024-09-29
Add Kover support

0xnm:
This PR adds Kover support declared in #3610.

Overall functionality is similar to what `ScoverageModule` and https://github.com/lefou/mill-jacoco are doing:

* Each test run gets instrumented by Kover and binary report is generated (this requires mixing in `KoverTests` trait)
* There is a possibility to generate HTML/XML report out of it using `__.kover.htmlReport` / `__.kover.xmlReport` respectively on each module having `KoverModule` mixed in.
* Also there is a possibility to declare module implementing `KoverReportModule` trait and use `htmlReportAll` / `xmlReportAll` tasks to collect coverage for all modules in the project.

Kover Agent documentation: https://kotlin.github.io/kotlinx-kover/jvm-agent
Kover Reporter documentation: https://kotlin.github.io/kotlinx-kover/cli/


example/kotlinlib/linting/4-kover/build.mill
```diff
@@ -0,0 +1,60 @@
...
+// coverage data.
+// This lets us perform the coverage operations but before that you
+// must first run the test.
+// `./mill test` then `./mill kover.htmlReport` and get your
+// coverage in HTML format.
+//
+// You can download this example project using the *download* link above
+// if you want to try out the commands below yourself. The only requirement is
+// that you have some version of the JVM installed; the `./mill` script takes
+// care of any further dependencies that need to be downloaded.
```
lihaoyi, line 39: This is probably a copy-paste error and should be removed

example/kotlinlib/linting/4-kover/build.mill
```diff
@@ -0,0 +1,60 @@
...
+... foo.FooTestskotlin - success started
+
+> ./mill resolve kover._ # List what tasks are available to run from kover
+...
+kover.htmlReport
+...
+kover.xmlReport
+...
+
+> ./mill kover.htmlReport
```
lihaoyi, line 54: Let's use `show kover.htmlReport` to make it clear where the output files live

example/kotlinlib/linting/4-kover/build.mill
```diff
@@ -0,0 +1,60 @@
...
+
+> cat out/kover/htmlReport.dest/kover-report/index.html
```
lihaoyi, line 56: I noticed that the `kover.htmlReport` target gives an incorrect/empty coverage report if `test` is not separately run first. Is that intentional? If that's the same as what `ScoverageModule` and `mill-jacoco` do I guess it's fine to follow suit, just kind of weird

kotlinlib/src/mill/kotlinlib/contrib/kover/KoverReportModule.scala
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * For example, define the following `kover` module and use the relevant
+ * reporting option to generate a report:
+ * {{{
+ * object kover extends KoverReportModule
+ * }}}
+ *
+ * - mill __.test                 # run tests for all modules
+ * - mill kover.htmlReportAll     # generates report in html format for all modules
+ * - mill kover.xmlReportAll      # generates report in xml format for all modules
```
lihaoyi, line 61: Having the `generates report in html format for all modules` task belong to a particular module's `.kover` submodule is kind of weird. If we're following the style of `ScalafmtModule` and `PalantirFormatModule`, these aggregate methods would belong to the external module via `kotlinlib.contrib.kover.KoverModule/htmlReportAll` or `kotlinlib.contrib.kover.KoverModule/xmlReportAll`, making it clear they're "global" and not owned by any one particular submodule of the build

example/kotlinlib/linting/4-kover/build.mill
0xnm: Deleted. The same was in `example/scalalib/linting/2-contrib-scoverage/build.mill`, so deleted there as well.

example/kotlinlib/linting/4-kover/build.mill
0xnm: I think both expect `test` to be run first indeed. This is what example of `Scoverage` is saying https://github.com/com-lihaoyi/mill/blob/bf81800270ece7adcec07215e1b0bf2735baf107/example/scalalib/linting/2-contrib-scoverage/build.mill#L29-L32 and `mill-jacoco` also runs `test` first: https://github.com/lefou/mill-jacoco/blob/c6d81d370bf780d8b972692fb1f7a7fd335958f6/build.sc#L142-L144.

Do you propose to add some console warning if `test` wasn't run before generating report?

kotlinlib/src/mill/kotlinlib/contrib/kover/KoverReportModule.scala
0xnm: Here I actually followed the idea of https://github.com/com-lihaoyi/mill/blob/bf81800270ece7adcec07215e1b0bf2735baf107/contrib/scoverage/src/mill/contrib/scoverage/ScoverageReport.scala.

Do you propose to take the approach of `mill-jacoco`: https://github.com/lefou/mill-jacoco/blob/c6d81d370bf780d8b972692fb1f7a7fd335958f6/README.adoc?plain=1#L37? This is essentially the same you proposing I believe.

lihaoyi: Yeah let's go with the mill-jacoco appraoch, which I believe is what we do for scalafmt/palantirformat. mill/contrib/scoverage is pretty old and crufty, so likely it just pre-dates us figuring out the best way to do these things

example/kotlinlib/linting/4-kover/build.mill
lihaoyi: How about a console message listing out which test module's coverage data got slurped up, with a big warning if no test module coverage data was detected at all? That would also help mitigate not just the "forgot to run tests at all" case, but also the "ran some tests but not others" failure mode.

Longer term, we should discuss if there's a better way of organizing things, e.g. making the coverage reports have an explicit task dependency on the test coverage data output. But that doesn't need to block this PR which just follows the (imperfect) status quo

0xnm: I added an error log message if binary report is missing: https://github.com/com-lihaoyi/mill/pull/3624/commits/3e00f9d7953ed55ba735e4cea980e05ae25d8957. Not sure how can module name can be reported.

kotlinlib/src/mill/kotlinlib/contrib/kover/KoverReportModule.scala
0xnm: Done. It will be like `mill.kotlinlib.contrib.kover.KoverModule/htmlReportAll`.

lihaoyi:
Looks good, thanks @0xnm ! Will transfer the bounty

PR #3626, opened 2024-09-30
Tidy up `init` command impl, make it unpack into current folder

lihaoyi:
Unpacking into the current folder is much more useful due to the fact that we recommend people use a bootstrap script. 

* With the previous approach, you end up with a bootstrap script in the parent dir and a subfolder with another bootstrap script with the real build inside. 
* With the new approach, you download a bootstrap script, run init, and it fleshes out the example in place

Since we now unpack into an existing folder that may or may not have things already in it, we add a new heuristic to fail early if the unpacking would over-write any existing files. This is a compromise, since we cannot easily assume the folder is 100% empty (what if there's a `.DS_STORE` in it???), nor can we remove any previous files we do not understand (what if there's already a `.gitignore` or `.git`?), and we also do not want to stomp over any existing files (what if there's already a `build.mill` and someone accidentally calls `init`?).

Tweak the implementation to avoid using so many nested `Try`s, normal exception handling is enough

Tested manually via `./mill installLocalCache`


lefou (commented):
There should be at least a message/warning for every file we delete or overwrite.

lihaoyi:
@lefou with the current impl, we do not delete/overwrite any file, and fail early if there's any conflict between the zip example and the files already on disk

lihaoyi:
There's something funky with `./mill -iw main.init.test` looping indefinitely, but that's already present in master and not unique to this PR. Will look into it

lefou:
I was referring to he deleting of a per-existing `mill` script.

PR #3633, opened 2024-09-30
Use build info for versions in kotlinlib

0xnm:
Some minor cleanup for the previous work on the Kotlin-related modules:

* Read tooling versions from build info (https://github.com/com-lihaoyi/mill/pull/3620#discussion_r1779445633 and https://github.com/com-lihaoyi/mill/pull/3620#discussion_r1779445811)
* Declare versions only once in the unit tests
* Change parent module to `JavaModule` for Ktfmt support, because we don't need the code to be compiled in this module, so declaring Kotlin version there was looking a bit strange. In fact only `sources()` method is needed, which is available in the `JavaModule`. In the future `KotlinModule` may be split into different traits: the one which collects Kotlin+Java files, but doesn't do any compilation (so Kotlin version is not needed) and the one which actually does compilation.

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -187,16 +187,35 @@ object Deps {
   val logback = ivy"ch.qos.logback:logback-classic:1.5.7"
   val sonatypeCentralClient = ivy"com.lumidion::sonatype-central-client-requests:0.3.0"
   val kotlinCompiler = ivy"org.jetbrains.kotlin:kotlin-compiler:1.9.24"
-  val koverVersion = "0.8.3"
-  val ktfmtVersion = "0.52"
-  val detektVersion = "1.23.7"
-  val dokkaVersion = "1.9.20"
 
   object RuntimeDeps {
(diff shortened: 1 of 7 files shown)


build.mill
```diff
@@ -188,6 +188,9 @@ object Deps {
   val sonatypeCentralClient = ivy"com.lumidion::sonatype-central-client-requests:0.3.0"
   val kotlinCompiler = ivy"org.jetbrains.kotlin:kotlin-compiler:1.9.24"
   val koverVersion = "0.8.3"
+  val ktfmtVersion = "0.52"
+  val detektVersion = "1.23.7"
+  val dokkaVersion = "1.9.20"
```
lefou, line 193: I think some or all of them are runtime versions, not needed to build the modules. So, they should go into the dedicated `object RuntimeDeps` below. Make sure to add them to `all` too, so we can check for updates automatically.

lihaoyi:
Thanks @0xnm !

PR #3635, opened 2024-10-01
Put back `withPromptPaused`, but make it `private[mill]`

lihaoyi:
Turns out we still need it for `console` and `repl` commands in order to disable the prompt auto updater thread. But we can limit its use to only within `EvaluatorCore` for every `exclusive` command, rather than letting it be used in any random user-land task

Also refactors it such that `exclusive = true` commands use `SystemStreams.original` by default, allowing `def repl` and `def console` to work without overriding the system streams themselves

Tested manually

```
$ ./mill -i dist.launcher && (cd example/scalalib/basic/1-simple  && ../../../../out/dist/launcher.dest/run -i path compile sources + console + version)

[2579/2579] ========================================================================= dist.launcher ============================================================================= 2s
[5/57] path
sources
allSources
allSourceFiles
compile
[56/57] console
Welcome to Scala 2.13.11 (OpenJDK 64-Bit Server VM, Java 17.0.6).

scala> 1 + 1
val res0: Int = 2

scala> :quit
[57/57] version
0.12.0-RC2-65-906911
[57/57] =============================================================== path compile sources + console + version ================================================================ 7s
lihaoyi mill$
```

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -116,7 +116,7 @@ object Deps {
     Seq(Play_3_0, Play_2_9, Play_2_8, Play_2_7, Play_2_6).map(p => (p.playBinVersion, p)).toMap
 
   val acyclic = ivy"com.lihaoyi:::acyclic:0.3.12"
-  val ammoniteVersion = "3.0.0"
+  val ammoniteVersion = "3.0.0-2-6342755f"
   val asmTree = ivy"org.ow2.asm:asm-tree:9.7"
   val bloopConfig = ivy"ch.epfl.scala::bloop-config:1.5.5"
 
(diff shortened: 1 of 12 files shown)


lefou:
Do we still print rhe task id and the task name before the exclusive task output? This would be especially helpful in case multiple exclusive tasks run sequentially.

lihaoyi:
@lefou yes, it looks like this:

```scala
$ ./mill -i dist.launcher && (cd example/scalalib/basic/1-simple  && ../../../../out/dist/launcher.dest/run -i path compile sources + console + version)

[2579/2579] ========================================================================= dist.launcher ============================================================================= 2s
[5/57] path
sources
allSources
allSourceFiles
compile
[56/57] console
Welcome to Scala 2.13.11 (OpenJDK 64-Bit Server VM, Java 17.0.6).

scala> 1 + 1
val res0: Int = 2

scala> :quit
[57/57] version
0.12.0-RC2-65-906911
[57/57] =============================================================== path compile sources + console + version ================================================================ 7s
lihaoyi mill$
```

lefou:
Great.

PR #3637, opened 2024-10-01
Standardize environment variables names on `_DIR` instead of `_FOLDER`

lihaoyi:
Follows suit from https://github.com/com-lihaoyi/mill/pull/3530. This hasn't been released in a stable version yet so we are free to change it


lefou (commented):
Looks good to me!

PR #3641, opened 2024-10-01
Optimize `SoftReference` caches using `ConcurrentHashMap`'s `compute`

lolgab:
@bjaglin raised a great point about concurrent initialization in mill-scalafix: https://github.com/joan38/mill-scalafix/pull/206#discussion_r1781402921
In `ZincWorkerImpl` were using `synchronized` to avoid that, but the built-in `compute` method in `java.util.concurrent.ConcurrentHashMap` is probably better.

Pull Request: https://github.com/com-lihaoyi/mill/pull/3641

### DIFF
--- a/scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
+++ b/scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
@@ -402,8 +402,7 @@ class ZincWorkerImpl(
 
   // for now this just grows unbounded; YOLO
   // But at least we do not prevent unloading/garbage collecting of classloaders
-  private[this] val classloaderCache =
-    new ZincWorkerImpl.AutoCloseableCache[Long, ClassLoader]
+  private[this] val classloaderCache = new ZincWorkerImpl.Cache[Long, ClassLoader]
 
   def getCachedClassLoader(compilersSig: Long, combinedCompilerJars: Array[java.io.File])(implicit
       ctx: ZincWorkerApi.Ctx
@@ -644,7 +643,7 @@ class ZincWorkerImpl(
   }
 
   override def close(): Unit = {
-    classloaderCache.clear()
+    classloaderCache.clear(closeAutoCloseables = true)
     javaOnlyCompilersCache.clear()
   }
 }
@@ -703,23 +702,20 @@ object ZincWorkerImpl {
         }
       )()
 
-    def clear(): Unit = cache.clear()
-  }
+    import scala.jdk.CollectionConverters._
+    def clear(closeAutoCloseables: Boolean): Unit = {
 
-  /** Cache that closes all `AutoClosable` values when `clear`ed */
-  private class AutoCloseableCache[A, B <: AnyRef] extends Cache[A, B] {
-    override def clear(): Unit = {
-      import scala.jdk.CollectionConverters._
+      if (closeAutoCloseables) {
+        cache
+          .values()
+          .iterator()
+          .asScala
+          .foreach { case SoftReference(v: AutoCloseable) =>
+            v.close()
+          }
+      }
 
-      cache
-        .values()
-        .iterator()
-        .asScala
-        .foreach { case SoftReference(v: AutoCloseable) =>
-          v.close()
-        }
-
-      super.clear()
+      cache.clear()
     }
   }
 }
--- a/scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
+++ b/scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
@@ -66,7 +66,7 @@ class ZincWorkerImpl(
 ) extends ZincWorkerApi with AutoCloseable {
   private val zincLogLevel = if (zincLogDebug) sbt.util.Level.Debug else sbt.util.Level.Info
   private[this] val ic = new sbt.internal.inc.IncrementalCompilerImpl()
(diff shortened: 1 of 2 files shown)


scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
```diff
@@ -708,4 +685,36 @@ object ZincWorkerImpl {
       toAnalyze = List((List(start), deps(start).toList))
     ).reverse
   }
+
+  private class AutoCloseableCache[A, B <: AnyRef] extends Cache[A, B] {
```
lefou, line 689: Why is this called `AutoCloseableCache`? Did you forget a `with AutoCloseable`?

Also, I'd prefer to have `Cache` defined before its derivatives.

lolgab: What I want to express is a cache that on `clear` also tries to find `AutoCloseable` values and `close` them. I don't what is the best name for it. Do you have suggestions? I will change the order.

lihaoyi:
I'm personally unsure about the motivation here. Are we seeing performance bottlenecks that we think will be solved by this optimization? How will we know that this PR actually achieves what we hope it will?

There's always a risk of bugs when making changes to this kind of concurrent code, both nondeterminsitic semantic bugs and performance bugs, and the risk typically goes up as your concurrency model gets more fine-grained. If we're not seeing any real bottlenecks, I would prefer to have coarse-grained synchronized blocks and locks over fine-grained concurrent data structures, the more coarse-grained the better

lolgab:
You are right, there is no particular performance reason behind this change. I should have named the PR `Simplify...` instead of `Optimize...`
Since I learned that we have a standard library function to do what we were manually doing (initialization of a cache disallowing concurrent creation), I thought it would simplify the code to use the std library function instead of reinventing it in Mill.
Some problems I noticed while making the PR:
- `ScalaJSWorkerImpl` uses a `mutable.Map` which is not thread-safe. What does it happen if two threads try to link Scala.js code at the same time? Its initialization doesn't use a synchronized block.
- `javaOnlyCompilersCache` uses another `mutable.Map` which is not thread-safe and doesn't wrap initialization with `synchronized`
- not sure about this one, but `classloaderCache` uses a non-thread-safe implementation. Maybe it's not a problem, but if you call `clear` and `getOrElseCreate` at the same time from two threads. Probably not possible at all.

A possibly problematic behavioral change of this PR that I haven't noticed until now:
Before it was using a `LinkedHashMap`, now a `ConcurrentHashMap`. Don't know if maintaining the order of insertion is important when we close classloaders. Is it?

lihaoyi:
I agree that there are probably tons of concurrency issues in the current implementation, but we still need to be clear about what we are trying to achieve here. Just dropping in `Concurrent*` data structures does nothing to make piece of code thread-safe or race-free, and unlike dealing with single-threaded correctness the bar of "seems to work and passes unit tests" tells us basically nothing about thread-safety.

To be clear, I think using `ConcurrentHashMap` sounds great, and could be a path to code that is both more cleaner and correct. But the PR description needs to be a lot more rigorous in identifying what the current problems are and arguing why the new approach fixes it

scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
```diff
@@ -708,4 +684,38 @@ object ZincWorkerImpl {
...
+      new java.util.concurrent.ConcurrentHashMap[A, SoftReference[B]]
+
+    def getOrElseCreate(a: A)(create: => B): B =
+      cache.compute(
+        a,
+        {
+          case (_, v @ SoftReference(_)) => v
+          case _ => SoftReference(create)
+        }
+      )()
```
bjaglin, line 703: Isn't there a tiny risk that the instance returned by `create` (or the existing one) is garbage-collected by the time the `SoftReference` is dereferenced? Having a var to maintain a strong reference throughout the lifecycle of `getOrElseCreate` feels safer.

Ichoran: Yeah, this doesn't look right.  The scheme that I used in `Soft` (which is a `SoftReference` that can recreate itself if it's needed after it's cleared) is better: https://github.com/Ichoran/kse3/blob/5a900ad9578008f0f3454c759cb83e73ec0e21a8/flow/src/Cached.scala#L105-L154

scalalib/worker/src/mill/scalalib/worker/ZincWorkerImpl.scala
```diff
@@ -708,4 +684,38 @@ object ZincWorkerImpl {
...
+
+    import scala.jdk.CollectionConverters._
+    def clear(closeAutoCloseables: Boolean): Unit = {
+
+      if (closeAutoCloseables) {
+        cache
+          .values()
+          .iterator()
+          .asScala
+          .foreach { case SoftReference(v: AutoCloseable) =>
```
lefou, line 713: `foreach` doesn't take a partial function, so we should provide a catch-all `case`. Not doing so will fail for non-`Autoclosable` entries.

lihaoyi:
Opened a ticker to centralize discussion of the more general issue here https://github.com/com-lihaoyi/mill/issues/3730

lihaoyi:
Superseded by https://github.com/com-lihaoyi/mill/pull/4001

PR #3643, opened 2024-10-01
Add Kotlin BSP integration test

alexarchambault:
This PR adds non-regression integration tests for Kotlin support via BSP. Kotlin modules are currently advertized as Java ones, and IntelliJ seems to handle fine Kotlin sources in those (using non-experimental BSP support).

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -449,7 +449,8 @@ trait MillBaseTestsModule extends TestModule {
       s"-DTEST_SCALATEST_VERSION=${Deps.TestDeps.scalaTest.dep.version}",
       s"-DTEST_TEST_INTERFACE_VERSION=${Deps.sbtTestInterface.dep.version}",
       s"-DTEST_ZIOTEST_VERSION=${Deps.TestDeps.zioTest.dep.version}",
-      s"-DTEST_ZINC_VERSION=${Deps.zinc.dep.version}"
+      s"-DTEST_ZINC_VERSION=${Deps.zinc.dep.version}",
+      s"-DTEST_KOTLIN_VERSION=${Deps.kotlinCompiler.dep.version}"
     )
   }
 
--- a/integration/ide/bsp-server/src/BspServerTests.scala
+++ b/integration/ide/bsp-server/src/BspServerTests.scala
@@ -15,6 +15,26 @@ object BspServerTests extends UtestIntegrationTestSuite {
   override protected def workspaceSourcePath: os.Path =
     super.workspaceSourcePath / "project"
 
+  def substitutions(
+      dependency: coursierapi.Dependency,
+      filter: coursierapi.Dependency => Boolean
+  ): Seq[(String, String)] = {
+    val fetchRes = coursierapi.Fetch.create()
+      .addDependencies(dependency)
+      .fetchResult()
+    fetchRes.getDependencies.asScala
+      .filter(filter)
+      .map { dep =>
+        val organization = dep.getModule.getOrganization
+        val name = dep.getModule.getName
+        val prefix = (organization.split('.') :+ name).mkString("/")
+        def basePath(version: String): String =
+          s"$prefix/$version/$name-$version"
+        basePath(dep.getVersion) -> basePath(s"<$name-version>")
+      }
+      .toSeq
+  }
+
   def tests: Tests = Tests {
     test("requestSnapshots") - integrationTest { tester =>
       import tester._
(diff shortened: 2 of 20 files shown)


alexarchambault:

alexarchambault:
Should preserve bin compat, unlike https://github.com/com-lihaoyi/mill/pull/3619 and https://github.com/com-lihaoyi/mill/pull/3644

lefou (commented):
TBH, to accept these weird kind of workaround, I'd require at least an upstream issue and/or discussion in the BSP project. Especially since this is a IDEA specific thing whereas BSP is a more general protocol, also serving other IDEs and editors. 

We already have some BSP client specific handling (e.g. enable semanticDB for metals, or if you dig in the git history handling of the `build.sc` for IDEA), so we should guard workaround behind a BSP client check.

alexarchambault:
@lefou https://github.com/com-lihaoyi/mill/pull/3644 restricts that workaround to IntelliJ, and only so if it seems IntelliJ lacks explicit Kotlin support over BSP. But it requires changes that break binary compatibility in `BspModule`.

alexarchambault:
In a similiar fashion, I'd like to propose to add the [synthetic root module](https://github.com/com-lihaoyi/mill/blob/26cbc8adee1bca52de9cb0c3fb8297c450c51d53/bsp/worker/src/mill/bsp/worker/SyntheticRootBspBuildTargetData.scala) only when we detect we're connected to IntelliJ (that one could be made without breaking bin compat).

lefou:
Can you a bit more specific? How do you detect the missing support in IDEA and why can't it be binary compatible. Or if this was discussed before plz provide a pointer.

lefou:
The whole BSP implementation is far from simple, so a bit of context and explanation helps a lot.

build.mill
```diff
@@ -823,26 +848,6 @@ object dist extends MillPublishJavaModule {
     PathRef(T.dest / raw.last)
   }
 
-  def prependShellScript = T {
```
lefou, line 826: Why is this moved to a local position? If you don't want the caching, you could make it a anonymous task instead.

alexarchambault:
> Can you a bit more specific? How do you detect the missing support in IDEA and why can't it be binary compatible. Or if this was discussed before plz provide a pointer.

Right now, BSP support in IntelliJ is provided by [intellij-scala](https://github.com/JetBrains/intellij-scala). Intellij-scala [advertizes support only for Java and Scala](https://github.com/JetBrains/intellij-scala/blob/6432b6e0b0ea3acb0aabb8eb0136abb9e49b45fc/bsp/src/org/jetbrains/bsp/protocol/BspCommunication.scala#L232-L233) in its `InitializeBuildParams` during BSP server initialization. Yet, in practice, it supports Kotlin sources in Scala modules, as I was able to check with [this project](https://github.com/alexarchambault/sbt-kotlin-test). The PR here makes Kotlin works in IntelliJ by relying on this.

Yet, advertizing Kotlin modules as Scala ones doesn't allow to pass kotlinc arguments (and other Kotlin-specific details) to the BSP client. It seems Jetbrains is working on a better solution, with explicit support for Kotlin via BSP, in the [intellij-bsp project](https://github.com/JetBrains/hirschgarten/tree/main/plugin-bsp) (I mean, it's formerly called "intellij-bsp", it's now part of a bigger repo). Note that intellij-scala has its own BSP implementation, that differs from the one of intellij-bsp.

@jastice [said](https://github.com/com-lihaoyi/mill/pull/3619#issuecomment-2381842143) in #3619 that intellij-bsp would probabaly be used mid-term by intellij-scala too.

So in #3644, we try to detect which IntelliJ BSP implementation we talk to, and advertize Kotlin modules either as Scala ones (for intellij-scala) or as Kotlin ones (for intellij-bsp). To detect which intellij BSP implementation we talk to, inspecting the BSP `InitializeBuildParams` seems to work: both implementation use `Intellij-BSP` as "display name", but they claim support for different language sets: [only Java and Scala](https://github.com/JetBrains/intellij-scala/blob/6432b6e0b0ea3acb0aabb8eb0136abb9e49b45fc/bsp/src/org/jetbrains/bsp/protocol/BspCommunication.scala#L232-L233) for intellij-scala, [Java, Scala, Kotlin, and more](https://github.com/JetBrains/hirschgarten/blob/3f8d48bb7ff023d5403c6b0b7067045ef993be82/protocol/src/main/kotlin/org/jetbrains/bsp/protocol/BspConstants.kt#L14) for intellij-bsp.

alexarchambault:
About the binary compatibility issue, we need details of the client `InitializeBuildParams` (at least, "display name" and supported languages) in `BspModule.{bspBuildTarget, bspBuildTargetData}`. Passing those details to these methods breaks bin compat.

alexarchambault:
The idea with the PR here is to advertize Kotlin module as Scala ones via BSP, rather than Java ones as is currently the case (as `KotlinModule` doesn't currently override `BspModule.{bspBuildTarget, bspBuildTargetData}`). That seemed to be an acceptable short-term workaround to me, until the bin compat of `BspModule` can be broken.

lefou:
Understood.

Regarding the detection logic. The data in `BspModule` isn't necessarily the one we need to send to the client. I think the server could simply translate it depending on its knowledge about the client. So we could be more BSP client specific without breaking `BspModule`.

Technically, I'd rather avoid investing into the current monolithic BSP implementation in favor of a more modular approach, but it's currently blocked by bsp4j not being leightweight enough, https://github.com/build-server-protocol/build-server-protocol/issues/668 / https://github.com/build-server-protocol/build-server-protocol/discussions/651.

lefou:
I think `KotlinModule` should definitely override `BspModule` `def`s to reflect its nature. This shouldn't break any compatibility.

Please note, that `KotlinModule` is currently unfinished. Also, I'd expect a rather poor behavior in any BSP client, since compilation of Kotlin modules isn't incremental and takes very long compared to any other languages supported by Mill. Hence, the `GenIdea` approach is still the go-to solution for Kotlin projects. I can't say much more about it, since I haven't finished the review of all the changes introduced lately to Mill in depth. This also includes the synthetic BSP module you mentioned.

alexarchambault:
One last note: I can't recall if I checked whether Kotlin sources alongside a Java module (as should be the case currently without this PR) were fine in Intellij. I checked as Kotlin module as done in #3644 in the general case, and as a Scala one as done here. We should able to do #3644 without breaking bin compat if it works fine alongside a Java module, by changing the `KotlinBuildTarget` data to `JvmBuildTarget` in `MillBuildServer` itself, rather than from the `BspModule` methods overrides.

lefou:
So is `intelli-scala` ignoring modules which support Kotlin as BSP language? Or only, or is just the Kotlin part not working, but they still work as Java modules?

lefou:
This constant rebasing make the review process harder than necessary. I'm going to set this PR to draft mode until PR #3608 is merged.

alexarchambault:
> This repeatedly rebasing makes the review process harder than necessary. I'm going to set this PR to draft mode until PR #3608 is merged.

The last commit is the only one specific to this PR. You can just look at this one. You'll have no diff between pushes, but that commit isn't that large.

build.mill
alexarchambault: Dealt with in https://github.com/com-lihaoyi/mill/pull/3608

integration/ide/bsp-server/src/BspServerTests.scala
```diff
@@ -15,6 +15,26 @@ object BspServerTests extends UtestIntegrationTestSuite {
   override protected def workspaceSourcePath: os.Path =
     super.workspaceSourcePath / "project"
 
+  def substitutions(
```
lihaoyi, line 18: Let's call this `substituteTransitiveDependencies` or something, the current name is a bit too concise given it's not obvious what it's trying to do

lihaoyi (approved):
One nit but looks good. What's the relationship between this and https://github.com/com-lihaoyi/mill/pull/3644? Does one depend on the other? Or are they unrelated?

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -314,6 +315,13 @@ trait KotlinModule extends JavaModule { outer =>
 
   private[kotlinlib] def internalReportOldProblems: Task[Boolean] = zincReportCachedProblems
 
+  @internal
+  override def bspBuildTarget: BspBuildTarget = super.bspBuildTarget.copy(
```
lihaoyi, line 319: What's the relation between this `bspBuildTarget` and the `jvmBuildTarget` you introduced in https://github.com/com-lihaoyi/mill/pull/3681?

alexarchambault: `bspBuildTarget` is defined in `BspModule`, and was there before my PRs. It contains details about the module, passed to the BSP clients via the `workspace/buildTargets` request. It goes along `bspBuildTargetData`, that contains language-specific data passed along in the response to this request too.

`bspJvmBuildTarget` contains JVM-specific details about the module. These are passed to clients via `bspBuildTargetData` for `JavaModule`, and as a field of what's passed in `bspBuildTargetData` for `ScalaModule` (same for `KotlinModule` in https://github.com/com-lihaoyi/mill/pull/3644). So `ScalaModule` (and `KotlinModule` in https://github.com/com-lihaoyi/mill/pull/3644) also use `bspJvmBuildTarget`, but uses it differently in `bspBuildTargetData`.

alexarchambault:
> What's the relationship between this and #3644? Does one depend on the other? Or are they unrelated?

#3644 includes the PR here. It adds things to it (it actually adds Kotlin-specific data in some BSP responses, to pass kotlinc arguments to clients in particular). It adds what [intellij-bsp](https://github.com/JetBrains/hirschgarten/tree/main/plugin-bsp) expects for Kotlin modules, although I wasn't able to use it in practice, and it breaks binary compatibility in order to detect which IntelliJ BSP implementation it's talking to.

PR #3679, opened 2024-10-06
Fixes:#3550; Added hello-world Android Kotlin example using Mill

himanshumahajan138:
# Pull Request 
Fixes: #3550 

## Description
Added `Kotlin Android "Hello World" Application Example` while reusing `AndroidSdkModule`, `KotlinModule` and creating new Module `AndroidAppModule` with Kotlin Support and Proper `Documentation`.

## Related Issues
- Link to related issue #3550.

## Checklist
- [x] Android Support for Kotlin Added in the example section for "HelloWorld" Android Application.
- [x] Added Kotlin Source files and Customized AndroidAppModule.
- [x] Code is clean and follows project conventions.
- [x] Documentation has been updated.
- [x] Tests have been added or updated.
- [x] All tests pass.

## Additional Notes
Actually, I was wondering that we reused the `AndroidSdkModule` but created a new Module `AndroidAppModule` for Andorid Workflow but what we can do is to provide `Kotlin support in the Existing AndroidAppModule` and then based on user query they can use the functionality.

So for this i need @lihaoyi Sir your Permission and Guidance, Please Review this PR suggest changes...


kotlinlib/src/mill/kotlinlib/android/AndroidAppModule.scala
```diff
@@ -0,0 +1,259 @@
...
+ * [[AndroidSdkModule]] (for Android SDK interactions) and [[KotlinModule]] (for standard Kotlin tasks).
+ *
+ * It provides a structured way to handle various steps in the Android app build process,
+ * including compiling Kotlin sources, creating DEX files, generating resources, packaging
+ * APKs, optimizing, and signing APKs.
+ *
+ * [[https://developer.android.com/studio Android Studio Documentation]]
+ */
+@mill.api.experimental
+trait AndroidAppModule extends KotlinModule {
```
lihaoyi, line 23: This should inherit from `mill.javalib.android.AndroidAppModule`, and should be able to re-use almost all that functionality without needing to copy it over

himanshumahajan138: actuatlly i was thinking the same thing but think of like if we combined KotlinModule and JavaModule Together then there will be clash with compile functions and we have to keep track of the files like what user wants to do 

please share your thoughts Sir...

lihaoyi: `KotlinModule` already inherits from `JavaModule`, so there should be no issue in making it work

himanshumahajan138: ooh now i got your point you wana say that allong with KotlinModule it should extend AndroidAppModule and then for compiling we will use KotlinModule and for other android related functionailty we will use AndroidAppModule 

ok then i will do updates thanks for review Sir...

himanshumahajan138:
### Final Changes

- [x] `mill.kotlinlib.android.AndroidAppModule` Renamed to `AndroidAppKotlinModule`
- [x] Extended `mill.javalib.android.AndroidAppModule` for App related tasks and redefine `androidJar` task for Kotlin Compilation
- [x] Fixed and `Updated` `Documentation`
- [x] All `Tests` `Passed` 

### Note 

@lihaoyi Thankyou Sir for the detailed feedback!

I’ve addressed all the requested changes and resolved the comments. Please let me know if anything else needs attention.

Please pardon me if I missed anything or made a mistake. Looking forward to your final review and any further suggestions!

kotlinlib/src/mill/kotlinlib/android/AndroidAppModule.scala
himanshumahajan138: done

kotlinlib/src/mill/kotlinlib/android/AndroidAppKotlinModule.scala
```diff
@@ -0,0 +1,64 @@
...
+    os.call(
+      Seq(
+        androidSdkModule().d8Path().path.toString, // Call D8 tool
+        "--output",
+        jarFile.toString, // Output JAR file
+        "--no-desugaring" // Disable desugaring
+      ) ++ os.walk(compile().classes.path).filter(_.ext == "class").map(
+        _.toString
+      ) // Get Kotlin class files, this calls compile task from Kotlinlib
+    )
```
lihaoyi, line 59: Why do we need this? Isn't this the same as the implementation in `AndroidAppModule`?

himanshumahajan138: sir it is same but think of situation like if we will implement the one from the `AndroidAppodule` then in that case our `AndoridAppModule` is extending `JavaModule` and in this we have `compile` function for `.java` extension only and we need this for `.kt` this time so like we extended the `KotlinModule` in `AndroidAppKotlinModule` then we need to call `compile` function here so that it will take function from the `KotlinModule` rather than the `JavaModule`

This is My Hypothesis or Approach is this correct ?

please extend the Conversation Sir with your thoughts...

himanshumahajan138: its shame for me actually it worked i tried this now 😅

no worries i will update the code just give me 5 mins

himanshumahajan138: @lihaoyi Sir it worked even without `AndroidSdkModule` Abstract 😅 

i removed this too...

this is what lack of experience looks like (shame on me) highly inspired by your expertise thanks for guidance Sir...

lihaoyi:
Look great @himanshumahajan138 , will merge it and close out that part of the bounty

PR #3680, opened 2024-10-07
Add Java 21 and 23 to CI

lefou:


.github/workflows/run-tests.yml
```diff
@@ -64,6 +64,15 @@ jobs:
           - java-version: 17
             millargs: "contrib.__.testCached"
 
+          - java-version: 21
```
lihaoyi, line 67: I'd suggest we pick some of the existing jobs and set them to 21/23. We are already not covering the entire version matrix, and we can't afford to if we want to keep CI times reasonable. So the best we can do is to scatter jobs at various points around the version matrix, which should hopefully be enough to catch most things

lefou: I'm solely interested in what fails and what succeeds. We can optimize the CI setup before we merge it.

lefou:
* #4388

lihaoyi:
Superseded by https://github.com/com-lihaoyi/mill/pull/4388

PR #3681, opened 2024-10-07
Pass BSP JvmBuildTarget stuff to ScalaBuildTarget

alexarchambault:
This adds the JVM we use in the Scala details in BSP build targets. I don't think Metals uses that for now, and I don't know about IntelliJ. But they might in the future.

### DIFF
--- a/scalalib/src/mill/scalalib/JavaModule.scala
+++ b/scalalib/src/mill/scalalib/JavaModule.scala
@@ -1057,7 +1057,7 @@ trait JavaModule
   )
 
   @internal
-  def jvmBuildTarget: JvmBuildTarget =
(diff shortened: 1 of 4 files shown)


alexarchambault:

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -1056,14 +1056,15 @@ trait JavaModule
     canRun = true
   )
 
+  @internal
+  def jvmBuildTarget: JvmBuildTarget =
```
lihaoyi, line 1060: Should we call this `bspJvmBuildTarget`, just so it's clear it's related to the other `bsp*` members

lihaoyi (approved):
One nit but otherwise looks good to me

PR #3683, opened 2024-10-07
Enable lock on output dir for BSP server too

alexarchambault:


main/eval/src/mill/eval/EvaluatorCore.scala
```diff
@@ -42,7 +42,28 @@ private[mill] trait EvaluatorCore extends GroupEvaluator {
...
+        OutLock.withLock(
+          noBuildLock = !outLock,
+          noWaitForBuildLock = false, // ???
+          out = outPath,
+          targetsAndParams = goals.toSeq.map {
+            case n: NamedTask[_] => n.label
+            case t => t.toString
+          },
+          streams = logger.systemStreams
+        ) {
```
lihaoyi, line 55: Can we keep the locking logic out of `Evaluator`? The granularity of `def evaluate` is too fine grained, since we need to lock around not just a single evaluation but the entire `MillBuildBootstrap` process that may encompass multiple evaluations.

I'm actually not sure how the BSP server interacts with evaluators and the bootstrap process. Is it able to properly re-run bootstrapping if the build config changes from under it, or would it need to be explicitly restarted?

lihaoyi: Got it.

In the short term, making the BSP server responsible for taking the lock whenever it calls `evaluate` sounds like a step forward: that is enough to ensure BSP evals do not overlap with CLI commands, and I assume that in any typical workflow there'll be a lot more BSP evals than there are BSP bootstraps

After that, medium term, let's see if there's some way to take the lock during the BSP bootstrap process as well, to ensure mutex between BSP bootstraps and CLI bootstraps. I suspect doing it nicely may require refactoring how the BSP server is spawned and managed. I see two options:

1. Make BSP server go through `MillBuildBootstrap` every time it wants something, rather than `evaluate`. 
    1. There will be some performance implications here, but maybe it can be optimized enough to be feasible. 
    2. This would be the "most elegant" solution in ensuring the BSP and CLI go through the same code paths, and ensure full mutex

2. Make BSP server take the lock go through `MillBuildBoostrap#evaluate` only for generating the `RunnerState` which contains the list of evaluators, then release the lock. After that, we can pass the evaluators to the BSP server, which can then take the lock as necessary when it calls evaluate on the bottom-most evaluator
    1. This would have no performance implications over what we do today
    2. This would leave some race conditions in place: if BSP bootstraps, build is changed, a CLI command bootstraps, then CLI evaluates, it would evaluate using stale state and may misbehave

Of the two options, maybe try (1) and see if we can make that work? And if not we can fall back to (2)

alexarchambault: I could try it, but I feel 1. would be impractical. Not having go through the meta-builds is actually a nice optimization.

To work around the issue you raise in 2.ii., I would try to avoid having a BSP server run on a stale meta-build as much as possible, but also find ways for it not crash if ever it does so.

To avoid having a BSP server run on a meta-build for too long, I'd make the command spawned for BSP in `MillMain` (`"mill.bsp.BSP/startSession"`) not block, so that the watch loop awakes as soon as the meta-build changes, and we can recompile the meta-build and update the evaluators ASAP. We could also prevent the BSP server from answering requests while the meta-build is being re-compiled.

To prevent rare uses of a stale meta-build to crash with NoClassDefFound and the like, I'd copy the class path of a meta-build before creating a class loader with it. In a directory specific to that evaluator. So that when other evaluators re-compile the meta-build, they don't remove class files of class loaders of other evaluators.

alexarchambault: > To prevent rare uses of a stale meta-build to crash with NoClassDefFound and the like, I'd copy the class path of a meta-build before creating a class loader with it. In a directory specific to that evaluator. So that when other evaluators re-compile the meta-build, they don't remove class files of class loaders of other evaluators.

About this one, Bloop does things along those lines already: in `.bloop` sub-directories, it has its own directories with compilation output, but it copies it to distinct directories for each of its clients.

lihaoyi: @alexarchambault what you proposed sounds reasonable, go for it

lihaoyi:
@alexarchambault is this ready to review/merge?

alexarchambault:
@lihaoyi I guess it can be reviewed / merged. I'd like to add an integration test to ensure Mill commands and BSP requests can be interleaved. It'd be mainly useful for non-regression, to ensure upcoming changes don't break things. So it can always be added later.

lihaoyi:
Sounds good, I went ahead and merged it for now

alexarchambault:
Just in case, I'd manually check if BSP import in IntelliJ works fine. I didn't with the changes here.

lihaoyi:
Just did, seems to work!

Issue #3695, opened 2024-10-09
Support Kotlin 2.0/Multiplatform artifacts during dependency resolution


lihaoyi:
Follow up to https://github.com/com-lihaoyi/mill/issues/3611

The basic issue is that Coursier is unable to resolve Kotlin 2.0 dependencies, such as https://repo1.maven.org/maven2/org/jetbrains/kotlinx/kotlinx-html-js/0.11.0/. These dependencies do not contain `.jar` files, but instead contain `.module` files and `.klib` files:

```
kotlin-stdlib-js-2.0.0-sources.jar                2024-05-20 16:12    620420      
kotlin-stdlib-js-2.0.0-sources.jar.asc            2024-05-20 16:12       679      
kotlin-stdlib-js-2.0.0-sources.jar.md5            2024-05-20 16:12        32      
kotlin-stdlib-js-2.0.0-sources.jar.sha1           2024-05-20 16:12        40      
kotlin-stdlib-js-2.0.0.klib                       2024-05-20 16:12   2635364      
kotlin-stdlib-js-2.0.0.klib.asc                   2024-05-20 16:12       679      
kotlin-stdlib-js-2.0.0.klib.md5                   2024-05-20 16:12        32      
kotlin-stdlib-js-2.0.0.klib.sha1                  2024-05-20 16:12        40      
kotlin-stdlib-js-2.0.0.module                     2024-05-20 16:12      2987      
kotlin-stdlib-js-2.0.0.module.asc                 2024-05-20 16:12       679      
kotlin-stdlib-js-2.0.0.module.md5                 2024-05-20 16:12        32      
kotlin-stdlib-js-2.0.0.module.sha1                2024-05-20 16:12        40      
kotlin-stdlib-js-2.0.0.pom                        2024-05-20 16:12      1544      
kotlin-stdlib-js-2.0.0.pom.asc                    2024-05-20 16:12       679      
kotlin-stdlib-js-2.0.0.pom.md5                    2024-05-20 16:12        32      
kotlin-stdlib-js-2.0.0.pom.sha1                   2024-05-20 16:12        40      
kotlin-stdlib-js-2.0.0.spdx.json                  2024-05-20 16:12      1064      
kotlin-stdlib-js-2.0.0.spdx.json.asc              2024-05-20 16:12       679      
kotlin-stdlib-js-2.0.0.spdx.json.md5              2024-05-20 16:12        32      
[...]

PR #3702, opened 2024-10-09
Fetch klib artifacts too

alexarchambault:
Quick workaround for https://github.com/com-lihaoyi/mill/issues/3695

### DIFF
--- a/main/util/src/mill/util/CoursierSupport.scala
+++ b/main/util/src/mill/util/CoursierSupport.scala
@@ -104,7 +104,7 @@ trait CoursierSupport {
             Agg.from(
               res.files
                 .map(os.Path(_))
-                .filter(path => path.ext == "jar" && resolveFilter(path))
+                .filter(resolveFilter)
                 .map(PathRef(_, quick = true))
             ) ++ localTestDeps.flatten
           )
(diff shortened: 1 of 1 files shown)


alexarchambault:
The fix here (currently, second commit) should only be temporary. Working on a proper fix, that should be two-fold:
- Mill should allow users to add extra artifact types (sbt does for sure, Maven and Gradle probably do too)
- coursier should fetch `klib` artifacts by default, given that these seem pretty widespread in the Kotlin ecosystem

main/util/src/mill/util/CoursierSupport.scala
```diff
@@ -104,7 +108,7 @@ trait CoursierSupport {
             Agg.from(
               res.files
                 .map(os.Path(_))
-                .filter(path => path.ext == "jar" && resolveFilter(path))
+                .filter(resolveFilter)
```
lefou, line 111: To preserve previous behavior, you should change the `resolveFilter` default to contain the `path.ext == "jar"` match, I guess.

lihaoyi:
Just tried this out, seems to satisfy the manual test described in ./mill -i dist.run example/kotlinlib/web/3-hello-kotlinjs foo.compile

main/util/src/mill/util/CoursierSupport.scala
alexarchambault: Answered in https://github.com/com-lihaoyi/mill/pull/3701#discussion_r1794821092

alexarchambault:
Closing this in favor of https://github.com/com-lihaoyi/mill/pull/3701 and https://github.com/com-lihaoyi/mill/pull/3705 (the two PRs are needed to fix https://github.com/com-lihaoyi/mill/issues/3695)

PR #3703, opened 2024-10-09
Add JavaModule.artifactTypes

alexarchambault:
Allowing users to ask for more artifact types to be fetched and put in the classpath


alexarchambault:
Might break bin compat :/

lefou (commented):
This doesn't need to break bin-compat as long as you provide a bin-compat overload without defaults delegating to the newest version. For that reason, the new parameter `artifactTypes` needs to come last.

PR #3704, opened 2024-10-09
Clean up out folder lock logic

lihaoyi:
* The locking logic doesn't belong on `Logger`, moved to `MillMain`
* Make it only start the logger once the lock is taken so the ticker doesn't show when Mill is not making progress
* Simplify the test, a lot, and kill the dummy locks since they aren't really used
* Store the `millActiveCommand` in the `out/` folder and print it out as part of the lock waiting message

```
lihaoyi mill$ /Users/lihaoyi/.cache/mill/download/0.12.0-RC3-35-c2dddd-DIRTYd553ab5d resolve _
Mill version 0.12.0-RC3-35-c2dddd-DIRTYd553ab5d is different than configured for this directory!
Configured version is 0.12.0-RC3-32-b4a0bf (/Users/lihaoyi/Github/mill/.config/mill-version)
Another Mill process is running '{scalalib,scalajslib,scalanativelib}.__.test', waiting for it to be done...
```


lihaoyi:
CC @alexarchambault

PR #3705, opened 2024-10-10
Update coursier to 2.1.14

alexarchambault:

PR #3709, opened 2024-10-10
Add kotlinx-html to `3-hello-kotlinjs` example

lihaoyi:
Takes advantage of the new klib support pulled in by https://github.com/com-lihaoyi/mill/pull/3705

We don't support the `module` file indirection, so for now we have to reference the `-js` version of the artifact directly


lihaoyi:
CC @0xnm

PR #3710, opened 2024-10-10
Try to use kotest in Kotlinjs example

lihaoyi:
Doesn't work yet because `mill foo.test` doesn't initially fail like it's meant to

### DIFF
--- a/example/kotlinlib/web/3-hello-kotlinjs/build.mill
+++ b/example/kotlinlib/web/3-hello-kotlinjs/build.mill
@@ -17,7 +17,11 @@ object foo extends KotlinJSModule {
   def ivyDeps = Agg(
     ivy"org.jetbrains.kotlinx:kotlinx-html-js:0.11.0",
   )
-  object test extends KotlinJSModule with KotlinJSKotlinXTests
+  object test extends KotlinJSTests{
+    def ivyDeps = Agg(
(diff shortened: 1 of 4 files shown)


lihaoyi:
@0xnm do you have any idea why `io.kotest:kotest-framework-engine-js:5.9.1` might not be running the tests on KotlinJS while the previous `org.jetbrains.kotlin:kotlin-test-js:${kotlinVersion()}` seems to run fine? Could this be related to the missing test result reporting components you mentioned earlier?

0xnm:
I believe `kotlin-test` has a special privilege in the compiler by looking on the places like [this](https://github.com/JetBrains/kotlin/blob/fb59c3ec69e8171a5599bb0e48ccc18a2bccbfa4/compiler/ir/backend.js/src/org/jetbrains/kotlin/ir/backend/js/lower/TestGenerator.kt#L278-L288), so compiler produces test executable in such case.

Kotest, obviously, doesn't have such support and it is needed to write a runner for this engine (although it [has](https://github.com/kotest/kotest/blob/4f2cea0cb378c06efaf37206759651d718f10bd1/kotest-framework/kotest-framework-engine/src/jvmMain/kotlin/io/kotest/engine/launcher/main.kt#L21) a `main` method for the launcher).

I will try to see what needs to be done once I have time.

PR #3716, opened 2024-10-10
Make JVM version configurable per-module

albertpchen:
Fixes https://github.com/com-lihaoyi/mill/issues/3480.

Changes:

- adds `coursier-jvm` as a dependency to the `util` module, this library is for fetching JVMs using coursier and the coursier [jvm-index](https://github.com/coursier/jvm-index)

- add a `def javaHome: Task[Option[PathRef]]` task to `ZincWorkerModule`. This defaults to `None`, which is the existing behavior of using mill's java home.

- Users who want to use a custom JVM need to define a custom `ZincWorkerModule` and override `def jvmId`, optionally `def jvmIndexVersion`

- updates the `mockito` and `commons-io` examples to use the new configuration options. Now these examples should run even when mill itself is not running on java 11.

- This also required adding `-encoding utf-8` to the projects' javac options.

- update the `run` and `test` tasks to use the `zincWorker`'s java home if a different one specified

Notes:

* `JavaModule#compile` forks a new JVM for each compilation to support custom JVM versions, and `ScalaModule#compile` does not support custom JVM versions at all. This would require a deeper refactoring of `ZincWorkerModule` to fix
*


albertpchen:
I think I'm running into the same issue as this https://github.com/sbt/sbt-assembly/issues/496. `coursier-jvm` pulls in shapless which has the problematic class. not sure what the correct fix is but I can think of a few options
1. add `export LC_ALL=C.UTF-8` to the workflow as mentioned in the issue
2. since we might not want shapeless as a dependency anyway, reimplement the parts of coursier-jvm that we need. I don't think this should be too hard, since coursier already has all the fetching and caching logic

lihaoyi:
At a first skim this looks great. I didnt know coursier could fetch JVMs or that it was so easy to integrate

Is it possible to configure a module to point it at a JVM not listed in coursier's index? I imagine there would be some use cases for that

The cross module stuff is lazily initialized then cached, so it should be OK to perform heavy downloads and initialization, but is it possible to move the download into a task? That would help benefit from parallelization and log handling and error reporting and all that

@alexarchambault should probably review this as well, since he knows more about coursier than i do

albertpchen:
> Is it possible to configure a module to point it at a JVM not listed in coursier's index? I imagine there would be some use cases for that

yes this is possible. I added `def javaHome: Task[Option[PathRef]]` task to the base `ZincWorkerModule` trait that can be overridden to set the JVM to an arbitrary path. we could also expose a configuration option to change the url we fetch the index from or make the in-memory `JvmIndex` mapping a configuration option. I'm not sure how many knobs should be exposed though, so I just added the one `javaHome` option.

> The cross module stuff is lazily initialized then cached, so it should be OK to perform heavy downloads and initialization, but is it possible to move the download into a task? That would help benefit from parallelization and log handling and error reporting and all that

How do you make the cross value for a cross module the output of a task? I'm not sure how, since it needs to be invoked outside of a `Task { ... }` block

lihaoyi:
Ah I see, you mean the index itself, and not the JVMs that the index references.

I think what we should do here is avoid using a Cross module to list out the various `ZincWorkerModule.ForJvm`s, and instead just make it something that the user has to instantiate in their build. Something like:

```scala
object myZincWorker = new ZincWorkerModule.ForJvm("graalvm-community:23.0.0") 
object core extends JavaModule {
  override def zincWorker = ModuleRef(myZincWorker)
  object test extends JavaTests with TestModule.Junit4
}
```

That is a tiny bit more boilerplate in the build file, but avoids the need for Mill to know about the contents of the coursier index entirely, and avoids having a huge cross module inside `ZincWorkerModule` that is mostly unused.

We still want to provide a way to download the index and list out its entries for debugging purposes, but that can be a normal task or command since it won't change the shape of the moduletree/taskgraph

lihaoyi:
@alexarchambault one question about https://github.com/coursier/jvm-index: do we offer any guarantees on that? e.g. does Coursier need to fetch it from Github every time? I'm concerned about having an operational dependency here, e.g. I don't want people's Mill builds to start failing if Github has an outage. 

To avoid a dependency on Github, maybe we can publish the index to maven central somewhere, since that's already a piece of infrastructure we are bound to? Or we can bundle snapshots of it with Mill when we make releases

alexarchambault:
It used to be pushed to Maven Central too, some time ago. This can always be re-instated (https://github.com/coursier/jvm-index/issues/277).

alexarchambault:
> Come to think of it, if the index is bundled with Mill, then having it used to initialize the Cross module becomes not an issue at all

It's up to you, but beware that it's getting more and more heavyweight (~1.8 MB uncompressed, 110 kB gzipped)

alexarchambault:
@albertpchen It looks like you inlined quite some code from coursier-jvm. Is it because of the shapeless issue you mentioned? It seems it's pulled by coursier-jvm by mistake (or not to break bin compat, I don't recall). You can safely exclude it manually when you pull it in Mill.

lihaoyi:
@alexarchambault I think 110kb is fine to bundle, the other issue is that it would fall stale relatively wuickly as new versions are added

I glanced through the indices, and it seems like a lot of the download paths are computed based on the artifact version. Could the logic to compute these be bundled into Mill, rather than bundling the generated JSON? That would have the effect that new versions of known distirbutions would automatically be supported, and we would only need to make a new release if new distributions are added or the URL pattern for a distribution changes (both of which should be much rarer than existing distirbutions adding new versions)

alexarchambault:
> I glanced through the indices, and it seems like a lot of the download paths are computed based on the artifact version

Somehow, except there are often subtleties. In `https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.7+7/OpenJDK17U-jdk_x64_mac_hotspot_17.0.7_7.tar.gz`, `17.0.7+7` is the "build" version, and it ends up as both `17.0.7+7` and `17.0.7_7` in the URL.

lihaoyi:
> > I glanced through the indices, and it seems like a lot of the download paths are computed based on the artifact version
> 
> Somehow, except there are often subtleties. In `https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.7+7/OpenJDK17U-jdk_x64_mac_hotspot_17.0.7_7.tar.gz` for example, `17.0.7+7` is the "build" version, and it ends up as both `17.0.7+7` and `17.0.7_7` in the URL.

Yes, but that is accounted for by the Scala code in coursier/jvm-index right? So if we bundle the few kb of classfiles we get all these special cases included, as well as support for any future versions (unless they add additional special cases)

alexarchambault:
> Yes, but that is accounted for by the Scala code in coursier/jvm-index right?

What do you mean? Right now, the index basically look like this:
```
$ cat "$(cs get https://github.com/coursier/jvm-index/raw/master/index.json)" | jq '.darwin.amd64["jdk@temurin"]'
…
  "1.17": "tgz+https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17%2B35/OpenJDK17-jdk_x64_mac_hotspot_17_35.tar.gz",
  "1.17.0.1": "tgz+https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.1%2B12/OpenJDK17U-jdk_x64_mac_hotspot_17.0.1_12.tar.gz",
  "1.17.0.10": "tgz+https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.10%2B7/OpenJDK17U-jdk_x64_mac_hotspot_17.0.10_7.tar.gz",
  "1.17.0.11": "tgz+https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.11%2B9/OpenJDK17U-jdk_x64_mac_hotspot_17.0.11_9.tar.gz",
  "1.17.0.12": "tgz+https://github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.12%2B7/OpenJDK17U-jdk_x64_mac_hotspot_17.0.12_7.tar.gz",
…
```

There's some logic in the coursier-jvm module, so that the `1.` is ignored, and versions such as just `17` actually match the highest `17.x` version… if there's no exact match (there's one here, but `17.0` will actually match the highest `17.0.x` version)

alexarchambault:
But we can probably "compress" that a bit, with a pattern and "build" version listings (in the snippet above, `17+35`, `17.0.1+12`, etc., rather than `17`, `17.0.1`, etc.) and allow users to pass newer build versions if they'd like. But these build versions don't look handy…

lihaoyi:
@alexarchambault i mean the scala code in https://github.com/coursier/jvm-index/tree/master/src; isnt that what generates the json?

alexarchambault:
> i mean the scala code in https://github.com/coursier/jvm-index/tree/master/src; isnt that what generates the json?

It does, yeah

lihaoyi:
Could that coxe be bundled with Mill to generate the download URL from a jdk version  that gets dynamically passed in? I see some parts of it are passing github tokens around, but its unclear to me how much of the logic needs to talk to github vs how much is just mangling strings locally

albertpchen:
> @albertpchen It looks like you inlined quite some code from coursier-jvm. Is it because of the shapeless issue you mentioned? It seems it's pulled by coursier-jvm by mistake (or not to break bin compat, I don't recall). You can safely exclude it manually when you pull it in Mill.

yeah thanks for the tip. I've updated the PR

main/package.mill
```diff
@@ -100,7 +100,13 @@ object `package` extends RootModule with build.MillStableScalaModule with BuildI
 
   object util extends build.MillStableScalaModule {
     def moduleDeps = Seq(api, client)
-    def ivyDeps = Agg(build.Deps.coursier, build.Deps.jline)
+    def ivyDeps = Agg(
+      build.Deps.coursier,
+      build.Deps.coursierJvm.exclude(
+        "com.github.alexarchambault" -> "argonaut-shapeless_6.3_2.13",
```
lihaoyi, line 106: Please add a comment here explaining why it is excluded, for future reference

scalalib/test/src/mill/scalalib/HelloJavaTests.scala
```diff
@@ -47,6 +63,23 @@ object HelloJavaTests extends TestSuite {
...
+      val Right(result) = eval.apply(HelloJavaJavaHomeOverride.core.compile)
+
+      val coreClassFile = os.walk(result.value.classes.path).find(_.last == "Core.class")
+
+      assert(
+        coreClassFile.isDefined,
+
+        // The first eight bytes are magic numbers followed by two bytes for major version and two bytes for minor version
+        // We are overriding to java 23 which corresponds to class file version 67
+        os.read.bytes(coreClassFile.get, 4, 4).toSeq == Seq[Byte](0, 0, 0, 67)
```
lihaoyi, line 79: Let's also add a test to exercise `.test` and `.run` and verify that `System.getProperty("java.version")` returns the expected value. We should also exercise this for two different java versions, to verify that it's not just picking up the environmental default

lihaoyi:
So I looked up how Gradle does this, and it appears they depend on the Adoptium service API https://api.adoptium.net/. I'm reluctant to have an additional service dependency from Mill, because it means that we need to start worrying about rate limiting, uptime, SLAs, outages, etc.. "Pull from latest master on Github" also definitely counts as "service dependency", because a Github outage or someone fat-fingering a bad push would then be able to cause an outage in Mill users.

1. @alexarchambault can we re-enable publishing stable versions of coursier/jvm-index to maven central? That would give us an immutable artifact to depend on and provide stronger guarantees than pulling stuff directly off of github, improving hermeticity/reproducibility reducing the blast radius if anything goes wrong (github down, someone fat-fingers a push, etc.). We pay a price of staleness, but I think that's a tradeoff that's worth it.

2. Once that's done, we move this PR's runtime dependency to maven central instead of github/jvm-index master, and we can merge this and close out the bounty.

3. We can have a longer-term follow up discussion of how to reduce the staleness issue. I have some ideas, e.g. breaking up the JVM index into separate sonatype artifacts per-version, so that newer artifact version-numbers/download-urls can be added piecemeal and be picked up by Mill/Coursier without having to re-publish the whole index and bump the index version in Mill. But that can be done later and shouldn't block the merging of this PR

scalalib/test/src/mill/scalalib/HelloJavaTests.scala
lihaoyi: Should overriding the java version also affect `.launcher` and `.assembly`? 

* `launcher` is explicitly not meant to be portable: it's meant to be something you run on the same machine where it was built, so any JVM you installed would still be present

* `assembly` is meant to be portable, e.g. run on a different machine from which it was built, so we probably can't rely on the configured java home for that

lefou (commented):
This should come with documentation and examples:
* How to configure a specific Java version?
* How to find available Java versions?
* What's the default?
* How can I use a locally installed JVM which path is hold in a system environment variable `MY_PROJECT_JAVA_HOME`?

lihaoyi:
thanks @alexarchambault!

lihaoyi:
> Right now, the coursier/jvm-index code recomputes the whole index every time it runs (it doesn't use previously computed indices). But maybe we could add a way to make it get just new entries for specific JVMs, and that could be used by Mill (and other coursier-jvm users too) 🤔
>
> Plus that would probably need less GitHub API requests, so that a GitHub token would probably not be needed.

Yeah that sounds like it should be doable: we probably still want a way to do a clean build, but if we can isolate the incremental portion of the logic sufficiently, we can ship snapshot+incremental-logic with Mill and it will be able to extrapolate for new JVM versions using the same logic a clean build would do. That would ensure the logic doesn't fall out of sync, and remains consistent and easy to maintain

alexarchambault:
One thing to note though: even if we get the index from Maven Central, the JVMs themselves often get downloaded from GitHub (for Temuring and GraalVM at least)

alexarchambault:
@albertpchen coursier `2.1.15`, with support for the new indices (https://github.com/coursier/jvm-index/pull/297, https://github.com/coursier/coursier/pull/3127, https://github.com/coursier/coursier/pull/3128), is out.

Something like this should allow to load new indices:
```scala
JvmIndex.load(
  cache = FileCache(), // the coursier.cache.Cache instance to use
  repositories = Resolve().repositories, // repositories to use
  indexChannel = JvmChannel.central(), // use new indices published to Maven Central
  os = None, // use defaults
  arch = None // use defaults
)
```

albertpchen:
@alexarchambault thanks! I've updated to use the new indices.

@lihaoyi I added three new main methods to `scalalib`: `DiscoverTestsMain`, `GetTestsTasksMain`, and `DiscoverMainClassesMain`. We need to load compiled class files to discover main and test classes, but since we aren't using the same jvm as mill, they might have an incompatible class file version. So I've changed those tasks to fork a process to call these main methods when we are using a different jvm.

lihaoyi:
@albertpchen thanks for your patience, will take a look!

lihaoyi:
@albertpchen @alexarchambault perhaps questions for both of you:

1. What happens for partially qualified names? e.g. I know what `"temurin:11.0.24"` is, but what about `"temurin:11.0"` or `"temurin:11"`? Do they pick the most recent 11.x.y, or oldest 11.x.y, or most-recent-at-time-of-coursier-index-update 11.x.y, or most-recent-at-time-of-coursier-index-update-when-mill-was-published 11.x.y?

2. Is there a way to list out possible versions to use?

3. If there is a downloadable JVM somewhere on the internet that isn't in the coursier JVM index, how do I use that?

lihaoyi:
@alexarchambault you should probably take a pass at reviewing this as well since you are the most familiar with Coursier and are also familiar with Mill

main/package.mill
alexarchambault: Note that bumping coursier to `2.1.15` or higher should make that exclude unnecessary (trying that in https://github.com/com-lihaoyi/mill/pull/3943)

alexarchambault: Seems you did it already in this PR, so the exclude can be removed

main/util/src/mill/util/CoursierSupport.scala
```diff
@@ -164,6 +167,47 @@ trait CoursierSupport {
...
+  ): Result[os.Path] = {
+    val coursierCache0 = coursierCache(ctx, coursierCacheCustomizer)
+    val jvmCache = JvmCache()
+      .withArchiveCache(
+        ArchiveCache().withCache(coursierCache0)
+      )
+      .withIndex(JvmIndex.load(
+        cache = coursierCache0, // the coursier.cache.Cache instance to use
+        repositories = Resolve().repositories, // repositories to use
+        indexChannel = JvmChannel.central(), // use new indices published to Maven Central
```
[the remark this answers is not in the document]
alexarchambault: Not much, maybe I should add one. The new indices live under [this organization](https://repo1.maven.org/maven2/io/get-coursier/jvm/indices)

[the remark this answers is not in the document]
lihaoyi: Yes please do. Especially for the questions I asked below, people will want to know how this stuff works in case they need to debug it when it goes wrong

alexarchambault:
> 1. What happens for partially qualified names? e.g. I know what `"temurin:11.0.24"` is, but what about `"temurin:11.0"` or `"temurin:11"`? Do they pick the most recent 11.x.y, or oldest 11.x.y, or most-recent-at-time-of-coursier-index-update 11.x.y, or most-recent-at-time-of-coursier-index-update-when-mill-was-published 11.x.y?

I think I didn't document that 🤔 This looks first for the most recent unpacked on disk 11.x.y (which may or may not be the highest in the index in cache). If there's none, coursier-jvm tries to download and unpack the highest 11.x.y in the index. This can be changed via the API: ensuring `javaHome.update` is true (via `javaHome.withUpdate(true)`) makes coursier-jvm use the highest 11.x.y in the index (downloading and unpacking it if necessary).

The index itself is obtained via the coursier cache. This follows a [TTL](https://get-coursier.io/docs/ttl), so that the index may not be the latest at this time (but it shouldn't be more than TTL-old). If the index is older than TTL, coursier-jvm tries to update the index (so it gets most up-to-date then).

alexarchambault:
> 2\. Is there a way to list out possible versions to use?

On the command-line, yes, with `cs java --available --jvm-index cs-central`

Via the API, `JvmIndex#available()` allows to list available JVMs.

alexarchambault:
> 3\. If there is a downloadable JVM somewhere on the internet that isn't in the coursier JVM index, how do I use that?

URLs of JVM archives are accepted as JVM ids, instead of `temurin:21` say

example/package.mill
```diff
@@ -36,6 +36,7 @@ object `package` extends RootModule with Module {
     object linting extends Cross[ExampleCrossModuleJava](build.listIn(millSourcePath / "linting"))
     object publishing extends Cross[ExampleCrossModuleJava](build.listIn(millSourcePath / "publishing"))
     object web extends Cross[ExampleCrossModule](build.listIn(millSourcePath / "web"))
+    object `java-home` extends Cross[ExampleCrossModule](build.listIn(millSourcePath / "java-home"))
```
lihaoyi, line 39: Let's combine these docs under the `example.javalib.module` folder

example/javalib/java-home/1-downloading-jvms/build.mill
```diff
@@ -0,0 +1,29 @@
+// :link-jvm-indices: https://github.com/coursier/jvm-index/tree/master/indices
```
lihaoyi, line 1: Let's also include a version of the `downloading-jvms` example for scalalib and kotlinlib

lihaoyi:
Got it. I think as next steps:

- @alexarchambault we need a documentation page for how the JVM indices work: 
    - what Maven Central org they're under, 
    - how they use maven-metadata.xml, 
    - how to use an explicit JDK index version, 
    - what the SLA is in terms of freshness, etc.

- @albertpchen:
    - We should pin the JDK index version by default to ensure reproducibility and make "download latest index" an opt-in flag/override
    - We should expose `JvmIndex#available()` as a Mill task `show availableJvms` or something and document it for people to use so they can discover what JVMs they can use 
    - We should show an example of using a full JVM download URL as the JVM version, in addition to the current examples of using the `name:version` string and local filesystem path
    - We should link to @alexarchambault 's documentation page from our example docs

main/util/src/mill/util/CoursierSupport.scala
```diff
@@ -164,6 +167,47 @@ trait CoursierSupport {
     (deps0, res.getOrThrow)
   }
 
+  def jvmIndex(
+      ctx: Option[mill.api.Ctx.Log] = None,
+      coursierCacheCustomizer: Option[FileCache[Task] => FileCache[Task]] = None
+  ): Result[JvmIndex] = {
+    val coursierCache0 = coursierCache(ctx, coursierCacheCustomizer)
+    JvmIndex.load().unsafeRun()(coursierCache0.ec)
```
alexarchambault, line 175: You should use a `JvmIndex.load` [overload](https://github.com/coursier/coursier/blob/f1840c21352aeeee3b3008d4a728156895d33196/modules/jvm/src/main/scala/coursier/jvm/JvmIndex.scala#L183-L187) that accepts a coursier cache instance (and pass it `coursierCache0`). Note that [I'm](https://github.com/coursier/coursier/pull/3160) is going to undeprecate the linked overload, it's fine to use it.

main/util/src/mill/util/CoursierSupport.scala
```diff
@@ -164,6 +167,47 @@ trait CoursierSupport {
...
+      id: String,
+      ctx: Option[mill.api.Ctx.Log] = None,
+      coursierCacheCustomizer: Option[FileCache[Task] => FileCache[Task]] = None
+  ): Result[os.Path] = {
+    val coursierCache0 = coursierCache(ctx, coursierCacheCustomizer)
+    val jvmCache = JvmCache()
+      .withArchiveCache(
+        ArchiveCache().withCache(coursierCache0)
+      )
+      .withIndex(JvmIndex.load(
```
alexarchambault, line 193: Shouldn't you use the `jvmIndex` method above here?

lihaoyi:
Took a pass at cleaning up the PR, I think it's in pretty good state now. @alexarchambault and @lefou can take another review pass.

One existing limitation is that setting the JVM version does not apply to Scala module compilation, as AFAICT Zinc does not support setting the javaHome for compiling Scala https://github.com/sbt/zinc/discussions/1498. We may need to manually fork and manage our own JVM subprocesses to make that work, which I think we can leave to a follow up

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -1126,14 +1130,18 @@ trait JavaModule
   )
 
   @internal
-  def bspJvmBuildTarget: JvmBuildTarget =
+  def bspJvmBuildTarget: Task[JvmBuildTarget] = Task.Anon {
```
lefou, line 1133: I think changing this signature might produce trouble, since plugins extending `JavaModule` in `trait`s could break.

lefou (commented):
Looks good from an quick scan. One concern about bin compat; see below.

scalalib/src/mill/scalalib/JavaModule.scala
lihaoyi: I'll leave it in place and add a new method

scalalib/src/mill/scalalib/RunModule.scala
```diff
@@ -44,7 +44,26 @@ trait RunModule extends WithZincWorker {
...
+        streamOut = false
+      )
+      if (process.exitCode == 0) {
+        val a = process.out.lines()
+        a
+      } else {
+        throw new Exception(
+          "Discover main classes subprocess failed (exit code " + process.exitCode + ")"
+        )
+      }
```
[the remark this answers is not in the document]
lihaoyi: I'll go with `check = true` and remove the  special handling instead. This should "never" fail, and if it does better to have a verbose stack trace than an anemic custom error

lihaoyi:
Thanks for your contribution @albertpchen! I'll pay out half the bounty for this PR since we split the work, email me your international bank transfer details to haoyi.sg@gmail.com and I will transfer the funds

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -1131,7 +1131,15 @@ trait JavaModule
...
-  def bspJvmBuildTarget: Task[JvmBuildTarget] = Task.Anon {
+  @deprecated("Use bspJvmBuildTargetTask instead", "0.12.3")
+  def bspJvmBuildTarget: JvmBuildTarget =
+    JvmBuildTarget(
+      javaHome = Option(System.getProperty("java.home")).map(p => BspUri(os.Path(p))),
+      javaVersion = Option(System.getProperty("java.version"))
+    )
+
+  @internal
+  def bspJvmBuildTargetTask: Task[JvmBuildTarget] = Task.Anon {
```
lefou, line 1142: This new impl will miss any customizations. It should instead reuse the value of deprecated method.

albertpchen:
@lihaoyi sorry about the slow responses, and thanks for finishing up this PR for me! I think coursier-jvm is doing a lot of the heavy lifting here, so if you'd like to give the bounty to @alexarchambault that's fine with me.

lihaoyi:
@albertpchen no worries! If you don't want to take the bounty, I'll pass it to the Scala Center to support their good work

Issue #502, opened 2018-12-11
Redesign tasks that get an evaluator instance


lefou:
See https://github.com/lihaoyi/mill/pull/480#issuecomment-436927941
and https://github.com/lihaoyi/mill/issues/491#issuecomment-444006357

Currently, `all`, `show`, `resolve`, `clean` and other built-in tasks take an evaluator instance and are part of the task graph itself, which is somewhat in-elegant. They take part of caching, force the existence of the `out`  folder (even the `clean` task) and trigger dependency resolution. Also, the interaction and ordering in case of parallel task execution is unclear.

lihaoyi:
I think all those tasks should not be `Task`s. That means:

- Not have automatic caching
- Not create an out folder
- Not need an Evaluator to run
- Not take part in the dependency analysis

Essentially, I'd like them to be "dumb" main/entrypoint methods. If they need to do stuff with the `Evaluator`, they can take an `Evaluator` instance and call `evaluate` themselves.

lefou:
Do we want to support an `Evaluator` magic task parameter at all? How do we want to contribute (by plugins) tasks like `mill.scalalib.GenIdea/idea`in the future?

lihaoyi:
I think `Evaluator` would need to not be a magic-task-parameter, but a magic-main-method-parameter; we can set up the implicits tasks and main methods can share most of the command-line deserializers, but can also have some that are unique to each (though for now I can only think of `Evaluator`)

`mill.scalalib.GenIdea/idea` would be a main-method rather than a T.command

lefou:
To call a main-method via cmdline, we need some proper syntax.

As buildfiles and external modules are just valid Scala, there should be no name collision when using the same syntax as for targets. When parsing the cmdline, we need to decide if the first parameter is a main-method or a target. But I have no idea how to elegantly solve this for built-in mains like `clean`, `all`, etc. Without some prefix or namespace, the user would expect those to be either targets or main-methods in `build.sc`.

* `mill mill.scalalib.GenIdea/idea` refers to a main-method or target in an `ExternalModule` named `mill.scalalib.GenIdea`. 
* `mill my-cmd` refers to a main-method or target in a buildfile `build.sc`

A magic `Evaluator` parameter is only accepted, if the def is not a target but a simple scala method (aka main-method).

How could we map built-ins to avoid name collisions:
* Move them into a namespace, e.g. `mill.clean`, `mill.version`, `mill mill.all t1 t2`
[...]

Issue #3359, opened 2024-08-09
`mill.scalalib.Dependency/showUpdates` seems much slower due to forced non-parallel execution


lefou:
I noticed this is a project where I regularily run:

```
> mill -j 0 mill.scalalib.Dependency/showUpdates --format PerDependency
...
[1/2] mill.scalalib.Dependency.updates > Resolving dependencies [99/109]: module.document.test
...
[1/2] mill.scalalib.Dependency.updates > Analyzing dependencies [59/1761]: module.document.test / org.fedorahosted.tennera:jgettext
...
```

The projects has a lot of modules and dependencies. The passed evaluator now seems to run with only one job. I spend some time to improve the performance of `showUpdates`, so it should ran much faster with parallelization turned on.

Related:
* https://github.com/com-lihaoyi/mill/pull/3014
* https://github.com/com-lihaoyi/mill/pull/3265

We probably needs some way to easily enable parallelism for evaluator commands. E.g. Tasks run with `show` will otherwise always run in single-job mode.

lihaoyi:
I guess we can configure the evaluator that gets passed to an evaluator command differently for the evaluator that runs it? Only the evaluator running `show` needs to be sequential, the evaluator that gets passed to it can follow the normal `--jobs` config

lihaoyi:
I'm actually not seeing the evaluator commands causing nested evaluations to run sequentially. Seems running `show __.compile` still seems to compile things in parallel, and `effectiveThreadCount` is still set to the number of cores :

```diff
lihaoyi mill$ git diff
diff --git a/main/eval/src/mill/eval/GroupEvaluator.scala b/main/eval/src/mill/eval/GroupEvaluator.scala
index 3ee01997a9..129526d572 100644
--- a/main/eval/src/mill/eval/GroupEvaluator.scala
+++ b/main/eval/src/mill/eval/GroupEvaluator.scala
@@ -41,6 +41,8 @@ private[mill] trait GroupEvaluator {
   val effectiveThreadCount: Int =
     this.threadCount.getOrElse(Runtime.getRuntime().availableProcessors())

+  pprint.log(effectiveThreadCount)
+
   /**
    * Synchronize evaluations of the same terminal task.
    * This isn't necessarily needed for normal Mill executions,
```

```scala
lihaoyi mill$ ./mill -i dev.run example/scalabuilds/9-realistic -i show __.compile
[2035/2035] dev.run
[...]

PR #3717, opened 2024-10-11
[WIP] Make `Evaluator` only available to `Task.Command(exclusive = true)`

lihaoyi:
This is a breaking change that tightens things up a bit following the introduction of `Task.Command(exclusive = true)`: now that we run commands in parallel, running multiple evaluator commands in parallel is dangerous, so they should be limited to `exclusive` commands

It's another step toward https://github.com/com-lihaoyi/mill/issues/502, by declaring that `exclusive` commands are different from normal commands, and thus need special treatment

Had to jump through some hoops to do this without breaking bincompat:

* We inject the Evaluator parameter during the resolution phase, before evaluation starts, 
* At that time, we do not yet know whether a command is exclusive or not.
* The solution is to instead pass a `ProxyEvaluator` that exposes all the same API, but looks up the current `Evaluator.currentEvaluator` on-demand
* `Evaluator.currentEvaluator` that only exists within exclusive commands, so we can fail with a nice error message if someone tries to use the evaluator inside a non-exclusive command
* We cannot remove other usages of `Evaluator.currentEvaluator`, 
    * e.g. in the `mill.main.Tasks.TokensReader` actually needs to resolve the tasks early on and return a `Seq`, which we cannot easily stub out.
    * But that's probably OK, because it doesn't expose the entire `Evaluator`, and the tasks themselves are immutable and probably safe to use concurrency


lihaoyi:
This might not be possible without breaking bincompat due to the fact that we inject the Evaluator parameter during the resolution phase, before evaluation starts.

lefou:
This would defeat the initial use case, that let us to introduce the `exclusive` tag in the first place.

* https://github.com/com-lihaoyi/mill/issues/3359

`Dependency/showUpdates` is an evaluator command, that is non-exclusive but profits heavily from being parallel-runnable.

lihaoyi:
@lefou If I'm not mistaken `Dependency/showUpdates` should be fine, since only the top-level `mill.scalalib.Dependency/showUpdates` command needs to be exclusive. Internally, `showUpdates` ends up calling a bunch of `javaModule.*` targets, and those both are neither exclusive nor are they evaluator commands. So I think it should continue to work unchange even if we land this limitation

PR #3721, opened 2024-10-12
Cherry-pick source compatible Scala 3 changes

lihaoyi:
Pulls in some changes from https://github.com/com-lihaoyi/mill/pull/3369 to try and reduce the size of that diff and simplify future backports


lihaoyi:
CC @bishabosha. For the upstream PR https://github.com/com-lihaoyi/mill/pull/3369, not sure whether you want to keep rebasing it, or whether just start merging in latest `main` (which might be less work?). I expect to follow this PR up with a few more PRs cherry picking the source/bin-compatible versions of that PR into `main`, so may be a hassle to shuffle each of those updates through a long rebase process

bishabosha:
I squashed a lot of commits in https://github.com/com-lihaoyi/mill/pull/3369 so that should make it simpler

PR #3723, opened 2024-10-12
Add Kotest support for Kotlin/JS

0xnm:
This PR brings [Kotest](https://github.com/kotest/kotest) test engine support for Kotlin/JS target test execution.

Replaces #3710.


kotlinlib/src/mill/kotlinlib/js/KotlinJSModule.scala
```diff
@@ -379,16 +382,23 @@ trait KotlinJSModule extends KotlinModule { outer =>
...
   // we already have cross-modules in the paths, so we don't need them here
-  private def moduleName() = millModuleSegments.value
-    .filter(_.isInstanceOf[Segment.Label])
-    .map(_.asInstanceOf[Segment.Label])
-    .last
-    .value
+  private def moduleName() = {
+    val labelSegments = millModuleSegments.value
+      .filter(_.isInstanceOf[Segment.Label])
+      .map(_.asInstanceOf[Segment.Label])
```
lihaoyi, line 388: This can be simplified to millModuleSegments.value.collect{case label: Segment.Label => label}`. Same below

example/kotlinlib/web/3-hello-kotlinjs/src/foo/Hello.kt
```diff
@@ -12,5 +12,5 @@ fun main() {
 }
 
 fun hello(): String {
-    return createHTML().h1 { +"Hello World" }.toString()
-}
\ No newline at end of file
+    return createHTML(prettyPrint = false).h1 { text("Hello World") }.toString()
```
lihaoyi, line 15: What's the difference between these two snippets? Is the new one more idiomatic than the old one?

lihaoyi:
Looks good to me, if I understand correctly the core thing that was missing was the `kotestProcessor`/`kotlincOptions`. 

Left some nits, but otherwise looks good

example/kotlinlib/web/3-hello-kotlinjs/src/foo/Hello.kt
0xnm: It is effectively the same: `+` == `text` call. I find `+` syntax a bit weird, but in my other PR I used it. Pretty print is disabled just to avoid adding line breaks after the tag to simplify string assertion.

0xnm:
Actually the 2 pieces were missing:

* Kotest compiler extension https://github.com/kotest/kotest/blob/master/kotest-framework/kotest-framework-multiplatform-plugin-embeddable-compiler/src/jvmMain/kotlin/io/kotest/framework/multiplatform/embeddablecompiler/JsTransformer.kt
* Even if applied, tests won't run without https://mochajs.org/ added (not sure why)

lihaoyi:
Thanks @0xnm !

Issue #3670, opened 2024-10-05
Write an example Mill build for `arrow-kt/arrow` (1500USD Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 1500USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

The goal of this ticket is to write an example Mill build for https://github.com/arrow-kt/arrow, in the spirit of the other example Mill builds for:

- Mockito https://github.com/com-lihaoyi/mill/tree/main/example/thirdparty/mockito
- Netty https://github.com/com-lihaoyi/mill/tree/main/example/thirdparty/netty
- Gatling https://github.com/com-lihaoyi/mill/tree/main/example/thirdparty/gatling

The other three examples cover Java and Scala for Gradle/Maven/SBT, the end goal of writing an example for arrow-kt is to have a reference example for Kotlin and Gradle-Kotlin

The goal of the example build isn't to reach 100% compatibility as a drop-in replacement for the original build. Rather, it is to be able to demonstrate a "representative" build: able to do most of the core, interesting functionality, without necessarily reproducing every quirk of the original. 

In particular, the example build should be able to compile, package, and run most existing tests. The only exception would be functionality not supported by Mill: e.g. using different JVM versions (at least until https://github.com/com-lihaoyi/mill/issues/3480 lands), Kotlin-JS (until https://github.com/com-lihaoyi/mill/issues/3611 lands), or android stuff (until https://github.com/com-lihaoyi/mill/issues/3550 lands).

Arrow's gradle build has 1 root `build.gradle.kts` and 20 sub-folder `build.gradle.kts` files, which should translate into 1 Mill `RootModule` and 20 nested modules.

Once the build is complete, and all modules compile and pass tests, the pull request should come with an ad-hoc performance comparison between the old Gradle build and the new example Mill build, in the spirit of the existing comparisons with Maven https://mill-build.org/mill/comparisons/maven.html Gradle https://mill-build.org/mill/comparisons/gradle.html and SBT https://mill-build.org/mill/comparisons/sbt.html. It should also contain a writeup of what parts of the build were _not_ covered in the example, and why (e.g. due to features not supported in Mill)

chaitanyawaikar:
@lihaoyi I would like to give this a shot.

lihaoyi:
@chaitanyawaikar go for it

rahat2134:
[...]

PR #3726, opened 2024-10-13
Example Mill build for `arrow-kt/arrow` project

chaitanyawaikar:
This PR addresses the [issue](https://github.com/com-lihaoyi/mill/issues/3670) 

It provides an example Mill build for [arrow-kt/arrow](https://github.com/arrow-kt/arrow), in the spirit of the other example Mill builds for:

[Mockito](https://github.com/com-lihaoyi/mill/tree/main/example/thirdparty/mockito)
[Netty](https://github.com/com-lihaoyi/mill/tree/main/example/thirdparty/netty)
[Gatling](https://github.com/com-lihaoyi/mill/tree/main/example/thirdparty/gatling)

### DIFF
--- a/example/thirdparty/arrowkt/build.mill
+++ b/example/thirdparty/arrowkt/build.mill
@@ -1,54 +1,72 @@
-package build
-import mill._, scalalib._
+import mill._
+import mill.scalalib._
 
 object Dependencies {
+  val kotlinVersion = "1.9.10"
+
   // Compile dependencies
-  private val arrowCore      = ivy"io.arrow-kt:arrow-core:1.2.0"
-  private val arrowFx        = ivy"io.arrow-kt:arrow-fx:1.2.0"
-  private val arrowSyntax    = ivy"io.arrow-kt:arrow-syntax:1.2.0"
-  private val arrowOptics    = ivy"io.arrow-kt:arrow-optics:1.2.0"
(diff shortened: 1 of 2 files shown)


example/thirdparty/arrowkt/build.mill
```diff
@@ -0,0 +1,72 @@
...
+
+object Dependencies {
+  val kotlinVersion = "1.9.10"
+
+  // Compile dependencies
+  val arrowCore = ivy"io.arrow-kt:arrow-core:1.2.4"
+  val arrowFxCoroutines = ivy"io.arrow-kt:arrow-fx-coroutines:1.2.4"
+  val arrowResilience = ivy"io.arrow-kt:arrow-resilience:1.2.4"
+  val arrowFxStm = ivy"io.arrow-kt:arrow-fx-stm:1.2.4"
+  val arrowCollectors = ivy"io.arrow-kt:arrow-collectors:1.2.0"
```
0xnm, line 12: I'm not sure it should be like that. I guess you need to build these from the sources since they are the part of the repo.

The only external dependencies are declared here https://github.com/arrow-kt/arrow/blob/main/gradle/libs.versions.toml.

lihaoyi: Yes, these should be internal `moduleDeps`

lihaoyi:
I only see 7 `extends KotlinModule` here, whereas `git ls-files | grep .gradle.kts` shows 21 gradle subprojects. Presumably we should have 21 Mill modules as well?

PR #3742, opened 2024-10-15
Refactor Reflect.reflect to try and optimize it a bit

lihaoyi:
JProfile shows it to be a hot path. I simplified the for comprehension, removed one redundant sort, and removed the final toIndexSeq to try and speed it up a bit

Tested by running CI with `assert(old == new)` to ensure it returns the same value in all code paths


lihaoyi:
Doesn't seem to make a measurable difference

Issue #3748, opened 2024-10-15
Investigate why changing the body of a task or method causes downstream files to re-compile


lihaoyi:
In the `CodeSigSubfolderTests` suite, we can see that modifying the body of task or helper method in the upstream `subfolder/package.mill` causes the downstream `build.mill` to recompile:

https://github.com/com-lihaoyi/mill/blob/5f617f053fbbad32072d69e8967a538bbfbd1e5c/integration/invalidation/codesig-subfolder/src/CodeSigSubfolderTests.scala#L42-L73

This is at odds with how incremental compilation is meant to work, and indeed when testing out simple two-file compilation modules standalone, changing the body of a method in an upstream file without changing the signature does not cause the downstream file to recompile

lihaoyi:
Here's a minimal reproduction of the problem with Zinc incremental compilation outside of Mill:


`build.scala`
```scala
package build

object outer {
  def foo = { build.subfolder.inner.helperFoo }
}
```
`subfolder/package.scala`
```scala
package build.subfolder
import build.{outer => unused}

class Unused

object inner  {
  def helperFoo = { 1 }
}
```

* With the `import build.{outer => unused}`, we see changes to `subfolder/package.scala` causing `build.scala` to be recompiled, even though the signatures in `package.scala` did not change
* If `class Unused` is _removed_, we see changes in `build.scala` causing `subfolder/package.scala` to be recompiled, even though it does not depend on `build.scala`

lihaoyi:
Seems like a regression in Zinc 1.10.x, reverting to 1.9.6 fixes the problem

lihaoyi:
https://github.com/sbt/zinc/issues/1461

PR #3751, opened 2024-10-16
Revert zinc to 1.9.6 to fix incremental compilation in the build

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/3748. Workaround for https://github.com/sbt/zinc/issues/1461#issuecomment-2415641212 that disables incremental compilation in multi-file builds 

With this PR, whitespace changes to `build.mill` recompile 1 build file in about 5 seconds, whereas without this PR they would recompile 20 build file taking 10 seconds. Presumably the difference would be larger on larger projects

Updates a number of tests to verify that incremental compilation in build files works correctly

Needed to downgrade Mill's own Scala version to 2.13.13 for compatibility concerns, but this doesn't affect projects built using Mill which can continue to use any Scala version. We may need to find some fix before upgrading to Scala 3.

Not terribly happy with the downgrade. Going to just leave this open for now since it's not a blocker for anything, and we may receive a fix from upstream which would allow us to tighten up the tests while negate the need for this downgrade

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -55,8 +55,8 @@ object Deps {
   // The Scala version to use
   // When updating, run "Publish Bridges" Github Actions for the new version
   // and then add to it `bridgeScalaVersions`
-  val scalaVersion = "2.13.14"
-  val scala2Version = "2.13.14"
+  val scalaVersion = "2.13.12"
+  val scala2Version = "2.13.12"
   // The Scala 2.12.x version to use for some workers
   val workerScalaVersion212 = "2.12.19"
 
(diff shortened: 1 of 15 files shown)


lihaoyi:
With this PR, whitespace changes to `build.mill` recompile 1 build file in about 5 seconds, whereas without this PR they would recompile 20 build file taking 10 seconds. Presumably the difference would be larger on larger projects

lihaoyi:
closing in favor of https://github.com/com-lihaoyi/mill/pull/3773

PR #3756, opened 2024-10-16
Extended `mill init` to support an existing Maven project

### LINKED ISSUE #3449: Support `mill init` from an existing Maven project (1500USD Bounty)
---

**From the maintainer Li Haoyi**: I'm putting a 1500USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

We should be able to run `./mill init` to do a best effort conversion of an existing Maven project to Mill: 

* Code-generate a `build.mill` containing a `MavenModule`for the root `pom.xml`, the `package.mill` for any subfolder `pom.xml`s
* Convert the main and test dependencies on third-party libraries to `ivyDeps`
* Convert inter-subproject dependencies to `moduleDeps`
* Set any necessary `javacOptions`
* Implement `PublishModule` with relevant metadata
* Support POM inheritance, dependency management, etc.

Such a conversion will never be 100%, but if we can automate 80% of the conversion that will already be a huge help for people migrating to Mill or even just trying it out

Some prior art that can be used as reference:

* Gradle's Maven Importer https://docs.gradle.org/current/userguide/build_init_plugin.html#sec:pom_maven_conversion
* SBT's POM Reader Plugin https://github.com/sbt/sbt-pom-reader

ajaychandran:
This PR adds support to convert a Maven build to Mill by generating Mill build file(s) from POM file(s).
The generator output should be considered scaffolding and will likely require edits to complete conversion.

### Capabilities
The conversion
 - handles deeply nested modules
 - captures project metadata
 - configures dependencies for scopes:
   - compile
   - provided
   - runtime
   - test
 - configures testing frameworks:
   - JUnit 4
   - JUnit 5
   - TestNG
 - configures multiple, compile and test, resource directories

### Limitations
The conversion does not support:
 - build extensions
 - build profiles
 - plugins, other than maven-compiler-plugin

### Future work
 Add support for:
 - [os-maven-plugin](https://github.com/trustin/os-maven-plugin)
 - [moditect-maven-plugin](https://github.com/moditect/moditect)
 - [hawtjni-maven-plugin](https://github.com/fusesource/hawtjni/tree/master/hawtjni-maven-plugin)
 
Resolves #3449.


ajaychandran:
For reference, below is generated `build.mill` for the root module:
```scala
package build

import mill._
import mill.javalib._
import mill.javalib.publish._

trait BaseMavenModule extends MavenModule with PublishModule {

  override def pomSettings = PomSettings(
    s"""Super CSV parent project""".stripMargin,
    "Super CSV",
    "https://github.com/super-csv/super-csv",
    Seq(
      License("Apache License, Version 2.0", "Apache License, Version 2.0", "http://www.apache.org/licenses/LICENSE-2.0.html", false, false, "repo")
    ),
    VersionControl(
      Option("https://github.com/super-csv/super-csv"),
      Option("scm:git:git@github.com:super-csv/super-csv.git"),
      Option("scm:git:git@github.com:super-csv/super-csv.git"),
      Option("v2.4.0")
    ),
    Seq(
      Developer("kbg", "Kasper Graversen", "", Option(null), Option(null)),
      Developer("jamesbassett", "James Bassett", "", Option(null), Option(null)),
      Developer("ddv36a78", "Dominique De Vito", "", Option(null), Option(null))
    )
  )

  override def publishVersion = "2.4.0"

  trait Tests extends TestModule.Junit4 with MavenTests {

    override def defaultCommandName() = "test"

    override def ivyDeps = super.ivyDeps() ++ Agg(
      ivy"com.novocode:junit-interface:0.11"
    )
  }
}

object root extends BaseMavenModule {

  override def artifactName = "super-csv-parent"

  override def pomPackagingType = PackagingType.Pom
}
```
Below is one of the `package.mill` files.
```scala
package build.`super-csv`

import mill._
import mill.javalib._
import mill.javalib.publish._

object `package` extends RootModule with build.BaseMavenModule {

  override def artifactName = "super-csv"

  override def javacOptions = Seq(
    "-source",
    "1.5",
    "-target",
    "1.5"
  )


  object test extends Tests {

    override def ivyDeps = super.ivyDeps() ++ Agg(
      ivy"junit:junit:4.12"
    )
  }

}```

**Comment by lihaoyi** [comment]
> For a multi-module project the package.mill files are generated for sub-modules but they are not recognized by Mill. See integration test MillInitMavenTests for details.

@ajaychandran for `package.mill` files to be recognized, you need the magic import `import $packages._` in the top-level `build.mill

> Should a new task be defined for provided dependencies? JavaModule.providedIvyDeps?

IIRC `provided` dependencies correspond to Mill's `compileIvyDeps`

**Comment by ajaychandran** [comment]
@lihaoyi Thanks for the tips.

Now, the conversion fails with
```
super-csv/package.mill:7:71: type BaseMavenModule is not a member of object build_.package_
```

I am not sure why this is happening. Could it be that `build.mill` has a `package` object for publishing the root module?

Below is the output of `generatedScriptSources`.
_build.mill_
```scala
package build_
import _root_.{build_ => $file}
import build_.{package_ => build}
import _root_.mill.runner.MillBuildRootModule
@_root_.scala.annotation.nowarn
object MillMiscInfo extends mill.main.RootModule.Info(
  Vector("/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.13.14/scala-library-2.13.14.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/mill-moduledefs_2.13/0.11.1/mill-moduledefs_2.13-0.11.1.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scalameta/scalafmt-dynamic_2.13/3.8.3/scalafmt-dynamic_2.13-3.8.3.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/modules/scala-xml_2.13/2.3.0/scala-xml_2.13-2.3.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/io/github/alexarchambault/windows-ansi/windows-ansi/0.0.5/windows-ansi-0.0.5.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/io/get-coursier/interface/1.0.19/interface-1.0.19.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/mainargs_2.13/0.7.6/mainargs_2.13-0.7.6.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/requests_2.13/0.9.0/requests_2.13-0.9.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/ch/qos/logback/logback-classic/1.5.7/logback-classic-1.5.7.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/jgrapht/jgrapht-core/1.4.0/jgrapht-core-1.4.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/guru/nidi/graphviz-java-min-deps/0.18.1/graphviz-java-min-deps-0.18.1.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/sourcecode_2.13/0.4.0/sourcecode_2.13-0.4.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/net/java/dev/jna/jna/5.14.0/jna-5.14.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/net/java/dev/jna/jna-platform/5.14.0/jna-platform-5.14.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/eed3si9n/jarjarabrams/jarjar-abrams-core_2.13/1.14.0/jarjar-abrams-core_2.13-1.14.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/scalaparse_2.13/3.1.1/scalaparse_2.13-3.1.1.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/os-lib_2.13/0.11.1/os-lib_2.13-0.11.1.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/upickle_2.13/3.3.1/upickle_2.13-3.3.1.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/pprint_2.13/0.9.0/pprint_2.13-0.9.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/fansi_2.13/0.5.0/fansi_2.13-0.5.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-sbt/test-interface/1.0/test-interface-1.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/jline/jline/3.26.3/jline-3.26.3.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/io/get-coursier/coursier_2.13/2.1.14/coursier_2.13-2.1.14.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-compiler/2.13.14/scala-compiler-2.13.14.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/ow2/asm/asm-tree/9.7/asm-tree-9.7.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/maven-embedder/3.9.9/maven-embedder-3.9.9.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/resolver/maven-resolver-connector-basic/1.9.22/maven-resolver-connector-basic-1.9.22.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/resolver/maven-resolver-supplier/1.9.22/maven-resolver-supplier-1.9.22.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/resolver/maven-resolver-transport-file/1.9.22/maven-resolver-transport-file-1.9.22.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/resolver/maven-resolver-transport-http/1.9.22/maven-resolver-transport-http-1.9.22.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/resolver/maven-resolver-transport-wagon/1.9.22/maven-resolver-transport-wagon-1.9.22.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scalameta/scalafmt-interfaces/3.8.3/scalafmt-interfaces-3.8.3.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scalameta/scalafmt-sysops_2.13/3.8.3/scalafmt-sysops_2.13-3.8.3.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/typesafe/config/1.4.3/config-1.4.3.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/fusesource/jansi/jansi/1.18/jansi-1.18.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/slf4j/slf4j-api/2.0.15/slf4j-api-2.0.15.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/modules/scala-collection-compat_2.13/2.12.0/scala-collection-compat_2.13-2.12.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/geny_2.13/1.1.1/geny_2.13-1.1.1.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/ch/qos/logback/logback-core/1.5.7/logback-core-1.5.7.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/jheaps/jheaps/0.11/jheaps-0.11.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/guru/nidi/graphviz-java/0.18.1/graphviz-java-0.18.1.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/eed3si9n/jarjar/jarjar/1.14.0/jarjar-1.14.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-reflect/2.13.14/scala-reflect-2.13.14.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/fastparse_2.13/3.1.1/fastparse_2.13-3.1.1.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/ujson_2.13/3.3.1/ujson_2.13-3.3.1.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/upack_2.13/3.3.1/upack_2.13-3.3.1.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/upickle-implicits_2.13/3.3.1/upickle-implicits_2.13-3.3.1.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/kohlschutter/junixsocket/junixsocket-native-common/2.10.0/junixsocket-native-common-2.10.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/kohlschutter/junixsocket/junixsocket-common/2.10.0/junixsocket-common-2.10.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/github/plokhotnyuk/jsoniter-scala/jsoniter-scala-core_2.13/2.13.5.2/jsoniter-scala-core_2.13-2.13.5.2.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/io/get-coursier/coursier-core_2.13/2.1.14/coursier-core_2.13-2.1.14.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/io/get-coursier/coursier-cache_2.13/2.1.14/coursier-cache_2.13-2.1.14.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/io/get-coursier/coursier-proxy-setup/2.1.14/coursier-proxy-setup-2.1.14.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/io/github/java-diff-utils/java-diff-utils/4.12/java-diff-utils-4.12.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/ow2/asm/asm/9.7/asm-9.7.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/maven-settings/3.9.9/maven-settings-3.9.9.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/maven-settings-builder/3.9.9/maven-settings-builder-3.9.9.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/maven-core/3.9.9/maven-core-3.9.9.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/maven-plugin-api/3.9.9/maven-plugin-api-3.9.9.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/maven-model/3.9.9/maven-model-3.9.9.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/maven-model-builder/3.9.9/maven-model-builder-3.9.9.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/maven-builder-support/3.9.9/maven-builder-support-3.9.9.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/resolver/maven-resolver-api/1.9.22/maven-resolver-api-1.9.22.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/resolver/maven-resolver-util/1.9.22/maven-resolver-util-1.9.22.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/shared/maven-shared-utils/3.4.2/maven-shared-utils-3.4.2.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/google/inject/guice/5.1.0/guice-5.1.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/google/guava/guava/33.3.0-jre/guava-33.3.0-jre.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/google/guava/failureaccess/1.0.2/failureaccess-1.0.2.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/javax/inject/javax.inject/1/javax.inject-1.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/javax/annotation/javax.annotation-api/1.3.2/javax.annotation-api-1.3.2.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/codehaus/plexus/plexus-utils/4.0.0/plexus-utils-4.0.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/codehaus/plexus/plexus-classworlds/2.8.0/plexus-classworlds-2.8.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/eclipse/sisu/org.eclipse.sisu.plexus/0.9.0.M3/org.eclipse.sisu.plexus-0.9.0.M3.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/codehaus/plexus/plexus-sec-dispatcher/2.0/plexus-sec-dispatcher-2.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/codehaus/plexus/plexus-cipher/2.0/plexus-cipher-2.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/codehaus/plexus/plexus-interpolation/1.27/plexus-interpolation-1.27.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/commons-cli/commons-cli/1.8.0/commons-cli-1.8.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/resolver/maven-resolver-spi/1.9.22/maven-resolver-spi-1.9.22.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/resolver/maven-resolver-named-locks/1.9.22/maven-resolver-named-locks-1.9.22.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/resolver/maven-resolver-impl/1.9.22/maven-resolver-impl-1.9.22.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/maven-resolver-provider/3.9.9/maven-resolver-provider-3.9.9.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/httpcomponents/httpclient/4.5.14/httpclient-4.5.14.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/httpcomponents/httpcore/4.4.16/httpcore-4.4.16.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/commons-codec/commons-codec/1.17.1/commons-codec-1.17.1.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/slf4j/jcl-over-slf4j/1.7.36/jcl-over-slf4j-1.7.36.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/wagon/wagon-provider-api/3.5.3/wagon-provider-api-3.5.3.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/modules/scala-parallel-collections_2.13/1.0.4/scala-parallel-collections_2.13-1.0.4.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/guru/nidi/com/eclipsesource/j2v8/j2v8_linux_x86_64/4.6.0/j2v8_linux_x86_64-4.6.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/google/code/findbugs/jsr305/3.0.2/jsr305-3.0.2.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/slf4j/jul-to-slf4j/1.7.30/jul-to-slf4j-1.7.30.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/ow2/asm/asm-commons/9.6/asm-commons-9.6.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/commons/commons-lang3/3.14.0/commons-lang3-3.14.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/upickle-core_2.13/3.3.1/upickle-core_2.13-3.3.1.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/io/github/alexarchambault/concurrent-reference-hash-map/1.1.0/concurrent-reference-hash-map-1.1.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/io/get-coursier/coursier-util_2.13/2.1.14/coursier-util_2.13-2.1.14.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/io/get-coursier/jniutils/windows-jni-utils/0.3.3/windows-jni-utils-0.3.3.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/codehaus/plexus/plexus-archiver/4.9.0/plexus-archiver-4.9.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/codehaus/plexus/plexus-container-default/2.1.1/plexus-container-default-2.1.1.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/virtuslab/scala-cli/config_2.13/0.2.1/config_2.13-0.2.1.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/maven-repository-metadata/3.9.9/maven-repository-metadata-3.9.9.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/maven/maven-artifact/3.9.9/maven-artifact-3.9.9.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/eclipse/sisu/org.eclipse.sisu.inject/0.9.0.M3/org.eclipse.sisu.inject-0.9.0.M3.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/codehaus/plexus/plexus-component-annotations/2.1.0/plexus-component-annotations-2.1.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/aopalliance/aopalliance/1.0/aopalliance-1.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/codehaus/plexus/plexus-xml/3.0.0/plexus-xml-3.0.0.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/codehaus/plexus/plexus-io/3.4.1/plexus-io-3.4.1.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/commons-io/commons-io/2.16.1/commons-io-2.16.1.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/commons/commons-compress/1.26.2/commons-compress-1.26.2.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/iq80/snappy/snappy/0.4/snappy-0.4.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/tukaani/xz/1.9/xz-1.9.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/com/github/luben/zstd-jni/1.5.5-10/zstd-jni-1.5.5-10.jar", "/home/aj/.cache/coursier/v1/https/repo1.maven.org/maven2/org/apache/xbean/xbean-reflect/3.7/xbean-reflect-3.7.jar", "/home/aj/github/ajaychandran/mill/out/runner/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/runner/client/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/main/server/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/main/codesig/compile.dest/classes", "/home/aj/github/ajaychandran/mill/runner/linenumbers/resources", "/home/aj/github/ajaychandran/mill/out/runner/linenumbers/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/bsp/buildInfoResources.dest", "/home/aj/github/ajaychandran/mill/out/bsp/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/scalanativelib/worker-api/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/scalanativelib/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/scalajslib/worker-api/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/scalajslib/buildInfoResources.dest", "/home/aj/github/ajaychandran/mill/out/scalajslib/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/kotlinlib/worker/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/kotlinlib/buildInfoResources.dest", "/home/aj/github/ajaychandran/mill/out/kotlinlib/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/testrunner/entrypoint/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/testrunner/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/scalalib/api/buildInfoResources.dest", "/home/aj/github/ajaychandran/mill/out/scalalib/api/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/main/resolve/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/main/util/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/main/client/buildInfoResources.dest", "/home/aj/github/ajaychandran/mill/out/main/client/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/main/api/buildInfoResources.dest", "/home/aj/github/ajaychandran/mill/out/main/api/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/main/define/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/main/eval/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/main/buildInfoResources.dest", "/home/aj/github/ajaychandran/mill/out/main/compile.dest/classes", "/home/aj/github/ajaychandran/mill/scalalib/resources", "/home/aj/github/ajaychandran/mill/out/scalalib/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/idea/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/main/init/exampleList.dest", "/home/aj/github/ajaychandran/mill/out/main/init/compile.dest/classes", "/home/aj/github/ajaychandran/mill/out/dist/runClasspath.dest"),
  "/home/aj/github/ajaychandran/mill/out/integration/feature/init/local/testOnly.dest/sandbox/run-2",
  "/home/aj/github/ajaychandran/mill/out/integration/feature/init/local/testOnly.dest/sandbox/run-2/out",
  "/home/aj/github/ajaychandran/mill/out/integration/feature/init/local/testOnly.dest/sandbox/run-2"
)
import MillMiscInfo._

//MILL_ORIGINAL_FILE_PATH=/home/aj/github/ajaychandran/mill/out/integration/feature/init/local/testOnly.dest/sandbox/run-2/build.mill
//MILL_USER_CODE_START_MARKER


import mill._
import mill.javalib._
import mill.javalib.publish._
import _root_._

trait BaseMavenModule extends MavenModule with PublishModule {

  override def pomSettings = PomSettings(
    s"""Super CSV parent project""".stripMargin,
    "Super CSV",
    "https://github.com/super-csv/super-csv",
    Seq(
      License("Apache License, Version 2.0", "Apache License, Version 2.0", "http://www.apache.org/licenses/LICENSE-2.0.html", false, false, "repo")
    ),
    VersionControl(
      Option("https://github.com/super-csv/super-csv"),
      Option("scm:git:git@github.com:super-csv/super-csv.git"),
      Option("scm:git:git@github.com:super-csv/super-csv.git"),
      Option("v2.4.0")
    ),
    Seq(
      Developer("kbg", "Kasper Graversen", "", Option(null), Option(null)),
      Developer("jamesbassett", "James Bassett", "", Option(null), Option(null)),
      Developer("ddv36a78", "Dominique De Vito", "", Option(null), Option(null))
    )
  )

  override def publishVersion = "2.4.0"

  trait Tests extends TestModule.Junit4 with MavenTests {

    override def defaultCommandName() = "test"

    override def ivyDeps = super.ivyDeps() ++ Agg(
      ivy"com.novocode:junit-interface:0.11"
    )
  }
}

abstract class package_  extends RootModule with BaseMavenModule {

  override def artifactName = "super-csv-parent"

  override def pomPackagingType = PackagingType.Pom
}

object package_ extends package_ {
  final lazy val `super-csv-distribution`: _root_.build_.`super-csv-distribution`.package_.type = _root_.build_.`super-csv-distribution`.package_ // subfolder module reference
final lazy val `super-csv-dozer`: _root_.build_.`super-csv-dozer`.package_.type = _root_.build_.`super-csv-dozer`.package_ // subfolder module reference
final lazy val `super-csv-joda`: _root_.build_.`super-csv-joda`.package_.type = _root_.build_.`super-csv-joda`.package_ // subfolder module reference
final lazy val `super-csv`: _root_.build_.`super-csv`.package_.type = _root_.build_.`super-csv`.package_ // subfolder module reference
final lazy val `super-csv-java8`: _root_.build_.`super-csv-java8`.package_.type = _root_.build_.`super-csv-java8`.package_ // subfolder module reference
final lazy val `super-csv-benchmark`: _root_.build_.`super-csv-benchmark`.package_.type = _root_.build_.`super-csv-benchmark`.package_ // subfolder module reference
  override lazy val millDiscover: _root_.mill.define.Discover = _root_.mill.define.Discover[this.type]

}
```
_super-csv/package.mill_
```scala
package build_.`super-csv`
import _root_.{build_ => $file}
import build_.{package_ => build}
object MillMiscSubFolderInfo
extends mill.main.SubfolderModule.Info(
  os.Path("/home/aj/github/ajaychandran/mill/out/integration/feature/init/local/testOnly.dest/sandbox/run-2/super-csv"),
  _root_.scala.Seq("super-csv")
)
import MillMiscSubFolderInfo._

//MILL_ORIGINAL_FILE_PATH=/home/aj/github/ajaychandran/mill/out/integration/feature/init/local/testOnly.dest/sandbox/run-2/super-csv/package.mill
//MILL_USER_CODE_START_MARKER


import mill._
import mill.javalib._
import mill.javalib.publish._

abstract class package_  extends mill.main.SubfolderModule with build.BaseMavenModule {

  override def artifactName = "super-csv"

  override def javacOptions = Seq(
    "-source",
    "1.5",
    "-target",
    "1.5"
  )


  object test extends Tests {

    override def ivyDeps = super.ivyDeps() ++ Agg(
      ivy"junit:junit:4.12"
    )
  }

}
object package_ extends package_ {
  
  
}```

**Comment by lihaoyi** [comment]
@ajaychandran looks like the issue is `trait BaseMavenModule` needs to live within the `object package` for it to get picked up correctly

**Comment by ajaychandran** [comment]

The issue was resolved by adding the `$file` import in `package.mill` files.

One criticism of this solution is that the root task namespace (`mill resolve _`) gets cluttered with the tasks of the root module. This could be a problem if a "name collision" occurs.
On the upside, this allows running tasks like `mill compile` (post initialization). IMO this offers a better user experience.

File: main/init/src/mill/init/maven/codegen.scala [hunk 1]
```diff
@@ -0,0 +1,353 @@
+package mill.init.maven
+
+import mill.*
+import mill.api.Ctx
+import org.apache.maven.model.{Dependency, Model}
+
+import scala.collection.immutable.SortedSet
+import scala.jdk.CollectionConverters.*
+
+private[maven] object codegen {
```
lihaoyi, line 10: Top-level objects should be Uppercase to match the rest of the repo

main/init/test/src/mill/init/maven/InitModuleTests.scala
```diff
@@ -0,0 +1,134 @@
...
+           |    "http://www.example.com",
+           |    Seq(),
+           |    VersionControl(
+           |      Option("http://github.com/gabrielf/maven-samples"),
+           |      Option("scm:git:git@github.com:gabrielf/maven-samples.git"),
+           |      Option("scm:git:git@github.com:gabrielf/maven-samples.git"),
+           |      Option("HEAD")
+           |    ),
+           |    Seq()
+           |  )
```
lihaoyi, line 38: Let's go all the way and use a golden file on disk that we use for comparisons. That way it's much clearer what is being compared (v.s. this long list of) and it can still be easily updated by just copying the generated file over the golden file

main/init/test/src/mill/init/maven/InitModuleTests.scala
```diff
@@ -0,0 +1,134 @@
...
+           |    "",
+           |    "http://www.example.com",
+           |    Seq(),
+           |    VersionControl(
+           |      Option("http://github.com/gabrielf/maven-samples"),
+           |      Option("scm:git:git@github.com:gabrielf/maven-samples.git"),
+           |      Option("scm:git:git@github.com:gabrielf/maven-samples.git"),
+           |      Option("HEAD")
+           |    ),
+           |    Seq()
```
lihaoyi, line 37: Let's generate parameter names here, since that'll make it much clearer what things like `""` and `Seq(0` represent or what the difference is between the two `Option("scm:git:git@github.com:gabrielf/maven-samples.git")`

Also let's use `Some()` instead of `Option()`

main/init/src/mill/init/maven/PomReader.scala
```diff
@@ -0,0 +1,116 @@
...
+    props
+  }
+
+  private def resolver(
+      system: RepositorySystem,
+      session: RepositorySystemSession,
+      remotes: Seq[RemoteRepository],
+      context: String
+  ): ModelResolver =
+    new ModelResolver {
```
lihaoyi, line 79: More idiomatic to make these factory-method-returning-anonymous-class-instance just a subclass `class foo() extends ModuleResolver`

main/init/src/mill/init/maven/codegen.scala
```diff
@@ -0,0 +1,353 @@
...
+    val empty: String = apply("")
+
+    /** Escapes `literal` with quotes. */
+    def apply(literal: String): String =
+      s"\"$literal\""
+
+    /** Escapes `lines` with triple quotes. */
+    def multi(lines: String): String =
+      s"s\"\"\"$lines\"\"\".stripMargin"
+  }
```
lihaoyi, line 352: These should just be normal `def`s probably, not nearly meaty enough to be worth encapsulating into `object`s

main/init/src/mill/init/maven/codegen.scala
```diff
@@ -0,0 +1,353 @@
...
+            Option(dev.getId).getOrElse("")
+          }", "${
+            Option(dev.getName).getOrElse("")
+          }", "${
+            Option(dev.getUrl).getOrElse("")
+          }", Option(${
+            Option(dev.getOrganization).fold("null")(escape)
+          }), Option(${
+            Option(dev.getOrganizationUrl).fold("null")(escape)
+          }))"""
```
lihaoyi, line 61: All the `${}` blocks should probably be one-lined

main/init/src/mill/init/maven/codegen.scala
```diff
@@ -0,0 +1,353 @@
...
+      }
+      val testModule = model0.getDependencies.iterator().asScala.collectFirst {
+        case dep if dep.getGroupId == "junit" => "TestModule.Junit4"
+        case dep if dep.getGroupId == "org.junit.jupiter" => "TestModule.Junit5"
+      }
```
lihaoyi, line 66: Are there any other cases we should worry about here, like `TestNG` or something? Or is `Junit` basically it in terms of Java testing frameworks?

main/init/src/mill/init/maven/codegen.scala
```diff
@@ -0,0 +1,353 @@
...
+    T.log.info("analyzing Maven project ...")
+
+    type ModelNode = (os.Path, Model, BuildPackage)
+    type BuildPackage = String
+
+    // identify Maven models and corresponding Mill packages
+    val (models, modules) = {
+      val (models, modules) = (Seq.newBuilder[ModelNode], Map.newBuilder[ModelId, BuildPackage])
+
+      def loop(pomDir: os.Path, pkg: String): Unit = {
```
lihaoyi, line 27: This should probably be named `recurse`

main/init/src/mill/init/maven/codegen.scala
```diff
@@ -0,0 +1,353 @@
...
+            }${
+              if (runIvyDeps.isEmpty) ""
+              else
+                s"""
+                   |  override def runIvyDeps = ${break.cs.indent(2, "Agg(", runIvyDeps, ")")}
+                   |""".stripMargin
+            }
+             |
+             |$testSource
+             |}""".stripMargin
```
lihaoyi, line 213: There is a lot of code here related to managing indentation and layout, and it really gets in the way of understanding the core logic of what we are generating. 

Let's instead just generate the code naively - i.e. not worrying about managing indentation or proper number of blank lines - and just running `ScalafmtWorkerModule.worker().reformat()` on the generated files (we may need to bump the scalafmt version to the latest release so it can pick up `.mill` files)

integration/feature/init/src/MillInitMavenTests.scala
```diff
@@ -0,0 +1,81 @@
...
+
+import mill.testkit.UtestIntegrationTestSuite
+import utest._
+
+object MillInitMavenTests extends UtestIntegrationTestSuite {
+
+  private def downloadExtractTo(workspace: os.Path, zipUrl: String, zipContentDir: String): Unit = {
+    val zipFile = os.temp(requests.get(zipUrl))
+    val contentDir = os.unzip(zipFile, os.temp.dir()) / zipContentDir
+    os.list(contentDir).foreach(os.move.into(_, workspace))
```
lihaoyi, line 11: It's a subjective style thing, but let's make most of these `foreach` blocks into `for` loops

main/init/src/mill/init/maven/codegen.scala
```diff
@@ -0,0 +1,353 @@
...
+      }
+      val testFrameworkIvyDeps = testModule.collect {
+        case "TestModule.Junit4" => "ivy\"com.novocode:junit-interface:0.11\""
+        case "TestModule.Junit5" => "ivy\"com.github.sbt.junit:jupiter-interface:0.13.0\""
+      }.toSeq
+      val testModuleSource =
+        s"""  trait Tests extends ${testModule.fold("MavenTests")(mod => s"$mod with MavenTests")} {
+           |
+           |    override def defaultCommandName() = "test"
+           |${
```
lihaoyi, line 75: As a first approximation, we should never have nested strings inside `${}` interpolation block, and `${}` blocks should never be multi-line. If they are getting complex, their contents should be pulled out into a named `val` earlier

lihaoyi:
@ajaychandran did a pass and left some comments. Probably will take a few more passes to get it merged, but looks promising.

Some high-level comments:

- Please update the PR description with details of what future maintainers or users would want to read. 
    * For me, I am most interested in the scope of this feature: it obviously cannot convert every Maven build under the sun with all possible plugins, but it can convert simple maven builds, so where is the line drawn for this particular implementation? 
    * There are probably other things that you would expect others sending you a large PR to write in the description

- Add an integration test that takes the Netty codebase we use in https://github.com/com-lihaoyi/mill/blob/main/example/thirdparty/netty/build.mill, runs your converter, and then runs the subset of the netty module tests that work. 
    * Not everything will work - some of the modules require calling `make` and compiling C code or running an H2Spec or Autobahn server - and that is expected. 
    * But we should assert that the modules that are supported do compile, the tests which are supported do pass, and the correct number of test suites are being run (by looking at `show __.discoveredTestClasses`).
    * We should also have comments enumerating the modules/tests that *not* pass, and why it is expected they do not pass (e.g. native C dependencies, requiring external servers running, etc.)

- Let's move all this maven logic and dependencies out of the main classpath into an isolated subprocess or classloader. This will avoid risking artifact collisions in the main classpath
    * You can see how it is done for GraphvizTools https://github.com/com-lihaoyi/mill/blob/main/main/graphviz/src/mill/main/graphviz/GraphvizTools.scala

main/init/src/mill/init/maven/codegen.scala
ajaychandran: Added `TestNg`.

ajaychandran:
@lihaoyi This is ready for another review.
- The PR description has been updated.
- I am not sure where to add the documentation for this feature. Any pointers?

integration/feature/init/src/MillInitMavenTests.scala
```diff
@@ -0,0 +1,221 @@
...
+        ) ==> true
+        initRes.isSuccess ==> true
+
+        val resolveRes = eval(("resolve", "_"))
+        Seq(
+          "dotenv",
+          "dotenv-guice"
+        ).forall(resolveRes.out.contains) ==> true
+
+        // JavaModule.JavaTests is not picking compileIvyDeps from outer module
```
ajaychandran, line 72: This can be fixed by adding the following in `JavaModule.JavaTests`,
```scala
override def compileIvyDeps = outer.compileIvyDeps()
```
Is this the correct solution?

integration/feature/init/src/MillInitMavenTests.scala
```diff
@@ -0,0 +1,221 @@
...
+        val compileRes = eval("__.compile")
+        compileRes.err.contains("package com.google.inject does not exist") ==> true
+        compileRes.isSuccess ==> false
```
ajaychandran, line 75: This can be fixed by adding the following in JavaModule.JavaTests,
```scala
override def compileIvyDeps = outer.compileIvyDeps()
```
Not sure if this is the right approach.

integration/feature/init/src/MillInitMavenTests.scala
```diff
@@ -0,0 +1,221 @@
...
+
+        val resolveRes = eval(("resolve", "_"))
+        Seq(
+          "fastexcel-reader",
+          "fastexcel-writer",
+          "e2e"
+        ).forall(resolveRes.out.contains) ==> true
+
+        // not sure why this happens but pom.xml has a hack to handle JPMS
+        val compileRes = eval("__.compile")
```
ajaychandran, line 104: The [hack](https://github.com/dhatim/fastexcel/blob/de56e786a1fe29351e2f8dc1d81b7cdd9196de4a/pom.xml#L252).

integration/feature/init/src/MillInitMavenTests.scala
```diff
@@ -0,0 +1,221 @@
...
+          "microbench",
+          "bom"
+        ).forall(resolveRes.out.contains) ==> true
+
+        // additional sources defined by build-helper-maven-plugin
+        val sourcesRes = eval(("show", "transport-native-epoll.sources"))
+        Seq(
+          s"$workspacePath/transport-native-epoll/src/main/java",
+          s"$workspacePath/transport-native-epoll/src/main/c"
+        ).forall(sourcesRes.out.contains) ==> true
```
ajaychandran, line 199: In this case, the additional source directory contains `C` sources. This is not useful.
The plugin supports additional directories for compile and test, sources and resources. Should sources be skipped?

integration/feature/init/src/MillInitMavenTests.scala
lihaoyi: My understanding is that `compileIvyDeps` is meant to be non-transitive, similar to Maven `provided` dependencies, but I think having the test submodule inherit them from the parent non-test module sounds reasonable. @lefou what do you think?

integration/feature/init/src/MillInitMavenTests.scala
lihaoyi: Could you elaborate a bit on how this affects us? The comment says `Run Java8 build with JPMS on JDK9+`, but AFAIK we already run Mill CI on JDK >=11, so I thought we would be on JDK9+ already.

integration/feature/init/src/MillInitMavenTests.scala
lihaoyi: Do you know how the `src/main/c/` gets handled by Maven? For example, if we put a `Foo.java` file in there, will it get picked up by the Java compiler? 

If so we can keep it, if not maybe we can skip `sources` and just assume the defaults

lihaoyi:
> I am not sure where to add the documentation for this feature. Any pointers?

Let's add a new top-level section `Migrating to Mill` in `nav.adoc`, underneath `Why Mill?`, with one page `Migrating A Maven Build to Mill` that documents the feature you are adding here: 

* One (small) example test (let's use the https://github.com/dhatim/fastexcel/ repo?)
* Listing the features and limitations of the feature (basically what you added to the PR description, but a bit more verbose)
* Standard caveats and disclaimers (e.g. this is a best effort converter, will never be 100% complete, especially for third-party plugins and customized builds, will likely need some manual touch-up after)

main/maven/src/mill/main/maven/BuildGen.scala
```diff
@@ -0,0 +1,442 @@
...
+      ) = (
+        SortedSet.newBuilder[String],
+        SortedSet.newBuilder[String],
+        SortedSet.newBuilder[String],
+        SortedSet.newBuilder[String],
+        SortedSet.newBuilder[String],
+        SortedSet.newBuilder[String],
+        SortedSet.newBuilder[String],
+        SortedSet.newBuilder[String]
+      )
```
lihaoyi, line 259: Doing these in 8 separate 1-line `val`s will probably be easier to read I think

main/maven/src/mill/main/maven/BuildGen.scala
```diff
@@ -0,0 +1,442 @@
...
+ *  - build profiles are not supported
+ *  - packaging support is limited to
+ *    - jar
+ *    - pom
+ */
+@mill.api.internal
+object BuildGen {
+
+  def main(args: Array[String]): Unit = {
+    implicit val cfg: BuildGenConfig = ParserForClass[BuildGenConfig].constructOrExit(args.toSeq)
```
lihaoyi, line 44: Let's make `cfg: BuildGenConfig` an explicit parameter and just pass it everywhere manually; it's a small enough part of our codebase that passing it explicitly isn't too much of a burden

main/maven/src/mill/main/maven/Plugins.scala
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  private implicit class IteratorDomOps(val self: Iterator[Xpp3Dom]) extends AnyVal {
+
+    def /(name: String): Iterator[Xpp3Dom] =
+      self.map(_.getChild(name))
+
+    def <>(f: String => Unit): Unit =
+      if (null != self) self.foreach(dom => f(dom.getValue))
+  }
```
lihaoyi, line 125: Let's give these non-operator names. How about something like this

```scala
def child
def children
def foreachValue
def foreachChildValue
```

In general we should limit our use of operators to things that will be both used extremely densely (i.e. more than one method call per line) and extremely commonly (e.g. all over a big part of the codebase). I understand what you are trying to do here but I don't think this quite makes the cut, so we should fall back to normal extension methods even if it's a bit more verbose

main/maven/test/resources/--deps-object/pom.xml
```diff
@@ -0,0 +1,226 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
lihaoyi, line 1: Let's remove the `--` from the file name and adjust the test code accordingly. Having files named with leading `-` or `--`s often causes problems from command-line usage, e.g. `rm -rf --deps-object` may fail to delete it, and although there are workarounds its best to just avoid the problem to begin with

integration/feature/init/src/MillInitMavenTests.scala
```diff
@@ -0,0 +1,221 @@
...
+  def tests: Tests = Tests {
+
+    // - Junit5
+    // - maven-compiler-plugin release option
+    test("jansi") {
+      prep("https://github.com/fusesource/jansi/archive/refs/tags/jansi-2.4.1.zip") { tester =>
+        import tester._
+
+        val initRes = eval(("init", "--deps-object", "Deps"))
+        initRes.out.contains(
```
lihaoyi, line 28: I think these checks will all be easier to read converted to normal `assert`s. `==>` is generally meant for user-facing test cases to make them prettier, but these are just internal so `assert` is fine

integration/feature/init/src/MillInitMavenTests.scala
```diff
@@ -0,0 +1,221 @@
...
+          "testsuite-osgi",
+          "testsuite-shading",
+          "testsuite-native",
+          "testsuite-native-image",
+          "testsuite-native-image-client",
+          "testsuite-native-image-client-runtime-init",
+          "transport-blockhound-tests",
+          "microbench",
+          "bom"
+        ).forall(resolveRes.out.contains) ==> true
```
lihaoyi, line 192: This kind of long-multi-line-expression followed by a single `.forall` is kind of hard to read; let's instead assign the `Seq` to a local `val` and then call `.forall` on that `val` after

integration/feature/init/src/MillInitMavenTests.scala
```diff
@@ -0,0 +1,221 @@
...
+
+        // additional sources defined by build-helper-maven-plugin
+        val sourcesRes = eval(("show", "transport-native-epoll.sources"))
+        Seq(
+          s"$workspacePath/transport-native-epoll/src/main/java",
+          s"$workspacePath/transport-native-epoll/src/main/c"
+        ).forall(sourcesRes.out.contains) ==> true
+
+        // compile succeeds for submodule
+        val compileRes = eval("codec.compile")
```
lihaoyi, line 202: Let's explicitly list out all the submodules that we expect `.compile` to work on, and loop over them and perform the `compile`, and for those submodules for which `.compile` does not work we can leave a comment about why it currently doesn't work (e.g. `codec.test` due to missing native dependency)

main/maven/src/mill/main/maven/Plugins.scala
```diff
@@ -0,0 +1,126 @@
...
+    execution.getConfiguration match {
+      case xpp3: Xpp3Dom => Some(xpp3)
+      case _ => None
+    }
+
+  /**
+   * @see [[https://maven.apache.org/plugins/maven-compiler-plugin/index.html]]
+   */
+  object `maven-compiler-plugin` {
+
```
lihaoyi, line 32: Let's give these `camelCase` names; I think the correspondence with the `kebab-case` maven plugin name is clear enough, especially with a comment linking to the relevant website

main/maven/test/resources/expected/--base-module/build.mill
```diff
@@ -0,0 +1,31 @@
+package build
```
lihaoyi, line 1: Let's compare the `scalafmt`-ed output with the `scalafmt`-ed expected build files, with an assertion to check that the expected build files are `scalafmt`-ed. That will make editing them much less fragile than now and make them much easier to read and understand

lihaoyi:
@ajaychandran took another pass at review. It's looking really good! A few more back and forth iterations on the code and I think we'll be ready merge it

integration/feature/init/src/MillInitMavenTests.scala
ajaychandran: This is from the [executions](https://github.com/dhatim/fastexcel/blob/de56e786a1fe29351e2f8dc1d81b7cdd9196de4a/pom.xml#L268) section.
```xml
<executions>
    <!-- First: compile all as Java9 -->
    <execution>
        <id>default-compile</id>
        <configuration>
            <release>9</release>
            <excludes>
                <!-- package-info.java is not recompiled as Java8 in the second step for an unknown reason -->
                <!-- this workaround just forces the second compilation by avoiding the first one -->
                <exclude>**/package-info.java</exclude>
            </excludes>
        </configuration>
    </execution>
    <!-- Second: recompile all but module-info as Java8 -->
    <execution>
        <id>base-compile</id>
        <goals>
            <goal>compile</goal>
        </goals>
        <configuration>
            <source>8</source>
            <target>8</target>
            <release>8</release>
            <excludes>
                <exclude>module-info.java</exclude>
            </excludes>
        </configuration>
    </execution>
</executions>
```

integration/feature/init/src/MillInitMavenTests.scala
ajaychandran: > Do you know how the src/main/c/ gets handled by Maven?

This is used to [publish](https://github.com/netty/netty/blob/64c52255c073d4a308d3dbb56a12d3cca7692c47/transport-native-epoll/pom.xml#L597) the `C` sources.

> For example, if we put a Foo.java file in there, will it get picked up by the Java compiler?

The compiler plugin is not used in this module since there are no Java sources.

I see 2 options for closing this:
- Limit `build-helper-maven-plugin` support to resources.
- Remove `build-helper-maven-plugin` support. Note that there are many ways to [configure additional sources](https://stackoverflow.com/questions/270445/maven-compile-with-multiple-src-directories).

main/maven/src/mill/main/maven/Plugins.scala
ajaychandran: Renamed extensions. Please review.

integration/feature/init/src/MillInitMavenTests.scala
ajaychandran: Added test for [avaje-config](https://github.com/avaje/avaje-config) that also has JPMS errors.

integration/feature/init/src/MillInitMavenTests.scala
ajaychandran: Thank you for the detailed explanation.

I agree that the end user should provide this.
We can mention this in the documentation (TBD).

lihaoyi:
Looks like some of the new tests are failing on windows; might be a path separator `/` vs `\` thing

ajaychandran:
> Looks like some of the new tests are failing on windows; might be a path separator `/` vs `\` thing

The issue occurs while prepping the workspace:
```scala
      for (p <- os.list(sourceDir)) os.move.into(p, tester.workspacePath)
```
```bash
[7855]   java.nio.file.DirectoryNotEmptyException: C:\Users\RUNNER~1\AppData\Local\Temp\1113725499934748432\netty-netty-4.1.114.Final\.devcontainer
[7855]     sun.nio.fs.WindowsFileCopy.ensureEmptyDir(WindowsFileCopy.java:265)
[7855]     sun.nio.fs.WindowsFileCopy.move(WindowsFileCopy.java:429)
[7855]     sun.nio.fs.WindowsFileSystemProvider.move(WindowsFileSystemProvider.java:293)
[7855]     java.nio.file.Files.move(Files.java:1432)
[7855]     os.move$.apply(FileOps.scala:98)
[7855]     os.move$into$.apply(FileOps.scala:113)
[7855]     mill.integration.MillInitMavenTests$.$anonfun$prep$2(MillInitMavenTests.scala:14)
```

ajaychandran:
@lihaoyi Please review.

Notably,
- added option to merge generated build files
- added "Future work" section in PR description

CI failures on Windows are due to EOL encodings in golden files. Any tips on how to fix this?

lihaoyi:
@ajaychandran for the golden files, we've typically just done `.replaceAll("\r\n", "\n")` to fix up line endings before comparison. Typically needs to be done on both the expected and actual output, but otherwise works great

lihaoyi:
I think this looks good overall. I tried it out on https://github.com/iluwatar/java-design-patterns, and apart from parts of the codebase that needed compiler plugins (e.g. `@lombok.extern.slf4j.Slf4j`) it seems that it mostly worked and I got most of the codebase compiling and testing. Nice work!

integration/feature/init/src/MillInitMavenTests.scala
ajaychandran: Went with option 2

ajaychandran:
Investigating the CI failure with [act](https://github.com/nektos/act).
Changed the PR to Draft.

lihaoyi:
@ajaychandran i ended up disabling the netty tests in windows for now. I'll close out this bounty and the previous one for the OS-Lib instrumentation

PR #3757, opened 2024-10-16
Add test reporter for Kotlin/JS

0xnm:
This PR adds test reporter for Kotlin/JS. It is based on [XUnit](https://mochajs.org/#xunit) reporter provided by Mocha - XML will be generated anyway, irrespective of the `TestModule.testReportXml` option, because structured output is needed in order to be able to parse it.

So as a result it will be XML file + test result message, which can be seen in the tests.

Adding HTML reporter is not yet possible without using 3rd party packages, because the one provided OOTB by Mocha is [not intended for CLI usage](https://mochajs.org/#html-reporter).

I also had to modify the `Jvm.runSubprocess` method, so it doesn't throw if exit code is not 0, but returns result. To comply with the old behavior, all the old call sites will just call `getOrThrow` on the result.

This change is needed, because if there is a test failure, `node` process will exit with code 1, but we still need to process and catching generic `Exception` is not an option (because maybe it is not because of the exit code).

Upd: Binary compatibility check now complains that `Jvm.runSubprocess` has a different return type, but I'm not sure if it should be a problem, because before it was simply returning `Unit`. But if it is a problem, I can introduce a new method instead.


lihaoyi:
> Upd: Binary compatibility check now complains that Jvm.runSubprocess has a different return type, but I'm not sure if it should be a problem, because before it was simply returning Unit. But if it is a problem, I can introduce a new method instead.

Yes let's give the new method a new name to preserve binary compatibility. The old one returning `: Unit` can just call the new one and ignore the return value

lihaoyi:
Seems like some of the `mill.kotlinlib.js.KotlinJSNodeRunTests` tests are failing

0xnm:
Oh, yeah, indeed, thank for spotting it. I was thinking it was a flaky test from `scalalib`, but turns out not only it.

This is the case where type safety cannot be checked during the compile time, because `Result.Failing` is itself `Exception`.

Fixed now.

kotlinlib/src/mill/kotlinlib/js/KotlinJsModule.scala
```diff
@@ -105,54 +107,68 @@ trait KotlinJsModule extends KotlinModule { outer =>
...
+    mill.api.Result.Failure("runMain is not supported in Kotlin/JS.")
+  }
+
+  override def runMain(@arg(positional = true) mainClass: String, args: String*): Command[Unit] =
+    Task.Command[Unit] {
+      mill.api.Result.Failure("runMain is not supported in Kotlin/JS.")
     }
 
-    val moduleKind = this.moduleKind()
+  protected[js] def runJsBinary(
```
lefou, line 133: I'd prefer the name `runKotlinJsBinary`, since chances are high the user project may also have some Scala.JS or other JS modules.

kotlinlib/src/mill/kotlinlib/js/KotlinJsModule.scala
```diff
@@ -457,7 +475,9 @@ trait KotlinJsModule extends KotlinModule { outer =>
 
     override def testFramework = ""
 
-    override def kotlinJSBinaryKind: T[Option[BinaryKind]] = Some(BinaryKind.Executable)
+    override def kotlinJSRunTarget: T[Option[RunTarget]] = outer.kotlinJSRunTarget()
```
lefou, line 478:
```suggestion
    override def kotlinJsRunTarget: T[Option[RunTarget]] = outer.kotlinJsRunTarget()
```

kotlinlib/test/src/mill/kotlinlib/js/KotlinJsKotlinTestPackageModuleTests.scala
```diff
@@ -20,6 +22,7 @@ object KotlinJsKotlinTestPackageModuleTests extends TestSuite {
 
     object foo extends KotlinJsModule {
       def kotlinVersion = KotlinJsKotlinTestPackageModuleTests.kotlinVersion
+      override def kotlinJSRunTarget = Some(RunTarget.Node)
```
lefou, line 25:
```suggestion
      override def kotlinJsRunTarget = Some(RunTarget.Node)
```

PR #3758, opened 2024-10-17
Update Discord link

lihaoyi:


lefou:
Can we instead have a central community page and maintain the Discord link only there?

Issue #299, opened 2018-04-16
Add ability to alias Modules (specially external modules)


caeus:
In the spirit of [issue 259](https://github.com/lihaoyi/mill/issues/259), wouldn't it be nice to give Mill the capability to alias commands and/or (external?) modules?
So if someone working creating an External Module with a long package name can therefore call an easy to remember alias than the whole package name plus slash plus target.

uosis:
As a workaround, something like this can be added to the `./mill` wrapper script:

```shell
if [ "$1" = "bloop" ] ; then
  exec $MILL_EXEC_PATH mill.contrib.Bloop/install
elif [ "$1" = "idea" ] ; then
  exec $MILL_EXEC_PATH mill.scalalib.GenIdea/idea
else
  exec $MILL_EXEC_PATH "$@"
fi
```
A bit better than global shell alias.

carlosedp:
Nice that this is "built-in" and documented but just a heads-up that my plugin allows one to define an alias that calls a sequence of other modules or commands.

Issue #3715, opened 2024-10-10
Storing module reference in module causes `resolve` to stack overflow (500USD Bounty)


KireinaHoro:
---

**From the maintainer Li Haoyi**: I'm putting a 500USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

The task is to appropriately detect module cycles during task resolution time, report a nice user-friendly error, and direct the user towards use of `ModuleRef` to fix the issue. This should be covered by unit tests, and we should verify that the module cycles in the code sample below gets properly recognized and handled.

---

I'm trying to work out a pattern that would allow the consumer of a foreign module swap out dependencies of modules.  Given the following module definition:

```scala
import mill._, scalalib._

trait CommonModule extends ScalaModule {
  def scalaVersion = "2.13.12"
  def m1 = mym1
  def m2 = mym2
}

object mym1 extends M1
trait M1 extends CommonModule

object mym2 extends M2
trait M2 extends CommonModule {
  override def moduleDeps = super.moduleDeps ++ Agg(m1)
}
```

One can then define another `build.sc`, import the above, instantiate new `M1` and `M2` (for example with new Scala versions), and then have the newly instantiated `M2` depend on `M1`:

```scala
import mill._, scalalib._
import $file.dep.build

trait MyModules {
  def m1 = newm1
  def m2 = newm2
}
object newm1 extends dep.build.M1 with MyModules
object newm2 extends dep.build.M2 with MyModules
```

However this breaks `resolve` with a stack overflow error:

```console
$ mill resolve __
No mill version specified.
You should provide a version via '.mill-version' file or --mill-version option.
Using mill version 0.11.12
[1/1] resolve 
Exception in thread "MillServerActionRunner" java.lang.StackOverflowError
        at java.base/java.security.AccessController.doPrivileged(Native Method)
        at java.base/java.lang.Class.getClasses(Class.java:1752)
        at mill.define.Reflect$.reflectNestedObjects0(Reflect.scala:77)
        at mill.resolve.ResolveCore$.resolveDirectChildren0(ResolveCore.scala:349)
        at mill.resolve.ResolveCore$.$anonfun$resolveDirectChildren$3(ResolveCore.scala:300)
        at scala.util.Either.flatMap(Either.scala:360)
        at mill.resolve.ResolveCore$.resolveDirectChildren(ResolveCore.scala:295)
        at mill.resolve.ResolveCore$.resolveTransitiveChildren(ResolveCore.scala:223)
[...]

PR #3760, opened 2024-10-17
Document and test workflow for aliasing external modules

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/299

I'm a bit surprised that this just works, but it does! 

The module's `millSourcePath` and `out/` folder path remain at their original locations as the `mill.define.Ctx` is still provided by the top-level `ExternalModule` constructor, it's just during resolution the added method `def` allows it to get picked up during object graph traversal


lefou:
* I think the solution space for https://github.com/com-lihaoyi/mill/issues/3715 is pretty identical to the effect you exploit here. If we find a way to detect, whether the rhs has already an `Ctx` and is assiciated with a instantiated module, we just need to provide an additional name for it, but not re-instantiate a new module.

lihaoyi:
I think #3715 is kind of adjacent, since it's more about preventing the stackoverflow. Whether the referenced module ends up discoverable or not isn't the main issue raised in that ticket, although it's certainly worth discussing.

One potential issue with `def` aliases is that one module can now have different selectors turn up when you `resolve _`. For external modules that's not a big deal since the external module namespace and internal module namespace are kept separate, but once you start aliasing internal modules then the duplication occurs. Even in scenarios without infinite loops, when someone does a `mill resolve __` we probably don't want to "expand out" every module alias and print a huge wide tree when it may actually just be a relatively narrow DAG

PR #3762, opened 2024-10-17
Fix java compile warning, again

lihaoyi:
Follow up to https://github.com/com-lihaoyi/mill/pull/3711

I swear this annotation didn't work last time, and the `match` on the `discoveredModuleType` did. But not the `match` doesn't work, and the annotation seems to work again. I have no idea what's going on

### DIFF
--- a/runner/src/mill/runner/CodeGen.scala
+++ b/runner/src/mill/runner/CodeGen.scala
@@ -235,7 +235,7 @@ object CodeGen {
   def discoverSnippet(segments: Seq[String]): String = {
     if (segments.nonEmpty) ""
     else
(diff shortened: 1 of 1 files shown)


.github/workflows/run-tests.yml
```diff
@@ -79,7 +79,7 @@ jobs:
 
             # Most of these integration tests should not depend on which mode they
             # are run in, so just run them in `local`
-          - java-version: '11'
+          - java-version: '17'
```
lihaoyi, line 82: The new test emits some benign warnings on Java 11, so just run on java 17 instead

Issue #1939, opened 2022-07-11
Windows - lingering java process locks file from deletion, breaks compilation


james-s-w-clark:
I'm checking compilation locally to reduce CI job runs (that's a [different story](https://github.com/com-lihaoyi/mill/pull/1933) though).

When I re-run a compilation (`mill contrib.flyway.compile`), I get an error - which I can solve by killing the only "java.exe" process. [This stackoverflow answer](https://stackoverflow.com/a/68556916/4261132) helped me get the compilation running again. Windows wouldn't let me delete the Jar:
![image](https://user-images.githubusercontent.com/12559216/178371872-bc20e70b-f3b8-443a-b350-149092046dc9.png)

```
mill contrib.flyway.compile 
...
[info] done compiling
[99/365] de.tobiasroeser.mill.vcs.version.VcsVersion.vcsState.overridden.de.tobiasroeser.mill.vcs.version.VcsVersion.vcsState
[177/365] main.moduledefs.manifest
Exception in thread "MillServerActionRunner" java.nio.file.FileSystemException: C:\Users\user\mill\out\main\moduledefs\jar.dest\out.jar: The process cannot access the file because it is being used by another process
        at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:92)
        at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
        at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
        at java.base/sun.nio.fs.WindowsFileSystemProvider.implDelete(WindowsFileSystemProvider.java:275)
        at java.base/sun.nio.fs.AbstractFileSystemProvider.deleteIfExists(AbstractFileSystemProvider.java:110)
        at java.base/java.nio.file.Files.deleteIfExists(Files.java:1191)
        at os.remove$.apply(FileOps.scala:290)
        at os.remove$.apply(FileOps.scala:284)
        at os.remove$all$.$anonfun$apply$5(FileOps.scala:301)
        at os.remove$all$.$anonfun$apply$5$adapted(FileOps.scala:301)
        at geny.Generator.$anonfun$foreach$1(Generator.scala:50)
        at geny.Generator$Mapped.$anonfun$generate$4(Generator.scala:283)
        at os.walk$stream$$anon$2$$anon$3.visitFile(ListOps.scala:249)
[...]

PR #3764, opened 2024-10-17
Improve file handling on Windows

sake92:
Changes:
- replace `new FileInputStream` with `Files.newInputStream`
- replace `new FileOutputStream` with `Files.newOutputStream`

Motivation: 
https://github.com/com-lihaoyi/mill/issues/1939

https://bugs.openjdk.org/browse/JDK-6357433 (in a comment) says:
> For jdk7, the workaround is to **use new file system API**. So instead of new FileInputStream(f) and new FileOutputStream(f), use f.toPath().newInputStream() and f.toPath().newOutputStream()

So seems like new NIO API does use, FILE_SHARE_DELETE, which is needed to delete files in unix-style.


lihaoyi:
Thanks @sake92! Might need to send a PR upstream to OS-Lib as well https://github.com/com-lihaoyi/os-lib/blob/f1bf225f8b110c4a561465717d8b13efcd9ee6d3/os/src/ZipOps.scala#L75

lihaoyi:
We also use `java.util.jar.JarOutputStream` in some places in Mill, I'm guessing that API predates NIO and may need to be updated as well

sake92:
> Thanks @sake92! Might need to send a PR upstream to OS-Lib as well https://github.com/com-lihaoyi/os-lib/blob/f1bf225f8b110c4a561465717d8b13efcd9ee6d3/os/src/ZipOps.scala#L75

Sure thing, I'll update it there too.

---

> We also use `java.util.jar.JarOutputStream` in some places in Mill, I'm guessing that API predates NIO and may need to be updated as well

JarOutputStream is not OS-dependent like the `java.io` input/output streams. I might be wrong tho.  
https://docs.oracle.com/javase/8/docs/api///?java/util/jar/JarOutputStream.html says:
> The JarOutputStream class is used to **write the contents of a JAR file to any output stream**. It extends the class java.util.zip.ZipOutputStream with support for writing an optional Manifest entry. The Manifest can be used to specify meta-information about the JAR file and its entries.

PR #3767, opened 2024-10-18
Comparison Documentation Review

lihaoyi:


liutikas:
Do you have the comparison projects available to verify the numbers you have shared? Specifically curious about Gradle vs Mill comparison

Issue #3761, opened 2024-10-17
`runIvyDeps` should be transitive for consistency with Maven resolution


lefou:
### Discussed in https://github.com/com-lihaoyi/mill/discussions/3722

<div type='discussions-op-text'>

<sup>Originally posted by **megri** October 12, 2024</sup>
Let's say we have the modules `A <- B <- C` where `<-` denotes a "depends on" relationship

```scala
// running with mill version 0.11.12
import mill._, mill.scalalib._

trait BaseModule extends ScalaModule {
  def scalaVersion = "3.5.1"
}

object C extends BaseModule {
  def ivyDeps = Agg(
    ivy"org.slf4j:slf4j-api:2.0.16"
  )
}

object B extends BaseModule {
  def moduleDeps = Seq(C)

  def runIvyDeps = Agg(
    ivy"ch.qos.logback:logback-classic:1.5.10"
  )
}

object A extends BaseModule {
  def moduleDeps = Seq(B)
}
```

So `C` introduces slf4j and `B` a logging implementation. `mill A.ivyDepsTree --withRuntime` shows

```
[build.sc] [54/59] compile 
[info] compiling 1 Scala source to /Users/martin/dev/mill-deps-test/out/mill-build/compile.dest/classes ...
[info] done compiling
[26/26] A.ivyDepsTree 
├─ org.scala-lang:scala3-library_3:3.5.1
│  └─ org.scala-lang:scala-library:2.13.14
└─ org.slf4j:slf4j-api:2.0.16
```

While the example may seem a little contrived—why not define the runtime dependency in the A module?—this has consequences with for instance assemblies. At work we have a project of several Spark applications (as separate modules) that are packaged into one single assembly for easy shipping. These modules have different but non-conflicting [runtime] dependencies, and the assembly won't collect the runtime parts without overriding for instance `transitiveIvyDeps` (which solves one level of transitivity—it has to be overridden in all modules to include "deep" runIvyDeps.)

Aside from backwards compatibility, would making runIvyDeps transitive be an issue? </div>

megri:
I think I can implement this, but I'd like us to map out which parts need to be touched and expected outcomes.
Today we basically have no transitivity and the runtime dep doesn't end up in packaged artefacts. This change would make it so that

- modules inherit runtime dependencies through something akin to `transitiveIvyDeps`
- runtime deps are included in packaged jars and Pom-files.

What else?

PR #3774, opened 2024-10-20
Make runtime deps transitive

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/3761

* Introduce `runModuleDeps`, the runtime version of `moduleDeps` and `compileModuleDeps`

* We introduce `transitiveRunIvyDeps`, the runtime version of `transitiveIvyDeps`, and use it in most places as a replacement for `runIvyDeps`

* I also flipped the ordering of `transitiveModuleDeps`/`transitiveRunModuleDeps` to put `this` on the right/last, to follow the rest of the module resolution logic where "upstream" things come on the left/first in the classpath ordering

There are some behavioral changes, e.g. the A module's direct `compileModuleDeps` no longer end up on your `runClasspath`, as shown by the change in the test case `mill.scalalib.ScalaMultiModuleClasspathsTests.modCompile`. I think the new behavior is more correct than the old one?

One (benign?) consequence of this change is that the contents of the various `*run*` tasks now have considerable overlap with the non-`run` tasks, e.g. `transitiveRunIvyDeps` overlaps with `transitiveIvyDeps`. 

The way transitive runtime dependencies are now managed, both "`run{Module,Ivy}Dep` of `{module,ivy}Dep`" and "`{module,ivy}Dep` of `run{Module,Ivy}Dep`" are both aggregated into the run classpath. I'm not sure if this matches what Maven does (it does according to chatgpt!) but it seems reasonable to me

Added some additional unit tests to cover the new transitive behavior. The whole dependency-wiring stuff is pretty gnarly but hopefully the existing test suite will stop us from breaking too much (especially `mill.scalalib.ScalaMultiModuleClasspathsTests` which is pretty rigorous and we add to).

This is a binary-compatible but semantically incompatible change, would be good to get it into 0.12.0

Best reviewed with `Hide Whitespace` enabled

### DIFF
--- a/scalalib/src/mill/scalalib/JavaModule.scala
+++ b/scalalib/src/mill/scalalib/JavaModule.scala
@@ -190,6 +190,16 @@ trait JavaModule
     moduleDeps
   }
 
+  /**
+   * Same as [[moduleDeps]] but checked to not contain cycles.
+   * Prefer this over using [[moduleDeps]] directly.
+   */
+  final def runModuleDepsChecked: Seq[JavaModule] = {
+    // trigger initialization to check for cycles
+    recRunModuleDeps
+    runModuleDeps
+  }
+
   /** Should only be called from [[moduleDepsChecked]] */
   private lazy val recModuleDeps: Seq[JavaModule] =
     ModuleUtils.recursive[JavaModule](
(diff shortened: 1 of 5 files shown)


scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -198,9 +208,26 @@ trait JavaModule
       _.moduleDeps
     )
 
-  /** The compile-only direct dependencies of this module. */
+  /** Should only be called from [[moduleDepsChecked]] */
+  private lazy val recRunModuleDeps: Seq[JavaModule] =
+    ModuleUtils.recursive[JavaModule](
+      (millModuleSegments ++ Seq(Segment.Label("moduleDeps"))).render,
```
myyk, line 214: should this be: `moduleDeps` -> `runModuleDeps`

Issue #3789, opened 2024-10-20
Flags created with --define appear to be incorrectly cached


hilcode:
Given this `build.sc`:

```scala
import mill.T

def myProperty: T[Boolean] = T.input {
    println(s"""System.getProperty("my-property", "X")=${System.getProperty("my-property", "X")}""")
    System.getProperty("my-property", "X") match {
        case result if result == "F" || result == "X" =>
            println(s"$result ::: my-property=false")
            false

        case "T" =>
            println("T ::: my-property=true")
            true

        case x =>
            println(s"my-property='$x'")
            throw new IllegalStateException("Invalid value")
    }
}

def foo: T[String] = T {
    if (myProperty()) {
        println("TRUE")
        "TRUE"
    } else {
        println("FALSE")
        "FALSE"
    }
}
```

I see the following:

```scala
~/c/s/mill-bug> ./mill --version
Mill Build Tool version 0.11.12
Java version: 21.0.1, vendor: Eclipse Adoptium, runtime: /nix/store/6m2wff2lm5pk892389ivgcbs1nqbsz6v-temurin-bin-21.0.1
Default locale: en_US, platform encoding: UTF-8
OS name: "Linux", version: 6.11.3-1-default, arch: amd64

~/c/s/mill-bug> rm -rf out

~/c/s/mill-bug> ./mill --define=my-property=T foo
[build.sc] [54/59] compile
[info] compiling 1 Scala source to .../out/mill-build/compile.dest/classes ...
[info] done compiling
[1/2] myProperty
System.getProperty("my-property", "X")=T
T ::: my-property=true
[2/2] foo
TRUE

~/c/s/mill-bug> ./mill foo
[1/2] myProperty
System.getProperty("my-property", "X")=T
T ::: my-property=true

~/c/s/mill-bug> ./mill --define=my-property=F foo
[1/2] myProperty
System.getProperty("my-property", "X")=F
F ::: my-property=false
[2/2] foo
FALSE

~/c/s/mill-bug> ./mill foo
[1/2] myProperty
System.getProperty("my-property", "X")=F
F ::: my-property=false
```
Note that when _not_ providing `--define=my-property=?`, it seems that the previous invocation of `myProperty` is reused somehow, including the `println` output.

lefou:
@lihaoyi just fixed this issue in `main` branch, which will become `0.12.0`. We might want to backport the fix to `0.11.x` too.

lefou:
[...]

PR #3791, opened 2024-10-21
Fix property handling and add tests and docs

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/3789. We just accidentally flipped the conditional and didn't have any tests that verify this behavior, so I added some example tests that also serve to explain the feature in the docs


lefou:
@lihaoyi Were you able to reproduce issue #3789 with these test cases, before you fixed it? How?

lihaoyi:
@lefou yeah the test cases failed without the included fix

PR #3795, opened 2024-10-21
Fix property handling and add tests and docs (backport #3791)

lefou:
Fixes https://github.com/com-lihaoyi/mill/issues/3789. We just
accidentally flipped the conditional and didn't have any tests that
verify this behavior, so I added some example tests that also serve to
explain the feature in the docs.

Backport of pull request: https://github.com/com-lihaoyi/mill/pull/3791

### DIFF
--- a/example/tasks/4-inputs/build.sc
+++ b/example/tasks/4-inputs/build.sc
@@ -29,12 +29,12 @@ def gitStatusTask = T {
 /** Usage
 
 > git init .
-> git commit --allow-empty -m "Initial-Commit"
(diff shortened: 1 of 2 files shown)


lefou:
CI tests fail most likely because the env var declaration at the beginning of the promt isn't supported.

lihaoyi:
can probably just skip the env var test and just leave the properties test which is what aftually covers the fix (env var test was just extra)

lefou:
> can probably just skip the env var test and just leave the properties test which is what aftually covers the fix (env var test was just extra)

Good point. This makes is a bit easier. :-)

PR #3808, opened 2024-10-22
Update coursier-interface to 1.0.22

lefou:
I don't know why we have coursier and coursier-interface at the same time in this project. Maybe we can get rid of one?


alexarchambault:
IIRC, that was added to bump the version that scalafmt pulls, and make sure we don't have stale coursier stuff around. Bumping coursier-interface in scalafmt needs a bit of work, https://github.com/scalameta/scalafmt/pull/4281.

PR #3809, opened 2024-10-22
Update error_prone_core to 2.34.0

lefou:


lefou:
Newer versions require Java 17.

PR #3811, opened 2024-10-22
Update javet to 4.0.0

lefou:
This is a major version change. I haven't tested manually, if it works. There is also PR #3810, which only bump to latest minor update.


lihaoyi:
Tested manually

Issue #3798, opened 2024-10-21
`TestModule.testForkGrouping` is not working for TestNG


lefou:
* Reproducer: https://github.com/com-lihaoyi/mill/pull/3799

lefou:
This seem to original how the `testng` contrib plugin is implementing the `TestFramework` API.

PR #3813, opened 2024-10-22
Fix TestNg test runner

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/3798, supersedes https://github.com/com-lihaoyi/mill/pull/3799

Not sure how this ever worked, but it seems like it works now

Converted the example test in https://github.com/com-lihaoyi/mill/pull/3799 to a unit test.

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -771,6 +771,7 @@ object dist0 extends MillPublishJavaModule {
     build.contrib.jmh.testDep(),
     build.contrib.playlib.testDep(),
     build.contrib.playlib.worker("2.8").testDep(),
+    build.contrib.testng.testDep(),
     build.bsp.worker.testDep(),
     build.testkit.testDep()
   )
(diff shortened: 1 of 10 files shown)


lihaoyi:
CC @lefou

lefou (dismissed):
Looks good to me. 

Tested with a reasonable large TestNG test suite spawning 98 test processes.

lefou:
@lihaoyi Looks like without custom `testForkGrouping`, the test suite is now running all tests over and over, probably as often as the count of discovered test classes.

lihaoyi:
Ah, we must have been running a single task because that task runs all tests within it for some reason

lefou (commented):
Looks good. Local tests seem to work for various fork groups. Thank you for this quick fix!

There is now no longer a single test summary at the end but one for each test class, which is a bit less convenient. Maybe Mill should print a summary, which would look more uniform over the various test frameworks anyways.

PR #3836, opened 2024-10-24
Move google tag into a partial

lihaoyi:


lefou:
@lihaoyi Could you document what this tag does? Is there an option to not build it in? Would be cool to leave it out for `localPages` or non-official build.

PR #3841, opened 2024-10-25
Allow `.mill-jvm-opts` to interpolate environment variables, add `.mill-opts`

lihaoyi:
This PR extends `.mill-jvm-opts` to support interpolating JVM variables. This provides a bit of flexibility in allowing `$PWD` and `$USER` to be used when passing `-Dproperty=` to Mill. This could in theory be fleshed out even further to run arbitrary shell snippets, but for now I this should provide enough flexibility to meet the immediate needs.

Also added a `.mill-opts`, which is similar to `.mill-jvm-opts` but used for Mill arguments rather than JVM arguments. This provides a way to persistent required Mill configuration in a format that is similar to what we already have today (CLI args). This would be a common place to put things like `--jobs=0.5C` or `--allow-positional` or `--no-build-lock`.

Updated the integration tests and documentation

related discussions

* https://github.com/com-lihaoyi/mill/discussions/1226 
* https://github.com/com-lihaoyi/mill/discussions/3840

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -643,8 +643,7 @@ def launcherScript(
     shellCommands = {
       val jvmArgsStr = shellJvmArgs.mkString(" ")
       def java(mainClass: String, passMillJvmOpts: Boolean) = {
-        val millJvmOpts = if (passMillJvmOpts) "$mill_jvm_opts" else ""
-        s"""exec "$$JAVACMD" $jvmArgsStr $$JAVA_OPTS $millJvmOpts -cp "${shellClassPath.mkString(
+        s"""exec "$$JAVACMD" $jvmArgsStr $$JAVA_OPTS -cp "${shellClassPath.mkString(
             ":"
           )}" $mainClass "$$@""""
       }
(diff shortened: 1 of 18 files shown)


build.mill
```diff
@@ -655,49 +654,19 @@ def launcherScript(
...
-         |          mill_jvm_opts="$${mill_jvm_opts} $$(echo $$line | grep -v "^[[:space:]]*[#]")"
-         |        done
-         |        # we are in a sub-shell, so need to return it explicitly
-         |        echo "$${mill_jvm_opts}"
-         |      )
-         |    )"
-         |    mill_jvm_opts="$${mill_jvm_opts} -Dmill.jvm_opts_applied=true"
-         |  fi
-         |}
-         |
```
lihaoyi, line 684: @lefou do you remember what this stuff is? AFAIK, we currently do all `MILL_JVM_OPTS` processing in the mill client Java code, so not sure why we have to do some bash logic here

lihaoyi: Also the relevant integration tests seem to pass with this stuff deleted

lihaoyi: I think it might be remnant from back when `mill -i` ran everything within a single process, so we needed to make sure it was initialized correctly before that process started. Should be unnecessary now

lefou: Exactly. This was prior to processing this in Java. This is a left over and should be no longer needed since we no longer run any task evaluation in the client JVM.

lefou (commented):
I think we should also (and only) support the delimited version `${VAR}`.

Also, since we now only have one place where this is implemented, we should also support files in the `.config` dir as we do for `.mill-version` vs. `.config/mill-version`.

lefou:
> Also added a `.mill-opts`, which is similar to `.mill-jvm-opts` but used for Mill arguments rather than JVM arguments. This provides a way to persistent required Mill configuration in a format that is similar to what we already have today (CLI args). This would be a common place to put things like `--jobs=0.5C` or `--allow-positional` or `--no-build-lock`.

How does this behave when we find an opion like `--jobs` in that file and specified on the CLI? We should not fail then but let the CLI override the default value from the file.

PR #3844, opened 2024-10-25
Make `.mill-opts`/`.mill-jvm-opts` env var interpolation require braces, add tests for overriding opts

lihaoyi:


lihaoyi:
CC @lefou

lefou (commented):
The `PWD` is the client working dir, right?

We seem to interpolate missing env vars to the empty String. Shouldn't we at least print some warning to STDERR?

lefou:
> Not sure about printing an error; having optional environment variables seems like it would be a common use case, in which case it's not an error for the env var to be missing

I find some kind of defensive programming can avoid lots of hard to spot issues. We could just log it when we run in `--debug` mode. But I think we don't fully parse the cli in the client. (If we do, we should also print which additional files we read in `--debug` mode. Can be another PR though.)

lefou:
If we don't log it, we should document what happens with undefined env vars.

lefou:
We don't interpolate `.mill-vars` currently. But I think we should. For consistency, but also because it's probably the natural way to pass env vars to system properties.

lihaoyi:
@lefou what's `.mill-vars`?

lefou:
> @lefou what's `.mill-vars`?

A typo. I meant `.mill-opts`.

PR #3856, opened 2024-10-28
Add stub `mill.define.Ctx` to trick IntelliJ into not erroring

lihaoyi:
Same as https://github.com/com-lihaoyi/mill/pull/3558, but for `mill.define.Ctx` rather than `mill.main.RootModule.Info`. Because it's provided by the enclosing code-generation, we need to trick IntelliJ into thinking there is something it can use, and so I define a low priority implicit to fall back upon if the main `def make` doesn't work (but `@compileTimeOnly` to show an error message if it ends up being needed)

I moved the bulk of the cacher logic from `mill-moduledefs` to `mill.define`. We just need a marker trait in `mill-moduledefs`, and having the rest of the logic in Mill itself simplifies maintenance and evolution of the code

Before
Screenshot 2024-10-28 at 5 55 39 PM

After
Screenshot 2024-10-28 at 5 54 22 PM

### DIFF
--- a/main/define/test/src/mill/define/MacroErrorTests.scala
+++ b/main/define/test/src/mill/define/MacroErrorTests.scala
@@ -94,8 +94,9 @@ object MacroErrorTests extends TestSuite {
             }
           }
         """)
+        pprint.log(e)
         assert(e.msg.contains(
-          "Modules and Tasks can only be defined within a mill Module"
+          "Task{} members must be defs defined in a Cacher class/trait/object body"
         ))
       }
       test("neg") {
@@ -139,7 +140,7 @@ object MacroErrorTests extends TestSuite {
           }
         """)
         assert(borkedCachedDiamond1.msg.contains(
-          "Modules and Tasks can only be defined within a mill Module"
+          "Task{} members must be defs defined in a Cacher class/trait/object body"
         ))
       }
     }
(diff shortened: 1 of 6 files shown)


main/define/test/src/mill/define/MacroErrorTests.scala
```diff
@@ -94,8 +94,9 @@ object MacroErrorTests extends TestSuite {
             }
           }
         """)
+        pprint.log(e)
         assert(e.msg.contains(
-          "Modules and Tasks can only be defined within a mill Module"
+          "Task{} members must be defs defined in a Cacher class/trait/object body"
```
lefou, line 109: Since `Cacher` never surfaces in our docs (and API), we probably should list the most likely candidate `Module`.

lefou:
```suggestion
          "Task{} members must be defs defined in a Cacher class/trait/object body like Module"
```

lihaoyi: This actually comes from the mill-moduledefs repo, so we'll need to fix it upstream there https://github.com/com-lihaoyi/mill-moduledefs/blob/33a27d470a8f27a1cb2ad9e71d8c25b24294eee8/moduledefs/src-2/Cacher.scala#L29

lihaoyi: https://github.com/com-lihaoyi/mill-moduledefs/pull/20

PR #3865, opened 2024-10-29
backtick when path segment is scala keyword

### LINKED ISSUE #3863: keyword as submodule name in 0.12
minimal repro: https://github.com/chikei/mill-keyword-package

if I declare package using ``package build.`import` ``
mill complains:
```
1 tasks failed
generateScriptSources Package declaration "package build.`import`" in import/package.mill does not match folder structure. Expected: "package build.import"
Package declaration "package build.`import`" in import/package.mill does not match folder structure. Expected: "package build.import"
```
if declare using `package build.import `
mill complains differently:
```
1 tasks failed
generateScriptSources import/package.mill:1:14 expected end-of-input
package build.import
             ^
import/package.mill:1:14 expected end-of-input
package build.import
```

note that submodule like ``package build.`name-hyphen` `` generally fine

chikei:
to allow submodule use scala keyword as name, fix #3863 

I am not sure this is the way we want to match scala keyword though

### DIFF
--- a/integration/feature/keyword-module/resources/import/package.mill
+++ b/integration/feature/keyword-module/resources/import/package.mill
@@ -1,0 +1,7 @@
+package build.`import`
+import mill._, scalalib._
+
+object `package` extends RootModule with ScalaModule {
+  def scalaVersion = "2.13.11"
+}
+
--- a/integration/feature/keyword-module/resources/import/package.mill
+++ b/integration/feature/keyword-module/resources/import/package.mill
@@ -1,7 +1,7 @@
 package build.`import`
 import mill._, scalalib._
 
-object `package` extends RootModule with ScalaModule {
-  def scalaVersion = "2.13.11"
+object `package` extends RootModule {
+  def task = Task { true }
 }
 
--- a/runner/src/mill/runner/FileImportGraph.scala
+++ b/runner/src/mill/runner/FileImportGraph.scala
@@ -5,7 +5,6 @@ import mill.main.client.CodeGenConstants._
 import mill.main.client.OutFiles._
 
 import scala.reflect.NameTransformer.encode
-import scala.reflect.runtime
 import scala.collection.mutable
 
 /**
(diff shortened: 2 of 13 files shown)


integration/feature/keyword-module/resources/import/package.mill
```diff
@@ -0,0 +1,7 @@
+package build.`import`
+import mill._, scalalib._
+
+object `package` extends RootModule with ScalaModule {
```
lihaoyi, line 4: Let's not make this a `ScalaModule`, since we don't care about the scala side at all. Instead, we can just add a trivial `def  task = Task { 123 }`

Furthermore, let's exercise a few more Scala keywords, with `this/package.mill`, `for/package.mill`, `if/package.mill`, just to make sure it works for more than just the single `import` keyword

runner/src/mill/runner/FileImportGraph.scala
```diff
@@ -33,7 +34,8 @@ case class FileImportGraph(
 object FileImportGraph {
   def backtickWrap(s: String): String = s match {
     case s"`$v`" => s
-    case _ => if (encode(s) == s) s else "`" + s + "`"
+    case _ => if (encode(s) == s) runtime.universe.asInstanceOf[runtime.JavaUniverse].quotedName(s)
```
lihaoyi, line 37: Rather than relying on `runtime.JavaUniverse`, let's instead port over the user-land implementation from Ammonite. That will make it a smoother upgrade to Scala 3 where `runtime.JavaUniverse` might not exist

https://github.com/com-lihaoyi/Ammonite/blob/b98bc05c0d90f43c46bafd69736ed22dd36f8734/amm/util/src/main/scala/ammonite/util/Model.scala#L127

chikei: So I just grab the alphabet keywords from ammonite since NameTransformer should handle the other case and that exists in scala 3 stdlib

integration/feature/keyword-module/resources/import/package.mill
chikei: I added few more case. One interesting point is that `` `package` `` will fail but I assume it's not due to backtick as mill complain this way:
```
[2758] [build.mill-57] [error] /home/Chikei/repos/mill/out/integration/feature/keyword-module/local/test.dest/sandbox/run-1/out/mill-build/generateScriptSources.dest/build_/build.mill:21:21: lazy value millDiscover overrides nothing
[2758] [build.mill-57] [error]   override lazy val millDiscover: _root_.mill.define.Discover = _root_.mill.define.Discover[this.type]
[2758] [build.mill-57] [error]                     ^
[2758] [build.mill-57] [error] one error found
```
So I just removed it from cases for now.

Issue #3862, opened 2024-10-29
Add an example Python Support case study (500USD bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 500USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

This would be similar to the example Typescript support case study we have in https://mill-build.org/mill/0.12.1/extending/new-language.html, but instead of typescript/npm/tsc/rollup we would use python/pip/mypy/pex (or equivalent). We should take care to make sure things are installed in the dest folders using virtualenvs or whatever, rather than globally as python traditionally defaults to

Doesn't need to be production ready, just complete enough for educational purposes illustrating the various Mill features

himanshumahajan138:
@lihaoyi i have a great knowledge of python and now for mill too (after your guidance and merged pr's) i would like to work around this i know bounties are open until merged pr but i just want to get a safe apart from clash so i commented 

Hope you understand !

c0d33ngr:
@lihaoyi  wouldn't it be better structure like this

```
newlang/
  typescript/
    ...
  python3/
    ...
```

c0d33ngr:
@himanshumahajan138 looked through your PR and you have already gone far. Lucky for you, I'll just let you move ahead with the issue. I don't want to do any catching up, after all there are other bounties that need to be resolved.

Goodluck

himanshumahajan138:
@c0d33ngr Buddy, there is one Kotlin PR of Mine actually its almost done 95% and only support of aar file needs resource compilation and merging 

If you want to taste bounty you can try that just Chery pick my commits and then start working ahead of it...

Best of luck...

c0d33ngr:
> @c0d33ngr Buddy, there is one Kotlin PR of Mine actually its almost done 95% and only support of aar file needs resource compilation and merging
> 
> If you want to taste bounty you can try that just Chery pick my commits and then start working ahead of it...
> 
> Best of luck...

I don't understand the last line

himanshumahajan138:
Actually cherry picking is just like cloning a part of the repo like the updates till a commit...
More you can find on google 🙃

PR #3873, opened 2024-10-30
[WIP] Add Python Examples

c0d33ngr:
A draft PR to close issue #3862

### DIFF
--- a/example/extending/newlang/5-hello-python/build.mill
+++ b/example/extending/newlang/5-hello-python/build.mill
@@ -1,3 +1,13 @@
+// This example demonstrates basic integration of https://www.python.org/[Python]
+// using Mill to run Python code. Mill
+// does not come bundled with a Python integration, so here we begin checking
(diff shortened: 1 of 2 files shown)


example/extending/newlang/5-hello-python/build.mill
```diff
@@ -0,0 +1,21 @@
+package build
+import mill._
+
+def installPython = Task{
```
myyk, line 4: I wonder if it would be better to just test that python is installed and if not, let the user deal with it since this solution here is architecture dependent. It wouldn't work for any macos or win users.

c0d33ngr: Alright.. noted

I was even thinking about finding way to install for the three main OSes but ran out of luck. Yours is better

himanshumahajan138:
@c0d33ngr well done bro i dont want to say but good usage of Ai.

But there is alot to add i think...

PR #3878, opened 2024-10-31
Detect when `resolve` needs `ModuleRef` when likely cyclic references

myyk:
This completes #3715 with all the cycle cases I could think of.

It also touches `resolveDirectChildren` where I made some changes that I would help with readability. I thought I was going to need to build on those but didn't. These can be reverted if it's unwanted.

### DIFF
--- a/main/resolve/src/mill/resolve/ResolveCore.scala
+++ b/main/resolve/src/mill/resolve/ResolveCore.scala
@@ -80,33 +80,23 @@ private object ResolveCore {
       case head :: tail =>
         def recurse(searchModules: Set[Resolved]): Result = {
           val searchModuleClasses = moduleClasses(searchModules)
-          val checkedSearchModules = if (seenModules.intersect(searchModuleClasses).nonEmpty) {
-            Left(s"Cyclic module reference detected: [${seenModules.intersect(searchModuleClasses).mkString(", ")}], it's required to wrap it in ModuleRef.")
-          } else { Right(searchModules) }
-
-          val errOrResults = for {
-            checkedSearchModules0 <- checkedSearchModules
-            searchModules <- Right(checkedSearchModules0)
-          } yield {
-            searchModules.map(r => resolve(rootModule, tail, r, querySoFar ++ Seq(head), seenModules ++ moduleClasses(Set(current))))
+          if (seenModules.intersect(searchModuleClasses).nonEmpty) {
+            Error(s"Cyclic module reference detected: [${seenModules.intersect(searchModuleClasses).mkString(", ")}], it's required to wrap it in ModuleRef.")
+          } else {
+            val results = searchModules.map(r => resolve(rootModule, tail, r, querySoFar ++ Seq(head), seenModules ++ moduleClasses(Set(current))))
               .partitionMap { case s: Success => Right(s.value); case f: Failed => Left(f) }
-          }
+            val (failures, successesLists) = results
+            val (errors, notFounds) = failures.partitionMap {
+              case s: NotFound => Right(s)
+              case s: Error => Left(s.msg)
+            }
 
-          errOrResults match {
-            case Left(msg) => Error(msg)
-            case Right(results) =>
-              val (failures, successesLists) = results
-              val (errors, notFounds) = failures.partitionMap {
-                case s: NotFound => Right(s)
-                case s: Error => Left(s.msg)
-              }
-
-              if (errors.nonEmpty) Error(errors.mkString("\n"))
-              else if (successesLists.flatten.nonEmpty) Success(successesLists.flatten)
-              else notFounds.size match {
-                case 1 => notFounds.head
-                case _ => notFoundResult(rootModule, querySoFar, current, head)
-              }
+            if (errors.nonEmpty) Error(errors.mkString("\n"))
+            else if (successesLists.flatten.nonEmpty) Success(successesLists.flatten)
+            else notFounds.size match {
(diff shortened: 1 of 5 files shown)


main/resolve/src/mill/resolve/ResolveCore.scala
```diff
@@ -240,29 +242,39 @@ private object ResolveCore {
...
+
+    val errOrModules = directTraverse.map { modules =>
+      modules.flatMap {
+        case m: Resolved.Module => Some(m)
+        case _ => None
+      }
+    }
+
+    if (seenModules.contains(cls)) {
+      Left(s"Cyclic module reference detected: ${cls.getName}, it's required to wrap it in ModuleRef. See documentation: https://mill-build.org/mill/0.12.1/fundamentals/modules.html#_abstract_modules_references")
```
myyk, line 263: It would be better if this link used the `MILL_VERSION` or I could not give the link. WDYT?

lefou: Ideally, the full link should be injected and generated in `build.sc`, since page refactoring happens now and then, and we could feed it to our link-checker when generating the docs.

main/resolve/test/src/mill/main/ResolveTests.scala
```diff
@@ -1110,5 +1110,54 @@ object ResolveTests extends TestSuite {
...
+        Right(Set(_.foo))
+      )
+    }
+    test("moduleRefCycle") {
+      val check = new Checker(TestGraphs.ModuleRefCycle)
+      test - check(
+        "__",
+        Right(Set(_.foo))
+      )
+    }
```
lihaoyi, line 1161: Let's also add a few test cases for non-transitive resolves, both by normal wildcard `foo.bar._` or `foo.bar._.qux`, as well as no-wildcard-at-all `foo.bar.qux`. In these cases, we should raise an error if we detect a cycle, even if there isn't a risk of stackoverflow

main/resolve/src/mill/resolve/ResolveCore.scala
myyk: I like @lefou's idea but I'll consider that out of scope. It would be nice if some warnings linked to docs, but I guess that would be part of a separate effort.

main/resolve/test/src/mill/main/ResolveTests.scala
myyk: I have some work on it here, it's very close: https://github.com/myyk/mill/tree/resolve-need-moduleref-2

I'm stuck on cases when it's `foo.__.bar`-shaped where `foo` or `bar` can be anything, even `_`.

myyk:
I added some cycle detection for the direct dependencies but I'm stuck on cases when it's `foo.__.bar`-shaped where `foo` or `bar` can be anything, even `_`.

You can run the tests with `./mill main.resolve.test`:

```
[960-2] ----------------------------------- Failures -----------------------------------
[960-2] X mill.resolve.ResolveTests.nested.wildcard3 3ms 
[960-2]   utest.AssertionError: resolvedTasks.map(_.map(_.toString).toSet[String]) ==
[960-2]   resolvedTasks: Either[String,List[mill.define.NamedTask[Any]]] = Left(Cyclic module reference detected: [class mill.util.TestGraphs$nestedModule$classInstance$], it's required to wrap it in ModuleRef.
[960-2]   Cyclic module reference detected: [class mill.util.TestGraphs$nestedModule$nested$], it's required to wrap it in ModuleRef.)
[960-2]   expected: scala.util.Either[String,scala.collection.immutable.Set[mill.define.NamedTask[Any]]] = Right(Set(classInstance.single, nested.single))
[960-2]     utest.asserts.Asserts$.assertImpl(Asserts.scala:30)
[960-2]     mill.resolve.ResolveTests$Checker.checkSeq(ResolveTests.scala:25)
[960-2]     mill.resolve.ResolveTests$Checker.apply(ResolveTests.scala:15)
[960-2]     mill.resolve.ResolveTests$.$anonfun$tests$66(ResolveTests.scala:230)
[960-2] X mill.resolve.ResolveTests.cross.double.wildcard.labelNeg4 3ms 
[960-2]   utest.AssertionError: resolvedTasks.map(_.map(_.toString).toSet[String]) ==
[960-2]   resolvedTasks: Either[String,List[mill.define.NamedTask[Any]]] = Left(Cyclic module reference detected: [class mill.util.TestGraphs$doubleCross$cross$], it's required to wrap it in ModuleRef.)
[960-2]   expected: scala.util.Either[String,scala.collection.immutable.Set[mill.define.NamedTask[Any]]] = Left(Cannot resolve cross.__.doesntExist. Try `mill resolve cross._` to see what's available.)
[960-2]     utest.asserts.Asserts$.assertImpl(Asserts.scala:30)
[960-2]     mill.resolve.ResolveTests$Checker.checkSeq(ResolveTests.scala:25)
[960-2]     mill.resolve.ResolveTests$Checker.apply(ResolveTests.scala:15)
[960-2]     mill.resolve.ResolveTests$.$anonfun$tests$113(ResolveTests.scala:365)
[960-2] Tests: 161, Passed: 159, Failed: 2
```

It's showing these 2 false positives because those are not cycles. These are the only noncyclic `foo.__.bar` test cases in there and I think that's the commonality.

myyk:
The cleanup looks good to me

lihaoyi:
Thanks @myyk! I'll close out the bounty over whatsapp

PR #3882, opened 2024-10-31
Fixes: #3862; Added an example Python Support case study

himanshumahajan138:
# Pull Request 
Added an example Python Support case study
Fixes: #3862 

## Description
Added `Python` Support for Mill From basic installation to multiple supports like virtual environment, mypy integration , pex integration

## Related Issues
- Link to related issue #3862.

## Checklist
- [x] Python-hello example
- [x] Python-modules example
- [x] Python-dependencies example
- [x] Python-mypy support example
- [x] Python-pex Support example
- [x] Documentation Updated


example/extending/newlang/5-hello-python/build.mill
```diff
@@ -0,0 +1,193 @@
...
+
+// Each component has been documented with inline comments for clarity.
+
+package build
+import mill._
+
+/**
+ * A trait for managing Python environments and dependencies in a Mill build.
+ */
+trait PythonModule extends Module {
```
lihaoyi, line 28: This first example should not use a `trait`, and every defined `Task` should be explained. See the Typescript example for reference

example/extending/newlang/5-hello-python/build.mill
```diff
@@ -0,0 +1,193 @@
...
+
+  /** Specifies dependencies to install in the virtual environment. */
+  def dependencies: T[Seq[String]] = Task { Seq.empty[String] }
+
+  /** Specifies command-line options to pass to the main Python file. */
+  def options: T[Seq[String]] = Task { Seq.empty[String] }
+
+  /** Specifies the name of the virtual environment directory. */
+  def venvName: T[String] = Task { "venv" }
+
```
lihaoyi, line 38: First example can get-rid-of/hardcode all of these for simplicity

example/extending/newlang/5-hello-python/build.mill
```diff
@@ -0,0 +1,193 @@
...
+      case None =>
+        throw new java.io.FileNotFoundException(
+          s"Main file '${mainFileName()}' not found in sources."
+        )
+    }
+    PathRef(mainFilePath)
+  }
+
+  /** Executes the main Python file using the virtual environment. */
+  def compile: T[Unit] = Task {
```
lihaoyi, line 153: Why is this called `compile` and why is it a `Task`?

lihaoyi:
Please clean this up to follow the Typescript example documentation. I cannot fix every detail for you, you need to go and real through those examples yourself and fix your code to match

himanshumahajan138:
Ok @lihaoyi  sir, Thanks for reviewing 

My silly mistake made one test fail its just f string error 😔

No worries i will update the code and documentation

example/extending/newlang/5-hello-python/build.mill
himanshumahajan138: You mean no need for virtual env and all just installation of pip and then run simple python file 👍

lihaoyi: Yes, they can be added in the later examples when we introduce resolving dependencies

example/extending/newlang/5-hello-python/build.mill
himanshumahajan138: silly me i have to give some output like PathRef or something else

actually i have edited this command to return PathRef for pex example but forgot this 

is there any other consideration?

himanshumahajan138: i think sir you are referring to have one run function which will run the file and compile will only compile all the sources and resources 

am i right ?

lihaoyi: I don't know, you have to decide, but what you have written now between the comment and the method is 100% wrong

himanshumahajan138: Ok sir, first i will re-search for it in already present examples and then will update it as per other and follow mill structure 👍

himanshumahajan138:
@lihaoyi Sir Final Testing has been done...

- Documentation Updated (Written by me 😅)
- Code Logic Updated as per TypeScript Example
- combined moduleDeps, virtualEnv & libraryDeps (coz they are dependent to each other)
- All Tests Passed and Code is checked thrice...

Please pardon me if anything missed or done in wrong way...

himanshumahajan138:
@lihaoyi Sir Final Testing has been done...

- Documentation Updated (Written by me 😅)
- Code Logic Updated as per TypeScript Example
- combined moduleDeps, virtualEnv & libraryDeps (coz they are dependent to each other)
- All Tests Passed and Code is checked thrice...

Please pardon me if anything missed or done in wrong way...

example/extending/newlang/5-hello-python/build.mill
```diff
@@ -0,0 +1,138 @@
...
+
+package build
+import mill._
+
+def installPip: T[PathRef] = Task {
+  val pipFile = Task.dest / "get-pip.py"
+  val pipPath = os.home / ".local" / "bin" / "pip3"
+
+  if (!os.exists(pipPath)) {
+    os.call(Seq("wget", "-q", "https://bootstrap.pypa.io/get-pip.py", "-O", s"$pipFile"))
```
lihaoyi, line 22: We should not need to install pip because Python3 bundles pip for all versions >3.4 and above. Let's just use the bundled one

example/extending/newlang/5-hello-python/build.mill
```diff
@@ -0,0 +1,138 @@
...
+
+// The `compile` Task  ensures `pip` is installed, retrieves the main file directory, and returns it for further processing.
+// This task is considered as a part of code which is responsible for all compilation of sources and pre-requirements.
+
+def compile: T[PathRef] = Task {
+  installPip()
+  val mainFileDir = sources().path
+  
+  PathRef(mainFileDir)
+}
```
lihaoyi, line 75: We should perform MyPy typecheck validation here, as mentioned in the original ticket

example/extending/newlang/5-hello-python/build.mill
```diff
@@ -0,0 +1,138 @@
...
+
+// Next, We define the `sources` and `allSources` tasks to help gather Python files for processing. 
+// `sources` sets a base directory for Python source files,
+// while `allSources` searches this directory (and its subdirectories) to collect all `.py` files.
+// Together, they enable easy access to Python files in a project,
+// making it straightforward to retrieve all relevant code files for tasks like building, testing, or analysis.
+
+def sources: T[PathRef] = Task.Source(millSourcePath / "src")
+
+def allSources: T[Seq[PathRef]] = Task {
```
lihaoyi, line 61: `allSources` is not used and should be removed

example/extending/newlang/5-hello-python/build.mill
lihaoyi: Maybe rename it `typecheck` instead of `compile`

example/extending/newlang/7-python-module-deps/build.mill
```diff
@@ -0,0 +1,263 @@
...
+  def sources: T[PathRef] = Task.Source(millSourcePath / "src")
+
+  /** Collects all Python source files from the sources directory. */
+  def allSources: T[Seq[PathRef]] = Task {
+    os.walk(sources().path).filter(_.ext == "py").map(PathRef(_))
+  }
+
+  /** Path to the requirements file, configurable by the user. */
+  def requirementsPath: T[PathRef] = Task {
+    PathRef(millSourcePath / "src" / "requirements.txt")
```
lihaoyi, line 57: This should not be in `src`, just `millSourcePath / "requirements.txt"` is fine

example/extending/newlang/7-python-module-deps/build.mill
```diff
@@ -0,0 +1,263 @@
...
+      println("pip is already installed, skipping installation.")
+    }
+  }
+
+  /** Creates a virtual environment using the installed Python version. */
+  def createVenv: T[PathRef] = Task {
+    installPip()
+    val venvDir = T.dest / venvName()
+
+    os.call(Seq(
```
lihaoyi, line 102: `venv` is bundled with Python3 so we should not need to install it

example/extending/newlang/7-python-module-deps/build.mill
```diff
@@ -0,0 +1,263 @@
...
+      "--break-system-packages",
+      "--no-warn-script-location"
+    ))
+    os.call(Seq("python3", "-m", "virtualenv", venvDir.toString))
+
+    PathRef(venvDir)
+  }
+
+  /** Activates the created virtual environment. */
+  def activateVenv: T[PathRef] = Task {
```
lihaoyi, line 117: We should not need to activate the venv, we should directly use the `bin/python` binary from within it

example/extending/newlang/8-python-mypy/build.mill
```diff
@@ -0,0 +1,225 @@
+// == Python MyPy Support
```
lihaoyi, line 1: Let's follow the typescript setup: MyPy support should be included in the first example. The last example should be for third-party dependencies and bundling

himanshumahajan138:
@lihaoyi Sir Let me tell you one thing that i have seen actually i  installed Ubuntu in WSL and when i run it for the first time it just only provides python which is installed with only setuptools and nothing else (no pip , no venv and others )

So when i searched on internet i came to know that we have to install pip and venv using `sudo apt install python3-pip` and `sudo apt install python3-venv`
 
but now here comes the problem of permissions we cant get sudo for every task and even this is not the good practise 

so to tackle this problem i have  used official guide for installing pip for users without requiring sudo permission and this will be installed for the user rather than the sudo 

i think this is a good approach, rest if you say i will follow your way...

example/extending/newlang/7-python-module-deps/build.mill
himanshumahajan138: i think you mean we should have only single `requirements.txt` in the project like if user have multiple sources then also
they have to mention all the requirements in the same file and that will be main requirements for the project

lefou:
A Mill build process should not install anything outside of the project directory, unless it's some kind of cache like `coursier`. We should just depend on installed tools and provide meaningful error messages if they are missing. We can provide installation/setup support in some external module, e.g. `build.PythonModule/installPip`, but that should be purely optional.

himanshumahajan138:
> A Mill build process should not install anything outside of the project directory, unless it's some kind of cache like `coursier`. We should just depend on installed tools and provide meaningful error messages if they are missing. We can provide installation/setup support in some external module, e.g. `build.PythonModule/installPip`, but that should be purely optional.

But Sir what about github actions tests ?

himanshumahajan138:
Sir Please see this image i just installed the ubuntu-24.04 version and when i tried these commands you can see that it don't comes with pip and venv 

**Moreover, i am not installing anything outside the T.dest actually i just downloaded `get-pip.py` in T.dest and then using this we installed pip for user originally that's the official way** 

image

lihaoyi:
You should be able to use `python3 -m pip` and `python3 -m venv` without needing installation

himanshumahajan138:
ok then let me recall the updates again:

- no need to install pip and virtualenv
- mypy will be added in first example (5-hello-python) change in compile function
- no need for activating venv
- no need for all sources
- and other minor updates...

Thanks For Review Sir (This community is filled with helpful members❤️)

lihaoyi:
@himanshumahajan138 yep! I expect this will take multiple rounds of back and forth review to polish into shape, but it's moving in the right direction so I think we'll eventually get there

himanshumahajan138:
> You should be able to use `python3 -m pip` and `python3 -m venv` without needing installation

Hat's off to your Experience Sir!  You said it should run, and it did 


One More question than i will Move to documentation part :

> Let's follow the typescript setup: MyPy support should be included in the first example. The last example should be for third-party dependencies and bundling

As per this you mean we should add mypy in first example i.e 5-hello-python but for this we need to install mypy dependency and this needs virtual env but we have introduced virtual env in example 7-python-module-deps 

so what should i do now ???

lihaoyi:
@himanshumahajan138 we need pip internally to install mypy in the first examples, but we do not need to expose it to the user to install their own things until the later examples. This is similar to the Typescript examples which use NPM from the start but only let users install stuff using NPM later

lihaoyi:
> oh so you mean we will download mypy internally without venv into main python on the system and then we will use it and we will type check that in all the other examples as a function named as type checking rather than compile

no, you should use venv

himanshumahajan138:
@lihaoyi Sir, Code and Documentation is Updated and Ready for Review...

Please Pardon me if I made any mistake or missed something...

Thanks...

example/extending/newlang/5-hello-python/build.mill
```diff
@@ -0,0 +1,158 @@
+// == Basic Python Pipeline 
+
+// This Example demonstrates integration of https://www.python.org[Python]
+// compilation into a Mill build to compile https://www.python.org[Python] Scripts. Mill
```
lihaoyi, line 4: As a rule of thumb, you should not link the same keyword more than once per page

example/extending/newlang/5-hello-python/build.mill
```diff
@@ -0,0 +1,158 @@
...
+// `MyPy` is a static type checker that uses Python `type hints` to enforce `type correctness` and `catch errors` during development.
+
+package build
+import mill._
+
+def createVenv: T[PathRef] = Task {
+  val venvDir = T.dest / "venv"
+
+  os.call(
+    Seq(
```
lihaoyi, line 23: Don't need the `Seq` Here, don't need the `.toString`, and the `call` should be a one-liner

example/extending/newlang/5-hello-python/build.mill
```diff
@@ -0,0 +1,158 @@
...
+      "-m",
+      "venv",
+      venvDir.toString
+    )
+  )
+
+  PathRef(venvDir)
+}
+
+def setup: T[PathRef] = Task {
```
lihaoyi, line 34: Let's combine this with `createVenv` above

example/extending/newlang/5-hello-python/build.mill
```diff
@@ -0,0 +1,158 @@
...
+//
+// ```graphviz
+// digraph G {
+//   rankdir=LR
+//   node [shape=box width=0 height=0 style=filled fillcolor=white]
+//   createVenv -> setup -> typeCheck
+//   sources -> typeCheck
+//   mainFileName -> typeCheck
+// }
+// ```
```
lihaoyi, line 98: We can probably skip this diagram since it's not yet interesting; the one below is enough

example/extending/newlang/5-hello-python/build.mill
```diff
@@ -0,0 +1,158 @@
...
+// and any additional arguments. The output is shown in the console, allowing for real-time interaction.
+  
+def run(args: mill.define.Args) = Task.Command {
+  val pythonVenv = typeCheck().path
+  
+  os.call(
+    Seq(
+      pythonVenv.toString,
+      s"${sources().path / mainFileName()}"
+    ) ++ args.value,
```
lihaoyi, line 120: No need `Seq`, `toSdtring`, or the `s"${}"` wrapper

lihaoyi:
@himanshumahajan138 took a pass, most of the comments apply to all the examples.

Getting pretty close I think, looks a lot better than when we started!

himanshumahajan138:
@lihaoyi sir, like i have used type checking in evey example is that a good and right way

lihaoyi:
Yes having typechecking in every example is right. The examples are meant to build upon each other, so we would expect them to get longer and more complex as it progresses down the list

example/extending/newlang/5-hello-python/build.mill
himanshumahajan138: i thought a lot for choosing name 😅

finally i choosed `Python Integration in Mill`

himanshumahajan138:
@lihaoyi Sir, finally I have Updated the Code as per reviews and now I think we can move further in review or merging part

Hope you Like it...

Thanks for Reviews Sir...

example/extending/newlang/5-hello-python/build.mill
```diff
@@ -0,0 +1,118 @@
...
+def sources: T[PathRef] = Task.Source(millSourcePath / "src")
+
+// === Type Checking
+
+// The `mainFileName` Task provides the name for the main python script file for running and 
+// The `typeCheck` Task  ensures that the python setup is set and Checks the main file for error free code.
+// This task is considered as a part of code which is responsible for Error Checking of the Main Script 
+// and setting up setup for python script runable environment.
+
+def mainFileName: T[String] = Task { "main.py" }
```
lihaoyi, line 51: We should move `mainFileName` definition and docs into the same section as `run`, since it is only needed later and is not necessary for `typeCheck`

example/extending/newlang/5-hello-python/build.mill
```diff
@@ -0,0 +1,118 @@
...
+  )
+
+  PathRef(pythonVenv)
+}
+
+// When `typeCheck` task sets up the Setup for Virtual Environment and checks the main file for error free code 
+// then we have `main.py` and running this file will result in either a generic greeting 
+// or a personalized one based on the command-line arguments passed, allowing for flexible interaction with the user.
+
+/** See Also: src/main.py */
```
lihaoyi, line 67: This snippet is far too verbose. We need something 5 lines max

example/extending/newlang/5-hello-python/build.mill
```diff
@@ -0,0 +1,118 @@
...
+  os.call(
+    (pythonVenv,sources().path / mainFileName(),args.value),
+    stdout = os.Inherit
+  )
+}
+
+// Note that we use `stdout = os.Inherit` since we want to display any output to the user,
+// rather than capturing it for use in our command.
+
+// ```graphviz
```
lihaoyi, line 88: This diagram is great, but we should have another diagram with just `typeCheck` and its upstream targets above under the `Type Checking` section

example/extending/newlang/6-python-modules/build.mill
```diff
@@ -0,0 +1,138 @@
+// == Re-usable PythonModule
+// This Example demonstrates the use of a `PythonModule` trait to manage Python Scripts within
+// different objects in a Mill build.
+
+// ==== Using Same PythonModule for Mutiple Objects
```
lihaoyi, line 5: `Mutiple` should be `Multiple`

example/extending/newlang/6-python-modules/build.mill
```diff
@@ -0,0 +1,138 @@
...
+
+// The `PythonModule` trait simplifies Python project management within the `Mill` build tool by providing essential
+// functionalities such as creating `virtual environment`, managing `source files`, `Checking code for Error`, and `executing Python` scripts.
+// It ensures that users have a streamlined workflow by automating the setup of necessary tools 
+// and organizing project files effectively.
```
lihaoyi, line 10: This whole paragraph is too repetitive, since we already discuss this above. We should discuss how `trait PythonModule` takes the same tasks we defined above and lets them be re-used for separate modules

example/extending/newlang/6-python-modules/build.mill
```diff
@@ -0,0 +1,138 @@
...
+
+// The `PythonModule` trait simplifies Python project management within the `Mill` build tool by providing essential
+// functionalities such as creating `virtual environment`, managing `source files`, `Checking code for Error`, and `executing Python` scripts.
+// It ensures that users have a streamlined workflow by automating the setup of necessary tools 
+// and organizing project files effectively. 
+
+package build
+import mill._
+
+/** PythonModule Trait for Basic Python Pipeline in Mill Build Tool */
```
lihaoyi, line 15: ~All these comments are useless and should be removed

example/extending/newlang/6-python-modules/build.mill
```diff
@@ -0,0 +1,138 @@
...
+    )
+  }
+  
+}
+
+// ==== Example
+
+// The script defines three key `objects` that extend the `PythonModule` trait, each object encapsulating a distinct runtime configuration.
+// By allowing objects like `foo`, `bar`, and `qux` to extend the `PythonModule`, it promotes code reuse and modular design, 
+// enabling easy access to common functionalities while facilitating customization for specific needs.
```
lihaoyi, line 65: This paragraph is bad and meaningless, as are the comments inside the below code block. Just say that we can now instantiate `PythonModule` three times to work with code in `foo/src/`, `foo/bar/src/`, and `qux/src/`

example/extending/newlang/6-python-modules/build.mill
```diff
@@ -0,0 +1,138 @@
...
+object qux extends PythonModule{
+  override def mainFileName: T[String] = Task { "qux.py" }
+  // Functions independently but has access to PythonModule methods
+}
+
+// We have used three different Python Scripts `foo/src/foo.py`, `foo/bar/src/bar.py`, `qux/src/qux.py` 
+
+/** See Also: foo/src/foo.py */
+/** See Also: foo/bar/src/bar.py */
+/** See Also: qux/src/qux.py */
```
lihaoyi, line 84: Again, these examples are far too long, and anyway we don't need to include them here since we already saw the file in the previous section

lihaoyi:
Took a review pass. Many of the comments apply to the following docs and examples as well

example/extending/newlang/5-hello-python/build.mill
himanshumahajan138: is this good which we have before...
image

example/extending/newlang/5-hello-python/build.mill
himanshumahajan138: i tried my best and converted to 10 lines 😅

example/extending/newlang/6-python-modules/build.mill
himanshumahajan138: i shortened them

example/extending/newlang/6-python-modules/build.mill
himanshumahajan138: removed

example/extending/newlang/6-python-modules/build.mill
himanshumahajan138: only new things have comments now

example/extending/newlang/6-python-modules/build.mill
himanshumahajan138: removed

example/extending/newlang/6-python-modules/build.mill
himanshumahajan138: done

example/extending/newlang/5-hello-python/build.mill
himanshumahajan138: done

example/extending/newlang/5-hello-python/build.mill
himanshumahajan138: done

example/extending/newlang/5-hello-python/build.mill
```diff
@@ -0,0 +1,118 @@
...
+
+> ./mill typeCheck
+Success: no issues found in 1 source file
+
+> ./mill run Mill Python
+Hello, Mill Python!
+
+*/
+
+// Uncomment the Result variable line(Line Number:13) in the `src/main.py` code to see `typeCheck` error
```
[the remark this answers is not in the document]
himanshumahajan138: actually sir i want to show error also but the problem is that if i write code for error then it will fail in ci test as this will me a major error from mypy side so how can i do this 

please Guide me a little Bit Sir...

himanshumahajan138:
@lihaoyi Sir, Ready for Next Reviews 

what i changed in last updates:
- updated documentation and some code 
- updated scripts and shorten them 
- rewrite the documentation
- removed irrelevant and repeative codes and docs

Thanks For Reviews Sir...

lihaoyi:
@himanshumahajan138 I think this looks good. Just need to run `./mill mill.scalalib.scalafmt.ScalafmtModule/`  to reformat everything and we can merge it

himanshumahajan138:
Actually sir before ci don't run for .mill build files but now it do for these files also and its was my fault that i didn't checked it...
No worries let me update and then finalize it...

Thanks sir for your guidance and reviews...

himanshumahajan138:
AND HERE WE GO🎉

Everything Ready for Merge Lets make this happen...

Thanks A Lot Sir ❤️
(To be Honest You have Changed my Life I was Nothing before contributing to Mill and now I am Confident with my Skills and alot financially also)

Issue #3874, opened 2024-10-30
`methodCodeHashSignatures` throw a `os.ResourceNotFoundException`


lefou:
Just before the build failure, I deleted the `tests` object from a Scala source file under `mill-build/src` (see below). I did not change the Mill version.

```
> mill --version
Mill Build Tool version 0.12.1-7-d8bbe5
Java version: 21.0.4, vendor: Eclipse Adoptium, runtime: /opt/openjdk-bin-21.0.4_p7
Default locale: de_DE, platform encoding: UTF-8
OS name: "Linux", version: 6.1.90-gentoo-x86_64, arch: amd64
```

This is probably related to
* https://github.com/com-lihaoyi/mill/pull/3849

Here's the stack trace:

```
1 tasks failed
methodCodeHashSignatures os.ResourceNotFoundException: java.net.URLClassLoader@1450ee23/comfisbuild/BuildTemplate$tests$.class
    os.ResourcePath.getInputStream(ResourcePath.scala:21)
    os.read$inputStream$.apply(ReadWriteOps.scala:245)
    mill.codesig.ExternalSummary$.load0$1(ExternalSummary.scala:55)
    mill.codesig.ExternalSummary$.$anonfun$apply$6(ExternalSummary.scala:48)
    scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
    scala.collection.mutable.HashMap.getOrElse(HashMap.scala:451)
    mill.codesig.ExternalSummary$.load$1(ExternalSummary.scala:48)
    mill.codesig.ExternalSummary$.$anonfun$apply$9(ExternalSummary.scala:69)
    mill.codesig.ExternalSummary$.$anonfun$apply$9$adapted(ExternalSummary.scala:69)
    scala.collection.immutable.BitmapIndexedSetNode.foreach(HashSet.scala:951)
    scala.collection.immutable.HashSet.foreach(HashSet.scala:958)
    mill.codesig.ExternalSummary$.apply(ExternalSummary.scala:69)
    mill.codesig.CodeSig$.compute(CodeSig.scala:17)
    mill.runner.MillBuildRootModule.$anonfun$methodCodeHashSignatures$2(MillBuildRootModule.scala:173)
    mill.define.Task$TraverseCtx.evaluate(Task.scala:215)
```

This project has a meta-build which defines some `trait`s to be used in the `build.sc`. 

Here is a sketch of the relevant parts of the project:

```scala
[...]

PR #3884, opened 2024-11-01
Drop incremental compilation state for Mill modules when version changes

lefou:
Mill used various tricks to map build scripts written by users to generated enhanced versions, including package objects and other stuff.
Since some of these techniques may result in Zinc undercompilation, we want avoid downstream issues,
especally in the `methodCodeHashSignatures` task, which is sensitive to inconsistent bytecode.

We trait potentially longer compilation against incorrect or too coarse code change detection, which may invalidate a lot more downstream tasks than needed due to invalid data.

Fix https://github.com/com-lihaoyi/mill/issues/3874


lefou:
Since we don't have a reproducer, this PR comes without a test.

lihaoyi (approved):

Issue #3879, opened 2024-10-31
Evaluator commands duff `out/mill-chrome-profile.json` 


lefou:
Evaluator commands make the `out/mill-chrome-profile.json` un-useful since they seem to overwrite it.

Here is an example, where `idea` is a customized derivate of `mill.idea.GenIdea/idea`.

While the command is still running, I can see a partially filled `out/mill-chrome-profile.json` with lots of details:

![image](https://github.com/user-attachments/assets/467cd7e7-d4c4-424b-b7bb-a83ecb5e75f5)

But after the command finishes, the `out/mill-chrome-profile.json` looks less useful:

![image](https://github.com/user-attachments/assets/f358db49-41e7-4a19-8b64-c99044bdc2fe)

lihaoyi:
Maybe every time we open `mill-chrome-profile.json`, we should check if one exists, and if one does then open `mill-chrome-profile-2.json`  `mill-chrome-profile-3.json` etc.. We can wipe out all `mill-chrome-profile*.json`s at the start of each run. This would ensure that all profiles are saved, even if multiple evaluation runs are started re-entrantly

The same should probably be done for the other `mill-*` debug files

lefou:
I think this is because evaluator command use there dedicated evaluator instances which don't use the outer profile logger. Instead, we could pass the profile logger around.

lihaoyi:
oh yeah that's also an option. Then the same profile would show tasks from all evaluations, which would be even better

PR #3885, opened 2024-11-01
Share profile loggers between nested evaluations

lihaoyi:
We broaden the scope of `profileLogger`/`chromeProfileLogger` from within `def evaluate` to instead last as long as the `Evaluator` itself, and share it with any sub-`Evaluator`s created e.g. to redirect output during `show`

`Evaluator` is now `AutoCloseable`, and needs to be closed after being used in `MillBuildBootstrap` and in `UnitTester`

For now I decided not to share the profilers between different evaluation levels of the meta-build, since they already don't overlap: one goes to `out/`, the parent to `out/mill-build/`, etc. We could do that if we think it would be useful to give people a consolidated profile, but it might be a bit more work (e.g. we would need to properly prefix entries in the profile to distinguish the different meta levels).

Added a unit test to ensure the `show compile` mill-profile and mill-chrome-profile contain the expected keys for both `show` and for `compile` and some of its upstream tasks

Fixes https://github.com/com-lihaoyi/mill/issues/3879

### DIFF
--- a/integration/feature/full-run-logs/src/FullRunLogsTests.scala
+++ b/integration/feature/full-run-logs/src/FullRunLogsTests.scala
@@ -1,10 +1,12 @@
 package mill.integration
 
+import mill.main.client.OutFiles
 import mill.testkit.UtestIntegrationTestSuite
-import utest._
+import utest.*
 
-// Run simple commands on a simple build and check their entire output,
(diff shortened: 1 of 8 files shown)


lefou (approved):
Looks good to me.

Issue #3867, opened 2024-10-29
Kotlin Multiplatform Support (2500USD Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 2000USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

The goal of this bounty is to flesh out the missing parts of Kotlin-Multi-Platform support. Previous Kotlin support in Mill is mostly for the JVM (https://github.com/com-lihaoyi/mill/issues/3451) with some support for web and very basic Kotlin-JS use cases (https://github.com/com-lihaoyi/mill/issues/3611). The goal here is to flesh out the Kotlin-JS functionality, add support for Kotlin-Native, and improve the overall multi-platform support e.g. publishing and consuming KMP modules and writing KMP applications

* [ ] Example for Kotlin/JS support of browser executables and test runner with selectors (500USD)
    * Currently we only support `node` executables

* [ ] Example for Kotlin/Native support (1000USD)
    * This should be the equivalent of `example/kotlinlib/basic/1-simple` as well as `example/scalalib/native/2-interop` but written using Kotlin-Native.

* [ ] Publishing Kotlin-Multi-Platform modules
    * Proper support for `.module` files such that they can be resolved by Gradle's kotlin-multi-platform integration (500USD)

* [ ] Example of a hello-world Kotlin-Multi-Platform app that uses Jetpack-Compose-Multi-Platform to support Desktop/Android/Web (500USD)
    * This can be a relatively trivial Hello-World application, but it should work on all three platforms using the same Kotlin code

These milestones are open for discussion, and can be modified or additional milestones added if necessary

lihaoyi:
@c0d33ngr i expect we would need a KotlinNativeModule, since kotlin native would have its own flags and configuration. This would be similar to ScalaNativeModule

0xnm:
Any target supported by `Kotlin/Native`, they are listed [here](https://kotlinlang.org/docs/native-overview.html). Obviously, some targets may require additional setup/compiler flags and moreover require a certain platform to be compiled on (e.g. Apple targets can be compiled only on Apple boxes), so if you want to try tackle it I would suggest starting with Linux or Windows target.

The exact use of `Kotlin/Native` depends on the way compiler is distributed, you may want to check first how Kotlin and `Kotlin/JS` code is implemented in `mill`.

lihaoyi:
[...]

PR #3886, opened 2024-11-01
Add `arrow-kt` example

0xnm:
This PR partially addresses #3670: it contains working example, but doesn't contain any docs.

This example contains compilation and test execution for all the modules of https://github.com/arrow-kt/arrow for the JVM and JS (Node) targets, except [arrow-optics-compose](https://github.com/arrow-kt/arrow/tree/main/arrow-libs/optics/arrow-optics-compose) module (it requires better Android support by Mill).

Full example run takes 6 minutes on my machine, so it is extracted into a dedicated CI job.

The following Gradle plugins are not added to the example build script, because their respective tooling is Gradle-dependent (no raw-jar or CLI):

* https://github.com/xvik/gradle-animalsniffer-plugin (although maybe it is possible to use options from https://www.mojohaus.org/animal-sniffer/)
* https://github.com/Kotlin/kotlinx-knit
* https://github.com/Kotlin/binary-compatibility-validator
* https://github.com/diffplug/spotless

There is also no publication support in this example (requires Kotlin Multiplatform publishing support, see https://github.com/com-lihaoyi/mill/issues/3867)

I faced the following issues while writing this example:

* There is quite a lot of duplication for JVM/JS configuration, because there is no support of [Kotlin Multiplatform hierarchies](https://kotlinlang.org/docs/multiplatform-hierarchy.html)
* Each Maven or Module dependency should be with JS/JVM/etc. target qualifier, which also quite explodes amount of code to be written. This is also a question of Kotlin Multiplatform hierarchies support and resolution.
* Until KMP hierarchies and target resolution support is not implemented, adding a new compilation target will bring quite a lot of new code in the build script. And there is a lot of targets in Arrow: https://github.com/arrow-kt/arrow-gradle-config/blob/97ba7b5eab810a336cf4070eb717f05533d208a8/arrow-gradle-config-kotlin/src/main/kotlin/io.arrow-kt.arrow-gradle-config-kotlin.gradle.kts#L34-L69
* There is no possibility to disable tests for the particular module (so I had to use a hack with overriding `compile` / `test` tasks) - this is because tests is dedicated module, but if comes as a part of the trait, we cannot add / remove it dynamically. Having such control is handy when certain multiplatform modules have a common trait for the main compilation unit, but some of them may have no tests.

Regarding docs: I think it is too early to write any comparison with Gradle, because clearly new functionality will be added to the Kotlin support in Mill which will affect overall execution time and also the comparison should be done not only for JVM targets, but for Kotlin/JS and Kotlin/Native targets as well. And ideally it should be a defined methodology for such testing (for example, certain Gradle plugins should be removed from the Arrow Gradle build script, to have the same set as Mill build script; compiler execution time should be excluded (it may be different because of the different flags passed to it under the hood)).

So I propose to open a dedicated bounty for writing such doc.

### DIFF
--- a/example/thirdparty/arrow/build.mill
+++ b/example/thirdparty/arrow/build.mill
@@ -127,18 +127,11 @@ object `package` extends RootModule {
       s"-Xcommon-sources=${files.mkString(",")}"
     }
 
-    trait ArrowPlatformModule extends KotlinMavenModule {  outer =>
+    trait ArrowPlatformModule extends KotlinMavenModule with PlatformKotlinModule {  outer =>
       def kotlinVersion = libraries.versions.kotlin
       def kotlinLanguageVersion = majorVersion(kotlinVersion())
       def kotlinApiVersion = majorVersion(kotlinVersion())
       def kotlinExplicitApi = ArrowMultiplatformModule.this.kotlinExplicitApi
-
-      def platformCrossSuffix: String = millModuleSegments
-        .value
-        .collect { case l: mill.define.Segment.Label => l.value }
-        .last
-
-      override def millSourcePath = super.millSourcePath / os.up
 
       override def sources: T[Seq[PathRef]] = Task.Sources {
         val sourcesRootPath = millSourcePath / "src"
--- a/example/thirdparty/arrow/build.mill
+++ b/example/thirdparty/arrow/build.mill
@@ -154,7 +154,9 @@ object `package` extends RootModule {
         sources
       }
 
-      override def kotlincOptions = super.kotlincOptions() ++ Seq("-Xexpect-actual-classes")
+      override def kotlincOptions =
+        super.kotlincOptions() ++
+        Seq("-Xexpect-actual-classes", commonSourcesArg(sources(), "commonMain"))
 
       trait ArrowPlatformTests extends KotlinMavenTests{
 
@@ -178,7 +180,6 @@ object `package` extends RootModule {
     def jvm: ArrowPlatformJvmModule
 
     trait ArrowPlatformJvmModule extends ArrowPlatformModule with KoverModule {
-      def kotlincOptions = super.kotlincOptions() ++ Seq(commonSourcesArg(sources(), "commonMain"))
       def test: ArrowPlatformJvmTests
       trait ArrowPlatformJvmTests extends KotlinMavenTests with  KoverTests with TestModule.Junit5  with ArrowPlatformTests{
       }
@@ -186,9 +187,7 @@ object `package` extends RootModule {
 
     def js: ArrowPlatformJsModule
     trait ArrowPlatformJsModule extends ArrowPlatformModule with KotlinJsModule {
-      def kotlincOptions = super.kotlincOptions() ++ Seq(
-        commonSourcesArg(sources(), "commonMain")
-      )
+
       def kotlinJsRunTarget = Some(RunTarget.Node)
 
       def test: ArrowPlatformJsTests
@@ -465,7 +464,7 @@ object `package` extends RootModule {
               libraries.kotestAssertionsCoreJs,
               libraries.kotestPropertyJs
             )
-            def testTimeout = 60_000L
+            def testTimeout = 120_000L
           }
         }
 
--- a/example/thirdparty/arrow/build.mill
+++ b/example/thirdparty/arrow/build.mill
@@ -112,15 +112,8 @@ object `package` extends RootModule {
   // TODO add more targets once available
   trait ArrowMultiplatformModule extends CoursierModule { multiplatformRoot =>
 
-    def jsTestModuleDeps: Seq[KotlinJsModule] = Nil
-    def jvmTestModuleDeps: Seq[KotlinModule] = Nil
-    def jsTestIvyDeps: T[Agg[Dep]] = Agg.empty[Dep]
-    def jvmTestIvyDeps: T[Agg[Dep]] = Agg.empty[Dep]
 
-    def jsTestKotlincOptions: T[Seq[String]] = Seq.empty[String]
-    def jvmTestKotlincOptions: T[Seq[String]] = Seq.empty[String]
 
-    def jsTestTimeout: T[Long] = 30_000L
 
     def explicitApi: T[Boolean] = true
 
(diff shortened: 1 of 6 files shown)


lihaoyi:
Thanks, will take a look.

I think splitting the bounty between example and doc is fine. Having an example gives us something to start off that we can critique and improve, and already gives a "realistic" example we can include in the documentation for educational purposes. Having it improved enough to be a meaningful comparison can come in follow ups

example/thirdparty/arrow/build.mill
```diff
@@ -0,0 +1,590 @@
+package build
```
lihaoyi, line 1: CC @lefou @lolgab in case you guys want to review this as well. We've never had a realistic kotlin multiplatform build before so this will be the flagship example.

lihaoyi:
@0xnm I pushed a small update to tidy it up a bit; tests remain passing on my laptop, hopefully will continue passing in CI.

I'll probably take a few more passes trying to tidy this up over the next few days

lihaoyi:
I opened separate bounties specifically for `Spotless` and `AnimalSniffer` since those seem to keep coming up in our examples

* https://github.com/com-lihaoyi/mill/issues/3889
* https://github.com/com-lihaoyi/mill/issues/3888

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -64,6 +64,21 @@ trait KotlinModule extends JavaModule { outer =>
    */
   def kotlinCompilerVersion: T[String] = Task { kotlinVersion() }
 
+  /**
+   * The compiler language version. Default is not set.
+   */
+  def languageVersion: T[String] = Task { "" }
```
lefou, line 70:
```suggestion
  def kotlinLanguageVersion: T[String] = Task { "" }
```

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -64,6 +64,21 @@ trait KotlinModule extends JavaModule { outer =>
...
+
+  /**
+   * The compiler API version. Default is not set.
+   */
+  def apiVersion: T[String] = Task { "" }
```
lefou, line 75:
```suggestion
  def kotlinApiVersion: T[String] = Task { "" }
```

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -64,6 +64,21 @@ trait KotlinModule extends JavaModule { outer =>
...
+
+  /**
+   * Flag to use explicit API check in the compiler. Default is `false`.
+   */
+  def explicitApi: T[Boolean] = Task { false }
```
lefou, line 80:
```suggestion
  def kotlinExplicitApi: T[Boolean] = Task { false }
```

example/thirdparty/arrow/build.mill
```diff
@@ -0,0 +1,689 @@
...
+        }
+
+        override def compile: T[CompilationResult] =
+          if (!modulesWithTestingEnabled.contains(multiplatformRoot))
+            Task { CompilationResult(T.dest, PathRef(T.dest)) }
+          else super.compile
+        override def test(args: String*): Command[(String, Seq[TestResult])] =
+          if (!modulesWithTestingEnabled.contains(multiplatformRoot))
+            Task.Command { ("", Seq.empty[TestResult]) }
+          else super.test(args: _*)
```
lefou, line 185: These are a bit strange. Can we at least add a comment, what and more important why it is happening here?

example/thirdparty/arrow/build.mill
```diff
@@ -0,0 +1,689 @@
...
```
lefou, line 213: Same here, please add a comment why that is needed.

example/thirdparty/arrow/build.mill
```diff
@@ -0,0 +1,689 @@
...
+    `arrow-libs`.fx.`arrow-collectors`,
+    `arrow-libs`.fx.`arrow-fx-coroutines`,
+    `arrow-libs`.fx.`arrow-fx-stm`,
+// Cannot test this one - it relies on the module dependency in the .jar format, while Mill provides .class files
+//    `arrow-libs`.optics.`arrow-optics-ksp-plugin`,
+    `arrow-libs`.optics.`arrow-optics-reflect`,
+    `arrow-libs`.optics.`arrow-optics`,
+
+    `arrow-libs`.resilience.`arrow-resilience`
+  )
```
lefou, line 107: After I better understand why that is needed at all, I'd like to move this into it's own support trait, so we can just mix in all the customized `compile` and `test` tasks, so the actual domain setup is free of these quirks.

lefou (commented):
Since `KotlinModule` mix-in at least `JavaModule`, task like `languageVersion` and `apiVersion` should be explicitly prefixing the language they refer to. Otherwise user may wonder which `api` is meant when they see that task. Is it Java API, or Test API or Kotlin API or domain API? There is also a risk of name clash.

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -326,15 +326,15 @@ trait KotlinModule extends JavaModule { outer =>
   def kotlincOptions: T[Seq[String]] = Task {
     val options = Seq.newBuilder[String]
     options += "-no-stdlib"
-    val kotlinLanguageVersion = languageVersion()
-    if (!kotlinLanguageVersion.isBlank) {
+    val kotlinkotlinLanguageVersion = kotlinLanguageVersion()
```
lefou, line 329: One `kotlin` too much

lihaoyi:
I pushed some more changes to try and DRY up the build further. I'm hitting some failures that I suspect are flakiness, but not sure yet. Still looking into it

0xnm:
JS tests may be flaky, because of NPM (sometimes if fails to download the artifact) and also because some tests may time out sometimes (no idea why).

For the _common_ part of the NPM-related JS tooling (`mocha`, `source-map`) it is probably better to use the common dest/task in the future, to avoid downloading it for every module.

example/thirdparty/arrow/build.mill
```diff
@@ -133,11 +127,11 @@ object `package` extends RootModule {
       s"-Xcommon-sources=${files.mkString(",")}"
     }
 
-    trait ArrowPlatformModule extends KotlinModule {
+    trait ArrowPlatformModule extends KotlinMavenModule {  outer =>
```
0xnm: `KotlinMavenModule` implies that layout is `src/main/...` while for multiplatform setup it is usually `src/commonMain/... ` + `src/<target>Main/...`. What is the advantage of applying it?

lihaoyi: Good question. I think it's pretty arbitrary, but it feels closer than the raw `src/` folder that `KotlinModule` provides, and at least it puts the `src/main/resources` and `src/test/resources` in the right place. I also ended up using it in `ArrowJvmModule` which has the exact correct layout, and so thought it would be nice to be consistent even if it doesn't quite fit

lihaoyi:
I think this looks pretty good, going to merge it for now and close out the bounty. We can continue to improve upon it in master and I can try using it to do a comparative study between Mill and Gradle-Kt

Issue #3837, opened 2024-10-24
Add Mill+Scala.js+WASM example (500USD Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 500USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

Scala.js now supports WebAssembly output (https://www.scala-js.org/news/2024/09/28/announcing-scalajs-1.17.0/). We should have an example doc-test in `example/scalalib/web/` exercising the hello-world Scala.js webassembly dev workflow using Mill and include it in `scalalib/web-examples.adoc` as documentation.

Quafadas:
To be honest, I'll likely have a look at this for my own curiosity in the coming days if no-one were to pick it up beforehand. 

I think the key question, is whether the scope of the example includes a "live reload" solution?

lihaoyi:
> I think the key question, is whether the scope of the example includes a "live reload" solution?

No, I just want a hello-world setup. Nothing fancy, just showing off that we can create a WASM output file and run it through some WASM interpreter or runtime

lihaoyi:
We don't have a browser in the example test suite, so you'll need to find some CLI wasm runtime to replace it

Quafadas:
> We don't have a browser in the example test suite, so you'll need to find some CLI wasm runtime to replace it

Hmmm, right, I hear you. The other option could be to wheel in playwright? Doesn't need an answer now. I'll explore it.

yadavshubham01:
hey @lihaoyi look quite interesting working on it

lihaoyi:
Go for it, bounties remain open until they are closed with a merged PR

Quafadas:
@yadavshubham01 

i guess for the avoidance of conflict - it’s all yours for a few days :-).

yadavshubham01:
hey @lihaoyi checkout this pr #3839

PR #3897, opened 2024-11-02
Emit WASM

Quafadas:
Seeks to address #3837 

I have some outstanding questions;

1. As I'm altering the scalaJSModule public API, a double check here would be welcomed. In particular, I've just called the module property `emitWasm`. sjrd pretty clearly called the linker API `withExperimentalUseWebAssembly`. However, if at some point it became non-experimental, I assume that's a compatibility problem. Advice welcomed - I've gone with the simplest name possible - `emitWasm`.
2. CI will fail, because node is < 22. The runnable test does work locally for me. Does mill require, assume or otherwise advertise compatibility with any particular node version? If not, I'll simply push the node version forward in the GHA definitions. OItherwise, I'd welcome advice - I'm not clear what the best way to manage the CI would be. 

Those are the two points where I'd welcome advice. I think this is the path to victory;

- [x] public API check? 
- [x] node version
- [x] documentation
- [x] CI green

### DIFF
--- a/scalajslib/src/mill/scalajslib/ScalaJSModule.scala
+++ b/scalajslib/src/mill/scalajslib/ScalaJSModule.scala
@@ -296,7 +296,7 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
   /** Whether to emit a source map. */
   def scalaJSSourceMap: T[Boolean] = Task { true }
 
-  /** Whether to emit WASM map. As of Nov 2024 scala JS wasm support is experimental */
+  /** Whether to emit WASM. As of Nov 2024 scala JS wasm support is experimental */
   def emitWasm: T[Boolean] = Task { false }
 
   /** Name patterns for output. */
--- a/scalajslib/src/mill/scalajslib/ScalaJSModule.scala
+++ b/scalajslib/src/mill/scalajslib/ScalaJSModule.scala
@@ -140,7 +140,8 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
       moduleSplitStyle = moduleSplitStyle(),
       outputPatterns = scalaJSOutputPatterns(),
       minify = scalaJSMinify(),
-      importMap = scalaJSImportMap()
+      importMap = scalaJSImportMap(),
+      emitWasm = emitWasm()
     )
   }
 
@@ -191,7 +192,8 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
       moduleSplitStyle: ModuleSplitStyle,
       outputPatterns: OutputPatterns,
       minify: Boolean,
-      importMap: Seq[ESModuleImportMapping]
+      importMap: Seq[ESModuleImportMapping],
+      emitWasm: Boolean
   )(implicit ctx: mill.api.Ctx): Result[Report] = {
     val outputPath = ctx.dest
 
@@ -212,7 +214,8 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
       moduleSplitStyle = moduleSplitStyle,
       outputPatterns = outputPatterns,
       minify = minify,
-      importMap = importMap
+      importMap = importMap,
+      emitWasm
     )
   }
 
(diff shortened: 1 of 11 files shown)


lihaoyi:
Thanks @Quafadas!

1. Updating the ScalaJSModule public API is fine. Let's prefix the name `scalaJSEmitWasm`, since it's a Scalajs specific task.
2. Mill does not advertise any required node version, so you can require whatever version you want, just make sure to document it (e.g. in `def scalaJSEmitWasm`)
3. Let's also include an `example/scalalib/web/` example test with a short illustrative example and some english prose explaining what is going on

Quafadas:
The documentation has been written somewhat blindly - I can't seem to get past this locally. 

```
simon@Simons-Mac-mini mill-1 % mill docs.localPages
[0753/1698] docs.oldDocSources
[0753] Cloning into '/Users/simon/Code/mill-1/out/docs/oldDocSources.dest/0.11.12'...
[0753] done.
[0753] error: pathspec '0.11.12' did not match any file(s) known to git
[1698/1698] ============================================================================================================================================================================================== docs.localPages ================================================================================================================================================================================================== 1s
1 tasks failed
docs.oldDocSources os.SubprocessException: Result of git…: 1

    os.proc.call(ProcessOps.scala:230)
    build_.docs.package_.$anonfun$oldDocSources$4(package.mill:247)
    build_.docs.package_.$anonfun$oldDocSources$3(package.mill:244)
```

lefou:
@Quafadas Could be you have a sparse git checkout. You can use `mill docs.fastPages` which should skip the older versions.

scalajslib/src/mill/scalajslib/ScalaJSModule.scala
```diff
@@ -293,6 +296,9 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
   /** Whether to emit a source map. */
   def scalaJSSourceMap: T[Boolean] = Task { true }
 
+  /** Whether to emit WASM. As of Nov 2024 scala JS wasm support is experimental */
+  def scalaJSEmitWasm: T[Boolean] = Task { false }
```
lolgab, line 300: I would stick to the same names used by the official Scala.js sbt plugin.
In this case since the method is `withExperimentalUseWebAssembly` then I would use:
```scala
def scalaJSExperimentalUseWebAssembly: T[Boolean] = Task { false }
```

lolgab: Then you could rename the fields and parameters accordingly in the private methods and files.

Quafadas: I'm happy to do this - the caveat is that if wasm were to become non-experimental, it would break compatibility to declare it as such?

Quafadas:
@lefou Thankyou! I think the docs works as I expect...

Quafadas:
Pending further feedback and making CI happy, I believe this now to be mergeable.

scalajslib/src/mill/scalajslib/ScalaJSModule.scala
lolgab: We will follow the same process as Scala.js, deprecate the field and tell people to use the new one.
We've done this already with [useECMAScript2015](https://github.com/com-lihaoyi/mill/pull/1671/files#diff-205bc876ab2361a662d87fc8432d83a29d06f56b88de596df28dd03ae5ddd7a8R200)

scalajslib/src/mill/scalajslib/ScalaJSModule.scala
```diff
@@ -293,6 +296,9 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
   /** Whether to emit a source map. */
   def scalaJSSourceMap: T[Boolean] = Task { true }
 
+  /** Whether to emit WASM. As of Nov 2024 scala JS wasm support is experimental */
```
lolgab, line 299: You can borrow some of the docs from scala.js: https://github.com/scala-js/scala-js/blob/b988843fd3718c26bcc41313a56632efdad1d616/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala#L189-L218

Quafadas: 👍

Quafadas:
I have not (sadly) won my little war with the CI. I think the functionality, tests and docs I introduced are working as intended.

I don't _think_ that the remaining failure, is related to my changes? Not 100% sure what to do next - any views?

lihaoyi:
Looks good to me, thanks @Quafadas! Email me your international bank transfer details and I will close out the bounty

Quafadas:
@lihaoyi r.e. bounty I'd say don't worry about it I get plenty out of mill so let's keep that resource inside the project. 

You can either add it to the scala 3 migration slush fund, or if you ever get the chance to have dinner with Tobias and/or Lorenzo - make it a nice one.

lihaoyi:
@Quafadas got it! In that case I'll pass the bounty to the scala center then to support their work

Issue #3910, opened 2024-11-05
Junit5 test discovery reports spurious test runs for non-test classes (500USD bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 500USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

First noticed when reviewing https://github.com/com-lihaoyi/mill/pull/3886

```
lihaoyi mill$ ./mill dist.run example/kotlinlib/basic/1-simple -i foo.test
[2617/2617] dist.run
======================================================== foo.test ====================================================
======================================================================================================================
[99/99] foo.test.test
[99] Warning: Kotest autoscan is enabled. This means Kotest will scan the classpath for extensions that are annotated with @AutoScan. To avoid this startup cost, disable autoscan by setting the system property 'kotest.framework.classpath.scanning.autoscan.disable=true'. In 6.0 this value will default to true. For further details see https://kotest.io/docs/next/framework/project-config.html#runtime-detection
[99] Test run started (Kotest)
[99] Test foo.FooTesttestSimple started
[99] Test foo.FooTesttestSimple finished, took 0.038s
[99] Test foo.FooTesttestEscaping started
[99] Test foo.FooTesttestEscaping finished, took 0.001s
[99] Test foo.FooTest finished, took 0.09s
[99] Test  finished, took 0.11s
[99] Test  finished, took 0.002s
[99] Test run finished: 0 failed, 0 ignored, 2 total, 0.14s
[99] Warning: Kotest autoscan is enabled. This means Kotest will scan the classpath for extensions that are annotated with @AutoScan. To avoid this startup cost, disable autoscan by setting the system property 'kotest.framework.classpath.scanning.autoscan.disable=true'. In 6.0 this value will default to true. For further details see https://kotest.io/docs/next/framework/project-config.html#runtime-detection
[99] Test  finished, took 0.001s
[99] Test  finished, took 0.0s
[99] Test run finished: 0 failed, 0 ignored, 0 total, 0.001s
[99] Warning: Kotest autoscan is enabled. This means Kotest will scan the classpath for extensions that are annotated with @AutoScan. To avoid this startup cost, disable autoscan by setting the system property 'kotest.framework.classpath.scanning.autoscan.disable=true'. In 6.0 this value will default to true. For further details see https://kotest.io/docs/next/framework/project-config.html#runtime-detection
[99] Test  finished, took 0.0s
[99] Test  finished, took 0.0s
[...]

PR #3501, opened 2024-09-10
Update jupiter-interface from 0.11.4 to 0.13.0

lihaoyi:

PR #3911, opened 2024-11-05
Filter out anonymous classes during test discovery

lihaoyi:
Mitigates https://github.com/com-lihaoyi/mill/issues/3910 by hackily filtering out all anonymous classes, but non-anonymous non-test classes still cause the problem

Added a unit test to check this specific misbehavior

### DIFF
--- a/kotlinlib/test/resources/hello-world-kotlin/main/kotest/src/FooTest.kt
+++ b/kotlinlib/test/resources/hello-world-kotlin/main/kotest/src/FooTest.kt
@@ -1,0 +1,15 @@
+package hello.tests
+
+import io.kotest.core.spec.style.FunSpec
+import io.kotest.matchers.shouldBe
+import hello.getHelloString
+
+class FooTest : FunSpec({
+    test("testFailure") {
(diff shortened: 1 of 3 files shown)


lefou:
Since we're more or less just reusing the `JupiterFramework`, we should first check upstream, whether these spurious discovered test classes are incidentally or not. THB, I have no idea of the complete feature set of Junit5.

lefou:
It would be nice to also update to the latest version of `jupiter-interface`. Maybe this is already fixed upstream?

* https://github.com/com-lihaoyi/mill/pull/3501

lihaoyi:
@lefou oh thsts a good point, should take another look at that pr

lihaoyi:
merging this first just as a mitigation. The upstream sbt-jupiter integrstion does do some special stuff beyond just being a test framework, so likely we need to see what its doing and replicate that logic in Mill

lefou:
I'm going to cut a snapshot release, as I have a project where we have lots of these spurious zero-ran test runs. This is also a nice test for the new publish pipeline (https://github.com/com-lihaoyi/mill/pull/3904).

PR #3769, opened 2024-10-18
Hello world Kotlin Jetpack Compose Example; Fixes:#3550

himanshumahajan138:
# Pull Request 
Fixes: #3550 

## Description
Added `Kotlin Android Jetpack Compose Example` insuring proper integration of `AndroidAppKotlinModule` with `Jetpack Compose`

## Related Issues
- Link to related issue #3550.

## Checklist
- [x] Android hello world application with Jetpack Compose integration.
- [x] Code is clean and follows project conventions.
- [x] Documentation has been updated (to be done after review)

## Additional Notes
Not Fully Complete and Correct

PR #3913, opened 2024-11-05
Improve Android SDK handling

0xnm:
This PR seeks to improve Android SDK handling approach.

The current issues are:

* Android SDK setup is [limited only to Linux](https://github.com/com-lihaoyi/mill/blob/9d11962eaf529474dd66e095620a97cff66ebcd3/scalalib/src/mill/javalib/android/AndroidSdkModule.scala#L24) and is also using [tooling which may be not available on other platforms](https://github.com/com-lihaoyi/mill/blob/9d11962eaf529474dd66e095620a97cff66ebcd3/scalalib/src/mill/javalib/android/AndroidSdkModule.scala#L102-L109).
* Each module [will have its own copy of SDK](https://github.com/com-lihaoyi/mill/blob/9d11962eaf529474dd66e095620a97cff66ebcd3/scalalib/src/mill/javalib/android/AndroidSdkModule.scala#L114). Imagine a project with 20 Android modules (19 lib modules + 1 app module), it will be 20x space usage than it should be. This is quite significant, considering that the basic setup (`platforms:android-35`, `platform-tools` and `build-tools;35.0.0`) is taking ~300 MB.
* Licenses are [accepted on the end-user behalf](https://github.com/com-lihaoyi/mill/blob/9d11962eaf529474dd66e095620a97cff66ebcd3/scalalib/src/mill/javalib/android/AndroidSdkModule.scala#L108), without the consent.

This PR:

* Aligns the SDK setup experience with AGP (Android Gradle Plugin):
  * Expects Android SDK (at least command line tools) to be present on the end-user machine and registered with the env variable (`ANDROID_HOME`).
  * Expects command line tools to be present there to utilize [sdkmanager](https://developer.android.com/tools/sdkmanager) to setup the necessary SDK components.
  * Expects user to accept the necessary licenses by themselves. If they are not accepted, `sdkmanager` will block `stdin` waiting for the user input before proceeding with installation. **Note**: when downloading [Android Studio](https://developer.android.com/studio), it will come with the necessary basic license already accepted + some basic SDK components. If downloading command line tools, it is necessary to accept the main license (`android-sdk-license`) manually.
* Installs all the necessary components to the location pointed by the `ANDROID_HOME` env variable, so that they are shared by all modules.
* Adds the necessary step on the CI to install Android SDK for the jobs where it is needed. **Note**: I was also thinking about extracting Android-related examples into a dedicated job, but then for the existing jobs doing something like `example.kotlinlib[__].local.testCached` ->`example.kotlinlib[__:^AndroidAppKotlinModule].local.testCached` doesn't work (at least `mill resolve` still shows Android module), all cross-segments should be listed manually.

Speaking of `sdkmanager`: it is also possible to utilize directly the [following artifact](https://mvnrepository.com/artifact/com.android.tools/sdklib) and put it in a worker, but for this one it is impossible to find a version on the Android Developers website, so end-user will have troubles updating it and anyway it is one-shot action normally, so keeping it even in the separate classpath looks like an overkill. [Command line tools](https://developer.android.com/tools) will be needed anyway further for the other tools it provides.

### DIFF
--- a/.github/workflows/run-mill-action.yml
+++ b/.github/workflows/run-mill-action.yml
@@ -27,6 +27,9 @@ on:
       env-bridge-versions:
         default: 'none'
         type: string
+      install-android-sdk:
+        default: false
+        type: boolean
 
 jobs:
   run:
@@ -62,6 +65,10 @@ jobs:
         with:
           node-version: '22'
 
+      - uses: android-actions/setup-android@v3
+        if: ${{ inputs.install-android-sdk }}
+        with:
+          log-accepted-android-sdk-licenses: false
 
       - name: Prepare git config
         run: |
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -65,10 +65,12 @@ jobs:
             millargs: "contrib.__.testCached"
 
           - java-version: 17
+            install-android-sdk: true
             millargs: "'example.javalib.__.local.testCached'"
           - java-version: 17
             millargs: "'example.scalalib.__.local.testCached'"
           - java-version: 17
+            install-android-sdk: true
             millargs: "'example.kotlinlib.__.local.testCached'"
           - java-version: '11'
             millargs: "'example.thirdparty[{mockito,acyclic,commons-io}].local.testCached'"
@@ -91,6 +93,7 @@ jobs:
 
     uses: ./.github/workflows/run-mill-action.yml
     with:
+      install-android-sdk: ${{ matrix.install-android-sdk }}
       java-version: ${{ matrix.java-version }}
       millargs: ${{ matrix.millargs }}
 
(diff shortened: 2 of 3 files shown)


0xnm:
By some reason CI fails (while there is an installation of SDK components in the logs), I will investigate.

himanshumahajan138:
@0xnm Sir your approach is very good but i want to add something that i did during kotlin example please once see this pr #3769

here please see the `AndroidAppModule` and `AndroidSdkModule`


main points:
- used cache for downloaded setup 
- updated the code for android setup 


actually in that PR i have done a alot of progress but the only thing i am stucked with library resources (aar files packaging) and almost all things are done i have updated the setup for cache and also updated the android app creation process but i was unable to solve the addition of jetpack compose fully(only aar packaging left)

i request you to please once have a look and take some insights i think this will help...

0xnm:
That was my miss, I forgot to add the line https://github.com/com-lihaoyi/mill/pull/3913/files#diff-7314d0ebbd2e9537ae4889316745b4fd2fa43cb86275c9caae18a86ba228b642R96, CI is fine now.

lihaoyi:
Thanks! Will take a look

Issue #3871, opened 2024-10-29
Re-run netty maven comparison with updated flags


lihaoyi:
* https://github.com/netty/netty/discussions/14420

We can either disable RevAPI and Animal Sniffer in the Maven build, or enable them in our Mill build. Either would help ensure a fair comparison. My cursory indicates the comparative numbers shouldn't change by too much.

PR #3918, opened 2024-11-07
Re-run Maven Netty comparison using `-Pfast` flag, fix SBT compile all benchmark

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/3871. I also moved the `./mvnw clean` into a separate command to avoid counting that time spent towards the Maven benchmark

The differences between Maven and Mill are no longer quite as stark, but they are still very significant. The Mill benchmarks are about the same, as expected, so I skipped running the last few

For the Gatling/SBT comparison, SBT `compile` does not compile test classes, while `test:compile` does. So Using SBT's `test:compile` is necessary for an apples-to-apples comparison with Mill's `__.compile` which does compile test classes


lihaoyi:
Failure seems unrelated

PR #3903, opened 2024-11-04
Apply Scalafmt to `examples/` in CI

myyk:
Partially completes: #3829

Tested locally by breaking format of `example/scalalib/basic/1-simple/foo/src/Foo.scala` by just adding a bunch of spaces.

* Observe break with `./mill mill.scalalib.scalafmt.ScalafmtModule/checkFormatAll __.sources` failure.
* Observe fix with `./mill mill.scalalib.scalafmt.ScalafmtModule/checkFormatAll __.sources`

Used `// format: off` rather than changing [docstrings.style = keep](https://scalameta.org/scalafmt/docs/configuration.html#docstringsstyle--keep) to our `.scalafmt.conf` since that would affect all files. There doesn't seem to be a built in way in ScalaFmt to have folder level configs or something like this. I thought it might be better to not have `example` be an exception cases if we can avoid it. I'm not sure if the `// format: off` comments break something else though.

Issue #3829, opened 2024-10-24
Enforce ScalaFmt/PalantirJavaFormat/KtLint on example doc/test suites (500USD Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 500USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

Currently managing the formatting for this is manual, and sometimes inconsistencies slip through. We should really just automate enforcement and compliance.

We should as much as possible re-use the infrastructure Mill has already built around these autoformatters, to dogfood it. This would mean doing the autoformatting as part of Mill's own Mill build, using the various `ScalaFmtModule`/`PalantirJavaFormatModule`/`KtLintModule`s as necessary

LevisNgigi:
Hello interesting but I think this might not be beginner friendly for someone new to the project.Will be on the lookout for other issues.

Sailsman63:
In this context, what does "Automate enforcement and compliance" look like?

- Just surface the errors in the build script CI, block PRs on fixing them
- Automatically apply and commit fixes on merge
- Other?

lihaoyi:
> Just surface the errors in the build script CI, block PRs on fixing them

Just this, with some manual command that can be run to fix them locally. Same as the current workflow.

myyk:
I'm going to see if I can figure this one out tomorrow

lihaoyi:
@myyk some pointers if you're going to look into this:

1. This should be purely a change in the `build.mill`/`package.mill` files, shouldn't need any application code changes

2. The default `mill.scalalib.scalafmt.ScalafmtModule` defaults to [looking for `__.sources`](https://github.com/com-lihaoyi/mill/blob/d5d56b5d55d59972bbe8bef2cb647e4c5f96fdd3/scalalib/src/mill/scalalib/scalafmt/ScalafmtModule.scala#L77-L99), so the easiest thing to do would be to hook into that by defining a `sources` in `ExampleCrossModule` that points to `testRepoRoot` so it can get picked up by `ScalafmtModule/formatAll` and `checkFormatAll`

3. To make this work for Java sources as well, we would need to run `PalantirFormatModule`  in addition to `ScalafmtModule` in CI [here](https://github.com/com-lihaoyi/mill/blob/d5d56b5d55d59972bbe8bef2cb647e4c5f96fdd3/.github/workflows/run-tests.yml#L155)

[...]

PR #3919, opened 2024-11-07
Autoformat Java examples

myyk:
* [x] Need to merge #3903 first because it's built directly on top of it to reduce conflicts.

Change 2/3 for #3829

* I updated the `PalantirFormatModule.scala` so that it didn't create a ton of noise when there's actually no issues such as no java code changing. I think this is a good thing to do, but I'm not 100% since some users might feel this change in some way.
* Bugfix: The Github Action now fails on any error instead of only `./mill -i __.fix --check` (was using `;` instead of `&&`)
* Added Palantir formatting into the Github Action
* Removed a lot of cruft so that the example java modules could use the `PalantirFormatModule`.


example/javalib/linting/2-checkstyle/src/InputWhitespaceCharacters.java
```diff
@@ -12,8 +12,10 @@ final class InputWhitespaceCharacters {
    * @return hack
    * @throws java.lang.Exception abc
    **/
-  int test1(int badFormat1, int badFormat2,
-  		final int badFormat3) // violation 'Line contains a tab character.'
+  int test1(
+      int badFormat1,
+      int badFormat2,
+      final int badFormat3) // violation 'Line contains a tab character.'
```
myyk, line 18: TODO: Need to restore this lint error and find a way to allow it as an exception. (Can't finish today.)

example/javalib/linting/3-palantirformat/src/A.java
```diff
@@ -1,6 +1,6 @@
 public class A {
 
-    public static void main(String[] args) {
+  public static void main(String[] args) {
     System.out.println("hello"); // indentation should be fixed
```
myyk, line 4: TODO: Need to restore this lint error and find a way to allow it as an exception. (Can't finish today.)

example/javalib/linting/3-palantirformat/src/A.java
```diff
@@ -1,6 +1,6 @@
...
-    }
+  }
 }
```
myyk, line 6: TODO: Need to restore this lint error and find a way to allow it as an exception. (Can't finish today.)

myyk:
@lihaoyi I'm missing some fixes on keeping the linting examples there. Otherwise I think this one is about ready.

example/javalib/linting/3-palantirformat/src/A.java
myyk: Done

example/javalib/linting/3-palantirformat/src/A.java
myyk: Done

example/javalib/linting/2-checkstyle/src/InputWhitespaceCharacters.java
myyk: Done

myyk:
@lihaoyi I think this one is ready for a review when you get a chance.

lihaoyi (approved):
Looks great, I think just need to update the test cases and we're good to go

lihaoyi:
Looks like a bunch of `PalantirFormatModuleTest` tests are failing. I'm guessing those expect their test data to be mis-formatted, so we might need to figure out a way to skip those

lihaoyi:
So PalantirJavaFormat does not provide any builtin flags to ignore files or folders, instead Mill manages the listing of files to pass to the tool.

https://github.com/com-lihaoyi/mill/blob/4c05156ccd8eca733411139b40886af7fcc485dc/scalalib/src/mill/javalib/palantirformat/PalantirFormatModule.scala#L173-L181

So we likely need to add some logic in Mill to allow `PalantirJavaFormatModule` to ignore particular globs, and then use that in Mill's own build (after rebootstrapping) to ignore the files and roll out enforcement internally.

To do this, I suggest we overload `PalantirFormatModule.formatAll` to take an additional optional parameter `excludedGlobs: Seq[String]`. We can then plumb those through to the java-file-listing above and use it to ignore files matching a glob (may need to find a glob-matching implementation online). We'd need to preserve the current `formatAll` signature for binary compatibility, but it can be a dumb forwarder to the new implementation

myyk:
I had actually already implemented skipping some files from the .mill side
but I guess I missed something.

myyk:
The glob solution seems like a good way to go but I'm not sure when I can
do it. Kotlin needs a similar mechanism and I think it would be good to
have all the linter modules have roughly the same interface (where
possible) which they don't currently.

lihaoyi:
Ah I just noticed your skipping code in Mill's own build. Not sure why it's not working

lihaoyi:
Pushed some fixes, let's see if this works

myyk:
Oh interesting. I see you changed the `sources` in `example/build.mill`, is that what added a lot more linting to this? That found a lot of files I missed.

lihaoyi:
@myyk yeah, previously we only used `lintableJavaSources` in `PalantirFormatModule/`, and didnt use it in `ScalafmtModule/` so the Scala `.mill` files were not picked up. Furthermore, `sources` was pointed at `testRepoRoot`, which for the `javalib/` examples contains a copy of the files and not the original ones so formatting them doesn't do anything.

By re-using the name `sources`, both `PalantirFormatModule/` and `ScalafmtModule/` pick up all of them, which is why now the `.mill` files in the Java folders are being picked up. And I re-targeted `sources` directly at `millSourcePath`, so it doesn't get affected by the override of `testRepoRoot` in `ExampleCrossModuleJava`

myyk:
That all makes sense. Oh, I think I fixed some of that in the next change #3923 and didn't back port it because I thought it was working enough in this PR. Sorry about that. But also, I wasn't sure about fully redefining `sources` like that because I'm new to the project. That makes a lot of sense though and I'm glad for the help.

PR #3922, opened 2024-11-07
Fix spelling errors

0xnm:
I noticed quite some spelling errors in the codebase, so I decided to fix some of them by running spell-checker.


testrunner/src/mill/testrunner/TestRunnerMain0.scala
```diff
@@ -47,7 +47,7 @@ import mill.util.PrintLogger
         println(e)
         e.printStackTrace()
     }
-    // Tests are over, kill the JVM whether or not anyone's threads are still running
```
lihaoyi, line 50: The old text was correct in this case

testrunner/src/mill/testrunner/TestRunnerUtils.scala
```diff
@@ -70,7 +70,7 @@ import scala.jdk.CollectionConverters.IteratorHasAsScala
             .flatten
         )
       }
-      // I think this is a bug in sbt-junit-interface. AFAICT, JUnit is not
```
lihaoyi, line 73: The old text was correct here

scalajslib/src/mill/scalajslib/ScalaJSModule.scala
```diff
@@ -221,7 +221,7 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
 
   override def mandatoryScalacOptions: T[Seq[String]] = Task {
     // Don't add flag twice, e.g. if a test suite inherits it both directly
-    // ScalaJSModule as well as from the enclosing non-test ScalaJSModule
```
lihaoyi, line 224: `as well as` is probably better here

lihaoyi:
Thanks! Looks good, just some false positives

PR #3923, opened 2024-11-08
Lint Kotlin examples

myyk:
* [x] Needs #3919 as this builds on it
* [x] Update this to use #3961, switching to `KtlintModule`
* [x] Needs #3966 to disable ktlint for some files

Last change I think needed for #3829

Overview:
* Changes `KtfmtModule` to set the error code when there's changes like is true of the other formatters.
* Some simplifications were made to the `example/package.mill` based on similarities with Java/Kotlin.
* Files were formatted
* CI was updated


example/kotlinlib/basic/1-simple/foo/test/src/foo/FooTest.kt
```diff
@@ -1,15 +1,11 @@
...
 
-import foo.generateHtml
 import io.kotest.core.spec.style.FunSpec
 import io.kotest.matchers.shouldBe
 
-class FooTest : FunSpec({
-    test("testSimple") {
-        generateHtml("hello") shouldBe "<h1>hello</h1>"
-    }
+class FooTest :
```
0xnm, line 6: this doesn't look like a reasonable change, but I don't know `ktfmt` default rules. Overall in the Kotlin system `ktlint` is much more popular than `ktfmt`.

myyk: This is just the default `ktfmt` as there's no configuration. I'm guessing it's better to any formatter in place and we can let which the formatter battle to happen in parallel since it should then be a fairly easy switch.

myyk:
I think I need to adjust the linter tests bc I've changed the implementation. Or change the default to not see the status code.

example/kotlinlib/basic/1-simple/foo/test/src/foo/FooTest.kt
0xnm: Then why not use `ktlint` (as it is proposed in https://github.com/com-lihaoyi/mill/issues/3829) provided by `KtlintModule`? It will require some changes to the run in the module-independent manner, but it will be easy to do by looking on `KtfmtModule`.

myyk: Oh yeah, it is `ktlint` in the issue. I misread that. Forgive me, I am actually dyslexic.

lihaoyi: No worries, let's try `ktlint` and see how it looks. Then we can pick the one that looks better.

I noticed `KtlintModule` does not have the companion `object extends External Module` that the other autoformatters do, making it more annoying to use ad-hoc. We should probably add such an companion external module. `KtfmtModule`, `ScalafmtModule`, and `PalantirFormatModule` can be used as a reference as to how it is done

myyk:
@lihaoyi I can fix up the tests if you think you're happy with the way the formatting looks now.

(I'd do it tonight but Starhub is having an outage *shakes fist*)

.editorconfig
```diff
@@ -13,3 +13,16 @@ insert_final_newline = true
...
+ktlint_standard_no-consecutive-blank-lines = disabled
+ktlint_standard_final-newline = disabled
+ktlint_standard_filename = disabled
+ktlint_standard_property-naming = disabled
+ktlint_standard_class-naming = disabled
+ktlint_standard_no-consecutive-comments = disabled
+ktlint_standard_import-ordering = disabled
+ktlint_standard_function-naming = disabled
+ktlint_standard_backing-property-naming = disabled
+ktlint_standard_no-empty-file = disabled
```
0xnm, line 28: what is the need for disabling so many defaults?

Also, to have a less of the change in this PR due to re-formatting you can change the default style `intellij_idea` (see https://pinterest.github.io/ktlint/latest/rules/configuration-ktlint/#code-style), because Intellij IDEA built-in formatter was used for the majority of Kotlin files in this repo.

myyk:
There's failures in here because `ktlint` is formatting the files in `kotlinlib/test/resources/contrib/ktfmt`. We would like to skip formatting on those files and possibly other files such as all of the ones in `out/`.

`ktlint` says they can ignore files, but it seems to not be working for me: https://pinterest.github.io/ktlint/latest/faq/#how-do-i-disable-ktlint-for-generated-code
```
[out/**/*]
ktlint = disabled
```

0xnm:
It should be working, it was added quite a time ago in `1.0.0` and I suppose it is stable. I would suggest checking what is the value of the working dir when running `ktlint`. Disabling so many rules to solve this issue seems fragile and quite hacky approach.

myyk:
@0xnm, yeah, I saw that too about it being added in `1.0.0`. I just did a deep dive. It's because the `.editorconfig` file isn't being configured to be used... This is all very strange to me because it's settings seem to be being picked up even when it's not set to be used.

This is true while directly calling `ktlint` from the command line using `java` as well. I think it's defaulting to `.editorconfig` for formatting but not using it for disabling. It's weird. I'll make a PR to set that path.

myyk:
I have 0 opinions on how Kotlin code should be formatted. I am happy for @0xnm and @lihaoyi to sort that out.

lihaoyi:
The reason i disabled all the rules was mostly to just focus on formatting, rather than other linters since thats what our other autoformatters do. I don't mind turning them back on if @0xnm prefers, it would just mean a bit more work to fix up the code to compliance

lihaoyi:
@myyk let's go with the intellij style that @0xnm suggested, and removing all the config overrides except one that we leave just to exercise the "ktlint picks up the config" code path. You'll need to resolve the various lint errors but it shouldnt be too hard

myyk:
Looks like we need to revert the expected broken ktlint code and ignore it in .editorconfig like for ktfmt test examples.

lihaoyi:
`example/kotlinlib/linting/2-ktlint` is still failing. The problem appears to be that ktlint traverses the filesystem hierarchy, starting from its PWD of `out/example/kotlinlib/linting/2-ktlint/local/test.dest/sandbox`, all the up the `.`, and then uses that `.editorconfig` file to ignore the files inside.

lihaoyi:
For now I skipped the `out/**/*` lint seems that's the problematic line that blocks all linting from happening in the out folder, and is picked up even when running inside a subfolder.

myyk:
Your finding is another interesting one. `ktlint` is full of a lot of surprises. I don't understand when it wants to follow the config and when it doesn't.

lihaoyi:
@myyk it appears we're not the only one to hit issues https://www.reddit.com/r/Kotlin/comments/1gdlc73/ktlint_alternatives/, but I guess it's the standard so that's not our problem to fix.

myyk:
@lihaoyi yeah, I agree.

lihaoyi:
Going to call this done and close out the bounty over PayNow

Issue #1975, opened 2022-07-28
Support for (transitive) dependency management / override


lefou:
Mill currently has no way to influence the transitive version management. Since we already have the `CoursierModule.mapDependencies` task, this is no issue for projects that build applications. But for libraries projects, controlling of downstream dependency selection might be a useful feature. 

Once we add such feature, we need to make sure, that dependency resolution behaves consistent between inner-module and ivy-module dependencies. Also, we want to write this information in `pom.xml` under `<dependencyManagement>` as well as into `ivy.xml` as dependency override section.

Also sbt does not have this features. It allows to force or specify (transitive) dependency versions, which is what we can do with `mapDependencies`. But it does not support to generate `dependencyManagement`-sections in `pom.xml`.

Some links: 
* Maven: https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-management
* Ivy: https://ant.apache.org/ivy/history/2.5.0/ivyfile/override.html
* sbt: https://www.scala-sbt.org/1.x/docs/Library-Management.html

alexarchambault:
> Mill currently has no way to influence the transitive version management.

There's at least two:
- adding dependencies ensures their version is >= to the added one
- exclusions

Post-processing dependencies is somewhat harmful: if users publish a module where they post-processed dependencies this way, users of the published module won't benefit from the post-processing (while they do benefit of the two points above, which do end up in the POM).

And https://github.com/coursier/coursier/issues/1390 won't change that: if a project uses it and is published, versions aren't forced the same way for its users (when pulling dependencies that have a BOM, the BOM is only used to fill "holes", that is missing versions in the POM - no version is otherwise forced whatsoever). https://github.com/coursier/coursier/issues/1390 is not a way to share forced dependencies for library authors / users, unless users explicitly add the BOM to their project (and if all libraries each supply their own BOM, it's likely some might conflict…)

lefou:
> > Mill currently has no way to influence the transitive version management.
> 
> There's at least two:
> 
>     * adding dependencies ensures their version is >= to the added one
> 
>     * exclusions
> 

Adding dependencies has the downside, that you misuse the mechanism, as you most likely list dependencies which are only transitively needed. There is a high risk those entries run out of date. The reasons may vary, but will be most likely that the upstream dependencies have changed. There is also the somewhat smaller risk to produce incompatibilities combinations.

[...]

PR #3924, opened 2024-11-08
Add BOM / dependency management support

alexarchambault:
This PR adds support for user-specified BOMs and dependency management in Mill.

BOM support allows users to pass the coordinates of an existing Maven "Bill of Material" (BOM), such as [this one](https://repo1.maven.org/maven2/io/quarkus/quarkus-bom/3.17.0/quarkus-bom-3.17.0.pom), that contains versions of dependencies, meant to override those pulled during dependency resolution. (They can also add exclusions to dependencies.)
```scala
def bomDeps = Agg(
  ivy"io.quarkus:quarkus-bom:3.17.0"
)
```

It also allows users to specify the coordinates of a parent POM, which are taken into account just like a BOM:
```scala
def parentDep = ivy"org.apache.spark::spark-parent:3.5.3"
```
(in line with `PublishModule#pomParentProject` that's been added recently)

It allows users to specify "dependency management", which act like the dependencies listed in a BOM: versions in dependency management override those pulled transitively during dependency resolution, and exclusions in its dependencies are added to the same dependencies during dependency resolution.
```scala
def dependencyManagement = Agg(
  ivy"com.google.protobuf:protobuf-java:4.28.3",
  ivy"org.java-websocket:Java-WebSocket:_" // placeholder version - this one only adds exclusions, no version override
        .exclude(("org.slf4j", "slf4j-api"))
)
```

BOM and dependency management also allow for "placeholder" versions: users can use `_` as version in their `ivyDeps`, and the version of that dependency will be picked either in dependency management or in BOMs:
```scala
def bomDeps = Agg(
  ivy"com.google.cloud:libraries-bom:26.50.0"
)
def ivyDeps = Agg(
  ivy"com.google.protobuf:protobuf-java:_"
)
```

A tricky aspect of that PR is that details about BOMs and dependency management have to be passed around via several paths:
- in the current module: BOMs and dependency management have to be taken into account during dependency resolution of the module they're added to
- via `moduleDeps`: BOMs and dependency management of module dependencies have to be applied to the dependencies of the module they come from
- ~to transitive modules pulled via `moduleDeps`: BOMs and dependency management of a module dependency have to be applied to the dependencies of modules they pull transitively (if A depends on B and B depends on C, from A, the BOMs and dep mgmt of B apply to C's dependencies too)~ (worked out-of-the-box with the previous point, via `transitiveIvyDeps`)
- via `ivy.xml`: when publishing to Ivy repositories (like during `pubishLocal`), BOMs and dep mgmt details need to be written in the `ivy.xml` file, so that they're taken into account when resolving that module from the Ivy repo
- via POM files: when publishing to Maven repositories, BOMs and dep mgmt details need to be written to POMs, so that they're taken into account when resolving that module from the Maven repo

Fixes https://github.com/com-lihaoyi/mill/issues/1975


himanshumahajan138:
Sir @alexarchambault, will this add bom support for jetpack compose libs also?

alexarchambault:
If it only relies on a "BOM", that should yes. But this PR isn't about the Gradle module metadata stuff, that was mentioned too in one of your PRs.

himanshumahajan138:
Actually i think this will work(not sure) lets see after it gets merged then i will try this once

lefou (commented):
Don't we need to take transitive `bomDeps` into account? In it's current draft, `bomDeps` is only used intransitively.

 As a rule of thumb, transitive `moduleDeps` should behave the same as transitive `ivyDeps`. If an upstream dependency import a BOM and that BOM is considered by coursier when resolving dependencies, it should also work, if that upstream dependency is a local module.

alexarchambault:
@lefou Yes, that's what I meant by "make things working for multi-module projects". In the ITs, I intend to publish locally things in a directory, and check that resolution by coursier of stuff published locally gives the same class path as Mill computes internally.

alexarchambault:
(Multi-module BOM support should be there)

alexarchambault:
@lihaoyi @lefou This is almost ready for review, you can already have a look if you wish.

There a two things I'd like to tweak though: some examples should be added in the doc, and there's an aspect of the description above that I'd like to check ("transitive modules pulled via `moduleDeps`")

This PR adds many features. I have a branch with these changes with a cleaner git history [here](https://github.com/alexarchambault/mill/tree/clean/bom-support). If needed, to ease review, I can drop some features (and send subsequent PRs later…), so these can be reviewed one-by-one.

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -140,10 +140,45 @@ trait JavaModule
...
+            .map(dep.dep.minimizedExclusions.join(_))
+            .getOrElse(dep.dep.minimizedExclusions)
+          dep.copy(
+            dep = dep.dep
+              .withVersion(versionOverride.getOrElse(dep.dep.version))
+              .withMinimizedExclusions(exclusions)
+          )
+        }
+    }
+  }
```
lefou, line 181: This large-ish task is hard to digest. Maybe, we can make it more compact and use some `def` s with meaningful names. Or add some in-line comments.

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -159,6 +194,109 @@ trait JavaModule
...
+   *   )
+   * }}}
+   */
+  def dependencyManagement: T[Agg[Dep]] = Task { Agg.empty[Dep] }
+
+  /**
+   * Data from dependencyManagement, converted to a type ready to be passed to coursier
+   * for dependency resolution
+   */
+  def dependencyManagementDict: Task[Seq[(DependencyManagement.Key, DependencyManagement.Values)]] =
```
lefou, line 254: This looks a bit like utility task we could move to `CoursierSupport` (might not work as-is due to uses of `Dep`) or `CoursierModule`.

lefou (commented):
Not a full review, but some remarks.

integration/feature/bom/resources/build.mill
```diff
@@ -0,0 +1,200 @@
...
+  def publishVersion = "0.1.0-SNAPSHOT"
+}
+
+object bom extends Module {
+  object placeholder extends JavaModule with TestPublishModule {
+    def bomDeps = Agg(
+      ivy"com.google.cloud:libraries-bom:26.50.0"
+    )
+    def ivyDeps = Agg(
+      ivy"com.google.protobuf:protobuf-java:_"
```
lihaoyi, line 25: Can we leave out the `:_` and just use `ivy"com.google.protobuf:protobuf-java"`? That would make our syntax consistent with Maven and Gradle, both of which just leave out the version segment without any placeholder

integration/feature/bom/resources/build.mill
```diff
@@ -0,0 +1,200 @@
+package build
```
lihaoyi, line 1: I don't think this needs to be an integration test. Let's cover all the edge cases in a unit test, and have a single example test under `example/fundamentals/dependencies/` as documentation for the happy paths.

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -159,6 +194,109 @@ trait JavaModule
    */
   def runIvyDeps: T[Agg[Dep]] = Task { Agg.empty[Dep] }
 
+  def parentDep: T[Option[Dep]] = Task { None }
```
lihaoyi, line 197: Let's call this `bomParentDep`? To make it clear what it's used for

scalalib/src/mill/scalalib/JavaModule.scala
lihaoyi: Maybe this logic could be moved into a helper method?

lihaoyi:
Left some comments. Some high level questions

1. How do BOMs interact with `moduleDeps`? Do they get aggregated and combined somehow, or do they need to be added to each individual module to apply? 

2. How do BOMs interact with `runClasspath` and `compileClasspath`? Do they apply universally to both/all? Would it be necessary to have a run-specific BOM and compile-specific BOM?

3. Do BOMs interact with `dependencyManagement` sections of third-party `ivyDeps`?

main/util/src/mill/util/CoursierSupport.scala
```diff
@@ -31,6 +31,20 @@ trait CoursierSupport {
         ctx.fold(cache)(c => cache.withLogger(new TickerResolutionLogger(c)))
       }
 
+  def isLocalTestDep(dep: Dependency): Option[Seq[PathRef]] = {
```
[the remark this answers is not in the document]
alexarchambault: Yes, `private` actually (with minor refactoring so that it's not used elsewhere)

integration/feature/bom/resources/build.mill
alexarchambault: Trying that

scalalib/src/mill/scalalib/JavaModule.scala
alexarchambault: That method isn't around any more now. What it used to do is now done in `JavaModule#{processDependency,processedIvyDeps}`.

alexarchambault: `processDependency` is a helper method somehow

main/util/src/mill/util/CoursierSupport.scala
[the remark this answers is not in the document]
alexarchambault: That's now private in `CoursierSupport`

scalalib/src/mill/scalalib/JavaModule.scala
alexarchambault: That method isn't around any more (more or less inlined elsewhere)

integration/feature/bom/resources/build.mill
alexarchambault: Things like `ivy"com.google.protobuf:protobuf-java"` are now accepted.

Note that missing version translates to the version field being `"_"`, which still acts as a placeholder for versions (that makes better error mssages IMO).

So we have
```scala
ivy"com.google.protobuf:protobuf-java" == ivy"com.google.protobuf:protobuf-java:_"
```

integration/feature/bom/resources/build.mill
alexarchambault: Still working on the examples / document update.

That being said, I'm not sure we need such end-to-end, or at least, not in the documentation. BOM / dependency management only impacts dependency resolution, and the unit tests assert that we end up with the right JARs in `compileClasspath` and `runClasspath` (and in the generated `ivy.xml` and POM files too).

Also, it's likely the documentation pages for Scala and Kotlin will just refer to the Java one, as there's almost specific to Scala and Kotlin in BOM / dependency management.

alexarchambault: I would add one integration test for each language just in case

scalalib/src/mill/scalalib/JavaModule.scala
alexarchambault: It's not only used as a BOM dependency: its value is used from `PublishModule`, so that it ends up advertized as "parent POM" in the POM. `parentDep` conveys that dual use better IMO.

alexarchambault:
> 2\. How do BOMs interact with `runClasspath` and `compileClasspath`? Do they apply universally to both/all? Would it be necessary to have a run-specific BOM and compile-specific BOM?

Handling this ended up being much less trivial than I thought (made me open https://github.com/com-lihaoyi/mill/pull/4028 along the way). I'm going to send changes for compile / runtime dependencies as a separate PR (that needs this one's changes).

alexarchambault:
> 3\. Do BOMs interact with `dependencyManagement` sections of third-party `ivyDeps`?

If I understand correctly your question, yes, they do. The BOM and dependency management added in a Mill module have precedence over those in their direct and transitive dependencies.

Things happen the same way in coursier during dependency resolution: if A depends on B and B depends on C, the BOMs / dependency management of A has precedence over those of B, and both those of A and B have precedence over those of C.

BOMs and dependency management basically force versions in their dependency graph.

alexarchambault:
This is still missing examples, but the rest can be reviewed if you want to (as soon as the CI is green)

scalalib/src/mill/scalalib/JavaModule.scala
alexarchambault: Actually, it became a private method (not a task anymore)

integration/feature/bom/resources/build.mill
lefou: I think `ivy"com.google.protobuf:protobuf-java:_"` could be arguably read as "any version will do", whereas the missing versions rather means, "version needs to be configured elsewhere and is not provided here".

scalalib/src/mill/scalalib/JavaModule.scala
lefou: I not convinced we need a `parentDep`. It's mapped to something Maven specific Mill doesn't even support. Can you elaborate why we need it and what it does exactly?

integration/feature/bom/resources/build.mill
lihaoyi: I think having one example test per language makes sense here, but if you want to go with a single shared example test you can put it in `fundamentals/library-deps.adoc` which contains the "shared" dependency management documentation

lihaoyi:
* What's the difference between `bomDeps` and `parentDep`?

* What happens if we have multiple `bomDeps` with conflicting dependency versions?

integration/feature/bom/resources/build.mill
alexarchambault: Support for `_` doesn't have to be advertized, it's only used internally (it's what I used in coursier for versions meant to be read in BOMs, and so that's what Mill passes to it).

If we want to use it for something else later on, that's still possible.

integration/feature/bom/resources/build.mill
alexarchambault: Just went for that

scalalib/src/mill/scalalib/JavaModule.scala
alexarchambault: I removed `parentDep`, given it doesn't bring much.

My initial idea was to complete early support for parent stuff added in https://github.com/com-lihaoyi/mill/pull/3756 (it adds `PublishModule#pomParentProject` in particular). But we might as well make Maven import add the parent to `bomDeps` alongside setting `pomParentProject`.

alexarchambault:
> * What's the difference between `bomDeps` and `parentDep`?

- `parentDep` allowed to add a single BOM, while `bomDeps` allows to add several (both `parentDep` and `bomDeps` could be used alongside each other)
- `parentDep` also used to define the default value of `PublishModule#pomParentProject` (marking the BOM as "parent" in the POM of the project)

I just removed `parentDep` given that it was confusing and didn't bring much on its own

alexarchambault:
> * What happens if we have multiple `bomDeps` with conflicting dependency versions?

For versions, the one from the first / earlist-in-the-list BOM wins.

For exclusions, all exclusions from all BOMs added to a dependency are added.

alexarchambault:
If users specify both `bomDeps` and `dependencyManagement`, versions from the latter ought to be picked, but I'd like to check that and add non-regression tests for it.

Versions from `dependencyManagement` should be picked IMO, so that things are consistent with `ivyDeps`: if a version is in `ivyDeps`, we use this one. Else the one from `dependencyManagement`. Else the one obtained via `bomDeps`. The version "specified the closest" to the `ivyDeps` wins.

scalalib/src/mill/scalalib/JavaModule.scala
lefou: Probably off-topic, about parent: I don't think we should model any parent-pom semantic. If we migrate from Maven, we should either ignore or somehow map to dependency imports, if possible. The addition of anything supporting "parent" to `PublishModule` doesn't look right to me. Since Mill does not have parents, we also don't need to generate `pom.xml` files referencing parents for publishing.

lihaoyi:
> But I just removed `parentDep` given that it was confusing and didn't bring much on its own

If this is gone, do we still support `marking the BOM as "parent" in the POM of the project`? Like by taking the first of the `bomDeps` or something? Or is it now unsupported?

example/fundamentals/library-deps/bom-1-external-bom/build.mill
```diff
@@ -0,0 +1,21 @@
...
+    ivy"io.grpc:grpc-protobuf"
+  )
+}
+
+// The version of grpc-protobuf (`io.grpc:grpc-protobuf`) isn't written down here, so the version
+// from the BOM, `1.67.1` is used.
+//
+// Also, by default, grpc-protobuf `1.67.1` pulls version `3.25.3` of protobuf-java (`com.google.protobuf:protobuf-java`) .
+// But the BOM specifies another version for that dependency, `4.28.3`, so
+// protobuf-java `4.28.3` ends up being pulled here.
```
lihaoyi, line 21: We should mention what happens if there are multiple `bomDeps` and how overlaps and resolved. You mentioned it in the PR discussion, but it should be included in the docs since people will want to know

lihaoyi: Also

> If users specify both bomDeps and dependencyManagement, versions from the latter ought to be picked, but I'd like to check that and add non-regression tests for it.

> Versions from dependencyManagement should be picked IMO, so that things are consistent with ivyDeps: if a version is in ivyDeps, we use this one. Else the one from dependencyManagement. Else the one obtained via bomDeps. The version "specified the closest" to the ivyDeps wins.

example/fundamentals/library-deps/bom-2-dependency-management/build.mill
```diff
@@ -0,0 +1,54 @@
+// Pass dependencies to `dependencyManagement` in a `JavaModule` / `ScalaModule` / `KotlinModule`, like
+
+//// SNIPPET:BUILD1
+package build
+import mill._, javalib._
+
+object dependencyManagement extends JavaModule {
```
lihaoyi, line 7: Let's rename this `foo` or something, otherwise the name is confusing when duplicated below in `def dependencyManagement`. The code and english explanation should be enough to describe what the example is about, and we can add a `== subheading` if necessary

example/fundamentals/library-deps/bom-1-external-bom/build.mill
lihaoyi: > For exclusions, all exclusions from all BOMs added to a dependency are added.

lihaoyi:
Some minor comments. I didn't review the code changes in detail, but I expect the existing and new test suites should give us confidence that things work

lihaoyi:
@lefou I'm planning on merging this once the next round of review updates is done, so take a look if you have time

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -159,6 +159,139 @@ trait JavaModule
...
+   * For example, the following forces com.lihaoyi::os-lib to version 0.11.3, and
+   * excludes org.slf4j:slf4j-api from com.lihaoyi::cask that it forces to version 0.9.4
+   * {{{
+   *   def dependencyManagement = super.dependencyManagement() ++ Agg(
+   *     ivy"com.lihaoyi::os-lib:0.11.3",
+   *     ivy"com.lihaoyi::cask:0.9.4".exclude("org.slf4j", "slf4j-api")
+   *   )
+   * }}}
+   */
+  def dependencyManagement: T[Agg[Dep]] = Task { Agg.empty[Dep] }
```
lihaoyi, line 210: Should we name this `def depManagement`? We use `dep`/`Dep` throughout the rest of the codebase, so this would make the API a bit more consistent

integration/feature/bom/resources/build.mill
lefou: I think we need to make sure the `_` used in coursier does not leak into Mill, which is probably not possible, since we forward coursier output in error messages and `ivyDepsTree` to name some. Out of curiosity, were there any limitations forcing you to choose `_` as a marker? Would it be possible to chnage it in coursier?

alexarchambault: I guess it could still be changed, but for what? I used a marker so that the dependency parsers in coursier don't need much change (compared to accepting things like `org:name`)

scalalib/src/mill/scalalib/JavaModule.scala
alexarchambault: I agree that there should be no need to model parent POM semantics in Mill. I don't know what motivated it in https://github.com/com-lihaoyi/mill/pull/3756

integration/feature/bom/resources/build.mill
lefou: For the user experience? For the consistency with Maven and Gradle?

alexarchambault:
> > But I just removed `parentDep` given that it was confusing and didn't bring much on its own
> 
> If this is gone, do we still support `marking the BOM as "parent" in the POM of the project`? Like by taking the first of the `bomDeps` or something? Or is it now unsupported?

That's unsupported. Users have to explicitly override `PublishModule#pomParentProject` themselves, with the right BOM coordinates.

integration/feature/bom/resources/build.mill
lefou: And for avoidance of extra work in Mill, since we already use `_` as a ubiquitous wildcard, not as a marker for absence. We only have the choise to
1. convert all dependency representation coming from coursier, which is tedious and leads to a lot of boilerplate code
2. use `_` and off-load the inconsistence to the user
3. ignore and make an inconsistent UX

lefou:
> > > But I just removed `parentDep` given that it was confusing and didn't bring much on its own
> > 
> > 
> > If this is gone, do we still support `marking the BOM as "parent" in the POM of the project`? Like by taking the first of the `bomDeps` or something? Or is it now unsupported?
> 
> That's unsupported. Users have to explicitly override `PublishModule#pomParentProject` themselves, with the right BOM coordinates.

I don't think there is any (valid) use case for that. The only I could imagine is a user trying to produce `pom`s with Mill (as it's easy) without actually building the artifacts with Mill.  Or some strange corporate setup, where using a specific parent pom is mandatory. But in such a scenario, Mill is most likely not allowed anyways. It would lead to inconsistencies in Mill vs. Maven usage too, if the parent pom has any significant content.

example/fundamentals/library-deps/bom-1-external-bom/build.mill
alexarchambault: > For exclusions, all exclusions from all BOMs added to a dependency are added.

Adding that

alexarchambault: > Also we should include
> 
> > If users specify both bomDeps and dependencyManagement, versions from the latter ought to be picked, but I'd like to check that and add non-regression tests for it.
> 
> > Versions from dependencyManagement should be picked IMO, so that things are consistent with ivyDeps: if a version is in ivyDeps, we use this one. Else the one from dependencyManagement. Else the one obtained via bomDeps. The version "specified the closest" to the ivyDeps wins.

For those, I'm going to add them in a subsequent PR. I wrote tests locally to assert that, and they don't pass (yet). That might need changes in coursier. I'd merge the PR here and leave that behavior unspecified for a (short) amount of time.

scalalib/src/mill/scalalib/JavaModule.scala
alexarchambault: `depMgmt` could work too

alexarchambault:
Not really sure what the exact use case would be, TBH

scalalib/src/mill/scalalib/JavaModule.scala
alexarchambault: Went for `depManagement` for now

lihaoyi:
This looks great, thanks @alexarchambault! Looking forward to the follow ups. I also chatted with the Bazel folks and they seem pretty interested in making use of the new BOM support in Coursier!

integration/feature/bom/resources/build.mill
lefou: So how we want to proceed now? Since `Dep` does not fully abstract over the inner `dep: Dependency`, the version is often used as is, and the forwarder `Dep.verison` wasn't touched in this PR, so we end up with a mix-up of defined or empty or `_` versions.

lefou:
> @lefou I'm planning on merging this once the next round of review updates is done, so take a look if you have time

So, I did take a look, I also commented and there were open unresolved remarks but this was merged anyways. I don't think we should compromise consistency for quickly merging features. It looks like the whole Scala build tool community seem to be interested in BOM support, so we should also try to find a nice user experience. It's technically complicated enough. The inconsistent empty version vs. `_` wildcard version isn't great to start with.

lihaoyi:
I thought the resolution was that `_` wildcard was just going to be an implementation detail and not a supported feature? That was my understanding but apologies if I'm mistaken.

lefou:
It's used in coursier `Dependency` class, which is a fully leaking abstraction. So Mill users see in all over the places. See this comment: https://github.com/com-lihaoyi/mill/pull/3924#discussion_r1869211040

lihaoyi:
@alexarchambault how hard would it be to tweak the coursier parser upstream to allow `foo:bar` without the trailing `:_`? Seems to me like it shouldn't be too much of a burden right?

alexarchambault:
Having a look, to me the main point of `_` as a version was to mark missing versions in the inputs of coursier. While empty versions originate from versions not written down in POMs (whose value should be somewhere in a BOM). So having a `_` in error messages means something went wrong in the upfront processing of inputs, while an empty version means something went wrong in the resolver internals.

Admittedly, the `_` can always be dropped… It shouldn't be that much of a big deal to tweak the coursier parser, I'm having a look

Issue #3868, opened 2024-10-29
Follow up Android Support Bounties (3000USD)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 3000USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

This is a follow up issue to https://github.com/com-lihaoyi/mill/issues/3550, to further flesh out Android support in the Mill build tool. The goal is to cover a variety of use cases that are common in Android development, beyond simple hello-world applications

* [x] Example of [App Bundles](https://developer.android.com/guide/app-bundle) (500USD)
* [ ] Example of [Build Variants](https://developer.android.com/build/build-variants), with flavor/variant-specific sources (500USD)
* [ ] Example of mixed NDK/Java/Kotlin android app (500USD)
* [x] Example of using [Android Linting](https://developer.android.com/studio/write/lint) (500USD)
* [x] Example using [Instrumented tests](https://developer.android.com/training/testing/instrumented-tests) on top of normal unit tests (500USD)
* [ ] Example using the [Hilt dependency injection framework](https://developer.android.com/training/dependency-injection/hilt-android) (500USD)

himanshumahajan138:
Its Been a week since this issue raised,
Now i think i can work around without any conflict 

PS: Working on App Bundle Fixture...

lihaoyi:
Reopening this since the linked PR only fixes the first bullet

himanshumahajan138:
Working on Build Variants (if any conflict then please let me know)

0xnm:
Linking the comment in the PR description which may be useful for the future work: https://github.com/com-lihaoyi/mill/pull/4188#issue-2761656394

0xnm:
@himanshumahajan138 Implementing build variants before Android lib modules support is implemented is not quite handy, because if there is app module declaring variants + lib module declaring variants and app module is using lib module, then there comes a question of variants matching between the two, which will impact the overall API. Also it is the most complex task from the list above because of the dynamic nature (at least in AGP) of the variants declaration rather than the static one.

PR #3931, opened 2024-11-10
Add Android Linting Example

c0d33ngr:
This is meant to to be a draft of item number 4 of issue #3868. The android linting

Hi @lihaoyi is this close to what you have in mind about the android linting or there's more?


lihaoyi:
@c0d33ngr looks good, need to include an example test and explanation so we can include it in the documentation

himanshumahajan138:
@c0d33ngr Buddy you did just basic linting but this is not the appropriate way tbh i have tried this on my example and its saying the miss of class file but class files only come after the generation

so @lihaoyi i request you to please wait till my new pull request(anndroid bundle support) gets merged coz in that i have updated the whole android app building process and made it professional (now we can include custom resources)

so @c0d33ngr sorry to interrupt you, but please understand...

lihaoyi:
@himanshumahajan138 completing bounties out of order is fine. I'll look at your PR if you wish and we can decide which one to merge first

himanshumahajan138:
> @himanshumahajan138 completing bounties out of order is fine. I'll look at your PR if you wish and we can decide which one to merge first

Ya I Agree We can decide which one to merge first 

i was saying before becoz there will be need for updated code as you can see in my pr i have updated the codes with resources 

so that's the only thing i was concerned about rest we can decide this...

lihaoyi:
@c0d33ngr the lint error should be fixed in latest main branch, if you merge in main it should work.

Give me some time to do a proper review, I'm not familiar with android haha

lihaoyi:
FYI I'm planning on landing https://github.com/com-lihaoyi/mill/pull/3935 first before coming back to this, since that one is a larger and more invasive PR so this one would be easier to rebase/merge after

scalalib/src/mill/javalib/android/AndroidLintModule.scala
```diff
@@ -0,0 +1,82 @@
...
+import mill.javalib.android.AndroidSdkModule
+
+/**
+ * Android Lint Module for integrating the Android Lint tool in a Mill build.
+ *
+ * This module provides configuration options for executing Android Lint, including setting paths,
+ * specifying lint rules, managing reports, and more.
+ */
+@mill.api.experimental
+trait AndroidLintModule extends AndroidSdkModule with JavaModule {
```
lihaoyi, line 15: Why are we inheriting from both `AndroidSdkModule` and `JavaModule`? Those are typically not mixed in together, so there's something weird here

example/javalib/android/2-linting/build.mill
```diff
@@ -0,0 +1,29 @@
...
+  def androidSdkModule = mill.define.ModuleRef(androidSdkModule0)
+}
+
+/** See Also: app/AndroidManifest.xml */
+/** See Also: app/src/main/java/com/helloworld/app/MainActivity.java */
+
+/** Usage
+
+> ./mill show app.androidLint  # Display full path to the linting report in HTML
+".../out/app/androidLint.dest/report.html"
```
lihaoyi, line 22: We should use `cat` on the HTML report to ensure certain lint errors get picked up and properly reported

example/javalib/android/2-linting/build.mill
```diff
@@ -0,0 +1,29 @@
...
+import mill.javalib.android.{AndroidSdkModule, AndroidLintModule}
+
+// Create and configure an Android SDK module to manage Android SDK paths and tools.
+object androidSdkModule0 extends AndroidSdkModule {
+  def buildToolsVersion = "35.0.0"
+}
+
+// Actual android application
+object app extends AndroidLintModule {
+  def androidSdkModule = mill.define.ModuleRef(androidSdkModule0)
```
lihaoyi, line 13: Let's include an example custom lint configuration here so people know what it looks like

scalalib/src/mill/javalib/android/AndroidLintModule.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * Path to the project for which lint should run.
+   */
+  def projectPath: T[PathRef] = Task.Source { millSourcePath }
```
lihaoyi, line 20: Do we need this? Or can we just use `millSourcePath` everywhere instead?

scalalib/src/mill/javalib/android/AndroidLintModule.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * Specifies the file format of lint report.
+   */
+  def lintReportFormat: T[String] = Task { "html" }
```
lihaoyi, line 25: The comment should list the possible formats we are aware of

scalalib/src/mill/javalib/android/AndroidLintModule.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * Enable or disable warnings in the lint report.
+   */
+  def warningsAsErrors: T[Boolean] = Task { false }
+
+  /**
+   * Additional options for lint (e.g., enabling/disabling specific checks).
+   */
+  def lintOptions: T[Seq[String]] = Task { Seq("--check", "NewApi,InlinedApi") }
```
lihaoyi, line 40: All tasks in this module should be prefixed by `androidLint` to make it clear where they come from and avoid collisions with tasks coming from other traits that get mixed in

lihaoyi:
Took a review pass. Looks like a straightforward change but left a few nits

docs/modules/ROOT/nav.adoc
```diff
@@ -11,6 +11,8 @@
 ** xref:javalib/publishing.adoc[]
 ** xref:javalib/build-examples.adoc[]
 ** xref:javalib/web-examples.adoc[]
+** xref:javalib/android-examples.adoc[]
+** xref:javalib/android-linting.adoc[]
```
lihaoyi, line 15: This is probably a merge conflict; we moved the stuff into the `android/` subfolder

example/javalib/android/2-linting/1-basic-linting/build.mill
```diff
@@ -0,0 +1,29 @@
...
+  def androidSdkModule = mill.define.ModuleRef(androidSdkModule0)
+}
+
+/** See Also: app/AndroidManifest.xml */
+/** See Also: app/src/main/java/com/helloworld/app/MainActivity.java */
+
+/** Usage
+
+> ./mill show app.androidLintRun  # Display full path to the linting report in HTML
+".../out/app/androidLintRun.dest/report.html"
```
lihaoyi, line 22: Let's `> cat out/app/androidLintRun.dest/report.html` and match some strings that we expect to be inside that report. I assume your examples contain some lint warnings, so we should be able to match them as part of the `cat` output

lihaoyi: We should also use `sed` to fix the lint warnings and run lint again to verify that no more warnings exist. This applies to the `2-custom-linting` setup as well

scalalib/src/mill/javalib/android/AndroidLintModule.scala
```diff
@@ -0,0 +1,127 @@
...
+import mill.scalalib.JavaModule
+
+/**
+ * Android Lint Module for integrating the Android Lint tool in a Mill build.
+ *
+ * This module provides configuration options for executing Android Lint, including setting paths,
+ * specifying lint rules, managing reports, and more.
+ */
+@mill.api.experimental
+trait AndroidLintModule extends JavaModule {
```
lihaoyi, line 15: This should probably inherit from `AndroidAppModule` and re-use the tasks from there, that should cut down on some of the tasks we need to define ourselves

example/android/javalib/3-linting/1-basic-linting/build.mill
```diff
@@ -0,0 +1,56 @@
...
+// the project is for demostration. There are no icons for the demo project.
+
+
+/** Usage
+
+> ./mill show app.androidLintRun  # Display full path to the linting report in HTML
+".../out/app/androidLintRun.dest/report.html"
+
+> cat out/app/androidLintRun.dest/report.html" # Display content of the linting report
+Avoid hardcoding the debug mode; leaving it out allows debug and release builds to automatically assign one
```
lihaoyi, line 42: You probably need to use `...` wildcards here, unless the HTML file really only contains this string and nothing else. See https://mill-build.org/mill/extending/writing-plugins.html#_example_tests

example/android/javalib/3-linting/1-basic-linting/build.mill
```diff
@@ -0,0 +1,56 @@
...
+
+> sed -i 's/ android:debuggable="true"//g' app/AndroidManifest.xml # Fix the HardcodedDebugMode warning issue from `AndroidManifest.xml`
```
lihaoyi, line 44: This needs to be `sed -i.bak` for cross-platform support, see elsewhere in the codebase

himanshumahajan138:
@c0d33ngr and @lihaoyi, 
We don't need android:debuggable parameter in the manifest you can remove this its intended only for developers to test the application using usb or wireless debugging.

I was using this only for my testing purposes nothing else...

PS: we dont need android debuggable parameter

himanshumahajan138:
> > @c0d33ngr and @lihaoyi, We don't need android:debuggable parameter in the manifest you can remove this its intended only for developers to test the application using usb or wireless debugging.
> > I was using this only for my testing purposes nothing else...
> > PS: we dont need android debuggable parameter
> 
> Yeah... It isn't needed as it's hardcoded which will make android lint cli tool unhappy. It seems you forgot to remove it in your PR
> 
> I'm using it to demo how the linting find and report issues.

Ya remove this from other examples also 🙃

himanshumahajan138:
> Is it in other examples too? @himanshumahajan138 I didn't know
> 
> I like to keep my PRs within the scope of any opened issues I intend to close. In this case, addition of Android linting examples

No worries i will update later on 👍

himanshumahajan138:
Actually i have spent more than 3 months now while contributing and learning mill and want to continue same in future so apart from bounties i take my contributions as a responsibility to enhance and groom mill to top-level 

And this community is filled with supporting members who provide great support and learning edge that's it 😊...

0xnm:
Side note for some improvement in the future PRs: some properties shouldn't be in `AndroidManifest`, even though they are allowed there by the namespace. Instead, things like `debuggable`, `minSdk`, etc. should be a part of the build configuration and passed to `aapt2` instead which will generate _final_ manifest, see [aapt2 link options](https://developer.android.com/tools/aapt2#link_options) for more details.

himanshumahajan138:
@c0d33ngr buddy what command you use in local can you please tell me.

Some commands that i use: 

- `./mill mill.scalalib.scalafmt.ScalafmtModule/`
- `./mill mill.javalib.palantirformat.PalantirFormatModule/`
- `./mill mill.kotlinlib.ktlint.KtlintModule/`
- `./mill -i docs.fastPages + docs.checkBrokenLinks`

himanshumahajan138:
> > @c0d33ngr buddy what command you use in local can you please tell me.
> > Some commands that i use:
> > ```
> > * `./mill mill.scalalib.scalafmt.ScalafmtModule/`
> > 
> > * `./mill mill.javalib.palantirformat.PalantirFormatModule/`
> > 
> > * `./mill mill.kotlinlib.ktlint.KtlintModule/`
> > 
> > * `./mill -i docs.fastPages + docs.checkBrokenLinks`
> > ```
> 
> Same `./mill -i mill.scalalib.scalafmt.ScalafmtModule/checkFormatAll __.sources + __.mimaReportBinaryIssues + __.fix --check + mill.javalib.palantirformat.PalantirFormatModule/ --check + mill.kotlinlib.ktlint.KtlintModule/checkFormatAll --check`
> 
> From the github action file

If you are running the full command then it will not work it's for testing not fixing 

Run those that i gave

himanshumahajan138:
@c0d33ngr if you want to include `3-linting` as a folder then add it in folder android as `example/android/linting` and then add two examples `1-basic-linting` and `2-custom-linting` and inside these two folders your build file should be present.

And if you want to add `linting` inside particular folders like inside `javalib` or `kotlinlib` then their respective folders will be available under them but these will be `3-basic-linting` and `4-custom-linting`.


Buddy please first understand the mill structure and internal building after that please do changes coz learning is the only key to success in this community 🤞

Hope you understand!

example/android/javalib/3-linting/build.mill
```diff
@@ -0,0 +1,84 @@
...
+  def buildToolsVersion = "35.0.0"
+  def bundleToolVersion = "1.17.2"
+}
+
+// Actual android application
+object app extends AndroidLintModule {
+  def androidSdkModule = mill.define.ModuleRef(androidSdkModule0)
+
+  // Set path to the location of the project source codes
+  def androidLintSources = Task { Some(PathRef(millSourcePath / "src" / "main" / "java")) }
```
lihaoyi, line 27: We should not need to specify this, we can just rely on `sources()`

scalalib/src/mill/javalib/android/AndroidLintModule.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  /**
+   * Specifies the resources directories argument for the Android Lint tool.
+   */
+  def androidLintResources: T[Option[PathRef]] = Task { None }
+
+  /**
+   * Specifies the sources directories argument for the Android Lint tool.
+   */
+  def androidLintSources: T[Option[PathRef]] = Task { None }
```
lihaoyi, line 44: Remove

scalalib/src/mill/javalib/android/AndroidLintModule.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  /**
+   * Specifies the libraries argument for the Android Lint tool.
+   */
+  def androidLintLibraries: T[Option[PathRef]] = Task { None }
```
lihaoyi, line 49: Let's replace this and all the other flags with just `def androidLintArgs: T[Seq[String]] = Nil`

example/android/javalib/3-linting/build.mill
```diff
@@ -0,0 +1,84 @@
...
+
+> ./mill show app.androidLintRun  # Display full path to the linting report in text file
+".../out/app/androidLintRun.dest/report.txt"
+
+> cat out/app/androidLintRun.dest/report.txt # Display content of the linting report
+AndroidManifest.xml:3: Error: Avoid hardcoding the debug mode; leaving it out allows debug and release builds to automatically assign one [HardcodedDebugMode]
+
+> sed -i.bak 's/ android:debuggable="true"//g' app/AndroidManifest.xml # Fix the HardcodedDebugMode warning issue from `AndroidManifest.xml`
+
+> ./mill clean && ./mill app.androidLintRun # Clean the cache and rerun the android lint
```
lihaoyi, line 63: Why do we need to `clean` here?

lihaoyi: Where does this cache live? Mill tasks should generally put any caches into their own `.dest` folder, so that they get automatically invalidated where necessary and do not need `clean` to work

lihaoyi:
@c0d33ngr what you suggest for flags sounds reasonable to me

lihaoyi:
We should not need to use `clean` at all

himanshumahajan138:
If i am not wrong, when we are creating linting example then we should also take care of linting the android specific files (like xml, proguard and more are there) i think we should also perform linting just like we do for java or kt files 

please ignore if this support is already added (i didn't tested)

lihaoyi:
@c0d33ngr you'll have to figure it out. I'm not an expert in the Android toolchain and I won't have the ability to become one on short notice, so you'll need to learn enough about the toolchian figure out how to fit it into the Mill build convention

scalalib/src/mill/javalib/android/AndroidLintModule.scala
```diff
@@ -0,0 +1,78 @@
...
+   * to analyze the source code for potential bugs, performance issues, and
+   * best practices compliance. It generates a report in the specified format.
+   *
+   * The lint tool requires the Android SDK's command-line tools to be installed.
+   * The report is saved in the task's destination directory as "report.html".
+   *
+   * For more details on the Android Lint tool, refer to:
+   * [[https://developer.android.com/studio/write/lint]]
+   */
+  def androidLintRun: T[PathRef] = Task.Input {
```
lihaoyi, line 47: Let's make this a `Task.Command`, for consistency with other linters which are all commands.

example/android/javalib/3-linting/build.mill
```diff
@@ -0,0 +1,95 @@
...
+  def androidSdkModule = mill.define.ModuleRef(androidSdkModule0)
+
+  // Set path to the custom `lint.xml` config file. It is usually at the root of the project
+  def androidLintConfigPath = Task { Some(PathRef(millSourcePath / "lint.xml")) }
+
+  // Set the linting report to be generated as a text file
+  def androidLintReportFmt = Task { "txt" }
+
+  // Set path to generated `.jar` files in this case (or `.class` files)
+  def classPath = Task.Source { androidJar() }
```
0xnm, line 33: what is the need to dedicated `classPath` which adds only a confusion, because there is already `compileClasspath` and `runClasspath` available and which already include everything what is needed to build the sources?

example/android/javalib/3-linting/build.mill
```diff
@@ -0,0 +1,95 @@
...
+
+  // Set path to the location of the project source codes
+  def sourcesPath = Task.Source { (millSourcePath / "src" / "main" / "java") }
```
0xnm, line 36: same here - why not reusing `sources`?

example/android/javalib/3-linting/build.mill
```diff
@@ -0,0 +1,95 @@
...
+
+// Actual android application with linting config
+object app extends AndroidLintModule {
+  def androidSdkModule = mill.define.ModuleRef(androidSdkModule0)
+
+  // Set path to the custom `lint.xml` config file. It is usually at the root of the project
+  def androidLintConfigPath = Task { Some(PathRef(millSourcePath / "lint.xml")) }
+
+  // Set the linting report to be generated as a text file
+  def androidLintReportFmt = Task { "txt" }
```
0xnm, line 30: lint tool has only 4 formats: xml, txt, sarif and html and I don't think anything else will be added soon. So, probably, it makes sense to use enum here instead of just string, to minimize the chance of mistake from the user side.

Also, AGP has a different way to configure it: it provides boolean properties allowing enable/disable the certain format, this allows to output multiple formats at the same time, while with such API module is limited only to the single output format.

This is not something critical, but quite handy.

scalalib/src/mill/javalib/android/AndroidLintModule.scala
```diff
@@ -0,0 +1,78 @@
...
+import mill.javalib.android.AndroidAppModule
+
+/**
+ * Android Lint Module for integrating the Android Lint tool in a Mill build.
+ *
+ * This module provides configuration options for executing Android Lint, including setting paths,
+ * specifying lint rules, managing reports, and more.
+ */
+@mill.api.experimental
+trait AndroidLintModule extends AndroidAppModule {
```
0xnm, line 14: what is the added value to have it as a dedicated trait and not a part of the `AndroidAppModule`?

scalalib/src/mill/javalib/android/AndroidLintModule.scala
```diff
@@ -0,0 +1,78 @@
...
+   * Specifies the file format of lint report. Available file formats are "html", "xml", "txt" and "sarif".
+   * Specifies additional arguments for the Android Lint tool.
+   * Allows for complete customization of the lint command.
+   */
+  def androidLintReportFmt: T[String] = Task { "html" }
+
+  /**
+   * Specifies the lint configuration XML file path. This allows setting custom lint rules or modifying existing ones.
+   */
+  def androidLintConfigPath: T[Option[PathRef]] = Task { None }
```
0xnm, line 26: there is no API to enable/suppress certain checks without using `lint.xml`, also this trait is missing baseline file API support. You can check this page for the configuration options https://developer.android.com/studio/write/lint#gradle

lihaoyi:
@c0d33ngr `runClasspath()` returns a `Seq[PathRef]` IIRC, so you need to `.map(_.path)` to turn it into raw paths suitable for external programs. Same with `sources()`

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -62,6 +88,32 @@ trait AndroidAppModule extends JavaModule {
    */
   def androidReleaseKeyPath: T[PathRef] = Task.Source(millSourcePath / androidReleaseKeyName())
 
+  /**
+   * Specifies the file format of lint report. Available file formats are "html", "xml", "txt" and "sarif".
```
0xnm, line 92: nit: probably you would like to reference `AndroidLintReportFormat` values here instead of string literals

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -62,6 +88,32 @@ trait AndroidAppModule extends JavaModule {
...
+
+  /**
+   * Specifies the lint baseline XML file path. This allows using a baseline to suppress known lint warnings.
+   */
+  def androidLintBaselinePath: T[Option[PathRef]] = Task { None }
+
+  /**
+   * Determines whether the build should fail if Android Lint detects any issues.
+   */
+  def abortOnError: Boolean = true
```
0xnm, line 109: should it be `androidLintAbortOnError` for the consistency with other API methods?

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -339,4 +391,47 @@ trait AndroidAppModule extends JavaModule {
...
+    ).getOrElse(Seq.empty)
+
+    // Prepare the lint baseline argument if the baseline path is set
+    val baselineArg = androidLintBaselinePath().map(baseline =>
+      Seq("--baseline", baseline.path.toString)
+    ).getOrElse(Seq.empty)
+
+    os.call(
+      Seq(
+        androidSdkModule().cmdlineToolsPath().path.toString + "/lint",
```
0xnm, line 426: I think you can directly define `lintToolPath` in `AndroidSdkModule`

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -339,4 +391,47 @@ trait AndroidAppModule extends JavaModule {
...
+
+  /**
+   * Runs the Android Lint tool to generate a report on code quality issues.
+   *
+   * This method utilizes Android Lint, a tool provided by the Android SDK,
+   * to analyze the source code for potential bugs, performance issues, and
+   * best practices compliance. It generates a report in the specified format.
+   *
+   * The lint tool requires the Android SDK's command-line tools to be installed.
+   * The report is saved in the task's destination directory as "report.html".
```
0xnm, line 403: It can be `report.xml`, etc. Depends on the format chosen.

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -339,4 +391,47 @@ trait AndroidAppModule extends JavaModule {
...
   }
+
+  /**
+   * Runs the Android Lint tool to generate a report on code quality issues.
+   *
+   * This method utilizes Android Lint, a tool provided by the Android SDK,
+   * to analyze the source code for potential bugs, performance issues, and
+   * best practices compliance. It generates a report in the specified format.
+   *
+   * The lint tool requires the Android SDK's command-line tools to be installed.
```
0xnm, line 402: probably this line can be deleted, not sure if it is useful.

example/android/javalib/3-linting/build.mill
```diff
@@ -0,0 +1,108 @@
...
+*/
+
+// == Linting with Custom Configurations
+
+// The commands below set the serverity level of `MissingApplicationIcon` to `warning` in the `lint.xml` file that was shown above
+// and run the `androidLintRun` task. Some output of the changes is shown below.
+
+/** Usage
+
+> sed -i.bak 's/severity="ignore"/severity="warning"/g' app/lint.xml # Set `MissingApplicationIcon` severity level to warning
```
0xnm, line 101: I think this example is also missing the usage of the `SuppressLint` annotation in the code (see https://developer.android.com/studio/write/lint#configure-lint-checking-in-java-or-kotlin), so it is not clear if it is working or not.

Same for suppression in XML https://developer.android.com/studio/write/lint#configure-lint-checking-in-xml

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -62,6 +88,32 @@ trait AndroidAppModule extends JavaModule {
    */
   def androidReleaseKeyPath: T[PathRef] = Task.Source(millSourcePath / androidReleaseKeyName())
 
+  /**
+   * Specifies the file format of lint report. Available file formats are "html", "xml", "txt" and "sarif".
+   */
+  def androidLintReportFmt: T[AndroidLintReportFormat.Value] = Task { AndroidLintReportFormat.Html }
```
0xnm, line 94: you can relatively easy support generating report in the multiple formats at the same time if this configuration option will be `T[Set[AndroidLintReportFormat.Value]]`

In case if it is helpful: tool source code is here https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:lint/cli/src/main/java/com/android/tools/lint/

example/android/javalib/3-linting/build.mill
```diff
@@ -0,0 +1,108 @@
...
+    val cp = runClasspath().map(_.path).filter(os.exists).mkString(":")
+
+    // Set path to the location of the project source codes
+    val src = sources().map(_.path).filter(os.exists).mkString(":")
+    Seq(
+      "--classpath",
+      cp,
+      "--sources",
+      src
+    )
```
0xnm, line 51: I think this should be moved to the `AndroidAppModule`, this will be the same for 99% of setups and end-user should have everything working out of the box with the reasonable defaults.

0xnm:
I suggest you to check the arguments you are calling `lint` tool with and verify that your `MainActivity.java` file path is in the list of the sources passed.

docs/modules/ROOT/nav.adoc
```diff
@@ -11,7 +11,6 @@
 ** xref:javalib/publishing.adoc[]
 ** xref:javalib/build-examples.adoc[]
 ** xref:javalib/web-examples.adoc[]
-
```
0xnm, line 14: probably a change to rollback? seems like this was a separator

example/android/javalib/3-linting/build.mill
```diff
@@ -0,0 +1,143 @@
...
+// The `lint.xml` file shown above ignores the `MissingApplicationIcon` error in the code as
+// the project is for demostration. There are no icons for the demo project.
+
+////SNIPPET:END
+
+/** Usage
+
+> ./mill app.androidLintRun
+
+> ls out/app/androidLintRun.dest  # Display full path to the linting report in text file
```
0xnm, line 57: nit: it is not only txt file, but also html

example/android/javalib/3-linting/build.mill
```diff
@@ -0,0 +1,143 @@
+// == Linting with Basic Config
+
+// This Mill build configuration includes a linting step, which is essential for ensuring code
+// quality and adhering to best practices in Android development projects. Running the `androidLintRun` task
+// produces a detailed HTML report by default, identifying potential issues in the code, such as performance,
+// security, and usability warnings. This helps maintain the health and quality of the codebase.
+
+// In this example, we set a custom config to generate the report file in txt format and not HTML.
```
0xnm, line 8: actually HTML is generated as well

example/android/javalib/3-linting/build.mill
```diff
@@ -0,0 +1,143 @@
...
+
+> cat out/app/androidLintRun.dest/report.txt # Check the content of report again
+lint-baseline.xml: Information: 1 errors/warnings were listed in the baseline file (lint-baseline.xml) but not found in the project; perhaps they have been fixed? Unmatched issue types: HardcodedDebugMode [LintBaseline]
+
+*/
+
+// == Linting with Custom Configurations
+
+// The commands below are demos on using `androidLintRun` task with some custom configuratins on a sample project.
+// The demos are adjusting severity lelvel of linting issue, enable and disable linting issue
```
0xnm, line 83:
```suggestion
// The demos are adjusting severity level of linting issue, enable and disable linting issue
```

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -62,6 +88,34 @@ trait AndroidAppModule extends JavaModule {
    */
   def androidReleaseKeyPath: T[PathRef] = Task.Source(millSourcePath / androidReleaseKeyName())
 
+  /**
+   * Specifies the file format(s) of the lint report. Available file formats are defined in AndroidLintReportFormat,
+   * such as AndroidLintReportFormat.Html, AndroidLintReportFormat.Xml, AndroidLintReportFormat.Txt, and AndroidLintReportFormat.Sarif.
+   */
+  def androidLintReportFmt: T[Seq[AndroidLintReportFormat.Value]] =
```
0xnm, line 95: I'm not sure if `fmt` is used in this codebase as a word, it seems it is rather `format` which is widely used.

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -339,4 +393,63 @@ trait AndroidAppModule extends JavaModule {
...
+    // Generate the alternating flag and file path strings
+    val reportArg: Seq[String] = formats.toSeq.flatMap { format =>
+      Seq(format.flag, (Task.dest / s"report.${format.extension}").toString)
+    }
+
+    // Set path to generated `.jar` files and/or `.class` files
+    val cp = runClasspath().map(_.path).filter(os.exists).mkString(":")
+
+    // Set path to the location of the project source codes
+    val src = sources().map(_.path / "main" / "java").filter(os.exists).mkString(":")
```
0xnm, line 423: I think you shouldn't hardcode path as `main/java`, because both pieces can be actually different. I would suggest to use `Lib.findSourceFiles` (extension can be both `.java` and `.kotlin`) at least with the given `sources`.

Ideally, in order to not have a huge path with each individual files there, it should be a way to find a common roots for the given set of the files, but I don't know if there are utility methods in the project's codebase to compute that.

0xnm (commented):
It looks reasonable to me. This first drop of lint support doesn't obviously support all the [configuration](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/gradle-api/src/main/java/com/android/build/api/dsl/Lint.kt) options available in AGP and [options](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:lint/cli/src/main/java/com/android/tools/lint/LintCliFlags.java) supported by the tool, but it is a reasonable basis for doing a simple lint run.

One more handy thing to have is maybe to provide module-independent runner/report collector (see existing verification tooling modules in the codebase, they have static methods with `Evaluator`), but it is not a blocker for me.

I left some comments, for the rest I think @lihaoyi can do a review.

scalalib/src/mill/javalib/android/AndroidAppModule.scala
0xnm: actual files can be under any path of `sources()` root, so hardcoding `main/java` doesn't work: nothing prevents having source files under `sources()/foo/bar/foobar`, for example.

0xnm (approved):

lihaoyi:
This looks great, I'll merge it and fix the checkBrokenLinks error myself. @c0d33ngr I'll close out the bounty using the bank information you sent me earlier

0xnm:
No, `.bak` stands for backup and it arbitrary file type which can be applied to any file, so `aapt2` won't pick it, obviously. It expects the original file ([link](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/tools/aapt2/cmd/Compile.cpp;l=709-729;drc=211bec2871da597f9f3fd81df7faffea1754437e)).

PR #3935, opened 2024-11-10
Fixes:#3868 Added Android Bundle Support

himanshumahajan138:
# Pull Request 
Added Android Bundle Support (with alot of updation in Android App Build Process)
Fixes: #3868

## Description
The new `AndroidAppBundle` module adds support for Android app bundles, extending `AndroidAppModule`. An example of the bundle process is now available in `JavaLib`. `AndroidAppModule` and `AndroidSdkModule` have been updated to support Android resources and the `aapt2` tool. Additionally, resource support has been added to the `JavaLib` and `KotlinLib` examples.

## Related Issues
- Link to related issue #3868.

## Checklist
- [x] New Module AndroidAppBundle (for adding android bundle support extends AndroidAppModule)
- [x] New Example for Android Bundle Process (Currently under JavaLib)
- [x] Updated AndroidAppModule(added resource support)
- [x] Updated AndroidSdkModule(related to Android aapt2 Tool)
- [x] Updated JavaLib and KotlinLib Android Examples (added resources support)

## Further Addition
Requires @lihaoyi Review for code satisfaction and then further move to Documentation Part(for Build.mill files(updations))


himanshumahajan138:
@lihaoyi Java Files are getting linting error is there some command to fix it like the way you gave last time for scala formatting

himanshumahajan138:
> > @lihaoyi Java Files are getting linting error is there some command to fix it like the way you gave last time for scala formatting
> 
> `./mill mill.javalib.palantirformat.PalantirFormatModule/` I think this should do. I got it from your CI test errors

Solved Already...
Thanks for Help

himanshumahajan138:
> > @lihaoyi Java Files are getting linting error is there some command to fix it like the way you gave last time for scala formatting
> 
> `./mill mill.javalib.palantirformat.PalantirFormatModule/` I think this should do. I got it from your CI test errors

Actually its still not solved 😅

himanshumahajan138:
@c0d33ngr buddy i don't know why extra ordinary things happen to me in this world 😂

can you tell me why my test docs failed ?

lihaoyi:
@0xnm not sure if you're available to help review this, if not I'll give it a shot (don't have recent android experience myself)

himanshumahajan138:
Great, btw I have followed official android latest docs (but they don't have good docs for someone to understand seriously)

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -124,33 +212,23 @@ trait AndroidAppModule extends JavaModule {
...
-        unsignedApk.toString // Output APK
-      ) ++ Seq(androidDex().path.toString) // Include DEX files
-    )
+    os.copy.over((T.dest / os.up / "androidResources.dest" / "res.apk"), unsignedApk)
+    os.call(Seq(
+      "zip",
+      "-j",
+      unsignedApk.toString,
+      s"${androidDex().path / "classes.dex"}"
+    ))
```
0xnm, line 231: `os.zip`?

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
```diff
@@ -0,0 +1,214 @@
...
+    // Download bundle tool and generate .aab using the bundle zip
+    os.write(bundleToolJar, requests.get(bundleToolUrl()).bytes)
+    os.call((
+      "java",
+      "-jar",
+      bundleToolJar,
+      "build-bundle",
+      s"--modules=${zipPath}",
+      s"--output=${bundleFile}"
+    ))
```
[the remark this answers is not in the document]
lihaoyi: I think for this kind of executable assembly using `java -jar` is fine; the `Jvm.*` stuff is mostly for making it convenient to execute classpaths

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -124,33 +212,23 @@ trait AndroidAppModule extends JavaModule {
...
-        "-f",
-        "-M",
-        androidManifest().path.toString, // Path to AndroidManifest.xml
-        "-I",
-        androidSdkModule().androidJarPath().path.toString, // Include Android JAR
-        "-F",
-        unsignedApk.toString // Output APK
-      ) ++ Seq(androidDex().path.toString) // Include DEX files
-    )
+    os.copy.over((T.dest / os.up / "androidResources.dest" / "res.apk"), unsignedApk)
```
lihaoyi, line 225: `os.copy` should be enough here, since the body of a `Task` gets wiped out every time before it runs (unless it has `persistent = true`

scalalib/src/mill/javalib/android/AndroidSdkModule.scala
```diff
@@ -66,6 +65,13 @@ trait AndroidSdkModule extends Module {
     PathRef(buildToolsPath().path / "aapt")
   }
 
+  /**
+   * Provides the path to AAPT2, used for resource handling and APK packaging.
```
lihaoyi, line 69: Is there any `aapt2` documentation we can link to here?

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
```diff
@@ -0,0 +1,214 @@
...
+   * Packages compiled resources and dex files into a zip for app bundle generation.
+   *
+   * @return PathRef to the created bundle zip.
+   */
+  def androidBundleZip: T[PathRef] = Task {
+    val dexFile = androidDex().path
+    val baseDir = Task.dest / "base"
+
+    // Create directory structure for Android bundle
+    Seq("manifest", "dex").foreach(d => os.makeDir.all(baseDir / d))
```
lihaoyi, line 104: Let's skip thie `makedir.All` and instead use `createFolders = true` below where necessary. That will be less fragile as we won't need to worry about them falling out of sync

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
```diff
@@ -0,0 +1,214 @@
...
+
+    // Copy dex and resource files into the bundle directory
+    os.copy(dexFile, Task.dest / dexFile.last)
+    os.unzip(Task.dest / os.up / "androidResources.dest" / "res.apk", Task.dest)
```
lihaoyi, line 108: We should never use `Task.dest / os.up`, as a task should only write to its own `dest` folder. Not sure what you are trying to do here, but if you want to write to `androidResources.dest` then the `def androidResources` task should do the writing

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
```diff
@@ -0,0 +1,214 @@
...
+    os.copy(Task.dest / "AndroidManifest.xml", baseDir / "manifest" / "AndroidManifest.xml")
```
lihaoyi, line 109: We can now write `baseDir / "manifest/AndroidManifest.xml"`

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
```diff
@@ -0,0 +1,214 @@
...
+
+    // Copy .pb files (compiled protobuf resources) and organize dex files
+    os.walk(Task.dest).filter(_.ext == "pb").foreach(file => os.copy(file, baseDir / file.last))
+    os.walk(Task.dest).filter(_.ext == "dex").zipWithIndex.foreach { case (file, idx) =>
+      os.copy(file, baseDir / "dex" / (if (idx == 0) "classes.dex" else s"classes${idx + 1}.dex"))
+    }
```
lihaoyi, line 115: This renaming of `.dex` files to `classes$n.dex` is a bit weird. Should we be preserving the original file names like we do for `pb`?

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
```diff
@@ -0,0 +1,214 @@
...
+
+    PathRef(Task.dest) // Return compiled resource directory as PathRef
+  }
+
+  /**
+   * Packages compiled resources and dex files into a zip for app bundle generation.
+   *
+   * @return PathRef to the created bundle zip.
+   */
+  def androidBundleZip: T[PathRef] = Task {
```
lihaoyi, line 99: Is there any documentation for what should go in an android bundle zip that we can link to here?

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
```diff
@@ -0,0 +1,214 @@
...
+        resDir,
+        "-o",
+        outputZip
+      ))
+      outputZip
+    }
+
+    // Separate main resource zip and library zips for later linking
+    val ResourceZip = compiledZips.find(_.toString.contains("resources"))
+    val libzips = compiledZips.filterNot(_.toString.contains("resources"))
```
lihaoyi, line 68: We can do `val (resourceZip, libZips) = compiledZips.partition(...)`

What are you trying to do with `.contains("resources")` though? Seems like there should be a better way to do it

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
```diff
@@ -0,0 +1,214 @@
...
+   */
+  def aapt2Url: T[String] = Task {
+    "https://dl.google.com/android/maven2/com/android/tools/build/aapt2/8.7.2-12006047/aapt2-8.7.2-12006047-linux.jar"
+  }
+
+  /**
+   * URL to download bundle tool, used for creating Android app bundles (AAB files).
+   */
+  def bundleToolUrl: T[String] = Task {
+    "https://github.com/google/bundletool/releases/download/1.17.2/bundletool-all-1.17.2.jar"
```
lihaoyi, line 26: Should this be on `AndroidSdkModule`? That way it would be in the same place as the other downloaded tools and maybe be easier to find

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -36,32 +36,120 @@ trait AndroidAppModule extends JavaModule {
...
+      val extractDir = T.dest / aarFile.baseName
+      os.unzip(aarFile, extractDir)
+
+      // Collect all .jar files in the AAR directory
+      jarFiles ++= os.walk(extractDir).filter(_.ext == "jar").map(PathRef(_))
+
+      // If the res folder exists, add it to the resource folders
+      val resFolder = extractDir / "res"
+      if (os.exists(resFolder)) {
+        resFolders :+= PathRef(resFolder)
```
lihaoyi, line 69: Let's use `resFolders ++= Seq(PathRef(resFolder))`

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
```diff
@@ -0,0 +1,214 @@
...
+    // Define paths for the AAPT2 jar and extracted tool
+    val aapt2Jar = Task.dest / "aapt2.jar"
+    val aapt2 = Task.dest / "aapt2"
+    val compiledResDir = Task.dest / "compiled"
+
+    // Get resource directories and filter for existing ones
+    val resourceDirs = resources().map(_.path).filter(os.exists)
+
+    // Download and extract AAPT2 tool
+    os.write(aapt2Jar, requests.get(aapt2Url()).bytes)
```
lihaoyi, line 45: I think we don't need the `.bytes` here, and don't need a local file if we do `os.unzip.stream(requests.get.stream(aapt2Url()), Task.dest`

lihaoyi: But yeah if it's already in `AndroidSdkModule` we don't need this here at all

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
```diff
@@ -0,0 +1,214 @@
...
+    os.unzip(aapt2Jar, Task.dest)
+    os.call(("chmod", "+x", aapt2)) // Make AAPT2 executable
+    os.makeDir.all(compiledResDir) // Ensure compiled resource directory exists
+
+    // Compile each resource directory using AAPT2
+    var count = 0
+    val compiledZips = resourceDirs.map { resDir =>
+      val outputZip = compiledResDir / s"${resDir.last}-${count}.zip"
+      count += 1
+      os.call((
```
lihaoyi, line 55: This can probably be one lined

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
```diff
@@ -0,0 +1,214 @@
...
+    os.copy(Task.dest / "AndroidManifest.xml", baseDir / "manifest" / "AndroidManifest.xml")
+
+    // Copy .pb files (compiled protobuf resources) and organize dex files
+    os.walk(Task.dest).filter(_.ext == "pb").foreach(file => os.copy(file, baseDir / file.last))
+    os.walk(Task.dest).filter(_.ext == "dex").zipWithIndex.foreach { case (file, idx) =>
+      os.copy(file, baseDir / "dex" / (if (idx == 0) "classes.dex" else s"classes${idx + 1}.dex"))
+    }
+
+    // Copy additional resources (res, lib, assets) if present
+    Seq("res", "lib", "assets").foreach { folder =>
```
lihaoyi, line 118: Better to use `for` loops than `.foreach` for this kind of thing, here and elsewhere in the PR

himanshumahajan138:
@0xnm and @lihaoyi i know you are concerned about the usage of aapt2 from the downloaded version 

Actually its clearly mentioned in the official docs that we can't use aapt2 which came bundled with build tools

See this: https://developer.android.com/build/building-cmdline#bundletool-build

scalalib/src/mill/javalib/android/AndroidAppModule.scala
himanshumahajan138: 😂 i think it will work but i didnt tried no worries I will  update...

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
```diff
@@ -0,0 +1,214 @@
...
+
+    PathRef(signedBundle)
+  }
+
+  /**
+   * Generates a universal APK from the signed Android App Bundle (AAB).
+   *
+   * @return PathRef to the generated universal APK.
+   */
+  def androidApk: T[PathRef] = Task {
```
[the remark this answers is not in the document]
himanshumahajan138: Why we created bundle ? 

To generate "apks" right, I know you will say that its for uploading straight to playstore or apple store but its our duty to show and check the validation of bundle from our side so that it will not cause error during uploading to stores 
And we are providing the universal apk not the device specific apk so think this should be done coz if usee created bundle then they should be able to create apk from it...

Rest what you say will be my priority...

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
```diff
@@ -0,0 +1,214 @@
...
+    val aapt2 = Task.dest / "aapt2"
+    val compiledResDir = Task.dest / "compiled"
+
+    // Get resource directories and filter for existing ones
+    val resourceDirs = resources().map(_.path).filter(os.exists)
+
+    // Download and extract AAPT2 tool
+    os.write(aapt2Jar, requests.get(aapt2Url()).bytes)
+    os.unzip(aapt2Jar, Task.dest)
+    os.call(("chmod", "+x", aapt2)) // Make AAPT2 executable
```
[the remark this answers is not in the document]
himanshumahajan138: https://github.com/com-lihaoyi/mill/pull/3935#issuecomment-2467205697

scalalib/src/mill/javalib/android/AndroidSdkModule.scala
himanshumahajan138: Yes i will add

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
himanshumahajan138: Actually i was doing this before but i don't know why it was clearing old content before adding new 
What you think is happening here 🤔

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
himanshumahajan138: Actually i want the reas.apk which was generated during the android resources process and in need to add dex file into it and as we know before running androiddex we will always run the android resources so think there will be no issue with the usage of os.up 

But still if you say i will call the android resources task rather than using hard-coded path to res.apk

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
himanshumahajan138: No sir its also mentioned in the official docs that how to manage multiple dex files 

See here: https://developer.android.com/build/building-cmdline#package_pre-compiled_code_and_resources

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
himanshumahajan138: Yes I will add

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
himanshumahajan138: Actually during compiling when we have multiple resources like from libs and main user defined resources so at that time we are renaming them res prefix used for libs resources and resources prefix used for user resources zip 

So to partition them (as they are in the same folder) i have used this

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
himanshumahajan138: It can be

scalalib/src/mill/javalib/android/AndroidAppModule.scala
himanshumahajan138: 😊 ok

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
himanshumahajan138: https://github.com/com-lihaoyi/mill/pull/3935#issuecomment-2467205697

himanshumahajan138:
I Request @0xnm and @lihaoyi to please check my replies for every review and again review them as most of these are already solved and some requires updation so rather then give small small updates lets firat figure out every possible update again and then finalize it

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
lihaoyi: Got it, in that case please add a comment linking to the relevant documentation

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
lihaoyi: In that case you should copy the `res.apk` in this task's `Test.dest` folder and modify your own copy

lihaoyi:
@himanshumahajan138 if the docs say to download it separately that's fine, but there should be a comment linking to the relevant documentation

himanshumahajan138:
> @himanshumahajan138 if the docs say to download it separately that's fine, but there should be a comment linking to the relevant documentation

Let me do experiment 😂, i will try the aapt2 which cane bundled and see if there is some issue if not then we can use the bundled one 🙃

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
```diff
@@ -0,0 +1,214 @@
...
+ */
+@mill.api.experimental
+trait AndroidAppBundle extends AndroidAppModule with JavaModule {
+
+  /**
+   * URL to download AAPT2, a tool for compiling and linking Android resources.
+   */
+  def aapt2Url: T[String] = Task {
+    "https://dl.google.com/android/maven2/com/android/tools/build/aapt2/8.7.2-12006047/aapt2-8.7.2-12006047-linux.jar"
+  }
```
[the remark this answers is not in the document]
0xnm: Even if you need to use custom `aapt2`, you shouldn't limit this PR functionality to Linux only by using tools which are not available on other platforms (e.g. by downloading Linux-variant of `aapt2` irrespective of the platform, or by using `chmod`, which don't exist on Windows system).

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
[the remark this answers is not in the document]
0xnm: I mean you can upload bundle to Play Store directly as `aab` file. And you can generate APK as well without going through the bundle flow. So what is the exact use-case of generating APK from AAB?

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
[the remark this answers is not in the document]
himanshumahajan138: Sir Give me 30 mins all the issues sorted let me give the updated code and in that everything will be sorted

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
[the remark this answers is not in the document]
himanshumahajan138: TBH I don't know what the use-case will be 😅

What you say?
- Remove Apk Generation using Bundle (no use-case)

My Doubt:
- Should we sign the bundle or not?

himanshumahajan138:
And Here We have all Updates...

@lihaoyi and @0xnm 
Please Check this now...

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -36,34 +36,118 @@ trait AndroidAppModule extends JavaModule {
...
+   *
+   * @return The updated classpath with `.jar` files only.
+   */
+  override def compileClasspath: T[Agg[PathRef]] = Task {
+    // Call the function to get jar files and resource paths
+    val (jarFiles, _) = androidUnpackArchives()
+    super.compileClasspath().filter(_.path.ext == "jar") ++ jarFiles
+  }
+
+  def linkOptions: T[Seq[String]] = Task {
```
lihaoyi, line 102: Let's rename this `androidLinkOptions`

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
lihaoyi: bump

himanshumahajan138:
@lihaoyi Don't Worry about Docs and Links i will Updated them after code satisfaction

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -36,34 +36,118 @@ trait AndroidAppModule extends JavaModule {
...
-      "-J",
-      genDir.toString, // Generate R.java files
-      "-M",
-      androidManifest().path.toString, // Use AndroidManifest.xml
-      "-I",
-      androidSdkModule().androidJarPath().path.toString // Include Android SDK JAR
-    ))
+    // Partition compiledZips into resourceZip and libZips
+    // "resources" is used to differ the libs res from app res
+    val (resourceZip, libZips) = compiledZips.partition(_.toString.contains("resources"))
```
lihaoyi, line 128: This `toString.contains("resources")` is still unacceptably sloppy. Since we construct the paths ourselves above in `outputZip`, we should be able to be more precise in exactly how we determine something is a resource vs a library. At the very least we should be able to look at `resDir.segments.toSeq` and check individual path segments rather than stringifying the whole thing

scalalib/src/mill/javalib/android/AndroidSdkModule.scala
lihaoyi: ?

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
```diff
@@ -0,0 +1,117 @@
...
+ * Defines an Android app bundle module with tasks for resource compilation,
+ * bundling using bundle tool.
+ */
+@mill.api.experimental
+trait AndroidAppBundle extends AndroidAppModule with JavaModule {
+
+  override def linkOptions: T[Seq[String]] = Task {
+    super.linkOptions() ++ Seq[String](
+      "--proto-format"
+    ) // protoformat for resources linking(required for bundle)
```
lihaoyi, line 18: This should be one line `super.linkOptions() ++ Seq[String]("--proto-format")`

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
```diff
@@ -0,0 +1,117 @@
...
+   * @return PathRef to the created bundle zip.
+   */
+  def androidBundleZip: T[PathRef] = Task {
+    val dexFile = androidDex().path
+    val resFile = androidResources().path / "res.apk"
+    val baseDir = Task.dest / "base"
+
+    // Copy dex and resource files into the bundle directory
+    os.copy(dexFile, Task.dest / dexFile.last)
+    os.unzip(resFile, Task.dest)
```
lihaoyi, line 33: This `unzip` into `Task.dest` is a bit sloppy. We don't know what `resFile` contains, and so if it ends up containing files or files named `manifest` or `base` everything might break. You should unzip it into a subfolder of `Task.dest` so we can be guaranteed that it will be an empty folder and there will be no collisions

scalalib/src/mill/javalib/android/AndroidAppModule.scala
0xnm: Word `link` is not used in the Android build pipeline, this seems like rather `aaptOptions` or something like that.

himanshumahajan138: i think resourcesLinkOptions is fine

lihaoyi: We should prefix it with `android` since `AndroidAppModule` is meant to be mixed into `JavaModule` so we need to try and avoid collisions. How about `androidAaptOptions`

himanshumahajan138: great `androidAaptOptions` is Good but i think this will tell that its for whole aapt options but in our case its only for linking

himanshumahajan138:
> This `toString.contains("resources")` is still unacceptably sloppy. Since we construct the paths ourselves above in `outputZip`, we should be able to be more precise in exactly how we determine something is a resource vs a library. At the very least we should be able to look at `resDir.segments.toSeq` and check individual path segments rather than stringifying the whole thing


I created Multiple resource folders situation for you and now explain me what you want to achieve 

here `app-resources.zip` is the zip which contains user resources 
and other zips are for libraries resources (dependencies)

image

lihaoyi:
If thats what you need to do the  you need to find abetter way to do it

himanshumahajan138:
@lihaoyi @0xnm Let's Review this again for more refining if required...

Overall code is working, apps + bundles are working as expected
Need Further Reviews and then Final Documentation Part for build Files...

example/kotlinlib/android/1-hello-world/app/src/main/java/com/helloworld/app/MainActivity.kt
```diff
@@ -23,10 +23,13 @@ class MainActivity : Activity() {
...
 
         // Set layout parameters (width and height)
-        textView.layoutParams = LayoutParams(
-            LayoutParams.MATCH_PARENT,
-            LayoutParams.MATCH_PARENT
-        )
+        textView.layoutParams = LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.MATCH_PARENT)
+
+        // Set the text color using a resource
+        textView.setTextColor(getColor(R.color.text_green)) // Using hex color code directly
```
lihaoyi, line 29: What does `// Using hex color code directly` mean?

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
```diff
@@ -0,0 +1,127 @@
...
+   * For Bundle Zip Format please See the official
+   * [[https://developer.android.com/guide/app-bundle/app-bundle-format Documentation]]
+   */
+  def androidBundleZip: T[PathRef] = Task {
+    val dexFile = androidDex().path
+    val resFile = androidResources().path / "res.apk"
+    val baseDir = Task.dest / "base"
+    val appDir = Task.dest / "app"
+
+    // Copy dex and resource files into the bundle directory
```
lihaoyi, line 35: All these comments are redundant and should be removed

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
```diff
@@ -0,0 +1,127 @@
...
+
+    // Copy .pb files (compiled protobuf resources) and organize dex files using for loops
+    for (file <- os.walk(appDir).filter(_.ext == "pb")) {
+      os.copy(file, baseDir / file.last)
+    }
+
+    for ((file, idx) <- os.walk(Task.dest).filter(_.ext == "dex").zipWithIndex) {
+      os.copy(
+        file,
+        baseDir / "dex" / (if (idx == 0) "classes.dex" else s"classes${idx + 1}.dex"),
```
lihaoyi, line 52: Where is the documentation link for the `classes{n}.dex` file names?

example/kotlinlib/android/1-hello-world/app/src/main/java/com/helloworld/app/MainActivity.kt
himanshumahajan138: 😂😂😂 actually these documentations are not updated i will update them with updation of build docs

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -22,48 +22,126 @@ import mill.scalalib.JavaModule
...
+
+    // Return both jar files and resource folders
+    (jarFiles, resFolders)
+  }
+
+  /**
+   * Combines module resources with those unpacked from AARs.
+   */
+  def resources: T[Seq[PathRef]] = Task {
+    // Call the function to unpack AARs and get the jar and resource paths
```
lihaoyi, line 75: This comments are redundant and can be removed

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
himanshumahajan138: sorry forgot will add this time

himanshumahajan138:
Please Confirm once if code is acceptable then should i move to documentation Part ???

himanshumahajan138:
@lihaoyi ThirdParty Arrow Test Failed 🥲 why these type of extra ordinary things happen with me...

lihaoyi:
That test is flaky, you can ignore it

himanshumahajan138:
ok then lest Finalize this PR

scalalib/src/mill/javalib/android/AndroidAppBundle.scala
0xnm: Such split is a part of multidex https://developer.android.com/build/multidex#mdex-on-l, you can reference this page if you want.

I'm not sure though why you don't preserve original file name: in your case original `classes2.dex` may become `classes3.dex` or `classes.dex`, for example, if the order of enumeration produced by `walk` is not the same as alphabetical ordering? Or are class files you are traversing have different naming?

I would suggest keeping original names, because `classes.dex` has a special treatment: it is always loaded first and some classes are packaged there specifically to allow application start.

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -22,48 +22,109 @@ import mill.scalalib.JavaModule
...
-   * This method must be implemented by the concrete class to specify the SDK paths.
-   *
-   * @return The Android SDK module that is used across the project.
+   * Provides access to the Android SDK configuration.
    */
   def androidSdkModule: ModuleRef[AndroidSdkModule]
 
   /**
-   * An XML file containing configuration and metadata about your android application
+   * Loads `AndroidManifest.xml` file.
```
0xnm, line 30: Does it load anything? It seems it just provides a path.

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -22,48 +22,109 @@ import mill.scalalib.JavaModule
...
    */
   def androidManifest: Task[PathRef] = Task.Source(millSourcePath / "AndroidManifest.xml")
 
   /**
-   * Generates the Android resources (such as layouts, strings, and other assets) needed
-   * for the application.
+   * Adds "aar" to the handled artifact types.
+   */
+  def artifactTypes: T[Set[coursier.Type]] = Task { super.artifactTypes() + coursier.Type("aar") }
```
0xnm, line 37: there is `aar` support added here, but examples lack any 3rd party dependency, is it intentional?

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -22,48 +22,109 @@ import mill.scalalib.JavaModule
...
+
+  /**
+   * Extracts JAR files and resources from AAR dependencies.
+   *
+   * @return Paths to extracted JAR files and resource folders.
+   */
```
0xnm, line 43: there is no any dependency usage in the examples, so I believe this method is not really tested?

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -95,62 +153,50 @@ trait AndroidAppModule extends JavaModule {
...
 
     os.call(
       Seq(
-        androidSdkModule().d8Path().path.toString, // Call d8 tool
+        androidSdkModule().d8Path().path.toString,
         "--output",
-        jarFile.toString, // Output JAR file
-        "--no-desugaring" // Disable desugaring
+        jarFile.toString,
+        "--no-desugaring"
```
0xnm, line 163: desugaring is a configurable parameter in the build setup in Gradle, so maybe it makes sense to make it configurable here as well

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -191,25 +235,23 @@ trait AndroidAppModule extends JavaModule {
...
+      "sign",
+      "--ks",
+      androidKeystore().path,
+      "--ks-key-alias",
+      "androidkey",
+      "--ks-pass",
+      "pass:android",
+      "--key-pass",
+      "pass:android",
+      "--out",
```
0xnm, line 251: this is only for the debug key, obviously. If someone want to build a signed **release** package, it should be an option to provide a singing configuration: release key and credentials for it.

himanshumahajan138: No worries we will add a function which will take args from user and then sign...

scalalib/src/mill/javalib/android/AndroidAppModule.scala
himanshumahajan138: Yes we can i will update to desugaring

scalalib/src/mill/javalib/android/AndroidAppModule.scala
himanshumahajan138: Actually its tested in my pr #3769 i have tested it my self and we will require this in future for kotlin jetpack compose or any other aar library support

himanshumahajan138: And if we want to test it again don't worry we will do that while merging #3769

scalalib/src/mill/javalib/android/AndroidAppModule.scala
himanshumahajan138: As mentioned below this will be used in future in jetpack compose or third-party addition examples

scalalib/src/mill/javalib/android/AndroidAppModule.scala
himanshumahajan138: 😂 it don't loads anything it provides path i used loads as to make it short and simple

himanshumahajan138:
https://github.com/com-lihaoyi/mill/pull/3935#discussion_r1838838171

Actually i do agree with preserving original names as they will be already present in different names like classes.dex or classes2.dex and there will be no duplicate name files as they all are made during same dex process and it will provide different names 

So no worries i will update the dex file handling...

Anything else ?

lihaoyi:
@himanshumahajan138 how are you testing this? I'm trying to load the aab file into my android emulator but it doesn't seem to be generating an icon on the home screen for me to open the app

himanshumahajan138:
@lihaoyi What you think should we add Android Apk Generation from AAB Bundle ?

lihaoyi:
So what is the appropriate way to deploy the AAB bundle so i csn verify it is working as expected?

lihaoyi:
Do i have to deploy it all the way to google play store? Ir is there some way to test it locally?

himanshumahajan138:
if you want to test locally then use this comment https://github.com/com-lihaoyi/mill/pull/3935#issuecomment-2473683367

else you can also try playstore or appstore but they are not suggested as you just want to test not to produce it for market

lihaoyi:
got it, let me try that out

himanshumahajan138:
@lihaoyi A Cup of Coffee always looks good if someone make it for you 😅

Here is the Exact Command For you 
```
 java -jar ~/opensource/mill/out/example/javalib/android/2-app-bundle/local/testCached.dest/sandbox/run-1/out/androidSdkModule0/bundleToolPath.dest/bundleTool.jar build-apks --ks ~/opensource/mill/out/example/javalib/android/2-app-bundle/local/testCached.dest/sandbox/run-1/bundle/releaseKey.jks --bundle=~/opensource/mill/out/example/javalib/android/2-app-bundle/local/testCached.dest/sandbox/run-1/out/bundle/androidBundle.dest/signedBundle.aab --output=try.apks --mode=universal --ks-key-alias releaseKey --ks-pass pass:MillBuildTool
 ```
 
 
 Just Replace the `~/opensource/mill` with the path you have in your laptop

himanshumahajan138:
@lihaoyi Can you please Explain me why `integration.invalidation[_].server.testCached` Failed ?

lihaoyi:
Tested it manually by generating the APK, seems to work. The code isn't perfect but good enough for now. @himanshumahajan138 I'll close out the bounty and send you the payslip

Issue #3894, opened 2024-11-02
Consolidate import $file.foo.bar into mill.foo.bar namespace


lihaoyi:
This would make it more consistent with how Scala source files behave in general and allow IDEs to understand the references without any customization. The old syntax can remain for migration compatibility

Might need to wait for 0.13.0 depending on whether this needs scala 3 or whether it needs to break bincompat (not sure)

PR #3942, opened 2024-11-11
Add stubs for `import $meta` and `import $packages` and `import $ivy`

lihaoyi:
This helps make the IDE happy

Before:

Screenshot 2024-11-12 at 12 57 33 AM

After:

Screenshot 2024-11-12 at 1 23 51 AM

We can't do similar for `import $file`, because it conflicts with the `=> $file` renaming import in the code generation. But hopefully that would be fixed later by https://github.com/com-lihaoyi/mill/issues/3894


lefou (commented):
That's a good idea!

PR #3948, opened 2024-11-12
NIT Use allIvyDeps task where it seems it's meant to be used

alexarchambault:
I was looking at this file, and it feels `allIvyDeps` was meant to be used in `transitiveIvyDeps `, but isn't. This fixes that. (Unless `allIvyDeps` is somehow scheduled for deletion?)


lihaoyi:
Seems reasonable, CC @lefou

lefou (commented):
Looks good to me. 

It was probably an oversight when `allIvyDeps` was introduced. Looks like we're missing some test coverage.

PR #3949, opened 2024-11-12
Allow users to customize coursier ResolutionParams

alexarchambault:
This adds a `JavaModule#resolutionParams` task, allowing users to customize coursier's [`ResolutionParams`](https://github.com/coursier/coursier/blob/b04937dda13cee80b3743c84f471a69eb430047e/modules/coursier/shared/src/main/scala/coursier/params/ResolutionParams.scala). These allow to customize some aspects of dependency resolution.

In particular, `ResolutionParams#enableDependencyOverrides`, when set to `Some(false)`, disables the recent BOM changes added (and enabled by default) in coursier `2.1.17`. This might be useful, if ever users aren't happy with the versions it selects, or want to wait a bit before enabling that, or if they find out performance suffers because of it for them.

The new task is anonymous (it's not a target, it's not serialized on disk), as `ResolutionParams` has a field that doesn't easily lend itself to serialization (`rules`, because of the `Rule` type, which is an open abstract class).


lefou (commented):
The PR description doesn't match what I can see in the diff. I can't find the `resolutionParams` task but a lot of `ResolutionParams` contructions in the default-arg position. I'd prefer to see the anonymous task since that matched all the other customization points we already provide for coursier.

lefou:
The new task should definitely come with some documentation, since some other customizations seem to affect the final `ResolutionParams` as well. What overrides what?

alexarchambault:
> What do you mean?

Oh, I think I see

lefou (commented):
Looks good to me. I restarted the failed test to see, if it was flakyness.

Issue #3921, opened 2024-11-07
Don't fail, if a `build.{sc,mill}` is found in sub-directories


lefou:
Mill currently fails with an error message when it detects a `build.sc` or `build.mill` file in a sub-directory. This isn't an issue in any way, so it should just ignore it. Printing a warning message might be an option, but I currently don't see why this is even a worth a warning. 

IMHO the most likely situation where users might want to be warned is, when we detect a `package.mill` file next to an `build.mill`. This means, that a directory is a root project as well as a sub-project. This is not an issue per se, but `package.mill` files further down the directory hierarchy might be ending up in both Mill projects, which can be wanted or not.

* See also dicsussion https://github.com/com-lihaoyi/mill/discussions/3349#discussioncomment-11180133

PR #3954, opened 2024-11-13
Ignore `build.mill` in subdir and `package.mill` in root dir

Tang-Haojin:
If a directory is both used as a root project and a sub-project, `build.{sc,mill,mill.scala}` and `package.{sc,mill,mill.scala}` may exist together. Simply ignore them in corresponding situations should be safe.

Fix https://github.com/com-lihaoyi/mill/issues/3921


lefou (approved):
Looks good to me. Thank you!

PR #3957, opened 2024-11-14
This is a forkArg not a compiler option

Quafadas:
Attempt 3 to make this trivial change. No rebasing this time.

The previous solution works, because both are a list of strings in the end, but I think this revised version more correctly follows mill nomenclature as forkedArgs rather than compiler flags. 

Apologies for the spamminess.


lihaoyi:
@Quafadas could you flesh out the PR description a bit with the summary of the problem and the current status

Quafadas:
@lihaoyi done.

Issue #3834, opened 2024-10-24
`--meta-level 1 show sources` triggers compilation


lefou:
In an existing project with a meta-build, running `mill --meta-level 1 show sources` triggers compilation in the same meta-level. This comes to a surprise, as compilation depends on the sources and I was in the process to analyze a compile error, for which I wanted to inspect the sources.

Reproducer:

```scala
// build.sc
import $meta._

this is a build error
```

```scala
// mill-build/build.sc
object `package` extends MillBuildRootModule
```

```
> mill --meta-level 1 show "{sources,allSourceFiles}"
No mill version specified.
You should provide a version via '.mill-version' file or --mill-version option.
Using mill version 0.12.0
[mill-build/build.mill-57/61] compile
[mill-build/build.mill-57] [info] compiling 1 Scala source to /tmp/mill2/out/mill-build/mill-build/compile.dest/classes ...
[mill-build/build.mill-57] [info] done compiling
[build.mill-57/61] compile
[build.mill-57] [info] compiling 1 Scala source to /tmp/mill2/out/mill-build/compile.dest/classes ...
[build.mill-57] [error] /tmp/mill2/build.mill:6:6: value is is not a member of build_.package_
[build.mill-57] [error] this is a build error
[build.mill-57] [error]      ^
[build.mill-57] [error] /tmp/mill2/build.mill:6:9: not found: value a
[build.mill-57] [error] this is a build error
[build.mill-57] [error]         ^
[build.mill-57] [error] /tmp/mill2/build.mill:6:17: not found: value error
[build.mill-57] [error] this is a build error
[build.mill-57] [error]                 ^
[build.mill-57] [error] three errors found
[61/61] ====================================================================================================== show {sources,allSourceFiles} ==========================================================================================================
1 tasks failed
compile Compilation failed
```

I would expect to see just the sources files, so I can examine them to probably spot any issues resulting in the compile error.

lefou:
This is no new issue. I can reproduce this with Mill `0.11.0` too. 

```
> MILL_VERSION="0.11.0" mill --meta-level 1 show sources
[mill-build/build.sc] [41/49] compile 
[info] compiling 1 Scala source to /tmp/mill2/out/mill-build/mill-build/compile.dest/classes ...
[info] done compiling
[build.sc] [41/49] compile 
[info] compiling 1 Scala source to /tmp/mill2/out/mill-build/compile.dest/classes ...
[...]

PR #3960, opened 2024-11-14
Avoid running subsequent meta-level builds after reaching the desired meta-level in mill-build-bootstrap

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/3834


lefou (approved):
Looks good to me.

PR #3966, opened 2024-11-14
Default KtlintModule to .editorconfig explicitly

myyk:
For some reason `ktlint` behaves differently if it's config is set vs unset to it's default of `.editorconfig`.

It seems to be unable to disable `ktlint` when not explicitly set, so I think we should set it instead.

Why not bump the version while we're at it?

### DIFF
--- a/kotlinlib/src/mill/kotlinlib/ktlint/KtlintModule.scala
+++ b/kotlinlib/src/mill/kotlinlib/ktlint/KtlintModule.scala
@@ -32,7 +32,7 @@ trait KtlintModule extends JavaModule {
    * Ktlint configuration file.
    */
   def ktlintConfig: T[Option[PathRef]] = Task {
-    Some(PathRef(millSourcePath / ".editorconfig"))
+    Some(PathRef(Task.workspace / ".editorconfig"))
   }
 
   /**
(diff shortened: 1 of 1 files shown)


kotlinlib/src/mill/kotlinlib/ktlint/KtlintModule.scala
```diff
@@ -32,14 +32,14 @@ trait KtlintModule extends JavaModule {
    * Ktlint configuration file.
    */
   def ktlintConfig: T[Option[PathRef]] = Task {
-    None
+    Some(PathRef(millSourcePath / ".editorconfig"))
   }
```
lihaoyi, line 37: How about defaulting to `Task.workspace / ".editorconfig"`? I'd guess most folks would have the config file at the root of their project (which is what Task.workspace is) rather than in each module's subfolder

myyk:
Sounds great. I'm offline now. Feel  free to change it now or I'll handle
it tomorrow when I'm up.

kotlinlib/src/mill/kotlinlib/ktlint/KtlintModule.scala
lefou: This should be an `Task.Input`

PR #3974, opened 2024-11-15
Added support for Revapi

### LINKED ISSUE #3929: RevApi support (500USD Bounty)
---

---

This has come up in both Netty and Mockito case studies, so we should properly support it with a built in Mill plugin

Will likely end up looking similar to our support for linters like ErrorProne and Checkstyle, as well as @lolgab's mill-mima plugin

ajaychandran:
Added [Revapi](https://revapi.org/revapi-site/main/index.html) support for API analysis and change tracking to identify incompatibilities.

Resolves #3929.

### DIFF
--- a/example/javalib/publishing/3-revapi/build.mill
+++ b/example/javalib/publishing/3-revapi/build.mill
@@ -1,0 +1,53 @@
+//// SNIPPET:BUILD
+package build
+import mill._, javalib._, publish._, revapi._
+
+object bar extends JavaModule with RevapiModule {
+  def publishVersion = "0.0.1"
+
+  def pomSettings = PomSettings(
+    description = "Hello",
+    organization = "com.lihaoyi",
+    url = "https://github.com/lihaoyi/example",
+    licenses = Seq(License.MIT),
+    versionControl = VersionControl.github("lihaoyi", "example"),
+    developers = Seq(Developer("lihaoyi", "Li Haoyi", "https://github.com/lihaoyi"))
+  )
+
+  override def revapiConfigFiles: T[Seq[PathRef]] =
+    // add Revapi config JSON file(s)
+    Task.Sources(millSourcePath / "conf/revapi.json")
+
+  override def revapiClasspath: T[Agg[PathRef]] = T {
+    // add folder containing logback.xml
+    super.revapiClasspath() ++ Seq(PathRef(millSourcePath / "conf"))
+  }
+}
+
+// This example uses the `revapi` task, provided by the `RevapiModule`, to run an
+// analysis on old and new archives of a module to identify incompatibilities.
+//
+// NOTE: For demonstration purposes, an archive, to compare against, is published locally.
+// In real usage, the old version would be downloaded from the publish repository.
+
+/** Usage
+
+> mill bar.publishLocal
+Publishing Artifact(com.lihaoyi,bar,0.0.1) to ivy repo...
+
+> cp dev/src/Visibility.java bar/src/Visibility.java
+
+> mill bar.revapi
+Starting analysis
+Analysis results
+----------------
+old: field Visibility.SuperClass.f @ Visibility.SubClass
+new: <none>
+java.field.removed: Field removed from class.
+... BREAKING
+old: field Visibility.f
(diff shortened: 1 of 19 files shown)


example/javalib/publishing/3-revapi/build.mill
```diff
@@ -0,0 +1,53 @@
...
+Analysis results
+----------------
+old: field Visibility.SuperClass.f @ Visibility.SubClass
+new: <none>
+java.field.removed: Field removed from class.
+... BREAKING
+old: field Visibility.f
+new: field Visibility.f
+java.field.visibilityReduced: Visibility was reduced from 'public' to 'protected'.
+... BREAKING
```
ajaychandran, line 51: The actual output lists incompatibilities at several levels but not always in he same order.
```bash
old: field Visibility.SuperClass.f @ Visibility.SubClass
new: <none>
java.field.removed: Field removed from class.
BINARY: BREAKING, SOURCE: BREAKING

old: field Visibility.f
new: field Visibility.f
java.field.visibilityReduced: Visibility was reduced from 'public' to 'protected'.
BINARY: BREAKING, SOURCE: BREAKING
```

lihaoyi: should be fine, the example test output assertions dont care about order as long as each expected line exists

ajaychandran:
@lihaoyi Please review.

scalalib/src/mill/javalib/revapi/RevapiModule.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  /**
+   * Runs [[revapiCliVersion Revapi CLI]] on this module's archives.
+   */
+  def revapi(args: String*): Command[Unit] = Task.Command {
+    val workingDir = millSourcePath
+
+    val oldFiles = revapiOldFiles()
+    val oldFile = oldFiles.head
+    val oldSupFiles = oldFiles.tail
```
lihaoyi, line 23: This (and below) can be written more concisely as `val oldFile +: oldSupFiles = revapiOldFiles()`

scalalib/src/mill/javalib/revapi/RevapiModule.scala
```diff
@@ -0,0 +1,121 @@
...
+ * Adds support for [[https://revapi.org/revapi-site/main/index.html Revapi checker]] to enable API analysis and change tracking.
+ */
+@mill.api.experimental // until Revapi has a stable release
+trait RevapiModule extends PublishModule {
+
+  /**
+   * Runs [[revapiCliVersion Revapi CLI]] on this module's archives.
+   */
+  def revapi(args: String*): Command[Unit] = Task.Command {
+    val workingDir = millSourcePath
```
lihaoyi, line 19: It seems that this is only used for caching according to the upstream code linked below; maybe we can use `Task.dest` instead?

lihaoyi:
Some nits but otherwise looks good!

scalalib/src/mill/javalib/revapi/RevapiModule.scala
ajaychandran: Updated.
This should now allow for specifying relative paths for any extension [output](https://github.com/revapi/revapi/blob/69445626881347fbf7811a4a78ff230fe152a2dc/revapi-reporter-file-base/src/main/resources/org/revapi/reporter/file/default-schema.json#L16).

ajaychandran: Not for `Agg` since it does not extend `Seq`.

scalalib/src/mill/javalib/revapi/RevapiModule.scala
ajaychandran: Updated.
This should now allow for specifying relative paths for any extension [output](https://github.com/revapi/revapi/blob/69445626881347fbf7811a4a78ff230fe152a2dc/revapi-reporter-file-base/src/main/resources/org/revapi/reporter/file/default-schema.json#L16).

lihaoyi (approved):

PR #3975, opened 2024-11-15
Handle case of .mill-version with Windows line ends

sideeffffect:
This is a precaution for cases when the `.mill-version` file can contain a Windows line end `\r\n` instead of the typical Unix line end `\n`. Without this fix, in such situation the script breaks in a spectacular and surprising way.

The error it was giving looked like this:
```sh
$ mill __.compile
curl: (3) URL using bad/illegal format or missing URL
```
Where `mill` is pointing to the wrapper script that I'm changing in this PR

### DIFF
--- a/mill
+++ b/mill
@@ -12,9 +12,9 @@ fi
 
 if [ -z "$MILL_VERSION" ] ; then
   if [ -f ".mill-version" ] ; then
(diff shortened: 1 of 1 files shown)


lihaoyi:
Can you paste the error message you get when the issue this PR solves occurs?

sideeffffect:
done

lihaoyi:
Thanks!

PR #3976, opened 2024-11-15
installation-ide.adoc: mill.bsp.BSP/install

sideeffffect:
Instructions to run `mill.bsp.BSP/install`. Without it, IntelliJ won't pick up the BSP model -- it won't show it as one of the possible ways how to import the project.

### DIFF
--- a/docs/modules/ROOT/pages/cli/installation-ide.adoc
+++ b/docs/modules/ROOT/pages/cli/installation-ide.adoc
@@ -62,6 +62,12 @@ as a drop-in replacement for `./mill` that supports running on all major platfor
 Mill supports IntelliJ and VSCode, both via the standard
 https://build-server-protocol.github.io/[Build Server Protocol]
 
+The first thing you need to do is to run this command to generat the BSP configuration files:
+[source,bash]
+----
+./mill mill.bsp.BSP/install
+----
+
 === IntelliJ
 
 To use Mill with IntelliJ, first ensure you have the free


lihaoyi:
I havent needed to do this for a clean checkout for maybe a year now

lihaoyi:
do you have a repro i can look at where running this command is necessary?

sideeffffect:
When you open the project as just a mere directory, IntelliJ doesn't prompt you to import -- regardless of whether with BSP or Maven or anything. You just see this and nothing happens.
[scala-mill-experiment.zip](https://github.com/user-attachments/files/17782727/scala-mill-experiment.zip)
image

lihaoyi:
The issue seems to be that IntelliJ uses the presence a `./mill` bootstrap script in order to detect it as a Mill build. After adding such a bootstrap script the intellij auto import seems to work

lihaoyi:
We can probably send a patch upstream to intellij-scala to make it a bit more flexible. In the meantime, this command may be useful to fix some of the misbehaviors I've seen when intellij gets into a bad state. Let me try to reproduce them to see

sideeffffect:
> presence [of] a `./mill` bootstrap script in order to detect it as a Mill build

Would you say it is more robust to rather detect Mill via detecting
 * `build.mill` or
 * `build.mill.scala` or
 * `mill.scala`
 * (did I forget any other case?)

lihaoyi:
@sideeffffect should be these cases

* `./mill`
* `./build.mill`
* `./build.mill.scala`

sideeffffect:
We may not need this PR if this gets merged
 * https://github.com/JetBrains/intellij-scala/pull/674

lefou:
Superseded by #3989

Issue #3928, opened 2024-11-09
First Class Python Support (4000USD Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 4000USD bounty on this issue, payable by bank transfer on merged PRs implementing this. [Standard bounty terms](https://github.com/orgs/com-lihaoyi/discussions/6) apply

---

Python is common in industry, and lacks good build tools. Mill could a be a good tool to help manage large multi-module Python projects  (see [What Makes Mill Unique?](https://mill-build.org/mill/comparisons/unique.html)) providing automatic caching and parallelism for the various workflows (dependency resolution, typechecking, testing, packaging, publishing) that are core to the local development experience to help keep them fast and responsive. 

The goal of this ticket is to generate a set of Python build examples that match the Java/Scala/Kotlin equivalents. We already have an [example PythonModule](https://mill-build.org/mill/extending/example-python-support.html) for demo purposes, but we would need to flesh it out using the equivalent Python tools and libraries.

For the purposes of this ticket, each examples should match as closely as possible the Java/Scala/Kotlin equivalents, to provide a useful minimal-but-still-educational code example, along with associated english documentation and explanations. You should read through the relevant sections of the [Building Java with Mill](https://mill-build.org/mill/javalib/intro.html) documentation before proceeding, even if you don't know Java, just to get a feel for what the documentation and examples for each section should cover. 

1. [x] `example/pythonlib/basic/` (500USD)
    * [x] `1-simple/`: A minimal Python module demonstrating typechecking/running/testing/pex
        * We should also demonstrate how to use the Python REPL hooked up to your module
    * [x] `2-custom-build-logic/`: A Python module with custom build logic
    * [x] `3-multi-modules/`: Multiple inter-related Python modules

2. [x] `example/pythonlib/dependencies/` (500USD)
    * [x] `1-pip-deps/`
    * [x] `2-unmanaged-wheels/`
    * [x] `3-downloading-unmanaged-wheels/`
    * [x] `4-repository-config/`: examples of how to use alternate PyPI mirrors or repositories

3. [x] `example/pythonlib/linting/` (1000USD)
    * ~1-black/`: equivalent to `1-scalafmt` but using Black~ (ruff handles this)
    * [x] `2-code-coverage/`: not sure what the popular Python code coverage lib is, but they should have something
    * [x] `3-ruff/`: using Ruff to lint the Python code

4. [x] `example/pythonlib/module/` (500USD)
[...]

PR #3986, opened 2024-11-18
[WIP] Fixes: #3928 Added First Class Python Support

himanshumahajan138:
# Pull Request 
Added First Class Python Support
Fixes: #3928

## Description
Adds Example for Python `1-simple`, `2-custom-build-logic` and `3-multi-module` Skeleton
**Documentation Not Updated Yet!**

## Related Issues

## Checklist
- [x] New Example for Python `1-Simple`
- [x] New Example for Python `2-custom-build-logic`
- [x] New Example for Python `3-multi-module`
- [ ] Updated Documentation

## Status
Require Reviews and Further fine-tuning


himanshumahajan138:
@lihaoyi Good Wishes Sir!

Actually, i have done some workaround after carefully checking other examples and everything is exact same but the way i used for testing is far different from the other examples.

I don't know this is right or not (as per mill structure), Require your Guidance 

Please Review and Guide me...

himanshumahajan138:
@lihaoyi, If Possible could you please review this PR coz further work requires your attention so please once provide the updates required so that i could bring this to mark...

pythonlib/src/mill/pythonlib/PythonModule.scala
```diff
@@ -69,11 +78,47 @@ trait PythonModule extends Module {
...
+        "RESOURCES" -> resources().map(_.toString).mkString(":")
+      ),
+      stdin = args.value,
+      stdout = os.Inherit
+    )
+  }
+
+  trait PythonTests extends PythonModule {
+
+    def pythonUnitTests: T[Unit] = Task {
```
lihaoyi, line 109: This should be a `Task.Command`

lihaoyi: Let's also call it `def test` for consistency with the other languages

pythonlib/src/mill/pythonlib/PythonModule.scala
```diff
@@ -69,11 +78,47 @@ trait PythonModule extends Module {
...
+        "SOURCES" -> allSourceFiles().map(_.toString).mkString(":"),
+        "RESOURCES" -> resources().map(_.toString).mkString(":")
+      ),
       stdout = os.Inherit
     )
 
     PathRef(pexFile)
   }
 
+  def pythonRepl(args: mill.define.Args) = Task.Command {
```
lihaoyi, line 92: This should probably be `exclusive = true`

lihaoyi: We can also just call it `repl` for consistency with Scala

example/pythonlib/basic/3-multi-module/build.mill
```diff
@@ -0,0 +1,39 @@
+//// SNIPPET:BUILD
```
lihaoyi, line 1: We don't need these snippet comments here since we're not sharing example code with the other languages

example/pythonlib/basic/1-simple/build.mill
```diff
@@ -1,27 +1,59 @@
 package build
 import mill._, pythonlib._
 
-// Docs for PythonLib
 object foo extends PythonModule {
-  object bar extends PythonModule {
-    def pythonDeps = Seq("pandas==2.2.3", "numpy==2.1.3")
+  def pythonDeps = Seq("pandas==2.2.3", "numpy==2.1.3")
```
lihaoyi, line 5: Let's get rid of pandas and numpy in these examples; they're really heavyweight packages and take a long time to download and install. We use HTML templating in our Java/Scala/Kotlin examples, so let's do so here as well. Jinja2 seems like a reasonable choice

example/pythonlib/basic/1-simple/build.mill
```diff
@@ -1,27 +1,59 @@
...
+
+  object test extends PythonTests {
+    def moduleDeps = Seq(foo)
+    def mainFileName: T[String] = Task { "test.py" }
```
lihaoyi, line 9: What does the test module require a `mainFileName`? Shouldn't the test framework discover the files automatically?

example/pythonlib/basic/1-simple/build.mill
```diff
@@ -1,27 +1,59 @@
...
 import mill._, pythonlib._
 
-// Docs for PythonLib
 object foo extends PythonModule {
-  object bar extends PythonModule {
-    def pythonDeps = Seq("pandas==2.2.3", "numpy==2.1.3")
+  def pythonDeps = Seq("pandas==2.2.3", "numpy==2.1.3")
+
+  object test extends PythonTests {
+    def moduleDeps = Seq(foo)
```
lihaoyi, line 8: The test module should not need an explicit dependency on the enclosing parent module. It should be automatic like it is in Java/Scala/Kotlin

example/pythonlib/basic/1-simple/build.mill
```diff
@@ -1,27 +1,59 @@
 package build
 import mill._, pythonlib._
 
-// Docs for PythonLib
 object foo extends PythonModule {
-  object bar extends PythonModule {
-    def pythonDeps = Seq("pandas==2.2.3", "numpy==2.1.3")
+  def pythonDeps = Seq("pandas==2.2.3", "numpy==2.1.3")
+
+  object test extends PythonTests {
```
lihaoyi, line 7: This should have an example of a third-party dependency being used in the test suite as well

himanshumahajan138:
@lihaoyi tbh all reviews are the expected ones coz i myself was not satisfied with this way of testing and want to acheive something like in other examples.


But the problem is with docs i didn't found anything related to this to get a learning spike...


If possible could you please provide me some docs link or guide on this please...

himanshumahajan138:
closing in favour of #4000

PR #3987, opened 2024-11-18
align SonatypeCentralPublishModule with SonatypePublisher

GeorgOfenbeck:
In the current version the SonatypeCentralPublishModule has non intented behaviour with PublishModule.defaultGpgArgs.

It will default to the args without a passphrase and afterwards ignore the passphrase env. var as a gpgarg is already set.
To get the "correct" behaviour one has to override the args with an empty string to then pick up the "PGP_PASSPHRASE" env. arg.

Also changed the env. name to match the one from SonaTypePublisher "PGP_PASSPHRASE" -> "MILL_PGP_PASSPHRASE".

Currently pgp and gpg are getting used both in terms of naming - guessing going for gpg in the future might be more clear.


lihaoyi:
Thanks!

PR #3990, opened 2024-11-19
Add a new doc page `Why Mill? / How Fast Does Java Compile?`

lihaoyi:


lefou (commented):
Interesting read. Would make a nice blog post. Do you strive for some specific conclusion regarding Mill? 

I think a lot of that "overhead" is not spend in wasteful work but in hose-keeping of dependencies and state, such that all possible build results are always up-to-date.

lihaoyi:
We don't have a blog section on the doc-site so I put it under `Why Mill?`. If that gets overcrowded we can break out a blog section later, but to some extent this post is an exploration of one facet of the "Why Mill?" question. I wanted it to go on the Mill docsite so it can draw attention and mindshare, which would be less effective if put on my personal blog 

The build tool overhead is definitely doing _something_, it's just a question of how much in can be optimized. For example, for Mill a lot of overhead that I am aware of is (a) starting the JVM client (b) starting the socket connection (somehow jprofiler says this takes 200ms???) and (c) serializing and de-serializing JSON files on disk. All of these things are necessary now, but they can all definitely be optimized if we really wanted to.

PR #3991, opened 2024-11-19
Use function from geny to pump streams

lolgab:


main/api/src/mill/api/StreamSupport.scala
```diff
@@ -22,15 +22,7 @@ trait StreamSupport {
...
-    val buffer = new Array[Byte](4096)
-    while ({
-      src.read(buffer) match {
-        case -1 => false
-        case n =>
-          dest.write(buffer, 0, n)
-          true
-      }
-    }) ()
+    geny.Internal.transfer(src, dest)
```
lihaoyi, line 25: `geny.Internal.transfer` calls `.close` on `src`. Not sure if we actually want to do that given where this function is used?

PR #3992, opened 2024-11-19
Some improvements to pythonlib

jodersky:
This is in view of #3928, specifically some essential things required for the first task, `example/pythonlib/basic/`.

You can look at commits individually if you like. The gist of this PR is to improve the scaffolding we currently have to unblock the previously mentioned tasks. It can be summarized by the following points (roughly each corresponding to one commit):

- [x] make `run` task the default, following the same convention used in `ScalaModule` and `JavaModule`
- [x] add a command to run an interactive REPL. This is named `console`, again following the conventions of scalalib
- [x] rework the way source files are handled:
  - support multiple source directories, again following similar conventions of scalalib
  - instead of aggreting python scripts in one syntetic directory, keep them where they are and instead manipulate PYTHONPATH or pass in parameters to various directories where appropriate
- [x] define a module for unit tests, similar to the TestModule of scalalib


pythonlib/src/mill/pythonlib/PythonModule.scala
```diff
@@ -25,52 +44,65 @@ trait PythonModule extends Module {
...
-    gatherScripts(Task.traverse(moduleDeps)(_.sources)().zip(moduleDeps))
-
     os.call(
-      (pythonExe().path, sources().path / mainFileName(), args.value),
-      env = Map("PYTHONPATH" -> Task.dest.toString),
-      stdout = os.Inherit
+      (pythonExe().path, script().path, args.value),
+      env = Map(
+        "PYTHONPATH" -> transitiveSources().map(_.path).mkString(":"),
+        "PYTHONPYCACHEPREFIX" -> (T.dest / "cache").toString
```
jodersky, line 58: I'm doing this for now to avoid polluting source directories, and hence both `run` and `console` won't share any caches. In the future, it could be interesting to see if this can be used in a global cache somehow, so that we don't need to recreate it for every task.

himanshumahajan138:
@jodersky TBH this is what i was looking for, Great !!!

Also i want to add something like the way i used Repl i think that is also good See #3986 

And the way of testing is now professional and matching with other examples (i didn't liked my way of testing)

So i think i should wait till this PR gets Merged and after wards i will work further on #3986 till then let me copy your changes and work around with the same pattern 

Thanks...

himanshumahajan138:
@jodersky ya that's great we will work together to bring the professional python first class support to mill 

Team Work is the Dream Work !!!

pythonlib/src/mill/pythonlib/PythonModule.scala
himanshumahajan138: btw have this worked ???

when i was doing the same thing like giving paths to the files which are module dependencies but getting error like module doesn't exist

any correction ???

pythonlib/src/mill/pythonlib/PythonModule.scala
```diff
@@ -25,55 +44,72 @@ trait PythonModule extends Module {
...
     )
   }
 
+  override def defaultCommandName(): String = "run"
+
+  /**
+   * Opens up a Python console with your module and all dependencies present,
+   * for you to test and operate your code interactively.
+   */
+  def console(): Command[Unit] = Task.Command(exclusive = true) {
```
himanshumahajan138, line 71: we should use repl i guess coz Li Hayoi said this in my PR

jodersky: hmm, well there's also a "console" task https://github.com/com-lihaoyi/mill/blob/5e3b6bf1c4a8c2c62bbaeb9398c311f252ee3f71/scalalib/src/mill/scalalib/ScalaModule.scala#L431.

The difference is that `repl` uses the ammonite interpreter, and `console` the basic scala interpreter. If we're trying to keep similarity, I would suggest to use `console`, since it's also the more bare-bones solution. Maybe we could have `repl` use something like `ipython`?

pythonlib/src/mill/pythonlib/PythonModule.scala
himanshumahajan138: https://github.com/com-lihaoyi/mill/tree/422ef4094600af1cc4c44027a8357dcd5e37bb01/example/extending/python/3-python-module-deps

Check this example tbh i haven't tested but i think this will not work coz when i was adding python support at that time i was getting module import error and then i tried to add `__init__.py` files it still didn't worked

if you have tested this already and it's working then i think this would be the best approach to handle this

pythonlib/src/mill/pythonlib/PythonModule.scala
himanshumahajan138: ya we can create two one is repl(ipython) and other is console

it's great !!!

pythonlib/src/mill/pythonlib/PythonModule.scala
himanshumahajan138: PS: i was copying the same structure like copying the files from on root folder like `foo/bar/src/bar.py` and `foo/src/foo.py` and then maintain the same structure in the `Task.dest`
and then call them in the `qux.py` where i include modules as :
`from foo.src.foo import someFunction` and same for bar `from foo.bar.src.bar import someFunction`

himanshumahajan138: and this is what we call `Experience` 

Hat's off, TBH i haven't even thought this

this is the best approach GREAT!!!

himanshumahajan138:
`./mill mill.scalalib.scalafmt.ScalafmtModule/`
`./mill mill.javalib.palantirformat.PalantirFormatModule/`
`./mill mill.kotlinlib.ktlint.KtlintModule/`

@jodersky  these commands will fix the lint error from the root

jodersky:
Alright, I think this is ready for review now.

example/pythonlib/basic/1-simple/build.mill
```diff
@@ -9,8 +9,15 @@ object foo extends PythonModule {
   def pythonDeps = Seq("numpy==2.1.3")
 }
 
-object qux extends PythonModule {
+object qux extends PythonModule { q =>
   def moduleDeps = Seq(foo, foo.bar)
+
+  object test extends PythonTests with TestModule.Unittest
+  object test2 extends PythonTests with TestModule.Pytest {
```
lihaoyi, line 16: For this intro example, just one `test` module is fine. Having examples of other test suites can come later in the `example/pythonlib/testing/` section

example/pythonlib/basic/1-simple/qux/src/main.py
```diff
@@ -1,11 +1,11 @@
 #!/usr/bin/python3
 import numpy as np
 
-from foo.src.foo import data
-from foo.bar.src.bar import df
+from foo import data
+from bar import df
```
lihaoyi, line 5: This looks great!

example/pythonlib/basic/1-simple/qux/src/main.py
```diff
@@ -1,11 +1,11 @@
 #!/usr/bin/python3
 import numpy as np
```
lihaoyi, line 2: I left this comment on the other PR as well, but let's swap out `numpy` for `jinja2` as our example third-party library. It is a lot more light weight and better mirrors what the other examples are doing. We should have some scientific/ml python stack examples somewhere but maybe not in the first example

example/pythonlib/basic/1-simple/qux/test/src/test_dummy.py
```diff
@@ -0,0 +1,17 @@
+import unittest
```
lihaoyi, line 1: This file should exercise some of the logic in the associated `PythonModule` rather than being entirely standalone

pythonlib/src/mill/pythonlib/PythonModule.scala
```diff
@@ -21,59 +40,81 @@ trait PythonModule extends Module {
...
 
-  def typeCheck: T[Unit] = Task {
-    Task.traverse(moduleDeps)(_.typeCheck)()
-
-    os.call(
-      (pythonExe().path, "-m", "mypy", "--strict", sources().path),
-      stdout = os.Inherit,
-      cwd = T.workspace
+  // TODO: right now, any task that calls this helper will have its own python
+  // cache. This is slow. Look into sharing the cache between tasks.
```
lihaoyi, line 44: What we've done elsewhere is provide a `Task(persistent = true)` whose sole purpose is to return a `os.Path` for people to scribble over. Maybe we could do that here to provide the `PYTHONPYCACHEPREFIX`?

pythonlib/src/mill/pythonlib/PythonModule.scala
```diff
@@ -21,59 +40,81 @@ trait PythonModule extends Module {
...
   }
 
-  def gatherScripts(upstream: Seq[(PathRef, PythonModule)]) = {
-    for ((sourcesFolder, mod) <- upstream) {
-      val destinationPath =
-        os.pwd / mod.millSourcePath.subRelativeTo(mill.api.WorkspaceRoot.workspaceRoot)
-      os.copy.over(sourcesFolder.path / os.up, destinationPath)
-    }
+  def typeCheck: T[Unit] = Task {
+    Task.traverse(moduleDeps)(_.typeCheck)()
```
lihaoyi, line 59: Looking at this again, since we do not actually use the upstream `typeCheck` results, should we skip this line? That would allow all typechecking to happen in parallel even for dependent modules, which would be nice

pythonlib/src/mill/pythonlib/PythonModule.scala
```diff
@@ -21,59 +40,81 @@ trait PythonModule extends Module {
...
-  def typeCheck: T[Unit] = Task {
-    Task.traverse(moduleDeps)(_.typeCheck)()
-
-    os.call(
-      (pythonExe().path, "-m", "mypy", "--strict", sources().path),
-      stdout = os.Inherit,
-      cwd = T.workspace
+  // TODO: right now, any task that calls this helper will have its own python
+  // cache. This is slow. Look into sharing the cache between tasks.
+  protected def runPythonExe(args: Task[Seq[String]]): Task[Unit] = Task.Anon {
```
lihaoyi, line 45: Rather than a method taking and returning `Task`s, we should have a zero-arg method returning a `Task` that returns a value that has more methods we can call. See how this is done for `runner().run(...)` in `JavaModule`s and `ScalaModule`s

example/pythonlib/basic/1-simple/build.mill
jodersky: ah, I don't think the example is representative. @himanshumahajan138 will use his own examples and structure in his upcoming PR, if I understood correctly

lihaoyi:
Looks pretty good overall. Left some comments. @jodersky you mentioned you looked at how Bazel handles Python, if in that research you found anything relevant to this PR can summarize whatever you learned that's relevant in the PR description

example/pythonlib/basic/1-simple/build.mill
lihaoyi: Ah ok, that's fine then

pythonlib/src/mill/pythonlib/PythonModule.scala
jodersky: indeed. We'll also need to look into generating stub files so that typechecking works when importing other modules defined in the same build. Right now `mypy` errors out when refering to other modules, lest we explicitly ignore import definitions. I think however that this can be done in a second PR

pythonlib/src/mill/pythonlib/PythonModule.scala
jodersky: If we have a task that returns a PathRef, then later we write to it after the task has run, wouldn't that mess with cache invalidation?

example/pythonlib/basic/1-simple/qux/src/main.py
himanshumahajan138: @lihaoyi Actually, i am carrying this after merging to finish line by adding examples and there i will add all the examples accordingly the reviews you made into my PR

pythonlib/src/mill/pythonlib/PythonModule.scala
lihaoyi: It isn't really kosher but it works. But I think we can punt on it for this PR: populating the `__pycache__` isn't particularly slow so I doubt anyone will notice, and if it proves to be problematic we can optimize it later

himanshumahajan138:
@jodersky i think we should work on linting before deps coz deps don't require much logical and backend code but linting is far harder and more complicated and in that case, experience is more required 😉

Rest is You's Choice...

pythonlib/src/mill/pythonlib/PythonModule.scala
jodersky: ok, I'll let it be as is for now then

jodersky:
> @jodersky i think we should work on linting before deps coz deps don't require much logical and backend code but linting is far harder and more complicated and in that case, experience is more required 😉

I'm not sure. I think dependencies are a very important feature, and they aren't all that easy when you consider that we'll need more than just building a pip command line.

himanshumahajan138:
> > @jodersky i think we should work on linting before deps coz deps don't require much logical and backend code but linting is far harder and more complicated and in that case, experience is more required 😉
> 
> I'm not sure. I think dependencies are a very important feature, and they aren't all that easy when you consider that we'll need more than just building a pip command line.
> 
> For example, we'll need to deal with things like installing wheels. Also, there are a couple of "quality-of-life" features that I'd like to include:
> 
> * ability to specifcy dependencies via an existing `requirements.txt` file (this is inspired by pants' python support)
> * on dependency conflicts, mention which modules requested the conflicting version (right now, since we're only building a pip command line, it's really hard to find out what module is the culprit)

Ok that's great afterall this all wil be available for mill 😊

jodersky:
@lihaoyi, I think I addressed all your comments. Would you like to go over it once more? Regarding the python example, it will all be reworked by @himanshumahajan138.

lihaoyi:
@jodersky looks great, I'll merge it and we can continue iterating on it after

PR #3996, opened 2024-11-20
Docs nit 1

myyk:
Fixes: https://mill-build.org/mill/extending/example-typescript-support.html#_npm_dependencies_and_bundling

image


lihaoyi:
Thanks!

Issue #3587, opened 2024-09-21
Make Agg[T] an alias for Seq[T]


lihaoyi:
`Agg` was a good idea at the time, but 7 years later I don't think the bespoke complexity is paying off. We should move all user facing APIs to `Seq` and make `Agg` an alias for backwards source compatibility

This is a bincompat breaking change and will need to go into 0.13.0

PR #3998, opened 2024-11-20
pythonlib: improved dependency management and examples

jodersky:
Make the dependency management more flexible:

- [x] allow users to re-use `requirements.txt` files (inspired by [the same pants feature](https://www.pantsbuild.org/dev/docs/python/overview/third-party-dependencies#requirementstxt))
- [x] separate library and tool requirements
  - library deps are transitive, and will in the future be included as requirements in distributions
  - tool requirements are not transitive and won't be included in distributions. They're used for things like type checkers and bundlers (which can all be extracted into separated traits later on if we want to).
- [x] support referencing wheel files directly
- [x] allow user-defined PYTHONPATH
- [x] custom package repos
-  ~lockfile support~ Removing this from this pull request, since it's likely a bit more complex than I initially thought. Pants delegates this behavior to PEX, but we should see if this can be done with more basic tools, or we decide to rely on pex for everything.
- [x] examples for all the above

Part of #3928


himanshumahajan138:
@jodersky what you think should i add repl or we use only console ???

himanshumahajan138:
@jodersky Sir I have done the Basic example and if you want/allow i can help in writing examples for this also...

jodersky:
thanks, however I already started working on the examples

example/pythonlib/dependencies/1-pip-deps/build.mill
```diff
@@ -0,0 +1,29 @@
+//// SNIPPET:BUILD
```
lihaoyi, line 1: For example doc/test files that don't inherit from an upstream example (i.e. they don't share the same relative path starting from `pythonlib` or `scalalib`), you don't need to do these `//// SNIPPET:` things. These are only useful if there is an upstream example that you are splicing in customizations into

example/pythonlib/dependencies/4-downloading-unmanaged-wheels/build.mill
```diff
@@ -0,0 +1,44 @@
...
+//// SNIPPET:END
+
+/** Usage
+
+> ./mill run
+Hello, world!
+
+*/
+
+// Tasks like `unmanagedClasspath` and `pythonDeps` are cached, so your wheel is downloaded only
```
lihaoyi, line 32: `unmanagedClasspath` should probably be `unmanagedWheels`

example/pythonlib/dependencies/4-downloading-unmanaged-wheels/build.mill
```diff
@@ -0,0 +1,44 @@
...
+
+object `package` extends RootModule with PythonModule {
+  def unmanagedWheels = Task {
+    val name = "jinja2-3.1.4-py3-none-any.whl"
+    os.write(
+      Task.dest / name,
+      requests.get.stream(
+        s"https://github.com/pallets/jinja/releases/download/3.1.4/$name"
+      )
+    )
```
lihaoyi, line 18: Probably would look prettier if we pull the whole URL out

```scala
val name = "jinja2-3.1.4-py3-none-any.whl"
val url = s"https://github.com/pallets/jinja/releases/download/3.1.4/$name"
os.write(Task.dest / name, requests.get.stream(url))
```

example/package.mill
```diff
@@ -65,7 +65,10 @@ object `package` extends RootModule with Module {
     object basic extends Cross[ExampleCrossModule](build.listIn(millSourcePath / "basic"))
   }
   object pythonlib extends Module {
-    object basic extends Cross[ExampleCrossModule](build.listIn(millSourcePath / "basic"))
+    //TODO: commented-out to prevent doc-generation from failing. Re-enable this.
+    // object basic extends Cross[ExampleCrossModule](build.listIn(millSourcePath / "basic"))
+
+    object dependencies extends Cross[ExampleCrossModule](build.listIn(millSourcePath / "dependencies"))
```
lihaoyi, line 71: If we're going to use `ExampleCrossModule`, this doesn't inherit from any upstream examples, so we can leave out all the `////SNIPPET` things

example/pythonlib/basic/1-simple/build.mill
```diff
@@ -4,9 +4,9 @@ import mill._, pythonlib._
 // Docs for PythonLib
 object foo extends PythonModule {
   object bar extends PythonModule {
-    def pythonDeps = Seq("pandas==2.2.3", "numpy==2.1.3")
+    def pythonDeps = Agg("pandas==2.2.3", "numpy==2.1.3")
   }
-  def pythonDeps = Seq("numpy==2.1.3")
+  def pythonDeps = Agg("numpy==2.1.3")
```
lihaoyi, line 9: This change isn't really necessary since we're planning on getting rid of `Agg` in 0.13.0 and just leaving an alias for `Seq` for backwards compatibility https://github.com/com-lihaoyi/mill/issues/3587

example/pythonlib/dependencies/5-repository-config/build.mill
```diff
@@ -0,0 +1,51 @@
...
+import mill._, pythonlib._
+
+object foo extends PythonModule {
+
+  def pythonDeps = Agg(
+    "testpkg-jodersky==0.0.1" // a test package, only available on test.pypi.org
+  )
+
+  // override this task to add or replace the package indexes
+  def indexes = Task {
```
lihaoyi, line 17: Can leave out the `Task{}` wrapper here since it's inferred

example/pythonlib/basic/1-simple/build.mill
jodersky: Ah, didn't know that. Do you have a pointer as to what you mean with "I don't think the bespoke complexity is paying off"?

example/pythonlib/dependencies/1-pip-deps/build.mill
jodersky: I don't follow what you mean by "upstream example".

jodersky: ok, I've looked throught the ExampleCrossModule and friends implementation a bit more. If I understand correctly, the subclasses of `ExampleCrossModule` would copy an upstream's example code as-is, only replacing snippets. Thus, effectively, comments would be the same, and only code is actually changed.

I can understand the appeal of that in some situations, but it also feels very constrained to enforce the same wording for all examples. What's your recommendation on this @lihaoyi? Since we can't really always translate concepts from python to java, I think it may be easier to stick with plain examples that don't use scalalib as the upstream.

Also, since you created the `ExampleCrossModulePython`, let me know your thoughts @himanshumahajan138

jodersky: I've removed the `ExampleCrossModulePython` trait and all SNIPPETS in pythonlib in my latest commit. I'll merge this if CI passes, but will keep the last commit separate so we can revert if necessary. I'm still curious to hear your opinions.

himanshumahajan138: @jodersky sir i think this is not right,
@lihaoyi said that in the case of dependencies we don't inherit any docs from the upstream but in the case of basic we have same upstream docs (for java, kotlin) and the way kotlin extended the base java module is the same way i did for python 
see the docs there are no such docs now why ??? is this educational ???
please resolve this...

lihaoyi: @jodersky yeah basically it only allows sharing of coarse-grained blocks, though they csn be snything: docs, code, use comments

I'd say if it's too restrictive just drop the sharing for now, we can reconsolidate later if nexessary

jodersky: > see the docs there are no such docs now why ??? is this educational ???

Did I break something? If so that wasn't intentional

himanshumahajan138: Actually you removed the `examplecrossmodule` and this leaded to docs removal and more over you removed snippets like build dependency and all 

But why ?

I think you have to undo this for showing docs for basic examples

jodersky: see #4042

PR #4000, opened 2024-11-20
Fixed: #3928 [Basic] Re-Work for First Class Python Support

himanshumahajan138:
# Pull Request 
Added First Class Python Support [Basic Example]
Fixes: #3928

## Description
Rework for Python `1-simple`, `2-custom-build-logic` and `3-multi-module` Examples
follow up for PR #3992 

## Related Issues

## Checklist
- [x] New Example for Python `1-Simple`
- [x] New Example for Python `2-custom-build-logic`
- [x] New Example for Python `3-multi-module`
- [x] Updated Documentation

## Status
Completed Addition & Require Review!!!


himanshumahajan138:
@lihaoyi why `//// SNIPPET:DEPENDENCIES` or `//// SNIPPET:TREE` and more like this are not working for my python docs???

 i was looking in other examples but there these are available how ?

Please Explain about this...

himanshumahajan138:
@lihaoyi Ready for Review!

example/pythonlib/basic/1-simple/foo/src/foo.py
```diff
@@ -1,2 +1,22 @@
-import numpy as np
-data = np.array([10, 20, 30, 40, 50])
\ No newline at end of file
+#!/usr/bin/python3
```
lihaoyi, line 1: We don't need the shebang line for this script do we? I assume the `run` task will use the correct venv python binary automatically

example/pythonlib/basic/2-custom-build-logic/foo/src/foo.py
```diff
@@ -0,0 +1,21 @@
...
+
+def line_count() -> int:
+    file_path = next(
+        (
+            os.path.join(p, "line-count.txt")
+            for p in os.environ.get("PYTHONPATH", "").split(":")
+            if os.path.exists(os.path.join(p, "line-count.txt"))
+        ),
+        None,
+    )
```
lihaoyi, line 12: This is pretty convoluted; is there some other way we can manage the `resource/` folders in Python to do this? Maybe `importlib.resources` or `pkgutil.get_data`? We also should ensure whatever approach we take here works for all of `run`, `test`, and `assembly` in the example test above

example/pythonlib/basic/1-simple/build.mill
```diff
@@ -1,34 +1,139 @@
...
-// Docs for PythonLib
-object foo extends PythonModule {
-  object bar extends PythonModule {
-    def pythonDeps = Seq("pandas==2.2.3", "numpy==2.1.3")
-  }
-  def pythonDeps = Seq("numpy==2.1.3")
-}
+object foo extends PythonModule { f =>
+
+  def script: T[PathRef] = Task.Source { millSourcePath / "src" / "foo.py" }
```
lihaoyi, line 6: Let's rename this `def mainScript`, so it's a bit more clear what it is used for

example/pythonlib/basic/1-simple/build.mill
```diff
@@ -1,34 +1,139 @@
...
 import mill._, pythonlib._
 
-// Docs for PythonLib
-object foo extends PythonModule {
-  object bar extends PythonModule {
-    def pythonDeps = Seq("pandas==2.2.3", "numpy==2.1.3")
-  }
-  def pythonDeps = Seq("numpy==2.1.3")
-}
+object foo extends PythonModule { f =>
```
lihaoyi, line 4: What is this `f =>` thing for?

lihaoyi (commented):
Took a pass at reviewing this, @jodersky should probably take a pass as well

himanshumahajan138:
Let me resolve these reviews 😊

example/pythonlib/basic/1-simple/build.mill
himanshumahajan138: just a typo 

fixed!

example/pythonlib/basic/1-simple/build.mill
himanshumahajan138: fixed!

example/pythonlib/basic/1-simple/foo/src/foo.py
himanshumahajan138: yes it will work! 

Fixed!

example/pythonlib/basic/2-custom-build-logic/foo/src/foo.py
himanshumahajan138: actually i added `resources` in `transitiveSources` but in `JavaModule` it's added in the `classpath`.

Now what you say should i do the same approach (coz i think the way you are expecting will be solved by this)

himanshumahajan138: @jodersky sir we know that the file will be present in the `PYTHONPATH` coz i have added `resources` in the `transitiveDependencies` so now when we call the `importlib.resource(package='foo', resource='line-count.txt')` there we will get the `no module named 'foo' defined`

and when i ran the command to check what is available as packages and what files are present there so we got this :
```
Python Path: ['/root/test/mill/out/example/pythonlib/basic/2-custom-build-logic/local/testCached.dest/sandbox/run-1/foo/src', '/root/test/mill/out/example/pythonlib/basic/2-custom-build-logic/local/testCached.dest/sandbox/run-1/foo/src', '/root/test/mill/out/example/pythonlib/basic/2-custom-build-logic/local/testCached.dest/sandbox/run-1/foo/resources', '/root/test/mill/out/example/pythonlib/basic/2-custom-build-logic/local/testCached.dest/sandbox/run-1/out/foo/resources.dest', '/usr/lib/python311.zip', '/usr/lib/python3.11', '/usr/lib/python3.11/lib-dynload', '/root/test/mill/out/example/pythonlib/basic/2-custom-build-logic/local/testCached.dest/sandbox/run-1/out/foo/pythonExe.dest/venv/lib/python3.11/site-packages']

Files in the current directory: ['mill', 'out', 'foo', 'build.mill']
```

here you can see the `path/to/resources.dest` is available in `PYTHONPATH` but the file `line-count.txt` is not available we call using package name foo and not even if we call the file name straight as `with open (line-count.txt) as f`

So now what should we do please guide...

himanshumahajan138: final reviews:

- for using importlib the criteria is that we have to follow python module structure means including `__init__.py` for converting a folder to package.
- we know file will be present in `sys.path` so the easiest workaround is loop with file name 
- if multiple files with same name will be present then need to use class approach means search for package or parent folder name in path (this all will be based on user python code and query)

himanshumahajan138:
@lihaoyi one file named `/mill/main/api/src/mill/api/ClassLoader.scala` got some lint changes while running tests for lint fixture 

what should i do ???

himanshumahajan138:
@lihaoyi i have done some changes and the code you mentioned for importlib usage is also marked with two comments for usage  

please review to refine more

example/pythonlib/basic/2-custom-build-logic/foo/src/foo.py
lihaoyi: Lets go with `importlib.resource`. This seems to be the closest thing that Python has to JVM resources, which share the same restriction that the resources need to be on the classpath. It also has the same issue with the same file appearing multiple times in the classpath, so although that's a problem it's a problem we and others should be familiar with

I do not think we need `__init__.py` files as long as the pythonpath is correct

lihaoyi:
The issue with `//// SNIPPET:DEPENDENCIES` or `//// SNIPPET:TREE` were not working because this needs to be an `ExampleCrossModuleJava` which contains the template inheritance logic https://github.com/com-lihaoyi/mill/blob/b1f6be0c0c2817e8c623fac75c09e1a232c44b89/example/package.mill#L68. 

Let's go back to using template inheritance and seeing how much of the example docs we can share; as we proceed through the rest of the Python example docs, I expect there will be large amounts of overlap between the docs and explanations, so we should get used to using the templated documentation to begin wtih

himanshumahajan138:
Working 🤞

himanshumahajan138:
> The issue with `//// SNIPPET:DEPENDENCIES` or `//// SNIPPET:TREE` were not working because this needs to be an `ExampleCrossModuleJava` which contains the template inheritance logic
> 
> https://github.com/com-lihaoyi/mill/blob/b1f6be0c0c2817e8c623fac75c09e1a232c44b89/example/package.mill#L68
> 
> .
> Let's go back to using template inheritance and seeing how much of the example docs we can share; as we proceed through the rest of the Python example docs, I expect there will be large amounts of overlap between the docs and explanations, so we should get used to using the templated documentation to begin wtih

@lihaoyi Sir, one more guide please:
see i am doing some workaround to fix this error :
```
[1830] [16:25:41.812] ERROR (asciidoctor): target of xref not found: pythonlib/web-examples.adoc
[1830]     file: /root/test/mill/out/docs/devAntoraSources.dest/modules/ROOT/pages/pythonlib/intro.adoc
[1830]     source: /root/test/mill (branch: issue-3928-final <worktree> | start path: out/docs/devAntoraSources.dest)
[1830] [16:25:41.812] ERROR (asciidoctor): target of xref not found: pythonlib/build-examples.adoc
[1830]     file: /root/test/mill/out/docs/devAntoraSources.dest/modules/ROOT/pages/pythonlib/intro.adoc
[1830]     source: /root/test/mill (branch: issue-3928-final <worktree> | start path: out/docs/devAntoraSources.dest)
```

Now this is happening because we don't have these examples yet, but they are linked in `scalalib` original docs for `1-basic`
So, to tackle this problem temporarily i am doing this 

```
object pythonlib extends Module {
  object basic extends Cross[ExampleCrossModulePython](build.listIn(millSourcePath / "basic"))
}
  
trait ExampleCrossModulePython extends ExampleCrossModuleJava {
  override def lineTransform(line: String) = this.millModuleSegments.parts.last match {
    case "1-simple" => line
      .replace("xref:{language-small}/web-examples.adoc","link:")
      .replace("xref:{language-small}/build-examples.adoc","link:")
    case _ => line
  }
}
```

but this is not working i am not getting this, why???

please give some guide please...

pythonlib/src/mill/pythonlib/PythonModule.scala
```diff
@@ -22,9 +22,21 @@ trait PythonModule extends Module with TaskModule { outer =>
...
+
+  /**
+   * All Python source files in this module, recursively discovered from the source directories.
+   */
+  def allSourceFiles: T[Seq[PathRef]] = Task {
+    sources().flatMap(src => os.walk(src.path).filter(_.ext == "py").map(PathRef(_)))
+  }
+
+  /**
+   * The mainScript to run. This file may not exist if this module is only a library.
```
jodersky, line 37: It looks like you did a search-replace of `script` to `mainScript`. Any reason for this?

If you look at `man python3`, you'll see that they call the file to execute `script` everywhere. It's a detail of course, but I'd favor staying closer to nomenclature that is used in the python ecosystem itself.

himanshumahajan138: @lihaoyi Said 😅

himanshumahajan138: @jodersky BTW Sir Can you help me with this https://github.com/com-lihaoyi/mill/pull/4000#issuecomment-2495441017

jodersky: ok, that's fine by me. But be careful when replacing this term in the docs. Right now it says `the mainScript` to run`.

pythonlib/src/mill/pythonlib/PythonModule.scala
```diff
@@ -49,7 +61,7 @@ trait PythonModule extends Module with TaskModule { outer =>
    */
   def transitiveSources: T[Seq[PathRef]] = Task {
     val upstreamSources = Task.traverse(moduleDeps)(_.transitiveSources)().flatten
-    sources() ++ upstreamSources
+    sources() ++ resources() ++ upstreamSources
```
jodersky, line 64: It feels a bit strange if `resources()` is included in `transitiveSources()` but then isn't included in `sources()`. 

Come to think of it, do we actually need a `resources()` task? Could we include a `resouces/` directory in the default `sources()` task? I'm not sure which would be the "best" (aka least surprising) behavior here. @lihaoyi any thoughts on this?

pythonlib/src/mill/pythonlib/PythonModule.scala
```diff
@@ -22,9 +22,21 @@ trait PythonModule extends Module with TaskModule { outer =>
...
-   * The script to run. This file may not exist if this module is only a library.
+   * The folders where the resource files for this module live.
    */
-  def script: T[PathRef] = Task.Source { millSourcePath / "src" / "main.py" }
+  def resources: T[Seq[PathRef]] = Task.Sources { millSourcePath / "resources" }
+
+  /**
+   * All Python source files in this module, recursively discovered from the source directories.
+   */
+  def allSourceFiles: T[Seq[PathRef]] = Task {
```
jodersky, line 32: Do you know if this task will be used in `PythonModule` in the future? Right now I think it's only used in your example, so I'd suggest to remove it from here unless we can anticipate a use-case to keep it here

pythonlib/src/mill/pythonlib/PythonModule.scala
himanshumahajan138: i think just like we added `transitiveSources` in the we need to add the resources to the JVM resource pool this will also allow us to use the `importlib.resource` 

also @jodersky sir tell me one thing why i we get error of `foo is not a package` if we do something like this 
`importlib.resource.open_file(package='foo', resource='line-count.txt')`

generally when we get `___ is not a package` this means that this folder don't include `__init__.py` or its not included in the `PYTHONPATH` and in our case we have used the `PYTHONPATH` Method but still we are getting same error 

any Thoughts on this ???

pythonlib/src/mill/pythonlib/PythonModule.scala
himanshumahajan138: btw for now its only used for example but may be in future it come to work (not sure)

jodersky: Let's remove it from `PythonModule` for now then. I think a good principal to follow is to keep the modules in `pythonlib` minimal, since that's what we'll need to maintain. It's too early to see what will be used and we can always move this task definition back into here if we need to.

himanshumahajan138: ok sir i will !

could you please once revisit the replies as some of them need attention

himanshumahajan138:
@jodersky Sir, I have something surprising for you tell me why these tests don't run ?

```
> mill foo._.compile # Runs `compile` for all direct sub-modules of `foo`

> mill foo.__.test # Runs `test` for all transitive sub-modules of `foo`

> mill {foo,bar}.__.testCached # Runs `testCached` for all sub-modules of `foo` and `bar`

> mill __.compile + foo.__.test # Runs all `compile` tasks ad all tests under `foo`.

```
these are from `3-multi-module` and they all fail coz the path its asking for is `foo/test/src` which don't exist and they throw error

Any Thoughts ???

himanshumahajan138:
@lihaoyi Please see the latest commit(https://github.com/com-lihaoyi/mill/pull/4000/commits/[f255cfa](https://github.com/com-lihaoyi/mill/pull/4000/commits/f255cfa5d848dc4a899305c4bde6cb0fa51caed2)5d848dc4a899305c4bde6cb0fa51caed2) and see how i am dealing with the documentation.

Furthermore the test cases will fail for `pythonlib` coz of above mentioned error(https://github.com/com-lihaoyi/mill/pull/4000#issuecomment-2495647958) 

So now the problem is with the testcases already present in the upstream `ExampleCrossModuleJava` but this should not be present

TBH i am surprised to see why its asking for `foo/test/src` Path please help and guide me....

himanshumahajan138:
@lihaoyi Sir Pending comments are there requiring your attention please resolve...

pythonlib/src/mill/pythonlib/PythonModule.scala
lihaoyi: I don't actually know which way would be better, but let's keep `sources` and `resources` separate tasks for now. I imagine there will be things people want to do to sources that they do not want to do to resources, e.g. you want to typecheck your sources but you do not want to typecheck your resources, or if you want to obfuscate your python sources you probably do not want to obfuscate random `.py` files in resources. 

While this stuff can generally be worked around on a case-by-case basis, I think having `sources` and `resources` be separate is worth maintaining

lihaoyi: @himanshumahajan138 you need to figure out what `importlib.resources` and `pkgutil.get_data` work, do what needs to be done, and explain to use why it needs to be done that way

jodersky: As part of #3998, I've reworked the way the pythonpath is created.

https://github.com/com-lihaoyi/mill/blob/abf23050ff9e659cbf4fee0d80486b19e504bb6a/pythonlib/src/mill/pythonlib/PythonModule.scala#L51-L106

This hierarchy could be used to include resources as well (see the TODO comment on line 81 of the snippet)

himanshumahajan138: yaa i checked and was also doing the same thing by coping your code from commits 😅

himanshumahajan138:
@jodersky sir if possible please give some light on this comment https://github.com/com-lihaoyi/mill/pull/4000#issuecomment-2495647958

example/pythonlib/basic/2-custom-build-logic/foo/src/foo.py
himanshumahajan138: @lihaoyi you were right we just need to structure the directory for PYTHONPATH it's done now 

FIXED!

pythonlib/src/mill/pythonlib/PythonModule.scala
himanshumahajan138: FIXED!

pythonlib/src/mill/pythonlib/PythonModule.scala
himanshumahajan138: FIXED!

pythonlib/src/mill/pythonlib/PythonModule.scala
himanshumahajan138: FIXED!

himanshumahajan138:
@lihaoyi @jodersky 
Ready for Review
Let's wrap it up!

pythonlib/test/src/mill/pythonlib/HelloWorldTests.scala
```diff
@@ -16,22 +16,22 @@ object HelloWorldTests extends TestSuite {
...
+    //   val baos = new ByteArrayOutputStream()
+    //   val eval = UnitTester(HelloWorldPython, resourcePath, outStream = new PrintStream(baos))
 
-      val Right(result) = eval.apply(HelloWorldPython.qux.run(Args()))
+    //   val Right(result) = eval.apply(HelloWorldPython.qux.run(Args()))
 
-      assert(baos.toString() == "Hello,  Qux!\n")
-    }
+    //   assert(baos.toString() == "Hello,  Qux!\n")
+    // }
```
lihaoyi, line 34: What's the error here?

himanshumahajan138: Actually it's very surprising it was giving the right output as "hello,  Qux!" But i don't know why assert was not working so then after trying manual debugging and some fixes it was not getting fixed 

Hence, i commented 😅

I will show you the exact error in some minutes...

jodersky: It's probably due to me streaming `pip`'s progress to stdout https://github.com/jodersky/mill/blob/3fb0730c08f76c1991c6424d56614555757aa1a7/pythonlib/src/mill/pythonlib/PythonModule.scala#L47 `stdout = os.Inherit`. I think it's useful for the user to see this output, but the caveat is that you can't match exactly on output anymore. You can update the test to be `assert(baos.toString().contains("Hello,  Qux!\n")), like I've also done in my PR.

himanshumahajan138: Fixed !

himanshumahajan138:
@lihaoyi and @jodersky please take a pass and then finalize this PR as this will open the gates for all the other Python Support PR's

lihaoyi:
I think this looks good enough for now. @himanshumahajan138 I'll merge this and close out the bounty using your existing banking details.

Going forward, I'll defer to @jodersky on how he wants to structure the work, since he's the one focusing on the Python language toolchain in Mill in the medium-long term. @himanshumahajan138 make sure to coordinate to ensure you guys don't duplicate work, since it's no longer only you working on this area of the bounties

himanshumahajan138:
@lihaoyi Ya Sure !

I Always try to ask @jodersky Sir for conflicts and resolutions we are working in manner such that we don't face duplicate and conflicting work 

Thanks for Merge...

Issue #3730, opened 2024-10-14
Improve framework support for cached multi-thread-safe workers


lihaoyi:
`Task.Worker`s in Mill are intended for managing mutable state, in-memory/on-disk/in-subprocesses. Before Mill was parallel that was easy, but Mill 0.12.0 is parallel-by-default. Thus defining a worker requires that the user return a multi-thread-safe object, but Mill provide no help in doing so. 

The basic problem is that producing a multi-thread-safe object containing mutable state that is hard, especially if you don't just `synchronize` all its members (e.g. because you don't want to only allow only 1 scala module to be compiled at a time), and if you want proper lifecycle management (e.g. caching and re-use of classloaders to allow them to get warm, eviction of the cache to avoid memory leaks, proper `close`ing of autoclosable workers)

Different workers end up with different ad-hoc solutions, e.g. 

* `ZincWorkerImpl` with its `javaOnlyCompilersCache: mutable.Map[Seq[String], SoftReference[Compilers]]`, `classloaderCache: collection.mutable.LinkedHashMap[Long, SoftReference[ClassLoader]]`, `compilerCache: KeyedLockedCache[Compilers]`

* `ScalaJSWorkerImpl` with its `ScalaJSLinker.cache: mutable.Map[LinkerInput, SoftReference[(Linker, IRFileCache.Cache)]] `

* `ScalaNativeWorker` with its `scalaInstanceCache: Option[(Long, workerApi.ScalaNativeWorkerApi)]`

* `VisualizeModule.worker` returns a tuple of `(LinkedBlockingQueue[(Seq[NamedTask[Any]], Seq[NamedTask[Any]], os.Path)], LinkedBlockingQueue[Result[Seq[PathRef]]])` that it expects you to use to pass arguments to the worker and receive results in a thread-safe manner

These all solve the problems above to varying degrees: some disallow parallelism, some have concurrency bugs. Really none of the concurrency concerns should be handled by users, since they're the same for most workers, and Mill should provide the scaffolding to let users plug in their own logic without a care for concurrency

We should provide a more opinionated API to `Task.Worker` that allows users to delegate these "generic worker" problems to Mill: exactly what that API would look like is TBD, but perhaps something like

```scala
// Helper
def createMultiThreadSafeWorker[T](cacheKey: Any, factory: () => T)

// Usage
def myCompilerWorker = Task.Worker(
  createMultiThreadSafeWorker(
    cacheKey = scalaVersion()
[...]

PR #4001, opened 2024-11-21
Overhaul worker concurrency helpers with `mill.api.CacheFactory`

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/3730

The old `FixSizedCache`/`KeyedLockedCache`/`RandomBoundedCache` were all a bit of a mess, and none of them provided one crucial component to long-lived cache entries: the ability to tear them down after you are done with them. This is necessary for a lot of long-lived components you may want to cache: classloaders, subprocesses, open file handles, etc. all need to be torn down after use. Also, their implementation was weird, probably overly convoluted, and not used consistently throughout Mill

This PR introduces a new `CachedFactory` class meant to replace the old ones. I use it in four places (ZincWorker Scala and Java compilers, ScalaJsWorker, ScalanativeWorker), and added docs so users can discover and use it too. It provides a hook for `def teardown`, which should hopefully fix some of the classloader leaks I've seen in our prior implementations (these normally manifest as CODE CACHE FULL, DISABLING JIT COMPILER warnings when doing large builds). The last previously-problematic worker defsite `VisualizeModule.worker` cannot be changed without breaking bincompat so I left it in place for now

`CachedFactory` also differs from the original implementation is that the cache contains the logic for `def setup`, rather than relying on the callsite of `.withCachedValue` to provide the setup and key together. This should be easier to understand and less likely to be mis-used, as the mapping from `key => value` is kept in one place, whereas previously every `withCachedValue` could have a different mapping.

Hopefully this new `CachedFactory` helper class can form the basis of how people define concurrency-friendly stateful workers going forward

Covered by some unit tests and example tests, as well as usage in the main `ZincWorker`/`ScalaJSWorker`/`ScalaNativeWorker` code paths

### DIFF
--- a/example/fundamentals/tasks/6-workers/build.mill
+++ b/example/fundamentals/tasks/6-workers/build.mill
@@ -3,8 +3,14 @@
 // tasks in that they let you cache things, but the fact that they let you
 // cache the worker object in-memory allows for greater performance and
 // flexibility: you are no longer limited to caching only serializable data
-// and paying the cost of serializing it to disk every evaluation. This example
-// uses a Worker to provide simple in-memory caching for compressed files.
+// and paying the cost of serializing it to disk every evaluation.
+//
+// This example uses a Worker to provide simple two-level cache: in-memory caching
+// for compressed files, in addition to caching them on disk. This means that if the
+// Mill process is persistent (e.g. with `--watch`/`-w`) the cache lookups are instant,
+// but even if the Mill process is restarted it can load the cache values from disk
+// without having to recompute them:
+
 package build
 import mill._, scalalib._
 import java.util.Arrays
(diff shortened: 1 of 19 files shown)


lihaoyi:
CC @jodersky @alexarchambault if you guys want to take a look

example/fundamentals/tasks/6-workers/build.mill
```diff
@@ -30,20 +36,25 @@ class CompressWorker(dest: os.Path) {
   val cache = collection.mutable.Map.empty[Int, Array[Byte]]
   def compress(name: String, bytes: Array[Byte]): Array[Byte] = {
     val hash = Arrays.hashCode(bytes)
-    if (!cache.contains(hash)) {
+    if (!synchronized(cache.contains(hash))) {
```
roman-mibex-2, line 39: Naive 'user level' complaint:

I'm not a fan the 'synchronized' in an example. Its so easy to get wrong and I'm always having a hard time to convince myself that it is correct.

What I would love instead is having the concurrent map operations on it:
`cache.computeIfAbsent(key, (key)=>createTheExpensiveThing())`

Note again: This is from a doc/end user perspective and not the internal use.

roman-mibex-2: Oh...this is a negative example / problem statement =). So,yes, then it makes sense.

Issue #3927, opened 2024-11-09
Add first class Javascript/Typescript support to the Mill build tool (4500 Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 4500USD bounty on this issue, payable by bank transfer on merged PRs implementing this. [Standard bounty terms](https://github.com/orgs/com-lihaoyi/discussions/6) apply

---

Mill is a multi-language build tool that provides automatic caching and parallelism of build tasks related to local development (see [What Makes Mill Unique?](https://mill-build.org/mill/comparisons/unique.html)). Traditionally it targeted JVM languages, but we think that JS/TS projects could benefit from automatic caching and parallelism as well, especially the larger multi-language codebases with both JVM and JS/TS modules that would benefit from having a single build tool able to build code in both languages incrementally and in parallel.

The goal of this ticket is to generate a set of Typescript build examples that match the Java/Scala/Kotlin equivalents. We already have an [example TypescriptModule](https://mill-build.org/mill/extending/example-typescript-support.html) for demo purposes, but we would need to flesh it out using the equivalent Typescript tools and libraries.

For the purposes of this ticket, each examples should match as closely as possible the Java/Scala/Kotlin equivalents, to provide a useful minimal-but-still-educational code example, along with associated english documentation and explanations. You should read through the relevant sections of the [Building Java with Mill](https://mill-build.org/mill/javalib/intro.html) documentation before proceeding, even if you don't know Java, just to get a feel for what the documentation and examples for each section should cover. 

* [x] `example/jslib/basic/` (500USD)
    * [x] `1-simple/`: A minimal Typescript module demonstrating typechecking/running/testing/bundling of a Node CLI tool
    * [x] `2-react/`: A minimal Typescript module demonstrating typechecking/running/testing/bundling of a React-based web application, served using a static HTML page
    * [x] `3-custom-build-logic/`: A Typescript module with custom build logic

* [x] `example/jslib/basic/`: (500USD)
    * [x] `4-multi-modules/`: Multiple inter-related Typescript modules
    * [x] `5-client-server-hello/`: Demonstrate a client-server application with two modules, one a minimal JS server and one a minimal JS client, wired up and interacting
    * [x] `6-client-server-realistic/`: Demonstrate a client-server application with two modules, one an Express server and one an React.js client, wired up and interacting, implementing a TodoMVC application

* [x] `example/jslib/dependencies/` (500USD)
    * [x] `1-npm-deps/`
[...]

PR #4003, opened 2024-11-21

monyedavid:
### Related Issues

#3927

### Checklist

- [x] **example/jslib/basic**
  - [x] 1-simple/: A minimal Typescript module demonstrating typechecking/running/testing/bundling of a Node CLI tool
  - [x] 2-react/: A minimal Typescript module demonstrating typechecking/running/testing/bundling of a React-based web application, served using a static HTML page
  - [x] 3-custom-build-logic/: A Typescript module with custom build logic


example/javascriptlib/basic/1-simple/build.mill
```diff
@@ -1,18 +1,159 @@
 package build
-import mill._, javascriptlib._
+import mill._
+
+trait TypeScriptModule extends Module {
```
lihaoyi, line 4: You should use `TypeScriptModule` from `javascriptlib` instead of copying it all here

lihaoyi:
Each top-level bullet of the original bounty should be its own PR. For this PR, just focus on `example/jslib/basic/`, and we can review/merge/payout the first part of the bounty first. Extending it for further top-level bullets can come later

example/javascriptlib/basic/1-simple/build.mill
monyedavid: using `import mill._, javascriptlib._`

exits with this error:

```
[build.mill-58] [error] /.../build.mill:3:16: not found: value javascriptlib
[build.mill-58] [error] import mill._, javascriptlib._

```

monyedavid: note, import scalalib._ and others works fine

monyedavid: I think documentation https://mill-build.org/mill/index.html
should include dist.run usage as it was my primary source for information and I am sure it would be for others

example/javascriptlib/basic/1-simple/build.mill
```diff
@@ -1,25 +1,125 @@
...
+  }
 }
 
-// Documentation for mill.javascriptlib
+// Documentation for mill.example.javascriptlib.basic
 
 /** Usage
 
-> mill qux.run James Bond prof
+> mill dist.run example/javascriptlib/basic/1-simple qux.run James Bond prof
```
lihaoyi, line 118: @monyedavid in the example tests you can use `mill` without `dist.run`. Please look at the other example tests for reference, the readme.adoc for how to run these examples, and make sure they are green

monyedavid: got it

example/javascriptlib/basic/1-simple/build.mill
```diff
@@ -1,25 +1,110 @@
...
+  def test =
+    Task {
+      os.call(("echo", "No tests defined for module " + millSourcePath.last), stdout = os.Inherit)
+    }
+
+}
 
-object foo extends TypeScriptModule {
-  object bar extends TypeScriptModule {
+object foo extends SimpleMod {
```
lihaoyi, line 34: The `1-simple` example should only have a single module, not three

example/javascriptlib/basic/1-simple/build.mill
```diff
@@ -1,25 +1,110 @@
...
+      .map(PathRef(_))
+  }
+
+  override def run(args: mill.define.Args) = Task.Command {
+    val (_, env) = prepareRun()
+    val mainFile = compile()._1.path / millSourcePath.last / "src" / mainFileName()
+    os.call(("node", mainFile, args.value), stdout = os.Inherit, env = env)
+  }
+
+  def bundle = Task {
```
lihaoyi, line 58: Why are we duplicating logic that is already defined in `TypescriptModule`?

example/javascriptlib/basic/1-simple/build.mill
```diff
@@ -1,25 +1,110 @@
...
+      javascriptOut,
+      npmInstall().path,
+      npmInstall().path / "node_modules"
+    ).mkString(":"))
+
+    os.call(
+      ("node", npmInstall().path / "node_modules/jest/bin/jest.js", javascriptOut),
+      stdout = os.Inherit,
+      env = env
+    )
```
lihaoyi, line 83: Most of this logic should be pushed upstream, let's say into a new `JestModule` so that we keep the possibilities open to support multiple kinds of test suite

lihaoyi:
It's a start and good that tests pass, but the code in the various `build.mill` example projects is far too messy to serve their purpose as documentation examples. 

For reference, look at the equivalent `build.mill` files in `example/pythonlib/basic/` or `example/kotlinlib/basic/` for what I would expect the `build.mill` files to look like. Everything else should be pushed upstream into `javascriptlib/src/` library traits and helpers, and those will have to be organized appropriately (see `pythonlib/src/` as an example)

example/javascriptlib/basic/1-simple/build.mill
```diff
@@ -1,27 +1,58 @@
...
+      "@babel/core@^7.26.0",
+      "@babel/preset-env@^7.26.0",
+      "typescript@5.6.3",
+      "jest@^29.7.0",
+      "ts-jest@^29.2.5",
+      "ts-node@^10.9.2",
+      "babel-jest@^29.7.0",
+      "esbuild@0.24.0",
+      "immutable@4.3.7",
+      transitiveNpmDeps()
```
lihaoyi, line 22: This list is very long. Where does it come from? Are all of them needed? How can a developer know to come up with this list themselves?

example/javascriptlib/basic/3-custom-build-logic/build.mill
```diff
@@ -0,0 +1,64 @@
+package build
+
+import mill._, javascriptlib._
+
+trait CustomModule extends JestModule {
```
lihaoyi, line 5: We should not have a `trait` here and just do things directly in `object foo`, for consistency with the other examples

example/javascriptlib/basic/1-simple/build.mill
```diff
@@ -1,27 +1,58 @@
 package build
+
 import mill._, javascriptlib._
 
-object foo extends TypeScriptModule {
-  object bar extends TypeScriptModule {
-    def npmDeps = Seq("immutable@4.3.7")
+trait SimpleMod extends JestModule {
+  override def npmInstall = Task {
```
lihaoyi, line 6: Can we override `npmDeps` rather than overriding `npmInstall`?

javascriptlib/src/mill/javascriptlib/ReactScriptsModule.scala
```diff
@@ -0,0 +1,174 @@
...
+      "typescript@^4.9.5",
+      "web-vitals@2.1.4",
+      "@testing-library/jest-dom@^5.17.0",
+      "@testing-library/react@^13.4.0",
+      "@testing-library/user-event@^13.5.0",
+      "@types/jest@^27.5.2",
+      "@types/node@^16.18.119",
+      "@types/react@^18.3.12",
+      "@types/react-dom@^18.3.1",
+      "serve@12.0.1",
```
lihaoyi, line 27: Should these be in `npmDeps` or `npmDevDeps` instead?

javascriptlib/src/mill/javascriptlib/ReactScriptsModule.scala
```diff
@@ -0,0 +1,174 @@
...
+      ujson.Obj.from(dependenciesObj)
+    }
+
+    val allPaths = Seq(
+      "*" -> Task.dest / "node_modules",
+      "app/*" -> "app",
+      "typescript" -> Task.dest / "node_modules" / "typescript"
+    )
+
+    // mk tsconfig.json
```
lihaoyi, line 60: Is this format documented anywhere? If so we should link it in a comment

javascriptlib/src/mill/javascriptlib/ReactScriptsModule.scala
```diff
@@ -0,0 +1,174 @@
+package mill.javascriptlib
+
+import mill._
+import os.*
+
+// create-react-app scripts
+trait ReactScriptsModule extends TypeScriptModule {
```
lihaoyi, line 7: If this is specific to `create-react-app`, should it be `CreateReactAppModule`?

example/javascriptlib/basic/1-simple/build.mill
monyedavid: ok, will move out the common packages JestModule depends on and leave "immutable" in SimpleMod since its project specific.

javascriptlib/src/mill/javascriptlib/ReactScriptsModule.scala
monyedavid: yeah this is create-react-app pretty standard. I will add a link

javascriptlib/src/mill/javascriptlib/ReactScriptsModule.scala
monyedavid: Will update the name

lihaoyi:
Left more comments. We're making progress, still more to be done

We need to figure out a proper interface between the user code in `build.mill` and the library code in `javascriptlib/src/`. Putting everything in either place is wrong, which is what we have so far: `1-simple` just puts everything in `build.mill`, `2-react` just puts everything in `javascriptlib/src/`. This is a tricky balance to strike, but it is the key to making `javascriptlib` useful at all, so we will need to continue discussing it until we find a decent place to draw the line

One thing you should do is start the english explanations of each examples, similar to the ones we see in `example/scalalib/basic` or `example/pythonlib/basic`. If you cannot come up with a concise understandable explanation of what the code is doing and how it can be extended, then the code likely needs to be reworked

javascriptlib/src/mill/javascriptlib/ReactScriptsModule.scala
monyedavid: hmm both. I have updated it to use npmDeps and npmDevDeps.

example/javascriptlib/basic/1-simple/build.mill
```diff
@@ -1,27 +1,36 @@
...
-    def npmDeps = Seq("immutable@4.3.7")
-  }
-}
+object foo extends JestModule {
+  def npmDeps = Seq("immutable@4.3.7")
 
-object qux extends TypeScriptModule {
-  def moduleDeps = Seq(foo, foo.bar)
+  override def mainFilePath =
+    Task { compile()._1.path / millSourcePath.last / "src" / mainFileName() }
```
lihaoyi, line 9: Why do we need to override this?

javascriptlib/src/mill/javascriptlib/TypeScriptModule.scala
```diff
@@ -56,28 +68,40 @@ trait TypeScriptModule extends Module {
...
       os.copy(jsDir.path, Task.dest / mod.millSourcePath.subRelativeTo(Task.workspace))
     }
-    val mainFile = compile()._1.path / mainFileName()
-    val env = Map("NODE_PATH" -> Seq(".", compile()._1.path, npmInstall().path).mkString(":"))
-    (mainFile, env)
+    (mainFilePath(), mkENV())
   }
 
-  def run(args: mill.define.Args) = Task.Command {
+  def argsOrder: Task[(Boolean, Seq[String])] = Task { (false, Seq("")) }
```
lihaoyi, line 86: What's this task about?

javascriptlib/src/mill/javascriptlib/CreateReactAppModule.scala
```diff
@@ -0,0 +1,171 @@
...
+    val allPaths = Seq(
+      "*" -> Task.dest / "node_modules",
+      "app/*" -> "app",
+      "typescript" -> Task.dest / "node_modules" / "typescript"
+    )
+
+    // mk tsconfig.json
+    os.write(
+      Task.dest / "tsconfig.json",
+      ujson.Obj(
```
lihaoyi, line 60: These JSON blobs should probably be exposed as top level `Task[ujson.Value]` for users to configure directly

example/javascriptlib/basic/1-simple/build.mill
monyedavid: for projects with nested directories, tsc will recreate the directories and compile attached js output within.
so a project with src/, test/ will have src/, test/ in out/**/compile.dest/

the original setup does not account for this.

javascriptlib/src/mill/javascriptlib/TypeScriptModule.scala
monyedavid: to include computed arguments during run, the boolean is to specify if/not to include before/after user defined arguments

lihaoyi:
@jodersky could you give this a review when you have some time? Since your work on the Pythoncside is very similar you may have ideas that are transferable to JS/TS

javascriptlib/src/mill/javascriptlib/TypeScriptModule.scala
lihaoyi: Why does whether we include them before or after matter?

example/javascriptlib/basic/1-simple/build.mill
lihaoyi: millSourcePath.last looks like the wrong eay to implement what you described, as it only returns the last segment. What if it is nested two levels deep?

Should we push the change upstream into TypescriptModule instead?

javascriptlib/src/mill/javascriptlib/TypeScriptModule.scala
lihaoyi: The task type signature looks bery weird. If we really need to do this, can we break it into teo separate tasks each one a `Seq[String]`, one that gets inserted before user args and one after?

javascriptlib/src/mill/javascriptlib/ReactScriptsModule.scala
monyedavid: I changed it back to `ReactScriptsModule` cra runs with react-scripts which is a package different from React

javascriptlib/src/mill/javascriptlib/TypeScriptModule.scala
monyedavid: I did instead: ` Task[(Seq[String], Seq[String])]`

example/javascriptlib/basic/1-simple/build.mill
monyedavid: my mistake tsconfig will dump code into: /out/foo/compile.dest/javascript/foo/...
hence the `millSourcePath.last `

Also since the tsconfig compiler options is fully customizable, ts configurations will vary and compile output dir will also vary.

I can update the tsconfig from foo, so the main file is at a different directory (/compile.dest/javascript/....) (without the foo after javascript). But I don't see the need, projects that use will most likely use varying tsconfigs and will have varying mainFilePaths

jodersky:
Will do!

example/javascriptlib/basic/1-simple/build.mill
monyedavid: no longer depends on millSourcePath.last

jodersky (commented):
I think this would be easier to review and improve if each example were its own pull request, starting with only the basic example.

A couple of things stand out to me:

- I think that testing can be factored out into its own module, without having to subclass modules and have custom tasks. Eg. instead of defining the "main" module to be a `JestModule` with `testSources` etc, the test sources could become the normal sources of a nested `JestModule` which would extend something like `TestModule. This is how Mill's Java and Scala support does it, and is what I used for the Python support as well. I'd recommend a look at how the [pythonlib PythonModule and TestModue testing modules](https://github.com/com-lihaoyi/mill/blob/029317d3884d1519703992ab3e3eebe6ce836e62/pythonlib/src/mill/pythonlib/TestModule.scala) are structured, since they're still quite minimal, and mostly follow what javalib/scalalib does.
 
- It will also be easier to approach this by documentation to examples first, and also commenting what every task is supposed to do.

monyedavid:
> I think this would be easier to review and improve if each example were its own pull request, starting with only the basic example.
> 
> A couple of things stand out to me:
> 
> * I think that testing can be factored out into its own module, without having to subclass modules and have custom tasks. Eg. instead of defining the "main" module to be a `JestModule` with `testSources` etc, the test sources could become the normal sources of a nested `JestModule` which would extend something like `TestModule. This is how Mill's Java and Scala support does it, and is what I used for the Python support as well.
> * It will also be easier to approach this by documentation to examples first, and also commenting what every task is supposed to do.
> 
> In general, I'd suggest to take a look at what we're doing with [`pythonlib`](https://github.com/com-lihaoyi/mill/blob/029317d3884d1519703992ab3e3eebe6ce836e62/pythonlib/src/mill/pythonlib/TestModule.scala) and use it as inspiration on how to approach this. The reason is that it's still fairly minimal and to-the-point, but it does strive to follow general tendencies and structure of the established `javalib` and `scalalib`.

the abstraction (TestModule) is done in another pr, as that extends beyond the scope of this task set (and many other things here really)

I can open that next after this is done / move them here whichever you prefer @jodersky 

Separating into smaller pr's would be somewhat difficult, its length is due to the modifications made to TypescriptModule and creating the other two modules. I promise the other pr's would be much smaller.

lihaoyi:
I think this looks in decent enough state to merge. It's not perfect, but it would be useful to get some of the follow on use cases in place first before polishing further.

@monyedavid there's a test failure that needs to be fixed, then I'll merge this and close out the bounty

lihaoyi:
Merged this, thanks @monyedavid for bearing with all our review and feedback! I think the PR is in a great state now to build upon for later, and I expect follow-up PRs to be easier since you are now more familiar with the Mill tool, codebase, and coding conventions.

Please send me your international bank transfer details over email and I will wire over the bounty

PR #4005, opened 2024-11-21
Add `isFullLinkJS` parameter to `scalaJSMinify`

arturaz:
Do not minify fast link JS by default, because that makes it harder to debug the application code in development.

I did not figure out a way to make this backwards compatible, so that existing builds that `override def scalaJSMinify` wouldn't have to update the method signature.

[Discord discussion](https://discordapp.com/channels/632150470000902164/1295786923410984981/1309286358224998401).

### DIFF
--- a/scalajslib/src/mill/scalajslib/ScalaJSModule.scala
+++ b/scalajslib/src/mill/scalajslib/ScalaJSModule.scala
@@ -139,7 +139,7 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
       esFeatures = esFeatures(),
       moduleSplitStyle = moduleSplitStyle(),
       outputPatterns = scalaJSOutputPatterns(),
-      minify = scalaJSMinify(),
+      minify = scalaJSMinify(isFullLinkJS)(),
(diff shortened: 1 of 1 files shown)


scalajslib/src/mill/scalajslib/ScalaJSModule.scala
```diff
@@ -326,8 +326,15 @@ trait ScalaJSModule extends scalalib.ScalaModule { outer =>
...
+   * 
+   * @param isFullLinkJS
+   *   whether we are in full-link mode
    */
-  def scalaJSMinify: T[Boolean] = Task { true }
+  def scalaJSMinify(isFullLinkJS: Boolean): T[Boolean] = Task { 
+    // Do not minify fast link JS by default, because that makes it harder to debug the application
+    // code in development.
+    isFullLinkJS
+  }
```
lefou, line 337: This can't work. Cached tasks can't accept parameters.

lolgab:
Since we are not in the breaking window yet, I proposed a binary compatible change in #4051

lihaoyi:
closing this for now since the other PR has landed

PR #4009, opened 2024-11-22
Optimize startup time using TCP sockets instead of junixsocket and `tput` instead of jline

lihaoyi:
The `AFUNIXServerSocket` library we are using seems to cause an extra ~500ms latency launching the Mill client, and also causes the Graal native-image to crash during generation. 

* Using TCP sockets seems to cut the launch overhead down from ~1000ms to ~500ms, and opens up the possibility of using native-image to cut it further. 

* Using localhost TCP sockets seems secure as far as I can tell  https://security.stackexchange.com/questions/108544/once-established-are-sockets-on-localhost-secure

* The JVM also ships with `UnixDomainSocketAddress` starting from JDK 17 that we can consider using, either by requiring JDK17 or by conditionally using it based on JDK version. But for now I want to keep supporting JDK11 and don't want to split the code paths, so we'll hold off on this for now

`jline.terminal.Terminal.getSize` also adds a few hundred milliseconds, so we instead use the same `tput` command that Ammonite uses which takes <10ms. 

* On windows where `tput` isn't support, it falls back to 100 cols, which I reduced it down from 120 to hopefully avoid line wrapping in most terminals while still providing a decent experience

Together these two changes cuts down the time taken for a hot `time ./mill version` on the Mill repo from ~1.05s to ~0.24s


main/server/src/mill/main/server/Server.scala
```diff
@@ -48,7 +46,8 @@ abstract class Server[T](
 
         while (
           running && {
-            val serverSocket = bindSocket()
+            val serverSocket = new java.net.ServerSocket(0)
```
roman-mibex-2, line 49: This binds to 0.0.0.0 for me, so it is listening to the world, including external network. At least on my machine.

So, better IMO:
```
        var serverSocket = new java.net.ServerSocket();
        serverSocket.bind(new java.net.InetSocketAddress("127.0.0.1",0));
```

I suggesting  the IP 127.0.0.1 over 'localhost', because localhost sometimes resolves to IP6 vs IP4, creating odd miss-matches.

main/api/src/mill/api/ClassLoader.scala
```diff
@@ -90,7 +90,9 @@ object ClassLoader {
         )
         retry {
           try os.copy(os.Path(Export.rt()), java90rtJar, createFolders = true)
-          catch { case e: FileAlreadyExistsException => /* someone else already did this */}
+          catch {
+            case e: FileAlreadyExistsException => /* someone else already did this */
+          }
         }
```
himanshumahajan138, line 96: @lihaoyi this file also got same changes in my PR while fixing linting error should i remove the changes from there or rebasing and merge will automatically fix this

PR #4019, opened 2024-11-24
Fix Detekt tests

lihaoyi:
Somehow the `PathRef(T.workspace).path` is causing issues, but it's totally redundant since we can just use `T.workspace`. Walking the `out/` folder filesystem and hashing stuff seems to be causing problems with the file locks we use to ensure mutual exclusion

Seems to make `example.kotlinlib.linting[1-detekt].local.test` pass reliably where previously it was very flaky

### DIFF
--- a/kotlinlib/src/mill/kotlinlib/detekt/DetektModule.scala
+++ b/kotlinlib/src/mill/kotlinlib/detekt/DetektModule.scala
@@ -21,7 +21,7 @@ trait DetektModule extends KotlinModule {
 
   private def detekt0() = T.task {
 
-    val args = detektOptions() ++ Seq("-i", PathRef(T.workspace).path.toString()) ++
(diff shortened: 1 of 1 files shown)


kotlinlib/src/mill/kotlinlib/detekt/DetektModule.scala
```diff
@@ -21,7 +21,7 @@ trait DetektModule extends KotlinModule {
 
   private def detekt0() = T.task {
 
-    val args = detektOptions() ++ Seq("-i", PathRef(T.workspace).path.toString()) ++
```
lefou, line 24: `PathRef(T.workspace)` is calculating an expensive (but unused) hash doomed to change as soon as we finish the task and write the cache to disc (into `T.workspace/"out"`).

PR #4027, opened 2024-11-25
Prepare JS Modules to be more extendable

myyk:
This is a very opinionated change which is inspired by my experiment with running NextJS on Mill in https://github.com/myyk/mill/tree/example-nextjs. It would be a smaller diff to merge that build off of this.

This change adds these features:
* rename `TypeScriptModule` to `TscModule` since there will be many types of TypeScript projects that wouldn't even want to extend it.
* Split out the node package managing stuff into `NodeModule`. In the future, this might even want to have flavors of `NpmModule`/`YarnModule`/etc but I don't want to over engineer this.
* `install` is a new command which can be more agnostic to `npm` or `yarn` or whatever and will install deps and copy 
`node_modules/` to the user's project outside of the cache to help the IDE. This will make auto-complete work a lot of the time.
* Splits out dev and normal dependencies like would be done in a node project.
* Ignores `node_modules/` in mill repo
* Adds ignoring `node_modules/` in examples, assuming user can copy example into their repo as a template.

Future work after this:
* Add a `NextJsModule` which extends `NodeModule` and looks a lot like this: https://github.com/myyk/mill/blob/example-nextjs/example/javascriptlib/basic/2-nextjs/build.mill

### DIFF
--- a/example/javascriptlib/basic/1-simple/build.mill
+++ b/example/javascriptlib/basic/1-simple/build.mill
@@ -1,13 +1,13 @@
 package build
 import mill._, javascriptlib._
 
-object foo extends TypeScriptModule {
-  object bar extends TypeScriptModule {
+object foo extends TscModule {
+  object bar extends TscModule {
     def npmDeps = Seq("immutable@4.3.7")
   }
 }
 
-object qux extends TypeScriptModule {
+object qux extends TscModule {
   def moduleDeps = Seq(foo, foo.bar)
 }
 


lihaoyi:
Let's merge https://github.com/com-lihaoyi/mill/pull/4003 before we proceed with this one, since that one has been WIP for a while and seems almost done

lihaoyi:
Closing this out of inactivity

Issue #3570, opened 2024-09-18
Ensure Mill names are only exported from one top-level package


lihaoyi:
Right now, importing both `javalib._` and `scalalib._` results in an error due to ambiguity when the same classes are exported from both:

```scala
package build
import mill._, javalib._, scalalib._

val x = Assembly
```
```
#06 [info] compiling 1 Scala source to /Users/lihaoyi/Github/mill/example/scalalib/basic/1-simple/out/mill-build/compile.dest/classes ...
#06 [error] /Users/lihaoyi/Github/mill/example/scalalib/basic/1-simple/build.mill:4:9: reference to Assembly is ambiguous;
#06 [error] it is imported twice in the same scope by
#06 [error] import scalalib._
#06 [error] and import javalib._
#06 [error] val x = Assembly
#06 [error]         ^
#06 [error] one error found
```

This isn't much of a problem now, as `scalalib` is a strict superset of `javalib`, so you can just use that all the time. But it may become a problem when https://github.com/com-lihaoyi/mill/issues/3567 breaks out `javalib` from `scalalib` so that they may each have their own unique features, and https://github.com/com-lihaoyi/mill/issues/3451 introduces `kotlinlib._`. When a user tries to have a multi-language codebase like `import javalib._, scalalib._, kotlinlib._`, they will inevitably hit the errors above

A solution to this would be to ensure that the different `*lib` packages do not export things that their upstream dependencies already export. So e.g. 

* If `scalalib` depends on `javalib`, you would need `import javalib._, scalalib._` to work. 
* And similarly if `kotlinlib` depends on `javalib`, `import javalib._, kotlinlib._`. 
* And for a project with both java/scala/kotlin, `import javalib._, kotlinlib._, scalalib._` would work without ambiguity

This will involve moving/removing a bunch of forwarders, and could happen together with https://github.com/com-lihaoyi/mill/issues/3567 in 0.13.0

jodersky:
Regarding

> When a user tries to have a multi-language codebase like import javalib._, scalalib._, kotlinlib._, they will inevitably hit the errors above

> A solution to this would be to ensure that the different *lib packages do not export things that their upstream dependencies already export. So e.g.

What should we do with types that describe generic functionality and that are reimplemented in various packages? E.g. there could be a `PublishModule` shared across `javalib`, `scalalib` and `kotlinlib`, but implemented as a separate trait with the same name in `pythonlib`.

[...]

PR #4032, opened 2024-11-26
pythonlib: publish modules

jodersky:
Adds the capability to build wheels and sdists, and publish them.

The idea is to roughly follow [Pant's approach to publishing python packages](https://www.pantsbuild.org/stable/docs/python/overview/building-distributions): act a build frontend which calls out to various specialized packaging tools.

Part of #3928

### DIFF
--- a/pythonlib/src/mill/pythonlib/PublishModule.scala
+++ b/pythonlib/src/mill/pythonlib/PublishModule.scala
@@ -1,0 +1,268 @@
+package mill.pythonlib
+
+import mill.api.Result
+import mill.{PathRef, Task, T, Command}
+
+/**
+ * A python module which also defines how to build and publish source distributions and wheels.
+ */
+trait PublishModule extends PythonModule {
+
+  override def moduleDeps: Seq[PublishModule] = super.moduleDeps.map {
+    case m: PublishModule => m
+    case other =>
+      throw new Exception(
+        s"PublishModule moduleDeps need to be also PublishModules. $other is not a PublishModule"
+      )
+  }
+
+  override def pythonToolDeps = Task {
+    super.pythonToolDeps() ++ Seq("setuptools>=75.6.0", "build>=1.2.2", "twine>=5.1.1")
+  }
+
+  /**
+   * Metadata about your project, required to build and publish.
+   *
+   * This is roughly equivalent to what you'd find in the general section of a `pyproject.toml` file
+   * https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#about-your-project.
+   */
+  def publishMeta: T[PublishModule.PublishMeta]
+
+  /**
+   * The artifact version that this module would be published as.
+   */
+  def publishVersion: T[String]
+
+  /**
+   * The content of the PEP-518-compliant `pyproject.toml` file, which describes how to package this
+   * module into a distribution (sdist and wheel).
+   *
+   * By default, Mill will generate this file for you from the information it knows (e.g.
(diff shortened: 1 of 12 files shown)


himanshumahajan138:
@jodersky Sir I am targeting web examples, is there any conflict???

himanshumahajan138:
> You're good to go on that. Note however that we still have some fundamentals missing, and compatibility may be broken while we work on the other missing parts

actually most of the code will rely on basic python and i think if there is some need of some function or logic i will add it on the go 

but if you want me to solve a particular bullet first then please once suggest me (don't think like you are giving me duty or work, i take mill as my responsibility and always ready to contribute so please once suggest me if you need particular bullet to be done first)

pythonlib/src/mill/pythonlib/SetupTools.scala
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Essentially, every `PythonModule` should have a way to build wheels and
+ * sdists, however how that is done can vary. As of now, we only support
+ * building artifacts via setuptools (generating and invoking a `setup.py`), but
+ * in the future that may change. For example, we may support building wheels by
+ * reading PEP-517 compliant `pyproject.toml` files, or even use "pure" Mill to
+ * build wheels and sdists. Hence, to separate concerns, we define some tasks in
+ * this mixin, and extend it in PythonModule.
+ */
+trait SetupTools extends Module { self: PythonModule =>
```
jodersky, line 18: not sure about this design

jodersky: Or, we KISS it, and flatten everything into one trait. Extracting the tasks will then break binary compatibility, but I'm sure that will happen at least once anyway, given that `pythonlib` is brand new anyway,

lihaoyi: I'd say dont worry about bincompat or stability at all for now; its still to early in the lifecycle of this part of the project. I'd just flatten it for now until we have a concrete reason to abstract it and then make the change

himanshumahajan138:
Sir @jodersky Can you please add background run to this please...

pythonlib/src/mill/pythonlib/PublishModule.scala
```diff
@@ -0,0 +1,276 @@
...
+ */
+trait PublishModule extends PythonModule {
+
+  // export these to make using them in subclasses possible without an import
+  type License = scalalib.publish.License
+  val License = scalalib.publish.License
+  type PublishMeta = PublishModule.PublishMeta
+  val PublishMeta = PublishModule.PublishMeta
+  type Developer = PublishModule.Developer
+  val Developer = PublishModule.Developer
```
lihaoyi, line 17: Should we export these from `mill.pythonlib`'s `package` object instead? We don't really do this kind of trait re-exporting elsewhere in Mill, so having people import from `mill.pythonlib`. https://github.com/com-lihaoyi/mill/issues/3570 is a problem, but I think better to just suffer that while we figure out a wholistic solution rather than doing ad-hoc mitigations for now

lihaoyi:
As a first pass this looks great, not much comments. `twine` seems to do most of the heavy lifting here and it's nice to be able to outsource it

pythonlib/src/mill/pythonlib/PublishModule.scala
jodersky: sounds good, I'll move it to a package oject

Issue #3638, opened 2024-10-01
Add javalib examples on using `jlink`, `jpackage`, and producing JPMS modules (500USD Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 500USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

These are newer Java tools that we should have documented/tested golden paths to using. The examples should go in `example/javalib/module` with `/** Usage */` tests and textual explanations of what the examples are doing and how they relate to the underlying tools and concepts (with links to the relevant upstream documentation pages)

lefou:
* https://github.com/com-lihaoyi/mill/pull/3201

ayewo:
Hi @lihaoyi 
I've just opened a PR.

When you have a moment, would appreciate it if you could take a look.

PR #4038, opened 2024-11-27
feat: add jlink & jpackage Java examples

ayewo:
This intended to close https://github.com/com-lihaoyi/mill/issues/3638.

The `JlinkModule` creates a runtime image in two steps:
1. it creates a `jlink.jmod` file for the module containing the `mainClass` using the [`jmod`](https://docs.oracle.com/en/java/javase/23/docs/specs/man/jmod.html) tool;
2. it then uses the [jlink](https://docs.oracle.com/en/java/javase/23/docs/specs/man/jlink.html) tool, to link the previously created `jlink.jmod` with a runtime image.

The generated runtime image can be executed with:
```sh
./out/foo/jlinkAppImage.dest/jlink-runtime/bin/jlink
```

As for the `JpackageModule` implementation, I decided to reuse most of the `JpackageModule` code @lefou  [linked](https://github.com/com-lihaoyi/mill/pull/3201) to the original issue. Of course his code completely avoids the `jmod` dance by using [`jpackage`](https://docs.oracle.com/en/java/javase/23/docs/specs/man/jpackage.html) to create a native package/installer directly.

`jpackage` supports up to 3 different package/installer outputs on macOS:
```sh
jpackageType = "dmg"
./out/foo/jpackageAppImage.dest/image/foo-1.0.dmg

jpackageType = "pkg"
./out/foo/jpackageAppImage.dest/image/foo-1.0.pkg

jpackageType = "app-image"
./out/foo/jpackageAppImage.dest/image/foo.app
```

Would be interested to hear your thoughts on the slight difference between how the 2 traits are implemented.


lihaoyi:
What happens after I run the `foo-1.0.pkg` installer? I assume it puts the file on disk somewhere, but I couldn't find it

scalalib/src/mill/scalalib/JlinkModule.scala
```diff
@@ -0,0 +1,90 @@
...
+    val jars = cp.filter(os.exists).zipWithIndex.map { case (p, idx) =>
+      val dest = libs / s"${p.last}"
+      os.copy(p, dest, createFolders = true)
+      dest
+    }
+    
+    val classPath = jars.map(_.toString).mkString(sys.props("path.separator"))
+    val args = {
+      val baseArgs = Seq(
+        "jmod",
```
lihaoyi, line 51: Is it possible to use `jlink` without first using `jmod`? For some reason the Java install I have on my laptop does not have `jmod` included (?)

```
lihaoyi mill$ jmod
zsh: command not found: jmod
lihaoyi mill$ java -version
openjdk version "17.0.6" 2023-01-17 LTS
OpenJDK Runtime Environment Corretto-17.0.6.10.1 (build 17.0.6+10-LTS)
OpenJDK 64-Bit Server VM Corretto-17.0.6.10.1 (build 17.0.6+10-LTS, mixed mode, sharing)
```

lihaoyi, line 51: Is it possible to use `jlink` without first using `jmod`? For some reason the Java install I have on my laptop does not have `jmod` included (?)

```
lihaoyi mill$ jmod
zsh: command not found: jmod
lihaoyi mill$ java -version
openjdk version "17.0.6" 2023-01-17 LTS
OpenJDK Runtime Environment Corretto-17.0.6.10.1 (build 17.0.6+10-LTS)
OpenJDK 64-Bit Server VM Corretto-17.0.6.10.1 (build 17.0.6+10-LTS, mixed mode, sharing)
```

scalalib/src/mill/scalalib/JlinkModule.scala
```diff
@@ -0,0 +1,90 @@
...
+  def jlinkModuleVersion: T[Option[String]] = T { None }
+
+  /** The main class to use as the runtime entry point. */
+  def jlinkMainClass: T[String] = T { finalMainClass() }
+
+  /** Compress level for the runtime image. 
+   * Valid values range between: 
+   *  "zip-0" (no compression) and "zip-9" (best compression)
+   * Defaults to "zip-6" */
+  def jlinkCompressLevel: T[String] = T { "zip-6" }
```
lihaoyi, line 29: This seems to be having problems in CI; are there any other values we could use?

scalalib/src/mill/scalalib/JpackageModule.scala
```diff
@@ -0,0 +1,83 @@
...
+      dest
+    }
+
+    val appName = jpackageName()
+    val appType = jpackageType()
+    val mainClass = jpackageMainClass()
+    val mainJarName = jars.head.last
+
+    val args: Seq[String] = Seq(
+      "jpackage",
```
lihaoyi, line 64: We should wire up all these CLI tools to the custom JVM versions that users can define https://mill-build.org/mill/fundamentals/configuring-jvm-versions.html. This probably just means using the executable from `this.zincWorker().javaHome()` rather than the one on `PATH`

scalalib/src/mill/scalalib/JlinkModule.scala
lihaoyi: Strangely enough, `jmod` does exist at `/Library/Java/JavaVirtualMachines/amazon-corretto-17.jdk/Contents/Home/bin/jmod`. So using the `javaHome()` version as mentioned below should solve this

example/javalib/module/16-jlink/build.mill
```diff
@@ -0,0 +1,23 @@
+//// SNIPPET:BUILD
+package build
+import mill._, javalib._
+import mill.javalib.Assembly._
+import mill.scalalib.JlinkModule
+
+object foo extends JavaModule with JlinkModule {
+  override def jlinkModuleName: T[String] = T { "foo" }
+  override def jlinkModuleVersion: T[Option[String]] = T { Option("1.0.0") }
```
lihaoyi, line 9: No need `override` keyword in `.mill` files

lihaoyi:
Overall the code looks reasonable I think. Some nits and some tweaks likely necessary to get it working reliably in CI and on my laptop.

One area that needs to be fleshed out is the english docs surrounding the examples. We should have a short explanation of what is going on, how these traits and the output files can be used from a user perspective, and with links to upstream documentation for more details

CC @lefou @alexarchambault

lihaoyi:
Managed to find the `foo` program in `Applications` after installing the `.pkg`, but running it seems to exit immediately before doing anything. Maybe we can make it pop up a Swing window or something just so it's visible to the user the program works?

lefou:
> As for the `JpackageModule` implementation, I decided to reuse most of the `JpackageModule` code @lefou [linked](https://github.com/com-lihaoyi/mill/pull/3201) to the original issue.

Please add me as [co-author](https://github.com/com-lihaoyi/mill?tab=readme-ov-file#commit-messages) to the relevant git commit. 

```
```

ayewo:
> Managed to find the `foo` program in `Applications` after installing the `.pkg`, but running it seems to exit immediately before doing anything. Maybe we can make it pop up a Swing window or something just so it's visible to the user the program works?

OK, I'll update the example code to be a basic desktop app in Swing.

> One area that needs to be fleshed out is the english docs surrounding the examples. We should have a short explanation of what is going on, how these traits and the output files can be used from a user perspective, and with links to upstream documentation for more details

In the original issue description, I interpreted your mention of "`/** Usage */` tests and textual explanations" as referring to adding illustrative examples inside `build.mill`, which I have done. Sounds like this is referring to updating the Mill documentation. Can you please point me to an example of what you have in mind?

ayewo:
@lihaoyi 
> What happens after I run the foo-1.0.pkg installer? I assume it puts the file on disk somewhere, but I couldn't find it

> Managed to find the foo program in Applications after installing the .pkg, but running it seems to exit immediately before doing anything. Maybe we can make it pop up a Swing window or something just so it's visible to the user the program works?

Double-clicking or doing an `open foo-1.0.pkg` in Terminal should show the installer.

I've updated the `jpackage` example to show a Swing GUI when the app is launched either from a `.dmg` or `.pkg` (after installing it to `/Applications`). 

For the `app-image` package, launching from the CLI (via `./out/foo/jpackageAppImage.dest/image/foo.app/Contents/MacOS/foo`) should also show the Swing GUI.

scalalib/src/mill/scalalib/JlinkModule.scala
ayewo: On my machine, I have OpenJDK installed:
```sh
java --version
openjdk 21.0.1 2023-10-17
OpenJDK Runtime Environment (build 21.0.1+12-29)
OpenJDK 64-Bit Server VM (build 21.0.1+12-29, mixed mode, sharing)
```
But a momentary switch to Oracle's JDK shows that `jlink` will only accept either of [`0`, `2`] as valid values for compression.

To summarize:
* On Oracle's build of the JDK, valid values are: `0` (no compression), `1` and `2` (zip compression).
* On recent builds of OpenJDK, valid values range between: [`"zip-0"` - `"zip-9"`].

I've updated the Usage examples so this subtle difference between JDK distributions is clearer.

scalalib/src/mill/scalalib/JlinkModule.scala
ayewo: Updated to use `javaHome()`.

scalalib/src/mill/scalalib/JpackageModule.scala
ayewo: Thanks for the pointer.

I've updated to use `javaHome()`.

ayewo:
@lihaoyi I’ve addressed all of your feedback.

lihaoyi:
Thanks, will take a look

lihaoyi:
Just looked through this, I think it looks pretty good.

Some last requests:

- Can you include your new examples into the documentation in https://github.com/com-lihaoyi/mill/blob/main/docs/modules/ROOT/pages/javalib/module-config.adoc?plain=1#L56-L58, and use `./mill docs.fastPages` to check that it looks good to you

- I imagine a common use case would be to generate `jpackage` outputs for all platforms, which the current `trait JpackageModule` makes difficult. Let's make a `def jpackageAppImage0` anonymous task that takes the `jpackageType: String` as a parameter, and then have `jpackageAppImage`/`jpackageDmg`/`jpackagePkg`/`jpackageExe`/`jpackageMsi`/`jpackageRpm`/`jpackageDeb` tasks that use `jpackageAppImage0`, and maybe a `jpackageAll` task that aggregates all the results. This should make it easier for the common use case where someone wants to generate all possible executable packages for the same Java module

lihaoyi:
We can leave it as is

ayewo:
Quick question regarding some test failures in [CI](https://github.com/com-lihaoyi/mill/actions/runs/12084145033/job/33698863549?pr=4038):

One of the usage example tests for `jpackage` looks like this:
```sh
java -jar ./out/foo/assembly.dest/out.jar
Nov 28, 2024 12:45:43 PM foo.Foo readConf
INFO: Loaded application.conf from resources: Foo Application Conf

Nov 28, 2024 12:45:44 PM foo.Bar lambda$main$0
INFO: Hello World application started successfully
``` 
In other words, it uses a logger that prints the current timestamp. I'm wondering if `UtestExampleTestSuite` is doing a string match against this for its expected output which might be why I'm seeing failures?

I'm guessing I'll need to remove the timestamp, right?

example/javalib/module/17-jpackage/build.mill
```diff
@@ -0,0 +1,107 @@
...
+
+/** Usage
+
+> mill foo.assembly
+
+> unzip -p ./out/foo/assembly.dest/out.jar application.conf
+Foo Application Conf
+
+> java -jar ./out/foo/assembly.dest/out.jar
+Nov 28, 2024 12:45:43 PM foo.Foo readConf
```
lihaoyi, line 75: @ayewo you can use `...` as wildcards when matching strings, e.g.

```
... foo.Foo readConf
```

ayewo:
@lihaoyi changes are ready to be merged.

lihaoyi:
@ayewo looks like some of the test failures are legit

ayewo:
@lihaoyi I believe I've fixed all the legit failures related to my changes.

lihaoyi:
Looks great, thanks @ayewo ! Please email me your international bank transfer details and I can close out the bounty

ayewo:
> Looks great, thanks @ayewo ! Please email me your international bank transfer details and I can close out the bounty

Thanks @lihaoyi! I've just emailed you using the email on your GH profile.

lefou:
While merging, you forgot to keep me in the co-authors. 😟

Issue #4036, opened 2024-11-27
Add a docsite page on manual migration to Mill


lihaoyi:
It's a bit tedious, but it isn't rocket science, if you kbow how to approach it. After migrating a few builds ourselves (netty/mockito/gatling/arrowkt), we should write up how they were done so there is a playbook for others to follow

PR #4040, opened 2024-11-28
Add `Migrating to Mill` doc page

lihaoyi:
This is a manual playbook for running such a migration, that will likely be useful even in the presence of semi-automated scaffolding generators

Resolves https://github.com/com-lihaoyi/mill/issues/4036


docs/modules/ROOT/pages/migrating/migrating.adoc
```diff
@@ -0,0 +1,283 @@
...
+tooling (e.g. see xref:migrating/maven.adoc[]), other automation is work-in-process
+(e.g. migrating from https://github.com/com-lihaoyi/mill/issues/3962[Gradle] or
+https://github.com/com-lihaoyi/mill/issues/3450[SBT]) but while automation helps with
+some of the scaffolding the general principles laid out on this page still apply.
+
+== How Long Does Migration Take?
+
+Migrating an existing project to Mill may take anywhere from an hour for small projects
+to several days or more for larger projects. These steps come experience doing proof-of-concept
+migrations a range of real-world projects to Mill, from simple single-module codebases to
```
lefou, line 16:
```suggestion
to several days or more for larger projects. These steps come from experience doing proof-of-concept
migrations of a range of real-world projects to Mill, from simple single-module codebases to
```

docs/modules/ROOT/pages/migrating/migrating.adoc
```diff
@@ -0,0 +1,283 @@
...
+
+** Sub-project `pom.xml` and `build.gradle` files need to be translated into Mill ``Module``s
+** Third-party dependencies need to be translated into Mill's def `ivyDeps`
+** Third-party Plugins need to be replaced by their Mill equivalent, or re-implemented
+** Custom build logic may need to be re-implemented in Mill
+
+3. *Once completed, the Mill build can be used as the default for some period of time*
+
+** This period gives to time to be confident in the robustness of the new Mill build system,
+   during which both old and new build systems should be maintained and kept up top date.
```
lefou, line 64:
```suggestion
** This period gives the time to be confident in the robustness of the new Mill build system,
   during which both old and new build systems should be maintained and kept up top date.
```

docs/modules/ROOT/pages/migrating/migrating.adoc
```diff
@@ -0,0 +1,283 @@
...
+whether bundled in Mill or your own DIY scripts, but there will always be a long
+tail of manual debugging and investigation necessary to reproduce every quirk and
+idiosyncrasy of your old build system in your new Mill build.
+
+However, while _tedious_, such migrations are not usually not _difficult_. Most
+build systems use a relatively small set of third-party tools with small amounts of
+custom logic, and Mill has built-in integrations with all the common JVM tools and
+makes custom logic easy to implement. In the end the decision to migrate comes down to
+the benefits of Mill (see xref:comparisons/why-mill.adoc[]) outweighing the cost of migration,
+which becomes more true as the lifespan and pace of development on a project grows.
```
lefou, line 283:
```suggestion
However, while _tedious_, such migrations are usually not _difficult_. Most
build systems use a relatively small set of third-party tools with small amounts of
custom logic, and Mill has built-in integrations with many common JVM tools and
makes custom logic easy to implement. In the end the decision to migrate comes down to
the benefits of Mill (see xref:comparisons/why-mill.adoc[]) outweighing the cost of migration,
which becomes more true as the lifespan and pace of development on a project grows.
```

lefou (commented):
Looks good. I spotted some nits and added suggestions.

docs/modules/ROOT/pages/migrating/migrating.adoc
```diff
@@ -0,0 +1,283 @@
...
+   be left in-place and fully working.*
+
+** This ensures you have the flexibility to back out of the migration at any point in time
+** On completion, this allows you to
+   perform head-to-head comparisons between old and new build systems
+
+2. *A parallel Mill build is set up for the project.*
+
+** Sub-project `pom.xml` and `build.gradle` files need to be translated into Mill ``Module``s
+** Third-party dependencies need to be translated into Mill's def `ivyDeps`
```
jodersky, line 57:
```
def `ivyDeps` => `def ivyDeps`
```

jodersky (commented):
Looks good! It might make sense to also mention using separate build files for large projects.

gamlerhart:
One challange I sometimes had to port builds: 
The existing build tasks rely on `stamping` all over the file system to collaborate between parts. Like Maven puts something in one place, then a `webpack` step picks up from these pieces and places its stuff yet in another place. And so on and so forth. Especially if tools (webpack, bazel, etc) then rely on 'implicit' paths to find configuration etc.

This are portable, but sometimes a bit a head ache. You can `give up` and do dirty writes outside the `Task.dest` but then get the more fragile builds. Or be somewhat creative on how to wire the tasks so these stuff works better in isolation.

Anyway, not sure how to formulate this at this point. Probably would need a public example of this tension.

PR #4042, opened 2024-11-28
pythonlib: revert some doc changes

jodersky:
This undoes my last commit of #3998, thereby re-establishing the `ExampleCrossModulePython`. It is required to get some sensible docs for the `basic/` usecase.

However, I'm still not convinced we should share documentation source code between scalalib and pythonlib examples. While it may make sense to share code between JVM technologies, I think python is a completely different beast, and not all concepts translate 1-1. We can see this just with the amount of `find-replaces` there are in `ExampleCrossModulePython` to get the `basic/` example working, and yet the resulting docs still feels off in some places. Also, with so many replacements, the blast radius of scalalib docs becomes unknown: if one day someone makes an innocent change to docs in scalalib, that may have non-sensical results in the python docs.

Preferably, I would like to drop the `ExampleCrossModulePython` and just copy-paste text when necessary, at least until we have a proven pattern of reuse.

### DIFF
--- a/example/package.mill
+++ b/example/package.mill
@@ -65,7 +65,7 @@ object `package` extends RootModule with Module {
     object basic extends Cross[ExampleCrossModule](build.listIn(millSourcePath / "basic"))
   }
   object pythonlib extends Module {
-    object basic extends Cross[ExampleCrossModule](build.listIn(millSourcePath / "basic"))
+    object basic extends Cross[ExampleCrossModulePython](build.listIn(millSourcePath / "basic"))
     object dependencies extends Cross[ExampleCrossModule](build.listIn(millSourcePath / "dependencies"))
   }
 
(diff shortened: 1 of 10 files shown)


jodersky:
@himanshumahajan138

himanshumahajan138:
@jodersky to be honest when first time i wrote the docs for python examples i didn't used the snippets and upstream inheritances but @lihaoyi pointed out at that time that we should have something like `exampleCrossModuleJava` so after looking at kotlin example i got that idea of creating `exampleCrossModulePython` which extends the upstream java examples and java extends upstream scala examples.


I think @lihaoyi can sought out this...

lihaoyi:
@jodersky @himanshumahajan138 let's go without the example doc sharing for now for Python and leave it separate. We can merge them later if necessary

himanshumahajan138:
> @jodersky @himanshumahajan138 let's go without the example doc sharing for now for Python and leave it separate. We can merge them later if necessary

Ok that's great 💯

lihaoyi:
We can probably close this then

PR #4044, opened 2024-11-28
[WIP] Build binaries for Mill Client using Graal VM native-image

### LINKED ISSUE #4007: Use GraalVM Native Image for Mill Client
Mill's client is currently a Java/JVM application. This means it suffers from the normal JVM slow startup (>100ms) and warmup times, and it prevents us from moving more logic into the client otherwise it will become even slower. 

We should explore using GraalVM to generate cross-platform native image launchers for the Mill client, and integrate that into our `./mill` script. This will allow us to make the client smarter:

- Using Scala rather than Java
- Parsing command line arguments in the client, which would remove the `-i must be passed as first argument` restriction
- Allow the client to use Coursier to bootstrap the JVM for the server to use

Reference https://blogs.oracle.com/developers/post/building-cross-platform-native-images-with-graalvm

ajaychandran:
This PR builds binaries for Mill using [Graal VM native-image](https://www.graalvm.org/latest/reference-manual/native-image/overview/Options/) and adds Java/Scala/Kotlin examples for building native images.

Execution time for `mill --version`, with server running, shows over 5x improvement.
```bash
$ time ./mill --version
Mill Build Tool version 0.12.3
Java version: 11.0.25, vendor: Eclipse Adoptium, runtime: ~/.sdkman/candidates/java/11.0.25-tem
Default locale: en_US, platform encoding: UTF-8
OS name: "Linux", version: 6.8.0-49-generic, arch: amd64

real	0m0.113s
user	0m0.171s
sys	0m0.037s

$ time ./out/dist/native.dest/mill --version
Mill Build Tool version 0.12.3-14-b0feea
Java version: 11.0.25, vendor: Eclipse Adoptium, runtime: ~/.sdkman/candidates/java/11.0.25-tem
Default locale: en_US, platform encoding: UTF-8
OS name: "Linux", version: 6.8.0-49-generic, arch: amd64

real	0m0.021s
user	0m0.005s
sys	0m0.016s

```

Resolves #4007.


ajaychandran:
TBD:
- Add Github actions for building and uploading binaries for different platforms.
- Update bootstrap script to download platform specific binary.

@lihaoyi 
Should the existing distribution assembly be retained?

lihaoyi:
@ajaychandran lets retain the existing assembly for now; we'll likely want to offer both runners for a while until we have confidence in the robustness of the nee one, and even then it would be nice to have a fallback (e.g. in OS/architectures the new one doesnt support)

lefou:
Maybe, we should just use the `mill.javalib` package for the new modules and avoid any new type re-exports from `scalalib` and `kotlinlib`. @lihaoyi WDYT?

lihaoyi:
@lefou I think for now following the re-export convention is probably best, once we decide what to do in https://github.com/com-lihaoyi/mill/issues/3570 we may need to break compatibility in 0.13.0 anyway, but until then we should at least keep things consistent

lihaoyi:
Going to leave out windows support and cross-platform mill artifact publishing jobs to a follow up PR, but the main user-land native image workflows and Mill native image construction seems to work

lihaoyi:
Merged this; thanks for the contribution @ajaychandran, I'll pay out 1/3 of the bounty (500USD) for the work you did. Will open up follow up tickets for the other parts (publishing Mill artifacts, windows, etc.)

PR #3988, opened 2024-11-19
Bump sbt-jupiter-interface to 0.13.x

lihaoyi:
Copy of https://github.com/com-lihaoyi/mill/pull/3501 opened from a separate repo to avoid double testing

Seems like the issue is that the jupiter-test-interface is not configured correctly and is thus missing the test class name as part of the success error reporting

```bash
./mill -w 'example.javalib.testing[3-integration-suite].local.test'
```

```
[2951]   Test run started (JUnit Jupiter)
[2951]   Test run started (JUnit Jupiter)
[2951]   Test #helloworld() started
[2951]   Test #hello() started
[2951]   Test #helloworld() finished, took 0.014s
[2951]   Test #hello() finished, took 0.013s
[2951]   Test #world() started
[2951]   Test  finished, took 0.025s
[2951]   Test #world() finished, took 0.0s
[2951]   Test  finished, took 0.025s
[2951]   Test  finished, took 0.038s
[2951]   Test run finished: 0 failed, 0 ignored, 1 total, 0.05s
[2951]   Test  finished, took 0.037s
[2951]   Test run finished: 0 failed, 0 ignored, 2 total, 0.049s
```

PR #4048, opened 2024-11-29
bump sbt-jupiter-interface to 0.13.3

jodersky:
Retrying this after fixes were merged upstream

Supersedes #3988


lihaoyi:
https://github.com/sbt/sbt-jupiter-interface/pull/128 should fix the kotlin side

PR #4051, opened 2024-11-30
Fix: Stop minifying in `fastLinkJS`

lolgab:
The alternative to this is to make `scalaJSMinify` an anonymous task, which is not binary compatible (we can have this change in a future major version).
Mill doesn't have `fastLinkJS` and `fullLinkJS` scopes as Sbt which makes it difficult to have specific settings for `fullLinkJS` only. This PR makes Mill behave like 99% of people want, which is what Sbt does by default. The downside is that it's now impossible to test what happens with minify on `fastLinkJS``, but the workaround is to directly test on `fullLinkJS` in that scenario.


lihaoyi:
Looks good to me

Issue #4016, opened 2024-11-24
Use Scala-Native for Mill Client (1000USD Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 1000USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

Sister ticket to https://github.com/com-lihaoyi/mill/issues/4007. 

The Mill client contains ~24 Java files with zero third party dependencies:

* https://github.com/com-lihaoyi/mill/tree/main/main/client/src/mill/main/client
* https://github.com/com-lihaoyi/mill/tree/main/runner/client/src/mill/runner/client

 We should try to see if we can port it all to Scala-Native for improved startup performance and to allow us to use Scala libraries in the client. e.g. using MainArgs in the Mill client would alleviate the `-i must be passed as the first argument` restriction which is very annoying

The success criteria for this ticket is a PR which (a) translates all Mill client code from Java to Scala (b) cross-compiles the Mill client code using Scala-Native and (c) bundles the Scala native launcher together with the Mill server jar so the client is able to launch the server.

Quafadas:
I'd be tempted to dot take a tilt at this, but I'll probably be slow.

scarf005:
i'm also interested in working on this, but pointers how to incrementally migrate would be welcome: https://discord.com/channels/632150470000902164/1295786923410984981/1310137269801320498

Quafadas:
@scarf005 I'll leave it to you for the time being - if you're willing to publish your PR, we can try and work through it together, but it sounds to me as though you're doing the same things I would have been trying to do. 

My knowledge is not more detailed than what Haoyi set out above right now.

scarf005:
> (b) cross-compiles the Mill client code using Scala-Native and (c) bundles the Scala native launcher together with the Mill server jar so the client is able to launch the 

@lihaoyi sorry, but could this be explained in more detail? still quite new to mill, i've rewritten runner.client at https://github.com/scarf005/mill/tree/refactor/scala-native-mill-client but unsure which command to run to ensure `(b)` and `(c)` works correctly.

Quafadas:
@scarf005 If you try running this command;

`mill show runner.client.compile`

and check the `compile.dest` target dir. I guess, that it will be empty. Did you make the client module a `ScalaModule`? Otherwise, it will ignore *.scala.

Quafadas:
@scarf005 Just want to check - are you still looking at this?

[...]

PR #4057, opened 2024-12-01
WIP: Native client

Quafadas:
The start of an effort, to get a mill Scala Native client.

Here's what (seems) to be in place. 

1. Converted java -> scala
2. Removed api calls which are not native koscher
3. Proved nativeLink works
4. Tests are runnable

Unfortunately, I think this might be quite some way from a solution. 

I've observed that the native client, appears to fire up a new server on every incarnation. Increasingly, I believe this _may_ come down to some difference  / bug / expected behaviour in how Scala Native interacts with files. 

I can't seem to port some of the tests which handle file tailing operations, which seems to lend support to that idea. 

If someone wants to play - on the branch of this PR branch. 

` mill runner.client.nativeLink`

will emit a file at 

`out/runner/client/nativeLink.dest/out`. 

If I run an `installLocalCache` and run that file in the root of (say) a clone of upickle. I appear to get a (semi) functioning mill client. That starts a server every time (!)... but it does compile and run things. And it starts very quickly, so there;s that I guess. 

for #4016


Quafadas:
https://discord.com/channels/632150470000902164/635668881951686686/1312831702384640010

PR #4058, opened 2024-12-01
[WIP] Fixes: #3928 [Module Examples] Add First Class Python Support

himanshumahajan138:
# Pull Request 
Adds First Class Python Support [Module Examples]
Fixes: #3928

## Description
Module Examples for Python `1-common-config`, `2-custom-tasks`, `3-override-tasks`, `4-compilation-execution-flags`, `5-resources`, `6-pex-config` Examples

## Related Issues

## Checklist

- [x]  1-common-config
- [x]  2-custom-tasks
- [x]  3-override-tasks
- [x]  4-compilation-execution-flags
- [x]  5-resources
- [ ]  6-pex-config
- [ ] Updated Documentation

## Status
WIP Addition & Require Review!!!

### DIFF
--- a/example/pythonlib/module/1-common-config/build.mill
+++ b/example/pythonlib/module/1-common-config/build.mill
@@ -1,0 +1,75 @@
+package build
+import mill._, pythonlib._
+
+object `package` extends RootModule with PythonModule {
+  // You can have arbitrary numbers of third-party libraries
+  def pythonDeps = Seq("MarkupSafe==3.0.2", "Jinja2==3.1.4")
+
+  // choose a main Script to run if there are multiple present
+  def mainScript = Task.Source { millSourcePath / "custom-src" / "foo2.py" }
+
+  // TODO: we have to choose whether to include `millSourcePath` by default in PYTHONPATH or not.
+  // def sources = Task.Sources {
+  // 	super.sources() ++ Seq(PathRef(millSourcePath))
+  // }
+
+  // def resources = Task.Sources {
+  //   super.resources() ++ Seq(PathRef(millSourcePath))
+  // }
+
+  def generatedSources: T[Seq[PathRef]] = Task {
+    val destPath = Task.dest / "generatedSources"
+    os.makeDir.all(destPath)
+    for (name <- Seq("A", "B", "C")) os.write(
+      destPath / s"foo$name.py",
+      s"""
+         |class Foo$name:
+         |    value = "hello $name"
+      """.stripMargin
+    )
+
+    Seq(PathRef(destPath))
+  }
+
+  def forkEnv: T[Map[String, String]] = Map("MY_CUSTOM_ENV" -> "my-env-value")
+
+  // Additional Python options e.g. to Turn On Warnings
+  // we can use -Werror to treat warnings as errors
+  def pythonOptions: T[Seq[String]] = Seq("-Wall")
+
+}
+
+/** Usage
+
+> ./mill run
+...
+Foo2.value: <h1>hello2</h1>
+Foo.value: <h1>hello</h1>
+FooA.value: hello A
+FooB.value: hello B
+FooC.value: hello C
+MyResource: My Resource Contents
+MyOtherResource: My Other Resource Contents
+MY_CUSTOM_ENV: my-env-value
+...
+
+> ./mill show bundle
+".../out/bundle.dest/bundle.pex"
+
+> ./out/bundle.dest/bundle.pex
+...
+Foo2.value: <h1>hello2</h1>
+Foo.value: <h1>hello</h1>
+FooA.value: hello A
+FooB.value: hello B
+FooC.value: hello C
+MyResource: My Resource Contents
+MyOtherResource: My Other Resource Contents
+...
+
+> sed -i.bak 's/import os/import os, warnings; warnings.warn("This is a test warning!")/g' custom-src/foo2.py
+
+> ./mill run
+...UserWarning: This is a test warning!...
+
+*/
--- a/example/pythonlib/module/4-compilation-execution-flags/build.mill
+++ b/example/pythonlib/module/4-compilation-execution-flags/build.mill
@@ -1,0 +1,15 @@
+package build
+import mill._, pythonlib._
+
+object `package` extends RootModule with PythonModule {
+  def mainScript = Task.Source { millSourcePath / "src" / "foo.py" }
+  def pythonOptions = Seq("-Wall", "-Xdev")
+  def forkEnv = Map("MY_ENV_VAR" -> "HELLO MILL!")
+}
+
+/** Usage
+
+> ./mill run
+HELLO MILL!
+
+*/
--- a/example/pythonlib/module/5-resources/foo/test/src/test.py
+++ b/example/pythonlib/module/5-resources/foo/test/src/test.py
@@ -1,0 +1,36 @@
+import os
+from pathlib import Path
+import importlib.resources
+import unittest
+from foo import Foo  # type: ignore
+
+
+class TestScript(unittest.TestCase):
+    def test_all(self) -> None:
+        appClasspathResourceText = Foo().PythonPathResourceText("resources", "file.txt")
+        self.assertEqual(appClasspathResourceText, "Hello World Resource File")
+
+        testClasspathResourceText = (
+            importlib.resources.files("resources")
+            .joinpath("test-file-a.txt")
+            .read_text()
+            .strip()
+        )
+        self.assertEqual(testClasspathResourceText, "Test Hello World Resource File A")
+
+        testFileResourceFile = next(
+            Path(os.getenv("MILL_TEST_RESOURCE_DIR")).rglob("test-file-b.txt"), None
+        )
+        with open(testFileResourceFile, "r", encoding="utf-8") as file:
+            testFileResourceText = file.readline()
+        self.assertEqual(testFileResourceText, "Test Hello World Resource File B")
+
+        with open(
+            Path(os.getenv("OTHER_FILES_DIR"), "other-file.txt"), "r", encoding="utf-8"
+        ) as file:
+            otherFileText = file.readline()
+        self.assertEqual(otherFileText, "Other Hello World File")
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/pythonlib/src/mill/pythonlib/PythonModule.scala
+++ b/pythonlib/src/mill/pythonlib/PythonModule.scala
@@ -75,6 +75,9 @@ trait PythonModule extends PipModule with TaskModule { outer =>
    */
   def unmanagedPythonPath: T[Seq[PathRef]] = Task { Seq.empty[PathRef] }
 
+  def generatedSources: T[Seq[PathRef]] = Task { Seq.empty[PathRef] }
+
+  // TODO: we have to choose whether to include `millSourcePath` by default in PYTHONPATH or not.
   /**
    * The directories used to construct the PYTHONPATH for this module, used for
    * execution, excluding upstream modules.
@@ -83,7 +86,9 @@ trait PythonModule extends PipModule with TaskModule { outer =>
    * directories.
    */
   def localPythonPath: T[Seq[PathRef]] = Task {
-    sources() ++ resources() ++ unmanagedPythonPath()
+    Seq(
+      PathRef(millSourcePath)
+    ) ++ sources() ++ resources() ++ generatedSources() ++ unmanagedPythonPath()
   }
 
   /**
@@ -95,17 +100,22 @@ trait PythonModule extends PipModule with TaskModule { outer =>
     localPythonPath() ++ upstream
   }
 
+  def forkEnv: T[Map[String, String]] = Task { Map.empty[String, String] }
+
+  def pythonOptions: T[Seq[String]] = Task { Seq.empty[String] }
+
   // TODO: right now, any task that calls this helper will have its own python
   // cache. This is slow. Look into sharing the cache between tasks.
   def runner: Task[PythonModule.Runner] = Task.Anon {
     new PythonModule.RunnerImpl(
       command0 = pythonExe().path.toString,
+      options = pythonOptions(),
       env0 = Map(
         "PYTHONPATH" -> transitivePythonPath().map(_.path).mkString(java.io.File.pathSeparator),
         "PYTHONPYCACHEPREFIX" -> (T.dest / "cache").toString,
         if (Task.log.colored) { "FORCE_COLOR" -> "1" }
         else { "NO_COLOR" -> "1" }
-      ),
+      ) ++ forkEnv(),
       workingDir0 = Task.workspace
     )
   }
@@ -194,6 +204,7 @@ object PythonModule {
 
   private class RunnerImpl(
       command0: String,
+      options: Seq[String],
       env0: Map[String, String],
       workingDir0: os.Path
   ) extends Runner {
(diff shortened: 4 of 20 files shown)


himanshumahajan138:
(@lihaoyi and @jodersky) Sir Let's Review this PR as Backbone support (Documentation will be updated after finalizing the Code)

Also @lihaoyi guide me for what kind of configs we needs to add for pex config example are we targeting specific bundle args or something else ???

himanshumahajan138:
@lihaoyi Sir please guide me,

**Why this error is coming where i have used show command in cli test case???**
```
[3297]   utest.AssertionError: else assert(evalResult.isSuccess)
[3297]   evalResult: mill.testkit.IntegrationTester.EvalResult = EvalResult(false,mill-server/ exitCode file not found
[3297]   ,)
[3297]     utest.asserts.Asserts$.assertImpl(Asserts.scala:30)
[3297]     mill.testkit.ExampleTester.validateEval(ExampleTester.scala:145)
[3297]     mill.testkit.ExampleTester.processCommand(ExampleTester.scala:134)
[3297]     mill.testkit.ExampleTester.processCommandBlock(ExampleTester.scala:95)
[3297]     mill.testkit.ExampleTester.$anonfun$run$2(ExampleTester.scala:194)
[3297]     mill.testkit.ExampleTester.$anonfun$run$2$adapted(ExampleTester.scala:194)
[3297]     scala.collection.ArrayOps$.foreach$extension(ArrayOps.scala:1323)
[3297]     mill.testkit.ExampleTester.run(ExampleTester.scala:194)
[3297]     mill.testkit.ExampleTester$.run(ExampleTester.scala:61)
[3297]     mill.testkit.UtestExampleTestSuite$.$anonfun$tests$3(UtestExampleTestSuite.scala:19)
[3297]     scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
[3297]     mill.api.Retry.$anonfun$apply$1(Retry.scala:32)
[3297]     mill.api.Retry.$anonfun$apply$1$adapted(Retry.scala:32)
[3297]     mill.api.Retry.$anonfun$indexed$2(Retry.scala:42)
[3297]     scala.util.Try$.apply(Try.scala:[217](https://github.com/com-lihaoyi/mill/actions/runs/12116726359/job/33777897135?pr=4058#step:11:218))
[3297]     mill.api.Retry.$anonfun$indexed$1(Retry.scala:42)
[3297]     java.lang.Thread.run(Thread.java:840)
```

i am not getting why this error is happening please help...

example/pythonlib/module/1-common-config/build.mill
```diff
@@ -0,0 +1,77 @@
+package build
+import mill._, pythonlib._
+
+// TODO: When we use `package` instead of foo then it fails for `show` command in CI test Why?
```
jodersky, line 4: If you use `package`, then you'd need to update the command `mill foo.XXX` to `mill XXX`. Does that still fail?

himanshumahajan138: Yaa i was doing exactly the same thing but its was not working

pythonlib/src/mill/pythonlib/PythonModule.scala
```diff
@@ -75,6 +75,9 @@ trait PythonModule extends PipModule with TaskModule { outer =>
    */
   def unmanagedPythonPath: T[Seq[PathRef]] = Task { Seq.empty[PathRef] }
 
+  def generatedSources: T[Seq[PathRef]] = Task { Seq.empty[PathRef] }
```
jodersky, line 78: You can include a comment what this does. I think it's fine to reuse the one from JavaModule
```
/**
   * Folders containing source files that are generated rather than
   * handwritten; these files can be generated in this target itself,
   * or can refer to files generated from other targets
   */
```

pythonlib/src/mill/pythonlib/PythonModule.scala
```diff
@@ -95,17 +100,22 @@ trait PythonModule extends PipModule with TaskModule { outer =>
     localPythonPath() ++ upstream
   }
 
+  def forkEnv: T[Map[String, String]] = Task { Map.empty[String, String] }
+
+  def pythonOptions: T[Seq[String]] = Task { Seq.empty[String] }
```
jodersky, line 105: please also add comments to these

pythonlib/src/mill/pythonlib/PythonModule.scala
```diff
@@ -83,7 +86,9 @@ trait PythonModule extends PipModule with TaskModule { outer =>
    * directories.
    */
   def localPythonPath: T[Seq[PathRef]] = Task {
-    sources() ++ resources() ++ unmanagedPythonPath()
+    Seq(
+      PathRef(millSourcePath)
```
jodersky, line 90: >   // TODO: we have to choose whether to include `millSourcePath` by default in PYTHONPATH or not.

PYTHONPATH is used by Python to look up source files (and resources), so I'm not sure why the mill module should be on there too. What is the argument for including millSourcePath?

example/pythonlib/module/5-resources/foo/test/src/test.py
```diff
@@ -0,0 +1,36 @@
+import os
+from pathlib import Path
+import importlib.resources
+import unittest
+from foo import Foo  # type: ignore
+
+
+class TestScript(unittest.TestCase):
+    def test_all(self) -> None:
+        appClasspathResourceText = Foo().PythonPathResourceText("resources", "file.txt")
```
jodersky, line 10: minor: maybe avoid the term "classpath", since it might be unfamiliar to python devs. I think "path" or "pythonpath" is fine

jodersky (commented):
In general this looks like it's on the right track.

If you can add explanations to the examples I can give it another look

pythonlib/src/mill/pythonlib/PythonModule.scala
himanshumahajan138: Actually i haven't updated the docs yet, i do that after code satisfaction 🙂

pythonlib/src/mill/pythonlib/PythonModule.scala
himanshumahajan138: See in jvm just like we have resources pool right, so in python we can also have the same thing that's by adding the millsourcepath in the pythonpath this eill allow us to load all the sources and resources using the class names 
Hope you understand

example/pythonlib/module/5-resources/foo/test/src/test.py
himanshumahajan138: Sorry it's just like typo coz i focused on code logics only and more tuning will be done on the go 
I'll update

pythonlib/src/mill/pythonlib/PythonModule.scala
jodersky: I don't understand how this would change anything.

Does my comment on https://github.com/com-lihaoyi/mill/issues/3928#issuecomment-2515525399 apply to this?

> regarding importlib: I don't think it's possible to read resources from the "root" package, but I also don't think it's a common thing to want to do. Typically, resources will live in at least one directory corresponding to the name of a python module. For example, you could have the following source layout:
>
>```
>foo/
>  - src/foo/main.py
>  - resources/foo/data.txt
>```
>
>then access the data.txt in main.py with importlib.resources.open_text("foo", "data.txt")

Also one thing I've noticed is that importlib doesn't play nice when python module and package have the same name. E.g a `foo/foo.py`, a folder named `foo` and a file named `foo.py`, then python will always look relative to the file. Could this be the cause of the issues you're having with resources? If so, I'd suggest renaming the files, since this is just the way python behaves.

himanshumahajan138: Ya you said right i think if the file and dir have the same name then it will clash for sure let me see a better way to handle this on official docs 🙂

himanshumahajan138: Thanks sir you are really very helpful ☺️

example/pythonlib/module/5-resources/foo/test/src/test.py
himanshumahajan138: done!

example/pythonlib/module/1-common-config/build.mill
himanshumahajan138: >   /**
>   * Command-line options to pass to the Python Interpreter defined by the user.
>   */
any more suggestion why it's not working

pythonlib/src/mill/pythonlib/PythonModule.scala
himanshumahajan138: done!

pythonlib/src/mill/pythonlib/PythonModule.scala
himanshumahajan138: for now i haven't fixed it properly but added paths just the same way we do in the build.mill file 

PS: We can do it in better way

himanshumahajan138:
@lihaoyi @jodersky 
Ready for Review!

himanshumahajan138:
@lihaoyi @jodersky anyone Up There ???

lihaoyi:
@jodersky I'll leave this review to you, feel free to merge this when you think it's in good state and I'll pay out the bounty

example/pythonlib/module/1-common-config/build.mill
```diff
@@ -0,0 +1,99 @@
...
+
+  // TODO: should we include only one(it is sufficient) or both tasks(for demo purpose)?
+  // Add (or replace) source folders for the module to use
+  def sources = Task.Sources {
+    super.sources() ++ Seq(PathRef(millSourcePath))
+  }
+
+  // Add (or replace) resource folders for the module to use
+  def resources = Task.Sources {
+    super.resources() ++ Seq(PathRef(millSourcePath))
```
jodersky, line 25: I would suggest not using `millSourcePath` in this example, since it has the potential to lead to overlapping PYTHONPATHS, which can be confusing. For example, as it currently stands, we can access `MyResource.txt` at the root of a pythonpath, by `MyOtherResources.txt` lives under the python module `custom-resources`.

For simplicity, I would suggest using a path at the same level as the existing `resouces/` path. Perhaps `custom-resources`, and you nest `MyOtherResoucre` in another folder.

example/pythonlib/module/6-pex-config/build.mill
```diff
@@ -0,0 +1,68 @@
...
+  def mainScript = Task.Source { millSourcePath / "src" / "foo.py" }
+
+  // These Dependencies will be used in the bundle
+  def pythonDeps = Seq("pyfiglet==1.0.2", "termcolor==2.5.0")
+
+  // Python Options to use while creating bundle
+  def pythonOptions: T[Seq[String]] =
+    Seq("-Wall", "-Wignore::ImportWarning", "-Wignore::ResourceWarning")
+
+  // TODO: Should we allow user to change the default config/function available: `bundle`?
```
jodersky, line 19: I think this is a good idea!

example/scalalib/module/2-custom-tasks/build.mill
```diff
@@ -61,7 +61,7 @@ object `package` extends RootModule with ScalaModule {
 // digraph G {
 //   rankdir=LR
 //   node [shape=box width=0 height=0]
-//   ivyDeps -> generatedSources -> compile -> "..."
```
jodersky, line 64: is this change to `scalalib` intentional?

example/pythonlib/module/4-compilation-execution-flags/build.mill
```diff
@@ -0,0 +1,40 @@
...
+// since you avoid the JVM startup, but does not support `forkEnv`.
+//
+// If you want to pass main-method arguments to `run` or `runLocal`, simply pass
+// them after the `foo.run`/`foo.runLocal`:
+//
+// [source,bash]
+// ----
+// mill foo.run arg1 arg2 arg3
+// mill foo.runLocal arg1 arg2 arg3
+// ----
```
jodersky, line 40: `runLocal` doesn't exist in `PythonModule`, and the concept doesn't really apply. You don't need to include it in the example.

jodersky (commented):
This is starting to look good for a merge. I left a couple of comments, but once those are cleared we can merge it.

example/scalalib/module/2-custom-tasks/build.mill
himanshumahajan138: 😂 sorry by mistake (haven't crosschecked)

example/pythonlib/module/6-pex-config/build.mill
himanshumahajan138: then we need three tasks one is `devBundleOptions`(default config), `bundleOptions`(user will add more customisation) and `bundle`(which will combine these two but if user want to change deafult then they can alter it)

Give me better names if you have in mind

example/pythonlib/module/1-common-config/build.mill
himanshumahajan138: i have used one approach before where i am adding the `resources` and `custom-src` and others in the PYTHONPATH(means in their tasks) and then add a **`src`**  folder to give nest feel this fix the problem 

is this acceptable ?

jodersky: If you put `src` in resources, and `resources` is in the `PYTHONPATH`, then `src` becomes a python module (which is more or less the same as a java package). That means, that if you want to load the resource with import lib, you need to call `importlib.resources.read_text("src", "data.txt")`. Instead of using `src`, why not reuse the same module you already have, `foo` in this case?

example/pythonlib/module/6-pex-config/build.mill
jodersky: Yeah, I'd say let's keep the default options hardcoded. We can revisit this later on if needed

himanshumahajan138: let me give you a better version of this give me 10 mins i'll be there with update

example/pythonlib/module/1-common-config/build.mill
himanshumahajan138: i tried `foo.resources` and `foo.custom-resources` (also with `resources` and `custom-resources`)and it said ModuleNotFoundError: No module named 'foo.resources'; 'foo' is not a package

This means we have to add `millSourcePath` to the PYTHONPATH for making it's subdirectories as Module and this bring us to again same thing as we have now

himanshumahajan138: @jodersky Sir let's do resources/ res and custom-resources/ custom-res 

is this fine ?

himanshumahajan138:
@jodersky Let's Finalize this if satisfied

example/pythonlib/module/1-common-config/build.mill
jodersky: > i tried foo.resources and foo.custom-resources (also with resources and custom-resources)and it said ModuleNotFoundError: No module named 'foo.resources'; 'foo' is not a package

You wouldn't need `foo.resources`. Essentially, if PYTHONPATH includes `/a/b/c`, and your resource is in `/a/b/c/x/y/z`, then you would access it in importlib with `x.y.z`.

Anyway, I think this example is fine as a first step, let's merge this

himanshumahajan138: Actually sir we both have same thoughts but i am unable to explain completely 😂

jodersky:
Thanks!

himanshumahajan138:
Great, @jodersky and @lihaoyi Thanks for your Guidance...

jodersky:
@himanshumahajan138, one minor comment: for the following PRs, could you please not say "fixed #3928" in the title, but instead "part of #3928". If you include the words "fixed" or "closes", then github will automatically close the linked issue, which we don't want in this case

himanshumahajan138:
> @himanshumahajan138, one minor comment: for the following PRs, could you please not say "fixed #3928" in the title, but instead "part of #3928". If you include the words "fixed" or "closes", then github will automatically close the linked issue, which we don't want in this case

OK Sir 😊 i will take care for future PR's...

lihaoyi:
Thanks @himanshumahajan138, I'll transfer this portion of the bounty using your existing bank details

Issue #2816, opened 2023-10-03
[BSP] Report completion percentage during compilation


lolgab:
Sbt sends compilation progress with a special `CompileProgress` trait implementation [here](https://github.com/sbt/sbt/blob/397c7a634a640cfad5c6d5dbfc08a3d1459ab4fc/main/src/main/scala/sbt/internal/server/BspCompileProgress.scala#L13).
Mill should do the same to send compilation progress to BSP clients.

PR #4064, opened 2024-12-02
[BSP] Report compilation's completion percentage

lolgab:
Fixes https://github.com/com-lihaoyi/mill/issues/2816

https://github.com/user-attachments/assets/8ff8f141-f217-40c0-a75f-8f4e87597391

Pull Request: https://github.com/com-lihaoyi/mill/pull/4064

### DIFF
--- a/main/api/src/mill/api/CompileProblemReporter.scala
+++ b/main/api/src/mill/api/CompileProblemReporter.scala
@@ -14,7 +14,9 @@ trait CompileProblemReporter {
   def fileVisited(file: os.Path): Unit
   def printSummary(): Unit
   def finish(): Unit
-  def notifyProgress(percentage: Long, total: Long): Unit
+  // TODO: cleanup once we break bin-compat in Mill 0.13
+  // Remove default implementation: `= ()`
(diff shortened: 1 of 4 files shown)


main/api/src/mill/api/CompileProblemReporter.scala
```diff
@@ -14,6 +14,7 @@ trait CompileProblemReporter {
   def fileVisited(file: os.Path): Unit
   def printSummary(): Unit
   def finish(): Unit
+  def notifyProgress(percentage: Long, total: Long): Unit
```
lefou, line 17: I'm surprised this is binary compatible. I would have expected that we need to provide a default impl.

lefou (commented):
Looks good to me.

main/api/src/mill/api/CompileProblemReporter.scala
lolgab: It wasn't. I added it a default impl in the last commit. Please review again.

lihaoyi (approved):

Issue #3930, opened 2024-11-09
Standardize usage of machine-wide caches


lihaoyi:
In the past we only had Coursier managing its own machine-global cache for maven central, but this is starting to grow:

- Custom Java version support needs somewhere to put the JVMs
- Android support needs somewhere to put the Android SDK and other heavyweight tools
- The example PythonModule needs somewhere to cache its PIP downloads
- The example TypescriptModule could use somewhere to cache its NPM downloads
- Once the filesystem-independent out/ folder layout lands, we could begin sharing out/ folder results between projects, and that would also need some standard place to put the artifacts

The basic idea is that when someone calls `clean`, they usually dont want to start all the way from scratch re-downloading every jar from Maven Central and re-downloading their JVM. These external dowloads typically are downloaded once and cached forever, and only very rarely do people want to clean them, in comparison to needing to clean local build outputs.

We should try and standardize how these "global" cached downloads are handled so people can just plug into the standard, rather than creatively coming up with their own solutions that end up being half-baked or inconsistent

0xnm:
My 2c on this: such cache structure vary a lot between the different ecosystems, because the particular ecosystem dictates the structure.

For example, in case of Maven dependencies the standard file cache layout supports artifact versioning as a part of the artifact's full path and it is easy to have different artifact versions out-of-project, somewhere on the machine.

If we take a look on the NPM dependencies, for example, `node_modules` doesn't support hosting different versions of the same artifact by default (unless package-aliasing is used? or maybe workspaces can help?). In this case such cache unlikely can be machine-wide and then becomes more project-wide (especially if we think that some dependencies are built-in, e.g. test reporting in Kotlin/JS, and versions can be different between different Mill installations). Now, let's say we have a project with several JS modules, and it won't be very wise to download the same NPM dependency several times (for each module), so it probably makes sense to have some folder at the root of the project, but there is no:

* built-in API to access root module (or root `out` folder). The closest way to access it (instead of using `T.dest` which is bound to the task) with the current API will be, probably, something like:

```
var m: mill.define.Module = this
while (m.millOuterCtx.enclosingModule != null) {
  m = m.millOuterCtx.enclosingModule.asInstanceOf[mill.define.Module]
}
val dir = m.millModuleBasePath.value / "out" / "js"
```

[...]

PR #4066, opened 2024-12-03
Download test repos via coursier

alexarchambault:
So that the repo archives are cached, and the tasks that use them work offline.

That's useful to build the website while being offline, for example.

### DIFF
--- a/ci/shared.mill
+++ b/ci/shared.mill
@@ -1,8 +1,15 @@
 package build.ci
 
+import coursier.cache.FileCache
+import coursier.util.Artifact
 
-
-def downloadTestRepo(label: String, commit: String, dest: os.Path) = {
-  os.unzip.stream(requests.get.stream(s"https://github.com/$label/archive/$commit.zip"), dest)
+def downloadTestRepo(label: String, commit: String, dest: os.Path)(implicit ctx: mill.api.Ctx.Log) = {
+  val cache = FileCache()
+  // Uncomment when rebootstrapping Mill, to get better logging
+  //   .withLogger(mill.util.CoursierSupport.millLogger())
+  val file = cache.file(Artifact(s"https://github.com/$label/archive/$commit.zip"))
+    .run.unsafeRun()(cache.ec)
+    .fold(ex => throw new Exception(ex), identity)
+  os.unzip(os.Path(file), dest)
   dest
 }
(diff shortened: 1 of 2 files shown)


ci/shared.mill
```diff
@@ -1,8 +1,15 @@
...
+import coursier.util.Artifact
 
-
-def downloadTestRepo(label: String, commit: String, dest: os.Path) = {
-  os.unzip.stream(requests.get.stream(s"https://github.com/$label/archive/$commit.zip"), dest)
+def downloadTestRepo(label: String, commit: String, dest: os.Path)(implicit ctx: mill.api.Ctx.Log) = {
+  val cache = FileCache()
+  // Uncomment when rebootstrapping Mill, to get better logging
+  //   .withLogger(mill.util.CoursierSupport.millLogger())
+  val file = cache.file(Artifact(s"https://github.com/$label/archive/$commit.zip"))
```
gamlerhart, line 10: Noob question: I just can create a file cache and download a arbitrary file?
Where is this then cached and what is its cache key? I assume the key is the URL?

Because this seems very useful in general. I have a 'manual' cache so far.

lihaoyi:
I think this should be part of a broader discussion in https://github.com/com-lihaoyi/mill/issues/3930. Should Coursier be the recommended way to download-and-cache stable URLs? All stable URLs or only some stable URLs? What are the considerations and alternatives here?

Obviously this PR works for this specific scenario and does what it says on the box, but we really should have a more rigorous analysis of what the options are, and if this is the recommended option then it should not be used just here but instead also makes its way into the user-facing documentation

alexarchambault:
@lihaoyi Sure

> I think this should be part of a broader discussion in #3930. Should Coursier be the recommended way to download-and-cache stable URLs? All stable URLs or only some stable URLs? What are the considerations and alternatives here?

I didn't think much about alternatives, but I don't think there are that many out there 🤔 As discussed in https://github.com/com-lihaoyi/mill/issues/3930, other ecosystems tend to have caches quite tied to their ecosystem. The cache of coursier is much more generic OTOH. AFAIK, only the Nix cache offers similar features.

alexarchambault:
Letting you decide if you'd rather merge this or not, while things are being discussed in https://github.com/com-lihaoyi/mill/issues/3930

lihaoyi:
as I mentioned in https://github.com/com-lihaoyi/mill/issues/3930, let's hold off on this for the moment and come back after general support in Mill is in place

PR #4067, opened 2024-12-03
Use bomDeps when importing Maven projects

alexarchambault:
This tries to take into account "import" dependencies when importing Maven projects (which correspond to `bomModulesDeps` and `bomIvyDeps` in Mill), and adds parent modules there too (parent modules are basically BOMs).

Checking if this breaks the current tests. Extra tests might need to be added, to test these changes.


alexarchambault:
CI error looks like a flake

lihaoyi:
Lemme kick the CI and see if it turns green

lihaoyi:
@alexarchambault seems the test is failing reliably on CI, does it pass on your machine?

alexarchambault:
So adding the parent POM to `bomDeps` during Maven imports is a problem. `bomDeps` can only be external BOMs, fetched from a Maven repository. If the parent is a module of the project being imported, then it's not found on Maven repos, and the generated Mill build fails.

This needs a missing part of BOM support, something like `BomModule`, where the dependency management defined in the Maven parent module would be written to (in `JavaModule#depManagement`), that other modules could refer to via a new `bomModules` say (like `bomDeps`, but for local modules), and that would get published as a BOM (no sources / no JAR, `pom` artifact type with `dependencyManagement` section).

alexarchambault:
The failing example is [this one](https://github.com/com-lihaoyi/mill/blob/b5255950a5d4c4f59db13e83b38349f8987350a0/example/javalib/migrating/2-maven-incomplete/build.mill). It runs into the parent POM issue above if we add `bomDeps` for parent POMs to it.

lihaoyi:
@alexarchambault perhaps `bomDeps` should be named `bomIvyDeps`, and then have a separate `bomModuleDeps` to refer to local modules? That would follow the naming convention we use elsewhere in `JavaModule`

alexarchambault:
About this PR: it tries to set `JavaModule#{bomModuleDeps,bomIvyDeps}` in a sensible way when importing Maven projects. It seems it doesn't break the current tests, but I didn't try it on actual Maven projects that need those tasks. So I'm not sure if it should be merged right now.

But if any one ran into issues with empty versions or faulty class paths, when importing Maven projects, this might help.

lihaoyi:
Closing this for now since it's stale, feel free to re-open a new one if necessary

PR #4068, opened 2024-12-03
Add BOM support for compileIvyDeps, runIvyDeps, and BOM test dependencies

alexarchambault:
~Checking what the CI says, this embeds https://github.com/com-lihaoyi/mill/pull/3924 for now~


alexarchambault:
Will need to be rebased on top of https://github.com/com-lihaoyi/mill/pull/4081

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -191,6 +200,23 @@ trait JavaModule
...
+   */
+  def extraBomIvyDeps: Task[Agg[BomDependency]] = Task.Anon { Agg.empty[BomDependency] }
+
+  /**
+   * Dependency management entries that are not meant to be overridden or changed by users.
+   *
+   * This is mainly used to add dependency management entries of the main module to its test
+   * modules.
+   */
+  def extraDepManagement: Task[Agg[Dep]] = Task.Anon { Agg.empty[Dep] }
```
lihaoyi, line 218: Do these really need to be `Task.Anon`s? Not sure about `BomDependency`, but at least `Agg[Dep]` is serializable and can be a normal `Task` right?

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -455,13 +504,62 @@ trait JavaModule
...
-    val bomDeps0 = allBomDeps().toSeq.map(_.withConfig(Configuration.compile))
+    val bomDeps0 =
+      allBomIvyDeps().toSeq.map(_.withConfig(Configuration.compile)) ++ extraBomIvyDeps().toSeq
+    val depMgmt = processedDependencyManagement(
+      (extraDepManagement().toSeq ++ depManagement().toSeq)
+        .map(bindDependency())
+        .map(_.dep)
+    )
+
+    addBoms(bomDeps0, depMgmt, overrideVersions = overrideVersions)
```
lihaoyi, line 521: Can we refactor `processDependency`, `processCompileDependency`, and `processRunDependency` to share the logic between them via an anonymous task. That will help make it easier to see the core differences between them.

I'm also not clear about the differences here:

- `processCompileDependency` uses `compileDepManagement()`, and `processRunDependency` uses `runDepManagement`
- `processDependency` and `processCompileDependency` both use `extraDepManagement` but `processRunDependency` does not. Why?
- Only `processDependency` seems to use `extraBomIvyDeps`, while `processCompileDependency` and `processRunDependency` do not. Why?

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -479,11 +577,38 @@ trait JavaModule
     }
   }
 
+  /**
+   * The compile-time Ivy dependencies of this module, with BOM and dependency management details
+   * added to them. This should be used when propagating the dependencies transitively
+   * to other modules.
+   */
+  def processedCompileIvyDeps: Task[Agg[BoundDep]] = Task.Anon {
```
lihaoyi, line 585: Let's make these normal `Task`s rather than `Task.Anon`s?

lihaoyi:
Left some comments. @alexarchambault also please fill out the PR description, that will make it easier to review and maintain since we won't need to reverse engineer what you already have in your head

scalalib/src/mill/scalalib/JavaModule.scala
alexarchambault: These can be made normal `Task`s, yes

scalalib/src/mill/scalalib/JavaModule.scala
alexarchambault: This should also simplify the way BOMs are handled when taking scopes into account. I've been having issues reconciling Mill's `compileIvyDeps` / `ivyDeps` / `runIvyDeps` (where scopes are known beforehand) with how scopes are handled in BOMs (not known beforehand, dependencies from all scopes are mixed, and BOMs can change the scope if it's undefined).

alexarchambault:
I'll probably close this PR. By relying on https://github.com/com-lihaoyi/mill/pull/4145, https://github.com/com-lihaoyi/mill/pull/4154 offers simpler / better support for BOMs / scopes.

alexarchambault:
I think I should close this one. https://github.com/com-lihaoyi/mill/pull/4154, based on https://github.com/com-lihaoyi/mill/pull/4145, is a cleaner approach, that I trust more than the developments I pushed here. IIRC, I was having a hard time making sense of the changes here, in particular in the light of https://github.com/com-lihaoyi/mill/pull/4068#discussion_r1886008054. Better go with https://github.com/com-lihaoyi/mill/pull/4154 and then https://github.com/com-lihaoyi/mill/pull/4155, even if this needs to wait a bit for `0.13`.

PR #4069, opened 2024-12-03
Don't swallow exception stack trace in mill.api.Exception

alexarchambault:
This keeps the original exception as cause of mill.api.Exception, so that its message, stack traces, and those of its causes, are usually printed when a mill.api.Exception gets thrown


alexarchambault:
See [this gist](https://gist.github.com/alexarchambault/bc363f4238057e43d29f7b017c2c5e72) for a before / after of an exception in unit tests. The latter has the original exception, so is easier to debug.

lihaoyi:
Thanks!

PR #4077, opened 2024-12-05
Write runtime dependencies in ivy.xml

alexarchambault:
It seems `runIvyDeps` were not written in `ivy.xml` files, and thus were not resolved when publishing a Mill project locally. This fixes that.


alexarchambault:
The coursier interface bump is necessary to be able to resolve compile scope dependencies with it from the added test (thanks to https://github.com/coursier/interface/pull/405)

PR #4082, opened 2024-12-06
Stop using "_" to mean empty version / version to read in BOM

alexarchambault:
Follow-up of https://github.com/com-lihaoyi/mill/pull/3924


alexarchambault:
This needs a new release of coursier, that has https://github.com/coursier/coursier/pull/3185

Issue #657, opened 2019-07-11
Generalize `.runBackground` to work with all kinds of subprocesses


lihaoyi-databricks:
Currently it's hardcoded to only work for the main method of Java/Scala modules, but it could be useful for running other sorts of long-running daemons too: webpack proxy, `docker run`, databases, etc.

We might need to replace the auto-shutdown thread with a auto-shutdown subprocess, since we'll be running non-JVM things

lihaoyi:
Maybe could even make it a general-purpose wrapper that works with all kinds of Tasks as well

lihaoyi:
We've hit the need for a non-JVM runBackground now that we're aiming to support Javascript https://github.com/com-lihaoyi/mill/pull/4022/files and Python https://github.com/com-lihaoyi/mill/issues/3928.

himanshumahajan138:
@lihaoyi actually yesterday i was working on this same issue for web examples and i was mimicing the same code as we have for java/scala like having backgroundwrapper in .py extension and then run that file with all the necessary args.

I will pull a request for this in some time and i think untill we don't get general support we can go with this one

Your thoughts?

lihaoyi:
@himanshumahajan138 I think it should be doable to generalize `MillBackgroundWrapper` to wrap a subprocess, lemme try and implement that

Issue #3772, opened 2024-10-19
Clean up `Jvm.*` subprocess/inprocess APIs (500USD Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 500USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

The various `Jvm.*` methods are a mess. https://github.com/com-lihaoyi/mill/pull/3770 makes them marginally more messy, but they've been messy for years and could use a cleanup.

The goal of this issue is to consolidate all subprocess and classloader spawning operations in `Jvm.scala` into (approximately) the following four signatures

```scala
  def spawn(mainClass: String,
                      classPath: Iterable[os.Path],
                      jvmArgs: Seq[String],
                      mainArgs: Seq[String],
                      env: Map[String, String] = null,
                      cwd: Path = null,
                      stdin: ProcessInput = Pipe,
                      stdout: ProcessOutput = Pipe,
                      stderr: ProcessOutput = os.Inherit,
                      mergeErrIntoOut: Boolean = false,
                      propagateEnv: Boolean = true,
                      useCpPassingJar: Boolean = false): os.SubProcess = {

    val commandArgs = jvmCommandArgs(javaExe, mainClass, jvmArgs, classPath, mainArgs, useCpPassingJar)
    os.spawn(commandArgs, env, cwd, stdin, stdout, stderr,  mergeErrIntoOut,  propagateEnv)
  }

  def call(mainClass: String,
                     classPath: Iterable[os.Path],
                     jvmArgs: Seq[String],
                     mainArgs: Seq[String],
                     env: Map[String, String] = null,
                     cwd: Path = null,
                     stdin: ProcessInput = Pipe,
                     stdout: ProcessOutput = Pipe,
                     stderr: ProcessOutput = os.Inherit,
                     mergeErrIntoOut: Boolean = false,
                     timeout: Long = -1,
                     check: Boolean = true,
                     propagateEnv: Boolean = true,
                     timeoutGracePeriod: Long = 100,
                     useCpPassingJar: Boolean = false): os.CommandResult = {

    val commandArgs = jvmCommandArgs(javaExe, mainClass, jvmArgs, classPath, mainArgs, useCpPassingJar)

    os.call(
      commandArgs,
      env,
      cwd,
      stdin,
      stdout,
      stderr,
      mergeErrIntoOut,
      timeout,
      check,
      propagateEnv,
      timeoutGracePeriod
    )
  }

[...]

PR #4085, opened 2024-12-08
Generalize `.runBackground` to work with all kinds of subprocesses

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/657

We re-use `MillBackgroundWrapper` and extend it to optionally spawn subprocess instead of calling the in-process main method. There's a bit of overhead in the JVM wrapper, but we need some kind of process wrapper to manage the mutex and termination even when the Mill process terminates, and eventually when https://github.com/com-lihaoyi/mill/issues/4007 lands we can use that to provide lighter-weight wrappers. We add a shutdown hook and copy the termination grace-period logic from OS-Lib to try and gently kill the wrapped process when necessary

Integrated this into `PythonModule#runBackground` and added a unit test to verify the asynchronous launch, background existence (by checking that a file lock is taken) and termination on deletion. We can integrate this into `TypescriptModule` as well but punting that for later

The subprocess APIs are a mess but leaving that to fix in 0.13.0 https://github.com/com-lihaoyi/mill/issues/3772

### DIFF
--- a/main/util/src/mill/util/Jvm.scala
+++ b/main/util/src/mill/util/Jvm.scala
@@ -394,8 +394,8 @@ object Jvm extends CoursierSupport {
       cwd = workingDir,
       env = envArgs,
       stdin = if (backgroundOutputs.isEmpty) os.Inherit else "",
-      stdout = backgroundOutputs.map(_._1).getOrElse(os.Inherit),
-      stderr = backgroundOutputs.map(_._2).getOrElse(os.Inherit)
+      stdout = os.Inherit,
+      stderr = os.Inherit
     )
   }
 
--- a/main/util/src/mill/util/Jvm.scala
+++ b/main/util/src/mill/util/Jvm.scala
@@ -394,8 +394,8 @@ object Jvm extends CoursierSupport {
       cwd = workingDir,
       env = envArgs,
       stdin = if (backgroundOutputs.isEmpty) os.Inherit else "",
-      stdout = os.Inherit,
-      stderr = os.Inherit
+      stdout = backgroundOutputs.map(_._1).getOrElse(os.Inherit),
+      stderr = backgroundOutputs.map(_._2).getOrElse(os.Inherit)
     )
   }
 
--- a/pythonlib/package.mill
+++ b/pythonlib/package.mill
@@ -9,4 +9,5 @@ object `package` extends RootModule with build.MillPublishScalaModule {
   // we depend on scalalib for re-using some common infrastructure (e.g. License
   // management of projects), NOT for reusing build logic
   def moduleDeps = Seq(build.main, build.scalalib)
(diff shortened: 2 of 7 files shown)


lihaoyi:
@jodersky do you happen to have a linux environment handy? Somehow `mill.pythonlib.RunBackgroundTests.runBackground` is failing in CI even though it passes on my laptop

jodersky:
I do, in fact it's my main development environment (Debian). I can take a
look at it.

lihaoyi:
thanks!

lihaoyi:
AFAIK the functionality seems to work from manual testing on an EC2 machine, just the test suite which uses a python file-lock to try and check whether the process is alive or not seems to be misbehaving

jodersky:
It's also failing on my machine

```
$ ./mill pythonlib.test mill.pythonlib.RunBackgroundTests.runBackground
X mill.pythonlib.RunBackgroundTests.runBackground 24862ms 
[1335]   java.lang.Exception: File never locked by python subprocess
[1335]     mill.pythonlib.RunBackgroundTests$.sleepIfTimeAvailable$1(RunBackgroundTests.scala:28)
[1335]     mill.pythonlib.RunBackgroundTests$.$anonfun$tests$2(RunBackgroundTests.scala:33)
```

jodersky:
I'm looking into it

jodersky:
I think linux supports multiple file locking mechanisms, so I wonder if this is due to the JVM's `channel.tryLock` not using a `flock()`

jodersky:
I checked the source code, and indeed OpenJDK uses `lockf` rather than `flock`. According to [this stackoverflow answer](https://stackoverflow.com/a/22411531/917519), on FreeBSD and Mac OSX, `lockf` and `flock` interact, but on Linux that is not the case.

Can you try replacing the line containing `flock` with `lockf` in the python script? That change works on my machine, but I don't have access to a mac, so I haven't tested if it also works there.

Alternatively, this should work on all systems
```
        if sys.platform == "linux":
            fcntl.lockf(file.fileno(), fcntl.LOCK_EX | fcntl.LOCK_NB)
        else:
            fcntl.flock(file.fileno(), fcntl.LOCK_EX | fcntl.LOCK_NB)
```

lihaoyi:
@jodersky nice find, let me try that

jodersky:
Unrelated, but have you had any success amending peoples PRs? I can see the check mark "Maintainers are allowed to edit this pull request. ", but when I tried pushing to your branch it was rejected with a "permission denied" error.

lihaoyi:
@jodersky maybe you need repo-specific committer access to amend PRs? So far you only have org-wide committer access, lemme try adding you to the Mill repo

I meant that the main code worked on my EC2 box, the test code failed same as on GHA. I pushed your suggested fix, let's see if it works

jodersky (approved):

Issue #4024, opened 2024-11-25
Support for Incremental Testing


lihaoyi:
Mill should support doing test/task selection based on changed files. This is an extremely common scenario when running CI workflows, and would even benefit Mill's own pull-request validation workflows.

The basic idea is that given two commits, we should be able to determine which test tasks to run based on the files that changed. This can be done via the following:

1. Compute the `methodCodeHashSignatures` on old and new commits to determine which task implementations changed

2. Run `git diff` between the two commits to see which source files changed

3. Map the changed source files to their respective changed source tasks

4. Traverse the task graph downstream from the changed tasks in (1) and changed source tasks in (3) and find the tests tasks you want to run

Issue #3534, opened 2024-09-13
Watch mode on multiple commands should only run commands downstream of changed files (1000USD Bounty)


nafg:
---

**From the maintainer Li Haoyi**: I'm putting a 1000USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

The goal of this is to make `-w`/`--watch` only run tasks/commands that are downstream of changed `Task.Source` files or `Task.Input`, so that if you watch multiple commands (e.g. `webserver.runBackground` and `webclient.deploy`), only the tasks relevant to a change get re-run.

---

### Background
I think my use case is a pretty common need. I'm developing in full-stack Scala: some JVM modules, some ScalaJS modules, and some shared dependencies of both. I want a fast feedback loop. When code in a JS or shared module changes, it needs to rebundle it, and when code in a JVM or shared module changes, it needs to recompile it and restart the server.

I don't need hot reload of the javascript code (that would be even better but let's get to first base first), but I do need to reload the page after rebundling JS or rebooting the server completes, but I'm accomplishing that outside of the build tool so let's ignore that aspect of things.

In my case, the target to bundle the JS and put it in the right place is `app_js.getScalaJs`, and restarting the backend is done with `app_lrbcol.runBackground`.

I've been doing this with two separate Mill processes. [tmuxp](https://tmuxp.git-pull.com/index.html) is useful for this; my config contained

```yaml
  panes:
  - mill -j 16 -w app_lrbcol.runBackground
  - docker compose up db
  - sleep 5; mill -j 16 -w app_js.getScalaJs
```

However, at the moment Mill isn't designed for multiple processes like this (see e.g., #3454). #3519 may fix this, but I recently was reminded that Mill supports multiple parallel targets.

### The issue

So I tried this command instead:

```shell
mill -j 0 -w app_lrbcol.runBackground + app_js.getScalaJs
```

However this doesn't do what I want. If I make a change to ScalaJS code, even if it doesn't compile, it still causes `runBackground` to run.

It would be better IMO if somehow watch mode could apply to each target independently, instead of what it seems to be doing, namely aggregating a single watch list of files and any of them cause the compound target to run.

nafg:
What is `watchLoop`? It ends with `???` and seems to only be called from `MillMain.main0`.

arturaz:
I'll grab this, as this is a feature I directly need.

lihaoyi:
[...]

PR #4091, opened 2024-12-09
Selective Execution based on input file and code changes

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/4024 and fixes https://github.com/com-lihaoyi/mill/issues/3534 

This PR adds support for selective execution based on build input changes and build code changes. This allows us to consider the changes to build inputs and build code and do a down-stream graph traversal to select the potentially affected tasks, without needing to go through the entire evaluation/cache-loading/etc. process to individually skip cached tasks. 

This is a major feature needed to support large monorepos, where you want to only run the portion of your build/tests relevant to your code change, but you want that selection of that portion to be done automatically by your build tool

## Motivation

Selective execution differs from the normal "cache things and skip cached tasks" workflow in a few ways:

1. We never need to actually run tasks in order to skip them
    * e.g. a  CI worker can run `selective.prepare` on `main`, `selective.run` on the PR branch, and skip the un-affected tasks without ever running them in the first place. 
    * Or if the tasks were run earlier on `main` on a different machine, we do not need to do the plumbing necessary to move the caches onto the right machine to be used

2. We can skip tasks entirely without going through the cache load-and-validate process.
    * In my experience with other build tools (e.g. Bazel), this can result in significantly better performance for skipped tasks

4. We can skip the running of `Task.Command`s as well, which normally never get cached

## Workflows

There are two workflows that this PR supports.

### Manual Task Selection

1. `./mill selective.prepare <selector>` saves an `out/mill-selective-execution.json` file storing the current `taskCodeSignatures` and `inputHashes` for all tasks and inputs upstream of `<selector>`

2. `./mill selective.run <selector>` loads `out/mill-selective-execution.json`, compares the previouos `taskCodeSignatures` and `inputHashes` with their current values, and then only executes (a) tasks upstream of `<selector>` and (b) downstream of any tasks or inputs that changed since `selective.prepare` was run

This workflow is ideal for things like CI machines, where step (1) is run on the target branch, (2) is run on the PR branch, and only the tasks/tests/commands that can potentially be affected by the PR code change will be run

The naming of `selective.prepare` and `selective.run` is a bit awkward and can probably be improved.

### Watch And Re Run Selection

1. When you use `./mill -w <selector>`, we only re-run tasks in `<selector>` if they are downstream of inputs whose values changed or tasks whose code signature changed

This is a common use case when watching a large number of commands such as `runBackground`, where you only want to re-run the commands related to the build inputs or build code you changed.

-----

It's possible that there are other scenarios where selective execution would be useful, but these were the ones I came up with for now

## Implementation

Implementation wise, we re-use of a lot of existing logic from `EvaluatorCore`/`GroupEvaluator`/`MainModule`, extracted into `CodeSigUtils.scala` and `SelectiveExecution.scala`. 

The _"store `inputHashes`/`methodCodeHashSignatures` (grouped together as `SelectiveExecution.Metadata`) from before, compare to value after, traverse graph to find affected tasks"_ logic is relatively straightforward, though the actual plumbing of the `SelectiveExecution.Metadata` data to the code where it is used can be tricky. We need to store it on disk in order to support `mill --no-server`, so we serialize it to `out/mill-selective-execution.json`

The core logic is shared between the two workflows above, as is most of the plumbing, although they hook into very different parts of the Mill codebase

Covered by some basic integration tests for the various code paths above, and one example test using 3 `JavaModule`s

## Limitations

Currently doesn't work with `-w show`, due to how `show` has two nested evaluations that are hard to keep track of which one should be selective and which one shouldn't. This is part of the general problem discussed in https://github.com/com-lihaoyi/mill/issues/502 and I think we can punt on a solution for now

Cannot yet be used on the com-lihaoyi/mill repo, due to the build graph unnecessarily plumbing `millVersion` everywhere which invalidates everything when the git dirty sha changes. Will clean that up in a follow up


lihaoyi:
This isn't ready to merge yet but I think it's worth starting the review/discussion, @jodersky @lefou @lolgab @alexarchambault

lihaoyi:
Most tests are green now, we needed to special case `Command(exclusive = true)` to make things like `--watch show <selector>` work, in the spirit of https://github.com/com-lihaoyi/mill/issues/502. However, a bit more thought is probably necessary in order to make `--watch show` support selective execution in a reasonable manner

example/depth/large/9-selective-execution/build.mill
```diff
@@ -0,0 +1,100 @@
...
+*/
+
+// Similarly, if we make a change `qux/`, using selective execution will only run tests
+// in `qux.test`, and skip those in `foo.test` and `bar.test`.
+// These examples all use `__.test` to selectively run tasks named `.test`, but you can
+// use selective execution on any subset of tasks by specifying them in the selector.
+//
+// Selective execution is very useful for larger codebases, where you are usually changing
+// only small parts of it, and thus only want to run the tests related to your changes.
+// This keeps CI times fast and prevents unrelated breakages from affecting your CI runs.
```
jodersky, line 100: Not sure if this is the right place, but it might be useful to mention a bluepront on *how* this could be implemented in CI, e.g. with github actions.

If I understand correctly, you'd want to save `mill-selective-execution.json` in the main branch and fetch it for every pull request CI run. Is that correct?

lihaoyi: The docs are meant to explain how to make use of it. Not sure if I did a good job tho

jodersky: I think they're pretty clear! I guess what my comment boils down to, Is that there's no mention of `mill-selective-execution.json` in the docs, which, if I understand correctly, is the file that needs to be created on CI-main runs and used in CI-PR runs.

But it's a detail and we can ignore it, maybe adding examples later on.

nafg:
Shouldn't the docs be on https://mill-build.org/mill/cli/builtin-commands.html?

PR #4103, opened 2024-12-11
make bsp build on TestModule success instead of fail

chikei:
IJ with BSP will make bsp call `BuildTargetCompile` with all `BuildTarget`, additional test module (introduced in #3064) will cause the bsp call fail as it falls into third case in the pattern match. This PR makes it does nothing and always success.


bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
```diff
@@ -385,6 +385,7 @@ private class MillBuildServer(
         case (m: SemanticDbJavaModule, ev) if clientWantsSemanticDb =>
           (m.compiledClassesAndSemanticDbFiles, ev)
         case (m: JavaModule, ev) => (m.compile, ev)
+        case (m: TestModule, ev) => (Task.Anon{()}, ev)
```
lefou, line 388: Alternatively, we could trigger `testClasspath`, which should trigger the compilation, if any. It might trigger a bit more that just compilation. I also can't imagine, why we not already caught the compilation by the matches above, so dong nothing might be valid too, although I'd like to leave a log message in such case, so whoever has issues and inspects the logs, she can see that we had to make a decision along the way.

lefou: Technically, calling compile on such a module is wrong. We either report the wrong module capabilities (`canCompile`) or IJ is doing it wrong and we should open an issue there.

lefou: @chikei Could you provide a reproducer for you initial issue?

lefou: So, I just read the older issue and PRs to that topic and I think this is an issue in IJ. Can you please open an issue there? We should also link that issue as a code comment here, since otherwise a future maintainer will reasonably decide to remove this case.

chikei:
reproducer mill project: https://github.com/chikei/mill-bsp-test-only-module
jetbrains issue: https://youtrack.jetbrains.com/issue/SCL-23352

bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
lefou: I guess a way to handle it instead of failing the whole request would be to just ignore the specific build target identifier which don't represent compilable modules and omit it in the result list. This is a defensive consequence of BSP working on bulk requests, an we already handle partial error the same way in other BSP requests.

lefou: > I guess a way to handle it instead of failing the whole request would be to just ignore the specific build target identifier which don't represent compilable modules and omit it in the result list. This is a defensive consequence of BSP working on bulk requests, and we already handle partial error the same way in other BSP requests.

Forget that. That won't work here, since the result value does not have a list with an item per build target identifier.

bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
```diff
@@ -385,6 +385,7 @@ private class MillBuildServer(
         case (m: SemanticDbJavaModule, ev) if clientWantsSemanticDb =>
           (m.compiledClassesAndSemanticDbFiles, ev)
         case (m: JavaModule, ev) => (m.compile, ev)
+        case (m: TestModule, ev) => (Task.Anon { () }, ev)
```
lefou, line 388:
```suggestion
        // See also https://youtrack.jetbrains.com/issue/SCL-23352
        case (m: TestModule, ev) => 
          val task = Task.Anon { 
            T.log.debug("Ignoring invalid compile request for test module ${m.bspBuildTarget.displayName}")
            }
          (task, ev)
```

lefou (approved):
Looks good to me.

lefou:
@chikei Sorry for the long delay. This will be part of Mill `0.12.6`.

PR #4110, opened 2024-12-11
Fix excessive CPU and I/O load from terminfo polling thread

BardurArantsson:


BardurArantsson:
As far as I can tell the test failures are spurious.

lihaoyi:
Looks greate, thanks!

PR #4117, opened 2024-12-12
Simplify CI config

lihaoyi:
Some cleanup in preparation for https://github.com/com-lihaoyi/mill/pull/4118

Set default `--jobs=0.5C` in `.mill-opts` so we don't need to specify it explicitly in CI, and it also works on laptops and other environments with different core counts correctly. TBH could probably go with even less parallelism due to how heavy our integration tests are - with all the compilation and JVM subprocess spawning they can probably easily take up all 4 cores on a Github Actions worker - but I'm leaving it at 0.5C X 4 cores = 2x parallelism in CI for now

Also simplified the windows CI setup: now that we have `./mill.bat` in the project root, we can just call `./mill` in the windows terminal. I had to delete the `.ps1` version because it was taking priority over `.bat` and failing with the following error:

```
Run ./mill -i -k __.compile
  
mill.ps1: D:\a\_temp\93d23[8](https://github.com/com-lihaoyi/mill/actions/runs/12292451871/job/34303135448#step:11:8)[8](https://github.com/com-lihaoyi/mill/actions/runs/12292451871/job/34303135448#step:11:9)4-066d-433e-bb26-8539b01a4aee.ps1:2
Line |
   2 |  ./mill -i -k __.compile
     |         ~~
     | Parameter cannot be processed because the parameter name 'i' is ambiguous. Possible matches include:
     | -InformationAction -InformationVariable.
Error: Process completed with exit code 1.
```


mill.ps1
lefou, line 1: Can we instead move it into a `wrapper` dir for now? I already archived `millw`, so it would be unfortunate for users to loose it. 

There is the `--no-server` option, that works the same way without collision.

PR #4130, opened 2024-12-13
Add `run-all-tests` label to disable selection execution on PRs

lihaoyi:


lihaoyi:
Still some issues in the selective execution logic, but  going to merge support for this label first and will deal with those issues in follow ups

PR #4145, opened 2024-12-16
Overhaul transitive module handling in dependency resolution

alexarchambault:
This PR changes the way transitive dependencies, but also transitive BOM dependencies and dependency management, are passed to coursier.

Before this commit, Mill was walking its own module graph, and dependencies / BOM dependencies / dep management were manually aggregated beforehand. During dependency resolution, coursier also does such aggregations for external dependencies.

Having both Mill and coursier handle that concern could lead to discrepancies or duplicate work (in https://github.com/com-lihaoyi/mill/pull/4068 in particular). So this commit passes non-processed dependencies / BOM dependencies / dep management as-is to coursier, and lets coursier handle / compose those.

With this PR, dependencies / BOM dependencies / dep management are passed around solely via `JavaModule#coursierProject` / `JavaModule#transitiveCoursierProjects`. This PR deprecates `JavaModule#{transitiveIvyDeps,transitiveCompileIvyDeps,transitiveRunIvyDeps,processDependency,processedIvyDeps}` (some only recently introduced in https://github.com/com-lihaoyi/mill/pull/3924).


alexarchambault:
This voluntarily removes `JavaModule#{transitiveIvyDeps,transitiveCompileIvyDeps,transitiveRunIvyDeps}`, in order to see what break

alexarchambault:
I'm going to try to rebase https://github.com/com-lihaoyi/mill/pull/4068 on top of that, to have a better sense of whether: the changes here are a nice-to-have, or are required or workarounds would be too hard to implement.

alexarchambault:
Some tests will fail until courier [`2.1.22`](https://github.com/coursier/coursier/releases/tag/v2.1.22) can be used (the changes here need https://github.com/coursier/coursier/pull/3189)

alexarchambault:
Going to describe more in length what these changes are about - how the Mill module graph and dependency details are now converted to coursier data structures. Maybe in a comment in the code.

lihaoyi:
I think this looks good to me. There are some classpath re-orderings, but we don't generally promise stable classpath orderings anyway.

The `transitiveIvyDeps` change is potentially breaking if people override `transitiveIvyDeps` and expect it to take effect, but I think that's probably a rare enough case that we can overlook it

lihaoyi (approved):

lefou (commented):
Can you explain the role of the internal repositories? It looks like some essential implementation detail, but is exposed as task. Maybe, it's visibility can be reduced or it can be moved into some worker. 

You introduced a new `def coursierDependency` which seems to represent the module itself. Is this in the spirit of `PublishModule.publishSelfDependency`? It seems to ignore any user settings regarding the artifact name and organization. Is it only used internally? What impact has it, when it's different from what `publishSelfDependency` returns?

alexarchambault:
> You introduced a new `def coursierDependency` which seems to represent the module itself. Is this in the spirit of `PublishModule.publishSelfDependency`? It seems to ignore any user settings regarding the artifact name and organization. Is it only used internally? What impact has it, when it's different from what `publishSelfDependency` returns?

It's only used internally, yes. I thought about making it use the same coordinates as for publishing, but I fear it might introduce ambiguity, if users mistakenly give the same coordinates to two modules. Also, all `JavaModule`s don't extend `PublishModule`, so we need to have an internal naming scheme for those.

There shouldn't be any impact… These internal coordinates leak in the dependency trees, but some post-processing (added in this PR) hides those (it actually only keeps the internal name, built out of the module segments, which should look familiar enough to users).

example/fundamentals/out-dir/1-out-files/build.mill
```diff
@@ -243,8 +243,7 @@ out/mill-server
 
 > cat out/mill-build/methodCodeHashSignatures.dest/current/spanningInvalidationTree.json
 {
-  "call scala.runtime.BoxesRunTime.boxToInteger(int)java.lang.Integer": {},
-  "call scala.Predef$#println(java.lang.Object)void": {
+  "call scala.runtime.BoxesRunTime.boxToInteger(int)java.lang.Integer": {
```
alexarchambault, line 246: Just wanted to mention that I'm not familiar enough with the hash signature stuff to really understand why this change is needed. Oddly enough, that needs to be reverted in https://github.com/com-lihaoyi/mill/pull/4155.

lefou (commented):
@alexarchambault Thanks for the insights. I think, we should mark all these internal methods with `@internal` and reduce their visibility as much as possible so that we can move/remove them later. Ideally, we can group and hide them in some inner `coursier` object in the future.

Besides, this PR looks good to me.

alexarchambault (commented):
Added a few comments about tricky points or odd changes, just in case

lihaoyi:
@alexarchambault feel free to merge this whenever you're ready

build.mill
```diff
@@ -591,6 +591,10 @@ trait MillStableScalaModule extends MillPublishScalaModule with Mima {
       "mill.main.MainModule.mill$define$BaseModule0$_setter_$evalWatchedValues_="
     ),
 
+    // Overrides added for new methods, ought to be safe
+    ProblemFilter.exclude[ReversedMissingMethodProblem]("mill.scalalib.JavaModule.mill$scalalib$JavaModule$$super$internalRepositories"),
+    ProblemFilter.exclude[ReversedMissingMethodProblem]("mill.scalanativelib.ScalaNativeModule.mill$scalanativelib$ScalaNativeModule$$super$coursierProject"),
+
```
lihaoyi, line 597: @alexarchambault these turn out not to be safe, we probably need to fix the code to not add calls to `super` since they are not binary compatible


CC @sake92

lefou: I'm not sure I said it before, but I suggest to generate the access to `super.<overridden>` from our macros as soon as we can break bin-compat, so that we better prepare the binary code for ad-hoc overriding of tasks. It can be left unused, so it should not affect functionality, but we at least we trigger the super-accessors to be present in the binary code.

alexarchambault: Trying to address that in https://github.com/com-lihaoyi/mill/pull/4380, although the workarounds aren't fully satisfying IMO (they assume users don't add overrides between the trait defining the method and the one overriding it)

alexarchambault: Other such changes adding calls to `super` have been introduced in the past, apparently:
https://github.com/com-lihaoyi/mill/blob/9815c3548a0ec05d90be659480757f185e2f2960/build.mill#L523-L580

PR #4162, opened 2024-12-18
Deprecate resolveFilter stuff

alexarchambault:
This PR makes Mill stop relying on the `resolveFilter` parameter of `CoursierSupport#resolveDependencies`.

This parameter manually filters the artifacts provided by coursier, but not the ones originating from "local dependencies" (Mill's hack to use freshly built workers and plugins during tests), where it seems such a filtering was done in the build (by customizing `testDepPaths`). "Local dependencies" cannot be kept as-is in https://github.com/com-lihaoyi/mill/pull/4145 (where such dependencies aren't necessarily root dependencies anymore, making it harder to handle them in a different fashion). Hence the need to stop using `resolveFilter`.

Rather than filtering artifacts after dependency resolution, this PR marks unwanted dependencies of workers as compile-time only. That way, they're not pulled by dependency resolution, and we don't need to manually filter artifacts.


alexarchambault:
The CI errors seem unrelated to these changes (I can't reproduce them locally)

lihaoyi:
Yes we've been seeing flakiness in ci, i'll rerun them

alexarchambault:
@lihaoyi @lefou Can any of you re-run the `Run Tests / linux (11, '{scalajslib,scalanativelib,kotlinlib,pythonlib,javascriptlib}.__.test', false)` job? It seems the error is unrelated to these changes and is a flake.

Also, I'm about to undraft the 3 other PRs that depend on this one (that actually depend on each other successively). Namely https://github.com/com-lihaoyi/mill/pull/4145, https://github.com/com-lihaoyi/mill/pull/4154, and https://github.com/com-lihaoyi/mill/pull/4155. Each adds something to the previous one. https://github.com/com-lihaoyi/mill/pull/4145 should be the one with the most important changes: it changes the way Mill interfaces with coursier basically, which impacts many other Mill parts.

The PR here should be the most straightforward to review I guess.

The other two are what motivated the first two: they consist in BOM-related changes.

If you'd like me to… split them further to ease review, or re-open them with a cleaner git history to spot more easily what each adds to the previous ones, just tell me.

lihaoyi:
@alexarchambault kicked the jobs, also gave you write access to the repo so you can do this yourself as necessary in future

alexarchambault:
> I've been holding off taking a look at these PRs since it looked like they were still in flux; should we look at all of them? Or just this one first?

As you wish, I'd say. One after the other should be easier, keeping in mind that the first two are not just nice-to-have refactorings / clean-ups, but are motivated by the other two, that they make easier to implement (first one is actually motivated by the second one, itself motivated by the other two).

lefou (commented):
I can't spot a single `@deprecated` anywhere. Shouldn't we provide some API that is free of deprecated stuff and have a message to an deprecated API user?

The change itself makes sense to me. But due to the informal deprecation, it'S rather hard to spot, where (else) we still rely on the filtering.

alexarchambault:
> I can't spot a single `@deprecated` anywhere. Shouldn't we provide some API that is free of deprecated stuff and have a message to an deprecated API user?

I was hesitating about this. Managing deprecations while removing a parameter is more of a burden than when adding one. Basically, in a signature like
```scala
def resolveDependencies(
    repositories: Seq[Repository],
    deps: IterableOnce[Dependency],
    force: IterableOnce[Dependency],
    sources: Boolean = false,
    customizer: Option[Resolution => Resolution] = None,
    ctx: Option[mill.api.Ctx.Log] = None,
    coursierCacheCustomizer: Option[FileCache[Task] => FileCache[Task]] = None,
    resolveFilter: os.Path => Boolean = _ => true,
    artifactTypes: Option[Set[Type]] = None,
    resolutionParams: ResolutionParams = ResolutionParams()
  ): Result[Agg[PathRef]]
```
we'd like to deprecate `resolveFilter`, while not breaking bin compat. Apparently, adding a `@deprecated` annotation to the parameter only deprecates its use in the method body, not its use when users call the method.

We can always remove the parameter, and add a shim of the former method with no default params for bin compat, but that means current users of the parameter are very likely to experience compilation errors when bumping Mill (if they rely on a default parameter), while deprecations usually only create a warning.

That said, I'm not too opiniated about doing this or what's currently in the PR (renaming of the parameter), it seems both have caveats.

lefou:
I see. You mean, the synthetic providers for the default values may change in an binary incompatible way. It's a real pita, that this is still an issue in Scala.

lefou:
But in the `@deprecated` annotation on the parameter doesn't hurt and is even recognized in the body, we should add it with actionable comment.

lihaoyi:
I've been trying to get `@unroll` support upstream into Scala 3 for about a year now, which would let us automate all these bincompat forwards. Hoping it can land soon in https://github.com/scala/scala3/pull/21693

lefou:
> I've been trying to get `@unroll` support upstream into Scala 3 for about a year now, which would let us automate all these bincompat forwards. Hoping it can land soon in [scala/scala3#21693](https://github.com/scala/scala3/pull/21693)

To my understanding, `@unroll` can't help if you want to *remove* a (not-last) parameter with defaults from the parameter list. Or was something added lately?

lihaoyi:
ah yes, it only helps for adding parameters

alexarchambault:
Just added `@deprecated` annotations on the `deprecatedResolveFilter` parameters, even though this should create warnings in users' code.

lefou (commented):
Maybe I overlook the obvious, but since we're now no longer declare any runtime dependencies in our workers, how so we ensure we properly load the workers transitive dependencies at runtime?

contrib/package.mill
```diff
@@ -126,18 +126,20 @@ object `package` extends RootModule {
...
-        Agg(
+        super.mandatoryIvyDeps() ++ Agg(
           // compile-time only, need to provide the correct scoverage version at runtime
           build.Deps.scalacScoverage2Plugin,
           build.Deps.scalacScoverage2Reporter,
           build.Deps.scalacScoverage2Domain,
           build.Deps.scalacScoverage2Serializer
         )
       }
+      def mandatoryIvyDeps = Agg.empty[Dep]
```
lihaoyi, line 142: What's going on with these changes? Did the semantics of the underlying resolution change in some way that we need to change this downstream code?

alexarchambault:
> Maybe I overlook the obvious, but since we're now no longer declare any runtime dependencies in our workers, how so we ensure we properly load the workers transitive dependencies at runtime?

We can still add them as standard dependencies. Their dependencies will be pulled alongside the worker. (We may want to add to them excludes for things that are already in Mill, like `scala-library` or the main Mill modules, as these are not needed to load the worker.)

The PR here actually enables that, while the prior `resolveFilter` and `testDepPaths` stuff would have needed to be changed to let the JARs of the new dependencies (and their transitive dependencies) through.

contrib/package.mill
alexarchambault: The goal here is to also make the `mandatoryIvyDeps` (`scala-library` basically) compile-time only too. But that's just it.

lefou:
> > Maybe I overlook the obvious, but since we're now no longer declare any runtime dependencies in our workers, how so we ensure we properly load the workers transitive dependencies at runtime?
> 
> We can still add them as standard dependencies. Their dependencies will be pulled alongside the worker. (We may want to add to them excludes for things that are already in Mill, like `scala-library` or the main Mill modules, as these are not needed to load the worker.)
> 
> The PR here actually enables that, while the prior `resolveFilter` and `testDepPaths` stuff would have needed to be changed to let the JARs of the new dependencies (and their transitive dependencies) through.

Ok. So, I didn't seen it because all workers you touched don't have any extra transitive dependencies not already in the Mill runtime.

alexarchambault:
> Ok. So, I didn't seen it because all workers you touched don't have any extra transitive dependencies not already in the Mill runtime.

Yeah, exactly. That surprised me.

alexarchambault:
Should I merge now that I have the rights?

lihaoyi:
@alexarchambault go for it

lihaoyi (approved):

PR #4163, opened 2024-12-18
Reduce Java Heap for Mill Client

gamlerhart:
The Mill Client only forwards requests and therefore does not need much memory.
However, Java doesn't know that an allocates a large heap based on the machines memory size.

Therefore, give a way lower heap limit,
creating a leaner client memory wise.

Example on my machine:
- Before: A Mill client reserved ~500MByte
- After: A Mill client reserves the specified 24Mbyte

This is mostly relevant for a long running mill client (examples  `--watch` clients, `-i --no-build-lock` and --bsp`).

### DIFF
--- a/dist/package.mill
+++ b/dist/package.mill
@@ -129,7 +129,11 @@ object `package` extends RootModule with InstallModule {
       .mkString("\r\n")
 
     os.write(vmOptionsFile, millOptionsContent)
-    val jvmArgs = otherArgs ++ List(s"-DMILL_OPTIONS_PATH=$vmOptionsFile")
+    val jvmArgs = otherArgs ++ List(
+      // Avoid reserving a lot of memory for the client, as the client only forward information
+      "-Xmx32m",
+      s"-DMILL_OPTIONS_PATH=$vmOptionsFile"
+    )
     val classpath = runClasspath().map(_.path.toString)
     val classpathJar = Task.dest / "classpath.jar"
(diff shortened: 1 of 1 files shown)


gamlerhart:
*facepalm*. Its also added to the test-runner, creating out of memory exceptions. Will take a look later.

gamlerhart:
Question: Are some tests flacky? And if so, how can I rerun that specific test suite/github action?

lihaoyi:
yes some tests are flaky. I'll give them a kick

lihaoyi:
@gamlerhart could you test this out with Mill's new ability to download JVMs on demand using Coursier (https://mill-build.org/mill/main-branch/cli/installation-ide.html#_customizing_mills_jvm)? I expect that will take more memory than the base Mill client, and we should adjust the number accordingly

gamlerhart:
The Memory is enough to download a JDK. It downloaded a Zulu JDK without issues for me.
I still increased it slightly, to have a bit more buffer.

However, the question is if this PR is still required. Now that we have native Graal clients, I assume most setups will use those. So, this optimization is less relevant.

lihaoyi:
I think it's still worth doing. We will need to continue support the JVM launcher anyway, either as a cross-platform launcher, or for platforms like Windows-ARM (e.g. my surface 7 laptop) that do not have Graal native image support yet

lihaoyi:
Thanks @gamlerhart !

Issue #3993, opened 2024-11-19
Coursier resolution error for Android aar dependencies


0xnm:
Repro steps:

* Create a `AndroidAppKotlinModule` (doesn't really need to be this one, just make sure that `artifactTypes` has also `coursier.Type("aar")`, but probably it doesn't matter at the POM resolution stage) with the following content:

```
def repositoriesTask: Task[Seq[Repository]] = Task.Anon {
  super.repositoriesTask() :+ MavenRepository("https://maven.google.com/")
}

def ivyDeps: T[Agg[Dep]] = Agg(
  ivy"androidx.core:core-ktx:1.15.0",
  ivy"androidx.activity:activity-compose:1.9.3",
  ivy"androidx.compose.ui:ui:1.7.5",
  ivy"androidx.compose.material3:material3:1.3.1"
)
```
* Run `ivyDepsTree` for this module.

The outcome will be an error:

```
app.ivyDepsTree coursier.error.ResolutionError$ConflictingDependencies: Conflicting dependencies:
androidx.collection:collection:1.0.0 or 1.1.0 or 1.4.0 or 1.4.2 or [1.2.0] wanted by
... <see attached file for the resolved tree>
```

At the same time if same dependencies are added in a Gradle build script via [implementation](https://developer.android.com/build/dependencies#dependency_configurations) dependency configuration, everything is resolved correctly (see attached file for the complete resolution tree).

I'm not sure where the issue is, probably rather on the Coursier side?

Gradle is using the following conflict resolution strategy by default ([link](https://docs.gradle.org/current/userguide/dependency_resolution.html#2_perform_conflict_resolution)):

> Gradle’s conflict resolution strategy, which defaults to selecting the highest version, selects a single version of a module when multiple are requested.

[mill-deps-tree.txt](https://github.com/user-attachments/files/17821750/mill-deps-tree.txt)
[gradle-compile-and-runtime-deps-tree.txt](https://github.com/user-attachments/files/17821751/gradle-compile-and-runtime-deps-tree.txt)

lihaoyi:
CC @alexarchambault

himanshumahajan138:
@0xnm i have the solution i guess coz when i was working on compose pr  i used particular version which are not prone to errors 

see this code and versions that i used they dont have any  error

```
package build

import mill._
import kotlinlib._
import coursier.maven.MavenRepository
import mill.kotlinlib.android.AndroidAppKotlinModule
[...]

PR #4169, opened 2024-12-21
Working towards android support

vaslabs:
### Relevant issues: 

- https://github.com/com-lihaoyi/mill/issues/3868
- https://github.com/com-lihaoyi/mill/issues/3550

### Outline

- Follow android studio + gradle standard for directory structure
- Unit test support (local runs - no android device needed)
- Integration test support (android tests, needs an android device/emulator)

### New features

- Ability to install an apk in the default available android device (tested against AVD)
- Ability to run all integration tests with the install command (android manifest with instrumentation is needed, this needs to be a functionality of integration tests support in the future instead)
- Classpath resolution (runtimeDependencies) for packaging with dependencies and user's source code (tested against integration test and running the android app itself via the emulator)
- Special classpath resolution and packaging for android tests (integration tests)
- Mill integrates with avdmanager, adb and emulator to automate the run of integration tests by creating virtual devices and controlling the emulator (start, stop). Examples are added in documentation of both kotlin and java app. This includes the ability to use different emulators for different apps, by specifying identifiers and ports, though this use case should be rare in app development lifecycles. 
- Github actions & mill support to run integration tests in (headless) emulator & differentiate from local mode (windowed emulator + default gpu settings)

### Github actions

A sample repo to demonstrate what's needed to run the emulator https://github.com/vaslabs-ltd/test-android-emulator

In the end what slowed me down was that for github actions, the AVD home env var was needed, which wasn't obvious until I've tested it in isolation . All the flags and kvm configs are also needed and it's what https://github.com/ReactiveCircus/android-emulator-runner does

### Stubs and missing features

An effort was made to standardise the structure faster, so follow up PRs will not use a non-standard structure. Things that are working but not "production ready" are:

1. Integration tests debug apk: This inherits the current apk construction, debug functionality needs to be added and separated from the release androidApk task
2. No testOnly for integration tests: this is easy to do, but was not added in the interest of not increasing the size of the PR and the testing surface
3. Classpath resolution: For kotlin, jvm specific dependencies are removed in a hardcoded way.
4. Desugaring: This shows as a warning in the android tests apk creation, while the flow works, the d8 will need to be configured to desugar annotations properly.
5. Emulator CI mode only recognises github actions

#### Running android tests

https://github.com/user-attachments/assets/0cb47ac7-32ff-4d5c-a678-11cd18807b10


himanshumahajan138:
@vaslabs Buddy can you please mention the issue you are working on ?

example/android/javalib/1-hello-world/app/src/test/java/com/helloworld/app/ExampleUnitTest.java
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Example local unit test, which will execute on the development machine (host).
+ *
+ * @see <a href="http://d.android.com/tools/testing">Testing documentation</a>
+ */
+public class ExampleUnitTest {
+  @Test
+  public void addition_isCorrect() {
+    assertEquals(4, 2 + 2);
```
lihaoyi, line 15: Can we make this unit test depend on the code in the `src/main/` folder? So we can ensure that the classpaths are properly wired up

example/android/javalib/3-linting/build.mill
```diff
@@ -41,12 +38,12 @@ object app extends AndroidAppModule {
...
 }
 
-/** See Also: app/AndroidManifest.xml */
+/** See Also: app/src/main/AndroidManifest.xml */
 
 // The `AndroidManifest.xml` file shown above is flawed with hardcoded string linting problem so it can be spoted by running the `androidLintRun` task.
 
-/** See Also: app/resources/values/colors.xml */
-/** See Also: app/resources/values/strings.xml */
+/** See Also: app/src/main/res/values/colors.xml */
```
lihaoyi, line 45: Just to confirm here that android uses `src/main/res/` as the primary resource folder? I thought most Java projects use `resources/`, but maybe Android has different conventions

example/android/javalib/1-hello-world/app/src/test/java/com/helloworld/app/ExampleUnitTest.java
lihaoyi: Same for the kotlin example below

example/android/kotlinlib/1-hello-kotlin/build.mill
```diff
@@ -27,10 +29,45 @@ object app extends AndroidAppKotlinModule {
...
+    def ivyDeps = super.ivyDeps() ++ Agg(
+      ivy"junit:junit:4.13.2"
+    )
+  }
+
+  object it extends AndroidAppKotlinIntegrationTests with AndroidTestModule.AndroidJUnit {
+    def repositoriesTask = Task.Anon {
+      super.repositoriesTask() ++
+        Seq(MavenRepository("https://maven.google.com"))
+    }
```
lihaoyi, line 51: Should this be set in `AndroidAppModule` by default? I assume most android projects would need it

example/android/kotlinlib/1-hello-kotlin/build.mill
```diff
@@ -27,10 +29,45 @@ object app extends AndroidAppKotlinModule {
...
+    def kotlinVersion = "2.0.20"
+    def androidSdkModule = mill.define.ModuleRef(androidSdkModule0)
```
lihaoyi, line 53: Should these be inherited from the enclosing `AndroidAppKotlinModule` by default? That's what we do in the unit tests above, which is why we do not need to set these configs there

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -47,6 +48,15 @@ object AndroidLintReportFormat extends Enumeration {
 @mill.api.experimental
 trait AndroidAppModule extends JavaModule {
 
+  final def root: Path = super.millSourcePath
+  private def src: Path = root / "src"
+  override def millSourcePath: Path = src / "main"
+
+  override def sources: T[Seq[PathRef]] = Task.Sources(millSourcePath / "java")
```
lihaoyi, line 55: If we want to follow the `src/main/java/` etc. convention, rather than configuring it ourselves we should be able to mix in `with MavenModule` to set that config

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -47,6 +48,15 @@ object AndroidLintReportFormat extends Enumeration {
...
+
+  private def bannedModules(classpath: PathRef): Boolean =
+    !classpath.path.last.contains("-jvm")
```
lihaoyi, line 58: This is a bit of a hack so we should leave a comment stating that and linking to the relevant issue https://github.com/com-lihaoyi/mill/issues/3867

scalalib/src/mill/javalib/android/AndroidSdkModule.scala
```diff
@@ -118,6 +120,25 @@ trait AndroidSdkModule extends Module {
...
+    PathRef(sdkPath().path / "platform-tools" / "adb")
+  }
+
+  /**
+   * List all attached devices and emulators.
+   *
+   * For more information, refer to the official Android documentation [[https://developer.android.com/tools/adb]]
+   */
+  def listAvailableDeviceTargets: Target[CommandResult] = Task {
+    os.proc(adbPath().path / "-l").call()
```
lihaoyi, line 139: This should probably be `os.proc((adbPath().path, "-l")).call()`, since `-l` is a CLI flag and not a filesystem path segment

scalalib/src/mill/javalib/android/AndroidAppModule.scala
lihaoyi: If we want to make `AndroidAppModule` have the `MavenModule` layout by default, we should introduce another trait (maybe `AndroidAppBasicModule`???) that keeps the current layout, to preserve the symmetry with `JavaModule`/`MavenModule`, `KotlinModule`/`KotlinMavenModule`, `ScalaModule`/`SbtModule`

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -4,6 +4,7 @@ import mill._
 import mill.scalalib._
 import mill.api.PathRef
 import mill.define.ModuleRef
+import os.{CommandResult, Path}
```
lihaoyi, line 7: Generally we should avoid importing these and instead use them fully qualified as `os.CommandResult` and `os.Path`

lihaoyi:
Did a quick review pass. Haven't tried to run the example builds manually yet, but will do so when I have a bit more time. Do make sure the comment documentation in each example's `build.mill` are sufficient for someone unfamiliar with Mill to figure out how to install the necessary things and run the examples

example/android/javalib/3-linting/build.mill
vaslabs: yes this is the default structure

image

scalalib/src/mill/javalib/android/AndroidAppModule.scala
vaslabs: I have to check if it's what Android follows and makes sense, after you mentioned it I bet it is! but if it's not, maybe we are coupling the android structure with the maven one, I need more time to check.

The android structure is that it has the java directory that hosts both kotlin and java classes . Maybe we'll want to support scala too, it could be a good idea to not couple them for now, but up to you

scalalib/src/mill/javalib/android/AndroidSdkModule.scala
vaslabs: ah this was a leftover sorry, instead I deleted it and added the necessary methods that are needed to run android tests

lihaoyi:
Also probably worth a review by @0xnm, who helped review a lot of the earlier experimental android work

scalalib/src/mill/javalib/android/AndroidAppModule.scala
vaslabs: I gave it a go with MavenModule but it entangled the paths even further and I gave up for now. Maybe we can consolidate later once we have basic working scenarios and tests

vaslabs:
Need to fix the GitHub action for android tests, I'm converting it to draft as I'll be away for the Christmas holidays

scalalib/src/mill/javalib/android/AndroidAppModule.scala
lihaoyi: Sounds good, we can just go with this for now

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -1,12 +1,14 @@
...
 import mill.api.PathRef
 import mill.define.ModuleRef
-
 import upickle.default._
 
+import scala.jdk.OptionConverters.RichOptional
+
 /**
  * Enumeration for Android Lint report formats, providing predefined formats
  * with corresponding flags and file extensions. Includes utility methods
```
sideeffffect, line 14: I know this doesn't change in this PR, but why is `AndroidLintReportFormat` implemented via `Enumeration` instead of the more common `sealed abstract case class` + `case object`s in the companion object?

lihaoyi: Probably no good reason. This stuff is a pretty rough work in progress and not polished at all

vaslabs: if you want I can take a note and change it in subsequent PRs that won't be so large as this one or on its own

vaslabs:
I think it's ready for another go on a review. Some of the tests seem flaky, but let me know if I nicked something. 

Also good to test it locally.

One note:

It seems that running 2 emulators is sometimes too much for github actions (although it passed at least once in my copy). So if you want we can use the android emulator action and use a shared one. But I wanted to have the emulator functionality in mill too, for a more out of the box CI experience

scalalib/src/mill/javalib/android/AndroidSdkModule.scala
```diff
@@ -118,13 +119,47 @@ trait AndroidSdkModule extends Module {
     PathRef(buildToolsPath().path / "apksigner")
   }
 
+  /**
+   * Provides the path for the Android Debug Bridge (adt) tool.
+   *
+   * For more information, refer to the official Android documentation [[https://developer.android.com/tools/adb]]
+   */
+  def adbPath: T[PathRef] = Task {
+    PathRef(sdkPath().path / "platform-tools" / "adb")
```
lihaoyi, line 128: You can consolidate multiple path segments into one string as long as it's a literal

```scala
PathRef(sdkPath().path / "platform-tools/adb")
```

Here and below

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -47,13 +49,32 @@ object AndroidLintReportFormat extends Enumeration {
 @mill.api.experimental
 trait AndroidAppModule extends JavaModule {
 
+  final def root: os.Path = super.millSourcePath
+  private def src: os.Path = root / "src"
+  override def millSourcePath: os.Path = src / "main"
```
lihaoyi, line 54: Let's leave `millSourcePath` unchanged, get rid of `def root`, and adjust the other parts of the code as necessary (e.g. `def sources`). That would make it most similar to how the other Mill modules work

example/android/kotlinlib/1-hello-kotlin/build.mill
```diff
@@ -27,10 +29,41 @@ object app extends AndroidAppKotlinModule {
...
+  object it extends AndroidAppKotlinIntegrationTests with AndroidTestModule.AndroidJUnit {
+
+    override def instrumentationPackage = "com.helloworld.app"
+
+    /* TODO this needs to change to something better once integration tests work with debug keystore */
+    override def releaseKeyPath = mainRoot
+    def androidReleaseKeyName: T[String] = Task { "releaseKey.jks" }
+    def androidReleaseKeyAlias: T[String] = Task { "releaseKey" }
+    def androidReleaseKeyPass: T[String] = Task { "MillBuildTool" }
+    def androidReleaseKeyStorePass: T[String] = Task { "MillBuildTool" }
```
lihaoyi, line 58: Should we make `AndroidAppKotlinIntegrationTests`  and `AndroidAppIntegrationTests` forward these by default to the enclosing `AndroidAppModule`. In these examples it seems like we always set them to the same value anyway, so might as well make that the default and anyone who wants something different can configure it accordingly

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -280,6 +313,22 @@ trait AndroidAppModule extends JavaModule {
     PathRef(unsignedApk)
   }
 
+  /**
+   * Gets all the jars from the classpath and creates an apk with them
+   * to be bundled with the app's code apk
+   */
+  def runtimeDependencies: T[Seq[PathRef]] = Task {
```
lihaoyi, line 320: Let's name this `androidRuntimeDependencies`.

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -452,4 +512,230 @@ trait AndroidAppModule extends JavaModule {
     PathRef(Task.dest)
   }
 
+  /** The name of the virtual device to be created by  [[createAndroidVirtualDevice]] */
+  def virtualDeviceIdentifier: String = "test"
```
lihaoyi, line 516: Let's prefix most of the task names here with `android`. `AndroidAppModule` is expected to be mixed into `JavaModule` or `KotlinModule` or `ScalaModule`, so it would be good to prefix the tasks to avoid conflicts.

lihaoyi (commented):
Some last nits but looks really good overall, once those are settled I think we can merge it and close out the relevant bounties

example/android/kotlinlib/1-hello-kotlin/build.mill
vaslabs: bear in mind that this will change completely on the next iteration, as the android tests (in gradle + android studio) generate a debug release key (with android as username and password, perhaps we'll do a mill-android credential pair). It will be part of the work for generating a debug APK which I stubbed here and there in this PR

scalalib/src/mill/javalib/android/AndroidAppModule.scala
vaslabs: thanks, I need to try this out, going for the sources was one of my early attempts , I'm not sure it will work as other things rely on it, but now that I'm more familiar maybe I missed something else

example/android/kotlinlib/1-hello-kotlin/build.mill
lihaoyi: got it, I'm fine leaving it as is then

0xnm:
@vaslabs You may consider to rebase on https://github.com/com-lihaoyi/mill/pull/4188 since it is a significant rework of Android support, but it can go after yours as well.

vaslabs:
> @vaslabs You may consider to rebase on #4188 since it is a significant rework of Android support, but it can go after yours as well.

Thanks for mentioning, I had a brief look and a few things conflict others we complement each other (e.g. I've changed everything to use the standard android structure after a brief discussion with @lihaoyi ) and you've done the android manifest merging.

I think my PR should go first to get the standard structure and the emulator work out, and then we could either work together to merge yours or any other arrangement you'd like

example/android/javalib/1-hello-world/build.mill
```diff
@@ -52,14 +105,58 @@ object app extends AndroidAppModule {
...
+> cat out/app/it/test.json | grep '"OK (1 test)"'
+..."OK (1 test)"...
+
+> ./mill show app.stopAndroidEmulator
+
+> ./mill show app.deleteAndroidVirtualDevice
+
+*/
+
+// The android tests (existing typically in androidTest directory, aka integration tests)
```
0xnm, line 156: minor: it is better to call them instrumented tests to follow the official Android lingvo https://developer.android.com/training/testing/instrumented-tests, integration test for the Android module can be a pure JVM test as well without the need of the Android VM.

Moreover, the text in `ExampleInstrumentedTest` in your PR is using a word `instrumented`.

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -47,13 +49,32 @@ object AndroidLintReportFormat extends Enumeration {
 @mill.api.experimental
 trait AndroidAppModule extends JavaModule {
 
+  final def root: os.Path = super.millSourcePath
+  private def src: os.Path = root / "src"
+  override def millSourcePath: os.Path = src / "main"
+
+  override def sources: T[Seq[PathRef]] = Task.Sources(millSourcePath / "java")
```
0xnm, line 56: you probably need to override this to `millSourcePath / "kotlin"` in the `AndroidAppKotlinModule`

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -47,13 +49,32 @@ object AndroidLintReportFormat extends Enumeration {
...
+  override def repositoriesTask: Task[Seq[Repository]] = Task.Anon {
+    super.repositoriesTask() ++
+      Seq(MavenRepository("https://maven.google.com"))
+  }
+
+  /* TODO this is a temporary hack to exclude any jvm only tagged dependencies
+   * which may conflict with android dependencies. For example, the
+   * kotlin coroutines which are provided by both kotlin-core and
+   * kotlin-core-jvm . See also https://github.com/com-lihaoyi/mill/issues/3867
+   */
```
0xnm, line 67: What is the exact issue here? Duplicate class definition?

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -139,20 +163,24 @@ trait AndroidAppModule extends JavaModule {
     (jarFiles, resFolders)
   }
 
+  def res: T[Seq[PathRef]] = Task.Sources { millSourcePath / "res" }
```
0xnm, line 166: imo this can be confusing for the end-user to have both `res` and `resources` in the build script API layout. Having just `resources` imo is fine, it doesn't cause issues.

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -452,4 +512,230 @@ trait AndroidAppModule extends JavaModule {
...
+    os.call((
+      androidSdkModule().avdPath().path,
+      "create",
+      "avd",
+      "--name",
+      virtualDeviceIdentifier,
+      "--package",
+      sdkInstallSystemImage(),
+      "--device",
+      "Nexus 5X",
```
0xnm, line 555: this is a pretty old device definition to use by default. Any particular reason to use this one and not recent Pixel definitions which bring more actual parameters like screen size?

scalalib/src/mill/javalib/android/AndroidAppModule.scala
vaslabs: This took me by surprise as well, but this is how android studio + gradle structure the project

Screenshot From 2024-12-28 13-53-38

scalalib/src/mill/javalib/android/AndroidAppModule.scala
vaslabs: yes, I am planning to look into consolidating the assembly functionality later (experimenting with a "fat jar", extracting that and see if d8 picks everything up), so I should be addressing it in a follow-up PR

scalalib/src/mill/javalib/android/AndroidAppModule.scala
vaslabs: there are a few issues that are mixing the resources and I'll look into untangling them on a subsequent PR, for now I separated them to make it easier to build everything , if it's ok with you

scalalib/src/mill/javalib/android/AndroidAppModule.scala
0xnm: This is rather a historical reason, because originally the primary language was Java, then, once Kotlin support was added, Kotlin files had to be under `main/kotlin` and then AGP added support of the mixed Java + Kotlin sources under the `main/java`.

If you try to create a new directory, `main/kotlin` will be in the dropdown list next to the `main/java` as well.

Anyway, since return type here is `T[Seq[PathRef]]`, it is definitely worth to have `Task.Sources(millSourcePath / "java")` + `Task.Sources(millSourcePath / "kotlin")` override in the Kotlin-specific module to support better ootb experience.

scalalib/src/mill/javalib/android/AndroidAppModule.scala
0xnm: This is pretty strange, I didn't face that in my PR which also has coroutines in the dependency graph. Are you able to post resolved dependency graph you have without this override?

scalalib/src/mill/javalib/android/AndroidAppModule.scala
vaslabs: ah thanks for spotting, I was trying to use a lower end device for running 2 emulators in parallel, changed to generic medium_phone instead and made it configurable. Although these will also change to subsequent PRs as we need to give more control to the user on how the start the android emulator

e.g. mill app.createAndroidDevice <options>

example/android/javalib/1-hello-world/build.mill
vaslabs: thanks

scalalib/src/mill/javalib/android/AndroidAppModule.scala
0xnm: The maybe it is worth to make this one `private` for now if there is a chance it goes away at some point?

scalalib/src/mill/javalib/android/AndroidAppModule.scala
vaslabs: done, let me know if I missed any

example/android/kotlinlib/1-hello-kotlin/build.mill
vaslabs: I've changed it anyway before seeing your comment and added more TODO comments around the areas this will need re-work

scalalib/src/mill/javalib/android/AndroidAppModule.scala
vaslabs: done and even spotted a bug with the previous compilation, thanks for your comment!

scalalib/src/mill/javalib/android/AndroidAppModule.scala
vaslabs: the test modules need to override to add there own res directories but I suppose they can be private nevertheless. let me give it a go.

also would be good to chat together in general if you are working in this as well to get in sync, please drop me an email if you'd like :)

scalalib/src/mill/javalib/android/AndroidAppModule.scala
vaslabs: have you managed to run the integration tests? this happens as a transitive dependency from androidx test something, let me change it and give you the exact error

scalalib/src/mill/javalib/android/AndroidAppModule.scala
0xnm: It can be `protected` then to support the override. Can you please explain the issue that leads to the need of this method?

vaslabs: wait, yeah I managed to get rid of it, thanks!

scalalib/src/mill/javalib/android/AndroidAppModule.scala
0xnm: No, I didn't pull this PR to run it yet. Also FYI there is the following issue in Mill (well, in Coursier), because Gradle under the hood relies on the `.module` files rather than `.pom` for the dependency graph resolution https://github.com/com-lihaoyi/mill/issues/3993.

scalalib/src/mill/javalib/android/AndroidAppModule.scala
vaslabs: makes sense, and I guess we could mix an android scala module in the future and make it work similarly. Let me make a change in the example too and test it out (with the Kotlin dir)

0xnm: To be precise, going until `src/main` is already enough, because unlikely it will be anything else with compilation sources under it other than `java` or `kotlin` (but AGP allows to defined custom source sets though) and `KotlinModule` will walk child hierarchy anyway https://github.com/com-lihaoyi/mill/blob/d663c0302f9474c4be924353462d7cc4d725f8b0/kotlinlib/src/mill/kotlinlib/KotlinModule.scala#L26-L44, but let's for now a strict whitelist of `java` and `kotlin` subfolders only.

vaslabs: I've added the kotlin path and enriched the example, let me know if you'd like anything else, otherwise please resolve conversation

scalalib/src/mill/javalib/android/AndroidAppModule.scala
vaslabs: the error is
```
[3074]   Caused by: com.android.tools.r8.internal.g: Type kotlinx.coroutines.StandaloneCoroutine is defined multiple times: /home/vnicolaou/.cache/coursier/v1/https/repo1.maven.org/maven2/org/jetbrains/kotlinx/kotlinx-coroutines-core-jvm/1.7.1/kotlinx-coroutines-core-jvm-1.7.1.jar:kotlinx/coroutines/StandaloneCoroutine.class, /home/vnicolaou/.cache/coursier/v1/https/repo1.maven.org/maven2/org/jetbrains/kotlinx/kotlinx-coroutines-core/1.3.4/kotlinx-coroutines-core-1.3.4.jar:kotlinx/coroutines/StandaloneCoroutine.class
```

and the relevant section of ivy dependency tree

```
├─ androidx.test:core:1.6.1
├─ androidx.concurrent:concurrent-futures-ktx:1.1.0
      └─ org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.4
      └─ org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.7.1
```

scalalib/src/mill/javalib/android/AndroidAppModule.scala
0xnm: LGTM, but I don't have an option to resolve this conversation.

scalalib/src/mill/javalib/android/AndroidAppModule.scala
vaslabs: I've read the thread in #3993 and I suspect you are right about the module. I also suspect that gradle or the android kotlin plugin (?) removes any dependencies that are not androidJvm
`"org.jetbrains.kotlin.platform.type": "androidJvm"`

as here
https://repo1.maven.org/maven2/org/jetbrains/kotlinx/kotlinx-coroutines-core-jvm/1.10.1/kotlinx-coroutines-core-jvm-1.10.1.module

the platform type is jvm

0xnm: > ├─ androidx.concurrent:concurrent-futures-ktx:1.1.0
      └─ org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.4
      └─ org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.7.1

How `kotlinx-coroutines-core-jvm` occurs in this graph? https://mvnrepository.com/artifact/androidx.concurrent/concurrent-futures-ktx/1.1.0 has a dependency on `org.jetbrains.kotlinx:kotlinx-coroutines-core`.

The problem here definitely is not in the presence of the `-jvm` suffix artifacts in the graph, but in the graph resolution itself.

vaslabs: sorry, made an indentation mistake, it comes from `androidx.test:core:1.6.1`

the issue is not the suffix, it's just happens that filtering them out works, but I think the proper implementation would be to look into the .module files in dependency resolution, and filter by platform type

0xnm: I would propose to remove this from `AndroidAppModule`, because it may cause the issues further, when some dependency with `-jvm` suffix _must_ be in the resolved graph.

Instead, you can manage dependencies resolution in the build script of the example, adding TODO marker like here https://github.com/com-lihaoyi/mill/pull/4188/files#diff-5f6fc77ef82493c72b1011020468273b24df3b3454dd7dba6b0c266a91dd2e02R42-R57.

vaslabs: tbh, I'm not happy with either, but I added the exclusion to the dependencies and the TODO for now.

Let me know if you are happy with this

0xnm: yes, looks better

lihaoyi:
I'm going to try and merge this PR before merging https://github.com/com-lihaoyi/mill/pull/4188, since it's been open WIP for a while and I don't want to drag it out too long

himanshumahajan138:
@vaslabs great now it's Merged 

Actually i was waiting for this merge coz i was working on the jetpack compose support, so @vaslabs are you familiar with how gradle or other build tools work with jetpack compose or in general how they manage the aar dependency resolution ?

Moreover, may be now we can move to jetpack compose (aar dependency resolution) part coz its required for further android support

Hope you Understand!

Issue #4171, opened 2024-12-22
Ugly log output in Mill 0.12.4


lefou:
After bumping a project from Mill `0.12.2` to `0.12.4`, the log often contains parts of the progress output ruler.

Example output (excerpt): `mill __.compile`

```
[3433] [warn] /home/lefou/work/bla/bla/framework/ui.binding.rcp/src/de/bla/richmodel/ui/binding/rcp/jface/formsupport/formeditor/OldRmfEditorPart.java:186:62: getMainBinding() in de.bla.richmodel.ui.binding.editor.EditorController ist veraltet
[3433] [warn] getEditorController().getMainBinding
[3433] [warn]                                     ^
[3433] [warn] /home/lefou/work/bla/bla/framework/ui.binding.rcp/src/de/bla/richmodel/ui/binding/rcp/jface/formsupport/formpage/RichModelFormPage.java:83:27: Mögliches "this"-Escape vor vollständiger Initialisierung der Unterklasse
[3433] [warn] initialize(editor)
[3433] [warn]                   ^
[3433] [warn] /home/lefou/work/bla/bla/framework/ui.binding.rcp/src/de/bla/richmodel/ui/binding/rcp/jface/userinteraction/JfaceListMultipleSelectionDialog.java:82:25: Mögliches "this"-Escape vor vollständiger Initialisierung der Unterklasse
[3433] [warn] setTitle("some title")
[3433] [warn]                       ^
[3433] [warn] /home/lefou/work/bla/bla/framework/ui.binding.rcp/src/org/vafada/swtcalendar/RepeatingButton.java:55:33: Mögliches "this"-Escape vor vollständiger Initialisierung der Unterklasse
[3433] [warn] addMouseListener(new MouseAdapter() {
[3433] [warn] ==================public void mouseDown(MouseEvent event) {
[3433] [warn] ==========================cancelRepeater();
[3433] [warn] 
[3433] [warn] ==========================if (event.button == 1) { // Left click
[3433] [warn] ============================ __.cobuttonPressed(event.stateMask, event.time);
[3433] [warn] 
[3433] [warn] ============================ __.corepeater = new Repeater(event.stateMask);
[3433] [warn] ============================ __.cogetDisplay().timerExec(initialRepeatDelay, repeater);
[3433] [warn] ==========================}
[3433] [warn] ==================}
[3433] [warn] 
[3433] [warn] ==================public void mouseUp(MouseEvent event) {
[...]

PR #4173, opened 2024-12-23
PromptLogger: Clean lines that contain a tab character

lefou:
Fix https://github.com/com-lihaoyi/mill/issues/4171


lihaoyi:
Should have my windows laptop back next week, when the time comes will do some manual testing before merging this

Issue #4176, opened 2024-12-24
Move override segment suffixing from planning time to task instantiation time


lihaoyi:
`example.thirdparty[gatling].testRepoRoot` seems to invalidate unnecessarily when running `./mill -w docs.fastPages` and editing `example/large/selective/9-selective-execution/build.mill`

The problem appears to come from the https://github.com/com-lihaoyi/mill/pull/1600, where we made tasks get labelled and assigned `.super.foo` suffixes during planning. Thus, depending on what exactly Mill is asked to evaluate, the overrid**den** `example.thirdparty[gatling].testRepoRoot` in the preparatory evaluation has the same label as the overrid**ing**`example.thirdparty[gatling].testRepoRoot` in the main evaluation. This causes them to collide on disk and results in the latter unnecessarily invalidating since the hashes appear to have changed (since the other task stomped over the json files on disk)

Even outside of `--watch`, this scenario can probably occur when running two different sets of tasks manually:

- An overridden task may have a `.super` suffix and the override task no `.super` suffix in the first run
- The overridden task has `.super` suffix with the override not evaluated in the second run the other
- Thus the override from the first run would share the overridden task segments/out-folder in the second run, potentially causing spurious invalidations since the two tasks will have different sets of inputs and input hashes

This whole "tasks are assigned their final label during planning" thing is an endless source of bugs, and was originally introduced because we started using the ordering of tasks during planning to determine override ordering which is necessary to assign `.super.Foo` suffixes as part of the labels in (https://github.com/com-lihaoyi/mill/pull/1600). Although that PR fixed stackable traits overrides from colliding in the same evaluation run, it introduced this issue which can cause overriden and overriding tasks to collide across multiple evaluation runs

We probably should try to detect move override detection and label allocation to task instantiation time to avoid this footgun.

lihaoyi:
Cross posted from the linked PR

Splitting out `downloadTestRepo` does mitigate the problem, but the underlying question remains why `"example.thirdparty[gatling].testRepoRoot"` seems to be an invalidation root when neither of its upstream tasks changed and it itself is not a source or input task

[...]

PR #4179, opened 2024-12-25
Split out `downloadTestRepo` from `testRepoRoot` in mill's own build

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/4176. `testRepoRoot.super` always needs to re-evaluate, because it is a `Task.Source`. But if we split the `downloadTestRepo` call into a separate cached task that ensures the download is cached and can be re-used every time our `testRepoRoot` override needs to re-evaluate

### DIFF
--- a/example/package.mill
+++ b/example/package.mill
@@ -286,23 +286,25 @@ object `package` extends RootModule with Module {
     val (repoPath, repoHash) = repoInfo(crossValue)
     def repoSlug = repoPath.split("/").last
 
-    def testRepoRoot = Task {
+    def downloadedRepo = Task{
       shared.downloadTestRepo(repoPath, repoHash, T.dest)
       val wrapperFolder = T.dest / s"$repoSlug-$repoHash"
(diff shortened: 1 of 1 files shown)


lihaoyi:
Splitting out `downloadTestRepo` does mitigate the problem, but the underlying question remains why `"example.thirdparty[gatling].testRepoRoot"` seems to be an invalidation root when neither of its upstream tasks changed and it itself is not a source or input task

lihaoyi:
The problem appears to come from the fact that the preparatory input evaluation done in selective execution uses a subset of the tasks, resulting in the inputs being labelled differently during planning, such that the overrid*den* `example.thirdparty[gatling].testRepoRoot` in the preparatory evaluation has the same label as the overrid*ing*`example.thirdparty[gatling].testRepoRoot` in the main evaluation, causing them to collide on disk and resulting in the latter unnecessarily invalidating

PR #4181, opened 2024-12-26
Move task override tracking from planning time to instantiation time

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/4176

- Make overriding of `Discover` mandatory, even in tests, so it always has useful data 
- Propagate `implicitMillDiscover` throughout the module hierarchy implicitly through `Ctx`
-  Combine `Discover#classInfo#declaredTasks#name` together with the linearization of the Scala trait hierarchy for each module, to decide whether or not a specific task is overridden or not
- Added a new unit test `optionalOverride` that fails on `main` and passes on this PR

Also took the opportunity to break up `TestGraphs`, `EvaluationTests` and `ResolveTests` a bit. Still pretty messy, but better than before


contrib/buildinfo/test/src/mill/contrib/buildinfo/BuildInfoTests.scala
```diff
@@ -17,6 +18,8 @@ object BuildInfoTests extends TestSuite {
     def scalaVersion = scalaVersionString
     def buildInfoPackageName = "foo"
     def buildInfoMembers = Seq.empty[BuildInfo.Value]
+
+    def millDiscover = Discover[this.type]
```
lefou, line 22: Shouldn't this be a `val` or `lazy val`. The object is already at the end of the inheritance hierarchy, and the `def` may cause wasteful re-initialisation of `Discover`.

lihaoyi: Probably can be a lazy val

PR #4201, opened 2024-12-28
Specify exact files to process in ktlint run

0xnm:
This PR improves `ktlint` support: it explicitly specifies the files to process.

Without that current `mill.kotlinlib.ktlint.KtlintModule/checkFormatAll` run, for example, will go through the files under the `out` folder, which is useless and pollutes logs.

Also this PR removes redundant flags from `reformatAll` and `checkFormatAll`: former will always be with `format=true` by definition, the latter will always be with `format=false`.

### DIFF
--- a/kotlinlib/src/mill/kotlinlib/ktlint/KtlintModule.scala
+++ b/kotlinlib/src/mill/kotlinlib/ktlint/KtlintModule.scala
@@ -1,10 +1,12 @@
 package mill.kotlinlib.ktlint
 
+import mainargs.arg
 import mill._
 import mill.api.{Loose, PathRef}
 import mill.define.{Discover, ExternalModule}
 import mill.javalib.JavaModule
 import mill.kotlinlib.DepSyntax
+import mill.main.Tasks
 import mill.util.Jvm
 
 /**
@@ -16,7 +18,13 @@ trait KtlintModule extends JavaModule {
    * Runs [[https://pinterest.github.io/ktlint/latest/install/integrations/ Ktlint]]
    */
   def ktlint(@mainargs.arg ktlintArgs: KtlintArgs): Command[Unit] = Task.Command {
-    KtlintModule.ktlintAction(ktlintArgs, ktlintConfig(), ktlintOptions(), ktlintClasspath())
+    KtlintModule.ktlintAction(
+      ktlintArgs,
+      sources(),
+      ktlintConfig(),
+      ktlintOptions(),
+      ktlintClasspath()
+    )
   }
 
   /**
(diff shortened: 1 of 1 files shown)


lihaoyi:
Thanks!

kotlinlib/src/mill/kotlinlib/ktlint/KtlintModule.scala
```diff
@@ -74,15 +81,14 @@ object KtlintModule extends ExternalModule with KtlintModule with TaskModule {
...
-   *              - when set, files are not formatted
    */
   def checkFormatAll(
-      check: mainargs.Flag = mainargs.Flag(value = false)
+      @arg(positional = true) sources: Tasks[Seq[PathRef]] =
+        Tasks.resolveMainDefault("__.sources")
   ): Command[Unit] = Task.Command {
     ktlintAction(
-      KtlintArgs(format = false, check = check.value),
+      KtlintArgs(format = false, check = false),
```
sideeffffect, line 90: Just a guess, but isn't this a typo? Shouldn't this be
```scala
KtlintArgs(format = false, check = true),
```

kotlinlib/src/mill/kotlinlib/ktlint/KtlintModule.scala
```diff
@@ -57,15 +65,14 @@ object KtlintModule extends ExternalModule with KtlintModule with TaskModule {
...
-   *              - when set, files are not formatted
    */
   def reformatAll(
-      check: mainargs.Flag = mainargs.Flag(value = false)
+      @arg(positional = true) sources: Tasks[Seq[PathRef]] =
+        Tasks.resolveMainDefault("__.sources")
   ): Command[Unit] = Task.Command {
     ktlintAction(
-      KtlintArgs(format = true, check = check.value),
+      KtlintArgs(format = true, check = true),
```
sideeffffect, line 74: Just a guess, but isn't this a typo? Shouldn't this be
```scala
KtlintArgs(format = true, check = false),
```
@0xnm

0xnm: no, because `check` flag controls exit code handling (see https://github.com/com-lihaoyi/mill/blob/d2df820a88b08933870e299231faa28bb94945ba/kotlinlib/src/mill/kotlinlib/ktlint/KtlintModule.scala#L136-L140). So if format succeeds, exit code will be 0, but we still want to throw exception if something went wrong.

sideeffffect: Oh, I see, cool, thank :)

kotlinlib/src/mill/kotlinlib/ktlint/KtlintModule.scala
0xnm: good catch, it is a copy-paste typo here. https://github.com/com-lihaoyi/mill/pull/4247

Issue #4164, opened 2024-12-19
Mill creates a rt-21.0.5.jar file under home directory


dragove:
Mill creates a rt-21.0.5.jar file in home directory on Linux. This behavior occurs in versions 0.12.3 and 0.12.4, but it doesn't seem to happen in earlier versions.

mill script is installed via

```shell
curl -L https://github.com/com-lihaoyi/mill/releases/download/0.12.4/0.12.4 > mill && chmod +x mill
```

my environment:

OS: Arch Linux
Scala Version: 3.6.2

lefou:
I also found some `rt-<version>.jar` files in my home dir.

sequencer:
I just touch the `rt-<version>.jar` to avoid this creation, see [here](https://github.com/chipsalliance/t1/blob/052caecee0dcb16a27b4efb93d0fdb83975edaaf/nix/pkgs/mill-builder.nix#L70)...

asarkar:
It seems to create a jar corresponding version of JDK is used. I've different JDKs for different projects, and Mill creates one for each. This is ugly, I don't want it to clutter my home directory.

lihaoyi:
Yes this is definitely a bug and regression somewhere, we used to put it in '~/.mill` but somewhere along we messed up the logic

lefou:
IIRC we use the location associated with `Ctx.home` and the `--home` option, which should default to `$HOME/.mill/ammonite`. We deprecated it for a while, since it was assumed an Ammonite legacy. I don't know why it's created in the home directory though.

PR #4206, opened 2024-12-30
Check if rt.jar is still necessary

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/4164

It's been around since we used to use Ammonite to compile scripts, and IIRC was necessary to make the in-memory Scala compiler work. We no longer go through that code path, so maybe we can get rid of it? AFAIK we now only use the Scala compiler through Zinc which manages this stuff itself


lihaoyi:
Seems like the answer is that it is no longer necessary!

Issue #4202, opened 2024-12-29
Define best practice for wiring up mill test runs to github actions


lihaoyi:
Currently Mill's own build just outputs things to stdout for us to scroll through. Other projects probably do something similar. We should figure out a better way of integrating the structured test.xml output files into github actions markdown reports, so we can easily see summaries of the test results without scrolling through thousands of lines of logs

lihaoyi:
https://github.com/com-lihaoyi/mill/pull/4218 helps a lot, remaining improvements would be to try and show stack traces (https://github.com/mikepenz/action-junit-report/issues/1278) and task names in the annotations

PR #4218, opened 2024-12-31
Add junit report action

sake92:
Improves #4202 

Example reports: https://github.com/sake92/mill/actions/runs/12563637492

If you click on a failed job on the left, you will see annotations section on the top.  
It shows which tests failed and the reason message.


lihaoyi:
Thanks @sake92 ! it looks great

Issue #4221, opened 2025-01-02
Troubleshooting hanging build


nafg:
I don't know if this is a bug in Mill or something dumb in my build. If the latter, then perhaps there is a straightforward way to troubleshoot it that either exists and I need to learn, or that could be implemented. For instance, a way to get a "call graph stack trace" -- exactly what tasks are running or waiting for other tasks to be able to run, right now.

I'm using Mill 0.12.4.

In my build I am using a Worker Task for a testcontainers database instance, using Flyway to run migrations against it, and then generating code from that with [slick-additions-codegen](https://github.com/nafg/slick-additions?tab=readme-ov-file#slick-additions-codegen).

I ran Mill with `-i`, something along the lines of `mill -i -w __.publishLocal`. I experienced this with any `-w` command line in this project though.

I generated a stack trace by pressing `Ctrl`+`\`. You can see the resulting terminal content here: https://gist.github.com/nafg/d3b2590ae27dea05aa1ec1224ac5abeb

I added my `build.mill` to the gist, where I appended `.scala` so GitHub should render syntax highlighting.

nafg:
I suppose I should just add a lot of logging to my tasks.

nafg:
BTW the container is `AutoCloseable`, if that makes a difference

lihaoyi:
@nafg the prompt at the bottom of the terminal is meant to tell you what tasks are actively running at any point in time, does it list any tasks?

lihaoyi:
Also, can you use `jstack` on the MillServerMain process to get its stack traces? `^\` seems to mix it yo with the client stack traces which makes it impossible to read

nafg:
> @nafg the prompt at the bottom of the terminal is meant to tell you what tasks are actively running at any point in time, does it list any tasks?

Just the "goal" task, e.g. `__.publishLocal` like at the end of the gist

nafg:
> Also, can you use `jstack` on the MillServerMain process to get its stack traces? `^\` seems to mix it yo with the client stack traces which makes it impossible to read

ok I added it here: https://gist.github.com/nafg/d3b2590ae27dea05aa1ec1224ac5abeb#file-jstack-txt

nafg:
Another thought, what if while a task is running (at least in watch mode), there was a key to press that would interrupt the current execution but either restart it, or go (back) to watch mode, waiting until sources change.

lihaoyi:
So this stack means the main evaluator is waiting for some tasks to complete

[...]

PR #4223, opened 2025-01-02
Catch fatal exceptions in futures

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/4221

Otherwise they just silently terminate the future and leave downstream `Await`s hanging, resulting in the process hanging waiting for a future that will never complete. `def reportFailure` doesn't seem to log stuff properly when this happens for some reason

Covered with additional integration tests


main/eval/src/mill/eval/EvaluatorCore.scala
```diff
@@ -124,72 +124,77 @@ private[mill] trait EvaluatorCore extends GroupEvaluator {
...
+                if (res.valueHashChanged) changedValueHash.put(terminal, ())
+
+                profileLogger.log(terminal, duration, res, deps)
+
+                Some(res)
+              }
+            } catch {
+              case e: Throwable if !scala.util.control.NonFatal(e) =>
+                throw new Exception(e)
             }
```
lefou, line 198: If this is an `OutOfMemoryError`, chances are high, creating a new object with `new` will again trigger a new `OutOfMemoryError`.

lihaoyi: Sure, but that's not any worse than the status quo, and for the other ones we have a chance to fail the build gracefully rather than hanging the process

nafg: Is that worse than the current behavior?

What do you suggest?

nafg: weird that I saw the first comment and not the second

PR #4246, opened 2025-01-03
Honor the NO_COLOR env variable

gamlerhart:
When the NO_COLOR env variable is set,
then avoid coloring the terminal.

--color argument does overrides the env variable

Inspired by https://no-color.org
and asked in discussion: https://github.com/com-lihaoyi/mill/discussions/4170

### DIFF
--- a/runner/src/mill/runner/MillMain.scala
+++ b/runner/src/mill/runner/MillMain.scala
@@ -162,7 +162,8 @@ object MillMain {
             (false, RunnerState.empty)
 
           case Right(config) =>
-            val colored = config.color.getOrElse(mainInteractive)
+            val noColorViaEnv = env.get("NO_COLOR").exists(v => v != "")
+            val colored = config.color.getOrElse(mainInteractive && !noColorViaEnv)
             val colors = if (colored) mill.util.Colors.Default else mill.util.Colors.BlackWhite
 
             if (!config.silent.value) {
--- a/runner/src/mill/runner/MillMain.scala
+++ b/runner/src/mill/runner/MillMain.scala
@@ -162,7 +162,7 @@ object MillMain {
             (false, RunnerState.empty)
 
           case Right(config) =>
-            val noColorViaEnv = env.get("NO_COLOR").exists(v => v != "")
+            val noColorViaEnv = env.get("NO_COLOR").exists(_.nonEmpty)
             val colored = config.color.getOrElse(mainInteractive && !noColorViaEnv)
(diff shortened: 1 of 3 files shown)


rolandtritsch (approved):
Thanks for the fix. NO_COLOR for the win!

rolandtritsch:
BTW ... are you also going to update the NO_COLOR website or do you want me to do that? - https://github.com/jcs/no_color

runner/src/mill/runner/MillMain.scala
```diff
@@ -162,7 +162,8 @@ object MillMain {
             (false, RunnerState.empty)
 
           case Right(config) =>
-            val colored = config.color.getOrElse(mainInteractive)
+            val noColorViaEnv = env.get("NO_COLOR").exists(v => v != "" && v != "0")
```
lefou, line 165: Spec (https://no-color.org/) says, the value must be non empty.

> Command-line software which adds ANSI color to its output by default should check for a NO_COLOR environment variable that, when present and not an empty string (regardless of its value), prevents the addition of ANSI color.

Technically even a `"0"` should enable the `NO_COLOR` request.

lefou (changes requested):

runner/src/mill/runner/MillMain.scala
gamlerhart: Yes...I found it a bit odd, as other env variables often have a way to disable it as well with a explicit value. Zero made sense to me.

Changed to follow the spec ;).

runner/src/mill/runner/MillMain.scala
```diff
@@ -162,7 +162,8 @@ object MillMain {
             (false, RunnerState.empty)
 
           case Right(config) =>
-            val colored = config.color.getOrElse(mainInteractive)
+            val noColorViaEnv = env.get("NO_COLOR").exists(v => v != "")
```
lolgab, line 165:
```suggestion
            val noColorViaEnv = env.get("NO_COLOR").exists(_.nonEmpty)
```

lefou (approved):
Looks good to me. Using `_.nonEmpty` (as @lolgab requested) would be even nicer. And a cross reference to the no-color.org site would be nice too.

lefou:
> And a cross reference to the no-color.org site would be nice too.

@gamlerhart Sorry for being ambiguous. I meant to add Mill to the no-color.org site. I think the link in the `--help` is a bit too much.

PR #4253, opened 2025-01-06

monyedavid:
This pr implements the examples for jslib/dependencies.

https://github.com/com-lihaoyi/mill/issues/3927

Checklist:
- [x] **example/jslib/testing**
     - [x]  3-integration-suite-cypress
     - [x]  3-integration-suite-playwright


example/javascriptlib/testing/3-integration-suite-cypress/server/src/server.ts
```diff
@@ -0,0 +1,82 @@
+import * as http from 'http';
```
lihaoyi, line 1: Can this server code be shortened? Most of this looks like boilerplate 404/500/200 return code management that should be handled by some framework, and should not be something the user needs to write manually. Minimizing dependencies is good, but I think in this case it results in verbosity that gets in the way of understanding the example

example/javascriptlib/testing/3-integration-suite-cypress/client/src/reportWebVitals.ts
```diff
@@ -0,0 +1,15 @@
+import { ReportHandler } from 'web-vitals';
```
lihaoyi, line 1: Is this a necessary part of the example?

lihaoyi:
I think at a high level, this example needs to be significantly shortened. Let's trim all the CSS, `robots.txt`, logos, all the boilerplate comments from the `.html` files: just leave a JS server, JS client, and the integration tests. 

This example doesn't need to be realistic, since there are other examples in the `javascriptlib/` folder that show the other parts, so we can just focus on the core of how the integration tests are wired up and written

example/javascriptlib/testing/3-integration-suite-cypress/server/src/server.ts
monyedavid: Sure

example/javascriptlib/testing/3-integration-suite-cypress/client/src/reportWebVitals.ts
monyedavid: nah, taking it out

example/javascriptlib/testing/3-integration-suite-cypress/server/src/server.ts
monyedavid: I just replaced, it with express. Forgot that was an option.

monyedavid:
@lihaoyi are you aware mill can no longer handle special characters?

when matching: "typescript/src/server/playwright/app.test.ts:4:9 › React App › displays the heading..."
the `›` amongst other will throw cause a failed matched. I am positive at some point that would work as expected.

lihaoyi:
I think this looks good, thanks!

PR #4254, opened 2025-01-06
Fix race condition of timeout thread interrupt to stabilize multi level build tests

lihaoyi:
The basic problem was that `thread.interrupt()` running before `interrupt = false` meant there was a chance the thread's `if (interrupt) {` conditional would be `true`, resulting in the timeout thread timing out the open socket immediately, and the Mill server process shutting down.

The solution is to move `interrupt = false` to before `thread.interrupt()`

This should hopefully fix the flakiness in multi-level-build tests, where the process shutting down would cause the next command to re-spawn a new process, resulting in all classloaders to be re-spawned, violating our assertions

Improved the test error checking so next time a similar thing happens, we get more precise reporting "an unwanted process restart occurred", rather than just "classloader invalidation was unexpected".

Tested manually with `while ./mill 'integration.invalidation[multi-level-editing].server.test'; do :; done`. This seems to reproduce the problem on my laptop in a few tens of minutes without this fix, after this fix I haven't managed to make it appear


lihaoyi:
I'm hoping this also fixes the flakiness in `example.fundamentals.tasks[6-workers].local.test` (e.g. https://github.com/com-lihaoyi/mill/actions/runs/12579809859/job/35060755776) and `integration.failure[fatal-error].local.test` (e.g. https://github.com/com-lihaoyi/mill/actions/runs/12591901838/job/35095817421), both of which could be explained by the Mill process from the previous invocation exiting unexpectedly before the subsequent invocation happens

PR #4261, opened 2025-01-07
Replace waiting for android emulator with a sleep and a list device call to verify emulator is running

vaslabs:
I've tried to replicate why the timeouts related to the android emulator that happen in github actions. It seems that  the `adb wait-for-device` is stuck although I cannot replicate it locally and the kotlin example does not suffer the same issue.

So for now, I added some wait time after start emulator and a call to `adb device` to verify the device is indeed present and it seems to be working.

This should work for the time being and make the job green until we create a better alternative.

Small change included: Since we throw an error if we don't detect the boot notification of start emulator, I changed the `startAndroidEmulator` signature from `Option[String]` to `String`

Linking the green job on my copy for reference: https://github.com/vaslabs-ltd/mill/actions/runs/12651834850/job/35253442650

### DIFF
--- a/example/android/javalib/1-hello-world/build.mill
+++ b/example/android/javalib/1-hello-world/build.mill
@@ -123,8 +123,8 @@ object app extends AndroidAppModule {
 
 > ./mill show app.startAndroidEmulator
 
-> ./mill show app.waitForDevice
-...emulator-5554...
+> sleep 20 && ./mill show app.adbDevices
+...emulator-5554...device...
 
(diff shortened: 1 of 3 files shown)


lihaoyi:
Looks like CI passed! Hope we start seeing it turn green again in main

0xnm:
what is the need for the `sleep` call in the example script? I don't think we can expect that all users will do this. Instead, the API should be self-contained: operation should be blocking until it is complete, e.g. `startAndroidEmulator` should finish only when emulator is up and completely ready for the interaction.

PR #4262, opened 2025-01-07
Add multi-module BSP tests

alexarchambault:
This PR adds multi-module non-regression tests of the BSP support. This is helpful for https://github.com/com-lihaoyi/mill/pull/4145, that touches some BSP things in `MillBuildServer`.


lefou:
I find it hard to read those dependency tests in general. There is often a motivation for a test, what should be tested, why were specific dependencies used and what is the significant part of the comparison? Maybe, you can add one or two lines, so the later dev understands the test better?

alexarchambault:
> I find it hard to read those dependency tests in general. There is often a motivation for a test, what should be tested, why were specific dependencies used and what is the significant part of the comparison? Maybe, you can add one or two lines of comments, so the later dev understands the test better?

Sure, I just added some comments

lefou (approved):
Looks good to me.

PR #4270, opened 2025-01-09
Update PublishModule.scala to additional check for MAVEN_LOCAL_REPO 

imdaveho:
This was referenced in #4269. This checks for m2 (Maven)


lihaoyi:
This looks fine. @imdaveho do you have a link to any reference about what `MAVEN_LOCAL_REPO` and `IVY_HOME` environment variables mean? Would be good to link them in a comment for future maintainers who aren't familiar with them

scalalib/src/mill/scalalib/PublishModule.scala
```diff
@@ -247,10 +247,14 @@ trait PublishModule extends JavaModule { outer =>
   /**
    * Publish artifacts to a local Maven repository.
    * @param m2RepoPath The path to the local repository  as string (default: `$HOME/.m2repository`).
+   *                   If set, environment variable `MAVEN_LOCAL_REPO` will be used.
```
lefou, line 250:
```suggestion
   *                   If not set, falls back to `maven.repo.local` system property or `MAVEN_LOCAL_REPO` environment variable or `~/.m2/repository`.
```

scalalib/src/mill/scalalib/PublishModule.scala
```diff
@@ -262,7 +266,9 @@ trait PublishModule extends JavaModule { outer =>
    */
   def publishM2LocalCached: T[Seq[PathRef]] = Task {
     publishM2LocalTask(Task.Anon {
-      os.Path(os.home / ".m2/repository", T.workspace)
+      sys.props.get("maven.repo.local").map(os.Path(_))
+        .getOrElse(sys.env.get("MAVEN_LOCAL_REPO").map(os.Path(_))
+          .getOrElse(os.Path(os.home / ".m2", T.workspace))) / "repository"
```
lefou, line 271: This needs to be put into a dedicated `Tast.Input`, otherwise cache-invalidation won't trigger when one of the variables change.

Once, we have that, we can reuse it for the default of `publishM2Local` to.

imdaveho:
> This looks fine. @imdaveho do you have a link to any reference about what `MAVEN_LOCAL_REPO` and `IVY_HOME` environment variables mean? Would be good to link them in a comment for future maintainers who aren't familiar with them

again, this might be against jvm conventions, but I did not find definitive references in the maven or ivy project pages, but there are several references in unofficial pages:

* https://javadoc.io/doc/org.owasp/dependency-check-utils/9.0.4/index-all.html
* https://docs.openshift.com/container-platform/3.11/using_images/s2i_images/java.html
* https://strongbox.github.io/user-guide/tool-integration/ant-ivy-example.html

I would say reject this PR and the other one (https://github.com/com-lihaoyi/mill/pull/4269)  if this isn't convention to use environment variables, and rather to pass in properties using JAVA_OPTS or MAVEN_OPTS or through -D using something like .mill_opts.

scalalib/src/mill/scalalib/PublishModule.scala
imdaveho: Do you mean a Tas*k*.Input? I'm not sure how to do that (fairly new to this). Is it just creating a new function that wraps this portion so that this function can call that function?

As I mentioned regarding conventions, this is more of a preference issue not a functional one. 

I'm OK with closing this, since I'm not seeing official documentation on the exact environment variable I used. I _am_ seeing references to setting the property with MAVEN/JAVA_OPTS. If that is the case, I would remove the environment variable and focus on adding the properties.

I don't know what you mean about Task.Input. I don't see this as part of the Mill API.

scalalib/src/mill/scalalib/PublishModule.scala
imdaveho: done in the latest commit

scalalib/src/mill/scalalib/PublishModule.scala
```diff
@@ -247,23 +247,27 @@ trait PublishModule extends JavaModule { outer =>
...
    * @return [[PathRef]]s to published files.
    */
   def publishM2LocalCached: T[Seq[PathRef]] = Task {
-    publishM2LocalTask(Task.Anon {
-      os.Path(os.home / ".m2/repository", T.workspace)
-    })()
+    publishM2LocalTask(getM2LocalRepoPath)()
+  }
+
+  def getM2LocalRepoPath: Task[os.Path] = Task.Input {
```
lefou, line 268: I'd rather name this `m2LocalRepoPath`

scalalib/src/mill/scalalib/PublishModule.scala
```diff
@@ -247,23 +247,27 @@ trait PublishModule extends JavaModule { outer =>
...
+   *                   If not set, falls back to `maven.repo.local` system property or `~/.m2/repository`
    * @return [[PathRef]]s to published files.
    */
-  def publishM2Local(m2RepoPath: String = (os.home / ".m2/repository").toString())
-      : Command[Seq[PathRef]] = Task.Command {
-    publishM2LocalTask(Task.Anon {
-      os.Path(m2RepoPath, T.workspace)
-    })()
-  }
+  def publishM2Local(m2RepoPath: String = getM2LocalRepoPath.toString()): Command[Seq[PathRef]] =
```
lefou, line 253: This doesn't look right. Since `getM2LocalRepoPath` is a task, the `toString` return it's task-name instead of the actual path.

lefou: Oh, this is already merged. Either I read the code wrong or we have no proper test coverage, otherwise we should see some test failures.

PR #4272, opened 2025-01-09
Enable GraalVM tests on Windows

lihaoyi:


scalalib/src/mill/scalalib/NativeImageModule.scala
```diff
@@ -31,14 +31,16 @@ trait NativeImageModule extends RunModule with WithZincWorker {
    */
   def nativeImage: T[PathRef] = Task {
     val dest = T.dest
-    val executableName = "native-image"
+    val executableName =
+      if (mill.main.client.Util.isWindows) "native-image.exe" else "native-image"
+
```
ayewo: I currently just have:
`val executableName = "native-image"` instead of the `if` check, because the Graal native-image script will automatically append an `.exe` extension for Windows builds.


If you don't remove the trailing `.exe`, the built file will have the name `native-image.exe.exe` leading to a file not found error in the Mill tests.

scalalib/src/mill/scalalib/NativeImageModule.scala
```diff
@@ -31,14 +31,16 @@ trait NativeImageModule extends RunModule with WithZincWorker {
...
       .+=(nativeImageTool().path.toString)
       .++=(nativeImageOptions())
       .+=("-cp")
       .+=(nativeImageClasspath().iterator.map(_.path).mkString(java.io.File.pathSeparator))
       .+=(finalMainClass())
-      .+=(executableName)
+      .+=((dest / executableName).toString())
       .result()
     os.proc(command).call(cwd = dest, stdout = os.Inherit)
     PathRef(dest / executableName)
```
ayewo: I'm currently using the `if-isWindows` check here instead:
```scala
PathRef(dest / (if (mill.main.client.Util.isWindows) s"${executableName}.exe" else executableName))
```

example/scalalib/publishing/7-native-image/build.mill
```diff
@@ -24,10 +24,11 @@ object foo extends ScalaModule with NativeImageModule {
...
 > ./mill show foo.ZincWorkerGraalvm.javaHome
 
-> ./mill foo.nativeImage
+> ./mill show foo.nativeImage
 GraalVM Native Image: Generating...native-image...
 Finished generating...native-image...
+???
 
-> ./out/foo/nativeImage.dest/native-image
+> ./out/foo/nativeImage.dest/native-image.exe
```
ayewo, line 31: `> ./out/foo/nativeImage.dest/native-image` without the .exe worked in my testing.

lihaoyi: Yeah, it seems that `native-image` appends the `.exe` automatically, but at the same time the windows terminal lets you run the file without writing out the `.exe` and it still works (at least until it hits the classloading issue earlier)

ayewo: I finally got it working here: https://github.com/ayewo/mill/actions/runs/12719302376/job/35459278855

I'll tidy up my changes and open a PR.

PR #4276, opened 2025-01-09
Cleanup after native image

lihaoyi:
* Bootstrap native image JVM using Coursier, rather than just hardcoding `java`
* Allow custom JVMs to be used for Mill server by specifying `.mill-jvm-version`
* Make sure we use separate download paths for `-native` and non`-native` mill executables
* Tidy up `dist/package.mill`
* Remove unnecessary `def fix` overrides now that `.fix` on empty modules no longer crashes

### DIFF
--- a/readme.adoc
+++ b/readme.adoc
@@ -177,12 +177,15 @@ remaining modules to `~/.ivy2/local`.
 This results in a more realistic test environment, but at the cost of taking tens-of-seconds
 more to run a test after making a code change.
 
-You can reproduce these tests manually using `dist.assembly`:
+You can reproduce these tests manually using `dist.installLocal`:
 
 [source,bash]
 ----
-./mill dist.assembly && (cd example/basic/1-simple && ../../../out/dist/assembly.dest/mill run --text hello)
+./mill dist.installLocal && (cd example/basic/1-simple && ../../../mill-assembly.jar run --text hello)
 ----
+
+ThYou can also use `dist.native.installLocal` for a Graal Native Image executable,
+which is slower to create but faster to start than the default executable assembly.
 
 There are two six of these tests, `.{local,assembly,native}.{server,fork}`.
 
--- a/readme.adoc
+++ b/readme.adoc
@@ -184,7 +184,7 @@ You can reproduce these tests manually using `dist.installLocal`:
 ./mill dist.installLocal && (cd example/basic/1-simple && ../../../mill-assembly.jar run --text hello)
 ----
 
-ThYou can also use `dist.native.installLocal` for a Graal Native Image executable,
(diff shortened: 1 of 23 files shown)


runner/client/src/mill/runner/client/MillProcessLauncher.java
```diff
@@ -106,15 +106,17 @@ static boolean isWin() {
...
   static String javaExe() {
-    final String javaHome = System.getProperty("java.home");
-    if (javaHome != null && !javaHome.isEmpty()) {
-      final File exePath = new File(
-          javaHome + File.separator + "bin" + File.separator + "java" + (isWin() ? ".exe" : ""));
-      if (exePath.exists()) {
-        return exePath.getAbsolutePath();
-      }
+    String javaHome = System.getProperty("java.home");
+    String jvmId = "temurin:17";
```
alexarchambault, line 110: FYI, in Scala CLI (and also maybe in the coursier CLI), there's a [check of the version of the local JVM](https://github.com/VirtusLab/scala-cli/blob/51bebb087f9adaf1f1f4760374feb1a212b63bc9/modules/core/src/main/scala/scala/build/internals/OsLibc.scala#L89) before downloading one. If the version of the local one is high enough, it just uses it rather than downloading a new one.

readme.adoc
```diff
@@ -177,13 +177,16 @@ remaining modules to `~/.ivy2/local`.
...
-You can reproduce these tests manually using `dist.assembly`:
+You can reproduce these tests manually using `dist.installLocal`:
 
 [source,bash]
 ----
-./mill dist.assembly && (cd example/basic/1-simple && ../../../out/dist/assembly.dest/mill run --text hello)
+./mill dist.installLocal && (cd example/basic/1-simple && ../../../mill-assembly.jar run --text hello)
 ----
 
+ThYou can also use `dist.native.installLocal` for a Graal Native Image executable,
```
lefou, line 187:
```suggestion
You can also use `dist.native.installLocal` for a Graal Native Image executable,
```

runner/client/src/mill/runner/client/CoursierClient.scala
```diff
@@ -0,0 +1,32 @@
...
+      .withIndex(jvmIndex0())
+
+    val javaHome = JavaHome().withCache(jvmCache)
+
+    javaHome.get(id).unsafeRun()(coursierCache0.ec)
+  }
+
+  def jvmIndex0(): Task[JvmIndex] = {
+    val coursierCache0 = FileCache[Task]()
+      .withLogger(coursier.cache.loggers.RefreshLogger.create())
```
lihaoyi, line 21: @alexarchambault for some reason this doesn't seem to be showing any progress logger on the terminal. Not sure why

PR #4277, opened 2025-01-09
Eliminate sleep from android emulator doc examples

vaslabs:
### Summary

This should fix the android job back to green and has some improvements on handling the android emulator in the CI (and locally) , part of which eliminates the need for sleep hack in documentation examples.

Job reference on my fork can be found [here](https://github.com/vaslabs-ltd/mill/actions/runs/12688329695/job/35364858469) . ~~Sometimes the device may be stuck in unauthorized, I don't know the reason yet and can't yet replicate it locally .~~

#### Unauthorized issue

In https://github.com/ReactiveCircus/android-emulator-runner, it seems that the same issue occurs when caching is used. ~~I'm wondering if there's any implicit caching that I have missed in the github actions~~

After experimenting with https://github.com/vaslabs-ltd/test-android-emulator , it seems that the runner avd directory is polluted. Although I'm not sure why, deleting it (at least in test-android-emulator) fixes the unauthorized issue . I've added the same in the github action here

### Improvements

- Removed the need for sleep in doc examples by implementing a new wait for device boot flag, that's used internally by startAndroidEmulator in addition to listening for the booted log message. Having both can help identify potential issues locally or in a CI . `Task.Command(exclusive = true)` as suggested by @lihaoyi seems to have made it very stable

### Fixes

- Fixed the bundle failing for the manifest/AndroidManifest.xml error message by flattening the zip file. I am not sure when this broke, but bundle is not in my radar in the immediate future, so my focus was to keep it green for now. Bundle build tool also seems to be marked as experimental, not that it's an issue, just a mention.

### Note

For the wait, I initially had it in a task and used a while loop but changed my mind and having it outside of task context seems easier to control the behaviour . I can convert it into a tailrec function if you prefer that style

EDIT: changed the outcome of unauthorized after some experimentation on a clean repo

### DIFF
--- a/.github/workflows/post-build-selective.yml
+++ b/.github/workflows/post-build-selective.yml
@@ -60,6 +60,10 @@ jobs:
           sudo udevadm control --reload-rules
           sudo udevadm trigger --name-match=kvm
 
+      - name: Cleanup any previous avd's to avoid signing key conflicts
+        if : ${{ inputs.install-android-sdk }}
+        run: rm -rf /home/runner/.config/.android/avd
+
       - name: Set AVD environment variable globally
         if: ${{ inputs.install-android-sdk }}
         run: echo "ANDROID_AVD_HOME=/home/runner/.config/.android/avd" >> $GITHUB_ENV
--- a/example/android/javalib/1-hello-world/build.mill
+++ b/example/android/javalib/1-hello-world/build.mill
@@ -123,7 +123,7 @@ object app extends AndroidAppModule {
 
 > ./mill show app.startAndroidEmulator
 
-> sleep 20 && ./mill show app.adbDevices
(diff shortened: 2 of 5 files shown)


lihaoyi:
Looks good to me. There's still a failure in `example.android`, but happy to merge regardless if you think this is a step forward @vaslabs

vaslabs:
> Looks good to me. There's still a failure in `example.android`, but happy to merge regardless if you think this is a step forward @vaslabs

I've managed to replicate the issue here https://github.com/vaslabs-ltd/test-android-emulator , seems consistent, I'll spend some more time on it and ping you when ready

vaslabs:
@lihaoyi seems to be ready

PR #4290, opened 2025-01-11
update `mill.bat` to handle `-native` extension

lihaoyi:


lihaoyi:
The download logic and url/path mangling logic seems to work, but I can't test the executable since my windows laptop is on ARM. @ayewo maybe you could give this a try and see if it works for you (as long as you have an x64 windows machine)?

ayewo:
@lihaoyi sure will test it on Windows today. I'm currently AFK.

lihaoyi:
CI is mostly green, which indicates that this does work at least on GHA, and I did a control experiment in https://github.com/com-lihaoyi/mill/pull/4290/commits/6f113576f6d8273d26692e6b49e5c76992e9708c to make sure we are exercising the new code path. But would still be good to test out on an actual person's machine

PR #4293, opened 2025-01-12

monyedavid:
This pr implements the examples for jslib/dependencies.

https://github.com/com-lihaoyi/mill/issues/3927

Checklist:
- [x] **example/jslib/publish**
     - [x]  1-publish
     - [x]  2-realistic
     
Key changes:
- Packages can now be imported without the node_modules prefix.
- PublishMeta - package.json (for publish) generation
- Prepare js code for publishing in publish directory


example/javascriptlib/publishing/2-realistic/build.mill
```diff
@@ -0,0 +1,124 @@
...
+
+      Seq(PathRef(Task.dest))
+    }
+
+    def npmDeps = Seq("immutable@4.3.7")
+  }
+
+}
+
+class qux extends TypeScriptModule {
```
lihaoyi, line 36: Generally modules should be `trait`s rather than `class`es

example/javascriptlib/publishing/2-realistic/build.mill
```diff
@@ -0,0 +1,124 @@
...
+
+object foo extends TypeScriptModule {
+  def generatedSources = Task {
+    os.write(
+      Task.dest / "foo.generated.ts",
+      s"""export default class FooGen {
+         |  static value: number = 123
+         |}
+      """.stripMargin
+    )
```
lihaoyi, line 13: We don't need to demonstrate `generatedSources` in every submodule in this file, just once is enough

example/javascriptlib/publishing/2-realistic/build.mill
```diff
@@ -0,0 +1,124 @@
...
+
+    Seq(PathRef(Task.dest))
+  }
+
+  object test extends TypeScriptTests with TestModule.Jest
+}
+
+object quxdev extends qux
+
+object quxpub extends qux with PublishModule {
```
lihaoyi, line 56: Can we publish `foo` and `bar` as separate modules with their own `publishMeta`?

example/javascriptlib/publishing/2-realistic/build.mill
```diff
@@ -0,0 +1,124 @@
...
+      """.stripMargin
+    )
+
+    Seq(PathRef(Task.dest))
+  }
+
+  object test extends TypeScriptTests with TestModule.Jest
+}
+
+object quxdev extends qux
```
lihaoyi, line 54: Do we still need `quxdev` if `quxpub` exists?

main/package.mill
```diff
@@ -90,6 +90,7 @@ object `package` extends RootModule with build.MillStableScalaModule with BuildI
 
     def ivyDeps = Agg(
       build.Deps.osLib,
+      build.Deps.sbtIo,
```
lihaoyi, line 93: What's this `sbtIo` dependency for?

javascriptlib/src/mill/javascriptlib/PublishModule.scala
```diff
@@ -0,0 +1,539 @@
...
+  private def mainType: T[String] = Task {
+    main().replaceFirst(bundledOut(), declarationOut()).replaceAll("\\.js", ".d.ts")
+  }
+
+  // Define exports for the package
+  // by default: mainFile is exported
+  // use this to define other exports
+  def exports: T[Map[String, String]] = Task { Map.empty[String, String] }
+
+  private def mkExports: T[Map[String, PublishModule.ExportEntry]] = Task {
```
lihaoyi, line 37: All these task names should be prefixed with `publish` so it's clear what they are for and what trait they came from

javascriptlib/src/mill/javascriptlib/PublishModule.scala
```diff
@@ -0,0 +1,539 @@
...
+  def publishMeta: T[PublishModule.PublishMeta]
+
+  override def npmDevDeps: T[Seq[String]] =
+    Task { Seq("glob@^10.4.5", "ts-patch@3.3.0", "typescript-transform-paths@3.5.3") }
+
+  def bundledOut: T[String] = Task { "dist" }
+
+  private def declarationOut: T[String] = Task { "declarations" }
+
+  override def mainFileName: T[String] = Task { s"${millSourcePath.last}.js" }
```
lihaoyi, line 22: This doesn't seem related to publishing; should it be a default in the upstream `TypescriptModule`? Or should we leave it for the user to configure?

javascriptlib/src/mill/javascriptlib/PublishModule.scala
```diff
@@ -0,0 +1,539 @@
...
+
+  override def symlinkNodeModules: Task[Unit] = Task {
+    tsPatchInstall() // patch typescript compiler => use custom transformers
+    os.call(
+      ("ln", "-s", npmInstall().path.toString + "/node_modules/", "node_modules"),
+      cwd = publishDir().path
+    )
+    if (os.exists(npmInstall().path / ".npmrc")) os.call(
+      ("ln", "-s", npmInstall().path.toString + "/.npmrc", ".npmrc"),
+      cwd = publishDir().path
```
lihaoyi, line 265: These should be using `os.symlink` rather than `os.call`

javascriptlib/src/mill/javascriptlib/PublishModule.scala
```diff
@@ -0,0 +1,539 @@
...
+      main = main(),
+      types = mainType(),
+      exports = Map("." -> PublishModule.Export("./" + main())) ++ mkExports(),
+      bin = json.bin.map { case (k, v) => (k, "./" + bundledOut() + "/" + v) },
+      typesVersions = mkTypesVersion(),
+      dependencies = transitiveNpmDeps().map { deps => splitDeps(deps) }.toMap,
+      devDependencies = transitiveNpmDeps().map { deps => splitDeps(deps) }.toMap
+    ).toJsonClean
+
+    os.write.over(publishDir().path / "package.json", updatedJson)
```
lihaoyi, line 75: You should never write to other task folders. You should instead return a `PathRef` to your own `Task.dest` folder and downstream tasks can copy and rename the files as necessary

javascriptlib/src/mill/javascriptlib/PublishModule.scala
```diff
@@ -0,0 +1,539 @@
...
+
+  def publish(): Command[CommandResult] = Task.Command {
+    // build package.json
+    packageJson()
+
+    // bundle code for publishing
+    bundle()
+
+    // run npm publish
+    os.call(("npm", "publish"), cwd = publishDir().path)
```
lihaoyi, line 383: Probably better to make this a `Command[Unit]` and make `os.call` have `stdout = os.Inherit` so users can see the logs

javascriptlib/src/mill/javascriptlib/PublishModule.scala
```diff
@@ -0,0 +1,539 @@
...
+          "afterDeclarations" -> true
+        )
+      ),
+      "moduleResolution" -> ujson.Str("node"),
+      "module" -> ujson.Str("CommonJS"),
+      "target" -> ujson.Str("ES2020")
+    )
+  }
+
+  // patch typescript
```
lihaoyi, line 248: Do we really need to patch typescript to get rid of the `node_modules` prefix? Are there other ways to do it?

javascriptlib/src/mill/javascriptlib/PublishModule.scala
```diff
@@ -0,0 +1,539 @@
...
+      Seq((
+        mod.millSourcePath.subRelativeTo(Task.workspace).toString + "/*",
+        s"typescript/${mod.millSourcePath.subRelativeTo(Task.workspace).toString}/src" + ":" + s"${declarationOut()}"
+      )) ++
+        res.map { rp =>
+          val resourceRoot = rp.path.last
+          val modName = mod.millSourcePath.subRelativeTo(Task.workspace).toString
+          // nb: resources are be moved in bundled stage
+          (
+            "@" + modName + s"/$resourceRoot/*",
```
lihaoyi, line 199: `s"@$modName/$resourceRoot/*"`

javascriptlib/src/mill/javascriptlib/PublishModule.scala
```diff
@@ -0,0 +1,539 @@
...
+  }
+
+  override def upstreamPathsBuilder: Task[Seq[(String, String)]] = Task.Anon {
+
+    val upstreams = (for {
+      (res, mod) <- Task.traverse(moduleDeps)(_.resources)().zip(moduleDeps)
+    } yield {
+      Seq((
+        mod.millSourcePath.subRelativeTo(Task.workspace).toString + "/*",
+        s"typescript/${mod.millSourcePath.subRelativeTo(Task.workspace).toString}/src" + ":" + s"${declarationOut()}"
```
lihaoyi, line 192:
```scala
val relative = mod.millSourcePath.subRelativeTo(Task.workspace)
s"typescript/$relative/src:${declarationOut()}"
```

javascriptlib/src/mill/javascriptlib/PublishModule.scala
```diff
@@ -0,0 +1,539 @@
...
+      os.makeDir.all(destination / os.up)
+      os.copy.over(
+        target.path,
+        destination
+      )
+    }
+  }
+
+  private def pubCpModDeps: T[Unit] = Task {
+    val cpTargets = pubModDeps()
```
lihaoyi, line 144: What's `cp`? Can we use a longer name for clarity?

javascriptlib/src/mill/javascriptlib/PublishModule.scala
```diff
@@ -0,0 +1,539 @@
...
+      for {
+        pr <- modSource
+        file <- os.walk(pr.path)
+        if fileExt(file)
+      } yield PathRef(file)
+    }).flatten
+  }
+
+  // mv generated sources for base mod and its deps
+  private def pubMvGenSources: T[Unit] = Task {
```
lihaoyi, line 124: What's `mv`?

javascriptlib/src/mill/javascriptlib/PublishModule.scala
```diff
@@ -0,0 +1,539 @@
...
+    for {
+      modSource <- Task.traverse(moduleDeps)(_.sources)
+    } yield modSource
+  }
+
+  private def pubModDepsGenSources: T[Seq[PathRef]] = Task {
+    (for {
+      modSource <- Task.traverse(moduleDeps)(_.generatedSources)()
+    } yield {
+      val fileExt: Path => Boolean = _.ext == "ts"
```
lihaoyi, line 114: Just inline these, the indirection they create just makes the code more confusing

javascriptlib/src/mill/javascriptlib/PublishModule.scala
```diff
@@ -0,0 +1,539 @@
...
+  private def pubModDepsSources: T[Seq[PathRef]] = Task {
+    for {
+      modSource <- Task.traverse(moduleDeps)(_.sources)
+    } yield modSource
+  }
+
+  private def pubModDepsGenSources: T[Seq[PathRef]] = Task {
+    (for {
+      modSource <- Task.traverse(moduleDeps)(_.generatedSources)()
+    } yield {
```
lihaoyi, line 113: `Task.traverse(moduleDeps)(_.generatedSources)().flatMap`

javascriptlib/src/mill/javascriptlib/PublishModule.scala
```diff
@@ -0,0 +1,539 @@
...
+            case _ => s"typescript/$resourceRoot"
+          }
+        )
+        result
+      }
+  }
+
+  private def pubModDeps: T[Seq[String]] = Task {
+    moduleDeps.map { x =>
+      x.millSourcePath.subRelativeTo(Task.workspace).toString.split("/").head
```
lihaoyi, line 100: `x.millSourcePath.subRelativeTo(Task.workspace).segments.head`

lihaoyi: Why are we computing this anyway? It seems wrong. What if the module is nested in a subfolder?

javascriptlib/src/mill/javascriptlib/PublishModule.scala
```diff
@@ -0,0 +1,539 @@
...
+    os.write.over(publishDir().path / "package.json", updatedJson)
+  }
+
+  // Package.Json construction
+
+  // Compilation Options
+  override def modulePaths: Task[Seq[(String, String)]] = Task.Anon {
+    val module = millSourcePath.last
+
+    Seq((s"$module/*", "typescript/src" + ":" + s"${declarationOut()}")) ++
```
lihaoyi, line 84: `s"typescript/src:${declarationOut()}")`

main/package.mill
monyedavid: I no longer need it, was using it to expand glob patterns, will remove

example/javascriptlib/publishing/2-realistic/build.mill
monyedavid: No. `foo` and `foo.bar` are intended to be apart of the same project (mill-realistic),  hence me listing them as moduleDeps in `qux`.

example/javascriptlib/publishing/2-realistic/build.mill
monyedavid: I was trying to demonstrate in the example that is it possible to have a dev (avoid potentially expensive transpile, path resolution process & publish prep. for very large project) and prod version of qux  as its a common practice in js/ts development.

I have removed it from the example.

javascriptlib/src/mill/javascriptlib/PublishModule.scala
monyedavid: couldn't get it to work, an error about symlink already existing kept popping up.
I was unable to resolve it, defaulted back to `os.call`.

build.mill
```diff
@@ -150,6 +150,7 @@ object Deps {
   val commonsIo = ivy"commons-io:commons-io:2.18.0"
   val log4j2Core = ivy"org.apache.logging.log4j:log4j-core:2.24.3"
   val osLib = ivy"com.lihaoyi::os-lib:0.11.4-M4"
+  val sbtIo = ivy"org.scala-sbt::io:1.9.0"
```
He-Pin, line 153: why sbt io?

javascriptlib/src/mill/javascriptlib/PublishModule.scala
monyedavid: - This is not to get rid of `node_modules` prefix, ts/js does not need the prefix to import node-modules, mill needs that. Node cannot resolve node_modules because we install them to `npmInstall.dest`, we have always needed the additional steps of pointing * to `npmInstall.dest` in our tsconfig, so it can resolve the modules, node expects your node_modules to be in the root of the project (or in a parent directory). we use the symlink to fix that.

- the node_modules/ prefix is also something not natural in js/ts development. I don't expect mill adopters for js/ts to prefix all their very many imports with node_module so they can use it.

- the ts-patch is to correctly resolve the ts-paths in our transpiled code, so they can run natively with node

example/javascriptlib/publishing/2-realistic/build.mill
monyedavid: no

javascriptlib/src/mill/javascriptlib/PublishModule.scala
monyedavid: - the mainFileName is from the `TscModule`
- it should be here, we are simply changing ts to js, as we publish js code.
- the user can change the mainFileName as always

javascriptlib/src/mill/javascriptlib/TypeScriptModule.scala
```diff
@@ -158,17 +202,35 @@ trait TypeScriptModule extends Module { outer =>
     combinedCompilerOptions
   }
 
+  // create a symlink for node_modules in compile.dest
+  // removes need for node_modules prefix in import statements `node_modules/<some-package>`
+  // import * as somepackage from "<some-package>"
+  def symlinkNodeModules: Task[Unit] = Task.Anon {
```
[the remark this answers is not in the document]
monyedavid: not really, renamed them.

javascriptlib/src/mill/javascriptlib/PublishModule.scala
monyedavid: move

build.mill
monyedavid: https://github.com/com-lihaoyi/mill/pull/4293#discussion_r1912599211

javascriptlib/src/mill/javascriptlib/PublishModule.scala
lihaoyi: If you want relative paths, why not just `x.millSourcePath.subRelativeTo(Task.workspace).toString`?

javascriptlib/src/mill/javascriptlib/PublishModule.scala
lihaoyi: Let's figure this out and either use `os.symlink` or file a minimized ticket as to why it's not working. `os.*` APIs are there for us to use, and we should use them where possible over subprocesses

javascriptlib/src/mill/javascriptlib/PublishModule.scala
monyedavid: in this case, we get `List(foo, foo/bar)` 
bar exist under foo, I just need `foo` as my target;

if we had a workspace with: List(foo, foo/bar, a, b, c, c/d, e)
I would only need List(foo, a, b, c, e) as the targets to copy

lihaoyi:
I think this looks great, thanks @monyedavid

PR #4294, opened 2025-01-12
Use `-native` Mill binary in Mill's own build

lihaoyi:
Needed some fixes in the bootstrap scripts to prevent the enclosing environment from leaking through. We can then remove the `setup-java` step from all of our workflows, since the Mill native executable will bootstrap the JVM as necessary

We can now run most of github actions jobs without installing Java beforehand, except for the `pre-build` step where `selective.prepare` step which runs on `main` and thus does not have these changes. Once this PR is merged, we can remove `setup-java` from `pre-build` as well


lihaoyi:
disabling the failing scalapb test on windows for now; it appears to be a test infra issue caused by adding `.mill-jvm-opts`, but it shouldn't affect the actual application code

PR #4299, opened 2025-01-13
Use `.local` integration tests in windows CI

lihaoyi:
- Need to move from `temurin:17` to `zulu:17` to get things working on my windows-arm laptop (temurin did not back-publish older JVMs for windows-arm)

- Get rid of the custom `pathingJar` we defined in `dist/package.mill` (using custom path rendering) which doesn't seem to work, in exchange for the default `Jvm.createClasspathPassingJar` (using path-as-url rendering) that does seem to work

- Make `def launcher` make use of `runUseArgsFile` similarly to the various `run` tasks

- Make the integration tests use the `./mill.bat` launcher on windows, rather than `./mill` as a shell script, as the shell script doesn't seem to work properly when used in a windows environment


lihaoyi:
- Need to move from `temurin:17` to `zulu:17` to get things working on my windows-arm laptop (temurin did not back-publish older JVMs for windows-arm)

- Get rid of the custom `pathingJar` we defined in `dist/package.mill` (using custom path rendering) which doesn't seem to work, in exchange for the default `Jvm.createClasspathPassingJar` (using path-as-url rendering) that does seem to work

- Make `def launcher` make use of `runUseArgsFile` similarly to the various `run` tasks

- Make the integration tests use the `./mill.bat` launcher on windows, rather than `./mill` as a shell script, as the shell script doesn't seem to work properly when used in a windows environment

Issue #3141, opened 2024-05-03
Refactor assembly logic into a `AssemblyModule`


lefou:
In the spirit of 
* https://github.com/com-lihaoyi/mill/pull/3064
* https://github.com/com-lihaoyi/mill/pull/3090

We should move the assembly logic into it's own trait, so we can easily provide multiply assembly configurations.

PR #4301, opened 2025-01-13
Break out `AssemblyModule` from `JavaModule`

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/3141

We cannot call `super.assembly` due to some weird bincompat stuff, so instead we make a `private[mill] AssemblyModule#assembly0` which contains the real implementation that both `AssemblyModule#assembly` and `JavaModule#assembly` can call


lihaoyi:
not sure why these changes are causing `methodCodeHashSignatures` to OOM in `test-mill-bootstrap.sh`, will look into it

lihaoyi:
OOMing appears to be a transient, seems like it passed now

PR #4303, opened 2025-01-13
Pretty print artifact on publishing

lolgab:


lihaoyi:
Looks fine, just need to update the example tests i think

lefou (commented):
Would be nice to keep it consistent with `LocalM2Publisher`.

PR #4304, opened 2025-01-13
More consistently propagate environment variables in integration and example tests.

lihaoyi:


lihaoyi:
Previously we copy-pasted it twice between Integration/Example-Tester and still missed a spot

PR #4313, opened 2025-01-13
Part of Android Studio Integration: Fix apk installation failure

vaslabs:
### Motivation 
The following small PR is part of integrating android studio with mill, so it is focused on developer experience. This was already mentioned as a TODO in the androidDebugKeystore method. The screenshot below is with an attempt to hook together android studio and mill and deleting the out directory, then trying to run the bsp test (work pending)

image

So far, the debug keystore, is stored in the task destination directory, which makes it ephemeral, even for a local development device, as out directory can be wiped frequently.

### Advice needed

We followed a similar approach that gradle and android studio use, excluding the lock file. So I'm marking this PR as draft and asking for opinions on:
1. should we handle multi-process access to the debug keystore (i.e. use a file lock)?
2. is it a problem that we break sandbox (we already do with android emulator) and if it is, should we maintain the current functionality for mill sandbox?

### Context

As part of Android studio integration, we are trying to make the bsp integration work for Mill Android module. With some crude adjustments, some functionality starts working (e.g. change bspCompileClasspath to be in sync with compile classpath) and then configure the testTask to work properly with test results etc (work pending). 

The aim is to find small things that are not working and send small PRs instead of an uber PR in the end.

This is co-authored with @irodotos7

### Update

For now, I've kept the keystore in debug destination, but its original location is the global one (so it's copied over), so the persistent is still working, but now it is also idempotent (assuming the original file in global position is not changed) . Not sure if this is actually needed, I think just having the global dest is much simpler but let's see how it goes and how this evolves

### DIFF
--- a/scalalib/src/mill/javalib/android/AndroidAppModule.scala
+++ b/scalalib/src/mill/javalib/android/AndroidAppModule.scala
@@ -897,10 +897,12 @@ trait AndroidAppModule extends JavaModule {
   private def androidDebugKeystore: Task[PathRef] = Task(persistent = true) {
     val debugFileName = "mill-debug.jks"
     val globalDebugFileLocation = os.home / ".android"
+
+
     if (!os.exists(globalDebugFileLocation)) {
       os.makeDir(globalDebugFileLocation)
     }
-    // TODO maybe we need a file lock here
+
     val debugKeystoreFile = os.home / ".android" / debugFileName
 
     if (!os.exists(debugKeystoreFile)) {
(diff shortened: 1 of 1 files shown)


lihaoyi:
The code looks fine. Test failures do seem related so probably should be investigated before merging

lihaoyi:
Looks good to me @vaslabs, will merge it if you're done

vaslabs:
@0xnm thanks for your comments, I will note them and follow up in subsequent PRs, I agree with them, just wanted someone to feedback before committing to specific implementations, these are very useful!

0xnm:
Sure, no problem. Feel free to tag me on the PRs if you want.

PR #4314, opened 2025-01-13
WIP: Generation Software Bill of Materials (SBOM)

gamlerhart:
Motivation: Often larger projects need a
Software Bill of Materials (SBOM) to track dependencies across their projects.
So the project can track things like vulnerable
components, licenses etc.

- Refactor Courier support to return the original dependency as well.
- Start of support for CycloneDX format

This is more a QnA pull request.


example/javalib/basic/1-simple/build.mill
```diff
@@ -2,7 +2,7 @@
 package build
 import mill._, javalib._
 
-object foo extends JavaModule {
+object foo extends JavaModule with javalib.sbom.CycloneDXModule {
```
gamlerhart, line 5: Just testing ad hoc testing, ignore in this draft.

main/util/src/mill/util/CoursierSupport.scala
```diff
@@ -106,10 +106,10 @@ trait CoursierSupport {
         case Right(res) =>
           Result.Success(
             Agg.from(
-              res.files
-                .map(os.Path(_))
-                .filter(resolveFilter)
-                .map(PathRef(_, quick = true))
+              res.fullDetailedArtifacts
```
gamlerhart, line 109: This is the root of the change: Not only return the path, but also what dependency.

Drawbacks:
- Will this break plugins. I'm not quite sure how backwards compatible these methods need to be?
- `.map(_._2)` in many locations, making the code more mysterious. Better use a `case class`. Other ideas?

Generally: Wondering if this is the right approach to return more info about the resolved dependencies. Or if I missed a obvious other way?

scalalib/src/mill/javalib/sbom/CycloneDX.scala
```diff
@@ -0,0 +1,114 @@
+package mill.javalib.sbom
```
gamlerhart, line 1: Not sure where to place this. Probably in the `contrib` area? 

Especially since I think it will need incubation time to get stable.

scalalib/src/mill/javalib/sbom/CycloneDX.scala
```diff
@@ -0,0 +1,114 @@
...
+import upickle.default.macroRW
+
+import java.math.BigInteger
+import java.nio.channels.FileChannel
+import java.nio.charset.StandardCharsets
+import java.security.MessageDigest
+import java.time.Instant
+import java.util.{Base64, UUID}
+
+object CycloneDXModule {
```
gamlerhart, line 17: Ignore the Module: Needs a bit more work =).

gamlerhart (commented):
Added my questions as code review annotations.

scalalib/src/mill/javalib/sbom/CycloneDX.scala
gamlerhart: Moved to contrib.

main/util/src/mill/util/CoursierSupport.scala
lihaoyi: I think in the past we would add a new task that returned the richer data type (e.g. a `case class` or something), and the old task would just call the new task and extract the information it needs. Then whoever needs the richer data type can call the new task

gamlerhart: Will do that. The only concern is that there are many 'flavors' of resolveDependencies. (CoursierSupport, Lib, JVM)
Anyway, can be cleared once the PR isn't draft anymore.

gamlerhart:
Closing this: I'll create a new PR with the latest approach once its ready.

PR #4340, opened 2025-01-16
Blog post: How to Compile Java into Native Binaries with Mill and Graal

lihaoyi:
This needs to wait until Graal support in Mill is released in 0.12.6


blog/modules/ROOT/pages/7-graal-native-executables.adoc
```diff
@@ -0,0 +1,475 @@
...
+```
+
+The column `373M` and `20M` are the respective memory footprints of the executable assembly
+and native image binary. In this small program, the native image uses almost 20x less memory
+than the JVM executable assembly! That is a very significant reduction in resource footprint
+
+### Portability and Hermeticity
+
+Executable assembly jars require a JVM installed globally in order to run. In a way they are
+not hermetic, since the globally-installed JVM can differ resulting in the assembly behaving
```
gamlerhart, line 393: Sentence is odd/hard to read with two words for 'difference' and passive voice.

My slight improvement, making it two sentences and reducing passive voice, so that it has less going on

`... not hermitic. A different globally-installed JVM can result in the assembly behaving different at runtime`.

PR #4362, opened 2025-01-17
Android Studio Integration: Attempt via bsp for Java android apps

vaslabs:
## This PR provides

- ability to import a mill android project to android studio via BSP, including sources and generated sources.
- necessary changes to the way sources are resolved (through module dependency instead of class inheritance).
- Changes to tests (e.g. number of sources compiled) reflect these changes

### Not provided

- Compilation still needs to be done from the command line in order for the paths to be resolved inside the IDE.

Note that to make compilation, running, testing and debugging work from android studio, we'll need to develop a plugin for it

## Note

This is an early draft and I need to cleanup or review a few rough edges, I submit this as a draft to get some early feedback. 

I've implemented java to get the basics to work, as kotlin has multiple kinds of sources and files that need to be taken care of, so this is just a small step.

Don't hesitate to point out any issues as I've really just braced through and hacked around to get this to work!

@0xnm @lihaoyi  any feedback will be appreciated!

## Android Studio Integration

The attempt focuses on BSP. Although bsp as a plugin does not seem to be available for android studio, it strangely becomes available if it is installed through intelliJ (see video)

### Basic changes

- Use the moduleDeps instead of resolving the classpath bits (sources, resources) through inheritance. This makes BSP and studio integration work (see the directories that are highlighted correctly)
- Crude implementation of bsp methods to get the IDE to report the imported modules correctly

### Tricky parts

- Resources are compiled twice due to how aapt linking and later the module deps compile hierarchy. In order to avoid this, I've added an empty generatedSources method in android instrumentation module . 

### Demo

You can see in the demo most of the IDE static import features work (R is not recognised but I'll look into it either in this PR or subsequent).

https://github.com/user-attachments/assets/77fe5e52-57e7-4480-9971-b50cc8d4d839

EDIT: I've fixed the R not being recognised and the kotlin import

image

EDIT 2: Added summary at the top

### DIFF
--- a/scalalib/src/mill/javalib/android/AndroidAppModule.scala
+++ b/scalalib/src/mill/javalib/android/AndroidAppModule.scala
@@ -422,7 +422,7 @@ trait AndroidAppModule extends JavaModule {
    * the Android resource generation step.
    */
   override def generatedSources: T[Seq[PathRef]] = Task {
-    androidLibsRClasses()
+    Seq(PathRef(androidResources()._1.path / rClassDirName)) ++ androidLibsRClasses()
   }
 
   /**
--- a/scalalib/src/mill/javalib/android/AndroidAppModule.scala
+++ b/scalalib/src/mill/javalib/android/AndroidAppModule.scala
@@ -3,8 +3,9 @@ package mill.javalib.android
 import coursier.Repository
 import mill._
 import mill.scalalib._
-import mill.api.{Logger, PathRef}
-import mill.define.ModuleRef
+import mill.api.{Logger, PathRef, internal}
+import mill.define.{ModuleRef, Task}
+import mill.scalalib.bsp.BspBuildTarget
 import mill.testrunner.TestResult
 import mill.util.Jvm
 import os.RelPath
@@ -266,8 +267,19 @@ trait AndroidAppModule extends JavaModule {
    */
   override def resources: T[Seq[PathRef]] = Task {
     val libResFolders = androidUnpackArchives().flatMap(_.resources)
-    super.resources() ++ libResFolders :+ PathRef(millSourcePath / "src/main/res")
+    libResFolders :+ PathRef(millSourcePath / "src/main/res")
   }
+
+  @internal
+  override def bspCompileClasspath: T[Agg[UnresolvedPath]] = Task {
+    compileClasspath().map(_.path).map(UnresolvedPath.ResolvedPath(_))
+  }
+
+  @internal
+  override def bspBuildTarget: BspBuildTarget = super.bspBuildTarget.copy(
+    baseDirectory = Some(millSourcePath / "src/main"),
+    tags = Seq("application"),
+  )
 
   /**
    * Replaces AAR files in classpath with their extracted JARs.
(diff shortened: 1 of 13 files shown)


vaslabs:
Saw [this](https://github.com/com-lihaoyi/mill/pull/4363) and I'm wondering if I should also give it a go with a maven import and maybe test the conversion of android apps to mill

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -816,6 +816,7 @@ trait JavaModule
    * Keep in sync with [[bspCompileClassesPath]]
    */
   def compile: T[mill.scalalib.api.CompilationResult] = Task(persistent = true) {
+    println(s"Compile with all source files being ${allSourceFiles()}")
```
vaslabs, line 819: Note: remove

lihaoyi:
> Saw [this](https://github.com/com-lihaoyi/mill/pull/4363) and I'm wondering if I should also give it a go with a maven import and maybe test the conversion of android apps to mill

Give it a try! I expect it will not work out of the box, but maybe it can be made to work with some effort

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -1007,15 +1045,23 @@ trait AndroidAppModule extends JavaModule {
   trait AndroidAppTests extends JavaTests {
     private def testPath = parent.millSourcePath / "src/test"
 
-    override def sources: T[Seq[PathRef]] = parent.sources() :+ PathRef(testPath / "java")
+    override def sources: T[Seq[PathRef]] = Seq(PathRef(testPath / "java"))
+
+    override def resources: T[Seq[PathRef]] = Task.Sources(Seq(PathRef(testPath / "res")))
```
0xnm, line 1050: similar update needs to be done in the Kotlin-related module

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -1030,11 +1076,15 @@ trait AndroidAppModule extends JavaModule {
 
     override def androidEmulatorPort: String = parent.androidEmulatorPort
 
-    override def sources: T[Seq[PathRef]] = parent.sources() :+ PathRef(androidTestPath / "java")
+    override def sources: T[Seq[PathRef]] = Seq(PathRef(androidTestPath / "java"))
```
0xnm, line 1079: This is a good change, but still, the way instrumented testing is implemented right now is different from the way AGP does it.

The current approach is to generate a single fat APK, consisting of production + test files.

AGP is doing it in a different way: it will be 2 APKs, one is a normal production APK, another one will contain _only_ test files and will also have a dedicated manifest obviously, with a dedicated package name attribute which will be `${main app package name}.test`. And this test APK should be installed with `-t` option (see [here](https://developer.android.com/tools/adb#-t-option)).

vaslabs: yes, you are right, I have decompiled the APKs from AGP as well and can work towards this if we want to have the exact same behaviour (which makes sense to have)

image

I want to set up a scaffold with some tests and examples working first and then do the underlying implementations on the next stage.

vaslabs:
> > Saw [this](https://github.com/com-lihaoyi/mill/pull/4363) and I'm wondering if I should also give it a go with a maven import and maybe test the conversion of android apps to mill
> 
> Give it a try! I expect it will not work out of the box, but maybe it can be made to work with some effort

will do after I give a closer look into it, this PR should be ready soon (once everything passes)

vaslabs:
getting a failure [on this job](https://github.com/vaslabs-ltd/mill/actions/runs/12869583654/job/35884130539#step:5:2944), so I haven't marked it as ready yet, but I don't think it's related to my changes

lihaoyi:
@vaslabs might be flakiness, feel free to ignore for now

vaslabs:
I'm done with the changes within the scope of this PR

lihaoyi:
Thanks @vaslabs and @0xnm for the review, looks good to me

PR #4365, opened 2025-01-18
Fix version typos in JVM versions example

NyuB:
Zinc modules' names were referring to a different java version than the one they actually defined

**Suggestion**: drop the version number in these names ?


lihaoyi:
@NyuB happy to drop the version numbers, that will help avoid them falling out of sync in future

PR #4376, opened 2025-01-20
Try to fix mima issues

lihaoyi:


lolgab:
Should we have TODOs to revert these changes when we are in binary breaking window?

PR #4380, opened 2025-01-20
Address bin compat issues introduced in #4145

alexarchambault:


lihaoyi:
Should be fixed in https://github.com/com-lihaoyi/mill/pull/4376

Issue #4378, opened 2025-01-20
Fix `PublishModule.publishM2Local` and rename `PublishModule.getM2LocalRepoPath`


lefou:
The default repo path is now `./<moudule>.getM2LocalRepoPath`, which is wrong.

The newly introduced target `getM2LocalRepoPath` should be named `m2LocalRepoPath` before the next release.

See
* https://github.com/com-lihaoyi/mill/pull/4270

PR #4385, opened 2025-01-21
Fix publishM2Local

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/4378

### DIFF
--- a/scalalib/src/mill/scalalib/PublishModule.scala
+++ b/scalalib/src/mill/scalalib/PublishModule.scala
@@ -327,12 +327,10 @@ trait PublishModule extends JavaModule { outer =>
    *                   If not set, falls back to `maven.repo.local` system property or `~/.m2/repository`
    * @return [[PathRef]]s to published files.
    */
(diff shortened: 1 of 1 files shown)


scalalib/src/mill/scalalib/PublishModule.scala
```diff
@@ -327,22 +327,20 @@ trait PublishModule extends JavaModule { outer =>
...
    * Publish artifacts to the local Maven repository.
    * @return [[PathRef]]s to published files.
    */
   def publishM2LocalCached: T[Seq[PathRef]] = Task {
-    publishM2LocalTask(getM2LocalRepoPath)()
+    publishM2LocalTask(publishM2LocalRepoPath)()
   }
 
-  def getM2LocalRepoPath: Task[os.Path] = Task.Input {
+  def publishM2LocalRepoPath: Task[os.Path] = Task.Input {
```
lefou, line 343: We should add some docs here, e.g. copy from `publishM2Local`.

lefou (approved):

Issue #4373, opened 2025-01-20
Stop publishing Github release artifacts that have been migrated to Maven Central


lihaoyi:
Follow up of https://github.com/com-lihaoyi/mill/pull/4372

lefou:
I'd love if we could update the Github Release page document so that it updates a list of  snapshot versions with every snapshot release.

PR #4386, opened 2025-01-21
Stop publishing artifacts to github releases since move to maven central

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/4373

This only affects people running clicking links or running commands manually to download the bootstrap scripts or example projects, or manually running `./mill init`, so it won't break any automated releases.

This lets us substantially simplify the github actions workflows and build config


nafg:
This broke Mise

nafg:
and `asdf`

PR #4391, opened 2025-01-24
Segments.last: Fix error message

nafg:


lihaoyi:
Thank @nafg !

lefou:
Can that be right? We need segments to start with a label, but ending with a cross is perfectly fine.

lefou:
Looks like `Segments.last` was only introduced recently in #4132. We should review whether it really needs to be a `Label`, or if it being a `Cross` would be ok too.

PR #4363, opened 2025-01-17
Added support for 'mill init' from existing Gradle project

ajaychandran:
Resolves #3962 by extending `mill init` with the ability to import a Gradle project.
* handles deeply nested modules
* configures dependencies for configurations:
  *  implementation / api
  * compileOnly / compileOnlyApi
  * runtimeOnly
  * testImplementation
  * testCompileOnly
* configures testing frameworks
* supports Gradle plugins
  * [java](https://docs.gradle.org/current/userguide/java_plugin.html)
  * [maven-publish](https://docs.gradle.org/current/userguide/publishing_maven.html)

Additionally, Maven support added in #3756 was improved.
* shared logic moved out to module `buildgen`
* added support for BOM dependencies
* eliminated creation of redundant test modules
* fixed `millSourcePath` for test module

PR #4404, opened 2025-01-25
Fix a possible typo in "maven.adoc"

ShreckYe:
This is from reviewing #4363. I couldn't comment on this because it was not part of the changes.


lihaoyi:
Thanks @ShreckYe !

PR #4410, opened 2025-01-27
Fix selective testing of native example tests by adding jvm version override

lihaoyi:
Minimized in https://github.com/com-lihaoyi/mill/pull/4357, it appears the difference in behavior comes from different JVM versions being used during `selective.prepare` and `selective.{resolve,run}` phases


lihaoyi:
Seems to have worked, as the `.native` targets are no longer being spuriously selected by `selective.resolve` https://github.com/com-lihaoyi/mill/actions/runs/13001357943/job/36260711097

PR #4411, opened 2025-01-27
Part of Android Support: screenshot testing

vaslabs:
# Background

Screenshot testing is a way to test UIs and in Android's case, typically without the need of running an emulator. This is achieved in recent developments with the usage of Compose and Previews (the same functionality that generates previews in android studio).

The typical flow is:

1. A base line is generated, with screenshots being committed into the application's repo
2. Regression tests subsequently, compare the screenshots
3. For new UIs that devs decide to add coverage or agreed changes, baseline is updated

Further Reading: https://developer.android.com/studio/preview/compose-screenshot-testing

# Provided in this PR

This PR, while it has in mind the full flow support, only provides the means to generate screenshots and run diffs with the same tools that AGP is using. At the same time, we keep the test task compatibility with Mill so we don't need to implement any custom runner tasks.

## Generating screenshots

Using the same tool as AGP, the [compose-preview-renderer](https://android.googlesource.com/platform/tools/base/+/61923408e5f7dc20f0840844597f9dde17453a0f/preview/screenshot/screenshot-test-gradle-plugin/src/main/java/com/android/compose/screenshot/tasks/PreviewRenderWorkAction.kt), which has undocumented cli arguments. However, these can be found by running the respective gradle task in a sample Android studio project

```bash
./gradlew  updateDebugScreenshotTest
```

Then, given the generated data, the PR worked towards generating the file computing classpaths and other needed files similarly

CLI arguments generated by Gradle for the preview tool

Currently these screenshots are ephemeral, and this PR is not concerned by adding support for version control on the screenshot base. The plan is for a future PR to make it easier for the users to customise the behaviour (e.g. choose the path)

## Testing screenshots

AGP uses the [screenshot-validation-junit-engine](https://android.googlesource.com/platform/tools/base/+/61923408e5f7dc20f0840844597f9dde17453a0f/preview/screenshot/screenshot-validation-junit-engine). This PR provides similar functionality but hooking into Mill's testing infrastructure instead of Gradle.

## Example App

The example App was set up initially with AGP and then a mill version was constructed with 2 differences:

1. Only 1 screen configuration is provided for the simplicity of this first attempt
2. Although compose preview is technically working, it is only kept in the screenTest module, again for the simplicity of this PR and not doing any changes to the compose support in the main android kotlin module yet. 

# Future developments / Not yet implemented in this PR

- Baseline management
- Updating previews (this can be done already, but with a crude overwrite)
- More examples and more screen sizes and UI configurations
- Discover classes and methods automatically


vaslabs:
there is a failure in the javalib native examples, but I don't think it's related to my changes.

kotlinlib/src/mill/kotlinlib/android/ComposeRenderer.scala
```diff
@@ -0,0 +1,57 @@
+package mill.kotlinlib.android
+
+import mill.api.internal
+
+private[android] object ComposeRenderer {
```
lihaoyi, line 5: Are these data structures written based on any external schema? If so we should add a link in the comment

example/android/kotlinlib/3-compose-screenshot-tests/app/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
```
lihaoyi, line 1: Is the app sources in `example/android/kotlinlib/3-compose-screenshot-tests/app/` based on any existing template or project generator? If so we should leave a comment mentioning that somewhere

lihaoyi:
I think the code looks fine, just some comments about the provenance of the code to better understand why it is the way it is.

For some reason Github is not letting me resolve the branch conflict, so you may have to tick the checkbox in the PR to let maintainers update it

kotlinlib/src/mill/kotlinlib/android/ComposeRenderer.scala
vaslabs: thanks! yes, added link!

example/android/kotlinlib/3-compose-screenshot-tests/app/src/main/AndroidManifest.xml
vaslabs: thanks! I've just pushed the sample AGP project in github and linked it here

vaslabs:
> I think the code looks fine, just some comments about the provenance of the code to better understand why it is the way it is.
> 
> For some reason Github is not letting me resolve the branch conflict, so you may have to tick the checkbox in the PR to let maintainers update it

thanks, there were conflicts with the main branch, T.log => Task.log, so I changed that bit of code on this PR as well to your recommendation

PR #4439, opened 2025-01-31
Add coverage module for python

jodersky:
The actual analysis is delegated to the well-known coverage.py package.

### DIFF
--- a/example/pythonlib/linting/3-coverage/build.mill
+++ b/example/pythonlib/linting/3-coverage/build.mill
@@ -1,0 +1,44 @@
+// Mill's support for code coverage analysis is implemented by
+// https://coverage.readthedocs.io/[the coverage.py package].
+//
+// You can use it by extending `CoverageTests` in your test module.
+
+import mill._, pythonlib._
+
+object `package` extends RootModule with PythonModule {
+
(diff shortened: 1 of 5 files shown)


example/pythonlib/linting/3-coverage/build.mill
```diff
@@ -0,0 +1,44 @@
+// Mill's support for code coverage analysis is implemented by
+// https://coverage.readthedocs.io/[the coverage.py package].
+//
+// You can use it by extending `CoverageTests` in your test module.
+
+import mill._, pythonlib._
+
+object `package` extends RootModule with PythonModule {
```
lihaoyi, line 8: Shouldn't this need to inherit from `CoverageModule` in order for `CoverageTests` to be available?

lihaoyi: Ah I misread it, `CoverageTests` is a sibling trait, not a nested one

lihaoyi:
I think this looks reasonable, @jodersky if you're done with it I'm happy to merge it

jodersky:
I'm happy with the current state, so feel free to go ahead :)

Issue #4433, opened 2025-01-30
Add Http4s TodoMVC Example (500USD bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 500USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

These should be example tests, similar to the exist one in `example/scalalib/web/1-todo-webapp`, but instead named `example/scalalib/web/9-todo-http4s`. It should have similar documentation/explanations, automated `Usage` test commands, and manual testing to verify the functionality works.

We should be able to reuse most of the existing frontend CsS and JS without changes. What is of interest is the server-side code and build configuration for the project

PR #4446, opened 2025-02-01
Add Http4s TodoMVC Example

armanbilge:
Closes https://github.com/com-lihaoyi/mill/issues/4433

### DIFF
--- a/docs/modules/ROOT/pages/scalalib/web-examples.adoc
+++ b/docs/modules/ROOT/pages/scalalib/web-examples.adoc
@@ -40,3 +40,7 @@ include::partial$example/scalalib/web/7-cross-platform-version-publishing.adoc[]
 == Scala.js WebAssembly Example
 
 include::partial$example/scalalib/web/8-wasm.adoc[]
+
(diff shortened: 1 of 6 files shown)


lihaoyi:
Thanks @armanbilge. Some notes:

* Trying to add a new todo and pressing Enter seems to make the whole UI disappear
* Can you use port 8084, to avoid conflict with the 1-webapp example that already uses 8080?

armanbilge:
> Trying to add a new todo and pressing Enter seems to make the whole UI disappear

Sorry, I stepped on a footgun 😅 (https://github.com/http4s/http4s/issues/7054#issuecomment-1502546233). I realized none of the JSON/Circe stuff is used, so I removed it.

> Can you use port 8084, to avoid conflict with the 1-webapp example that already uses 8080?

👍

lihaoyi:
Is the `.debug` call still necessary?

lihaoyi:
going to close thtis for now. @armanbilge can email me your international bank transfer details and i will close out the bounty

PR #4456, opened 2025-02-02
Consolidate jvm subprocess and inprocess functions

### LINKED ISSUE #3772: Clean up `Jvm.*` subprocess/inprocess APIs (500USD Bounty)

sake92:
Changes:
- added functions based on #3772 description
  - the process related ones now return os-lib process result or subprocess handle
  - `callClassLoader` uses a loan pattern function and closes the classloader after usage
- added `mill.util.ProcessUtil.toResult(processResult)` to handle process result in same way https://github.com/com-lihaoyi/mill/blob/main/main/util/src/mill/util/Jvm.scala#L352-L357

Some minor questions/suggestions:
1. current `spawnClassloader` is more flexible, you can pass your own classloader rather that use `getClass.getClassLoader`, this is ok I guess?
1. suggestion to name `spawnClassLoader` as `createClassLoader`
1. suggestion to name `callClassloader` as `withClassloader` (common loan pattern name)
1. `getMainMethod` is not public so `runLocal` can't be removed.. so expose `getMainMethod` or leave `runLocal` as is?
1. Some calls are inside build.mill files.  
    I have to wait for new version of Mill with these changes included, and then update them?
1. Not sure if `closeClassLoaderWhenDone` is needed, I don't understand why it is left unclosed in few places.

Closes #3772


lihaoyi:
The naming is pulled a bit in two directions

1. On one hand, we want classloader methods  to be consistent with the equivalent subprocess methods, which are named `spawn` and `call`
2. On the other hand, we want classloader methods to be consistent with the equivalent Scala patterns, which call them `create` and `with`

TBH I'm not sure which one is better. @lefou @lolgab not sure if you guys have any opinions here?

lihaoyi:
> current spawnClassloader is more flexible, you can pass your own classloader rather that use getClass.getClassLoader, this is ok I guess?

This is to let you select the parent classloader; at the minimum it lets you decide whether or not to include the current Mill classloader on the child classpath, or have the child classpath entirely isolated. So we probably need to keep it as necessary complexity in the API

lihaoyi:
> Some calls are inside build.mill files.
I have to wait for new version of Mill with these changes included, and then update them?

That's right. For now can just leave them in place

lihaoyi:
> getMainMethod is not public so runLocal can't be removed.. so expose getMainMethod or leave runLocal as is?

Seems that `Jvm.runLocal` is only called in one place in `RunModule.scala`, so maybe we can move `getMainMethod` there?

lihaoyi:
> Not sure if closeClassLoaderWhenDone is needed, I don't understand why it is left unclosed in few places.

Maybe we can try closing it all the time and see if any tests break? I don't understand why we wouldn't close it either haha

lihaoyi:
The name `Jvm` is also somewhat generic, could we come up with a more specific name? e.g. maybe `Classpath.withClassLoader`/`Classpath.spawnSubprocess`?

sake92:
> > getMainMethod is not public so runLocal can't be removed.. so expose getMainMethod or leave runLocal as is?
> 
> Seems that `Jvm.runLocal` is only called in one place in `RunModule.scala`, so maybe we can move `getMainMethod` there?

Good point.  
Copied it to `RunModule`, we can remove the one from `Jvm` when we remove these deprecated functions.

sake92:
@lihaoyi this is ready for a first pass.  
Only the android tests are failing, but they fail in main branch too.  
Have to admit, this was a bit hairy.. lots of nulls, ifs, getorelses and defaults to figure out 😄

lihaoyi:
Thanks @sake92 ! will take a look

main/util/src/mill/util/Jvm.scala
```diff
@@ -14,10 +14,173 @@ import os.CommandResult
...
+   * @param mainArgs Args passed to the `mainClass` main method
+   * @param env Environment variables used when starting the forked JVM
+   * @param propagateEnv If `true` then the current process' environment variables are propagated to subprocess
+   * @param cwd The working directory to be used by the forked JVM
+   * @param stdin Standard input override
+   * @param stdout Standard output override
+   * @param stderr Standard error override
+   * @param mergeErrIntoOut If `true` then the error output is merged into standard output
+   */
+  def spawn(
```
lihaoyi, line 123: Let's call it `spawnSubprocess`

main/util/src/mill/util/Jvm.scala
```diff
@@ -14,10 +14,173 @@ import os.CommandResult
...
+      mainArgs: Seq[String] = Seq.empty,
+      env: Map[String, String] = Map.empty,
+      propagateEnv: Boolean = true,
+      cwd: os.Path = null,
+      stdin: os.ProcessInput = os.Inherit,
+      stdout: ProcessOutput = os.Inherit,
+      stderr: ProcessOutput = os.Inherit,
+      mergeErrIntoOut: Boolean = false,
+      destroyOnExit: Boolean = true
+  )(implicit ctx: Ctx): CommandResult = {
```
lihaoyi, line 58: Let's get rid of the `ctx: Ctx` parameter here and below in `spawn`, and also get rid of the `ctx.log.debug` calls. The occasional benefit of turning on debug logging probably isn't worth the complexity this adds to the signature and calling context

main/util/src/mill/util/Jvm.scala
```diff
@@ -454,6 +630,23 @@ object Jvm extends CoursierSupport {
     )(new Ctx.Home { override def home = os.home })
   }
 
+  def callClassLoader[T](
```
lihaoyi, line 633: Let's call this `withClassLoader` as you suggested

main/util/src/mill/util/Jvm.scala
```diff
@@ -454,6 +630,23 @@ object Jvm extends CoursierSupport {
...
+      classPath: Iterable[os.Path],
+      sharedPrefixes: Seq[String] = Seq.empty,
+      parent: ClassLoader = null
+  )(f: ClassLoader => T)(implicit ctx: mill.api.Ctx.Home): T = {
+    val oldClassloader = Thread.currentThread().getContextClassLoader
+    val newClassloader = spawnClassloader(classPath, sharedPrefixes, parent)
```
lihaoyi, line 639: Let's define a `Jvm.createClassloader` and deprecate `mill.api.ClassLoader.create`. This will help make it discoverable, since it'll be defined in the same place as all the other  classloader/subprocess methods in `Jvm.*`.

main/util/src/mill/util/ProcessUtil.scala
```diff
@@ -0,0 +1,15 @@
+package mill.util
+
+import mill.api.Result
+
+object ProcessUtil {
+
+  def toResult(processResult: os.CommandResult): Result[Int] = {
```
lihaoyi, line 7: If we make `Jvm.callSubprocess` take `check: Boolean = true`, then we can just rely on the underlying `os.call` throwing an exception and won't need to call this `ProcessUtil.toResult` everywhere

lihaoyi:
@sake92 took a first pass at reviewing this. Looks great!

main/util/src/mill/util/Jvm.scala
```diff
@@ -14,10 +14,173 @@ import os.CommandResult
...
+      useCpPassingJar: Boolean = false,
+      mainArgs: Seq[String] = Seq.empty,
+      env: Map[String, String] = Map.empty,
+      propagateEnv: Boolean = true,
+      cwd: os.Path = null,
+      stdin: os.ProcessInput = os.Inherit,
+      stdout: ProcessOutput = os.Inherit,
+      stderr: ProcessOutput = os.Inherit,
+      mergeErrIntoOut: Boolean = false,
+      destroyOnExit: Boolean = true
```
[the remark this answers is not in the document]
sake92: Do you want to match the defaults too?  
For example `stdout: ProcessOutput = os.Inherit` is mostly used in mill codebase, but in os-lib it is 
```
stdin: ProcessInput = Pipe,
stdout: ProcessOutput = Pipe,
stderr: ProcessOutput = os.Inherit,
```

I am fine with either.

sake92:
Hmm, I found this comment about leaving classloader open.  
https://github.com/com-lihaoyi/mill/blob/1c100f7acfea86687006fa885ee47781163e7dac/testrunner/src/mill/testrunner/TestRunner.scala#L17

But the tests are passing, so should be fine..? 🤷🏻‍♂️

main/util/src/mill/util/Jvm.scala
sake92: Ok, np.  
Should I leave deprecation annotation "since 0.12.7" as is now? `@deprecated("Use callClassLoader", "Mill 0.12.7")`  
Or put `0.13.0-M1` or?

lihaoyi: @sake92 let's say deprecated `since 0.12.8`, with the goal of backporting this PR from `main` back to `0.12.x` so that it can go out in the next `0.12.8` release

lihaoyi:
@sake92 yeah I dunno, that config is from 2018 so the code has evolved significantly since then. Let's just close the classloaders everywhere, and if tests pass we can ship it. If there are new issues we can deal with them when they arrive

main/util/src/mill/util/Jvm.scala
```diff
@@ -14,10 +14,173 @@ import os.CommandResult
...
+   * @param mainArgs Args passed to the `mainClass` main method
+   * @param env Environment variables used when starting the forked JVM
+   * @param propagateEnv If `true` then the current process' environment variables are propagated to subprocess
+   * @param cwd The working directory to be used by the forked JVM
+   * @param stdin Standard input
+   * @param stdout Standard output
+   * @param stderr Standard error
+   * @param mergeErrIntoOut If `true` then the error output is merged into standard output
+   */
+  def call(
```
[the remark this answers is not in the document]
sake92:
```
[17623] [error] -- Error: D:\projects\forks\mill\main\util\src\mill\util\Jvm.scala:184:6 -------
[17623] [error] 184 |  def callSubprocess(
[17623] [error]     |      ^
[17623] [error]     |two or more overloaded variants of method callSubprocess have default arguments
[17623] [error] -- Error: D:\projects\forks\mill\main\util\src\mill\util\Jvm.scala:538:6 -------
[17623] [error] 538 |  def spawnSubprocess(
[17623] [error]     |      ^
[17623] [error]     |two or more overloaded variants of method spawnSubprocess have default arguments
```

Seems like we can't have both the deprecated one and this new one..  
We can't remove these deprecated ones right now?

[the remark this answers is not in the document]
lihaoyi: How about we tweak the name to `callProcess` and `spawnProcess`? That should avoid the collision right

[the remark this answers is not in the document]
sake92: Yeah, or I can delegate to the new function?  
Not sure if it will break binary/source compatibility..?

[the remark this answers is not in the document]
lihaoyi: Using a new name and leaving the old one in place should preserve compatibility

sake92:
In the last commit I only changed a comment.  
But the previous commit tests worked fine..

main/util/src/mill/util/Jvm.scala
```diff
@@ -454,6 +627,44 @@ object Jvm extends CoursierSupport {
...
+          sharedLoader.loadClass(name)
+        } else super.findClass(name)
+      }
+    }
+
+  def withClassLoader[T](
+      classPath: Iterable[os.Path],
+      sharedPrefixes: Seq[String] = Seq.empty,
+      parent: ClassLoader = null
+  )(f: ClassLoader => T)(implicit ctx: mill.api.Ctx.Home): T = {
```
lihaoyi, line 655: We can get rid of the `ctx: mill.api.Ctx.Home` implicit here

main/util/src/mill/util/Jvm.scala
```diff
@@ -454,6 +627,44 @@ object Jvm extends CoursierSupport {
     )(new Ctx.Home { override def home = os.home })
   }
 
+  def createClassLoader(
+      classPath: Iterable[os.Path],
+      sharedPrefixes: Iterable[String] = Seq(),
+      parent: java.lang.ClassLoader = null,
+      sharedLoader: java.lang.ClassLoader = getClass.getClassLoader,
+      logger: Option[mill.api.Logger] = None
```
lihaoyi, line 635: Let's drop this `logger` parameter

main/util/src/mill/util/Jvm.scala
```diff
@@ -454,6 +627,44 @@ object Jvm extends CoursierSupport {
     )(new Ctx.Home { override def home = os.home })
   }
 
+  def createClassLoader(
+      classPath: Iterable[os.Path],
+      sharedPrefixes: Iterable[String] = Seq(),
+      parent: java.lang.ClassLoader = null,
```
lihaoyi, line 633: Lets move `sharedPrefixes` next to `sharedLoader` so it's more obvious they're related

main/util/src/mill/util/Jvm.scala
```diff
@@ -454,6 +627,44 @@ object Jvm extends CoursierSupport {
     )(new Ctx.Home { override def home = os.home })
   }
 
+  def createClassLoader(
```
lihaoyi, line 630: Let's add a short Scaladoc here describing what it does, with docs for the various parameters

scalalib/test/src/mill/scalalib/AssemblyTestUtils.scala
```diff
@@ -52,16 +52,23 @@ trait AssemblyTestUtils {
...
-        commandArgs = Seq(file.toString(), "--text", "tutu"),
-        envArgs = Map.empty[String, String],
-        workingDir = wd
+      os.call(
+        cmd = Seq(file.toString(), "--text", "tutu"),
+        env = Map.empty[String, String],
+        cwd = wd,
+        stdin = os.Inherit,
+        stdout = os.Inherit,
+        stderr = os.Inherit
```
lihaoyi, line 71: No need `stderr = os.Inherit` since it's already the default (above as well)

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -1165,10 +1165,13 @@ trait JavaModule
...
-        commandArgs = Seq(Jvm.jdkTool("javadoc")) ++ cmdArgs,
-        envArgs = Map(),
-        workingDir = Task.dest
+      os.call(
+        cmd = Seq(Jvm.jdkTool("javadoc")) ++ cmdArgs,
+        env = Map(),
+        cwd = Task.dest,
+        stdin = os.Inherit,
+        stdout = os.Inherit,
+        stderr = os.Inherit
```
lihaoyi, line 1174: No need `stderr = os.Inherit`

pythonlib/src/mill/pythonlib/PythonModule.scala
```diff
@@ -261,11 +266,16 @@ object PythonModule {
...
-        envArgs = Option(env).getOrElse(env0),
-        workingDir = Option(workingDir).getOrElse(workingDir0)
+    )(implicit ctx: Ctx): Unit = {
+      os.call(
+        cmd = Seq(Option(command).getOrElse(command0)) ++ options ++ args.value,
+        env = Option(env).getOrElse(env0),
+        cwd = Option(workingDir).getOrElse(workingDir0),
+        stdin = os.Inherit,
+        stdout = os.Inherit,
+        stderr = os.Inherit,
```
lihaoyi, line 276: No need `stderr = os.Inherit`

lihaoyi:
Last few comments, after that we can merge it! Test failures are probably flakes and can be ignored.

sake92:
Should be good now.

lihaoyi:
@sake92 looks like there's a compile error somewhere

PR #4458, opened 2025-02-02
Remove all @deprecated classes, methods and fields

lolgab:


lefou:
I think we should support older cli options a while longer. Those are often used in other tools or scripts and typically evolve slower that build scripts. E.g. Scala Steward already has a hard time to handle all supported sets of Mill cli options, since it aims to support older Mill versions specifically. I had the same issues in the `mill-integrationtest` plugin, which needed to run much older Mill versions when cross-testing plugins. Keeping older cli options has a long tradition in classic Linux or GNU cli tools, too. It also kind-of makes sure, we not add new options that potentially collide with previously removed ones.

lihaoyi:
Superseded by https://github.com/com-lihaoyi/mill/pull/4499

Issue #3888, opened 2024-11-02
Add Mill Spotless Plugin (500USD Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 500USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

Spotless is used by many of our third-party example builds: Gatling, Mockito, and now Arrow-Kt. We should have official support in Mill for a `SpotlessModule` to allow usage of Spotless in `JavaModule`, `ScalaModule`, and `KotlinModule`

ayewo:
Hey @lihaoyi 

I've actually been working on this task for a bit and I've been able to get most of it working locally.

It's clear the effort is much more than the examples `PalantirFormatModule`, `ScalafmtModule`, `ErrorProneModule` you want it to be styled after. Good thing is that you acknowledged as much in this [thread](https://github.com/com-lihaoyi/mill/issues/3448#issuecomment-2326555273).

For instance, the code inside the main `spotless.jar` [does not expose](https://github.com/diffplug/spotless/issues/268#issuecomment-407523105) a main class that can be invoked from a separate sub-process. They expect anyone trying to integrate it with [a new build tool](https://github.com/diffplug/spotless/blob/main/CONTRIBUTING.md#how-to-add-a-new-plugin-for-a-build-system) to create all that plumbing, by looking at how it was done for `maven` or `gradle`.

This meant that I first had to build such a `main` class as a standalone Java project so I could create a `spotless` based Java formatter. The same approach is used for Scala and Kotlin, which are also in scope of this ticket.

So I'm wondering if you are open to upping the bounty amount?

PR #4464, opened 2025-02-03
Add Spotless plugin support

ayewo:
This PR adds official support in Mill for a `SpotlessModule` via the addition of 3 Spotless traits:

- `JavaSpotlessModule`
- `KotlinSpotlessModule` and 
- `ScalaSpotlessModule`  
 
which can be used with `JavaModule`, `KotlinModule` and `ScalaModule` respectively. This is because each of those 3 JVM languages have different 3rd-party dependencies.

It works similar in style to `PalantirFormatModule`.

This is intended to close https://github.com/com-lihaoyi/mill/issues/3888


example/javalib/linting/4-spotless/build.mill
```diff
@@ -0,0 +1,173 @@
...
+/*
+// See https://github.com/diffplug/spotless/tree/main/plugin-gradle#kotlin
+object `package` extends RootModule with KotlinSpotlessModule {
+  val ktfmtOptions: Map[String, Any] = Map(
+    "maxWidth" -> Some(80), // Expects an Option[Int]
+    "blockIndent" -> Some(4), // Expects an Option[Int]
+    "continuationIndent" -> Some(4), // Expects an Option[Int]
+    "removeUnusedImports" -> Some(false), // Expects an Option[Boolean]
+    "manageTrailingCommas" -> Some(false) // Expects an Option[Boolean]
+  )
```
ayewo, line 58: @lihaoyi  Can I leave this as-is, change to a `case class` or import the actual Java class  `com.diffplug.spotless.kotlin.KtfmtStep.KtfmtFormattingOptions` directly?

lefou: I think making it a Mill owned case class would suite our API better. It could hold a `additionalOptions` accepting a `Option[String]` to support future options, if this seems appropriate.

scalalib/src/mill/javalib/spotless/settings.scala
```diff
@@ -0,0 +1,270 @@
...
+    ktfmtVersion: String = "0.53",
+    ktfmtOptions: Option[KtfmtOptions] = None,
+    klintVersion: String = "1.5.0"
+) extends JVMLangConfig {
+  require(
+    !(licenseHeader.isDefined && licenseHeaderFile.isDefined),
+    "Please specify only licenseHeader or licenseHeaderFile but not both"
+  )
+
+  def withKtfmtVersion(version: String): KotlinConfig =
```
lihaoyi, line 239: Let's drop all the `.with*` updaters. Once `@unroll` comes out in Scala 3.7, we can add fields in a binary compatible way by directly adding fields

scalalib/src/mill/javalib/spotless/settings.scala
```diff
@@ -0,0 +1,270 @@
...
+        val header: String = os.read(Path(file))
+        LicenseHeaderStep.headerDelimiter(header, licenseHeaderDelimiter)
+      }))
+      .foreach(step => steps.add(step.build()))
+  }
+}
+
+sealed trait JavaFormatter
+
+case class GoogleJavaFormat private (
```
lihaoyi, line 40: Maybe let's organize this as

```scala
trait Config
object Config{
  case class GoogleJavaFormat() extends Config
  case class PalantirJavaFormat() extends Config
  case class JavaConfig() extends Config
  case class ScalaConfig() extends Config
  case class KtFmtConfig() extends Config
  case class KotlinConfig() extends Config
}
```

Also, what's the difference between e.g. `KotlinConfig` and `KtFmtConfig`?

ayewo: A. I'm a little confused as to why you want `GoogleJavaFormat` and `PalantirJavaFormat` to each extend `Config` when they both currently `extends JavaFormatter`? 
```scala
case class GoogleJavaFormat() extends Config
case class PalantirJavaFormat() extends Config
```

They will only be consumed by `JavaConfig` as `formatter: Option[JavaFormatter]`.

---

B. Not sure where you got `KtFmtConfig` from. Perhaps you meant `KtfmtOptions`?

Anyway, `KotlinConfig` is used to instantiate these Spotless Java classes:

1. `com.diffplug.spotless.kotlin.KtfmtStep` and 
2. `com.diffplug.spotless.kotlin.KtLintStep`

which in turn use Mill and the Spotless class loader to load 3rd-party dependencies from these Maven coordinates:
1. `com.facebook:ktfmt`
2. `com.pinterest.ktlint:ktlint-cli`

`KtfmtOptions` is used to customize the options passed to `KtfmtStep` (i.e. in 1. above).

IOW, `KtfmtOptions` is a Scala wrapper for the Java class:
- `com.diffplug.spotless.kotlin.KtfmtStep.KtfmtFormattingOptions`

so devs only have to import `mill.javalib.spotless.KtfmtOptions` instead of trying to use `com.diffplug.spotless.kotlin.KtfmtStep.KtfmtFormattingOptions` directly in their `build.mill` file.

lihaoyi: I think I'm a bit confused about how the classes fit together. Let me soend some more time looking at the PR to try and figure it out

scalalib/src/mill/javalib/spotless/settings.scala
```diff
@@ -0,0 +1,270 @@
...
+    manageTrailingCommas: Option[Boolean] = Some(false)
+)
+case class KotlinConfig(
+    target: String = ".kt",
+    licenseHeader: Option[String] = None,
+    licenseHeaderFile: Option[String] = None,
+    override val licenseHeaderDelimiter: String = "(package |@file|import )",
+    ktfmtVersion: String = "0.53",
+    ktfmtOptions: Option[KtfmtOptions] = None,
+    klintVersion: String = "1.5.0"
```
lihaoyi, line 232: Is there some way to configure whether Spotless uses `ktfmt` or `ktlint`? AFAIK both of them provide formatters. Or is Spotless hardcoded to use one or the other?

ayewo: There's is currently no way to configure whether Spotless uses `ktlint`. It's currently set as the default.

If a valid config for `ktfmt` is provided via `KotlinConfig(...)` in the `build.mill`, then `ktfmt` will also be used, alongside the default formatter i.e. `ktlint`.

You want both of them to be configurable?

lihaoyi: No need then, I think we can just follow the upstream convention

ayewo: Based on your answer, it sounds like you were asking about upstream Spotless, not my implementation?
My answer was regarding my implementation.

If you were specifically asking regarding the Spotless project, this is how they do it for [Kotlin](https://github.com/diffplug/spotless/tree/main/plugin-gradle#kotlin):
```groovy
  kotlin {
    // by default the target is every '.kt' and '.kts` file in the java sourcesets
    ktfmt()    // has its own section below
    ktlint()   // has its own section below
    diktat()   // has its own section below
    prettier() // has its own section below
    licenseHeader '/* (C)$YEAR */' // or licenseHeaderFile
  }
```

In the example `build.gradle` file above, the presence of `ktfmt()` would correspond to the invocation of the default version of `ktfmt` via this config method [`ktfmt()`](https://github.com/diffplug/spotless/blob/5a5db69db6254de76c2b33ca4c1f76d3da5d62b3/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/BaseKotlinExtension.java#L60).

```groovy
  kotlin {
    ktfmt("0.51") 
    ...
  }
```

In the example above, passing a specific version i.e. `ktfmt("0.51")` would invoke a similar config method that takes a version argument: [`ktfmt(String version)`](https://github.com/diffplug/spotless/blob/5a5db69db6254de76c2b33ca4c1f76d3da5d62b3/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/BaseKotlinExtension.java#L68).

```groovy
  kotlin {
    licenseHeader '/* (C)$YEAR */' // or licenseHeaderFile
  }
```
In this example, `ktfmt` will be skipped entirely since it is completely absent from the steps.

ayewo: Following the upstream convention would mean refactoring `KotlinConfig` and similar case classes to look like this:
```scala
case class KotlinConfig(
    target: String = ".kt",
    licenseHeader: Option[String] = None,
    licenseHeaderFile: Option[String] = None,
    override val licenseHeaderDelimiter: String = "(package |@file|import )",
    ktfmt: Boolean = false,
    ktfmtVersion: String = "0.53",
    ktfmtOptions: Option[KtfmtOptions] = None,
    klint: Boolean = false,
    klintVersion: String = "1.5.0"
) extends JVMLangConfig {
  require(
    !(licenseHeader.isDefined && licenseHeaderFile.isDefined),
    "Please specify only licenseHeader or licenseHeaderFile but not both"
  )    

  def ktfmt(): KotlinConfig =
    copy(ktfmt = true)

  def ktfmt(version: String): KotlinConfig =
    copy(ktfmt = true, ktfmtVersion = version)

  def ktfmtOptions(options: KtfmtOptions): KotlinConfig =
    copy(ktfmtOptions = Some(options))

  def ktlint(): KotlinConfig =
    copy(klint = true)

  def ktlint(version: String): KotlinConfig =
    copy(klint = true, klintVersion = version)
...
}
```
With this change, `ktlint` would no longer added as a formatter step by default.

lihaoyi:
I think the config looks fine as is. Let's drop the `with` updaters as mentioned earlier, then we can merge it

scalalib/src/mill/javalib/spotless/settings.scala
ayewo: I've pushed changes accordingly.

lihaoyi:
Is there a reason you want to use forwarders rather than using the `.copy` method directly?

ayewo:
No reason other than I thought you wanted it to be similar to the style of Spotless in Groovy.

lihaoyi:
Let's use the `.copy` methods directly then. We need to find a balance between the upstream style and the local Scala style, and in this case I think `.copy` wins out

ayewo:
I've updated to use `.copy()` directly now.

ayewo:
@lihaoyi Anything else left for this to be merged ?

example/javalib/linting/4-spotless/build.mill
```diff
@@ -0,0 +1,101 @@
...
+    .copy(formatJavadoc = true)
+
+  def jvmLangConfig = new JavaConfig()
+    .copy(importOrder = Some(Seq()))
+    .copy(formatter = Some(googleFormatter))
+    .copy(formatter =
+      Some(palantirFormatter)
+    ) // this replaces `googleFormatter` as only 1 `JavaFormatter` can be active
+    .copy(licenseHeader = Some("/* (C) 2025. Licensed under Apache-2.0. */\n"))
+}
```
lihaoyi, line 35: We should not need to use `.copy` here at all I think, we can just use the `GoogleJavaFormat(...)` constructor. Same as the others?

scalalib/src/mill/javalib/spotless/SpotlessModule.scala
lefou, line 1: I'd prefer to have the top-level classes in their own files.

lefou (commented):
I'd prefer to have the top-level classes in their own files.

example/javalib/linting/4-spotless/build.mill
ayewo: Removed as requested.

ayewo:
@lefou Done.

scalalib/src/mill/javalib/spotless/JavaSpotlessModule.scala
```diff
@@ -0,0 +1,35 @@
...
+        ivy"com.palantir.javaformat:palantir-java-format:${palantirJavaFormatVersion()}"
+      )
+    )
+  }
+
+  /**
+   * Google Java Format version. Defaults to `1.25.2`.
+   */
+  def googleJavaFormatVersion: T[String] = Task {
+    "1.25.2"
```
lefou, line 26: Where does this version come from? Is this the newest version? Do users expect it to stay stable or use always the newest? 

We should not hardcode any versions. We either should leave it undefined or use a `BuildInfo` value that comes from the outer Mill build. That way, we can keep it up-to-date more easily.

scalalib/src/mill/javalib/spotless/JavaSpotlessModule.scala
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  /**
+   * Google Java Format version. Defaults to `2.50.0`.
+   */
+  def palantirJavaFormatVersion: T[String] = Task {
+    "2.50.0"
```
lefou, line 33: Same as for `googleJavaFormatVersion`, versions should not be hardcoded. Have a look at PR https://github.com/com-lihaoyi/mill/pull/4552, which replaces another instance of a hardcoded version.

scalalib/src/mill/javalib/spotless/KotlinSpotlessModule.scala
```diff
@@ -0,0 +1,35 @@
...
+        ivy"com.pinterest.ktlint:ktlint-cli:${ktlintVersion()}"
+      )
+    )
+  }
+
+  /**
+   * Defaults to `0.53`.
+   */
+  def ktfmtVersion: T[String] = Task {
+    "0.53"
```
lefou, line 26: Should not be hardcoded. This version is already managed. 

https://github.com/com-lihaoyi/mill/blob/facbe11cf4d2ab176aa16780ddcf509e67d5a2cd/kotlinlib/src/mill/kotlinlib/ktfmt/KtfmtModule.scala#L26

scalalib/src/mill/javalib/spotless/KotlinSpotlessModule.scala
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  /**
+   * Defaults to `1.5.0`.
+   */
+  def ktlintVersion: T[String] = Task {
+    "1.5.0"
```
lefou, line 33: Use managed `Versions.ktlintVersion` instead.

scalalib/src/mill/javalib/spotless/settings.scala
```diff
@@ -0,0 +1,224 @@
...
+    formatJavadoc: Boolean = true,
+    reorderImports: Boolean = true,
+    groupArtifact: String = "com.google.googlejavaformat:google-java-format"
+) extends JavaFormatter {}
+object GoogleJavaFormat {
+  def apply(): GoogleJavaFormat = new GoogleJavaFormat()
+}
+
+case class PalantirJavaFormat(
+    version: String = "2.50.0",
```
lefou, line 59: Don't hardcode versions.

scalalib/src/mill/javalib/spotless/settings.scala
```diff
@@ -0,0 +1,224 @@
...
+case class KotlinConfig(
+    target: String = ".kt",
+    licenseHeader: Option[String] = None,
+    licenseHeaderFile: Option[String] = None,
+    override val licenseHeaderDelimiter: String = "(package |@file|import )",
+    ktfmtFlag: Boolean = false,
+    ktfmtVersion: String = "0.53",
+    ktfmtOptions: Option[KtfmtOptions] = None,
+    ktlintFlag: Boolean = false,
+    ktlintVersion: String = "1.5.0"
```
lefou, line 190: Don't hardcode versions. Theese are already available via the `Versions` object.

scalalib/src/mill/javalib/spotless/JavaSpotlessModule.scala
ayewo: I've removed the hard-coding by making it undefined and deferring the setting of its value inside `build.mill`.

ayewo:
I believe I've fixed all instances of versions being hard-coded.

This is ready to be merged @lihaoyi.

lihaoyi:
@ayewo the example test seems to be failing, you can find instructions to run it locally in the `contributing.adoc`. The example tests serve as both tests and documentation, so you need to make sure the tests pass and the english explanations are properly written up and included in `website/docs/modules/ROOT/pages/javalib/linting.adoc`, `website/docs/modules/ROOT/pages/scalalib/linting.adoc`, `website/docs/modules/ROOT/pages/kotlinlib/linting.adoc`. See the other example tests and their place in the documentation for reference

lefou (commented):
It might have slipped thru in previous reviews, but it look like you expecting the spotless library on the build classpath, whereas what we want it the `SpotlessModule` to load all necessary libraries dynamically on demand. the whole point of abstracting over the spotless API is, that spotless itself is not on the classpath and will only be loaded dynamically based on the version configured in the module. All code that is coded against any spotless API needs therefore to reside in a dedicated modules.

For example, `ZincWorkerModule` is resposible to abstract over a concrete zinc dependency. The shared API (`ZinkWorkerAii`) is loaded from the `mill.scalalib.api` module, whereass the implemntation that has access to the actual zinc API resides in `mill.scalalib.worker` and is loaded on demand by `ZincWorkerModule`.

ayewo:
> It might have slipped thru in previous reviews, but it looks like you expecting the spotless library on the build classpath, whereas what we want is the `SpotlessModule` to load all necessary libraries dynamically on demand. The whole point of abstracting over the spotless API is, that spotless itself is not on the classpath and will only be loaded dynamically based on the version configured in the module. All code that is coded against any spotless API needs therefore to reside in a dedicated module.
> 
> For example, `ZincWorkerModule` is resposible to abstract over a concrete zinc dependency. The shared API (`ZinkWorkerApi`) is loaded from the `mill.scalalib.api` module, whereass the implemntation that has access to the actual zinc API resides in `mill.scalalib.worker` and is loaded on demand by `ZincWorkerModule`.

I don't understand what you are saying here.

The main module doing all the heavy-lifting is `SpotlessModule` and it already depends on `CoursierModule`:
```scala
trait SpotlessModule extends CoursierModule
```

mill-build/build.mill
```diff
@@ -12,6 +12,7 @@ object `package` extends MillBuildRootModule {
     ivy"com.goyeau::mill-scalafix::0.5.0",
     ivy"com.lihaoyi::mill-main-graphviz:${mill.api.BuildInfo.millVersion}",
     // TODO: document, why we have this dependency
-    ivy"org.jsoup:jsoup:1.18.1"
+    ivy"org.jsoup:jsoup:1.18.1",
+    ivy"com.diffplug.spotless:spotless-lib:3.0.2"
```
lefou, line 16: We should not required to add `spotless-lib` to the Mill classpath.

scalalib/src/mill/javalib/spotless/SpotlessModule.scala
```diff
@@ -0,0 +1,192 @@
...
+import mill.scalalib.CoursierModule
+import java.util.Set
+import java.util.stream.Collectors
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.Paths
+import java.io.File
+import java.io.IOException
+import com.diffplug.spotless.{Provisioner, Formatter, LineEnding}
```
lefou, line 14: This import means, the spotless API is already on the Mill classpath.

scalalib/src/mill/javalib/spotless/settings.scala
```diff
@@ -0,0 +1,218 @@
...
+import com.diffplug.spotless.java.{
+  GoogleJavaFormatStep,
+  PalantirJavaFormatStep,
+  ImportOrderStep,
+  RemoveUnusedImportsStep,
+  CleanthatJavaStep,
+  FormatAnnotationsStep
+}
+import com.diffplug.spotless.generic.LicenseHeaderStep
+import com.diffplug.spotless.Provisioner
```
lefou, line 18: These imports mean, we need spotless on the Mill classpath.

scalalib/package.mill
```diff
@@ -18,7 +18,7 @@ object `package` extends RootModule with build.MillStableScalaModule {
 
   def moduleDeps = Seq(build.main, build.scalalib.api, build.testrunner)
   def ivyDeps = {
-    Agg(build.Deps.scalafmtDynamic, build.Deps.scalaXml) ++ {
+    Agg(build.Deps.scalafmtDynamic, build.Deps.scalaXml, build.Deps.spotless) ++ {
```
lefou, line 21: This dependency forced spotless on the Mill classpath.

lefou (commented):
Let me point out some issues or indicators, that we load or are required to load the spotless API on the Mill classpath. Beside the fact that `SpotlessModule` `extends` `CoursierModule`, all ` com.diffplug.spotless._`  imports don't get loaded via that `CoursierModule` but are rather already on the classpath, as you added spotless to the dependencies of `mill.scalalib` module.

lefou:
> It's not nice at all, to point out nits, a contributor fixes them, the cycle repeats, then you come swinging pointing out a fresh set of nits. Please be respectful of contributors time.

@ayewo I'm sorry if you feel that way. And I understand your emotions under the assumptions, these are nits. But those points are not nits. And it's not only your time that is spent here. Until now I only commented on smaller parts under the assumption that the main feature was already reviewed by @lihaoyi. Now, I recognized some major issues and felt the need to point them out.

I understand, that the amount of work to finish this PR is likely higher compared to e.g. implementing `PalantirModule`, but my review is not bound to any assumptions you might have based on the amount of a bounty. 

Just to make it clear, my latest review points are not nits. Your PR description claims to do it the same way as other modules, yet all those other modules don't add any dependencies to `mill.scalalib`. That does not mean the general design and all review that went into it, is lost work. It only means, we need to move some of these API classes into dedicated modules and load them at runtime.

I'm glad to help you with the module split-up, if you need help. My main focus is to get a new feature into Mill without breaking all other properties and users of Mill.

lihaoyi:
@lefou's is correct, the issues he raised are important, I missed them in my own review, and we should resolve them before merging this. @lefou has offered to help get it in if you would like it.

Some friction in getting PRs in is expected. If we were all sitting in the same room this would be smoother, but we're scattered around the world, so we'll have to live with some inefficiency.

lihaoyi:
closing this since the PR appears to be inactive for a month. Thanks ayewo for your work on this, perhaps someone in future may make more progress on this. Feel free to open a new PR if necessary

lefou:
@ayewo you never reached out to me, so I was inactive on this issue. @lihaoyi If we still want that feature, I could add the necessary worker stuff and see if the rest is usable.

lihaoyi:
@lefou I think the feature would be nice to have. Spotless is widely used, so having some kind of support for it definitely checks a box and should answer a frequently asked question 

Feel free to pick it up to see if you can implement the worker stuff and get it over the finish line

ayewo:
@lihaoyi 

1. > [@lefou](https://github.com/lefou)'s is correct, the issues he raised are important, I missed them in my own review, and we should resolve them before merging this.

Of course the issue he raised on March 4 (rather belatedly) is important. I know because I also tried to raise it with you—on Jan 29. This was prior to opening my PR, albeit with a different framing. I posted a reminder on Feb 4, but all I got was crickets.

I framed the [question](https://github.com/com-lihaoyi/mill/issues/3888#issuecomment-2620271231) in terms of time-&-effort:

> "Hey @lihaoyi 
> 
> I've actually been working on this task for a bit and I've been able to get most of it working locally.
> 
> It's clear the effort is much more than the examples `PalantirFormatModule`, `ScalafmtModule`, `ErrorProneModule` you want it to be styled after. Good thing is that you acknowledged as much in this [thread](https://github.com/com-lihaoyi/mill/issues/3448#issuecomment-2326555273).
> 
> For instance, the code inside the main `spotless.jar` [does not expose](https://github.com/diffplug/spotless/issues/268#issuecomment-407523105) a main class that can be invoked from a separate sub-process. They expect anyone trying to integrate it with [a new build tool](https://github.com/diffplug/spotless/blob/main/CONTRIBUTING.md#how-to-add-a-new-plugin-for-a-build-system) to create all that plumbing, by looking at how it was done for `maven` or `gradle`.
> 
> This meant that I first had to build such a `main` class as a standalone Java project so I could create a `spotless` based Java formatter. The same approach is used for Scala and Kotlin, which are also in scope of this ticket.
> 
> So I'm wondering if you are open to upping the bounty amount?"

In the original ticket, you wrote:

> Should be similar in style of `PalantirFormatModule`, `ScalafmtModule`, `ErrorProneModule`, etc.

Based on your guidance and my research, any PR adding support for Spotless could only mimic the style of either `PalantirFormatModule` or `ScalafmtModule` (but not `ErrorProneModule` since Spotless doesn't expose a compiler plugin). 

It was clear that it was going to require more time/effort to mimic the design of `PalantirFormatModule`because Spotless doesn't ship a main class that would have made it trivial to consume it from a sub-process as was done with `PalantirFormatModule`(or `CheckstyleModule`), which is why I framed my question accordingly.

Once I didn't get any response from you regarding my question, I simply mimicked the style of `ScalafmtModule`.
Now my PR has been closed because of a design decision that I **tried to be up-front about, prior to opening it**.

---

2. > [@lefou](https://github.com/lefou) has offered to help get it in if you would like it.

That’s fine. Tobias' [explanation](https://github.com/com-lihaoyi/mill/pull/4464#issuecomment-2697980798) as to why my PR shouldn't add any dependencies to `mill.scalalib` is really just a consequence of what I was trying to ask in #1. 

---

3. > Some friction in getting PRs in is expected. If we were all sitting in the same room this would be smoother, but we're scattered around the world, so we'll have to live with some inefficiency.

Yes, PR friction is a thing but this PR hasn’t **gotten in** yet, so I don’t think what you are alluding to is what happened here.

Tobias started out by requesting minor changes instead of anything substantive. A few weeks pass and after many of those minor issues have been fixed, he starts pointing out substantive changes that would cause the PR to be reworked.

When reviewing PRs, many folks start with pointing out major issues (that could prevent the PR from being accepted) and end with minor issues, while some do it in reverse (like Tobias did here), so there is no right or wrong way to review PRs.

But in this instance, because I’m still **a new contributor** to the project, the first style would have saved everyone—on both sides—a lot of time, if Tobias had started with pointing out substantive issues, instead of minor ones. (Yes, I saw his explanation where he assumed you had already done this).

IMHO, two things could have prevented this from going off the rails:
1. try to respond to contributor questions, even if you might disagree with its framing;
2. add a checklist to CONTRIBUTING.md and link it to the PR template that will guide new contributors and reviewers on common mistakes to avoid (and perhaps also a style guide).

This situation has nothing to do with everyone not being in the same room. I’ve plenty of experience working with distributed teams, so I was really furious when I saw that non sequitur in your reply, which is why I avoided responding in the moment.

ayewo:
> @ayewo you never reached out to me, so I was inactive on this issue. @lihaoyi If we still want that feature, I could add the necessary worker stuff and see if the rest is usable.

@lefou you can go right ahead, thanks.

lihaoyi:
This PR is closed, as incomplete and failing to fulfil the "Standard PR requirements" of responding to code review change requests. PRs get non-trivial code change requests all the time, and these requests don't always come in some idealized ordering. It is not rare for PRs to be closed due to inability of the author to satisfy code review. Mill needs people who are able to contribute despite such setbacks, and not respond to technical code review with a long airing of grievances.

Fundamentally, it seems like we're just not compatible. As you said you have had better luck working on distributed teams elsewhere, and I have had better luck working with other distributed contributors as well.  So although I appreciate the work you have done on this PR and in previous bounties, I think it would be best to stop collaborating and we can each find better counterparties to collaborate with.

ayewo:
I’ve not had much availability lately, which is why I only got round to reading your 2nd reply yesterday.

Anyway, I’m glad you posted it because it brought much needed clarity to the discussion.

## Impedance Mismatch
The goal in my original comment was to provide actionable feedback, from the perspective of a new contributor, but it clearly didn’t land how I intended it. 

In fact, it didn’t land at all especially since you characterized it as a “long list of grievances” :D.

My frustration dwelt on the fact that my PR wasn’t done after it had open for 2 months, but your 1st reply dwelt on something else entirely.

Based on both of your replies, I completely agree that it’s clear there’s an impedance mismatch. And it’s not hard to see why.

## The Wrong Mental Model
I guess my mistake—and I apologize for this—was to approach Mill as if it were a COSS (Commercial Open Source Software) project. COSS is about having your ducks in a row, moving fast, time is money etc. 

Also in COSS projects, the bounty amount is often a signal of two things:
- the complexity of an issue;
- the priority of the issue to the business, relative to other open issues.

Your 2nd reply makes it abundantly clear that as an OSS project, Mill’s expectations of maintainers and contributors are the (complete?) opposite of the expectations of a COSS project.

If my reading is correct, then once again I sincerely apologize. My unsolicited feedback was completely warranted in this context, especially since this is more of a hobby than anything resembling a business.

I don’t have much availability right now, but if you agree, when I have availability to push this over the finish line, I can open a new PR to continue where this left off.

(Replying on a phone so pls excuse any typos.)

PR #4485, opened 2025-02-05
Kotlin KSP support with basic dependency injection example using Dagger

vaslabs:
# Motivation

While exploring the [android architecture samples](https://github.com/android/architecture-samples/tree/main) to identify functionality that is missing  from Mill, I've realised that the [Kotlin Symbol Processing](https://kotlinlang.org/docs/ksp-command-line.html) support is missing .

In order to support such setups, we need to support KSP on the Kotlin side and then work our way to Android.

# Background

KSP is an API that allows for compiler extensions in Kotlin. It is similar (in end result) to annotation processors in Java, and one such example is dagger.

## Dagger, Kotlin and Android

Dagger is static dependency injection framework, i.e. avoiding runtime errors. Android works with Dagger via [Hilt](https://dagger.dev/hilt/), which reduces the Dagger boilerplate. Hilt support is also referenced from [this issue](https://github.com/com-lihaoyi/mill/issues/3868)

The example demonstrates these boilerplate problems  as, in this first basic setup, the PR just aims for the dagger components and modules to be compiled.

## Gradle

Before starting the work, I've put together a basic example using Gradle which can be found in [this repo](https://github.com/vaslabs-ltd/kotlin-dagger-demo) .

If you'd like to observe the Gradle behaviour as a black box you can run

```bash
./gradlew --debug compileKotlin
```

Or view the logs [here](https://github.com/vaslabs-ltd/kotlin-dagger-demo/actions/runs/13371896338/job/37342109287)

By observing the logs, one should see 2 compilation tasks (via the kotlin-compiler-embedded) , one executes the KSP and the next one the normal compilation.

In addition, the classpaths and plugin arguments can also be observed. I'm not gonna add log snippets here for the sake of clarity, but for the reader to know what to look for in terms of command line arguments, the following is how one would run KSP from the command line. More documentation can be found [here](https://kotlinlang.org/docs/ksp-command-line.html)

```bash
#!/bin/bash

KSP_PLUGIN_ID=com.google.devtools.ksp.symbol-processing
KSP_PLUGIN_OPT=plugin:$KSP_PLUGIN_ID

KSP_PLUGIN_JAR=./com/google/devtools/ksp/symbol-processing-cmdline/2.1.0-1.0.29/symbol-processing-cmdline-2.1.0-1.0.29.jar
KSP_API_JAR=./com/google/devtools/ksp/symbol-processing-api/2.1.0-1.0.29/symbol-processing-api-2.1.0-1.0.29.jar
KOTLINC=./kotlinc/bin/kotlinc

AP=/path/to/your-processor.jar

mkdir out
$KOTLINC \
        -Xplugin=$KSP_PLUGIN_JAR \
        -Xplugin=$KSP_API_JAR \
        -Xallow-no-source-files \
        -P $KSP_PLUGIN_OPT:apclasspath=$AP \
        -P $KSP_PLUGIN_OPT:projectBaseDir=. \
        -P $KSP_PLUGIN_OPT:classOutputDir=./out \
        -P $KSP_PLUGIN_OPT:javaOutputDir=./out \
        -P $KSP_PLUGIN_OPT:kotlinOutputDir=./out \
        -P $KSP_PLUGIN_OPT:resourceOutputDir=./out \
        -P $KSP_PLUGIN_OPT:kspOutputDir=./out \
        -P $KSP_PLUGIN_OPT:cachesDir=./out \
        -P $KSP_PLUGIN_OPT:incremental=false \
        -P $KSP_PLUGIN_OPT:apoption=key1=value1 \
        -P $KSP_PLUGIN_OPT:apoption=key2=value2 \
        $*
```

## Kotlin compiler classpath

Now, in the documented example, the `symbol-processing-cmdline` is used. Gradle uses `symbol-processing` and `kotlin-compiler-embedded` .

I've used the same dependencies as Gradle, because kotlin-compiler-embedded safeguards the compiler classpath from the application classpath. For example, if `kotlin-compiler` is used with `symbol-processing-cmdline` , classes from the kotlin compiler (`com.google.common.base.BaseGraph`) conflict with guava classpaths (inherited from the dagger plugin).

# Provided in this PR

- Support for KSP and example usage using Dagger
- Auto-configuration of the needed symbol-processing plugins along with their dependencies, including kotlin configuration such as language level
- Multi-step compilation, with KSP done as part of the generatedSources


vaslabs:
@lihaoyi this should be ready, there's a test failing https://github.com/vaslabs-ltd/mill/actions/runs/13240848505/job/36959007556 but I think it's unrelated (I re-fixed a failing test on android, don't know why this zip behaviour keeps changing occasionally)

lihaoyi:
Thanks, will take a look

vaslabs:
some android tests started failing after the Agg -> Seq changes introduced duplicates, if this change should stay, I'll need to look into removing these duplicates

fyi @lihaoyi

lihaoyi:
Yes the migration from Agg to Seq will stay. May need to sprinkle in some `.distinct`s if you're finding unwsnted duplicates slipping in

vaslabs:
This is also ready

A followup PR could work to make use of the embedded kotlin compilers universally

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -74,11 +74,59 @@ trait KotlinModule extends JavaModule { outer =>
    */
   def kotlinApiVersion: T[String] = Task { "" }
 
+  def symbolProcessingVersion: T[String] = Task { "1.0.29" }
```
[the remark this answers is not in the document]
vaslabs: I like the build info idea, if you could point me to some examples I'll explore the options. The reason for hardcoding this (partial version, because the full version is a combination of kotlin compiler version and this one) is that we know it works with the plugin options in [here](https://github.com/com-lihaoyi/mill/pull/4485/files#diff-6d7da547530735d1e00dbce596fdd8265d733dabf357ff2a59ce92ab5b1708c2R359)

But if we can have it generated in a build info and the tests pass, then that would be ideal I think!

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -74,11 +74,59 @@ trait KotlinModule extends JavaModule { outer =>
...
+
   /**
    * Flag to use explicit API check in the compiler. Default is `false`.
    */
   def kotlinExplicitApi: T[Boolean] = Task { false }
 
+  /** Flag to use KSP */
+  def kotlinSymbolProcessing: Boolean = false
```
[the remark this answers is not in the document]
vaslabs: I did struggle with both options in my head for a bit. Technically we can do either, I'm still 50-50. So the reasons I considered for Pros and Cons on the current solution are

+ Feature toggle aside, the experience is similar to gradle, just declaring KSPs without the symbol processing api
- The KotlinModule is a bit more complex, although
+ it forced me (feature toggle aside) to find a better solution with using the embedded compiler which I think should be the default once we make sure everything else is stable

In summary, as I see it, I put this as a feature flag, but everything should work without it . Going forward I could do one of these options:

1. Remove the feature flag and use the user defined KSPs (if they are present, then kotlin symbol processing should be enabled), and hopefully all other kotlin tests pass without issues too
2. Leave the feature flag, do a bit of testing on Android, see any gaps and come back to this with new information
3. Create a new mixing trait for KSPs, which effectively will work the same as this feature flag

All in all, I don't have a strong opinion, but I'd prefer option 2, it's up to you though!

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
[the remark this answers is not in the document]
lefou: For an example, have a look at `KtfmtBaseModule`. The buildinfo is named `Versions`.

https://github.com/com-lihaoyi/mill/blob/facbe11cf4d2ab176aa16780ddcf509e67d5a2cd/kotlinlib/src/mill/kotlinlib/ktfmt/KtfmtModule.scala#L25-L27

The value is setup here:

https://github.com/com-lihaoyi/mill/blob/facbe11cf4d2ab176aa16780ddcf509e67d5a2cd/kotlinlib/package.mill#L23

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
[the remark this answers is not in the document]
lefou: I'm not a fan of the feature toggle, TBH. What are the reasons for it? Do users need to disable KSP sometimes? Is it slow? Does it even make sense to toogle it off in a module that configured KSP and most likely depends on KSP?

Since we implicitly need to add some compiler plugins to enable KSP (which has some potential to make things slower or collide with other plugins), I tend to prefer a dedicated `KotlinSymbolProcessingModule` (I'm open for a better name) that can be mixed in. This makes the additional task more transparent and easier to review. It also keeps the `KotlinModule` lean. Of course, this assumes, users need this feature seldomly. Do you know if this KSP is needed often?

Unfortunately I'm not deep enough into that topic to make an elaborated choice about the embedded compiler. Is there some info about a preferred way? What I know is, that we're currently not able to profit from Kotlin incremental compilation. Maybe, we can fix this by using the embedded version? We certainly don't need to support very old Kotlin versions, if that is an issue.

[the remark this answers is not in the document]
vaslabs: the feature toggle was merely for my benefit to be able to organise my work with Android (e.g. I'd go next and prepare an android setup using Hilt, and once everything was working I planned to remove it). But a separate trait would have the same effect for me, so I have no objections to your suggestion, will try to get it ready tomorrow!

I think in Kotlin/Java world, dependency injection is quite frequent, and compile time DI is even should be preferable for many, although to be honest I haven't used such a pattern in years (I was using FP patterns with Scala). Android has a complementary plugin (Hilt) to dagger and I see it in android samples being used a lot, but I can ask around if you wanna make sure! Although, it won't affect anything, unless an actual preprocessor is plugged in the compiler (in this case, dagger), which uses these KSP APIs. So I don't think it affects performance or anything, it just remains in the classpath!

The embedded compiler is used by gradle as well, and one of the reasons is it's basically a shadowed version of the non embedded one. In this particular example, it avoids the conflict between the kotlin compiler declared BaseGraph with guava (being pulled from dagger) . So it's just the safest choice, [this discussion](https://discuss.kotlinlang.org/t/kotlin-compiler-embeddable-vs-kotlin-compiler/3196 ) should be helpful to get more context . 

So this is what I propose:

I'll split this PR into two:
1. switch kotlin module to use the embeddable compiler and make sure the tests pass
2. do a new trait (KSPModule) that adds the necessary configuration for KSP to work with the above example

 WDYT?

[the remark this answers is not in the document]
lefou: > So it's just the safest choice, [this discussion](https://discuss.kotlinlang.org/t/kotlin-compiler-embeddable-vs-kotlin-compiler/3196) should be helpful to get more context .

Looks like, both compiler packages might be valid options, but they're not compatible in terms of their API due to the shading/relocating. Fortunately, you already made it a dedicated config task, so a user can easily switch to the other dependency if their use-case needs them to.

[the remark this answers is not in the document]
vaslabs: having tried this https://github.com/com-lihaoyi/mill/pull/4557

I think we need to change the plan, instead of having the feature switch reference KSP, I can change it to embeddable (just naming) so we don't break apps, and manage the KSP in a separate module.

Let me know what you think in the new PR

lihaoyi:
@vaslabs I'll defer to @lefou for review on this one, he's much more familiar with Mill's Kotlin support than I am and will be able to give much more insightful feedback

vaslabs:
@lefou I've adjusted the implementation after the rebase to include the embeddable compiler work and moved the ksp logic to a separate module.

I'm not 100% happy as the user needs to enable ksp on every submodule something not needed in Gradle's case (https://github.com/vaslabs-ltd/kotlin-dagger-demo/blob/main/build.gradle.kts)

Do you have any proposals on how to enhance this?

lefou:
> I'm not 100% happy as the user needs to enable ksp on every submodule something not needed in Gradle's case (https://github.com/vaslabs-ltd/kotlin-dagger-demo/blob/main/build.gradle.kts)
> 
> Do you have any proposals on how to enhance this?

I'm not sure I quite understand the issue. Since in Mill you always specify the type of each submodule, what's the difference? The gradle project you linked doesn't contain subnodule either. Can you explicitly point out the pain points?

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -0,0 +1,157 @@
...
+/**
+ * Sets up the kotlin compiler for using KSP (Kotlin Symbol Processing)
+ * by plugging in the symbol-processing and symbol-processing-api dependencies.
+ *
+ * Use of kotlin-compiler-embedded is also recommended (and thus enabled by default)
+ * to avoid any classpath conflicts between the compiler and user defined plugins!
+ */
+trait KspModule extends KotlinModule {
+
+  def symbolProcessingVersion: T[String]
```
lefou, line 20: Can we name this `kspVersion` and add some docs?

example/kotlinlib/basic/7-dependency-injection/build.mill
```diff
@@ -0,0 +1,89 @@
...
+  def ivyDeps = Agg(
+    ivy"com.google.dagger:dagger-compiler:2.55"
+  )
+
+  def kotlinSymbolProcessors = Agg(
+    ivy"com.google.dagger:dagger-compiler:2.55"
+  )
+
+  object test extends KotlinTests with TestModule.Junit5 with KspModule {
+    def kotlinSymbolProcessing = true
```
lefou, line 30: This line looks like a left-over and can be removed.

example/kotlinlib/basic/7-dependency-injection/build.mill
```diff
@@ -0,0 +1,89 @@
...
+
+  def symbolProcessingVersion = "1.0.29"
+
+  def kotlinLanguageVersion = "1.9"
+
+  def mainClass = Some("com.example.dagger.MainKt")
+
+  def kotlincOptions = super.kotlincOptions() ++ Seq("-no-reflect", "-verbose")
+
+  def kotlinSymbolProcessing = true
```
lefou, line 19: This line looks like a left-over and can be removed.

example/kotlinlib/basic/7-dependency-injection/build.mill
```diff
@@ -0,0 +1,89 @@
...
+
+  def ivyDeps = Agg(
+    ivy"com.google.dagger:dagger-compiler:2.55"
+  )
+
+  def kotlinSymbolProcessors = Agg(
+    ivy"com.google.dagger:dagger-compiler:2.55"
+  )
+
+  object test extends KotlinTests with TestModule.Junit5 with KspModule {
```
lefou, line 29: If using KSP in test modules is the default or needed often, we should think about providing a pre-configured `KspTests` trait, so we don't need to refer to `outer` to make consistent settings.

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  def kspProjectBaseDir: T[PathRef] = Task {
+    PathRef(moduleDir)
+  }
+
+  /** KSP output dir */
+  def kspOutputDir: T[PathRef] = Task {
+    PathRef(T.dest / "generated" / "ksp" / "main")
+  }
```
lefou, line 42: Why is this needed. I results in a design we try to discourage in Mill. A task should not write into the dest of another task. Since this task is only used from the `generateSourcesWithKSP` task, why it it externalized? Instead, the `generateSourcesWithKSP` task should manage that dir by itself. If later, other tasks need access to it, them may as well just refer to that task.

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -0,0 +1,157 @@
...
+
+  /** KSP caches dir */
+  def kspCachesDir: T[PathRef] = Task {
+    PathRef(T.dest / "main")
+  }
```
lefou, line 47: What is this for? This is likely a detail of `generateSourcesWithKSP`.

vaslabs:
> > I'm not 100% happy as the user needs to enable ksp on every submodule something not needed in Gradle's case (https://github.com/vaslabs-ltd/kotlin-dagger-demo/blob/main/build.gradle.kts)
> > Do you have any proposals on how to enhance this?
> 
> I'm not sure I quite understand the issue. Since in Mill you always specify the type of each submodule, what's the difference? The gradle project you linked doesn't contain subnodule either. Can you explicitly point out the pain points?

Ideally this 
https://github.com/com-lihaoyi/mill/pull/4485/files#diff-8a2c86545176fed4cade80be52eb036a936a4a184ca5657f64a9a4a3e20702c5R29

and the following 2 lines should not be needed , for instance I'd prefer if KotlinTests worked with the same compiler options selected in the main module.

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -0,0 +1,157 @@
...
+
+  /** ksp generated sources */
+  private def kspGeneratedSources: T[Seq[PathRef]] = { generateSourcesWithKSP }
```
lefou, line 50: This looks redundant to `generateSourcesWithKSP`.

lefou (commented):
Below I have some review comments. This is not a complete review but there are already some fundamental questions, so let's go with them.

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
vaslabs: yes, I need to group these better, they are all part of the ksp domain, I'll come back with something better

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
vaslabs: done, let me know if you'd like some different wording

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
vaslabs: I've removed all of these and have them locally defined in the `generateSourcesWithKSP` task

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
vaslabs: removed

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
vaslabs: removed

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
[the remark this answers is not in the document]
vaslabs: resolved now

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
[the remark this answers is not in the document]
vaslabs: this is user defined now after the Ksp module separation, let me know if you'd prefer the build info route!

lefou (commented):
Looks good to me!

I think there will be some revisions in the future, as I just now realized, that we don't have proper support for Kotlin compiler plugins (except juggling with options) in our `KotlinModule`. 

There is also the semantics of `kspPlugins` and `kotlinSymbolProcessorId` which seem to be tightly coupled, but we nowhere make sure they match. Are there other known symbol processors? Maybe, we can think through these options to find a nicer solution:
* derive one task from the other
* combine them into one single task, to make sure, they stay consistent
* if there are more than one commonly used processors, provide pre-configured traits that configure the processor ID and dependencies (as we do with test frameworks)

vaslabs:
> Looks good to me!
> 
> I think there will be some revisions in the future, as I just now realized, that we don't have proper support for Kotlin compiler plugins (except juggling with options) in our `KotlinModule`.
> 
> There is also the semantics of `kspPlugins` and `kotlinSymbolProcessorId` which seem to be tightly coupled, but we nowhere make sure they match. Are there other known symbol processors? Maybe, we can think through these options to find a nicer solution:
> 
> * derive one task from the other
> * combine them into one single task, to make sure, they stay consistent
> * if there are more than one commonly used processors, provide pre-configured traits that configure the processor ID and dependencies (as we do with test frameworks)

that was my reasoning for a single module, I believe we should strive towards that in the future. 

From reading the docs and reverse engineering gradle, I think to support ksp and kotlin plugins in a generic way, we would need the embedded compiler. 

My understanding is that the processor id is fixed for KSP , to change it, one would need to replace the default symbol processing logic with something else.

I think it will inevitably have future iterations, I want to stress the implementation using Android and Hilt (which is what I'll do next) and see where that would leave us

lefou:
> From reading the docs and reverse engineering gradle, I think to support ksp and kotlin plugins in a generic way, we would need the embeddable compiler.

That may be. But beside KSP, other Kotlin plugins don't necessarily need the embeddable compiler. In another client project, we also had a Mill module to setup Kotlin compiler plugins (namely `kotlinx-serialization` and `allopen`), but we never upstream it. Probably because it's from the Kotlin 1.x time without proper plugin releases. We had to download and unpack the kotlin compiler distribution to get the plugin jars. We also applied other plugins (e.g. `dataframe`) but applied it by just mangling the `kotlincOptions`. 

> My understanding is that the processor id is fixed for KSP , to change it, one would need to replace the default symbol processing logic with something else.

I that case, we shouldn't have the `kotlinSymbolProcessorId` task at all.

> we can even annotate it as experimental?

Makes sense. Please add that annotation for now.

vaslabs:
> > From reading the docs and reverse engineering gradle, I think to support ksp and kotlin plugins in a generic way, we would need the embeddable compiler.
> 
> That may be. But beside KSP, other Kotlin plugins don't necessarily need the embeddable compiler. In another client project, we also had a Mill module to setup Kotlin compiler plugins (namely `kotlinx-serialization` and `allopen`), but we never upstream it. Probably because it's from the Kotlin 1.x time without proper plugin releases. We had to download and unpack the kotlin compiler distribution to get the plugin jars. We also applied other plugins (e.g. `dataframe`) but applied it by just mangling the `kotlincOptions`.
> 

they'd need it only if they have classpath conflicts, but we won't know that beforehand, a user could add whatever plugin they'd like, but it would break at runtime (of the compilation process) on the first conflict with probably an IllegalAccessException (same that happened with the KSP plugin

> > My understanding is that the processor id is fixed for KSP , to change it, one would need to replace the default symbol processing logic with something else.
> 
> In that case, we shouldn't have the `kotlinSymbolProcessorId` task at all.
> 
> > we can even annotate it as experimental?
> 
> Makes sense. Please add that annotation for now.

done

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -0,0 +1,156 @@
...
+      kspCompilerArgs,
+      // parameters
+      sourceFiles.map(_.toString())
+    ).flatten
+
+    // currently if we don't delete the already generated sources
+    // several layers are problematic such as the KSP giving a FileAlreadyExists
+    // and test compilation complaining about duplicate classes
+    // TODO maybe find a better way to do this
+    os.remove.all(kspOutputDir)
```
lefou, line 143: This should be obsolete since this is inside the task and the task is not persistent.

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -0,0 +1,156 @@
...
+
+    kotlinWorkerTask().compile(KotlinWorkerTarget.Jvm, compilerArgs)
+
+    os.walk(kspOutputDir).filter(os.isFile).map(PathRef(_))
```
lefou, line 147: IIUC, this collects all files under `T.dest / "generated/ksp/main"`. It's not, what we typically expect from `generatedSources`. Instead we should just return source root dirs, which is probably just `kspOutputDir`.

We already have `allSourceFiles` or `allKotlinSourceFiles` which list all individual files.

Returning files without base dirs may be even not correct in some cases. E.g. for Java files, it's required that their package declarations match their location in the directory layout.

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -0,0 +1,156 @@
...
+  private val kspPluginId: String =
+    "com.google.devtools.ksp.symbol-processing"
+
+  /*
+   * The Kotlin compile task with KSP.
+   * This task should run as part of the [[generatedSources]] task to
+   * so that the generated  sources are in the [[compileClasspath]]
+   * for the main compile task.
+   */
+  private def generateSourcesWithKSP = Task {
```
lefou, line 85: Since generating files is the whole point of this module, this task should be `public`.

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -0,0 +1,156 @@
...
+
+  /**
+   * The symbol processors to be used by the Kotlin compiler.
+   * Default is empty.
+   */
+  def kotlinSymbolProcessors: T[Agg[Dep]] = Task {
+    Agg.empty[Dep]
+  }
+
+  final def kotlinSymbolProcessorsResolved: T[Agg[PathRef]] = Task {
```
lefou, line 65: What is the rational in making this `final`. Overriding this is IMHO the only way to get a processor only available as a JAR working.

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
vaslabs: good catch, I forgot about this, but since I put this in generated sources and not in the compilation stage, it's not needed!

lefou (approved):
Thank you, @vaslabs ! This is ready for merge.

One thing I need to find out is where the example page shows up or if we need to touch the documentation to wire it up. But this can be fixed in a follow up PR.

Issue #4479, opened 2025-02-05
Improve Spring-Boot integration, support for Spring-Boot Assemblies


lefou:
Being a Java framework, Spring-Boot not necessarily requires dedicated support. Yet, The Spring-Boot framework provides features and tooling which could use a better Mill integration, namely Spring-Boot Assemblies.

Mill already comes with Assembly support. Spring-Boot provides an alternative implementation with different trade-offs. For example, in Spring-Boot Assemblies, library dependencies are not unpacked but embedded as-is, which has various benefits:

* no re-packaging
* no special setup for file collisions needed
* no licensing issues due to repackaging
* works for Non-Spring applications as well
* potentially more docker-friendly

This comes at the price, that a Spring-Boot application uses it's own main class as an entry point and a specialized classloader.

It would be nice to have a `SpringBootAssemblyModule`, that supports those features out of the box. There is a prior-art external plugin [`mill-spring-boot`](https://github.com/lefou/mill-spring-boot) which could be integrated for that.

A Spring-Boot specific tutorial should be added and discuss the features and its pros and cons. Also, other tutorials using Spring-Boot should be reviewed if they can benefit from the new module.

lihaoyi:
I think what we can aim for is 4-5 example projects showing off how to use Mill with Spring-Boot to support the most common use cases, that we can include on a dedicated page in the documentation under `Building Java with Mill`. 

We should also try to quantify and state the value prop of using Mill for Spring-Boot projects: are compile times better? is `--watch` more convenient? Easier integrations with non-JVM languages in the build? I'm not actually sure what concrete comparisons would make the most sense here, but I feel we should be able to find some benefits of using Mill rather than Maven or Gradle

lihaoyi:
@lefou are you still working on this? If not I'd like to put a bounty on it to open it up to the community

lefou:
Yes. I have the impl ready as PR, but need to write or adapt an example.

PR #4489, opened 2025-02-06
Add `RepackageModule` to build assemblies based on Spring Boot tools

lefou:
This adds a `mill.javalib.repackage.RepackageModule` which enhances a `Module` with a `repackagedJar` task to produce self-executable JAR files. If the module is a `JavaModule` everything should work out of the box. Otherwise, the use may need to override various tasks, which is analogue to using `RunModule` or `AssemblyModule`.

The Spring-Boot tools are encapsulated into a separate `mill.javalib.spring.boot.SpringBootToolsModule` to avoid instantiating multiple instance. A pre-configured version using the tools version present at built-time is available as `ExternalModule` and used by default.

Fix https://github.com/com-lihaoyi/mill/issues/4479


lefou:
This is an early draft for a `SpringBootAssemblyModule`

This derives from `JavaModule` for now. I tried to just derive from AssemblyModule or RunModule, but could get it right. Since we revisit the module structure anyways, I think this can be tried later again.

The Spring-Boot tools are encampsulated into a separate Module to avoid multiple instance. A pre-setup version using the tools that were available at built-time is available as `ExternalModule` and used by default.

lihaoyi:
@lefou I wonder if we should give `SpringBootAssemblyModule` a more generic name? Having an assembly that doesn't merge classpath entries seems like something that could be generally useful, and not limited to spring boot

lefou:
> @lefou I wonder if we should give `SpringBootAssemblyModule` a more generic name? Having an assembly that doesn't merge classpath entries seems like something that could be generally useful, and not limited to spring boot

Yeah, the name is quite long. And beside the tools coming under the Spring Boot brand, the utility is independent. We could name it `RepackageModule` and the task `repackage`, which is also the name of the Maven goal (`spring-boot:repackage`, whereas in Gradle it's `assemble`). All the spring-boot-tools specific stuff is already handled by the external `SpringBootToolsModule`, so a user doesn't need to manage anything Spring related.

lihaoyi:
Maybe we can all it `RepackageModule` and document it next to the `AssemblyModule` in the `Packaging and Publishing` doc pages, and link to that from the part of the `Java Web Build Examples` page discussing spring-boot?

lihaoyi:
That's assuming that this isn't hard-coded to be used for spring boot of course. If it does do stuff specific to spring-boot, we should just leave the name as it is

lefou:
The repackaging does not require the project to be a Spring project. The documentation is already in the `package and publish` section.

lihaoyi (approved):

lefou:
Updated PR description

alexarchambault:
FYI coursier has very similar support, illustrated by its [`bootstrap` command](https://get-coursier.io/docs/cli-bootstrap) in particular. I hope to open very soon a Mill PR using it in the Mill server, to launch the server with a two class loader setup, in order to workaround issues with the current `MillURLClassLoader#{sharedLoader,sharedPrefixes}`. That should illustrate how to use it.

alexarchambault:
coursier actually even [re-uses some code from springboot](https://github.com/coursier/coursier/tree/836bc113fa849de5d9141d7e6e822f74e1904d70/modules/bootstrap-launcher/src/main/java/coursier/bootstrap/launcher/jar)

PR #4495, opened 2025-02-06
Add file count to `PalantirFormatModule` output

lefou:
Having a quick quantitative feedback what goes on is always nice and helps spotting issues.

For example, I tried to manually format the code base with two different source selectors, from which I assumed they should have the same effect. But they didn't. After this change, the difference would be obvious.

```
> mill mill.javalib.palantirformat.PalantirFormatModule/formatAll --check __:JavaModule.sources

[437] checking format in 109 java sources ...

> mill mill.javalib.palantirformat.PalantirFormatModule/formatAll --check __.sources

[947] checking format in 259 java sources ...
```

Obviously, there are Java sources in non-Java modules.


lefou:
Failing test seems unrelated.

PR #4497, opened 2025-02-06
Use system calls to get terminal size on Linux / Mac (0.12.x)

alexarchambault:
This makes Mill use my [native-terminal](https://github.com/alexarchambault/native-terminal) library (formerly windows-ansi) to get the terminal size. That way, Mill doesn't run `tput` commands every ~100 ms to query the terminal size.


runner/client/src/mill/runner/client/MillProcessLauncher.java
```diff
@@ -248,15 +249,31 @@ static int getTerminalDim(String s, boolean inheritError) throws Exception {
...
+
+    canUseNativeTerminal = canUse;
+  }
+
   static void writeTerminalDims(boolean tputExists, Path serverDir) throws Exception {
     String str;
 
     try {
-      if (java.lang.System.console() == null) str = "0 0";
+      if (!mill.main.client.Util.hasConsole()) str = "0 0";
```
lihaoyi, line 272: It seems we're checking `mill.main.client.Util.hasConsole()` twice here; once in the `static` blockl and once in `writeTerminalDims`. We can probably skip one of those checks right?

lihaoyi:
Trying this out on my macbook seems to add about ~200ms latency (~150ms -> ~350ms), running `./mill dist.installLocal` then `time ~/Github/mill/mill-assembly.jar version` repeatedly in an empty folder. @alexarchambault can you see if you get similar numbers? I wonder where that 200ms is going, but if we can't figure out we may have to stick with the shelling-out-to-`tput` approach even if it's a bit ugly

alexarchambault:
I'm getting the same numbers :/ Printing the duration of some of the calls to get the terminal size, I get things like
```text
First NativeTerminal.getSize duration: 182
Second NativeTerminal.getSize duration: 0
First tput thing duration: 19
Second tput thing duration: 13
```

So once the native things are setup, getting the terminal size is sub-ms.

Upon initialization, jansi reads a binary file (like `libjansi.jnilib`) in its resources, writes it in a temporary directory, and loads it with `System.load`. Manually circumventing the resources / temp dir stuff speeds things up a lot, and gives back the same total time for `time …/mill/mill-assembly.jar version`.

lihaoyi:
If the resource/tempdir stuff is circumventable, does that mean that it isn't necessary for the jni functionality?

alexarchambault:
> If the resource/tempdir stuff is circumventable, does that mean that it isn't necessary for the jni functionality?

Not necessarily, we can load the native library from elsewhere on disk if we want to. I just pushed code that loads it via the coursier cache (this needs https://github.com/coursier/coursier/pull/3272, so this is going to need new coursier and [coursier-interface](https://github.com/coursier/interface) releases). If the native library is already in cache, there's no need to write a temporary file.

From native images, we can also statically link the native library, so that nothing needs to be loaded at runtime. But this needs the JNI library to push static libraries somewhere (like [here](https://repo1.maven.org/maven2/io/github/alexarchambault/tmp/ipcsocket/ipcsocket/1.4.1-aa-5-1/) or [here](https://repo1.maven.org/maven2/io/github/alexarchambault/tmp/libsodiumjni/libsodiumjni/0.0.2/)), and jansi doesn't. It worked fine for other JNI libraries in Scala CLI, although I haven't seen that being done elsewhere.

alexarchambault:
From Java 22 onwards, it should be possible to use [FFM](https://github.com/jline/jline3?tab=readme-ov-file#ffm), JLine is able to for sure. This doesn't need to load custom native libraries upfront (although I have very little experience with it for now).

lihaoyi:
Unpacking the native library and caching it via coursier sounds reasonable. We just need to make sure Coursier is not on the hot path so we don't need to classload all the coursier/scala code when the library is already unpacked.

alexarchambault:
(Seems this is surfacing issues with newer versions of coursier)

lihaoyi:
I'm having trouble trying to `./mill dist.installLocal` this PR on my M1 macbook

```
runner.client.buildInfoMembers java.lang.RuntimeException: Cannot get jansi version from compile class path
    scala.sys.package$.error(package.scala:27)
    build_.runner.package_$client$.$anonfun$buildInfoMembers$7(package.mill:25)
    scala.Option.getOrElse(Option.scala:201)
    build_.runner.package_$client$.$anonfun$buildInfoMembers$3(package.mill:25)
```

runner/client/src/mill/runner/client/MillProcessLauncher.java
```diff
@@ -248,15 +250,81 @@ static int getTerminalDim(String s, boolean inheritError) throws Exception {
 
   private static AtomicReference<String> memoizedTerminalDims = new AtomicReference();
 
+  private static final boolean canUseNativeTerminal;
+
+  static {
```
lihaoyi, line 255: This block is pretty messy, and I'm wondering if there's some way we can make it obvious that the fast path is fast and does not classload any heavy libraries. That is the case now, but the fact that it's not obvious from the code makes it likely to regress accidentally if people touch this code.

Perhaps we could do something like what we do for the JVM versions in https://github.com/com-lihaoyi/mill/blob/392f32d9f3d7e7cb4a4873a1f6c5c337dc63432b/runner/client/src/mill/runner/client/MillProcessLauncher.java#L134-L152, where we do a fast-path check whether a cache/marker file exists on disk, and put the complex code inside the conditional so it only runs in the slow path?

lihaoyi:
Added some rough measurements and this cuts down the time for `getDimensions` from ~5ms to ~0ms, which is nice

runner/client/src/mill/runner/client/MillProcessLauncher.java
alexarchambault: I'm going to try to make that clearer from the code, but it already goes first through a fast path, then through a slower one. The fast path goes from the beginning of the `static {` block up to `File jansiLib = …`. It calls a method on `coursier.paths.CachePath`, but it's a fast thing, that shouldn't load much more coursier classes (the method is [here](https://github.com/coursier/coursier/blob/8e864315c6879cbccab964323eedf21251906a95/modules/paths/src/main/java/coursier/paths/CachePath.java#L120), and it loads [that class](https://github.com/coursier/coursier/blob/8e864315c6879cbccab964323eedf21251906a95/modules/paths/src/main/java/coursier/paths/CoursierPaths.java), which itself loads [directories-jvm](https://github.com/dirs-dev/directories-jvm) classes - it's all pure Java classes).

alexarchambault:
Seems [this](https://github.com/com-lihaoyi/mill/blob/3e2f5814b4524792c2c540599cddbd4e58240aad/integration/feature/init/src/MillInitMavenTests.scala#L37) puts jansi in `~/.ivy2/local`…

alexarchambault:
It seems these consistently fail on CI, but not locally, I'm not sure why
```text
integration.feature[output-directory].packaged.server.test 2 tests failed: 
  mill.integration.OutputDirectoryTests mill.integration.OutputDirectoryTests.Output directory elsewhere in workspace
  mill.integration.OutputDirectoryTests mill.integration.OutputDirectoryTests.Output directory outside workspace
```

lihaoyi:
> It seems these consistently fail on CI, but not locally, I'm not sure why
> 
> ```
> integration.feature[output-directory].packaged.server.test 2 tests failed: 
>   mill.integration.OutputDirectoryTests mill.integration.OutputDirectoryTests.Output directory elsewhere in workspace
>   mill.integration.OutputDirectoryTests mill.integration.OutputDirectoryTests.Output directory outside workspace
> ```

Not sure,  but you can use the commented-out `debug` section of run-tests.yml to try and minimize what is running in CI and figure add logging to debug it

runner/client/src/mill/runner/client/MillProcessLauncher.java
lihaoyi: Yes I'm aware it currently goes through all lightweight JVM classes, and the manual startup-time benchmarks verify it. But we should refactor the code to make it obvious at a glance rather than having to dig through third-party libraries, otherwise it is prone to regress when someone overlooks it and makes changes later

runner/client/src/mill/runner/client/MillProcessLauncher.java
alexarchambault: `hasConsole` caches its result now, so calling it twice shouldn't incur any cost

runner/client/src/mill/runner/client/MillProcessLauncher.java
alexarchambault: Just put the loading logic aside in a separate class, with two methods, `tryLoadFast` and `loadSlow` (did you mean that kind of thing?)

PR #4500, opened 2025-02-07
More cleanup

lihaoyi:


contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBModule.scala
```diff
@@ -113,7 +113,7 @@ trait ScalaPBModule extends ScalaModule {
                 if (os.exists(protoDest))
                   Task.log.error(s"Warning: Overwriting ${dest} / ${os.SubPath(entry.getName)} ...")
                 Using.resource(os.write.over.outputStream(protoDest, createFolders = true)) { os =>
-                  IO.stream(zip, os)
+                  _root_.os.Internals.transfer(zip, os)
```
lefou, line 116: Wondering, why this API is called internals. Sound like it should not be used outside. OTOS, we have a close channel to os-lib. :-D

main/api/src/mill/api/DummyStreams.scala
```diff
@@ -16,23 +16,3 @@ object DummyOutputStream extends OutputStream {
   override def write(b: Array[Byte], off: Int, len: Int): Unit = ()
 }
 
-trait StreamSupport {
```
lefou, line 19: We should deprecated it in `0.12.x` branch.

contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBModule.scala
lihaoyi: yeah, we're already using all sorts of experimental stuff from that library, so I guess it's fine haha

PR #4501, opened 2025-02-07
Native Client

Quafadas:
This would be the first meaningful  step toward #4016 

It should do the heavy lifting of porting the 
- client .java files 
- JUnit tests  

to 
- .scala
- Utest

With the idea being, that we can then cross compile a SN client. Currently, tests on native will fail, so they are not enabled. Waiting on SN 0.5.7.

The idea, is that this PR should keep the CI green, as it _should_ integrate seamlessly.


Quafadas:
Well, this hasn't gone very well. it doesn't appear to be possible to do this "non-intrusively"... 

My current problem, is that I can't seem to override a test module. 

```
      override lazy val test = new MillScalaTests with SharedTests { // This works
      // object test2 extends MillScalaTests with SharedTests { // This works
        def scalaVersion = build.Deps.scalaVersion
        override def sources = T.sources {
          Seq(millSourcePath / os.up / os.up / "test" ).map(PathRef(_)) ++
          super.sources()
        }
        override def resources = T.sources {
          Seq(millSourcePath / os.up / os.up / "test" / "resources" ).map(PathRef(_)) ++
          super.resources()
        }
      }
```
The `test2` definition works, but I'm not sure that's terribly idiomatic. 

```
[build.mill-61] [info] compiling 4 Scala sources to /Users/simon/Code/mill-1/out/mill-build/compile.dest/classes ...
[build.mill-61] [error] /Users/simon/Code/mill-1/out/mill-build/generateScriptSources.dest/build_/build.mill:31:92: class type required but mill.define.Module.BaseClass with build_.main.package_.client.jvm.MillScalaTests with build_.main.package_.client.SharedTests found
[build.mill-61] [error]   override lazy val millDiscover: _root_.mill.define.Discover = _root_.mill.define.Discover[this.type]
[build.mill-61] [error]                                                                                            ^
[build.mill-61] [error] one error found
[65/65] ============================== main.clien

```
is the error without it. 

Aside from that the CI fails... but I don't understand why. I must have mucked something up pretty badly, but I fear the PR is too large to sift through and figure out why.

Quafadas:
If anyone is willing to have a look, I would need help. It comiles, some tests pass, but apparently  I also but have also broken something quite fundamental which was what I hoped to avoid. 

`mill mill.main.server.ClientServerTests` fails for example on client interrupt, but I can't see why.

sideeffffect:
Just a wild guess, but have you tried rewriting .java files to .scala one by one and running test after each? Then you could identify the place/cause which breaks the test, no?

lihaoyi:
Closing due to inactivity

PR #4503, opened 2025-02-07
Add scalalib.ApeModule to produce APE (actually portable executable) assembly jar

kiendang:
Add an `ApeModule` to support producing actually portable executable assembly jars.

```scala
object foo extends JavaModule with scalalib.ApeModule
```

```sh
./mill show foo.assemblyApe
```

`out.jar` would run cross-platform (tested with arm Mac, x86 Linux and Windows (on cmd) with [this example hello world project](https://mill-build.org/blog/5-executable-jars.html#_trying_out_mills_executable_assemblies)).


scalalib/src/mill/scalalib/ApeModule.scala
```diff
@@ -0,0 +1,111 @@
+package mill.scalalib
+
+import mill.api.Loose.Agg
+import mill.api.{PathRef, Result}
+import mill.define.{Target => T, _}
+
+trait ApeModule extends mill.Module with AssemblyModule {
+  def cosmoccVersion: T[String] = Task { "" }
+
+  def cosmoccZip: T[PathRef] = Task {
```
lefou, line 10: This task should be at least `persistent`, to avoid re-downloads. Better would be to use some machine-wide cache. (https://github.com/com-lihaoyi/mill/issues/3930)

scalalib/src/mill/scalalib/ApeModule.scala
```diff
@@ -0,0 +1,111 @@
...
+    )
+    PathRef(Task.dest / "cosmocc.zip")
+  }
+
+  def cosmocc: T[PathRef] = Task {
+    os.call(("unzip", cosmoccZip().path, "-d", Task.dest / "cosmocc"))
+    PathRef(Task.dest / "cosmocc" / "bin" / "cosmocc")
+  }
+
+  def launcherScript: T[Option[String]] = Task {
```
lefou, line 24: This needs a prefix. Either `ape` or `cosmo`.

scalalib/src/mill/scalalib/ApeModule.scala
```diff
@@ -0,0 +1,111 @@
+package mill.scalalib
+
+import mill.api.Loose.Agg
+import mill.api.{PathRef, Result}
+import mill.define.{Target => T, _}
+
+trait ApeModule extends mill.Module with AssemblyModule {
```
lefou, line 7: How about namings this `CosmoModule` ?

scalalib/src/mill/scalalib/ApeModule.scala
```diff
@@ -0,0 +1,111 @@
...
+           |  def prependShellScript = ""
+           |""".stripMargin,
+        Some(created.pathRef)
+      )
+    } else {
+      Result.Success(created.pathRef)
+    }
+  }
+
+  def assemblyApe: T[PathRef] = Task {
```
lefou, line 108: `cosmoAssembly` or `apeAssembly`. Depends on which name we choose above. Should be consistent though.

scalalib/src/mill/scalalib/Assembly.scala
```diff
@@ -192,11 +192,11 @@ object Assembly {
     ).pathRef
   }
 
-  def create(
+  private[mill] def create0(
       destJar: os.Path,
       inputPaths: Agg[os.Path],
       manifest: mill.api.JarManifest = mill.api.JarManifest.MillDefault,
-      prependShellScript: Option[String] = None,
+      prepend: Option[os.Source] = None,
```
lefou, line 199: I assume you need a byte array instead of a string. Woud be a `Option[String|Array[Byte]]` sufficient? Or do you expect large data here? A `os.Source` feels a bit too generic.

lefou (commented):
I can't tell what's the best name, but `Ape` feels a bit cryptic. It's IMHO not a well established term. Since we seem to use just one  tool/technology jere, we might as well name the module after it. This is what we do in many other modules and it also leaves room for alternative implementations for APE. `CosmoModule` might be the better choice.

scalalib/src/mill/scalalib/ApeModule.scala
kiendang: Agree. A machine-wide cache is best here. Downloading a >400MB `cosmocc.zip` for every project is not ideal.

scalalib/src/mill/scalalib/Assembly.scala
kiendang: `Option[String|Array[Byte]]` would work as well. The data is around 300K in this case.

kiendang:
@lefou Thanks for the review. Admittedly this is just a POC PR I quickly drafted up to demonstrate the idea. I haven't thought too much about organizing things yet, which obviously needs fleshing out if this is to be developed further.

Just want to mention a few drawbacks to this approach
- The prepended compiled binary is around 300KB, quite a bit larger than the current shell script.
- As I have tested, the assembly `out.jar` runs on arm Mac, arm and x86 Linux. On x86 Windows it runs on CMD, but doesn't on PowerShell (it just doesn't show anything). I'm not a Windows user so not yet sure what went wrong or even where to look. I'm using the same hello world example project [here](https://mill-build.org/blog/5-executable-jars.html#_trying_out_mills_executable_assemblies).
- Even binaries compiled with cosmocc have hiccups with a couple of environments (https://github.com/jart/cosmopolitan?tab=readme-ov-file#platform-notes)
- My launcher script use `execvp` which evokes the `execve` syscall. According to [this table](https://justine.lol/cosmopolitan/functions.html). `execve` has some problem on Windows, but that might have been solved in [a recent update](https://github.com/jart/cosmopolitan/commit/42a3bb729aecc1e45d2e560d8a26712dc3bf7dc5).

lihaoyi:
In terms of naming, I think `ActuallyPortableExecutableModule` would be best. It's bit of an unusual/niche thing so having the name be verbose is probably better than conciseness.

300kb is fine; any Scala program gets 4-5mb from scala-library anyway, and most Java programs will have similar dependency trees. Mill's own executable assemblies bundle Coursier with Graal for a >60mb jar prepend and it works great

We need to see what's going on with Powershell to see if it can be resolved. Currently Powershell is the default shell for Windows Terminal, and the whole point of APEs is to work seamlessly on Windows environments where previously you would need to rename `out.jar` to `out.bat`. If it's technically impossible at the moment I don't mind closing out the bounty regardless, but we should dig in a bit to see why it doesn't work.

kiendang:
@lihaoyi I figured out the problem with PowerShell. It looks like PowerShell strictly follows `PATHEXT`. It won't run any binary if the file extension is not included in `PATHEXT`. Changing the file name to `out.jar.com` works. This means that we would have this naming problem on PowerShell regardless of the file content. It's impossible to make `./out.jar` run as is on PowerShell, be it APE or any other format, unless the user already set `$Env:PATHEXT += ';.JAR'`. Now I understand why in most of Cosmopolitan examples the files have the `.com` extension.

kiendang:
[This section on `PATHEXT`](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_environment_variables?view=powershell-7.5#path-information) is interesting

> To ensure that scripts for another scripting language run in the current console session, add the file extension used by the scripting language. For example, to run Python scripts in the current console, add the .py extension to the environment variable. For Windows to support the .py extension as an executable file you must register the file extension using the ftype and assoc commands of the CMD command shell.

and I checked

```pwsh
cmd /c ftype jarfile

jarfile="C:\Program Files\Java\jdk-17\bin\javaw.exe" -jar "%1" %*
```

This means that if the user installs java using the Windows installer, which registers jarfile with ftype, and we could change the assembly zip to be able to run `java -jar out.jar` instead of `java -cp out.jar <main class>` then we could do `./out.jar` on PowerShell. That's a big condition though. I suppose it wouldn't work if you get java via other methods, i.e. download the archive, through coursier, ...

lefou:
> In terms of naming, I think `ActuallyPortableExecutableModule` would be best

I don't like that or a shorter version `ProtableExecModule`. We should name it after the tool at work, `CosmoModule`. We also have `ScalafmtModule`, `ProguardModule` or `JmhModule` instead of `FormatModule`, `ShrinkAndObfuscateModule` or `BenchmarkModule`.

lihaoyi:
@kiendang if the extension matters, what if we name the output file `.exe`? That should be executable on all windows shells right? And since Linux/Mac doesn't look at extensions, it should work there as well?

sake92:
This is the default for PATHEXT: `.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC`
So any of these should work: mill.com, mill.exe, mill.cmd.
Depends what we think makes most sense. Mill.cmd feels like middle ground to me.
Mill.com sounds like a website, mill.exe sounds windows specific.. might confuse people

kiendang:
@lihaoyi Yup that would work.

@sake92 I actually think `.EXE` is the more appropriate here. The APE file format is a binary file format. It even starts with `MZ` which is the executable format for `.EXE` files. `.CMD` would be shell script files to be run by the Windows CMD shell. I don't think using either extension would mean much on Linux/Unix, but on Windows using `.CMD` instead of `.EXE` extension for these binaries would not be fitting.



@ all, I have a bit of problem doing this
>One end-to-end test that should be done is to build the Mill dist.executable jar with an APE prefix

One way of enabling `dist.ape.executable` with the least amount of code I can think of is having `dist.ape` "extends" `dist` via an export clause

`dist/build.mill`
```scala
package build.dist

object `package` extends RootModule with InstallModule {
  ...
  
  object ape
    extends RootModule
    with InstallModule
    with mill.scalalib.ApeModule {

    export `package`.{executableRaw as _, *}

    override def executableRaw = ...
  }
}
```

but looks like mill doesn't support export clauses? If I do `./mill show dist.ape.executable` I got

```
1 tasks failed
generateScriptSources dist/package.mill:334:21 expected "}"
    export `package`.{executableRaw as _, *}
                    ^
dist/package.mill:334:21 expected "}"
    export `package`.{executableRaw as _, *}
```

also inside `dist/build.mill` I can't seem able to access the `mill.scalalib.ApeModule` trait I added. Even putting something simple like `trait x extends mill.scalalib.ApeModule` would give me

```
[build.mill-61/65] compile
[build.mill-61] [info] compiling 1 Scala source to /home/kien/mill/out/mill-build/compile.dest/classes ...
[build.mill-61] [error] /home/kien/mill/dist/package.mill:333:35: type ApeModule is not a member of package mill.scalalib
[build.mill-61] [error] did you mean BomModule, RunModule, or SbtModule?
[build.mill-61] [error]   trait ape extends mill.scalalib.ApeModule {
[build.mill-61] [error]                                   ^
[build.mill-61] [error] one error found
```

Not sure if I miss something.

lihaoyi:
@kiendang  Let's go with @lefou's suggestion of `CosmoModule`?

I'm not sure about `export`, but your snippet looks wrong. You should not need to have nested `RootModule`s, and you shouldn't need to use `export`s to share tasks between modules as you should use `trait`s instead.

However, for now you can't use stuff defined in your PR in Mill's own `{build,package}.mill` scripts until we publish a new version and update `.config/mill-version`. So it should be fine to just exercise your code in unit/integration/example tests, and we can optionally integrate it into mill proper in a follow up

kiendang:
Added the tests to `integration/feature/cosmopolitan`.

I moved `mill.scalalib.ApeModule` to `mill.scalalib.cosmopolitan.CosmopolitanModule`. Almost everywhere in the Cosmopolitan official repo and documentation it's referred to as Cosmopolitan instead of Cosmo (which I think is just a informal shortened name) so I think it's a better fit. If you think this is too long I'll change it to `CosmoModule` (sorry for the bikeshedding).

Right now where should I cache the Cosmo toolchain? I see #3930 but looks like nothing is standardized yet?

lihaoyi:
Yes there's no standard global cache yet, so caching i  the task.dest folder is fine for now

scalalib/src/mill/scalalib/cosmopolitan/CosmopolitanModule.scala
```diff
@@ -0,0 +1,131 @@
...
+  def cosmocc: T[PathRef] = Task(persistent = true) {
+    val version = if (cosmoccVersion().isEmpty) "" else s"-${cosmoccVersion()}"
+    os.write(
+      Task.dest / "cosmocc.zip",
+      requests.get.stream(s"https://cosmo.zip/pub/cosmocc/cosmocc${version}.zip")
+    )
+
+    os.call(("unzip", Task.dest / "cosmocc.zip", "-d", Task.dest / "cosmocc"))
+    PathRef(Task.dest / "cosmocc" / "bin" / "cosmocc")
+  }
```
lefou, line 21: The idea behind the `persistent = true` is, that `T.dest` doesn't get automatically removed. We still need to check, if the wanted zip-file is present and only download it if not.

scalalib/src/mill/scalalib/cosmopolitan/CosmopolitanModule.scala
```diff
@@ -0,0 +1,131 @@
...
+    // See https://github.com/com-lihaoyi/mill/pull/2655#issuecomment-1672468284
+    val problematicEntryCount = 65535
+    if (
+      prepend.isDefined &&
+      (upstream.addedEntries + created.addedEntries) > problematicEntryCount
+    ) {
+      Result.Failure(
+        s"""The created assembly jar contains more than ${problematicEntryCount} ZIP entries.
+           |JARs of that size are known to not work correctly with a prepended shell script.
+           |Either reduce the entries count of the assembly or disable the prepended shell script with:
```
lefou, line 122: We should reword this, since it's mentioning of "shell script" is misleading. Best make this specific, since the user doesn't what to prepend a shell script, she wants to build a universal/portable executable.

lefou: It's probably warranted to just fail the task in case the JAR has to many entries.

kiendang: Good catch!

scalalib/src/mill/scalalib/cosmopolitan/CosmopolitanModule.scala
```diff
@@ -0,0 +1,135 @@
...
+    val zip = Task.dest / s"${versionedCosmocc}.zip"
+    val dest = Task.dest / versionedCosmocc / "bin" / "cosmocc"
+
+    if (!os.exists(dest)) {
+      os.write.over(
+        zip,
+        requests.get.stream(s"https://cosmo.zip/pub/cosmocc/${versionedCosmocc}.zip")
+      )
+      os.remove.all(Task.dest / versionedCosmocc)
+      os.call(("unzip", zip, "-d", Task.dest / versionedCosmocc))
```
lefou, line 24: Let's use `os.unzip` instead.

lefou (commented):
The name you settled with is fine. But I'd like that have some more consistent naming. We now have the `Cosmopolitan`, `cosmo` and `ape` as prefix. Would be nice to get rid of `ape` in favor to `cosmo`.

scalalib/src/mill/scalalib/cosmopolitan/CosmopolitanModule.scala
kiendang: I did try `os.unzip` but it doesn't preserve permissions. Would have to have an extra step to give `cosmocc` executable permission. `cosmoc.zip` also contains other files. I'm not sure if not having the original permissions changes anything (though unlikely).

lihaoyi: Let's go with `os.zip` then and `os.perms.set` to fix anything necessary, and open an issue on OS-Lib about `os.unzip` not preserving permissions so we can fix that

scalalib/src/mill/scalalib/Assembly.scala
kiendang: Basic question: What is the simplest way to pass a `x: String | Array[Byte]` to `os.write`? Looks like there is no implicit conversion for unions. `os.write(..., x)` does not work. `x: os.Source`, `x.asInstanceOf[os.Source]` doesn't work either. Is there any other way that's less clunky than `x match { case x: String => os.write(..., x); case x: Array[Byte] => os.write(..., x) }`?

lefou: I don't know the answer. Could be, this isn't fully implemented in upickle. But we could also either fallback to always use a byte array, which sounds a bit inconvenient for the other cases, or go with a `Option[Either[Array[Byte],String]]` for now.

lihaoyi: I think just using byte array is fine. Strings can be converted with just `.getBytes`, so it's no big hardship. Definitely easier than wrapping in `Either`s and `Left`s and `Right`s

scalalib/src/mill/scalalib/cosmopolitan/CosmopolitanModule.scala
lihaoyi: ok, using a subprocess is probably fine for now, but can  you open an issie upstream

kiendang:
One more thing. The current shell script approach uses the `JAVA_OPTS` env var: `java $JAVA_OPTS -cp ...`. It's not possible to do the exact same thing with C's `execvp`. `execvp("java", {"java", getenv("JAVA_OPTS", ...)})` just doesn't work if `JAVA_OPTS` contains multiple values. Parsing and splitting `JAVA_OPTS` ourselves is out of the question.

The cleanest approach is to use [java argument files](https://docs.oracle.com/en/java/javase/17/docs/specs/man/java.html#java-command-line-argument-files), i.e. write the values of `JAVA_OPTS` to a temp file then do `java @<path/to/temp/file> ...`. Would have the exact same effect. However argument files require Java 9 minimum.

Another approach is to append `JAVA_OPTS` to [`JAVA_TOOL_OPTIONS`](https://docs.oracle.com/javase/8/docs/technotes/guides/troubleshoot/envvars002.html) or [`JDK_JAVA_OPTIONS`](https://docs.oracle.com/en/java/javase/17/docs/specs/man/java.html#using-the-jdk_java_options-launcher-environment-variable) which will be picked up by `java`. The former works with Java 8.

There is a very small edge case with appending to `JAVA_TOOL_OPTIONS` approach in which the behavior differs from the current shell script approach though. Command line arguments take precedence over `JDK_JAVA_OPTIONS`, which in turns takes precedence over `JAVA_TOOL_OPTIONS`, so in case the same argument is set in both `JAVA_OPTS` and `JDK_JAVA_OPTIONS`, the value in `JDK_JAVA_OPTIONS` will win whereas with the current shell script approach the value in `JAVA_OPTS` wins. This looks like a rare edge case though.

In summary there are 3 approaches

- Argument files. Cleanest and ensures same behavior with the current shell script approach, but requires Java 9.
- Append to `JDK_JAVA_OPTIONS`. Also requires Java 9, so should just use argument files instead.
- Append to `JAVA_TOOL_OPTIONS`. Works with Java 8, but there's a small edge case where the behavior differs from the current shell script approach.

sideeffffect:
> Parsing and splitting `JAVA_OPTS` ourselves is out of the question.

Maybe it's possible after all:

```c
#include <stdio.h>
#include <stdlib.h>
#include <wordexp.h>

int main() {
    char *env_var = getenv("JAVA_OPTS");
    if (!env_var) {
        printf("JAVA_OPTS is not set.\n");
        return 1;
    }

    wordexp_t p;
    if (wordexp(env_var, &p, 0) != 0) {
        perror("wordexp failed");
        return 1;
    }

    // Print parsed options
    printf("Parsed JAVA_OPTS:\n");
    for (size_t i = 0; i < p.we_wordc; i++) {
        printf("  Arg[%zu]: %s\n", i, p.we_wordv[i]);
    }

    wordfree(&p); // Free allocated memory
    return 0;
}
```

Courtesy of ChatGPT
image


While `wordfree` doesn't seem to be available in Cosmpolitan, there are permissively-licensed implementations:
 * https://github.com/leleliu008/libwordexp

https://www.gnu.org/software/libc/manual/html_node/Wordexp-Example.html

kiendang:
@sideeffffect Thanks! Didn't know about `wordexp`. Looks quite handy. I don't think this is something that warrants an external dependency though. I'd rather just let the jvm handles it.

kiendang:
Trying to make running `cosmocc` on Windows but this turned out to be difficult. Running it inside the mill task keeps failing despite running the same command myself in PowerShell works. However, besides that, I think supporting building an APE assembly on Windows would be hard to maintain because:

- `cosmocc` is a shell script which [requires a POSIX-compatible shell to run](https://github.com/jart/cosmopolitan/tree/master/tool/cosmocc#gotchas). Cosmo does provide one (which needs to be downloaded separately from the toolchain), but `cosmocc` also requires other utilities such as `rm`, `mv`, ... These are provided by Cosmo, but are not packaged into the toolchain and need to be downloaded separately ([https://cosmo.zip/pub/cosmos/bin](https://cosmo.zip/pub/cosmos/bin/dash)). Different versions of `cosmocc` might require different tools. For example the doc says it depends on `mkdir`, `less`, `cat`, and `kill` (non-exhaustive) but I found for the latest `cosmocc` version I need `mv` and `rm`.
- Building on Windows also seems lightly supported, not a top priority based on what I gathered from Discord. Means it works, but requires tinkering.
- From experience, running `cosmocc` on Windows seems very brittle. Even when it works for me when I run it in PowerShell, it also throws SIGTERM randomly (even when it does produces a working binary). Things like long path names also fail the compilation.

I think we can provide support for building APE assemblies on Linux/Unix only for now. These assemblies would still run on Windows. Not sure how to set up the CI for that though, building the assembly on Linux then testing on Windows.

lihaoyi:
@kiendang I think skipping windows support for building APEs is fine. For example tests, you can use the `# mac/linux` suffix for commands that should be skipped on windows, which for APE I guess would be all of them. 

I think no need to have automated tests for building on one OS and running on another; we can just trust the upstream tool to do what it says on the box

sideeffffect:
@kiendang 
> I'd rather just let the jvm handle it.

So that means we'd have to go with `JAVA_TOOL_OPTIONS`, right? Maybe we can get the original behavior based on whether we pre-pend or a-pend?

kiendang:
@sideeffffect yes either `JAVA_TOOL_OPTIONS` or argument files, just passing the string to the JVM and let it do its thing.

@lihaoyi what do you think about the `JAVA_OPTS` thing (https://github.com/com-lihaoyi/mill/pull/4503#issuecomment-2658673612)?

lihaoyi:
@kiendang I think let's go with argument files? Mill only supports Java >=11 now anyway, so needing >=9 is not an issue

kiendang:
Should be done now. Ready for review 👍

scalalib/src/mill/scalalib/cosmo/CosmoModule.scala
```diff
@@ -0,0 +1,155 @@
...
+    val zip = Task.dest / s"${versionedCosmocc}.zip"
+    val dest = Task.dest / versionedCosmocc / "bin" / "cosmocc"
+
+    if (!os.exists(dest)) {
+      os.write.over(
+        zip,
+        requests.get.stream(s"https://cosmo.zip/pub/cosmocc/${versionedCosmocc}.zip")
+      )
+      os.remove.all(Task.dest / versionedCosmocc)
+      os.call(("unzip", zip, "-d", Task.dest / versionedCosmocc))
```
lihaoyi, line 31: Can you update the issue upstream in OS-Lib with the details of what doesn't work, and link the issue in a comment here?

scalalib/src/mill/scalalib/cosmo/CosmoModule.scala
```diff
@@ -0,0 +1,155 @@
...
+        s"""char* s${i};
+           |asprintf(&s${i}, "${arg}", argv[0]);
+           |all_argv[${i}] = (char*)malloc((strlen(s${i}) + 1) * sizeof(char));
+           |strcpy(all_argv[${i}], s${i});
+        """.stripMargin
+      }.mkString("\n");
+
+    val preArgvSize = size + 3;
+
+    s"""#include <stdlib.h>
```
lihaoyi, line 62: Let's move this to a `launcher.c` file in `scalalib/resources`; It might not be exactly valid C due to the template interpolations, but at least people editing it in IDEs would get some syntax highlighting to help

scalalib/src/mill/scalalib/cosmo/CosmoModule.scala
```diff
@@ -0,0 +1,155 @@
...
+    if (created.entries > problematicEntryCount) {
+      Result.Failure(
+        s"""The created assembly jar contains more than ${problematicEntryCount} ZIP entries.
+           |Prepended JARs of that size are known to not work correctly.
+         """.stripMargin
+      )
+    } else {
+      Result.Success(created.pathRef)
+    }
+  }
```
lihaoyi, line 154: Let's re-use the existing `assembly` task; it seems only a few things need to be tweaked (the `prepend` and the `destJar` name), so those can be moved to separate tasks that `CosmoModule` overrides

example/scalalib/publishing/9-cosmo-assembly/build.mill
```diff
@@ -0,0 +1,20 @@
...
+package build
+import mill._, scalalib._
+
+object foo extends ScalaModule with cosmo.CosmoModule {
+  def scalaVersion = "2.13.16"
+}
+//// SNIPPET:END
+
+// This example uses the `scalalib.cosmo.CosmoModule` to build a cross-platform executable assembly by prepending
+// a launcher in the Actually Portable Executable (APE) format compiled with the Cosmopolitan's toolchain.
```
lihaoyi, line 11: This explanation needs to be fleshed out. What does cross-platform mean? What platforms does it support? How does it work? Where can a user go to read more about this?

example/scalalib/publishing/9-cosmo-assembly/build.mill
```diff
@@ -0,0 +1,20 @@
+//// SNIPPET:BUILD
```
lihaoyi, line 1: No need these `SNIPPET` directives since we're not sharing any code with other templates

example/scalalib/publishing/9-cosmo-assembly/build.mill
```diff
@@ -0,0 +1,20 @@
...
+package build
+import mill._, scalalib._
+
+object foo extends ScalaModule with cosmo.CosmoModule {
```
lihaoyi, line 5: Let's make this example a `JavaModule`, move it to `example/javalib/`, include it in `website/docs/modules/ROOT/pages/javalib/publishing.adoc`, and link to that section in `website/docs/modules/ROOT/pages/scalalib/publishing.adoc` and `website/docs/modules/ROOT/pages/kotlinlib/publishing.adoc`

lihaoyi:
Looks good to me overall. Just some nits and docs and I think we can merge it

integration/feature/cosmo/resources/build.mill
```diff
@@ -0,0 +1,12 @@
+package build
+
+import mill._, javalib._, scalalib._
```
lihaoyi, line 3: Let's put an alias `javalib.cosmo` so people don't need to import it from `scalalib._` in Java projects

scalalib/src/mill/scalalib/cosmo/CosmoModule.scala
kiendang: Use `os.unzip` now. Turns out I need to `+x` files inside `libexec` too, not just `bin`. Also commented on the issue with more details.

scalalib/src/mill/scalalib/cosmo/CosmoModule.scala
kiendang: After refactoring `Assembly.create` is not used anywhere anymore, all replaced with `Assembly.create0`, but I guess we keep it around for backward compat?

lihaoyi:
Just tried this on my macbook, seems to work!

Perhaps one last request @kiendang: let's move `def cosmocc` from `CosmoModule` into the `CosmoModule` companion object, by making it an `ExternalModule`. e.g. see how `mill.scalalib.ZincWorkerModule` works. That will ensure that the heavyweight `cosmocc` download only happens once per Mill workspace, even if multiple `CosmoModule`s are defined

lihaoyi:
Also managed to take an executable built on my macbook and run it on a amazon linux EC2 machine, seems to work great

kiendang:
Done moving `cosmoc` to `ExternalModule`, but doesn't this prevent users from overriding the task, e.g. downloading another version, using their already downloaded copy on disk, ...?

Another thing, in the shell script approach `java` is replaced with `"$JAVA_HOME"/bin/java` if `JAVA_HOME` is set. I didn't do that here since that's how `java` behaves anyway. If you have `java` on PATH pointing to java 8 and `JAVA_HOME` set to java 11, running `java` would invoke java 11 anyway. The only case where there would be a difference is when `JAVA_HOME` is set to something invalid, in which case `java` would be whatever on `PATH` while the shell script would fail.

lihaoyi:
Thanks @kiendang ! I'll leave the PR unmerged for now since we don't have https://github.com/com-lihaoyi/mill/issues/3930 yet, and without it the huge cosmo distributable will slow down test times locally and in CI by quite a lot. But I'll consider the ticket completed and wire you the bounty using the details we used before

PR #4506, opened 2025-02-07
feat: support test case selection in TestNG framework

kasiaMarek:
needed for: https://github.com/scalameta/metals/pull/7200

This allows to select and run single tests in Metals using TestNGFramework.
TestNGSingleTest

Previously only whole suite selection was possible.

Tested it manually as visible on the GIF above.

### DIFF
--- a/contrib/testng/src/mill/testng/ResultEvent.java
+++ b/contrib/testng/src/mill/testng/ResultEvent.java
@@ -27,7 +27,7 @@ public class ResultEvent {
       }
 
       public Selector selector() {
-        return new SuiteSelector();
+        return new TestSelector(testNGResult.getName());
       }
 
       public Status status() {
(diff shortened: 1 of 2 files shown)


lihaoyi:
Looks reasonable, could you add a test case in https://github.com/com-lihaoyi/mill/blob/main/contrib/testng/test/src/mill/testng/TestNGTests.scala?

kasiaMarek:
> Looks reasonable, could you add a test case in https://github.com/com-lihaoyi/mill/blob/main/contrib/testng/test/src/mill/testng/TestNGTests.scala?

I'm honestly not sure how to do it reasonably. An execution of a test using mill passes only specific classes to select (uses only `SuiteSelector`). Using other selectors like `TestSelector` is not needed, since such a selection can be passed as an argument to test execution (e.g. using `-method` for TestNG). This change is only needed when bloop or Metals uses mill's implementation of `sbt.testing.Framework` for `TestNG`.

lihaoyi:
@kasiaMarek ok, in that case can you update the description to describe how this was tested? manual testing is fine

lihaoyi:
Thanks @kasiaMarek !

PR #4515, opened 2025-02-08
Enable fatal warnings

lihaoyi:
The existing warnings were dealt with by a mix of fixing and silencing. We can tweak the set of silenced warnings later as necessary, but this is a significant step in ensuring all warnings are dealt with one way or another, and should prevent them uselessly piling up in the console logs


build.mill
```diff
@@ -477,14 +477,21 @@ trait MillScalaModule extends ScalaModule with MillJavaModule with ScalafixModul
...
 
   def scaladocOptions = Seq("-Xsource:3")
+
   def scalacOptions =
     super.scalacOptions() ++ Seq(
       "-deprecation",
       "-feature"
     ) ++ (
       if (ZincWorkerUtil.isScala3(scalaVersion())) Seq(
         // "-Werror",
```
som-snytt, line 487: This is the progressive spelling of "warnings as error".

build.mill
```diff
@@ -477,14 +477,21 @@ trait MillScalaModule extends ScalaModule with MillJavaModule with ScalafixModul
...
-        "-Wunused:all"
+        "-Wunused:all",
+        "-Xfatal-warnings",
+        "-Wconf:msg=An existential type that came from a Scala-2 classfile:silent",
+        "-Wconf:msg=import scala.language.implicitConversions:silent",
+        "-Wconf:msg=IterableOnceExtensionMethods:silent",
+        "-Wconf:msg=is deprecated:silent",
```
som-snytt, line 493: `-Wconf:cat=deprecation:s` is one of a few portable "categories".

Issue #4121, opened 2024-12-12
Enforce that `Task.Sources` cannot have upstream task dependencies


lihaoyi:
This would simplify the data model considerably, make things easier to reason about internally, and avoid common mistakes (e.g. https://github.com/com-lihaoyi/mill/pull/4120). e.g. for watch-and-rerun or selective execution, it would mean we can just inspect the hashes of the input tasks in isolation, rather than running through an entire evaluation workflow to get their values

I wanted to do this a while back, but in the past the fact that `T` and `T.sources` were different types meant that you had to override `T.sources` with `T.sources`, so sources needed to support depending on tasks to allow the common override-and-add workflow. Now, we can simply override `Task.Sources` with `Task`, so we no longer need `Task.Sources` to support upstream task dependencies

This is a breaking change and would need to wait for 0.13.0

lefou:
So how do I add new source dirs (that should be watched) to an inherited sources task then? 

```scala
object Common {
  def sources = Task.Sources(millSourcePath / "src" )
}
object foo extends Common {
  def sources = Task.Sources {
    super.sources() ++ Seq(PathRef(millSourcePath / "src-foo"))
  }
}
```

lihaoyi:
Something like this:

```scala
object Common {
  def sources = Task.Sources(millSourcePath / "src" )
}
object foo extends Common {
  def newSources = Task.sources(millSourcePath / "src-foo")
  def sources = Task { super.sources() ++ newSources() }
}
```

A bit more verbose than the current setup, but not too bad

lefou:
That's really inconvenient. Since this change is only motivated by the implementation impact, maybe we can restrict the `super` call to when `super` is a `Task.Sources`, so that the outcome can be merged easily?

lihaoyi:
The dynamism required for those can still be done with the API above I think. They just require `crossValue`/`crossScalaVersion` and the `millSourcePath`, all of which can be computed at resolution time without needing to evaluate any tasks.

Not sure if there are other use cases that do require task-results to do their computation, but at least those discussed here do not.

I'd like to remove the ability of `Task.Input`s to depend on upstream tasks as well.

Overall these issues we'll need to sort out during implementation to see if it hits any blockers, but at least for now it still seems feasible

lefou:
[...]

PR #4524, opened 2025-02-10
Restrict sources and inputs to not accept upstream tasks

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/4121

This PR makes `Task.Source`, `Task.Sources`, and `Task.Input` to not accept any upstream tasks. Sources and Inputs thus must be the upstream-most nodes in our task graph: although they still contain values computed during the `resolution` phase, they no longer can contain values computed during `execution`. This is a significant simplification of Mill's internal build graph data model, and I expect it to allow a ton of internal code to be streamlined and simplifierd

In practice, following this restriction doesn't seem to be too onerous:

1. You can no longer `override def sources = Task.Sources` with another `Task.Sources` and call `super.sources() ++ Seq(PathRef(...))`. Instead, you need to define a separate `def customSources =  Task.Sources{}` and do `override def sources = Task { super.sources() ++ customSources() }`

2. You can no longer `override def sources = Task.Sources { super.sources().flatMap{...} }`. Instead, you need to have a separate `def sourcesFolders: Seq[os.SubPath]` that you can override and manipulate, before it gets fed into `Task.Sources(sourcesFolders*)` once at the end.

The various places where we were previously chaining `Task.Sources` or `Task.Input`s together inside Mill's own codebase have been updated to use the new pattern. This was a relatively straightforward and mechanical process. Only `MillBuildRootModule#scriptSources` took a bit more refactoring, which I took as an opportunity to drop the `$file` syntax to allow us to separate `.mill` file discovery and processing (which previously had to be interleaved due to parsing and de-referencing `import $file` statements)

There's some inconvenience, but overall this way of doing things is a lot more correct: e.g. 

1. Previously, if you did `override def Task.Sources { super.sources().map{...} }`, Mill's task graph would depend on `super.sources()` even if the final paths that end up getting returned to `Task.Sources` do not include the originals due to being transformed somehow. 

2. With this PR, you can manipulate `def sourcesFolders` all you want during inheritance using overrides, and only the final `sourceFolders` that get passed to `Task.Sources` ends up being part of the dependency graph.

3. The current `--watch` behavior expects that the paths returned by `Source` and `Sources` are constant, which is not strictly true. It is possible that each time you call `Source`/`Sources` you get a different set of paths, but the current `--watch` behavior does not account for that possibility. With this PR, we formalize the existing assumption that `Source`/`Sources` paths are fixed at module initialization time, and only the content can change.

One scenario where we do lose something is `private def pullAndHash = Task.Input {`. This depends on upstream tasks to determine what to pull, and then acts as a `Task.Input`: that means every time you run `./mill foo` that depends on `pullAndHash` it will re-run, and if you run `./mill -w foo` it will run `pullAndHash` continuously in a loop. While whether or not that is good idea for this specific scenario is questionable (dockerhub may [rate limit you](https://medium.com/@PlanB./docker-hubs-new-rate-limits-what-it-means-for-kubernetes-users-2748e9ff632e)!) it does demonstrate a use case that would not be possible if we merge this PR to restrict `Task.Input`s to not allow upstream tasks

In general you can still do computations to determine the paths to your `Task.Source` and `Task.Sources` folders, just that these computations must happen during the module-initialization/task-resolution phase rather than during the execution phase

This simplification allows significant cleanups

1. `mill.api.TaskResult` is no longer necessary, since we can trivially check the `os.Path`s of `Task.Sources` for updates and the `() => T` of `Task.Input`s for updates without needing to awkwardly capture things in a lambda that we can re-run later

2. `selective.run` no longer needs to run a full evaluation cycle in order to collect the hashes of all the `Task.Input`s and `Task.Sources`, since now you can easily collect those tasks and get each return value individually

There are probably other follow ups I can't think of off the top of my head that will become apparent later


alexarchambault:
Does that mean we can't use `scalaVersion` any more from `sources`?

lihaoyi:
@alexarchambault if we decide to go ahead with this, yes. You would need to do the computation of `sources` without tasks, e.g. using `crossScalaVersion: String` to compute the `sourcesFolders: Seq[os.SubPath]`. You wouldn't be able to use `scalaVersion: T[String]` to compute `sources = Task.Sources`

lihaoyi:
Going to merge this first. We can loosen the restriction again in future if it proves to be too limiting, but even if we do we can probably find a much cleaner implementation than the previous `TaskResult` class and its sketchy `recalc` lambda that violates all sorts of scoping and lifecycle expections

joan38:
@lihaoyi I was looking into a bug with the `import $file` in `.sc` file which I tied to [this change](https://github.com/com-lihaoyi/mill/commit/351fc0fe09ba7391fc664ec2822ea47f771860e8#diff-799f98d1fe0166d49b113cf0bb6c774df0d29b95a390f1c3bf1fdf6ef853189fL277-L284):
```scala
    // Manually do the foldLeft to work around bug in os-lib
    // https://github.com/com-lihaoyi/os-lib/pull/160
    val restSegments = rest
      .map {
        case "^" => os.up
        case s => os.rel / s
      }
      .foldLeft(os.rel)(_ / _)
```
This bug is still in os-lib I guess.

Now this PR just removed this code and you stated:
> I took as an opportunity to drop the $file syntax to allow us to separate .mill file discovery and processing

I'm not sure what this means for users but we are still using `.sc` (instead of `.mill` files) because we are struggling with imports in `.mill` files which also are not supported by Intellij.
I think we cannot drop `.sc` files until there is support in Intellij (right now everything is red with `.mill` imports) and also we should figure out a way to be able to import easily without having to add `package.mill` files everywhere.

joan38:
After looking more closely at this PR, I understand we are basically just taking all `.mill` files in all folders recessively and relying on the `package` declaration to refer to code that sits in different files? The same as for a normal project? Am I right?

lihaoyi:
@joan38 yes that's right. You do need `package.mill` files to mark things for discovery of the mill files, because we do not want to recursively traverse the entire filesystem which may be arbitrarily deep and expensive. But once discovered, the `.mill` files basically behave like normal `.scala` files, `package` statement and all.

Issue #4527, opened 2025-02-11
Split up `mill.api.Result`


lihaoyi:
Currently `Result` serves double duty as an internal model of Task evaluation status, and an external user-facing success-or-failure type. We should split those into two separate types so the user-facing type can be simpler and not have the various `Skipped`/`Aborted`/`Exception`/`Failure#value` cases that the internal task status result needs

lefou:
We should also remove the `Exception` inheritance. Currently the API is ambiguous due to the `Exception` kind.

PR #4528, opened 2025-02-11
Split out `Result` and `ExecResult` types, replace `Either[String, T]` with `Result`

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/4527

`Result` should hopefully provide a simple user-facing result type, replacing the extensive use of `Either[String, T]` throughout the codebase. `ExecResult` models the internals of the `exec/` module and really shouldn't be used elsewhere.

Exactly how rich to make the `Result` type is uncertain; this PR's `Either[String, T]` version is easy and simple, but it's arguable whether we want `Either[Err, T]` to be able to hold non-string errors, or `Either[List[String], T]` to be able to aggregate multiple errors (which we currently discard and pick the first one most of the time). But such changes can happen in future PRs any time before 0.13.0 is released

For now I just got rid of `ExecResult.Failure#value: Option[T]`. We weren't using it anywhere but in unit tests, and it wasn't really suitable to downstream use by users AFAICT. We probably need to come up with a different way to e.g. gather data from failing test results, but this one probably was a failed experiment


core/api/src/mill/api/Result.scala
```diff
@@ -1,138 +1,60 @@
...
-        makeResultException(e.getCause, new java.lang.Exception())
-      case e: Exception => makeResultException(e, new java.lang.Exception())
-    }
-  }
-
-  def makeResultException(e: Throwable, base: java.lang.Exception): Left[String, Nothing] = {
-    val outerStack = new mill.api.Result.OuterStack(base.getStackTrace)
-    Left(mill.api.Result.Exception(e, outerStack).toString)
-  }
+  class Exception(val error: String) extends java.lang.Exception(error)
```
lefou, line 59: Some ScalaDoc would be nice. What's the purpose of and when to use this exception?

lihaoyi: This is meant to replace the `Result.Failure extends Exception` we wanted to get rid of. I'll add Scaladoc later, I expect this to evolve significantly

scalajslib/test/src/mill/scalajslib/UtestTests.scala
```diff
@@ -1,62 +1,62 @@
...
-        .groupBy(_.fullyQualifiedName)
-        .view
-        .mapValues(_.map(e => e.selector -> e).toMap)
-        .toMap
+//      val (doneMsg, testResults) = res
+//      testResults
+//        .groupBy(_.fullyQualifiedName)
+//        .view
+//        .mapValues(_.map(e => e.selector -> e).toMap)
+//        .toMap
```
lefou, line 21: This and some other tests were commented out. We should revisit this.

PR #4529, opened 2025-02-11
Cleanup `Evaluator` public API

lihaoyi:
* Fold `evalOrThrow` into normal `evaluate` code path, rename to `evaluateValues`
* Rename the various methods in `exec` from `evaluate` to `execute`
* Fold `MainModule.resolve*` into `Evaluator#resolve*`
* Break out an `mill.exec.Execution` object so that `Evaluator` doesn't need to directly inherit all of `ExecutionCore`/`GroupExecution`

### DIFF
--- a/core/exec/src/mill/exec/Execution.scala
+++ b/core/exec/src/mill/exec/Execution.scala
@@ -14,11 +14,23 @@ import scala.concurrent._
 /**
  * Core logic of evaluating tasks, without any user-facing helper methods
  */
-private[mill] trait ExecutionCore extends GroupExecution {
-
-  def baseLogger: ColorLogger
-  protected[exec] def chromeProfileLogger: ChromeProfileLogger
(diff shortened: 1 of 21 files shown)


core/exec/src/mill/exec/Execution.scala
```diff
@@ -14,18 +14,34 @@ import scala.concurrent._
 /**
  * Core logic of evaluating tasks, without any user-facing helper methods
  */
-private[mill] trait ExecutionCore extends GroupExecution {
-
-  def baseLogger: ColorLogger
-  protected[exec] def chromeProfileLogger: ChromeProfileLogger
-  protected[exec] def profileLogger: ProfileLogger
+private[mill] case class Execution(
```
lefou, line 17: Shouldn't this be an normal non-case class, since some parameters refer to mutable data structures?

Issue #4534, opened 2025-02-12
`mill.scalalib.checkstyle.CheckstyleModuleTest` always fails


lefou:
Reproduce:

```
> mill -j1 scalalib.test.testOnly mill.javalib.checkstyle.CheckstyleModuleTest
...
[1444] mill.testkit.UnitTester: running checkstyle ...
[1444] Checkstyle endet mit 2 Fehlern.
[1444] mill.testkit.UnitTester: checkstyle output report at /home/lefou/work/opensource/mill/out/scalalib/test/testOnly.dest/sandbox/out/mill-test-base-module/16381572422961682342/out/checkstyleOutput.dest/checkstyle-output.xml
[1444] mill.testkit.UnitTester: checkstyle found 2 violation(s)
[1444] X mill.javalib.checkstyle.CheckstyleModuleTest.arguments.sources 1042ms 
[1444]   utest.AssertionError: #1: testJava(
[1444]               resources / "non-compatible",
[1444]               violations = violations.take(2),
[1444]               sources = Seq("src/blocks")
[1444]             )
[1444]   resources: os.Path = /home/lefou/work/opensource/mill/scalalib/test/resources/checkstyle
[1444]   violations: Seq[String] = List(Array should contain trailing comma, Array should contain trailing comma, Empty statement, Empty statement)
[1444]     utest.asserts.Asserts$.assertImpl(Asserts.scala:30)
[1444]     mill.javalib.checkstyle.CheckstyleModuleTest$.tests$$anonfun$1$$anonfun$1$$anonfun$3(CheckstyleModuleTest.scala:12)
...
[1444] mill.testkit.UnitTester: running checkstyle ...
[1444] Checkstyle endet mit 4 Fehlern.
[1444] mill.testkit.UnitTester: checkstyle output report at /home/lefou/work/opensource/mill/out/scalalib/test/testOnly.dest/sandbox/out/mill-test-base-module/8340530497038255299/out/checkstyleOutput.dest/checkstyle-output.plain
[1444] mill.testkit.UnitTester: checkstyle found 4 violation(s)
[1444] X mill.javalib.checkstyle.CheckstyleModuleTest.settings.format 976ms 
[1444]   utest.AssertionError: #1: testJava(resources / "non-compatible", "plain", violations = violations)
[1444]   resources: os.Path = /home/lefou/work/opensource/mill/scalalib/test/resources/checkstyle
[...]

PR #4535, opened 2025-02-12
Explicitly set locale in scalalib tests to `LC_ALL=C.utf8`

lefou:
Should fix locale specific isues in checkstyle tests

Fix https://github.com/com-lihaoyi/mill/issues/4534

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -392,6 +392,7 @@ trait MillJavaModule extends JavaModule {
   }
 
   def testIvyDeps: T[Agg[Dep]] = Agg(Deps.TestDeps.utest)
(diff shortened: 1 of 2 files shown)


alexarchambault:
Still investigating why, but it seems this broke this test (among others):
```text
$ ./mill 'scalalib.test.test' "mill.scalalib.CrossVersionTests.StandaloneScala213"
```

(it fails explicitly in the CI of https://github.com/com-lihaoyi/mill/pull/4626, but was broken before that apparently)

alexarchambault:
https://github.com/com-lihaoyi/mill/pull/4747 fixes it for me

PR #4540, opened 2025-02-12
Android/fix android paths

vaslabs:
Fixes android build + tests following Agg -> Seq migration

This should be ready for review. Android tests should be green again


vaslabs:
the tests are passing, but the CI is timing out now because the sequential testing has changed and one of the tests depending on emulation gets stuck

PR #4543, opened 2025-02-12
Use adoc syntax to mark graphviz blocks in examples

lefou:
Fixed graphviz block to use AsciiDoc syntax, and other fixes.


lefou:
> what's the differnece in the two syntaxes? dont they mean the same thing?

One is from Markdown, the other from AsciiDoc. Asciidoctor does support some Markdown, so that most markdown documents render out of the box, but the latter works better and has more options. Also in InteliiJ, only the AsciiDoc syntax will be supported and render graphviz diagrams or use syntax highlighting (and even jump to code actions) for source blocks.

Issue #4532, opened 2025-02-11
Breakage in GenIdea tests due to coursier local module resolution error


lihaoyi:
Applying the following patch to surface the error

```diff
lihaoyi mill$ git diff
diff --git a/idea/src/mill/idea/GenIdea.scala b/idea/src/mill/idea/GenIdea.scala
index 6801bd4609a..dbb12ad8dbe 100644
--- a/idea/src/mill/idea/GenIdea.scala
+++ b/idea/src/mill/idea/GenIdea.scala
@@ -17,8 +17,6 @@ object GenIdea extends ExternalModule with mill.define.TaskModule {
       ).run())
     } catch {
       case GenIdeaImpl.GenIdeaException(m) => Result.Failure(m)
-      case NonFatal(e) =>
-        ExecResult.Exception(e, new ExecResult.OuterStack(new java.lang.Exception().getStackTrace))
     }
   }
 
diff --git a/integration/ide/gen-idea/src/GenIdeaTests.scala b/integration/ide/gen-idea/src/GenIdeaTests.scala
index fe6b569fd5a..2e7abbd639d 100644
--- a/integration/ide/gen-idea/src/GenIdeaTests.scala
+++ b/integration/ide/gen-idea/src/GenIdeaTests.scala
@@ -15,7 +15,7 @@ object GenIdeaTests extends UtestIntegrationTestSuite {
       val expectedBase = workspacePath / "idea"
       val resources = os.walk(expectedBase).filter(os.isFile).map(_.subRelativeTo(expectedBase))
 
-      eval("mill.idea.GenIdea/")
+      eval("mill.idea.GenIdea/", stdout = os.Inherit, stderr = os.Inherit, check = true)
 
       for (resource <- resources) assertIdeaXmlResourceMatchesFile(workspacePath, resource)
     }

```

and running

```
./mill 'integration.ide[gen-idea].packaged.server.test' mill.integration.GenIdeaTests.genIdeaTests
```

results in

```[5873] ----------- Running Tests mill.integration.GenIdeaTests.genIdeaTests -----------
[5873] Copying integration test sources from /Users/lihaoyi/Github/mill/integration/ide/gen-idea/resources/hello-idea to /Users/lihaoyi/Github/mill/out/integration/ide/gen-idea/packaged/server/test.dest/sandbox/run-1
[...]

PR #4499, opened 2025-02-07
Cleanup deprecated

lihaoyi:

PR #4546, opened 2025-02-12
Rework millEmbeddedDeps stuff

alexarchambault:
This fixes an issue with GenIdea stuff that's been around since https://github.com/com-lihaoyi/mill/pull/4499 (that uncommented suggested code from #4145 to address some deprecations).

IIUC, in `RootModule#bindDependency`, Mill excludes all of its dependencies, so that these are not pulled by plugins. The PR here excludes not only Mill dependencies, but also the Mill modules themselves (anything that the `dist.dist0` module depends on, itself included).

Also, GenIdea tries to pull the class path of Mill itself, to pass it to IntelliJ. This class path is now pulled via `com.lihaoyi:mill-dist-dist0`, that should pull the whole class path of Mill.

Fixes https://github.com/com-lihaoyi/mill/issues/4532


lihaoyi:
Thanks Alex!

PR #4557, opened 2025-02-14
Add option to use the embeddable compiler in `KotlinModule`

vaslabs:
Adds a new `kotlinCompilerEmbeddable` task, which switches the used compiler toolchain to use the alternative "embeddable" jars.

This is in preparation for https://github.com/com-lihaoyi/mill/pull/4485 .

In summary, these changes are needed to avoid classpath conflicts between the compiler and the compiler plugins (as it happens [here](https://github.com/com-lihaoyi/mill/pull/4485) with the dagger-compiler , which is used in Android DI setups as well (with the addition of Hilt)

### DIFF
--- a/example/kotlinlib/module/8-kotlin-compiler-plugins/build.mill
+++ b/example/kotlinlib/module/8-kotlin-compiler-plugins/build.mill
@@ -15,6 +15,8 @@ object foo extends KotlinModule {
     ivy"org.jetbrains.kotlinx:kotlinx-serialization-json:1.6.3"
   )
 
+  def kotlinCompilerEmbeddable = true
+
   def processors = Task {
     defaultResolver().resolveDeps(
       Seq(
--- a/example/kotlinlib/module/8-kotlin-compiler-plugins/build.mill
+++ b/example/kotlinlib/module/8-kotlin-compiler-plugins/build.mill
@@ -15,12 +15,10 @@ object foo extends KotlinModule {
     ivy"org.jetbrains.kotlinx:kotlinx-serialization-json:1.6.3"
   )
 
-  def kotlinCompilerEmbeddable = true
-
   def processors = Task {
     defaultResolver().resolveDeps(
       Seq(
-        ivy"org.jetbrains.kotlin:kotlin-serialization-compiler-plugin-embeddable:1.9.24"
+        ivy"org.jetbrains.kotlin:kotlin-serialization-compiler-plugin:1.9.24"
       )
     )
   }
--- a/example/kotlinlib/module/8-kotlin-compiler-plugins/build.mill
+++ b/example/kotlinlib/module/8-kotlin-compiler-plugins/build.mill
@@ -18,7 +18,7 @@ object foo extends KotlinModule {
   def processors = Task {
     defaultResolver().resolveDeps(
       Seq(
-        ivy"org.jetbrains.kotlin:kotlin-serialization-compiler-plugin:1.9.24"
+        ivy"org.jetbrains.kotlin:kotlin-serialization-compiler-plugin-embeddable:1.9.24"
       )
     )
   }
--- a/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
+++ b/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
@@ -412,6 +412,8 @@ trait KotlinModule extends JavaModule { outer =>
       outer.kotlincOptions().filterNot(_.startsWith("-Xcommon-sources")) ++
         Seq(s"-Xfriend-paths=${outer.compile().classes.path.toString()}")
     }
+
+    override def kotlinCompilerEmbeddable: Task[Boolean] = outer.kotlinCompilerEmbeddable
   }
 
 }
--- a/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
+++ b/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
@@ -107,14 +107,14 @@ trait KotlinModule extends JavaModule { outer =>
    * Default is derived from [[kotlinCompilerVersion]].
    */
   def kotlinCompilerIvyDeps: T[Seq[Dep]] = Task {
-    Seq(ivy"org.jetbrains.kotlin:kotlin-compiler:${kotlinCompilerVersion()}") ++
+    Seq(ivy"org.jetbrains.kotlin:kotlin-compiler-embeddable:${kotlinCompilerVersion()}") ++
       (
         if (
           !Seq("1.0.", "1.1.", "1.2.0", "1.2.1", "1.2.2", "1.2.3", "1.2.4").exists(prefix =>
             kotlinVersion().startsWith(prefix)
           )
         )
-          Seq(ivy"org.jetbrains.kotlin:kotlin-scripting-compiler:${kotlinCompilerVersion()}")
+          Seq(ivy"org.jetbrains.kotlin:kotlin-scripting-compiler-embeddable:${kotlinCompilerVersion()}")
         else Seq()
       )
   }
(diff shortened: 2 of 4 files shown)


kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -102,19 +102,48 @@ trait KotlinModule extends JavaModule { outer =>
     )().toSeq ++ kotlinWorkerClasspath()
   }
 
+  /** Flag to use the embeddable kotlin compiler */
+  def kotlinCompilerEmbeddable: Boolean = false
```
lefou, line 106: I think this could and should be a `Task[Boolean]`, with the benefit, that it can depend on other tasks to make the decision. I could imagine so logic to analyze the plugins for compatibility, or to refer to some compatibility matrix.

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -102,19 +102,48 @@ trait KotlinModule extends JavaModule { outer =>
     )().toSeq ++ kotlinWorkerClasspath()
   }
 
+  /** Flag to use the embeddable kotlin compiler */
```
lefou, line 105:
```suggestion
  /** 
   * Flag to use the embeddable kotlin compiler. 
   * This can be necessary to avoid classpath comflicts or ensure 
   * compatibility to the used set of plugins. 
   *
   * See also https://discuss.kotlinlang.org/t/kotlin-compiler-embeddable-vs-kotlin-compiler/3196
   */
```

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -102,19 +102,48 @@ trait KotlinModule extends JavaModule { outer =>
...
+  def kotlinCompilerEmbeddable: Boolean = false
+
+  /**
+   * If ksp plugins are used, switch to embeddable to avoid
+   * any classpath conflicts.
+   *
+   * @return kotlin-compiler or kotlin-compiler-embeddable dependency
+   */
```
lefou, line 113: The mention of `ksp` is misleading here.

```suggestion
  /**
   * The kotlin-compiler dependency. It uses the embeddable version, if [[kotlinCompilerEmbeddable]] is `true`.
   */
```

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -102,19 +102,48 @@ trait KotlinModule extends JavaModule { outer =>
...
+  def kotlinCompilerDep: T[Dep] = Task {
+    if (kotlinCompilerEmbeddable)
+      ivy"org.jetbrains.kotlin:kotlin-compiler-embeddable:${kotlinCompilerVersion()}"
+    else
+      ivy"org.jetbrains.kotlin:kotlin-compiler:${kotlinCompilerVersion()}"
+  }
+
+  /**
+   * If ksp plugins are used, switch to embeddable to avoid
```
lefou, line 122: The mention of `ksp` is misleading here.

example/kotlinlib/module/8-kotlin-compiler-plugins/build.mill
```diff
@@ -15,10 +15,12 @@ object foo extends KotlinModule {
     ivy"org.jetbrains.kotlinx:kotlinx-serialization-json:1.6.3"
   )
 
+  def kotlinCompilerEmbeddable = true
```
lefou, line 18: Do you enable it to show-case it, or does is actually fix/improves anything?

lefou (commented):
Thanks for opening this PR. I left some comments below.

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
vaslabs: thanks! do you know why when I switch it to a task, the underlying tasks are executed as well?

e.g. in

```scala
if (kotlinCompilerEmbeddable())
   thisTaskEvalIsExecuted()
else
   thisTaskEvalIsAlsoExecuted()
```

example/kotlinlib/module/8-kotlin-compiler-plugins/build.mill
vaslabs: just to show case it, there are 2 such examples, I've enabled it only here. I can skip this and do it in the dagger example of the following PR

lefou: If it's not really needed, we should not use it, esp. since we plan to show it in the dagger example. In case we want to show it in isolation, there should be at least some docs pointing out the fact.

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
lefou: You're example is expanded to the following: 

```scala
Task {
   if (kotlinCompilerEmbeddable())
      thisTaskEvalIsExecuted()
   else
      thisTaskEvalIsAlsoExecuted()
   }
```

It's a task with three dependencies. Since Mill calculated the dependency graph before it starts to evaluate it, there is no way to make the graph dependent on the results of task dependencies.

But in this PR, I can't see any downsteam task where this limitation could causes any issue.

vaslabs: no, I had this issue on the other PR, for KSP

example/kotlinlib/module/8-kotlin-compiler-plugins/build.mill
vaslabs: reverted!

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
lefou: Ok, I didn't recognized that. Let me check.

lefou: I checked. You probably meant the feature switch to enable KSP. The nice thing is, that by removing the initally proposed feature switch and implementing the `KspModule` as a stackable trait, we won't have this issue. It's now encode in the types. We change the task graph by adding the `KspModule` or not. Sounds ok to me.

lefou: To elaborate, your initial feature switch was a on/off decision resulting in different work to do, whereas this task here is a A/B decision resulting in the same kind of work on different data.

vaslabs: that was my understanding, but I'm new here and wanted to clarify :D 
either that or I was going crazy, lol

thanks for the info!

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -102,19 +102,53 @@ trait KotlinModule extends JavaModule { outer =>
...
+   * See also https://discuss.kotlinlang.org/t/kotlin-compiler-embeddable-vs-kotlin-compiler/3196
+   */
+  def kotlinCompilerEmbeddable: Task[Boolean] = Task { false }
+
+  /**
+   * The kotlin-compiler dependency.
+   *
+   * It uses the embeddable version, if [[kotlinCompilerEmbeddable]] is `true`.
+   */
+  def kotlinCompilerDep: T[Dep] = Task {
```
lefou, line 119: We probably should make this return a `Seq[Dep]`, in case, multiple deps or more control is needed.

lefou (commented):
This looks really good. 

In the PR description you wrote:

> This would break any modules that use non embeddable plugins (see serialization plugin changes for examples). I believe we shouldn't break them and have a feature switch

Can you elaborate? Since the user have control over the fact which toolchain is used, this PR should not break any existing workflow.

Can you add a (unit) test that covers the newly added embedabble toolchain to early catch issues. (see `kotlin.HelloWorldTests`)

vaslabs:
> In the PR description you wrote:
> 
> > This would break any modules that use non embeddable plugins (see serialization plugin changes for examples). I believe we shouldn't break them and have a feature switch
> 
> Can you elaborate? Since the user have control over the fact which toolchain is used, this PR should not break any existing workflow.
> 
> Can you add a (unit) test that covers the newly added embedabble toolchain to early catch issues. (see `kotlin.HelloWorldTests`)

Sorry, it was before I decided to add the toggle! Without it, 2 example apps needed to switch to using the embeddable serialization plugin!

lefou (approved):
Looks good to me. Thanks!

Issue #4564, opened 2025-02-15
Separate `resolvedRunIvyDeps` and `transitiveLocalClasspath` into two stages when build assembly jar


lihaoyi:
This will allow modifications to upstream local modules to benefit from the incremental assembly construction

Noticed when writing https://github.com/com-lihaoyi/mill/pull/4558

PR #4565, opened 2025-02-15
Separate `resolvedRunIvyDeps` and `transitiveLocalClasspath` into two stages when build assembly jar

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/4564

This makes it such that changes to upstream modules can avoid re-assembling all the external thirdparty dependencies, whereas previously only changes to the exact module running `.assembly` would benefit

Tested manually via ` /Users/lihaoyi/Github/mill/mill-assembly.jar -w integration.assembly` in `example/javalib/web/3-todo-springboot/`. 

- Previously incremental updates after changes to `src/` would take ~20s, because `integration` was a downstream module of the root `build` and so it wouldn't benefit from incremental assembly construction. 
- With this PR, we separately construct the `resolvedIvyAssembly.dest/out.jar` containing the bulk of the Spring classfiles before adding transitive local classpath and then the module's own local classpath, and incremental updates take ~1s

### DIFF
--- a/scalalib/src/mill/scalalib/AssemblyModule.scala
+++ b/scalalib/src/mill/scalalib/AssemblyModule.scala
@@ -63,7 +63,7 @@ trait AssemblyModule extends mill.Module {
 
   private[mill] def assemblyRules0: Seq[Assembly.Rule] = Assembly.defaultRules
 
-  def resolvedIvyAssemblyClasspath: T[Seq[PathRef]]
+  def upstreamIvyAssemblyClasspath: T[Seq[PathRef]]
 
   def upstreamLocalAssemblyClasspath: T[Seq[PathRef]]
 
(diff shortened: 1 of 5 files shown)


scalalib/src/mill/scalalib/AssemblyModule.scala
```diff
@@ -63,10 +63,28 @@ trait AssemblyModule extends mill.Module {
 
   private[mill] def assemblyRules0: Seq[Assembly.Rule] = Assembly.defaultRules
 
-  def upstreamAssemblyClasspath: T[Seq[PathRef]]
+  def upstreamIvyAssemblyClasspath: T[Seq[PathRef]]
+
+  def upstreamLocalAssemblyClasspath: T[Seq[PathRef]]
```
lefou, line 68: Some scaladoc would be nice.

PR #4570, opened 2025-02-16
Rebootstrap on 0.12.8

lihaoyi:
Pulls in https://github.com/com-lihaoyi/os-lib/pull/359 from upstream to fix a problem with leaked processes causing hangs

PR #4545, opened 2025-02-12
Rebootstrap on latest 0.12.x snapshot

lefou:

PR #4571, opened 2025-02-16
Fix hang on `mill.integration.invalidation[selective-execution]`

lihaoyi:
Should fix the rebootstrap problems found in https://github.com/com-lihaoyi/mill/pull/4545 and https://github.com/com-lihaoyi/mill/pull/4570 that seems to be caused by a mistake in this backport https://github.com/com-lihaoyi/mill/pull/4504. Not sure why this causes a hang, but for now just fixing it to revert to the previous semantics

Tested manually via

```bash
./mill dist.installLocal
git checkout main
./mill-assembly.jar 'integration.invalidation[selective-execution].local.fork.test' mill.integration.SelectiveExecutionWatchTests.watch.changed-code
```

Doing this manually on `0.12.x` hangs, whereas doing it on this PR succeeds


lihaoyi:
Seems this didn't fix the issue, still seeing the hang here https://github.com/com-lihaoyi/mill/pull/4570

PR #4575, opened 2025-02-17
Add init for sbt

sake92:
Got basic init setup for sbt working.

Depends on https://github.com/[sake92/sbt-build-extract](https://github.com/sake92/sbt-build-extract) sbt plugin.
Users have to add it to their `plugins.sbt` file.
The plugin exports *each sbt project* to root `target/build-export` folder, in JSON format.

Then mill picks up those JSON files and parses them to build the `build.mill` and `package.mill` files.


lihaoyi:
The approach of using a custom SBT plugin sounds reasonable. Can we include it in this PR and keep it in the Mill repo?

Adding the plugin manually seems like it can be automated by having the command write a file `project/mill_export.sbt` to include it during the export, and remove the file after it's done

sake92:
I've embedded it inside this repo:
- added CI to run on PR
- added CD to publish on tag

Also applied `project/exportBuildStructure.sbt` temporarily as you suggested, works good!

sake92:
Seems like I didnt copy the sbt plugin sources properly 🤦🏻 

But anyways, seems like https://github.com/com-lihaoyi/mill/pull/4586  got it further.

PR #4586, opened 2025-02-18
Support `mill init` from an existing sbt project, and improve related Maven and Gradle support

### LINKED ISSUE #3450: Support `mill init` from an existing SBT project (2000USD Bounty)
---

**From the maintainer Li Haoyi**: I'm putting a 2000USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

We should be able to run `./mill init` to do a best effort conversion of an existing SBT project to Mill: 

* Code-generate a `build.mill` containing a `MavenModule`for the root `build.sbt`, the `package.mill` for any subfolder `build.sbt`s or SBT subprojects in the root `build.sbt`
* Convert the main and test dependencies on third-party libraries to `ivyDeps`
* Convert inter-subproject dependencies to `moduleDeps`
* Set any necessary `javacOptions`
* Implement `PublishModule` with relevant metadata

Such a conversion will never be 100%, but if we can automate 80% of the conversion that will already be a huge help for people migrating to Mill or even just trying it out

ShreckYe:
Resolves #3450.

## Feature summary

The conversion
- handles deeply nested modules
- captures publish settings
- configures dependencies for configurations:
  - no configuration
  - Compile
  - Test
  - Runtime
  - Provided
  - Optional
- configures testing frameworks:
  - Java:
    - JUnit 4
    - JUnit 5
    - TestNG
  - Scala:
    - ScalaTest
    - Specs2
    - µTest
    - MUnit
    - Weaver
    - ZIOTest
    - ScalaCheck

The conversion does not support:
- custom dependency configurations
- custom sbt settings including custom tasks
- sources other than Scala on JVM and Java, such as Scala.js and Scala Native
- cross builds

## Implementation summary

To resolve the issue, I use sbt's [addPluginSbtFile command](https://www.scala-sbt.org/1.x/docs/Combined+Pages.html#addPluginSbtFile+command) on Linux or add a temporary `.sbt` script in the sbt `project` directory on Windows, to add a plugin jar assembled from the `exportplugin` module, which adds an sbt task `millInitExportBuild` to extract the settings for a build and serialize them with uPickle, and then the rest of `SbtBuildGenMain` imports the file and generates the build. The `exportplugin` sbt plugin module depends on sbt which requires Scala 2.12, so the `models` module is set up as a cross module.

Code in the common `buildgen`, Maven, and Gradle modules is refactored and improved. Major changes include:

- The project hierarchy representation is updated from `Tree[Node[T]]` to `Tree[Node[Option[T]]]` and related code is refactored.
- Some general integration test helpers such as `testMillInit` are added in `MillInitUtils`, and the integration tests for Maven and Gradle and updated using them.
- Generated module settings that are different from those in the base module are overridden, and extended from their super ones if possible.
- Update some nullable types with Scala 3 explicit nulls.
- Inter-project dependencies for Gradle and sbt are imported in an improved direct way rather than resolved from a GAV-module map, so dependencies on subprojects with the same name are properly imported.

## Notes for further development

* I left some TODO comments in the code related to alternative implementations, unimplemented features I don't know whether are necessary at the moment and may involve some work, and refactoring/renaming suggestions.


ShreckYe:
The CI seems to fail because there is no sbt set up. (See "main.init.sbt.sbtPluginJarResources failed" in https://github.com/ShreckYe/mill/actions/runs/13398010494/job/37429413361.) Shall I update the CI to set it up? Or shall I pull and add a script from sbt-extras to use instead?

lihaoyi:
@ShreckYe yes please update the CI scripts

ShreckYe:
Hi, I have fixed the CI failures except for `itest (17, ci/test-mill-bootstrap.sh)` and I think this PR is basically ready for review. The CI job fails with the following message:

```text
1 out of 6 hunks failed--saving rejects to 'example/package.mill.rej'
```

I couldn't figure out how to resolve this. I have only updated the `package.mill` with a `migrating` module and added 2 example test `build.mill`s in the "example" directory without modifying other files in it. Could you maintainers help me resolve this or offer me some guidance?

integration/feature/init/src/MillInitSbtTests.scala
```diff
@@ -0,0 +1,230 @@
...
+      bumpSbtTo1107(workspacePath)
+
+      val initResult = eval(initCommand, stdout = os.Inherit, stderr = os.Inherit)
+      assert(initResult.isSuccess)
+
+      val compileResult = eval("compile")
+      assert(compileResult.isSuccess)
+
+      val compileSubmodulesResult = eval("_.compile")
+      if (System.getProperty("java.version").split('.').head.toInt <= 11)
```
lihaoyi, line 103: Can use implement a `--jvm-id 11` flag similar to what the Gradle import supports, so the user can force this to use a working version of JDK11?

main/init/package.mill
```diff
@@ -86,4 +93,53 @@ object `package` extends RootModule with build.MillPublishScalaModule {
     )
     def testModuleDeps = super.testModuleDeps ++ Seq(build.scalalib, buildgen.test)
   }
+
+  object sbt extends build.MillPublishScalaModule {
+    def moduleDeps = Seq(buildgen, models(build.Deps.scalaVersion))
+
+    val sbtPluginProjectPath = millModuleBasePath.value / "sbt-mill-init-export-build"
```
lihaoyi, line 100: Could we build this plugin using Mill instead of SBT? A SBT plugin is just a `ScalaModule` with some dependencies configured after all, so we should be able to integrate it into Mill as a normal `ScalaModule` without needing to shell out to `sbt`

integration/feature/init/src/MillInitSbtTests.scala
```diff
@@ -0,0 +1,230 @@
...
+
+      val compileSubmodulesResult = eval("_.compile")
+      // Some dependencies with currently unsupported `CrossVersion` `For3Use2_13` are not imported properly
+      assert(!compileSubmodulesResult.isSuccess)
+    }
+  }
+}
+
+// Scala.js and scala-native projects are not properly imported
+object MillInitSbtScalazTests extends BuildGenTestSuite {
```
lihaoyi, line 141: When I run this built in SBT, I see the following subprojects:

```
[info] In file:/Users/lihaoyi/Github/mill/out/integration/feature/init/local/fork/test.dest/sandbox/run-1/
[info] 	   coreJS
[info] 	   coreJVM
[info] 	   coreNative
[info] 	   effectJS
[info] 	   effectJVM
[info] 	   effectNative
[info] 	   exampleJS
[info] 	   exampleJVM
[info] 	   exampleNative
[info] 	   iterateeJS
[info] 	   iterateeJVM
[info] 	   iterateeNative
[info] 	   rootJS
[info] 	   rootJVM
[info] 	   rootNative
[info] 	   scalacheckBindingJS
[info] 	   scalacheckBindingJVM
[info] 	   scalacheckBindingNative
[info] 	 * scalaz
[info] 	   testsJS
[info] 	   testsJVM
[info] 	   testsNative
```

But the converted Mill code only shows:

```
root
rootJS
rootJVM
rootNative
```

How come the other SBT subprojects are not being converted?

main/init/sbt/test/resources/expected/config/sbt-multi-project-example/build.mill
```diff
@@ -0,0 +1,155 @@
...
+    def ivyDeps = super.ivyDeps() ++ Seq(
+      Deps.`ch.qos.logback:logback-classic`,
+      Deps.`com.typesafe.akka:akka-stream`,
+      Deps.`com.typesafe.scala-logging:scala-logging`,
+      Deps.`com.typesafe:config`,
+      Deps.`net.logstash.logback:logstash-logback-encoder`,
+      Deps.`org.slf4j:jcl-over-slf4j`
+    )
+
+    object tests extends MavenTests with TestModule.ScalaTest {
```
lihaoyi, line 51: Why is this `MavenTests` and not `SbtTests`?

integration/feature/init/src/MillInitSbtTests.scala
```diff
@@ -0,0 +1,230 @@
...
+
+    test - integrationTest(url) { tester =>
+      import tester.*
+
+      bumpSbtTo1107(workspacePath)
+
+      val initResult = eval(initCommand, stdout = os.Inherit, stderr = os.Inherit)
+      assert(initResult.isSuccess)
+
+      val compileResult = eval("compile")
```
lihaoyi, line 99: These examples should also run tests and asserting on the stdout/stderr to make sure we're correctly picking up the test sources and wiring up the test suites.

lihaoyi:
Feel free to ignore `ci/test-mill-bootstrap.sh` for now. The breakage is incidental and I'll fix it myself when it's time to merge this

integration/feature/init/src/MillInitSbtTests.scala
```diff
@@ -0,0 +1,230 @@
...
+      assert(compileResult.isSuccess)
```
lihaoyi, line 131: Apart from `compile` on the root module, does `.compile` work on any other modules? which ones?

main/init/buildgen/src/mill/main/buildgen/BuildGenBase.scala
```diff
@@ -4,26 +4,29 @@ import mill.main.buildgen.BuildGenUtil.{buildPackages, compactBuildTree, writeBu
 
 import scala.collection.immutable.SortedMap
 
-trait BuildGenBase[M, D] {
+trait BuildGenBase[M, D, I] {
```
lihaoyi, line 7: What does this new type parameter do? I don't see us using it anywhere

integration/feature/init/src/MillInitSbtTests.scala
```diff
@@ -0,0 +1,230 @@
...
+     */
+    val url = "https://github.com/typelevel/cats/archive/refs/tags/v2.13.0.zip"
+
+    test - integrationTest(url) { tester =>
+      import tester.*
+
+      val initResult = eval(initCommand, stdout = os.Inherit, stderr = os.Inherit)
+      assert(initResult.isSuccess)
+
+      val compileResult = eval("compile")
```
lihaoyi, line 179: Similarly, SBT lasts the following subprojects

```
[info] 	   algebraJS
[info] 	   algebraJVM
[info] 	   algebraLawsJS
[info] 	   algebraLawsJVM
[info] 	   algebraLawsNative
[info] 	   algebraNative
[info] 	   alleycatsCoreJS
[info] 	   alleycatsCoreJVM
[info] 	   alleycatsCoreNative
[info] 	   alleycatsLawsJS
[info] 	   alleycatsLawsJVM
[info] 	   alleycatsLawsNative
[info] 	   bench
[info] 	   binCompatTest
[info] 	 * cats
[info] 	   catsJS
[info] 	   catsJVM
[info] 	   catsNative
[info] 	   coreJS
[info] 	   coreJVM
[info] 	   coreNative
[info] 	   docs
[info] 	   freeJS
[info] 	   freeJVM
[info] 	   freeNative
[info] 	   kernelJS
[info] 	   kernelJVM
[info] 	   kernelLawsJS
[info] 	   kernelLawsJVM
[info] 	   kernelLawsNative
[info] 	   kernelNative
[info] 	   lawsJS
[info] 	   lawsJVM
[info] 	   lawsNative
[info] 	   testkitJS
[info] 	   testkitJVM
[info] 	   testkitNative
[info] 	   testsJS
[info] 	   testsJVM
[info] 	   testsNative
[info] 	   unidocs
```

So how is it the Mill build only contains a subset

```
bench.compile
binCompatTest.compile
binCompatTest.test.compile
compile
site.compile
unidocs.compile
```

lihaoyi:
> Conventionally in an sbt project the common settings, corresponding to the base module, are set in ThisBuild. However, not all projects seem to follow this convention, which can lead to very cumbersome build files with long duplicate settings, such as javacOptions. I can add the similar baseProject config arg from Gradle's implementation to resolve this if this is necessary.

Yes `baseModule` is meant to support this use case. That should let us move all the common config into a `trait` and have the individual modules inherit it and only override the cases where they differ

main/init/package.mill
ShreckYe: I first tried this but it didn't work well. You can see from the commit message of 3c9f155d1fd5e4146e22284d7104e02da83bbbcf that after reviewing the contents of https://repo1.maven.org/maven2/com/eed3si9n/sbt-assembly_2.12_1.0/2.3.1/sbt-assembly_2.12_1.0-2.3.1.jar, I see there is an "sbt/sbt.autoplugins" file which has the contents:

```
sbtassembly.AssemblyPlugin

```

This probably means that such a file has to be included for an sbt plugin and a plain Mill module doesn't support this. I guess if using Mill a task could be added to write the file to the jar, but I was afraid that there were other files packaged in the jar that I am not aware of, so I switched to sbt.

integration/feature/init/src/MillInitSbtTests.scala
ShreckYe: I think it is due to [this TODO which is not resolved yet](https://github.com/ShreckYe/mill/blob/6b717b7423dd802a50bc25a2110fc6f1843d4d30/main/init/sbt/src/mill/main/sbt/SbtBuildGenMain.scala#L150-L159):

```scala
    /*
    TODO This does not support converting child projects nested in a parent directory which is not a project yet.
     Supporting this may involve refactoring `Node[Project]` into `Node[Option[Project]]`
     or adding a subproject as a direct child with its ancestor directory project (I am not sure whether this works).
     */
    val input = Tree.from(projectNodesByParentDirs(None).head) { node =>
      val dirs = node.dirs
      val children = projectNodesByParentDirs.getOrElse(Some(dirs), Seq.empty)
      (node, children)
    }
```

main/init/sbt/test/resources/expected/config/sbt-multi-project-example/build.mill
ShreckYe: Sorry I think I missed this in common buildgen code.

integration/feature/init/src/MillInitSbtTests.scala
ShreckYe: I will resolve these after fixing https://github.com/com-lihaoyi/mill/pull/4586#discussion_r1965045922.

main/init/buildgen/src/mill/main/buildgen/BuildGenBase.scala
ShreckYe: In my current information sbt's `IrBaseInfo` is not retrieved from a project but from `mill.main.sbt.BuildInfo`, which is extracted from sbt's `ThisBuild`, so for this to work the input type `I` has to be `(BuildInfo, Tree[Node[Project]])` instead of just `Tree[Node[Project]]`.

integration/feature/init/src/MillInitSbtTests.scala
ShreckYe: I will look into this. This might be related to https://github.com/com-lihaoyi/mill/pull/4586#discussion_r1965045922 too.

ShreckYe:
In the current implementation, `baseModule` is supported but the settings are extracted from sbt's `ThisBuild`. Shall I also support extracting them from a project specified by `baseProject` like how it's done for Gradle?

lihaoyi:
> In the current implementation, baseModule is supported but the settings are extracted from sbt's ThisBuild. Shall I also support extracting them from a project specified by baseProject like how it's done for Gradle?

@ShreckYe I'm actually not sure what the different between these two approaches is. Could you elaborate a bit?

integration/feature/init/src/MillInitSbtTests.scala
lihaoyi: While that refactoring sounds plausible, I'm not sure it's the cause here. e.g. the `effect{JVM,JS.Native}` projects are in the `effect/`, directly inside the project root. But you'll have to play around and see what you can do

ShreckYe:
Sure. sbt is configured through settings, and project's settings delegate from `ThisBuild` if they are not set. sbt calls this [scope delegation](https://www.scala-sbt.org/1.x/docs/Scope-Delegation.html), it's like inheriting or overriding non-abstract methods in OOP. So my current implementation extracts `IrBaseInfo` from `ThisBuild`'s settings.

Gradle doesn't have something like `ThisBuild` and common configurations are shared by applying common script plugins (usually called convention plugins like [buildlogic.java-common-conventions.gradle.kts](https://github.com/com-lihaoyi/mill/blob/601c5ae657b12dd0e7ad84eba74afa4555268388/main/init/gradle/test/resources/application-library/buildSrc/src/main/kotlin/buildlogic.java-common-conventions.gradle.kts)), the Gradle implementation therefore selects a subproject to extract the common `IrBaseInfo`, which is configuration with the `baseProject` config parameter.

Not all sbt projects follow the convention of putting common settings in `ThisBuild`, some just put the common settings in a common `val` like [here](https://github.com/pbassiner/sbt-multi-project-example/blob/152b31df9837115b183576b0080628b43c505389/build.sbt#L93). I am not an old user of sbt, but I suspect that it's probably because `ThisBuild` was introduced in a relatively late version. So to support such projects as well, we can support the `baseProject` config parameter when converting from sbt too.

lihaoyi:
Got it. I think in general we will never be able to replicate the logic of the arbitrary SBT code passing around settings lists before they end up finally being passed to a project. Letting the user select a `baseProject` sounds reasonable, but what should we do if there is a project that contains a _subset_ of the `baseProject`'s settings?

ShreckYe:
I think we can diff the settings and override the ones that are different, which is also what I found that the current Gradle implementation doesn't handle perfectly ([This line](https://github.com/com-lihaoyi/mill/blob/0c9b749e60855d6314df0c6140fdc020ef16d6a3/main/init/gradle/src/mill/main/gradle/GradleBuildGenMain.scala#L179) for example just ignores the project settings, assuming it's the same as the `baseProject` one).

However, a potential issue with this is that some settings like `PomSettings` are quite long, and if one single member is different and the whole thing is overriden, it can produce very long Mill code. A possible solution I can think of is to use `super.pomSettings.copy(...)` and specify only the different members, but this definitely complicates the build generation code.

lihaoyi:
How about something like:

1. For a setting which are `Seq`s, we consider two special cases
    * If the `setting == baseProject.setting`, then we do not need to emit any override
     * If the `setting.startsWith(baseProject.setting)`, then we emit an `override  def setting = super.setting() ++ Seq(...)`
2. Otherwise, we emit `override def setting = ...` from scratch

This build-file-generation can get arbitrarily sophisticated, so where we want to draw the line is arbitrary. I think this should cover many common usage patterns

ShreckYe:
I think this solution is cool. I will work on it then. To achieve this, I may have to refactor some of Maven and Gradle's build generation code too.

integration/feature/init/src/MillInitSbtTests.scala
ShreckYe: This is fixed after the TODO is resolved. See:

https://github.com/com-lihaoyi/mill/blob/44c7381189454a35ed83040c8dc83e03edaa7297/integration/feature/init/src/MillInitSbtTests.scala#L195-L197

main/init/sbt/test/resources/expected/config/sbt-multi-project-example/build.mill
ShreckYe: Fixed.

https://github.com/com-lihaoyi/mill/blob/1d8011cec7464e1dc87136e29ca4ed8c07fa0451/main/init/sbt/test/resources/expected/config/without-base-project/sbt-multi-project-example/build.mill#L77

integration/feature/init/src/MillInitSbtTests.scala
ShreckYe: I have added the expected `compile` and `test` tests that succeed or fail and comments explaining why in each integration test. A summary:

* library-example's `compile` tasks succeed, but its test mudule using scalaprops is not supported in conversion yet.
* scala-csv v2.0.0's tasks fail because cross-builds supporting multiple Scala versions are not supported in conversion yet.
* scala-csv v1.3.6's `compile` tasks succeed, but its `test` task fails as paths relative to the workspace are used in the test sources such as `new File("src/test/resources/simple.csv")` causing `FileNotFoundException`.
* sbt-multi-project-example's all `compile` and `test` tasks succeed when Mill is run with JDK 11.
* zio-http's tasks fail because common sources shared among multiple platforms (JVM and Scala.js) in "shared" directories in cross-builds are not supported in conversion yet.
* scalaz's `compile` tasks fail because common sources shared among multiple platforms (JVM, Scala.js, and Scala Native) in cross-builds are not supported in conversion yet, its testes use ScalaCheck and are not supported in conversion yet.
* cats is a compilciated build. Its sbt cross projects with `CrossType.Pure` are not imported properly as their platform project directory name start with ".". Most other modules depend on these modules, and therefore don't compile as well.

integration/feature/init/src/MillInitSbtTests.scala
ShreckYe: As expained in https://github.com/com-lihaoyi/mill/pull/4586#discussion_r1970355857.

main/init/buildgen/src/mill/main/buildgen/BuildGenBase.scala
ShreckYe: What do you think of converting all these type parameters into abstract type members? See the TODO in the code for more:

https://github.com/com-lihaoyi/mill/blob/22033e909dd461689b374815ca4602b087b8830f/main/init/buildgen/src/mill/main/buildgen/BuildGenBase.scala#L7-L23

integration/feature/init/src/MillInitSbtTests.scala
ShreckYe: Modules without `CrossType.Pure` are converted now. See the updated code and https://github.com/com-lihaoyi/mill/pull/4586#discussion_r1970355857.

ShreckYe:
Both `baseProject` and https://github.com/com-lihaoyi/mill/pull/4586#issuecomment-2674530292 are supported now.

ShreckYe:
Hi @lihaoyi I have finished the changes requested and I think it's ready for another round of review.

lihaoyi:
@ShreckYe at a glance, looks like a lot of your new tests are failing in CI?

integration/feature/init/src/MillInitSbtTests.scala
```diff
@@ -0,0 +1,332 @@
...
+    /*
+    - 21 KB
+    - sbt 1.5.2
+     */
+    val url = "https://github.com/scalacenter/library-example/archive/refs/tags/v1.0.1.zip"
+
+    test - integrationTest(url)(
+      _.testMillInit(
+        expectedCompileTasks = Some(SplitResolvedTasks(Seq("compile"), Seq())),
+        expectedTestTasks = None // scalaprops not supported in `TestModule`
```
lihaoyi, line 26: Maybe we can pick another example repo that has a supported test framework? e.g. https://github.com/scala/scala3-example-project seems to use munit which is supported by `TestModule.Munit`

integration/feature/init/src/IntegrationTesterExt.scala
```diff
@@ -0,0 +1,78 @@
...
+import utest.*
+
+val defaultInitCommand =
+  Seq("init", "--base-module", "BaseModule", "--deps-object", "Deps", "--merge")
+
+case class SplitResolvedTasks(successful: Seq[String], failed: Seq[String]) {
+  val all = (successful ++ failed).sorted
+}
+
+extension (tester: IntegrationTester)
```
lihaoyi, line 13: Lets move this to a `object InitUtils` object with static methods

integration/feature/init/src/IntegrationTesterExt.scala
```diff
@@ -0,0 +1,78 @@
...
+        for (task <- expected.failed)
+          Predef.assert(!eval(task).isSuccess, s"task $task succeeded")
+      })
+    }
+
+    testAllResolvedTasks("compile", expectedCompileTasks)
+    testAllResolvedTasks("test", expectedTestTasks)
+  }
+
+extension (module: String) {
```
lihaoyi, line 63: Again, let's use normal methods rather than extensions

integration/feature/init/src/MillInitSbtTests.scala
```diff
@@ -0,0 +1,332 @@
...
+            /*
+            `multi1.compile` doesn't work well when Mill is run with JDK 17 and 21:
+            ```text
+            1 tasks failed
+            multi1.compile java.io.IOError: java.lang.RuntimeException: /packages cannot be represented as URI
+                java.base/jdk.internal.jrtfs.JrtPath.toUri(JrtPath.java:175)
+                scala.tools.nsc.classpath.JrtClassPath.asURLs(DirectoryClassPath.scala:183)
+                ...
+            ```
+            Passing a `jvmId` 11 doesn't work here.
```
lihaoyi, line 119: Does this work with JDK 11? If so, could we try writing a `.mill-jvm-version` file to `temurin:11` to see if that makes it work?

main/init/buildgen/src/mill/main/buildgen/BuildGenBase.scala
```diff
@@ -2,68 +2,102 @@ package mill.main.buildgen
...
+  def convertWriteOut(cfg: C, shared: BuildGenUtil.BasicConfig, input: I): Unit = {
     val output = convert(input, cfg, shared)
     writeBuildObject(if (shared.merge.value) compactBuildTree(output) else output)
   }
 
+  /**
+   * A possibly optional module model which might be a parent directory of an actual module without its own sources.
+   */
+  type OM
+  extension (om: OM) def toOption(): Option[M]
```
lihaoyi, line 23: Can we just use `Option[M]` rather than defining this alias?

main/init/buildgen/src/mill/main/buildgen/BuildGenBase.scala
lihaoyi: I think leaving them as type parameters is fine for now. In the end it doesn't really matter, and although such a conversion may or may not be desirable, it's unrelated to SBT and can come in a follow up

main/init/buildgen/src/mill/main/buildgen/BuildGenUtil.scala
```diff
@@ -46,7 +52,7 @@ object BuildGenUtil {
 
   }
 
-  def renderIrPom(value: IrPom): String = {
+  def renderIrPom(value: IrPom | Null): String = {
```
lihaoyi, line 55: Do these `| Null` types get checked by the compiler? Previously we didn't have the `| Null` and it seems we were sometimes passing `null` anyway

integration/feature/init/src/MillInitSbtTests.scala
```diff
@@ -0,0 +1,332 @@
...
+        )),
+        expectedTestTasks = Some(SplitResolvedTasks(
+          Seq(),
+          /*
+          Paths relative to the workspace are used in the test sources such as `new File("src/test/resources/simple.csv")`
+          and they seem to cause the test to fail with Mill:
+          ```text
+          java.io.FileNotFoundException: src/test/resources/simple.csv (No such file or directory)
+          ```
+           */
```
lihaoyi, line 76: Could we try generating `def testSandboxWorkingDir = false` to see if that fixes this issue? Setting that should make the test process run in the repo root IIRC rather than in some sandbox subfolder in `out/`

integration/feature/init/src/MillInitSbtTests.scala
```diff
@@ -0,0 +1,332 @@
...
+          }),
+          expectedTestTasks = Some(SplitResolvedTasks(Seq.empty, submodules.map(_.testTask)))
+        )
+    }
+  }
+}
+
+// relatively large libraries
+
+object MillInitSbtZioHttpTests extends BuildGenTestSuite {
```
lihaoyi, line 135: Let's drop most of these large libraries and just pick 1 large library and try to make it work. 

Apart from the big backend frameworks you've chosen here which probably are very sophisticated, how about some of the other projects listed on https://github.com/EvanLi/Github-Ranking/blob/master/Top100/Scala.md?
https://github.com/gatling/gatling is one we used in our case study (https://mill-build.org/mill/comparisons/sbt.html) that may have a simpler build configuration since it's an application rather than a core framework. https://github.com/lichess-org/lila could be worth a try as well.

Again, I don't expect to have everything compiling and all tests passing, but would be nice to get at least one (1) single test passing to verify that this conversion works at all!

main/init/package.mill
lihaoyi: Let's try to build it with Mill. Adding resources that will be included in the jar is easy enough (https://mill-build.org/mill/scalalib/module-config.html#_common_configuration_overrides), so we just need to look at what the SBT-build jar contains and make sure our Mill-built jar has the same contents

integration/feature/init/src/MillInitSbtTests.scala
lihaoyi: Alternatively we could try a different repo for the "small library" test, e.g. https://github.com/scopt/scopt or https://github.com/scallop/scallop or https://github.com/typelevel/jawn

main/init/package.mill
ShreckYe: OK I will give this a try.

ShreckYe:
> @ShreckYe at a glance, looks like a lot of your new tests are failing in CI?

I will try to fix them. I commented before they finish and haven't run them on Windows.

Some CI tests such as the `example.android` test sometimes fails randomly in my branch but succeeds after a rerun.

main/init/buildgen/src/mill/main/buildgen/BuildGenBase.scala
ShreckYe: I added this so I could refactor less code and get it to work at the time being. I can switch to `Option[M]` but more Maven and Gradle implementation code will be refactored and generalized.

main/init/buildgen/src/mill/main/buildgen/BuildGenUtil.scala
ShreckYe: Yes they do. I see you bumped the Scala version to 3 so I adopted [explicit nulls](https://docs.scala-lang.org/scala3/reference/experimental/explicit-nulls.html) to make the nullable types more explicit.

integration/feature/init/src/MillInitSbtTests.scala
ShreckYe: Yes it does. I will give it a try.

integration/feature/init/src/MillInitSbtTests.scala
lihaoyi: Let's just remove them

main/init/buildgen/src/mill/main/buildgen/BuildGenBase.scala
lihaoyi: Let's go with that then

integration/feature/init/src/MillInitSbtTests.scala
ShreckYe: I added some code to directly write `def testSandboxWorkingDir = false` to the converted `build.mill` file and it works. Shall I update `BuildGenUtil` to render this line for the test modules automatically?

integration/feature/init/src/MillInitSbtTests.scala
ShreckYe: I added a `MillInitSbtGatlingTests` as requested. Some modules `compile` and `test` tasks run successfully on my machine, while others fail because the sbt configuration "compile->compile;test->test" is not supported yet. Shall I try to support this?

lihaoyi: Yes let's try to support `"compile->compile;test->test"`. It means that both the `foo` depends on `bar` and also `foo.test` depends on `bar.test`. Should be straightforward to implement using `moduleDeps` on the main and test modules

integration/feature/init/src/MillInitSbtTests.scala
lihaoyi: Yes let's generate that by default. That flag is meant to ease in migration, so this is exactly what it's meant for

ShreckYe:
Hi I have finshed the changes. Please review again.

The CI test `windows (17, 'integration.{feature,failure}.__.packaged.fork.test')` fails because of a `TimeoutException`:

```text
 X mill.integration.MillInitSbtGatlingTests.0 1200243ms 
[9320-08]   java.util.concurrent.TimeoutException: Future timed out after [600000 milliseconds]
```

I have tested that this passes on my Windows device with JDK 17. Is there anyway to extend CI time for this?

`linux (17, 'example.android.__.local.server.test', true)` somehow fails too, but this is kind of random and has somtimes succeeded with other commits as I haven't edited any code related to Android.

main/init/package.mill
```diff
@@ -63,8 +64,16 @@ object `package` extends RootModule with build.MillPublishScalaModule {
   }
 
   object buildgen extends build.MillPublishScalaModule {
-    def moduleDeps = Seq(build.runner)
+    def moduleDeps = Seq(build.runner /*, tree(build.Deps.scalaVersion)*/ )
     def testModuleDeps = super.testModuleDeps ++ Seq(build.scalalib)
+
+    // I tried moving `Tree` into this module, but it doesn't compile with Scala 2.12.20.
```
lihaoyi, line 70: Could you elaborate on what this is about? I thought this module should be compiled with Scala 3.6.3

main/init/package.mill
```diff
@@ -86,4 +95,36 @@ object `package` extends RootModule with build.MillPublishScalaModule {
...
+        extends Cross[ModelsModule](build.Deps.sbtScalaVersion212, build.Deps.scalaVersion)
+    trait ModelsModule extends build.MillPublishCrossScalaModule {
+      // def moduleDeps = Seq(buildgen.tree())
+      def ivyDeps = Agg(build.Deps.upickle)
+      def compileIvyDeps = Agg(build.Deps.sbt) // for definition references only
+    }
+
+    // no need to publish this
+    // `test` fails with `build.MillScalaModule`, and since it's an SBT plugin project it's most likely not needed.
+    object `sbt-mill-init-export-build` extends ScalaModule {
```
lihaoyi, line 122: We can probably just call this `exportplugin`. It's already in `main.init.sbt` in the Mill codebase, so no need to repeat all of that again

main/init/package.mill
lihaoyi, line 70: Could you elaborate here? What is the compile error? What was the workaround?

lihaoyi:
@ajaychandran could I ask you to help review this? Since it touches a lot of code that you contributed for the Maven and Gradle exporters

integration/feature/init/src/MillInitSbtTests.scala
```diff
@@ -0,0 +1,198 @@
...
+      testMillInit(
+        tester,
+        expectedCompileTasks = Some(SplitResolvedTasks(
+          all = Seq("compile")
+            ++ submodulesWithTests.flatMap(allCompileTasks)
+            ++ submodulesWithoutTests.map(compileTask),
+          failed = Seq(
+            /*
+            `...gatling-benchmarks/src/main/scala/io/gatling/Utils.scala:26:19: no arguments allowed for nullary method toString: (): String`
+            This fails to compile with sbt too. See https://github.com/gatling/gatling/issues/4612.
```
lihaoyi, line 174: Looks like this was fixed! https://github.com/gatling/gatling/commit/40094059e644978624648c261e20723eb05afba7 we can pull it in the latest version and remove this special case

integration/feature/init/src/MillInitSbtTests.scala
```diff
@@ -0,0 +1,198 @@
...
+      "gatling-quicklens",
+      "gatling-samples",
+      "gatling-test-framework"
+    )
+    val submodulesWithTests = submodules diff submodulesWithoutTests
+
+    test - integrationTest(url) { tester =>
+      testMillInit(
+        tester,
+        expectedCompileTasks = Some(SplitResolvedTasks(
```
lihaoyi, line 167: Could we add assertions on the number of source files in the `allSourceFiles` task, for both the `main` and the `test` submodules? That would help us catch any issues with the source folders being mis-configured, resulting in no source files being passed and thus compile/test passing by default

integration/feature/init/src/MillInitUtils.scala
```diff
@@ -0,0 +1,104 @@
...
+      initCommand: Seq[String] = defaultInitCommand,
+      modifyConvertedBuild: () => Unit = () => (),
+      expectedInitResult: Boolean = true,
+      // expectedCompileResult: Boolean,
+      expectedCompileTasks: Option[SplitResolvedTasks],
+      expectedTestTasks: Option[SplitResolvedTasks]
+  ) = {
+    import tester.*
+
+    val initResult = eval(initCommand, stdout = os.Inherit, stderr = os.Inherit)
```
lihaoyi, line 50: Could we not stream the output via `os.Inherit` by default, and only print it on failure? It's very verbose and causing Github Actions to have problems in my browser

integration/feature/init/src/MillInitUtils.scala
```diff
@@ -0,0 +1,104 @@
...
+        if (failed != null) failed else all diff successful
+      )
+  }
+
+  /**
+   * @param expectedCompileTasks [[ None ]] to denote that the `resolve __.compile` task fails
+   * @param expectedTestTasks [[ None ]] to denote that the `resolve __.test` task fails
+   * @return
+   */
+  def testMillInit(
```
lihaoyi, line 39: If we're going to introduce this new helper, let's also use it in `MillInitMavenTests.scala` and `MillInitGradleTests.scala` to keep things consistent

main/init/sbt/models/src/mill/main/sbt/Models.scala
```diff
@@ -0,0 +1,171 @@
...
+    devConnection: Option[String]
+)
+object ScmInfo {
+  implicit val rw: RW[ScmInfo] = macroRW
+}
+
+/**
+ * @see [[sbt.librarymanagement.Developer]]
+ */
+case class Developer(id: String, name: String, email: String, url: URL)
```
lihaoyi, line 104: This seems to duplicate a bunch of stuff in `ir.scala`; can we re-use some of that? Or does the data model diverge enough it can't work?

main/init/package.mill
ShreckYe: OK. I named it this way after sbt's [naming convention](https://www.scala-sbt.org/1.x/docs/Plugins-Best-Practices.html#Artifact+naming+convention).

Shall I add a dash between the 2 words and make it `export-plugin`? Or what about still following the naming convetion and making it shorter like `sbt-export`?

main/init/package.mill
ShreckYe: Here is the full error message after I uncommented the module and move `Tree` into it:

```text
[45/45] main.init.buildgen.tree[2.12.20].compile
[45] [info] compiling 1 Scala source to .../mill/out/main/init/buildgen/tree/2.12.20/compile.dest/classes ...
[45] [error] .../mill/main/init/buildgen/tree/src/mill/main/buildgen/Tree.scala:11:7: object api is not a member of package mill
[45] [error] @mill.api.experimental
[45] [error]       ^
[45] [error] .../mill/main/init/buildgen/tree/src/mill/main/buildgen/Tree.scala:23:32: not found: type IterableOnce
[45] [error]   def transform[Out](f: (Node, IterableOnce[Tree[Out]]) => Tree[Out]): Tree[Out] = {
[45] [error]                                ^
[45] [error] .../mill/main/init/buildgen/tree/src/mill/main/buildgen/Tree.scala:30:7: object api is not a member of package mill
[45] [error] @mill.api.experimental
[45] [error]       ^
[45] [error] .../mill/main/init/buildgen/tree/src/mill/main/buildgen/Tree.scala:34:61: not found: type IterableOnce
[45] [error]   def from[Input, Node](start: Input)(step: Input => (Node, IterableOnce[Input])): Tree[Node] = {
[45] [error]                                                             ^
[45] [error] .../mill/main/init/buildgen/tree/src/mill/main/buildgen/Tree.scala:37:23: value iterator is not a member of Any
[45] [error]       Tree(node, next.iterator.map(recurse).toSeq)
[45] [error]                       ^
[45] [error] .../mill/main/init/buildgen/tree/src/mill/main/buildgen/Tree.scala:37:36: missing argument list for method recurse
[45] [error] Unapplied methods are only converted to functions when a function type is expected.
[45] [error] You can make this conversion explicit by writing `recurse _` or `recurse(_)` instead of `recurse`.
[45] [error]       Tree(node, next.iterator.map(recurse).toSeq)
[45] [error]                                    ^
[45] [error] 6 errors found
[45/45] ============================== main.init.buildgen.tree[2.12.20].compile ==============================
1 tasks failed.buildgen.tree[2.12.20].compile
main.init.buildgen.tree[2.12.20].compile Compilation failed

```

So the main reason seems to be that `IterableOnce` is not available in Scala 2.12. However, I think this is not important anymore and I can remove this block of comments here, because I first tried to move `Tree` to a dependency module of the `sbt/models` module and reuse it to serialize the subprojects, but it turns out that sbt offers a list of subprojects rather than a tree and doing this is no longer needed.

integration/feature/init/src/MillInitSbtTests.scala
```diff
@@ -0,0 +1,198 @@
...
+    /*
+    - 34 KB
+    - originally sbt 1.10.0
+     */
+    val url = "https://github.com/tototoshi/scala-csv/archive/refs/tags/2.0.0.zip"
+
+    test - integrationTest(url) { tester =>
+      bumpSbtTo1107(tester.workspacePath)
+
+      // Cross-builds are not supported yet.
```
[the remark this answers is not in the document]
ShreckYe: v2.0.0 is a cross-build [supporting multiple Scala versions](https://github.com/tototoshi/scala-csv/tree/2.0.0/src/main) and v1.3.6 [is not](https://github.com/tototoshi/scala-csv/tree/1.3.6/src/main).

Shall I update the comments in the code about this?

integration/feature/init/src/MillInitSbtTests.scala
ShreckYe: It's not released yet though. Shall I change the version tag to a commit in the url to achieve this?

integration/feature/init/src/MillInitSbtTests.scala
ShreckYe: OK I will try.

main/init/sbt/models/src/mill/main/sbt/Models.scala
ShreckYe: I think we call reuse some but others diverge. For example, `organization` refers to the Maven group in both sbt and Mill, but it refers to the name of the organization in Maven and Gradle, equivalent to sbt's `organizationName`.

Another difficulty here is that the IR models depend on some other modules and libraries such as `build.runner` and OS-Lib, and don't have uPickle configured.

Also to reduce duplication I think it's also needed to remove and unify some Java models in the Gradle module, as I commented last time reviewing the Gradle PR.

So in overall, I think this may involve some work can come in a cleanup and refactor PR after this.

ajaychandran:
> @ajaychandran could I ask you to help review this? Since it touches a lot of code that you contributed for the Maven and Gradle exporters

@lihaoyi I am working on an alternate solution (with a goal to support cross Scala modules). I will help with the review if my attempt fails or if this PR is selected.

lihaoyi:
`MillInitSbtGatlingTests` seems to be timing out in CI after 10 minutes. Does it take 10 minutes to run locally as well? I wonder if there's some way we can speed it up, since even if it passes 10 minutes to run the tests is pretty long

ShreckYe:
All the sbt tests together take about 9 minutes on my machine':

```text
[3552] integration.feature[init].local.server.test 530s
[3552-0] + mill.integration.MillInitSbtGatlingTests.0 539399ms
[3552-0] Tests: 1, Passed: 1, Failed: 0
```

 I can try concatenating all the successful task commands into one command and see if that solves the problem.

lihaoyi:
Yeah that sounds reasonable, by combining them into a single command you should get some parallelism that would speed things up a bit, and reduce the overhead of repeated commands

ShreckYe:
I have removed the special case and combined the successful commands to a `__.compile` command in `MillInitSbtGatlingTests` but it still times out. Shall I just skip this test on Windows like how it's done in `MillInitGradleEhcache3Tests`?

lihaoyi:
If you have a windows machine would be good to try and debug it, but if not can just skip it for now

core/constants/src/mill/constants/Util.java
```diff
@@ -10,8 +10,9 @@
 
 public class Util {
 
-  public static boolean isWindows =
-      System.getProperty("os.name").toLowerCase(Locale.ROOT).startsWith("windows");
+  private static String lowerCaseOsName = System.getProperty("os.name").toLowerCase(Locale.ROOT);
```
lefou, line 13: I'd rather not keep this as a field. If we want to be DRY, make it a private method.

ShreckYe:
Finshed with the changes then. Please review again.

integration/feature/init/src/MillInitUtils.scala
ShreckYe: I have done this for the tests except for these 3 Maven tests, as their testing logic is quite different making assertions on whether thether `out` or `err` contain certain strings:

https://github.com/com-lihaoyi/mill/blob/8ba432279ac643d98e8484aeae3e92b8d7b83f6f/integration/feature/init/src/MillInitMavenTests.scala#L14-L133

integration/feature/init/src/MillInitSbtTests.scala
```diff
@@ -1,16 +1,21 @@
...
-  def bumpSbtTo1107(workspacePath: os.Path) =
-    // bump sbt version to resolve compatibility issues with lower sbt versions and higher JDK versions
-    os.write.over(workspacePath / "project" / "build.properties", "sbt.version = 1.10.7")
+  /**
+   * bump sbt version to our tested version to resolve compatibility issues with lower sbt versions and higher JDK versions
+   */
+  def bumpSbt(workspacePath: os.Path) =
+    os.write.over(
+      workspacePath / "project" / "build.properties",
+      s"sbt.version = ${sys.props("TEST_SBT_VERSION")}"
```
lefou, line 18: Unfortunately, we need this ugly `getOrElse`, as `sys.props` is returning `null` for undefined properties, making this issue of a missing property hard to diagnose.

```suggestion
      s"sbt.version = ${sys.props.getOrElse("TEST_SBT_VERSION", ???)}"
```

ShreckYe:
Updated.

One of the 2 CI test on Linux with JDK 11 failed randomly and the other somehow got cancelled. They should work after some reruns.

main/init/sbt/src/mill/main/sbt/SbtBuildGenMain.scala
```diff
@@ -123,7 +123,7 @@ object SbtBuildGenMain
     } catch {
       case e: os.SubprocessException => throw RuntimeException(
           "The sbt command to run the `millInitExportBuild` sbt task has failed, " +
-            s"please update the project's sbt version to the latest or our tested version v${sys.props("TEST_SBT_VERSION")}, " +
+            s"please update the project's sbt version to the latest or our tested version v${sys.props.getOrElse("TEST_SBT_VERSION", ???)}, " +
```
lefou, line 126: It's not a good idea to throw an exception while we're about to construct another exception. Also, this is not in the context of the test project, so the `TEST_SBT_VERSION` variable is undefined. 

If we want to hint the sbt version used at Mill build time, we need some generated constant, e.g. by using the `BuildInfo` module. Here is an usage example:

https://github.com/com-lihaoyi/mill/blob/59764e88a2168db8eeb61ecc0f10a171f4c160e4/scalalib/src/mill/scalalib/ZincWorkerModule.scala#L133

It's defined here:

https://github.com/com-lihaoyi/mill/blob/59764e88a2168db8eeb61ecc0f10a171f4c160e4/scalalib/package.mill#L45

ShreckYe: Sorry I didn't realize this. Fixed.

lefou (commented):
Looks good to me. I didn't test it manually though.

main/init/buildgen/src/mill/main/buildgen/BuildGenUtil.scala
```diff
@@ -428,7 +538,13 @@ object BuildGenUtil {
   val testModulesByGroup: Map[String, String] = Map(
     "junit" -> "TestModule.Junit4",
     "org.junit.jupiter" -> "TestModule.Junit5",
-    "org.testng" -> "TestModule.TestNg"
+    "org.testng" -> "TestModule.TestNg",
+    "org.scalatest" -> "TestModule.ScalaTest",
+    "org.specs2" -> "TestModule.Specs2",
+    "com.lihaoyi.utest" -> "TestModule.UTest",
```
sake92, line 544: Shouldnt this be just com.lihaoyi?

main/init/buildgen/src/mill/main/buildgen/BuildGenUtil.scala
```diff
@@ -428,7 +538,13 @@ object BuildGenUtil {
...
+    "org.scalameta" -> "TestModule.Munit",
+    "com.disneystreaming" -> "Weaver",
```
sake92, line 546: TestModule.Weaver ?

lihaoyi:
@ShreckYe so I did some manual testing, and hit an issue trying to import https://github.com/ruippeixotog/scala-scraper. It seems the dependency from the `modules/config` module on the `core` module (defined by SBT [here](https://github.com/ruippeixotog/scala-scraper/blob/bee9c02ae2bcd73960d604f8a5c981bcd1f5b107/build.sbt#L42)) is being dropped, and `modules/config/package.mill` is generating a dependency on the (empty) root module instead `def moduleDeps = super.moduleDeps ++ Seq(build)`. Do you have any idea why this might be happening?

ShreckYe:
> @ShreckYe so I did some manual testing, and hit an issue trying to import https://github.com/ruippeixotog/scala-scraper. It seems the dependency from the `modules/config` module on the `core` module (defined by SBT [here](https://github.com/ruippeixotog/scala-scraper/blob/bee9c02ae2bcd73960d604f8a5c981bcd1f5b107/build.sbt#L42)) is being dropped, and `modules/config/package.mill` is generating a dependency on the (empty) root module instead `def moduleDeps = super.moduleDeps ++ Seq(build)`. Do you have any idea why this might be happening?

I don't see the reason at a glance. I will try to debug and find out.

lihaoyi:
Also tried this on the https://github.com/foundweekends/giter8 and it works great. Some manual fixes needed (`launcher` module needed to be renamed to avoid conflicting with the task, `org.scala-sbt:sbt` dependency is added by a plugin so needed to be added manually, `plugin.compile` fails due to some weird magic) but otherwise most things compile and all tests pass (or at least pass as much as they do with SBT).

lihaoyi:
I tried this on https://github.com/nscala-time/nscala-time and it doesn't seem to be picking up the `test` module. Seems like it's generating a `trait test extends SbtTests` rather than an `object`

main/init/buildgen/src/mill/main/buildgen/BuildGenUtil.scala
ShreckYe: Thanks for pointing out. Fixed.

ShreckYe:
> I tried this on https://github.com/nscala-time/nscala-time and it doesn't seem to be picking up the `test` module. Seems like it's generating a `trait test extends SbtTests` rather than an `object`. Is it because we don't have `Scalacheck` in the list of `TestModule`s? Could we add it?

I remember seeing this too at some point but didn't reproduce it later after some iterations. I will try to find out why it happens.

lihaoyi:
@ShreckYe manual testing is fine; integration tests are pretty expensive since they run every time and slow down CI, so we want to minimize them even at the cost of test coverage

ShreckYe:
> @ShreckYe so I did some manual testing, and hit an issue trying to import https://github.com/ruippeixotog/scala-scraper. It seems the dependency from the `modules/config` module on the `core` module (defined by SBT [here](https://github.com/ruippeixotog/scala-scraper/blob/bee9c02ae2bcd73960d604f8a5c981bcd1f5b107/build.sbt#L42)) is being dropped, and `modules/config/package.mill` is generating a dependency on the (empty) root module instead `def moduleDeps = super.moduleDeps ++ Seq(build)`. Do you have any idea why this might be happening?

This is caused by the `core` project with a setting `name := "scala-scraper" having the same GAV ID as the root project.

In its `build.sbt` there is a workaround of not publishing the root project to make it work with sbt:

```sbt
// do not publish the root project
publish / skip := true
```

I don't see an easy and general way to fix this issue though: we can't just ignore projects with `publish / skip := true`; and I don't think either project (the root project or `core`) takes precedence over the other.

lihaoyi:
What about if there is a conflict between two modules and one has `publish / script := true` and the other doesn't, we ignore the one that is skipped and use the other one?

lihaoyi:
Perhaps another question: looking at that repo, only the `core` module has `name := "scala-scraper"`, and the `root` module has no `name` assigned. How can `root` have the same GAV as `core` when one has a name and the other doesn't?

ShreckYe:
> Perhaps another question: looking at that repo, only the `core` module has `name := "scala-scraper"`, and the `root` module has no `name` assigned. How can `root` have the same GAV as `core` when one has a name and the other doesn't?

If I understand correctly, in sbt a root project is added by default, and each project takes the name of its directory if it's not set explicitly. So in this case the root project takes the directory name `scala-scraper`, which is the same as the `core` project's, and both of them then have the same GAV `net.ruippeixotog:scala-scraper:version`.

ShreckYe:
> What about if there is a conflict between two modules and one has `publish / script := true` and the other doesn't, we ignore the one that is skipped and use the other one?

I can try to implement this approach but I think it's a bit niche and not straighforward to the user, as `publish` is only one of the publish tasks, it's not the SSOT for all the publish settings as there are others such as `publishLocal` and `publishM2`. This is also the reason I just let all the converted modules extend `PublishModule` instead of judging from `publish / skip`.

ShreckYe:
I just checked again and I was wrong about this, `publishLocal / skip` depends on `publish / skip`.

lihaoyi:
@ShreckYe FYI I have pushed fixed for `test-mill-bootstrap.sh`, pull it down before you make further changes so you preserve them

> I have been going through the related sbt settings, and a better approach as it occurs to me now, might be to use buildDependencies and get ProjectRefs instead of ModuleIDs for the inter-project dependencies, instead of retrieving the ModuleIds and getting the corresponding projects from a GAV-project map.

Let's try implementing that in this PR and see if we can get it to work. It does indeed seem like a better way of doing things, since unlike Maven, SBT and Gradle don't always publish all modules as artifacts by default, so it seems likely to get conflicts of the sort I hit in `scala-scraper`

ShreckYe:
OK no problem. Shall I do this just for sbt here in this PR or for both Gradle and sbt?

lihaoyi:
Try doing it for both; hopefully a lot of the code is very similar, and you've reviewed the gradle PR so you should be somewhat familiar. But if you can't get gradle to work then at least let's do it for SBT

sake92:
I think I used that approach in my PR https://github.com/com-lihaoyi/mill/pull/4575
A bit ugly with mutable variables.. might be useful

ShreckYe:
@lihaoyi I've finished the updates. `test-mill-bootstrap.sh` still fails and also conflicts with the main branch. Could you check what's wrong?

lihaoyi:
@ShreckYe thanks I'll take a look, and I'll fix up the bootstrap test for you

lihaoyi:
I think this is ready to merge. @ShreckYe can you update the PR description to bring it up to date with the current state of the PR after all the discussion and changes, and open another PR to backport this to 0.12.x

lihaoyi (approved):

main/init/sbt/src/mill/main/sbt/SbtBuildGenMain.scala
```diff
@@ -0,0 +1,507 @@
...
+      else
+        throw new RuntimeException(s"No system-wide `$systemSbt` found")
+    } else {
+      val systemSbt = "sbt"
+      // resolve the sbt executable
+      // https://raw.githubusercontent.com/paulp/sbt-extras/master/sbt
+      if (os.exists(workspace / "sbt"))
+        "./sbt"
+      else if (os.exists(workspace / "sbtx"))
+        "./sbtx"
```
ajaychandran, line 84: An alternative would be to add a `customSbt` config argument.

ShreckYe: Added and tested manually. Thanks.

ShreckYe:
I have updated the PR description, taken Ajay's review suggestion, and added the backport PR.

There are some more possible changes for this PR I found while backporting, as listed below. Shall I add them in this PR?

1. The `OptionNodeTree.scala` file contains the difinitions directly instead of in an `Utils` object. Shall I put them in an `Utils` object?
1. Also in `OptionNodeTree.scala`, there are some curly braces missing and the code is parsed based on Scala 3's support for indentation-based parsing / optional braces, which I didn't notice. Shall I add the curly braces?
1. The type parameter `T` in `Node` has to be made covariant in the backport PR as at some places `Node[Some[...]]` is inferred where `Node[Option[...]]` is needed with Scala 2's older type inference. Shall I do this in this PR too?

lihaoyi:
@ShreckYe thanks, will take a look at the two PRs.

I think for now we can punt on your proposed changes. They're nice to have, but honestly not a big deal, and the PR is big enough we can clean up later if necessary

lihaoyi:
@ShreckYe seems like both PRs have some consistent failures. The android job you can ignore but could you look into the other failure on each? We need to get that green before we can merge them

ShreckYe:
The `gatling-http-client.test` fails in the converted project probably due to some new changes in `main` as it used to pass. I have manually run `gatling-http-client.test` with my system Mill installed with Homebrew (0.12.5) and it passes, so the conversion logic should be fine. When I run `gatling-http-client.test` with `mill-assembly.jar` built with `./mill dist.installLocal` it fails with:

```text
[213-5] Test io.gatling.http.client.proxy.HttpsRequestThroughHttpProxyTest failed: java.io.IOException: Failed to bind to 0.0.0.0/0.0.0.0:8888, took 1.038s
[213-5]     at org.eclipse.jetty.server.ServerConnector.openAcceptChannel(ServerConnector.java:349)
[213-5]     at org.eclipse.jetty.server.ServerConnector.open(ServerConnector.java:310)
[213-5]     at org.eclipse.jetty.server.AbstractNetworkConnector.doStart(AbstractNetworkConnector.java:80)
[213-5]     at org.eclipse.jetty.server.ServerConnector.doStart(ServerConnector.java:234)
[213-5]     at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:73)
[213-5]     at org.eclipse.jetty.server.Server.doStart(Server.java:401)
[213-5]     at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:73)
[213-5]     at io.gatling.http.client.proxy.HttpsRequestThroughHttpProxyTest$HttpsProxy.start(HttpsRequestThroughHttpProxyTest.java:60)
[213-5]     at io.gatling.http.client.proxy.HttpsRequestThroughHttpProxyTest.start(HttpsRequestThroughHttpProxyTest.java:74)
[213-5]     at java.lang.reflect.Method.invoke(Method.java:569)
[213-5]     at java.util.ArrayList.forEach(ArrayList.java:1511)
[213-5] Caused by: java.net.BindException: Address already in use
[213-5]     at sun.nio.ch.Net.bind0(Native Method)
[213-5]     at sun.nio.ch.Net.bind(Net.java:555)
[213-5]     at sun.nio.ch.ServerSocketChannelImpl.netBind(ServerSocketChannelImpl.java:337)
[213-5]     at sun.nio.ch.ServerSocketChannelImpl.bind(ServerSocketChannelImpl.java:294)
[213-5]     at sun.nio.ch.ServerSocketAdaptor.bind(ServerSocketAdaptor.java:89)
[213-5]     at org.eclipse.jetty.server.ServerConnector.openAcceptChannel(ServerConnector.java:344)
[213-5]     ... 10 more
[213-5] Test junit-jupiter finished, took 1.062s
[213-5] Test run finished: 1 failed, 0 ignored, 1 total, 1.106s
```

Do you have ideas what change broke this? Or shall I just move it to the failed tests.

ShreckYe:
I think the port binding failure is probably due to some changes that enable the tests to run in parallel. I browsed the code and I see that both `HttpRequestThroughHttpProxyTest` and `HttpsRequestThroughHttpProxyTest` bind to port 8888.

ShreckYe:
I had a look here too and both `./mill example.scalalib.__.local.server.test` and `integration.{feature,failure}.__.packaged.fork.test` on CI seem to have failed due to network issues.

lihaoyi:
Merged both PRs, thanks @ShreckYe for all your work! I'll send the bounty to the same bank details used for the previous payment, unless you have any other instructions for the transfer

ShreckYe:
A pleasure work on this and learnt something myself 😄. Thanks! The bank details are the same.

Issue #4414, opened 2025-01-28
Make `mill`/`mill.bat` launcher scripts fall back to non-`native` binary when used on unsupported OS/CPU platforms 


lihaoyi:
In particular, we need to make sure that `mill.bat` ignores the `-native` suffix when used on Windows/ARM, at least until https://github.com/oracle/graal/issues/9215 is resolved upstream

lefou:
If we would add support for a `-jvm` suffix, we could let the launcher autmatically decide which platform to choose, when it find a pure version (without any platform suffix).

PR #4589, opened 2025-02-19
Ignore `-native` suffix on Windows-ARM and fall back to JVM

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/4414

We only need to do this on windows, since on Linux and Mac we support both ARM and Intel


lihaoyi:
This was tested manually on my microsoft surface laptop 7

PR #4591, opened 2025-02-19
Update coursier to 2.1.25-M2 (adds early Gradle Module support)

alexarchambault:


alexarchambault:
I don't like the new error message when the version is empty (`No version available in (,)`, while it was something like this before: `not found: https://repo1.maven.org/maven2/com/google/protobuf/protobuf-java//protobuf-java-.pom`). Going to try to fix that before coursier `2.1.25` final.

lihaoyi:
Looks good to me, I assume usage of the gradle module support will come in follow ups

lihaoyi (approved):
Looks good to me

Issue #3693, opened 2024-10-09
`RootModule` and `Cross` cannot be applied to the same module


lihaoyi:
Both of them are `class`s, so neither one can be mixed in.

This probably can't be fixed in Scala 2, as they both need to take parameters, and so cannot be traits. This may be fixable in Scala 3 which allows trait parameters

PR #4593, opened 2025-02-19
feat: RootModule and Cross can be applied to the same module

jbwheatley:
Implements #3693 

These seem to be the minimal changes to allow RootModule and Cross to be applied to the same module. 
it was only necessary to make `Cross` a trait. I did try making `RootModule` also a trait but it proved a little trickier due to it usign the class constructor for `BaseModule` that it extends. 

The main issue I ran into that caused me to have to remove the implicit parameter on `Cross` was the Ctx implicit was not being filled in properly and i was seeing `Modules and Tasks can only be defined within a mill Module` from the dummy being used. There may have been a way around this in the `CrossMacros`, but I couldn't see it, and this solution appears to work, at least for the examples shown in the integration tests added.

### DIFF
--- a/core/define/src/mill/define/Cross.scala
+++ b/core/define/src/mill/define/Cross.scala
@@ -158,9 +158,9 @@ object Cross {
  * }
  * }}}
  */
-class Cross[M <: Cross.Module[?]](factories: Cross.Factory[M]*)(implicit
-    ctx: mill.define.Ctx
-) extends mill.define.Module {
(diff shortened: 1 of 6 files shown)


jbwheatley:
all checks passing on my fork: https://github.com/jbwheatley/mill/pull/1

lihaoyi:
Thanks @jbwheatley , please email me your wire transfer details to haoyi.sg@gmail.com and I'll transfer the bounty

PR #4597, opened 2025-02-20
Provide default JVM version to be used by Mill server process, when `.mill-jvm-version` and system `java` are not provided

lihaoyi:
This PR makes Mill's server use a fixed JVM version by default in the case where no other Java version is available:

1. No `.mill-jvm-version` file present
2. No `java` command available environmentally
3. The Mill client has no `java.home` set (i.e. it is a graal native image)

This makes running Mill on a clean environment without a JVM installed work by default. Previously, it either needed you to define a `.mill-jvm-version` or install a JVM, either of which is a manual step we now can now avoid. Now it instead uses the version of Java configured in Mill's own `.mill-jvm-version` file

We still allow any system `java` command to take precedence over the default, since that's what most people would expect when using the JVM.

Added an integration test in CI `no-jvm-bootstrap` that runs Mill with the native launcher with no `.mill-jvm-version` file and no system `java` installed to verify that it falls back to the Mill default JVM version. Needed to add some hacky test hooks to make sure we exercise the code path in question


runner/client/src/mill/runner/client/MillProcessLauncher.java
```diff
@@ -114,12 +114,17 @@ static boolean isWin() {
...
 
   static String javaHome() throws IOException {
-    String jvmId;
+    String jvmId = null;
     Path millJvmVersionFile = millJvmVersionFile();
 
     String javaHome = null;
     if (Files.exists(millJvmVersionFile)) {
       jvmId = Files.readString(millJvmVersionFile).trim();
+      if (jvmId.equals("global")) jvmId = null;
```
alexarchambault, line 123: FYI in coursier, `system` stands for "use the already installed / available JVM". Don't know which is better, if `global` feels natural to enough people, I could accept it in coursier too.

runner/client/src/mill/runner/client/MillProcessLauncher.java
```diff
@@ -114,12 +114,17 @@ static boolean isWin() {
...
+    } else {
+      jvmId = "zulu:17.0.3";
```
alexarchambault, line 125: Is there a reason to use `zulu` rather than `temurin`? I recall defaulting to `zulu` sometimes, because it has JDK 8 builds for macOS ARM64 while `temurin` doesn't, but that shouldn't be a problem here.

runner/client/src/mill/runner/client/MillProcessLauncher.java
lihaoyi: `system` works too, I was unaware of Coursier's default. I can change this over

runner/client/src/mill/runner/client/MillProcessLauncher.java
lihaoyi: Here is the issue: Temurin doesn't publish JDKs for versions <21 on Windows-ARM:

- https://github.com/coursier/jvm-index/blob/master/indices/windows-arm64.json

lihaoyi:
Seems like CI is hitting `No space left on device` errors with this PR. I assume it has to be due to us leaking downloaded JVMs somewhere in Coursier, and now that we're calling Coursier every integration/example test it's adding up enough to be a problem. @alexarchambault probably will need your help to look into coursier see if we're leaking any temporary download files or something

alexarchambault:
Going to try to tweak how `COURSIER_ARCHIVE_CACHE` is handled in workflows in a separate PR

lihaoyi:
Good catch with the `COURSIER_ARCHIVE_CACHE`, maybe the empty value is making it live in the `cwd` rather than just having the value unset.

lihaoyi:
Let me see if setting the default in github actions to `null` instead of `""` helps

lihaoyi:
Looks like passing in `null` doesn't seem to help https://github.com/lihaoyi/mill-1/actions/runs/13458684002/job/37608878277

lihaoyi:
Seems like setting `coursierarchive` to `/tmp` seems to have worked! A few other errors to sort out, but at least it seems we're unblocked now

lihaoyi:
One possible point of confusion with this PR if you use Mill without configuring a java version on your `JavaModule`s, and without configuring a `.mill-jvm-version` file in your build, you will get the Mill default `zulu:17.0.3` JVM rather than your environmentally-installed `java` JVM. 

This would be pretty confusing, and maybe in the absence of explicit configuration it's worth making application code build/run/tested with mill default to the system `java` rather than the Mill default. Not sure TBH

lihaoyi:
Another more lightweight change would be to only fall back to the hardcoded `zulu:17.0.3` if there is no existing JVM installed. That would preserve the ability to use whatever JDK people have installed, while also keeping the Mill JDK consistent with the `JavaModule` default JDK, and avoid downloading two separate JDKs. 

I'm not sure how to check whether there is already a JVM available, apart from shelling out to `java -help` which is pretty expensive (50-100ms). @alexarchambault do you know if there's any easy way to do such a check?

alexarchambault:
Both coursier and Scala CLI do such things. I can't have a look right now, but IIRC, they might just look at `JAVA_HOME` or for a `java` binary in `PATH` (which needs quite some care on Windows, with `PATHEXT` and all), and maybe read the content of a file under `JAVA_HOME` 🤔

alexarchambault:
coursier and Scala CLI don't require a minimum Java version (they do, like >= 8, but don't really check it IIRC), but Scala CLI does look at the Java version when it spawns a Bloop server (it needs >= 17).

lihaoyi:
Decided to go with the more conservative approach of just providing a fallback default if there is no JVM available.

Issue #4469, opened 2025-02-04
`mill.contrib.bloop.Bloop/install` run very slow in some case


harry-xi:
In project [ScAllay/ScalaPluginTemplate-WithMill](https://github.com/ScAllay/ScalaPluginTemplate-WithMill) it always take me more then 10 minutes to run it after I change the list of dependencies (whether removing or adding). When using sbt with bloop in a project with similar dependencies, It takes less than a minute to do it. (you can find the project [here](https://github.com/AllayMC/ScalaPluginTemplate))

I checked the code for this task and didn't find anything significant. But I still have a suspicion that maybe it re-downloads all the dependencies when there is any change in the dependency list.

lefou:
Have you tried the Mill built-in BSP server, which is the preferred BSP server and can be installed with `mill mill.bsp.BSP/install`. See also https://mill-build.org/mill/cli/installation-ide.html#_ide_support

harry-xi:
I haven't traid it. If you think a try to it can help to find out the problem,l will try it as soon as I have time (maybe a week later).
What more I'm workin on vscode, so bloop seems like my only choice.

lefou:
You most likely using Metals in vscode, which connects to Mill via BSP the Build Server Protocol. It requires a BSP Server to run. Bloop is on BSP server that is independent of Mill, hence the contrib plugin needs to generate a config for it. Mill BSP is another BSP server, directly supported by Mill. Since not all concepts of Mill can be mapped easily to the Bloop config files, and the bloop contrib plugin haven't seen much contributions (e.g. it is still using a `bloop-config-1.5.5`), you should definitely give Mill BSP a try. It will also heloful to identify the issues you see. E.g. is Mill BSP also taking that long?

harry-xi:
> You most likely using Metals in vscode, which connects to Mill via BSP the Build Server Protocol. It requires a BSP Server to run. Bloop is on BSP server that is independent of Mill, hence the contrib plugin needs to generate a config for it. Mill BSP is another BSP server, directly supported by Mill. Since not all concepts of Mill can be mapped easily to the Bloop config files, and the bloop contrib plugin haven't seen much contributions (e.g. it is still using a `bloop-config-1.5.5`), you should definitely give Mill BSP a try. It will also heloful to identify the issues you see. E.g. is Mill BSP also taking that long?

this problem didn't repoduce in mill bsp.

lefou:
[...]

PR #4600, opened 2025-02-20
Improve performances of `Bloop/install`

Baccata:
Prevents having to iteratively querying the location of each artifact's resolved files. This makes a big difference in large mill projects.

### DIFF
--- a/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
@@ -317,63 +317,45 @@ class BloopImpl(evs: () => Seq[Evaluator], wd: os.Path) extends ExternalModule {
       import coursier._
       import coursier.util._
 
-      def source(r: Resolution) = Resolution(
-        r.dependencies
-          .map(d =>
-            d.withAttributes(
-              d.attributes.withClassifier(coursier.Classifier("sources"))
-            )
-          )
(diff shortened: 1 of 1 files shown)


lolgab:
Does this close https://github.com/com-lihaoyi/mill/issues/4469 ?
If not it seems for sure related.

lihaoyi:
CC @alexarchambault and @lefou , not sure if you have any opinions on this since you know more about IDE stuff than I do. Otherwise it looks good to me

lefou (approved):
Looks reasonable to me.

alexarchambault (commented):
This switches that bit of code from the v1 to the v2 API of coursier, which is nice. The former code was apparently running the resolution twice, so that's better now 👍

Baccata:
Arguably it could be made event better if the task graph of java/scala modules were keeping the coursier dependencies around alongside the paths to the downloaded artifacts ... but 🤷‍♂️

alexarchambault:
> Arguably it could be made event better if the task graph of java/scala modules were keeping the coursier dependencies around alongside the paths to the downloaded artifacts ... but 🤷‍♂️ 

https://github.com/com-lihaoyi/mill/pull/4608 adds helpers to get artifact details without calling too directly the coursier API, that could help here.

We could try adding tasks keeping all those details, but I fear these might be quite heavyweight 🤔

PR #4606, opened 2025-02-21
Don't change COURSIER_ARCHIVE_CACHE at all on CI

alexarchambault:
That might be a problem on Windows, might need to add some of this stuff back

### DIFF
--- a/.github/actions/post-build-selective/action.yml
+++ b/.github/actions/post-build-selective/action.yml
@@ -8,10 +8,6 @@ inputs:
     required: true
     type: string
 
-  coursierarchive:
-    default: ''
-    type: string
-
 runs:
   using: "composite"
   steps:
@@ -38,8 +34,6 @@ runs:
     #  shell: ${{ inputs.shell }}
 
     - run: ./mill -i -k selective.run ${{ inputs.millargs }}
-      env:
-        COURSIER_ARCHIVE_CACHE: ${{ inputs.coursierarchive }}
       shell: ${{ inputs.shell }}
 
     - name: Publish Test Report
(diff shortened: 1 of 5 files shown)


lihaoyi:
IIRC we tweaked `COURSIER_ARCHIVE_CACHE` because Graal native image was hitting path length issues where the default `COURSIER_ARCHIVE_CACHE` + the coursier http URL suffix was >256 characters. Not sure if there's some other way to work around that issue, but that's why we starting fiddling with it to begin with

alexarchambault:
I had the same problem in coursier and Scala CLI, where I added [hacks](https://github.com/coursier/coursier/blob/5cc65ac9ec581c347a7fe531ea8d4a973b162099/build.sc#L1106) to copy the GraalVM distrib under a shorter path

alexarchambault:
Strangely enough, CI [seems fine](https://github.com/alexarchambault/mill/actions/runs/13458224478/job/37606933837) with the changes here

alexarchambault:
(undrafting to check if the CI in the main repo is green too)

lihaoyi:
If CI is green without this hack maybe we can just remove it? But it's puzzling why the problem would just go away on its own

lihaoyi:
Oh, I think we may be getting confused by selective testing. Mill doesn't know that `.github/` changes can affect test runs, and so skips all the tests. If you add the `run-all-tests` label and push a commit that should force the tests to actually run

lihaoyi:
Yeah if you look at the `selective.resolveTree` output in https://github.com/com-lihaoyi/mill/actions/runs/13458812216/job/37608835354?pr=4606 you'll see no tests actually got selected due to lack of code changes

lihaoyi:
Superseded by https://github.com/com-lihaoyi/mill/pull/4597

PR #4607, opened 2025-02-21
Change millBinPlatform to 0.13

joan38:
This will unblock lib maintainers to start supporting M versions


build.mill
```diff
@@ -317,7 +317,7 @@ def millBinPlatform: T[String] = Task {
   //  val pos = if (tag.startsWith("0.")) 2 else 1
   //  tag.split("[.]", pos + 1).take(pos).mkString(".")
   // }
-  "0.11"
+  "0.13"
```
lefou, line 320: Instead of hardcoding, removing the hardcoded version and uncommenting the logic based on `millLastTag` seems what you want.

lihaoyi:
Since unstable releases don't promise any compatibility forward or backwards, merging this is basically a no-op in terms of what we promise right?

lefou:
> Since unstable releases don't promise any compatibility forward or backwards, merging this is basically a no-op in terms of what we promise right? At least until we cut `0.13.0` final

This only hold's true if you're not using any thirdparty plugins. But the whole point of the binary platform is supporting the use on thirdparty plugins. Even if there is no guarantee that a `-MXXX` bin platform is compatible, we want to detect incompatible bin platforms. So we try to bump the `-M<count>` whenever we know we changed the bin-compatibility. So using a meaningless number wouldn't non only help anybody, it would break any confidence we gave our early adopters in the past.

lefou (approved):
Looks good to me.

PR #4608, opened 2025-02-21
Pass Mill source JARs to Bloop

alexarchambault:
The goal of this PR is to pass source JARs corresponding to the Mill class path to Bloop. These are added in the root modules, where they're needed to get go-to-source to work for Mill build files.

### DIFF
--- a/contrib/bloop/src/mill/contrib/bloop/Bloop.scala
+++ b/contrib/bloop/src/mill/contrib/bloop/Bloop.scala
@@ -8,5 +8,6 @@ import mill.api.WorkspaceRoot
  */
 object Bloop extends BloopImpl(
       () => Evaluator.allBootstrapEvaluators.value.value,
-      WorkspaceRoot.workspaceRoot
+      WorkspaceRoot.workspaceRoot,
+      addMillSources = None
     )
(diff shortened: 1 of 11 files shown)


alexarchambault:
The Mill BSP server already adds those to the root modules. But using Bloop is handy sometimes. And if I'm not mistaken, I think using Bloop rather than BSP is still the default in Metals.

alexarchambault:
Note that this fixes things for me… when I use my own local fork of Metals, not with Metals mainline. With mainline Metals, on a simple project, Bloop import is slightly better (hover works, not go-to-source). With BSP import, neither hover not go-to-source work.

alexarchambault:
cc @Baccata, who wrote large parts of the Bloop import stuff, if I'm not mistaken

lefou (commented):
Sounds reasonable. If you say it helps, feel free to merge it.

Don't know why adding Mill source JARs to the root modules help though, as they most likely should be added to the `MillBuildRootModule`s.

alexarchambault:
> Don't know why adding Mill source JARs to the root modules help though, as they most likely should be added to the `MillBuildRootModule`s.

Seems those that are seen as "root" in the added logic are `MillBuildRootModule`s (they're `MillBuildRootModule.BootstrapModule`)

Issue #4611, opened 2025-02-23
Mill prompt shows many more open tasks than the parallelism limit


lihaoyi:
See the image below, where I ran `./mill -k integration.__.local.server`  with repo the config `-j0.5C` and 10 cores on my laptop, and I would expect 5 ongoing tasks

<img width="976" alt="Image" src="https://github.com/user-attachments/assets/391289a5-2872-46d2-9f3b-ab7a36e7bcec" />

It's not surprising to have more than 5, because some tasks may be blocked by their child `ctx.fork.async` futures are running. But I wouldn't expect more than 10 (5 blocked and 5 running). Maybe our scheduling is a bit wonky and letting the blocked tasks pile up rather than clearing them promptly

lihaoyi:
I think the problem here may be the first-in-first-out nature of `ThreadPoolExecutor` with `LinkedBlockingQueue`: if every test task that spawns sub-tasks is submitted to the queue first, the sub-tasks will all run last only after every test task is blocked. This would cause a huge pile of test tasks to enter the started-but-waiting-for-subtasks blocked state, which they will stay in while the sub-tasks complete over time

If that's the problem, then a solution could be to change `ThreadPoolExecutor` to use a first-in-last-out model, which would prioritize recently-spawned subtasks over earlier-queued test tasks

PR #4614, opened 2025-02-23
Use process pool for test runner

### LINKED ISSUE #4590: Implement a dynamic work-stealing test fork runner (1500USD Bounty)
---

**From the maintainer Li Haoyi**: I'm putting a 1500USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

The goal of this ticket is to implement a variant of `testForkGrouping` that performs work-stealing: rather than splitting test classes into fixed groups with each grouping running serially in a separate JVM, we want to spawn 1-or-more JVMs for each test group with the JVMs in each group performing work-stealing on the group's test classes, shutting down when that group's test classes have all been completed.

Currently, `testForkGrouping` only allows static allocation of test classes to various subprocesses. This means it cannot be easily turned on by default, since there will always be codebases with many fast test classes where forking is a negative due to the JVM startup overhead, and others with fewer slow test classes where the JVM startup overhead is less costly. A work-stealing test runner as described above would avoid this problem by letting fast test classes share the JVM, only forking as many parallel JVMs as necessary to use saturate Mill's thread count config (`--jobs`) which defaults to `NUM_CORES`. This would likely be self-tuning enough to turn it on by default, so everyone can benefit from running test classes in parallel regardless of the runtime characteristics of their test classes.

As described above, work stealing would only occur within each test group. Thus a user would still be able to use `testForkGrouping` to separate test classes that should _never_ run in the same JVM, but if there are no such restrictions they could stick with the default "everything in one group" config and Mill's test runners will work-steal the test classes to complete them as soon as possible.

See earlier discussion in https://github.com/com-lihaoyi/mill/issues/4419

HollandDM:
fixes #4590.

This PR introduce `testEnableWorkStealing` setting for `TestModule`. When enabled, tests in the module will be distributed by a work-stealing process pool, and run in parallel with each other.

Internally, when set `testEnableWorkStealing` to `true`, Mill will try to spawn at most `--job` test runner for each test groups. Each runner will compete to "steal"  a test classes from shared folder to run. This increase throughput of the whole test group, as slow tests now don't block fast tests from running. Each test runner have it own working folder and log files used to communicate with the Mill parent process.


HollandDM:
Local run on my machine yield similar test time when run `scalalib.test.testForked` with grouping on and off. But do I need some kind of benchmark for this?

lihaoyi:
Let's go with the multiple-JVMs-one-thread-each architecture described in the original ticket, rather than the one-JVM-multiple-threads architecture you have in this PR. 

Having each JVM be single-threaded running tests makes things like resource usage much more attributable and deterministic, since the resource footprint of each JVM can only be due to the one test it is currently running. Whereas if you put a bunch of tests into a single JVM and it crashes due to OOM, you have no idea which test is at fault.

In terms of performance, I think the goal should be:

1. Running heavyweight test classes like `scalalib.test.testForked`, it should have no significant performance difference with `testForkGrouping` enabled or disabled

2. Running lightweight test classes, say 1000 lightweight test classes that do nothing, it should be much more performant than having `testForkGrouping` configured to one-test-per-class, and comparable to `testForkGrouping` disabled (although maybe not quite the same, depending on whether the overhead of N different JVMs or whether the speedup from running N times parallel wins out)

HollandDM:
Hmm, look like i misunderstood the ideal of the original ticket. From what you said, it seems like each test groups can now spawn multiple sub processes (with respect to total `--jobs` config), and they can perform work stealing between them, right?
Also, for the performance goal, number 1 is easy to setup. But for number 2, do we have existing test like this, or should I create new one, and add it in this PR also?

lihaoyi:
@HollandDM that's right.

In terms of the (2) benchmarks, feel free to include it in this PR as an integration test. You can have the test code generate the necessary project files before running the test to avoid having to commit all the boilerplate code

HollandDM:
@lihaoyi I've update this into multiple-JVMs-one-thread-each architect, with the work stealing capability.
The PR is still kind of like a PoC, to make sure the approach is correct, please verify it first.
I'll try clean up and pass all test cases in following commits

lihaoyi:
@HollandDM can you write an english summary of how the implementation works? That would make it much easier to review vs. trying to reverse engineer your diff from scratch

lihaoyi:
At a first glance, the IPC protocol looks far too complicated for what this task requires. 

For example, do we really need 8 different cluster states? Since the test module has to complete running every test class eventually anyway, each subprocess could just continually take work off its (possibly shared) queue until there is no more work, and then shut down. There should be no need for subprocesses ever to be blocked, stopped, or in other states: either each subprocess is actively running tests, or it's done (when there are no more test classes to claim)

In terms of managing concurrency, the parent Mill process can claim the async slot before each subprocess starts, and release it when the subprocess ends. The subprocess shouldn't need to worry about this: the fact that the subprocess exists means that a slot is claimed for it, and the fact that the subprocess hasn't exited means it is doing work running some test class. Once the subprocess exits, the Mill parent process can release the async slot. This whole thing could be automated just by running the subprocess in a `ctx.fork.async(dest, key, msg){...}` block.

In terms of stealing work, we can get around the whole locking-stealing-file protocol by writing each test group's test class names to files on disk, that each subprocess tries to atomically `os.move` to claim the class. Such a simple disk-based queue would be more than sufficient for the scales we're looking at (_O(num-test-source-files)_) and would have a lot fewer moving parts than the mem-mapped version you have here

HollandDM:
Let's me explain the IPC protocol, as it is the heart of all of this implementation:
In this impl:
- the host mill will decide: 1. when to spawn a subprocess; 2. when to signal the subprocess to stop; 3. what subprocess it gonna steal from.
- the subprocess will signal cluster: 1. if it is running normally or slow (blocking), if it want to steal work
- the subprocess will signal other process: 1. if it want to steal; 2. if it permit other to steal.

For host, at first the host spawns 1 subprocess to handle all the test cases. As time go, if any subprocess want to steal, the host will find another subprocess to be the victim, prioritize blocking one. The host also keep track of the state each subprocess is in at the moment: if more blocking than steal, the host will try spawn new one to accommodate the need to off load work, if more stealing and blocking, it will signal denial to the stealers.

For subprocesses, they periodically notify host their running state. If they're allow to steal from a victim, a 2nd channel will be used between them to coordinate the steal (write test to steal to disk, and signal other to check this). If they are denied from stealing, they can keep retry for sometimes, but ultimately they will stop.

Yes, this is quite a complicated work stealing system. The reason that I went this way are:
- It is a work stealing system, the stealing happen without knowledge of the host, the host only coordinate the stealer - victim pair, the pain of coordinating stealing is off loaded to that pair of subprocesses.
- This stealing happen in a batch, and if one test is blocked because its running to slow, other tests are guaranteed to be picked by other subprocesses.  If we coordinated by using host as the test provider, unless we provide 1 test at a time (which I think is worse because the of the overload in I/O compare to running test), we can't guaranteed this behavior. This of course contains more moving part, and a test can be moved multiple times (but at most only O(log(num-test-source-files)))).
- The number of states is quite overwhelming, but it is needed to convey the state of each parties, and because I don't want to use any locking mechanism while doing IPC, the number of states is expected to be bigger than using one.
- In terms of managing concurrency, I did it like this to respect the --job parameter to the fullest. The host when decided to spawn new subprocess need to "ask" executor for a slot, and it can be denied. Spawning normally by `ctx.fork.async` (which is a decorated `new Future[A]`) will leave the number of subprocess for one group test to be as much as --jobs. At first I though this can lead to jobs * jobs subprocess to be spawn in the worst cases. But now after a closer look, I think this is not the case. So indeed the managing concurrency problem can be reduced to just use `ctx.fork.async`.

Your suggestion is way simpler, and should work fine. The only thing that I'm not too sure is how many test should we write on a single file for subprocess to pick up. If it is one then subprocess can scanning and contending for files on disk heavily. If it is more then quick test cases can be blocked by one slow test case, and I think work stealing is required in this situation, which will lead to something similar to this

lihaoyi:
Got it. What you say makes sense, but lets go with the approach i suggested and see if we can make that work without the complexity of a full peer-to-peer work stealing model

> The only thing that I'm not too sure is how many test should we write on a single file for subprocess to pick up. If it is one then subprocess can scanning and contending for files on disk heavily. If it is more then quick test cases can be blocked by one slow test case, and I think work stealing is required in this situation, which will lead to something similar to this

Most things work on the granularity of test classes, so that would be the natural unit of work here. In that case the workers can just pick work directly from the shared queue folder, until the folder is empty.

I don't think we need to worry too much about performance of the disk queue; we won't have more than a few thousand test classes per module, and the operations we need to do are cheap ones (just checking if the folder is, trying doing `os.move` on an arbitrary file to claim it).

HollandDM:
@lihaoyi I updated the PR to use the simplify version I came up with.
The performance is also great, it seems that we indeed don't need the full blow work stealing cluster I designed originally.
The only down side for now, is that the number of await thread will be super big if testGroup is used.
I'll try fix the CI in the following commit if we're good to go with this impl

lihaoyi:
@HollandDM thanks will take a look

testrunner/src/mill/testrunner/TestRunnerUtils.scala
```diff
@@ -208,6 +210,145 @@ import scala.jdk.CollectionConverters.IteratorHasAsScala
...
+
+      val taskQueue = mutable.Queue.empty[Task]
+
+      @tailrec
+      def stealFromSelectorFolder(): Boolean = {
+        val files = os.list(selectorFolder)
+        if (files.nonEmpty) {
+          var shouldRetry = false
+          val offset = Random.nextInt(files.size)
+          try {
```
lihaoyi, line 231: We can do `val shouldRetry = try {`

testrunner/src/mill/testrunner/TestRunnerUtils.scala
```diff
@@ -208,6 +210,145 @@ import scala.jdk.CollectionConverters.IteratorHasAsScala
...
+      stealFolder: os.Path,
+      selectorFolder: os.Path
+  )(implicit ctx: Ctx.Log): (String, Iterator[TestResult]) = {
+    val events = new ConcurrentLinkedQueue[Event]()
+    val doneMessage = {
+
+      val taskQueue = mutable.Queue.empty[Task]
+
+      @tailrec
+      def stealFromSelectorFolder(): Boolean = {
```
lihaoyi, line 226: Can we move `stealFromSelectorFolder` out of `stealTasks` into its own top-level method and make it return a `Option[String]` with the successfully stolen class name? That will help encapsulate things a bit more and keep the code easier to understand

testrunner/src/mill/testrunner/TestRunnerUtils.scala
```diff
@@ -208,6 +210,145 @@ import scala.jdk.CollectionConverters.IteratorHasAsScala
...
+              // fail to steal, try again
+              shouldRetry = true
+          }
+          if (shouldRetry) {
+            stealFromSelectorFolder()
+          } else {
+            // we successfully stole a selector
+            val selectors = os.read.lines(stealFolder / s"selector-stolen")
+            val classFilter = TestRunnerUtils.globFilter(selectors)
+            val tasks = runner.tasks(
```
lihaoyi, line 248: Is `tasks` always of length `1` here? If so we can add an assert and simplify the code accordingly

testrunner/src/mill/testrunner/TestRunnerUtils.scala
```diff
@@ -208,6 +210,145 @@ import scala.jdk.CollectionConverters.IteratorHasAsScala
...
+      testClasses: Seq[(Class[?], Fingerprint)],
+      testReporter: TestReporter,
+      runner: Runner,
+      stealFolder: os.Path,
+      selectorFolder: os.Path
+  )(implicit ctx: Ctx.Log): (String, Iterator[TestResult]) = {
+    val events = new ConcurrentLinkedQueue[Event]()
+    val doneMessage = {
+
+      val taskQueue = mutable.Queue.empty[Task]
```
lihaoyi, line 223: With the changes suggested for `stealFromSelectorFolder` and `stealTaskLoop`, we should be able to remove this mutable `taskQueue` as well

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -43,46 +43,91 @@ private[scalalib] object TestModuleUtil {
...
-        outputPath = outputPath,
-        colored = Task.log.colored,
-        testCp = testClasspath.map(_.path),
-        globSelectors = selectors2
-      )
+    def prepareTestSelectorFolder(selectors2: Seq[String], base: os.Path): os.Path = {
+      val selectorFolder = base / "selectors"
+      os.makeDir.all(selectorFolder)
+      selectors2.zipWithIndex.foreach { case (s, i) =>
+        os.write.over(selectorFolder / s"selector-$i", s)
```
lihaoyi, line 50: How about we name the files according to the class name `s` rather than the index `i`? If we need to debug things being able to `ls` to see the queued test classes would be much easier than having to dig into each file individually

testrunner/src/mill/testrunner/TestRunnerUtils.scala
```diff
@@ -208,6 +210,145 @@ import scala.jdk.CollectionConverters.IteratorHasAsScala
...
+          )
+
+          taskQueue.enqueueAll(next)
+          stealTaskLoop()
+        } else if (stealFromSelectorFolder()) {
+          stealTaskLoop()
+        } else {
+          ()
+        }
+      }
```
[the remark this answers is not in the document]
HollandDM: This is more about the coding style used in these repositories. In my other open-source PRs, people tend to prefer tailrec over while loops. It looks like it's the other way around here. I'm happy to change all the tailrec to while loops if you'd like. I don't really have a preference.

testrunner/src/mill/testrunner/TestRunnerUtils.scala
HollandDM: I'm not really sure if the tasks returned always has the same length as the taskdefs input.

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -43,46 +43,91 @@ private[scalalib] object TestModuleUtil {
...
+
+    def runTestRunnerSubprocesses(selectors2: Seq[String], base: os.Path) = {
+      val selectorFolder = prepareTestSelectorFolder(selectors2, base)
+
+      val resultFutures = Range(0, selectors2.length).map { index =>
+        val indexStr = index.toString
+        Task.fork.async(base / indexStr, indexStr, s"Test process $indexStr") {
+          (indexStr, runTestRunnerSubprocess(index, base / indexStr, selectorFolder))
+        }
+      }
```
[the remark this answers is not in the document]
HollandDM: The implementation already have the check before spawn, so it's true that the number of meaningful threads that spawn subprocess is small.
What I meant was that whew using `testGroup`, the server call `runTestRunnerSubprocesses` foreach test groups. These `Future` then only `await` for `runTestRunnerSubprocess` (the real test running work) to return. So we got a lot of awaiting Future hanging arounds, and these will all show on the cli when run (I'll provide screenshot example later)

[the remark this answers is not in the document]
HollandDM: It's look exactly like the thing you saw #4611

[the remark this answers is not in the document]
lihaoyi: Got it. Since `runTestRunnerSubprocesses` already spawns `fork.async` blocks, what if we move the caller out into normal synchronous code? So instead of

```scala
          Task.fork.async(Task.dest / folderName, paddedIndex, groupPromptMessage) {
            (folderName, runTestRunnerSubprocesses(testClassList, Task.dest / folderName))
          }
```

Just

```scala
          (folderName, runTestRunnerSubprocesses(testClassList, Task.dest / folderName))
```

That way there won't be a middle-layer of async blocks that do nothing except wait for their children, which should reduce the number of blocked tasks showing up in the command line prompt

testrunner/src/mill/testrunner/TestRunnerUtils.scala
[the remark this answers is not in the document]
HollandDM: Now I think about this, this logic can only be changed to do while without changing it's sematic, which is not supported by scala

HollandDM:
@lihaoyi Updated, I replaced the @tailrec with while loop. Also flatten out all subprocess calls to the outer most level and await all of them

HollandDM:
CI failed massively, so I did some investigations.
It's seem that example test can be troublesome, because now we always have multiple test runner for a group, the order of output log can be non deterministic, this will make all example tests fail because they rely on deterministic output log.
One way to keep the same behavior is that for each subprocess in a group, we sort the log by the order of `discoveredTestClasses`. This can be done with 2 strategies:
- Defer all the log, after everything end, write the logs by the correct order. This delay the output at the end, and I don't think we want that
- Write the log in order as soon as possible, eg: if we received log in the order of: 1 4 2 6 3 5, then we will out put 4 times, [1], [2], [3, 4], [5, 6]. This will be more responsive, but require some kind of concurrent algo.
Another way to to change ExampleTester to check `expectedLine` by existence only, but I'm afraid this can affect too many area outside of scala test (kotlin, python, javascript, etc...)

lihaoyi:
@HollandDM could you elaborate on the CI failures? AFAIK the example tests already check output in a non-ordered fashion: as long as each line in the example expected output appears in the actual output, the test passes, regardless of what order the actual output is in. So if your tests are failing, it seems like there must be some other issue there other than the difference in output ordering

HollandDM:
Oh true, now I check the ode carefully, it does check existence. I guess looking at all the `--- Expected output ----------` make me dizzy. Sorry for the fail alert.
Let me try fixing the CI

HollandDM:
@lihaoyi Ok, so a lot of example tests are failing. This is because we changed the test process to always spawn new sub-processes with `Fork.async`, it also changed the output of the test (added [...] at the front, like when we use -j > 1).
We also changed the test folder structure (now contains multiple sub folders).
After some though, I think the only logically thing to do is to go through each example and make adjustments, as other changes may creates some implicit conversion that will confuse others.
It's quite time consuming, and I will have to abuse CI checks to identify the errors (my machine cannot run all of them).

lihaoyi:
@HollandDM regarding the leading `[...]`, I think that may have gone away with some of my recent changes to the `--ticker false` flag (see the PR https://github.com/com-lihaoyi/mill/pull/4661/files#diff-8309a341d7e1369fb83edf1b37ba298fb13d61d612e4794bfe5d8a8c67c59788R32-R35). If that's the case, that particular change in behavior in the test suite may disappear if you merge in `main`, since all integration and example tests are run with `--ticker false`

We should also probably gate the new functionality on a config flag like `def workStealingTestRunners: Boolean = true` (or something like that? Not sure about the naming).

That will let us try out the new functionality for a while opt-in without changing everyone's behavior immediately. And then when we're ready to make it the default, we could flip the default from `false` to `true` and roll it out by default with an opt-out

lihaoyi:
For CI, they should already be running on your fork of the Mill repo https://github.com/HollandDM/mill/commits/work-stealing-test-runner. That uses up your own Github Actions quota of ~20 concurrent workers, so you don't need to worry about abusing it since it won't inconvenience anybody

HollandDM:
@lihaoyi Updated, for now I'll hide this feature behind the `useTestStealingScheduler` task. In case you want to switch it to flag, I think the adjustment will be straight forward.
I also updated some test cases and added examples, which caused the size of the PR to explode, but that should be fine since it's test related.

example/scalalib/testing/5-test-stealing/foo/test/src/RandomTestsB.scala
```diff
@@ -0,0 +1,42 @@
+package foo
+import utest._
+object RandomTestsB extends TestSuite {
+  def tests = Tests {
+    test("test1") {
+      val result = Foo.greet("Dakota")
+      assert(result == "Hello Dakota")
+      Thread.sleep(22)
+      result
```
lihaoyi, line 9: Could we DRY this up with a helper method? That would let each test case take 1 line instead of 7, and shrink the size of this PR considerably

lihaoyi: How are these `sleep`s chosen? Are they arbitrary? Or is there some thought behind it? Either way we should leave a comment somewhere explaining where they came from (e.g. in the helper method)

example/scalalib/testing/6-test-group-stealing/build.mill
```diff
@@ -0,0 +1,58 @@
...
+/** See Also: foo/test/src/GroupY1.scala */
+/** See Also: foo/test/src/GroupY2.scala */
+/** See Also: foo/test/src/GroupY3.scala */
+/** See Also: foo/test/src/GroupY4.scala */
+/** See Also: foo/test/src/GroupY5.scala */
+/** See Also: foo/test/src/GroupY6.scala */
+/** See Also: foo/test/src/GroupY7.scala */
+/** See Also: foo/test/src/GroupY8.scala */
+/** See Also: foo/test/src/GroupY9.scala */
+/** See Also: foo/test/src/GroupY10.scala */
```
lihaoyi, line 37: `See Also` inlines the referenced file in the docs. I think we can just elide all of this; there's a `browse` link anyway anyone who wants can still get to the detailed sources

scalalib/src/mill/scalalib/TestModule.scala
```diff
@@ -97,15 +97,24 @@ trait TestModule
...
   def testForkGrouping: T[Seq[Seq[String]]] = Task {
     Seq(discoveredTestClasses())
   }
 
+  /**
+   * Whether to use the test stealing scheduler to run tests in multiple JVM processes.
+   * When used in combination with [[testForkGrouping]], every JVM test running process
+   * will guarantee to never steal tests from different test groups.
+   */
+  def useTestStealingScheduler: T[Boolean] = T(false)
```
lihaoyi, line 116: Let's call this `testEnableWorkStealing`? I'd like it to start with the `test*` prefix following the convention of the rest of the module's tasks

testrunner/src/mill/testrunner/Model.scala
```diff
@@ -11,7 +11,7 @@ import mill.api.internal
     outputPath: os.Path,
     colored: Boolean,
     testCp: Seq[os.Path],
-    globSelectors: Seq[String]
+    globSelectors: Either[Seq[String], (os.Path, os.Path)]
```
lihaoyi, line 14: Leave a comment here explaining what the two cases mean, as it's no longer obvious what `(os.Path, os.Path)` has to do with glob selectors

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -24,6 +25,7 @@ private[scalalib] object TestModuleUtil {
       testClasspath: Seq[PathRef],
       args: Seq[String],
       testClassLists: Seq[Seq[String]],
+      useTestStealingScheduler: Boolean,
```
lihaoyi, line 28: Let's add this new parameter to the end of the parameter list

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -136,6 +96,332 @@ private[scalalib] object TestModuleUtil {
...
+      forkEnv: Map[String, String],
+      testSandboxWorkingDir: Boolean,
+      forkWorkingDir: os.Path,
+      javaHome: Option[os.Path],
+      props: Map[String, String],
+      jvmArgs: Seq[String],
+      testRunnerClasspathArg: String,
+      resourceEnv: Map[String, String],
+      baseFolder: os.Path,
+      selector: Either[Seq[String], os.Path]
```
lihaoyi, line 182: I think these parameter lists are getting copy-pasty enough that it's time to create a `class TestModuleUtil` that takes all these in the constructor and then has methods than use the various configuration values without needing to tediously pass around the whole list every defsite and callsite.

testrunner/src/mill/testrunner/Model.scala
HollandDM: I actually want to rename this because of what you said, but I'm bad at naming.

example/scalalib/testing/5-test-stealing/foo/test/src/RandomTestsB.scala
lihaoyi: LLM is fine, feel free to use whatever you like, just need to clean it up and justify it after the LLM is done

testrunner/src/mill/testrunner/TestRunnerUtils.scala
```diff
@@ -213,6 +228,111 @@ import scala.jdk.CollectionConverters.IteratorHasAsScala
...
+          val tasks = runner.tasks(
+            for ((cls, fingerprint) <- testClasses.iterator.toArray if classFilter(cls.getName))
+              yield new TaskDef(
+                cls.getName.stripSuffix("$"),
+                fingerprint,
+                false,
+                Array(new SuiteSelector)
+              )
+          )
+          os.remove(stealFolder / s"selector-stolen")
```
lihaoyi, line 264: Let's leave the files on disk once stolen, no need to delete them. That would help debugging since people can `ls` each worker's baseFolder and see what test classes that worker claimed

lihaoyi: We can reuse the original file name to avoid conflicts, rather than renaming them to `selector-stolen` as you have done here.

testrunner/src/mill/testrunner/TestRunnerUtils.scala
```diff
@@ -213,6 +228,111 @@ import scala.jdk.CollectionConverters.IteratorHasAsScala
...
+            )
+            true
+          } catch {
+            case e: Exception => false
+          }
+        if (stole) {
+          val selectors = os.read.lines(stealFolder / s"selector-stolen")
+          val classFilter = TestRunnerUtils.globFilter(selectors)
+          val tasks = runner.tasks(
+            for ((cls, fingerprint) <- testClasses.iterator.toArray if classFilter(cls.getName))
```
lihaoyi, line 256: By the time we get here the files should each be a single test class name, so we shouldn't need to run `globFilter` and iterate over all the `testClasses` to find the right one; we should be able to turn `testClasses` into a `Map` keyed on the class name and do a direct lookup

testrunner/src/mill/testrunner/TestRunnerUtils.scala
```diff
@@ -213,6 +228,111 @@ import scala.jdk.CollectionConverters.IteratorHasAsScala
...
+    if (doneMessage != null && doneMessage.nonEmpty) {
+      if (doneMessage.endsWith("\n"))
+        println(doneMessage.stripSuffix("\n"))
+      else
+        println(doneMessage)
+    }
+
+    val results = parseRunTaskResults(events.iterator().asScala)
+
+    (doneMessage, results)
```
lihaoyi, line 311: There's a bunch of duplication here between `runTasks` and `stealTasks`; let's consolidate it into a helper method or something to keep things DRY

lihaoyi:
Trying this out manually, the UI isn't great because we don't have any information about what each worker is running

Screenshot 2025-03-07 at 12 26 44 PM

To fix that, you can update the signature of `ctx.fork.async` from `(t: => T)` to `(t: Logger => T)` to expose the sub-logger than each async block creates, and use `log.ticker` on the `Logger` to set the dynamic progress text that shows up to the right of the task message/run-duration. The worker process can dump the name of the currently running test class to disk before running each test class, and have the parent Mill process spawn a thread to poll it for updates, since we need  to somehow get that information back across the process boundary from the subprocess to the parent.

testrunner/src/mill/testrunner/TestRunnerUtils.scala
HollandDM: What the subprocesses are stealing is the selector of test classes, not the test class names. I know TestModuleUtil run the glob filter once before passing them into the subprocesses, so they basically the same. But I think subprocesses themself need to handle the case if someone called them without the preprocessing step (I think the current test runner also make the assumption, right?).
How about the hybrid way? We first create a map of single test classes with the name, then if we encounter unknown selector, we run through it and cache it for further use.

lihaoyi: The test runner subprocess is only used internally by ourselves, so we just need to make sure it works for what we pass it. Since we're always running the glob filter to generate `filteredClassLists` before passing them to the subprocess, we can assume that going forward they will be concrete class names rather than selectors. We shouldn't try too hard to support use cases we don't have since we can just refactor the code later if need to anyway

HollandDM:
@lihaoyi I believe I adjust all your feedback until now.
For the UI, I managed to tick current running tests every 1 second.
The color itself seem quite dark, other than that, it works well.

core/api/src/mill/api/Ctx.scala
```diff
@@ -141,9 +141,13 @@ object Ctx {
        *                terminal prompt to display what this future is currently computing.
        * @param t The body of the async future
        */
-      def async[T](dest: os.Path, key: String, message: String)(t: => T)(implicit
+      def async[T](dest: os.Path, key: String, message: String)(t: Logger => T)(implicit
           ctx: mill.api.Ctx
       ): Future[T]
+
+      @inline def async[T](dest: os.Path, key: String, message: String)(t: => T)(implicit
```
HollandDM, line 148: keep this, as I don't want to disrupt the current API too much

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -173,15 +205,190 @@ private[scalalib] object TestModuleUtil {
...
+          stealLogMap.forEach { case (_, (stealLog, stealLogCallback)) =>
+            try {
+              os.read.lines(stealLog).lastOption.foreach(stealLogCallback)
+            } finally ()
+          }
+        }
+      },
+      0,
+      1,
+      java.util.concurrent.TimeUnit.SECONDS
```
HollandDM, line 315: I think looping every 1 second is enough for logging purpose here, and it only causes a little impact to the whole server

HollandDM:
@lihaoyi It should have. To be sure, have you set testEnableWorkStealing to true yet.
For some reason, when testEnableWorkStealing is set to false, the prompts are gone. I’ll have look at this

lihaoyi:
Ah that is likely my mistake, let me try adding it

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -173,15 +209,217 @@ private[scalalib] object TestModuleUtil {
...
         }
+      },
+      0,
+      1,
+      java.util.concurrent.TimeUnit.SECONDS
+    )
+
+    val outputs = Task.fork.awaitAll(subprocessFutures)
+
+    logScheduler.shutdown()
```
lihaoyi, line 349: This probably should be in a `finally` block in case `awaitAll` fails

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -173,15 +209,217 @@ private[scalalib] object TestModuleUtil {
...
     }
+
+    subprocessResult
+  }
+
+}
+
+private[scalalib] object TestModuleUtil {
+
+  def runTests(
```
lihaoyi, line 384: No need this forwarder, let's just have the callsites use `new TestModuleUtil` directly

example/scalalib/testing/6-test-group-stealing/build.mill
```diff
@@ -0,0 +1,39 @@
...
+  def scalaVersion = "2.13.8"
+  object test extends ScalaTests {
+    def ivyDeps = Seq(ivy"com.lihaoyi::utest:0.8.5")
+    def testFramework = "utest.runner.Framework"
+
+    def testForkGrouping = discoveredTestClasses().grouped(10).toSeq
+    def testEnableWorkStealing = true
+  }
+}
+
```
lihaoyi, line 17: This needs a more verbose explanation about how `testForkGrouping` and `testEnableWorkStealing` interact

example/scalalib/testing/5-test-stealing/build.mill
```diff
@@ -0,0 +1,53 @@
+// Test stealing is an opt-in feature that allows you to take a single
```
lihaoyi, line 1: This example docs should be included in `scalalib/testing.adoc`, and referenced from `javalib/testing.adoc` and `kotlinlib/testing.adoc`

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -173,15 +209,217 @@ private[scalalib] object TestModuleUtil {
...
+  private def runTestStealingScheduler(
+      filteredClassLists: Seq[Seq[String]]
+  )(implicit ctx: mill.api.Ctx) = {
+
+    type StealLogCallback = String => Unit
+    val stealLogMap =
+      new java.util.concurrent.ConcurrentHashMap[(Int, Int), (os.Path, StealLogCallback)]()
+
+    def prepareTestSelectorFolder(selectors2: Seq[String], base: os.Path): os.Path = {
+      val selectorFolder = base / "selectors"
```
lihaoyi, line 224: Maybe call this `test-classes` since we know they all refer to concrete class names

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -173,15 +209,217 @@ private[scalalib] object TestModuleUtil {
...
-        catch {
-          case e: Throwable => Result.Failure("Test reporting failed: " + e)
+    subprocessResult
+  }
+
+  private def runTestStealingScheduler(
+      filteredClassLists: Seq[Seq[String]]
+  )(implicit ctx: mill.api.Ctx) = {
+
+    type StealLogCallback = String => Unit
```
lihaoyi, line 219: Lets just inline this type, the alias doesn't really provide much value here

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -173,15 +209,217 @@ private[scalalib] object TestModuleUtil {
...
+    val stealLogMap =
+      new java.util.concurrent.ConcurrentHashMap[(Int, Int), (os.Path, StealLogCallback)]()
```
lihaoyi, line 221: Let's store the `Logger` instead of the `String => Unit`, that might make things a bit more concrete and easy to follow

lihaoyi: Also, I think it can be simplified into `ConcurrentHashMap[os.Path, StealLogCallback]`? the `os.Path`s are unique so we don't need to separately key it based on `groupIndex` and `processIndex`

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -173,15 +209,217 @@ private[scalalib] object TestModuleUtil {
...
-          case e: Throwable => Result.Failure("Test reporting failed: " + e)
+    subprocessResult
+  }
+
+  private def runTestStealingScheduler(
+      filteredClassLists: Seq[Seq[String]]
+  )(implicit ctx: mill.api.Ctx) = {
+
+    type StealLogCallback = String => Unit
+    val stealLogMap =
```
lihaoyi, line 220: Let's call this `workerStatusMap`, since it's really more a single current status rather than an append-only log file

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -173,15 +209,217 @@ private[scalalib] object TestModuleUtil {
...
+      if (processIndex == 0) || (processIndex < numTests)
+    } yield {
+      val paddedGroupIndex = mill.internal.Util.leftPad(groupIndex.toString, maxGroupLength, '0')
+      val paddedProcessIndex =
+        mill.internal.Util.leftPad(processIndex.toString, maxProcessLength, '0')
+      val processFolder = baseFolder / processIndex.toString
+
+      Task.fork.async(
+        processFolder,
+        s"${paddedGroupIndex}-${paddedProcessIndex}",
```
lihaoyi, line 314: If there is only one group, which is likely a common case, lets shorten this to only be the `paddedProcessIndex`

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -173,15 +209,217 @@ private[scalalib] object TestModuleUtil {
...
+
+    val maxProcessLength = jobs.toString.length
+    val groupLength = groupFolderData.length
+    val maxGroupLength = groupLength.toString.length
+
+    val subprocessFutures = for {
+      processIndex <- 0 until jobs
+      groupIndex <- 0 until groupLength
+      (baseFolder, selectorFolder, numTests, groupPromptMessage) = groupFolderData(groupIndex)
+      if (processIndex == 0) || (processIndex < numTests)
```
lihaoyi, line 305: Rather than doing this filter, it would be simpler to instead loop over `processIndex <- 0 until math.min(math.max(jobs, 1) numTests)` to just pick the correct number of processes up front

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -173,15 +209,217 @@ private[scalalib] object TestModuleUtil {
...
+    }
+
+    val maxProcessLength = jobs.toString.length
+    val groupLength = groupFolderData.length
+    val maxGroupLength = groupLength.toString.length
+
+    val subprocessFutures = for {
+      processIndex <- 0 until jobs
+      groupIndex <- 0 until groupLength
+      (baseFolder, selectorFolder, numTests, groupPromptMessage) = groupFolderData(groupIndex)
```
lihaoyi, line 304: This can be moved into the body of the `yield` block as a normal local variable

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -173,15 +209,217 @@ private[scalalib] object TestModuleUtil {
...
+          logger
+        )
+      }
+    }
+
+    val logScheduler = new java.util.concurrent.ScheduledThreadPoolExecutor(1)
+
+    logScheduler.scheduleWithFixedDelay(
+      new Runnable {
+        override def run(): Unit = {
```
lihaoyi, line 334: I think this can be shortened using the lambda syntax `() => {...}` without explicitly instantiating an anonymous subclass of `Runnable`

testrunner/src/mill/testrunner/TestRunnerUtils.scala
```diff
@@ -213,6 +234,108 @@ import scala.jdk.CollectionConverters.IteratorHasAsScala
     (doneMessage, results.toSeq)
   }
 
+  private def stealTaskFromSelectorFolder(
+      globSelectorCache: Map[String, ClassWithFingerprint],
+      runner: Runner,
+      stealFolder: os.Path,
+      selectorFolder: os.Path
+  ): Array[Task] = {
+    val stealLog = stealFolder / os.up / "steal.log"
```
lihaoyi, line 243: This can now be written `stealFolder / "../steal.log"`, or rather `stealFolder / "../status.txt"` since it's not really a log file but a snapshot of the current status

testrunner/src/mill/testrunner/TestRunnerUtils.scala
```diff
@@ -213,6 +234,108 @@ import scala.jdk.CollectionConverters.IteratorHasAsScala
...
+    while (true) {
+      val files = os.list(selectorFolder)
```
lihaoyi, line 245: I think we can simplify this further given the current design: we just need to run `os.list` once and then loop over all the tests to try and steal each one in order. Since the files in `selectorFolder` never grows, running `os.list` a second time will never pick up anything the first list didn't.

Let's also get rid of the `Random.nextInt` selection and just pick up the tests in order from top to bottom. That will give a bit more determinism to how things run, and although it still nondeterministic due to parallelism that would help people follow the order that tests are being run in

lihaoyi:
Right now the `ctx.fork.async` tasks all have no name. We should fix that for the default case when work stealing is disabled,  and when work stealing is enabled let's give them simple names `worker-$processIndex` if there's one group and `worker-$groupIndex-$processIndex` if there's multiple groups. We don't need to list the test classes, since we don't know up front which test classes each worker will take, and the in-progress test will show up in the ticker on the right anyway

lihaoyi:
I think this looks pretty good. Adding `def testEnableWorkStealing = true` in my manual testing does indeed make it work, though the missing labels when `testEnableWorkStealing` is not enabled needs to be fixed. Left some last comments, but once those are resolved I think the code is ready to merge.

Next steps are probably:

1. Update the PR description to accurately reflect the design of the PR after all the revisions and updates
2. Open a second PR backporting it to the `0.12.x` branch. Looking at the code I think all the APIs changed are internal or private so there shouldn't be binary compat issues.

Once both PRs are ready we can merge it and close out the bounty

testrunner/src/mill/testrunner/TestRunnerUtils.scala
HollandDM: The file was appended while runner's working. So it can be use to check the order of test that was executed by the running.

HollandDM: I also like "/ os.up" more, I think it more verbose and eplicit

lihaoyi: Got it, it makes sense then. We can leave it as is

testrunner/src/mill/testrunner/TestRunnerUtils.scala
lihaoyi: The `os.move` already effectively checks whether the file is still left before taking it, so we don't need to run an entire `os.list` each time we steal a single file. We could add a `os.exists` check before `os.move` if we want to avoid throwing and catching an error in the common case

testrunner/src/mill/testrunner/TestRunnerUtils.scala
lihaoyi: Maybe have a comment somewhere explaining how this works?

HollandDM:
> Right now the `ctx.fork.async` tasks all have no name. We should fix that for the default case when work stealing is disabled, and when work stealing is enabled let's give them simple names `worker-$processIndex` if there's one group and `worker-$groupIndex-$processIndex` if there's multiple groups. We don't need to list the test classes, since we don't know up front which test classes each worker will take, and the in-progress test will show up in the ticker on the right anyway

I think there is a bug in `MultiLogger` that cause this.
when creating new`MultiLogger`, it does not have its own `message`/`keySuffix`, so when we call `setPromptLine`, it uses the default value of "". I tried a simple fix like this:
```scala
  private[mill] override def message = logger1.message ++ logger2.message
  private[mill] override def keySuffix = logger1.keySuffix ++ logger2.keySuffix
```
And the log is good again.

Would you like me to include these fixes in this PR? If not, I think I can skip the problem because it'll be fixed in other PR

lihaoyi:
@HollandDM your proposed fix looks reasonable. Lets include it in this PR

example/scalalib/testing/5-test-stealing/build.mill
HollandDM: Not sure how I can do reference correctly, should I create a mill only file, or will I have to port all test example into java & kotlin, that will make this Pr quite big.

HollandDM:
@lihaoyi updated, please check if you are happy with the current state of the PR

example/javalib/testing/5-test-stealing/foo/test/src/foo/RandomTestsA.java
```diff
@@ -0,0 +1,20 @@
...
+
+  @Test
+  public void test2() throws Exception {
+    testGreeting("Bella", 25);
+  }
+
+  @Test
+  public void test3() throws Exception {
+    testGreeting("Cameron", 32);
+  }
```
lihaoyi, line 19: Do we really need more than one test per file? If not we could consolidate each file to a single test case to greatly reduce the verbosity of these test examples

example/scalalib/testing/5-test-stealing/build.mill
```diff
@@ -0,0 +1,38 @@
+// Test stealing is an opt-in, powerful feature that enables parallel test execution while maintaining complete reliability and debuggability.
```
lihaoyi, line 1: `while maintaining complete reliability and debuggability.` is meaningless. You should write why people would turn this on: because it provides both the speedups from parallelism and the efficiency of re-using JVMs between tests, allowing it to be used on all sorts of test suites without downside. And we will likely make it the default in future

HollandDM:
update test example to only have 1 test case, and fix some wording in adoc

HollandDM:
@lihaoyi backport for 0.12.x readied at #4697

lihaoyi:
I think this looks good, thanks @HollandDM! Let's get https://github.com/com-lihaoyi/mill/pull/4679 green and merged as well, and send me your international bank transfer details and I will send you the bounty

PR #4616, opened 2025-02-24
Added module docs for selective execution

lefou:
The hope is, to have some meaningfull `mill inspect selective` output, but it seems, inspect does not output the Scaladoc for modules yet.

Fixed some typos along the way.

### DIFF
--- a/example/large/selective/9-selective-execution/build.mill
+++ b/example/large/selective/9-selective-execution/build.mill
@@ -10,11 +10,11 @@
 //
 // * `mill selective.run <selector>`: run on the codebase after the code change,
 // runs tasks in the given `<selector>` which are affected by the code changes
-// that have happen since `selective.prepare` was run
+// that have happened since `selective.prepare` was run
 //
 // * `mill selective.resolve <selector>`: a dry-run version of `selective.run`, prints
 // out the tasks in `<selector>` that are affected by the code changes and would have
-// run, without actually tunning them.
+// run, without actually running them.
 //
 // For example, if you want to run all tests related to the code changes in a pull
 // request branch, you can do that as follows:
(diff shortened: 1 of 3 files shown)


lihaoyi:
The scaladoc inspect thing should be fixed in https://github.com/com-lihaoyi/mill/pull/4576

lihaoyi:
At least for some modules; I didn't test it with the `lazy val` approach to defining modules so may need more tweaks

lefou:
> At least for some modules; I didn't test it with the `lazy val` approach to defining modules so may need more tweaks

Looking into it. 

It works nicely with `object`, but not with `lazy val`. 

```
[62/66] ============================== inspect selective ============================== 60s
selective(MainModule.scala:327)

Inherited Modules:
    mill.main.SelectiveExecutionModule
```

```diff
diff --git a/main/src/mill/main/MainModule.scala b/main/src/mill/main/MainModule.scala
index 8e6974e331e..26149d1d76d 100644
--- a/main/src/mill/main/MainModule.scala
+++ b/main/src/mill/main/MainModule.scala
@@ -324,7 +324,8 @@ trait MainModule extends BaseModule {
    *   prints out the tasks in `<selector>` that are affected by the code changes
    *   and would have run, without actually running them.
    */
-  lazy val selective: SelectiveExecutionModule = new SelectiveExecutionModule {}
+  object selective extends SelectiveExecutionModule
 
 }
```

```
============================== inspect selective ==============================
[1/1] inspect
selective(MainModule.scala:328)
    Commands related to selective execution, where Mill runs tasks selectively
    depending on what task inputs or implementations changed.
    
    Read more about it at: https://mill-build.org/mill/large/selective-execution.html
    
    Here are the essential commands:
    
    - `mill selective.prepare <selector>`:
    run on the codebase before the code change,
    stores a snapshot of task inputs and implementations
    
    - `mill selective.run <selector>`:
    run on the codebase after the code change,
    runs tasks in the given `<selector>` which are affected by the code changes
    that have happened since `selective.prepare` was run
    
    - `mill selective.resolve <selector>`:
    a dry-run version of `selective.run`,
    prints out the tasks in `<selector>`` that are affected by the code changes
    and would have run, without actually running them.

Inherited Modules:
    mill.main.SelectiveExecutionModule
```

PR #4621, opened 2025-02-25
Add a `sourceGenerators` task to help out IDE config generation

Baccata:
Context : https://github.com/com-lihaoyi/mill/discussions/4530

* Adds a `DeferredGeneratedSourcesModule` that extends `JavaModule`, providing a `def deferredGeneratedSourceTasks` alternative to `generatedSources`. This allows BSP/GenIdea/Bloop to predict source roots without actually running the code generation logic. 
* Adds a `def out` to the `Ctx` class, allowing to query the location `out` folder within tasks. 
* Adds unit tests / integration tests where relevant 

#### Previous description (kept for the record) : 

@lihaoyi, I want to open this PR to continue the discussion. At the time of writing this, this adds a `generatedSourceRoots` parameter for tasks, as you've described, and uses it to compute the source directories for the bloop configuration, as opposed to actually running the `allSources` task. A similar change would need to be achieved in `GenIdea` and `BSP`. 

So now, the problem I see is as follows : in order for this to be useful, it needs to be applied to all source generators provided OOTB by the mill codebase, but also third party providers of source-generator (like smithy4s), as the ability to jump to the sources is paramount for the user experience in IDEs. 

Applying this change is a fair bit of work in this codebase, and a big ask for the larger ecosystem, and although I think SBT's approach of treating all source generators lazily is preferable to mill's current approach of running all source generators eagerly, I don't think I can weigh the benefits of changing the approach against the cost induced by breaking IDE behaviour for existing plugins. 

With that in mind here's what I think would be a good mitigation : 

```scala
// Rather than traversing the transitive inputs of `allSources`, we define this setting as a way to register lazy 
// source generators. We force the contract to return `Unit` to prevent the assumption that a returned `PathRef` would be
// added to the sources. 
def lazySourceGenerators = Seq[T[Unit]] 
 
 // Runs the lazy source generators, returning path refs computed from their `generatedSourceRoots`. 
 // When the tasks don't have `generatedSourceRoots` set, we assume that the `T.dest` is to be added as source root. 
 final def runLazySourceGenerators : T[Seq[PathRef]] = ???
 
// Does not run the lazy source generators, but rather computes the source roots they'll generate code in. 
// Wrapped in T because the evaluator is needed to get the T.dest of the lazy source generators. I presume that is 
// something we could thread in through the ctx (like `def destOf(task: T[_]) : Path`)  
final def lazyGeneratedSourceRoots: T[Seq[Path]] = ???
 
 // Keeps the current behaviour intact 
def allSources = T {
    sources() ++ generatedSources() ++ runLazySourceGenerators()
} 
```

The `Bloop`, `GenIdea` and `BSP` then get amended to call `sources()`, `generatedSources()` and `lazyGeneratedSourceRoots()`, but avoids `runLazySourceGenerators`. 

This approach would allow to alleviate the pain point raised in the discussion above, whilst retaining the current behaviour of the ecosystem. Mill plugin providers could be encouraged to use `runLazySourceGenerators` instead of `generatedSources`, but users would have a recourse to turn eager source generators into lazy ones if they need it to be (provided a `withGeneratedSourceRoots` is provided to them, that is)

### DIFF
--- a/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
+++ b/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
@@ -232,6 +232,7 @@ private class MillBuildServer(
       targetIds = _ => sourcesParams.getTargets.asScala.toSeq,
       tasks = {
         case module: MillBuildRootModule =>
+          // TODO
           Task.Anon {
             module.sources().map(p => sourceItem(p.path, false)) ++
               module.generatedSources().map(p => sourceItem(p.path, true))
--- a/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
+++ b/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
@@ -232,15 +232,16 @@ private class MillBuildServer(
       targetIds = _ => sourcesParams.getTargets.asScala.toSeq,
       tasks = {
         case module: MillBuildRootModule =>
-          // TODO
           Task.Anon {
             module.sources().map(p => sourceItem(p.path, false)) ++
-              module.generatedSources().map(p => sourceItem(p.path, true))
+              module.generatedSources().map(p => sourceItem(p.path, true)) ++
+              module.predictedGeneratedSourceRoots().map(p => sourceItem(p.path, true))
           }
         case module: JavaModule =>
           Task.Anon {
             module.sources().map(p => sourceItem(p.path, false)) ++
-              module.generatedSources().map(p => sourceItem(p.path, true))
+              module.generatedSources().map(p => sourceItem(p.path, true)) ++
+              module.predictedGeneratedSourceRoots().map(p => sourceItem(p.path, true))
           }
       }
     ) {
--- a/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
+++ b/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
@@ -235,13 +235,13 @@ private class MillBuildServer(
           Task.Anon {
             module.sources().map(p => sourceItem(p.path, false)) ++
               module.generatedSources().map(p => sourceItem(p.path, true)) ++
-              module.predictedGeneratedSourceRoots().map(p => sourceItem(p.path, true))
+              module.predictedGeneratedSources().map(p => sourceItem(p.path, true))
           }
         case module: JavaModule =>
           Task.Anon {
             module.sources().map(p => sourceItem(p.path, false)) ++
               module.generatedSources().map(p => sourceItem(p.path, true)) ++
-              module.predictedGeneratedSourceRoots().map(p => sourceItem(p.path, true))
+              module.predictedGeneratedSources().map(p => sourceItem(p.path, true))
           }
       }
     ) {
--- a/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
+++ b/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
@@ -235,7 +235,7 @@ private class MillBuildServer(
           Task.Anon {
             module.sources().map(p => sourceItem(p.path, false)) ++
               module.generatedSources().map(p => sourceItem(p.path, true)) ++
-              module.predictedGeneratedSources().map(p => sourceItem(p.path, true))
+              module.predictedGeneratedSources().map(p => sourceItem(p, true))
           }
         case module: JavaModule =>
           Task.Anon {
--- a/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
+++ b/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
@@ -14,7 +14,7 @@ import mill.eval.Evaluator.TaskResult
 import mill.main.MainModule
 import mill.runner.MillBuildRootModule
 import mill.scalalib.bsp.{BspModule, JvmBuildTarget, ScalaBuildTarget}
-import mill.scalalib.{JavaModule, SemanticDbJavaModule, TestModule}
+import mill.scalalib.{JavaModule, DeferredGeneratedSourcesModule, SemanticDbJavaModule, TestModule}
 import mill.util.ColorLogger
 import mill.given
 
@@ -231,7 +231,7 @@ private class MillBuildServer(
       hint = s"buildTargetSources ${sourcesParams}",
       targetIds = _ => sourcesParams.getTargets.asScala.toSeq,
       tasks = {
-        case module: MillBuildRootModule =>
+        case module: DeferredGeneratedSourcesModule =>
           Task.Anon {
             module.sources().map(p => sourceItem(p.path, false)) ++
               module.generatedSources().map(p => sourceItem(p.path, true)) ++
@@ -240,8 +240,7 @@ private class MillBuildServer(
         case module: JavaModule =>
           Task.Anon {
             module.sources().map(p => sourceItem(p.path, false)) ++
-              module.generatedSources().map(p => sourceItem(p.path, true)) ++
-              module.predictedGeneratedSources().map(p => sourceItem(p.path, true))
+              module.generatedSources().map(p => sourceItem(p.path, true))
           }
       }
     ) {
--- a/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
@@ -116,7 +116,7 @@ class BloopImpl(evs: () => Seq[Evaluator], wd: os.Path) extends ExternalModule {
     JavaModuleUtils.transitiveModules(mod, accept)
   }
 
-  private lazy val modulesToOutFolderMap: Map[JavaModule, os.Path] = {
+  private def modulesToOutFolderMap: Map[JavaModule, os.Path] = {
     val evals = evs()
     evals.flatMap { eval =>
       if (eval != null)
@@ -137,10 +137,6 @@ class BloopImpl(evs: () => Seq[Evaluator], wd: os.Path) extends ExternalModule {
       !module.asInstanceOf[outer.Module].skipBloop
     else true
 
-  private def deferredAllSources(m: JavaModule): T[Seq[PathRef]] = Task {
-    m.sources() ++ m.generatedSources() ++ m.allDeferredSourceRoots()
-  }
-
   /**
    * Computes sources files paths for the whole project. Cached in a way
    * that does not get invalidated upon source file change. Mainly called
@@ -149,7 +145,7 @@ class BloopImpl(evs: () => Seq[Evaluator], wd: os.Path) extends ExternalModule {
   def moduleSourceMap = Task.Input {
     val sources = Task.traverse(computeModules) { m =>
       // We're not using `allSources` here, one purpose. See https://github.com/com-lihaoyi/mill/discussions/4530
-      deferredAllSources(m).map { paths =>
+      m.ideSources.map { paths =>
         name(m) -> paths.map(_.path)
       }
     }()
@@ -413,7 +409,6 @@ class BloopImpl(evs: () => Seq[Evaluator], wd: os.Path) extends ExternalModule {
     // //////////////////////////////////////////////////////////////////////////
 
     val project = Task.Anon {
-
       val mSources = moduleSourceMap()
         .get(name(module))
         .toSeq
--- a/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
@@ -116,19 +116,15 @@ class BloopImpl(evs: () => Seq[Evaluator], wd: os.Path) extends ExternalModule {
     JavaModuleUtils.transitiveModules(mod, accept)
   }
 
-  private def modulesToOutFolderMap: Map[JavaModule, os.Path] = {
+  protected def computeModules: Seq[JavaModule] = {
     val evals = evs()
     evals.flatMap { eval =>
       if (eval != null)
         JavaModuleUtils.transitiveModules(eval.rootModule, accept)
-          .collect { case jm: JavaModule => (jm, eval.outPath) }
+          .collect { case jm: JavaModule => jm }
       else
         Seq.empty
-    }.toMap
-  }
-
-  protected def computeModules: Seq[JavaModule] = {
-    modulesToOutFolderMap.keys.toSeq
+    }
   }
 
   // class-based pattern matching against path-dependant types doesn't seem to work.
--- a/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
+++ b/contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
@@ -116,15 +116,19 @@ class BloopImpl(evs: () => Seq[Evaluator], wd: os.Path) extends ExternalModule {
     JavaModuleUtils.transitiveModules(mod, accept)
   }
 
-  protected def computeModules: Seq[JavaModule] = {
+  private lazy val modulesToOutFolderMap: Map[JavaModule, os.Path] = {
     val evals = evs()
     evals.flatMap { eval =>
       if (eval != null)
         JavaModuleUtils.transitiveModules(eval.rootModule, accept)
-          .collect { case jm: JavaModule => jm }
+          .collect { case jm: JavaModule => (jm, eval.outPath) }
       else
         Seq.empty
-    }
+    }.toMap
+  }
+
+  protected def computeModules: Seq[JavaModule] = {
+    modulesToOutFolderMap.keys.toSeq
   }
 
   // class-based pattern matching against path-dependant types doesn't seem to work.
@@ -140,7 +144,8 @@ class BloopImpl(evs: () => Seq[Evaluator], wd: os.Path) extends ExternalModule {
    */
   def moduleSourceMap = Task.Input {
     val sources = Task.traverse(computeModules) { m =>
-      m.allSources.map { paths =>
+      // We're not using `allSources` here, one purpose. See https://github.com/com-lihaoyi/mill/discussions/4530
+      m.sources.map { paths =>
         name(m) -> paths.map(_.path)
       }
     }()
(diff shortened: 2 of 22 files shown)


lihaoyi:
@Baccata one new development worth mentioning is that https://github.com/com-lihaoyi/mill/pull/4524 targeting 0.13.x, we've moved the computation of `JavaModule` source paths in general to static `def sourcesFolders: Seq[os.SubPath]` members computer before evaluation. This was done for reasons unrelated to codegen, but perhaps now that we're doing it a similar approach could be taken for generated sources? 

Following the naming convention would just define a `def JavaModule#generatedSourcesFolders: T[Seq[os.SubPath]]` and use that for IDE codegen, which I think is similar to what you are proposing here, just to a slightly different name

Baccata:
See [c262ea6](https://github.com/com-lihaoyi/mill/pull/4621/commits/c262ea6a20cfad8bed39f1392e2c70f6e5a84947) for a showcase of what I meant in https://github.com/com-lihaoyi/mill/pull/4621#issue-2877900833

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -847,10 +847,52 @@ trait JavaModule
...
+      deferredSourceGenerators.map { t => t.map((_: Unit) => deferredSourceRoots(t)) }
+    }().flatMap(_.apply(ws)).map(PathRef(_))
+  }
+
   /**
    * The folders containing all source files fed into the compiler
    */
-  def allSources: T[Seq[PathRef]] = Task { sources() ++ generatedSources() }
+  def allSources: T[Seq[PathRef]] =
+    Task { sources() ++ generatedSources() ++ deferredGeneratedSources() }
```
Baccata, line 895: Keeps the current (0.12.x) behaviour of `generatedSources` to avoid breaking the ecosystem

contrib/bloop/src/mill/contrib/bloop/BloopImpl.scala
```diff
@@ -133,14 +137,19 @@ class BloopImpl(evs: () => Seq[Evaluator], wd: os.Path) extends ExternalModule {
       !module.asInstanceOf[outer.Module].skipBloop
     else true
 
+  private def deferredAllSources(m: JavaModule): T[Seq[PathRef]] = Task {
+    m.sources() ++ m.generatedSources() ++ m.allDeferredSourceRoots()
```
Baccata, line 141: Keeps the current (0.12.x) behaviour of `generatedSources` being run to produce IDE config, in order to avoid breaking the ecosystem

Baccata:
See later changes for a showcase of what I meant in https://github.com/com-lihaoyi/mill/pull/4621#issue-2877900833 ... the names are horribly bad but will change them later

bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
```diff
@@ -234,12 +234,14 @@ private class MillBuildServer(
         case module: MillBuildRootModule =>
```
Baccata, line 234: Not sure why this warrants a specific case, considering it extends JavaModule anyway

lihaoyi:
@Baccata can you try reproducing the failure locally? You can run `ci/test-mill-bootstrap.sh`

Baccata:
Yes, it reproduces locally. Still extremely puzzled as to what's wrong though.

lihaoyi:
Oh I just realized this was targeting 0.12.x. Maybe its fixed by https://github.com/com-lihaoyi/mill/pull/4625 ?

bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
lefou: Could be some left over from a time, where `MillBuildRootModule` was either not a `JavaModule` or needed different treatment. If both cases are identical now, we can merge them.

Baccata:
Should I close this PR and re-open it anew to remove the noise of the previous conversations ?

lihaoyi:
No leaving the PR as is is fine, I'll take a look when I have time

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -847,10 +847,61 @@ trait JavaModule
...
+    val explicitSourceRoots = task.generatedSourceRoots
+    if (explicitSourceRoots.isEmpty) Seq(dest)
+    else explicitSourceRoots.map(subPath => dest / subPath)
+  }
+
+  /**
+   * Computes the list of source roots that will be produced by [[sourceGenerators]] without
+   * actually running the generators in question.
+   */
+  def predictedGeneratedSourceRoots: T[Seq[PathRef]] = T {
```
lihaoyi, line 875: The terminology is kind of weird here, as are the types. How about

```scala
def deferredGeneratedSourcePaths: T[Seq[os.Path]]
``` 

I don't like `predicted`, since it sounds like it could refer to all generated sources, which it doesn't. The earlier term `deferred` is a bit better since it sounds more like it is a separate thing from the other generated sources.

If it's just returning the paths, we can just return `os.Path`s rather than `PathRef`s

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -847,10 +847,61 @@ trait JavaModule
...
+    val ws = T.workspace
+    sourceGenerators.flatMap(t => predictGeneratedSourceRoots(t)(ws)).map(PathRef(_))
+  }
+
+  /**
+   * Runs the lazy source generators, returning references to the expanded generated source roots
+   * they are supposed to have written code in.
+   */
+  final def generatedSourceRoots: T[Seq[PathRef]] = T {
```
lihaoyi, line 884: We can call this `deferredGeneratedSources`, to be in line with `sources`, `generatedSources`, `allSources`, etc.

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -847,10 +847,61 @@ trait JavaModule
...
 
+  /**
+   * A set of tasks producing generated code. The difference between this and
+   * [[generatedSources]] lies in the fact that IDE-related tasks
+   * (BSP/Bloop/GenIdea) will not run the tasks, but will use them to statically
+   * determine the location of the source code they will produce by looking up their
+   * [[Task#generatedSourceRoots]], knowing that these roots will be present under
+   * the dest directories once the task is run.
+   */
+  def sourceGenerators: Seq[NamedTask[_]] = Seq.empty
```
lihaoyi, line 858: Maybe `deferredSourceGenerators`?

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -847,10 +847,52 @@ trait JavaModule
...
+   * [[Task#generatedSourceRoots]], knowing that these roots will be present under
+   * the dest directories once the task is run.
+   */
+  def deferredSourceGenerators: Seq[NamedTask[Unit]] = Seq.empty
+
+  private def deferredSourceRoots(task: NamedTask[Unit])(workspace: os.Path): Seq[os.Path] = {
+    val dest = mill
+      .eval
+      .EvaluatorPaths
+      .resolveDestPaths(workspace / "out", task)
```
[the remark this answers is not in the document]
lihaoyi: Yeah this is fine. A bit verbose but that's the best we got

main/define/src/mill/define/Task.scala
```diff
@@ -819,8 +850,13 @@ class PersistentImpl[+T](
     t: Task[T],
     ctx0: mill.define.Ctx,
     readWriter: RW[_],
-    isPrivate: Option[Boolean]
-) extends TargetImpl[T](t, ctx0, readWriter, isPrivate) {
+    isPrivate: Option[Boolean],
+    generatedSourceRoots: Seq[os.SubPath]
```
lihaoyi, line 854: `deferredGeneratedSourceRoots`? It's a bit annoyingly verbose, but we do want to use the same terminology throughout so it's clear what tasks it is related to

lihaoyi:
I think the implementation is fine. The main thing I would say is we should try and standardize the naming so all the related tasks have the same name prefix so users can see that they are related. `deferredGeneratedSource*` is annoyingly long, maybe we can find a shorter prefix that is still clear to users. But if not I suppose this is a niche enough feature that being a bit verbose is fine.

Baccata:
Happy to defer to you on the naming decision (pun very much intended). See [7ceb97f](https://github.com/com-lihaoyi/mill/pull/4621/commits/7ceb97ffe64513aecadf0cee517611fc0307faec)

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -847,10 +847,61 @@ trait JavaModule
...
+   * [[Task#deferredGeneratedSourceRoots]], knowing that these roots will be present under
+   * the dest directories once the task is run.
+   */
+  def deferredGeneratedSourceTasks: Seq[NamedTask[_]] = Seq.empty
+
+  private def predictGeneratedSourceRoots(task: NamedTask[_])(workspace: os.Path): Seq[os.Path] = {
+    val dest = mill
+      .eval
+      .EvaluatorPaths
+      .resolveDestPaths(workspace / "out", task)
```
lefou, line 864: Since the `out`-path is configurable, this must depend on the real out-path in some way.

lefou (commented):
What do you think of moving all changes currently done in `JavaModule` into a dedicated trait `DeferredGeneratedSources` and extends in from `JavaModule`. This will keep them closely together. It will also keep `JavaModule` more lean and free of that stuff.

scalalib/src/mill/scalalib/JavaModule.scala
Baccata: Yeah that's my concern, but I've not found an API for retrieving this bit of information

lihaoyi: Currently `Evaluator` has an `outPath` field, but it's not available in normal tasks like this one. Maybe we can expose a `out: os.Path` member on `mill.api.Ctx` for such advanced use cases, similar to how we already expose `workspace: os.Path`?

Baccata:
> What do you think of moving all changes currently done in JavaModule into a dedicated trait DeferredGeneratedSources and extends in from JavaModule? This will keep them closely together. It will also keep JavaModule more lean and free of that stuff. We could even avoid default-extending it and just mix it in in case it's needed.

Well my personal opinion is really that the current `generatedSources` should be deprecated in favour of the deferred version, for a similar argument that lead me to start the discussion : I think any kind of IDE integration should be as resilient as possible. I believe that if I have errors in a protobuf file that I use to generate Java or Scala code, I should still be able to load my project into the IDE and be able to change code that's not necessarily related to the protobuf pipeline. I therefore believe that the current `generatedSources` task, although intuitive, was ill-designed from the get-go, and that the use of the `deferred` methods should be encouraged instead within the mill ecosystem. 

Therefore, I would rather the `JavaModule` had them than for them to be separate in a mix-in that IDE-config code needs to be aware of. Now I'd personally be fine having them in a separate trait, as conceptually they don't have to be tied to Java/Scala and other integrations could benefit from them. 

Now that being said, that's just a personal opinion. If yourself and Haoyi make an executive decision to make this a mix-in, I'll abide and will amend the contribution in consequence.

lihaoyi:
For now lets go with a separate trait. I think it's still a bit early to deprecate the existing workflow and tell people to use the new one. We probably want someone using it for a while and happy with it before we deprecate the old way, which would probably put it some time after the 0.13.0 release

Baccata:
@lihaoyi : just to be clear : you want a separate trait that `JavaModule` inherits from, or a separate trait that would be mixed-in in userland ?

lihaoyi:
I don't have a strong opinion between those two options. @lefou do you have a preference?

lefou:
TBH, I find the new proposed solution (although being a solution to some immediate issue) ugly and complicated. (I also don't think the current `generatedSources` is broken in any way. If you have generated sources that aren't required to be properly compiled, those are probably not sources but just resources. I admit, the IDE integration currently doesn't handle failures well.) 

Therefore I tend to not force it on all users of `JavaModule` and make it completely opt-in. So, let user-land mix-in `DeferredGeneratedSources` if they really need to.

lihaoyi:
mixin sounds good to me

Baccata:
> TBH, I find the new proposed solution (although being a solution to some immediate issue) ugly and complicated 

I agree I don't like the naming, but don't really want to bikeshed on it, as long as the maintainers are happy with it. 

> If you have generated sources that aren't required to be properly compiled, those are probably not sources but just resources

I think that's a mis-interpretation of the problem. To simplify it as intuitively as possible : 

1. Should I be able to load a project in my IDE if the Scala code it contains doesn't compile ? => everyone agrees on yes
2. Should I be able to load a project in my IDE if the protobuf code it depends on to generate Scala code is invalid => ??? 

Saying that you don't think the current `generatedSources` is broken in any way implies that your answer to 2) is "no", whereas I think "yes". 

> I tend to not force it on all users of JavaModule and make it completely opt-in. 

Happy to abide 👌

scalalib/src/mill/scalalib/JavaModule.scala
Baccata: Added an `out` member in [2d54b8f](https://github.com/com-lihaoyi/mill/pull/4621/commits/2d54b8fdab43e269ef4589471f01a945bce6e99a)

lefou:
> 2. Should I be able to load a project in my IDE if the protobuf code it depends on to generate Scala code is invalid => ???
> 
> Saying that you don't think the current `generatedSources` is broken in any way implies that your answer to 2) is "no", whereas I think "yes".

My answer is of course "yes". I do think our IDE integration should not fail, if some source generators don't work. We want see some warnings, of course. Also, compilation most likely won't work, even if the IDE bypasses the build tool to compile things, but that not the question here. 

I think we need to solve the issue in the IDE integration, not in the task signature. If IDE integration just deals with a failed source generator and continues with either a stub-path or without the path of the generated sources at all (depends on how we fail or how we are able to recover). Once we repaired the source generator, we re-import and the missing generated sources should appear in the IDE. I don't see where in that picture a different task signature for `generatedSources` is needed. Once we give away an incomplete result (here a target path without the files that are supposed to be on that path), we give up on our principle of a immutable, cachable and reproducible results.

lihaoyi:
> I think we need to solve the issue in the IDE integration, not in the task signature. If IDE integration just deals with a failed source generator and continues with either a stub-path or without the path of the generated sources at all (depends on how we fail or how we are able to recover). Once we repaired the source generator, we re-import and the missing generated sources should appear in the IDE. I don't see where in that picture a different task signature for generatedSources is needed. Once we give away an incomplete result (here a target path without the files that are supposed to be on that path), we give up on our principle of a immutable, cachable and reproducible results.

The basic problem this PR is trying to solve is that currently, as I understand it, is that Mill does not provide the path to generated sources if generation fails: you only get a `PathRef` on success and nothing on failure. So there's no way to continue with a stub path, because no such path is available.

This PR doesn't change the idea of immutable/cacheable/reproducible results, all it does is let you list out which tasks will contain generated sources, and at what paths, without needing to actually run them. That is what allows the IDE to have paths it can use even if the tasks themselves fail

lefou:
Might be there is an unspoken assumption: We don't want to re-import/refresh the project in the IDE after we fixed source generation? Which means, all source paths need to stay stable. This is equivalent with what we modelled in `bspCompileClassesPath` or `bspCompileClasspath` which both return some `UnresolvedPath`s. They're are typically implemented as follows: In case we know the implementation, we provide some realistic path (without producing the result we later expect in that location), otherwise we fall back to call the backing task.

lefou:
I don't want to hold this PR any longer with discussions. The current solution as dedicated trait is isolated enough that it can be replaced with some later concept, once we have it.

I think we need to sit another round on the design table and find out, how a nice solution could look like. I think https://github.com/com-lihaoyi/mill/discussions/1779 might be a part of the solution.

Just one last though. Have you considered to explicitly encode the fallback into the task result, e.g. 

```scala
def instableGeneratedSources: Task[Either[Seq[os.Path], Seq[PathRef]]]
```

It's appealing as it is much less complex as a concept. Don't know, if it is enough to represent the problem?

lihaoyi:
I think the issue with `def instableGeneratedSources: Task[Either[Seq[os.Path], Seq[PathRef]]]` would be how does one actually implement it? Presumably you'd do a `try-catch`, and in the `catch` case you would go look up the `.dest` folders of all the generated sources and return them as a list. That's basically identical to what this PR does, except it skips the `try-catch` part and just returns the `.dest` folders directly

lefou:
> except it skips the `try-catch` part and just returns the `.dest` folders directly

Do we make sure, the real implementation is returning the `.dest` folder? Do we fail if not?

lihaoyi:
> Do we make sure, the real implementation is returning the .dest folder? Do we fail if not?

The PR doesn't do such checks IIRC. It could, but it doesn't

Baccata:
> Do we make sure, the real implementation is returning the same .dest folder? Do we fail if not? 

Actually that's why I've elected to to require the NamedTasks yield a `PathRef`. Instead, the interface [takes](https://github.com/com-lihaoyi/mill/pull/4621/files#diff-a0b63d74a7afa9663b4cd76aaf3e8e1781a83659f02abe7fd04a0f29c0579101R24) an existential. Initially I wanted to force them to return `Unit` to prevent the assumption that the returned value would be used in any way, but figured it may not be necessary. 

I could change the interface to have it require `Seq[NamedTask[PathRef]]` which would allow to perform a check, but I think it'd be weird to force the user to essentially return `PathRef(T.dest)` (most of the time).

lihaoyi:
I'd say for now just tag `DeferredGeneratedSourcesModule` with `@experimental` and try it out for now. That'll give us some experience to see how well it works in practice, and whether we want to pursue it further, tweak it, or try something else later

scalalib/src/mill/scalalib/DeferredGeneratedSourcesModule.scala
```diff
@@ -0,0 +1,69 @@
...
+    val explicitSourceRoots = task.deferredGeneratedSourceRoots
+    if (explicitSourceRoots.isEmpty) Seq(dest)
+    else explicitSourceRoots.map(subPath => dest / subPath)
+  }
+
+  /**
+   * Computes the list of source roots that will be produced by [[deferredGeneratedSourceTasks]] without
+   * actually running the generators in question.
+   */
+  final def predictedGeneratedSources: T[Seq[PathRef]] = T {
```
lefou, line 41: Since this tasks does not depend on the paths content, shouldn't it return `os.Path`s?

Baccata: Fair point !

Baccata: [9e736fa](https://github.com/com-lihaoyi/mill/pull/4621/commits/9e736fafc6517031d5c68a392957fe9b3f619f02)

Baccata:
Module marked @experimental in https://github.com/com-lihaoyi/mill/pull/4621/commits/3b7b174115bbf2f86a1da09127d7bf558ab6f5fd

lefou (approved):
Looks good to me.

lihaoyi:
Thanks @Baccata ! I kicked off a publish run on the `0.12.x` branch here https://github.com/com-lihaoyi/mill/actions/runs/13694389375 so once that's done the logs will have the a new development release you can try out. I think we don't need to port it to `main` immediately, since I wouldn't be surprised if it needed further changes or tweaks before we end up happy with it

PR #4624, opened 2025-02-26
Support reading the Mill version from a using directive

lefou:
This experimentally adds support to read the Mill version from a using directive.

The line 

```scala
//> using mill.version <version>
``` 
can be used to specify the Mill version in the `build.mill` file.

* See also Discussion https://github.com/com-lihaoyi/mill/issues/3062

## Testing

To exercise that the change is working, I removed the `.config/mill-version` file and added the version to the `build.mill` file.

```diff
diff --git a/.config/mill-version b/.config/mill-version
deleted file mode 100644
index 3061735abf5..00000000000
--- a/.config/mill-version
+++ /dev/null
@@ -1 +0,0 @@
-0.12.8-1-46e216-native

diff --git a/build.mill b/build.mill
index e64f5c62f98..7d1cadcfeb0 100644
--- a/build.mill
+++ b/build.mill
@@ -1,3 +1,5 @@
+//> using mill.version 0.12.8-1-46e216-native
+
 package build
 // imports
 import com.github.lolgab.mill.mima.Mima
```

All CI jobs pass as before and run with the configured Mill version, so Mill version selection seems to work.

## Why `using directives`?

I wanted an way to configure the Mill version in the build script for a long time. Does it need to use the `using directive`? _Maybe not._ 

Using directives were newly invented by Scala CLI project. There were earlier versions supporting richer definitions, but the scala-cli devs settled with a rather primitive key-value format. I don't want to question that. Fact is, it makes processing easy, although it may not be suitable for all possible config values that may come in the future. Just let me highlight the major benefits:

* It is a (rather new) standard in the Scala community. Most users know how to read a using directive.
* It's easy to parse. This PR comes with a shell and a Windows batch script that already can parse the using directive. 
* It's hard to break by formatters, and since it's a Scala standard, formatters are aware of it being a config format
* It's already supported by IDEs and syntax highlighters. For example, look that Github renders the given snippet already with some syntax colors:

```scala
//> using mill.version 0.12.8
```

## Other changes

* Removed the `--mill-version` options from the bootstrap scripts. Since it's not a regular option that Mill supports when run without a bootstrap script.

* Fixed the precedence of `.mill-version` over `.config/mill-version` in Mill's own runtime check.

## Precedence

If a Mill version is given via multiple configuration options, the following precedence rules apply. First option wins:

* Use a `MILL_VERSION` env var
* Use a `.mill-version` file
* Use a `.config/mill-verison` file
* Read a `mill.version` directive from `build.mill`
* Read a `mill.version` directive from `build.mill.scala`
* Read a `mill.version` directive from `build.sc`


lihaoyi:
I said as much on the original discussion, but let's keep this PR open for now without merging it.

PR #4472, opened 2025-02-04
Break the dependency between repositoriesTask and ivyDeps

alexarchambault:
Same as #4465, but against the `0.12.x` branch

---

mill-scalablytyped (but possibly other plugins or user setups too) adds a dependency on `repositoriesTask` from `ivyDeps`. Since the introduction of `JavaModule#coursierProject`, Mill had a dependency the other way around, `repositoriesTask` depends on `ivyDeps`.

This creates a cycle, leading to StackOverflowException-s.

In order to work around that, this PR splits both `repositoriesTask` and `defaultResolver`, adding:
- `allRepositoriesTask`: basically `repositoriesTask`, with the Mill internal repository added
- `internalResolver`: same as `defaultResolver`, with the Mill internal repository added (via `allRepositoriesTask`)

If users need to resolve purely external modules (most common case it seems), they can keep using `repositoriesTask` or `defaultResolver`.

If they need to resolve some Mill internal modules (usually brought in via `JavaModule#coursierDependency`), they now need to use `allRepositoriesTask` and `internalResolver` instead of `repositoriesTask` and `defaultResolver`.

That way, no cycle is introduced when users only need to resolve external modules.

Fixes #4457

PR #4625, opened 2025-02-26
Fix 0.12.x branch bootstrap job

alexarchambault:
Seems the bootstrap job on CI consistently fails, like [here](https://github.com/com-lihaoyi/mill/actions/runs/13409951292/job/37459400148). This should fix it.


lefou (approved):
Makes sense, since `allRepositories` was only introduced lately: https://github.com/com-lihaoyi/mill/pull/4472

This should be made permanent, once we rebootstrap on Mill 0.12.9.

PR #4626, opened 2025-02-26
Enable Gradle module support for Android and Kotlin

alexarchambault:
This enables Gradle Module support in `KotlinModule` and `AndroidModule`, and allows users to enable it in any `CoursierModule` if they feel like it.

Users can enable or disable Gradle Module support via `CoursierModule#checkGradleModules`, like
```scala
def checkGradleModules: T[Boolean] = true
```

Gradle Modules rely on so called "variant attributes" to select a "variant" in a Gradle Module definition. Commonly used attributes are already set by this PR on `KotlinModule`, `AndroidModule`, `JavaModule`, etc. Variant attributes can be set like
```scala
def resolutionParams: Task[ResolutionParams] = Task.Anon {
  super.resolutionParams().addVariantAttributes(
    "org.jetbrains.kotlin.platform.type" -> VariantMatcher.Equals("jvm"),
    "thing.foo" -> VariantMatcher.AnyOf(VariantMatcher.Equals("first"), VariantMatcher.Equals("second"))
  )
}
```

To change a previously set attribute, just set a new value for it this way.

Missing variant attributes can result in resolution errors like
```text
Error downloading org.jetbrains.kotlin:kotlin-stdlib:1.9.22
  Found too many variants in org.jetbrains.kotlin:kotlin-stdlib:1.9.22 for { org.gradle.category=library, org.gradle.usage=runtime }:
  jsRuntimeElements: {
    org.gradle.category: library
    org.gradle.jvm.environment: non-jvm
    org.gradle.usage: kotlin-runtime
    org.jetbrains.kotlin.js.compiler: ir
    org.jetbrains.kotlin.platform.type: js
    $relocated: true
  }
  jsV1RuntimeElements: {
    org.gradle.category: library
    org.gradle.jvm.environment: non-jvm
    org.gradle.usage: kotlin-runtime
    org.jetbrains.kotlin.js.compiler: legacy
    org.jetbrains.kotlin.platform.type: js
    $relocated: true
  }
  jvmRuntimeElements: {
    org.gradle.category: library
    org.gradle.jvm.environment: standard-jvm
    org.gradle.libraryelements: jar
    org.gradle.usage: java-runtime
    org.jetbrains.kotlin.platform.type: jvm
  }
  wasmJsRuntimeElements: {
    org.gradle.category: library
    org.gradle.jvm.environment: non-jvm
    org.gradle.usage: kotlin-runtime
    org.jetbrains.kotlin.platform.type: wasm
    org.jetbrains.kotlin.wasm.target: js
    $relocated: true
  }
  wasmWasiRuntimeElements: {
    org.gradle.category: library
    org.gradle.jvm.environment: non-jvm
    org.gradle.usage: kotlin-runtime
    org.jetbrains.kotlin.platform.type: wasm
    org.jetbrains.kotlin.wasm.target: wasi
    $relocated: true
  }
```

Errors like this are addressed by adding variant attributes like above. Among the specified attributes, an attribute not found in the Gradle Module is ignored, while an attribute found in the Gradle Module must have the same value (more precisely, a value matched by the passed "variant matcher"). If the variant with the most matching attributes gets selected. If several variants correspond to this case, resolution fails the same way, and users should pass extra attributes to disambiguate things.

cc @vaslabs


vaslabs:
looks promising, some issues on android don't seem serious, I'll have time to have a go tomorrow!

alexarchambault:
(Didn't mean to push the `Clean-up more stuff` commit in this PR, nothing to see for Gradle Modules in it)

alexarchambault:
This includes both https://github.com/com-lihaoyi/mill/pull/4725 and https://github.com/com-lihaoyi/mill/pull/4727 for now, so the diff might be a bit confusing…

vaslabs:
We are also ready to test this, @irodotos7 will give it a go for Android

alexarchambault:
@vaslabs @irodotos7 About this PR, it's kind of in flux right now. I opened it with the first iterations of Gradle Module support in coursier. This required explicitly enabling Gradle Module support for individual dependencies, by calling `addVariantAttributes` on them (and passing attribute values).

Since then, I made it possible to enable Gradle Module support for a whole resolution (rather than on individual dependencies), and managed to enable them on all tests of coursier (https://github.com/coursier/coursier/pull/3352, that's just a check, not meant to be merged), which allowed to find and address some issues.

This PR enables that. Now the missing part is the actual variant attributes to pass (which should differ for Kotlin JVM, Kotlin JS, Android, …).

These can be passed this way
```scala
def resolutionParams = Task.Anon {
  super.resolutionParams().addVariantAttributes(
    "org.gradle.foo" -> VariantSelector.VariantMatcher.Equals("thing")
  )
}
```

Once [the CI](https://github.com/alexarchambault/mill/actions/runs/14520300094) has passed, I'm going to proceed to add those where they're needed (`KotlinModule`, `AndroidModule`, etc.), in order to make the CI green.

vaslabs:
> @vaslabs @irodotos7 About this PR, it's kind of in flux right now. I opened it with the first iteration of Gradle Module support in coursier. This required explicitly enabling Gradle Module support for individual dependencies, by calling `addVariantAttributes` on them (and passing attribute values).
> 
> Since then, I made it possible to enable Gradle Module support for a whole resolution (rather than on individual dependencies), and managed to enable them on all tests of coursier ([coursier/coursier#3352](https://github.com/coursier/coursier/pull/3352), that's just a check, not meant to be merged), which allowed to find and address some issues.
> 
> This PR enables that. Now the missing part is the actual variant attributes to pass (which should differ for Kotlin JVM, Kotlin JS, Android, …).
> 
> These can be passed this way
> 
> ```scala
> def resolutionParams = Task.Anon {
>   super.resolutionParams().addVariantAttributes(
>     "org.gradle.foo" -> VariantSelector.VariantMatcher.Equals("thing")
>   )
> }
> ```
> 
> Once [the CI](https://github.com/alexarchambault/mill/actions/runs/14520300094) has passed, I'm going to try to add those where they're needed (`KotlinModule`, `AndroidModule`, etc.), in order to make the CI green. I already ran into common attributes needed for Kotlin or Android in [coursier/coursier#3352](https://github.com/coursier/coursier/pull/3352).

would there be a possibility to much by precedence? e.g. try to match `androidJvm`, if not `common` if not `jvm` ?

core/util/src/mill/util/Jvm.scala
```diff
@@ -495,7 +495,7 @@ object Jvm {
       }
       .pipe { cache =>
         if (ctx.fold(false)(_.offline)) cache.withCachePolicies(Seq(CachePolicy.LocalOnly))
-        else cache
+        else cache.withCachePolicies(Seq(CachePolicy.FetchMissing))
```
alexarchambault, line 498: CI seems to be green [here](https://github.com/alexarchambault/mill/actions/runs/14642052919/job/41087262442) (mostly, the red is transient GraalVM issues on Windows that should be fixed by https://github.com/com-lihaoyi/mill/pull/4917).

Once https://github.com/coursier/coursier/pull/3389 is merged and released, I'm planning to remove this line, and this PR should be good to review and merge.

vaslabs:
> > would there be a possibility to match by precedence? e.g. try to match `androidJvm`, if not `common` if not `jvm` ?
> 
> There is precedence now. In the current state of this PR, it's `androidJvm` then `jvm`, but I could add `common` between the two.

thanks, I had not yet have time to see how exactly things work with gradle irt android and kotlin multi-platform, I'd guess as long as there's a way to be flexible to we can experiment , should be alright, but in any case we can always test and come back to it

libs/androidlib/src/mill/androidlib/AndroidModule.scala
```diff
@@ -118,6 +120,14 @@ trait AndroidModule extends JavaModule {
     super.repositoriesTask() :+ AndroidSdkModule.mavenGoogle
   }
 
+  override def checkGradleModules: T[Boolean] = true
+  override def resolutionParams: Task[ResolutionParams] = Task.Anon {
+    super.resolutionParams().addVariantAttributes(
+      "org.jetbrains.kotlin.platform.type" ->
+        VariantMatcher.AnyOf(Seq(VariantMatcher.Equals("androidJvm"), VariantMatcher.Equals("jvm")))
```
vaslabs, line 127: ok this is what I meant in https://github.com/com-lihaoyi/mill/pull/4626#issuecomment-2827984709, that looks fine with me!

core/util/src/mill/util/Jvm.scala
alexarchambault: That's going to need https://github.com/coursier/coursier/pull/3392 too

vaslabs:
I tried setting common as well
```
  override def resolutionParams: Task[ResolutionParams] = Task.Anon {
    super.resolutionParams().addVariantAttributes(
      "org.jetbrains.kotlin.platform.type" ->
        VariantMatcher.AnyOf(Seq(VariantMatcher.Equals("androidJvm"), VariantMatcher.Equals("common"), VariantMatcher.Equals("jvm")))
    )
  }
  ```
  
  which [annotation](https://maven.google.com/web/index.html?q=androidx.annotation#androidx.annotation:annotation:1.7.0-alpha01) should have
  
  but I think some of our assumptions might be wrong.
  
  For example annotation which is referenced by androidx.test.ext:junit (which does not have a gradle module file published) has common, native and jvm platform types for its dependencies. With common in the mix it should have selected this variant:

```
    {
      "name": "metadataApiElements",
      "attributes": {
        "org.gradle.category": "library",
        "org.gradle.usage": "kotlin-metadata",
        "org.jetbrains.kotlin.platform.type": "common"
      },
      "dependencies": [
        {
          "group": "org.jetbrains.kotlin",
          "module": "kotlin-stdlib",
          "version": {
            "requires": "1.8.10"
          }
        }
      ],
      "files": [
        {
          "name": "annotation-metadata-1.7.0-alpha01.jar",
          "url": "annotation-1.7.0-alpha01.jar",
          "size": 8090,
          "sha512": "b1ec90df4a382357d8a0179511f5028890da753a7f79f3e7082c02e7859fadd9eed7fcb7e4dfc2f81e8fa872ecd4c4ae50dcb41399ca9fdbdaa8872fdf350a1a",
          "sha256": "ec0ad3a875902d25640e7bd07de02322354fabc92de97c6d0919e79bd30d2b27",
          "sha1": "60c3a9112a7c4afd3968f52b9f10deed9e395a94",
          "md5": "1409db7bf41bf36dacc5a68919e0d25e"
        }
      ]
    },
    ```
    
    but it doesn't see it (I removed jvm to quickly verify).
    
    Which is probably due to the

```    
                      "org.gradle.usage" -> cs.VariantSelector.VariantMatcher.Api
```

in Java module. I need to run more tests to have a better idea how to untangle this, but maybe you'll know better

alexarchambault:
> For example annotation which is referenced by androidx.test.ext:junit (which does not have a gradle module file published) has common, native and jvm platform types for its dependencies. With common in the mix it should have selected this variant:

The variant with `"org.jetbrains.kotlin.platform.type": "common"` also has `"org.gradle.usage": "kotlin-metadata"`. It's not selected because we only keep variants whose `org.gradle.usage` either equals `api` or ends with `-api` (for compile dependencies), or either equals `runtime` or ends with `-runtime` (for runtime dependencies).

alexarchambault:
I don't know what `kotlin-metadata` corresponds too. Do you think we should keep that?

alexarchambault:
> I tried to give it a go on a clean branch but I still get this jvm dependency

In the JAR list you give, I think `kotlinx-coroutines-core-1.3.4.jar` is the problem. `kotlinx-coroutines-core-jvm-1.7.1.jar` is there, but both `kotlinx-coroutines-core` and `kotlinx-coroutines-core-jvm` versions should have been the same. When they are, `org.jetbrains.kotlinx:kotlinx-coroutines-core:1.7.1` actually only brings `kotlinx-coroutines-core-jvm-1.7.1.jar`, there's no `kotlinx-coroutines-core-1.7.1.jar`.  It might be possible to handle that during resolution, by taking `endorseStrictVersions` into account in the [module of `org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm`](https://repo1.maven.org/maven2/org/jetbrains/kotlinx/kotlinx-coroutines-core-jvm/1.7.1/kotlinx-coroutines-core-jvm-1.7.1.module). Looking into it.

alexarchambault:
https://github.com/coursier/coursier/pull/3402 should fix that problem

vaslabs:
hm, I think we are still missing something. I'm still investigating, but I did some testing with gradle. I can see the -jvm dependencies, but they are marked as [dependency constraints](https://docs.gradle.org/current/userguide/dependency_constraints.html#dependency-constraints)

e.g.

```
|    |    |    |    +--- org.jetbrains.kotlinx:kotlinx-coroutines-android:1.6.4 -> 1.7.1
|    |    |    |    |    +--- org.jetbrains.kotlinx:kotlinx-coroutines-core:1.7.1
|    |    |    |    |    |    \--- org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.7.1
|    |    |    |    |    |         +--- org.jetbrains:annotations:23.0.0
|    |    |    |    |    |         +--- org.jetbrains.kotlinx:kotlinx-coroutines-bom:1.7.1
|    |    |    |    |    |         |    +--- org.jetbrains.kotlinx:kotlinx-coroutines-android:1.7.1 (c)
|    |    |    |    |    |         |    +--- org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.7.1 (c)
|    |    |    |    |    |         |    \--- org.jetbrains.kotlinx:kotlinx-coroutines-core:1.7.1 (c)

```

vaslabs (approved):

lihaoyi (approved):
shipit

alexarchambault:
Alright 😅

core/util/src/mill/util/Jvm.scala
```diff
@@ -556,6 +563,7 @@ object Jvm {
       repositories: Seq[Repository],
       deps: IterableOnce[Dependency],
       force: IterableOnce[Dependency],
+      checkGradleModules: Boolean,
```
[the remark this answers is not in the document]
alexarchambault: Will fix that in a subsequent PR, I think

PR #4632, opened 2025-02-28
Update Scaladocs for NativeImageModule

ajaychandran:
Update Scaladocs to reflect that `nativeImage` is not a command.

### DIFF
--- a/scalalib/src/mill/scalalib/NativeImageModule.scala
+++ b/scalalib/src/mill/scalalib/NativeImageModule.scala
@@ -27,10 +27,6 @@ trait NativeImageModule extends WithZincWorker {
   /**
    * [[https://www.graalvm.org/latest/reference-manual/native-image/#from-a-class Builds a native executable]] for this
    * module with [[finalMainClass]] as the application entry point.
-   *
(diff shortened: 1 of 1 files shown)


zelosleone:
https://www.graalvm.org/latest/reference-manual/native-image/

It's a compiler and they are options for the compiler, native image itself is not meant to be used as a command...

ajaychandran:
I meant `NativeImageModule.nativeImage` was originally defined as a `Task.Command`.

zelosleone (commented):
I can see the changes, you deleted the documentation about a compiler's, native image, options.

ajaychandran:
The `NativeImageModule.nativeImage` task was originally defined as a `Task.Command` that accepted arguments passed as-is to the `native-image` tool. The intention was to allow users to pass in platform specific options from the CLI and use `NativeImageModule.nativeImageOptions` for options shared across platforms.

Later, the command arguments were removed. This PR updates the relevant Scaladocs to reflect this by 
- removing comments for `args` parameter that no longer exist
- removing recommendation for options that is no longer applicable

PR #4634, opened 2025-02-28
Pass Mill source JARs to Bloop (#4608)

alexarchambault:
Same as https://github.com/com-lihaoyi/mill/pull/4608, but against `main`

The goal of this PR is to pass source JARs corresponding to the Mill class path to Bloop. These are added in the root modules, where they're needed to get go-to-source to work for Mill build files.


lihaoyi (approved):
Looks good to me

scalalib/src/mill/scalalib/CoursierModule.scala
```diff
@@ -317,6 +318,28 @@ object CoursierModule {
 
       res.orderedDependencies
     }
+
+    def getArtifacts[T: CoursierModule.Resolvable](
```
gamlerhart, line 322: Thanks for this. I was working on SBOM generation and I was not sure what the best way was to expose the detailed info. This will provide all the info I need.

PR #4636, opened 2025-03-01
move client tests to utest / scala

Quafadas:
This seeks to port the client tests to utest / scala. 

It should be simple baby step, but is a pre / requisite for being able to test a scala native client.


lihaoyi:
I think this looks reasonable, but lets keep it open for now until there's downstream changes that require this, and then review them together

Quafadas:
Hmmm... 



My goal is to gradually port the client parts of the codebase to scala, and then ultimately compile it to scala native in pursuit of 
https://github.com/com-lihaoyi/mill/issues/4016.

The problem I had previously, is that I did a bulk change, but some of the changes must have been problematic, as it broke the CI, and I think it became un-reviewable.

This new strategy is to try and port smaller chunks whilst keeping the CI green at each stage. A kind of binary search on the previous set of changes to narrow down where it went wrong as I couldn't figure it out from the end state.

Is there a reason to avoid a merge? I'm fine if there is, but then I also need to factor that into the approach, as it means the smaller, incremental strategy is also problematic...

lihaoyi:
You are free to do everything you mentioned to bisect your changes to figure out which parts are problematic. However, you should do that on your own fork of the Mill repo. The `main` branch is not a debugging ground for bisecting broken PRs, you can use your own branch for that and once the code is ready we can review it properly and get it into main (if that's what we decide to do)

Quafadas:
Fair enough. I'm willing to argue this isn't a "debug", it's an incremental change I didn't think had any knock on impacts. Nothing here is "broken". Equally, I can respect the principle you are defending, and this PR in isolation, I would accept is valueless. 

I would note that the previous work followed the strategy suggested above, but is essentially sunk cost against the rate of change on main. That's a risk I took on my own time, but not one I can take a second time. I'll wait for the rate of change to slow before looking at picking this up again.

PR #4645, opened 2025-03-04
Reduce number of os.Checker.Nop usages in javascript lip

pawelsadlo:


monyedavid:
@lihaoyi this jslib is outdated PublishModule and TypeScriptModule (now TscModule) have been significantly modified to accommodate the new features - package root and relative paths usage between mod dependencies.

The only javascript module currently in violation of the restriction is the `TestModule`.
We are generating the configuration files for each test suite and inserting them into the `out/test/compile.dest`.

Seeing the goal of this PR is to reduce the violations (which did not throw those errors at the time), #4647 `TestModule`s implementation can be easily modified to avoid the violation.

I misunderstood previously, I thought I was meant to fix the implementation of `GroupExecutions` and the error I was getting an unintended bug not a restriction to be enforced.

pawelsadlo:
@lihaoyi I have seen that  #4647 already addressed the issue, I'm ok with closing this PR

lihaoyi:
Got it. Thanks for taking a crack at this @pawelsadlo !

Issue #4327, opened 2025-01-14
Improve "building" of Mill bootstrap scripts


lefou:
Currently, the various start files are used as-is. They are cluttered in different locations and must be edited manually, which is error-prone and cumbersome.

To properly finish up https://github.com/com-lihaoyi/mill/issues/2161 and esp. comment https://github.com/com-lihaoyi/mill/issues/2161#issuecomment-2509720499, we should do the following:

* Move scripts (`mill`, `mill.bat`, `ci/mill.ps1`) into new dir `scripts` (or `scripts/src`)
* Replace all movable parts with placeholders (script version, default versions, links to web-site and download locations)
* Add a Mill build step to build final script from these templates
* Consolidate further testing under the `scripts` project
* Automate, when the `<project-root>/mill{.bat}` scripts need updates and fail CI or autofix it.

He-Pin:
I still not get how to local build a mill and make it been used in another project locally :(

PR #4648, opened 2025-03-04
Fix curl check in Windows wrapper script

SuperSodaSea:
The curl check in the `mill.bat` wrapper script is actually broken:

```batch
if not exist "%MILL%" (
    ...
    where /Q curl
    if %ERRORLEVEL% EQU 0 (
        curl -f -L "!DOWNLOAD_URL!" -o "!DOWNLOAD_FILE!"
    )
```

The `%ERRORLEVEL%` is inside of nested `if`, so it will be the value of `ERRORLEVEL` when `if not exist "%MILL%"` is executed, not the result of `where /Q curl`. It should be replaced with `!ERRORLEVEL!` to get the correct result.

Some minor indentation fixes are also included.

### DIFF
--- a/mill.bat
+++ b/mill.bat
@@ -140,7 +140,7 @@ if not exist "%MILL%" (
 
     for /F "delims=- tokens=1" %%A in ("!MILL_VERSION!") do set MILL_VERSION_BASE=%%A
     for /F "delims=- tokens=2" %%A in ("!MILL_VERSION!") do set MILL_VERSION_MILESTONE=%%A
-	set VERSION_MILESTONE_START=!MILL_VERSION_MILESTONE:~0,1!
+    set VERSION_MILESTONE_START=!MILL_VERSION_MILESTONE:~0,1!
     if [!VERSION_MILESTONE_START!]==[M] (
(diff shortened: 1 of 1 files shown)


lefou (approved):
Nice catch. Looks good to me. Thank you!

lefou:
Tested manually on a Window VM. Instead of `bitsadmin`, it's now using `curl` for download.

lihaoyi:
Nice catch! I had always just assumed `bitsadmin` was the windows implementation of `curl`, I didn't realize that it was meant as a fallback and that a real `curl` command also exists

SuperSodaSea:
BTW should we maintain the Script Version in the wrapper script? Seems that it was not changed after merged from the original millw repo.

<https://github.com/com-lihaoyi/mill/blob/9d35aa0f3eb6a7876b3aab7f583fedb42db2af94/mill.bat#L8>

lefou:
> BTW should we maintain the Script Version in the wrapper script? 

We don't want to hand-maintain this version, but having it in the script is helpful for users. We'd like to have a proper build process for all bootstrap scripts, which is tracked here:

* https://github.com/com-lihaoyi/mill/issues/4327

PR #4654, opened 2025-03-05
Add the organization name and url/homepage members to `PomSettings`

ShreckYe:
The `PomSettings.organization` member corresponds to a Maven group. The organization name and url properties are available in [Maven](https://maven.apache.org/pom.html#Organization), [Gradle](https://docs.gradle.org/current/dsl/org.gradle.api.publish.maven.MavenPomOrganization.html), and [sbt](https://www.scala-sbt.org/1.x/docs/Howto-Project-Metadata.html#Set+the+project+organization). Shall we add them in `PomSettings` for Mill too?

I haven't added the `organizationUrl`/`organizationHomepage` member yet as I am not sure whether we stick to the former name (more similar to Maven and Gradle) or the latter (consistent with sbt). I have added a backward-comaptible companion object `apply` method for the code to work with minimal modification and haven't updated the callers yet.

### DIFF
--- a/scalalib/src/mill/scalalib/publish/settings.scala
+++ b/scalalib/src/mill/scalalib/publish/settings.scala
@@ -69,13 +69,32 @@ case class Developer(
 case class PomSettings(
     description: String,
     organization: String,
+    organizationName: String,
     url: String,
     licenses: Seq[License],
     versionControl: VersionControl,
-    developers: Seq[Developer],
-    @deprecated("Value will be ignored. Use PublishModule.pomPackagingType instead", "Mill 0.11.8")
-    packaging: String = PackagingType.Jar
+    developers: Seq[Developer]
 ) derives RW
+
+object PomSettings {
+  @deprecated(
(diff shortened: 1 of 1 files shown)


lihaoyi:
Not sure i understand this PR; what's the difference between `organization` and `organizationName`?

ShreckYe:
From [sbt's corresponding setting keys](https://github.com/sbt/sbt/blob/78bffa22b63cf6eb9539894faa192d85111167f1/main/src/main/scala/sbt/Keys.scala#L345-L347):

```scala
  val organization = settingKey[String]("Organization/group ID.").withRank(APlusSetting)
  val organizationName = settingKey[String]("Organization full/formal name.").withRank(BMinusSetting)
  val organizationHomepage = settingKey[Option[URL]]("Organization homepage.").withRank(BMinusSetting)
```

lefou:
https://maven.apache.org/pom.html#Organization

lihaoyi:
Closing due to inactivity

PR #4655, opened 2025-03-05
Add an `email` member to `Developer`

ShreckYe:
Similar to #4654. The `email` member is available in [Maven](https://maven.apache.org/pom.html#Developers), [Gradle](https://docs.gradle.org/current/dsl/org.gradle.api.publish.maven.MavenPomDeveloper.html#org.gradle.api.publish.maven.MavenPomDeveloper:email), and [sbt](https://github.com/sbt/librarymanagement/blob/9bcfbef2b7f4268fb48bbf6a1c17748f1e198eb1/core/src/main/contraband-scala/sbt/librarymanagement/Developer.scala). Shall we add it here too?

In sbt's official code, however, it's unclear to me why [they pass the (GitHub) ID to the `email` member](https://github.com/sbt/sbt/blob/3529e202799abc6a381a1f02cd527d7b8ff88034/build.sbt#L29-L40).

Compared to #4654, I don't think there is a way to make this change in a backward-compatible way with the default arguments though.

This is just a proposal so the changes are far from complete. Please let me know whether it's necessary.

### DIFF
--- a/scalalib/src/mill/scalalib/publish/settings.scala
+++ b/scalalib/src/mill/scalalib/publish/settings.scala
@@ -67,20 +67,6 @@ case class Developer(
     organizationUrl: Option[String] = None
 ) derives RW
 
-object Developer {
-  @deprecated(
-    "This method is for backward compatibility. Use the one with `email` instead."
-  )
-  def apply(
-      id: String,
-      name: String,
-      url: String,
-      organization: Option[String],
-      organizationUrl: Option[String]
-  ) =
-    new Developer(id, name, None, url, organization, organizationUrl)
-}
-
 case class PomSettings(
     description: String,
     organization: String,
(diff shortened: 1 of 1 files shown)


lefou (commented):
Instead, we should encode an email with the `mailto` scheme into an url.

ShreckYe:
So by "elegant" do you mean all 3 cases (email only, url only, both email and url) should be supported, they shouldn't be encoded with an extra sum type or `Either`, and the constraint should happen at compile time instead of runtime?

lefou:
> So by "elegant" do you mean all 3 cases (email only, url only, both email and url) should be supported, they shouldn't be encoded with an extra sum type or `Either`, and the constraint should happen at compile-time instead of runtime?

That's what I meant. But looking at https://maven.apache.org/pom.html#Developers, non of these fields are really required and it's up to the specific repository to accept or reject a pom.

In that case, we should also add all the other missing fields like `roles`, `timezone` and `properties`.

ShreckYe:
Got it. Thanks. Feel free to close this PR.

PR #4666, opened 2025-03-07
Fix multiline ANSI color preservation in PromptLogger

### LINKED ISSUE #4115: Multiline ANSI colors are not properly preserved in Mill PromptLogger (500USD Bounty)
---

---

I thought I fixed it in https://github.com/com-lihaoyi/mill/pull/3797, but it seems it still appears sometimes, albeit much less frequently.

Screenshot 2024-12-12 at 12 18 16 PM

The completion criteria for this is a consistent reproduction of the issue, along with a PR that demonstrates the issue fixed

onyedikachi-david:
# Fix Multiline ANSI Color Preservation in PromptLogger

This PR fixes the issue with multiline ANSI colors not being properly preserved in Mill's PromptLogger.

## Changes

1. Enhanced `LinePrefixOutputStream` to better handle ANSI color codes:
   - Added tracking of both start and end colors for each line
   - Implemented proper ANSI escape sequence parsing
   - Improved color state management across line boundaries
   - Added buffer handling for split ANSI sequences

The issue with multiline ANSI colors not being properly preserved stems from several factors in the `LinePrefixOutputStream` implementation:

1. **Color State Tracking Limitation**: The original implementation only tracked the end-of-line color state (`endOfLastLineColor`), which was insufficient for handling complex multiline color sequences. This meant that when a color sequence spanned multiple lines, the color state could be lost or incorrectly applied after line breaks.

2. **ANSI Sequence Handling**: ANSI escape sequences could be split across multiple writes, causing the color codes to be processed incorrectly. The original implementation didn't maintain the partial sequence state between writes.

3. **Line Prefix Interference**: When adding line prefixes, the color state wasn't being properly reapplied, causing color sequences to be interrupted or reset unintentionally.

## Steps to Reproduce
The issue can be reproduced using multiline colored output in Mill tasks:

1. Create a Mill task that outputs multiline text with ANSI colors

2. The output shows incorrect color preservation:
- Colors might be reset after line breaks
- Line prefixes might interrupt color sequences
- Split ANSI sequences might not be handled correctly

## Solution Explanation
The fix addresses these issues through several key improvements:

1. **Enhanced Color State Management**:
   - Added tracking of both start and end colors for each line (`currentLineStartColor` and `currentLineEndColor`)
   - This ensures proper color state preservation across line boundaries
   - Colors are correctly reapplied after line prefixes

2. **Proper ANSI Sequence Handling**:
   - Added `escapeSequenceBuffer` to properly accumulate and process ANSI sequences
   - Implemented proper state machine for ANSI sequence parsing
   - Handles split sequences correctly across multiple writes

## Why This Fix is Correct
1. **Complete State Tracking**: By tracking both start and end colors of each line, we maintain full context of color state transitions.

2. **ANSI Spec Compliance**: The implementation now properly handles ANSI escape sequences according to spec, including:
   - Proper sequence parsing (ESC [ ... m)
   - Maintaining state between partial writes
   - Preserving nested and complex color combinations

## Related Issues

Fixes #4115 - Multiline ANSI colors are not properly preserved in Mill PromptLogger

image

image

### DIFF
--- a/core/internal/src/mill/internal/LinePrefixOutputStream.scala
+++ b/core/internal/src/mill/internal/LinePrefixOutputStream.scala
@@ -61,7 +61,7 @@ private[mill] class LinePrefixOutputStream(
     if (inEscapeSequence) {
       escapeSequenceBuffer.append(b.toChar)
       // Check if this is the end of the escape sequence
-      if ((b >= 0x40 && b <= 0x7E) && b != '[') {
+      if ((b >= 0x40 && b <= 0x7e) && b != '[') {
         inEscapeSequence = false
         val sequence = escapeSequenceBuffer.toString
         // Process the complete escape sequence
@@ -90,7 +90,7 @@ private[mill] class LinePrefixOutputStream(
     while (i < max) {
       writeLinePrefixIfNecessary()
       processEscapeSequence(b(i))
-      
+
       if (b(i) == '\n') {
         i += 1 // +1 to include the newline
         buffer.write(b, start, i - start)
--- a/core/internal/src/mill/internal/LinePrefixOutputStream.scala
+++ b/core/internal/src/mill/internal/LinePrefixOutputStream.scala
@@ -21,17 +21,21 @@ private[mill] class LinePrefixOutputStream(
   private var isNewLine = true
   val buffer = new ByteArrayOutputStream()
 
-  // Make sure we preserve the end-of-line ANSI colors every time we write out the buffer, and
-  // re-apply them after every line prefix. This helps ensure the line prefix color/resets does
-  // not muck up the rendering of color sequences that affect multiple lines in the terminal
-  private var endOfLastLineColor: Long = 0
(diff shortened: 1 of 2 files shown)


lihaoyi:
@onyedikachi-david can you add to the PR description an explanation of what the root cause of the problem was, what steps you took to reproduce the issue, and why your fix is the correct one

onyedikachi-david:
done @lihaoyi

lihaoyi:
I don't understand your explanation; why does tracking the color at the start of each line matter?

Your reproduction steps are not a repro; please give a proper repro that I can use to verify that your fix works

lihaoyi:
While the problem you describe sounds reasonable, it doesn't seem like it would apply to the problem screenshotted in the linked ticket, where there were no changes in the ansi colors and so no ansi codes being emitted in at the point in which the colors were lost

onyedikachi-david:
I'll update accordingly when I am able to reproduce what you have in the issue, could you please point me to how best to reproduce that.

Issue #3571, opened 2024-09-18
Turn `mill.api.Ctx` into a `trait`


lihaoyi:
We shouldn't need to expose the constructor to the end user, but this would be a bincompat breaking change

PR #4670, opened 2025-03-08
Make `mill.api.Ctx` a `trait`

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/3571

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -1,3 +1,4 @@
+//> mill-version =
 package build
 // imports
 import com.github.lolgab.mill.mima.Mima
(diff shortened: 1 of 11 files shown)


build.mill
```diff
@@ -1,3 +1,4 @@
+//> mill-version =
```
lefou, line 1: @lihaoyi Was this line addition intentional?

lihaoyi: @lefou oh probably not, must have committed it accidentally. Let me remove it

PR #4671, opened 2025-03-08
Update Scala from 3.6.2 to 3.6.4

lefou:
This most likely also requires some plugin bumps. I'm on it...

* https://github.com/com-lihaoyi/mill-moduledefs/pull/23
* https://github.com/com-lihaoyi/acyclic/pull/156


lefou:
The tests work on my Linux machine but not on the GHA CI Linux machines. There are differences in `.idea/modules/mill-module.iml` files. Since we use coursier to assemble the classpaths, we should get same results. 

I think I need to make these tests more relaxed, since this could be the result of different external resources (Maven repos), I opened https://github.com/com-lihaoyi/mill/issues/4978.

Issue #4674, opened 2025-03-08
.mill-version cannot be read on Windows in Spanish


joshuakfarrar:
It appears there's an encoding issue on Windows 11, in Spanish, where the .mill-version file is not read properly by Java.

```shell
PS C:\Users\joshua> $filePath = "C:\Users\joshua\g\test\.mill-version"
PS C:\Users\joshua> $stream = [System.IO.StreamReader]::new($filePath, [System.Text.Encoding]::Default)
PS C:\Users\joshua> $encoding = $stream.CurrentEncoding
PS C:\Users\joshua> $encoding


IsSingleByte      : True
BodyName          : iso-8859-1
EncodingName      : Europeo occidental (Windows)
HeaderName        : Windows-1252
WebName           : Windows-1252
WindowsCodePage   : 1252
IsBrowserDisplay  : True
IsBrowserSave     : True
IsMailNewsDisplay : True
IsMailNewsSave    : True
EncoderFallback   : System.Text.InternalEncoderBestFitFallback
DecoderFallback   : System.Text.InternalDecoderBestFitFallback
IsReadOnly        : True
CodePage          : 1252
```

https://github.com/com-lihaoyi/mill/blob/e009563f99e9189fd4c864e1949f2203e5b4b5cc/runner/src/mill/runner/MillMain.scala#L410

I believe the solution is to use StandardCharsets.UTF_16 to convert the raw file bytes into a String, e.g.:

```scala
def checkMillVersionFromFile(projectDir: os.Path, stderr: PrintStream): Unit = {
  Seq(
    projectDir / ".config/mill-version",
    projectDir / ".mill-version"
  ).collectFirst {
    case f if os.exists(f) => {
      val rawBytes = os.read.bytes(f).toSeq
      val decodedLines = new String(rawBytes.toArray, StandardCharsets.UTF_16)
        .split(System.lineSeparator())
        .toSeq
      (f, decodedLines.find(l => l.trim().nonEmpty).get)
    }
  }.foreach { case (file, version) =>
    if (BuildInfo.millVersion != version.stripSuffix("-native")) {
      val msg =
        s"""Mill version ${BuildInfo.millVersion} is different than configured for this directory!
           |Configured version is ${version} (${file})""".stripMargin
      stderr.println(
        msg
      )
    }
  }
}
```

lefou:
Who created that file? I think we want stick to UTF-8.

joshuakfarrar:
[...]

PR #4675, opened 2025-03-08
use UTF_16 for correct decoding of .mill-version on Windows 11

joshuakfarrar:
See #4674.

### DIFF
--- a/runner/src/mill/runner/MillMain.scala
+++ b/runner/src/mill/runner/MillMain.scala
@@ -406,8 +406,13 @@ object MillMain {
       projectDir / ".config/mill-version",
       projectDir / ".mill-version"
     ).collectFirst {
-      case f if os.exists(f) =>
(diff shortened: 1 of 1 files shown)


joshuakfarrar:
Per the comment thread and further diagnosis on #4674, the issue is that PowerShell 5.1 specifically defaults to UTF-16LE. Either the documentation should be updated to reflect specific problems with PowerShell 5.1, which ships with Windows 10 and 11 by default, or this code should be updated to attempt to assess the user's platform.

lefou:
I'll close this, as I think we don't want to support UTF-16 for `.mill-version`, just because there is no reason we have to. 

In general, If we're going to be smart about file encoding, we should peek the first bytes of the file and check if it's a [byte-order marker](https://en.wikipedia.org/wiki/Byte_order_mark), and then switch to the correct encoding. But it's IMHO over-engineering for a simple version file, esp. for one which we specified ourselves.

PR #4690, opened 2025-03-11
Remove unused files

alexarchambault:
Unless I'm mistaken, it seems these files are unused (and they make scalafmt crash no matter the dialect, so they're somewhat harmful too)


alexarchambault:
Wrong base branch

PR #4693, opened 2025-03-11
Format more files

alexarchambault:
This reformats the whole codebase with the changes of https://github.com/com-lihaoyi/mill/pull/3511.

Only files that were not formatted by `./mill mill.scalalib.scalafmt.ScalafmtModule/` actually need formatting here.


lefou (approved):
Looks good to me.

After being merged, the commit should be added to `.git-blame-ignore-revs`.

PR #4697, opened 2025-03-11
Add --doc, --sources, and --transitive options to publishLocal

alexarchambault:
This is a continuation of https://github.com/com-lihaoyi/mill/pull/3512. In a new PR, against `0.12.x`, as I'm currently having issues running tests on the `main` branch.


lefou (commented):
Looks good to me.

lihaoyi:
This looks reasonable, could you update the docs with a one-liner about the use case? I assume this is to speed up performance in the common case you don't care about local scaladoc/javadoc?

Once this has landed, we should probably wire up `dist.installLocal` and other local dev workflows to disable local doc generation

lihaoyi (approved):

PR #4698, opened 2025-03-11
Optimize BSP initialization stuff

alexarchambault:
This optimizes the `workspaceBuildTargets` BSP request. On my machine, this makes it go from around 15s to about 3s on the Mill repo. Which is time saved when opening or reloading Mill in one's IDE. Note that gains are more modest on smaller code bases, but still there.

In more detail, two things are optimized:
- the `MillBuildServer.groupList` method, whose first implementation (that I wrote) was too naive (don't know how the code it replaced at the time fared)
- it reduces the number of calls to `Evaluator.evaluate` when the BSP server needs to evaluate tasks. It does so by really grouping them by evaluator, rather than by (evaluator, BuildTargetId). Seems the former grouping was intended, but the grouping was actually more loose.

Lastly, not an optimization per se, but this makes the BSP server less verbose. It was printing some very large objects in particular, that tend to flood the output / BSP server log. These messages are now only printed if the env var `MILL_BSP_DEBUG` is `true`.

(Opened against `0.12.x` first, as this branch is the easiest to test with existing Mill projects.)


alexarchambault:
@lihaoyi Just did so

lihaoyi (approved):
Looks good to me

joan38:
Thanks @alexarchambault. We need more focus on the IDE/BSP sync. IDE setup is the first cause of issues people hit in my experience.

PR #4701, opened 2025-03-12
Configure thread pool executor to run in LIFO order to minimize blocked task count

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/4611

We basically reverse the execution order of tasks submitted to the `mill.exec.ExecutionContexts.ThreadPool` executor to be LIFO order instead of FIFO order. Since child tasks are always spawned after parent tasks, this prioritizes scheduling child tasks (which can actually run and do work) over parent tasks (which are sometimes blocked waiting for child tasks to be scheduled, run, and complete). 

There's no `LinkedBlockingStack` equivalent of `LinkedBlockingQueue`, but `LinkedBlockingDeque` can serve the same purpose by forwarding `poll` and `take` to `pollLast` and `takeLast`. `ThreadPoolExecutor` only seems to call `size`, `isEmpty`, `remove`, `poll`, `take`, `offer` on it's `workQueue` reference, so we don't need to worry about forwarding other methods.

This LIFO order is similar to the approach that `ForkJoinPool` and other schedulers that prioritize child tasks take.

Ideally we would use a `PriorityBlockingQueue` to properly prioritize all child tasks to run before all parent tasks, but because of our usage of Scala `Future`s tasks end up being wrapped and it gets hard to distinguish exactly where the `Runnable`s submitted to the queue actually come from. Using a LIFO is a crude approximation of the prioritization we actually want, but has the benefit of working without any knowledge of the individual tasks that are submitted, with the downside that later-scheduled parent tasks can take priority over earlier-scheduled child tasks spawned from an even-earlier parent.

Tested locally via `./mill-assembly.jar -i integration.__.local.server`. Without this patch, we see large numbers of tasks buffering up in the prompt. with this PR, it seems to be limited to 5. Also tweaked some of the `test-grouping-parallel` tests


core/exec/src/mill/exec/ExecutionContexts.scala
```diff
@@ -38,7 +38,15 @@ private object ExecutionContexts {
...
       TimeUnit.SECONDS,
-      new LinkedBlockingQueue[Runnable]()
+      // Use a `Deque` rather than a normal `Queue`, with the various `poll`/`take`
+      // operations reversed, providing elements in a LIFO order. This ensures that
+      // child `fork.async` tasks always take priority over parent tasks, avoiding
+      // large numbers of blocked parent tasks from piling up
+      new LinkedBlockingDeque[Runnable]() {
+        override def poll(): Runnable = super.pollLast()
+        override def poll(timeout: Long, unit: TimeUnit): Runnable = super.pollLast(timeout, unit)
+        override def take(): Runnable = super.takeLast()
```
He-Pin, line 48: maybe we can use a forkjoin pool instead, which can configure the queue with fifo or lifo with `async` parameter.

lihaoyi: Forkjoin has weird behavior where `Await` is re-entrant, we used to use it and had to swap it out

PR #4706, opened 2025-03-12
Replace uppercase "SBT" with lowercase "sbt"

ShreckYe:
The lowercase spelling "sbt" seems to be the official spelling used in their docs and other places. Shall we replace the uppercase "SBT"s with this instead?

This is done for all occurences of "SBT" except for the link http://www.lihaoyi.com/post/SowhatswrongwithSBT.html which will be broken if updated.


integration/ide/gen-idea/resources/extended/idea/libraries/sbt_ junit_junit_3_6_4_13_2_jar.xml
```diff
@@ -1,5 +1,5 @@
 <component name="libraryTable">
-    <library name="SBT: junit:junit_3.6:4.13.2:jar">
+    <library name="sbt: junit:junit_3.6:4.13.2:jar">
```
lefou, line 2: Don't know if those still have any special meaning, but I'd say let's keep the upper case versions in those technical use places.

integration/ide/gen-idea/resources/extended/idea/libraries/sbt_ org_scalameta_munit_3_3_6_0_7_29_jar.xml
```diff
@@ -1,5 +1,5 @@
 <component name="libraryTable">
-    <library name="SBT: org.scalameta:munit_3_3.6:0.7.29:jar">
+    <library name="sbt: org.scalameta:munit_3_3.6:0.7.29:jar">
```
lefou, line 2: Revert.

idea/src/mill/idea/GenIdeaImpl.scala
```diff
@@ -386,7 +386,7 @@ class GenIdeaImpl(
      *
      * @see [[https://github.com/JetBrains/intellij-scala/blob/idea223.x/scala/worksheet/src/org/jetbrains/plugins/scala/worksheet/ammonite/AmmoniteUtil.scala#L240]]
      * @example {{{
-     *   //SBT: com.lihaoyi:ammonite-ops_2.13:2.2.0:jar
+     *   //sbt: com.lihaoyi:ammonite-ops_2.13:2.2.0:jar
```
lefou, line 389: revert

idea/src/mill/idea/GenIdeaImpl.scala
```diff
@@ -404,7 +404,7 @@ class GenIdeaImpl(
           // Default to the scala binary version used by mill itself
           s"${artifactId}_${BuildInfo.scalaVersion.split("[.]").take(2).mkString(".")}"
       }
-      s"SBT: ${pom.module.organization.value}:$artifactWithScalaVersion:${pom.version}:jar"
+      s"sbt: ${pom.module.organization.value}:$artifactWithScalaVersion:${pom.version}:jar"
```
lefou, line 407: revert

website/docs/modules/ROOT/pages/comparisons/sbt.adoc
```diff
@@ -1,46 +1,46 @@
-= Case Study: Mill vs SBT
-:page-aliases: Case_Study_Mill_vs_SBT.adoc
+= Case Study: Mill vs sbt
+:page-aliases: Case_Study_Mill_vs_sbt.adoc
```
lefou, line 2: We probably should keep the old link in addition, to not break external links.

website/docs/modules/ROOT/pages/depth/design-principles.adoc
```diff
@@ -367,11 +367,11 @@ reduced to simply instantiating a `trait` over and over, with tweaks.
...
 == Prior Work
 
-=== SBT
+=== sbt
 
-Mill is built as a substitute for SBT, whose problems are
-http://www.lihaoyi.com/post/SowhatswrongwithSBT.html[described here].
-Nevertheless, Mill takes on some parts of SBT (builds written in Scala, Task
+Mill is built as a substitute for sbt, whose problems are
+http://www.lihaoyi.com/post/Sowhatswrongwithsbt.html[described here].
```
lefou, line 373: This looks like we should revert?

lefou (commented):
Looks good. I commented in some places, where I think we should keep the upper case variant.

website/docs/modules/ROOT/pages/comparisons/sbt.adoc
lefou, line 2:
```suggestion
:page-aliases: Case_Study_Mill_vs_sbt.adoc, Case_Study_Mill_vs_SBT.adoc
```

website/docs/modules/ROOT/pages/depth/design-principles.adoc
ShreckYe: True. I made a mistake here.

ShreckYe: I just noticed that this link is still an `http` link. Shall we make it `https`?

lihaoyi:
TBH I find the lowercase `sbt` very hard to parse when reading. I think it's better to leave it uppercase, even if that might conflict with whatever upstream is doing. `SBT` is unambiguous in what it means and how it is meant to be pronounced.

lefou:
I think I multiple time saw in the sbt channels, that `sbt` is the correct spelling. I personally find it important to respect the correct names, and if sbt folks stated that it's not called `SBT`, I'd rather like to accept and apply it.

What do you think of writing of always writing it in `monospace` font as we do for app names: "`sbt`" instead of "sbt"? That should be easier to parse.

lihaoyi:
monospace helps, I think that's a good way of working around the visual parsing problem

lefou (changes requested):
@ShreckYe Can you replace all mentions of 

```
sbt
``` 

by 

```
`sbt`
```

ShreckYe:
OK. Shall I still revert the changes such as `SBT: junit:junit_3.6:4.13.2:jar`? As commented in https://github.com/com-lihaoyi/mill/pull/4706#discussion_r1991657695 I found that the tests pass as long as I replace them in the filenames.

lefou:
> OK. Shall I still revert the changes such as `SBT: junit:junit_3.6:4.13.2:jar` -> `sbt: junit:junit_3.6:4.13.2:jar`? As commented in [#4706 (comment)](https://github.com/com-lihaoyi/mill/pull/4706#discussion_r1991657695) I found that the tests pass as long as I replace them in the filenames.

Yes. These are conventions used by `sbt` and IntelliJ. They might be outdated, still we should refer to what we used to use. We don't have integration tests for it.

ShreckYe:
I see some inconsistencies for the words  "Hot sbt" and "Cold sbt" in "sbt.adoc". Which one shall I go with:

1. `Hot sbt`
1. "Hot sbt"
1. "Hot `sbt`"

lefou:
> I have done this for the Scaladoc and comments in code. Shall I replace those in program output strings too? They will not be rendered in the terminal and users will just see backticks around.

No, the backticks is just needed for rending AsciiDoc.

main/init/sbt/src/mill/main/sbt/SbtBuildGenMain.scala
```diff
@@ -108,13 +108,12 @@ object SbtBuildGenMain
...
          */
         val sbtFile = writeTempSbtFileInSbtProjectDirectory(workspace)
-        val commandResult = os.call(
+        os.call(
           (sbtExecutable, "millInitExportBuild"),
           cwd = workspace,
           stdout = os.Inherit
         )
         os.remove(sbtFile)
-        commandResult
```
lefou, line 117: This looks wrong. We now no longer return the result of `os.call` but `os.remove`.

website/docs/modules/ROOT/pages/depth/design-principles.adoc
lefou: It it works, yes.

main/init/sbt/src/mill/main/sbt/SbtBuildGenMain.scala
ShreckYe: IntelliJ IDEA warned me that this variable is not used. I think the return value is not needed here. The variable was added when I first used the `scala.sys.process` APIs and had to judge whether the command was successful from the exit code, but later I refactored this part to use OS-Lib and this is no longer needed.

ShreckYe:
> I see some inconsistencies for the words "Hot sbt" and "Cold sbt" in "sbt.adoc". Which one shall I go with:
> 
> 1. `Hot sbt`
> 2. "Hot sbt"
> 3. Hot `sbt`
> 4. "Hot `sbt`"

How about this question? Which one shall I choose here?

main/init/sbt/src/mill/main/sbt/SbtBuildGenMain.scala
lefou: If the return value isn't supposed to transport semantics, let's explicitly make it `: Unit`.

lefou: I see.

lefou:
> I see some inconsistencies for the words "Hot sbt" and "Cold sbt" in "sbt.adoc". Which one shall I go with:

Hot `sbt`

lefou (approved):
Looks good to me. Thanks!

PR #4724, opened 2025-03-14
Optimize BSP initialization stuff (forwardport #4698)

alexarchambault:
Report of https://github.com/com-lihaoyi/mill/pull/4698 in `main`

---

This optimizes the `workspaceBuildTargets` BSP request. On my machine, this makes it go from around 15s to about 3s on the Mill repo. Which is time saved when opening or reloading Mill in one's IDE. Note that gains are more modest on smaller code bases, but still there.

In more detail, two things are optimized:
- the `MillBuildServer.groupList` method, whose first implementation (that I wrote) was too naive (don't know how the code it replaced at the time fared)
- it reduces the number of calls to `Evaluator.evaluate` when the BSP server needs to evaluate tasks. It does so by really grouping them by evaluator, rather than by (evaluator, BuildTargetId). Seems the former grouping was intended, but the grouping was actually more loose.

Lastly, not an optimization per se, but this makes the BSP server less verbose. It was printing some very large objects in particular, that tend to flood the output / BSP server log. These messages are now only printed if the env var `MILL_BSP_DEBUG` is `true`.


alexarchambault:
@lihaoyi @lefou Feel free to have a look if you wish. Given this is a report (and the CI is green), maybe this doesn't need a review / approval?

lihaoyi:
@alexarchambault go ahead and merge it, unless there's something surprising in the forwardport we can skip the review

PR #4727, opened 2025-03-14
Prefer `defaultResolver().classpath` over `Lib.resolveDependencies`

alexarchambault:
Whenever possible, this tries to make modules defined in Mill use `CoursierModule#defaultResolver` or `CoursierModule#millResolver`  to resolve dependencies, rather than calling lower level helpers like `Lib. resolveDependencies`. The former automatically take into account customizations of resolution defined on `CoursierModule` and `JavaModule` (`repositoriesTask`, `mapDependencies`, etc.), while these have to be passed manually to `Lib.resolveDependencies`.

This helps when adding more parameters - these have to be passed to `defaultResolver` / `millResolver`, and that's it, rather than also having to inspect all calls to `Lib.resolveDependencies`, and passing the new parameter there too.

This also tries to make resolution log to the task logger of the task requesting the resolution, rather than the logger of `CoursierModule#{defaultResolver,millResolver}`.


lefou (commented):
Looks good to me. But we need to introduce the new API in Mill `0.12` too and mark the old as deprecated. Otherwise migration can become hard.

alexarchambault:
(Not sure what's going on with the CI right now. Seems the Mill CI assumes this PR is a draft, while it's not anymore, and cancelling [this former CI run](https://github.com/alexarchambault/mill/actions/runs/14089463674/job/39462677466) in my fork CI has no effect, blocking CI runs of the newer commits)

alexarchambault:
Actual CI on my fork [here](https://github.com/alexarchambault/mill/actions/runs/14090121566)

alexarchambault:
The `0.12.x` version of this PR is [here](https://github.com/com-lihaoyi/mill/pull/4784). Unlike here, it has to maintain binary compatibility. Because of that, it keeps passing `Ctx.Log` around rather than just `Ctx`, as we can only provide the former in deprecated / bin-compat methods (and whose implementations call the new methods).

lefou:
@alexarchambault I think this is good to merge. Since I also touch some of the same locations in concurrent PRs, we should merge soon (after resolving the conflicts) to avoid piling-up merge conflicts.

PR #4728, opened 2025-03-14
[0.12] Add `@deprecated` annotation to bin-compat shims

lolgab:


lihaoyi:
IIRC we didn't do this universally because sometimes Scala overload resolution would pick the shim rather than picking the method we wanted, resulting in people getting spurious deprecation warnings.

Issue #4109, opened 2024-12-11
Fine-grained selective testing at a class-level granularity


lihaoyi:
https://github.com/com-lihaoyi/mill/pull/4091 implements selective execution and testing at a granularity of mill build modules, but we should be able to do better. 

There are two main implementation strategies we could use

## Method Code Hash Signature Bytecode Analysis

* We already have the ability to compute code signatures for JVM bytecode and do basic callgraph analysis, which we currently use for selectively invalidating targets in the `build.mill` file. 

* We should be able to apply the same bytecode analysis to classes and methods in the `JavaModule ` and `ScalaModule` compile output, automatically determining which test classes within a module could potentially be affected by a code change and would need to be executed to validate it

This would provide much finer grained selective testing than the initial module-level implementation, and help greatly in messy real-world projects where the module structure is not as fine grained as it could be. It should also work for Java 
and Kotlin-JVM modules, since the analysis happens on JVM bytecode which is universal across JVM languages

The implement strategy would be:

- Add a new task `JavaModule#methodCodeHashSignatures`, that runs the same `mill.core.codesig` logic we currently run on `MillBuildRootModule`, but makes the signatures available for any `JavaModule`.

- Add a new command `TestModule#testQuick`, that compares the previous and current `methodCodeHashSignatures` (maybe using a persistent task to store the old value), and selects the test classes to run based on those who have a method in `methodCodeHashSignatures` that changed

- The first time `testQuick` is run it would run all the tests, but if you change code subsequent runs of `testQuick` would only run test classes that either (1) failed previously or (2) were affected by the code changes, via the hash signature computation above

## Zinc Incremental Compilation Dependency Graph

* We already have analysis files that the Zinc incremental compiler generates, which contains file-by-file dependency information in the local codebase. We can use this dependency graph similarly as above, and find the test classes that were affected by any code change

HollandDM:
I'm currently trying to implement the first approach. The implementation is basically what you've highlighted: create `methodCodeHashSignatures` for JavaModule, and move from there. I did some adjustment to call graph analyst to let it splits out changed classes only. This data will be used to select which test should run.
One thing I've noticed while testing is that: I've updated `TestRunnerTestUtils.testrunnerWorkStealing` by update the `enableParallelism` from `true` to `false`. The `spanningInvalidationTree.json` is as follow:
```json
{
  "def mill.scalalib.TestRunnerTestUtils$testrunnerWorkStealing$#enableParallelism()boolean": {
[...]

PR #4731, opened 2025-03-15
Add `testQuick` command

HollandDM:
Add `testQuick` command that leverage `CallGraphAnalysis`.

This pull request introduces the `testQuick` command, which utilizes `CallGraphAnalysis`. The following changes have been implemented:
- The outputs of `CallGraphAnalysis` now include `invalidClassNames`, which contains the names of test classes extracted from the `spanningInvalidationTree`. This list indicates the classes that are affected by code changes.
- A `callGraphAnalysis` task has been added to `JavaModule`. This is a persistent task that, when executed, generates a `callGraphAnalysis` folder containing data related to the `CallGraphAnalysis` feature.
- The `codeSignatures` of `MillBuildRootModule` have been updated to incorporate the results from `callGraphAnalysis`.
- The `testQuick` command has been added to `JavaTests`. It uses the `invalidClassNames` output from `callGraphAnalysis` and a dedicated log file, `quickTestFailedClasses.log`, which stores the failed test classes from previous runs. `testQuick` combines the information from these two sources to determine the set of test classes to run.

/claim #4109

One limitation of the current implementation is after the first time `testQuick` is run (test all classes), `callGraphAnalysis` will only create the `current` folder. If we run `testQuick` again, because `callGraphAnalysis`'s inputs haven't changed, it will not run provide `previous` folder. In that case `testQuick` will not have enough information and have to run all the test classes again.


scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -1380,6 +1472,90 @@ trait JavaModule
...
+                // In Scala 3, lambdas are implemented by private instance methods,
+                // not static methods, so they fall through the crack of "isSimpleTarget".
+                // Here make the assumption that a zero-arg lambda called from a simpleTarget,
+                // should in fact be tracked. e.g. see `integration.invalidation[codesig-hello]`,
+                // where the body of the `def foo` target is a zero-arg lambda i.e. the argument
+                // of `Cacher.cachedTarget`.
+                // To be more precise I think ideally we should capture more information in the signature
+                isSimpleTarget(callSiteSig.desc) && calledSig.name.contains("$anonfun")
+              )
+            }
```
lihaoyi, line 1521: All the other special cases except this one are heuristics unique to `MillBuildRootModule`, and so probably should live there. We can have a `def codeSigIgnoreCall` method that gets fed into `CodeSig.compute`, with a default implementation for `JavaModule` and a more detailed override for `MillBuildRootModule`

lihaoyi:
> One limitation of the current implementation is after the first time testQuick is run (test all classes), callGraphAnalysis will only create the current folder. If we run testQuick again, because callGraphAnalysis's inputs haven't changed, it will not run provide previous folder. In that case testQuick will not have enough information and have to run all the test classes again.

I don't understand the problem are describing here, could you go into more detail

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -92,6 +95,95 @@ trait JavaModule
           case _: ClassNotFoundException => // if we can't find the classes, we certainly are not in a ScalaJSModule
         }
     }
+
+    private def quickTest(args: Seq[String]): Task[(String, Seq[TestResult])] =
+      Task(persistent = true) {
```
lihaoyi, line 100: `Task(persistent = true)`  is probably the wrong thing here, since `def quickTest` takes arguments. We should add a `Command(persistent = true)` flag and keep it as a command similar to the other test tasks

lihaoyi:
> The outputs of CallGraphAnalysis now include invalidClassNames, which contains the names of test classes extracted from the spanningInvalidationTree. This list indicates the classes that are affected by code changes.

This seems like a post-processing step, could we put it in a separate method that takes the output of `spanningInvalidationTree` and parses out the parts we need?

HollandDM:
> > One limitation of the current implementation is after the first time testQuick is run (test all classes), callGraphAnalysis will only create the current folder. If we run testQuick again, because callGraphAnalysis's inputs haven't changed, it will not run provide previous folder. In that case testQuick will not have enough information and have to run all the test classes again.
> 
> I don't understand the problem are describing here, could you go into more detail

Here is the reproduce steps:
- `rm -rf out` to clean the output folder.
- run `scalalib.callGraphAnalysis` for the first time.
- run `scalalib.callGraphAnalysis` again.

In this case, the first `callGraphAnalysis` will run normally and produce `current` folder, but  because we din't modify any source code, the second `callGraphAnalysis` will do nothing (I guess because it's input, `compile()` doesn't change). There for `previous` folder will not be produced.
The problem is that this PR will check for `previous` folder, and it will consider `testQuick` to be a first time command when it doesn't find one. So with this case especially, invoke `testQuick` consecutively will run all test cases like normal `testForked`.

=================================

> > The outputs of CallGraphAnalysis now include invalidClassNames, which contains the names of test classes extracted from the spanningInvalidationTree. This list indicates the classes that are affected by code changes.
> 
> This seems like a post-processing step, could we put it in a separate method that takes the output of `spanningInvalidationTree` and parses out the parts we need? Or could it be a different helper method that works directly on `prevTransitiveCallGraphHashes` and `transitiveCallGraphHashes0`, so we don't need to construct a `spanningInvalidationTree` at all?

I also consider post process this at first, but the output of `spanningInvalidationTree` is too easy to work with, so I went with this approach. I will update this into a separate post-processing step

lihaoyi:
@HollandDM the problem with `callGraphAnalysis` caching should be fixed by making `testQuick` a command. That way it will run every time regardless of changes, and then can decide what to do based on whether or not tue upstream tasks changed

lihaoyi:
> I also consider post process this at first, but the output of pre-jsonifiedspanningInvalidationTree is too easy to work with, so I went with this approach. I will update this into a separate post-processing step

Let's split it into three methods then: one shared one, one that uses the shared helper to generate the tree, and one that uses the shared helper to generate a list of affected classes

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -92,6 +94,117 @@ trait JavaModule
...
+        // we cannot parse the json back to its concrete type, we only know that all of them are
+        // encoded with class name as prefix.
+        // so we can do a prefix checking to get the affected class.
+        // naive implementation won't work well as it has loop through all `testClasses` and check prefix
+        // for all of them, so we employ a prefix trie to speed up the prefix checking.
+
+        val jsonValueQueue = mutable.ArrayDeque[(String, ujson.Value)]()
+        val prefixTrie = new PrefixTrie[String]()
+        jsonValueQueue.appendAll(spanningInvalidationTreeObj.value)
+        while (jsonValueQueue.nonEmpty) {
```
HollandDM, line 120: > > I also consider post process this at first, but the output of pre-jsonifiedspanningInvalidationTree is too easy to work with, so I went with this approach. I will update this into a separate post-processing step
> 
> Let's split it into three methods then: one shared one, one that uses the shared helper to generate the tree, and one that uses the shared helper to generate a list of affected classes

I went with this approach first, as I think leaving `callGraphAnalysis` folder as it is right now is quite good. This pre-processing seem quick enough when I run `scalalib.test.testQuick`, so I think it is fine. WDYT?

core/define/src/mill/define/Task.scala
```diff
@@ -121,7 +121,16 @@ object Task extends TaskBase {
...
       inline w: W[T],
       inline ctx: mill.define.Ctx
-  ): Command[T] = ${ TaskMacros.commandImpl[T]('t)('w, 'ctx, exclusive = '{ false }) }
+  ): Command[T] = ${ TaskMacros.commandImpl[T]('t)('w, 'ctx, exclusive = '{ false }, persistent = '{ false }) }
+
+
+  /**
+   * This version allow [[Command]] to be persistent
+   */
+  inline def Command[T](inline persistent: Boolean)(inline t: Result[T])(implicit
```
HollandDM, line 130: add a Command with `persistent` option.

scalalib/src/mill/scalalib/JavaModule.scala
lihaoyi: Looking at the code, I think splitting it into three methods is better. The various string operations necessary to do the post-processing look pretty fragile, and while it no doubt works, it probably will be harder to maintain and refactor in future compared to working with typed data structures

lihaoyi:
I think the approach looks reasonable. Questions I have, more on the semantics and expected behavior rather than the code:

1. I assume you tested that this works for changes in the test code, but does this work for code changes in the non-test module code? Or upstream modules? The callgraph analysis we do for `MillBuildRootModule` is focused on one module, though it should in theory work for multi-module codebases too if given all the necessary classfiles

2. What kind of performance cost does this incur when used on various `.test` modules in the Mill codebase? You can run `./mill dist.installLocal && ci/patch-mill-bootstrap.sh` to get an executable you can try out locally

3. Are there any major edge cases that we should be aware of that this doesn't work well?

Eventually all this should probably make its way into a unit/integration test suite, but for now manual testing is fine just to figure out what the boundaries and limitations of this featue

HollandDM:
For question 1, it seems that changes from upstream modules does not trigger the test. The main reason is that call graph analysis does not compute necessary data for it

lihaoyi:
The callgraph analysis should be able to handle cross-module analysis if all the classfiles are available on disk; you should be able to aggregate the `transitiveLocalClasspath` and pass it to CodeSig and it should just work (in theory)

HollandDM:
I've update the code to include all upstream module while running the `testQuick` command. Also clean up some shared logic.
I've also run some local test, e.g: if I update `UnitTester.scala`, then run `scalalib.test.testQuick`, all test cases that use `UnitTester` will re-run the test. This ensure cross module is working.
But for some reason, if I update both `UnitTester` and `TestRunnerTestUtils`, only changes in `TestRunnerTestUtils` are discovered. I will try to debug this

core/codesig/src/mill/codesig/ExternalSummary.scala
```diff
@@ -47,7 +48,8 @@ object ExternalSummary {
 
     def load(cls: JCls): Unit = methodsPerCls.getOrElse(cls, load0(cls))
 
-    def load0(cls: JCls): Unit = {
+    // Some macros implementations will fail the ClassReader, we can skip them
```
lihaoyi, line 51: What kind of error do those macro implementations produce? We should try and be specific about what errors we catch here, to avoid silencing unexpected errors that may indicate real issues

core/codesig/src/mill/codesig/ReachabilityAnalysis.scala
```diff
@@ -78,44 +73,45 @@ class CallGraphAnalysis(
...
+          transitiveCallGraphHashes0,
+          indexToNodes,
+          indexGraphEdges
+        )
+      case None => ujson.Obj()
+    }
   }
 
-  logger.mandatoryLog(spanningInvalidationTree)
+  def calculateInvalidClassName(
```
lihaoyi, line 91: Probably should call this `calculateInvalidatedClassName` to avoid confusion

core/codesig/src/mill/codesig/CodeSig.scala
```diff
@@ -3,31 +3,63 @@ package mill.codesig
...
+    implicit val st: SymbolTable = new SymbolTable()
+
+    val callAnalysis = callGraphAnalysis(classFiles, upstreamClasspath, ignoreCall)
+
+    logger.log(callAnalysis.localSummary)
+    logger.log(callAnalysis.externalSummary)
+    logger.log(callAnalysis.resolved)
+
+    logger.mandatoryLog(callAnalysis.methodCodeHashes)
+    logger.mandatoryLog(callAnalysis.prettyCallGraph)
```
lihaoyi, line 52: Are these two `mandatoryLog`s necessary? It seems we only use `transitiveCallGraphHashes` and `spanningInvalidationTree`, at least as far as I can tell

core/codesig/src/mill/codesig/ReachabilityAnalysis.scala
```diff
@@ -135,12 +131,75 @@ object CallGraphAnalysis {
...
+        indexGraphEdges
+      ),
       k => indexToNodes(k).toString
     )
   }
 
+  /**
+   * Get all class names that have their hashcode changed compared to prevTransitiveCallGraphHashes
+   */
+  def invalidClassNames(
```
lihaoyi, line 170: `invalidatedClassNames`

runner/src/mill/runner/MillBuildRootModule.scala
```diff
@@ -138,74 +140,45 @@ abstract class MillBuildRootModule()(implicit
...
+
+    // Skip calls to `millDiscover`. `millDiscover` is bundled as part of `RootModule` for
+    // convenience, but it should really never be called by any normal Mill module/task code,
+    // and is only used by downstream code in `mill.eval`/`mill.resolve`. Thus although CodeSig's
+    // conservative analysis considers potential calls from `build_.package_$#<init>` to
+    // `millDiscover()`, we can safely ignore that possibility
+    def isMillDiscover =
+      calledSig.name == "millDiscover$lzyINIT1" ||
+        calledSig.name == "millDiscover" ||
+        callSiteOpt.exists(_.sig.name == "millDiscover")
```
lihaoyi, line 164: Can these be moved into `callGraphAnalysisIgnoreCalls` as well?

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -1380,6 +1490,60 @@ trait JavaModule
...
+  ): Boolean = {
+    // We can ignore all calls to methods that look like Targets when traversing
+    // the call graph. We can do this because we assume `def` Targets are pure,
+    // and so any changes in their behavior will be picked up by the runtime build
+    // graph evaluator without needing to be accounted for in the post-compile
+    // bytecode callgraph analysis.
+    def isSimpleTarget(desc: mill.codesig.JvmModel.Desc) =
+      (desc.ret.pretty == classOf[mill.define.Target[?]].getName ||
+        desc.ret.pretty == classOf[mill.define.Worker[?]].getName) &&
+        desc.args.isEmpty
```
lihaoyi, line 1507: This should be in the `MillBuildRootModule` override method

scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -1380,6 +1490,60 @@ trait JavaModule
...
+          // In Scala 3, lambdas are implemented by private instance methods,
+          // not static methods, so they fall through the crack of "isSimpleTarget".
+          // Here make the assumption that a zero-arg lambda called from a simpleTarget,
+          // should in fact be tracked. e.g. see `integration.invalidation[codesig-hello]`,
+          // where the body of the `def foo` target is a zero-arg lambda i.e. the argument
+          // of `Cacher.cachedTarget`.
+          // To be more precise I think ideally we should capture more information in the signature
+          isSimpleTarget(callSiteSig.desc) && calledSig.name.contains("$anonfun")
+        )
+      }
```
lihaoyi, line 1533: These should be in the `MillBuildRootModule` override method I think; they look specific to Mill's own "ignore `Task`-returning methods" heuristic that won't apply to normal code

lihaoyi:
Left some comments. Let's also add some unit tests in `scalalib.test` to exercise this logic, next to the various `TestRunnerTests` that exercise the rest of the testing logic:

1. single-module Java project
2. single-module Scala project
3. multi-module Java project
4. multi-module Scala project

Since we don't need to make any changes to the `build.mill` to exercise this logic, unit tests are fine and we don't need an integration test.

scalalib/src/mill/scalalib/JavaModule.scala
HollandDM: If we allow analysis to go through `Task`-liked classes/objects, then by design it will spread into all `task`s. In `mill` project there is tests that create a custom module of `Test Module` (`TestModuleTestUtils`, etc...). So in this case it will trigger every `task`s.
Should we keep it like this, or allow `testQuick` to trigger `task`s like that?

scalalib/src/mill/scalalib/PrefixTrie.scala
```diff
@@ -0,0 +1,66 @@
+package mill.scalalib
+
+import scala.collection.mutable
+
+private[scalalib] final class PrefixTrie[A](using CanEqual[A, A]) {
```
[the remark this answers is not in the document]
HollandDM: True, I must have missed it

core/codesig/src/mill/codesig/CodeSig.scala
HollandDM: I have no ideal either, I put them here to mimic the original code, which log everything like this

scalalib/src/mill/scalalib/JavaModule.scala
lihaoyi: The issue is that this problem exists for other interfaces as well. For Mill and Mill plugins the heuristic is to ignore `Task`, but for other projects with other base abstractions the heuristic would need to be something different to be useful. Let's think if we can come up with a more general heuristic that isn't hardcoded to depend on Mill-specific types like `Task`, using the data we already have

What if rather than using a method-based callgraph, we did the analysis on a more coarse grained class-based dependency graph? This is something we can't really do for invalidating tasks since tasks are individual methods, but since test selection is at a granularity of classes that might work. Presumably if `ClassA` never directly or transitively references `ClassB`, and `ClassA` doesn't take any constructor parameters (since it's a test suite class and those usually have zero parameters) then changes to `ClassB` should never affect `ClassA`? This feels like an approach that could either replace the existing method-level approach, or be stacked on top as an additional heuristic (e.g. only invalidate a test or task if both the method-level codesig and the class-level codesig both change)

HollandDM: > The issue is that this problem exists for other interfaces as well. For Mill and Mill plugins the heuristic is to ignore Task, but for other projects with other base abstractions the heuristic would need to be something different to be useful. Let's think if we can come up with a more general heuristic that isn't hardcoded to depend on Mill-specific types like Task, using the data we already have

This is true, we do need to stop using Mill-specific heuristic

HollandDM:
@lihaoyi I've update the code to address the comments and answer some questions.

For using class based graph instead of method call, I've tried using the class dep graph produced from zinc as reference, while it works, I still need to use the `ReachabilityAnalysis` to check the hash code (not sure if zinc can do it itself). And after some thoughts, while leaving method call to pass the `Task` boundary result in a bigger call graph, in the end we only check if the method's class name  contains the module test classes discovered from `discoverTestClasses`, so the test classes that need to be run will never more than the normal `testForked`. So I guess leave it as it is for now is also good enough.

If the code is good to go, I'll proceed with writing test cases

lihaoyi:
Thanks @HollandDM ! will take a look

lihaoyi:
I think the code looks reasonable, at least as I would expect.

However I would like to push a bit more on the design of the feature. How does Zinc implement `testQuick`? Presumably they must have made similar tradeoffs as we are considering here. We should find out what approach they chose and why they chose it, so we can make a better decision rather than just arbitrarily picking a design on our own.

HollandDM:
Not sure what zinc's `testQuick` is, I don't think zinc has that, but `sbt` does have `testQuick` command.
If going zinc way, I think we can use the output analysis of previous `compile` and current `compile`, compare there `stamps` to extract changes classes (`product`s), then use the current analysis again to get all relations, and do something similar to spanning forest on the graph.
Comparing zinc stamp meaning we'll compare mostly the hash of the whole file content.

HollandDM:
@lihaoyi this is also updated to include the similar integration tests

integration/feature/test-quick/src/TestQuickTests.scala
```diff
@@ -0,0 +1,122 @@
...
+        val exists = firstRunOutLines.exists(_.contains(expectedLines))
+        assert(exists)
+      }
+
+      // Second run, nothing should run because we're not changing anything
+      val secondRun = eval("app.test.testQuick")
+      assert(secondRun.out.isEmpty)
+
+      // Third run, MyNumber.scala changed, so MyNumberDefaultValueTests should run
+      modifyFile(
```
HollandDM, line 30: This is one win of this approach, because we do a more fine grain graph invalidation, only relevant test class will need to be rerun

integration/feature/test-quick/src/TestQuickTests.scala
```diff
@@ -0,0 +1,122 @@
...
+        val exists = firstRunOutLines.exists(_.contains(expectedLines))
+        assert(exists)
+      }
+
+      // Second run, nothing should run because we're not changing anything
+      val secondRun = eval("app.test.testQuick")
+      assert(secondRun.out.isEmpty)
+
+      // Third run, Combinator.scala changed syntactically, should not run
+      modifyFile(
```
HollandDM, line 95: another win for this approach is that it work on JVM bytecode, so some syntax changes do not invalidate node in graph, although source hashcode can be changed

HollandDM:
I run this with time command for a few time, here is the result:

./mill-assembly.jar --no-server scalalib.test.testForked: 147,53s; 148.53s; 151.19s (avg: 149.08s)
./mill-assembly.jar --no-server scalalib.test.testQuick: 154.19s; 150.36s; 169.86s (avg: 158.14s)

So in average, the cost to calculate the re-run test is about ~9 extra seconds. Compare to the overall runtime of the command, this is about ~5.8% slow down.
The testQuick was triggered by modifying UnitTester.scala file, which will invalidate all test classes

HollandDM:
@lihaoyi the example test and adoc are pushed, please have a look.
For some reason, I cannot make Java example to run correctly, despite Scala and Kotlin does. I will look into further when rebasing.
But I do want to ask for your experience on Java `JupiterFramework` framework. In the example, the test run correctly until the line `> mill -j 1 foo.test.testQuick # Fourth run re-runs failing FooTest2 even without code changes`. Some debugging yields that `FooTest2.java` is not even run, hence it cannot throw error.

lihaoyi:
@HollandDM I know that Junit5/Jupiter has some special casing in most build tools, including Mill, not sure if that's what you're seeing, but AFAIK it should take place entirely before the test filtering logic done by `testQuick` so I don't see how it could affect your test:

- https://github.com/com-lihaoyi/mill/blob/d2a31808da38682dba74f118209bc842f30b0382/libs/scalalib/src/mill/scalalib/TestModule.scala#L374

Is `FooTest2` not being run the second time, or is it not being run at all even the first time? Is it run if you use normal `testForked`?

example/javalib/testing/7-test-quick/build.mill
```diff
@@ -0,0 +1,33 @@
...
+  }
+  // Ultilities for replacing text in files
+  def replaceBar(args: String*) = Task.Command {
+    val relativePath = os.RelPath("../../../foo/src/Bar.java")
+    val filePath = Task.dest() / relativePath
+    os.write.over(filePath, os.read(filePath).replace(
+      """return String.format("Hi, %s!", name);""",
+      """return String.format("Ciao, %s!", name);"""
+    ))
+  }
```
lihaoyi, line 22: Let's move these into the `/** Usage` blocks using `sed -i.bak` rather than having it as part of the Scala code; can grep for that string to see examples of that in the codebase

HollandDM:
> Is `FooTest2` not being run the second time, or is it not being run at all even the first time? Is it run if you use normal `testForked`?
It does run the first time, and stop running the second time if there is no update

PR #4736, opened 2025-03-16
Support `nativeServiceProviders` in `ScalaNativeModule`

lolgab:
Scala Native supports `ServiceProvider` which allows loading implementations for interfaces using reflection. To avoid bloating binaries with all available implementations, Scala Native requires to set at build time the implementations that are used.
This PR adds support for it in Mill.

Pull Request: https://github.com/com-lihaoyi/mill/pull/4736

### DIFF
--- a/scalanativelib/src/mill/scalanativelib/ScalaNativeModule.scala
+++ b/scalanativelib/src/mill/scalanativelib/ScalaNativeModule.scala
@@ -225,7 +225,13 @@ trait ScalaNativeModule extends ScalaModule { outer =>
    */
   def nativeMultithreading: T[Option[Boolean]] = Task { None }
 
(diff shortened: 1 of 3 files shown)


lihaoyi (approved):

lihaoyi:
Could you flesh out the scaladoc? What is this `Map[String, Seq[String]]` datastructure and what do the various elements in that map mean?

lefou (approved):
Looks good to me.

Out of curiosity. Are the typical files under `META-INF` still relevant?

lolgab:
> Looks good to me.
> 
> Out of curiosity. Are the typical files under `META-INF` still relevant?

Yes, they are :)
This is a feature to contain binary sizes avoid avoid bundling all possible implementations in a binary.

Issue #4735, opened 2025-03-16
Add progress and failure indicators to Mill test runner (500USD Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 500USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

See https://github.com/orgs/com-lihaoyi/discussions/6 for other bounties and the terms and conditions that bounties operate under 

---

Currently, the prompt shows the test classes that each test task is running, the duration of the task, and the duration of each test runner worker

```scala
[1427/1427] ============================== scalalib.test ============================== 54s
[1427] scalalib.test.test 25s
[1427-0] 23s mill.javalib.revapi.RevapiModuleTests
[1427-1] 23s mill.scalalib.BomTests
[1427-2] 23s mill.scalalib.AssemblyExeTests
[1427-3] 23s mill.scalalib.HelloJavaTests
[1427-4] 23s mill.scalalib.AssemblyNoExeTests
```

We should extend this UI in three ways:

1. Show the time each test class has been running for, on the right of that class name
2. Show the number of tests that have completed so far , on the right of the test task via the ticker
3. Show the count of test class failures so far

So it should look something like this

```scala
[1427/1427] ============================== scalalib.test ============================== 54s
[1427] scalalib.test.test 25s 14/58, 5 failures
[1427-0] 23s mill.javalib.revapi.RevapiModuleTests 5s
[1427-1] 23s mill.scalalib.BomTests 4s
[1427-2] 23s mill.scalalib.AssemblyExeTests 1s
[1427-3] 23s mill.scalalib.HelloJavaTests 1s
[1427-4] 23s mill.scalalib.AssemblyNoExeTests
```

PR #4737, opened 2025-03-16
Add test runner indicators

HollandDM:
/claim #4735

This PR add some calculation to add test progress & failure indicator for each test runner while `testParallelism` is set to `true`

Screenshot 2025-03-16 at 23 50 22

To add progress indicators, this pull request introduces a new file within each worker's claim folder. This file is overwritten every time `runClaimedTestClass` is called. The parent process will checks these files in each worker, accumulates their contents, and outputs them via a `ticker`.

For test timing, this pull request uses a separate map, similar to `workerStatusMap`, to store the starting time. The time calculation is then straightforward, and the result is output via a `ticker` alongside the test class name.

### DIFF
--- a/scalalib/src/mill/scalalib/TestModuleUtil.scala
+++ b/scalalib/src/mill/scalalib/TestModuleUtil.scala
@@ -225,7 +225,7 @@ private final class TestModuleUtil(
   )(implicit ctx: mill.api.Ctx) = {
 
     val workerStatusMap = new java.util.concurrent.ConcurrentHashMap[os.Path, String => Unit]()
-    val workerStatsSet = new java.util.concurrent.ConcurrentHashMap[os.Path, Unit]()
+    val workerResultSet = new java.util.concurrent.ConcurrentHashMap[os.Path, Unit]()
     val testClassTimeMap = new java.util.concurrent.ConcurrentHashMap[String, Long]()
 
     def prepareTestClassesFolder(selectors2: Seq[String], base: os.Path): os.Path = {
@@ -266,13 +266,13 @@ private final class TestModuleUtil(
         workerStatusMap.put(claimLog, logger.ticker)
         val claimStats = claimFolder / os.up / s"${claimFolder.last}.stats"
         os.write.over(claimStats, upickle.default.write((0L, 0L)))
-        workerStatsSet.put(claimStats, ())
+        workerResultSet.put(claimStats, ())
         val result = callTestRunnerSubprocess(
           base,
           Right((startingTestClass, testClassQueueFolder, claimFolder))
         )
         workerStatusMap.remove(claimLog)
-        // We don't remove workerStatsSet entry, as we still need them
+        // We don't remove workerResultSet entry, as we still need them
         // for calculation of total success/failure counts
         Some(result)
       } else {
@@ -376,16 +376,12 @@ private final class TestModuleUtil(
               os.read.lines(claimLog).lastOption.foreach { currentTestClass =>
                 testClassTimeMap.putIfAbsent(currentTestClass, now)
                 val last = testClassTimeMap.get(currentTestClass)
-                val suffix = ((now - last) / 1000).toInt match {
-                  case 0 => ""
-                  case n => s" ${n}s"
-                }
-                callback(s"$currentTestClass$suffix")
+                callback(s"$currentTestClass${mill.internal.Util.renderSecondsSuffix(now - last)}")
               }
             }
             var totalSuccess = 0L 
             var totalFailure = 0L
-            workerStatsSet.forEach { (claimStats, _) =>
+            workerResultSet.forEach { (claimStats, _) =>
               val (success, failure) = upickle.default.read[(Long, Long)](os.read.stream(claimStats))
               totalSuccess += success
               totalFailure += failure
--- a/scalalib/src/mill/scalalib/TestModuleUtil.scala
+++ b/scalalib/src/mill/scalalib/TestModuleUtil.scala
@@ -123,17 +123,25 @@ private final class TestModuleUtil(
       //     - list of glob selectors to feed to the test runner directly.
       // - Right((startingTestClass, testClassQueueFolder, claimFolder)):
       //     - first test class to run, folder containing test classes for test runner to claim from, and the worker's base folder.
-      selector: Either[Seq[String], (Option[String], os.Path, os.Path)]
+      selector: Either[Seq[String], (Option[String], os.Path, os.Path)],
+      resultSet: java.util.concurrent.ConcurrentHashMap[os.Path, Unit]
   )(implicit ctx: mill.api.Ctx) = {
     os.makeDir.all(baseFolder)
 
     val outputPath = baseFolder / "out.json"
+
+    // test runner will log success/failure test class counter here while running
+    val resultPath = baseFolder / s"result.log"
+    os.write.over(resultPath, upickle.default.write((0L, 0L)))
+    resultSet.put(resultPath, ())
+
     val testArgs = TestArgs(
       framework = testFramework,
       classpath = runClasspath.map(_.path),
       arguments = args,
       sysProps = props,
       outputPath = outputPath,
+      resultPath = resultPath,
       colored = Task.log.prompt.colored,
       testCp = testClasspath.map(_.path),
       globSelectors = selector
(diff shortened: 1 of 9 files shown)


testrunner/src/mill/testrunner/TestRunnerUtils.scala
```diff
@@ -347,4 +361,26 @@ import java.io.PrintStream
       filters.exists(f => f(name))
     }
   }
+
+  @internal private[TestRunnerUtils] final class TestEventSummary(outputFileOpt: Option[os.Path])
+      extends AtomicBoolean(true) {
```
lihaoyi, line 366: This class wrapper and inheritance is kind of weird. Let's convert them to just static helper methods and pass in `outputFileOpt` and the `AtomicBoolean` as arguments where necessary

testrunner/src/mill/testrunner/TestRunnerUtils.scala
```diff
@@ -347,4 +361,26 @@ import java.io.PrintStream
...
+        case _ => val _ = compareAndSet(true, true) // consider success as a default
+      }
+    }
+
+    def summary(): Boolean = {
+      val isSuccess = get()
+      outputFileOpt.foreach { outputFile =>
+        val (success, failure) = upickle.default.read[(Long, Long)](os.read.stream(outputFile))
+        val (newSuccess, newFailure) =
+          if (isSuccess) (success + 1, failure) else (success, failure + 1)
```
lihaoyi, line 380: This workflow to both read and write to the file on disk is unnecessarily complicated. Let's just keep the success/failure count as in-memory `Long` (or `AtomicLong`) and have the file be write-only purely for the purpose of passing the information to the parent process

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -374,15 +403,20 @@ private final class TestModuleUtil(
...
           }) Thread.sleep(1)
         }
         subprocessFutures.flatMap(_.value).map(_.get)
-      } finally executor.shutdown()
+      } finally {
+        executor.shutdown()
+        // help gc
+        workerStatusMap.clear()
+        workerStatsSet.clear()
+        testClassTimeMap.clear()
```
lihaoyi, line 418: This should be unnecessary, let's skip it

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -262,11 +264,15 @@ private final class TestModuleUtil(
...
         workerStatusMap.put(claimLog, logger.ticker)
+        val claimStats = claimFolder / os.up / s"${claimFolder.last}.stats"
+        os.write.over(claimStats, upickle.default.write((0L, 0L)))
+        workerStatsSet.put(claimStats, ())
         val result = callTestRunnerSubprocess(
           base,
           Right((startingTestClass, testClassQueueFolder, claimFolder))
         )
         workerStatusMap.remove(claimLog)
+        // We don't remove workerStatsSet entry, as we still need them for calculation
```
lihaoyi, line 275: Can elaborate a bit more `...for calculation of total success/failure counts`

lihaoyi:
@HollandDM Looks pretty good, some small changes requested. Please also update the PR description with a summary of your implementation strategy

testrunner/src/mill/testrunner/TestRunnerUtils.scala
HollandDM: This is a trick to better the memory layout of object, I guess it's kind of overkill here

HollandDM:
Updated to adrees the feedbacks

testrunner/src/mill/testrunner/TestRunnerUtils.scala
```diff
@@ -135,17 +136,22 @@ import java.io.PrintStream
...
     while (taskQueue.nonEmpty) {
       val next = taskQueue.dequeue().execute(
         new EventHandler {
           def handle(event: Event) = {
             testReporter.logStart(event)
             events.add(event)
+            event.status match {
+              case Status.Error => taskStatus.set(false)
+              case Status.Failure => taskStatus.set(false)
+              case _ => taskStatus.compareAndSet(true, true) // consider success as a default
```
lihaoyi, line 153: I think we can leave out this `compareAndSet`; this branch of the `match` being a no-op would be sufficient

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -225,6 +225,8 @@ private final class TestModuleUtil(
   )(implicit ctx: mill.api.Ctx) = {
 
     val workerStatusMap = new java.util.concurrent.ConcurrentHashMap[os.Path, String => Unit]()
+    val workerStatsSet = new java.util.concurrent.ConcurrentHashMap[os.Path, Unit]()
```
lihaoyi, line 228: Maybe call this `workerResultsSet`? `Stats` and `Status` sounds too similar

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -355,16 +362,39 @@ private final class TestModuleUtil(
...
               // the last one is always the latest
-              os.read.lines(claimLog).lastOption.foreach(callback)
-            },
+              os.read.lines(claimLog).lastOption.foreach { currentTestClass =>
+                testClassTimeMap.putIfAbsent(currentTestClass, now)
+                val last = testClassTimeMap.get(currentTestClass)
+                val suffix = ((now - last) / 1000).toInt match {
+                  case 0 => ""
+                  case n => s" ${n}s"
+                }
```
lihaoyi, line 382: Let's move `PromptLoggerUtil.renderSecondsSuffix` to `mill.internal.Util` so we can re-use it here

HollandDM:
Updated again. After the ticket is good to merge, I'll squash them back into 1 commit, I think it'll be better

lihaoyi:
I think this looks good for `testParallelism = true`. Could you also make the progress count and failure count show in `testParallelism = false`? I think a lot of code would be re-usable, and it would help improve the UX of the 0.12.x branch where test parallelism is false by default

lihaoyi:
No need to worry about squashing your commits, we'll squash it during the merge using github's squash-and-merge button

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -355,16 +362,35 @@ private final class TestModuleUtil(
...
+            var totalSuccess = 0L
+            var totalFailure = 0L
+            workerResultSet.forEach { (claimStats, _) =>
+              val (success, failure) =
+                upickle.default.read[(Long, Long)](os.read.stream(claimStats))
+              totalSuccess += success
+              totalFailure += failure
+            }
+            ctx.log.ticker(
+              s"${totalSuccess + totalFailure}/${filteredClassCount} completed, ${totalFailure} failures"
```
lihaoyi, line 391: Let's only print `n failures` for `n > 0`, since that's what we do in the main prompt for failed tasks

HollandDM:
Updated, I added progress indicator (complete/failure) to default run as well.

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -172,59 +180,92 @@ private final class TestModuleUtil(
...
+        () => {
+          val (totalSuccess, totalFailure) = TestModuleUtil.calculateTestProgress(workerResultSet)
+          ctx.log.ticker(s"${totalSuccess + totalFailure}/${filteredClassCount} completed${
+              if totalFailure > 0 then s", ${totalFailure} failures" else ""
+            }")
+        },
+        0,
+        20,
+        java.util.concurrent.TimeUnit.MILLISECONDS
+      )
```
lihaoyi, line 201: Let's consolidate the whole `val workerResultSet`, `newScheduledThreadPool`, `scheduleWithFixedDelay`, `finally executor.shutdown` logic into a single 

```scala
def withTestProgressTickerThread(updateWorkerStatuses: Boolean)(block: => T): T
```

That should help DRY up the duplicate code, and also shorten some of these `def runTestQueueScheduler`/`def runTestDefault` methods which are getting rather long

testrunner/src/mill/testrunner/TestRunnerUtils.scala
```diff
@@ -203,15 +211,34 @@ import java.io.PrintStream
   }
 
   def runTasks(
-      tasks: Seq[Task],
+      tasksSeq: Seq[Seq[Task]],
```
lihaoyi, line 214: What's the reason for this change in signature from `Seq[Task]` to `Seq[Seq[Task]]`? I've been looking at the code and haven't managed to figure out the intention

lihaoyi: I guess each element of the outer `Seq` represents one test class, and each inner `Seq` represents the tasks for that test class? If so, probably worth a comment somewhere

lihaoyi:
Some last nitpicks. Tried it out locally and works great. Turned on CI, once it's green it should be good to merge

testrunner/src/mill/testrunner/TestRunnerUtils.scala
HollandDM: Yeah, I'm adding the comments in the backport PR, will add it in this as well. But it's meaning is exactly what you've said

scalalib/src/mill/scalalib/TestModuleUtil.scala
HollandDM: I though about this, but because it has codes that belong to both inside & outside `scheduleWithFixedDelay`, so I struggle with some naming/signature.
I think it can be `def withTestProgressTickerThread(updateWorkerStatuses: Boolean)(fixedDelayBlock: => Unit)(block: => T): T`

lihaoyi:
The `mill.scalalib.TestRunnerTests.doneMessage.failure` seems to happen reliably on my laptop as well, might need to tweak the test or code to make it pass

HollandDM:
updated, the test failing is due to the changes of `Seq[Task]` to `Seq[Seq[Task]]`, I left a comment there

scalalib/src/mill/scalalib/TestModuleUtil.scala
```diff
@@ -262,17 +296,29 @@ private final class TestModuleUtil(
...
+        // test runner will log success/failure test class counter here while running
+        val resultPath = base / s"result.log"
+        os.write.over(resultPath, upickle.default.write((0L, 0L)))
+        workerResultSet.put(resultPath, ())
+
         val result = callTestRunnerSubprocess(
           base,
+          resultPath,
           Right((startingTestClass, testClassQueueFolder, claimFolder))
         )
```
lihaoyi, line 308: Can we consolidate this logic into `runTestRunnerSubprocess` as well?

HollandDM: `runTestRunnerSubprocess`'s are quite different between `queueRunner` and `defaultRunner`, so I have to create 2 versions for it, each in the corresponding runners function.
I'm also thinking of a way to further clean up the logic, if we can generalize both runners to just produce blueprint for subprocess, and do the spawn + logging in a shared function, then the code should be clearer. But that required some refactoring.
If you want, I can try to go that way and see if we can do any better.

lihaoyi: Got it, I think for now let's just leave it as is then

lihaoyi:
Thanks @HollandDM , i will send the bounty using your existing transfer details

HollandDM:
Thanks @lihaoyi, lets use that as my transfer detail

PR #4744, opened 2025-03-17
[0.12.x] Support `nativeServiceProviders` in `ScalaNativeModule` (Backport #4736)

lolgab:
Scala Native supports `ServiceProvider` which allows loading implementations for interfaces using reflection. To avoid bloating binaries with all available implementations, Scala Native requires to set at build time the implementations that are used. This PR adds support for it in Mill.

Pull Request: https://github.com/com-lihaoyi/mill/pull/4736


lefou (commented):
Looks good to me, but there is a failing test in the scalanative group, so would be good to check it before merging.

PR #4745, opened 2025-03-18
Try out Scala 3.7.0-RC1

lihaoyi:


lihaoyi:
Basically all tests pass now, so we can count this as done until 3.7.0 final is out

PR #4747, opened 2025-03-18
Use en_US.UTF-8 locale in scalalib tests on macOS

alexarchambault:
Seems C.UTF-8 isn't available there


lefou:
> Seems C.UTF-8 isn't available there

I `POSIX` available? That should be the equal to classic `C`.

E.g.

```
> locale -a
C
C.utf8
POSIX
de_DE
de_DE.iso88591
de_DE.iso885915@euro
de_DE.utf8
de_DE@euro
en_US
en_US.iso88591
en_US.utf8
```

alexarchambault:
There's both `C` and `POSIX`. But that might not be enough, we need a better character encoding than just ASCII I think. The tests involving dependency trees involve non-ASCII characters to make the tree.

lefou (approved):

PR #4748, opened 2025-03-18
Set user.language in scalalib tests on macOS and Windows…

alexarchambault:
…and propagate this property in the JVM app spawned by `CheckstyleModule`.

In the same vein https://github.com/com-lihaoyi/mill/pull/4535 and https://github.com/com-lihaoyi/mill/pull/4747. Along those, this makes `./mill scalalib.test` pass on my local machine (macOS).


lefou:
Is this still needed when https://github.com/com-lihaoyi/mill/pull/4747 will be merged? Shouldn't Java derive its language settings from those standard env variables?

alexarchambault:
@lefou Precisely, it doesn't always follow `LC_ALL`. On my machine (Mac ARM)
```text
$ LC_ALL="en_US.UTF-8" scala-cli -e 'println(System.getenv("LC_ALL")); println(sys.props.get("user.language"))' --jvm 8
Starting compilation server
Compiling project (Scala 3.6.3, JVM (8))
Compiled project (Scala 3.6.3, JVM (8))
en_US.UTF-8
Some(fr)
```

lefou:
Which test is it, that depends on `user.language`. I never saw anything except Eclipse UI using these properties.

alexarchambault:
> Which test is it, that depends on `user.language`. I never saw anything except Eclipse UI using these properties.

This fails on `main` on my machine, and succeed with these changes:
```
$ ./mill scalalib.test.testOnly "mill.javalib.checkstyle.CheckstyleModuleTest"
```

It seems checkstyle prints some stuff in French, when the locale is French too.

build.mill
```diff
@@ -388,7 +389,12 @@ trait MillJavaModule extends JavaModule {
 
   // Workaround for Zinc/JNA bug
   // https://github.com/sbt/sbt/blame/6718803ee6023ab041b045a6988fafcfae9d15b5/main/src/main/scala/sbt/Main.scala#L130
-  def testArgs: T[Seq[String]] = Task { Seq("-Djna.nosys=true") }
+  def testArgs: T[Seq[String]] = Task {
+    val userLang =
+      if (Properties.isMac || Properties.isWin) Seq("-Duser.language=en")
+      else Nil
+    Seq("-Djna.nosys=true") ++ userLang
+  }
```
lefou, line 397: Can you keep the comment (above this block) close to the `-Djna.nosys=true`, where it belongs.

scalalib/src/mill/javalib/checkstyle/CheckstyleModule.scala
```diff
@@ -31,6 +31,9 @@ trait CheckstyleModule extends JavaModule {
       Seq("-f", checkstyleFormat()) ++
       (if (stdout) Seq.empty else Seq("-o", output.toString())) ++
       (if (leftover.value.nonEmpty) leftover.value else sources().map(_.path.toString()))
+    val jvmArgs = sys.props.get("user.language")
+      .map(lang => s"-Duser.language=$lang")
+      .toSeq
```
lefou, line 36: Using `sys.props` here adds a hidden input to this anonymous task. It won't trigger a re-evaluating if the user changes this property. It's currently called from a command, so it should be no issue. But it's a hidden code smell.

Maybe, we should pass the language as parameter?

alexarchambault: Just added a `Task.Input` reading this property, `CheckstyleModule#checkstyleLanguage`

lefou (approved):
Looks good to me. Would be good, if you can update the title/comment and refer to the addition to `CheckstyleModule` and merge it.

PR #4749, opened 2025-03-18
Use minDependencies instead of orderedDependencies in allDeps

joan38:
We have a plugin that was doing:
```scala
  defaultResolver()
    .allDeps(dependencies)
    .find(_.module.name.value == "flink-core")
    .getOrElse(
      throw new IllegalStateException("Dependency flink-core not found to determine the Flink version")
    )
```
And it's not working anymore. I noticed with `minDependencies` it works. I don't understand everything so maybe @alexarchambault can provide a review?

### DIFF
--- a/scalalib/src/mill/scalalib/CoursierModule.scala
+++ b/scalalib/src/mill/scalalib/CoursierModule.scala
@@ -300,7 +300,7 @@ object CoursierModule {
      */
     def allDeps[T: CoursierModule.Resolvable](
         deps: IterableOnce[T]
-    ): Seq[coursier.core.Dependency] = {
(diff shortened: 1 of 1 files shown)


alexarchambault:
`orderedDependencies` returns dependencies in a sorted fashion, ready to be used as class path.

In the Mill code base, `allDeps` is used by the BSP server, where we should keep using `orderedDependencies` IMO.

@joan38 What does `allDeps` return in your case? `orderedDependencies` and `minDependencies` ought to contain the same modules at least.

joan38:
This is resolved with the new `defaultResolver().resolution()` API.
Thanks

PR #4759, opened 2025-03-21
Android: Hilt support with tranform ASM and todo external app sample build

vaslabs:
# Summary

~~This PR is for encouraging discussion on how to proceed with Hilt support on Android. It composes of 2 main things:~~

~~1. How far I could go for Hilt support in Mill with current provided tools~~
~~2. How Gradle does it and what's missing from Mill to achieve the same result~~

This PR is now ready for review.

Covers part of https://github.com/com-lihaoyi/mill/issues/3868 (Hilt)

# How far this PR goes

With Kotlin KSP support added in https://github.com/com-lihaoyi/mill/pull/4557 and https://github.com/com-lihaoyi/mill/pull/4485 generating the Hilt source files is relatively easy.

Java files are generated and I'm demonstrating in the PR how each compilation step (Kotlin KSP, Kotlin compose, Java compilation and Java annotation processing) are generated, following the exact steps gradle is doing.

So in the end we have:

1. Kotlin compiled sources
2. Java sources generated from the KSP and Hilt compiler
4. Java compiled sources (including the _ComposedTreeDeps - an aggregation of all dependencies that hilt does)
5. Transforms ASM for the dependency injection step in compile time  (only for main sources)
 
image

Debug and test sources are not covered due to the amount of changes, I've decided to add them in a future PR.

## The Transform ASM step

There is one step, that is done via a gradle plugin (android-hilt-gradle-plugin) that is not straightforward to reuse. This is what makes dependency injection possible.

In the demonstrated example, TodoActivity's constructor should look like this after the final stage of compilation:

```
.method public constructor <init>()V
    .registers 1

    .line 29
    invoke-direct {p0}, Lcom/example/android/architecture/blueprints/todoapp/Hilt_TodoActivity;-><init>()V

    return-void
.end method
```

instead of the (normally) compiled

```
.method public constructor <init>()V
    .registers 1

    .line 29
    invoke-direct {p0}, Landroidx/activity/ComponentActivity;-><init>()V

    .line 28
    return-void
.end method
```

# How gradle does it

Gradle does the following steps on the release task (assembleRelease):

0. Fetch dependencies and jetify them (this I've skipped as it shouldn't be needed with new android API and dependency versions)
1. Run KSP (kotlin with symbol processing plugin) which runs Dagger and Hilt source generation, here the sources are written
2. Run Kotlin with compose plugin
3. Run java compilation on generated sources
4. Run java compilation on the component tree classes (there should just be 2) with hilt-compiler as the annotation processor
6. Run transform ASM which achieves what is described in critical step

# What is needed

In my opinion, to achieve Hilt support for Mill (and have the android architecture examples working) we need to implement Mill's version of [AndroidEntryPointClassVisitor](https://github.com/google/dagger/blob/b3d3443e3581b8530cd85929614a1765cd37b12c/java/dagger/hilt/android/plugin/main/src/main/kotlin/dagger/hilt/android/plugin/transform/AndroidEntryPointClassVisitor.kt#L122)

With 2 instrumentation related dependencies, one specific for android (`com.android.build.api.instrumentation`) and  (`org.objectweb.asm`) which is already in Mill if I'm not mistaken!

I would like your opinions on how to proceed!

## TL;DR 

Hilt should be possible after we implement or somehow reuse android-hilt-gradle-plugin's transform ASM step to achieve the dependency injection step for Activities 

## Update

The transform asm step, is done in `mill.main.android.hilt.AndroidHiltTransformAsm` which is provided as an external program, that takes a directory to scan for classes and re-write them to a destination directory.

The rewrite happens only if a class `isInstrumentable` , i.e. it contains any of the following annotations:

```
        descriptor == "Ldagger/hilt/android/AndroidEntryPoint;" ||
        descriptor == "Ldagger/hilt/android/HiltAndroidApp;"
```

which is similar to [hilt's gradle plugin](https://github.com/google/dagger/blob/b3d3443e3581b8530cd85929614a1765cd37b12c/java/dagger/hilt/android/plugin/main/src/main/kotlin/dagger/hilt/android/plugin/transform/AndroidEntryPointClassVisitor.kt#L122):

```
  override fun isInstrumentable(classData: ClassData) =
      classData.classAnnotations.any { ANDROID_ENTRY_POINT_ANNOTATIONS.contains(it) }
  }
  ```
  
## Other changes

I've introduced androidApplicationNamespace and androidApplicationId , but probably this needs a cleanup (we need a user defined package name to add to the android manifest). Gradle accepts these configs as:

```
android {
    namespace = "com.example.android.architecture.blueprints.todoapp"
    compileSdk = libs.versions.compileSdk.get().toInt()

    defaultConfig {
        applicationId = "com.example.android.architecture.blueprints.main"
```

and since I have not used the application id in this PR after all, I can remove it until its need is more evident 

I've also changed the androidInstall to a command, firstly it can be run now more than once (instead of being cached) and in the future it can also accept parameters (e.g. to target a specific emulator)

EDIT: Application Id can be used and altered in build variants, we are gonna look into this soon, whether it stays or it's added later is not an issue for me

### CODE REVIEW

**Comment by vaslabs** [comment]
@lihaoyi @lefou @0xnm 

I've done some exploration on [android's architecture samples](https://github.com/android/architecture-samples) to stretch what's missing from Mill and what we need to do. A major obstacle is supporting Hilt and you can check my findings in the PRs description.

I'd like opinions and advise on how you think we can proceed! 

This of course is not something we need to do or support now, but let me know your thoughts

**Comment by lefou** [comment]
TBH, I don't understand why tooling is so often bound to specific build tools. This is not unique to Gradle, to be fair. Also many Scala tools came as `sbt` plugin first, and only some where split into an generic library and a specific `sbt`-plugin later. 

I really don't know (yet) what HILT is all about, but if we're going to make HILT support working outside of Gradle, we should strive for an Mill-independent library or CLI tool (ideally both) and a Mill integration.

**Comment by vaslabs** [comment]
> TBH, I don't understand why tooling is so often bound to specific build tools. This is not unique to Gradle, to be fair. Also many Scala tools came as `sbt` plugin first, and only some where split into an generic library and a specific `sbt`-plugin later.
> 
> I really don't know (yet) what HILT is all about, but if we're going to make HILT support working outside of Gradle, we should strive for an Mill-independent library or CLI tool (ideally both) and a Mill integration.

I haven't tried it, but from what I see it shouldn't be too hard to extract a cli tool from [AndroidEntryPointClassVisitor](https://github.com/google/dagger/blob/b3d3443e3581b8530cd85929614a1765cd37b12c/java/dagger/hilt/android/plugin/main/src/main/kotlin/dagger/hilt/android/plugin/transform/AndroidEntryPointClassVisitor.kt#L122) .

Basically there are 3 critical functions, `isInstrumentable` , `visit` and `visitMethod`

Might be worth checking if there's already an initiative to extract this functionality, but I think the `com.android.build.api.instrumentation` package is also [part of gradle](https://developer.android.com/reference/tools/gradle-api/8.11/com/android/build/api/instrumentation/package-summary) , which could make things trickier

**Comment by lihaoyi** [comment]
I don't have a strong preference for how the code is supported in Mill, but the nature of `AndroidEntryPointClassVisitor` seems like it would fit fine in a `worker` module in Mill that we call using `Jvm.callProcess` or `Jvm.createClassLoader`. That would satisfy @lefou's "separate re-usable module" idea while also helping e.g. not bloat the main Mill executable.

It's definitely a long-term maintenance risk implementing this in Mill and then having to keep up to date as upstream evolves, but that's the nature of targeting a moving platform like Android, and isn't unique to this particular part of the problem

**Comment by vaslabs** [comment]
> I don't have a strong preference for how the code is supported in Mill, but the nature of `AndroidEntryPointClassVisitor` seems like it would fit fine in a `worker` module in Mill that we call using `Jvm.callProcess` or `Jvm.createClassLoader`. That would satisfy @lefou's "separate re-usable module" idea while also helping e.g. not bloat the main Mill executable.
> 
> It's definitely a long-term maintenance risk implementing this in Mill and then having to keep up to date as upstream evolves, but that's the nature of targeting a moving platform like Android, and isn't unique to this particular part of the problem

cool, I'll give it a go with a worker module, could you point me to a couple of examples?

**Comment by lihaoyi** [comment]
Perhaps the best example is https://github.com/com-lihaoyi/mill/blob/main/main/graphviz/src/mill/main/graphviz/GraphvizTools.scala. It's a separate module containing Graphviz and the necessary Node.js runtime to run it, and is resolved and downloaded on-demand via coursier only for people who use it.

**Comment by vaslabs** [comment]
@lefou @lihaoyi 

Finally I managed to make it work for the main sources of the android architecture samples!

This was a lot of work and I'm kind of fatigued with the changes :smile: 

So please have a review, it still needs a cleanup, but I've replicated the AGP logic for instrumenting the hilt classes using the way @lihaoyi recommended! So if you do a review pass, it would help me clean things up!

I've tested it manually and it works:

Screenshot From 2025-03-31 16-30-07

Let me know what you think!

**Comment by lihaoyi** [comment]
@vaslabs before I review the code, could you update the PR description summarizing the current implementation? That would help me get my bearings so I know what I'm looking at in this PR

**Comment by lihaoyi** [comment]
One thing to start is that we probably don't want to include the entire android todoapp example as part of this repo in the test suite; we should minimize the part that is interesting and leave the "entire android app" either for manual testing or as a `thirdparty/` test which clones the repo at build time. For the purposes of this work, I think manual testing is fine for now, since it's pretty experimental and I expect it to evolve

**Comment by vaslabs** [comment]
> @vaslabs before I review the code, could you update the PR description summarizing the current implementation? That would help me get my bearings so I know what I'm looking at in this PR

I have updated the description, although I'm happy if we had a call and we go through things, because this wasn't easy and it would help me as well to talk through it and clean things up.



> One thing to start is that we probably don't want to include the entire android todoapp example as part of this repo in the test suite; we should minimize the part that is interesting and leave the "entire android app" either for manual testing or as a `thirdparty/` test which clones the repo at build time. For the purposes of this work, I think manual testing is fine for now, since it's pretty experimental and I expect it to evolve

I have deleted anything that's not needed, left only the main sources, and tested again manually

**Comment by lihaoyi** [comment]
Sure let me put some time on your calendar

File: kotlinlib/src/mill/kotlinlib/KotlinModule.scala [hunk 1]
```diff
@@ -402,6 +404,10 @@ trait KotlinModule extends JavaModule { outer =>
     )
   }
 
+  def kotlinPrecompiledClasses: Task[Seq[PathRef]] = Task {
```
lefou, line 407: What are these? Who is supposed to pre-compile them?

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -288,7 +288,9 @@ trait KotlinModule extends JavaModule { outer =>
       val isJava = javaSourceFiles.nonEmpty
       val isMixed = isKotlin && isJava
 
-      val compileCp = compileClasspath().map(_.path).filter(os.exists)
+      val compileCp =
+        (compileClasspath() ++ kotlinPrecompiledClasses()).map(_.path).filter(os.exists)
```
lefou, line 292: Why are "pre-compiled" classes not part of the `compileClasspath`?

kotlinlib/src/mill/kotlinlib/android/AndroidHiltSupport.scala
```diff
@@ -0,0 +1,305 @@
...
+  }
+
+  override def androidGeneratedCompiledClasses: T[Seq[PathRef]] =
+    androidHiltGeneratedClasses
+
+  def hiltProcessorClasspath: T[Seq[PathRef]] = Task {
+    kspApClasspath() ++ compileClasspath()
+  }
+
+  override def kotlinPrecompiledClasses: Task[Seq[PathRef]] = androidGeneratedCompiledClasses
```
lefou, line 284: Don't reuse tasks, this results in invalid (misleading) task contexts. Instead, refer to their results.

```suggestion
  override def kotlinPrecompiledClasses: Task[Seq[PathRef]] = androidGeneratedCompiledClasses()
```

of 


```suggestion
  override def kotlinPrecompiledClasses: Task[Seq[PathRef]] = Task {
    androidGeneratedCompiledClasses()
  }
```

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -76,14 +76,57 @@ trait KspModule extends KotlinModule { outer =>
...
+  /**
+   * Kotlinc arguments used with KSP
+   * @return
+   */
+  def kspKotlinc: T[Seq[String]] = Seq(
+    "-Xallow-unstable-dependencies",
+    "-no-reflect",
+    "-no-stdlib",
+    "-language-version",
+    "1.9"
```
lefou, line 115: Shouldn't we re-use `kotlinLanguageVersion` here?

main/src/mill/main/AndroidHiltModule.scala
```diff
@@ -0,0 +1,23 @@
...
+import mill.api.PathRef
+import mill.define.{Discover, ExternalModule, Target}
+import mill.util.MillModuleUtil.millProjectModule
+
+object AndroidHiltModule extends ExternalModule {
+  def repositories: Seq[Repository] = Seq(
+    LocalRepositories.ivy2Local,
+    MavenRepository("https://repo1.maven.org/maven2"),
+    MavenRepository("https://maven.google.com/")
+  )
```
lefou, line 15: This looks suspicious and ignored any user configuration. I assume, you need it to resolve dependencies later.

Instead you could hold a reference to a `JvmWorkerModule` by extending `WithJvmWorker` and use `jvmWorker().defaultResolver()` to resolve dependencies.

main/src/mill/main/AndroidHiltModule.scala
```diff
@@ -0,0 +1,23 @@
...
+
+  lazy val millDiscover = Discover[this.type]
+
+  def toolsClasspath: Target[Seq[PathRef]] = Target {
+    millProjectModule("mill-main-android-hilt", repositories)
```
lefou, line 20: This should give you a deprecation warning. 

Instead use `Dep.millProjectModule` to get the dependency representing the module and use `jvmWorker().defaultResolver().classpath` to resolve it as usual.

scalalib/src/mill/javalib/android/AndroidSdkModule.scala
```diff
@@ -54,6 +54,18 @@ trait AndroidSdkModule extends Module {
     s"https://github.com/google/bundletool/releases/download/${bundleToolVersion()}/bundletool-all-${bundleToolVersion()}.jar"
   }
 
+  /** The version of the jetifier standalone */
+  def jetifierStandaloneVersion: T[String] = "1.0.0-beta10"
```
lefou, line 58: Code smell: hard-coded version.

lefou (commented):
I just took a look at the changes to `KotlinModule` and Co. and commented below. Sorry, if this is premature, but I hope it helps. 

Can you add some comments (here and probably also to the code) about pre-compiled classes? What are these? Who is supposed to create them? Is this a concept you invented or does is it state of the art? Are there any references? Why are they handled specially and not part of the compile classpath?

vaslabs:
> I just took a look at the changes to `KotlinModule` and Co. and commented below. Sorry, if this is premature, but I hope it helps.
> 
> Can you add some comments (here and probably also to the code) about pre-compiled classes? What are these? Who is supposed to create them? Is this a concept you invented or is it state of the art? Are there any references? Why are they handled specially and not part of the compile classpath?

thanks, no it definitely helps, a lot are just leftovers from me trying to reverse engineer gradle and make things work

main/src/mill/main/AndroidHiltModule.scala
vaslabs: I've removed the module from main, it didn't make sense to me to be there, and moved it to the hilt module file which I think now resolves the dependencies properly (my tests were also failing because of this)

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
vaslabs: I've removed that, the compiled classpath should work now properly, it was an early effort to make things work quickly without cyclic calls and stack overflow, but I separated the kspClasspath from the compile classpath now

vaslabs:
https://github.com/user-attachments/assets/e2deb093-55b4-410e-ae80-ff1b3300cb8e

I've reached a point where the pipeline is green , so this should be ready for a review. I've also cleaned up a few things, thanks to @lefou 's comments

The video is me manually testing the example app on the latest changes

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
lefou: I still see `def kotlinPrecompiledClasses` and it's use sites.

kotlinlib/src/mill/kotlinlib/android/AndroidHiltSupport.scala
```diff
@@ -0,0 +1,331 @@
...
+
+@mill.api.experimental
+trait AndroidHiltSupport extends KspModule with AndroidAppKotlinModule {
+
+  override def sources: T[Seq[PathRef]] = Task { Seq.empty[PathRef] }
+
+  override def kspSources: T[Seq[PathRef]] = Task { super.sources() }
+
+  override def generatedSources: T[Seq[PathRef]] =
+    super[AndroidAppKotlinModule].generatedSources
```
lefou, line 25: Don't reused tasks, use their results.

```suggestion
    super[AndroidAppKotlinModule].generatedSources()
```

kotlinlib/src/mill/kotlinlib/android/AndroidHiltSupport.scala
```diff
@@ -0,0 +1,331 @@
...
+            )
+          else
+            Seq(dep)
+      }
+    )
+  }
+
+  override def kspPluginsResolved: T[Seq[PathRef]] = Task {
+    super.kspPluginsResolved()
+  }
```
lefou, line 47: this looks superfluous.

kotlinlib/src/mill/kotlinlib/android/AndroidHiltSupport.scala
```diff
@@ -0,0 +1,331 @@
...
+
+  override def kotlinSymbolProcessorsResolved: T[Seq[PathRef]] = Task {
+    super.kotlinSymbolProcessorsResolved()
+  }
```
lefou, line 51: Looks superfluous.

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -76,16 +76,61 @@ trait KspModule extends KotlinModule { outer =>
...
+  /**
+   * The classpath when running Kotlin Symbol processing
+   *
+   * For more info go to [[https://kotlinlang.org/docs/ksp-command-line.html]]
+   */
+  def kspClasspath: T[Seq[PathRef]] = Task {
+    super.compileClasspath()
+  }
+
+  override def kotlinLanguageVersion: T[String] = "1.9"
```
lefou, line 106: Why is this hard-coded here? If we only support "1.9", we should check it where we expect it to be "1.9". But silently overriding a possibly already configured values (via a trait mixed in before `KspModule` is bad style and hard to debug for users).

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -76,16 +76,61 @@ trait KspModule extends KotlinModule { outer =>
...
+
+  /**
+   * Kotlinc arguments used with KSP
+   * @return
+   */
+  def kspKotlinc: T[Seq[String]] = Seq(
```
lefou, line 112:
```suggestion
  def kspKotlincOptions: T[Seq[String]] = Seq(
```

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -76,16 +76,61 @@ trait KspModule extends KotlinModule { outer =>
...
+
   /**
    * The Kotlin compile task with KSP.
    * This task should run as part of the [[generatedSources]] task to
    * so that the generated  sources are in the [[compileClasspath]]
    * for the main compile task.
    */
-  def generateSourcesWithKSP: Target[Seq[PathRef]] = Task {
-    val sourceFiles = sources().map(_.path)
+  def generateSourcesWithKSP: Target[GeneratedKSPSources] = Task {
```
lefou, line 130:
```suggestion
  def generatedSourcesWithKSP: Target[GeneratedKSPSources] = Task {
```

kotlinlib/src/mill/kotlinlib/android/AndroidHiltSupport.scala
```diff
@@ -0,0 +1,331 @@
...
+import mill.kotlinlib.ksp.KspModule
+import mill.kotlinlib.worker.api.KotlinWorkerTarget
+import mill.scalalib.{Dep, JvmWorkerModule}
+import mill.{T, Task}
+import os.Path
+
+import java.io.File
+
+@mill.api.experimental
+trait AndroidHiltSupport extends KspModule with AndroidAppKotlinModule {
```
lefou, line 18: This trait / Mill module should have some docs.

lefou (commented):
Leaving a partial review here to avoid loosing comment due to updates.

kotlinlib/src/mill/kotlinlib/android/AndroidKotlinModule.scala
```diff
@@ -8,6 +8,23 @@ import mill.kotlinlib.{Dep, DepSyntax, KotlinModule}
...
+        "-P",
+        "plugin:androidx.compose.compiler.plugins.kotlin:generateFunctionKeyMetaClasses=false," +
+          "plugin:androidx.compose.compiler.plugins.kotlin:traceMarkersEnabled=true",
+        "-verbose",
+        "-opt-in=kotlin.RequiresOptIn",
+        "-opt-in=kotlin.Experimental",
+        "-Xallow-unstable-dependencies"
+      )
+    } else Seq.empty
+  }
```
lefou, line 26: Something is fishy here. We add a plugin but we don't bring it it. We already set up inherited `kotlincPluginJars` was most likely applied already.

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
vaslabs: the default is empty, and 1.9 is one that works with KSP . If it's not provided, KSP stops working

kotlinlib/src/mill/kotlinlib/android/AndroidHiltSupport.scala
```diff
@@ -0,0 +1,331 @@
...
+
+  override def sources: T[Seq[PathRef]] = Task { Seq.empty[PathRef] }
```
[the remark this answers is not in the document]
vaslabs: it's a hack for now, because we compile the sources during the hilt transformations and then do an asm transform. If we don't override, the compiler will bring the original compilation in.

The hilt flow is just - pick the sources, generate code, compile code, transform asm and then compile the rest.

So one solution to avoid overriding the sources, would be to override the compile task. Do you have any more suggestions?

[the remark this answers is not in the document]
vaslabs: or maybe override allSources by not including sources() to it?

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
vaslabs: I think I can just do, kspKotlinLanguageVersion

kotlinlib/src/mill/kotlinlib/android/AndroidHiltSupport.scala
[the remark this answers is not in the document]
lefou: The ultimate source for compilation is `allSourceFiles` in most cases and sometimes more specific versions of it, e.g. module like `KotlinModule` also support Java compilation, so they have `allKotlinSourceFiles` and `allJavaSourceFiles`.

If you think you need to further customize that list, but it is in most cases almost the same as `allSourceFiles`, you should start with your own input task, e.g. `allHiltSourceFiles` and derive it from the value of `allSourceFiles` or `allKotlinSourceFiles`. Either by filtering, or by inspecting and erroring out if something is inconsistent or unexpected. 

You should not modify the outcomes of tasks that are already important to other parts of the toolchain. E.g. `allSourceFiles` and `allKotlinSourceFiles` is still needed for the proper compilation (after the HILT toolchain ran). 

If that doesn't match you expectation, you might want to start with describing how HILT tooling is meant to work and we can find the best architecture for it. E.g. just hypothetically, if you need to customize a full `KotlinModule` workflow just for HILT, but then still need to run another `KotlinModule` workflows for the outer/other files combined with the result of the HILT processor, you probably want to model it as dedicated sub-module (`hilt`), which outcome contributes to the outer `generatedSources`, or somehting like this. In such a case, you have full control over the inner `hilt` module and can customize as you wish.

There are other examples in Mill where we do something equivalent. For example the `ScoverageModule` defines an inner `scoverage` module which copies most of it's setup from the outer module, but customizes the compiler setting by adding a compiler plugin specifically. Another example is the `Bloop` contrib module, which (implicitly) defines an inner `bloop` module for each `JavaModule` to hold additional computed data.

kotlinlib/src/mill/kotlinlib/android/AndroidHiltSupport.scala
```diff
@@ -0,0 +1,345 @@
...
+   * on the compiled sources, so we don't want to recompile them.
+   */
+  override def allSources: T[Seq[PathRef]] = generatedSources()
+
+  /**
+   * In the case of hilt, we are not adding any generated
+   * KSP files in generated sources - yet
+   */
+  override def generatedSources: T[Seq[PathRef]] =
+    super[AndroidAppKotlinModule].generatedSources()
```
lefou, line 41: Do we really need to specify a specific `super` here? Can you add a comment what issue this solves. Why isn't a normal `super.generatedSources()` enough?

kotlinlib/src/mill/kotlinlib/android/AndroidHiltSupport.scala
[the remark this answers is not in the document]
vaslabs: I see what you mean, I have an idea to make this better. basically right now we have:

image

This in the future, will have a few issues that I was planning to solve later. For example, the generated sources need to be part of the compilation, so that the IDE sees them. I'll try to transform this to:

image

We'll compile the kotlin classes twice, but I think that should be ok, as the first part will be during the generation process, so generated java files that reference types in Kotlin sources will work and do the transform last, by overriding compile and doing `super.compile()` + `transformASM`

I'm not 100% it will work but I can give it a go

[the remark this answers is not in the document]
vaslabs: your comment @lefou was extremely helpful, I reached to all that point and stages by following gradle to replicate the process trying to make things work!

I think now, it is much simpler, I've removed all the redundant tasks and it looks like just the KSP and Hilt in classpath + the transformASM at the end it's all that's needed, and this should also solve the generated sources

kotlinlib/src/mill/kotlinlib/android/AndroidHiltSupport.scala
vaslabs: this is resolved after the latest simplification

kotlinlib/src/mill/kotlinlib/android/AndroidHiltSupport.scala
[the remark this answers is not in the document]
vaslabs: I've retested it and it's working, please let me know if you are happy with the approach

lihaoyi:
As mentioned earlier, can we drop the `example/android/kotlinlib/5-todo-architecture-sample` app, or at least move it to `example/thirdparty`? I'm not against vendoring code sometimes, but vendoring an entire 4kLOC android codebase seems a bit much

vaslabs:
> Yeah thirdparty should work; anyone who runs `./mill example.thirdparty.androidtodo.server.local` or whatever can look inside their `out/` folder to see the unpacked sources

can you point me to an example where I can configure the third party to get the sources?

vaslabs:
> > Yeah thirdparty should work; anyone who runs `./mill example.thirdparty.androidtodo.server.local` or whatever can look inside their `out/` folder to see the unpacked sources
> 
> can you point me to an example where I can configure the third party to get the sources?

ah this is what I'm looking for (I think) https://github.com/com-lihaoyi/mill/blob/main/example/package.mill#L300

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
lefou: I changed the defaults in the meantime. https://github.com/com-lihaoyi/mill/pull/4890

vaslabs: ok, I tried it, but still doesn't seem to work without 1.9 source compatibility

vaslabs:
This should be ready unless you want to have the androidtodo thirdparty test running in the pipeline

vaslabs:
I've resolved the conflicts and retested. Let me know if there are things you need to change @lefou @lihaoyi , otherwise we can merge it so I can move to the next thing

lihaoyi:
Will take a look!

lihaoyi:
I think this looks good. Happy to merge it unless @lefou has any other comments

lefou:
> I think this looks good. Happy to merge it unless @lefou has any other comments

Please wait with merging this, I want to review the latest changes. I don't think we want the new `androidhilt` module under `main`, esp. since it depends on `kotlinlib`.

kotlinlib/src/mill/kotlinlib/android/AndroidHiltSupport.scala
```diff
@@ -0,0 +1,122 @@
...
+
+  }
+
+  def hiltProcessorClasspath: T[Seq[PathRef]] = Task {
+    kspApClasspath() ++ kspClasspath()
+  }
+
+}
+
+object AndroidHiltTransform extends ExternalModule with JvmWorkerModule {
```
lefou, line 107: This module should have it's own source file and some ScalaDoc.

kotlinlib/src/mill/kotlinlib/android/AndroidHiltSupport.scala
```diff
@@ -0,0 +1,122 @@
...
+    super.repositoriesTask() :+ AndroidSdkModule.mavenGoogle
+  }
+
+  def toolsClasspath: T[Seq[PathRef]] = Task {
+    defaultResolver().classpath(
+      Seq(
+        Dep.millProjectModule("mill-main-androidhilt")
+      )
+    )
+  }
```
lefou, line 119: It's a bit strange to have a `toolsClasspath` here, but no code using it. This is hard to understand and more important, hard to maintian.

I'd expect at least some `Task.Anon` or a `Worker` using this classpath.

kotlinlib/src/mill/kotlinlib/android/AndroidHiltSupport.scala
```diff
@@ -0,0 +1,122 @@
...
+
+  /**
+   * Transforms the Kotlin classes with Hilt dependency injection context
+   * and returns the new path of the kotlin compiled classpath. This uses
+   * the [mill.main.android.hilt.AndroidHiltTransformAsm] that uses
+   * the hilt gradle plugin and the android build tools
+   */
+  private def androidHiltTransformAsm(
```
lefou, line 76: This should most likely be a `Task.Anon` residing in `AndroidHiltTransform` module. The `toolsClasspath` param can then be a normal task reference.

kotlinlib/src/mill/kotlinlib/android/AndroidHiltSupport.scala
lefou: It should also be a `trait` and a `object`, so that users have a chance to customize.

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -77,16 +77,64 @@ trait KspModule extends KotlinModule { outer =>
...
     "com.google.devtools.ksp.symbol-processing"
 
+  /**
+   * The KSP ap classpath
+   *
+   * For more info go to [[https://kotlinlang.org/docs/ksp-command-line.html]]
+   */
+  def kspApClasspath: T[Seq[PathRef]] = Task {
+    kotlinSymbolProcessorsResolved()
+  }
```
lefou, line 87: Should we add a call to the `super` task here?

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -77,16 +77,64 @@ trait KspModule extends KotlinModule { outer =>
...
+
+  /**
+   * The sources for being used in KSP, in case
+   * the user wants to separate KSP specific sources
+   * from others. Defaults to [[sources]] (i.e. no splitting)
+   */
+  def kspSources: T[Seq[PathRef]] = Task {
+    sources()
+  }
```
lefou, line 96: Should we add a call to the `super` task here?

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -77,16 +77,64 @@ trait KspModule extends KotlinModule { outer =>
...
+   */
+  def kspClasspath: T[Seq[PathRef]] = Task {
+    super.compileClasspath()
+  }
+
+  /**
+   * Kotlinc arguments used with KSP
+   * @return
+   */
+  def kspKotlincOptions: T[Seq[String]] = Task {
```
lefou, line 111: Should we add a call to the `super` task here?

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -77,16 +77,64 @@ trait KspModule extends KotlinModule { outer =>
...
+    Seq(
+      "-Xallow-unstable-dependencies",
+      "-no-reflect",
+      "-no-stdlib",
+      "-language-version",
+      kotlinLanguageVersion()
+    )
+  }
+
+  def kspPluginParameters: T[Seq[String]] = Task {
```
lefou, line 124: Should we add a call to the `super` task here?

main/androidhilt/src/mill/main/android/hilt/AndroidHiltTransformAsm.scala
```diff
@@ -0,0 +1,96 @@
...
+
+  def main(args: Array[String]): Unit = {
+
+    val scanDirectory = os.Path(args.head)
+
+    val destination = os.Path(args.last)
+
+    transformAsm(scanDirectory, os.walk(scanDirectory).filter(_.ext == "class"), destination)
+
+  }
```
lefou, line 18: This is a uncommon entry-point for a Mill "worker". It this because it can run stand-alone (without any Mill infrastructure)? Please add some Scaladoc and include what `args` are supported/expected.

main/package.mill
```diff
@@ -111,5 +112,16 @@ object `package` extends RootModule with build.MillStableScalaModule {
...
+  object androidhilt extends build.MillPublishScalaModule {
+    def repositoriesTask = Task.Anon {
+      super.repositoriesTask() :+ AndroidSdkModule.mavenGoogle
+    }
+    def moduleDeps = Seq(build.main, build.kotlinlib)
+    def ivyDeps = Agg(
+      build.Deps.hiltGradlePlugin,
+      build.Deps.androidTools
+    )
+  }
```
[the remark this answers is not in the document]
vaslabs: I was mainly following the example suggested [here](https://github.com/com-lihaoyi/mill/pull/4759#issuecomment-2747276678)

I can move it to kotlinlib  but I have a few questions:

This is just a mini executable which transforms any class that has hilt annotations by re-writing the byte code to point to hilt generated classes (e.g. Hilt_Activity instead of ComponentActivity)

It is not to be extended by users.

Where should I put it in kotlinlib? In a new module? Given that it depends to gradle api and hilt gradle plugin which we don't want to pull if the users don't use hilt.

Shall I create a new module under kotlinlib? how should it be called, is there another example I can follow?

kotlinlib/src/mill/kotlinlib/android/AndroidHiltSupport.scala
vaslabs: This is used [here](https://github.com/com-lihaoyi/mill/pull/4759/files#diff-7f222b857b1c9388c66abfe3bd8a712120e9fe2298e9df9d4f23e56aee12dfadR60), it's the classpath of the transform asm executable. Is there a more idiomatic way of doing it? I pulled this from another external module that was doing something similar

main/package.mill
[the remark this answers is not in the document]
lefou: I think the module itself doesn't need to change. It just should reside as a sub-module of `kotlinlib` instead of `main`, since it's utility is bound to the Kotlin tooling.

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
lefou: I misread. Ignore this.

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
lefou: Can you add some docs here?

kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
lefou: No.

kotlinlib/src/mill/kotlinlib/android/AndroidHiltSupport.scala
```diff
@@ -0,0 +1,115 @@
...
+  def androidHiltModule = ModuleRef(AndroidHiltTransform)
+
+  /** Compile and then transform asm for Hilt DI */
+  override def compile: T[CompilationResult] = Task {
+    val transformClasses = androidHiltModule().androidHiltTransformAsm(
+      Task.Anon {
+        super.compile().classes
+      }
+    )()
+    CompilationResult(super.compile().analysisFile, transformClasses)
```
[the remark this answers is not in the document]
vaslabs: I think the kotlin one is creating a dummy file, we can create a dummy file here too and come back to it later?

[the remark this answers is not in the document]
lefou: Yeah, if we expect a dummy, we better make that explicit. Please use a dummy directly.

scalalib/src/mill/javalib/android/AndroidAppModule.scala
```diff
@@ -54,6 +55,25 @@ trait AndroidAppModule extends AndroidModule {
 
   private val parent: AndroidAppModule = this
 
+  def androidApplicationNamespace: String
+  def androidApplicationId: String
```
lefou, line 59: These two should have docs, even if they're almost self-explaining.

vaslabs:
> Can you please make sure, that you don't re-use tasks. If you return a (non-anon) task defined elsewhere directly when you define a task, The compiler can't insert it's implicit task wrapping, resulting in a incorrect task context (e.g. a wrong source location). I marked one of those places below, but there are more.

I made this mistake from the very beginning, I've fixed a few I could find quickly, but we have a few upcoming cleanup tasks, we can fix the rest in follow up PRs

kotlinlib/src/mill/kotlinlib/android/AndroidHiltSupport.scala
```diff
@@ -68,25 +68,17 @@ trait AndroidHiltSupport extends KspModule with AndroidAppKotlinModule {
...
     super.repositoriesTask() :+ AndroidSdkModule.mavenGoogle
   }
 
-  def toolsClasspath: T[Seq[PathRef]] = Task {
-    defaultResolver().classpath(
-      Seq(
-        Dep.millProjectModule("mill-kotlinlib-androidhilt")
-      )
-    )
-  }
```
lefou: We should keep this task, for multiple reasons:
* User can override to customize, e.g. inject their own patched tooling
* We can cache the classpath in one single location
* We can easily fetch the classpath as part of `prepareOffline` requests

vaslabs: ah I thought you wanted it removed, sorry, putting it back

lefou (approved):
Looks good to me. This has gone many rounds. Thank you @vaslabs! Let's see if CI is green and merge this.

Any further changes, e.g. proper support for `OfflineSupportModule`, can be done in separate PRs.

kotlinlib/androidhilt/src/mill/kotlinlib/android/hilt/AndroidHiltTransformAsm.scala
```diff
@@ -0,0 +1,109 @@
+package mill.kotlinlib.android.hilt
+
+import com.android.build.api.instrumentation.{ClassContext, ClassData}
+import dagger.hilt.android.plugin.transform.AndroidEntryPointClassVisitor
+import mill.*
```
lefou, line 5: Is this `mill`-import used anywhere?

vaslabs:
I'm already trying :)

kotlinlib/androidhilt/src/mill/kotlinlib/android/hilt/AndroidHiltTransformAsm.scala
vaslabs: no, good catch, I removed it with the rebase

vaslabs:
looks like the mill-bootstrap.patch breaks because I have changes in build.mill, I don't know how to fix that

lihaoyi:
@vaslabs just ignore the test-mill-bootstrap job for now, I can help fix it when it's time to merge

vaslabs:
> @vaslabs just ignore the test-mill-bootstrap job for now, I can help fix it when it's time to merge

this should be ready par the patch

lihaoyi:
Working on it

lihaoyi:
@vaslabs you need to give me permission to push to your pull request I think

vaslabs:
> @vaslabs you need to give me permission to push to your pull request I think

That option is supposed to be here

image


but I don't see it, I'll just give you access to my fork

lihaoyi:
ok pushed my changes, let's see if this works

vaslabs:
note: now that the android job is stable and runs in ~12 minutes, I'll follow with another PR to fully test hilt, but won't add more into this one, let's merge it if the pipeline is happy :sweat_smile:

lihaoyi:
Going to go ahead and merge this to unblock concurrent work

Issue #4659, opened 2025-03-05
`No main class specified or found` when overriding `zincWorker`


dontgitit:
Hello,

I put together a very simple hello world app, but experienced strange behavior once I added a `zincWorker` override; my main class could no longer be detected. The full sample app can be found [here](https://github.com/dontgitit/TestMillMain). Commenting out [this line](https://github.com/dontgitit/TestMillMain/blob/master/build.sc#L11) makes the app run as expected. 

Let me know if any additional info is needed.

lefou:
I can reproduce it.

`build.sc`
```scala
import mill.scalalib._
import mill.define.ModuleRef

object ZincWorkerJava21 extends ZincWorkerModule {
  def jvmId = "temurin:21.0.6"
}

object foo extends ScalaModule {
  def scalaVersion = "3.6.3"
  // commenting out the zincWorker makes the app run
  override def zincWorker = ModuleRef(ZincWorkerJava21)
  // alternatively, uncomment this for an explicit main class
  //  def mainClass = Some("Foo")
}
```

`.mill-version`
```
0.12.8
```

`foo/src/Foo.scala`
```scala
object Foo {
  def main(args: Array[String]): Unit = println("hello world")
}
```

```
> mill --version
Mill Build Tool version 0.12.8
Java version: 21.0.5, vendor: Eclipse Adoptium, runtime: /opt/openjdk-bin-21.0.5_p11
Default locale: de_DE, platform encoding: UTF-8
OS name: "Linux", version: 6.6.74-gentoo-x86_64, arch: amd64

> mill foo.run
[build.sc-61/65] compile
[build.sc-61] [info] compiling 1 Scala source to /tmp/mill-2/out/mill-build/compile.dest/classes ...
[build.sc-61] [info] done compiling
[58/66] foo.compile
[58] [info] compiling 1 Scala source to /tmp/mill-2/out/foo/compile.dest/classes ...
[58] [info] done compiling
[66/66] ============================== foo.run ============================== 62s
1 tasks failed
foo.finalMainClass No main class specified or found
```

PR #4760, opened 2025-03-21
Fix main class discovery in scala 3

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/4659

Tested manually, the problem reproduces reliably on `main` and goes away with this PR


lefou:
@lihaoyi Can you tell how you tested this o could you re-test it against `main`, since I replaced the main class discovery logic in #4822 ?

lefou:
@lihaoyi Nevermind, in #4659 is a reproducer.

lihaoyi:
@lefou I used your reproducer in https://github.com/com-lihaoyi/mill/issues/4659

Issue #4768, opened 2025-03-25
`ArrayIndexOutOfBoundsException` in `InputPumper`


lefou:
I saw it in a customers CI log. After it, the job was stalled.

Here are two occurences:

Log 1:

```
> ./mill --version
Mill Build Tool version 0.12.9
Java version: 17.0.14, vendor: Ubuntu, runtime: /usr/lib/jvm/java-17-openjdk-amd64
Default locale: en_US, platform encoding: UTF-8
OS name: "Linux", version: 5.15.0-134-generic, arch: amd64
...
[2469] [INFO] [2469] Exception in thread "/var/lib/jenkins/workspace/xxx-multibranch_mill-0.12/cloud/vaadin/frontend/xxx/STAGE/mvnw package stdout thread" java.lang.RuntimeException: java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 0
[2469] 	at mill.main.client.InputPumper.run(InputPumper.java:65)
[2469] 	at java.base/java.lang.Thread.run(Thread.java:840)
[2469] Caused by: java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 0
[2469] 	at fansi.Str.getColor(Fansi.scala:111)
[2469] 	at mill.util.LinePrefixOutputStream.writeOutBuffer(LinePrefixOutputStream.scala:45)
[2469] 	at mill.util.LinePrefixOutputStream.flush(LinePrefixOutputStream.scala:87)
[2469] 	at java.base/java.io.PrintStream.flush(PrintStream.java:423)
[2469] 	at mill.util.MultiStream$$anon$1.flush(MultiLogger.scala:138)
[2469] 	at java.base/java.io.PrintStream.flush(PrintStream.java:423)
[2469] 	at mill.main.client.InputPumper.run(InputPumper.java:57)
[2469] 	... 1 more
[2469/2475] ========================== docker.comfis-app.dockerBuild ========================= 60s
[2469] cloud.vaadin.frontend.xxx.STAGE.mvnPackage 58s
[2469/2475] ========================= docker.comfis-app.dockerBuild ========================= 120s
...
[2469/2475] ========================= docker.comfis-app.dockerBuild ========================= 840s
[2469] cloud.vaadin.frontend.xxx.STAGE.mvnPackage 838s
```

Log 2:

```
[...]

PR #4769, opened 2025-03-25
Fix potential ArrayIndexOutOfBoundsException

lefou:
Fix https://github.com/com-lihaoyi/mill/issues/4768

Ensure we don't use a negative index in case a string is empty.


lihaoyi (approved):
ship it. Maybe add a comment about when it happens? I guess when there's a empty line (i.e. two back-to-back newlines) with colors being applied before?

lefou:
Unfortunately, I don't know in which situations it can happen. I just added a check for a potential issue, since we don't check the length of `s` before we substract `1`.

lefou:
Could be the process (Maven) is itself messing with ansi sequences and the `ErrorMode.Sanitize` results in a resulting length of `0`.

lihaoyi:
Sounds good, go ahead then

PR #4771, opened 2025-03-25
Add Kotlin support for BuildInfo module

lefou:
Add support for Kotlin source file generator in `BuildInfo` module.

In addition to Java and Scala, `BuildInfo` now also supports a Kotlin source generator.

The default generator is selected based on the module type, but can now also be configured to a specific type. E.g. users of `ScalaModule` or `KotlinModule` might want to generate a Java source file. 

```scala
// Always generate a Java file
def buildInfoLanguage = BuildInfo.Language.Java
```

Or modules implementing `ScalaModule` and `KotlinModule` might want to generate a Kotlin file instead of a Scala file.

### DIFF
--- a/contrib/buildinfo/src/mill/contrib/buildinfo/BuildInfo.scala
+++ b/contrib/buildinfo/src/mill/contrib/buildinfo/BuildInfo.scala
@@ -240,7 +240,7 @@ object BuildInfo {
            |      .getResourceAsStream("${buildInfoObjectName}.buildinfo.properties")
            |
            |    if(buildInfoInputStream == null)
-           |      throw new RuntimeException("Could not load resource ${buildInfoObjectName}.buildinfo.properties")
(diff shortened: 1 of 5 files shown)


contrib/buildinfo/src/mill/contrib/buildinfo/BuildInfo.scala
```diff
@@ -88,7 +88,9 @@ trait BuildInfo extends JavaModule {
       )
 
       os.write(
-        Task.dest / buildInfoPackageName.split('.') / s"${buildInfoObjectName}.${buildInfoLanguage.ext}",
+        Task.dest / buildInfoPackageName.split(
+          '.'
+        ) / s"${buildInfoObjectName}.${buildInfoLanguage.ext}",
```
lefou, line 93:
```suggestion
        Task.dest / buildInfoPackageName.split('.') / 
          s"${buildInfoObjectName}.${buildInfoLanguage.ext}",
```

lihaoyi (approved):

Issue #4776, opened 2025-03-25
`KotlinModule` need first class plugin support


lefou:
We just want to provide plugin dependencies, everything else should be automatic. Analogue to how we do it in `ScalaModule`.

PR #4779, opened 2025-03-25
First-class support for Kotlin compiler plugins

lefou:
Fix https://github.com/com-lihaoyi/mill/issues/4776

This introduces multiple new tasks, analog to the ones in `ScalaModule`.

* `kotlincPluginIvyDeps` - Dependencies resembling compiler plugins
* `kotlincPluginJars` - The coursier-resolved jars of the compiler plugins; resolved intransitively; also usable to add plugins not available via coursier/Maven
* `mandatoryKotlincOptions` - compiler options derived from other module settings, like Kotlin versions and compiler plugin enablement
* `allKotlincOptions` - all compiler options

Revised the existing modules and used the new tasks to provide Kotlin compiler plugins and their configuration. Removed `composeProcessor` and `kotestProcessor` tasks.

Also added some comments for stuff that need further work like hard-coded versions.


lefou:
I'd like to add some documentation how to use Kotlin compiler plugins. Since this is analogue to our Scala setup, I wonder where we have the Scala documentation for plugins and if I can share some bits.

kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -153,6 +155,23 @@ trait KotlinModule extends JavaModule { outer =>
...
+   */
+  def kotlinCompilerPluginJars: T[Seq[PathRef]] = Task {
+    val jars = defaultResolver().resolveDeps(
+      kotlinCompilerPluginIvyDeps()
+        // Don't resolve transitive jars
+        .map(d => d.exclude("*" -> "*"))
+    )
+    jars.toSeq
+  }
+
```
lefou, line 174: I'd rather use the shorter `kotlinc` prefix instead of `kotlinCompiler`. We use both prefixes in the modules though.

lihaoyi (approved):

PR #4784, opened 2025-03-26
Prefer `defaultResolver().classpath` over `Lib.resolveDependencies` (backport #4727)

alexarchambault:
Whenever possible, this tries to make modules defined in Mill use `CoursierModule#defaultResolver` or `CoursierModule#millResolver`  to resolve dependencies, rather than calling lower level helpers like `Lib. resolveDependencies`. The former automatically take into account customizations of resolution defined on `CoursierModule` and `JavaModule` (`repositoriesTask`, `mapDependencies`, etc.), while these have to be passed manually to `Lib.resolveDependencies`.

This helps when adding more parameters - these have to be passed to `defaultResolver` / `millResolver`, and that's it, rather than also having to inspect all calls to `Lib.resolveDependencies`, and passing the new parameter there too.

This also tries to make resolution log to the task logger of the task requesting the resolution, rather than the logger of `CoursierModule#{defaultResolver,millResolver}`.

---

Original pull request: https://github.com/com-lihaoyi/mill/pull/4727 

Pull request: https://github.com/com-lihaoyi/mill/pull/4784

### DIFF
--- a/scalalib/src/mill/scalalib/CoursierModule.scala
+++ b/scalalib/src/mill/scalalib/CoursierModule.scala
@@ -58,7 +58,7 @@ trait CoursierModule extends mill.Module {
    * A `CoursierModule.Resolver` to resolve dependencies.
    *
    * Can be used to resolve external dependencies, if you need to download an external
-   * tool from Maven or Ivy repositories, by calling `CoursierModule.Resolver#resolveDeps`.
+   * tool from Maven or Ivy repositories, by calling `CoursierModule.Resolver#classpath`.
    *
    * @return `CoursierModule.Resolver` instance
    */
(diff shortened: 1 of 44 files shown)


scalalib/src/mill/scalalib/CoursierModule.scala
```diff
@@ -318,6 +368,10 @@ object CoursierModule {
      * @param resolutionParams coursier resolution parameters
      * @return dependencies with empty version filled
      */
+    @deprecated(
+      "Use resolution instead, and extract the values you're interested in from it",
+      "Mill after 0.12.9"
```
lefou, line 373: We already know the next version.

```suggestion
      "Mill 0.12.10"
```

alexarchambault: Should be fixed

PR #4785, opened 2025-03-26
Fix Android Example Tests

gamlerhart:
The tests now run concurrently with workers,
therefore the `out.json` is separated per worker.

### DIFF
--- a/example/android/javalib/1-hello-world/build.mill
+++ b/example/android/javalib/1-hello-world/build.mill
@@ -108,7 +108,7 @@ object app extends AndroidAppModule {
 > ./mill show app.test
 ...compiling 2 Java source...
 
-> cat out/app/test/testForked.dest/out.json
+> cat out/app/test/testForked.dest/worker-0/out.json
 ["",[{"fullyQualifiedName":"com.helloworld.ExampleUnitTest.textSize_isCorrect","selector":"com.helloworld.ExampleUnitTest.textSize_isCorrect","duration":...,"status":"Success"}]]
 
 */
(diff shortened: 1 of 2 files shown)


example/android/javalib/1-hello-world/build.mill
```diff
@@ -108,7 +108,7 @@ object app extends AndroidAppModule {
 > ./mill show app.test
 ...compiling 2 Java source...
 
-> cat out/app/test/testForked.dest/out.json
+> cat out/app/test/testForked.dest/worker-0/out.json
```
gamlerhart, line 111: My question is, how did it work in `main`? It maybe passed due to selective tests or something?

lefou: Most likely selective tests. I saw it fail very often too, mostly in PR run.

lihaoyi: I think they were broken, and @alexarchambault  disabled them by adding a `ignoreErrorsOnCI` marker file. Could you try removing the file in this PR and see if they pass?

vaslabs: The timeout I see in the last run is quite common, last time I was trying to fix it , my hypothesis was running the emulator task in parallel with other tasks was causing slowness.

We could have a simpler test for the java example, without the emulator and keep only the kotlin one, or try to force the android tasks to run sequentially again

gamlerhart: Removed the markers, build is green.

lefou (approved):

alexarchambault:
It seems the test that was re-enabled is erroring again, see [here](https://github.com/com-lihaoyi/mill/actions/runs/14172102833/job/39701764805?pr=4800) for example (with two runs failing in a row)

gamlerhart:
Hmm...this is indeed a different failure. This pull request fixed the failing because the reference files are not found. 

This failure seems more to do with the emulator. Should we disable these example tests again?

PR #4790, opened 2025-03-27
Add `Dep.millProjectModule` to refer to mill project modules

lefou:
This is meant as a replacement for `mill.util.ModuleUtils.millProjectModule`.

The main motivation for this change is to handle mill module dependencies more regular. Since they can be resolved together with other dependencies, this should result in better conflict resolution. The older API rather forces us to concat two potential redundant or conflicting classpaths. 

There is one last usage of the older `ModuleUtils.millProjectModule` from `VisualizeModule`, which resides in `mill.main` and can't access `CoursierModule` or `Dep`. Don't know how to best handle it.

### DIFF
--- a/bsp/src/mill/bsp/BSP.scala
+++ b/bsp/src/mill/bsp/BSP.scala
@@ -4,15 +4,14 @@ import mill.api.{Ctx, PathRef}
 import mill.{T, Task, given}
 import mill.define.{Command, Discover, Evaluator, ExternalModule}
 import mill.main.BuildInfo
-import mill.util.MillModuleUtil.millProjectModule
-import mill.scalalib.CoursierModule
+import mill.scalalib.{CoursierModule, Dep}
 
 object BSP extends ExternalModule with CoursierModule {
 
   lazy val millDiscover = Discover[this.type]
 
   private def bspWorkerLibs: T[Seq[PathRef]] = Task {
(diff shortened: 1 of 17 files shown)


alexarchambault (approved):

lihaoyi (approved):
If the tests pass looks good to me I guess, this whole `millModule` thing is purely to support testing using local classes

lihaoyi:
> There is one last usage of the older ModuleUtils.millProjectModule from VisualizeModule, which resides in mill.main and can't access CoursierModule or Dep. Don't know how to best handle it.

That's actually an interesting question to ask. Maybe we should have a "global" `defaultResolver()` to use when you are not part of any `CoursierModule`? That would be useful in cases such as `PalantirFormatBaseModule`  where the companion object is a  `ExternalModule`s  inherits from `CoursierModule` to get access to `defaultResolver()`, but really isn't a `CoursierModule` at all in the sense that the `JavaModule`/`ScalaModule`s in your build config are.

lefou:
@lihaoyi That sound like a good idea. Since coursier is currently part of Mill core anyway, it should work. But it means we also need to move `Dep` an maybe some other class up. Maybe into `mill.api`. I think I will leave the deprecated call for now, so we can discuss and refactor in a follow up.

lefou:
The one test failing in CI is running successfully on my machine. Merging.

PR #4792, opened 2025-03-27
Log command to be run in Jvm.callProcess when debugging is enabled

alexarchambault:
(Context is I'm currently having issues running specific tests on the main branch, and I was hoping `--debug` would help me, but it doesn't 😅)


lefou (commented):
Looks good to me. I thought we already had it. Probably in the (now removed) `runSubprocess` version.

PR #4800, opened 2025-03-28
Use system calls to get terminal size on Linux / Mac

alexarchambault:
This makes Mill use my [native-terminal](https://github.com/alexarchambault/native-terminal) library (formerly windows-ansi) to get the terminal size. That way, Mill doesn't run `tput` commands every ~100 ms to query the terminal size.

A few things to note, about JVM warnings first:
- native-terminal relies on [jline-native](https://github.com/jline/jline3/tree/114e9a8f86102245ed9e2e642603f97e11ac962b/native) to make native calls. jline-native itself relies on JNI to make native calls
- loading JNI libraries works both on the JVM and from native-images
- on recent JVM versions (>= 22), using JNI can print warnings at runtime, like
```text
WARNING: A restricted method in java.lang.System has been called
WARNING: java.lang.System::load has been called by org.fusesource.jansi.internal.JansiLoader in an unnamed module (file:~/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/fusesource/jansi/jansi/2.4.1/jansi-2.4.1.jar)
WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled
```
- it seems we don't get those warnings from native-image. When spawning JVMs, we could silence those warnings with `--enable-native-access=ALL-UNNAMED`. We should pass add that option when needed, but we don't yet.

About the optimization to load the jline-native native library faster:

By default, jline-native loads its JNI library in a somewhat slow way (it has native libraries for all major platforms as resources, it writes the one of the current system in a temporary location, and loads it from there - it does this upon initialization in every new JVM process). To work around that slow init process, we write the native library ourselves under `~/.cache/mill/jline/${jline-version}/…`. That way, upon initialization, if we find the right native library there, we just load it. So if the right native library is already on disk, initialization is fast, no new file is written on disk.

We use reflection to tell jline-native that it's initialized, and doesn't need to load its native library on its own. We could have it load the native library that we wrote ourselves (we set the `library.jline.path` and `library.jline.name` Java properties for that), but it would still [spawn a "clean-up" thread](https://github.com/jline/jline3/blob/114e9a8f86102245ed9e2e642603f97e11ac962b/native/src/main/java/org/jline/nativ/JLineNativeLoader.java#L72), that would slow things down. So we use reflection to tell jline-native to not do anything.

---

Same as https://github.com/com-lihaoyi/mill/pull/4497, but against the `main` branch - hoping CI would go smoother here

### DIFF
--- a/integration/ide/bsp-server/resources/snapshots/workspace-build-targets.json
+++ b/integration/ide/bsp-server/resources/snapshots/workspace-build-targets.json
@@ -26,7 +26,7 @@
       },
       "dataKind": "jvm",
       "data": {
-        "javaHome": " java-home",
+        "javaHome": "java-home",
         "javaVersion": "<java-version>"
       }
     },
@@ -188,8 +188,7 @@
           "file:///coursier-cache/https/repo1.maven.org/maven2/org/jline/jline-terminal/<jline-terminal-version>/jline-terminal-<jline-terminal-version>.jar",
           "file:///coursier-cache/https/repo1.maven.org/maven2/org/jline/jline-terminal-jna/<jline-terminal-jna-version>/jline-terminal-jna-<jline-terminal-jna-version>.jar",
           "file:///coursier-cache/https/repo1.maven.org/maven2/org/scala-lang/scala-library/<scala-version>/scala-library-<scala-version>.jar",
-          "file:///coursier-cache/https/repo1.maven.org/maven2/org/scala-sbt/util-interface/<util-interface-version>/util-interface-<util-interface-version>.jar",
-          "file:///coursier-cache/https/repo1.maven.org/maven2/org/jline/jline-native/<jline-native-version>/jline-native-<jline-native-version>.jar"
+          "file:///coursier-cache/https/repo1.maven.org/maven2/org/scala-sbt/util-interface/<util-interface-version>/util-interface-<util-interface-version>.jar"
         ],
         "jvmBuildTarget": {
           "javaHome": "java-home",
--- a/integration/ide/gen-idea/resources/extended/idea/libraries/scala_SDK_3_6_2.xml
+++ b/integration/ide/gen-idea/resources/extended/idea/libraries/scala_SDK_3_6_2.xml
@@ -3,7 +3,6 @@
         <properties>
             <language-level>Scala_3_6</language-level>
             <compiler-classpath>
-                <root url="file://$USER_HOME$/COURSIER_CACHE/v1/https/repo1.maven.org/maven2/org/jline/jline-native/3.27.0/jline-native-3.27.0.jar"/>
                 <root url="file://$USER_HOME$/COURSIER_CACHE/v1/https/repo1.maven.org/maven2/org/jline/jline-reader/3.27.0/jline-reader-3.27.0.jar"/>
                 <root url="file://$USER_HOME$/COURSIER_CACHE/v1/https/repo1.maven.org/maven2/org/jline/jline-terminal-jna/3.27.0/jline-terminal-jna-3.27.0.jar"/>
                 <root url="file://$USER_HOME$/COURSIER_CACHE/v1/https/repo1.maven.org/maven2/org/jline/jline-terminal/3.27.0/jline-terminal-3.27.0.jar"/>
--- a/integration/ide/gen-idea/resources/hello-idea/idea/libraries/scala_SDK_3_6_2.xml
+++ b/integration/ide/gen-idea/resources/hello-idea/idea/libraries/scala_SDK_3_6_2.xml
@@ -3,7 +3,6 @@
         <properties>
             <language-level>Scala_3_6</language-level>
             <compiler-classpath>
-                <root url="file://$USER_HOME$/COURSIER_CACHE/v1/https/repo1.maven.org/maven2/org/jline/jline-native/3.27.0/jline-native-3.27.0.jar"/>
                 <root url="file://$USER_HOME$/COURSIER_CACHE/v1/https/repo1.maven.org/maven2/org/jline/jline-reader/3.27.0/jline-reader-3.27.0.jar"/>
                 <root url="file://$USER_HOME$/COURSIER_CACHE/v1/https/repo1.maven.org/maven2/org/jline/jline-terminal-jna/3.27.0/jline-terminal-jna-3.27.0.jar"/>
                 <root url="file://$USER_HOME$/COURSIER_CACHE/v1/https/repo1.maven.org/maven2/org/jline/jline-terminal/3.27.0/jline-terminal-3.27.0.jar"/>
--- a/integration/ide/gen-idea/src/GenIdeaExtendedTests.scala
+++ b/integration/ide/gen-idea/src/GenIdeaExtendedTests.scala
@@ -19,6 +19,7 @@ object GenIdeaExtendedTests extends UtestIntegrationTestSuite {
 
       for (resource <- resources) {
         GenIdeaUtils.assertIdeaXmlResourceMatchesFile(
+          tester.workspaceSourcePath,
           workspacePath,
           resource
         )
--- a/integration/ide/gen-idea/src/GenIdeaTests.scala
+++ b/integration/ide/gen-idea/src/GenIdeaTests.scala
@@ -12,12 +12,13 @@ object GenIdeaTests extends UtestIntegrationTestSuite {
   def tests: Tests = Tests {
     test("genIdeaTests") - integrationTest { tester =>
       import tester._
-      val expectedBase = workspacePath / "idea"
+      val expectedBase = tester.workspaceSourcePath / "idea"
       val resources = os.walk(expectedBase).filter(os.isFile).map(_.subRelativeTo(expectedBase))
 
       eval("mill.idea.GenIdea/", check = true, stdout = os.Inherit, stderr = os.Inherit)
 
-      for (resource <- resources) assertIdeaXmlResourceMatchesFile(workspacePath, resource)
+      for (resource <- resources)
+        assertIdeaXmlResourceMatchesFile(tester.workspaceSourcePath, workspacePath, resource)
     }
   }
 
--- a/integration/ide/gen-idea/src/GenIdeaUtils.scala
+++ b/integration/ide/gen-idea/src/GenIdeaUtils.scala
@@ -7,36 +7,52 @@ import utest.assert
 object GenIdeaUtils {
 
   /**
+   * Set this to true to update the gen-idea tests snapshot data
+   */
+  def updateResources: Boolean = false
+
+  /**
    * The resource content will be loaded from the classpath and matched against the file.
    * It may contain the `<!-- IGNORE -->` String, to simulate wildcard-matches.
    */
   def assertIdeaXmlResourceMatchesFile(
+      workspaceSourcePath: os.Path,
       workspacePath: os.Path,
       resource: os.SubPath
   ): Unit = {
-    val expectedResourcePath = workspacePath / "idea" / resource
+    val expectedResourcePath = workspaceSourcePath / "idea" / resource
     val actualResourcePath = workspacePath / ".idea" / resource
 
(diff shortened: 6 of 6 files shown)


integration/migrating/init/src/MillInitMavenTests.scala
```diff
@@ -40,7 +40,8 @@ object MillInitMavenJansiTests extends BuildGenTestSuite {
         testRes.isSuccess
       )
 
-      val publishLocalRes = eval("publishLocal")
+      val ivy2Repo = tester.baseWorkspacePath / "ivy2Local"
+      val publishLocalRes = eval(("publishLocal", "--localIvyRepo", ivy2Repo.toString))
```
lihaoyi, line 44: Is this change necessary?

runner/client/src/mill/runner/client/MillProcessLauncher.java
```diff
@@ -274,15 +275,54 @@ static int getTerminalDim(String s, boolean inheritError) throws Exception {
...
+    boolean canUse;
+    if (mill.constants.Util.hasConsole()) {
+      try {
+        NativeTerminal.getSize();
+        canUse = true;
+      } catch (Throwable ex) {
+        canUse = false;
+      }
+    } else canUse = false;
+
```
lihaoyi, line 313: Most of this logic seems like it should be moved into `JansiLoader`

runner/src/mill/runner/MillMain.scala
```diff
@@ -432,6 +433,10 @@ object MillMain {
     for ((k, v) <- desiredProps) System.setProperty(k, v)
   }
 
+  def adjustLogbackOutputFile(): Unit =
+    if (!sys.props.contains("mill.log.file"))
+      sys.props("mill.log.file") = s"${OutFiles.out}/mill.log"
```
lihaoyi, line 438: Could you leave a comment about what this is for?

lihaoyi (approved):
Looks good to me, seems to work fine in my manual testing

runner/client/src/mill/runner/client/JansiLoader.java
```diff
@@ -0,0 +1,50 @@
+package mill.runner.client;
+
+import java.io.File;
+import java.io.IOException;
+import org.fusesource.jansi.internal.OSInfo;
+
+final class JansiLoader {
```
lihaoyi, line 7: Let's add a top-level comment in this file explaining what we're doing here: we're loading Jansi and extracting the necessary `.dylib` or `.jnilib`'s manually through Coursier so the extraction can be cached and we can avoid the cost of performing it every time at startup

runner/client/src/mill/runner/client/JansiLoader.java
```diff
@@ -0,0 +1,50 @@
...
+
+  private String jansiVersion;
+
+  JansiLoader(String jansiVersion) {
+    this.jansiVersion = jansiVersion;
+  }
+
+  private String jansiLibPathInArchive() {
+    return "org/fusesource/jansi/internal/native/" + OSInfo.getNativeLibFolderPathForCurrentOS()
+        + "/" + System.mapLibraryName("jansi").replace(".dylib", ".jnilib");
```
lihaoyi, line 17: This `String#replace` call looks a bit weird; maybe add a comment explaining why we're doing this?

runner/client/src/mill/runner/client/JansiLoader.java
```diff
@@ -0,0 +1,50 @@
...
+  File tryLoadFast() {
+    File archiveCacheLocation;
+    try {
+      archiveCacheLocation = coursier.paths.CachePath.defaultArchiveCacheDirectory();
+    } catch (IOException ex) {
+      throw new RuntimeException(ex);
+    }
+    File jansiLib = new File(
+        archiveCacheLocation,
+        "https/repo1.maven.org/maven2/org/fusesource/jansi/jansi/" + jansiVersion + "/jansi-"
```
lihaoyi, line 29: These hardcoded references to `repo1.maven.org` are potentially problematic, e.g. some folks work in companies with access to maven central blocked to force their developers to use their internal mirror. Is there some way we can make this obey the `COURSIER_REPOSITORIES` configuration to support such use cases?

lihaoyi: e.g. see https://github.com/com-lihaoyi/mill/issues/3408

alexarchambault: That's a good point… In a first time, maybe we could detect if the default repositories of coursier were overridden or not. If they're overridden, keep doing what we were doing before (calling `tput` every n-th mlliseconds). If they're not, then proceed to do what the PR currently does. Maybe we could also check if `~/.ivy2/local` contains the version of jansi that we use, and do as if repositories were overridden. That way, we could merge now and benefit from this PR in many cases already.

In a second time, in a subsequent PR, I could manage to pull some more coursier logic in Java classes (in the coursier codebase), so that we could stay in Java land here, even when the coursier repositories are overridden.

lihaoyi: Could we bundle the relevant binaries with Mill and unpack them from the Mill executable by reading them from the classpath? That would save us the whole trip-to-maven-central thing and make it nicely self-contained

lihaoyi: Writing it out to `~/.cache/mill` sounds reasonable

integration/migrating/init/src/MillInitMavenTests.scala
alexarchambault: Just added a comment

runner/client/src/mill/runner/client/MillProcessLauncher.java
alexarchambault: Should be fixed

runner/src/mill/runner/MillMain.scala
alexarchambault: Just added a comment. That was necessary to make `OutputDirectoryTests` pass. The library that coursier uses to extract archives logs things with slf4j. So when unpacking the jansi JAR, some logs where sent. Logback is configured in Mill to handle those. But its configuration file hardcoded `out/` as out directory, so that a file with logs was written there no matter what, and the tests checking that out directory customization works broke.

If we stop using coursier to unpack the jansi JAR, that wouldn't be required for this PR. But it's still nice to fix that (we could add a new non-regression test for it then).

runner/client/src/mill/runner/client/JansiLoader.java
alexarchambault: Just did that

alexarchambault:
I'm still investigating, but not sure why, it's much simpler now, and `./mill version` takes around ~200ms, versus ~150ms without those changes

lefou:
Does this have any compatibility restrictions? E.g. does it work on Java 23?

alexarchambault:
> Does this have any compatibility restrictions? E.g. does it work on Java 23?

It does work, although we get a warning at runtime:
```text
$ scala-cli repl --amm --scala 2.13.16 --jvm 24 --dep io.github.alexarchambault.native-terminal:native-terminal:0.0.8
Loading...
Welcome to the Ammonite Repl 3.0.2 (Scala 2.13.16 Java 24)
@ import io.github.alexarchambault.nativeterm.NativeTerminal
import io.github.alexarchambault.nativeterm.NativeTerminal

@ val size = NativeTerminal.getSize()
WARNING: A restricted method in java.lang.System has been called
WARNING: java.lang.System::load has been called by org.fusesource.jansi.internal.JansiLoader in an unnamed module (file:~/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/fusesource/jansi/jansi/2.4.1/jansi-2.4.1.jar)
WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled

size: io.github.alexarchambault.nativeterm.TerminalSize = TerminalSize{width=80, height=25}
```

I've been wanting to [make the native-terminal JAR a multi-JVM JAR](https://github.com/alexarchambault/native-terminal/issues/123), and use project panama stuff in Java >= 22 or 23. It's not done yet, but that should fix the runtime warning.

lihaoyi:
The warning is definitely a bit annoying, I think we should try and find a solution before merging this. 

Is it feasible to vendor and patch Jansi's Java code to allow us to control over the native library loading without using reflection? We could send a patch upstream and see if they would be OK with adding a `JansiLoader.setLoaded(boolean)` method, and in the mean time use our own patched version that includes it

alexarchambault:
@lihaoyi Reflection isn't the issue here. We get a warning because of the `System.load` call. In the PR here, we call it ourselves (so the warning should point at our code, rather than at jansi, in my comment above).

alexarchambault:
That being said, it'd still be nice not to have to rely on reflection. I agree we should try to make jansi add a helper for that.

Addressing https://github.com/alexarchambault/native-terminal/issues/123 should fix the warning here.

lihaoyi:
@alexarchambault got it. I assume we can't just add `--enable-native-access=ALL-UNNAMED` blindly in the bootstrap script since it would fail on older version of Java? In that case using a multi-release jar seems like the only way then

lihaoyi:
If it's just about native loading, how come Jansi doesn't seem to cause this warning when used directly? IIRC we've used JLine/Jansi in the past to load the terminal settings (before switching to `tput`) and I don't recall this warning ever being shown. Or does even using JLine/Jansi directly end up displaying this error if run on a sufficiently new version of Java?

alexarchambault:
> If it's just about native loading, how come Jansi doesn't seem to cause this warning when used directly? IIRC we've used JLine/Jansi in the past to load the terminal settings (before switching to `tput`) and I don't recall this warning ever being shown. Or does even using JLine/Jansi directly end up displaying this warning if run on a sufficiently new version of Java?

The latter, it seems we only get the warning for JDK >= 22

lihaoyi:
Yeah it seems like older versions of the JVM fail when run with that flag

```
/Users/lihaoyi/Library/Caches/Coursier/arc/https/github.com/adoptium/temurin11-binaries/releases/download/jdk-11.0.25%252B9/OpenJDK11U-jdk_aarch64_mac_hotspot_11.0.25_9.tar.gz/jdk-11.0.25+9/Contents/Home/bin/java -cp /Users/lihaoyi/Github/mill/example/depth/javahome/1-custom-jvms/out/foo/compile.dest/classes --enable-native-access=ALL-UNNAMED foo.Foo
```

lihaoyi:
Although from Java 17 it seems to accept the flag without error

```
/Users/lihaoyi/Library/Caches/Coursier/arc/https/github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.13%252B11/OpenJDK17U-jdk_aarch64_mac_hotspot_17.0.13_11.tar.gz/jdk-17.0.13+11/Contents/Home/bin/java -cp /Users/lihaoyi/Github/mill/example/depth/javahome/1-custom-jvms/out/foo/compile.dest/classes --enable-native-access=ALL-UNNAMED foo.Foo
Foo running on Java 17.0.13
```

We need to make the decision to pass the flag or not super early, since it needs to be passed when launching the client JVM.

@alexarchambault does the warning print when the Java client is compiled into a Graal native image? If we make Graal the default in 0.13.0, maybe it's ok to have this error message in the less common JVM-launcher case. Alternatively, if we bump the required version to Java 17, that would allow us to hardcode the flag in the launcher

alexarchambault:
I'm not seeing the warning when using the native client, so both would work

lihaoyi:
https://github.com/com-lihaoyi/mill/pull/4858 makes the native client the default going forward, so I think as long as the native client doesn't print the warning we're good to go. The JVM client should only be used on windows-arm, which would be a minority of users.

When you're done with this PR please update the PR description to summarize all the changes, discussion, and decisions that have taken place as part of the review. Turns out the reasoning around this surprisingly non-trivial and we should capture all that as part of the commit

PR #4801, opened 2025-03-28
Build cleanup: move shared traits/classes into proper Scala files

lefou:
- Moved ExampleParser in dedicated source dir
- Replace `resolveDeps` with defaultResolver usage
- Cleanup mill build: move shared traits/classes into proper Scala files


lefou:
@lihaoyi What do you think of moving shared module traits under `mill-build/src/`?

lihaoyi:
@lefou i dont have a strong opinion, fine with whatever you decide to do here

Issue #4772, opened 2025-03-25
`KotlinModule` can't detect a main class


lefou:
Noticed here: https://github.com/com-lihaoyi/mill/pull/4771

https://github.com/com-lihaoyi/mill/blob/b93f8a5e728e9a60f38e56615a5b87f3b6db26b2/contrib/buildinfo/test/src/mill/contrib/buildinfo/BuildInfoTests.scala#L113-L114

PR #4803, opened 2025-03-28
Backport "Fix main class discovery for `KotlinModule`"

lihaoyi:
Backports https://github.com/com-lihaoyi/mill/pull/4797

Fixes https://github.com/com-lihaoyi/mill/issues/4772

Previously it inherited the Zinc-based main class discovery via JavaModule, which overrode the generic classpath-scanning approach from RunModule. This PR adjusts the overrides such the KotlinModule uses the disk-based discovery API.

Removed all the now-unnecessary def mainClass annotations in the kotlinlib example tests, except for the one in example/kotlinlib/module/1-common-config that is useful for educational purposes.

### DIFF
--- a/example/kotlinlib/basic/1-simple/build.mill
+++ b/example/kotlinlib/basic/1-simple/build.mill
@@ -6,8 +6,6 @@ import mill._, kotlinlib._
 object foo extends KotlinModule {
   def kotlinVersion = "1.9.24"
 
-  def mainClass = Some("foo.FooKt")
-
   def ivyDeps = Agg(
     ivy"com.github.ajalt.clikt:clikt-jvm:4.4.0",
     ivy"org.jetbrains.kotlinx:kotlinx-html-jvm:0.11.0"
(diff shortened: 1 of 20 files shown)


lefou (commented):
Do we really need to backport the `allLocalMainClasses0` task?

lihaoyi:
We don't need to, the alternative would be to copy-paste the `allLocalMainClasses` implementation from RunModule into KotlinModule, which seemed like a messier way of doing things than extracting the `allLocalMainClasses0` helper

lefou:
We could, as you already suggested, just always use the classpath-based main class finder. I think it doesn't add much to the runtime.

lihaoyi:
Yeah I think that's probably the next step. But this is the just backport of the previous PR, so didn't want to get too creative here

lefou:
I wanted to avoid the introduction of deprecated API. But if we do the next step before the release, it's fine too.

Issue #4119, opened 2024-12-12
Allow invocation of `foo.super` tasks from the command line (500USD Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 500USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

See https://github.com/orgs/com-lihaoyi/discussions/6 for other bounties and the terms and conditions that bounties operate under 

---

We already give them unique selectors and paths on disk, should be possible to make them callable as well. This could be useful for debugging

We should be able to say `./mill foo.bar.super` if there is only one overridden task, or if there are multiple overridden tasks then `./mill foo.bar.super.qux.Baz`  where `qux.Baz` is the super-suffix computed in `mill.define.internal.OverrideMapping.assignOverridenTaskSegments` (basically the partially-qualified class name long enough to uniquely define the defining class of the task)

jseteny:
@lihaoyi I start to work on this

jseteny:
@lihaoyi I did not really progress with this and stopped to work on it.

rekudyu:
hey @lihaoyi I would like to work on this.

lihaoyi:
closed by https://github.com/com-lihaoyi/mill/pull/6482

PR #4807, opened 2025-03-28
Enhance supertask resolution and file path handling

zelosleone:
So far from my testings, I saw we only supported the `/super` path to call parent tasks, which worked in CLI but wasn’t ideal for Scala code from my testings (tests in my PR prove this)

We can now use `.super` in Scala code to call the parent implementation of a task which is —much cleaner and more intuitive

### 1. **Path-based (`/super`) (This will still stay the same for CLI support.)**
- **Use:** CLI  
- **Example:**  
  ```bash
  ./mill foo.compile/super (as an example)

2. Method-style (.super)

```
override def compile = T {
  super.compile() // Calls parent
}
```

 What Happens Now:
.super and /super resolve to the same task internally still.

```
`trait BaseModule extends mill.Module {
  def compile = T {
    println("Base compile")
    Seq()
  }
}

object CustomModule extends BaseModule {
  override def compile = T {
    super.compile()
    println("Extra steps")
    Seq()
  }
}
```
`
What I wanted to implement with these changes are:

1. Modify the task resolution system to recognize both syntaxes. 
2. When a task name segment contains .super (like in compile.super), we strip this portion to find the base task definition but preserve the full segment in the task path for parent-overridden trait access.

Both manual testing and unit tests ensure both syntaxes work the same and resolve correctly to parent tasks.

Fixes https://github.com/com-lihaoyi/mill/issues/4119

It also correctly returns the empty string as the overridden version from the trait 

`./mill-assembly.jar show contrib.testng.artifactSuffix.super`


core/define/src/mill/define/ExecutionPaths.scala
```diff
@@ -13,7 +13,23 @@ object ExecutionPaths {
       outPath: os.Path,
       segments: Segments
   ): ExecutionPaths = {
-    val segmentStrings = segments.parts
+    val adjustedSegments =
```
lihaoyi, line 16: Can you describe in english what you are trying to do here?

core/resolve/src/mill/resolve/ResolveCore.scala
```diff
@@ -141,6 +141,75 @@ private object ResolveCore {
         }
 
         (head, current) match {
+          case (Segment.Label(".super"), nt: Resolved.NamedTask) =>
```
lihaoyi, line 144: When is this branch taken and why is it a copy-paste of the branch below?

core/resolve/src/mill/resolve/Resolve.scala
```diff
@@ -126,19 +126,34 @@ private[mill] object Resolve {
       p: Module,
       cache: ResolveCore.Cache
   ): Result[NamedTask[?]] = {
+    // Get the last segment value (task name)
+    val taskName = r.segments.last.value
+    val baseTaskName = if (taskName.contains(".super")) {
```
lihaoyi, line 131: When does this conditional return true?

core/resolve/src/mill/resolve/ResolveCore.scala
```diff
@@ -141,6 +141,75 @@ private object ResolveCore {
...
+            }
+
+            val newLastSegment = lastSegment match {
+              case Segment.Label(name) => Segment.Label(s"$name.super$superSuffix")
+              case other => other
+            }
+            val newSegments = Segments(nt.segments.value.dropRight(1) :+ newLastSegment)
+            Success(Seq(Resolved.NamedTask(newSegments)))
+
+          case (Segment.Label(taskName), m: Resolved.Module) if taskName.endsWith(".super") =>
```
lihaoyi, line 174: When does this branch happen?

core/resolve/test/src/mill/resolve/ResolveTests.scala
```diff
@@ -250,5 +273,117 @@ object ResolveTests extends TestSuite {
...
+        test("superTaskResolution") - check(
+          "outputFile.super",
+          Result.Success(Set(_.outputFile)),
+          Set("outputFile.super")
+        )
+
+        // This test verifies that the fix for the NoSuchFileException issue works correctly.
+        // The issue was that when a task with a .super suffix was resolved, the system was
+        // trying to access a file with the .super suffix included in the path, which didn't exist.
+        // The fix ensures that tasks with a .super suffix use the same file paths as their base tasks.
```
lihaoyi, line 368: This seems incorrect. Tasks with a `.super` suffix have different file paths as their overriding tasks

core/resolve/src/mill/resolve/ResolveCore.scala
zelosleone: Branch below? I simply worked on my deleted branch by copying-pasting my own deleted branch

core/resolve/src/mill/resolve/ResolveCore.scala
zelosleone: It happens when you are trying to access a parent class's implementation of a task from a child class

core/resolve/src/mill/resolve/Resolve.scala
zelosleone: it returns true when resolving a call to a parent task implementation using `super.taskName()` in an overridden task

core/define/src/mill/define/ExecutionPaths.scala
zelosleone: that code modifies "super" segments to support parent task access, without it it only supported /super, now it also properly supports parent task with .super path

lihaoyi: I still don't understand

lihaoyi:
I generally do not understand any of your responses. Please provide proper explanations of what you are tryingn to do, as without understanding we cannot review your code regardless of how correct it is

lihaoyi:
You should be able to implement this without needing to perform string manipulation on the `Segments` data structure. We should parse the input query directly into the necessary data structure and operate on that data structure directly, identifying the `"super"` segment. The current approach is hacky and cannot be merged

zelosleone:
> You should be able to implement this without needing to perform string manipulation on the `Segments` data structure. We should parse the input query directly into the necessary data structure and operate on that data structure directly, identifying the `"super"` segment. The current approach is hacky and cannot be merged

Alright, let me check if i can do it in the way you want, I will close it myself if i cannot.

zelosleone:
@lihaoyi Ready for another review, I believe i solved this by usign segments directly instead of suffix manipulation.

core/resolve/src/mill/resolve/ResolveCore.scala
```diff
@@ -141,6 +141,49 @@ private object ResolveCore {
...
+              Some(tail.head.asInstanceOf[Segment.Label].value)
+            } else {
+              None
+            }
+
+            val newSegments = Segments.superTask(nt.segments, superTarget)
+            Success(Seq(Resolved.NamedTask(newSegments)))
+
+          case (Segment.Label(taskName), m: Resolved.Module) if taskName.endsWith(".super") =>
+            val baseTaskName = taskName.stripSuffix(".super")
```
lihaoyi, line 155: I'm still seeing string manipulation here

zelosleone: Yeah sorry, removed that part as well. It was not doing anything anyways after i switched to segments.

lihaoyi:
@zelosleone please review your own code first before I take a look

lihaoyi:
I'm going to declare failure on this PR; we've been going over it for weeks and the quality of code and comments is still nowhere near what is required. Thanks @zelosleone for your efforts, perhaps others in future may have better luck on that ticket

PR #4810, opened 2025-03-28
Standardize on filesystem-based main class discovery

lihaoyi:
`DiscoverMainClassesMain` was previous missing top-level classes, which this PR fixes by replacing its unnecessarily complicated recursion code


lihaoyi:
@lefou the filesystem-based main class discovery logic fails in some cases, seems like it isn't quite ready to totally replace the zinc-based approach in all our test cases, so swapping over isn't trivial at the moment. Going to close this PR for now but feel free to look into it if you're interested

PR #4811, opened 2025-03-29
Fix uTest test task creation for multi-test-class test modules

lihaoyi:
Seems to be caused by the `def getTestTasks` changes in https://github.com/com-lihaoyi/mill/pull/4737. The problem appears to be in uTest, which is strict about ensuring the `TaskDef`s passed to `.tasks` must cover all tasks selected by the `args` selector, which fails when we make multiple separate calls to `.tasks` each with the tasks related to an individual test class:

https://github.com/com-lihaoyi/utest/blob/76e7fe5a72464ab44df3f829bd2b3e7699d522bf/utest/src/utest/runner/BaseRunner.scala#L67-L75

Arguably this is an issue in uTest for being too strict, but for now a workaround is to do all the work in a single call to `.tasks` with all test classes involved, and break it up into the per-test-class sublists after the fact

Adjusted `example.scalalib.testing[1-test-suite]` to cover this case by having two test classes in the `foo` test suite


lihaoyi:
This test failure passes on my laptop, will merge first and if it breaks in main we can disable the test

PR #4819, opened 2025-03-30
Don't use Zinc analysis file to find main classes

lefou:


scalalib/worker/src/mill/scalalib/worker/DiscoverMainClassesMain.scala
```diff
@@ -67,8 +67,11 @@ import scala.util.Using
         new ClassPathFactory(settings, registry).classesInExpandedPath(cp)
       )
 
+      // always include the Java default package
+      val defaultPackage = ""
+
       val mainClasses = for {
-        foundPackage <- recursive("", (p: String) => path.packages(p).map(_.name))
+        foundPackage <- defaultPackage +: recursive("", (p: String) => path.packages(p).map(_.name))
```
lihaoyi, line 74: This whole `def recursive` thing that was copy-pasted from elsewhere can probably be replaced by the following snippet from the other PR

```scala
      def recursivePackages(p: String) = Seq(p) ++ path.packages(p).map(_.name)

      val mainClasses = for {
        foundPackage <- recursivePackages("")
```

lefou: Got it. I just wanted to see all test green before I change that "working" part of the code.

lefou:
* Superseded by https://github.com/com-lihaoyi/mill/pull/4822

PR #4822, opened 2025-03-31
Find main classes with ClassGraph

lefou:
Since the scalap approach doesn't work for jlink/jpackage modules and scalap seems discontinued in Scala 3 anyways,
this is an attempt to use ClassGraph (https://github.com/classgraph/classgraph) to find main classes.

Added tests for allLocalMainClasses (in the default package).

Removed all alternative implementations to find main classes.


lihaoyi (approved):
Looks fine to me, if the tests pass on the various languages and JVM versions that gives pretty good confidence

scalalib/src/mill/scalalib/RunModule.scala
```diff
@@ -46,27 +48,37 @@ trait RunModule extends WithZincWorker {
...
+    classgraphWorker().discoverMainClasses(localRunClasspath().map(_.path))
+  }
+
+  def classgraphWorkerClasspath: T[Seq[PathRef]] = T {
+    zincWorker().defaultResolver().classpath(Seq(
+      Dep.millProjectModule("mill-scalalib-classgraph-worker")
+    ))
+  }
+
+  def classgraphWorker: Worker[ClassgraphWorker] = Task.Worker {
```
lihaoyi, line 63: Should this go on `ZincWorkerImpl`? It's mostly shared across all `RunModule`s, and doesn't seem to have any local configuration

scalalib/src/mill/scalalib/RunModule.scala
```diff
@@ -46,27 +48,37 @@ trait RunModule extends WithZincWorker {
...
+
+    object Worker extends ClassgraphWorker with AutoCloseable {
+      override def discoverMainClasses(classpath: Seq[Path])(implicit ctx: Ctx): Seq[String] =
+        worker.discoverMainClasses(classpath)
+
+      override def close(): Unit = {
+        cl.close()
+      }
     }
+    Worker
```
lihaoyi, line 81: This code is pretty awkward.

* I'd just make it one big `new ClassgraphWorker extends AutoCloseable{}` and put everything inside.
* The logging can probably be removed; anyone who wants to debug stuff can add it, but it really makes the code much harder to read
* The naming needs to be improved. `klass` vs `cls` is super confusing

lefou: That means I need two "worker" classes. The `AutoClosable` handling obviously needs to happen outside the real `ClassgraphWorker`, but since the `ClassgraphWorker` is instantiated from the managed classloader, it can't (or shouldn't) provide the cleanup logic.

I can cleanup further...

lefou: fixed

lihaoyi (approved):

lefou:
@lihaoyi Do we want this backported to 0.12?

lihaoyi:
@lefou sure let's do that, it's not a breaking change and doesn't rely on any of the other new stuff

PR #4823, opened 2025-03-31
Adjust logback outfile file upon startup

alexarchambault:
If anything in Mill logs with slf4j, logback is in charge of handling those logs. If ever the Mill out directory is customized, we need to adjust the logback log file location too, so that logs aren't written to the default `out/mill.log`.

### DIFF
--- a/runner/client/src/mill/runner/client/MillClientMain.java
+++ b/runner/client/src/mill/runner/client/MillClientMain.java
@@ -17,6 +17,10 @@ import mill.constants.Util;
  */
 public class MillClientMain {
   public static void main(String[] args) throws Exception {
+
+    if (System.getProperty("mill.log.file") == null)
+      System.setProperty("mill.log.file", OutFiles.out + "/mill.log");
+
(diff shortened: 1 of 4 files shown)


lefou:
Could you clarify how this matches or bypasses https://github.com/com-lihaoyi/mill/discussions/4680?

alexarchambault:
@lefou This doesn't do anything for https://github.com/com-lihaoyi/mill/discussions/4680. It's more a follow-up of #3530. Before this PR, if any library used by the build logs with slf4j, the log output goes to `out/mill.log`, without respecting any out directory customization. This PR fixes that, by sending the logs to `mill.log` under the right out directory (taking into account out directory customization).

I stumbled upon that by working on https://github.com/com-lihaoyi/mill/pull/4800, even though it doesn't need this anymore.

alexarchambault:
Still a draft, as I'd like to add a non-regression test for this

lihaoyi:
Closing this for now

PR #4824, opened 2025-03-31
Don't publish jansi in ~/.ivy2/local in integration test

alexarchambault:
This makes a test in `MillInitMavenTests` test local publishing by writing to a throw away directory, rather than the user's `~/.ivy2/local`, where the written files might be a problem.


lefou (commented):
Looks good.

lihaoyi (approved):

Issue #3982, opened 2024-11-18
Make custom JVM versions used during Scala compilation


lihaoyi:
Follow up from https://github.com/com-lihaoyi/mill/pull/3716

Zinc only supports customizing the javaHome for Java compiler invocations, and does not support customizing it for Scala compiler invocations. We likely either need to implement Scala compiler javaHome support upstream in Zinc, or we'll need to do so in Mill. Either case will likely involve spawning and managing long-lived JVM worker processes (one-per-JVM-version) to run ZincWorkerImpl, and RPC to these workers from the main Mill process

lihaoyi:
Taking this off 0.13.0 for now. Would be nice to fix it before then, but can be done in a binary compatible manner if we're careful

lihaoyi:
Unassigning Jakob since he won't have time to get to this

lihaoyi:
Now that https://github.com/com-lihaoyi/mill/pull/5327 has landed, this should be doable purely in the implementation of `JvmWorkerImpl` without changing the public API

arturaz:
I'm taking this on.

Issue #3567, opened 2024-09-17
Break out `javalib/` module from `scalalib/`


lihaoyi:
Right now we kind of fake it with a `javalib` package in the `scalalib` module, but ideally we would have `javalib` be a separate module and have `scalalib` depend on it

lefou:
Since separating `javalib` and `scalaib` it more to please our users but not so much a technical requirement (since we don't really have different dependencies, as these are all loaded dynamically), we could as well rename the current `scalalib` module to `jvmlib` and just shuffle the classes around, so we have proper packages `mill.jvmlib` (e.g. for `RunModule`), `mill.javalib` and `mill.scalalib` packages.

PR #4834, opened 2025-04-01
Rename `ZincWorker` to `JvmWorker`

lihaoyi:
It no longer is exclusive to zinc, and nowadays also handles custom JVM versions for compile/run/test as well. The new name is more generic to reflect that, and probably easier to understand for the vast majority of people who don't know what "Zinc" is

We could consider splitting out a `JvmWorker` from `ZincWorker`, but it is likely the two will remain tightly bound together, e.g. to allow https://github.com/com-lihaoyi/mill/issues/3982


lefou:
This is a good change to rename the `worker` module into `jvm-worker` or `compiler-worker`.

changelog.adoc
lefou, line 1: changelog should stay unchanged

lihaoyi:
Currently it's `scalalib-worker`, I guess renaming that would be part of https://github.com/com-lihaoyi/mill/issues/3567

lefou:
Probably part of #3567, yes. We now also have `scalalib-classgraph-worker` and other "worker" modules, so it might help to disambiguate if we name it `scalalib-compiler-worker` or later `javalib-compiler-worker`.

PR #4842, opened 2025-04-02
Vendor `lefou::mill-vcs-version` as part of `mill-main`

lihaoyi:
This PR vendors a copy of https://github.com/lefou/mill-vcs-version inside the Mill repo. 

`mill-vcs-version` is necessary for re-bootstrapping on top of `main`. `mill-mima` or `mill-scalafix` are linters that can be disabled during the re-bootstrapping process and turned on again later when the third-party plugins are available, but `mill-vcs-version` is needing to compute versions during the bootstrapping/dogfooding process and cannot simply be skipped.

An alternative would be copy-pasting code into `build.mill` as a temporary hack, but I think it's better to just vendor a full copy of the upstream codebase. `mill-vcs-version` is a pretty small and stable codebase, so the overhead of maintaining a vendored copy should be manageable.


lefou (commented):
I'm on board. Looks good to me in general, but a few nits.

mill-vcs-version was published under the Apache License v2 and also got external contributions, although small. The license requires, that we mention it. I think a simple "Originally developed/distributed under the Apache License ..." should be good.

We should remove all `@deprecated` API, since we change the package anyways.

PR #4844, opened 2025-04-02
Replace deprecated API: log.outputStream and log.errorStream

lefou:
Depends on a Mill version containing
* https://github.com/com-lihaoyi/mill/pull/4840


lefou:
Rebased on https://github.com/com-lihaoyi/mill/pull/4845, to be merged after it.

PR #4848, opened 2025-04-02
Rebootstrap on `main` branch targeting `0.13.x`

lihaoyi:


lihaoyi:
I assume it'll take more work to get there, since not all your forwards-compat PRs have landed and there'll probably be more required that we haven't noticed yet. I guess the goal would be to be able to swap back to Mill 0.12.latest and build the Mill repo without changes to the `.mill` files

lefou:
Providing a patchfile do downgrade to Mill 0.12.latest is a good alternative.

lefou:
All renamed/moved changes affecting `def`s that are support to be overridden, can't be fully mapped to some compatibity API. E.g. `def millSourcePath = super...`

PR #4851, opened 2025-04-02
MillBuildRootModule: Don't include generated wrapper files in sources

lefou:
The generated files under the `wrapped` directory wrap and duplicate the original files. Although they're sources to the compiler, they're not sources for the user and/or editor. 

Also, when compiling the wrapped files, we already translate compiler error messages back to the original files, hence the user/editor should not need to deal with these intermediate files. Before this change, the IDE sometimes shows the generated `build.mill` instead of the original `build.mill` when jumping to definitions or using other code navigation options. If the user edited these files, changes were lost later.

This PR removes the wrapped files from `generatedSources`, so they no longer appear in the IDE. Those files are still fed via `allScalaSourceFiles` to the compiler.

Tested manually with Metals 1.5.2+97-5a09352f-SNAPSHOT/vscode  and IntelliJ 2025.1.


lefou:
* https://github.com/com-lihaoyi/mill/issues/5044

lefou:
Going to merge this unless @lihaoyi or @alexarchambault have any objections.

lihaoyi:
got for it

Issue #4536, opened 2025-02-12
Cyclic module reference detected


lefou:
This is from the `main` branch of the Mill repo.

```
> mill __:^IntegrationCrossModule.__.ivyDepsTree
[65/65] ============================== __:^IntegrationCrossModule.__.ivyDepsTree ============================== 7s
Cyclic module reference detected at , it's required to wrap it in ModuleRef.

```

Issue #4514, opened 2025-02-08
`__:Module` should properly select modules


lefou:
* `mill resolve __:Module` should list all modules
* `mill inspect __:Module` should show inspections for modules only

Issue #3618, opened 2024-09-27
Bug in `__:Type` syntax


Shri333:
Resolve is working incorrectly. If I have the following `build.mill`:
```scala
package build
import mill._, javalib._

trait TypeA extends Module {
  def foo = Task { "foo" }
}
trait TypeB extends Module {
  def bar = Task { "bar" }
}
trait TypeC extends Module {
  def baz = Task { "baz" }
}
trait TypeAB extends TypeA with TypeB

object typeA extends TypeA
object typeB extends TypeB
object typeC extends TypeC {
  object typeA extends TypeA
}
object typeAB extends TypeAB
```
Running `mill resolve _:TypeA._` produces the following output:
```
typeA.foo
typeAB.bar
typeAB.foo
```
But, running `mill resolve __:TypeA._` produces the following output:
```
clean
init
inspect
path
plan
resolve
show
showNamed
shutdown
typeA
typeA.foo
typeAB
typeAB.bar
typeAB.foo
typeB
typeC
typeC.typeA.foo
version
visualize
visualizePlan
```

lefou:
That's strange, we have a test for this case. https://github.com/com-lihaoyi/mill/blob/0e50ff7ee3216f8224958e150dbfeac6113a2ce1/main/resolve/test/src/mill/main/ResolveTests.scala#L973-L976

lihaoyi:
I think I've hit similar issues when working on unrelated tasks, but haven't had time to investigate

lefou:
Releated:

* #4514

PR #4861, opened 2025-04-04
Fix wildcard and type selector query bugs

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/4536 and https://github.com/com-lihaoyi/mill/issues/4514 and https://github.com/com-lihaoyi/mill/issues/3618

- `__` was incorrectly adding the current module to `seenModules`, which causes problems when it recurses on the current module. This PR skips updating `seenModules` when recursing on the current module. There are probably other edge cases we haven't noticed in this cycling module graph detection logic, but for now this just fixes the reported issue and the test suite should give confidence we didn't regress anything

- `typePatterns` was incorrectly being resolved only on modules during the `__` wildcard resolution, and allowing tasks to pass through without filtering. This PR moves the filtering to after `resolveDirectChildren`/`resolveTransitiveChildren` has completed, ensuring we filter out unwanted tasks as well.

Added unit tests for these cases

### DIFF
--- a/core/resolve/src/mill/resolve/ResolveCore.scala
+++ b/core/resolve/src/mill/resolve/ResolveCore.scala
@@ -153,7 +153,6 @@ private object ResolveCore {
                     m.cls,
                     None,
                     current.segments,
-                    Nil,
                     seenModules,
                     cache
                   )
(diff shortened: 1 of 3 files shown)


lefou:
Wow, you fixed three issues I reported in one strike!

PR #4864, opened 2025-04-04
[WIP] Mill launcher classpath rework

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/2703. 

`build.mill` code is no more strongly encapsulated from the `mill.runner` code, only sharing `scala-library` and a minimal `mill.runner.api` interface. Overall this gives us and users much more fine-grained control over the compile-time and run-time classpath available to `build.mill` files: 

1. We can now elide Mill modules from being visible to user code (without needing to tediously mark everything as `private`) 

2. Users can evolve or upgrade versions of libraries without being pinned to the version used internally by Mill (modulo normal backwards compatibility concerns)

3. The Mill distribution should become smaller as a result speeding up the initial download, and bootstrapping should be faster since libraries can later be downloaded in parallel by Coursier

## Implementation

This PR implements a stricter separation between the `build.mill` files and the Mill assembly classpaths:

1. `MillBuildRootModule` no longer takes the enclosing Mill assembly jar as it's `unmanagedClasspath`, and instead resolves necessary jars via Coursier

2. The `build.mill` classloader now also contains the `mill.eval.Evaluator` and all third-party dependencies (OS-Lib, uPickle, etc.) rather than inheriting them as part of the Mill assembly

3. `mill.runner` now interacts with the `build.mill` classloader through a very narrow interface defined in `mill.runner.api`, which contains minimal logic and only depends on the Scala standard library

4. First-party mill modules like `mill-scalalib`, `mill-kotlinlib`, `mill-scalajslib`, and other modules used in the `build.mill` files are now resolved from Maven Central

5. The Mill assembly no longer depends on most of those first-party modules, with the exception of `mill-scalalib` and other modules necessary to implement the `MillBuildRootModule.BootstrapModule`

6. Now every `build.mill` file can now have different versions of any third-party library except for `scala-library`, and the versions of libraries used in `build.mill` no longer needs to align with the versions used in `mill-runner`


lihaoyi:
All tests pass except for the BSP and GenIdea subsystems. Those will need an overhaul to break the tight coupling between them and all the meta-build evaluators: 

1. BSP currently directly references and manipulates `Evaluator` and `Module` and `Task`s from every meta-build's classloader at the same time, forcing them to be on the same classpath with the `mill.runner` and with each other

2.  We would need to do a similar refactor to bring most of the logic into each meta-build's classloader, exposing only a thin interface for `MillBuildServer` to interact with each meta-build and aggregate the results to respond to the BSP client

Will be a pretty big refactor, but seems doable

PR #4867, opened 2025-04-04
Generate scripts from templates

lefou:


lefou:
Fix https://github.com/com-lihaoyi/mill/issues/4327

PR #4871, opened 2025-04-05
Remove TestModule.compile

lefou:
We couldn't removed in 0.12 due to bin-compat. Later we overlooked it.

Turns out, it is still used in the `JUnit5` case. I tried to come up with a alternative implementation, without inspecting whether there is a better way.


lefou:
I suspect, the failing test isn't something new, but seldomly seen due to selective testing. Since we want the change in this PR nevertheless, let's deal with the test issue afterwards.

PR #4876, opened 2025-04-06
More cleanup for Mill BSP integration

lihaoyi:


lihaoyi:
@lefou seems the arrow test you hit in your earlier PR was a real breakage, seeing it in other PRs now

lefou:
Hm. I think I reproduced it locally with the `main` branch before merging. I can look into it.

PR #4877, opened 2025-04-06
[WIP] avoid usage of `Task.Anon` in Mill BSP server

lihaoyi:
This is a requirement to split up the BSP server classpath from the evaluator classpath, since they may each have their own `Task` classes and transitive dependencies. It's generally straightforward to move the logic into `BspModule`

### DIFF
--- a/scalalib/src/mill/scalalib/RunModule.scala
+++ b/scalalib/src/mill/scalalib/RunModule.scala
@@ -209,6 +209,22 @@ trait RunModule extends WithJvmWorker {
    */
   def launcher: T[PathRef] = Task { launcher0() }
 
+  def bspJvmRunTestEnvironment = {
+    val moduleSpecificTask = this match {
+      case m: (TestModule & JavaModule) => m.getTestEnvironmentVars()
+      case _ => allLocalMainClasses
+    }
+    Task.Anon {
+      (
+        runClasspath(),
+        forkArgs(),
+        forkWorkingDir(),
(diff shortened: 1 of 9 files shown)


scalalib/src/mill/scalalib/RunModule.scala
```diff
@@ -209,6 +209,22 @@ trait RunModule extends WithJvmWorker {
...
+    Task.Anon {
+      (
+        runClasspath(),
+        forkArgs(),
+        forkWorkingDir(),
+        forkEnv(),
+        mainClass(),
+        moduleSpecificTask()
+      )
+    }
```
lefou, line 226: I feel uncomfortable with this return type (and possible in other locations). We refactored it multiple times already. Instead, we should use a class which can be more easily changed/extended. It would better support our binary compatibility story.

In general, we should annotate the return types to avoid accidental changed due to inference.

lihaoyi: Yeah this code isn't great, but it will take some time and a lot of refactoring to get `MillBspServer` and friends into a decent state. Definitely will take more than one PR.

lefou: Ideally, we can have exactly one bspXxx-method for each bsp request type. That way, we are on the safe side in terms of what we need to return. In case later BSP version introduce changes, we can start adding versions to the tasks.

PR #4881, opened 2025-04-07
Make BSP/install only create the .bsp/mill-bsp.json file

alexarchambault:
This makes `./mill mill.bsp.BSP/install` only create a `.bsp/mill-bsp.json` file, and not write the BSP worker class path to a `.bsp/mill-….resources` file any more, which shouldn't be needed after #4948.

This also makes `./mill mill.bsp.BSP/install` work even when the build doesn't compile, so that IDEs can load Mill projects in that case too.


lihaoyi:
@alexarchambault for these improvements maybe you can start off targeting 0.12.x? I expected the BSP setup for 0.13.x will look very different once I'm done with the current refactorings, so we should hold off on feature work on 0.13.x until the dust has settled since the approach we take may change substantially

alexarchambault:
Another change is https://github.com/alexarchambault/mill/commit/9dc3c8cf07dc2b028d55e1bbd71006dd75da1749, that I already rebased locally on the latest `main`. It adds the ability to broadcast Mill build changes to BSP clients, and adds some new BSP integration tests, which could help refactor BSP support.

alexarchambault:
The changes in the PR here are in line with https://github.com/com-lihaoyi/mill/pull/4879, it seems. They allow to fetch the BSP worker class path with coursier right when we need it, rather than upfront before users even attempt to start the BSP server.

lefou:
So, if you only load the dependencies when the IDE is about to start the server, how do you ensure to respond to the BSP client in a timely manner? 

In our first version of the BSP server, we had a similiar approach (but via Mill task dependencies, not direct coursier) we ran into timeouts. IIUC, we can only speak the BSP protocol after we assembled to full classpath, so our BSP server is going dark from the clients POV.

lihaoyi:
We need to be careful to not break existing IDE support that may involve running `mill.bsp.BSP/install`. Perhaps we can detect that query string and special-case it as a no-op that doesn't run `MillBuilBootstrap` or run any tasks? That way existing IDEs that run it can continue to work, but since it's a no-op it won't crash on a broken build.mill, and eventually when IDEs remove the call to `mill.bsp.BSP/install` things can continue to work seamlessly

alexarchambault:
> @alexarchambault what's the status on this PR? Do you think we can merge it?

Checking what the CI says, but it could be merged, yes. I have [more BSP changes to come](https://github.com/alexarchambault/mill/tree/bsp-stuff), I should open a PR for those tomorrow or the next day.

alexarchambault:
@lihaoyi Feel free to undraft and merge if [the CI](https://github.com/alexarchambault/mill/actions/runs/14480356058/job/40616237067) is green-enough for you.

alexarchambault:
> We need to be careful to not break existing IDE support that may involve running `mill.bsp.BSP/install`. Perhaps we can detect that query string and special-case it as a no-op that doesn't run `MillBuilBootstrap` or run any tasks? That way existing IDEs that run it can continue to work, but since it's a no-op it won't crash on a broken build.mill, and eventually when IDEs remove the call to `mill.bsp.BSP/install` things can continue to work seamlessly

This PR doesn't do this currently. I could add it either tomorrow or the next day. Feel free to add it on your own in the mean time if you want.

runner/src/mill/runner/MillMain.scala
```diff
@@ -326,16 +326,15 @@ object MillMain {
...
-            val logDir = wsRoot / OutFiles.out / "mill-bsp"
-            os.makeDir.all(logDir)
-            worker.startBspServer(
-              define.WorkspaceRoot.workspaceRoot,
-              splitStreams,
-              logDir,
-              true
-            )
-          }
+          val worker = mill.bsp.BspClasspathWorker(wsRoot, logger = ???)
```
alexarchambault, line 329: Actually, this shouldn't be merged for now because of this (added while rebasing)

alexarchambault: Should be fixed

alexarchambault:
> So, if you only load the dependencies when the IDE is about to start the server, how do you ensure to respond to the BSP client in a timely manner?
> 
> In our first version of the BSP server, we had a similiar approach (but via Mill task dependencies, not direct coursier) we ran into timeouts. IIUC, we can only speak the BSP protocol after we assembled the full classpath, so our BSP server is going dark from the clients POV.

That might be a problem, yes. This runs fast enough on my machine on an empty cache (< 10 s), but that could take longer on a slower connection. Maybe the recent class path isolation work could allow to embed BSP directly in Mill, so that we don't need to handle it via a worker, and nothing new needs to be downloaded?

lihaoyi:
Sounds good, I'll let you do it then

lihaoyi:
Yes embedding the BSP code seems reasonable if the download time is a worry. Most people will be using and IDE and will need to download it anyway, so we don't save anything by making it download lazily on-demand, and now with classpath isolation it wouldnt affect user code

lefou:
> Yes embedding the BSP code seems reasonable if the download time is a worry. Most people will be using and IDE and will need to download it anyway, so we don't save anything by making it download lazily on-demand, and now with classpath isolation it wouldnt affect user code

As long as we make sure, we don't mix the BSP classpath with anything a use might want to load into their own builds (without isolation classloaders) this should work. 

I think we should keep it separated though, but trigger the download while building the project. This should be a small slowdown only on the first build, but it ensures the jars are in place.

runner/src/mill/runner/MillMain.scala
```diff
@@ -160,6 +162,25 @@ object MillMain {
...
+                  config.leftoverArgs.value.headOption.contains("mill.bsp.BSP/install")
+              ) {
+                val expectedSelector = Seq((
+                  Some(Segments.labels("mill", "bsp", "BSP")),
+                  Some(Segments.labels("install"))
+                ))
+
+                for {
+                  parsed <- ParseArgs(config.leftoverArgs.value, SelectMode.Separated).toOption
+                  case Seq((`expectedSelector`, args)) <- Option(parsed)
```
lihaoyi, line 177: This nested code to try and detect the `install` command looks very convoluted. Would just checking `config.leftoverArgs.value.headOption == Some("mill.bsp.BSP/install")` be enough?

alexarchambault: I did that to parse any `--jobs` argument passed to it (whose value ends up being passed to the `./mill --bsp` command). I don't rely on it, but I don't know for other people out there.

lihaoyi: Seems like IntelliJ doesn't pass the flag resulting in the default of `--jobs 1` https://github.com/search?q=repo%3AJetBrains%2Fintellij-scala+%22mill.bsp.BSP%22&type=code, while Metals passes `--jobs 0` https://github.com/search?q=repo%3Ascalameta%2Fmetals%20%22mill.bsp.BSP%22&type=code

I think for now let's just ignore the `--jobs` config here to simplify our code, and either hardcode 1 or hardcode 0 (probably doesn't really matter).

bsp/src/mill/bsp/BSP.scala
```diff
@@ -85,4 +73,18 @@ object BSP extends ExternalModule with CoursierModule {
     )
   }
 
+  private[mill] case class InstallArgs(jobs: Int = 1)
```
lefou, line 76: I'd expect that you use this class in the `install` command?

runner/src/mill/runner/MillMain.scala
lefou: What is the situation in which this is needed at all?

lefou: I guess this is the special handling needed to make `install` work even when the build does not work. If we keep it here, we should add some comment sufficiently explaining the situation.

I don't like this, tbh. If we're going to intercept the `mill.bsp.BSP/install` and just fake it, we might as well remove the original `Task.Command`. Ideally, we can use the same code in both situations, to avoid some config args to be accepted on one situations but not in the other. Of course, I'd like to keep the proper `--jobs` parsing, since this is currently the only way for a user to explicitly configure how the BSP server should run.

lihaoyi: The goal is in general moving the BSP logic out of `ExternalModule`s into the `mill-runner`, so really we should remove the `extends ExternalModule` from bsp/src/mill/bsp/BSP.scala and just make it normal helper methods

alexarchambault: Just removed the current `mill.bsp.BSP/install` command, and simplified the handling of `--jobs` in the manual handling of `mill.bsp.BSP/install` in Mill main

bsp/src/mill/bsp/BSP.scala
alexarchambault: Just removed the original `BSP/install` command altogether (and this class too actually)

lefou:
I think we should provide a shorter way to install the BSP connection file, e.g. `--bsp-install`, in addition to the compatibility `mill.bsp.BSP/install`. We might want to revive the `mill.bsp.BSP` module at a later point. Not for installation, but maybe for ad-hoc querying/debugging of the BSP server. E.g. just issuing a BSP request on the CLI instead of the need to work over HTTP.

lihaoyi:
@alexarchambault looks like some of the tests may need to be updated I think

runner/src/mill/runner/MillMain.scala
```diff
@@ -491,4 +517,9 @@ object MillMain {
     for (k <- systemPropertiesToUnset) System.clearProperty(k)
     for ((k, v) <- desiredProps) System.setProperty(k, v)
   }
+
+  private val bspInstallAliases = Seq(
+    "mill.bsp.BSP/install",
+    "mill.bsp/install"
```
lefou, line 523: Under the assumption, that we also add a dedicated cli option, I would only keep the first alias. The second one was added after any stable release.

lefou (commented):
We currently emulate a command task to stay compatible to previous Mill versions. But in fact we implement a cli option, so we should also make it one. I suggest `--bsp-install`.

runner/src/mill/runner/MillMain.scala
alexarchambault: Should be fixed

runner/src/mill/runner/MillMain.scala
```diff
@@ -163,6 +164,33 @@ object MillMain {
...
+                  case other =>
+                    streams.err.println(
+                      s"Warning: ignoring leftover arguments passed to ${config.leftoverArgs.value.head}"
+                    )
+                    BSP.defaultJobCount
+                }
+              }
+
+              viaEmulatedExternalCommand.orElse {
+                Option.when(config.bspInstall.value)(BSP.defaultJobCount)
```
lefou, line 191: I think we should use the value of `config.threadCount` (after parsing) when defined, and oly use `BSP.defaultJobCount` as a fallback.

lefou (approved):

PR #4884, opened 2025-04-07
Add TestModule#testLogLevel, make netty tests less verbose

alexarchambault:
This adds a new `testLogLevel` task on `TestModule`, that allows users to silence test debug output. It's useful to make the netty tests less verbose, which is what the second commit does.

My goal here is to make the netty tests, that fail from time to time, stop unnecessarily flooding the CI output, like [they do sometimes](https://github.com/alexarchambault/mill/actions/runs/14306280494/job/40092395363).

### DIFF
--- a/example/thirdparty/netty/build.mill
+++ b/example/thirdparty/netty/build.mill
@@ -59,6 +59,8 @@ trait NettyBaseTestSuiteModule extends NettyBaseModule with TestModule.Junit5 {
     "-enableassertions"
   )
 
+  def testLogLevel = mill.api.TestReporter.LogLevel.Info
+
   def compile = Task {
     // Hack to satisfy fragile tests that look for /test-classes/ in the file paths
     val sup = super.compile()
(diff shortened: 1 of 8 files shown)


alexarchambault:
I was hesitating adding a new task on `TestModule` for that, but it seems to be useful for the netty test, so I'm thinking it might be useful in other contexts too

alexarchambault:
Should I merge this one too? The extra task on `TestModule` might require careful review.

lihaoyi:
I think it looks reasonable, not sure if @lefou has any opinions

example/thirdparty/netty/build.mill
```diff
@@ -636,16 +638,18 @@ object `transport-udt` extends NettyModule {
 
 > ./mill -j5 __.compile
 
-> ./mill 'codec-{dns,haproxy,http,http2,memcache,mqtt,redis,smtp,socks,stomp,xml}.test'
+> ./mill 'codec-stomp.test.testForked'
```
lefou, line 641: I think the idea was that all tests must pass. This is now no longer checked.

lefou (approved):
Looks ok. Don't know if we want to reduce the test coverage though.

example/thirdparty/netty/build.mill
alexarchambault: It is still checked, right after this line. The diff doesn't make it obvious, but the test coverage shouldn't be lower.

lefou: Thanks for pointing this out.

PR #4897, opened 2025-04-09
Implement prepareOffline for ClassgraphWorkerModule

unlsycn:
#4822 introduced `ClassgraphWorkerModule` and the corresponding package `classgraph-worker`, but `prepareOffline` does not install it, causing offline build to fail. This PR fixes that problem.

### DIFF
--- a/scalalib/src/mill/scalalib/JavaModule.scala
+++ b/scalalib/src/mill/scalalib/JavaModule.scala
@@ -1245,6 +1245,7 @@ trait JavaModule
     Task.Command {
       super.prepareOffline(all)()
       resolvedIvyDeps()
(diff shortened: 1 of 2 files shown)


scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -1246,7 +1246,7 @@ trait JavaModule
       super.prepareOffline(all)()
       resolvedIvyDeps()
       jvmWorker().prepareOffline(all)()
-      resolvedRunIvyDeps()
```
lefou, line 1249: Why did you remove `resolveRunIvyDeps`?

lefou (approved):
Looks good to me. Thank you!

PR #4902, opened 2025-04-10
Don't attempt to get the color of the empty string

lefou:
Fix https://github.com/com-lihaoyi/mill/issues/4903

Follow up to https://github.com/com-lihaoyi/mill/pull/4769

### DIFF
--- a/core/internal/src/mill/internal/LinePrefixOutputStream.scala
+++ b/core/internal/src/mill/internal/LinePrefixOutputStream.scala
@@ -42,7 +42,10 @@ private[mill] class LinePrefixOutputStream(
       val bufferString = buffer.toString
       if (bufferString.length > 0) {
         val s = fansi.Str.apply(bufferString, errorMode = fansi.ErrorMode.Sanitize)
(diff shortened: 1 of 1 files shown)


core/internal/src/mill/internal/LinePrefixOutputStream.scala
```diff
@@ -42,7 +42,10 @@ private[mill] class LinePrefixOutputStream(
       val bufferString = buffer.toString
       if (bufferString.length > 0) {
         val s = fansi.Str.apply(bufferString, errorMode = fansi.ErrorMode.Sanitize)
-        endOfLastLineColor = s.getColor(math.max(0, s.length - 1))
+        if (s.length > 0) {
+          // only change last value if we have some proper ansi string
+          endOfLastLineColor = s.getColor(s.length - 1)
```
lefou, line 47: If the string a length of zero, there is also no color information as index `0`. 

Now, we only attempt to get the color of the last string position, if there is a character at all. Otherwise we don't update the color info.

Issue #4885, opened 2025-04-07
Limit `build.mill` classpath to minimal set of Mill modules


lihaoyi:
Follow up of https://github.com/com-lihaoyi/mill/pull/4879

Now that `mill-runner` and `build.mill` have separate classpaths, we should trim down the `build.mill` classpath to the minimum set necessary. That probably means including:

- `mill.constants`
- `mill.runner.api`
- `mill.api`
- `mill.define`
- `mill.util`
- `mill.main`
- `mill.*lib`

And _not_ including

- `mill.codesig`
- `mill.exec`
- `mill.resolve`
- `mill.internal`
- `mill.server`
- `mill.server.client`
- `mill.client`
- `mill.runner`

Currently the dependencies of `build.mill` are much broader than this, but that should be fixable.

Some things may need to be refactored:

- `mill.runner` should be broken up. Meta-build `build.mill` files need the stuff necessary for `MillBuildRootModule`, but it shouldn't need the `mill.server` stuff or `mill.bsp` stuff that also gets pulled in
- `mill.runner.api` and `mill.api` could probably be combined, with the parts of `mill.api` that don't fit being moved to `mill.define`

Once we have tightened things up, we should have an integration test that scans the `build.mill` classpath and asserts that the jars or folders that are present are as we expect

PR #4906, opened 2025-04-10
Add integration test for `build.mill` classpath

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/4885

Not quite as clean as I would like, as `mill-core-internal_3.jar` is still on the classpath. But it's pretty good overall

This PR also fixes handling of `local-test-overrides`, which previously was incorrectly applying to `.packaged` tests


lihaoyi:
Tests passed on my fork, merging this

PR #4907, opened 2025-04-10
Fix some recurrent CI errors

alexarchambault:
This should fix different kinds of CI errors that we run into quite often:
- some YAML / workflow definition errors
- junit tests failing on purpose in Mill test cases, that were reported as CI errors in the workflow summary page by the action-junit-report action
- GraalVM images, downloaded in the build-windows job, that went missing in the examples / integration jobs (fixed by revalidation of some `PathRef`s)
- Android linting errors printed on purpose by the Android tests, that were matched anyway by problem matchers set up by the setup-android action, and reported as errors on the workflow summary page
- timeout errors in the BSP tests (this one was my own mistake initially 😅)
- JVM version and JVM index-related errors: pinning the index version isn't enough for the exact same JVM version (like `17.0.14`) to be picked when passed only a major version (like `17`). If a lower version is already in cache, coursier uses it, except if `JavaHome#update` is set to true, which we do here. (This led to issues in the main repo CI, that had older JVMs in cache, while forks' CI were usually fine)


lefou (approved):
Nice catches and fixes! Hoping, CI likes it too.

alexarchambault:
> Hoping, CI likes it too.

The CI on my fork liked it, the one here has still some red :/

alexarchambault:
I think I understand why [this job](https://github.com/com-lihaoyi/mill/actions/runs/14391262722/job/40358992977) was failing on the main repo CI and not on the CI of my fork (weird interplay between the `coursierarchive` param in the workflow, how the JvmCache in coursier selects the highest already-in-cache JVM version when given a short version like `17`, and the state of caches fought by coursier/cache-action here vs on my fork). I could probably manage to get the CI green here without the "Copy GraalVM under shorter path on Windows CI" and "Remove coursierarchive thing" commit.

alexarchambault:
Seems the CI is green here, and on my fork too ([this PR branch](https://github.com/alexarchambault/mill/actions/runs/14404961684/job/40399769232) about to be green, [different branch with same content](https://github.com/alexarchambault/mill/actions/runs/14404941303/job/40399739603) is green already)

The latest changes, to `NativeImageModule`, might need careful review. They basically add a `nativeImageMaybeCopyGraal` that copies the GraalVM distribution about to be used to a shorter path, like `C:\jvms\5aa5d09c`, `5aa5d09c` being the hash of the path of the original GraalVM distribution. This only happens if `NativeImageModule#nativeImageGraalVmCopyBase: Option[os.Path]` is non-empty. It's non-empty by default only in Windows CI environments, where it's `Some(os.Path("C:/jvms"))`. Then, `NativeImageModule#nativeImageTool` is able to automatically pick this copied distribution if it finds it.

lihaoyi:
For the graal copying stuff, I wonder if it's possible to make it more seamless? e.g. could we make give coursier the option to give a shorter path to the downloaded files? This won't just be a problem in our CI, but might also cause problems to users on windows who want to create Graal images

alexarchambault:
> For the graal copying stuff, I wonder if it's possible to make it more seamless? e.g. could we make give coursier the option to give a shorter path to the downloaded files? This won't just be a problem in our CI, but might also cause problems to users on windows who want to create Graal images

Mill users can benefit from it too, given it's done in `NativeImageModule`. They just have to run the `nativeImageMaybeCopyGraal` command on it.

alexarchambault:
> I think maybe we can separate the graal stuff from the rest of this PR. The rest of this PR is relatively straightforward internal fixes, the graal stuff is a bit more user facing and should be discussed and reviewed separately

Yeah, I'm moving this stuff to a separate PR

alexarchambault:
Going to go ahead and merge, unless anyone want to have a closer look

lihaoyi:
go fori t

Issue #4863, opened 2025-04-04
`dist.run` not printing expected stdout


lihaoyi:
```
$ ./mill -i dist.run example/javalib/basic/1-simple -i show version
[3454/3454] dist.run
============================== show version ==============================
[1/1] show
[1/1] ============================== show version ==============================
[3454/3454] ========================= dist.run example/javalib/basic/1-simple -i show version ======================== 4s
```

I would expect this to print `0.13.0-M1-2-c0b707` or `SNAPSHOT` or something, but nothing gets printed at all

PR #4909, opened 2025-04-10
Fix `dist.run` stdout forwarding

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/4863

Seems like it was broken by #4837 which replaced `Jvm.runSubprocess` with `os.call`, which unfortunately had different stream forwarding defaults. This class of problem should be impossible in future thanks to https://github.com/com-lihaoyi/mill/pull/4456 from @sake92 which standardizing the stream forwarding defaults of `Jvm.*Process` APIs with `os.*`


lefou:
Yeah. I run in similar issues in other project. I though the default where already aligned as part of the `Jvm` API overhaul. My bad.

Issue #4778, opened 2025-03-25
Add `--offline` mode on best effort basis


lefou:
The idea is to have a new `--offline` option, which instructs Mill to try to work offline (if possible).

We could expose it via `mill.api.Ctx.offline`, such that task can try to adapt their behavior accordingly.

E.g. `CoursierModule.defaultResolver` could use a different coursier config which doesn't check for updates (e.g. ignore TTL values) and fails in case some resource isn't available offline.

PR #4914, opened 2025-04-11
Introduce a `--offline` option and expose it via `TaskCtx`

lefou:
This adds support for an offline mode which instructs Mill to try to work offline (if possible). It can be enabled via the `--offline` CLI option.

The offline mode is exposes via `mill.define.TaskCtx.offline`, such that tasks can try to adapt their behavior accordingly. Since Mill is extensible, we can't fully control whether extensions really respect when the user wants to be offline. With the deprecation of the `SecurityManager` in newer Java versions, there is also no means to detect and fail hard, in case some code tries to access the internet. Nevertheless, in the Mill code-base we should aim to fully support and respect the offline mode. Beside, offline mode will always on a best-effort basis, as it is typically for other build tools like Maven or Gradle.

Since `mill --offline` won't be able to fetch dependencies, users can use the `OfflineSupportModule.prepareOffline` task, to download and cache remote dependencies ahead of time.

As a start, this PR implements offline mode for all `CoursierModule`-based dependency operations. If offline, it configures the cache manager of coursier to work in local-only mode.

Additionally the `mill.scalalib.Dependency` module and the `mill.scalalib.dependency` package were revised to respects the offline mode.

I also reviewed all occurrences of `requests.get` and made the respective tasks offline aware.

Other places we need to review in follow up PRs are
* artifact publishing, e.g. `PublishModule` and the various repository handlers
* all uses of the `java.net.http.HttpClient`
* all uses of `URL.openConnection` API

Fix https://github.com/com-lihaoyi/mill/issues/4778


lihaoyi:
Maybe we should give it a more descriptive name like `--coursier-offline` or similar? `--offline` seems a bit generic for what looks like a very specific use case.

I'm also wondering whether this should be a top-level flag at all; could we make it e.g. a JVM property instead that a `Task.Input` reads? Top-level flags are both scarce as well as especially powerful, and are typically used for cross-cutting configuration that affects all tasks or evaluation. This seems a bit specific and I wonder if there's another way we can expose this kind of more specific configuration options to the user

lefou:
The idea is to support `--offline` everywhere, not only in coursier. E.g. we should guard `requests.get` when offline. I just started with coursier, since this is our main way to fetch dependencies so it has the biggest impact. Having a `--offline` CLI option is also state of the art. Maven has it, Gradle has it, scala-cli has it. Lots of other tools have it.

lihaoyi:
In that case you should explain in the PR description: what is `--offline` meant to do, broadly? What does it do in other tools? What does this PR do, and what does it not do yet?

lefou:
I updated the PR description. As a start, I implemented offline mode for all `CoursierModule`-based dependency operations. If offline, the cache manager of coursier is set to work in local-only mode.

Additionally the `mill.scalalib.Dependency` module and the `mill.scalalib.dependency` package were revised to respects the offline mode.

I also reviewed all occurrences of `requests.get` and made the respective tasks offline aware.

Other places we need to review in follow up PRs are
* artifact publishing, e.g. `PublishModule` and the various repository handlers
* all uses of the `java.net.http.HttpClient`
* all uses of `URL.openConnection` API

lihaoyi:
What does "offline aware" mean? Is it meant to just fail if `--offline` is provided, or do we expect these code paths to have fallback logic to work differently if `--offline` is provided?

lefou:
> What does "offline aware" mean? Is it meant to just fail if `--offline` is provided, or do we expect these code paths to have fallback logic to work differently if `--offline` is provided?

We basically allow each task to be aware of the user request to stay offline and handle it accordingly. For coursier, this not necessarily means to fail, as we can try to use already present resources and only fail if something is missing. For remote-publishing this means to fail. But in the ends, it's up to each specific task to decide how to handle it.

lihaoyi:
Sounds good!

scalalib/src/mill/javalib/android/AndroidSdkModule.scala
```diff
@@ -59,10 +66,22 @@ trait AndroidSdkModule extends Module {
...
+    // TODO: Use caching API once available, https://github.com/com-lihaoyi/mill/issues/3930
+    val cache = FileCache()
+      .pipe { cache =>
+        if (Task.offline) cache.withCachePolicies(Seq(LocalOnly)) else cache
+      }
+    cache.file(Artifact(url)).run.unsafeRun()(cache.ec) match {
+      case Right(file) =>
+        Result.Success(PathRef(os.Path(file)).withRevalidateOnce)
+      case Left(ex) if Task.offline =>
+        Result.Failure(s"Can't fetch bundle tools (from ${url}) while in offline mode.")
```
lefou, line 80: This is me trying to get a feeling for the current coursier cache API. I'd like to work on that API next. Ideally, we can built-in offline suport.

Avimitin:
Is it possible to dump coursier or other dependencies information, like URL, name and sha256 sums, when user trigger `_.prepareOffline`. And add a integrity check to those dependencies when user run any target with `--offline` flag.

lefou:
@Avimitin can you please a new issue(s)? This PR is already merged.

PR #4916, opened 2025-04-11
More classpath/bsp/idea cleanup

lihaoyi:
* Remove `mill.internal` from `build.mill` classpath
* Move `main.util` to `core.util`
* Move `core` to `core.eval`
* Remove hardcoded usage of `mill-dist` in BSP/IDEA integration, instead we just treat it as any other module
* Uncomment `buildTargetCleanCache` and `buildTargetScalaTestClasses`
* Syntax cleanups for `GenIdeaImpl`
* Remove redundant `.toSeq`s throughout `MillBuildServer` since all `Agg`s are now replaced by `Seq`s by default
* Consolidate duplicate `def withOutLock`s
* Move `mill.main.server` to `mill.server` for consistency with filesystem path


lihaoyi:
Merging this since it's green on my fork's CI

PR #4917, opened 2025-04-11
Copy GraalVM under shorter path on Windows CI

alexarchambault:
This adds a `nativeImageMaybeCopyGraal` command on `NativeImageModule`. This command only has an effect if `NativeImageModule#nativeImageGraalVmCopyBase: Option[os.Path]` is non-empty. By default, it's non-empty only in Windows CI environments, where it's `Some(os.Path("C:/jvms"))`. `nativeImageMaybeCopyGraal` copies the GraalVM distribution under a sub-directory of it, like `C:\jvms\5aa5d09c`. `5aa5d09c` is a hash of the original GraalVM distribution path.

If this directory exists, `NativeImageModule#nativeImageTool` uses it rather than the original GraalVM distribution, to get the path of `native-image`. This offers a workaround for too long paths of GraalVM header files, that we run into sometimes on the CI (more on fork CIs than in the main repo, in my experience).

This allows to get rid of the coursier archive cache customizations on the CI, that were meant to work around that problem too. The resulting paths were sometimes still too long, given they contain the URL of the GraalVM distribution, like `C:\coursier-arc\https\github.com\graalvm\graalvm-ce-builds\releases\download\jdk-23.0.1\graalvm-community-jdk-23.0.1_windows-x64_bin.zip\`. In constrast, with this PR, we get GraalVM distribution paths like `C:\jvms\5aa5d09c`.


alexarchambault:
This also copies the changes in `NativeImageModule` in `dist/package.mill`, so that we can benefit from them right away. Applying `ci/mill-bootstrap.patch` when re-bootstrapping should remove those.

alexarchambault:
About https://github.com/com-lihaoyi/mill/pull/4907#issuecomment-2797217209, this could be added in coursier. Maybe with something like
```scala
ArchiveCache().withShortPathDir(Some(new File("C:/jvms")))
```

That way, we'd only have to keep `NativeImageModule#nativeImageGraalVmCopyBase` here (maybe moved to `JvmWorkerModule`)

lihaoyi:
Looking at this PR it feels like it's exposing too much complexity in the API: asking the user to configure an explicit path, and then asking the user to call a separate command to copy things to that explicit path, both seem like incidental complexity that we should try and avoid.

Are there any other options? e.g. Can we always copy the Graal binary to a shorter path in `~/.cache/mill/graal/<hash>` before running it?

alexarchambault:
i reverted all changes, and made it use https://github.com/coursier/coursier/pull/3373 (this needs coursier `2.1.25-M5`, whose release is on-going). Going to try to enable it in the bootstrap job via `ci/mill-bootstrap.patch` (and we should benefit from it here after re-bootstrapping).

main/util/src/mill/util/Jvm.scala
```diff
@@ -617,12 +617,19 @@ object Jvm {
...
       ctx: Option[mill.define.TaskCtx.Log] = None,
       coursierCacheCustomizer: Option[FileCache[Task] => FileCache[Task]] = None,
-      jvmIndexVersion: String = "latest.release"
+      jvmIndexVersion: String = "latest.release",
+      useShortPaths: Boolean = false
   ): Result[os.Path] = {
     val coursierCache0 = coursierCache(ctx, coursierCacheCustomizer)
+    val shortPathDirOpt = Option.when(useShortPaths) {
+      if (isWin) os.Path(System.getenv("UserProfile")) / ".mill/cache/jvm"
+      else os.home / ".cache/mill/jvm"
```
lihaoyi, line 626: This snippet here looks a bit unusual. I thought we just use `os.home` for most things? Or are we using `UserProfile` in other parts of the code as well?

scalalib/src/mill/scalalib/JvmWorkerModule.scala
```diff
@@ -51,6 +53,11 @@ trait JvmWorkerModule extends OfflineSupportModule with CoursierModule {
 
   def zincLogDebug: T[Boolean] = Task.Input(Task.ctx().log.debugEnabled)
 
+  def useShortJvmPath(jvmId: String): Boolean =
+    Properties.isWin && (
+      jvmId.startsWith("graalvm") || jvmId.startsWith("liberica-nik")
```
lihaoyi, line 58: What's `liberica-nik`?

main/util/src/mill/util/Jvm.scala
alexarchambault: Yes, `mill.bat` uses it [here](https://github.com/com-lihaoyi/mill/blob/58ab3d68b33b7ddaf6b63165bf5a08571494923b/mill.bat#L85). I also [used it](https://github.com/com-lihaoyi/mill/blob/58ab3d68b33b7ddaf6b63165bf5a08571494923b/runner/client/src/mill/runner/client/JLineNativeLoader.java#L41) in `JLineNativeLoader`, added in https://github.com/com-lihaoyi/mill/pull/4882.

Ideally, that directory should be obtained by a native Windows call rather than via the environment. Coursier does it via [directories-jvm](https://github.com/coursier/directories-jvm), but it's quite cumbersome (requires JNI or newer JVM foreign API stuff, or running a powershell script…). Given `mill.bat` relies on the env for that, it's better just to do the same in Mill IMO.

scalalib/src/mill/scalalib/JvmWorkerModule.scala
alexarchambault: It's an alternative GraalVM distribution. Coursier uses it, as it has support for more platforms (like Alpine Linux on arm64, but not Windows on arm64 though)

alexarchambault:
Recently run into this issue on my fork [here](https://github.com/alexarchambault/mill/actions/runs/14642052919/job/41087262486) and [here](https://github.com/alexarchambault/mill/actions/runs/14642052919/job/41087262442), with a somewhat cryptic error message looking like
```text
dist.native.nativeImage os.SubprocessException: Result of C:\coursier-arc\https\github.com\graalvm\graalvm-ce-builds\releases\download\jdk-23.0.1\graalvm-community-jdk-23.0.1_windows-x64_bin.zip\graalvm-community-openjdk-23.0.1+11.1\bin\native-image.cmd�: 1
```

lihaoyi (approved):
looks good to me

core/util/src/mill/util/Jvm.scala
```diff
@@ -620,12 +620,19 @@ object Jvm {
...
       ctx: Option[mill.define.TaskCtx] = None,
       coursierCacheCustomizer: Option[FileCache[Task] => FileCache[Task]] = None,
-      jvmIndexVersion: String = mill.api.BuildInfo.coursierJvmIndexVersion
+      jvmIndexVersion: String = mill.api.BuildInfo.coursierJvmIndexVersion,
+      useShortPaths: Boolean = false
   ): Result[os.Path] = {
     val coursierCache0 = coursierCache(ctx, coursierCacheCustomizer)
+    val shortPathDirOpt = Option.when(useShortPaths) {
+      if (isWin) os.Path(System.getenv("UserProfile")) / ".mill/cache/jvm"
+      else os.home / ".cache/mill/jvm"
```
lefou, line 629: This should respect the `XDG_CACHE_HOME` env var. We respect it in other places like the bootstrap scripts and I think coursier also repects it.

alexarchambault: Should be fixed

core/util/src/mill/util/Jvm.scala
```diff
@@ -620,12 +620,23 @@ object Jvm {
...
-      jvmIndexVersion: String = mill.api.BuildInfo.coursierJvmIndexVersion
+      jvmIndexVersion: String = mill.api.BuildInfo.coursierJvmIndexVersion,
+      useShortPaths: Boolean = false
   ): Result[os.Path] = {
     val coursierCache0 = coursierCache(ctx, coursierCacheCustomizer)
+    val shortPathDirOpt = Option.when(useShortPaths) {
+      if (isWin)
+        os.Path(System.getenv("UserProfile")) / ".mill/cache/jvm"
+      else {
+        val cacheBase = sys.env.get("XDG_CACHE_HOME").map(os.Path(_)).getOrElse(os.home / ".cache")
```
lefou, line 631: Two things.

1. When `ctx` is defined, `ctx.env` is probably the better source for environment variables.
2. We should either use `System.getenv` or `sys.env.get` but not both.

lefou:
```scala
val env = ctx.map(_.env).getOrElse(sys.env)
```

alexarchambault: We can check `ctx.env` for `XDG_CACHE_HOME`, but I'm more wary of it for `UserProfile`. On Windows, env var names are case-insensitive, so that the path can be set as `Path` for example. In that case, `System.getenv("PATH")` respects the case-insensitiveness on Windows, and finds the variable. `sys.env.get("PATH")` doesn't. And IIUC, `ctx.env.get("PATH")` wouldn't either.

So I'd stick to `System.getenv` for `UserProfile`, as this is used on Windows only. But `ctx.env` and both `System.getenv` and `sys.env` are fine outside of Windows, for `XDG_CACHE_HOME`.

lefou: Can you please add a comment, since this isn't obvious. 

Also, I'd assume Mill already suffers form this nifty details. Ideally, we have a map with case-insensitive keys semantics in `TaskCtx.env`. Also, `sys.env` seems to be a bad abstraction when it doesn't work correctly under Windows, so we should always use `System.getenv`. Nothing we can solve here, but I think we should review Mill regarding this issue.

lefou: Looks like the case-insensitive lookup is already implemented in `sys.env`.

https://www.scala-lang.org/api/current/scala/sys.html#env-0

> An immutable Map representing the current system environment.
>
> If lookup fails, use System.getenv(_) for case-insensitive lookup on a certain platform. If that also fails, throw NoSuchElementException.

alexarchambault: Just checked that again in a VM to be 100% sure

Capture d’écran 2025-04-25 à 14 53 22

The JDK sources [confirm that too](https://github.com/openjdk/jdk11u/blob/856bab316a63ce3acddd08d3fbe76b9c3223cdb6/src/java.base/windows/classes/java/lang/ProcessEnvironment.java#L209-L210).

lefou (approved):

lihaoyi:
@alexarchambault seems like this has caused some tests to break in `main` https://github.com/com-lihaoyi/mill/actions/runs/14680562076/job/41202842913

alexarchambault:
> @alexarchambault seems like this has caused some tests to break in `main` https://github.com/com-lihaoyi/mill/actions/runs/14680562076/job/41202842913

These were broken before merging this PR. The broken task is `dist.native.nativeImage`, not an example or an integration test. The changes here couldn't break this task (not until a re-bootstrap, that is). I'd expect a re-bootstrap to actually fix those jobs (thanks to the PR here).

lihaoyi:
Publishing seems to be making more progress now since the revert https://github.com/com-lihaoyi/mill/actions/runs/14700178878/job/41248197683

alexarchambault:
Apparently, the coursier version bump seems to be the problem, rather than the use of shorter paths in `NativeImageModule` (that wasn't used in the build of Mill itself here). https://github.com/com-lihaoyi/mill/pull/5015 has issues with long GraalVM paths both on the CI of both my fork and the main repo. I suspect some weird issues with reproduciblity and the state of the coursier cache for JVMs, but can't pinpoint where exactly.

But enabling the use of short GraalVM paths alongside bumping coursier (making sure the Mill build itself relies on shorter paths this time) seems to work, see https://github.com/com-lihaoyi/mill/pull/5018.

PR #4918, opened 2025-04-11
Add ability to define `PackageExternalModule`s that can be called via just `pkg.path/` rather than `pkg.path.ObjectName/`

lihaoyi:
This simplifies a common case where there is only one `ExternalModule` in a package, e.g. 

1. `./mill mill.scalalib.scalafmt.ScalafmtModule/` can now be called via `./mill mill.scalalib.scalafmt/`
2. `./mill mill.kotlinlib.ktlint.KtlintModule/` can now be called via `./mill mill.kotlinlib.ktlint/`
3. `./mill mill.javalib.palantirformat.PalantirFormatModule/` can now be called via `./mill mill.javalib.palantirformat/`
4. `./mill mill.idea.GenIdea/` can now be called via `./mill mill.idea/`

This removes a huge source of verbosity from the Mill command line interface, and basically fixes one of Mill's oldest issues https://github.com/com-lihaoyi/mill/issues/299 by default. Referencing external modules via their package path is sufficiently concise that it basically removes the need for defining your own `def` external module aliases, though you can always define your own aliases to make it more concise still if you so desire

For cases where there is more than one `ExternalModule` in a single `package`, they will need to be called fully qualified e.g. `./mill mill.scalalib.PublishModule/`

`PackageExternalModule` is the spiritual equivalent to Scala `package object` or Mill `TaskModule#defaultCommandName`, in that it allows you to take a `package` in the hierarchy and designate one direct child as the "blessed" entrypoint for that `package`.

The trailing `/` in the selector is still a bit annoying and weird, and I might try making it optional a follow up PR. 

## Implementation

The "blessed" `ExternalModule` in a package is specified by defining a 

```scala
object PackageExternalModule extends mill.define.PackageExternalModule(MyExternalModule)
```

Where `MyExternalModule` is the actual `ExternalModule` in question. The `Resolve` logic when given a `foo.bar.baz/` selector first looks for the class `foo.bar.baz.PackageExternalModule$` before looking for the class `foo.bar.baz$`. 

This adds a tiny amount of ambiguity in what a selector `foo.bar.baz` means, but in practice I don't expect it to be a problem. I expect such collisions between normal `ExternalModule` names and `PackageExternalModule` names to happen basically never, so from a practical perspective there will only ever be one obvious target for each selector

I chose the indirection of defining a separate `PackageExternalModule` object over renaming all the various external modules to `PackageExernalModule`. Renaming them is possible, but then we lose the meaningful name when referencing them in code, and breaks backwards compatibility with the fully `foo.bar.baz.MyExternalModule/` selector being passed at the command line.

Using an `object PackageExternalModule` as a marker lets us avoid doing any sort of classpath scanning, which would be required if we wanted to use an `@Annotation` or marker-trait for designating the primary `ExternalModule` for a `package`. It just adds one additional `ClassLoader#loadClass` check, which is much cheaper than scanning the classpath with `classgraph` to try and find classes with particular annotations

## Testing

Added a unit test case to cover this new resolution behavior, and updated all the eligible example tests to use the shorter selector.

This could be backported to 0.12.x since it is backwards compatible


core/define/src/mill/define/ExternalModule.scala
```diff
@@ -22,3 +22,5 @@ abstract class ExternalModule(implicit
     Segments(millModuleEnclosing0.value.split('.').map(Segment.Label(_)).toIndexedSeq)
   }
 }
+
+class PackageExternalModule(val value: ExternalModule)
```
lefou, line 26: Can we have some ScalaDoc here?

lefou: Should we have the work "default" in the class name? It's IMHO a expected catch-word. We also have `TaskModule.defaultCommandName`. 

Maybe: `DefaultExternalModule` or `DefaultModule`. `DefaultPackageExternalModule` is too long.

lihaoyi: Renamed it `PackageDefaultExternalModule extends ExternalModule.Alias` and added some scaladoc, a bit long but not too bad. Still a one-liner

lefou: I love the `Alias` idea! (Thinking if we somehow could allow it on all real modules, so we could give users the chance to lift their entry points (often burried deep in the hierarchy) into some other locations. This would reduce the need for local `justfile`s or `Makefile`s.)

The name is rather long honestly. We need a `Default` for a `Package` which is a  `Module`. Since it must be some instance, it doesn't matter whether it's a `ExternalModule` anymore, so the `External` is the superfluous part in that name.

lihaoyi: For now it's basically only useful for `ExternalModule`s, since the only time it comes into lay is when resolving external modules via `foo.bar.baz/`. It doesn't work for modules in `build.mill` for example, and we already have `def` for defining aliases inside your `build.mill` file

lihaoyi: the current documented workflow for aliasing external modules https://mill-build.org/mill/fundamentals/modules.html#_aliasing_external_modules

Issue #4045, opened 2024-11-28
Rename ivy stuff


alexarchambault:
It seems using "Ivy" to refer to dependencies is confusing users. All the more that
- Apache Ivy isn't used any more to fetch dependencies
- 99.9% of the time, these refer to dependencies hosted on Maven repositories (putting `~/.ivy2/local` aside)

I'm not sure what should replace it: just "dependencies" seems right, but might confuse people using Mill for non-JVM languages. "Maven dependencies" or "JVM dependencies" might work. That could give:
```scala
def jvmDeps = Agg(
  dep"com.lihaoyi::os-lib:0.11.3"
)
// or
def jvmDeps = Agg(
  jvm"com.lihaoyi::os-lib:0.11.3"
)
```

alexarchambault:
Plus no one refers to those as "Ivy dependencies" in JVM ecosystems AFAIK. "Maven dependencies" seems to be more used.

lihaoyi:
I agree

lefou:
I think `mvn` would be a nice option since we target Maven repositories. `dep` is a bit to generic, and  jvm a bit misleading (we don't select JVMs and we also use if for Scala.JS and Scala Native artifacts). We could introduce the alternative `StringContext` right away. No need to wait for `0.13`.

lihaoyi:
We should rename ivyDeps at the same time? That one may be more tricky to do backwards compatibly, but we could have a new `def mvnDeps` task that defaults to `def ivyDeps` so people can override either

lihaoyi:
There's a lot of `ivy` terminology spread throughout Mill, might be easier to do a big-bang rename in 0.13.0 rather than trying to shim `jvm` versions of everything

megri:
I'm a bit off when it comes to the finer details of maven- versus ivy-deps, but wouldn't `mvnDeps` be confusing as well if artefacts are being pulled from let's say sonatype? Why not just `deps` ?

On the same note, one nice thing that could come with having string context prefixes would be to solve the "colon-hell" (which unintentionally sounds like a bowel affliction); that is we could have `java"group:artifact:version"`, `scala"group:artefact:version"`, `sjs"group:artefact:version"` etc without fiddling around with colons. It may be too many variants to be comfortable, but it's worth considering.

lefou:
[...]

PR #4930, opened 2025-04-13
Rename `ivyDeps` to `mvnDeps`

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/4045

This PR replaces the `ivyDeps` nomenclature throughout Mill with `mvnDeps`. This is a more accurate name, and the old name has long been both confusing and inaccurate since Mill stopped using Ivy as the dependency resolver many years ago.

An alternative I considered was `libraryDeps`: this has the upside of being more consistent with SBT, but has the downside of being very long as a string interpolator `library"..."`. And we can't easily get rid of the string interpolator due to how type inference works, as implicit conversions do not apply to the elements of a `Seq(...)` when the `Seq` itself is being implicitly converted to `Task[Seq]`

We considered `jvmDeps`, but that's a bit confusing for things like Scala.js/Scala-Native/Kotlin.js/etc. that aren't JVM

For now I left `import $ivy` in place, but I'm hoping to replace that entirely as part of https://github.com/com-lihaoyi/mill/issues/3062. No need to both renaming it now when it'll hopefully be gone soon anyway


lefou:
I don't like `jvmDeps`, mostly because there is not JVM-blessed repository. There is a JVM packaging system though, but that's something completely different.

I propose `mvnDeps` instead. We deal with Maven dependencies and get them from Maven repositories. (If we look closely, we still support Ivy repositories, but we mostly just support those features which are compatible to Maven anyway.)

lihaoyi:
@lefou I renamed it `mvn""`/`mvnDeps`

lihaoyi:
Updated the PR description

joan38:
Why don't we align to scala cli `dep""`?

PR #4931, opened 2025-04-13
Added MILL_TEST_FREE_PORT  

albassort:
Made in accordance to https://github.com/com-lihaoyi/mill/issues/3802


albassort:
RE: There is some failing tests on HEAD right now. I think 

> [Run Tests / windows (17.0.14, 'integration.bootstrap[no-java-bootstrap].native.server') / run (pull_request)](https://github.com/com-lihaoyi/mill/actions/runs/14433601322/job/40471544846?pr=4931)Failing after 9m

Might be related to the PR but im not sure. I will be investigating,

core/constants/src/mill/constants/EnvVars.java
```diff
@@ -52,4 +52,8 @@ public class EnvVars {
    * the local disk and not from Maven Central
    */
   public static final String MILL_BUILD_LIBRARIES = "MILL_BUILD_LIBRARIES";
+  /**
+   * Used to store free ports for Mill's tests to use.
+   */
+  public static final String MILL_TEST_FREE_PORT = "MILL_TEST_FREE_PORT";
```
lefou, line 58: Is this a single port or multiple. The comment is a bit vague. If it is multi, that the variable should be `MILL_TEST_FREE_PORTS` and we should document here, which separator is used.

runner/src/mill/runner/MillMain.scala
```diff
@@ -348,6 +362,69 @@ object MillMain {
...
+              None
+            }
+          }
+        }
+        Some(ports)
+      } else {
+        var i = 0
+        var ports = Set.empty[Int]
+        for (_ <- 1 to 100) {
+          if (i == 6) {
```
lefou, line 403: Magic number `6`! A parameter or a named constant might be better.

runner/src/mill/runner/MillMain.scala
```diff
@@ -348,6 +362,69 @@ object MillMain {
...
+              i += 1
+            }
+
+          } finally {
+            socket.close()
+          }
+        }
+
+        streams.err.println(
+          "Failed to generate enough ports for servers to use. Either there isn't 6 free ports on your computer, or there is a firewall issue. "
```
lefou, line 420: Magic number `6`.

albassort: > Magic number `6`.

Fixed but, maybe we don't need to do this at all.

lihaoyi:
This doesn't satisfy the original issue

> The end result is that Mill's own example suite can remove all hardcoded ports and replace them with usages of the MILL_TEST_FREE_PORT environment variable

albassort:
> This doesn't satisfy the original issue
> 
> > The end result is that Mill's own example suite can remove all hardcoded ports and replace them with usages of the MILL_TEST_FREE_PORT environment variable

I misunderstood, I thought that were to be done in a second PR. I will work on this. Setting as draft until that is done.

albassort:
@lihaoyi 

So there is numerous limitations with what I have here.

1. The PR uses the PORT environmental variable
2. Ports are ephemeral with no central manager for the following reasons
    1. The desire to have it in a ExternalModule, any build.mill can access the code 
    2. The lack of any existing central manager for Tasks as well was UTestManager
    3. The lack of access to the ExecutionContext from the UTestManager
 3. Only works for UTESTFramework
 
I experimented with adding some code in ExecutionImpl, thats where it could be managed if I knew 
1. How I can add such env variable from GroupExecute or Execute0. There the ExecutionContext can be modified 
2. How to route the ExecutionContext into UTestFramework. From what I can tell, there is no way
3. How I can coordinate from all parts of the code to a specific location easily. This is also an issue in my other PR. I have no way of getting the out/ folder from all parts in the codebase, so I cant move the code to the testkit or server or runner and use a lock to coordinate ports over a JSON port.

Still, you can easily add some code to Task.Command {} (and I did) To allow a centrally managed Port dispatcher for those

---

Regardless, I did manage to replace hard coded ports for most of the TESTS. However, there is still the chance there can be conflicting ports.

lihaoyi:
I must admit I am having trouble following the code. There's a bunch of random `"!"` strings, an apparently unused `MILL_TEST_FREE_PORT` variable, and other things. 

As far as I can tell, the current implementation is unique to UTestExampleTestSuite. We should try and make it work with any `Task` with some kind of `ports = true` or `ports = 1` flag, so it's useful to users as ewll

albassort:
> I must admit I am having trouble following the code. There's a bunch of random `"!"` strings, an apparently unused `MILL_TEST_FREE_PORT` variable, and other things.
> 
> As far as I can tell, the current implementation is unique to UTestExampleTestSuite. We should try and make it work with any `Task` with some kind of `ports = true` or `ports = 1` flag, so it's useful to users as ewll

I will mark as WIP until I figure out a way to make it work with all Tasks

PR #4934, opened 2025-04-14
fix android cmd-line tools on CI

irodotos7:
## Motivation

- It seems like, github actions has already installed an old version of the android tools, so the android-setup action was running and trying to install in the same locations as the pre-installed tools ( android-setup couldn't override the locations but instead it created a new folder and installed it there )
- The old pre-exsisting android tools doesn't have the R8 tool 

## Solution

Change the location of where android home is located and change the cmd-line tools version to latest

### DIFF
--- a/.github/actions/setup-android/action.yml
+++ b/.github/actions/setup-android/action.yml
@@ -3,11 +3,15 @@
 runs:
   using: "composite"
   steps:
+    - name: Set ANDROID_SDK_ROOT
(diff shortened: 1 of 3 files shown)


vaslabs:
@irodotos7 screenshot tests have a classpath issue, I didn't get into fixing that yet, so it's not gonna pass

vaslabs:
nice one, seems the android job is more stable with this updates .

vaslabs:
@lefou @lihaoyi this is green, we can get it in so we can update our PRs with these fixes

PR #4944, opened 2025-04-15
Add helpers to debug flaky tests, that might have fixed them

alexarchambault:
This adds more checks in two currently flaky tests, and makes them print some details when they fail. My hope is this could help understand what's going on when they fail for no obvious reason. This also retries the failing assertions in those tests after a delay, which might fix things.


alexarchambault:
@lihaoyi @lefou Don't know if you want to have a look. Going to merge if the CI is green else, I guess. This adds some debugging messages when those flaky tests fail, but also might fix things, thanks to the small delay before checking the assertions again.

lihaoyi:
go for it

PR #4951, opened 2025-04-16
Failing test for broken ANSI escape handling in `LinePrefixOutputStream`

lihaoyi:
If the ANSI escape is split over multiple writes, the current color tracking fails


lihaoyi:
The issue seems to be that `flush()` calls `writeOutBuffer()`, so a single line may be written more than once if broken into multiple `write`s with `flush`es in between, and if the ANSI escape code is broken over the multiple writes then the `fansi.Str.apply(` call does not properly recognize the partial escape codes on each side of the flush

lihaoyi:
This is no longer relevant since we no longer log partial lines to the terminal, and complete lines cannot have a partial ansi escape code at the end because they must end with a newline

Issue #4716, opened 2025-03-13
Review all deprecations before cutting Mill `0.13.0`


lefou:
Ideally, we can backport all deprecations into `0.12.x` and remove them for Mill `0.13.0`. All that's needed then, is to tell users to first upgrade to latest Mill `0.12.x` before upgrading to `0.13`.

lihaoyi:
Another important thing to do here is take a pass at `private`izing everything that can be made `private`, to allow easier evolution later in the 0.13.x cycle

lihaoyi:
Closing this since 1.0.0 is out

PR #4954, opened 2025-04-18
Move `ScalaFmt` into its own classloader

lihaoyi:
Cuts `mill-assembly.jar` from 55mb to 53mb, so ~2mb of savings, with a somewhat cleaner classpath. the `scalafmt-dynamic` code will be downloaded on-demand

### DIFF
--- a/scalalib/src/mill/scalalib/scalafmt/ScalafmtWorker.scala
+++ b/scalalib/src/mill/scalalib/scalafmt/ScalafmtWorker.scala
@@ -1,20 +1,33 @@
 package mill.scalalib.scalafmt
 
-import mill._
-import mill.define.{Discover, ExternalModule, Worker}
-import mill.define.TaskCtx
-import org.scalafmt.interfaces.Scalafmt
+import mill.*
+import mill.define.{Discover, ExternalModule, PathRef, TaskCtx, Worker}
+import mill.scalalib._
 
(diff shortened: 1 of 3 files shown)


scalalib/src/mill/scalalib/scalafmt/ScalafmtWorker.scala
```diff
@@ -1,20 +1,33 @@
...
+  /**
+   * Classpath for running Palantir Java Format.
+   */
+  def scalafmtClasspath: T[Seq[PathRef]] = Task {
+    defaultResolver().classpath(
+      Seq(mvn"org.scalameta:scalafmt-dynamic_2.13:${mill.scalalib.api.Versions.scalafmtVersion}")
+    )
+  }
+
+  def scalafmtClassLoader: Worker[java.net.URLClassLoader] = Task.Worker {
```
lefou, line 21: Nice trick to have the classloader in its own worker, since this avoids any explicit  ÀutoCloesable` handling. I think we should keep this `private` though.

lihaoyi: Whoops sorry I missed this, I guess we may come back to it as part of https://github.com/com-lihaoyi/mill/issues/4716 when we go around privatizing everything we can

PR #4955, opened 2025-04-18
Tighten up `mill-runner-meta` dependencies

lihaoyi:
Also enabled assertions for the meta-build in the `GenIdeaExtended` tests, so we have some test coverage of what's on the classpath for the meta-build


lihaoyi:
Seems no more broken then `main`, so going to merge this

PR #4957, opened 2025-04-18
Send BSP onBuildTargetDidChange notifications to clients when the build changes

alexarchambault:
This makes Mil send onBuildTargetDidChange BSP notifications to clients when the build change. If ever the build compilation fails at some point, this tries to keep the former evaluators of build levels that can't be loaded anymore, so that there's less interruption of IDE features for end users (that is, if `build.mill` is broken, this keeps using an old evaluator for the main build tasks, same for the first meta-build and the main build if `mill-build/build.mill` is broken, etc.).

In order for the preserved evaluators to keep working, but also to help with concurrency, this makes `MillBuildBootstrap` copy meta-builds' classes to a temporary directory right before loading them in a class loader, so that concurrent sessions don't remove class files that we need.

Lastly, when `--bsp` is passed, this makes sure we don't print messages for users in it early on, only BSP messages.


bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
```diff
@@ -45,21 +45,98 @@ private class MillBuildServer(
...
   private var bspEvaluators: Promise[BspEvaluators] = Promise[BspEvaluators]()
+  private var savedPreviousEvaluators = Option.empty[BspEvaluators]
   // Set when a session is completed, either due to reload or shutdown
-  private[worker] var sessionResult: Option[BspServerResult] = None
+  private[worker] var sessionResult = Promise[BspServerResult]()
 
   def initialized = sessionInfo != null
 
-  def updateEvaluator(evaluatorsOpt: Option[Seq[EvaluatorApi]]): Unit = {
+  def updateEvaluator(evaluatorsOpt: Option[Seq[EvaluatorApi]], errored: Boolean): Unit = {
```
lefou, line 59: Can we have ScalaDoc explaining what  error` means.

bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
```diff
@@ -45,21 +45,98 @@ private class MillBuildServer(
   // progresses through:
   //
   // Set when the client connects
-  protected var client: BuildClient = scala.compiletime.uninitialized
+  protected var client: BuildClient = null
```
lefou, line 48: Why is `uninitialized` not enough? Can we use a better type then, e.g. `Option`?

lefou (commented):
Do we signal/log (over BSP) in any way when we need to fallback to the previous evaluator? If not, I think we should, so the user has a chance to verify.

lihaoyi:
I haven't looked at the code in detail since it hasn't been cleaned up yet, but one thing we should do is try to re-use as much logic as possible from the normal `--watch` codepath. We already have everything necessary to watch sources (including build sources), rebootstrap the meta-builds, and perform an action on their change. We should make use of that logic for updating the BSP server, rather than coming up with our own custom logic here

alexarchambault:
Superseded by https://github.com/com-lihaoyi/mill/pull/5305

PR #4958, opened 2025-04-18
Fix `main` build and CI

lihaoyi:
Not sure why these dependencies were listed as `compileMvnDeps` rather than just `mvnDeps`, but that seems to be the cause of the `java.lang.NoSuchMethodError`s that were plaguing CI since we removed scalafmt/parallel-collections from the classpath. Hopefully this helps make things work again

Seems like the `compile` prefix was added in https://github.com/com-lihaoyi/mill/pull/4162

### DIFF
--- a/scalajslib/package.mill
+++ b/scalajslib/package.mill
@@ -48,7 +48,7 @@ object `package` extends RootModule with build.MillStableScalaModule with BuildI
     def moduleDir: os.Path = super.moduleDir / scalajsWorkerVersion
     def compileModuleDeps = Seq(build.scalajslib.`worker-api`, build.core.constants, build.core.api)
     def mandatoryMvnDeps = Agg.empty[Dep]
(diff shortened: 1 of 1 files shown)


lihaoyi:
CC @alexarchambault . How did using `compileMvnDeps` for `build.scalajslib.worker[1]` work at all? Wouldn't all those jars be missing from the runtime classpath, and we'd end up being unable to run anything at all?

lihaoyi:
CI is green on my fork, merging this

alexarchambault:
> CC @alexarchambault . How did using `compileMvnDeps` for `build.scalajslib.worker[1]` work at all? Wouldn't all those jars be missing from the runtime classpath, and we'd end up being unable to run anything at all?

At the time, the change was basically just a refactoring. All dependencies of the worker class path were filtered out at some point, so marking them as compile-time only made it more explicit.

alexarchambault:
That surprised me at first, but that was the state of things at the time

PR #4967, opened 2025-04-19
Consolidate Mill runtime libraries under `libs/` subfolder

lihaoyi:
This tidies up our Mill repo root, and groups together all these related libraries with the same properties: not core to Mill's evaluation logic, but used to support various languages and toolchains. 

We keep the JVM package path unchanged for backwards compatibility, which has some precedence e.g. the stuff in `core/` doesn't have `.core` in the package path.


lihaoyi:
Green on my fork, merging this

PR #4969, opened 2025-04-20
Add a YAML header comment syntax for configuration of `.mill-` files, `import $ivy`, `mill-version`

lihaoyi:
This PR adds a YAML 1.2 header comment syntax that subsumes many of the ad-hoc configuration mechanisms we have today: `.mill-version`, `.mill-opts`, `.mill-jvm-opts`, `.mill-jvm-version`, `import $ivy`, and possibly more meta-build overrides or other configuration options in future. 

```scala
//| mill-version: 0.13.0
//| mill-jvm-version: 17
//| repositories: [$PWD_URI/custom-repo]
//| mvnDeps: [org.thymeleaf:thymeleaf:3.1.1.RELEASE]
package build
import mill._, javalib._
import org.thymeleaf.TemplateEngine
import org.thymeleaf.context.Context

object foo extends JavaModule {
  def htmlSnippet = Task {
    val context = new Context
    context.setVariable("heading", "hello")
    context.setVariable("paragraph", "world")
    new TemplateEngine().process(
      """<div><h1 th:text="${heading}"></h1><p th:text="${paragraph}"></p></div>""",
      context
    )
  }

  def resources = Task {
    os.write(Task.dest / "snippet.txt", htmlSnippet())
    super.resources() ++ Seq(PathRef(Task.dest))
  }
}
```

The YAML header for any `.mill` file must be a line-comment block starting from the first line of the file and prefixed by `//|`. The prefix is chosen to avoid conflicts with normal comments (`//`), Scala-CLI directives (`//>`), and Java markdown comments (`///`), and to be reminiscent of Scala `|`s commonly used in `.stripMargin`

We use YAML 1.2 (via `snakeyaml-engine`) rather than ScalaCLI directives because:

- It can map more directly to Mill concepts: `mill-version`, `mill-jvm-version`, `mvnDeps`, and other config keys can be directly in YAML without any mapping. 

- YAML's hierarchical list/dict/primitive data model maps clearly to Mill's JSON data model for tasks, whereas Scala-CLI directives do not have such a clean correspondence and will require that we manually maintain an explicit mapping

- YAML will also be more familiar to non-Scala developers than Scala-CLI's directive syntax, including all the edge cases that are moderately well defined (quoting? escaping? comments? type coercions? multi-line strings?)

- Processing by third-party tools will be easier: anyone can write a Python script to strip the prefixes and parse out the YAML using pyyaml or yamlcore, or their equivalents in Node.js/Ruby/Rust/Go/whatever, whereas with Scala-CLI directives you are restricted to a single JVM implementation without spec or documentation

All of these concerns were raised in the original Scala-CLI directives discussion but were ignored ([link](https://contributors.scala-lang.org/t/pre-sip-using-directives/5700/5?u=lihaoyi), [link](https://contributors.scala-lang.org/t/pre-sip-scala-cli-as-new-scala-command/5628/92?u=lihaoyi)), so that brings us to today where we have to diverge from their syntax.

The choice of YAML over TOML was largely arbitrary. Python chose TOML for their script-header metadata in [PEP-723](https://peps.python.org/pep-0723/), whereas all Markdown implementations ([Github](https://docs.github.com/en/contributing/writing-for-github-docs/using-yaml-frontmatter), [RMarkdown](https://zsmith27.github.io/rmarkdown_crash-course/lesson-4-yaml-headers.html), [Jekyll](https://jekyllrb.com/docs/front-matter/)) chose YAML for their header-metadata syntax. YAML 1.2 seems to fix/mitigate most of the problems with YAML 1.1 (e.g. the `no == false` thing). Kotlin has their own syntax for file-level metadata [`@file:JvmName("Foo")`](https://kotlinlang.org/docs/annotations.html#annotation-use-site-targets), as does Swift [`@Metadata { @DocumentationExtension(mergeBehavior: override) }`](https://www.swift.org/documentation/docc/metadata)

Builds on top of https://github.com/com-lihaoyi/mill/pull/4624


lolgab:
To avoid having every line to start with `//|` which is not really YAML, you could use a syntax similar to the `RMarkdown` example you linked.
Instead of:

```scala
//|mill-version: x.y.z
//|other-setting:
//|  sub: value
```

we could have:


```scala
/*yaml
mill-version: x.y.z
other-setting:
  sub: value
*/
```

I added a `yaml` after the comment start `/*` since we want this to be a smart comment (as you did with `//|`).

The advantage is that the part in the middle is pure yaml.

Another alternative would be a second file, it would break the single build.mill file we always had. But having a second `mill.yml` would allow us to make it instantaneously work with YAML tooling, instead of requiring our tooling to support yet another standard.

I honestly prefer the separate file approach.

lihaoyi:
Separate file approach is definitely the easiest, but I do think there is a lot of value in having everything be crammed into a single file. Gradle, SBT, and other tools use separate `.properties` files and the indirection of having to jump back and forth definitely adds to the difficulty of using those tools

For the multi-line comment `/* */` style, the challenge there is we have to parse the `mill-version` string in Bash/Cmd even before we have a JVM process or Graal executable bootstrapped. The other is that it adds two lines of overhead, which in the common case of specifying a single `mill-version` is a lot. Definitely something we can consider though

lefou:
The YAML frontmatter almost works for me, as long as we keep the single-line approach initially proposed with this PR. These lines are roughly equivalent to Scala-CLI using directives but have a well spec'ed format. Why I said "almost"? There are multiple ways to represent the same thing, e.g. optional quoting, lists and awkward stuff resulting from it, like the Norway problem, which makes parsing without a dedicated YAML library harder. But restricting the features might irritate users as well, since everyone has preferences when it comes to YAML. We should have a dedicated section in our Documentation, where we provide best practices and some Mill-blessed preferences, so the user has a guide what a typical Mill buildfile header should look like. With that I can live with using YAML as frontmatter.

I hope I speak for us all, but at least I myself aim to follow the single source of truth philosophy with Mill, meaning we only have a single build file `build.mill`, and any additional build setup needs to be derived from its content. No assumptions.

Moving some external info back into the main build file (which this PR effectively does) is a good thing. We could talk about  support for loading external YAML files if we add some include concept, which means, it must be explicitly enabled in the `build.mill` header. But I'm not sure we should, as it makes parsing with primitive tools (bash/cmd) harder.

ci/mill-bootstrap.patch
```diff
@@ -1,3 +1,18 @@
...
+--- a/build.mill
++++ b/build.mill
+@@ -19,9 +19,8 @@ import mill.define.Cross
+ import scala.util.Properties
+ 
+ // plugins and dependencies
+-import $meta._
++//| meta: true
+ 
+-import $packages._
```
lefou, line 12: Did you remove the need for `packages` enablement? Also I find just `meta` a bit short and undescriptive. What about `meta-build: true` or `mill.features: [meta,packages]`.

runner/src/mill/runner/MillMain.scala
```diff
@@ -435,14 +438,37 @@ object MillMain {
     !(whitelistMatch || extraPlugins)
   }
 
+  def readVersionFile(file: os.Path): Option[String] = file match {
+    case f if os.exists(f) => os.read.lines(f).find(l => l.trim().nonEmpty)
+    case _ => None
+  }
+
+  private val usingMillVersionPattern = "//[|] +mill-version: +([^ ]+) *$".r
```
lefou, line 446: We need to take care of quoting and yaml comments here.

lihaoyi: I'm not sure we should; in the end the lowest common denominator is the bash/bat parser, and that one can't really handle much beyond simple regex space-splitting. 

If we try to best effort handle as much of YAML as possible, it will end up being some arbitrary subset, since there's no way we can handle the whole of YAML in a bash regex. It might be better to just throw in the towel and spec that the `mill-version` key-value pair must have this specific syntax, at least that way we can clearly define what syntax is allowed (i.e. not much)

core/util/src/mill/util/Jvm.scala
```diff
@@ -725,25 +725,11 @@ object Jvm {
...
-          .toSeq
-          .filter(_.nonEmpty)
-      else
-        spaceSep
-          .split(str)
-          .toSeq
-          .filter(_.nonEmpty)
-
+  private[mill] def reposFromStrings(repoList: Seq[String]): Result[Seq[Repository]] = {
     RepositoryParser.repositories(repoList).either match {
```
lihaoyi, line 729: @alexarchambault is the syntax supported by `RepositoryParser.repositories` documented anywhere? I'd like to link it in the documentation for the `//| repositories:` configuration key

core/exec/src/mill/exec/GroupExecution.scala
```diff
@@ -33,6 +33,40 @@ private trait GroupExecution {
...
+      "PWD_URI" -> workspace.toNIO.toUri.toString
+    )
+    def rec(x: Any): ujson.Value = {
+      import collection.JavaConverters._
+      x match {
+        case d: java.util.Date => ujson.Str(d.toString)
+        case s: String => ujson.Str(mill.constants.Util.interpolateEnvVars(s, envWithPwd.asJava))
+        case d: Double => ujson.Num(d)
+        case d: Int => ujson.Num(d)
+        case d: Long => ujson.Num(d)
```
He-Pin, line 52: does java.lang.Long also matched here?

ci/mill-bootstrap.patch
lefou: Maybe it just about more automated assistance that's needed. I could think of in case we hit an error, we check if we can find those files and ask the user if she meant to enable those features. Or we always check if the feature is not disabled, and warn the user that we are going to ignore those files.

runner/src/mill/runner/MillMain.scala
lihaoyi: I guess the question is if two-completely-separate-syntaxes is easier to understand, or if yaml-but-with-constraints is easier to understand. I'm guessing that 99% of people who write YAML will fall into writing `mill-version` "raw" and un-quoted, so although in theory it is a bit awkward in practice I think it'll be reasonably seamless for most people.

We could add support for simple quotes and comments as you mention, I think that actually won't be too hard even in Bash and Cmd. That should cover 99.9% of users since nobody is going to be using e.g. YAML multi-line strings for this even if in theory they could

ci/mill-bootstrap.patch
lihaoyi: Maven and Gradle have the explicit listing of subpackages similar to what you propose here, e.g. Maven's `<modules>` config or Gradle's `include` statements, so there is precedence. Though I must say I've hit the same confusion working with Maven/Gradle builds where I move stuff around the filesystem and forget to update the `<modules>` or `include` statements, or vice versa

lihaoyi:
Going to merge this as is, including the removal of the `import $packages` and `import $meta` requirement.

core/constants/src/mill/constants/Util.java
```diff
@@ -71,4 +74,48 @@ public static String hexArray(byte[] bytes) {
...
+    return yamlString;
+  }
+
+  private static String envInterpolatorPattern0 = "(\\$|[A-Z_][A-Z0-9_]*)";
+  private static Pattern envInterpolatorPattern =
+      Pattern.compile("\\$\\{" + envInterpolatorPattern0 + "\\}|\\$" + envInterpolatorPattern0);
+
+  /**
+   * Interpolate variables in the form of <code>${VARIABLE}</code> based on the given Map <code>env</code>.
+   * Missing vars will be replaced by the empty string.
```
lefou, line 94: Can't we fail instead?

core/exec/src/mill/exec/GroupExecution.scala
```diff
@@ -33,6 +33,40 @@ private trait GroupExecution {
...
   def exclusiveSystemStreams: SystemStreams
   def getEvaluator: () => EvaluatorApi
+  def headerData: String
+  lazy val parsedHeaderData: Map[String, ujson.Value] = {
+    import org.snakeyaml.engine.v2.api.{Load, LoadSettings}
+    val loaded = new Load(LoadSettings.builder().build()).loadFromString(headerData)
+    // recursively convert java data structure to ujson.Value
+    val envWithPwd = env ++ Seq(
+      "PWD" -> workspace.toString,
+      "PWD_URI" -> workspace.toNIO.toUri.toString
```
lefou, line 43: `PWD` is a bit unfortunate, as we mean the workspace dir. Should be `WORKSPACE` instead.

example/extending/imports/3-contrib-dep/build.mill
```diff
@@ -1,6 +1,7 @@
+//| mvnDeps: ["com.lihaoyi::mill-contrib-buildinfo:$MILL_VERSION"]
```
lefou, line 1: Isn't the empty version supported here?

runner/meta/src/mill/runner/meta/MillIvy.scala
```diff
@@ -13,12 +13,13 @@ object MillIvy {
...
-
-    replaced
+    millSigs.map(substituteMillVersion)
   }
+
+  def substituteMillVersion(str: String) = str
+    .replace("$MILL_VERSION", mill.util.BuildInfo.millVersion)
+    .replace("${MILL_VERSION}", mill.util.BuildInfo.millVersion)
+    .replace("$MILL_BIN_PLATFORM", mill.util.BuildInfo.millBinPlatform)
+    .replace("${MILL_BIN_PLATFORM}", mill.util.BuildInfo.millBinPlatform)
```
lefou, line 23: We should also make `MILL_BIN_PLATFORM` available in the YAML parser.

lefou (commented):
We need a documentation about all supported header data entries and all supported env variables. Also, we should fail if an undefined env variable is used instead of silently using an empty strings. This is very error prone, possibly dangerous and hard to debug.

mill
lefou, line 1: These changed need to happen in `scripts/src/mill.sh`

mill.bat
lefou, line 1: These changes need to happen in `scripts/src/mill.bat`.

lefou: * #4982

mill
lefou: * #4982

core/constants/src/mill/constants/Util.java
lefou: * https://github.com/com-lihaoyi/mill/pull/4983

runner/meta/src/mill/runner/meta/MillIvy.scala
lefou: * https://github.com/com-lihaoyi/mill/pull/4983

PR #4980, opened 2025-04-23
More careful stream handling when starting BSP server

alexarchambault:
When Mill is started with `--bsp`, this makes sure that only the BSP server ever writes something to stdout (and ever tries to read something from stdin). This fixes weird output issues I've been seeing while debugging the BSP tests in https://github.com/com-lihaoyi/mill/pull/4626, where the BSP server output and my debugging output weren't printed anywhere.


lefou:
Not a direct comment to your changes, but I remember to have fixed almost the same issues before and it was intricate work. We should apply more care when bulk-refactoring this code-base.

lefou (commented):
~Can you tell why the `--bsp` option must come first?~

Nevermind, it's still to avoid proper parsing in Java and/or shell code. I though we had replaced that.

runner/src/mill/runner/MillMain.scala
```diff
@@ -79,6 +79,40 @@ object MillMain {
 
   lazy val maybeScalaCompilerWorker = ScalaCompilerWorker.bootstrapWorker()
 
+  private def withStreams[T](
+      bspMode: Boolean,
+      streams0: SystemStreams
```
lefou, line 84: Can you give `streams0` a more descriptive name? Since it's not just `streams`, it must be something special.

runner/src/mill/runner/MillMain.scala
```diff
@@ -91,8 +125,8 @@ object MillMain {
       systemExit: Int => Nothing,
       serverDir: os.Path
   ): (Boolean, RunnerState) = {
-    val streams = streams0
-    mill.define.SystemStreams.withStreams(streams) {
+    val bspMode = args.headOption.contains("--bsp")
```
lihaoyi, line 128: I removed this check in the earlier BSP cleanup because it wasn't clear what the purpose of this check was, and it appears this code path was not being triggered anyway when I imported the project into IntelliJ.

What's the purpose of this check when we already do the stream redirection after argument parsing below?

alexarchambault: > What's the purpose of this check when we already do the stream redirection after argument parsing below?

If all goes well, that's even. But if anything goes wrong, in particular while debugging / developing on Mill, it's better to have the redirection in place early, to make it less likely to miss anything.

Plus it wasn't done right after argument parsing before this PR, but a bit after that, which increases the odd of messages going to the wrong stream.

runner/src/mill/runner/MillMain.scala
alexarchambault: (renamed it back to `streams`)

runner/src/mill/runner/MillMain.scala
```diff
@@ -94,8 +128,14 @@ object MillMain {
       serverDir: os.Path
   ): (Boolean, RunnerState) = {
 
-    val streams = streams0
-    mill.define.SystemStreams.withStreams(streams) {
+    val bspMode = {
```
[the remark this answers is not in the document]
lefou: Why do we avoid to parse the command line again? In the other parts I understand it (Java: avoid classloading and parsing time; shell: avoid complexity and extra tools) but here, we should just run our cli parser and use whatever it returns.

runner/src/mill/runner/MillMain.scala
```diff
@@ -305,75 +350,52 @@ object MillMain {
 
   def runBspSession(
       streams0: SystemStreams,
-      runMillBootstrap: (Option[RunnerState], SystemStreams) => RunnerState
+      logStreams: SystemStreams,
```
lefou, line 353: Can we please add some scaladoc stating the purpose of `streams0` and `logStreams` including their difference? This is also useful for devs.

lefou: Maybe rename `streams0` to `bspProtocolStreams`

lefou: I should include some hints when it is ok to send something to `bspStreams.err` (as it appears in the BSP client log output).

alexarchambault: Should be fixed

runner/src/mill/runner/MillMain.scala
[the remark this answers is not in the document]
alexarchambault: Just pushed something that I think should work (if the command-line argument parser doesn't print things to stdout on its own, which it seems it doesn't)

runner/src/mill/runner/MillMain.scala
```diff
@@ -136,6 +177,10 @@ object MillMain {
             )
             (false, RunnerState.empty)
 
+          case Result.Success(config) if config.bsp.value != bspMode =>
```
lefou, line 180:
```suggestion
          // The first-option requirement results from sloppy parsing 
          // in MillClientMain and our executable-jar-prepend script
          case Result.Success(config) if config.bsp.value != bspMode =>
```

lefou (commented):
Looks good to me! I suggested one additional comment below. Feel free to merge then.

runner/src/mill/runner/MillMain.scala
alexarchambault: That check can actually be removed, now that the actual CLI arg parser is used upfront

alexarchambault: Just removed it

lihaoyi (approved):

PR #4983, opened 2025-04-24
Fail on missing interpolatable variables

lefou:
Interpolating missing variable to the empty string (`""`) is error prone, possibly dangerous and hard to debug. Hence, this change makes it a hard failure when a variable is missing.


lefou:
We should use whatever is common in POSIX shells for that task. I think in `bash` it is `${VAR:-default_value}`.

lefou:
Since default value support is a new feature, I think it's best to do it in a separate PR.

lihaoyi:
Sounds good, feel free to merge this then

PR #4984, opened 2025-04-24
More robust readYamlHeader

sideeffffect:
Small improvement on top of
 * https://github.com/com-lihaoyi/mill/pull/4969

Hopefully will lead to less surprising behavior (and thus less user confusion) in weird edge cases.

### DIFF
--- a/core/constants/src/mill/constants/Util.java
+++ b/core/constants/src/mill/constants/Util.java
@@ -78,7 +78,7 @@ public class Util {
   public static String readYamlHeader(java.nio.file.Path buildFile) throws java.io.IOException {
     java.util.List<String> lines = java.nio.file.Files.readAllLines(buildFile);
     String yamlString = lines.stream()
-        .filter(line -> line.startsWith("//|"))
+        .takeWhile(line -> line.startsWith("//| "))
         .map(line -> line.substring(4)) // Remove the `//|` prefix
         .collect(java.util.stream.Collectors.joining("\n"));
 


lefou:
In our examples projects, we have the situation that `build.mill` files serve two purposes: buildfiles and documentation. The failing tests we see here are cases, where we want to have some introductory text before the actual buildfile. 

E.g. `example/extening/imports/2-mvn-deps-scala/build.mill` starts with the following lines:

```scala
// `//| mvnDeps:` can also be used with Scala libraries. The following example
// replaces `org.thymeleaf:thymeleaf:3.1.1.RELEASE` above with `com.lihaoyi::scalatags:0.12.0`,
// a HTML generation library that does the string concatenation/construction internally
// and also escapes the input strings on your behalf:

//| mvnDeps: ["com.lihaoyi::scalatags:0.13.1"]
import mill._, scalalib._
import scalatags.Text.all._

object bar extends ScalaModule {
  def scalaVersion = "2.13.8"

  def mvnDeps = Seq(mvn"com.lihaoyi::os-lib:0.10.7")
  def htmlSnippet = Task { div(h1("hello"), p("world")).toString }
  def resources = Task {
    os.write(Task.dest / "snippet.txt", htmlSnippet())
    super.resources() ++ Seq(PathRef(Task.dest))
  }
}
```

With your change, this leads to a build failure due to the missing `scalatags` library on the classpath. 

One solution could be to just repeat the header at the start of the file, but that raises the maintenance costs. Also, we need to check how that renders. We also could move the header to a marker line when rendering the documentation pages.

lefou:
My attempt to address the CI issues.
* https://github.com/com-lihaoyi/mill/pull/4986

lihaoyi (approved):
Looks good to me, may need to update tests to be compliant with the new restriction and ensure they don't have leading newlines or something, or maybe make example tests more lenient since it doesn't always make sense to have the `//|` frontmatter at the top of the example file before the explanatory comments.

lefou:
Superseded by #4986.

PR #4986, opened 2025-04-24
Parse YAML frontmatter only from file head

lefou:
Adapt `ExampleParser` to also capture the YAML frontmatter.
Defer rendering of YAML frontmatter until the first Scala code block is rendered.
Revise all contrib plugin examples to use YAML frontmatter to load the plugin.

This supersedes https://github.com/com-lihaoyi/mill/pull/4984


lihaoyi (approved):
ship it

PR #4892, opened 2025-04-08
Use R8 optimization tool in Android modules

irodotos7:
A step back from this PR https://github.com/com-lihaoyi/mill/pull/4743

I created a smaller, clear PR with the R8 task only and without the separation of Debug and Release Install paths

## Motivation
As part of Mill’s Android integration, we want to use the R8 optimization tool to create smaller dex files

image

image

## Provided in this PR
1) Run the R8 task which will shrink the dex files according to the proguard rules

PR #4996, opened 2025-04-25
Android/cleanup app release install

vaslabs:
## Summary

- **better organisation** between androidmodule and androidappmodule (separate properly what things are used in apps and things that are shared between lib building and app building)
- R8 and D8 **dex generation** depending on build type settings
- **Instrumented tests** now use the android application namespace and id and adding a "variant" indicator
- introduced a lite version of **build type settings** with main motivation to control r8 and d8 but also serve as a scaffolding for build variants
- **Android manifest** has less hardcoded parameters (uses-sdk and instrumentation tags are now generated)
- Minor change into **testTask** to keep it consistent with the rest of modules(Task.Anon) and thus the test-report is now in `testForked.dest`

## R8 and Proguard

Followed similar configuration style with gradle

image

Although there's still work to be done in that area, if minification is enabled, we use the R8 tool to generate the dex otherwise the d8 (which also uses proguard for now) . This needs further work and clarification as AGP seems to be using R8 by default when minification is enabled, otherwise d8 is used.

Proguard default files are also fetched (as an opt-in setting) from the Android SDK directories

## Build settings

I've introduced a first very minimal step for doing build variants, mainly out of necessity to manage R8 and D8 variations better, but this will be expanded once we need to fully support build variants.

The settings so far concerned the R8 use case at hand and it's a case class that has the parameters to feed proguard files and flags into R8.

## R8 enhancements

Extending the work from https://github.com/com-lihaoyi/mill/pull/4892 , I've made some  usability enhancements and enriched the examples to demonstrate how instrumented tests can be used alongside R8 and release builds.

I think R8 now can be fairly usable to generate minified and optimised release builds without getting in the middle of development (e.g. running tests and testing in the emulator) although more work needs to be done towards that area, but we'll follow up after we do more exploratory usability testing!

## Future work

Following this work, we can move to complete the work on architecture samples by testing instrumented tests, unit tests and later on debug variants.

The cleanup also reduced a lot of duplication between `AndroidAppModule` and `AndroidModule` so it will make our development efforts a bit easier to add more features!

More discovery work will follow to align the instrumented test apk installation more in line with AGP

### DIFF
--- a/example/androidlib/java/1-hello-world/app/src/main/AndroidManifest.xml
+++ b/example/androidlib/java/1-hello-world/app/src/main/AndroidManifest.xml
@@ -1,5 +1,4 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" android:versionCode="1" android:versionName="1.0">
-    <uses-sdk android:minSdkVersion="9" android:targetSdkVersion="35"/>
     <application android:label="@string/app_name" android:theme="@android:style/Theme.Light.NoTitleBar" android:debuggable="true">
         <activity android:name=".MainActivity"
                 android:exported="true">
@@ -9,7 +8,4 @@
             </intent-filter>
         </activity>
     </application>
-    <instrumentation
-            android:name="androidx.test.runner.AndroidJUnitRunner"
-            android:targetPackage="com.helloworld.app" />
 </manifest>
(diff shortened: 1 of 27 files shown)


vaslabs:
Pipeline failures don't seem related, I think they're the same as in main branch

lihaoyi:
Yeah looks not causes by this PR

PR #5014, opened 2025-04-28
Clean up `mill-dist` generation

lihaoyi:
- Rename `trait InstallModule` to `DistModule`
- Rename `testDep` to `localTestOverrides`
- No longer include `mill.runner` in `dist.localTestOverrides`, since those are bundled with the executable classpath and should never be resolved via coursier
- `DistModule` is now a bare `Module`, not a `MillPublishJavaModule`. `build.dist` is a non-publish module which defines `def publishArtifacts` itself rather than going through `PublishModule`
- `DistModule` defines `def publishArtifacts` directly rather than inheriting/overriding it from `MillPublishJavaModule`. The various publishing `ExternalModule`s all just look for a duck-typed `def publishArtifacts: T[PublishModule.PublishData]`, so we shouldn't actually need to inherit from `PublishModule` for publishing to work
- `def uploadToGithub` since we distribute all artifacts on maven central now
- Remove `dist0` since it's no longer necessary
- Consolidate handling of `def transitiveLocalTestOverrides` in `dist/package.mill`

Overall this narrows down the logic around `dist` publishing considerably: we no longer publish them as full `JavaModule`s with associated poms and sourcejars and javadoc and metadata, and instead publish only the executable assembly jar or native image for people to use. Which is what we want, since the `dist` files are all standalone and should never be part of a larger JVM dependency graph.


lihaoyi:
Mill's entire task query system is based on names though, the type query stuff was only introduced more recently and isn't nearly as widely used. All command line work goes through queries and almost all those queries are based on names

PR #5016, opened 2025-04-28
Respect XDG_CACHE_HOME in JLineNativeLoader

alexarchambault:
Comes from https://github.com/com-lihaoyi/mill/pull/4917, but shouldn't be a problem


alexarchambault:
Like https://github.com/com-lihaoyi/mill/pull/5017, merging given the CI is fully green

PR #5017, opened 2025-04-28
Try to fix transient issue on CI

alexarchambault:
Same workaround as a test nearby (see #4944)

Comes from https://github.com/com-lihaoyi/mill/pull/4917, but shouldn't be a problem


alexarchambault:
(Merging as the CI is - really - green this time)

PR #5020, opened 2025-04-28
Copy meta-build classes in an evaluator specific directory

alexarchambault:
This makes `MillBuildBootstrap` copy in a temporary directory the class path it's about to load for an evaluator, so that it doesn't get removed if some of its entries get recompiled. Only the class path entries that live under the Mill `out/` directory get copied, not the JARs living in the coursier cache for example.

This should make concurrency issues less likely, especially when a Mill BSP server is running alongside Mill watch commands for example, and both might be using the meta-build build products.

This had been discussed a bit around https://github.com/com-lihaoyi/mill/pull/3683#discussion_r1797104302.

### DIFF
--- a/runner/src/mill/runner/MillBuildBootstrap.scala
+++ b/runner/src/mill/runner/MillBuildBootstrap.scala
@@ -12,10 +12,12 @@ import mill.runner.worker.api.MillScalaParser
 import mill.runner.meta.{CliImports, FileImportGraph, MillBuildRootModule, ScalaCompilerWorker}
 import mill.util.BuildInfo
 
-import java.io.File
+import java.io.{File, IOException}
 import java.net.URLClassLoader
+import java.util.UUID
+import scala.annotation.tailrec
 import scala.jdk.CollectionConverters.ListHasAsScala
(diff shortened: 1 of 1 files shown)


lihaoyi:
I'm trying to think if there's another way to approach this, as having multiple sets of concurrent evaluators floating around in random UUID folders seems like something we would regret in future. 

We already take a global lock to try and block concurrent evaluation. Could we be more strict about invalidating `MillBuildServer#bspEvaluators` whenever we run `MillBuildBootstrap`? e.g. every time we run `MillBuildBootstrap`, we take a lock and set `MillBuildServer#bspEvaluators` to a `new Promise` so any future requests have to wait for the bootstrapping to complete?

In general we have a "assume things are single threaded" concurrency model, so I don't feel like we should random start allowing concurrency in poorly understood edge cases just to work around bugs in our locks. The default solution to single-threaded locking bugs should be to tighten to single-threaded locking restrictions unless we have a very good reason to allowing concurrency

alexarchambault:
Superseded by https://github.com/com-lihaoyi/mill/pull/5200

Issue #5002, opened 2025-04-25
Split out `mill-runner` from `mill-client` assembly


lihaoyi:
Now that `mill-client` has coursier, we can resolve and download the `mill-runner` jars on-demand rather than bundling them as part of the assembly. As many of these jars are shared with `mill-libs`, this would reduce the total size of the downloaded artifacts, and allow the separate jar downloads to be parallelize via coursier instead of curl

PR #5025, opened 2025-04-29
Split runner out of client assembly

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/5002

This decouples the Mill executable and the Mill server classpath. 

* Previously the Mill server took on the Mill executable as part of the classpath, with this change the Mill server classpath is resolved from Maven Central and the Mill client may not even be a jar anymore (e.g. if it is a graal native image)

* Reduces the native executable download size from ~108mb to ~49mb, with the difference being the runner assembly jar we previously prepended to the executable that now gets resolved from maven central.

* Removes the messy `MILL_CLASSPATH` environment variable handling that was necessary to reliably get the classpath of the client JVM across all different execution modes (local, packaged, native)

* Introduces a `cachedComputedValue` helper in our client Java code to make it easy to wrap arbitrary expensive computations, and used to to add some caching to the `yamlHeader` processing


lihaoyi:
Going to merge this despite the messiness and try to clean up in a follow up

PR #5026, opened 2025-04-30
Fix/actually limit memory for client

arturaz:
Properly implements fix done in https://github.com/com-lihaoyi/mill/pull/4163

Also adds JVisualVM support for native mill image.

Tested manually by launching both the JVM and native versions and looking at the memory usage during compilation via JVisualVM


lefou (commented):
Looks good to me.

PR #5028, opened 2025-04-30
Wrong paths inside diagnostics sent from bsp-server

### LINKED ISSUE #4988: mill-bsp wrong diagnostics path
original issue: https://github.com/scalameta/metals/issues/7347

**Reproduction:**
https://github.com/NPCRUS/reproductions/tree/mill-metals-compilation-lag
open repo in vs-code(might be reproducible in other editors as well)
import project
switch to mill-bsp
check that autocomplete and code inspection works
change the name of the Base trait
adjust name of the trait in package object where this trait is extended
now you should be able to see not found: type Base mill, which signals some wrong compilation state
**Expected behavior:**
mill.build should recompile properly on change

As @kasiaMarek found mill-bsp push diagnostic with such path, that might be an issue:
```
Received notification 'build/publishDiagnostics'
Params: {
  "textDocument": {
    "uri": "file:///Users/kmarek/Desktop/Tests/reproductions/out/mill-build/generateScriptSources.dest/build_/build.mill"
  },
  "buildTarget": {
    "uri": "file:///Users/kmarek/Desktop/Tests/reproductions/mill-build"
  },
  "originId": "METALS-$01b1f6a3-a9b5-47ae-b89d-b58f3b13e98a",
  "diagnostics": [],
  "reset": true
}
```

NPCRUS:
fixes #4988 

Hi folks, I have started to look into issue of wrong diagnostics reporting from mill by trying first to implement possible test case. I wrote an approximate test scenario with some missing bits I'm not sure how to implement yet, I would appreciate any advice on how to move forward, my current questions are:
1. Is this the right place to write such a test case?
2. What would be the right way to introduce new `mill.build` file into the test without touching existing one from resources
3. Is there any common way to change build.mill file in the middle of test execution?

### DIFF
--- a/integration/ide/bsp-server/src/BspServerTestUtil.scala
+++ b/integration/ide/bsp-server/src/BspServerTestUtil.scala
@@ -112,7 +112,8 @@ object BspServerTestUtil {
 
   def withBspServer[T](
       workspacePath: os.Path,
-      millTestSuiteEnv: Map[String, String]
+      millTestSuiteEnv: Map[String, String],
+      client: b.BuildClient = DummyBuildClient
   )(f: (MillBuildServer, b.InitializeBuildResult) => T): T = {
 
     val bspMetadataFile = workspacePath / Constants.bspDir / s"${Constants.serverName}.json"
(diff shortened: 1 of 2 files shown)


lihaoyi:
@NPCRUS You can make a new integration test case if you want something that's totally unrelated. You can use `modifyFile` to modify any file during test execution

lihaoyi:
Superseded by https://github.com/com-lihaoyi/mill/pull/5349

PR #5048, opened 2025-05-02
Bump OS-Lib

lihaoyi:
Split out from https://github.com/com-lihaoyi/mill/pull/5043 due to apparent breakage in the upgrade

### DIFF
--- a/integration/ide/gen-idea/resources/extended/idea/mill_modules/mill-build.iml
+++ b/integration/ide/gen-idea/resources/extended/idea/mill_modules/mill-build.iml
@@ -80,6 +80,7 @@
         <orderEntry type="library" name="munit_3-0.7.29.jar" level="project"/>
         <orderEntry type="library" name="native-terminal-no-ffm-0.0.9.1.jar" level="project"/>
         <orderEntry type="library" name="os-lib_3-0.11.5-M8.jar" level="project"/>
+        <orderEntry type="library" name="os-zip-0.11.5-M8.jar" level="project"/>
         <orderEntry type="library" name="plexus-archiver-4.10.0.jar" level="project"/>
         <orderEntry type="library" name="plexus-classworlds-2.6.0.jar" level="project"/>
         <orderEntry type="library" name="plexus-container-default-2.1.1.jar" level="project"/>
--- a/integration/ide/gen-idea/resources/extended/idea/mill_modules/mill-build.iml
+++ b/integration/ide/gen-idea/resources/extended/idea/mill_modules/mill-build.iml
@@ -79,7 +79,7 @@
         <orderEntry type="library" name="mill-moduledefs_3-0.11.4.jar" level="project"/>
         <orderEntry type="library" name="munit_3-0.7.29.jar" level="project"/>
         <orderEntry type="library" name="native-terminal-no-ffm-0.0.9.1.jar" level="project"/>
-        <orderEntry type="library" name="os-lib_3-0.11.5-M2.jar" level="project"/>
+        <orderEntry type="library" name="os-lib_3-0.11.5-M8.jar" level="project"/>
         <orderEntry type="library" name="plexus-archiver-4.10.0.jar" level="project"/>
         <orderEntry type="library" name="plexus-classworlds-2.6.0.jar" level="project"/>
         <orderEntry type="library" name="plexus-container-default-2.1.1.jar" level="project"/>
--- a/integration/ide/gen-idea/resources/extended/idea/mill_modules/mill-build.mill-build.iml
+++ b/integration/ide/gen-idea/resources/extended/idea/mill_modules/mill-build.mill-build.iml
@@ -81,7 +81,7 @@
         <orderEntry type="library" name="mill-runner-codesig_3.jar" level="project"/>
         <orderEntry type="library" name="mill-runner-meta_3.jar" level="project"/>
         <orderEntry type="library" name="native-terminal-no-ffm-0.0.9.1.jar" level="project"/>
-        <orderEntry type="library" name="os-lib_3-0.11.5-M2.jar" level="project"/>
+        <orderEntry type="library" name="os-lib_3-0.11.5-M8.jar" level="project"/>
         <orderEntry type="library" name="plexus-archiver-4.10.0.jar" level="project"/>
         <orderEntry type="library" name="plexus-classworlds-2.6.0.jar" level="project"/>
         <orderEntry type="library" name="plexus-container-default-2.1.1.jar" level="project"/>
--- a/integration/ide/gen-idea/resources/extended/idea/mill_modules/mill-build.mill-build.iml
+++ b/integration/ide/gen-idea/resources/extended/idea/mill_modules/mill-build.mill-build.iml
@@ -82,6 +82,7 @@
         <orderEntry type="library" name="mill-runner-meta_3.jar" level="project"/>
         <orderEntry type="library" name="native-terminal-no-ffm-0.0.9.1.jar" level="project"/>
         <orderEntry type="library" name="os-lib_3-0.11.5-M8.jar" level="project"/>
+        <orderEntry type="library" name="os-zip-0.11.5-M8.jar" level="project"/>
         <orderEntry type="library" name="plexus-archiver-4.10.0.jar" level="project"/>
         <orderEntry type="library" name="plexus-classworlds-2.6.0.jar" level="project"/>
         <orderEntry type="library" name="plexus-container-default-2.1.1.jar" level="project"/>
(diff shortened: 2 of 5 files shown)


lihaoyi:
@vaslabs seems like upgrading OS-Lib is hitting issues in the android test. Do you know what zip file it is trying to unzip? If we could get the zip standalone outside of the mill test suite, that would make it much easier to debug the os-lib issue upstream

vaslabs:
I'll check, give me some time
 In the past there was an issue with nested directories when unzipping after a version bump, might be the same

vaslabs:
@lihaoyi looks like the permissions are not set properly

image

vaslabs:
looks like the androidx junit 1.2.1 doesn't have any unix file attributes (`zipinfo -v`) which may contribute to that

```
  androidx/

  offset of local header from start of archive:   139
                                                  (000000000000008Bh) bytes
  file system or operating system of origin:      MS-DOS, OS/2 or NT FAT
  version of encoding software:                   1.0
  minimum file system compatibility required:     MS-DOS, OS/2 or NT FAT
  minimum software version required to extract:   1.0
  compression method:                             none (stored)
  file security status:                           not encrypted
  extended local header:                          no
  file last modified on (DOS date/time):          2010 Jan 1 00:00:00
  32-bit CRC value (hex):                         00000000
  compressed size:                                0 bytes
  uncompressed size:                              0 bytes
  length of filename:                             9 characters
  length of extra field:                          0 bytes
  length of file comment:                         0 characters
  disk number on which file begins:               disk 1
  apparent file type:                             binary
  non-MSDOS external file attributes:             000000 hex
  MS-DOS file attributes (00 hex):                none

  There is no file comment.

```

I'll check if it's originally like that or we mess it up somehow

lihaoyi:
Got it, it could be a weakness in the new zip implementation that @kiendang worked on, but probably fixable

vaslabs:
> Got it, it could be a weakness in the new zip implementation that @kiendang worked on, but probably fixable

Yeah it's a very peculiar use case, I checked the newest version of androidx junit, has the same issue. 

Let me know if you need any help to try things to fix it or how to reproduce. The problematic jar can be found [here](https://mvnrepository.com/artifact/androidx.test.ext/junit/1.2.1)

(classes.jar in the aar)

kiendang:
If someone points me to the code where the jar is unzipped, I can take a look.

lihaoyi:
@kiendang the problematic callsite is https://github.com/com-lihaoyi/mill/blob/0e0282a939acbe14c87a9523901cb356e35ac1c0/libs/androidlib/src/mill/androidlib/AndroidAppModule.scala#L227, the problematic zip should be the one @vaslabs linked. I think you should be able to reproduce the problem outside of Mill to fix it

kiendang:
The problem is `classes.jar` contains zip entries for directories with permission information. However the permission is missing `x`

Examples:

`tracing-1.1.0/classes.jar`
```
zipinfo /home/kien/mill/out/example/androidlib/java/1-hello-world/local/server/testForked.dest/sandbox/run-1/out/app/it/androidTransformAarFiles.dest/transform/tracing-1.1.0/classes.jar
Archive:  /home/kien/mill/out/example/androidlib/java/1-hello-world/local/server/testForked.dest/sandbox/run-1/out/app/it/androidTransformAarFiles.dest/transform/tracing-1.1.0/classes.jar
Zip file size: 3967 bytes, number of entries: 6
-rw-rw-rw-  0.0 unx        0 b- stor 81-Jan-01 01:01 androidx/
-rw-rw-rw-  0.0 unx        0 b- stor 81-Jan-01 01:01 androidx/tracing/
-rw-rw-rw-  0.0 unx     4693 b- defN 81-Jan-01 01:01 androidx/tracing/Trace.class
-rw-rw-rw-  0.0 unx      719 b- defN 81-Jan-01 01:01 androidx/tracing/TraceApi18Impl.class
-rw-rw-rw-  0.0 unx     1093 b- defN 81-Jan-01 01:01 androidx/tracing/TraceApi29Impl.class
-rw----     2.4 fat        6 b- stor 81-Jan-01 01:01 META-INF/androidx.tracing_tracing.version
```

`junit-1.2.1/classes.jar`
```
zipinfo /home/kien/mill/out/example/androidlib/java/1-hello-world/local/server/testForked.dest/sandbox/run-1/out/app/it/androidTransformAarFiles.dest/transform/junit-1.2.1/classes.jar
Archive:  /home/kien/mill/out/example/androidlib/java/1-hello-world/local/server/testForked.dest/sandbox/run-1/out/app/it/androidTransformAarFiles.dest/transform/junit-1.2.1/classes.jar
Zip file size: 10824 bytes, number of entries: 13
-rw----     2.0 fat       24 bX defN 10-Jan-01 00:00 META-INF/ext_junit_java_androidx_test_ext_junit-junit_kt.kotlin_module
-rw----     1.0 fat        0 b- stor 10-Jan-01 00:00 androidx/
-rw----     1.0 fat        0 b- stor 10-Jan-01 00:00 androidx/test/
-rw----     1.0 fat        0 b- stor 10-Jan-01 00:00 androidx/test/ext/
-rw----     1.0 fat        0 b- stor 10-Jan-01 00:00 androidx/test/ext/junit/
-rw----     1.0 fat        0 b- stor 10-Jan-01 00:00 androidx/test/ext/junit/rules/
-rw----     2.0 fat     1810 bl defN 10-Jan-01 00:00 androidx/test/ext/junit/rules/AppComponentFactoryRule.class
-rw----     2.0 fat      614 bl defN 10-Jan-01 00:00 androidx/test/ext/junit/rules/ActivityScenarioRule$Supplier.class
-rw----     2.0 fat     4330 bl defN 10-Jan-01 00:00 androidx/test/ext/junit/rules/ActivityScenarioRule.class
-rw----     2.0 fat     2789 bl defN 10-Jan-01 00:00 androidx/test/ext/junit/rules/DeleteFilesRule$1.class
-rw----     2.0 fat     3262 bl defN 10-Jan-01 00:00 androidx/test/ext/junit/rules/DeleteFilesRule.class
-rw----     1.0 fat        0 b- stor 10-Jan-01 00:00 androidx/test/ext/junit/runners/
-rw----     2.0 fat     6685 bl defN 10-Jan-01 00:00 androidx/test/ext/junit/runners/AndroidJUnit4.class
13 files, 19514 bytes uncompressed, 8630 bytes compressed:  55.8%
```

`os.unzip` does nothing wrong here. It sees `androidx/` with `-rw-rw-rw- ` and thus creates the `androidx` directory with `-rw-rw-rw- `.

I tried using the `unzip` command on linux and the same problem occurs.

Previously it works because `os.unzip` did not use the permission information when unzipping files. We can add a flag to `os.unzip` to ignore permissions.

vaslabs:
strange, for me the unzip command (Fedora 41 6.13.12-200.fc41.x86_64) gives proper permissions for directories

image

image

lihaoyi:
Thanks for the investigation @kiendang. So for the given `classes.jar`, we're seeing the crash inside OS-Lib's `os.unzip` code, so we don't really get a chance to manually fix up the permissions issue in Mill:

```[4611-0]       java.base/java.nio.file.Files.createDirectories(Files.java:793)
[4611-0]       os.makeDir$all$.apply(FileOps.scala:55)
[4611-0]       os.unzip$.apply$$anonfun$6(ZipOps.scala:363)
[4611-0]       scala.runtime.function.JProcedure1.apply(JProcedure1.java:15)
[4611-0]       scala.runtime.function.JProcedure1.apply(JProcedure1.java:10)
[4611-0]       scala.collection.IterableOnceOps.foreach(IterableOnce.scala:619)
[4611-0]       scala.collection.IterableOnceOps.foreach$(IterableOnce.scala:617)
[4611-0]       scala.collection.AbstractIterable.foreach(Iterable.scala:935)
[4611-0]       scala.collection.IterableOps$WithFilter.foreach(Iterable.scala:905)
[4611-0]       os.unzip$.apply(ZipOps.scala:393)
[4611-0]       mill.androidlib.AndroidAppModule.androidLibsClassesJarMetaInf$$anonfun$1$$anonfun$1$$anonfun$2(AndroidAppModule.scala:227)
[4611-0]       scala.collection.immutable.List.flatMap(List.scala:294)
[4611-0]       scala.collection.immutable.List.flatMap(List.scala:79)
[4611-0]       mill.androidlib.AndroidAppModule.androidLibsClassesJarMetaInf$$anonfun$1$$anonfun$1(AndroidAppModule.scala:225)
[4611-0]       mill.define.NamedTask.evaluate(Task.scala:248)
[4611-0]       mill.define.NamedTask.evaluate$(Task.scala:236)
[4611-0]       mill.define.TargetImpl.evaluate(Task.scala:382)
```

I guess the fix would be to make OS-Lib add the "executable" bit to all folders regardless of the zip file's permissions, so at least we are able to create the nested directory contents without crashing?

kiendang:
Yup maybe "`os.unzip` did nothing wrong" wasn't totally accurate. It created the outer directory with the specified permissions then crashed when attempted to unzipping the directory contents.

Adding `OWNER_EXECUTE` to directory perms sounds like a good default.

lihaoyi:
Got it. @kiendang could you send a PR?

lefou:
Shouldn't we first create the content, and then apply the permissions? If that means, the directory won't be accessible any longer, so be it.

lihaoyi:
Updated this PR with @kiendang's latest updates to OS-Lib, let's see how it goes

lihaoyi:
@kiendang your patch seems to fix things, thanks! CC @arturaz this should fix the failure on your PR

PR #5013, opened 2025-04-28
native-libs for android

irodotos7:
## MOTIVATION

As part of the Mill Android Integration, we want to be able to use and create Native Libs that are written in C/C++

## Provided in this PR

1) Ability to create the .so files for every different ABI supported by Android SDK

image

2) Use the new library in your android app

image

PR #5053, opened 2025-05-03
Third party example for android native: endless-tunnel

vaslabs:
Following the work on https://github.com/com-lihaoyi/mill/pull/5013 I've added a third party example for Android native. I included an `androidRun` command as a testing motivation . (which we can connect to run and subsequently to IDE BSP run in the future) 

The example increases confidence on building android native apps with mill and it is another milestone for Mill's android support .

https://github.com/user-attachments/assets/a442a185-4b45-4342-9d78-3c1fe726873b

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -182,6 +182,11 @@ jobs:
             install-sbt: false
 
           - java-version: 17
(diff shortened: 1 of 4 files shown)


vaslabs:
san-angeles ndk example works too, but the newest compiler version has fatal warnings and I had to delete 4 lines with unused variables

image


but it works without further changes and that example combines Java and c++

Issue #5058, opened 2025-05-04
Loading Mill into IntelliJ via BSP is broken in `main`


lihaoyi:
Seems like we generate a `.bsp/mill-bsp.json` file with the runner classpath in the `argv`, rather than the binary

Current workaround is to checkout a720c3782ea8d7f2a8430d5b39a9a23c7268c218 to import into IntelliJ before checking out the branch you are working on

lihaoyi:
Seems like an unfortunate side effect of https://github.com/com-lihaoyi/mill/pull/5025

PR #5061, opened 2025-05-05
Fix BSP install `argv` path

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/5058

It seems the `mill.main.cli` property and `MILL_MAIN_CLI` environment variable were not getting propagated properly, and the server-side `java.class.path` property we were falling back to is now invalid since https://github.com/com-lihaoyi/mill/pull/5025. 

The solution is to propagate a single `MILL_EXECUTABLE_PATH` variable computed from `class         .getProtectionDomain().getCodeSource().getLocation()`, which returns the assembly jar in `packaged` on or native binary in `native` mode. This doesn't work in `local` mode, but that's fine since the `.ide` tests don't pass in `local` mode anyway.

As we now handle `MILL_EXECUTABLE_PATH` in a narrow code path from the Mill launcher to the Mill server, we no longer need special handling in all the Mill bootstrap scripts to pass in this variable

Added a test case to `ide[bsp-modules]` to exercise this at least on the JVM; tested the native image case manually. Also tested the JVM launcher in intellij via `MILL_STABLE_VERSION=1 ./mill dist.installLocalCache` and verified the BSP import no longer immediately crashes


lefou (commented):
I thought, you changed the generated script files, but I overlooked some changes.

PR #5068, opened 2025-05-05
Draft: Implement watching file paths via `oslib.watch`

arturaz:
Currently watching files uses polling which is inefficient with large codebases. On my laptop `./mill --watch` uses 20% CPU just to watch the files for changes.

This uses `oslib`'s `watch` module to watch for the file changes instead of polling them.


lefou:
What are the CPU numbers on your laptop with this change applied?

arturaz:
> What are the CPU numbers on your laptop with this change applied?

I haven't tested it yet, as my repo uses mill 0.12.10. With https://github.com/com-lihaoyi/mill/pull/5073 I'll be able to test it and report it later.

runner/daemon/src/mill/daemon/MillCliConfig.scala
```diff
@@ -97,6 +97,11 @@ case class MillCliConfig(
       doc = """Watch and re-run the given tasks when when their inputs change."""
     )
     watch: Flag = Flag(),
+    @arg(
+      name = "watch-via-fs-notify",
```
lihaoyi, line 101: Let's shorten this to `--notify-watch` here and in the other PR, the current name is a bit of a mouthful

core/api/src/mill/api/Watchable.scala
```diff
@@ -8,6 +8,24 @@ package mill.api
...
 private[mill] sealed trait Watchable
 private[mill] object Watchable {
-  case class Path(p: java.nio.file.Path, quick: Boolean, signature: Int) extends Watchable
-  case class Value(f: () => Long, signature: Long, pretty: String) extends Watchable
+
+  /** A [[Watchable]] that is being watched via polling. */
+  private[mill] sealed trait Pollable extends Watchable
+
+  /** A [[Watchable]] that is being watched via a notification system (like inotify). */
+  private[mill] sealed trait Notifiable extends Watchable
```
lihaoyi, line 16: Do we actually need these two additional marker traits? It seems to me that working directly with `Watchable.Path` and `Watchable.Value` will be enough. There aren't any compatibility/extensibility concerns since all this stuff is `private[mill]` anyway

runner/daemon/src/mill/daemon/Watching.scala
```diff
@@ -109,13 +250,23 @@ object Watching {
...
-  def poll(w: Watchable) = w match {
+  /** @return true if the watchable did not change. */
+  inline def validate(w: Watchable.Pollable): Boolean = validateAnyWatchable(w)
+
+  /**
+   * As [[validate]] but accepts any [[Watchable]] for the cases when we do not want to use a notification system.
+   *
+   * Normally you should use [[validate]] so that types would guide your implementation.
+   */
+  def validateAnyWatchable(w: Watchable): Boolean = poll(w) == signature(w)
```
lihaoyi, line 261: This name is pretty weird; it wasn't great before but if we're going to change it we should use a better one. Maybe `def checkForUpdate` or something?

runner/daemon/src/mill/daemon/Watching.scala
```diff
@@ -17,72 +19,209 @@ import scala.annotation.tailrec
...
+          isUnderWorkspaceRoot
+        }
+
+        // If I have 'root/a/b/c'
+        //
+        // Then I want to watch:
+        //   root/a/b/c
+        //   root/a/b
+        //   root/a
+        //   root
```
lihaoyi, line 161: This comment is a bit confusing. We're only setting one `os.watch.watch` on the root, and this makes it sound like we're setting multiple. What we're actually doing is choosing the paths we need to watch recursively in Linux since inotify is non-recursive by default, since changes in any enclosing folder could result in the watched file or folder disappearing (e.g. if the enclosing folder was renamed) and we want to pick up such changes. Not sure how to phrase it in the comment but maybe you can come up with something

PR #5073, opened 2025-05-06
Draft: Implement watching file paths via `oslib.watch` (for `0.12.x` branch)

arturaz:
Backport of https://github.com/com-lihaoyi/mill/pull/5068

The fs watching is *disabled* by default and can be disabled via `--watch-via-fs-notify=true`.


runner/src/mill/runner/Watching.scala
```diff
@@ -15,72 +15,194 @@ import scala.annotation.tailrec
...
 
   def watchAndWait(
       streams: SystemStreams,
       setIdle: Boolean => Unit,
       stdin: InputStream,
       watched: Seq[Watchable],
-      colors: Colors
+      colors: Colors,
+      useNotify: Boolean,
+      serverDir: os.Path
```
lihaoyi, line 99: Let's just pass in `WatchArgs` directly rather than unpacking it, since we already have it in a convenient `case class`

runner/src/mill/runner/Watching.scala
```diff
@@ -15,72 +15,194 @@ import scala.annotation.tailrec
...
+    streams.err.println {
+      val viaFsNotify = if (useNotify) " (via fsnotify)" else ""
       colors.info(
-        s"Watching for changes to ${watchedPaths.size} paths$watchedValueStr... (Enter to re-run, Ctrl-C to exit)"
+        s"Watching for changes to ${watchedPathsSeq.size} paths$viaFsNotify$watchedValueStr... (Enter to re-run, Ctrl-C to exit)"
       ).toString
-    )
+    }
+
+    def doWatch(notifiablesChanged: () => Boolean) = {
```
lihaoyi, line 119: Let's rename this `doPollingWatch` or something to distinguish it from the notify watch

lihaoyi, line 119: Let's rename this `doPollingWatch` or something to distinguish it from the notify watch code path

runner/src/mill/runner/Watching.scala
```diff
@@ -15,72 +15,194 @@ import scala.annotation.tailrec
...
+      //   root/a/b/c
+      //   root/a/b
+      //   root/a
+      //   root
+      val filterPaths = pathsUnderWorkspaceRoot.flatMap { path =>
+        path.relativeTo(workspaceRoot).segments.inits.map(segments => workspaceRoot / segments)
+      }
+      writeToWatchLog(s"[watched-paths:filtered] ${filterPaths.toSeq.sorted.mkString("\n")}")
+
+      Using.resource(os.watch.watch(
```
lihaoyi, line 163: This whole block of code should probably be moved into a separate helper method, maybe `doNotifyWatch`, in contrast to `doPollingWatch`

Issue #5076, opened 2025-05-06
Mill scripts broke BSP support for older Mill versions


lefou:
The regression was introduced in https://github.com/com-lihaoyi/mill/pull/5061

The script no longer passes the `mill.main.cli` system property to the actual JVM process, resulting in a invalid `BSP/mill-bsp.json` connection file.

lefou:
Although this property is no longer in use, we still need it for older Mill versions. The Mill bootstrap scripts are supposed to be universal launchers for whatever Mill version the user selects, so they can't just drop required features of the latest stable Mill version. In fact, we need to support these features for a very long time.

PR #5077, opened 2025-05-06
Fix support for older Mill version in bootstrap scripts

lefou:
Fix https://github.com/com-lihaoyi/mill/issues/5076

Although this property is no longer in use,
we still need it for older Mill versions,
including the latest available stable Mill release `0.12.10`.

This reverts some changes from PR https://github.com/com-lihaoyi/mill/pull/5061

### DIFF
--- a/ci/mill.ps1
+++ b/ci/mill.ps1
@@ -134,4 +134,5 @@ if ($null -ne $remainingArgs) {
     }
 }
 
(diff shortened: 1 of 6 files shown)


lihaoyi (approved):
Maybe let's also add some comments so next time we see it "unused" we know not to delete it?

lefou:
> Maybe let's also add some comments so next time we see it "unused" we know not to delete it?

Already on it.

PR #5080, opened 2025-05-07
More Unit test classloader leak detection and cleanup

lihaoyi:
* Move `MillURLClassLoader` into its own file with related logic

* Use `Jvm.createClassLoader` in more places, replace `RunnerState.URLClassLoader`


libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -162,8 +160,11 @@ trait KotlinModule extends JavaModule { outer =>
     jars.toSeq
   }
 
+  def kotlinWorkerClassLoader = Task.Worker {
+    mill.util.Jvm.createClassLoader(kotlinCompilerClasspath().map(_.path), getClass.getClassLoader)
+  }
```
lefou, line 165: This is probably not what we want. It means, we create a classloader for each `KotlinModule` instead of sharing classes between all modules using the same Kotlin toolchain. The Classloader should be managed in the `KotlinWorkerManager`.

lihaoyi: Yes, we probably want to make it a `CachedFactory` like we have for `JvmWorkerImpl`

lefou: Was this peculiar piece of code leaking or could we just revert the part affecting `KotlinModule` for now?

lihaoyi: IIRC I noticed it because it was getting caught by some of the new unit test leak checks. It shouldn't be too hard to make it a `CachedFactory`, I can do it after my current PR is done

lefou: That would be great. Kotlin compiler is already the slowest compiler in our tool box, and dropping possible gains from shared hot classloaders would be a a real loss.

Issue #4522, opened 2025-02-10
Intermittent code cache full messages in main (500USD Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 500USD bounty on this issue, payable by bank transfer on a merged PR fixing this.

See https://github.com/orgs/com-lihaoyi/discussions/6 for other bounties and the terms and conditions that bounties operate under 

---

e.g.

```
[12107] [info] compiling 1 Scala source to /Users/lihaoyi/Github/mill/out/core/server/test/semanticDbData.dest/classes ..
OpenJDK 64-Bit Server VM warning: CodeHeap 'non-profiled nmethods' is full. Compiler has been disabled.
OpenJDK 64-Bit Server VM warning: Try increasing the code heap size using -XX:NonProfiledCodeHeapSize=
[99.636s][warning][codecache] CodeHeap 'non-profiled nmethods' is full. Compiler has been disabled.
[99.636s][warning][codecache] Try increasing the code heap size using -XX:NonProfiledCodeHeapSize=
CodeHeap 'non-profiled nmethods': size=120032Kb used=119871Kb max_used=119871Kb free=161Kb
 bounds [0x0000000112248000, 0x0000000119780000, 0x0000000119780000]
CodeHeap 'profiled nmethods': size=120016Kb used=99387Kb max_used=120016Kb free=20628Kb
 bounds [0x000000010a780000, 0x0000000111cb4000, 0x0000000111cb4000]
CodeHeap 'non-nmethods': size=5712Kb used=1932Kb max_used=2116Kb free=3779Kb
 bounds [0x0000000111cb4000, 0x0000000111f24000, 0x0000000112248000]
 total_blobs=69508 nmethods=68283 adapters=1132
 compilation: disabled (not enough contiguous free space left)
              stopped_count=1, restarted_count=0
 full_count=1
```

Seems to have started only after merging the Scala 3 update. Probably leaking a classloader somewhere.

The completion criteria is a consistent reproduction, and a PR which can be shown to fix the issue.

gamlerhart:
I had this one on 0.12.7 as well, after upgrading from 0.12.5. 

However, I had it once, and then couldn't reproduce it.

zelosleone:
I was able to reproduce this both locally and in github ci, I will try to see if i can fix it

lihaoyi:
[...]

PR #5082, opened 2025-05-08
More work to fix classloader leaks

lihaoyi:
WIP to address https://github.com/com-lihaoyi/mill/issues/4522

* Move `MillURLClassLoader` into `mill.api` so the `openClassloaders` count is global across the entire Mill JVM process
* Add `label: String`s to `MillURLClassLoader`s, defaulting to `sourcecode.Enclosing`, to provide a more human-readable experience than storing the long classpaths
* Remove `AllBootstrapEvaluators`: it is no longer used since we made `mill.idea.GenIdea` and `mill.bsp.BSP` hardcoded parts of `MillMain`
* Move some logic in `GroupExecution` and `MillBuildBootstrap` into their companion objects to try and limit the lambdas in the code from capturing the enclosing `this` object and causing leaks when they in turn are captured by the coursier thread pool picking up `DynamicVariable`s/`InheritableThreadLocal`s
* Name all threads
* Set `-Dscalac.filebasedcache.defer.close.ms=0` on the Mill server process to fix https://github.com/com-lihaoyi/mill/issues/5083, which was leaking threads that were in turn capturing and leaking their referenced classloaders

## Testing

* Added an integration smoketest `integration/feature/leak-hygiene` to try and detect classloader and thread leaks

* Manual Testing: with this PR running `rm -rf out` then `./mill-assembly.jar clean && ./mill-assembly.jar __.compile`, taking a JProfiler heap snapshot, and then repeating it a second time on the same `MillServerMain` process results in the same number of `java.net.URLClassLoader`s being reachable in both, whereas previously that number would rise by ~20 each iteration


lefou:
@lihaoyi 
> * Remove `AllBootstrapEvaluators`: it is no longer used since we made `mill.idea.GenIdea` and `mill.bsp.BSP` hardcoded parts of `MillMain`

Doesn't that mean, no Mill Module or plugin can provide comparable functionality anymore, e.g. an hypothetical Eclipse project generator.

lihaoyi:
@lefou yes that is right. Those APIs were meant for us to use ourselves, and I'm not aware of anyone who is relying on those APIs. But if someone needs it, or we explicitly want to support this functionality, I can put it back.

PR #5092, opened 2025-05-09
Rebootstrap on yaml header mill-jvm-version parsing fixes

lihaoyi:


lihaoyi:
Going to close this as wontfix since we actually use `.mill-jvm-version` as a `Task.Source` to set the default Java version in the Mill launcher JVM bootstrap code path, so it's not trivial to move it into the YAML header as `//| mill-jvm-version:` without more work to make the YAML header metadata available to the build

PR #5102, opened 2025-05-10
bugfix: All git cmds should be run with cwd=vcsBasePath

forever043:
When mill `out/` directory was moved out of srctree by MILL_OUTPUT_DIR env, all git cmds w/o `cwd` set to `vcsBasePath` will be failed.


forever043:
same as https://github.com/lefou/mill-vcs-version/pull/177

lihaoyi:
Thanks!

PR #4928, opened 2025-04-13
Mill sonatype central publishing doc

c0d33ngr:
This PR is to close #4754

PR #5107, opened 2025-05-12
Backport "Mill sonatype publishing doc"

lihaoyi:
Backport of https://github.com/com-lihaoyi/mill/pull/4928 to 0.12.x


lihaoyi:
The failed scalalib test passes locally on my laptop, going to merge this regardless

lefou:
I'm surprised about the removal of the sonatypecentral contrib plugin. We should avoid removing anything between 0.12.11 and 0.12.12, which gives a bad taste, esp. those who fear the larger breakages towards 1.0.0. Given how important this plugin becomes we hurt all users who already adopted it. I think we should keep the plugin functional, maybe by aliasing it.

lihaoyi:
Good point, let me put it back. It wasn't a conscious decision to remove, it just disappeared as part of the `git cherry-pick` and I didn't notice it

PR #5109, opened 2025-05-12
Rebootstrap on top of sonatype central published artifacts

lihaoyi:
File extension for the download URL needed to change, because the new sonatype central publishing pipeline doesn't accept shell-prefixed assemblies or native image binaries with the `.jar` prefix, so I instead suffixed it with `.exe`


mill
```diff
@@ -23,10 +23,10 @@
 # into a cache location (~/.cache/mill/download).
 #
 # Mill Project URL: https://github.com/com-lihaoyi/mill
-# Script Version: 0.13.0-M2-63-e8edbd
+# Script Version: debug-badge-21-6ec959-DIRTY9d9881ba
```
lefou, line 26: This version doesn't look right.

PR #5146, opened 2025-05-19
Use named tuples in Tasks

lefou:
Use named tuples where we use tuples in Task types. The following task were changed:

* `TestModule.test*` tasks now return `(msg: String, results: Seq[TestResult])`

* `AndroidAppKotlinModule.androidDiscoveredPreviews` now returns `(previewsDiscoveredJsonFile: PathRef, screenshotConfigs: Seq[ComposeRenderer.Screenshot])`
* `AndroidAppModule.android{D,R}8Dex` now returns `(outPath: PathRef, dexCliArgs: Seq[String])`
* `AndroidModule.androidCompiledResources` now returns named tuple and includes all relevant resources directly.

* Also use some named tuples in `GenIdeaImpl`, which hopefully improve reading/editing it.

### DIFF
--- a/libs/scalanativelib/src/mill/scalanativelib/ScalaNativeModule.scala
+++ b/libs/scalanativelib/src/mill/scalanativelib/ScalaNativeModule.scala
@@ -17,7 +17,7 @@ import mill.scalanativelib.worker.{
 import mill.T
 import mill.constants.EnvVars
 import mill.scalanativelib.worker.api.ScalaNativeWorkerApi
-import upickle.implicits.namedTuples.default.given 
+import upickle.implicits.namedTuples.default.given
 
 trait ScalaNativeModule extends ScalaModule with ScalaNativeModuleApi { outer =>
   def scalaNativeVersion: T[String]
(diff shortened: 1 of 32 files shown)


lihaoyi (commented):
shipit

libs/scalanativelib/src/mill/scalanativelib/ScalaNativeModule.scala
```diff
@@ -17,6 +17,7 @@ import mill.scalanativelib.worker.{
 import mill.T
 import mill.constants.EnvVars
 import mill.scalanativelib.worker.api.ScalaNativeWorkerApi
+import upickle.implicits.namedTuples.default.given
```
bishabosha, line 20: Is there possibly something that can be done to make this import visible by default?

lefou: It would be nice, yes. I'm open for suggestions. IIUC, if we `export` it from the location where it is needed (e.g. `TestModule`), we risk getting it into scope multiple times. We could just export it from `mill`, then at least `import mill._` will get it automatically?

lihaoyi: I think exporting it from JsonFormatters (which ends up inherited by mill._) is probably the best thing to do. That's where we put our other upickle implicits anyway so it'll fit right in

lefou: * https://github.com/com-lihaoyi/mill/pull/5258

PR #5148, opened 2025-05-19
Add latest unstable release to docsite

lihaoyi:
This isn't stored in any tag in the Git repo, so we need to fetch it from Maven Central's `maven-metadata.xml` in an input task

I also broke up the `oldDocSources` into multiple smaller tasks using a `Cross` module, so we can get some parallelism on what is typically a pretty slow code path

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -45,10 +45,14 @@ def millVersion: T[String] = Task {
   } else "SNAPSHOT"
 }
 
-def millLastTagTruth: T[String] = Task {
-  VcsVersion.calcVcsState(Task.log).lastTag.getOrElse(
-    sys.error("No (last) git tag found. Your git history seems incomplete!")
-  )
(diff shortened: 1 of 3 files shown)


build.mill
```diff
@@ -45,10 +45,14 @@ def millVersion: T[String] = Task {
...
-    sys.error("No (last) git tag found. Your git history seems incomplete!")
-  )
+def latestUnstableVersion = Task.Input {
+  val mavenMetadataXml = requests
+    .get("https://repo1.maven.org/maven2/com/lihaoyi/mill-dist/maven-metadata.xml")
+    .text()
+
+  val xml = scala.xml.XML.loadString(mavenMetadataXml)
+
+  (xml \\ "version").map(_.text).last
```
lefou, line 55: Would be nice to use coursier for that, so we can re-use the cache.

PR #5152, opened 2025-05-19
Fix possible race condition around file locks

alexarchambault:
`FileLock#lock` doesn't always block until it can acquire a lock. If the lock is held by the same JVM, it just throws an `OverlappingFileLockException`. To work around that, this replaces a call to `FileLock#lock` by a while loop that calls `FileLock#tryLock`, that stops when the lock is actually acquired.

Please open all PRs as drafts and ensure that your fork of Mill has 
`settings/actions` / `Allow all actions and reusable workflows` enabled to run CI on
your own fork of the Mill repo. Only once CI passes mark the PR as `Ready for review`
and CI will run on the main Mill repo before we merge it.

PR #5153, opened 2025-05-19
Fix race condition in DoubleLocked#release

alexarchambault:
This PR makes some uses of `DoubleLock` safer, and fixes a race condition with it.

`DoubleLock` is basically used as a wrapper around a file lock (second lock), using it alongside a memory lock (first lock), that acts as a mutex for it. That way, the same JVM process doesn't try to acquire the file lock twice at the same time in a given JVM, which `java.nio.channels.FileChannel` doesn't allow (it throws an `OverlappingFileLockException` in that case).

This PR makes sure `MillMain` and `MillDaemonMain` use the same `Lock.memory()` instance as a guard when trying to acquire a lock for the output directory. This doesn't solve the issue I was running into (see below), but seems to be safer nonetheless.

It also tries to dispose of locks in `DoubleLock` in case calls to `Lock#{lock,tryLock}` throw.

Lastly, it releases the two locks in `DoubleLocked` in the reverse order that they were acquired. This fixes the test case below. Doing so fixes a race condition, where
- a first thread acquires the lock
- a second one tries to acquire it, and waits for it
- the first thread releases the lock

During the third step, releasing the memory lock before the file lock makes the second thread acquire the memory lock, and then possibly try to acquire the file lock, before the first thread is done releasing the file lock. The JVM doesn't allow this, and the second thread gets an `OverlappingFileLockException`.

See the discussion in https://github.com/com-lihaoyi/mill/pull/5152 too

This fixes
```text
$ ./mill 'integration.feature[output-directory].packaged.daemon.testForked' "mill.integration.OutputDirectoryLockTests.basic"
```
for me, locally.

### DIFF
--- a/runner/client/src/mill/client/lock/DoubleLock.java
+++ b/runner/client/src/mill/client/lock/DoubleLock.java
@@ -12,22 +12,44 @@ public class DoubleLock extends Lock {
 
   @Override
   public Locked lock() throws Exception {
-    return new DoubleLocked(lock1.lock(), lock2.lock());
+    Locked l1 = null;
+    Locked l2 = null;
+    Locked result = null;
+    try {
+      l1 = lock1.lock();
+      l2 = lock2.lock();
+      result = new DoubleLocked(l1, l2);
+    } finally {
+      if (result == null) {
+        if (l2 != null) l2.release();
+        if (l1 != null) l1.release();
+      }
+    }
+
+    return result;
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


lihaoyi:
@alexarchambault can you fill out the PR description with your understanding of the problem and why this is the correct fix for it

alexarchambault:
@lihaoyi Should be done

lihaoyi (approved):
shipit

PR #5154, opened 2025-05-19
Prefix logging for BSP

alexarchambault:
When Mill runs as a BSP server, this makes it log in a similar way as when running Mill from the command-line, with `[…] ` prefixes.

For example, the small BSP session run in the test added in this PR logs the following to stderr:
```text
[bsp] Trying to load BSP server...
[bsp] BSP server started
[1-buildInitialize] Entered buildInitialize
[1-buildInitialize] buildInitialize took 2 msec
[2-workspaceBuildTargets] Entered workspaceBuildTargets
[bsp-init-build.mill-61] [info] compiling 3 Scala sources to /Users/alexandre/projects/mill/mill-class-loading/out/integration/ide/bsp-server/local/daemon/testForked.dest/sandbox/run-1/out/mill-build/compile.dest/classes ...
[bsp-init-build.mill-61] [warn] package scala contains object and package with same name: caps.
[bsp-init-build.mill-61] [warn] This indicates that there are several versions of the Scala standard library on the classpath.
[bsp-init-build.mill-61] [warn] The build should be reconfigured so that only one version of the standard library is on the classpath.
[bsp-init-build.mill-61] [warn] one warning found
[bsp-init-build.mill-61] [info] done compiling
[bsp-init] SNAPSHOT
[2-workspaceBuildTargets] Evaluating 5 tasks
[2-workspaceBuildTargets] Done
[2-workspaceBuildTargets] Evaluating 1 task
[2-workspaceBuildTargets] Done
[2-workspaceBuildTargets] workspaceBuildTargets took 4197 msec
[3-loggingTest] Entered loggingTest
[3-loggingTest] Evaluating 1 task
[3-loggingTest-1] bspLoggingTest from System.out
[3-loggingTest-1] bspLoggingTest from System.err
[3-loggingTest-1] bspLoggingTest from Console.out
[3-loggingTest-1] bspLoggingTest from Console.err
[3-loggingTest-1] bspLoggingTest from Task.log.info
[3-loggingTest] Done
[3-loggingTest] loggingTest took 11 msec
[4-buildShutdown] Entered buildShutdown
[5-onBuildExit] Entered onBuildExit
Reload finished, result: mill.api.internal.BspServerResult$Shutdown$@6726cc69
Shutting down executor
[bsp] BSP session returned with mill.api.internal.BspServerResult$Shutdown$@6726cc69
Stopping server via handle...
[bsp] Exiting BSP runner loop. Stopping BSP server. Last result: Success(mill.api.internal.BspServerResult$Shutdown$@6726cc69)
```

This helps make sense of what's going on when working on BSP-releated features, but could also help Mill BSP users in case anything goes wrong.


alexarchambault:
Includes https://github.com/com-lihaoyi/mill/pull/5153 for now

alexarchambault:
In the output above, I was expecting the `SNAPSHOT` to have a prefix to, looking at that…

alexarchambault:
About https://github.com/com-lihaoyi/mill/pull/5154/commits/11136022bd3e3e0a4e738f58339182694a98947b, without it, it seems `System.{out,err}` don't get captured in tasks run by the BSP server. In the example above, `bspLoggingTest from System.out` and `bspLoggingTest from System.err` get no prefix in that case. I wasn't able to reproduce that issue outside of it.

core/api/src/mill/api/Logger.scala
```diff
@@ -29,25 +29,37 @@ trait Logger {
    * Prints miscellaneous logging output which isn't part of the main output
    * a user is looking for, but useful to provide context on what Mill is doing
    */
-  def info(s: String): Unit
+  final def info(s: String): Unit =
+    info(logKey, s)
+
+  def info(logKey: Seq[String], s: String): Unit
```
lihaoyi, line 36: Do we need these changes to the `Logger` interface?

`Logger` is shared throughout Mill as part of `core.api`, and thus it really needs to be as simple as possible. Rather than adding a more verbose version of `info`/`debug`/etc., could we instead instantiate sub-classes of `Logger` inside `MillBuildServer` that have the correct `logKey` set?

runner/bsp/worker/src/mill/bsp/worker/MillJvmBuildServer.scala
```diff
@@ -41,9 +41,12 @@ private trait MillJvmBuildServer extends JvmBuildServer { this: MillBuildServer
       targetIds: collection.Seq[BuildTargetIdentifier],
       agg: java.util.List[JvmEnvironmentItem] => V
   )(implicit name: sourcecode.Name): CompletableFuture[V] = {
+    val logger = createLogger()
     handlerTasks(
       targetIds = _ => targetIds,
-      tasks = { case m: RunModuleApi => m.bspJvmRunTestEnvironment }
+      tasks = { case m: RunModuleApi => m.bspJvmRunTestEnvironment },
+      requestDescription = "jvmRunTestEnv",
```
lihaoyi, line 48: Do we need custom names for these? We already take the `sourcecode.Name` of each endpoint method, which I would have thought would give us enough information about what endpoint each log message is comming from

runner/client/src/mill/client/lock/DoubleLock.java
```diff
@@ -12,22 +12,44 @@ public DoubleLock(Lock lock1, Lock lock2) throws Exception {
 
   @Override
   public Locked lock() throws Exception {
-    return new DoubleLocked(lock1.lock(), lock2.lock());
+    Locked l1 = null;
```
lihaoyi, line 15: Let's remove this from this PR since it's unrelated

runner/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
```diff
@@ -54,24 +57,28 @@ private class MillBuildServer(
...
-  def print(msg: String): Unit = logStream.println(msg)
-  def debug(msg: => String): Unit = if (debugMessages) logStream.println("[debug] " + msg)
-
   def onConnectWithClient(buildClient: BuildClient): Unit = client = buildClient
 
   override def buildInitialize(request: InitializeBuildParams)
-      : CompletableFuture[InitializeBuildResult] =
-    handlerRaw(checkInitialized = false) {
+      : CompletableFuture[InitializeBuildResult] = {
+    val logger = createLogger()
```
lihaoyi, line 80: Do we really need to call `createLogger()` and pass it in to every `handlerRaw`? Could we instead make `handlerRaw` take the `sourcecode.Enclosing` and create the logger internally?

core/api/src/mill/api/Logger.scala
lihaoyi: Could we even use the `PrefixLogger` class for this? Adding a prefix to log lines is exactly what that class is meant to do

runner/client/src/mill/client/lock/DoubleLock.java
alexarchambault: Should be done

core/api/src/mill/api/Logger.scala
alexarchambault: Will try that now, but I was under the impression that `PrefixLogger` doesn't allow what I wanted here

runner/bsp/worker/src/mill/bsp/worker/MillJvmBuildServer.scala
alexarchambault: They're used to build messages that end up as dialog messages when using VSCode (lower right corner). `requestDescription` is meant to be used to build more user-friendly messages. I tweaked it a bit in the latest push.

runner/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
alexarchambault: Should be fixed

alexarchambault:
Closing the PR here, superseded by https://github.com/com-lihaoyi/mill/pull/5180

PR #5155, opened 2025-05-19
Use layered class loaders rather than "shared" class loader hack for class loader isolation

alexarchambault:
This PR makes Mill start its daemon with the help of [coursier bootstraps](https://almond.sh/docs/install-advanced.html#isolation-of-almond-internal-dependencies), to isolate some Mill dependencies. These bootstraps are generated so that they load Mill using three class loaders:
- a first one with `org.scala-sbt:compiler-interface` and `org.scala-sbt:test-interface`
- a second one, with the first as parent, that loads the `core-api` module of Mill (alongside its dependencies)
- a third one, loading the rest of the Mill class path

The first class loader is the parent of the second one, and the second one is the parent of the third one.

When loading scala-compiler JARs (in `JvmWorkerImpl#scalaCompilerCache`) or test frameworks (like in `TestModule#bspBuildTargetScalaTestClasses`, to discover tests from within the Mill process, but it's also done in other places), Mill uses the first class loader as a parent loader for the class loader that loads scala-compiler or the class path that has tests. That way, the compiler-interface and test-interface classes are shared between Mill's internals and scalac or the test class path.

When loading the build class path in `MillBuildBootstrap#processRunClasspath`, Mill uses the second class loader, so that the core-api class path is shared between Mill's internals and the user build, but the rest of the Mill class path is hidden from user builds.

This has two benefits, compared to what's done without this PR:
- avoiding mistakes or issues that can arise with the use of package prefixes to hide classes (like done [here](https://github.com/com-lihaoyi/mill/blob/aa7125f81d4952717fa42e863e0311b2b39d1631/runner/daemon/src/mill/daemon/MillBuildBootstrap.scala#L284-L289))
- make the class loader hierarchy more "standard", consisting solely of `URLClassLoader`s (and the JVM's app and bootstrap loaders), so that it's easier to inspect by users if they need to

Initially, these development were motivated by suspicious things seen around `SystemStreams.ThreadLocalStreams.current` involving thread local variables, in https://github.com/com-lihaoyi/mill/pull/5154, but it turns out the changes in the PR here are not necessary to address these issues.

So this PR only makes things safer or more standard, but it isn't required for any feature as I initially thought. If you've seen anything suspicious with the handling of class loaders, it might be helpful…

### DIFF
--- a/contrib/scoverage/src/mill/contrib/scoverage/ScoverageReportWorker.scala
+++ b/contrib/scoverage/src/mill/contrib/scoverage/ScoverageReportWorker.scala
@@ -35,10 +35,7 @@ class ScoverageReportWorker {
       )(implicit
           ctx: TaskCtx
       ): Unit = {
-        mill.util.Jvm.withClassLoader(
-          classpath.map(_.path).toVector,
-          getClass.getClassLoader
-        ) { cl =>
+        mill.util.Jvm.withClassLoader(classpath.map(_.path).toVector) { cl =>
 
           val worker =
             cl
--- a/core/util/src/mill/util/Jvm.scala
+++ b/core/util/src/mill/util/Jvm.scala
@@ -219,7 +219,7 @@ object Jvm {
    */
   def createClassLoader(
       classPath: Iterable[os.Path],
-      parent: ClassLoader = null,
+      parent: ClassLoader = Thread.currentThread().getContextClassLoader,
       sharedLoader: ClassLoader = getClass.getClassLoader,
       sharedPrefixes: Iterable[String] = Seq(),
       label: String = null
@@ -228,6 +228,25 @@ object Jvm {
     parent,
     sharedLoader,
     sharedPrefixes,
+    Option(label).getOrElse(e.value)
+  )
+
+  /**
+   * Creates a `java.net.URLClassLoader` with specified class path
+   *
+   * Only basic JVM classes are shared with this loader.
+   *
+   * @param classPath URLs from which to load classes and resources
+   * @return new classloader
+   */
+  def createIsolatedClassLoader(
+      classPath: Iterable[os.Path],
+      label: String = null
+  )(implicit e: sourcecode.Enclosing): MillURLClassLoader = new MillURLClassLoader(
+    classPath.map(_.toNIO),
+    parent = null,
+    sharedLoader = null,
+    sharedPrefixes = Nil,
     Option(label).getOrElse(e.value)
   )
 
@@ -241,7 +260,7 @@ object Jvm {
    */
   def withClassLoader[T](
       classPath: Iterable[os.Path],
-      parent: ClassLoader = null,
+      parent: ClassLoader = Thread.currentThread().getContextClassLoader,
       sharedLoader: ClassLoader = getClass.getClassLoader,
       sharedPrefixes: Seq[String] = Seq.empty
   )(f: ClassLoader => T): T = {
@@ -252,6 +271,26 @@ object Jvm {
       sharedLoader = sharedLoader,
       sharedPrefixes = sharedPrefixes
     )
+    Thread.currentThread().setContextClassLoader(newClassloader)
+    try {
+      f(newClassloader)
+    } finally {
+      Thread.currentThread().setContextClassLoader(oldClassloader)
+      newClassloader.close()
+    }
+  }
+
+  /**
+   * @param classPath URLs from which to load classes and resources
+   * @param f function that will be called with newly created classloader
+   * @tparam T the return type of this function
+   * @return return value of the function `f`
+   */
+  def withIsolatedClassLoader[T](
+      classPath: Iterable[os.Path]
+  )(f: ClassLoader => T): T = {
+    val oldClassloader = Thread.currentThread().getContextClassLoader
+    val newClassloader = createIsolatedClassLoader(classPath)
     Thread.currentThread().setContextClassLoader(newClassloader)
     try {
       f(newClassloader)
@@ -656,7 +695,7 @@ object Jvm {
     val testOverridesClassloaders = System.getenv("MILL_LOCAL_TEST_OVERRIDE_CLASSPATH") match {
       case null => Nil
       case cp =>
-        cp.split(';').map { s => createClassLoader(Array(os.Path(s))) }.toList
+        cp.split(';').map { s => createIsolatedClassLoader(Array(os.Path(s))) }.toList
     }
 
     try {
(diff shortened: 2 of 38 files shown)


core/util/src/mill/util/Jvm.scala
```diff
@@ -219,7 +219,7 @@ object Jvm {
    */
   def createClassLoader(
       classPath: Iterable[os.Path],
-      parent: ClassLoader = null,
+      parent: ClassLoader = Thread.currentThread().getContextClassLoader,
```
lihaoyi, line 222: I think better to keep this default as is; flipping the default will likely cause a lot of confusion since the code will all compile and run but possibly fail with subtle classloading issues

runner/client/src/mill/client/ServerLauncher.java
```diff
@@ -108,8 +113,33 @@ Socket launchConnectToServer(Path daemonDir) throws Exception {
...
         mill.client.lock.Locked locked = locks.launcherLock.lock()) {
 
-      if (locks.daemonLock.probe()) initServer(daemonDir, locks);
-      while (locks.daemonLock.probe()) Thread.sleep(1);
+      Process serverProcess = null;
+
+      if (locks.daemonLock.probe()) serverProcess = initServer(daemonDir, locks);
+      while (locks.daemonLock.probe()) {
+        if (serverProcess != null && !serverProcess.isAlive()) {
+          System.err.println("Mill server exited!");
```
lihaoyi, line 121: I assume this is just debugging logging that should be cleaned up?

runner/daemon/src/mill/daemon/MillBuildBootstrap.scala
```diff
@@ -281,12 +281,25 @@ class MillBuildBootstrap(
           // Make sure we close the old classloader every time we create a new
           // one, to avoid memory leaks
           prevFrameOpt.foreach(_.classLoaderOpt.foreach(_.close()))
-          val cl = mill.util.Jvm.createClassLoader(
-            runClasspath.map(p => os.Path(p.javaPath)),
-            null,
-            sharedLoader = classOf[MillBuildBootstrap].getClassLoader,
-            sharedPrefixes = Seq("java.", "javax.", "scala.", "mill.api")
-          )
+          val hasLayeredClassLoader =
```
lihaoyi, line 284: These conditionals are new, and seem to against the goal of making the classloader hierarchy easier to understand. Why do we need them now, if we did not need them before?

lihaoyi:
Is it possible to set up the layered classloaders without using coursier's bootstrap infrastucture? It seems to be straightforward to do it "manually", as we do a similar thing in `TestRunnerMain.java`, and if it's not difficult it would be nice to have the logic encapsulated here rather than calling out to a third-party library

Overall the high-level goal of this PR seems reasonable, but the details of the PR seem to contradict the notion that this makes the classloader setup cleaner or easier to understand. At first glance, this PR seems to introduce a lot more complexity (including usage of arbitrarily-complex third-party library logic in coursier-bootstrap) as well as more confusion in the Mill codebase (e.g. the `if` conditionals scattered throughout the codebase). Not sure if that is fundamental or incidental, but if it's incidental perhaps after some cleanup we could demonstrate an improvement in clarity over the status quo

contrib/scoverage/src/mill/contrib/scoverage/ScoverageReportWorker.scala
```diff
@@ -35,10 +35,7 @@ class ScoverageReportWorker {
       )(implicit
           ctx: TaskCtx
       ): Unit = {
-        mill.util.Jvm.withClassLoader(
-          classpath.map(_.path).toVector,
-          getClass.getClassLoader
-        ) { cl =>
+        mill.util.Jvm.withClassLoader(classpath.map(_.path).toVector) { cl =>
```
lefou, line 38: This looks suspicious. We should not refactor an explicitly known classloader into some implicitly known. This reduces clarity. Although there is a concept of a context class loader, we should not rely on it if we can manage it directly.

core/util/src/mill/util/Jvm.scala
lefou: I'm also in favor of avoiding the context classloader if we can specify one explicitly.
Since this PR also adds `createIsolatedClassLoader`, we should just remove the default argument here.

core/util/src/mill/util/Jvm.scala
```diff
@@ -231,6 +231,25 @@ object Jvm {
...
+   * Creates a `java.net.URLClassLoader` with specified class path
+   *
+   * Only basic JVM classes are shared with this loader.
+   *
+   * @param classPath URLs from which to load classes and resources
+   * @return new classloader
+   */
+  def createIsolatedClassLoader(
+      classPath: Iterable[os.Path],
+      label: String = null
```
lefou, line 244: Please add some docs, since `null` is always required to be explained.

core/util/src/mill/util/Jvm.scala
```diff
@@ -241,7 +260,7 @@ object Jvm {
    */
   def withClassLoader[T](
       classPath: Iterable[os.Path],
-      parent: ClassLoader = null,
+      parent: ClassLoader = Thread.currentThread().getContextClassLoader,
```
lefou, line 263: We should remove the default. Using the context classloader should be a conscious decision at the use site.

libs/kotlinlib/src/mill/kotlinlib/KotlinWorkerManager.scala
```diff
@@ -15,7 +15,7 @@ class KotlinWorkerFactory()(implicit ctx: TaskCtx)
     extends CachedFactory[Seq[os.Path], (URLClassLoader, KotlinWorker)] {
 
   def setup(key: Seq[os.Path]) = {
-    val cl = mill.util.Jvm.createClassLoader(key, getClass.getClassLoader)
+    val cl = mill.util.Jvm.createClassLoader(key)
```
lefou, line 18: Keep it explicitly.

lefou (commented):
I stopped review after I realized, that the direction of the refactorings is to prefer the context classloader. I don't like it. It seem there is no real need, but using the context classloader is a can of worms. We are in the fortunate current situation that we exactly know which classloader we use in all code locations. We should not give up this now. It's a valuable quality. Even if you provide a stronger motivation to not use`getClass.getClassloader`, I'd argue that we should use some dedicated API to provide the "correct" classloader instead of just using the context classloader, which is some uncontrollable wishy-washy resource.

lihaoyi:
Closing this for now, feel free to open a new one if you'd like to propose this again

PR #5162, opened 2025-05-20
Split `MillMain` into `MillNoDaemonMain` and `MillMain0`

lihaoyi:


runner/daemon/src/mill/daemon/MillMain0.scala
```diff
@@ -21,7 +21,7 @@ import scala.util.control.NonFatal
 import scala.util.{Properties, Using}
 
 @internal
-object MillMain {
+object MillMain0 {
```
lefou, line 24: Why the rename to `MillMain0`?

lefou (commented):
Can we have some docs on the classes? Having multiple mains raises the question which class to use, which should be clarified in the docs.

PR #5163, opened 2025-05-20
Rename `publishM2Local` to `publishLocal`, remove old `~/.ivy2` `publishLocal` implementation

lihaoyi:
Follow up to recent work to get rid of legacy `ivy` terminology (e.g. https://github.com/com-lihaoyi/mill/pull/4930), we should publish to the local `~/.m2` maven repo rather than `~/.ivy2` ivy repo by default.


lefou (commented):
IMHO, we should keep support for publishing to the local ivy repo for a while, maybe as `publishIvyLocal`. I think sbt is still using it. 

Also, why not keep the name `publishM2Local`. Since there are multiple local repository standards, being more explicit leaves less room for confusion.

(Ideally, `publishLocal` would just be an alias to the concrete publish method and defaults to `publishM2Local`.)

alexarchambault:
About using `~/.m2/repository`, there have been issues when using this repository in coursier see [here](https://github.com/coursier/coursier/blob/3eeec1a00f2faea4ba94b42da62d08491abf2d70/doc/docs/other-repositories.md?plain=1#L32).

But a Mill-specific local Maven repository could work, like `~/.mill/maven-local`. I think I could add it as a default in coursier (alongside `~/.ivy2/local`), and we could give instructions to Maven / Gradle users about how to use it in these.

lihaoyi:
Probably going to close this for now since we're not fully confident using `~/.m2` with Coursier

PR #5166, opened 2025-05-21
Remove `T.*` APIs

lihaoyi:
These were replaced by `Task.*` in 0.12.x, but kept around for compatibility. We should remove them in 1.0.0's cleanup


He-Pin:
But the type us still there, would it be Task too？

lihaoyi:
Keeping the `T[_]` type for now, since that's pretty widely used

He-Pin:
Will you mark it as deprecated then?

lihaoyi:
Not sure TBH, but for now no plans to remove it so can keep it for now

PR #5169, opened 2025-05-21
Android/hilt unit tests

vaslabs:
## Summary

Prove that we can run the architecture-sample unit tests

The major feature missing here is to allow for kotlin modules to mark any internal modules they depend on as friend modules (there was already an isolated case for KotlinTests and the main source code) . Following how the shared-tests of architecture samples depends on app main module and in turn tests depend on shared-tests and main sources (as they do) , I've covered the missing gaps to get everything to work properly.

## Features added

- Kotlin `-X-friend-paths` is now used depending on transitive module deps, for AndroidKotlinModules (this can later be expanded to KotlinModule too)
- An androidNamespace is required for internal libs to function
- Android manifest for internal libs/modules needs to add metadata to the manifest too
- AndroidHiltSupport should also be available to all android modules, not just apps

## Fixes

- The AndroidHiltTransformAsm was dropping the META-INF, but it is required for `-X-friend-paths` to function , as the kotlin module contains the symbols necessary for internal members to be available to the depending modules
- Ironed out the usage of androidApplicationNamespace vs androidApplicationId as their usage becomes clearer to me the more apps we integrate - this is under development and I expect their usage to be stabilised once build variants are added too

## Unlocked workflows

This PR proves the following to be working:
- Complex unit test setup with Hilt
- A shared android module (shared-tests) can be given as a dependency (it's a dependency of `app.test` and `app.androidTest` which will be wired in in a follow up PR)
- Android Kotlin internal and top level members can be used between depended modules (app.test can use app top level members as app compiled path is passed as a friend path)

## Previous work reference

This is a continuous development for supporting android worfklows using Hilt and by extension Kotlin KSP features. 

Previous related work items:
- https://github.com/com-lihaoyi/mill/pull/4759
- https://github.com/com-lihaoyi/mill/pull/4485
- https://github.com/com-lihaoyi/mill/pull/4557

## Kotlin build issue

Need to check why arrow is not building, but probably I'll move the friend paths functionality to AndroidKotlinModule if I don't manage to identify the issue for the time being

Due to the use of outer in KotlinTest
```
    override def kotlincOptions: T[Seq[String]] = Task {
      outer.kotlincOptions().filterNot(_.startsWith("-Xcommon-sources")) ++
        Seq(s"-Xfriend-paths=${outer.compile().classes.path.toString()}")
    }
```

The `-Xfriend-paths` are picked from the parent and not computed properly (so the module that a kotlin test directly depends on is missing. e.g.

```
-Xfriend-paths=/home/vnicolaou/mill/out/example/thirdparty/arrow/downloadedRepo.dest/arrow-bc9bf92cc98e01c21bdd2bf8640cf7db0f97204a/out/arrow-libs/core/arrow-annotations/jvm/compile.dest/classes,/home/vnicolaou/mill/out/example/thirdparty/arrow/downloadedRepo.dest/arrow-bc9bf92cc98e01c21bdd2bf8640cf7db0f97204a/out/arrow-libs/core/arrow-atomic/jvm/compile.dest/classes,/home/vnicolaou/mill/out/example/thirdparty/arrow/downloadedRepo.dest/arrow-bc9bf92cc98e01c21bdd2bf8640cf7db0f97204a/out/arrow-libs/core/arrow-core/jvm/compile.dest/classes
```

but the kotlinc friend paths are from the outer module

```
[
  "-Xfriend-paths=/home/vnicolaou/mill/out/example/thirdparty/arrow/downloadedRepo.dest/arrow-bc9bf92cc98e01c21bdd2bf8640cf7db0f97204a/out/arrow-libs/core/arrow-annotations/jvm/compile.dest/classes,/home/vnicolaou/mill/out/example/thirdparty/arrow/downloadedRepo.dest/arrow-bc9bf92cc98e01c21bdd2bf8640cf7db0f97204a/out/arrow-libs/core/arrow-atomic/jvm/compile.dest/classes,/home/vnicolaou/mill/out/example/thirdparty/arrow/downloadedRepo.dest/arrow-bc9bf92cc98e01c21bdd2bf8640cf7db0f97204a/out/arrow-libs/core/arrow-core/jvm/compile.dest/classes",
  "-Xexpect-actual-classes"
]
```

I'm not gonna handle this in this PR, instead I've moved this logic to AndroidKotlinModule and we can deal with it at a later stage

### DIFF
--- a/libs/androidlib/src/mill/androidlib/AndroidAppKotlinModule.scala
+++ b/libs/androidlib/src/mill/androidlib/AndroidAppKotlinModule.scala
@@ -26,25 +26,34 @@ import upickle.implicits.namedTuples.default.given
 @mill.api.experimental
 trait AndroidAppKotlinModule extends AndroidKotlinModule with AndroidAppModule { outer =>
 
-  def kotlinSources = Task.Sources("src/main/kotlin")
+  def kotlinSources: T[Seq[PathRef]] = Task.Sources("src/main/kotlin")
   override def sources: T[Seq[PathRef]] =
     super[AndroidAppModule].sources() ++ kotlinSources()
 
-  trait AndroidAppKotlinTests extends KotlinTests with AndroidAppTests {
-    def kotlinSources = Task.Sources("src/test/kotlin")
+  trait AndroidAppKotlinTests extends AndroidAppKotlinModule with AndroidAppTests {
+    override def kotlinVersion: T[String] = outer.kotlinVersion
+    
+    override def kotlinSources: T[Seq[PathRef]] = Task.Sources("src/test/kotlin")
+    
     override def sources: T[Seq[PathRef]] =
       super[AndroidAppTests].sources() ++ kotlinSources()
+      
+    override def kotlincPluginMvnDeps: T[Seq[Dep]] = outer.kotlincPluginMvnDeps()
   }
 
   trait AndroidAppKotlinInstrumentedTests extends AndroidAppKotlinModule
       with AndroidAppInstrumentedTests {
 
-    override final def kotlinVersion = outer.kotlinVersion
-    override final def androidSdkModule = outer.androidSdkModule
+    override final def kotlinVersion: T[String] = outer.kotlinVersion
+    override final def androidSdkModule: ModuleRef[AndroidSdkModule] = outer.androidSdkModule
 
-    def kotlinSources = Task.Sources("src/androidTest/kotlin")
+    override def kotlinSources: T[Seq[PathRef]] = Task.Sources("src/androidTest/kotlin")
+
+    override def kotlincPluginMvnDeps: T[Seq[Dep]] = outer.kotlincPluginMvnDeps()
+    
     override def sources: T[Seq[PathRef]] =
       super[AndroidAppInstrumentedTests].sources() ++ kotlinSources()
+
   }
 
   trait AndroidAppKotlinScreenshotTests extends AndroidAppKotlinModule with TestModule with Junit5 {
--- a/libs/androidlib/src/mill/androidlib/AndroidAppKotlinModule.scala
+++ b/libs/androidlib/src/mill/androidlib/AndroidAppKotlinModule.scala
@@ -32,12 +32,12 @@ trait AndroidAppKotlinModule extends AndroidKotlinModule with AndroidAppModule {
 
   trait AndroidAppKotlinTests extends AndroidAppKotlinModule with AndroidAppTests {
     override def kotlinVersion: T[String] = outer.kotlinVersion
-    
+
     override def kotlinSources: T[Seq[PathRef]] = Task.Sources("src/test/kotlin")
-    
+
     override def sources: T[Seq[PathRef]] =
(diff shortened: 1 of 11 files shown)


lefou:
I think the `-Xfriend-paths` is something we should directly support in `KotlinModule`. It is some generic feature of Kotlin, mostly used in tests I think, which needs tight coupling to the build tool to get the proper path and reduce the chance of misconfiguration. In Bazel, this is called `associates`.

See also:
* https://github.com/lefou/mill-kotlin/issues/110

vaslabs:
> I think the `-Xfriend-paths` is something we should directly support in `KotlinModule`. It is some generic feature of Kotlin, mostly used in tests I think, which needs tight coupling to the build tool to get the proper path and reduce the chance of misconfiguration. In Bazel, this is called `associates`.
> 
> See also:
> 
> * [Add convenience support for Kotlin `-Xfriend-paths` aka `associates` (in Bazel) lefou/mill-kotlin#110](https://github.com/lefou/mill-kotlin/issues/110)
> * [Support for Kotlin build-server-protocol/build-server-protocol#520 (comment)](https://github.com/build-server-protocol/build-server-protocol/issues/520#issuecomment-1845345952) and following comments

Yes, I agree it needs to be, I had trouble with the current kotlinc wiring as described above, I'm happy to do it in a follow up PR as I'll need more time to work on it

libs/androidlib/src/mill/androidlib/AndroidAppKotlinModule.scala
```diff
@@ -26,25 +26,34 @@ import upickle.implicits.namedTuples.default.given
...
 
   trait AndroidAppKotlinInstrumentedTests extends AndroidAppKotlinModule
       with AndroidAppInstrumentedTests {
 
-    override final def kotlinVersion = outer.kotlinVersion
-    override final def androidSdkModule = outer.androidSdkModule
+    override final def kotlinVersion: T[String] = outer.kotlinVersion
+    override final def androidSdkModule: ModuleRef[AndroidSdkModule] = outer.androidSdkModule
+
+    override def kotlinSources: T[Seq[PathRef]] = Task.Sources("src/androidTest/kotlin")
```
lefou, line 50: I think `kotlinSources` is meant to be private and should therefore not be overridden.

libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -117,7 +117,7 @@ trait KspModule extends KotlinModule { outer =>
       "-no-stdlib",
       "-language-version",
       kotlinLanguageVersion()
-    )
+    ) ++ kotlincOptions()
```
lefou, line 120: Should this be added at the start, so that in case something is duplicated, the `kspKotlincOptions` value overrides the other?

libs/androidlib/src/mill/androidlib/AndroidAppKotlinModule.scala
vaslabs: Thanks, I made it private instead then

Issue #5159, opened 2025-05-19
`NullPointerException` in scalac when compiling a project with `--release 8`


alexarchambault:
When trying to use my [case-app](https://github.com/alexarchambault/case-app) project with Mill 1.0 pre versions, I'm consistently getting `NullPointerException`s  when running `./mill __.compile`.

The NPE always happens at the same location in scalac (and similar locations between 2.12 and 2.13), when scalac tries to read a `CtSymClassPath`. scalac creates this ClassPath instance itself. It allows it to read the JVM class path when `--release` is used (for case-app, `--release 8`). Looking at [the location in `ZipFileSystem.java`](https://github.com/adoptium/jdk17u/blob/e1cda219b3fea2aa9db7953645299153e4e086df/src/jdk.zipfs/share/classes/jdk/nio/zipfs/ZipFileSystem.java#L1892) where the NPE happens, it seems we can get an NPE iff the `ZipFileSystem` [was closed](https://github.com/adoptium/jdk17u/blob/e1cda219b3fea2aa9db7953645299153e4e086df/src/jdk.zipfs/share/classes/jdk/nio/zipfs/ZipFileSystem.java#L506).

It seems the problem happens with both Scala 2.12.20 and 2.13.16. It's not always happening for the same module (not totally reproducible), but I seem to always run into it for some module when running this command from clean sources.

Bisecting Mill releases, it seems `0.13.0-M2-74-132761` is the first Mill version to suffer from this, while `0.13.0-M2-59-17fa7b` doesn't. Bisecting commits, it seems this happens since https://github.com/com-lihaoyi/mill/pull/5082 has been merged. I'll keep investigating, but I'm not sure what's the problem yet.

Reproduction:

Clone the branch of [this PR](https://github.com/alexarchambault/case-app/pull/615), and try to compile all modules at once
```text
$ git clone https://github.com/alexarchambault/case-app.git -b mill-1.0

$ cd case-app

$ ./mill -j 1 -w __.compile
…
[3495/3495, 1 failed] ============================== __.compile ============================== 40s
1 tasks failed
cats.js[2.12.20].test.compile java.lang.NullPointerException: Cannot invoke "java.util.LinkedHashMap.get(Object)" because "this.inodes" is null
    jdk.zipfs/jdk.nio.zipfs.ZipFileSystem.getInode(ZipFileSystem.java:1892)
[...]

PR #5171, opened 2025-05-21
Manually stop scalac timers

alexarchambault:
IIUC, in order to cache opened zip files, scalac delays the closing of those with [a timer](https://github.com/scala/scala/blob/6d0b1346d645ea79ad6aaedba90a098b4af5b01f/src/compiler/scala/tools/nsc/classpath/ZipAndJarFileLookupFactory.scala#L326C15-L326C20). The thread created by this timer keeps hanging around, making Mill leak threads. So https://github.com/com-lihaoyi/mill/pull/5082 disabled that timer. It seems disabling that timer creates new issues: https://github.com/com-lihaoyi/mill/issues/5159 (which look like scalac bugs to me).

So that PR handles those timers another way: it uses reflection to get the timers created by scalac, and stops them manually, when we dispose of the scala compiler.

### DIFF
--- a/libs/scalalib/worker/src/mill/scalalib/worker/JvmWorkerImpl.scala
+++ b/libs/scalalib/worker/src/mill/scalalib/worker/JvmWorkerImpl.scala
@@ -118,6 +118,10 @@ class JvmWorkerImpl(
       classloaderCache.updateWith(compilersSig) {
         case Some((cl, groupOpt, 1)) =>
           for (group <- groupOpt) {
+            // Trying to find the timer created by scala.tools.nsc.classpath.FileBasedCache
+            // and its thread. We cancel the timer and interrupt the thread, so that
+            // the thread cleanly exits, and doesn't keep hanging around.
+
             val threadOpt = Thread.getAllStackTraces()
               .asScala
               .keysIterator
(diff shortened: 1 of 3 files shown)


alexarchambault:
(Will add more details in the summary…)

alexarchambault:
The CI test checking for thread leaks fails, but manual tests show that the timer threads do get stopped 🤔

lihaoyi:
Seems like the tests pass and indicate that after `mill shutdown`, the timer threads are gone even without the original flag and without the cleanup code that was proposed in this PR. I'm not sure why, but I think it's reasonable to revert the original flag for now

libs/scalalib/worker/src/mill/scalalib/worker/JvmWorkerImpl.scala
```diff
@@ -657,13 +683,9 @@ class JvmWorkerImpl(
...
   override def close(): Unit = {
-    val urlClassLoaders = classloaderCache
-      .map(_._2._1)
-      .collect { case v: AutoCloseable => v }
-
-    // pprint.log(classloaderCache.map(_._2._1))
-    // pprint.log(urlClassLoaders.map(_.getURLs))
-    urlClassLoaders.foreach(_.close())
+    scalaCompilerCache.close()
+    javaOnlyCompilerCache.close()
```
lihaoyi, line 687: @alexarchambault this is what your PR was missing

PR #5175, opened 2025-05-22
Fix `transitiveCoursierProjects` cause OOM on large build (#4452)

### LINKED ISSUE #4451: OOM when running compile in large build
As discussed in Mill's discord, I'm having OOM issue when updating Mill to version `0.12.7` and running `__.compile` command.
There might be some issues in https://github.com/com-lihaoyi/mill/pull/4145 as I tried Mill version `0.12.5-131-1e6358` and it's stable for my build.

Heapdump file is huge (> 7GB) so below are some heap dump and usage screenshots from VisualVM
### 0.12.5

Image
Image
Image

### 0.12.7

Image
Image
Image

lefou:
Resolve #4451 

It seems that `transitiveCoursierProjects` is populated with repetitive
transitive modules, that's why it only causes problem for large build.


lefou:
Wrong branch

PR #5186, opened 2025-05-25
New cut at manually stopping scalac timers 

alexarchambault:
This is making the leaking thread test green without needing to ignore the timer threads in it. I pushed this code at some point in #5171, but thought this could be reverted.


alexarchambault:
Might look deeper into it later, but [CI is green](https://github.com/alexarchambault/mill/actions/runs/15240021963) with these changes (that drop the `Timer` mentions in `LeakHygieneTests`) 🤔

lihaoyi:
got it, but in that case why are no timer threads running after running `__.compile`? Since we haven't `clean`ed and the classloader used to `compile` is still hanging around, shouldn't the timer thread be still running as well?

lihaoyi:
Closing this out of inactivity

PR #5198, opened 2025-05-26
Factor thread pool used to run tasks

alexarchambault:
This makes `mill.exec.Execution` accept a thread pool in its constructor, rather than creating one every time tasks need to be run. In BSP and in watch mode, this should save the thread pool creation across task runs. I don't have numbers to back the impact of this change, but factoring the thread pool was part of [a bunch of optimizations](https://github.com/coursier/sbt-coursier/pull/256) in the coursier integration in sbt a few years ago. Plus given we follow more of a single-execution-at-a-time model, there's no reason not to re-use the thread pool across executions.

### DIFF
--- a/core/exec/src/mill/exec/Execution.scala
+++ b/core/exec/src/mill/exec/Execution.scala
@@ -6,7 +6,7 @@ import mill.constants.OutFiles.{millChromeProfile, millProfile}
 import mill.define._
 import mill.internal.PrefixLogger
 
-import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.{ConcurrentHashMap, ThreadPoolExecutor}
 import java.util.concurrent.atomic.{AtomicBoolean, AtomicInteger}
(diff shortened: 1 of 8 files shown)


lihaoyi:
Could you ad some timing logs to see how long creating the threadpool takes? spawning a new one wverytime has some nice isolation properties, e.g. not leaking threadlocals from one run to another, so if it's not too expensive i would prefer starting afresh each time

lihaoyi (approved):
lgtm

alexarchambault:
About the benchmark, the extra 0.25 ms cost is actually only paid by execution (once per BSP request or Mill command). I spawned 8 runnables because that's needed for the thread pool to actually spawn 8 threads. So it should only be paid once when we start to use a thread pool. But after that, sharing the thread pool or not should be even.

So while I'd merge because I'm used to manage threads and thread pools carefully, it's very unlikely Mill users will notice 😅

PR #5200, opened 2025-05-26
Rework MillBuildServer concurrency handling

alexarchambault:
This reworks and I think simplifies concurrency handling in `MillBuildServer`. Basically, with these changes, `MillBuildServer` doesn't rely on thread pools and scala futures any more. It starts a single thread upon start-up, that waits for incoming requests via a concurrent queue. BSP requests that use evaluators don't do any processing and simply enqueue messages (more precisely, they enqueue things like `BspEvaluators => Unit`). The thread processes incoming messages one-by-one, and makes sure we own the lock on the `out/` directory before processing a message and running tasks for it.

This also helps further changes that recreate the evaluators upon meta-build changes (better version of https://github.com/com-lihaoyi/mill/pull/4957, for an upcoming PR).

As a bonus, this brings the ability to cancel requests whose processing hasn't started yet. See the changes in integration tests.


alexarchambault:
In principle, we could have kept using scala futures along with a single-threaded thread pool, but that wouldn't have given us cancellation support, and it's also a problem with support for reloading the meta-build (upcoming PR, see description above), where we have to basically suspend request processing when doing so, which is trickier to achieve with futures it seems.

runner/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
```diff
@@ -660,83 +661,126 @@ private class MillBuildServer(
...
+  private val queue = new LinkedBlockingQueue[(BspEvaluators => Unit, Logger, String)]
+  private var stopped = false
+  private val evaluatorRequestsThread: Thread =
+    new Thread("mill-bsp-evaluator") {
+      setDaemon(true)
+      def waitForEvaluators(): BspEvaluators =
+        bspEvaluatorsOpt match {
+          case Some(ev) => ev
+          case None =>
+            Thread.sleep(1000L)
```
lefou, line 673: Each `Thead.sleep` should come with a comment: Why and why the value?

lefou (commented):
Looks reasonable to me, from looking from the browser.

lihaoyi (commented):
Looks reasonable to me, nice to strip out the layer of `concurrent.Future` complexity that we weren't really getting any benefits from

runner/bsp/worker/src/mill/bsp/worker/BspWorkerImpl.scala
```diff
@@ -93,4 +96,21 @@ object BspWorkerImpl {
         )
     }
   }
+
+  private val executorCounter = new AtomicInteger
+  private def createJsonrpcExecutor(): ExecutorService =
+    Executors.newCachedThreadPool(
```
lihaoyi, line 102: What's this for? I thought the PR description said we were trying to avoid using thread pools

runner/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
alexarchambault: The last push gets rid of it, by using a promise and awaiting it to have a value there

runner/bsp/worker/src/mill/bsp/worker/BspWorkerImpl.scala
alexarchambault: `MillBuildServer` itself doesn't use a thread pool any more, but the jsonrpc library still needs one. Before this PR, both shared a common thread pool, now it's specific to jsonrpc handling.

I think the jsonrpc library uses it to accept incoming connections, to call methods on our build server instance (even though they don't block), etc.

lihaoyi (approved):

alexarchambault:
Merging, as the CI failure seems unrelated to these changes

PR #5207, opened 2025-05-27
Avoid printing "Some()" in Server messages

lolgab:


lefou (commented):
USing `.orNull` when we assume it can't be a `null` (because we checked above) is a bit misleading. We could use `.get` and make our assumption explicit. (Which could lead to failure after the next refactoring, though)

lefou:
If the option might be `null`, lets use `.getOrElse("_")`. We replace the unset-value with a `_`. Or use `none` if you prefer a word.

> I wouldn't comment only because it's a debug print.

IMHO, that doesn't matter. The code must be read and understood nevertheless.

lefou:
I changed it to use `0.0.0` for the unknown last Mill version and `<system>` for the unspecified JVM version. @lolgab WDYT?

lolgab:
Maybe `<unknown>` instead of `0.0.0`?

lolgab:
Thank you! @lefou

PR #5214, opened 2025-05-28
Refactor internal BSP tasks out of `JavaModuleApi`

lefou:
We use a dedicated `trait` to hold internal BSP tasks. This is, to free the binary compatibly Module API from these likely changing tasks.

Also refined the package layout in `mill.api.internal` and introduced sub-packages `bsp` and `idea`.


lefou:
Moved various internal BSP tasks into a dedicated trait and update the PR description.

lefou:
I also identified some internal tasks to support `GenIdea`, which I'd like the refactor following the same pattern but in a separate PR. I will wait with that until we have merged this PR, so ensure the pattern we use is consistent.

lefou:
@lihaoyi I'd like to merge this one. It only juggles with unreleased API and should decouple internals from the public `JavaModuleApi` hierarchy.

lihaoyi:
@lefou i havent look at it but go ahead if CI and manual testing passes

PR #5219, opened 2025-05-28
Android: working android tests for architecture samples

vaslabs:
Completes the missing features to allow for android tests (instrumented tests) to run for architecture-samples (hilt android Todo application)

## Changes

These were the necessary changes needed to make instrumented tests for architecture samples work.

### Instrumented manifest

AGP seems to be creating a hardcoded temporary file (which is easy to miss as the task is quite fast and the file is deleted once is finished), but I've managed to pick it up with some hackery and it looks like this

```XML
<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.example.android.architecture.blueprints.main.test">
  <uses-sdk android:minSdkVersion="21" android:targetSdkVersion="35"/>
  <application>
    <uses-library android:name="android.test.runner"/>
  </application>
  <instrumentation android:name="com.example.android.architecture.blueprints.todoapp.CustomTestRunner" android:targetPackage="com.example.android.architecture.blueprints.main" android:handleProfiling="false" android:functionalTest="false" android:label="Tests for com.example.android.architecture.blueprints.main"/>
</manifest>
```

The manifest merging for instrumented tests, seem to only be picking manifest files from androidx.test packages, so I hardcoded that in and I ended up with the same manifest as in AGP . The task to get is is
```bash
./gradlew :app:processDebugAndroidTestManifest
```

### Debug sources

I've added the debug sources as part of the AndroidAppModule, to keep the scope of this PR small. I'll separate the debug from release similarly to how R8 was done in this PR in the future. In order to avoid making a lot of changes for the debug sources, I've included the debug sources as part of the build.mill, which will probably be removed when the release/debug separation work is completed.

### R8

I haven't managed to make the architecture-samples work with D8, whether some meta-inf is missing or not, with R8 and the same configuration choices as in the architecture-samples repo, everything works properly (app, unit tests, instrumented tests). I've then separated the R8 functionality into its own trait.

It seems that by using R8 we can drop some of the manual code we have (like picking the meta-inf files manually) and creating other needed directories, as R8 seems to be doing all of that work. We can keep purely d8 builds for legacy purposes if we want to of course, but we'll see how it goes in the future when we integrate more apps.

 For R8, desugaring being enabled is also the sane default, so I switched that too, following up to failed builds of architecture-samples with R8.

### Cleanups

- Some manifest generated values are not handled by the merger tool instead of manually constructing them
- Printing out ignored for stream output of instrumented tests along with each line was confusing when dealing with failed tests, which I had initially, so I removed it.

### Fixes
- Instrumented tests should run with androidApplicationId as the package, as that identifies the android app, was missed [from the previous PR](https://github.com/com-lihaoyi/mill/pull/5210)
- Instrumented packaged class files, also need to include the classfiles of the  modules they depend on (e.g. shared-tests in case of the architecture-samples).

## Running experience

Performance-wise, this feels similar to AGP (I didn't do any benchmarks)

https://github.com/user-attachments/assets/94e29098-91c8-4fb2-af62-8294833d261a

## Summary

All in all, this is another milestone, where we have a fairly complex android app, which uses KSP and R8, can be built, run and tested with both unit and instrumented tests


vaslabs:
this should be ready for review

PR #5236, opened 2025-05-31
Make query-selector resolution recognize tasks and commands that do not take parameters

lihaoyi:
This PR removes the need for `+` when running multiple tasks or commands in common cases:

Before:

```bash
./mill foo + bar + baz
```

After:

```bash
./mill foo bar baz
```

Mill is able to recognize scenarios where:

1. `foo` is not a command, and thus cannot take parameters on the right
2. `foo` is a command but it does not take positional or leftover arguments, and thus cannot take any parameters on the right

This means that in most common cases, you no longer need to separate tasks with `+` at the command line, and can just list them one after another like with most build tools

Mill resolves the query-selector tokens from left to right, and if it finds a token that resolves to a command that takes parameters, it then takes the remaining tokens and assigns them to be the arguments of that command

While this heuristic isn't perfect, it should work well in 99% of cases, which should reduce the friction of people running multiple tasks or commands from Mill's command-line interfact


lihaoyi:
> > For now we continue supporting the `+` separator, but I expect most people will not need to use it, and we might want to replace it with `--` since we no longer need two separators
> 
> `--` is already used in a lot of commands, e.g. `test` and `testOnly` but also all those that forwards their args to external tools.
> 
> (I still have a todo to add a `++` to add tasks that should "run after" all previously selected tasks.)

In that case we'll probably need to stick with `+` then

lefou:
We might want to print the finally resolved tasks in `debug` (verbose) mode, so we can easily verify if the heuristics match our expectation (when a user reports some strange behavior).

core/resolve/src/mill/resolve/Resolve.scala
```diff
@@ -332,7 +394,7 @@ private[mill] trait Resolve[T] {
       nullCommandDefaults: Boolean,
       allowPositionalCommandArgs: Boolean,
       resolveToModuleTasks: Boolean
-  ): Result[Seq[T]] = {
+  ): Result[(Seq[T], Seq[Resolved])] = {
```
lefou, line 397: Can we use a named tuple here, so it's more clear what we get at each position?

core/resolve/src/mill/resolve/ResolveCore.scala
```diff
@@ -448,15 +448,15 @@ private object ResolveCore {
 
     val namedTasks = Reflect
       .reflect(cls, classOf[Task.Named[?]], namePred, noParams = true, cache.getMethods)
-      .map { m =>
-        Resolved.NamedTask(Segments.labels(cache.decode(m.getName))) ->
-          None
+      .collect {
+        case m if !classOf[Task.Command[?]].isAssignableFrom(m.getReturnType) =>
+          Resolved.NamedTask(Segments.labels(cache.decode(m.getName))) -> None
```
lefou, line 453: `collect` may silently drop some items, while the previous `map` was complete. I think we want a default case to catch that case or a comment explaining, why this isn't needed.

core/resolve/test/src/mill/resolve/ParseArgsTests.scala
lefou, line 1: I there a replacement for these tests?

lihaoyi:
Probably not going to proceed with this for now. The basic problem is that we can make `./mill foo.compile bar.compile qux.compile` work, but we cannot make `./mill foo.test bar.test qux.test` work, because `test` is a command and thus will always need to take parameters (e.g. to pass to the test suite). Unless we require users to quote test tokens e.g. `./mill "foo.test arg1 arg2 arg3"`, there is no way of squaring that, and I expect  having `foo.compile bar.compile qux.compile` work but `foo.test bar.test qux.test` not work will likely be a big cause of confusion.

There are probably other things we can do to improve the parameter parsing logic, but for now going to close this

PR #5245, opened 2025-06-02
Impore `Task.{Source,Sources}` API

lefou:
Accept the following tpye:
* `os.SubPath` - an sub-path resolved against `moduleDir`
* `os.RelPath` - an relative path resolved against `moduleDir`
* `os.Path` - an absolute path
* `String` - a sub-path relative to `moduleDir`; the special string `"."` is treated as the `moduleDir`.

### DIFF
--- a/core/define/src/mill/define/Task.scala
+++ b/core/define/src/mill/define/Task.scala
@@ -148,34 +148,31 @@ object Task {
    * signature for you source files/folders and decides whether or not downstream
    * [[Task.Computed]]s need to be invalidated and re-computed.
    */
-  inline def Sources(inline values: Result[os.Path]*)(implicit
+  inline def Sources(inline values: (os.Path | os.SubPath | os.RelPath | String)*)(implicit
       inline ctx: mill.define.ModuleCtx
   ): Simple[Seq[PathRef]] = ${
-    Macros.sourcesImpl('{ Result.sequence(values.map(_.map(PathRef(_)))) })('ctx)
+    Macros.sourcesImpl('{ values.map(p => PathRef(mapToPath(p))) })('ctx)
   }
 
-  inline def Sources(inline values: os.SubPath*)(implicit
-      inline ctx: mill.define.ModuleCtx,
-      dummy: Boolean = true
-  ): Simple[Seq[PathRef]] = ${
-    Macros.sourcesImpl(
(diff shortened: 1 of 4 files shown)


core/define/src/mill/define/Task.scala
```diff
@@ -148,34 +148,31 @@ object Task {
    * signature for you source files/folders and decides whether or not downstream
    * [[Task.Computed]]s need to be invalidated and re-computed.
    */
-  inline def Sources(inline values: Result[os.Path]*)(implicit
+  inline def Sources(inline values: (os.Path | os.SubPath | os.RelPath | String)*)(implicit
```
lihaoyi: This big union type looks pretty out of place. Could we use `os.FilePath` here? Maybe if we add a few more implicits upstream in OS-Lib to make the conversions from string work?

lefou: Yeah. Once we use just `os.FilePath`, we can't profit from `String` to `os.SubPath` translation. We either want to support a more generic conversion for `String` to `os.FilePath`, which also means we want to handle absolute and relative paths, or find some other way to support conversion.

I committed some compromise acepting `os.SubPath | os.FilePath` to see if CI likes it.

lihaoyi:
Isn't `os.SubPath | os.FilePath` just `os.FilePath`, since `os.SubPath` is a subclass of it?

lefou:
@lihaoyi 

> Isn't `os.SubPath | os.FilePath` just `os.FilePath`, since `os.SubPath` is a subclass of it?

Yeah. Here's my comment from commit 5a28bbb43d20ed815ade7de597e496f82685385a

> By using `os.SubPath | os.FilePath`, which is redundant as `os.SubPath` is already a sub-type of `os.FilePath`, we allow implicit converstions from `String` to `os.SubPath`.

Also from https://github.com/com-lihaoyi/mill/pull/5245#discussion_r2123756412

> Once we use just `os.FilePath`, we can't profit from `String` to `os.SubPath` translation. We either want to support a more generic conversion for `String` to `os.FilePath`, which also means we want to handle absolute and relative paths, or find some other way to support conversion.
>
> I committed some compromise acepting `os.SubPath | os.FilePath` to see if CI likes it.

lihaoyi (approved):
Looks good to me!

Issue #5095, opened 2025-05-09
Mill since `0.12.6` no longer includes optional compile dependencies


lefou:
A project which uses optional compile-time dependencies, a switch from Mill `0.12.5` to `0.12.6` suddenly no longer compiles due to missing dependencies on the classpath.

The project dependencies are defined like this: 

```scala
override def compileIvyDeps = Agg(
  ivy"de.tototec:de.tototec.utils.functional:2.3.0".optional(true)
)
```

The failure with Mill `0.12.6` (same with `0.12.10`)

```console
> MILL_VERSION="0.12.6" mill __.compile
...
[52/56] compile
[52] [info] compiling 17 Java sources to /home/lefou/work/opensource/domino-java/out/compile.dest/classes ...
[52] [error] /home/lefou/work/opensource/domino-java/src/domino/java/OsgiContext.java:3:42: package de.tototec.utils.functional does not exist
[52] [error] de.tototec.utils.functional.FList
[52] [error]                                  ^
...
2 tasks failed
compile javac returned non-zero exit code
docJar Interactive Subprocess Failed (exit code 1)

```

With Mill `0.12.5`, compilation works.

```console
> MILL_VERSION="0.12.5" mill __.compile
...
[52/56] compile
[52] [info] compiling 17 Java sources to /home/lefou/work/opensource/domino-java/out/compile.dest/classes ...
[54] [info] done compiling
```

According to https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html, optional dependencies are not transitively included, so marking them optional is relevant for the generated and published `pom.xml`. It's a way to communicate to a user that a certain dependency might be necessary for some specific functionionality. But they should still be resolved and added to the classpath in the same module where they are defined.

lefou:
@alexarchambault This is most likely related to your changes. From looking at `ivyDepsTree` ( without any extra opitons) no optional dependency (from `ivyDeps` or `compileIvyDeps`) is shown.

alexarchambault:
@lefou Likely a misconfiguration in Mill. I'm having a look right now.

PR #5246, opened 2025-06-02
Add optional deps to the class path of their modules

alexarchambault:
Optional dependencies are marked as optional for consumers of a module.

They should still be added to the class path of the module that has them as optional dependencies, as Maven does, as mentioned by @lefou in https://github.com/com-lihaoyi/mill/issues/5095

Fixes https://github.com/com-lihaoyi/mill/issues/5095

### DIFF
--- a/libs/scalalib/src/mill/scalalib/JavaModule.scala
+++ b/libs/scalalib/src/mill/scalalib/JavaModule.scala
@@ -1554,6 +1554,11 @@ object JavaModule {
   final case class InternalRepo(projects: Seq[cs.Project])
       extends cs.Repository {
 
+    // Reversing the sequence before calling toMap, so that earlier elements have precedence
+    // over later one, in case they have the same module / version.
+    // That's useful for the handling of optional dependencies, where the main project, with
+    // initially optional dependencies marked as non-optional, is put upfront, but might still
+    // be pulled transitively, in that case without the special handling of optional dependencies.
     private lazy val map = projects.reverseIterator.map(proj => proj.moduleVersion -> proj).toMap
 
     override def toString(): String =
--- a/libs/scalalib/test/src/mill/scalalib/ResolveDepsTests.scala
+++ b/libs/scalalib/test/src/mill/scalalib/ResolveDepsTests.scala
@@ -201,29 +201,30 @@ object ResolveDepsTests extends TestSuite {
         val dependsOnOptionalRunCp = eval(TestCase.optional.dependsOnOptional.runClasspath)
           .fold(_.get, _.value).map(_.path)
 
-        // optional dependencies in mvnDeps should be added to the current module class path,
-        // but not to the class path of its dependees
+        // optional dependencies in mvnDeps should be added to the current module class path
         assert(optionalCompileCp.exists(_.last == "interface-1.0.29-M1.jar"))
         assert(optionalRuntimeCp.exists(_.last == "interface-1.0.29-M1.jar"))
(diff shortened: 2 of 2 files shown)


libs/scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -647,7 +653,18 @@ trait JavaModule
...
+    // class paths of this module
+    val project0 = project.withDependencies0(
+      project.dependencies0.map {
+        case (conf, dep) if dep.optional =>
+          (conf, dep.withOptional(false))
+        case other =>
+          other
+      }
+    )
+    JavaModule.InternalRepo(Seq(project0) ++ transitiveCoursierProjects())
```
lefou, line 669: I'm a bit surprised, we need to manually manage the handling of direct optional dependencies. Isn't this something that also must work in `sbt` and `coursier` cli tool?

libs/scalalib/src/mill/scalalib/JavaModule.scala
```diff
@@ -1633,7 +1650,7 @@ object JavaModule {
   final case class InternalRepo(projects: Seq[cs.Project])
       extends cs.Repository {
 
-    private lazy val map = projects.map(proj => proj.moduleVersion -> proj).toMap
+    private lazy val map = projects.reverseIterator.map(proj => proj.moduleVersion -> proj).toMap
```
lefou, line 1562: Is this a performance or a logic improvement?

libs/scalalib/test/src/mill/scalalib/ResolveDepsTests.scala
```diff
@@ -168,5 +184,50 @@ object ResolveDepsTests extends TestSuite {
         }
       }
     }
+
+    test("optional") {
```
lefou, line 188: The various cases are hard to read. Can we group them by module?

libs/scalalib/src/mill/scalalib/JavaModule.scala
alexarchambault: Not really. The coursier code-base mainly deals with fetching dependencies, not publishing them out of local modules. Optional dependencies in the input are handled the same way as those coursier stumbles upon during resolution: that is, by ignoring them (although there's an option to keep them nonetheless, but it applies to input dependencies as well as transitive ones equally).

And I'm not sure sbt handles optional dependencies like this, I'm pretty sure there isn't a special handling for those in [sbt-coursier](https://github.com/coursier/sbt-coursier).

But doing like Maven does, as you suggest in https://github.com/com-lihaoyi/mill/issues/5095, seems correct.

Maybe coursier could retain root dependencies that are marked as optional, but we'd still need a special handling in Mill: since #4145, Mill modules's direct dependencies are seen as transitive dependencies by coursier (the Mill modules themselves are the root dependencies).

libs/scalalib/src/mill/scalalib/JavaModule.scala
alexarchambault: Logic one. I'm going to add a comment for that.

libs/scalalib/test/src/mill/scalalib/ResolveDepsTests.scala
alexarchambault: Sure

lefou:
@alexarchambault It would be nice, if we could backport this to `0.12.x` as well.

PR #5247, opened 2025-06-02
Android: Example with versions from boms (with compose-bom)

vaslabs:
Example with versions to be resolved via Boms for android architecture samples. E.g. our build.mill for architecture samples can now declare the dependencies in the [compose-bom without a version](https://github.com/android/architecture-samples/blob/main/gradle/libs.versions.toml#L83) .

This is just an improvement on `build.mill` for the architecture-sample thirdparty example , getting it closer to the gradle configuration (no versions on the androidx.compose)

### DIFF
--- a/libs/androidlib/src/mill/androidlib/AndroidModule.scala
+++ b/libs/androidlib/src/mill/androidlib/AndroidModule.scala
@@ -158,7 +158,7 @@ trait AndroidModule extends JavaModule {
           VariantMatcher.Equals("androidJvm"),
           VariantMatcher.Equals("jvm")
         ))
-    ).withForceDepMgmtVersions(Some(true))
+    )
   }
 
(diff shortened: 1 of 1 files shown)


lefou (commented):
Looks good to me. I think the `def composeDeps` makes reading more complicated in this simple example. Since it's not re-used in multiple places, as simple comment before the deps managed by the BOM could be an alternative.

vaslabs:
> Looks good to me. I think the `def composeDeps` makes reading more complicated in this simple example. Since it's not re-used in multiple places, a simple comment before the deps managed by the BOM could be an alternative.

fair, I'm currently working on getting [compose-samples ](https://github.com/android/compose-samples) , I can do it the same there and change it in the next PR if you are ok with it

Issue #5039, opened 2025-05-02
`MILL_TEST_RESOURCE_DIR` not passed in when running tests in IntelliJ


lihaoyi:

lihaoyi:
@arturaz is working on this. Current investigation shows that the `resources/` folder is not being included in the runtime classpath when performing tests in IntelliJ, but it is included when running a main method. Not sure if this is Intellij's fault or Mill's

arturaz:
What we've learned so far:

- SBT successfully passes the ENV vars with the given `build.sbt`
```scala
val scala3Version = "2.13.16"

lazy val root = project
  .in(file("."))
  .settings(
    name := "mill-resources-test-sbt",
    version := "0.1.0-SNAPSHOT",

    scalaVersion := scala3Version,

    fork := true,
    envVars := Map("DOGGIES" -> "woof"),
    (Test / envVars) := Map("DOGGIES" -> "woof from test"),

    libraryDependencies += "com.lihaoyi" %% "os-lib" % "0.10.7",
    libraryDependencies += "org.scalameta" %% "munit" % "1.1.1" % Test
  )
```

-  Metals invokes `buildTarget/jvmRunEnvironment` for SBT but not for mill:
```
[Trace - 00:00:29 PM] Received response 'buildTarget/jvmRunEnvironment - (15)' in 14ms
Result: {
  "items": [
    {
      "target": {
        "uri": "file:/home/arturaz/work/mill-resources-test-sbt/#root/Test"
      },
      "classpath": [
        "file:/home/arturaz/work/mill-resources-test-sbt/target/scala-2.13/test-classes/",
        "file:/home/arturaz/work/mill-resources-test-sbt/target/scala-2.13/classes/",
         ...
      ],
      "jvmOptions": [],
      "workingDirectory": "file:/home/arturaz/work/mill-resources-test-sbt/",
      "environmentVariables": {
        "DOGGIES": "woof from test"
      }
    }
  ]
}
Error: null
```

The only difference so far seems to be that when invoking `buildTarget/scalaTestClasses` mill and SBT responds a bit differently in the `uri` field.


The SBT responds with:
```
[Trace - 00:08:55 PM] Received response 'buildTarget/scalaTestClasses - (33)' in 341ms
Result: {
  "items": [
    {
      "target": {
        "uri": "file:/home/arturaz/work/mill-resources-test-sbt/#root/Test"
      },
      "framework": "munit",
      "classes": [
[...]

PR #5265, opened 2025-06-06
Export `MILL_TEST_RESOURCE_DIR` to BSP project definitions

arturaz:
Fixes https://github.com/com-lihaoyi/mill/issues/5039.

We've introduced `allForkEnv` to `JavaModule`, which now contains the environment variables that Mill adds, in addition to user-defined environment variables.

Updated the integration tests to include the newly-exported environment variables so that `./mill integration.ide[bsp-server].local.__.testForked mill.integration.BspServerTests.requestSnapshots` would work.

Tested with Metals (requires a fix in Metals: https://github.com/scalameta/metals/pull/7544) and Idea (only works with Junit for now: https://youtrack.jetbrains.com/issue/SCL-23964/Running-utest-test-in-Intellij-does-not-pass-environment-variables-configured-in-Mill)


integration/ide/bsp-server/resources/snapshots/build-targets-jvm-run-environments.json
```diff
@@ -18,14 +18,52 @@
...
           "arguments": []
         }
       ]
     },
+    {
+      "target": {
+        "uri": "file:///workspace/app/test"
+      },
+      "classpath": [
+        "/coursier-cache/https/repo1.maven.org/maven2/ch/qos/logback/logback-core/1.5.15/logback-core-1.5.15.jar",
```
arturaz: This is questionable, why do classpath entries here don't have `file://` prefix?

integration/ide/bsp-server/resources/snapshots/build-targets-jvm-run-environments.json
```diff
@@ -18,14 +18,52 @@
...
+        "/workspace/app/test/compile-resources",
+        "/workspace/app/test/resources",
+        "/workspace/out/app/test/compile.dest/classes",
+        "/coursier-cache/https/repo1.maven.org/maven2/org/scala-sbt/test-interface/1.0/test-interface-1.0.jar",
+        "/mill-workspace/out/libs/testrunner/entrypoint/compile.dest/classes"
+      ],
+      "jvmOptions": [],
+      "workingDirectory": "/workspace",
+      "environmentVariables": {
+        "MILL_TEST_RESOURCE_DIR": "/workspace/app/test/resources"
```
arturaz: Should this be exported here, or only in `jvm-test-environments.json`?

lefou: The test runner is a dedicated runner, hence `trait TestModule extends RunModule`. `forkEnv` and `forkArgs` come from `RunModule`, so it's correct that these are also exposed here. It's a Mill-specific redundancy, since we always use dedicated test modules in contrast to sbt, Maven and Gradle, where tests are some feature/scope of normal modules.

integration/ide/bsp-server/resources/snapshots/build-targets-jvm-test-environments.json
```diff
@@ -18,14 +18,52 @@
...
...2.13-1.1.3.jar,file:/coursier-cache/https/repo1.maven.org/maven2/org/fusesource/jansi/jansi/2.4.1/jansi-2.4.1.jar,file:/coursier-cache/https/repo1.maven.org/maven2/commons-codec/commons-codec/1.17.1/commons-codec-1.17.1.jar,file:/coursier-cache/https/repo1.maven.org/maven2/org/apache/commons/commons-lang3/3.16.0/commons-lang3-3.16.0.jar,file:/mill-workspace/out/core/coursierutil/compile.dest/classes/,file:/mill-workspace/out/core/define/compile.dest/classes/,file:/mill-workspace/out/core/constants/buildInfoResources.dest/,file:/mill-workspace/out/core/constants/compile.dest/classes/,file:/mill-workspace/out/core/api/buildInfoResources.dest/,file:/mill-workspace/out/core/api/compile.dest/classes/,file:/mill-workspace/out/core/util/buildInfoResources.dest/,file:/mill-workspace/out/core/util/compile.dest/classes/,file:/mill-workspace/out/libs/scalalib/api/buildInfoResources.dest/,file:/mill-workspace/out/libs/scalalib/api/compile.dest/classes/,file:/mill-workspace/out/libs/testrunner/entrypoint/compile.dest/classes/,file:/mill-workspace/out/libs/testrunner/compile.dest/classes/,file:/mill-workspace/out/libs/scalalib/compile.dest/classes/,file:/mill-workspace/libs/scalalib/resources/",
```
lefou: Regarding the single slash after `file:`: Could be a missing `sanitizeUri` call, that should ensure the same formatting as everywhere else.

integration/ide/bsp-server/resources/snapshots/build-targets-jvm-run-environments.json
lefou: Tricky case. The BSP spec often uses `URI[]`, but here a `string[]` is required. I still think using an URI-string would be more consistent.

lefou: E.g. in `JavaOptionsItem` is also a `classpath: string[]` specified, but we return URI-strings.

arturaz: But why does it differ between the app and test projects?

lefou: Don't know. Looks like a bug.

lihaoyi:
CC @alexarchambault if you could take a look, since you're familiar with the BSP parts of the codebase

alexarchambault:
Just pushed a few commits that fix some confusion there was around BSP run and test environments, I think (that was there in `main` already), and makes sure we pass the class path as URIs for test environments

core/api/src/mill/api/internal/README.md
```diff
@@ -0,0 +1,4 @@
+These `*ModuleApi` traits are only allowed to use Scala standard library to decouple the `build.mill` code from
+the `mill.runner` code.
+
+https://github.com/com-lihaoyi/mill/pull/4879
```
lihaoyi: This duplicates docs already present in the `core/api/package.mill` scaladoc. Can we combine them there? The build scaladoc is visible to `./mill inspect` and in the IDE when working with build files, which is better than the readme.md which doesn't integrate with anything

lihaoyi:
@arturaz looks like some tests need to be updated but otherwise the PR is probably in good shape

lihaoyi:
> No run env for test modules, no test env for run modules

@alexarchambault Is this correct? Right now Mill `JavaTests` or `ScalaTests` modules all can have `public static void main(String[] args)` methods , and if they are defined the module can be run from the command line.

lefou:
What is the rationale for adding `allForkEnv`? Is there an issue with overriding `forkEnv`?

It is a bit misleading. We have various `all`-prefixed tasks in Mill, but they typically don't introduce new data but just aggregate other tasks. Here, we use the `all` prefix to provide some new stuff. Why can't we provide the env variables by just overriding `forEnv`?

lefou:
@lihaoyi 

> > No run env for test modules, no test env for run modules
> 
> @alexarchambault Is this correct? Right now Mill `JavaTests` or `ScalaTests` modules all can have `public static void main(String[] args)` methods , and if they are defined the module can be run from the command line.

Don't know where the first quote comes from, but I agree with @lihaoyi here in that test modules should always provide a run environment. This is a consequence of `TestModule` being a `RunModule`.

lihaoyi:
> What is the rationale for adding allForkEnv? Is there an issue with overriding forkEnv?

> It is a bit misleading. We have various all-prefixed tasks in Mill, but they typically don't introduce new data but just aggregate other tasks. Here, we use the all prefix to provide some new stuff. Why can't we provide the env variables by just overriding forEnv?

The basic goal is the same as the other `all` tasks: to avoid people accidentally overriding `forkEnv` and forgetting to call `super`, which would result in confusing misbehaviors when the "builtin" environment variables are missing. This is similar to `ivyDeps` and `mandatoryIvyDeps` being combined into `allIvyDeps`, but without a `mandatoryForkEnv` since it's just included directly in `allForkEnv`

alexarchambault:
> > No run env for test modules, no test env for run modules
> 
> @alexarchambault Is this correct? Right now Mill `JavaTests` or `ScalaTests` modules all can have `public static void main(String[] args)` methods , and if they are defined the module can be run from the command line.

It feels there was some duplication. `jvmRunEnvironment` and `jvmTestEnvironment` were both returning the same thing. With this commit, we only return actual main classes in `jvmRunEnvironment`, and commands to run tests in `jvmTestEnvironment`.

Do we support the scenario you're describing (main methods in test modules) already in BSP? It seems some more tweaks are needed for that. I could add a test case, and have a look.

alexarchambault:
> @lihaoyi
> 
> > > No run env for test modules, no test env for run modules
> > 
> > 
> > @alexarchambault Is this correct? Right now Mill `JavaTests` or `ScalaTests` modules all can have `public static void main(String[] args)` methods , and if they are defined the module can be run from the command line.
> 
> Don't know where the first quote comes from, but I agree with @lihaoyi here in that test modules should always provide a run environment. This is a consequence of `TestModule` being a `RunModule`.

My changes don't split modules between RunModules on the one hand, and test ones on the other. For the `jvmRunEnvironment` request, it keeps `RunModule`s (which should include `TestModule`s), using discovered main classes. For `jvmTestEnvironment`, it looks for `TestModule`s, and uses their test environment.

But the code [around here](https://github.com/com-lihaoyi/mill/blob/7b634aef10a30a8b690fd686c53b138c902b614c/libs/scalalib/src/mill/scalalib/bsp/BspRunModule.scala#L47-L58) currently prevents `TestModule`s from being seen as pure `RunModule`s (by matching them as `TestModule` first, and retaining only their test environment), and using their discovered main classes. That might need fixing (and a test case).

alexarchambault:
I pushed my commits, alongside a fix for (what I think is) an issue and that I discussed above, in https://github.com/com-lihaoyi/mill/pull/5287

alexarchambault:
The additional snapshot data build-targets-jvm-test-environments.json embeds parts of the Mill class path, whose versions should change quite often. https://github.com/com-lihaoyi/mill/pull/5287 addresses that. Maybe it should be merged before the PR here (I can take care of pushing a merge commit to disentangle things here if needed).

lihaoyi:
@alexarchambault go it, I'll merge this first and we can do the other PR as a follow up

integration/ide/bsp-server/resources/snapshots/build-targets-jvm-test-environments.json
```diff
@@ -18,8 +18,8 @@
         "file:///workspace/app/test/compile-resources",
         "file:///workspace/app/test/resources",
         "file:///workspace/out/app/test/compile.dest/classes",
-        "file:///coursier-cache/https/repo1.maven.org/maven2/org/scala-sbt/test-interface/1.0/test-interface-1.0.jar",
-        "file:///mill-workspace/out/libs/testrunner/entrypoint/compile.dest/classes"
+        "file:///Users/lihaoyi/.ivy2/local/com.lihaoyi/mill-libs-testrunner-entrypoint/SNAPSHOT/jars/mill-libs-testrunner-entrypoint.jar",
```
lefou: This path doesn't look reproducible

integration/ide/bsp-server/src/BspServerTestUtil.scala
```diff
@@ -22,7 +22,7 @@ object BspServerTestUtil {
    * Set to true when running tests to update the snapshots on the disk. Do not forget to set it back to
    * false after that.
    */
-  val updateSnapshots = false
+  val updateSnapshots = true
```
lefou: This should not be `true` in the repo/CI, only in local dev.

integration/ide/bsp-server/resources/snapshots/build-targets-jvm-test-environments.json
lihaoyi: Should be fixed in the merged PR, I added a new `file://user-home` prefix substitution

integration/ide/bsp-server/src/BspServerTestUtil.scala
lihaoyi: should also be fixed in `main`

arturaz: Created https://github.com/com-lihaoyi/mill/pull/5292 to avoid this in the future.

PR #5274, opened 2025-06-09
Add optional deps to the class path of their modules (backport #5246)

alexarchambault:
Backport of https://github.com/com-lihaoyi/mill/pull/5246

---

Optional dependencies are marked as optional for consumers of a module.

They should still be added to the class path of the module that has them as optional dependencies, as Maven does, as mentioned by @lefou in https://github.com/com-lihaoyi/mill/issues/5095

Fixes https://github.com/com-lihaoyi/mill/issues/5095


lefou (commented):
Thank you, @alexarchambault !

Issue #5271, opened 2025-06-09
`bspBuildTargetCompileMerged` fails when module has no classes


arturaz:
If you have:
```scala
object foo extends JavaModule {
  object test extends JavaTests with TestModule.Junit4
}
```

And no classes in the `foo` module.

You'll get:
```
[12-compile] Entered buildTargetCompile
[12-compile] Evaluating 1 task
[12-compile] 1 tasks failed
[12-compile] foo.bspBuildTargetCompileMerged java.nio.file.NoSuchFileException: /home/arturaz/work/mill-example-javalib-module-7-resources/out/foo/compile.dest/classes
[12-compile]     java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
[12-compile]     java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
[12-compile]     java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
[12-compile]     java.base/sun.nio.fs.UnixFileSystem.copy(UnixFileSystem.java:1005)
[12-compile]     java.base/sun.nio.fs.UnixFileSystemProvider.copy(UnixFileSystemProvider.java:300)
1 tasks failed
foo.bspBuildTargetCompileMerged java.nio.file.NoSuchFileException: /home/arturaz/work/mill-example-javalib-module-7-resources/out/foo/compile.dest/classes
    java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
    java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
    java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
    java.base/sun.nio.fs.UnixFileSystem.copy(UnixFileSystem.java:1005)
    java.base/sun.nio.fs.UnixFileSystemProvider.copy(UnixFileSystemProvider.java:300)
    java.base/java.nio.file.Files.copy(Files.java:1304)
    os.copy$.copyOne$1(FileOps.scala:208)
    os.copy$.apply(FileOps.scala:212)
    mill.scalalib.JavaModule.bspBuildTargetCompileMerged$$anonfun$1$$anonfun$2(JavaModule.scala:1411)
    mill.define.Task$Named.evaluate(Task.scala:342)
    mill.define.Task$Named.evaluate$(Task.scala:327)
    mill.define.Task$Computed.evaluate(Task.scala:353)
[...]

PR #5276, opened 2025-06-09
Fix: `bspBuildTargetCompileMerged` fails when module has no classes

arturaz:
If a module doesn't have any compilable sources or resources, the `bspBuildTargetCompileMerged` task would fail. Now it won't, because we check for the existence of the source directory

Fixes https://github.com/com-lihaoyi/mill/issues/5271.


testkit/src/mill/testkit/UnitTester.scala
```diff
@@ -27,17 +36,36 @@ object UnitTester {
...
+
+    /** Do nothing. */
+    case DoNothing
+
+    /** Reset the module dir to an empty directory. */
+    case Reset
+
+    /** Reset the module dir to an empty directory and copy the source tree from the given path. */
+    case ResetAndCopyFrom(path: os.Path)
+  }
```
lefou: Name and purpose don't really match. `DoNothing` still needs to point the `moduleDir` to somewhere, but where? Maybe `UsePwd`? `Reset` might be a `UseEmpty` (or `CreateEmpty`) and `ResetAndCopyFrom` might be `UseTemplate` (or `CreateFromTemplate`).

testkit/src/mill/testkit/UnitTester.scala
```diff
@@ -27,17 +36,36 @@ object UnitTester {
...
+  enum SourceRoot {
+
+    /** Do nothing. */
+    case DoNothing
+
+    /** Reset the module dir to an empty directory. */
+    case Reset
+
+    /** Reset the module dir to an empty directory and copy the source tree from the given path. */
+    case ResetAndCopyFrom(path: os.Path)
```
lihaoyi: Do we really need this? It seems to me like we can just stick with `Reset` and do whatever copying we need to in the body of the integration test

testkit/src/mill/testkit/UnitTester.scala
arturaz: I think it matches exactly. If `DoNothing` is selected, nothing happens with `moduleDir`.

testkit/src/mill/testkit/UnitTester.scala
arturaz: What do you mean? I just noticed that there were 4 cases (sourceDir = path vs null and reset = true and false), one of which silently failed (sourceDir = path && reset = false). I just changed it to an enum which doesn't have this capability to fail.

lihaoyi: I think let's just leave the original 4 cases in place for now, and keep this PR to solving the `bspBuildTargetCompileMerged` failure

testkit/src/mill/testkit/UnitTester.scala
lefou: So, what does is mean? As an API user, wat means "do nothing"? `moduleDir == null` ?

Issue #5248, opened 2025-06-03
Add flag to globally disable restricted os operations


hoangmaihuy:
Can we have a flag similar to `--no-build-lock` to globally disable os restricted read/write operations?

I'm facing some issues when using Mill 1.0.0-RC1 in our build. It sometimes throw exception on writing to valid `Task.dest` but when we rerun the task again it passed. 

Please clone [mill-caliban-test](https://github.com/hoangmaihuy/mill-caliban-test) and run `./mill -j 1 __.compile` to reproduce the issue.

lihaoyi:
Looks like the problem is the ZIO fiber scheduling isn't cooperating well with our threadlocal variables. A workaround would be to wrap https://github.com/hoangmaihuy/mill-caliban/blob/86ce8fdee1f7ca998ed755b25ffc90641aba9a43/mill-caliban/src/io/github/hoangmaihuy/mill/caliban/CalibanSourceGenModule.scala#L30-L35 in a `BuildCtx.withFilesystemCheckerDisabled{...}` block. 

We should still have a global flag as an escape hatch, but this might be able to unblock you while I get that implemented

PR #5278, opened 2025-06-10
Add `--no-filesystem-checker` flag to globally disable filesystem read/write checks

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/5248

I kept all the `os.Checker`s in place and made them check a threadlocal to see whether to perform a check or not, rather than swapping out the `Checker` based on the flag. This is necessary to support cases like `BuildFileCls` where plumbing in the flag is not easy.

Updated the `os-checker` integration tests to cover this, and mentioned it in the sandbox example tests

### DIFF
--- a/example/depth/sandbox/1-task/build.mill
+++ b/example/depth/sandbox/1-task/build.mill
@@ -83,6 +83,12 @@ def bannedWriteTaskOverridden = Task {
 hello
 */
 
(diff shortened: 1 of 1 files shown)


lefou (commented):
Can you update `flags.adoc` to reflect the new option?

PR #5279, opened 2025-06-10
[WIP] Add bash completions

lihaoyi:


lihaoyi:
@lefou do you know how bash completions are normally distributed and installed? It seems all the instructions involve curling some script and hooking it into `~/.bash_profile` or `~/.zshrc`, which seems a bit sketchy, but I don't see any other alternative

lefou:
> @lefou do you know how bash completions are normally distributed and installed? It seems all the instructions involve curling some script and hooking it into `~/.bash_profile` or `~/.zshrc`, which seems a bit sketchy, but I don't see any other alternative

It differs for each package and distribution. In Gentoo and probably other distributions, a `/etc/bash-completion.d/ ` directory is scanned and each file found is automatically sourced, so you can place a complete-script there and it should work.

lefou:
I think we should also provide a way to manually install the complete scripts. E.g. don't write  anything outside of the project dir but produce something under `out/`. Looks to me the `tabcomplete/install` command is doing both in one go. 

A `tabcomplete/script` could just materialize the embedded script under its `.dest`, so users and distributors can do with it whatever they want.

lihaoyi:
Added an ability to `install --dest <path>` so you can specify where you want to put the completiobn script

website/docs/modules/ROOT/pages/cli/installation-ide.adoc
```diff
@@ -65,6 +65,25 @@ folder with one of the Mill example projects. There are a wide range of example
...
 
+=== Bash/Zsh Tab Completion
+
+You can set up Bash and Zsh tab completions by subsequently running the command:
+
+[source,console,subs="verbatim,attributes"]
+----
+> ./mill mill.tabcomplete/install
+----
+
```
lefou, line 76: Here we need a line explaining what `install` does:
* copy-ing complete-script to a specific location ~/.cache/mill/download
* modifying shell config files ~/.bash_profile and ~/.zshrc

website/docs/modules/ROOT/pages/cli/installation-ide.adoc
```diff
@@ -65,6 +65,25 @@ folder with one of the Mill example projects. There are a wide range of example
...
+After that, doing `./mill <tab>` should begin showing possible modules and tasks for you
+to select at the command line.
+
+You can also choose to write the tab-completion script to a manually-specified path via:
+
+[source,console,subs="verbatim,attributes"]
+----
+> ./mill mill.tabcomplete/install --dest <path>
```
lefou, line 85: We might need to say, that we don't modify shell config files if `--dest` is given.

lolgab:
Would be nice to support also fish shell. @ckipp01 had a package already: https://github.com/ckipp01/mill-fish-completions/blob/main/completions/mill.fish

PR #5282, opened 2025-06-10
Improve error message for discover parens checks

lihaoyi:
Print the name of the task method which is at fault, so it's easier for the user to figure out which method is being problematic


Quafadas:
wow. i think it took me longer to type it the description. Thankyou 🙏

PR #5287, opened 2025-06-10
Tweak BSP JVM run / test environment handling

alexarchambault:
This is a follow-up of some changes added in https://github.com/com-lihaoyi/mill/pull/5265. This further changes how we handle `buildTarget/jvmRunEnvironment` and `buildTarget/jvmTestEnvironment` requests, by making sure main classes in test modules are taken into account in `buildTarget/jvmRunEnvironment`, and by cleaning-up some test snapshot data of `buildTarget/jvmTestEnvironment`, so that they don't change too often and are more maintainable.


alexarchambault:
I should add sources with a standard main class in one of the added test modules. We should see it appear in the `buildTarget/jvmRunEnvironment` request snapshot.

PR #5290, opened 2025-06-11
Migrate `GenIdeaModule.intellijModulePath` to `intellijModulePathJava`

lefou:
This change enhances forward compatibility with Mill 1.

Deprecated `intellijModulePath`.


lefou:
Can't do it in a binary compatible way. Under the assuption that this setting will be tweaked rather seldomly, it's probably ok to not deprecate it before its removal in `1.0`.

PR #5292, opened 2025-06-11
Update BSP test snapshots via MILL_TESTS_BSP_UPDATE_SNAPSHOTS env var

arturaz:
The current way of changing the code and then changing it back is error prone, as witnessed in https://github.com/com-lihaoyi/mill/pull/5265#discussion_r2140429333 .

I changed it to use an environment variable, which has no chance of being committed to the codebase.

### DIFF
--- a/integration/package.mill
+++ b/integration/package.mill
@@ -39,7 +39,14 @@ object `package` extends mill.Module {
       def mode: String = moduleSegments.parts.last
       def scalaVersion = Deps.scalaVersion
 
-      def forkEnv =
+      def forkEnv = {
+        def exportFromEnv(key: String) = {
+          sys.env.get(key) match {
+            case Some(value) => Map(key -> value)
+            case None => Map.empty
+          }
+        }
+
         super.forkEnv() ++
           IntegrationTestModule.this.forkEnv() ++
           Map(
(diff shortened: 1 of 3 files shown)


lefou (commented):
Looks  reasonable

lefou:
Did you test this? I think without proper setup Mill will not pass the env variable though to the test.

arturaz:
> Did you test this? I think without proper setup Mill will not pass the env variable though to the test.

I did not, thought it should work out of the box. You are right, it seems that the variables aren't passed. :(

arturaz:
@lefou, now it works.

integration/package.mill
```diff
@@ -39,7 +39,14 @@ object `package` extends mill.Module {
       def mode: String = moduleSegments.parts.last
       def scalaVersion = Deps.scalaVersion
 
-      def forkEnv =
+      def forkEnv = {
+        def exportFromEnv(key: String) = {
+          sys.env.get(key) match {
```
lefou: `sys.env` may return stale values. Use `Task.env` (or `TaskCtx.env`) instead.

```suggestion
          Task.env.get(key) match {
```

lefou: Also, we still may suffer from stale cached values, since this isn't an input task.

lefou: I'll push a fix

lefou: done

PR #5304, opened 2025-06-12
ScalaJS API cleanup

lefou:
Use some Scala 3 features like enums and the `derives ReadWriter` notation, to make code simpler.

The remaining files are still convoluted and contain some undocumented patterns to make binary compatible evolution easier.


libs/scalajslib/src/mill/scalajslib/api/ModuleKind.scala
```diff
@@ -0,0 +1,9 @@
+package mill.scalajslib.api
+
+import upickle.default.ReadWriter
+
+enum ModuleKind derives ReadWriter {
+  case NoModule
+  case CommonJSModule
+  case ESModule
+}
```
lihaoyi, line 9: Damn this new `enum derives` syntax is really nice

PR #5305, opened 2025-06-12
Automatically reload BSP server upon meta-build change

alexarchambault:
This makes Mill send an `onBuildTargetDidChange` notification to BSP clients when the meta-build changes. In Metals, that should make it automatically reload the Mill build, with newer up-to-date module metadata (dependencies, …).

This is achieved by tweaking the way we start the Mill BSP server: rather than blocking until the BSP server exits in `MillMain0` (if we receive a `workspace/reload` or a `build/exit` notification), we get a `Future` completed upon those notifications. Then `MillMain0` watches both the build sources and this future, until it detects a change. When build sources changed, it suspends BSP requests processing, rebuilds the build, passes new evaluators to the Mill BSP server, and resumes BSP requests processing.


alexarchambault:
(Sorry for the force push, forgot there was an open PR for this branch…)

lihaoyi:
I assume this was manually tested in VSCode and IntelliJ?

runner/daemon/src/mill/daemon/MillBuildBootstrap.scala
```diff
@@ -76,15 +76,18 @@ class MillBuildBootstrap(
...
       error = runnerState.errorOpt,
       result = runnerState
     )
   }
 
-  def evaluateRec(depth: Int): RunnerState = {
-    // println(s"+evaluateRec($depth) " + recRoot(projectRoot, depth))
+  def evaluateRec(
+      depth: Int,
+      closeEvaluator: Boolean = true
```
lihaoyi: Does this need to be individually configured for every `evaluateRec` call? If not we should pass it in the constructor instead

testkit/src/mill/testkit/IntegrationTesterBase.scala
```diff
@@ -77,7 +77,9 @@ trait IntegrationTesterBase {
       } else {
         val serverPath0 = outDir / millNoDaemon
 
-        for (serverPath <- os.list.stream(serverPath0)) os.remove(serverPath / processId)
+        if (os.exists(serverPath0))
```
lihaoyi: Is there a way to avoid this conditional? Adding it makes this code very fragile, since if the server path changes or is misconfigured the code will fail silently and cause process leaks in our integration test suite causing other random failures

runner/daemon/src/mill/daemon/MillMain0.scala
```diff
@@ -301,25 +305,83 @@ object MillMain0 {
...
+                          case Some(Failure(ex)) =>
+                            streams.err.println("BSP server threw an exception, exiting")
+                            ex.printStackTrace(streams.err)
+                            errored = true
+                            keepGoing = false
+                          case Some(Success(BspServerResult.ReloadWorkspace)) =>
+                          // reload asked by client
+                          case Some(Success(BspServerResult.Shutdown)) =>
+                            // shutdown asked by client
+                            keepGoing = false
```
lihaoyi: Should we log something in the BSP logs in these cases, so anyone debugging something can see the server was reloaded?

core/define/src/mill/define/internal/WatchSig.scala
```diff
@@ -0,0 +1,22 @@
+package mill.define.internal
```
lihaoyi, line 1: Why does this stuff need to be moved to `mill.define`? AFAIK we only use it in the `runner` codebase which is not exposed to the `build.mill` classpath that `mill.define` is meant for

runner/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
```diff
@@ -60,25 +61,108 @@ private class MillBuildServer(
...
   // Set when a session is completed, either due to reload or shutdown
   private[worker] var sessionResult: Promise[BspServerResult] = Promise()
 
   private val requestCount = new AtomicInteger
 
   def initialized = sessionInfo != null
 
-  def updateEvaluator(evaluatorsOpt: Option[Seq[EvaluatorApi]]): Unit = {
-    baseLogger.debug(s"Updating Evaluator: $evaluatorsOpt")
+  def updateEvaluator(
```
lihaoyi, line 99: What's all this new code about? It's not clear to me from reading it

runner/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
```diff
@@ -60,25 +61,108 @@ private class MillBuildServer(
...
-        topLevelProjectRoot,
-        evaluators,
-        s => baseLogger.debug(s())
-      ))
+    val updatedEvaluators =
+      if (errored)
+        previousEvaluatorsOpt.map(_.evaluators) match {
+          case Some(previous) =>
+            evaluators.headOption match {
+              case None => // ???
```
lihaoyi: What are these `// ???` comments?

lihaoyi:
The concurrency model of how `MillMain0` interacts with `MillBuildServer` is not clear to me. Can you explain it in the PR description? Previously only one or the other was running at a time so that was trivial, but now we have both running and returning futures and watching things and re-triggering each other we need to make sure it's as simple as possible to avoid becoming a mess

alexarchambault:
> I assume this was manually tested in VSCode and IntelliJ?

Briefly, yes. It seems IntelliJ doesn't really care about those changes. Unless I'm mistaken, it loads the build upfront using BSP, but then doesn't emit BSP requests any more. The extra notification has no effect on it.

About Metals, it takes it into account: in VSCode, users get a dialog in the lower right corner, asking them if they want to reload the build or not.

(In practice, I'm working more with my own LSP server, which is private, but that I should open source at some point)

runner/daemon/src/mill/daemon/MillBuildBootstrap.scala
alexarchambault: My bad, this is code from former changes, that have been dropped. Just removed this parameter.

testkit/src/mill/testkit/IntegrationTesterBase.scala
alexarchambault: I think I added that because I must have run in this conditional not to be true at some point when working on this. I'm going to revert it, and see what happens.

runner/daemon/src/mill/daemon/MillMain0.scala
alexarchambault: Sure

core/define/src/mill/define/internal/WatchSig.scala
alexarchambault: I added this here because I need to use some of it from `MillBuildServer`. See the explanation about concurrency above for more details.

runner/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
alexarchambault: Just refactored it a bit, and added comments

runner/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
alexarchambault: Just removed them

alexarchambault:
> The concurrency model of how `MillMain0` interacts with `MillBuildServer` is not clear to me. Can you explain it in the PR description? Previously only one or the other was running at a time so that was trivial, but now we have both running and returning futures and watching things and re-triggering each other we need to make sure it's as simple as possible to avoid becoming a mess

Just updated the PR description. Hopefully, the concurrency model doesn't feel too messy. (It feels simple enough to me, but I might be biased here.)

lihaoyi (approved):

Issue #5221, opened 2025-05-29
Scaladoc links broken


lihaoyi:
https://mill-build.org/api/latest/mill/coursierutil/TestOverridesRepo.html links to https://github.com/com-lihaoyi/mill/blob/main/home/runner/work/mill/mill/core/coursierutil/src/TestOverridesRepo.scala, which is invalid

We're meant to pass in the HTTP URL to serve as the base for links to target, but somehow that config must have stopped working in the Scala 3.7 update.

arturaz:
Working on it.

PR #5308, opened 2025-06-13
Fix multiple issues with `UnidocModule`

arturaz:
Originally started because Scaladoc links to the `.scala` files were broken (https://github.com/com-lihaoyi/mill/issues/5221).

Fixed multiple issues along the way:
- `UnidocModule` did not support Scala 3. The appropriate code was in mill's build, moved that code to `UnidocModule`.
- `UnidocModule` did not collect transitive dependencies, all modules had to be listed. Now it does that by default and you can still override `unidocModuleDeps` to bet the old behaviour back.
- The source links option changed it's semantics from Scala 2 to Scala 3, updated it to work properly.
- Added tests for this functionality.

Breaking API changes:
- `unidocDocumentTitle` had a default of "Mill", now we force the user to specify it.


lihaoyi:
@arturaz seems like some tests need to be updated

lihaoyi:
bump @arturaz

PR #5310, opened 2025-06-13
Don't fail fast when evaluating tasks in BSP server

alexarchambault:
This enables keep-going mode when evaluating tasks in the BSP server. When running tasks to answer BSP requests, we want to run as many tasks as possible, to return as detailed as possible responses. Without this flag, a single module that doesn't compile could make BSP responses much more sparse than they should have been. Plus this might have been made worse by #4724, which grouped tasks more aggressively.

### DIFF
--- a/integration/ide/bsp-server/src/BspServerTests.scala
+++ b/integration/ide/bsp-server/src/BspServerTests.scala
@@ -188,6 +188,11 @@ object BspServerTests extends UtestIntegrationTestSuite {
               new b.CompileParams(
                 targetIds
                   .asScala
+                  // No need to attempt to compile the failing targets.
+                  // The snapshot data for this request basically only contains
+                  // a global status, errored or success. By excluding these,
+                  // we get a success status, and ensure compilation succeeds
+                  // as expected for all other targets.
                   .filter(!_.getUri.endsWith("/errored/exception"))
                   .filter(!_.getUri.endsWith("/errored/compilation-error"))
                   .filter(!_.getUri.endsWith("/delayed"))


integration/ide/bsp-server/src/BspServerTests.scala
```diff
@@ -190,7 +183,18 @@ object BspServerTests extends UtestIntegrationTestSuite {
...
         compareWithGsonSnapshot(
-          buildServer.buildTargetCompile(new b.CompileParams(targetIds)).get(),
+          buildServer
+            .buildTargetCompile(
+              new b.CompileParams(
+                targetIds
+                  .asScala
+                  .filter(!_.getUri.endsWith("/errored/exception"))
+                  .filter(!_.getUri.endsWith("/errored/compilation-error"))
+                  .filter(!_.getUri.endsWith("/delayed"))
```
lihaoyi, line 198: What's this?

alexarchambault: It's a way to drop the targets (modules) whose compilation fails on purpose. The response of compilation requests (that is snapshotted in `build-targets-compile.json`) is only a global value, that states if the whole compilation was successful or not. Removing these targets makes the check ensure that compilations expected to succeed to succeed.

alexarchambault: Maybe I could add a comment for that

lihaoyi: I still don't really understand. Is this a test fixture specifically for integration tests? Or is it for usage in production as well?

lihaoyi: If it's an integration test fixture, it should have the word "test" somewhere in the code or comment

alexarchambault: I mean, this lives in the tests, that's not part of the Mill code that users run

lihaoyi (approved):
shipit

PR #5311, opened 2025-06-13
Add tests for semanticdb generation in BSP server and make them work for build.mill

alexarchambault:
This adds non-regression tests for semanticdb generation in the BSP server, by enabling the generation of semanticdb in the BSP integration tests. This also make sure these are generated for generated sources in the build (like `build.mill`), by creating the semanticdb of original sources (like `build.mill`) from the semanticdb of generated ones, like Scala CLI does.

This also tweaks related things: this ensure scala-specific BSP requests only return results for Scala modules, and restricts job parallelism in the BSP integration tests to one job at a time, so that the BSP logs, which are validated, are more reproducible. This is necessary when enabling semanticdb generation, as compilation happens twice in that case (once for classes, and a second time for semanticdbs), and their output might be mixed without this.


alexarchambault:
Includes https://github.com/com-lihaoyi/mill/pull/5310 for now

runner/bsp/worker/src/mill/bsp/worker/MillScalaBuildServer.scala
```diff
@@ -27,7 +27,7 @@ private trait MillScalaBuildServer extends ScalaBuildServer { this: MillBuildSer
       targetIds = _ => p.getTargets.asScala.toSeq,
       tasks = {
         // We ignore all non-JavaModule
-        case m: JavaModuleApi =>
+        case m: ScalaModuleApi =>
```
lefou, line 29: This just makes the comment above a lie. Either update or delete it.

runner/daemon/src/mill/daemon/MillMain0.scala
```diff
@@ -247,7 +247,9 @@ object MillMain0 {
                             new MillBuildBootstrap(
                               projectRoot = BuildCtx.workspaceRoot,
                               output = out,
-                              keepGoing = config.keepGoing.value,
+                              // In BSP server, we want to evaluate as many tasks as possible,
+                              // in order to give as many results as available in BSP responses
+                              keepGoing = bspMode || config.keepGoing.value,
```
lefou: While I agree this may make sense, we should decide what the best way is to handle defaults. We currently set many BSP specifics via CLI args (e.g. we set `--disable-ticker` and `--color false` in `mill.bsp.BSP/install`). Here we implicitly change defaults for `keepGoing` which means a `--keep-going` is having no effect. We should either handle all defaults via CLI args or as derived default values, but not a mix of both.

alexarchambault: I agree. In case-app, there is a way to define flags as `Option[Boolean]`. Then users can specify either `--flag` or `--flag=true` to get `Some(true)`, or `--flag=false` to get `Some(false)`. That way, we can decide to compute the default value when it's `None` depending on the other options. I wish there was something similar in mainargs (there isn't if I'm not mistaken).

alexarchambault: Should `keepGoing`'s type be changed to `Option[Boolean]` here?

lihaoyi: `mainargs` supports `Option[Boolean]` as well, but that would make it require `--keep-going=true` or `--keep-going=false`. It doesn't support google-style `--keep-going` and `--nokeep-going` for `mainargs.Flag`s

I think for the purposes of fixing the problem adding it hard-coded is fine. We do need to clean up `MillMain0.scala`, but that's not something we need to bundle into this PR

lihaoyi (approved):

Issue #5314, opened 2025-06-14
Move `Task.workspace`/`TaskCtx.workspace` to `BuildCtx.workspace`


lihaoyi:
The `workspace` is global across the entire build, and does not vary task-by-task or module-by-module, so we should move it to the appropriate `*Ctx` that matches its scope

PR #5322, opened 2025-06-15
Replace `Task.workspace`/`TaskCtx.workspace` with `BuildCtx.workspaceRoot`

lihaoyi:
resolves https://github.com/com-lihaoyi/mill/issues/5314


lihaoyi:
Turns out that `BuildCtx.workspaceRoot` does not work in unit tests because they do not have the `MILL_WORKSPACE_ROOT` variable set. Might need some more work to make it usable

Issue #5318, opened 2025-06-14
Simplify specification of custom JVM versions in JavaModule


lihaoyi:
Currently this involves defining a custom `object myJvmWorker extends JvmWorkerModule` with a custom `def jvmId` then wiring it up with `override def jvmWorker = ModuleRef(myJvmWorker)` on the `JavaModule`.

We should try and simplify it to just `def jvmId` on `JavaModule`, and Mill should be able to do the rest

PR #5327, opened 2025-06-15
Simplify configuration of java home 

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/5318

I moved the `jvmId`, `jvmIndexVersion` and `javaHome` onto `JavaModule` rather than having it on `jvmWorker`. That lets people just configure it directly via `def jvmId` rather than jumping through hoops defining a custom `object MyJvmWorker extends JvmWorkerModule { def jvmId }` and `def jvmWorker = ModuleRef(MyJvmWorker)`. Common `def jvmId` settings can be moved to shared `trait`s just like any other `def`

The assumption is that although we are resolving the JVM once-per-module and potentially-in-parallel, rather than just once like we did before, Coursier's will keep things reasonably snappy via caching and concurrency-safe via locking

CC @alexarchambault

### DIFF
--- a/example/javalib/publishing/7-native-image/build.mill
+++ b/example/javalib/publishing/7-native-image/build.mill
@@ -4,12 +4,9 @@ import mill._, javalib._
 import mill.define.ModuleRef
 
 object foo extends JavaModule with NativeImageModule {
-  def jvmWorker = ModuleRef(JvmWorkerGraalvm)
+  def jvmId = "graalvm-community:17.0.7"
 
   def nativeImageOptions = Seq("--no-fallback")
 }
 
-object JvmWorkerGraalvm extends JvmWorkerModule {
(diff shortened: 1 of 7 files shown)


lihaoyi:
@alexarchambault can we rely on Coursier here for caching and concurrency-safety and deduplication of JVM downloads? Or should we implement some of that within Mill before calling out to Coursier?

alexarchambault:
> @alexarchambault can we rely on Coursier here for caching and concurrency-safety and deduplication of JVM downloads? Or should we implement some of that within Mill before calling out to Coursier?

JVM management in coursier is supposed to be concurrency-safe. Downloads of the JVM index and JVM archive and unpacking of the JVM archive should all be concurrency-safe.

Issue #5335, opened 2025-06-16
Suggestion: use .jvmopts or convert to .mill-jvm-opts


ThijsBroersen:
After an awesome presentation of Haoyi on Mill, I decided to convert one of my projects. However, I was running into a compiler error which turned out to be caused by missing jvm options. I wrongly assumed Mill was using the same .jvmopts after running `./mill init` on my SBT project.
This got me thinking: perhaps a small disclaimer/instruction when running init, inclusion of .jvmopts (perhaps as fallback) or conversion to .mill-jvm-opts would make migration a bit more smooth?

lefou:
If both files share the same syntax (one line per arg), we could automatically convert the `.jvmopts` to `.mill-jvm-opts`.

PR #5339, opened 2025-06-16
init: copy `.jvmopts` to `.mill-jvm-opts` when converting a sbt project

lefou:
Fix https://github.com/com-lihaoyi/mill/issues/5335

### DIFF
--- a/integration/migrating/init/src/MillInitSbtTests.scala
+++ b/integration/migrating/init/src/MillInitSbtTests.scala
@@ -56,7 +56,7 @@ object MillInitScala3ExampleProjectWithJvmOptsTests extends BuildGenTestSuite {
       "https://github.com/scala/scala3-example-project/archive/853808c50601e88edaa7272bcfb887b96be0e22a.zip"
 
     test - integrationTest(url)(it =>
-      os.write(it.workspacePath / ".jvmopts", "-Ddummy=prop")
+      os.write(it.workspacePath / ".jvmopts", "-Ddummy=prop -Ddummy2=prop2")
       testMillInit(
         it,
         expectedAllSourceFileNums = Map("allSourceFiles" -> 13, "test.allSourceFiles" -> 1),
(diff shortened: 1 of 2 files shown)


lihaoyi:
Are they the same format? The Mill file is an odd line-separated format because i was too lazy to properly toenize multiple arguments on a single line, not sure what the SBT file format is

lefou:
> Are they the same format? The Mill file is an odd line-separated format because i was too lazy to properly toenize multiple arguments on a single line, not sure what the SBT file format is


Yes, same format. But we currently don't filter out comments.

https://www.scala-sbt.org/1.x/docs/Command-Line-Reference.html  and an example: https://www.scala-sbt.org/1.x/docs/GitHub-Actions-with-sbt.html#Custom+JVM+options

lefou:
> > Are they the same format? The Mill file is an odd line-separated format because i was too lazy to properly toenize multiple arguments on a single line, not sure what the SBT file format is
> 
> Yes, same format. But we currently don't filter out comments.
> 
> https://www.scala-sbt.org/1.x/docs/Command-Line-Reference.html and an example: https://www.scala-sbt.org/1.x/docs/GitHub-Actions-with-sbt.html#Custom+JVM+options

I probably jumped to conclusion too fast... Let me ask the sources...

lefou:
Here is the parser code: 

https://github.com/sbt/sbt/blob/2db4b1af0f381c7f915e726054b2ecc43f00742b/sbt#L736

```shell
loadConfigFile() {
  # Make sure the last line is read even if it doesn't have a terminating \n
  cat "$1" | sed $'/^\#/d;s/\r$//' | while read -r line || [[ -n "$line" ]]; do
    eval echo $line
  done
}
```

Called from here: https://github.com/sbt/sbt/blob/2db4b1af0f381c7f915e726054b2ecc43f00742b/sbt#L831

```shell
[[ -f ".jvmopts" ]] && export JAVA_OPTS="$JAVA_OPTS $(loadConfigFile .jvmopts)"
```

lefou:
So, they allow multi-line and that's what I remembered. But the code also allows multiple args on a single line, so there is a change it will not work.

lihaoyi:
If SBT supports multiple tokens per line, it's probably worth trying to follow that in Mill as well. Shouldn't be too hard to come up with a shell lexer using Fastparse that understands `'`, `"`, `\`, and `#`

ThijsBroersen:
Awesome work!

lefou:
TBH, I'd rather stick to the simple format we currently have. I'd like to support comments, though.

We could still add that parser for the sbt-Mill converter, or just detect spaces and ask the user to review the file.

lihaoyi:
Sounds good to me

lefou:
Oh, we already support comments.

PR #5341, opened 2025-06-16
Android compose samples: Jet Lagged

vaslabs:
## Summary

This PR builds on top of https://github.com/com-lihaoyi/mill/pull/5285 and https://github.com/com-lihaoyi/mill/pull/5285 , and extends the work by creating a new linking android resources method that triggers the merging algorithm (with flat files instead of zips and auto overlay, with -R to add each library resource)

Screenshot From 2025-06-16 18-24-28

## Android Dependency Resolution changes

- New configuration parameters were needed to include "common" dependencies otherwise some of them were failing to be resolved for android

##  Android Resources

This is the major change that allows for [JetLagged](https://github.com/android/compose-samples/tree/main/JetLagged) from compose samples to work (only to run the App for now). The key change was to overlay the flat files of the libraries (with `-R` parameter) instead of linking them together as command inputs (without -R)

This allows for any conflicts of values that are the same to be resolved and it took some trial and error (I have a few versions of compose sample branches with a lot of failing attempts).

Another change is to use the android resources from resolved run maven deps although I have a suspicion we'll come back to that later on, I don't think the dependency resolution and packaging work is finished.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -187,6 +187,11 @@ jobs:
             install-sbt: false
 
           - java-version: 17
+            millargs: "'example.thirdparty[android-compose-samples].local.daemon'"
+            install-android-sdk: true
+            install-sbt: false
+
+          - java-version: 17
             millargs: "'{example,integration}.migrating.__.local.daemon'"
             install-android-sdk: false
             install-sbt: true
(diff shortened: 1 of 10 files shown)


vaslabs:
this should be ready although more PRs should follow, I don't think the test failure is relevant

Issue #4938, opened 2025-04-15
Remove or repurpose `--silent` cli option


lefou:
It currently seems unused and is documented with:

> Make ivy logs during script import resolution go silent instead of printing

We now renamed `ivy` to `mvn` and don't use ivy to manage dependencies at all.

lihaoyi:
Yeah `--silent` hasn't really been respected. IMO we can probably get by with removing it, and relying on stderr/stdout redirection if anyone wants to silence the build tool

Issue #2158, opened 2022-11-29
ScalafmtModule does not respect --silent


kierendavies:
I have a module that extends `mill.scalalib.scalafmt.ScalafmtModule`. When my Coursier cache is empty, and I run `mill --silent __.checkFormat`, I see many lines of output like
```
Downloading https://repo1.maven.org/maven2/org/scala-lang/scala-reflect/2.13.10/scala-reflect-2.13.10.pom
```

I would expect these lines to be suppressed when using the `--silent` flag, just like they usually are for other targets.

lefou:
These lines are probably produced by Scalafmt itself, which is downloading the required jars  based on the settings of your `.scalafmt.conf`.

Issue #1946, opened 2022-07-13
Scala.js targets `fullLinkJS` and `fastLinkJS` does not respect silent mode


carlosedp:
Running mill with silent and disable-ticker mode, still prints output to console (on success):

```
❯ mill -s --disable-ticker frontend.fullLinkJS
[info] compiling 1 Scala source to repos/scala-playground/zio-scalajs-stack/out/frontend/compile.dest/classes ...
[info] done compiling
Linker: Compute reachability: 419107 us
Linker: Assemble LinkedClasses: 24123 us
Linker: 474801 us
Optimizer: Batch mode: true
Optimizer: Incremental part: 151552 us
Optimizer: Optimizing 8124 methods.
Optimizer: Optimizer part: 689155 us
Optimizer: 850909 us
Refiner: Compute reachability: 79216 us
Refiner: Assemble LinkedClasses: 6955 us
Refiner: 97589 us
Module Splitter: Calculate Dependency Info: 2797 us
Module Splitter: Analyze Modules: 53800 us
Module Splitter: Assemble Modules: 9472 us
Module Splitter: 67405 us
Emitter: Generate Classes: 612803 us
Emitter: Write trees: 35800 us
Emitter: Class tree cache stats: reused: 0 -- invalidated: 744
Emitter: Method tree cache stats: reused: 0 -- invalidated: 4151
Emitter: 676614 us
BasicBackend: Write result: 307168 us
```

It's expected no output is generated from logs.

Use case: I have a user-defined task that prints out the generated files full-path that is sent to another process input. This issue impacts this since all output (printed above) is sent to the other process together with the path.

PR #5343, opened 2025-06-17
Clean up Mill CLI flags

lihaoyi:
I went and marked the various "internal" flags as `hidden = true`: things like `--bsp`, `--tab-complete`, and so on aren't really intended to be used directly by users, so no need to show them in the CLI `--help` message. Others like `--no-build-lock` or `--no-wait-for-build-lock` are kind of advanced features

All the internal and advanced flags no longer show in the normal `--help` output, but instead in the output of `--help-advanced` which is mentioned in the `--help` message so people can find it.

Deprecated flags are no longer shown in either `--help` or `--help-advanced`, as the assumption is that only legacy scripted callsites will hit these and anyone interactive should have no reason to use them:

- `--no-server` is replaced by `--no-daemon`
- `--enable-ticker` and `--disable-ticker` are replaced by `--ticker true/false`
- `--home` is unused
- `--repl` is not supported
- `--disable-callgraph` no longer works
- `--disable-prompt` no longer works
- `--silent` seems to have been unused for quite some time

Fixes https://github.com/com-lihaoyi/mill/issues/4938 and https://github.com/com-lihaoyi/mill/issues/2158 and https://github.com/com-lihaoyi/mill/issues/1946

### DIFF
--- a/runner/daemon/src/mill/daemon/MillCliConfig.scala
+++ b/runner/daemon/src/mill/daemon/MillCliConfig.scala
@@ -129,8 +129,18 @@ case class MillCliConfig(
     tabComplete: Flag = Flag(),
 
     // ==================== DEPRECATED CLI FLAGS ====================
+    @arg(hidden = true, short = 'h', doc = "Unsupported")
+    home: os.Path = os.home,
+    @arg(hidden = true, doc = "Unsupported")
+    repl: Flag = Flag(),
     @arg(hidden = true, doc = "Unsupported")
     noServer: Flag = Flag(),
+    @arg(short = 's', doc = "Unsupported")
+    silent: Flag = Flag(),
+    @arg(name = "disable-callgraph", doc = "Unsupported")
+    disableCallgraph: Flag = Flag(),
+    @arg(hidden = true, doc = "Unsupported")
+    disablePrompt: Flag = Flag(),
     @arg(hidden = true, doc = "Unsupported")
     enableTicker: Option[Boolean] = None,
     @arg(hidden = true, doc = "Unsupported")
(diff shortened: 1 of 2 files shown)


lefou:
There is a long history in CLI tools to try to support outdated CLI options for as long as possible, to avoid breaking older scripts (which typically don't have a means to specific version compatibility). E.g. in `mill-integrationtest` plugin, which tries to invoke various older Mill versions, it's really cumbersome to maintain a matrix of the best way to enforce a "no-server"-mode. 

If we keep those flags as `hidden` flags, which still make sense but are renamed, that would slightly improve the user experience, esp. when users have to switch between maintenance branches which host much older versions of a project and tools. Showing a deprecation or unused notice is fine.

lefou (commented):
Lets error out on `--repl` and `--diable-callgraph`. Both request an explicit behaviour we can no longer provide.

lihaoyi:
> Lets error out on `--repl` and `--diable-callgraph`. Both request an explicit behaviour we can no longer provide.

Most of those flags are no longer supportable though: `home`, `silent`,`disable-prompt`. Should we remove those too?

lefou:
> > Lets error out on `--repl` and `--diable-callgraph`. Both request an explicit behaviour we can no longer provide.
> 
> Most of those flags are no longer supportable though: `home`, `silent`,`disable-prompt`. Should we remove those too?

Don't know. It's a matter of taste. Those won't harm the user, so there is no real need to fail on them. Also, still being able to parse them avoids any accidental misinterpretation as task.

There is an essential difference between removing a flag and actively failing on one. In both cases, Mill will error out. But only in the latter case, we can provide a meaningful error message.

lefou:
Flags such as `--disable-prompt` have a high change to end up in some CI script and will cause unnecessary failure and debugging when removed. I myself and also coworkers of mine have lots of `justfile`s with Mill commands that need to work for multiple branches.

PR #5349, opened 2025-06-17
BSP diagnostics fixes / tweaking

alexarchambault:
This adds tests related to diagnostics in BSP integration tests, and fixes or adjusts several things related to diagnostics in BSP.

In particular, this ensures that a `CompileProblemReporter` is always passed around when running tasks for BSP. This ensures that users get all diagnostics related to compilation happening in BSP requests, not only diagnostics happening in the `buildTarget/compile` request. This includes compilation of the build itself, that happens upfront right after having started the BSP server. This allows users to get diagnostics for errors that make the build not compile for example (which can help address those errors).

This fixes things related to generated files. Some diagnostics notifications were not sent for those, which was making diagnostics stick around in VSCode, even when the errors were addressed. Line adjustment was incorrect for `build.mill` files having no `package` directive. This fixes that too.

There was also a hardcoded limit of maximum 10 errors being reported (both in BSP and in the console). This removes this limit for BSP, where a large amount of errors shouldn't be a problem, given they're reported inline in sources.

Lastly, the "Don't send diagnostics twice" commit ensures we don't send duplicate diagnostics via BSP. Ensuring a CompileProblemReporter is passed around everywhere (see above) can sometimes trigger that (when semanticdb generation is enabled, compilation happens twice: once for classes, a second time for semanticdbs, which would report each diagnostic twice).


core/api/src/mill/api/internal/EvaluatorApi.scala
```diff
@@ -7,6 +7,7 @@ trait EvaluatorApi extends AutoCloseable {
   def evaluate(
       scriptArgs: Seq[String],
       selectMode: SelectMode,
+      reporter: Int => Option[CompileProblemReporter] = _ => None,
```
lihaoyi, line 10: I wonder if we should make this a constructor param on `Evaluator` instead? We can e.g. mark every evaluator going into BspEvaluators with `.withCompileProblemReporter`, which would simplify the user-facing `Evaluator` API and ensure BSP code doesn't forget to pass in `reporter` at any use sites

lefou: Sound like we should try such a refactoring. But let's do this is a separate PR.

lihaoyi: the refactor does preserve the public interface of `Evaluator` and prevent it from becoming unnecessarily complicated, so it's actually user facing and not just of interest internally

lefou: Makes sense.

lefou: Is BSP currently using the daemon process? I think the evaluator is long-lived in the daemon, so ctr-args may not be suitable.

lihaoyi: BSP shares the same daemon process as the CLI does, so depending on whether or not the evaluators are shared this may or may not work. But it's definitely worth exploring

alexarchambault: Even if we pass the reporter builder directly to `Evaluator`, it seems we'll still need to pass the "origin id", that can differ for each BSP request, to individual methods. The origin id has to be added to the notifications that the compile problem reporter reports.

lihaoyi: Ok!

Issue #5229, opened 2025-05-30
Coursier default repositories break mill init


jdsalchow:
### Issue and reproduction
Coursier support setting default repositories by setting the [repositories.default](https://github.com/VirtusLab/scala-cli/blob/7a73dd18de041d32880d3d2ddd2597cf06d9d19f/modules/config/src/main/scala/scala/cli/config/Keys.scala#L129) key. For example by adding
```
{ "repositories": { "default": ["ivy2Local", "central"] } } 
```
to your scala-cli config (`~/Library/Application Support/ScalaCli/secrets/config.json` on a mac).

When explicitly specifying this, mill fails to resolve internal dependencies. For example, with the above config `mill init` fails in an empty directory with

```
mill init
[65/65, 2 failed] ============================== init ============================== 1s
2 tasks failed
resolvedRunIvyDeps
Resolution failed for 1 modules:
--------------------------------------------
  mill-internal::0+mill-internal
        not found: /Users/jds/.ivy2/local/mill-internal/0+mill-internal/ivys/ivy.xml
        not found: https://repo1.maven.org/maven2/mill-internal//0+mill-internal/-0+mill-internal.pom

--------------------------------------------

For additional information on library dependencies, see the docs at
https://mill-build.org/mill/Library_Dependencies.html

resolvedIvyDeps
Resolution failed for 1 modules:
--------------------------------------------
  mill-internal::0+mill-internal
        not found: /Users/jds/.ivy2/local/mill-internal/0+mill-internal/ivys/ivy.xml
        not found: https://repo1.maven.org/maven2/mill-internal//0+mill-internal/-0+mill-internal.pom

--------------------------------------------

For additional information on library dependencies, see the docs at
https://mill-build.org/mill/Library_Dependencies.html
```

Removing the default repositories from the config fixes the issue.

### Use-case
I have coursier configured locally to use a private repository by adding credentials and an extra default repository. That's a handy setup I'd like to re-use with mill.

### Extra information
I've tried reproduction above is with mill `0.12.11` and `0.12.14`. If I run `mill init` with `repositories.default` removed and add them back after, mill works "normally". I stumbled upon this trying to `mill __.compile` a project, getting the same behaviour, but the init reproduction seems to be the same issue and is easier to report.

jdsalchow:
@alexarchambault at first glance I don't quite understand the changes that you made in course. Does it imply that you understand the issue? What's missing for a fix?

alexarchambault:
[...]

Issue #5134, opened 2025-05-15
Properly handle coursier-supported env variables in Mill daemon


mberndt123:
Some settings in mill, like COURSIER_REPOSITORIES and COURSIER_CREDENTIALS, are specified as env variables. Env variables are read when a process starts, and later changes to the env variable are invisible to that process.
Since mill automatically starts a background process to avoid paying the startup overhead every time the user types `./mill`, users will be confused when they change these variables but they don't take effect. Hence, the mill background process should be restarted if the environment variables have changed since the last time the launcher script was invoked, and perhaps also if the launcher script itself was modified.

lefou:
Restarting the Mill daemon with every env var change is IMHO overkill. First, the assumption that those change infrequently might not hold for all users. But more importantly, Mill already captures env variable changes and has an API to read up-to-date env variables via `Task.env`/`TaskCtx.Env` API. Workers and other long running tasks need to capture those env variables via input tasks (thus making them proper dependencies), and Mill will automatically take care of them, e.g. by restarting the worker. This is properly documented: https://mill-build.org/mill/fundamentals/tasks.html#_environment_variable_inputs

The fact, that the embedded coursier implementation currently ignores dynamic env variable changes is IMHO a bug (or an absent feature), which we should fix.

lihaoyi:
Hardcoding support for coursier env vars seems fine for now, since those are by far the most common ones people will have trouble with

lefou:
Coursier falls back to read Java system properties, so we could forward the env vars to the respective system properties. It uses `coursier.repositories` and `cursier.credentials`. But since the env vars have higher priority, we have the issue that we must clear them before coursier reads them, otherwise the stale values env var values still overshadow the up-to-date system properties. But I don't think we can clear env vars programmatically.

We need some API to set these values then.

lihaoyi:
@alexarchambault could you take a look at this one? This seems like it should be straightforward for someone who is familiar with the coursier codebase and APIs

alexarchambault:
coursier currently reads a number of env vars and Java properties, that are used to build default values for a number of parameters (default repositories, TTL of snapshots and version listings in cache, …)

I guess reading those via `Task.Input`s ought to make Mill aware of their changes. I'm trying that.

lihaoyi:
Re-opening since the fix got reverted

PR #5350, opened 2025-06-17
Read coursier env vars via Task.Input

alexarchambault:
This makes Mill read all environment variables and Java properties used by coursier via a `Task.Input`.

In more detail, this adds a `CoursierConfigModule` external module in scalalib. It contains a single `Task.Input`, that reads all environment variables used by coursier via `Task.env` and all Java properties used by coursier via `sys.props`.

Helper tasks compute default values for various parameters (repositories, cache location, etc.). `CoursierConfigModule.coursierConfig` returns most of these parameters in a case class defined in Mill, `CoursierConfig`. For convenience, down-the-line, we pass `CoursierConfig` instances around.

Fixes https://github.com/com-lihaoyi/mill/issues/5134 (by managing coursier env vars / Java properties via `Task.Input`)

Fixes https://github.com/com-lihaoyi/mill/issues/5229 (via https://github.com/coursier/coursier/pull/3426 from coursier `2.1.25-M14` and the small rework of default repository handling)

### DIFF
--- a/website/docs/modules/ROOT/pages/cli/installation-ide.adoc
+++ b/website/docs/modules/ROOT/pages/cli/installation-ide.adoc
@@ -274,9 +274,6 @@ The below command should pass the environment variable to the `mill` command.
 > COURSIER_REPOSITORIES=https://packages.corp.com/artifactory/maven/ mill resolve _
 ----
 
-NOTE: Currently, you may need to `./mill shutdown` before calling Mill with the environment variable to make sure it gets properly propagated to Mill's background daemon.
-There is an https://github.com/com-lihaoyi/mill/issues/5134[open issue] to let Mill  detect these changes automatically.
-
 If you are using bootstrap script, a more permanent solution could be to set the environment variable
 at the top of the bootstrap script, or as a user environment variable.
 


example/scalalib/dependencies/5-repository-config/build.mill
```diff
@@ -15,7 +15,7 @@ object foo extends ScalaModule {
     mvn"com.lihaoyi::mainargs:0.6.2"
   )
 
-  def repositories = Seq("https://oss.sonatype.org/content/repositories/releases")
+  def repositories = Seq("https://oss.sonatype.org/content/repositories/releases", "default")
```
lihaoyi: Do we need to force people to add `default` here? It seems like most people will want the defaults to be available, and overriding and forgetting to add them would be a easy mistake for people to make

alexarchambault: Just changed that

core/util/src/mill/util/CoursierConfig.scala
```diff
@@ -0,0 +1,53 @@
...
+    archiveCacheLocation: String,
+    credentials: Seq[Credentials],
+    ttl: Option[Duration],
+    cachePolicies: Seq[CachePolicy]
+)
+
+object CoursierConfig {
+  // as much as possible, prefer using CoursierConfigModule.coursierConfig(),
+  // that reads up-to-date values from the environment and from Java properties
+  def default(): CoursierConfig =
```
lefou: In which situation do we need this directly? Maybe, making this `private[mill]` is an option?

libs/scalalib/src/mill/scalalib/CoursierConfigModule.scala
```diff
@@ -0,0 +1,143 @@
...
+import coursier.core.Repository
+import coursier.credentials.Credentials
+import coursier.params.Mirror
+import coursier.util.{EnvEntry, EnvValues}
+import mill.define.*
+import mill.{T, Task}
+
+import scala.concurrent.duration.Duration
+
+object CoursierConfigModule extends ExternalModule {
```
lefou: Can you add some ScalaDoc explaining which entries we support and maybe a little example how they can be cutsomized?

We get asked for this every now and then, and we could answer: Just use `mill inspet mill.scalalib.CoursierConfigModule/`

libs/scalalib/src/mill/scalalib/dependency/versions/VersionsFinder.scala
```diff
@@ -76,7 +76,8 @@ private[dependency] object VersionsFinder {
         coursierCacheCustomizer = cacheCustom,
         resolutionParams = coursier.params.ResolutionParams(),
         boms = Nil,
-        checkGradleModules = javaModule.checkGradleModules()
+        checkGradleModules = javaModule.checkGradleModules(),
+        config = CoursierConfigModule.coursierConfig()
```
lefou: I think we should have that as a class-level `coursierConfigModule: ModuleRef`.

lefou (commented):
The most tricky case with long running daemon is the case where the env variable is defined initially, but later gets removed from the env, in which case we need to fall back to the other defaults. Do we have this in a test case already?

libs/scalalib/src/mill/scalalib/CoursierConfigModule.scala
```diff
@@ -0,0 +1,180 @@
...
+    val (env, props) = coursierEnv()
+    CacheEnv.defaultCachePolicies(
+      CacheEnv.cachePolicy.readFrom(env, props)
+    )
+  }
+
+  /**
+   * Aggregates coursier parameters read from the environment
+   */
+  def coursierConfig: Task[mill.util.CoursierConfig] = Task.Anon {
```
lefou, line 168: Looks like a candidate for caching, to me.
```suggestion
  def coursierConfig: Task[mill.util.CoursierConfig] = Task {
```

libs/scalalib/src/mill/scalalib/dependency/versions/VersionsFinder.scala
```diff
@@ -13,7 +13,8 @@ private[dependency] object VersionsFinder {
   def findVersions(
       evaluator: Evaluator,
       ctx: TaskCtx,
-      rootModule: BaseModule
+      rootModule: BaseModule,
+      coursierConfigModule: ModuleRef[CoursierConfigModule]
```
lefou: A simple `coursierConfigModule: CoursierConfigModule` should be enough.

libs/scalalib/src/mill/scalalib/dependency/versions/VersionsFinder.scala
```diff
@@ -43,7 +44,12 @@ private[dependency] object VersionsFinder {
     def next(): Int = counter.getAndIncrement()
   }
 
-  private def classpath(progress: Progress, offline: Boolean, clock: Clock)(
+  private def classpath(
+      progress: Progress,
+      offline: Boolean,
+      clock: Clock,
+      coursierConfigModule: ModuleRef[CoursierConfigModule]
```
lefou: A simple `coursierConfigModule: CoursierConfigModule` should be enough.

libs/scalalib/src/mill/scalalib/dependency/DependencyUpdatesImpl.scala
```diff
@@ -14,13 +15,14 @@ object DependencyUpdatesImpl {
       evaluator: Evaluator,
       ctx: TaskCtx,
       rootModule: BaseModule,
+      coursierConfigModule: ModuleRef[CoursierConfigModule],
```
lefou: A simple `coursierConfigModule: CoursierConfigModule` should be enough.

lefou (commented):
Looks good. Found two other nits. 

Could you also clean `installation-ide.adoc` which references the issues you just fixed? https://github.com/com-lihaoyi/mill/blob/2ee59e4e5c374b465d67e23a2efe3e5520e1a7b7/website/docs/modules/ROOT/pages/cli/installation-ide.adoc?plain=1#L277-L278

lihaoyi:
I think this looks fine, once the last set of review comments have been resolve we can merge this

libs/scalalib/src/mill/scalalib/CoursierConfigModule.scala
alexarchambault: I've been trying that. Some `abstract class`es are involved (`coursier.core.Repository`, `coursier.params.Mirror`, `coursier.credentials.Credentials` in particular), which can be worked around, given this class is parsed from a bunch of strings, and the class instances are actually only simple easy-to-serialize classes.

But it seems this can also involve credentials (parsed from files on disk in particular). This raises security concerns: we don't adjust permissions of files where task results are written, so we could end up writing passwords in files with too loose permissions.

lefou: Let's leave it as-is then. Would be nice to have some support for secret caching (e.g. via encryption, or just in memory), but we currently don't.

Issue #5309, opened 2025-06-13
Make bootstrap scripts more flexible with `//| mill-version` key


lihaoyi:
Currently, the `//| mill-version: value` syntax has a lot of limitations:

- It must be on the first line of the build header
- `mill-version` must be unquoted
- `value` must be unquoted
- There can be no YAML comments or anything after the `value`

These limitations come from the necessity to parse this line in the `./mill` and `./mill.bat` bootstrap scripts, but they are very restrictive. While we cannot hope to parse the entirety of YAML in bash and bat code, we should try making the scripts a bit smarter to loosen these restrictions:

- Allow `//| mill-version` on any line of the build header
- Allow `mill-version` and `value` to be quoted by single or double quotes
- Allow YAML comments to be placed after the build header

arturaz:
> Allow YAML comments to be placed after the build header

Can you give an example for this?

Did you mean: `//| mill-version: 1.0.0-RC2 # this is a comment` ?

lihaoyi:
> > Allow YAML comments to be placed after the build header
> 
> Can you give an example for this?
> 
> Did you mean: `//| mill-version: 1.0.0-RC2 # this is a comment` ?

Something like `//| mill-version: 1.0.0 # yaml comment`

PR #5354, opened 2025-06-21
Make bootstrap scripts more flexible with `//| mill-version` key

arturaz:
Removes these limitations from the `//| mill-version: value` frontmatter:

- Allows `//| mill-version` on any line of the build header
- Allow `mill-version` and `value` to be quoted by single or double quotes
- Allows YAML comments to be placed after the build header: `//| mill-version: 1.0.0-RC2 # this is a comment`

It currently allows nonsensical quoting like `//| "mill-version': "1.0.0'`, but I figured it's not worth complicating the parser, given that it's written in sh/cmd scripts.

Fixes https://github.com/com-lihaoyi/mill/issues/5309


dist/scripts/test/src/MillVersionFrontmatterTests.scala
```diff
@@ -0,0 +1,88 @@
+package mill.scripts
+
+import utest.framework.TestPath
+import utest.*
+
+object MillVersionFrontmatterTests extends TestSuite {
```
lihaoyi, line 6: Let's test this using `MILL_TEST_DRY_RUN_LAUNCHER_SCRIPT=true`. That would make the tests run much faster than actually running Mill and bootstrapping a Mill daemon which is what the code does now, while still exercising the important part of the code which is the `mill.bat`/`mill.sh` parser logic

Issue #5097, opened 2025-05-09
Make `build.mill` Scala version configurable


lihaoyi:
Since https://github.com/com-lihaoyi/mill/pull/4879 the `build.mill` classpath is mostly isolated from the `mill.daemon` classpath, with one exception: they both share a dependency on the `scala-library`. This means that the version of Scala used in `build.mill` is pinned to the version used in `mill.daemon`, and `build.mill` is thus unable to use jars compiled against newer Scala versions, or make use of newer language features or library APIs in their `build.mill`.

But we can work around that: since https://github.com/com-lihaoyi/mill/pull/5025 `mill.daemon` is resolved from Maven Central by Coursier in `mill.launcher`. That means that we can ask Coursier to resolve artifacts with arbitrarily newer Scala versions, which can then be used as the pinned version shared between `mill.daemon` and `build.mill`. Since Scala is backwards compatible, such upgrades can continue to use jars compiled against earlier Scala versions, and will also allow users to use jars compiled against newer Scala versions, and make use of newer language features and library APIs in their `build.mill`

This doesn't need to go into 1.0.0, since it can be done without breaking compatibility.

lihaoyi:
probably going to call this a wontfix, in favor of pinning each mill version to a single scala version

PR #5366, opened 2025-06-25
Make `build.mill` Scala version configurable

arturaz:
Allows overriding the `scala-library` version used by `build.mill` via the `//| mill-scala-library-version: 3.7.2-RC1` configuration frontmatter.

Tested manually.

Implements https://github.com/com-lihaoyi/mill/issues/5097.

### DIFF
--- a/runner/launcher/src/mill/launcher/MillProcessLauncher.java
+++ b/runner/launcher/src/mill/launcher/MillProcessLauncher.java
@@ -118,11 +118,13 @@ public class MillProcessLauncher {
                 Object conf = mill.launcher.ConfigReader.readYaml(
                     buildFile, buildFile.getFileName().toString());
                 if (!(conf instanceof Map)) return new String[] {};
-                @SuppressWarnings("unchecked") Map<String, Object> conf2 = (Map<String, Object>) conf;
+                @SuppressWarnings("unchecked")
+                Map<String, Object> conf2 = (Map<String, Object>) conf;
 
                 if (!conf2.containsKey(key)) return new String[] {};
                 if (conf2.get(key) instanceof List) {
-                  @SuppressWarnings("unchecked") List<String> list = (List<String>) conf2.get(key);
+                  @SuppressWarnings("unchecked")
+                  List<String> list = (List<String>) conf2.get(key);
                   String[] arr = new String[list.size()];
                   for (int i = 0; i < arr.length; i++) {
                     arr[i] = mill.constants.Util.interpolateEnvVars(list.get(i), env);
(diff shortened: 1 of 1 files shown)


arturaz:
@lihaoyi I've tested this manually as there doesn't seem to be any testing infrastructure set up for other frontmatter flags?

lihaoyi:
@arturaz the integratiom and example tests cover frontmatter flags, this one could be an example test that we embed in the `Build Header Config` adoc page

lihaoyi:
Let's call it `mill-scala-version`? AFAIK it should affect both the library and the compiler

arturaz:
> @arturaz the integratiom and example tests cover frontmatter flags, this one could be an example test that we embed in the `Build Header Config` adoc page

I'll write an integration test then that:
- makes sure that the flag works
- makes sure that if you change it the build is rebuilt.

> Let's call it mill-scala-version? AFAIK it should affect both the library and the compiler

It does not seem so.

from `out/mill-resolve-runner`:
```
"mill:SNAPSHOT,scala-library-version:3.7.2-RC1"
"/home/arturaz/.ivy2/local/com.lihaoyi/mill-runner-daemon_3/SNAPSHOT/jars/mill-runner-daemon_3.jar"                                                                                        
"/home/arturaz/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala3-library_3/3.7.2-RC1/scala3-library_3-3.7.2-RC1.jar"                                                                                                                              
"/home/arturaz/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala3-compiler_3/3.7.1/scala3-compiler_3-3.7.1.jar"      
```

arturaz:
I guess we should make it `mill-scala-version` and change the compiler dependency and not the `scala-library` dependency?

lihaoyi:
It should change both I assume. Scala library and compiler versions generally have to be kept in sync AFAIK, otherwise weird bugs can happen

lefou:
I wonder what' the motivation for this change is. It might not be possible to just update the Scala version unless you're on a very older Mill version already. The compiler plugins Mill uses for it's buildfiles need to be release for newer Scala versions first, and since these plugins are maintained by Mill devs, it's very likely that they will also update the Scala version in Mill itself, once it is possible.

lihaoyi:
The motivation is to make it possible for users to bump the scala version even before Mill does. For example, when Scala 3.8.0 is released, users may want to use it with Mill even before a new version of Mill has been released that uses it.

Before this change, Mill users can upgrade any library in their classpath - upickle, oslib, mainargs - all except scala library and compiler which are pinned. This PR should make those two upgradable as wel

lefou:
> The motivation is to make it possible for users to bump the scala version even before Mill does. For example, when Scala 3.8.0 is released, users may want to use it with Mill even before a new version of Mill has been released that uses it.
> 
> Before this change, Mill users can upgrade any library in their classpath - upickle, oslib, mainargs - all except scala library and compiler which are pinned. This PR should make those two upgradable as wel

Which most likely means, the required compiler plugins are not yet released for these Scala versions. Making the Mill meta-build fail.

lihaoyi:
You are right that the compiler plugins are a problem. A solution would be to publish the same plugin version for new scala versions like we do for the Scala 2 compiler bridges. That would allow an older Mill version to use a newet Scala version without needing to upgrade Mill, or wait for a new Mill release that forces a Scala version upgrade

It does call into question how useful this is though. e.g. you wouldnt be able to try out Scala nightlies
or RCs without getting the plugins published first, which is one of the use cases I imagined.

lefou:
I [opened a discussion](https://github.com/com-lihaoyi/mill/discussions/5264) a while back about source-plugins. Following that idea, the meta-meta-build could ad-hoc compile the compiler-plugin(s) for a newer Scala version, if there is none available. Sounds a bit complicated and niche, but it might be possible.

lihaoyi:
closing as a wontfix with https://github.com/com-lihaoyi/mill/issues/5097

Issue #4421, opened 2025-01-28
Support to publish snapshots to Central Sonatype


chitralverma:
Since snapshots are now supported by central, can you please add support for publishing snapshots to sonatype central ?
https://central.sonatype.org/publish/publish-portal-snapshots/

chitralverma:
CC @lihaoyi

lefou:
What exactly are you missing to publish artifacts with a `-SNAPSHOT` version to Maven Central?

chitralverma:
@lefou  Im not exactly sure, but the [sonatype contrib extension shows that snapshots are ignored](https://github.com/com-lihaoyi/mill/blob/b69b695851ddcbdb2b6e40bade7fbc320c2d371c/contrib/sonatypecentral/src/mill/contrib/sonatypecentral/SonatypeCentralPublisher.scala#L45)

lefou:
According to [this page](https://central.sonatype.org/publish/publish-portal-snapshots/#enabling-snapshot-releases-for-your-namespace)., to support `-SNAPSHOT` releases, multiple steps need to be taken:

1. The publishing namespace must have snapshot publishing enabled. For that, navigate to the [namespace page](https://central.sonatype.com/publishing/namespaces) and activate snapshot publishing
2. A `mvn deploy`-compatible API is used. Whatever that means

For Mill that means we need to find out what a `mvn deploy`-compatible API means. Do we already have a publisher using this method, or do we need to implement one?

arturaz:
From looking at SBT sources, it seems that snapshot publishing does not go through the same Sonatype API.

From [SBT docs](https://www.scala-sbt.org/1.x/docs/Using-Sonatype.html#step+4%3A+Configure+build.sbt):
```scala
// new setting for the Central Portal
ThisBuild / publishTo := {
  val centralSnapshots = "https://central.sonatype.com/repository/maven-snapshots/"
  if (isSnapshot.value) Some("central-snapshots" at centralSnapshots)
  else localStaging.value
}
```

As I understand, `localStaging` [is a directory where artifact is put before uploaded to Sonatype](https://github.com/sbt/sbt/blob/f37aea07b2105d49cdab5c9b5e59a44664791c98/main/src/main/scala/sbt/Keys.scala#L572).

```scala
val localStaging = settingKey[Option[Resolver]]("Local staging resolver for Sonatype publishing").withRank(CSetting)
```

[...]

PR #5367, opened 2025-06-25
Add some logs & error reporting for (unsupported) publishing snapshots to Sonatype Central

Gedochao:
Some much missed logs and error handling for:
- https://github.com/com-lihaoyi/mill/pull/3130

Also relevant to:
- https://github.com/com-lihaoyi/mill/issues/4421
- https://github.com/VirtusLab/scala-cli/issues/3742

With this, we should:
- not have a green CI, when trying to publish snapshots only 🙃 
- not upload empty bundles/silently not upload anything, when publishing in such a case
- have any logs to follow up on


libs/scalalib/src/mill/scalalib/SonatypeCentralPublisher.scala
```diff
@@ -42,7 +42,26 @@ class SonatypeCentralPublisher(
...
+    val (snapshots, releases) = mappings.partition(_._1.isSnapshot)
+    if (snapshots.nonEmpty) {
+      val snapshotNames = snapshots.map(_._1)
+        .map { case Artifact(group, id, version) => s"$group:$id:$version" }
+      log.warn(
+        s"""Publishing snapshots to Sonatype Central Portal is currently not supported by Mill.
+           |This is tracked under https://github.com/com-lihaoyi/mill/issues/4421
+           |The following snapshots will not be published:
+           |  ${snapshotNames.mkString("\n  ")}""".stripMargin
+      )
```
lefou: Let's fail in this case, too. `Task.fail` is the convenient API to use here.

libs/scalalib/src/mill/scalalib/SonatypeCentralPublisher.scala
```diff
@@ -42,7 +42,26 @@ class SonatypeCentralPublisher(
...
+    }
+    if (releases.isEmpty) {
+      log.error("No releases to publish to Sonatype Central.")
+      val errorMessage =
+        "No releases to publish to Sonatype Central." +
+          (if (snapshots.nonEmpty)
+             "It seems there were only snapshots to publish, which is not supported by Mill, currently."
+           else "Please check your build configuration.")
+      throw new RuntimeException(errorMessage)
```
lefou: Let's use `Task.fail` here.

libs/scalalib/src/mill/scalalib/SonatypeCentralPublisher.scala
Gedochao:
```
Task APIs can only be used with a Task { ... } block
```
it doesn't seem I can use it here, though...

Gedochao: I mean, I guess I can wrap it all in a `Task.Anon`, if that's what you mean

Gedochao: done

libs/scalalib/src/mill/scalalib/SonatypeCentralPublisher.scala
Gedochao: done

libs/scalalib/src/mill/scalalib/SonatypeCentralPublisher.scala
lefou: Sorry, I was too quick and overlooked that we are not in a task context, so `Task.fail` is not available, and making the method a task is not the right thing, at least not without further changes.

Instead, we could `throw Result.Exception(msg)`, which should result in a proper task failure with message.

Gedochao: done

lefou:
We might want to backport this to 0.12.

Gedochao:
> We might want to backport this to 0.12.

@lefou Yes, please!

PR #5387, opened 2025-06-27
Privatization and cleanup  pass

lihaoyi:
https://github.com/com-lihaoyi/mill/issues/4716


libs/kotlinlib/src/mill/kotlinlib/KotlinMavenModule.scala
```diff
@@ -13,7 +13,7 @@ trait KotlinMavenModule extends KotlinModule with MavenModule {
   override def sources = super.sources() ++ sources0()
 
   trait KotlinMavenTests extends KotlinTests with MavenTests {
-    override def intellijModulePathJava: Path = (moduleDir / "src/test").toNIO
+    private[mill] override def intellijModulePathJava: Path = (moduleDir / "src/test").toNIO
```
lefou, line 16: This is meant to be overridable by the user. Otherwise, it's not possible to have multiple test sub-modules in one module.

PR #5392, opened 2025-06-27
Android: Replace hardcoded versions with relevant versions from build info

souvlakias:
- Extracted hardcoded values of each android module subclass into the build info component of androidlib

### DIFF
--- a/libs/androidlib/package.mill
+++ b/libs/androidlib/package.mill
@@ -35,10 +35,10 @@ object `package` extends MillPublishScalaModule with BuildInfo {
       "Version of NDK (Native Development Kit)"
     ),
     BuildInfo.Value("cmakeVersion", Deps.AndroidDeps.cmakeVersion, "Version of cmake"),
-    BuildInfo.Value("layoutLibVersion", Deps.AndroidDeps.cmakeVersion, "Version of LayoutLib"),
+    BuildInfo.Value("layoutLibVersion", Deps.AndroidDeps.layoutLibVersion, "Version of LayoutLib"),
     BuildInfo.Value(
       "composePreviewRendererVersion",
-      Deps.AndroidDeps.cmakeVersion,
+      Deps.AndroidDeps.composePreviewRendererVersion,
       "Version of ComposePreviewRenderer"
     ),
     BuildInfo.Value("uiToolingVersion", Deps.AndroidDeps.uiToolingVersion, "Version of uiTooling")


libs/androidlib/package.mill
```diff
@@ -1,27 +1,71 @@
...
-
   def transitiveLocalTestOverrides =
     super.transitiveLocalTestOverrides() ++ Seq(
       build.libs.kotlinlib.worker.localTestOverride()
     )
 
-  object hilt extends MillPublishScalaModule {
+  def buildInfoPackageName = "mill.androidlib"
+  def buildInfoObjectName = "Versions"
+  // Add versions?
```
vaslabs: is this a todo or a leftover?

souvlakias: leftover thanks. will remove it

libs/androidlib/package.mill
```diff
@@ -1,27 +1,71 @@
...
+    BuildInfo.Value(
+      "ndkVersion",
+      Deps.AndroidDeps.ndkVersion,
+      "Version of NDK (Native Development Kit)"
+    ),
+    BuildInfo.Value("cmakeVersion", Deps.AndroidDeps.cmakeVersion, "Version of cmake"),
+    BuildInfo.Value("layoutLibVersion", Deps.AndroidDeps.cmakeVersion, "Version of LayoutLib"),
+    BuildInfo.Value(
+      "composePreviewRendererVersion",
+      Deps.AndroidDeps.cmakeVersion,
```
vaslabs: wrong reference here

mill-build/src/millbuild/Deps.scala
```diff
@@ -280,4 +280,15 @@ object Deps {
...
+
+  object AndroidDeps {
+    val manifestMergerVersion = "31.10.0"
+    val bundleToolVersion = "1.17.2"
+    val ndkVersion = "27.0.12077973"
+    val cmakeVersion = "3.22.1"
+    val layoutLibVersion = "15.1.2"
+    val composePreviewRendererVersion = "0.0.1-alpha09"
+    val uiToolingVersion = "1.7.6"
+  }
```
lefou, line 292: What's missing here is to also use these dependencies somewhere, so we can auto-detect updates. Please add these to the `dummy` cross module in the top-level `build.mill`.

vaslabs: cool, we'll do, what happens for the tools that we can't match to maven deps?  (I think it's just cmake and ndk)

Is it better to have those in package.mill instead?

lefou: Currently, nothing. If you have an idea how to check them for updates, you can add a task that check it, of course.

I already started to add some deps to the dummy module. I'll send a review request.

lefou: Here's my draft:

* https://github.com/com-lihaoyi/mill/pull/5397


@vaslabs Don't know, if we need to accommodate to Android specifics. Feel free to comment or push to that PR branch.

vaslabs: no worries, @souvlakias is on it, he's also referencing the dependencies where appropriate (e.g. for layout lib and compose renderer)

vaslabs: we can do the android specifics later, for now let's go for the easy maven deps. We'll need to do some investigation for those as some of them depend on the android Sdk too.

lefou: If we need to customize the repos or other stuff, just create a second dummy module.

PR #5393, opened 2025-06-27
PMD integration & test

GingerYouth:
- Implemented PmdModule analogical to CheckstyleModule
- Added test with some PMD violations and simple ruleset

Resolves #5379

PR #5394, opened 2025-06-28
Added support for analyzing sources using PMD

### LINKED ISSUE #5379: PMD support (300USD Bounty)
---

**From the maintainer Li Haoyi**: I'm putting a 300USD bounty on this issue, payable by bank transfer on a merged PR implementing this.

---

### Discussed in https://github.com/com-lihaoyi/mill/discussions/5371

<div type='discussions-op-text'>

<sup>Originally posted by **GingerYouth** June 26, 2025</sup>
Hi, I'm new to Mill, have been using it recently for my Java project, and I'm wondering if it's possible to integrate PMD checks into Mill. 
For checkstyle I'm using `with CheckstyleModule`, and it would be really nice to have similar stuff for PMD. Do you have any plans regarding such thing? </div>

ajaychandran:
This PR adds support for analyzing sources using [PMD](https://docs.pmd-code.org/latest/index.html) via the `mill.javalib.PmdModule`.

- Custom rules/rulesets are supported.
- Companion external module included for global usage.
- Offline mode is supported.

All [PMD languages](https://docs.pmd-code.org/latest/tag_PmdCapableLanguage.html) are supported, theoretically, but test coverage is limited to Java.

- There are only 2 rules available for [Kotlin](https://docs.pmd-code.org/latest/pmd_rules_kotlin.html).
- There are no rules for [Scala](https://docs.pmd-code.org/latest/pmd_rules_scala.html).

Accordingly, the API has been marked experimental for now.

Resolves #5379.


libs/scalalib/src/mill/javalib/pmd/PmdModule.scala
```diff
@@ -0,0 +1,206 @@
...
+      - Java preview features require passing a runtime option.
+        See https://docs.pmd-code.org/latest/pmd_languages_java.html#using-java-preview-features.
+     */
+    defaultResolver().classpath(Seq(mvn"net.sourceforge.pmd:pmd-dist:${pmdVersion()}"))
+  }
+
+  /**
+   * PMD version to use. Defaults to ''7.15.0''.
+   */
+  def pmdVersion: Task[String] = Task { "7.15.0" }
```
lefou: We should not hardcode versions. Either leave it abstract or use a `BuildInfo` value that is managed at Mill-built-time, so we can keep it up-to-date as part of normal dependency maintenance.

libs/scalalib/src/mill/javalib/pmd/PmdModule.scala
```diff
@@ -0,0 +1,206 @@
...
+
+  /**
+   * Paths to [[https://docs.pmd-code.org/latest/pmd_userdocs_making_rulesets.html rulesets]]
+   * defining analysis rules to be executed. Defaults to XML files, starting with name ''ruleset'',
+   * under
+   *  - [[moduleDir]], if rulesets exist
+   *  - ''workspace'', otherwise
+   * @note Path values can be specified relative to [[moduleDir]].
+   */
+  def pmdRulesets: Task[Seq[String]] = Task {
```
lefou: This should not return `String`s if these are meant to be paths. This looks like it should be a `Task.Sources` with returns a `Seq[PathRef]`.

Since we also apply some filtering, I could image a two-step config, first the paths's where to look and second a concrete list of files, like we do in `JavaModule.sources` and `JavaModule.allSourceFiles`.

libs/scalalib/src/mill/javalib/pmd/PmdModule.scala
```diff
@@ -0,0 +1,206 @@
...
+        .orElse(rulesetsIn(Task.ctx().workspace))
+        .getOrElse(Task.fail("rulesets not found"))
+    }
+  }
+
+  /**
+   * Output [[https://docs.pmd-code.org/latest/pmd_userdocs_report_formats.html format]] of the
+   * analysis report. Defaults to ''text''.
+   */
+  def pmdFormat: Task[String] = Task("text")
```
lefou: This seems only of relevance for the generated output which is runtime-specific. Can this be a parameter of the `pmd` command instead?

ajaychandran: Nice tip! Updated example to demonstrate this.

libs/scalalib/src/mill/javalib/pmd/PmdModule.scala
ajaychandran: Updated both `pmdRulesets` and `pmdInputs`.

The former is a list of configuration files. The default value was selected to handle the two use-cases where rulesets are managed on a module basis or globally in the workspace module. It is not safe to use files at different levels, we don't know what the user intends in these cases.
 
The latter is the list of a combination of source files and/or folder roots with source files to be analyzed. In practice, this will always be `JavaModule.sources`.

Added  [--exclude <filepath+>](https://docs.pmd-code.org/latest/pmd_userdocs_cli_reference.html#file-collection-options) option for skipping paths. This in combination with `pmdInputs` should be sufficient for file selection.

libs/scalalib/src/mill/javalib/pmd/PmdModule.scala
ajaychandran: The positional values for inputs were replaced with an explicit option (`-d`). This eliminated the quirk and the `format` option.

ajaychandran:
@lefou This is ready for review.

libs/scalalib/src/mill/javalib/pmd/PmdModule.scala
lefou: This should be a `Task.Sources` to capture changes. This will likely eliminate the need for the `withFilesystemCheckerDisabled` tweak, too.

I don't get why this might end up with `JavaModule.sources` though, since it typically starts with `src` and not `ruleset` and you don't search recursively.

libs/scalalib/src/mill/javalib/pmd/PmdModule.scala
```diff
@@ -0,0 +1,196 @@
...
+  /**
+   * The specific [[https://docs.pmd-code.org/latest/pmd_userdocs_cli_reference.html#supported-languages language]]
+   * and version to use when parsing source code for a given language.
+   */
+  def pmdUseVersion: Task[Map[String, String]] = Task(Map.empty[String, String])
+
+  /**
+   * File to which output report is written. If empty, the report is rendered to standard output.
+   */
+  def pmdReportFile: Task[Option[PathRef]] = Task(Option.empty[PathRef])
```
lefou: What is this task for? If the `pmd` command needs to write to a file, it should be located under `pmd`'s dest folder.

libs/scalalib/src/mill/javalib/pmd/PmdModule.scala
```diff
@@ -0,0 +1,196 @@
...
+    case _ => Task {
+        BuildCtx.withFilesystemCheckerDisabled {
+          os.list.stream(moduleDir)
+            .collect:
+              case path
+                  if os.isDir(path) && path.relativeTo(moduleDir).segments.head.startsWith("src") =>
+                PathRef(path)
+            .toSeq
+        }
+      }
```
lefou: This should be a `Task.Sources` to account for changed files.

libs/scalalib/src/mill/javalib/pmd/PmdModule.scala
ajaychandran: Updated `pmdRulesets` to use `Task.Sources`.

Ignore the comment mentioning `JavaModule.sources`. It was meant for `pmdInputs`.

libs/scalalib/src/mill/javalib/pmd/PmdModule.scala
ajaychandran: This has been removed.

libs/scalalib/src/mill/javalib/pmd/PmdModule.scala
ajaychandran: Done.

ajaychandran:
This is mostly ready.

The idiomatic way to support [third party rulesets](https://docs.pmd-code.org/latest/pmd_userdocs_3rdpartyrulesets.html) would be to add a member (`pmdMvnDeps`) for corresponding Maven dependencies. I will work on this over the weekend.

lefou:
Meanwhile, we merged PR #5393 which added basic support for PMD. I think this PR adds valuable features, which we might want to merge, like out-of-the-box support for per-module configurations, language specific versions, PMD runner options and third-party rulesets.

ajaychandran:
Added support for defining and using custom rules/rulesets with `pmdMvnDeps` and `pmdModuleDeps`. Added an example for the latter.

The existing implementation from #5393 was removed and the PR description was updated.

libs/jvmlib/src/mill/javalib/pmd/PmdModule.scala
```diff
@@ -1,175 +1,165 @@
...
+   *  - ''pmd-dist'', for a [[BaseModule root module]]
+   *  - ''pmd-cli'' and ''pmd-java'', for a [[JavaModule]]
+   *  - ''pmd-cli'', otherwise
+   * @note [[https://docs.pmd-code.org/latest/pmd_userdocs_3rdpartyrulesets.html 3rd party rules]]
+   *       can be used after adding the dependencies here.
+   */
+  def pmdMvnDeps: Task[Seq[Dep]] = this match
+    case _: BaseModule => Task(Seq(mvn"$pmdDist"))
+    case _: JavaModule => Task(Seq(mvn"$pmdCli", mvn"$pmdJava"))
+    case _ => Task(Seq(mvn"$pmdCli"))
```
ajaychandran, line 92: Or this could be simplified to always use `pmdDist`.

ajaychandran:
Mima check is failing. Should I add exceptions?
```
[1496] Found 2 issue when checking against com.lihaoyi:mill-libs-jvmlib-api_3:VersionConstraint.Lazy(1.0.0-RC3-10-2e07f5, VersionInterval(None, None, false, false), Some(Version(1.0.0-RC3-10-2e07f5)), None)
[1496]  * static method pmdVersion()java.lang.String in class mill.jvmlib.api.Versions does not have a correspondent in current version
[1496]    filter with: ProblemFilter.exclude[DirectMissingMethodProblem]("mill.jvmlib.api.Versions.pmdVersion")
[1496]  * method pmdVersion()java.lang.String in object mill.jvmlib.api.Versions does not have a correspondent in current version
[1496]    filter with: ProblemFilter.exclude[DirectMissingMethodProblem]("mill.jvmlib.api.Versions.pmdVersion")
[1473] Binary compatibility check passed
[1473/1656, 1 failed] ======================= __.mimaReportBinaryIssues ====================== 21s
1 tasks failed
libs.jvmlib.api.mimaReportBinaryIssues Failed binary compatibility check! Found 2 potential problems
```

lihaoyi:
@ajaychandran yes just add an exclusion for now

ajaychandran:
CI failure seems unrelated.
```
[7065]     File "/home/runner/.cache/nce/ba7257887eb1da08c71c410700e65849b7716ee6237f3e86558d33802728b3ec/bindings/shiv_root/science.pyz_ee33d73d113bc985a5864e6f8cba0864a510cb1e41b1ee0166167c2dc244968c/site-packages/science/providers/python_build_standalone.py", line 204, in create
[7065]       release = release_data["name"]
[7065]                 ~~~~~~~~~~~~^^^^^^^^
[7065]   builtins.KeyError: 'name'
[7065]   1 tasks failed
[7065]   show 1 tasks failed
[7065]   qux.bundle os.SubprocessException: Result of /home/runner/work/mill/mill/out/example/extending/python/4-python-libs-bundle/local/daemon/testForked.dest/sandbox/run-2/out/qux/pythonExe.dest/venv/bin/python3?: 1
[7065]   
[7065]     os.SubprocessException$.apply(Model.scala:211)
[7065]       os.proc.call(ProcessOps.scala:232)
[7065]       os.call$.apply(ProcessOps.scala:42)
[7065]       build_.package_$PythonModule.bundle$$anonfun$1$$anonfun$2(build.mill:86)
```

lefou:
> The existing implementation from #5393 was removed and the PR description was updated.

That wasn't the intention of merging the other PR first.

If you replace something existing, it should come with a good motivation. E.g. you replaced the example test case with another one. Why this might be a better show case is unclear to me. Also, we settled with a default `pmd-ruleset.xml` file. This PR renames that to just `ruleset.xml`. Without further insights or discussion, this change isn't going to make it. Unless we're breaking some widely used best practices which you need to point out. I think the former name is much better as it clearly indicates which tools or part of the toolchain is going to use that file. You removed the `pmdVersion` task, which is clearly a step back for the user, who need to use a corporate-defined ruleset library for a different version. ...

Please try to keep stuff you don't touch stable (e.g. avoid reordering methods), otherwise the diff gets unnecessarily large and I can't spot the essential change/improvement.

ajaychandran:
Apologies. I did misread the intention.
I will submit another PR with the suggested improvements.

> I think this PR adds valuable features, which we might want to merge, like out-of-the-box support for per-module configurations, language specific versions, PMD runner options and third-party rulesets.

PR #5417, opened 2025-07-02
Use comma syntax for multiple inheritance in examples

lihaoyi:
Follow up of https://github.com/com-lihaoyi/mill/pull/5412, the new Scala 3 `,` syntax is more similar than the Scala 2 `with` syntax to the syntax used in Java or Python, and so we should use it where possible

One complication is that `extends` clauses can have the parent followed by either `with` or `,`
separators, but it needs to be consistent. So we need to try and see if
any separators are already present and if so follow suite. Not 100%
precise, but probably works well enough people will rarely hit issues


lihaoyi:
Turns out this may be a bit less trivial than I expected, due to interactions with the code generator that adds the additional `MillBuildRootModule` superclass to the parameter list using `with`

Issue #5208, opened 2025-05-27
Conside exclusive-mode by default for `RunModule.run`.


lefou:
`run` tasks are almost always tasks executing user code, for which `exclusive = true` is a better default. Especially, since customizing of the run-commands isn't trivial.

In those rare cases, where `run`-tasks are really meant to run build steps and can run non-exclusive, it should be rather easy to just create a dedicated non-exclusive task.

lefou:
This is blocked by the fact, that our BSP-integration can't invoke exclusive tasks.

PR #5424, opened 2025-07-02
Make most `RunModule.runXxx` commands exclusive

lefou:
Fix https://github.com/com-lihaoyi/mill/issues/5208

### DIFF
--- a/libs/jvmlib/src/mill/javalib/RunModule.scala
+++ b/libs/jvmlib/src/mill/javalib/RunModule.scala
@@ -110,7 +110,7 @@ trait RunModule extends WithJvmWorkerModule with RunModuleApi {
   /**
    * Runs this module's code in a subprocess and waits for it to finish
    */
-  def run(args: Task[Args] = Task.Anon(Args())): Task.Command[Unit] = Task.Command {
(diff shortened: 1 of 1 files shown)


lefou:
This change breaks BSP support, since when we try to invoke the `run` command from an annonymous task, we get the error message: 

```
Task.Anon@53839a78(mill.scalalib.bsp.BspRunModule.EmbeddableBspRunModule#internalBspRunModule.bspRun)
Non-exclusive task Task.Anon@53839a78(mill.scalalib.bsp.BspRunModule.EmbeddableBspRunModule#internalBspRunModule.bspRun) 
cannot depend on exclusive task app.run
```

lefou:
That also means, BSP will always break if users have `exclusive=true` `run`-tasks. Ideally, we handle the `exclusive` flag differently when running from BSP.

lihaoyi:
One option is to do what we did with `bspClean`, which is avoid depending on the exclusive `def clean` command and instead depend on a `def cleanTask` anonymous task. A similar refactor could be done for the `def run` commands to allow the anonymous task implementations to be used by BSP

lihaoyi:
Closing due to inactivity

PR #5425, opened 2025-07-02
Draft: support publishing snapshots to central sonatype

arturaz:
TODO: description

Current status: publishing works, needs tests and cleanup

Resolves https://github.com/com-lihaoyi/mill/issues/4421


core/util/src/mill/util/FileSetContents.scala
```diff
@@ -0,0 +1,117 @@
+package mill.util
+
+import os.{Path, SubPath}
+
+/**
+ * A bunch of relative file paths with their contents.
+ *
+ * @tparam Contents Allows to specify in the type system that we only have a certain type of contents.
+ */
+case class FileSetContents[+Contents](
```
lihaoyi: At a first glance, I don't think this abstraction is worth the complexity it brings. The previous `Seq[(os.Path, String)]`, while a bit low level, definitely seems easier to understand to me than trying to unpack the `FileSetContents` data structure and its generics, type-level logic, and bespoke utility methods.

Maybe we can pretty it up a bit by tightening the types and consolidating duplicate helper logic, but the `FileSetContents` you have here definitely goes too far in the other direction

arturaz: > definitely seems easier to understand to me

I beg to differ. Every time I encountered that signature I went "uhh, wait, what is what"? Also, it's a Seq, that means you can actually have correctness bugs, as in couple of destinations pointing to the same filename.

What exactly is hard with `FileSetContents`? The generic bound? From my perspective, it's using the most basic Scala generics, along with some variance declarations. This can be written in Kotlin or C# as well. Also, why use Scala if we're going not to use it's most basic features to make abstractions that aren't leaky?

lihaoyi: If you're worried about it being a `Seq` due to duplication and lack of uniqueness, we could make it a `Map` instead?

It's not about the language features used, just the cost of introducing custom collection data structures with their own custom operations and invariants is pretty high. It can be worth it in some cases where the need is clear and the semantics are obvious, but this does not seem one of those cases. if this was written in Kotlin or C#, I would similarly question whether a custom data structure is needed

lihaoyi:
Looking at the amount of refactoring here, I don't think I want to make such a wide refactor this late in the 1.0.0 RC release train, and I don't want to rush through such a refactoring since the whole point of refactoring is to take time and get it right. 

Could you split out the SNAPSHOT support into a separate PR from the refactoring cleanup? We can still perform the refactoring after 1.0.0 is released, we'd just need to leave deprecated forwarders in place that call your new method signatures, which isn't the end of the world.

core/util/src/mill/util/FileSetContents.scala
arturaz: > just the cost of introducing custom collection data structures with their own custom operations and invariants is pretty high

The data structure is essentially a `Map` where the key is fixed to a certain type. What is the cost here? Why aren't the semantics obvious?

lihaoyi: If it's really just a `Map` with a few helper methods, then it sounds like we can just use a `Map` with a few static helper methods.

I don't know why the semantics aren't obvious, but they aren't to me. The biggest question is "what does this give us over a `Map` with a few helper methods?" So it's on you to argue why it's necessary enough we can't do without, or we can just do without.

jackkoenig:
FWIW, it would also be ideal if this could be backported to 0.12 in which case the old public APIs are needed anyway (although I guess it could be a heavily modified backport if you did stick with the new types).

core/util/src/mill/util/FileSetContents.scala
arturaz: >  then it sounds like we can just use a Map with a few static helper methods.

So you propose:
```scala
object Foo {
  def doThing[A](map: Map[SubPath, A]): Int = ???
}
```

Or
```scala
def doThing[A](map: Map[SubPath, A]): Int = ???
```

Instead of:
```scala
case class MyThing[A](map: Map[SubPath, A]) {
  def doThing: Int = ???
}
```

?

lihaoyi: That's right! Basically `MyThing` should either be a dumb wrapper, or it should truly be its own data type with its own invariants and other things that it needs to encapsulate. As written, it's neither here nor there, which I think is why I find it very confusing.

arturaz:
> The removal of the bintray contrib module should be a separate PR with the motivation in the description.

@lefou https://github.com/com-lihaoyi/mill/pull/5467

arturaz:
Closing this as it has been split into parts and merged as separate PRs.

PR #5428, opened 2025-07-02
Revised build importer implementations

ajaychandran:
This PR rewrites importer implementations using a new architecture that is easy to maintain and extend.

## Motivation
While working on adding support for cross projects, the build models used in the existing implementations turned out to be insufficient for the use-case.
This prompted work on an alternate architecture that culminated in this pull request.

## Solution
New data models are introduced to define a build specification. This includes an ADT whose members have a one-to-one correspondence with module types defined within Mill. Importer implementations were updated to target the new models.

## New and dropped features
- Added support for converting cross-platform and cross version SBT projects.
- Added support for configuring `ErrorProneModule` for Gradle and Maven projects.
- Added support for configuring `jvmId`.
- Dropped support for mapping `resources`.
- Redundant command line options were removed.

## Other improvements
- Fixed mapping for `PomSettings` in Maven conversion.
- Fixed mapping for `artifactName` in SBT conversion.
- Publish is skipped for Maven modules where it is explicitly disabled. Example, `testsuite-osgi` module in [netty](https://github.com/netty/) project.
- Prevent runtime failure by handling Maven `ModelBuildingException`.

## Limitations
- As before, build customizations are not supported. This includes custom sources, custom Gradle configurations, and Maven profiles.
- SBT cross-platform support is limited to projects using `sbt-crossproject` plugin.

## Summary of code changes
The previous implementation has been completely removed. Source files for application entry points and some tests were retained, but the contents have changed.
- The new shared models are housed in the `buildgen.api` module. Together with new code in `buildgen` these account for ~1300 LOC.
- The new Gradle plugin is isolated in the `gradle.exportplugin` module. The Gradle importer implementation and unit tests account for ~430 and ~700 LOC, respectively.
- The Maven importer implementation and unit tests account for ~500 and ~1350 LOC, respectively.
- The SBT importer implementation and unit tests account for ~600 and ~2200 LOC, respectively.
- The integration tests and golden files account for ~150 and ~11500 LOC, respectively.
- The manual integration tests account for ~400 LOC.

## Testing
- Updated integration tests to use golden testing for checking settings.
- Added `integration.manual[migrating]` module to develop and test the conversions against real world projects.

Summary of fixes/enhancements made to improve support for test projects.
- Fixed `init` error in path handling for `spring-ai`.
- Fixed `init`/configuration error by removing custom resources for `byte-buddy`/`joda-beans`.
- Added `spring-boot-dependencies` BOM dependency for `microconfig`.
- Generated `ErrorPluginModule` configurations for `mockito`.
- Generated cross-platform/version configurations for `Airstream`, `cats`, `enumeratum`, `fs2`, `nscala-time`, `refined`, `scala-logging`, `scopt`, `scrypto`.

## Future work
- Add support for converting Kotlin/Android projects.
- Add support for converting SBT projects that use the `sbt-projectmatrix` plugin.


ajaychandran:
@lihaoyi As per the Maven POM [schema](https://maven.apache.org/xsd/maven-4.0.0.xsd), the `description`, `organization` and `url` fields are not mandatory. Is there a reason why these are required in Mill `PomSettings`?

lihaoyi:
No reason I'm aware of, we can make them optional

lefou:
I think the Maven Central validator rejects projects without these fields. Although it's specific to Maven Central, having these fields mandatory avoids some late pitfalls, when publishing artifacts. 

We only use `PomSettings` for publishing/sharing, so having some minimal standards on meta-data is probably not bad.

ajaychandran:
We could allow `null` values and handle in the render method. This would satisfy the empty use-case without requiring changes in user code.

ajaychandran:
A new SBT importer implementation has been added that can import projects that use `sbt-crossproject` plugin for cross platform development. The implementation also generates cross modules for cross Scala versions.

This was tested against 12 projects and all but 2 were successfully imported.

- `scala/scala3`: Multiple SBT projects use the same base directory.
- `scalapb/ScalaPB`: Uses `sbt-projectmatrix` for cross development.

One of the tests can be run using the following command:
```console
./mill standalone.migrate[sbt].testOnly mill.standalone.MillInitAirstreamTests
```

ajaychandran:
@lihaoyi 

I am going to be working on adding support for some more settings such as `jvmId` and generating base trait feature.

What are your thoughts on supporting more plugins? We already have tests that require the following:

- `sbt-projectmatrix`
- `sbt-play`
- `ScalablyTypedConverterGenSourcePlugin`

lihaoyi:
@ajaychandran update the PR title and description to summarize the motivation, major changes, how it integrates with the existing codebase, and testing strategy. It's a large PR and that would make it easier to review

libs/init/migrate/src/mill/init/migrate/PackageWriter.scala
```diff
@@ -0,0 +1,334 @@
...
+  def writeModuleConfigs(ps: PrintStream, configs: Seq[ModuleConfig]) = {
+    for config <- configs do
+      config match
+        case config: CoursierModuleConfig =>
+          import config.*
+          if (repositories.nonEmpty)
+            ps.println()
+            ps.print("def repositories = super.repositories() ++ ")
+            ps.print("Seq(")
+            ps.print(literalize(repositories.head))
```
lihaoyi: Please use string interpolation for this and all similar code snippets so it's not so verbose

standalone/testkit/src/mill/standalone/StandaloneTester.scala
```diff
@@ -0,0 +1,104 @@
+package mill.standalone
+
+import mill.constants.DaemonFiles.processId
+import mill.constants.OutFiles.{millDaemon, millNoDaemon, out}
+import os.{Path, ProcessInput, ProcessOutput, Shellable}
+
+class StandaloneTester(
```
lihaoyi: Why does this need to be different from normal IntegrationTester? We should make these normal integration tests using the normal `IntegrationTester` utilities

standalone/package.mill
```diff
@@ -0,0 +1,33 @@
+package build.standalone
```
lihaoyi: Let's put this in the `integration/manual/`, indicating that these are tests meant to be run manually but won't be run in CI, but are otherwise normal integration tests

libs/init/migrate/api/src/mill/init/migrate/ModuleConfig.scala
```diff
@@ -0,0 +1,108 @@
+package mill.init.migrate
```
lihaoyi: How does this file related to the code we already have in `libs/init/buildgen/src/mill/main/buildgen/ir.scala`? Does it replace it entirely, or does it re-use parts of the existing model?

libs/init/migrate/api/src/mill/init/migrate/ModuleTypedef.scala
```diff
@@ -0,0 +1,56 @@
...
+    supertypes
+  }
+
+  val testSupertypeByDepOrg = Map(
+    "junit" -> "TestModule.Junit4",
+    "org.junit.jupiter" -> "TestModule.Junit5",
+    "org.testng" -> "TestModule.TestNg",
+    "org.scalatest" -> "TestModule.ScalaTest",
+    "org.specs2" -> "TestModule.Specs2",
+    "com.lihaoyi" -> "TestModule.Utest",
```
lihaoyi: This seems a bit imprecise. For example a lot of people use `com.lihaoyi::os-lib` but don't use `utest`. Could we match on a whitelist of artifact names (or prefixes, ignoring the `_{scala-version}`) as well?

lihaoyi: Also, this seems to duplicate logic we already have in `BuildgenUtil`. We should consolidate it and avoid copy-pasting code

libs/init/migrate/api/src/mill/init/migrate/Tree.scala
```diff
@@ -0,0 +1,24 @@
+package mill.init.migrate
+
+import upickle.default.{ReadWriter, macroRW}
+
+case class Tree[+A](root: A, children: Seq[Tree[A]] = Nil) {
```
lihaoyi: How is this different from `libs/init/buildgen/src/mill/main/buildgen/Tree.scala`?

libs/init/migrate/sbt/src/mill/init/migrate/sbt/SbtPackageWriter.scala
```diff
@@ -0,0 +1,191 @@
...
+    if (platformCrossTypeBySegments.nonEmpty)
+      ps.println()
+      ps.println(renderSbtPlatformModule)
+      if (crossScalaVersionsBySegments.nonEmpty)
+        ps.println()
+        ps.println(renderCrossSbtPlatformModule)
+  }
+
+  def renderSbtPlatformModule =
+    s"""trait SbtPlatformModule extends SbtModule { outer =>
```
lihaoyi: How is this different from `SbtModule with PlatformScalaModule`? If it's necessary, we should put it (and `CrossSbtPlatformModule`) in `libs/scalalib/src/mill/scalalib/` rather than codegen-ing it every import

libs/init/migrate/sbt/exportscript/src/mill/init/migrate/sbt/ExportSbtBuildScript.scala
```diff
@@ -0,0 +1,225 @@
+package mill.init.migrate.sbt
```
lihaoyi: How does this relate to the existing `libs/init/sbt/exportplugin/src/mill/main/sbt/ExportBuildPlugin.scala`? Can we consolidate them?

libs/init/migrate/src/mill/init/migrate/PackageTree.scala
```diff
@@ -0,0 +1,90 @@
+package mill.init.migrate
+
+import mill.constants.CodeGenConstants.{nestedBuildFileNames, rootBuildFileNames}
+
+import java.io.PrintStream
+import scala.collection.mutable
+import scala.util.Using
+
+opaque type PackageTree = Tree[PackageRepr]
```
lihaoyi: Let's just use a normal `case class` here with normal methods, rather than this `opaque type` with `extension`s

lihaoyi:
@ajaychandran Overall this looks decent as a first pass, but what is missing is how this interacts with the existing `libs/init/sbt` implementation

- Does the new implementation pass all the same tests? I saw some tests failing on your fork's CI
- Does the new implementation replace the old one? If so can we delete the old one entirely?
- Can we share code between the old and new implementations? I see a bunch of duplicate logic that can be consolidated

lihaoyi:
> @lihaoyi
> 
> I am going to be working on adding support for some more settings such as jvmId and generating base trait feature.
> 
> What are your thoughts on supporting more plugins? We already have tests that require the following:
> 
> sbt-projectmatrix
> sbt-play
> ScalablyTypedConverterGenSourcePlugin


@ajaychandran Let's focus on getting this PR merged first before we start looking at follow ups, I think there's enough work here to keep you occupied for at least a few more days, and I don't want us to get distracted and end up leaving work half finished

libs/init/migrate/src/mill/init/migrate/PackageWriter.scala
```diff
@@ -0,0 +1,334 @@
+package mill.init.migrate
+
+import mill.constants.CodeGenConstants.rootModuleAlias
+import mill.internal.Util.backtickWrap
+import pprint.Util.literalize
+
+import java.io.PrintStream
+import scala.collection.immutable.SortedSet
+
+trait PackageWriter {
```
lihaoyi: We should re-use as much of `BuildGenUtil` as possible here, which already has logic for rendering `def mvnDeps`, `def moduleDeps`, etc. We should not need to re-implement it from scratch and end up maintaining two different implementations. We should only need to implement that parts that `BuildGenutil` does not already implement

libs/init/migrate/sbt/exportscript/src/mill/init/migrate/sbt/ExportSbtBuildScript.scala
```diff
@@ -0,0 +1,225 @@
...
+          import dep._
+          val bin = crossVersion match {
+            case _: librarymanagement.Full => ":::"
+            case _: librarymanagement.Binary => "::"
+            case _ => ":"
+          }
+          val platform = crossVersion match {
+            case v: librarymanagement.Full if v.prefix.nonEmpty => "::"
+            case v: librarymanagement.Binary if v.prefix.nonEmpty => "::"
+            case v: librarymanagement.For2_13Use3 if v.prefix.nonEmpty => "_3" + "::"
```
lihaoyi: Why not just `"_3::"`? Same for the `"_2.13" + "::"` below

standalone/testkit/src/mill/standalone/StandaloneTester.scala
ajaychandran: It doesn't have the same requirements as `IntegrationTester`. The workspace creation and initialization happens outside the tester.
Given that this PR proposal will take some time, I chose to isolate this to avoid any conflicts when re-basing.

libs/init/migrate/api/src/mill/init/migrate/ModuleConfig.scala
ajaychandran: It is meant to replace it.
In essence, it is a refactoring of the existing model that organizes settings into groups that correspond to actual module types.
Also, the IR types with the rendered code are not required and do not exist in the new model. This was primarily introduced, by me, for file merging. The new implementation achieves the same in a sane way.

libs/init/migrate/api/src/mill/init/migrate/ModuleTypedef.scala
ajaychandran: Yes, it is the same as the list in `BuildGenUtil`.
Given that this is a rewrite of the core design, I have kept the new implementation separate. If the proposal is approved, this can be consolidated.

libs/init/migrate/api/src/mill/init/migrate/Tree.scala
ajaychandran: It's in Scala 2 and the unused features are removed.
But basically it is the same and only exists because the new implementation is isolated.

libs/init/migrate/sbt/src/mill/init/migrate/sbt/SbtPackageWriter.scala
ajaychandran: `SbtModule` resets `sources`
```scala
  override def sources = Task.Sources("src/main/scala", "src/main/java")
```
and `PlatformModuleBase` operates on `sourceFolders`.
```scala
  override def sourcesFolders: Seq[os.SubPath] = super.sourcesFolders.flatMap {
    source => Seq(source, source / os.up / s"${source.last}-${platformCrossSuffix}")
  }
```
Are these compatible in any scenario?

This is being used to replicate the layout used by `sbt-crossproject` plugin. But now we also have `sbt-projectmatrix` for cross development which seems to use a completely different structure. Is it okay to add a trait to `scalalib` per plugin?

libs/init/migrate/sbt/exportscript/src/mill/init/migrate/sbt/ExportSbtBuildScript.scala
ajaychandran: Functionally they are the same.
But the script offers certain advantages such as not requiring any extra build files to add a plugin. The idea was borrowed from [IDEA](https://github.com/JetBrains/sbt-structure?#sideloading) :)

standalone/testkit/src/mill/standalone/StandaloneTester.scala
lihaoyi: What workspace creation and initialization do we expect with IntegrationTester? If run on an empty folder, the workspace starts empty

libs/init/migrate/src/mill/init/migrate/PackageTree.scala
ajaychandran: A `PackageTree` is supposed to have the following invariants:
- The root package is located at the workspace root.
- The tree structure represents the filesystem structure.
An `opaque type` seemed the "right" choice.
The extension methods can be defined as normal methods.

Could you provide the reason behind the comment? This is just for my learning.

lihaoyi: In general, we aim to use simpler language features over more powerful ones, unless there is a strong reason otherwise. In this case, there is no strong reason to use an `opaque type`, so we use `case class` which people are more familiar with

standalone/testkit/src/mill/standalone/StandaloneTester.scala
ajaychandran: `IntegrationTester` clears the workspace and copies resources. I suppose this can be circumvented with an empty resource folder.

I am not opposed to the idea of sharing existing code. I just kept all new code isolated initially. I will merge the testkit classes and move the tests to `integration/manual`.

libs/init/migrate/sbt/src/mill/init/migrate/sbt/SbtPackageWriter.scala
lihaoyi: We should make `SbtModule` use `sourceFolders` as well, just for consistency with the rest.

I think having traits in scalalib for each of them is fine as long as they're meaningful and well documented .

ajaychandran:
The PR title and description have been updated.

lihaoyi:
@ajaychandran Thanks for updating the PR description. Given this is a cleanroom implementation that aims to completely replace the old one, the PR description will need to be a lot more detailed to explain what the differences between the old implementation and new one are, to justify why it needs to be replaced and why the new implementation is better.

As mentioned over email, this will need to be integrated into the existing utils and Maven/Gradle implementation before merging. While it's fine for you to experiment with a new design in a separate module, it needs to be fully integrated into the existing code before merging, so we can see the results of your new "better" architecture in a realistic scenarios with the same requirements. The old SBT implementation should be fully removed, and whatever code that can be shared between the new SBT implementation and the Maven/Gradle implementations should be consolidated

libs/init/migrate/api/src/mill/init/migrate/ModuleConfig.scala
```diff
@@ -0,0 +1,108 @@
...
+
+import upickle.default.{ReadWriter, macroRW}
+
+sealed trait ModuleConfig
```
lihaoyi: Now that we are on Scala 3, can this be an `enum` with `derives ReadWriter` instead of a raw `sealed trait`? Most of the other `implicit val`s can probably be converted to `derives ReadWriter` as well

libs/init/migrate/src/mill/init/migrate/PackageWriter.scala
```diff
@@ -0,0 +1,334 @@
...
+      ps.print(" extends ")
+      ps.print(supertypes.head)
+      for supertype <- supertypes.tail do
+        ps.print(" with ")
+        ps.print(supertype)
+  }
+
+  def writeModuleConfigs(ps: PrintStream, configs: Seq[ModuleConfig]) = {
+    for config <- configs do
+      config match
```
lihaoyi: This can probably be `configs.foreach{`

libs/init/migrate/api/src/mill/init/migrate/ModuleConfig.scala
ajaychandran: The `api` is cross compiled with Scala 2.12 for use within SBT.

lihaoyi: got it

libs/init/migrate/sbt/exportscript/src/mill/init/migrate/sbt/ExportSbtBuildScript.scala
```diff
@@ -0,0 +1,225 @@
...
+          ),
+        mainConfigs = mainConfigs,
+        testModuleName = testModuleName,
+        testModuleBase = mvnDependencies.iterator.map(
+          _.organization
+        ).collectFirst(ModuleTypedef.testSupertypeByDepOrg),
+        testConfigs = testConfigs
+      )
+    }
+  }
```
lihaoyi: Where does the JSON serialization for these case classes happen? I assume they must be serialized to extract them out of SBT, but I can't seem to find where the code for that happens

ajaychandran:
> @ajaychandran Overall this looks decent as a first pass, but what is missing is how this interacts with the existing `libs/init/sbt` implementation
> 
>     * Does the new implementation pass all the same tests? I saw some tests failing on your fork's CI
> 
>     * Does the new implementation replace the old one? If so can we delete the old one entirely?
> 
>     * Can we share code between the old and new implementations? I see a bunch of duplicate logic that can be consolidated

- CI failures are for `sbt` unit tests. To run and compare old and new implementations, I added a new `@mainarg` in the existing implementation to redirect the control flow. Not sure why this would cause a failure, will investigate further.

- Yes, this implementation is meant to supplant the old one. For now though, I would like to retain the existing implementation for testing. The old implementation can be run by removing the `--poc` option from the `init` command.

- Of course we can share code where possible. If required, we can even reuse the existing package structure and files.

libs/init/migrate/sbt/exportscript/src/mill/init/migrate/sbt/ExportSbtBuildScript.scala
ajaychandran: See `exportModules` defined above this method.

lihaoyi:
> Yes, this implementation is meant to supplant the old one. For now though, I would like to retain the existing implementation for testing. The old implementation can be run by removing the --poc option from the init command.

Let's remove the old implementation as part of this PR. You can keep the old implementation around locally for testing and experimentation, but for the new implementation either we merge it to replace the old one, or we don't merge it at all. I'm confident enough in our testing (both existing tests and your new tests) that anything we merge should work, and I don't want two implementations of the same thing in the `main` branch in varying states of correctness and completeness

lihaoyi:
@ajaychandran let me know when you'd like me to take another look at this PR

ajaychandran:
I am planning to push a commit today with some refinements and missing features. Another commit will be required for documentation. The PR should be ready for review by tomorrow.

ajaychandran:
@lihaoyi
With the last commit, the features planned for the first milestone are complete.
I am going to need another day to complete testing and updating the documentation.

libs/init/migrate/src/mill/init/migrate/PackageWriter.scala
ajaychandran: I was hoping to port the other importers to the new design. This might eliminate `BuildGenUtil` entirely.
Can we discuss this?

ajaychandran:
@lihaoyi This is ready for another review.

A pending issue is that the Gatling example is missing 2 modules, `benchmarks` and `samples`, with the new importer. I am still diagnosing the problem.

lihaoyi:
@ajaychandran could you update the PR description with a summary of the major changes and how they fit into the existing codebase? In particular:

- What code paths were re-used
- What code paths were modified
- What code paths were replaced
- What code paths were duplicated
- What we expect the new implementation to be able to do that the old one cannot
- How this was tested

That would make this much easier to review than reverse-engineering your intent from the raw line diff

ajaychandran:
@lihaoyi The SBT importer implementation is ready for review. The PR title and description have been updated per your instructions.

lihaoyi:
> The current importer implementations consist of several models, roughly one per build tool, and conversions between these models. This is quite unnecessary and can be replaced with a single model that is optimized for code generation.

What do you mean by this? My understanding is that all current importers use the same model defined in `ir.scala`

libs/init/buildgen/src/mill/main/buildgen/BuildWriter.scala
```diff
@@ -0,0 +1,282 @@
...
+        writeNestedPackage(_, pkg)
+  }
+
+  protected def writeRootPackage(
+      ps: PrintStream,
+      pkg: Tree[ModuleRepr],
+      baseTraits: Seq[BaseTrait],
+      depsObject: Option[DepsObject]
+  ): Unit = {
+    ps.println(s"package $rootModuleAlias")
```
lihaoyi: Please use string concatenation rather than `println` for all this codegen

lihaoyi:
@ajaychandran what's the status of the existing unit and integration tests for importing SBT projects? Do they still pass?

ajaychandran:
> > The current importer implementations consist of several models, roughly one per build tool, and conversions between these models. This is quite unnecessary and can be replaced with a single model that is optimized for code generation.
> 
> What do you mean by this? My understanding is that all current importers use the same model defined in `ir.scala`

Additional models are used to extract the information and then these models are converted to the target model in `ir.scala`.
- SBT implementation uses models defined in `libs.init.sbt.models`.
- Gradle implementation uses models in `mill.main.gradle` package. These were added since Gradle requires data-types exchanged between client and server to be defined as Java interfaces that extend `Serializable`. But we can use the approach used from the SBT implementation and eliminates these, by writing the target model to file and reading it on the client side.

ajaychandran:
> @ajaychandran what's the status of the existing unit and integration tests for importing SBT projects? Do they still pass?

Yes but new tests have to be added to check cross development support.

libs/init/buildgen/src/mill/main/buildgen/BuildWriter.scala
ajaychandran: Do you mean the entire type should be rendered as a string?

lihaoyi:
Got it, let me turn on Ready to Review to see how CI looks

libs/init/buildgen/src/mill/main/buildgen/BuildWriter.scala
lihaoyi: yes

lihaoyi: In general, we should use strings rather than printstreams, unless there is some particular reason we need streaming

lihaoyi:
@ajaychandran from your PR description, I suppose the immediate follow up would be to convert the Gradle and Maven importers to use the same data model you are introducing for SBT here?

ajaychandran:
> @ajaychandran from your PR description, I suppose the immediate follow up would be to convert the Gradle and Maven importers to use the same data model you are introducing for SBT here?

Yes. I will start with the Gradle implementation.

lihaoyi:
Ok, I'll wait for you to resolve them println/string-interpolation comment and then we can merge this

ajaychandran:
> Ok, I'll wait for you to resolve them println/string-interpolation comment and then we can merge this

Does this mean that the other implementations should be updated in a separate PR?

lihaoyi:
Either way is fine, either as part of this PR or in follow ups.

lihaoyi:
@ajaychandran I'll leave this PR open if you want to add the Maven/Gradle implementationss to it, just let me know

ajaychandran:
> @ajaychandran I'll leave this PR open if you want to add the Maven/Gradle implementationss to it, just let me know

Yes, I would like to update the other implementations in this PR.

lihaoyi:
Ok will wait for thise changes to appear before taking another look

ajaychandran:
I changed the PR to a draft. Will update the status once the pending changes are completed.

lihaoyi:
@ajaychandran can you explain in more detail in the PR descriptions

1. What the technical changes are and why they were necessary
2. What the user-facing impact of the new approach is: what builds import now that didn't before? 
3. Any limitations of the new approach compared to the old one. Are there any regressions or open issues?
4. How was this tested. Manually? Through some script or command? In CI?

@lefou could you help review this? I'll look at it too, but it's a big PR so would be good to get another pair of eyes on it

lihaoyi:
> Added support for generating meta-build.

What scenarios would we need to generate a meta-build?

example/migrating/javalib/3-maven-complete-large/build.mill
```diff
@@ -6,7 +6,9 @@
 > git remote add -f origin https://github.com/iluwatar/java-design-patterns
 > git checkout ede37bd05568b1b8b814d8e9a1d2bbd71d9d615d
 
-> ./mill init --jvm-id 21 # Repo needs Java >=21 to build and test
+> echo 21 > .mill-jvm-version # Repo needs Java >=21 to build and test
+
+> ./mill init
```
lihaoyi: Do we no longer support `--jvm-id`?

ajaychandran:
> > Added support for generating meta-build.
> 
> What scenarios would we need to generate a meta-build?

Meta-build refers to base traits and `Deps` object defined under the `mill-build` folder. These are generated by default now.
This can be disabled with the `--no-meta-build` option resulting in build files that do not share anything.

example/migrating/javalib/3-maven-complete-large/build.mill
ajaychandran: The conversion now auto-configures `JavaHomeModule.jvmId` for Gradle and Maven projects.
Based on the projects tested, _JDK version enforcement is quite rare_. It is much more common to ensure compatibility with `javacOptions` like `-target`.
The Netty project is an example where the version, `1.8`, is enforced. But we  cannot support this value due to #5782.

The option was also being used to configure the JVM version in the build header. We could rename it to `--mill-jvm-version` and use it for this purpose.
I am not convinced that the conversion should configure this property since, in the typical case, _the preferred JDK is the one already installed_.
Consider the Mockito example in `5-gradle-incomplete`.
```
> git checkout v5.19.0 # multi-module Java project that requires Java 11

> echo "17" > .mill-jvm-version # Gradle version in use requires Java 17
```
Since `init` runs the Gradle daemon, we need at least version 17. Post conversion, JDK 11 would suffice (this won't be configured at the module level since it is not enforced in the build).
But why not use the latest version? The Mill project itself uses `zulu:21`.
The reason the version needs to be specified here is that we do not know the JDK version in the CI environment. We could move the value to `.mill-jvm-version` file to "hide" this from the user.

lihaoyi: Could we look at versions below 11 like `1.8` and generate `mill-jvm-version: 11`? That won't work for all projects, but it should hopefully work for many of them which aren't too picky about the version they are using. Then the user won't have to create the `.mill-jvm-version` file manually.

I think trying to specify a version as close as possible to that configured within their build via `javacOptions` sounds best. Not all projects work with newer versions, e.g. Scala, Kotlin, Spring, etc. all can potentially have problems if we pick a too-new version of the JVM to run them on

libs/scalalib/src/mill/scalalib/CrossSbtPlatformModule.scala
```diff
@@ -0,0 +1,21 @@
+package mill.scalalib
+
+/**
+ * A [[SbtPlatformModule]] suited to be used with [[mill.api.Cross]]. It supports additional source
+ * directories with the scala version pattern as suffix (src/main/scala-{scalaversionprefix}), e.g.
+ *  - src/main/scala-2
+ *  - src/main/scala-2.12
+ *  - src/main/scala-2.13
```
lihaoyi: How does the `-2`, `-2.12`, `-2.13` suffixes interact with the `-jvm`, `-js`, and `-native` suffixes? Can they be used together? I assume so, but we should mention how it works in the doccoment

mill-build/src/millbuild/MillStableJavaModule.scala
```diff
@@ -11,7 +11,18 @@ trait MillStableJavaModule extends MillPublishJavaModule with Mima {
...
+    // refactored layout modules for extensibility
+    ProblemFilter.exclude[NewMixinForwarderProblem]("mill.javalib.MavenModule.sources"),
+    ProblemFilter.exclude[NewMixinForwarderProblem]("mill.javalib.MavenModule#MavenTests.sources"),
+    ProblemFilter.exclude[NewMixinForwarderProblem]("mill.scalalib.SbtModule.sources"),
+    ProblemFilter.exclude[ReversedMissingMethodProblem](
+      "mill.scalalib.SbtModule.mill$scalalib$SbtModule$$super$sourcesFolders"
+    ),
+    ProblemFilter.exclude[NewMixinForwarderProblem]("mill.scalalib.SbtModule#SbtTests.sources"),
+    ProblemFilter.exclude[ReversedMissingMethodProblem](
+      "mill.scalalib.SbtModule#SbtTests.mill$scalalib$SbtModule$SbtTests$$super$sourcesFolders"
```
lihaoyi: We are not allowed to make these changes due to binary compatibility. You can normally work around this by adding your new code in helper methods that both the super-class and sub-class call

libs/scalalib/src/mill/scalalib/SbtPlatformModule.scala
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * A cross-platform [[SbtModule]] that supports multiple source root folders each having a sbt
+ * compatible directory layout.
+ *
+ * For `sbt-crossproject` plugin layout, use one of the following presets:
+ *  - [[SbtPlatformModule.CrossTypeFull]]
+ *  - [[SbtPlatformModule.CrossTypePure]]
+ *  - [[SbtPlatformModule.CrossTypeDummy]]
+ */
+trait SbtPlatformModule extends PlatformScalaModule with SbtModule { outer =>
```
lihaoyi, line 26: Could you explain in the doc-comment what this provides that the underlying `PlatformScalaModule` and `SbtModule` do not?

mill-build/src/millbuild/MillStableJavaModule.scala
ajaychandran: We are overriding the `sourcesFolders` member here as per a [suggestion](https://github.com/com-lihaoyi/mill/pull/5428#discussion_r2241907164) made earlier.
I don't think we can work around this. Should the changes be rolled back?

lihaoyi: We are allowed to override `sourceFolders`, the only problem is we are not allowed to add a `super` call within the override. The solution is to add a `private[mill] def sourceFolders0` task in the parent class and then call `sourceFolders0` in both classes without using `super`; that should have the same semantics as you want but avoid breaking binary compatibility

example/migrating/javalib/3-maven-complete-large/build.mill
ajaychandran: > Then the user won't have to create the .mill-jvm-version file manually.

> I think trying to specify a version as close as possible to that configured within their build via javacOptions sounds best. 

But why should the conversion explicitly specify the `.mill-jvm-version` at all? Isn't the default Mill behavior of using the installed JDK sufficient for real world usage?

The conversion should be concerned with configuring the JVM only at the module level and only if it is strictly enforced in the [Gradle](https://docs.gradle.org/current/userguide/toolchains.html#sec:using-java-toolchains) / [Maven](https://maven.apache.org/enforcer/enforcer-rules/requireJavaVersion.html) build.
I don't think `javacOptions` should be used to determine a version since a low value is targeted in many cases. Downgrading the JDK will likely adversely affect build times and require a fix in the build file (assuming the cause for the slowdown is identified).

lihaoyi: In the Mill `main` branch, we need to opt-into using the installed JDK via `mill-jvm-version: system`. But in general that is discouraged (hence making it no longer the default) and so we shouldn't generate it on-import if possible.

Pulling a low Java version is fine as long as it works. People can tweak and adjust it later, but our goal for the import _"a version that is most likely to work while avoiding `mill-jvm-version: system`"_. So if there is a `-target` or `-release` flag we should use as close to a version as possible, and if there is no flag we maybe just leave it unspecified and Mill will default to Java 21

ajaychandran:
> @ajaychandran can you explain in more detail in the PR descriptions
> 
>     1. What the technical changes are and why they were necessary
> 
>     2. What the user-facing impact of the new approach is: what builds import now that didn't before?
> 
>     3. Any limitations of the new approach compared to the old one. Are there any regressions or open issues?
> 
>     4. How was this tested. Manually? Through some script or command? In CI?
> 

The PR description has been updated.

libs/scalalib/src/mill/scalalib/CrossSbtPlatformModule.scala
ajaychandran: Updated Scaladocs.

libs/scalalib/src/mill/scalalib/SbtPlatformModule.scala
ajaychandran: Updated Scaladocs.

example/migrating/javalib/3-maven-complete-large/build.mill
ajaychandran: Reinstated `mill-jvm-version` in build header and added lower bound for Mill/module JVM version.

lihaoyi:
@ajaychandran can you include in your description a summary of:

* How the new data model handles cross-platform and cross-version Scala modules: how does it determine the Scala platforms and versions?
* How does it model single-platform single-version Scala modules?
* How does it handle Java modules, which cannot have platform or version?
* What example projects did you use to test these various cases?

mill-build/src/millbuild/MillStableJavaModule.scala
ajaychandran: The suggested changes were implemented and the MiMa filters were eliminated. But this caused `./mill example.thirdparty[netty].native.daemon` to fail with
```console
/home/runner/work/mill/mill/out/example/thirdparty/netty/native/daemon/testForked.dest/sandbox/run-2/build.mill:80:10
[7498]   [error] 80 ?  object test extends NettyBaseTestSuiteModule, MavenTests {
[7498]   [error]    ?         ^
[7498]   [error]    ?error overriding method sourcesFolders0 in trait MavenModule of type => Seq[os.SubPath];
[7498]   [error]    ?  method sourcesFolders0 in trait MavenTests of type => Seq[os.SubPath] module class test$ inherits conflicting members:
[7498]   [error]    ?  method sourcesFolders0 in trait MavenModule of type => Seq[os.SubPath]  and
[7498]   [error]    ?  method sourcesFolders0 in trait MavenTests of type => Seq[os.SubPath]
[7498]   [error]    ?(Note: this can be resolved by declaring an override in module class test$.)
[7498]   [warn] -- /home/runner/work/mill/mill/out/example/thirdparty/netty/native/daemon/testForked.dest/sandbox/run-2/build.mill:460:22
[7498]   [warn] 460 ?    } finally server.destroyForcibly()
[7498]   [warn]     ?                     ^^^^^^^^^^^^^^^
[7498]   [warn]     ?method destroyForcibly in class SubProcess is deprecated: Use destroy(shutdownGracePeriod = 0)
[7498]   [warn] one warning found
[7498]   [error] one error found
[7498]   1 tasks failed
[7498]   compile Compilation failed
```
I suppose this can be fixed by updating the test build definition or using a different name for one of the duplicates.

@lihaoyi Is there a recommended solution for such cases?

lihaoyi: Could we make `MavenTests` inherit from `MavenModule`? I think that should make it work by prioritizing the `MavenTests` version of the method. You may need to tweak some overrides in `MavenTests` to make that work and preserve the current behavior

lihaoyi: We can probably do the same with `SbtTests` and `SbtModule`

ajaychandran: Fixed. Thanks for the suggestions.

ajaychandran:
> @ajaychandran can you include in your description a summary of:
> 
>     * How the new data model handles cross-platform and cross-version Scala modules: how does it determine the Scala platforms and versions?
> 
>     * How does it model single-platform single-version Scala modules?
> 
>     * How does it handle Java modules, which cannot have platform or version?
> 
>     * What example projects did you use to test these various cases?

I have updated the description with the mechanism used.

The testing strategy is being reworked. Marked this as pending in the description.

ajaychandran:
The PR has been marked as draft. The status will be updated once the pending tasks are completed.

lihaoyi:
@ajaychandran any updates here?

ajaychandran:
@lihaoyi The test coverage for Gradle import was expanded and the issues uncovered were fixed. I am working on doing the same for Maven and SBT conversions.

example/migrating/javalib/3-maven-complete-large/build.mill
ajaychandran: The module `jvmId` derivation was updated to consider `javacOptions`.
The `mill-jvm-version` setting has been removed from the build header since no ideal value can be determined.

ajaychandran:

lihaoyi:
@ajaychandran could you write a bulleted list of what user-facing improvements this PR adds as part of the PR description? I saw there's `New features`, but are there any other things we should be aware of reviewing this?

lihaoyi:
For `Testing`, could you list out in the PR description what repos fail to import with the previous implementation, that now are able to import successfully? And for each one, what feature was missing before that this PR adds to make it succeed

integration/manual/migrating/src/MillInitGradleAsmTests.scala
```diff
@@ -0,0 +1,35 @@
...
+
+      // custom sourceSets not supported
+      eval(
+        "tools.retrofitter.compile",
+        stdout = os.Inherit,
+        stderr = os.Inherit
+      ).isSuccess ==> false
+    }
+  }
+}
```
lihaoyi: Can we consolidate the boilerplate code in these tests into a helper class? Each test object should only need to configure the git repo, version, and a list of `(command: String, success: Boolean)` tuples to specify what the test is meant to do

libs/init/buildgen/src/mill/main/buildgen/BuildWriter.scala
```diff
@@ -0,0 +1,769 @@
...
+
+    println(
+      s"NOTE: It is recommended to set mill-jvm-version in the header section in $rootBuildFile."
+    )
+  }
+
+  def renderBuildHeader = {
+    import build.*
+    s"""//| mill-version: $millVersion
+       |//| mill-jvm-opts: [${millJvmOpts.iterator.map(literalize(_)).mkString(", ")}]""".stripMargin
```
lihaoyi: Let's skip the `mill-jvm-opts` key if the value is empty

libs/init/buildgen/src/mill/main/buildgen/BuildWriter.scala
```diff
@@ -0,0 +1,769 @@
...
+      os.write(os.pwd / nestedBuildFile, renderPackage(pkg))
+    }
+
+    metaBuild.foreach { metaBuild =>
+      val rootDir = os.sub / millBuild
+      val srcDir = rootDir / "src"
+      os.makeDir.all(os.pwd / srcDir)
+      val rootBuildFile = rootDir / rootBuildFileNames.get(0)
+      println(s"writing Mill meta-build file to $rootBuildFile")
+      os.write(os.pwd / rootBuildFile, renderMetaBuildRoot)
```
lihaoyi: The `mill-build/build.mill` file is not necessary anymore if you just need to put things in `mill-build/src/`, since https://github.com/com-lihaoyi/mill/pull/5951, unless you actually want to customize something in there

libs/init/buildgen/src/mill/main/buildgen/BuildWriter.scala
```diff
@@ -0,0 +1,769 @@
...
+      case config: SbtPlatformModuleConfig => renderSbtPlatformModuleConfig(config, crossConfig)
+    })
+  }
+
+  def renderCoursierModuleConfig(
+      config: CoursierModuleConfig,
+      crossConfigs: Seq[(String, CoursierModuleConfig)]
+  ) = {
+    def get[A](f: CoursierModuleConfig => A) = crossConfigs.map((k, v) => (k, f(v)))
+    import config.*
```
lihaoyi: Let's skip the `import config.*` here and just refer to things directly via `config.repositories`. Here and elsewhere in this file

libs/init/buildgen/src/mill/main/buildgen/BuildWriter.scala
```diff
@@ -0,0 +1,769 @@
...
+    s"""${renderRepositories(repositories, get(_.repositories))}
+       |""".stripMargin
+  }
+
+  def renderJavaHomeModuleConfig(
+      config: JavaHomeModuleConfig,
+      crossConfigs: Seq[(String, JavaHomeModuleConfig)]
+  ) = {
+    def get[A](f: JavaHomeModuleConfig => A) = crossConfigs.map((k, v) => (k, f(v)))
```
lihaoyi: These local `def`s are kind of weird in these scenarios where there is only one call, can we just inline them unless it gets used more than once?

libs/init/buildgen/src/mill/main/buildgen/BuildWriter.scala
```diff
@@ -0,0 +1,769 @@
...
+
+  def renderArtifact(value: ModuleConfig.Artifact) = {
+    import value.*
+    s"Artifact(${literalize(group)}, ${literalize(id)}, ${literalize(version)})"
+  }
+
+  def renderPomSettings0(value: ModuleConfig.PomSettings) = {
+    import value.*
+    s"PomSettings(${
+        literalize(if (description == null) "" else description)
```
lihaoyi: Could we make these nullable fields `Option`s? Then we can just `.getOrElse("")` rather than this explicit `if-else` check

libs/init/gradle/test/resources/expected/gradle-9-0-0/mill-build/src/`9.0.0BaseModule`.scala
```diff
@@ -0,0 +1,36 @@
+package millbuild
+
+import mill.javalib._
+import mill.javalib.publish._
+
+trait `9.0.0BaseModule` extends MavenModule with PublishModule {
```
lihaoyi: Let's avoid backticks, maybe cal it `Gradle900BaseModule`  or something?

libs/init/gradle/test/resources/gradle-6-0/gradlew
```diff
@@ -0,0 +1,183 @@
+#!/usr/bin/env sh
```
lihaoyi: Are these `gradlew`, `gradlew.bat`, and `gradle-wrapper.jar` files available for download anywhere? If so we should download them as part of the build in `libs/init/package.mill` rather than vendoring them in our codebase

libs/init/buildgen/src/mill/main/buildgen/BuildWriter.scala
```diff
@@ -0,0 +1,769 @@
...
+      config: SbtPlatformModuleConfig,
+      crossConfigs: Seq[(String, SbtPlatformModuleConfig)]
+  ) = {
+    def get[A](f: SbtPlatformModuleConfig => A) = crossConfigs.map((k, v) => (k, f(v)))
+    import config.*
+    s"""${renderSourcesRootFolders(sourcesRootFolders, get(_.sourcesRootFolders))}
+       |""".stripMargin
+  }
+
+  def renderRepositories(values: Seq[String], crossValues: Seq[(String, Seq[String])]) =
```
lihaoyi: This long list of `render*` methods that just forward to `renderTask`/`renderTaskSeq` don't seem to do anything useful, especially since they're only called in one place each. Let's just inline them at the callsite

lihaoyi:
@lefou would you like to take a review pass on this as well? I've reviewed it, but it's a big PR and could use another set of eyes

ajaychandran:
@lihaoyi @lefou 

I will be pushing in some updates shortly. Please hold on to the review for now.

libs/init/buildgen/src/mill/main/buildgen/BuildWriter.scala
ajaychandran: Fixed

libs/init/gradle/test/resources/gradle-6-0/gradlew
ajaychandran: The files were removed since they are not used by the conversion.

libs/init/gradle/test/resources/expected/gradle-9-0-0/mill-build/src/`9.0.0BaseModule`.scala
ajaychandran: Fixed.
Using "Project" prefix since this is common to all conversions. The directory name is ignored in this case.

libs/init/buildgen/src/mill/main/buildgen/BuildWriter.scala
ajaychandran: Fixed.

libs/init/buildgen/src/mill/main/buildgen/BuildWriter.scala
ajaychandran: Fixed.

libs/init/buildgen/src/mill/main/buildgen/BuildWriter.scala
ajaychandran: Fixed.

libs/init/buildgen/src/mill/main/buildgen/BuildWriter.scala
ajaychandran: Fixed. The file is not generated anymore.

libs/init/buildgen/src/mill/main/buildgen/BuildWriter.scala
ajaychandran: Fixed.

integration/manual/migrating/src/MillInitGradleAsmTests.scala
ajaychandran: I replaced the tests with notes describing known issues.
I am not sure why we need these tests anymore. Do you have a use-case in mind?

ajaychandran:
@lihaoyi @lefou
This is ready for review.

The new core datatypes were refactored and simplified.
- `ModuleRepr` and `TestModuleRepr` were unified into `ModuleSpec`. This recursive type is a specification for a module and its nested modules.
- Added `PackageSpec` to define the path for a root module.
- `BuildRepr` was renamed to `BuildSpec`.
- The generic `Tree` type was removed. 

The PR description has been updated.

ajaychandran:
Also, the SBT script was changed to a plugin.
The classpath for plugins is isolated from scripts. This prevented the use of a reflection based technique to read plugin settings.

lihaoyi:
> For `Testing`, could you list out in the PR description what repos fail to import with the previous implementation, that now are able to import successfully? And for each one, what feature was missing before that this PR adds to make it succeed

@ajaychandran bump on this

ajaychandran:
> > For `Testing`, could you list out in the PR description what repos fail to import with the previous implementation, that now are able to import successfully? And for each one, what feature was missing before that this PR adds to make it succeed

Failure could be
- 'init` command fails
- compilation fails for generated build files
- imported configuration is invalid

Which one are you referring to here?

Should cross Scala projects be included?

ajaychandran:
@lihaoyi 
Added 2 entries for fixes for `init` failures in the "Testing" section.
Also, added an entry in "Other improvements" section for skipping publish.

lihaoyi:
> Failure could be
> 
> * 'init` command fails
> * compilation fails for generated build files
> * imported configuration is invalid
> 
> Which one are you referring to here?
> 
> Should cross Scala projects be included?

Any of them, what I want in the PR description is to know which of the features added in this PR is the one that made each new project's import succeed

ajaychandran:
Updated PR description with summary for test projects.

integration/manual/migrating/src/MillInitMavenCheckstyleTests.scala
```diff
@@ -0,0 +1,25 @@
...
+      import tester.*
+      eval("init", stdout = os.Inherit, stderr = os.Inherit).isSuccess ==> true
+      eval("__.showModuleDeps", stdout = os.Inherit, stderr = os.Inherit).isSuccess ==> true
+
+      s"""Requires support for code generation using antlrv4.
+         |
+         |[error] .../src/main/java/com/puppycrawl/tools/checkstyle/JavaParser.java:43:52: cannot find symbol
+         |[error]   symbol:   class JavaLanguageLexer
+         |[error]   location: package com.puppycrawl.tools.checkstyle.grammar.java
+         |""".stripMargin
```
lihaoyi: @ajaychandran as I mentioned twice already, these should be in the PR description, not in the code as random string literals

integration/manual/migrating/src/MillInitGradleAsmTests.scala
lihaoyi: We want the manual tests in case people want to run them in future. If not, how will anyone working on this code know whether these repos succeed or fail the import?

lihaoyi:
@ajaychandran please clean up the boilerplate in the manual tests as I mentioned earlier. They are part of the testing plan in the PR description, and so they need to be of decent quality and not just copy-paste boilerplate

testkit/src/mill/testkit/GitRepoIntegrationTestSuite.scala
```diff
@@ -0,0 +1,36 @@
+package mill.testkit
```
lihaoyi: This should not be in `mill.testkit`, it should live inside the `integration/manual/migrating/src/` since that's the only place that is using it

lihaoyi:
@ajaychandran given that you are adding 12,000 lines of code to the repo, can you summarize in the PR description what the breakdown of these 12,000 lines is? Is it all tests? test cases? Implementation code? 12,000 lines is a large burden to maintain going forward, and we need to ensure we minimize that burden where possible

ajaychandran:
> @ajaychandran please clean up the boilerplate in the manual tests as I mentioned earlier. They are part of the testing plan in the PR description, and so they need to be of decent quality and not just copy-paste boilerplate

Fixed, both manual and CI integrations.

ajaychandran:
> @ajaychandran given that you are adding 12,000 lines of code to the repo, can you summarize in the PR description what the breakdown of these 12,000 lines is? Is it all tests? test cases? Implementation code? 12,000 lines is a large burden to maintain going forward, and we need to ensure we minimize that burden where possible

Updated PR description. The "Layout changes" section was replaced with "Summary of code changes".

testkit/src/mill/testkit/GitRepoIntegrationTestSuite.scala
ajaychandran: Fixed.

integration/manual/migrating/src/MillInitMavenCheckstyleTests.scala
ajaychandran: Fixed. The comments have been removed.

integration/manual/migrating/src/MillInitGradleAsmTests.scala
ajaychandran: Fixed. The code was refactored.

libs/init/maven/src/mill/main/maven/MavenBuildGenMain.scala
```diff
@@ -1,300 +1,224 @@
...
 
+  /**
+   * @see [[MavenBuildGenArgs Command line arguments]]
+   */
   def main(args: Array[String]): Unit = {
-    val cfg = ParserForClass[Config].constructOrExit(args.toSeq)
-    run(cfg)
-  }
-
-  private def run(cfg: Config): Unit = {
```
lefou: Can we keep a second (public) entry-point accepting a `MavenBuildGenArgs`? This essentially split args parsing from the other logic and allows it to run with an explicit config. Same for the other `XxxGenMain` classes.

libs/init/sbt/src/mill/main/sbt/SbtBuildGenMain.scala
```diff
@@ -1,542 +1,216 @@
...
-    import upickle.*
-    val buildExport = read[BuildExport](buildExportPickled)
+    // For cross projects, multiple specs are exported with settings for each Scala version.
+    type ExportedPackageSpec = (
+        // A flag that is set for cross-platform projects and the package location.
+        // When the flag is set, the location is the shared root directory for member projects.
+        // Otherwise, the location is the project base directory.
+        (Boolean, Seq[String]),
+        ModuleSpec
+    )
```
lefou: Please use named tuples here, so we better see what we're dealing with in the use sites.

lefou (commented):
Lots of `public def`s do not have a return type annotation, which is unnecessary hard to read. Please add them.

libs/init/src/mill/init/Util.scala
```diff
@@ -1,27 +1,22 @@
...
-        |newlines.topLevelStatementBlankLines = [
-        |  {
-        |    blanks { before = 1 }
-        |  }
-        |]
-        |""".stripMargin
-    )
+  def scalafmtConfig: String =
+    """version = "3.8.5"
+      |runner.dialect = scala213
```
lefou:
```suggestion
      |runner.dialect = scala3
```

ajaychandran: This should be a separate PR. A consideration would be the syntax to use for the generated source.

libs/init/maven/src/mill/main/maven/MavenBuildGenMain.scala
ajaychandran: Added `runImport` method in all 3 applications.

libs/init/sbt/src/mill/main/sbt/SbtBuildGenMain.scala
ajaychandran: Using named tuples here would break JSON de-serialization. The SBT plugin is on Scala 2 and is limited to tuples.

Using another alias for this purpose can solve the problem.
```scala
    type ExportedPackageSpec0 = (
        metadata: (Boolean, Seq[String]),
        module: ModuleSpec
    )
    val exportedBuild: Seq[ExportedPackageSpec0] = os.list.stream(exportDir)
      .map(path => upickle.default.read[ExportedPackageSpec](path.toNIO))
      .toSeq
```
This might turn out to be a new source of confusion though. WDYT?

ajaychandran:
> Lots of `public def`s do not have a return type annotation, which is unnecessary hard to read. Please add them.

Added return type annotation for public methods.

ajaychandran:
Diagnosis of a problem reported in a test case revealed an issue in Mill.

For testing, the `jcommander` project uses `TestNg` that itself has a dependency on the project.
Post import, this resulted in a situation where the classes in `compile.dest` were permanently overshadowed due their trailing position in `compileClasspath`.
Perhaps Mill, like Gradle, should prioritize the `localCompileClasspath`.

lefou:
> Diagnosis of a problem reported in a test case revealed an issue in Mill.
> 
> For testing, the `jcommander` project uses `TestNg` that itself has a dependency on the project. Post import, this resulted in a situation where the classes in `compile.dest` were permanently overshadowed due their trailing position in `compileClasspath`. Perhaps Mill, like Gradle, should prioritize the `localCompileClasspath`.

This is an issue specific to jcommander / TestNG. I saw it multiple times. IMHO, there is no appropriate generic fix for that in tooling. You can either workaround it (by reordering the classpath, if you're in the lucky case that jcommander versions are binary compatible) or properly fix it by changing the design of TestNG. Both approaches are out of scope for a generic build converter.

libs/init/src/mill/init/Util.scala
lefou: Isn't this for the Mill Scala files, which are now Scala 3?

ajaychandran:
> This is an issue specific to jcommander / TestNG. I saw it multiple times. IMHO, there is no appropriate generic fix for that in tooling. You can either workaround it (by reordering the classpath, if you're in the lucky case that jcommander versions are binary compatible) or properly fix it by changing the design of TestNG. Both approaches are out of scope for a generic build converter.

I didn't mean this is an issue with respect to conversion; this is an issue in Mill.
Shouldn't `localCompileClasspath` be prioritized over other dependencies? Should I open a separate ticket for this discussion?

lefou:
@ajaychandran

> > This is an issue specific to jcommander / TestNG. I saw it multiple times. IMHO, there is no appropriate generic fix for that in tooling. You can either workaround it (by reordering the classpath, if you're in the lucky case that jcommander versions are binary compatible) or properly fix it by changing the design of TestNG. Both approaches are out of scope for a generic build converter.
> 
> I didn't mean this is an issue with respect to conversion; this is an issue in Mill. Shouldn't `localCompileClasspath` be prioritized over other dependencies? Should I open a separate ticket for this discussion?

If unrelated, please open a new issue.

libs/init/sbt/src/mill/main/sbt/SbtBuildGenMain.scala
lefou: Having two separate upickle setups for the same format is a call for trouble. It's not obvious and might break late after future refactorings.

I'd prefer, if we code-share the serialization format between the sbt plugin and this Mill implementation. Defining a case class in `api` module looks like the best approach.

libs/init/src/mill/init/Util.scala
ajaychandran: Yes.
Should the code generator be updated to replace deprecated syntax? AFAICT, only the wildcard imports need to be changed.

lefou: IIUC, you're saying that we need to change the generator code to be able to change the formatter setting? I don't get it. But it's probably a minor issue, so let's just keep it as is and concentrate on more essential parts of this already big PR.

> Should the code generator be updated to replace deprecated syntax? AFAICT, only the wildcard imports need to be changed.

In general yes. It might confuse users to start with already deprecated build configs.

ajaychandran: Fixed.

libs/init/sbt/src/mill/main/sbt/SbtBuildGenMain.scala
ajaychandran: Fixed. Added `api` module with custom types for the serialization format.

ajaychandran:
@lihaoyi Is there anything pending here?

libs/init/gradle/test/resources/gradle-6-0/build.gradle
```diff
@@ -0,0 +1,39 @@
+/*
```
lihaoyi: For all these files `was generated by the Gradle 'init' task`, can we generate them as part of our `libs/init/package.mill` in a [generated resource folder](https://mill-build.org/mill/javalib/intro.html#_custom_build_logic)? That would let us avoid committing all this generated code into `main` and having it potentiall drift or fall out of sync over time

integration/migrating/init/src/MillInitImportTestSuite.scala
```diff
@@ -0,0 +1,152 @@
...
+        os.list(cwd).head
+      }
+      override def initWorkspace() = {}
+    }
+    try {
+      import tester.{eval, workspaceSourcePath as resources}
+      val initRes = eval("init" +: initArgs)
+      assert(initRes.isSuccess)
+
+      if (configsGoldenFile != null) {
```
lihaoyi, line 35: Let's replace this and the one below with a single `val taskNames: Seq[String]` that we loop over and then `.mkString` into a single big string. That would remove all the duplication that comes from assigning things to local `val`s then interpolating them

lihaoyi: Also, how is this different from the version in `IntegrationTesterUtil`, and the version below in `renderAllImportedConfigurations`? Can they be combined somehow?

lihaoyi: Also, rather than generating a flat file, we can do a single `./mill show __.{repositories,jvmId,mvnDeps,...}` and which should return a JSON object we can parse and re-render with `indent = 4, sortKeys = true` as one big JSON object to use as our golden file contents. That will make it much easier for anyone reading these golden files later to understand what all these values mean, and avoid the messiness of the line-separated JSON objects you are using now

lihaoyi:
@ajaychandran i left some last comments, let's give a day or so to see if @lefou has anything else, after that we can merge it

integration/migrating/init/src/MillInitImportTestSuite.scala
ajaychandran: Refactored test method to remove duplication.
Removed unused `IntegrationTesterUtil` and `renderAllImportedConfigurations`.

libs/init/gradle/test/resources/gradle-6-0/build.gradle
ajaychandran: This is not possible since the output of Gradle `init` command varies across versions and is not guaranteed to be backwards compatible.

The test projects were "minified" by removing comments and unused source files. The current line count is at 245.

ajaychandran: Alternatively, the unit tests can be revised to run the conversion on real world projects. We could target older releases to ensure Gradle version compatibility.

lihaoyi:
@ajaychandran i think we can move forward. Could you resolve the conflicts one last time and then I'll merge it

lihaoyi:
@ajaychandran seeing some failures in the `example.migrating.javalib` job

ajaychandran:
@lihaoyi Could you please re-run the job? I tried the same on my fork and it [succeeded](https://github.com/ajaychandran/mill/actions/runs/18819113087/job/53695086582).

lihaoyi:
Retriggered the job, lets see how it goes

lihaoyi:
@ajaychandran still seeing failures, does your branch include the latest `main`?

ajaychandran:
Yes, the `main` branch was merged.

As per the log, the Mill server process fails to start.
```
[8148] Running:
[8148] D:\a\mill\mill\out\example\migrating\javalib\3-maven-complete-large\packaged\daemon\testForked.dest\sandbox\run-1> ./mill.bat  --ticker false init
[8148] --- Expected output ----------
[8148] 
[8148] ------------------------------
[8148] 
[8148] Running:
[8148] D:\a\mill\mill\out\example\migrating\javalib\3-maven-complete-large\packaged\daemon\testForked.dest\sandbox\run-1> ./mill.bat  --ticker false shutdown
[8148] --- Expected output ----------
[8148] 
[8148] ------------------------------
[8148] Success:
 [8148] D:\a\mill\mill\out\example\migrating\javalib\3-maven-complete-large\packaged\daemon\testForked.dest\sandbox\run-1> ./mill.bat  --ticker false shutdown
[8148] 
[8148] java.util.concurrent.ExecutionException: Boxed Exception
[8148] 	at scala.concurrent.impl.Promise$.scala$concurrent$impl$Promise$$resolve(Promise.scala:99)
[8148] 	at scala.concurrent.impl.Promise$DefaultPromise.tryComplete(Promise.scala:278)
[8148] 	at scala.concurrent.Promise.complete(Promise.scala:57)
[8148] 	at scala.concurrent.Promise.complete$(Promise.scala:56)
[8148] 	at scala.concurrent.impl.Promise$DefaultPromise.complete(Promise.scala:104)
[8148] 	at mill.util.Retry.$anonfun$1(Retry.scala:49)
[8148] 	at java.base/java.lang.Thread.run(Thread.java:829)
[8148] Caused by: utest.AssertionError: evalResult.isSuccess
[8148] evalResult: EvalResult = EvalResult(
[8148]   exitCode = 1,
[8148]   out = """Server died during startup:
[8148] ProcessDied{server=LaunchedServer.OsProcess{process=7204}, outputs=ServerLaunchOutputs{No daemon stdout
[8148] 
[8148] Daemon stderr:
[8148] 
[8148] Exception in thread "main" java.lang.Exception: Mill server process already present
[8148] 	at mill.server.Server.run$$anonfun$4(Server.scala:261)
[8148] 	at mill.server.Server.run$$anonfun$adapted$1(Server.scala:261)
[8148] 	at scala.Option.getOrElse(Option.scala:201)
[8148] 	at mill.server.Server.run(Server.scala:261)
[8148] 	at mill.daemon.MillDaemonMain$.main$$anonfun$1(MillDaemonMain.scala:60)
[8148] 	at mill.daemon.MillDaemonMain$.main$$anonfun$adapted$1(MillDaemonMain.scala:64)
[8148] 	at mill.api.SystemStreamsUtils$.withTopLevelSystemStreamProxy(SystemStreamsUtils.scala:75)
[8148] 	at mill.daemon.MillDaemonMain$.main(MillDaemonMain.scala:64)
[8148] 	at mill.daemon.MillDaemonMain.main(MillDaemonMain.scala)
[8148] 
[8148] 
[8148] }}
[8148] """,
[8148]   err = ""
[8148] )
```

ajaychandran:
Below is the error for a different test.
```
[8150] D:\a\mill\mill\out\example\migrating\javalib\5-gradle-incomplete\packaged\daemon\testForked.dest\sandbox\run-1> ./mill.bat  --ticker false init --gradle-jvm-id 17
[8150] --- Expected output ----------
[8150] converting Gradle build
[8150] init completed, run "mill resolve _" to list available tasks
[8150] ------------------------------
[8150] 
[8150] Running:
[8150] D:\a\mill\mill\out\example\migrating\javalib\5-gradle-incomplete\packaged\daemon\testForked.dest\sandbox\run-1> ./mill.bat  --ticker false shutdown
[8150] --- Expected output ----------
[8150] 
[8150] ------------------------------
[8150] java.util.concurrent.ExecutionException: Boxed Exception
[8150] 	at scala.concurrent.impl.Promise$.scala$concurrent$impl$Promise$$resolve(Promise.scala:99)
[8150] 	at scala.concurrent.impl.Promise$DefaultPromise.tryComplete(Promise.scala:278)
[8150] 	at scala.concurrent.Promise.complete(Promise.scala:57)
[8150] 	at scala.concurrent.Promise.complete$(Promise.scala:56)
[8150] 	at scala.concurrent.impl.Promise$DefaultPromise.complete(Promise.scala:104)
[8150] 	at mill.util.Retry.$anonfun$1(Retry.scala:49)
[8150] 	at java.base/java.lang.Thread.run(Thread.java:829)
[8150] Caused by: utest.AssertionError: evalResult.isSuccess
[8150] evalResult: EvalResult = EvalResult(
[8150]   exitCode = 1,
[8150]   out = """Server died during startup:
[8150] ProcessDied{server=LaunchedServer.OsProcess{process=1648}, outputs=ServerLaunchOutputs{No daemon stdout
[8150] 
[8150] Daemon stderr:
[8150] 
[8150] Error: could not open `C:\Users\runneradmin\AppData\Local\Coursier\cache\arc\https\cdn.azul.com\zulu\bin\zulu21.44.17-ca-jdk21.0.8-win_x64.zip\zulu21.44.17-ca-jdk21.0.8-win_x64\lib\jvm.cfg'
[8150] 
[8150] 
[8150] }}
[8150] """,
[8150]   err = ""
[8150] )
```

lihaoyi:
Let me try a more aggressive restart of all jobs rather than just the one that failed

PR #5446, opened 2025-07-04
Make `mill.idea/` work again

lihaoyi:


vaslabs:
is it also worth having a test like https://github.com/vaslabs-ltd/mill/blob/idea-gen-test/.github/workflows/run-tests.yml#L225 ?

(Not necessarily in that form and place, I'm just experimenting to see if it works)

lihaoyi:
Yes it's possible, not sure if it's worth it though. Tests are expensive and CI is already pretty slow

vaslabs:
> Yes it's possible, not sure if it's worth it though. Tests are expensive and CI is already pretty slow

ok, yeah, if it's not too serious...

it could be part of the `test-install-local.sh` or `test-mill-bootstrap.sh` if we wanted to have it (doesn't work as a millarg in the form I shared)

Issue #5448, opened 2025-07-04
`bspClean` fails


lihaoyi:
Reported by @unkarjedy in https://youtrack.jetbrains.com/issue/SCL-23974

```
[29-cleanCache] Entered buildTargetCleanCache
[29-cleanCache] Evaluating 1 task
[29-cleanCache] 1 tasks failed
[29-cleanCache] Task.Anon@35a88a0d(mill.main.MainModule#bspClean) Non-exclusive task Task.Anon@35a88a0d(mill.main.MainModule#bspClean) cannot depend on exclusive task clean
[29-cleanCache] Failed
1 tasks failed
Task.Anon@35a88a0d(mill.main.MainModule#bspClean) Non-exclusive task Task.Anon@35a88a0d(mill.main.MainModule#bspClean) cannot depend on exclusive task clean
Cleaning cache of bar failed, see Mill logs for more details
[29-cleanCache] buildTargetCleanCache took 7 msec
[29-cleanCache] buildTargetCleanCache caught exception: mill.api.Result$Exception: Non-exclusive task Task.Anon@35a88a0d(mill.main.MainModule#bspClean) cannot depend on exclusive task clean
[29-cleanCache] mill.api.Result$Exception: Non-exclusive task Task.Anon@35a88a0d(mill.main.MainModule#bspClean) cannot depend on exclusive task clean
[29-cleanCache] 	at mill.api.ExecResult$Failing.throwException(ExecResult.scala:69)
[29-cleanCache] 	at mill.api.ExecResult$Failing.throwException$(ExecResult.scala:64)
[29-cleanCache] 	at mill.api.ExecResult$Failure.throwException(ExecResult.scala:80)
[29-cleanCache] 	at mill.api.ExecResult.get(ExecResult.scala:18)
[29-cleanCache] 	at mill.api.ExecResult.get$(ExecResult.scala:11)
[29-cleanCache] 	at mill.api.ExecResult$Failure.get(ExecResult.scala:80)
[29-cleanCache] 	at mill.bsp.worker.MillBuildServer.$anonfun$23(MillBuildServer.scala:550)
[29-cleanCache] 	at scala.collection.IterableOnceOps.foldLeft(IterableOnce.scala:727)
[29-cleanCache] 	at scala.collection.IterableOnceOps.foldLeft$(IterableOnce.scala:721)
[29-cleanCache] 	at scala.collection.AbstractIterable.foldLeft(Iterable.scala:935)
[...]

PR #5449, opened 2025-07-04
Fix bsp clean

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/5448

`Task.Anon`s cannot call `Task.Command`s, so we move the `clean` logic into a separate `def cleanTask: Task.Anon` that can be shared between both


lefou:
> `Task.Anon`s cannot call `Task.Command`s

I think they can, but if the commands are `exclusive = true`, it won't work. This is the same issue I hit in https://github.com/com-lihaoyi/mill/pull/5424

While working around might help, we eventually need to properly handle that case.

PR #5466, opened 2025-07-07
Rename `mill.api.shared` to `mill.api.daemon`

lihaoyi:
`shared` feels too generic and friendly for an API that is pretty niche and internal. `daemonapi` feels more appropriately unnatural


lefou (commented):
Just make it `mill.api.daemon` since `daemonapi` is redundant.

PR #5471, opened 2025-07-08
Fix `gpg` hanging waiting for input when more than one key is present in keychain

arturaz:
Mill's artifact signing for publishing needs `gpg` to import the signing keys. However, the current implementation naively assumes that the key it imports is the only key.

If you have other keys in your keychain, a couple of things could happen:

- `gpg` by default signs using the first key in your keychain, which isn't necessarily the key you just imported, so your artifact can be signed using the wrong key.
- If the key `gpg` picks has a passphrase, `gpg` will hang indefinitely waiting for you to enter the passphrase to stdin, which will never happen, as it's mill handling the stdin.

This PR ensures that when the key is imported into GPG, its ID is extracted and then used in the signing commands to specify which key to use.

Tested manually on Linux and Mac.

### DIFF
--- a/contrib/sonatypecentral/readme.adoc
+++ b/contrib/sonatypecentral/readme.adoc
@@ -30,7 +30,7 @@ $ mill -i \
 mill.contrib.sonatypecentral/publishAll \
 --username myusername \
 --password mypassword \
---gpgArgs --passphrase=$MILL_PGP_PASSPHRASE,--no-tty,--pinentry-mode,loopback,--batch,--yes,-a,-b \
+--gpgArgs --passphrase=$MILL_PGP_PASSPHRASE,--no-tty,--pinentry-mode,loopback,--batch,--yes,--armor,--detach-sign \
 --publishArtifacts __.publishArtifacts \
 --readTimeout  36000 \
 --awaitTimeout 36000 \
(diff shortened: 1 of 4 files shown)


lihaoyi:
This was reverted in 1.0.0 due to breakage in Mill's publishing pipeline

```
[277] gpg: WARNING: no command supplied.  Trying to guess what you mean ...
[277] usage: gpg [options] [filename]
[277/277, 1 failed] =============== mill.javalib.Sonaty...ve.publishArtifacts =============== 158s
1 tasks failed
mill.javalib.SonatypeCentralPublishModule.publishAll os.SubprocessException: Result of gpg…: 2

    os.SubprocessException$.apply(Model.scala:211)
    os.proc.call(ProcessOps.scala:232)
    os.call$.apply(ProcessOps.scala:42)
    mill.javalib.publish.SonatypeHelpers$.gpgSigned(SonatypeHelpers.scala:55)
    mill.javalib.publish.SonatypeHelpers$.$anonfun$2(SonatypeHelpers.scala:30)
    mill.javalib.publish.SonatypeHelpers$.getArtifactMappings$$anonfun$1(SonatypeHelpers.scala:28)
    scala.collection.immutable.Vector1.map(Vector.scala:2141)
    scala.collection.immutable.Vector1.map(Vector.scala:386)
    mill.javalib.publish.SonatypeHelpers$.getArtifactMappings(SonatypeHelpers.scala:19)
    mill.javalib.SonatypeCentralPublisher.publishAll(SonatypeCentralPublisher.scala:46)
    mill.javalib.SonatypeCentralPublishModule$.publishAll$$anonfun$1(SonatypeCentralPublishModule.scala:151)
    mill.api.Task$Named.evaluate(Task.scala:328)
    mill.api.Task$Named.evaluate$(Task.scala:313)
    mill.api.Task$Command.evaluate(Task.scala:392)
Error: Process completed with exit code 1.
```

PR #5473, opened 2025-07-08
keytool functionalities for androidlib

souvlakias:
This PR replaces the use of the `keytool` CLI with a JVM-based implementation that handles keystore creation, key pair and certificate generation directly in code.

### Motivation

We want to be compatible with keystores created by Android Studio. Gradle seems to use the `KeystoreHelper` class for this ([source here](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:sdk-common/src/main/java/com/android/ide/common/signing/KeystoreHelper.java)), which we looked into.

While we could’ve pulled that in as a dependency, it comes with some Android/Gradle-specific stuff we don’t need. So for now, we’ve gone with a custom implementation that does just what we need.

## Notes

We tried using BouncyCastle as the provider for the Keystore handling but that resulted to unexpected behaviour with key passwords, so we just use default provider instead.

### What this changes

- Removes the `keytool` call used for debug keystore creation
- Adds code (and Tests) to:
- Create a new keystore
- Add a key pair and certificate
- Save and load a created keystore
- Load a keystore created by AGP and verify the keypair

### What’s next

We'll add a command in a future PR to let users generate a release keystore on their own.


libs/androidlib/src/mill/androidlib/keytool/CertUtil.scala
```diff
@@ -0,0 +1,41 @@
...
+import org.bouncycastle.cert.jcajce.{JcaX509v3CertificateBuilder, JcaX509CertificateConverter}
+import org.bouncycastle.operator.jcajce.JcaContentSignerBuilder
+import org.bouncycastle.jce.provider.BouncyCastleProvider
+
+object CertUtil:
+  Security.addProvider(new BouncyCastleProvider())
+  def createSelfSignedCertificate(
+      dname: String,
+      keyPair: KeyPair,
+      validityDays: Int = 365
```
vaslabs: we can use FiniteDuration

libs/androidlib/src/mill/androidlib/keytool/CertUtil.scala
```diff
@@ -0,0 +1,41 @@
...
+  ): X509Certificate = {
+    val now = new Date()
+    val notAfter = new Date(now.getTime + validityDays.toLong * 24 * 60 * 60 * 1000)
```
vaslabs: and here with FiniteDuration we can just say `.toMillis`

libs/androidlib/src/mill/androidlib/keytool/RSAKeyGen.scala
```diff
@@ -0,0 +1,16 @@
+package mill.androidlib.keytool
+
+import java.security.{KeyPair, KeyPairGenerator, Security}
+import org.bouncycastle.jce.provider.BouncyCastleProvider
+
+object RSAKeyGen {
+
+  Security.addProvider(new BouncyCastleProvider())
```
vaslabs: could potentially avoid this with 
`JcaX509CertificateConverter converter = new JcaX509CertificateConverter().setProvider(newBouncyCastleProvider());`

Not sure how different they are, but that's what AGP seems to be doing

libs/androidlib/src/mill/androidlib/keytool/Keystore.scala
```diff
@@ -0,0 +1,65 @@
...
+    val ks = KeyStore.getInstance(ksType)
+    val fis = new FileInputStream(file)
+    try ks.load(fis, password)
+    finally fis.close()
+    ks
+
+  def addKeyPair(
+      ks: KeyStore,
+      alias: String,
+      keyPair: java.security.KeyPair,
```
vaslabs: this can be imported at the top, or is their a duplicate name?

libs/androidlib/src/mill/androidlib/keytool/Keystore.scala
```diff
@@ -0,0 +1,65 @@
...
+      dname: String,
+      password: String
+  ): Unit =
+    val cert = CertUtil.createSelfSignedCertificate(dname, keyPair)
+    ks.setKeyEntry(alias, keyPair.getPrivate, password.toCharArray, Array(cert))
+
+  def listAliases(ks: KeyStore): List[String] =
+    val aliases = ks.aliases()
+    Iterator
```
vaslabs: maybe there's a scala conversion to do `ks.aliases().asScala.toList`

libs/androidlib/test/src/mill/androidlib/KeytoolTest.scala
```diff
@@ -0,0 +1,87 @@
+package mill.androidlib
```
vaslabs: can move this to keytool package to match the same package as the Keystore being tested, so it also won't need the 
`import mill.androidlib.keytool.*`

souvlakias:
@lihaoyi @lefou This is ready for review, I'd be happy if you could take a look, thank you

integration/ide/gen-idea/resources/extended/idea/mill_modules/mill-build.iml
```diff
@@ -12,6 +12,9 @@
         <orderEntry type="sourceFolder" forTests="false"/>
         <orderEntry type="library" name="scala-SDK-3.7.1" level="project"/>
         <orderEntry type="library" scope="RUNTIME" name="aircompressor-0.27.jar" level="project"/>
+        <orderEntry type="library" name="bcpkix-jdk18on-1.81.jar" level="project"/>
+        <orderEntry type="library" name="bcprov-jdk18on-1.81.jar" level="project"/>
+        <orderEntry type="library" name="bcutil-jdk18on-1.81.jar" level="project"/>
```
lihaoyi: Let's keep these libraries off of the Mill classpath by moving them into classloaders (https://mill-build.org/mill/extending/running-jvm-code.html#_in_process_isolated_classloaders) that would help keep the `build.mill` classpath clean and avoid conflicts with user code. You can see examples of how this is done in

- classgraph-worker
- jarjarabrams-worker
- maven-worker
- spotless-worker
- spring-boot-worker

souvlakias: Would an approach similar to the one used in [Hilt integration](https://github.com/com-lihaoyi/mill/blob/main/libs/androidlib/hilt/src/mill/androidlib/hilt/AndroidHiltTransformAsm.scala) (https://github.com/com-lihaoyi/mill/blob/main/libs/androidlib/package.mill#L73C1-L73C42) be okay?
Since there are multiple classes, it would be difficult to do it exclusively via reflection

lihaoyi: @souvlakias yes an approach like that would be fine. That's similar to the various `-worker` examples I mentioned above

libs/androidlib/package.mill
```diff
@@ -60,16 +65,6 @@ object `package` extends MillPublishScalaModule with BuildInfo {
...
 
-  object api extends MillAndroidModule {
-    def moduleDeps = Seq(build.libs.javalib.testrunner)
-
-    override def compileMvnDeps: T[Seq[Dep]] = Task {
-      super.compileMvnDeps() ++ Seq(
-        Deps.androidTools
-      )
-    }
-  }
```
lefou, line 71: What was this? Some unused left-over? There is no removed code associated to this removal.

souvlakias: Yes i left it over accidentally on [this PR](https://github.com/com-lihaoyi/mill/pull/5392)

PR #5476, opened 2025-07-08
Support publishing snapshots to central sonatype (0.12.x)

arturaz:
Backport of https://github.com/com-lihaoyi/mill/pull/5425 to Mill 0.12.x. (issue https://github.com/com-lihaoyi/mill/issues/4421)

---

Sonatype Central has an API for publishing artifact releases. However, that API does not accept snapshots. Instead, they have a Maven API compatible endpoint where SNAPSHOT artifacts should be sent.

This PR adds support for publishing SNAPSHOT releases to the Sonatype Central snapshots repository.

To do that, it uses [Apache Maven Artifact Resolver](https://maven.apache.org/resolver/) (ex-Eclipse Aether) library. To prevent classpath issues with code that supports `mill init` from Maven repositories (that also uses Maven Artifact Resolver, but a different version of it), a `maven-worker` is introduced.

A test has been added, however, because Sonatype API can be flaky and/or incur network errors, it is not run by default. It can be run with `./mill integration.feature[publish-sonatype-central-snapshots]`. You will have to provide your own namespace to publish to, as well as your Sonatype username and password.


arturaz:
Fixed the tests:
https://github.com/com-lihaoyi/mill/blob/4c76092a9b734d9556f30bab51ee0fea61b148a1/build.mill#L192-L195

These tests seem to fail even on the `0.12.x` branch (from https://github.com/com-lihaoyi/mill/actions/runs/16147204985/job/45569713411?pr=5476):
```
./mill example.android.javalib[2-app-bundle].local.server.test
./mill example.android.kotlinlib[3-compose-screenshot-tests].local.server.test 
```

lihaoyi:
Yes some of the android tests are borked in 0.12.x, we should probably just disable them

PR #5483, opened 2025-07-09
Update Scala 2.13 to 2.13.16

lefou:


lefou:
This requires support in mill-moduledefs.

PR #5495, opened 2025-07-09
Android compose samples: JetNews

souvlakias:
Added [JetNews](https://github.com/android/compose-samples/tree/main/JetNews) to the compose samples.
image

The object structure is similar to the JetLagged sample. R8 still currently throws some errors but converting them to warnings allows the app to function and the tests to pass successfully. 
This expands the test coverage and will be useful for detecting regressions in the future

### DIFF
--- a/example/thirdparty/android-compose-samples/build.mill
+++ b/example/thirdparty/android-compose-samples/build.mill
@@ -267,7 +267,7 @@ object JetNews extends mill.api.Module {
   }
 }
 
-/** Usage
+/** Usage JetLagged
 
 > ./mill JetLagged.app.androidApk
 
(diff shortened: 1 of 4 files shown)


vaslabs:
@souvlakias I pushed some scaffolding for the compose sample docs, let's add the hilt example too (there's not much going on in this PR so I think we can make it doc oriented + the JetNews docs and tests you added)

vaslabs:
@lihaoyi 

I've been thinking about adding more android samples - in this case for compose samples, we seem to be getting a timeout - on my machine the compose samples with both JetLagged and JetNews pass under 5 minutes.

So I'm thinking, instead of adding more and more samples here and making the Mill CI heavier, should we create an external repo just dedicated to android and mill ? We are not gonna have any regression coverage, but we can have a list of apps and configurations that can be built with mill 1.0.0

In the case of this PR, I'm thinking removing JetLagged and having JetNews in its place as it has more going on with android tests and would be a better example in terms of cases it covers

lihaoyi:
@vaslabs I think let's keep the examples in this repo, but just pick a subset of them to run in CI. That will keep them discoverable, and convenient to run manually when necessary, but without the performance and flakiness cost of running all of them every commit in CI

lihaoyi:
Concretely, that would mean replacing `example.androidlib.__.local.daemon` in our `run-tests.yml` file with specific examples, and just leaving the rest for manual runs when necessary

vaslabs:
> Concretely, that would mean replacing `example.androidlib.__.local.daemon` in our `run-tests.yml` file with specific examples, and just leaving the rest for manual runs when necessary

ok, I would prefer if we had them as a regression suite, but I think we can always keep the most complex third party examples and we should be ok . So in this case, I will leave the build configurations of both JetLagged and JetNews, but only run JetNews for better "regression coverage"

vaslabs:
I've added the both the compose samples and the architecture samples third party examples  in the android docs.

I'll follow up with more documentation and a blog post in future PRs

PR #5514, opened 2025-07-11
Rework test overrides repo

alexarchambault:
"Test overrides" stuff is being used to make integration tests or examples use modules and artifacts from the build rather than downloading those from Maven Central. Until now, the way it was injecting modules from the build was somehow low-level: the injected modules don't have dependencies anymore, and their full class path as provided by the build is put in the resolution artifacts as is.

The PR here makes it more principled: the build now provides a local Maven repository (`out/dist/localRepo.dest`, resulting in things like `out/dist/localRepo.dest/com/lihaoyi/mill-libs-javalib_3/SNAPSHOT/mill-libs-javalib_3-SNAPSHOT.{pom,jar}`), and the path to that local repository is passed around via an environment variable. Then, that local repository is added to the repository list, where coursier can pick things from it, and we don't blindly drop all dependencies and inject class paths as is any more.

This makes the resolutions happening in examples and ITs more similar to how it happens when using a Mill release. 

I've been needing while working on #5398, although I don't recall exactly which part of the fix requires it. But this makes things more principled, and less hackish IMO (no more `transitiveLocalTestOverrides` thing to maintain), so I think it's a good thing in itself.


contrib/playlib/src/mill/playlib/RouterModule.scala
```diff
@@ -47,13 +47,18 @@ trait RouterModule extends ScalaModule with Version {
   def routerClasspath: T[Seq[PathRef]] = Task {
     defaultResolver().classpath(
       playMinorVersion() match {
-        case "2.6" | "2.7" | "2.8" =>
-          Seq(mvn"com.typesafe.play::routes-compiler:${playVersion()}")
+        case "2.6" =>
```
alexarchambault, line 50: We make the Scala version suffixes explicit here, in accordance with what's done in `playRouteCompilerWorkerClasspath` a bit below. Without this, `playRouteCompilerWorkerClasspath` and `routerClasspath`, which are merged in `playRouterToolsClasspath`, might bring artifacts for different Scala versions. I don't recall if this breaks the CI here or only with some more work for https://github.com/com-lihaoyi/mill/issues/5398, but this needs fixing anyway.

alexarchambault: This makes the resolution of `com.lihaoyi:mill-contrib-playlib-worker-*`, which now happens with the new local repo in examples and integration tests, more consistent

lihaoyi:
One concern about using a single `localRepo` for everything is that it results in every test depending on every non-test module, which would result in poor selective testing precision since any code change will trigger every test to run. Part of the goal of the original `transitiveTestOverrides` infrastructure was to try and keep the dependencies narrow such that tests only depended on modules upstream of their enclosing non-test module

Is there some way to preserve some of the precision that `transitiveTestOverrides` gave us? Maybe having a separate `localRepo` for each module that aggregates only the transitive upstream module's classpaths, with a "global" `localRepo` used for integration/example tests? That would mirror the prior setup, just migrated to a `localRepo` folder rather than `MILL_LOCAL_TEST_OVERRIDES_*` env var

alexarchambault:
> Is there some way to preserve some of the precision that `transitiveTestOverrides` gave us? Maybe having a separate `localRepo` for each module that aggregates only the transitive upstream module's classpaths, with a "global" `localRepo` used for integration/example tests? That would mirror the prior setup, just migrated to a `localRepo` folder rather than `MILL_LOCAL_TEST_OVERRIDES_*` env var

We probably can do something like that, yes. Looking into it.

mill-build/src/millbuild/MillJavaModule.scala
```diff
@@ -29,13 +31,55 @@ trait MillJavaModule extends JavaModule {
...
-    val upstream = Task.traverse(moduleDeps ++ compileModuleDeps) {
-      case m: MillJavaModule => m.transitiveLocalTestOverrides.map(Some(_))
-      case _ => Task.Anon(None)
-    }().flatten.flatten
-    val current = Seq(localTestOverride())
-    upstream.toMap ++ current
+  def localTestExtraModules: Seq[MillJavaModule] = Nil
+  def localTestRepositories: T[Seq[PathRef]] = {
+
+    def recursive[T](start: T, deps: T => Seq[T]): Seq[T] = {
```
lihaoyi: I find this code surprisingly confusing. What is going on here? Is it just a depth-first or breadth-first search? If so, could we implement it with a more traditional while-loop and mutable queue/stack?

mill-build/src/millbuild/MillJavaModule.scala
```diff
@@ -29,13 +31,55 @@ trait MillJavaModule extends JavaModule {
...
+
+      rec(Set(), Nil, start :: Nil).reverse
+    }
+
+    val allModules = recursive[MillJavaModule](
+      this,
+      m => {
+        val localTestExtraModules0 = m match {
+          case m0: MillJavaModule => m0.localTestExtraModules
+          case _ => Nil
```
lihaoyi: Given the type signatures involved, it seems like this `case` will never occur? And the `.collect` below can just be a `.map`?

libs/javalib/src/mill/javalib/PublishModule.scala
```diff
@@ -600,6 +600,28 @@ trait PublishModule extends JavaModule { outer =>
...
+  /**
+   * Creates a Maven repository directory with main and source artifacts for this module
+   *
+   * This results in a directory which contains things like
+   * `my/organization/my-module/VERSION/my-module-VERSION.{pom,jar}`.
+   * Such directories can later be used as Maven repositories during
+   * artifact resolution, or be merged together and then used as repository,
+   * for example.
+   */
+  def stagePublish: Task[PathRef] = Task {
```
lihaoyi: Let's put the `publish` prefix first, to namespace the task a bit better. `publishLocal` is already taken, maybe `publishStage`? `publishTestFolder`?

PR #5519, opened 2025-07-11
Fix version sources in CrossSbtTests

jackkoenig:
They were pointing to src/main instead of src/test.

I was about to file an issue but the fix is easy enough. Of course this probably needs a test.


seldridge (approved):
LGTM

PR #5524, opened 2025-07-14
Refactor publishing code to use a more appropriate data structure to represent the publishing data

arturaz:
A part of https://github.com/com-lihaoyi/mill/pull/5425 extracted from it.

The API has been changed from `Seq[(Contents, String)]` to `Map[os.SubPath, Contents]`.

- `String` is replaced with `os.SubPath` to make sure the paths are relative without any `..` components.
- `Seq` is replaced with `Map` to ensure multiple contents cannot be written to the same file path.

Additionally, this turns on mima signature checking which ensures that generics are also checked. This has caught an issue in `defaultGpgArgs`, which has been fixed by this PR as well.

### DIFF
--- a/libs/javalib/src/mill/javalib/SonatypeCentralPublisher.scala
+++ b/libs/javalib/src/mill/javalib/SonatypeCentralPublisher.scala
@@ -1,10 +1,6 @@
 package mill.javalib
 
-import com.lumidion.sonatype.central.client.core.{
-  DeploymentName,
-  PublishingType,
-  SonatypeCredentials
-}
+import com.lumidion.sonatype.central.client.core.{DeploymentName, PublishingType, SonatypeCredentials}
 import com.lumidion.sonatype.central.client.requests.SyncSonatypeClient
 import mill.api.Logger
 import mill.javalib.internal.PublishModule.GpgArgs
@@ -15,6 +11,7 @@ import mill.javalib.publish.SonatypeHelpers.getArtifactMappings
 import java.nio.file.Files
 import java.util.jar.JarOutputStream
 import java.util.zip.ZipEntry
+import scala.annotation.targetName
 
 /**
  * Publishing logic for the standard Sonatype Central repository `central.sonatype.org`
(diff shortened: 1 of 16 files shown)


libs/javalib/src/mill/javalib/SonatypeCentralPublisher.scala
```diff
@@ -53,23 +54,51 @@ class SonatypeCentralPublisher(
   private val sonatypeCentralClient =
     new SyncSonatypeClient(credentials, readTimeout = readTimeout, connectTimeout = connectTimeout)
 
+  // binary compatibility forwarder
```
lefou, line 57: I think we want a `@deprecated` here. Alternatively, we could add a `@publicInBinary` and make it `private[javalib]`, so it's not accidentally used by new code?

libs/javalib/src/mill/javalib/SonatypeCentralPublisher.scala
```diff
@@ -53,23 +54,51 @@ class SonatypeCentralPublisher(
...
       artifacts: (Seq[(os.Path, String)], Artifact)*
+  ): Unit = {
+    val mappedArtifacts = artifacts.iterator.map { case (fileMapping, artifact) =>
+      val mapping = fileMapping.iterator.map { case (path, name) => os.SubPath(name) -> path }.toMap
+      mapping -> artifact
+    }.toArray
+    publishAll(publishingType, singleBundleName, mappedArtifacts*)
+  }
+
+  @targetName("publishAllByMap")
```
lefou, line 88: Can we have a comment here, why we need the `@targetName`? What is the intention?

arturaz: Because otherwise it doesn't compile, as after the generic erasure, both functions have the signature of `artifacts: Seq`.

libs/javalib/src/mill/javalib/SonatypeCentralPublisher.scala
arturaz: I'll add the `@deprecated` in https://github.com/com-lihaoyi/mill/pull/5525

PR #5528, opened 2025-07-14
Pretty print artifact on publishing

lolgab:


lihaoyi:
Looks fine but the tests all need to be updated

libs/javalib/src/mill/javalib/publish/LocalIvyPublisher.scala
```diff
@@ -39,7 +39,9 @@ class LocalIvyPublisher(localIvyRepo: os.Path) {
       artifact: Artifact,
       contents: Map[os.SubPath, FileSetContents.Writable]
   )(implicit ctx: TaskCtx.Log): Seq[os.Path] = {
-    ctx.log.info(s"Publishing ${artifact} to ivy repo ${localIvyRepo}")
+    ctx.log.info(
+      s"Publishing ${artifact.group}:${artifact.id}:${artifact.version} to ivy repo ${localIvyRepo}"
+    )
```
lefou, line 44: Should we override `Artifact.toString()` instead?

Issue #5538, opened 2025-07-15
Publishing data structure situation needs improvement


arturaz:
Currently, we have a couple of data structures to represent data for publishing:
- `PublishModule.PublishData`:
  https://github.com/com-lihaoyi/mill/blob/5e650f6b78d903f34e122a9f97c6b223c6251d8f/libs/javalib/src/mill/javalib/PublishModule.scala#L633-L642
- `PublishInfo`:
   https://github.com/com-lihaoyi/mill/blob/5e650f6b78d903f34e122a9f97c6b223c6251d8f/libs/javalib/src/mill/javalib/publish/PublishInfo.scala#L5-L20

   The javadocs on it say "An extra resource artifact to publish.", but in reality it's used in the codebase to represent a file to publish from the JVM ecosystem with its metadata.

To date, it seems that `PublishData` has been used as a de facto container for "things that need publishing."

The problem is that, depending on what we are publishing, we need more metadata about the files than `PublishData` can provide.

For example, when publishing snapshots to Sonatype Central, we use `PublishInfo`s and the POM file instead of `PublishData`:
https://github.com/com-lihaoyi/mill/blob/5e650f6b78d903f34e122a9f97c6b223c6251d8f/libs/javalib/src/mill/javalib/SonatypeCentralPublishModule.scala#L63-L67

Whereas when doing a release, we use the `PublishData`:
https://github.com/com-lihaoyi/mill/blob/5e650f6b78d903f34e122a9f97c6b223c6251d8f/libs/javalib/src/mill/javalib/SonatypeCentralPublishModule.scala#L83-L85

This works if using default settings, but the user wants to override some tasks, there's not one clear place what he needs to override. Even worse, the places that he/she needs to override depend on the internal implementation details of Mill.

Ideally, this should be refactored into data structures that make sense and bring the necessary metadata with them until the last second, when the actual publishing is happening.

However, with the 1.0.0 released, there is no way to do this without breaking binary compatibility.

So I do not have a proposed solution here, but I am surfacing the issue for visibility.

PR #5539, opened 2025-07-15
Snapshot handling in `SonatypeCentralPublishModule/publishAll`

arturaz:
Currently, there are two ways to publish artifacts to Sonatype Central:
- `./mill __.publishSonatypeCentral`
- `./mill SonatypeCentralPublishModule/publishAll`

Unfortunately, they have different code paths, and when adding snapshot support, I only added it to `publishSonatypeCentral`, which resulted in people being confused about why things do not work when they use `SonatypeCentralPublishModule/publishAll`.

I looked at whether snapshot support can be added to `publishAll`, but it seems it can't, as the `publishArtifacts: mill.util.Tasks[PublishModule.PublishData] = Tasks.resolveMainDefault("__:PublishModule.publishArtifacts")` value does not have enough metadata for snapshot publishing (we require `PublishInfo`, but by the time we get `PublishData` it is already gone).

The current solution is to:
- abort the task if you're only trying to publish snapshots.
- warn the user and skip the snapshots if you are mixing snapshots and releases.

The solution isn't great, but the whole publishing pipeline needs improvements, as mentioned in https://github.com/com-lihaoyi/mill/issues/5538.

### DIFF
--- a/libs/javalib/src/mill/javalib/PublishModule.scala
+++ b/libs/javalib/src/mill/javalib/PublishModule.scala
@@ -488,6 +488,10 @@ trait PublishModule extends JavaModule { outer =>
   /** @see [[PublishModule.sonatypeCentralSnapshotUri]] */
   def sonatypeCentralSnapshotUri: String = PublishModule.sonatypeCentralSnapshotUri
 
+  @deprecated(
+    "Do not override this task, override `artifactMetadata`, `publishArtifactsDefaultPayload` or `extraPublish` instead.",
+    "1.0.1"
+  )
   def publishArtifacts: T[PublishModule.PublishData] = Task {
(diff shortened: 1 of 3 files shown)


arturaz:
After discussing this with @lihaoyi we decided that we'll add support for publishing snapshots to `publishAll`. It did require extracting metadata from strings and is not guaranteed to work with all cases:
https://github.com/arturaz/mill/blob/e88240354d4ba417748aedfebda6f7c8b1c92d62/libs/javalib/src/mill/javalib/publish/PublishInfo.scala#L69-L101

But it seems to work fine for the default publishing workflow, which is what most people use.

https://github.com/com-lihaoyi/mill/issues/5538 is still relevant, though.

lihaoyi (approved):
I think this looks fine, @arturaz could you update the PR description with the latest summary and then I'll merge it

PR #5543, opened 2025-07-15
Don't accept CI secrets on CLI in SonatypeCentralPublishModule

alexarchambault:
This PR makes `SonatypeCentralPublishModule` refuse credentials passed via `--username` and `--password` if it detects it's running on a CI. Using these arguments poses security issues: the Mill's prompt tries to print them back, like in
```text
[4032/4037] =============== mill.scalalib.SonatypeC...N/lEdX9diiWxHsua8VLfDzw =============== 480s
```
(prompt obtained by running a command like `./mill -i mill.scalalib.SonatypeCentralPublishModule/ --username … --password …`, containing parts of a now revoked token)

These secrets might be truncated in particular, which can make parts of these secrets slip through output secret detection.

### DIFF
--- a/libs/javalib/src/mill/javalib/SonatypeCentralPublishModule.scala
+++ b/libs/javalib/src/mill/javalib/SonatypeCentralPublishModule.scala
@@ -53,10 +53,11 @@ trait SonatypeCentralPublishModule extends PublishModule with MavenWorkerSupport
 
   def publishSonatypeCentral(
       username: String = defaultCredentials,
-      password: String = defaultCredentials
+      password: String = defaultCredentials,
+      force: Boolean = false
   ): Task.Command[Unit] = Task.Command {
     val artifact = artifactMetadata()
-    val finalCredentials = getSonatypeCredentials(username, password)()
+    val finalCredentials = getSonatypeCredentials(username, password, force)()
 
     def publishSnapshot(): Unit = {
       val uri = sonatypeCentralSnapshotUri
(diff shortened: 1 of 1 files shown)


arturaz:
Perhaps we should make Mill support taking in `mill.util.Secret[String]` as the command argument, and that would prevent Mill from printing it out in any place?

alexarchambault:
@arturaz Maybe at some point, yes. We need better secret management overall, some secrets might leak in serialized task results under `out/` too. I think we should merge this in the mean time.

libs/javalib/src/mill/javalib/SonatypeCentralPublishModule.scala
```diff
@@ -279,8 +292,17 @@ object SonatypeCentralPublishModule extends ExternalModule with DefaultTaskModul
...
+      force: Boolean
   ): Task[SonatypeCredentials] = Task.Anon {
+    val isCI = Task.env.get("CI").nonEmpty
+    if (!force && isCI && (usernameParameterValue.nonEmpty || passwordParameterValue.nonEmpty))
+      sys.error(
+        "--username and --password options forbidden on CI. " +
+          "Their use might leak secrets. " +
+          s"Pass those values via environment variables instead ($USERNAME_ENV_VARIABLE_NAME and $PASSWORD_ENV_VARIABLE_NAME), or pass --force alongside them. " +
+          "You might want to check the output of this job for a leak of those secrets or parts of them."
+      )
```
lefou, line 305: Since we consider this unsafe, we might as well print the same message as warning in all other cases.

Also, instead of `sys.error` use `Task.fail`.

alexarchambault:
Closing this for now. I might get back to it later (or would be ok if anyone wants to take over it).

PR #5546, opened 2025-07-16
Minor fixes for project importer

lihaoyi:
- If no test framework can be identified, default to `TestModule.Junit4` to allow the build to at least compile
- Add a space between the `lazy val $lhs` and the following `:` to avoid parsing errors when `lhs` ends with a `_`

Noticed these while trying to get MIMA imports working


lihaoyi:
Looks like the failure in `example.migrating.javalib[3-maven-complete-large].packaged.daemon`'s `page-object` test suite is that it is incorrectly configured in Maven: it does not a test framework or test dependencies configured, and thus doesn't run when you use `./mvnw clean test -pl page-object -am`.

Issue #5534, opened 2025-07-15
Remake `intellijModulePathJava` public


lefou:
It is meant to be overridable by the user. Otherwise, it's not possible to have multiple test sub-modules in one module.

_Originally posted by @lefou in https://github.com/com-lihaoyi/mill/pull/5387#discussion_r2170907410_

PR #5550, opened 2025-07-17
Make `GenIdeaModule.intellijModulePathJava` public

lefou:
Fix https://github.com/com-lihaoyi/mill/issues/5534

### DIFF
--- a/core/api/daemon/src/mill/api/daemon/internal/idea/GenIdeaModuleApi.scala
+++ b/core/api/daemon/src/mill/api/daemon/internal/idea/GenIdeaModuleApi.scala
@@ -7,6 +7,6 @@ trait GenIdeaModuleApi {
   /**
    * The path denoting the module directory in generated IntelliJ projects. Defaults to [[moduleDir]].
    */
-  private[mill] def intellijModulePathJava: java.nio.file.Path
+  def intellijModulePathJava: java.nio.file.Path
 
 }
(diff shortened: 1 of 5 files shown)


lihaoyi:
Could we expose a `os.Path` version of this to users? athe `*Java` version using java types isnt really idiomatic, and is only necessary as part of the internal daemon api

Could we also flesh out the doc comment explaining what the user case is. I don't actuslly understand why this needs to be configurable by users

chikei:
my use case is to define a separated test (that have its own source code) module from the usual `src/test` one. For example, an integration test under `src/it`

lihaoyi:
@chikei why do you need to override `intellijModulePathJava` to define a separate integration test, and not for normal tests?

chikei:
maybe because I am using maven/sbt layout and mill already did that for `src/test` in `MavenModule#MavenTests` ?

lihaoyi:
@chikei seems like we should be able to do it automatically on your behalf without the user needing to configure stuff. I opened a PR that does that https://github.com/com-lihaoyi/mill/pull/5587

lihaoyi:
@chikei I've tagged an unstable release `1.0.1-8-3f52cf` that includes https://github.com/com-lihaoyi/mill/pull/5587. I think that should satisfy your use case without needing to customize `intellijModulePathJava` yourself. Please try it out and let me know if it works for you

chikei:
@lihaoyi tested `1.0.1-8-3f52cf`, I think it generates correct idea project structure

lihaoyi:
@chikei great, I'll close this and the linked issue for now, but feel free to open a new issue if there are other problems that need fixing

PR #5551, opened 2025-07-17
Cleanup compiler warnings in bsp worker module

lefou:


runner/bsp/worker/src/mill/bsp/worker/MillBuildServer.scala
```diff
@@ -374,7 +374,7 @@ private class MillBuildServer(
         s"Getting sources of ${sourcesParams.getTargets.asScala.map(_.getUri).mkString(", ")}",
       originId = ""
     ) {
-      case (_, _, id, _: JavaModuleApi, result) => new SourcesItem(
+      case (_, _, id, _, result) => new SourcesItem(
```
lefou, line 377: Unnecessary type match, since we already applied this type filter in the `tasks` partial function above.

PR #5582, opened 2025-07-22
Remove need for `mvn` interpolator in deps

lihaoyi:


lihaoyi:
Looks like the main problem here is that we use overloading for `def apply` in order to make `Task{...}` and `Task(persistent = true){...}` work. This overloading disables target-typing within the `...` body of the task, which means implicit conversions like `Dep.parse` are not kicking in.

lihaoyi:
Here's a minimized repro of what we need to make work

```scala
implicit def makeOption(t: Seq[Dep]): Option[Seq[Dep]] = Some(t)
class Dep(val s: String)
implicit def makeDep(s: String): Dep = new Dep(s)
  
def foo: Option[Seq[Dep]] = Seq("hello", "world")
```

Basically given a `Seq("hello", "world")`, we need to both apply an implicit conversion _within_ it to `Seq(makeDep("hello"), makeDep("world"))`, as well as implicit conversion _around_ it to `makeOption(Seq("hello", "world"))`. In Mill, `makeOption` is the `Task.Simple.create` macro, and `makeDep` is the `Dep.parse` conversion, but the problem is the same

PR #5589, opened 2025-07-22
Fixes incorrect parsing of `PublishInfo` from string when artifact id or version contains regex-sensitive characters

arturaz:
This PR fixes parsing of strings like "chisel-plugin_3.3.4-7.0.0-RC3+5-c4ca27e5-SNAPSHOT.pom". Previously `chisel-plugin_2.13.14-7.0.0-RC3+4-4815de48-SNAPSHOT` would be identified as a `classifier`.

This happened because when constructing a regular expression we did not escape the id/version parts and thus symbols like + were interpreted as regex instructions instead of literals, which made the regex never match.

Additionally, a small refactoring has been done where `PublishInfo.parseFromFile` is now `PublishInfo.IvyMetadata.parseFromFile`, as it's actually the metadata we are parsing.

### DIFF
--- a/libs/javalib/src/mill/javalib/publish/PublishInfo.scala
+++ b/libs/javalib/src/mill/javalib/publish/PublishInfo.scala
@@ -2,6 +2,8 @@ package mill.javalib.publish
 
 import mill.PathRef
 
+import java.util.regex.Pattern
+
 /**
  * An extra resource artifact to publish.
  *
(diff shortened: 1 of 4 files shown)


lihaoyi:
I don't quite understand the underlying bug here; could you elaborate on it a bit in the PR description?

libs/javalib/src/mill/javalib/publish/PublishInfo.scala
```diff
@@ -74,6 +76,39 @@ object PublishInfo {
...
+     * extensions are without the dot, so things like ".tar.gz" aren't supported, but they do not seem to be used in Maven
+     * ecosystem, so that works out for most of the cases.
+     *
+     * This should eventually be rehauled, see https://github.com/com-lihaoyi/mill/issues/5538 for more information.
+     *
+     * @param fileName        name of the file to use. Can be different from the actual filename of the `file`.
+     * @param artifactId      for example, "mill"
+     * @param artifactVersion version of the artifact, for example "1.0.0-RC3"
+     */
+    def parseFromFile(
```
jackkoenig, line 94: Should this remain package private?

arturaz: The enclosing object is package private, so this should be too.

Issue #5592, opened 2025-07-23
Publishing to maven repository other than central


chikei:
It used to be done through `SonatypePublisher` with `stagingRelease = false` and can publish without gpg signing. 

Now we're deprecating `SonatypePublisher`, maybe we can make a `MavenPublishModule` which work just like `SonatypeCentralPublishModule` but instead of doing staging in `publishAll#publishReleases`, we just do the same as in `publishSnapshot` (i.e. publish to given uri using aether)? 

After this we probably can deprecate `contrib.Artifactory` as well.

lihaoyi:
@lefou tagging you here and in https://github.com/com-lihaoyi/mill/issues/5464 to resolve these. I think we have most of the code in place, and what is needed is probably updated docs that explicitly cover how to cover these non-maven-central use cases. Not sure if we'll need any actual code changes.

chikei:
https://github.com/com-lihaoyi/mill/pull/5601#issuecomment-3240984931 explains what changes in the PR, basically we need a way to 1. instead of using repo1 staging flow, upload directly with aether (the `worker.publishToRemote` call in current `publishSnapshot`) in publishing release version. 2. call `publishAll` without gpg (or a dummy gpg args that does not trigger errors). But otherwise I don't think we need any changes

lihaoyi:
Closed by https://github.com/com-lihaoyi/mill/pull/6023

lihaoyi:
https://github.com/com-lihaoyi/mill/pull/6023 was reverted

PR #5601, opened 2025-07-25
Add MavenPublishModule for maven compatiable repository

chikei:
more like a POC for #5592, I think there are rooms for sharing code/task between this and SonatypeCentralPublishModule

### DIFF
--- a/libs/javalib/src/mill/javalib/MavenPublishModule.scala
+++ b/libs/javalib/src/mill/javalib/MavenPublishModule.scala
@@ -1,0 +1,207 @@
+package mill.javalib
+
+import com.lihaoyi.unroll
+import com.lumidion.sonatype.central.client.core.SonatypeCredentials
+import mill.*
+import javalib.*
+import mill.api.{ExternalModule, Task}
+import mill.util.Tasks
+import mill.api.DefaultTaskModule
+import mill.api.Result
+import mill.javalib.MavenPublishModule.{
+  defaultAwaitTimeout,
+  defaultConnectTimeout,
+  defaultCredentials,
+  defaultReadTimeout,
+  getSonatypeCredentials
+}
+import mill.javalib.publish.Artifact
+import mill.javalib.publish.SonatypeHelpers.{PASSWORD_ENV_VARIABLE_NAME, USERNAME_ENV_VARIABLE_NAME}
+import mill.api.daemon.Logger
+import mill.javalib.PublishModule.PublishData
+
+trait MavenPublishModule extends PublishModule, MavenWorkerSupport {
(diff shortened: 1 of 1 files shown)


chikei:
It's copied from existing `SonatypeCentralPublishModule` and replace the `publishReleases` in `publishAll` using almost identical from `publishSnapshot`. Effectively changing the publish flow for release from repo1's staging flow to direct http upload.

Edit: and remove the gpg args from `publishAll` as well

lefou (commented):
So, the general approach looks good. 

I'd like a have the implementations of `object MavenPublishModule` moved into a trait and just extends that. That way, it should be easier to customize and reuse it in projects. 

Also, the hard-coded use of `MavenPublishModule.publishAll` should instead use a `ModuleRef` instead.

maxstreese:
Hi @lefou and @chikei , do you still consider this PR actively worked on? If not I would like to attempt this (though it does feel a bit out of my depth).

@lefou regarding your last comment about (1) moving the implementations of `object MavenPublishModule` into a trait and (2) replace the call to `MavenPublishModule.publishAll` with a `ModuleRef`: If I look at [the sources](https://github.com/com-lihaoyi/mill/blob/0feb2b6deaada69fb0bf4154ce27353f6032478f/libs/javalib/src/mill/javalib/SonatypeCentralPublishModule.scala) of `SonatypeCentralPublishModule` (which @chikei said `MavenPublishModule` is in large part a copy of) then both of these points apply to that as well. Therefore I am wondering if adjusting things only here is really desirable? Or should a larger refactoring be done that factors out the commonalities between these two modules and implements both of your remarks?

lefou:
> Hi @lefou and @chikei , do you still consider this PR actively worked on? If not I would like to attempt this (though it does feel a bit out of my depth).

It would be nice to get this over the finish line. Since the base functionality already done, left work should be mostly mechanical editing. 

> @lefou regarding your last comment about (1) moving the implementations of `object MavenPublishModule` into a trait and (2) replace the call to `MavenPublishModule.publishAll` with a `ModuleRef`: If I look at [the sources](https://github.com/com-lihaoyi/mill/blob/0feb2b6deaada69fb0bf4154ce27353f6032478f/libs/javalib/src/mill/javalib/SonatypeCentralPublishModule.scala) of `SonatypeCentralPublishModule` (which @chikei said `MavenPublishModule` is in large part a copy of) then both of these points apply to that as well. Therefore I am wondering if adjusting things only here is really desirable? Or should a larger refactoring be done that factors out the commonalities between these two modules and implements both of your remarks?

Sharing common parts between both traits would be the best case, of course. Whether we do it right now or in a subsequent PR is both fine for me, as long as we get it merged rather soon.

PR #5604, opened 2025-07-25
Draft: Use custom JVM versions during Scala compilation, avoid forking for Java compilation

arturaz:
Adds support for compiling projects under different JVMs and JVM runtime options.

Previously:
- Compiling java sources with a different JVM/options worked, but a compiler was forked for each compilation, which meant that it was always JIT-cold.
- Specifying `jvmId` in a `ScalaModule` did not make the compiler run with that JVM. The option was ignored.

This PR fixes both of those issues.

To do that, it introduces an RPC framework to mill, located under `mill.rpc` namespace.

### RPC overview

`MillRpcClient.create` allows you to create an RPC client using a specific `MillRpcWireTransport`. A transport is a simple trait that provides a way to read and write protocol messages to the wire. RPC uses line-delimited JSON for exchanging messages.

Currently supported transports are:
- Using stdin/stdout.
- Using stdin/stdout of a subprocess.
- Using concurrent blocking queues (mostly used for testing).
- Using Java streams (for example, from a TCP connection).

Client, upon establishing a connection to the server, sends an `Initialize` message. Then it sends `ClientToServer` messages, which always return a response (which can be `Unit`). A server, when responding to a client's message, can also send a `ServerToClient` request to the client (for example, to obtain additional data from the client needed to fulfill the request). However, in the current implementation, a server cannot push messages to the client without the client initiating a request first (as the client implementation does not read from the transport unless it is waiting for a server's response).

### TCP server infrastructure

Part of the code has been extracted from `runner.client` and `runner.server` modules into `libs.daemon.client` and `libs.daemon.server` respectively.

This code handles:
- Client starting a server process if needed, or connecting to a running process if it's there.
- Initial handshake between the client and the server via the filesystem to exchange on which free port the server did bind.
- Server accepting connections on separate threads and the client establishing those connections.
- Multiplexing stdout/stderr from the server into a single TCP stream and demultiplexing it on the client side.
- Infrastructure for returning exit codes to the client and other conveniences.

### ZincWorker

Keeps the in-memory Java/Scala compilers and performs the actual compilations. There is one local `ZincWorker` in `JvmWorkerImpl` which is used if the compilation is performed without a custom JVM and custom JVM runtime options, and each `ZincWorkerRpcServer` also contains a `ZincWorker`.

### ZincWorkerRpcServer

This handles requests to perform Java/Scala compilations and Scaladoc generation. It uses `ZincWorker` internally to do that.

### ZincWorkerMain

Entry point for the TCP-based RPC server. `JvmWorkerImpl` spins up these as JVM subprocesses and then communicates with them via TCP to perform java/scala compilations and scaladoc generation.

Supports multiple clients, each client is handled in it's own thread.

--- 

Fixes https://github.com/com-lihaoyi/mill/issues/3982


libs/javalib/worker/src/mill/javalib/zinc/ZincWorkerRpcServer.scala
```diff
@@ -0,0 +1,224 @@
...
+  }
+  object ReporterMode {
+    case object NoReporter extends ReporterMode {
+      override def reportCachedProblems: Boolean = false
+    }
+
+    case class Reporter(reportCachedProblems: Boolean, maxErrors: Int) extends ReporterMode
+  }
+
+  sealed trait ClientToServer extends MillRpcMessage derives ReadWriter
```
lihaoyi, line 155: Now that we're on Scala 3, can these be `enum`s?

lihaoyi:
One thing I had a hard time figuring out from the PR: what messaging protocol are you using? Line-delimited JSON? length-prefixed JSON? Something else?

libs/javalib/worker/src/mill/javalib/zinc/ZincWorkerRpcServer.scala
```diff
@@ -0,0 +1,224 @@
...
+      override type Response = Boolean
+    }
+  }
+
+  sealed trait ServerToClient extends MillRpcMessage derives ReadWriter
+  object ServerToClient {
+    case class AcquireZincCompilerBridge(scalaVersion: String, scalaOrganization: String)
+        extends ServerToClient {
+      override type Response = ZincCompilerBridge.AcquireResult[os.Path]
+    }
```
lihaoyi: Could we move the downloading of the compiler bridge to `JvmWorkerImpl`, before we begin RPCing to `ZincWorker`? We already have all the necessary data (`scalaVersion` and `scalaOrganization`) pretty early, so moving it to `JvmWorkerImpl` would save us one back-and-forth and simplify the RPC data flow

libs/javalib/worker/src/mill/javalib/zinc/ZincWorkerRpcServer.scala
lihaoyi: Not sure if enums can take `type` members, but if not I think they should be replaceable with type parameters

libs/javalib/worker/src/mill/javalib/worker/JvmWorkerImpl.scala
```diff
@@ -444,623 +63,378 @@ class JvmWorkerImpl(
...
-      val addColorNever = !ctx.log.prompt.colored &&
-        compilers.scalac().scalaInstance().version().startsWith("3.") &&
-        !scalacOptions.exists(_.startsWith("-color:")) // might be too broad
-      if (addColorNever)
-        "-color:never" +: scalacOptions
-      else
-        scalacOptions
+    if (jOpts.runtime.options.isEmpty && javaHome.isEmpty) runWithLocalZinc()
+    else runWithSpawned(javaHome, jOpts.runtime, zincCtx, log) { worker =>
+      f(worker, jOpts.compiler)
```
lihaoyi: This branch is kind of confusing because it's asymmetric: `runWithLocalZinc` takes no args because it's a inner `def` and so can reference things lexically, while `runWithSpawned` needs to take args because it is a separate `def`. Can we make both of them work the same way so the symmetry between the two branches is more obvious?

libs/javalib/worker/src/mill/javalib/worker/JvmWorkerImpl.scala
```diff
@@ -444,623 +63,378 @@ class JvmWorkerImpl(
...
 
-    def analysisMap(f: VirtualFile): Optional[CompileAnalysis] = {
-      val analysisFile = f match {
-        case pathBased: PathBasedFile => analysisMap0.get(os.Path(pathBased.toPath))
-        case _ => None
-      }
-      analysisFile match {
-        case Some(zincPath) => fileAnalysisStore(zincPath).get().map(_.getAnalysis)
-        case None => Optional.empty[CompileAnalysis]
+      val api = new ZincWorkerApi {
```
lihaoyi: All the forwarding of `ZincWorkerApi` params to `zincLocalWorker` calls or `ZincWorkerRpcServer.ClientToServer` messages is pretty tedious. What if rather than making `ZincWorkerApi` a `trait`, we standardized on the `ZincWorkerRpcServer.ClientToServer` case classes as the way to pass data to both local and remote zinc workers? 

Then the entrypoint for `ZincWorker` can be a single method taking `ZincWorkerRpcServer.ClientToServer` and matching on the various cases, that can be re-used for both local and remote workers, without constantly having to convert between loose method parameters and bundled case class instances

core/rpc/src/mill/rpc/MillRpcRequestId.scala
```diff
@@ -0,0 +1,92 @@
...
+ * c0:s1
+ * // Client responds, server asks for even more data
+ * c0:s2
+ * // Client sends a request to server
+ * c0:s2:c0
+ * // Everyone responds, client sends a new request
+ * c1
+ * }}}
+ */
+case class MillRpcRequestId private (parts: Vector[MillRpcRequestId.Part]) {
```
lihaoyi: Do we even need a `RequestId` data type? As I understand it, when the client sends a request to the server,  any responses from the server must be related to the last-sent request from that client, until the server sends a `Response` message, so we do not need any metadata or ID to know which client request a particular server message is attached to

libs/javalib/worker/src/mill/javalib/worker/JvmWorkerImpl.scala
```diff
@@ -444,623 +63,378 @@ class JvmWorkerImpl(
...
+          ),
+          log,
+          (requestId, msg) =>
+            throw new IllegalStateException(
+              s"Server message handler is not ready to handle request $requestId: $msg"
+            )
+        )
+      ) { client =>
+        // Exchange the handler from the cached value.
+        client.withServerToClientHandler(handler)
```
lihaoyi: Since the call to `f(client)` blocks the current thread, what thread do the `handler` callbacks execute on? The current thread, a newly spawned one, or some shared threadpool?

libs/javalib/worker/src/mill/javalib/worker/JvmWorkerImpl.scala
```diff
@@ -444,623 +63,378 @@ class JvmWorkerImpl(
...
-    if oi.isPresent then oi.get().intValue()
-    else default
+    override def teardown(
+        key: SubprocessCacheKey,
+        client: MillRpcClient[
+          ZincWorkerRpcServer.ClientToServer,
+          ZincWorkerRpcServer.ServerToClient
+        ]
+    ): Unit = {
+      client.close()
```
lihaoyi: When you add support for shared server processes, you probably will need logic similar to what is currently in `RefCountedClassLoaderCache` to keep the server processes running until the number of active clients drops to zero. Perhaps we can take `RefCountedClassLoaderCache` and refactor out a more generic `RefCountedCache` that can shared the ref-counting logic between ref-counting classloaders and ref-counting subprocesses

core/api/daemon/src/mill/api/daemon/MillURLClassLoader.scala
```diff
@@ -19,9 +19,10 @@ class MillURLClassLoader(
   import MillURLClassLoader.*
   addOpenClassloader(label)
 
-  override def findClass(name: String): Class[?] =
+  override def findClass(name: String): Class[?] = {
     if (sharedPrefixes.exists(name.startsWith)) sharedLoader.loadClass(name)
     else super.findClass(name)
+  }
```
lihaoyi: Let's revert these unrelated changes (and also in `CompileProblemReporter.scala` and elsewhere) since the PR is already pretty large and these unrelated changes make it harder to scan and review

core/rpc/src/mill/rpc/MillRpcClientToServer.scala
```diff
@@ -0,0 +1,18 @@
...
+
+  /** Client is asking the server to do something. */
+  case Ask(id: MillRpcRequestId, data: Data)
+
+  /** Response to server's [[MillRpcServerToClient.Ask]] which either succeeded or failed. */
+  case Response(id: MillRpcRequestId, data: Either[RpcThrowable, Data])
+}
+
+object MillRpcClientToServer {
+  given reader[Data: Reader]: Reader[MillRpcClientToServer[Data]] = Reader.derived
```
lihaoyi: Can we use `enum MillRpcClientToServer[+Data] derives ReadWriter`?

core/rpc/src/mill/rpc/MillRpcMessage.scala
```diff
@@ -0,0 +1,18 @@
+package mill.rpc
+
+import pprint.TPrint
+import upickle.default.ReadWriter
+
+trait MillRpcMessage {
+  type Response
+
+  given responseRw: ReadWriter[Response] = compiletime.deferred
+  given responseTypeName: TPrint[Response] = compiletime.deferred
```
lihaoyi: I'm not familiar with `compiletime.deferred`. What does this mean?

core/rpc/src/mill/rpc/MillRpcRequestId.scala
```diff
@@ -0,0 +1,92 @@
...
+    val parts = str.split(":").iterator.map(Part.unapply).toVector
+    if (parts.contains(None)) Left(s"invalid request id: \"$str\" ($parts)")
+    else if (parts.isEmpty) Left(s"empty request id: \"$str\"")
+    else Right(MillRpcRequestId(parts.flatten))
+  }
+
+  given rw: upickle.default.ReadWriter[MillRpcRequestId] =
+    upickle.default.readwriter[String].bimap(
+      _.toString,
+      fromString(_) match {
```
lihaoyi: Would `fromString(_).fold(sys.error(_), identity)` work here?

core/rpc/src/mill/rpc/MillRpcRequestId.scala
lihaoyi: Even if we do find we want to keep this code, it feels over engineered. A single `Seq[Int]` should be enough to model the examples you show above, everything else can be derived from that (e.g. whether a particular `Int` is a client or server request is simply whether the index is odd or even. We can wrap the `Seq[Int]` in a `case class` with a nice `toString` or `upickle.default.ReadWriter` if we want it to look pretty in logs

arturaz:
> One thing I had a hard time figuring out from the PR: what messaging protocol are you using? Line-delimited JSON? length-prefixed JSON? Something else?

Line delimited json.

libs/javalib/worker/src/mill/javalib/zinc/ZincWorkerRpcServer.scala
arturaz: > Now that we're on Scala 3, can these be `enum`s?
> Not sure if enums can take `type` members, but if not I think they should be replaceable with type parameters

I don't think these can be enums: https://scastie.scala-lang.org/pG8bRIWVT72grDXdxGx13g

As for replaceable with type parameters, if you mean this:
```scala
enum Foo1[-Response] {
  case FooInt extends Foo1[Int]
}
```

I've tried using this pattern when I hit this problem:
```scala
          msg match {
            case ServerToClient.GiveMeExtraInfo(what) =>
              // Note the cast.
              s"Extra $what for $requestId".asInstanceOf[msg.Response]
          }
```

However, I did not find a way to express the needed patterns. Maybe it is possible, we can try going over that in a life session, perhaps?

core/rpc/src/mill/rpc/MillRpcRequestId.scala
arturaz: > Do we even need a RequestId data type? As I understand it, when the client sends a request to the server, any responses from the server must be related to the last-sent request from that client, until the server sends a Response message, so we do not need any metadata or ID to know which client request a particular server message is attached to

The original modeling for this type was done with these goals:
- When debugging, have an easy way to see how requests relate to each other.
- Support parallel requests in the future. You fire off c0, c1, c2 and then the server could have process them in parallel and respond to then out of order.

> e.g. whether a particular Int is a client or server request is simply whether the index is odd or even.

That is true in the current model, where server must wait for a client request to respond. However, in general case (in future) we might want to have RPC functionality where the server sends requests on it's own behalf, even without the client requesting the first message.

libs/javalib/worker/src/mill/javalib/worker/JvmWorkerImpl.scala
arturaz: Handler callbacks execute on the same blocking thread.

Here's how it works:
- client sends a request
- it starts reading from the transport, awaiting a response
- however, it can receive other things instead of response, like log/console messages or servers' asks
- if it gets a server ask, it runs the server to client message handler in a blocking way and once that gets back, resumes waiting for a response

core/rpc/src/mill/rpc/MillRpcRequestId.scala
lihaoyi: If we have multiple clients talking to the same server, then we can consider having every client having a separate connection to the server that talks back and forth in a single-threaded fashion, as clients will be cheap and if we want multiple concurrent requests we can spawn multiple clients. The only cost would be the TCP port, which should be plentiful enough we don't need to worry about it

core/rpc/src/mill/rpc/MillRpcClientToServer.scala
arturaz: That fails the compilation:

```
No given instance of type upickle.default.Writer[mill.rpc.MillRpcClientToServer[ClientToServer]] was found for a context parameter of method writeSerialized in trait MillRpcWireTransport.
I found:

    mill.rpc.MillRpcClientToServer.derived$ReadWriter[ClientToServer](
      upickle.default.ReadWriter.join[ClientToServer](
        upickle.default.superTypeReader[ClientToServer, V](
          /* missing */summon[scala.deriving.Mirror.ProductOf[ClientToServer]],
          ???, ???, ???),
      ???)
    )

But Failed to synthesize an instance of type scala.deriving.Mirror.ProductOf[ClientToServer]: trait MillRpcMessage is not a generic product because it is not a case class

where:    ClientToServer is a type in method create with bounds <: mill.rpc.MillRpcMessage
.

The following import might make progress towards fixing the problem:

  import upickle.default.ReadWriter.join

 [100:93]
 ```

This works, though.
```scala
enum MillRpcServerToClient[+Data] derives Reader, Writer
```

core/rpc/src/mill/rpc/MillRpcMessage.scala
arturaz: It defers the lookup of a given until you extend this trait.

So in this site it cannot summon the given because the type of Response isn't known, but once you override that in a concrete class, it can.

core/rpc/src/mill/rpc/MillRpcRequestId.scala
arturaz: It would, but that throws generic RuntimeException. I've changed it to `fromString(_).fold(err => throw IllegalArgumentException(err), identity)`

core/api/daemon/src/mill/api/daemon/MillURLClassLoader.scala
arturaz: Sorry, I didn't have a chance to clean it up, as it's still a draft and there's still work to be done on it.

captainvredina:
Worth checking out Zinc's classpath alignment...
To avoid runtime mismatches.

arturaz:
> Worth checking out Zinc's classpath alignment... To avoid runtime mismatches.

@captainvredina can you elaborate?

libs/javalib/worker/src/mill/javalib/worker/JvmWorkerImpl.scala
arturaz: I've extracted the common stuff into `mill.javalib.api.internal.{ZincCompileJava, ZincCompileMixed, ZincScaladocJar}` classes, but unfortunately there's still some juggling happening.

libs/javalib/worker/src/mill/javalib/zinc/ZincWorkerRpcServer.scala
arturaz: I don't think we should touch that logic now. I tried doing as little as possible with the actual compilation logic and this seems like another, unrelated, refactoring.

libs/javalib/worker/src/mill/javalib/worker/JvmWorkerImpl.scala
arturaz: Extracted as `RefCountedCache`

libs/javalib/src/mill/javalib/internal/ZincCompilerBridge.scala
```diff
@@ -0,0 +1,143 @@
...
+  /**
+   * Provides the compiler bridge.
+   *
+   * @tparam Data extra data to pass to the acquire function.
+   */
+  trait Acquire[-Data] {
+    def apply(scalaVersion: String, scalaOrganization: String, data: Data): AcquireResult[os.Path]
+  }
+
+  sealed trait AcquireResult[+Path] {
```
lihaoyi: Make it an enum?

libs/javalib/worker/src/mill/javalib/worker/JvmWorkerImpl.scala
```diff
@@ -1,1066 +1,293 @@
...
-           |  scalacOptions: ${scalacOptions.map("'" + _ + "'").mkString(" ")}
-           |  sources: ${sources.map("'" + _ + "'").mkString(" ")}
-           |  classpath: ${compileClasspath.map("'" + _ + "'").mkString(" ")}
-           |  output: ${classesDir}"""
-          .stripMargin
+    override def setup(
+        key: SubprocessCacheKey,
+        init: SubprocessCacheInitialize
+    ): MillRpcClient[ZincWorkerRpcServer.ClientToServer, ZincWorkerRpcServer.ServerToClient] = {
+      println(s"Spawning JVM worker subprocess: ${key.debugStr}")
```
lihaoyi: Remove the println

arturaz:
@lihaoyi current status with tests:

[kotlin tests](https://github.com/com-lihaoyi/mill/actions/runs/16850947479/job/47737171416?pr=5604) seem flaky, they passed previously and also pass on my machine.

[js tests](https://github.com/com-lihaoyi/mill/actions/runs/16850947479/job/47737171456?pr=5604) fail with some random node error, but they seem to do the same on `main` branch. And also my changes do not seem to interact with javascript support codepaths.

[windows tests](https://github.com/com-lihaoyi/mill/actions/runs/16850947479/job/47737219898?pr=5604) seem to fail because two tests are being run in parallel and they both try to clean the `moduleDir`, which results in directory sharing exception on windows.

https://github.com/com-lihaoyi/mill/blob/58d60b4f27235f38846229d05da36d226f653e02/libs/javalib/test/src/mill/javalib/LauncherTests.scala#L55-L57
https://github.com/com-lihaoyi/mill/blob/58d60b4f27235f38846229d05da36d226f653e02/libs/javalib/test/src/mill/javalib/LauncherTests.scala#L22-L23
https://github.com/com-lihaoyi/mill/blob/58d60b4f27235f38846229d05da36d226f653e02/testkit/src/mill/testkit/UnitTester.scala#L76-L78

Otherwise, I've successfully compiled mill's own source with `mill-assembly.jar` from this branch, so at least this smoke test passes. And Mockito import example triggered a realistic parallel compilation case which caught a locking bug, which has been fixed, so I'm reasonably confident that parallel compilation works properly.

I was wondering on whether it's possible to write a test to exercise the parallel computation code path, but we actually need something large to compile, so it would take enough time for the compilations to actually be parallel. Maybe you have any suggestions here?

lihaoyi:
For the windows problem, why is it causing problems now and not before? `mill.javalib.LauncherTests.assembly` is not a new test, and it seems to have been passing before

lihaoyi:
As far as I can tell, all the code in `LauncherTests` should run sequentially. Mill only parallelizes suites on a per-class level by default, so I'm not seeing where the parallelism in `LauncherTests` causing the failure is coming from

core/exec/src/mill/exec/Execution.scala
```diff
@@ -254,8 +254,14 @@ private[mill] case class Execution(
...
               } catch {
+                // Wrapping the fatal error in a non-fatal exception, so it would be caught by Scala's Future
+                // infrastructure, rather than silently terminating the future and leaving downstream Awaits hanging.
                 case e: Throwable if !scala.util.control.NonFatal(e) =>
-                  throw new Exception(e)
+                  val nonFatal = new Exception(s"fatal exception occurred: $e", e)
+                  // Set the stack trace of the non-fatal exception to the original exception's stack trace
+                  // as it actually indicates the location of the error.
+                  nonFatal.setStackTrace(e.getStackTrace)
+                  throw nonFatal
```
lefou, line 264: I don't think we should handle fatal exceptions any further (so deep in the call stack). E.g. allocating new objects after a fatal exception (e.g. a out of memory situation) can cause new exceptions, which will also hide the original cause.

lefou (commented):
I'd prefer to keep the unrelated changes small or zero, to avoid discussion of the unrelated changes. E.g. I don't like all those changes where you removed parentheses from Java method calls and overrides. They are initially defined _with_ parenthesis, so they should also be called with parentheses, even if the Scala compiler supports both styles.

core/exec/src/mill/exec/Execution.scala
arturaz: @lihaoyi should make a call on this, but just FYI "NoClassDefFoundError" is also considered a fatal exception and that definitely benefits from the `setStackTrace` we do here. In fact, we stumbled on this while developing this PR, which has lead to https://github.com/com-lihaoyi/mill/pull/5659.

lefou: Maybe, we handle "known good" candidates explicitly?

lihaoyi: I think this is fine for now. "Just let threads die when a fatal exception occurs" has never been practical advice, nor have we been following it, and it's been fine

Issue #5600, opened 2025-07-24
AndroidSdkModule.findLatestSdkManager misuses baseName


wingdeans:
Codepath for getting the latest Android sdkManager (when there is no `latest` directory) errors:
```
1 tasks failed
androidSdkModule0.installAndroidSdkComponents java.lang.UnsupportedOperationException: empty.maxBy
    scala.collection.IterableOnceOps.maxBy(IterableOnce.scala:1160)
    scala.collection.IterableOnceOps.maxBy$(IterableOnce.scala:1158)
    scala.collection.AbstractIterable.maxBy(Iterable.scala:935)
    mill.androidlib.AndroidSdkModule.findLatestSdkManager(AndroidSdkModule.scala:412)
    mill.androidlib.AndroidSdkModule.installAndroidSdkComponents$$anonfun$1$$anonfun$1(AndroidSdkModule.scala:274)
    mill.api.Task$Named.evaluate(Task.scala:328)
    mill.api.Task$Named.evaluate$(Task.scala:313)
    mill.api.Task$Computed.evaluate(Task.scala:339)
```
### fix
`p.baseName`should be `p.last` in order to match `major.minor` directories (like `19.0`)
https://github.com/com-lihaoyi/mill/blob/b252270a1e60392c1453564d98f88c7a4e813543/libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala#L407-L411

lihaoyi:
CC @vaslabs

vaslabs:
thanks, this did came up in @souvlakias recent vanilla tests, we'll prepare a fix soon

EDIT: this aligns with another task that we were looking at, to make mill android features self sufficient and auto-download needed android sdk dependencies, this issue will be fixed as a consequence, instead of doing a quick fix now, we'll have a PR soon

vaslabs:
@souvlakias [has reworked](https://github.com/com-lihaoyi/mill/pull/5608) a major part of interacting with Android Sdk, including making Mill download all the necessary dependencies, which should also resolve this issue.

PR #5608, opened 2025-07-25
Android: Auto-Download CmdlineTools

souvlakias:
## Motivation  
Previously, `cmdline-tools` had to be downloaded manually (either via Android Studio or from the official website). In CI, a plugin was used to handle this step. This PR automates the process entirely, removing the need for manual setup.  
This should also resolve https://github.com/com-lihaoyi/mill/issues/5600

## What this changes  
- Adds automatic downloading of `cmdline-tools` if not found.  
- The version to be downloaded is hardcoded, with the option for users to override it by providing their own.  

## TBC
- Update the documentation.  
- Add interactive license acceptance prompting for users.

### DIFF
--- a/libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala
+++ b/libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala
@@ -322,7 +322,7 @@ trait AndroidSdkModule extends Module {
     // Accept Licenses
     // TODO: Prompt the user to accept licenses interactively
     if (isCI)
-      os.proc("echo", "y").pipeTo(os.proc(sdkManagerExe.toString, "--licenses")).call()
+      os.proc("echo", "y\n" * 10).pipeTo(os.proc(sdkManagerExe.toString, "--licenses")).call()
     sdkManagerExe
   }
 
--- a/libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala
+++ b/libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala
@@ -72,6 +72,13 @@ trait AndroidSdkModule extends Module {
   def platformsVersion: T[String] = Task { "android-" + buildToolsVersion().split('.').head }
 
   /**
+   * Specifies the version of the Android Command Line Tools to be used.
+   */
+  def cmdlineToolsVersion: T[String] = Task {
+    Versions.cmdlineToolsVersion
+  }
+
+  /**
    * URL to download bundle tool, used for creating Android app bundles (AAB files).
    */
   def bundleToolUrl: T[String] = Task {
@@ -147,7 +154,7 @@ trait AndroidSdkModule extends Module {
    */
   def lintToolPath: T[PathRef] = Task {
     installAndroidSdkComponents()
-    PathRef(sdkPath().path / "cmdline-tools/latest/bin/lint")
+    PathRef(cmdlineToolsPath().path / "bin/lint")
   }
 
   /**
@@ -205,7 +212,7 @@ trait AndroidSdkModule extends Module {
    *  For more information refer to the official Android documentation [[https://developer.android.com/tools/avdmanager]]
    */
   def avdPath: T[PathRef] = Task {
-    PathRef(sdkPath().path / "cmdline-tools/latest/bin/avdmanager")
+    PathRef(cmdlineToolsPath().path / "bin/avdmanager")
   }
 
   /**
@@ -226,21 +233,12 @@ trait AndroidSdkModule extends Module {
   }
 
   /**
-   * Provides the path for the Android SDK Manager tool
-   *
-   * @return A task containing a [[PathRef]] pointing to the SDK directory.
-   */
-  def sdkManagerPath: T[PathRef] = Task {
-    PathRef(sdkPath().path / "cmdline-tools/latest/bin/sdkmanager")
-  }
-
-  /**
    * Provides the path for the r8 tool, used for code shrinking and optimization.
    *
    * @return A task containing a [[PathRef]] pointing to the r8 directory.
    */
   def r8Exe: T[PathRef] = Task {
-    PathRef(sdkPath().path / "cmdline-tools/latest/bin/r8")
+    PathRef(cmdlineToolsPath().path / "bin/r8")
   }
 
   def ndkPath: T[PathRef] = Task {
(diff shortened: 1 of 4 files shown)


libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala
```diff
@@ -263,6 +261,104 @@ trait AndroidSdkModule extends Module {
...
+      "GITHUB_ACTIONS",
+      "GITLAB_CI",
+      "BITBUCKET_PIPELINE",
+      "TEAMCITY_VERSION"
+    )
+    ciEnvironments.exists(env => sys.env.contains(env))
+  }
+
+  private def cmdlineToolsURL(versionLong: String): String = {
+    val osName = System.getProperty("os.name").toLowerCase
```
vaslabs: Might be worth checking if os lib provides this functionality

libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala
```diff
@@ -263,6 +261,104 @@ trait AndroidSdkModule extends Module {
...
+
+    s"https://dl.google.com/android/repository/commandlinetools-$platform-${versionLong}_latest.zip"
+  }
+
+  private def installCmdlineTools(
+      sdkPath: os.Path,
+      versionShort: String,
+      remoteReposInfo: os.Path
+  ) = {
+    object CmdlineToolsHelper {
```
vaslabs: I think there's no need for this nested object, can just do a tuple
`val (millAndroidCmdlineToolsVersion, millAndroidCmdlineToolsLongVersion) = "19.0->13114758"`

vaslabs: Alternatively , this can be a separate task (maybe just returning a named tuple) so "system" or "mill" versions stay separated from user's choice and also can be independently customised

libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala
```diff
@@ -263,6 +261,104 @@ trait AndroidSdkModule extends Module {
...
+      os.makeDir.all(extractTo)
+      os.unzip(downloaded, extractTo)
+
+      // Move the extracted tools to the version-specific directory
+      os.move(extractTo / "cmdline-tools", cmdlineToolsHelperPath)
+
+      // Accept Licenses
+      if (isCI)
+        os.proc(
+          "echo",
```
vaslabs: Doesn't it work with `yes` ?

mill-build/src/millbuild/Deps.scala
```diff
@@ -307,7 +307,7 @@ object Deps {
     val uiTooling = mvn"androidx.compose.ui:ui:1.7.6"
     val screenshotValidationJunitEngine =
       mvn"com.android.tools.screenshot:screenshot-validation-junit-engine:0.0.1-alpha09"
-
+    val cmdlineToolsVersion = "19.0"
```
vaslabs: This should be the version mill uses to do the initial setup but https://github.com/com-lihaoyi/mill/pull/5608/files#diff-02a26e3d4f6a4df457e3e14bc7da46644c3b6be0d7c95a8d04ff7b5eb53c4a17R298 is hardcoded

libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala
```diff
@@ -71,6 +71,13 @@ trait AndroidSdkModule extends Module {
    */
   def platformsVersion: T[String] = Task { "android-" + buildToolsVersion().split('.').head }
 
+  /**
+   * Specifies the version of the Android Command Line Tools to be used.
+   */
+  def cmdlineToolsVersion: T[String] = Task {
```
vaslabs, line 84: Adding to all my relevant comments, I would have also a `millCmdlineToolsVersion` to use it for the initial setup, instead of the nested object, maybe making it a private task

vaslabs:
Great improvements, left a few comments around managing versions

libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala
vaslabs: I checked, seems no, and I found in other places it's either this ir sys.props.get.

I think it's safer to do `sys.props.get("os.name").map(_.toLowerCase)` and then use `contains`

libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala
souvlakias: I used `echo` so it works for windows too

vaslabs:
One little nitpick:

if for example `/home/vnicolaou/Android/Sdk/cmdline-tools/19.0/` exists but it is empty the setup doesn't go through and the android tasks fail. 

I think we should at least cover this scenario by a checksum or at least check `/home/vnicolaou/Android/Sdk/cmdline-tools/19.0/bin/sdkmanager` exists and if not let the user know and recommending either removing the directory and retrying with Mill or fixing it manually with android studio.

For a checksum based approach, we could do it on a subsequent PR using [the repository info](https://dl.google.com/android/repository/repository2-3.xml)
```
<remotePackage path="cmdline-tools;19.0">
<type-details xsi:type="generic:genericDetailsType"/>
<revision>
<major>19</major>
<minor>0</minor>
</revision>
<display-name>Android SDK Command-line Tools</display-name>
<uses-license ref="android-sdk-license"/>
<channelRef ref="channel-0"/>
<archives>
<archive>
<complete>
<size>164760899</size>
<checksum type="sha1">5fdcc763663eefb86a5b8879697aa6088b041e70</checksum>
<url>commandlinetools-linux-13114758_latest.zip</url>
```

souvlakias:
> One little nitpick:
> 
> if for example `/home/vnicolaou/Android/Sdk/cmdline-tools/19.0/` exists but it is empty the setup doesn't go through and the android tasks fail.
> 
> I think we should at least cover this scenario by a checksum or at least check `/home/vnicolaou/Android/Sdk/cmdline-tools/19.0/bin/sdkmanager` exists and if not let the user know and recommending either removing the directory and retrying with Mill or fixing it manually with android studio.
> 
> For a checksum based approach, we could do it on a subsequent PR using [the repository info](https://dl.google.com/android/repository/repository2-3.xml)
> 
> ```
> <remotePackage path="cmdline-tools;19.0">
> <type-details xsi:type="generic:genericDetailsType"/>
> <revision>
> <major>19</major>
> <minor>0</minor>
> </revision>
> <display-name>Android SDK Command-line Tools</display-name>
> <uses-license ref="android-sdk-license"/>
> <channelRef ref="channel-0"/>
> <archives>
> <archive>
> <complete>
> <size>164760899</size>
> <checksum type="sha1">5fdcc763663eefb86a5b8879697aa6088b041e70</checksum>
> <url>commandlinetools-linux-13114758_latest.zip</url>
> ```

image

PR #5614, opened 2025-07-26
[WIP] Add generator for Eclipse JDT project files

thahnen:
> [!WARNING]  
> This is a first push of my changes for the Eclipse JDT project files generator
> I had no time yet to think about good commit messages or this PR message, will be changed later

References #3372 regarding the generator for Eclipse JDT project files (Java projects only)


thahnen:
@lihaoyi may I ask for a first round of review from you?

This is currently in good shape. I will add **integration tests** and **documentation** in the next couple of days.

I have one question, tho: Since this should only be enabled for Java Modules, I was thinking about implementing something similar to [this](https://github.com/com-lihaoyi/mill/blob/main/core/api/daemon/src/mill/api/daemon/internal/idea/GenIdeaModuleApi.scala#L5) for IntelliJ. But I'm not sure how to plug it in so that it is disabled for Scala / Kotlin, etc.
For now, it will not fail when it tries to generate Eclipse JDT Project files for content other than Java. Either this could be solved by implementing something similar to IntelliJ but more restrictive, or adding this to the docs.

I'm also not sure, but for now I would discard [THIS](https://github.com/thahnen/mill/actions/runs/16627265337/job/47047532482) failure. Of course this is a new API and I guess it fails therefore.

core/api/daemon/src/mill/api/daemon/internal/JavaModuleApi.scala
```diff
@@ -38,6 +39,11 @@ trait JavaModuleApi extends ModuleApi with GenIdeaModuleApi {
    */
   private[mill] def genIdeaInternal: () => GenIdeaInternalApi
 
+  /**
+   *  Internal access to some GenEclipse helper tasks. These are used when constructing the necessary information to
+   *  create a resolved module. This in turn will be used later for creating the actual Eclipse JDT project files!
+   */
+  private[mill] def genEclipseInternal: () => GenEclipseInternalApi
```
lihaoyi: I think we can make MIMA happy by making these return `() => null` by default, and overriding it in the actual implementation

runner/eclipse/src/mill/eclipse/GenEclipseImpl.scala
```diff
@@ -0,0 +1,435 @@
...
+          (rootModule.moduleDirJava ->
+            JavaModuleDto(evaluator, api, mutable.Set.empty[JavaModuleApi]))
+      case _ =>
+    }
+    iterateModuleChildren(
+      aggregatedJavaModules,
+      evaluator,
+      rootModule.moduleDirJava,
+      rootModule.moduleDirectChildren
+    )
```
lihaoyi: This interleave traversal and processing of the `JavaModuleApi` objects is kind of confusing. We can do something similar to what we did in `GenIdeaImpl`:

- First, traverse all modules recursively
```scala
  def transitiveModules(module: ModuleApi): Seq[ModuleApi] = {
    Seq(module) ++ module.moduleDirectChildren.flatMap(transitiveModules)
  }
  transitiveModules(rootModule)
```

- Then, filter out the modules we care about. We can also filter out Scala and Kotlin modules here since we don't want them, and perform whatever processing we need in a `collect` call
```
.collect{ case api: JavaModuleApi if !api.isInstanceOf[ScalaModuleApi] && !api.isInstanceOf[KotlinModuleApi] => 
  ... fo something with api... 
}
```

runner/eclipse/src/mill/eclipse/GenEclipseImpl.scala
```diff
@@ -0,0 +1,435 @@
...
+  ExecutionResultsApi,
+  JavaModuleApi,
+  ModuleApi,
+  TaskApi,
+  TestModuleApi
+}
+
+import java.nio.file.{Files, Path, Paths}
+import scala.collection.mutable
+
```
lihaoyi, line 18: Could we have some Javadoc on the class quickly summarizing what kinds files it generates and what format they are in, with a link to the eclipse online documentation for the filesystem format?

lihaoyi:
For documentation, a section under `Installation and IDE Setup / IDE Support` for `Eclipse` with instructions for CLI usage and screenshots of how to import the project would work. I assume this implementation won't provide assistance in `build.mill` files since those are written in Scala, which eclipse doesn't support, so we should mention that in the docs that it only works on `JavaModule`s and `.java` source files

For automated testing, an integration test in the same style as `./mill integration.ide[gen-idea].packaged.daemon'` should suffice. It would run `mill mill.eclipse/` on a few example `JavaModules` and make sure their source folders, internal dependencies, and external dependencies are all wired up correctly in the generated files.

The PR description should also include screenshots of manual testing, with you using `mill.eclipse/` on some non-trivial open-source Java project, loading it into eclipse, and showing that navigation etc. works. I'll run through that workflow myself before merging just as a sanity check.

Of course, you should also test this on the project you want to use it on to make sure it satisfies your own requirements

runner/eclipse/src/mill/eclipse/GenEclipseImpl.scala
thahnen: Done

runner/eclipse/src/mill/eclipse/GenEclipseImpl.scala
thahnen: Done, I agree that the first implementation was a bit too complicated.

core/api/daemon/src/mill/api/daemon/internal/JavaModuleApi.scala
thahnen: This did not work, it complained about `null` not working here. Additionally I disagree with this MIMA tool here as I'd like to keep it on par with the IntelliJ implementation. I have to say I see the purpose of this tool only to a certain degree since it checks for binary compatibility but would fail on API changes anyway if I'm not mistaken.

If there is another way, please point it out (I'm not familiar with Scala enough to build a working method descriptor / signature here that satisfies the build) or otherwise, I would keep it this way for now.

thahnen:
@lihaoyi the user facing documentation I added alongside screenshots, I will add integration tests next.

For now I relied mostly on the examples when manually testing, can I copy them (in the IT build, not on disk) into the integration tests and re-use them there?

core/api/daemon/src/mill/api/daemon/internal/JavaModuleApi.scala
lihaoyi: Can you paste the error message here using `null? It seems like it should work

lihaoyi:
@thahnen for integration tests, you should try and minimize a small example project to use in your test. Maybe `example/javalib/basic/3-multi-module` and `example/javalib/basic/4-compat-modules` so we exercise both projects with the inter-module dependencies as well as projects with the `MavenModule` source layout

core/api/daemon/src/mill/api/daemon/internal/JavaModuleApi.scala
thahnen: When changing the signature to `private[mill] def genEclipseInternal: () => Null` in `JavaModuleApi` and calling something simple like

> ./mill "integration.ide[gen-eclipse].packaged.daemon"

I get the following errors:

> [5715] [error] -- [E007] Type Mismatch Error: /Users/thahnen/VCS/thahnen/mill/runner/eclipse/src/mill/eclipse/GenEclipseImpl.scala:96:52 
[5715] [error] 96 |      val moduleTask = dto.module.genEclipseInternal().genEclipseModuleInformation()
[5715] [error]    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[5715] [error]    |Found:    Null
[5715] [error]    |Required: ?{ genEclipseModuleInformation: ? }
[5715] [error]    |Note that implicit conversions were not tried because the result of an implicit conversion
[5715] [error]    |must be more specific than ?{ genEclipseModuleInformation: [applied to () returning <?>] }
[5715] [error]    |
[5715] [error]    | longer explanation available when compiling with `-explain`
[5715] [error] -- [E007] Type Mismatch Error: /Users/thahnen/VCS/thahnen/mill/runner/eclipse/src/mill/eclipse/GenEclipseImpl.scala:108:57 
[5715] [error] 108 |        sourceSetModuleTasks += module.genEclipseInternal().genEclipseModuleInformation()
[5715] [error]     |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[5715] [error]     |Found:    Null
[5715] [error]     |Required: ?{ genEclipseModuleInformation: ? }
[5715] [error]     |Note that implicit conversions were not tried because the result of an implicit conversion
[5715] [error]     |must be more specific than ?{ genEclipseModuleInformation: [applied to () returning <?>] }
[5715] [error]     |
[5715] [error]     | longer explanation available when compiling with `-explain`
[5715] [error] two errors found

lihaoyi: can you do `null.asInstanceOf[GenEclipseInternalApi]`?

thahnen: When changing the signature to `private[mill] def genEclipseInternal: () => GenEclipseInternalApi = null` there is no error, running

> ./mill -i __.mimaReportBinaryIssues

afterwards also shows no errors. I'll push for now and if the CI differs from the local Mima call, I don't know either ^^

lihaoyi:
I tried this out, seems to work!
 
 @thahnen one thing I noticed was that the `build.mill` file is not visible within Eclipse at all. Is there some way to make eclipse display it in the project explorer, even as an un-highlighted text file? I imagine being able to edit the `build.mill` config within an IDE would be a common ask from users. Maybe we could generate a synthetic project at the root of the repository that contains that file?

lihaoyi:
Also, I noticed that source files get duplicated in the project view: once under the source root, and once under the generic filesystem tree. Is that expected for how the Eclipse UI works? Or is there some way to avoid the duplication?

Screenshot 2025-08-04 at 8 45 50 AM

thahnen:
Hi @lihaoyi,

regarding the point of duplication: This is expected since this is due to the Eclipse UI for that view (Project Explorer).
It is possible to filter out these duplicated files by configuring it manually or to use the Java-specific "Package Explorer" view that does not duplicate the resources. It can happen that it shows the folders twice, I'm not sure what is the reason for this then happening once or not at all.

The duplication is due to the fact that that in the Project Explorer the directory is once kept as a source set (that is why there is this little icon above the folder) and once as a lame folder. I think this is a useless duplication but hey, who am I to judge ^^
In case you just have a `src` folder and not the nested Maven structure, it is not shown twice in most cases as Eclipse seems to be smart enough in that case to not show the source set again as a separate folder.

Regarding the point of having the projects either link to the `build.mill` file with a linked resource or having a separate synthetic project for it, I think the latter one makes more sense with the following cases:

- in case there will be a Eclipse project created in that folder anyway (e.g. for the `integration/ide/gen-eclipse/resources/simple-java-project`, don't create one.
- in case there will be no project created in that folder, create a synthetic one with only a simplified `.project` file - as it won't contain Java sources most of it does not need to be created

lihaoyi:
Your approach to include the build.mill sounds reasonable

thahnen:
Hey @lihaoyi,

I'm done. Feel free to validate the PR and provide one last round of feedback!
The instructions in the PR description should be sufficient I hope.

Best,
Tobias

lihaoyi:
@thahnen I have merged this and tagged `1.0.2-23-d3f2ba` for you to try out

Issue #5430, opened 2025-07-02
Pass process environment variables to the integration tests


arturaz:
Currently mill's integration tests do not get all of the environment variables from the parent process.

This leads to `MY_ENV_VAR=foo ./mill integration.feature[foo].local.nodaemon.testForked` to not have access to `MY_ENV_VAR`, unless explicitly exposed via `forkEnv` .
https://github.com/com-lihaoyi/mill/blob/4540af45ce5eb7e702e0301d1ceec02271011224/integration/package.mill#L51-L63

This first reared it's head in https://github.com/com-lihaoyi/mill/pull/5292 and now again in [#5425](https://github.com/com-lihaoyi/mill/pull/5425/files#diff-eb2c0efd777821140103facd38b418599e0bba8e7f80f6828cc00a744dec9689).

This can lead to subtle bugs. For example, I'm not sure the `CI` flag is ever actually passed to the test here:
https://github.com/com-lihaoyi/mill/blob/4540af45ce5eb7e702e0301d1ceec02271011224/integration/invalidation/watch-source-input/src/WatchSourceInputTests.scala#L24

I suggest forwarding all env variables, as we currently do with `MILL_TESTS_BSP_UPDATE_SNAPSHOTS`. Is there any downside to that?

lefou:
I think, we want to control the environment for integration tests from Mill, and having an explicit and reproducible setup that doesn't change much based on the host system, is what we should aim for. Of course, this comes at a prices in terms of flexibility. There is probably some middle ground, but we should outline and discuss the requirements and the implications.

PR #5617, opened 2025-07-28
Propagate environment variables to run and test unless `def propagateEnv = false`

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/5430, and would be necessary to make `UTEST_UPDATE_GOLDEN_TESTS` work smoothly

Previously, we relied on the default `propagateEnv = true`, but that doesn't work well because it picks up the Mill daemon process's env vars which do not get updated when you re-run the launcher. This PR explicitly sets `propagateEnv = false`, and passes in the `Task.env` explicitly to `Jvm.callProcess` when the `RunModule#propagateEnv` task is configured to be true (the default). If `RunModule#propagateEnv` is `false`, then we do not propagate any environment variables from the Mill process to the subprocess

Covered by a new `integration/feature/env-vars/` test that fails on main and passes on this PR

### DIFF
--- a/integration/feature/env-vars/src/MillOptsTests.scala
+++ b/integration/feature/env-vars/src/MillOptsTests.scala
@@ -1,7 +1,6 @@
 package mill.integration
 
-import mill.constants.Util
-import mill.testkit.{IntegrationTester, UtestIntegrationTestSuite}
+import mill.testkit.UtestIntegrationTestSuite
 import utest.*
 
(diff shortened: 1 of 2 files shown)


arturaz:
Is all of the passing in https://github.com/com-lihaoyi/mill/blob/ee74831f8f48ff3599e65c4e6a058b76786118ba/integration/package.mill#L96 still needed?

Issue #5616, opened 2025-07-27
Android Compose example does not build a new apk when sources change


lavrov:
Mill version: 1.0.1
Example project: `example/androidlib/kotlin/2-compose`.

It works well on the first (clean) run. Subsequent execution of `./mill app.androidInstall`, provided modified the app sources, does not update the app.

lihaoyi:
CC @vaslabs

vaslabs:
issue is due to a failure in d8 the second time, complaining about duplicate classes on the second run, although I can't see at the moment where these are being duplicated. Happens on the java example too.

```
/home/vnicolaou/Android/Sdk/build-tools/35.0.0/d8 --debug --output /home/vnicolaou/mill/example/androidlib/java/1-hello-world/out/app/androidD8Dex.dest --lib /home/vnicolaou/Android/Sdk/platforms/android-35/android.jar --min-api 19 --main-dex-rules /home/vnicolaou/mill/example/androidlib/java/1-hello-world/out/app/androidD8Dex.dest/proguard-rules.pro /home/vnicolaou/mill/example/androidlib/java/1-hello-world/out/app/compile.dest/classes/com/helloworld/app/MainActivity.class /home/vnicolaou/mill/example/androidlib/java/1-hello-world/out/app/compile.dest/classes/com/helloworld/SampleLogic.class /home/vnicolaou/mill/example/androidlib/java/1-hello-world/out/app/androidCompiledRClasses.dest/classes/com/helloworld/app/R$color.class /home/vnicolaou/mill/example/androidlib/java/1-hello-world/out/app/androidCompiledRClasses.dest/classes/com/helloworld/app/R$string.class /home/vnicolaou/mill/example/androidlib/java/1-hello-world/out/app/androidCompiledRClasses.dest/classes/com/helloworld/app/R.class
Error in /home/vnicolaou/mill/example/androidlib/java/1-hello-world/out/app/androidCompiledRClasses.dest/classes/com/helloworld/app/R.class:
Type com.helloworld.app.R is defined multiple times: /home/vnicolaou/mill/example/androidlib/java/1-hello-world/out/app/androidCompiledRClasses.dest/classes/com/helloworld/app/R.class, /home/vnicolaou/mill/example/androidlib/java/1-hello-world/out/app/androidCompiledRClasses.dest/classes/com/helloworld/app/R.class
Compilation failed
[...]

PR #5623, opened 2025-07-29
Fix android install

vaslabs:
Fixes caches not being invalidated for androidDex when source code changes

ref: https://github.com/com-lihaoyi/mill/issues/5616

## Explanation

- androidD8Dex was just returning the cli arguments with the paths as strings, thus while it was executed, androidDex was not, because the result from `androidD8Dex` was identical even after source changes. After adding the list of PathRef of the compiled classes the androidDex was executed on altering source files as in turn, compiled files were changed and detection is propagated by path ref hashing.
- androidDex was forwarding the PathRef of `androidD8Dex` which was being created on an empty directory, thus the PathRef signature was identical. Recreating it gives a fresh hash.

## Fix demo

https://github.com/user-attachments/assets/76a85669-c62b-43a7-85e4-5931f2f53b6d


vaslabs:
this is ready for review

lihaoyi:
Thanks @vaslabs !

PR #5628, opened 2025-07-30
Blog Post: Supporting android architectures with Hilt

vaslabs:
Blog article for our experience at supporting Hilt for Android.


vaslabs:
Maybe I should put a diagram to show relationships of 
AsmAdapter->Hilt->Dagger->KSP->Generated sources

EDIT: added diagram

lihaoyi:
I think this is a good start, but needs a bit more work before we publish it.

1. This post feels like it should be a follow-on post to one that explores the Android tooling ecosystem at a higher level
    - Should we first have a "Building Android Apps with Mill" post or similar, that we can link to at the start of this one to set the context for people who are curious?
    - Ideally we want a post that a mildly-curious Android dev would be interested in reading, before we dive deep into technical details like Hilt/Dagger code generation
    - After an initial "overview" post, we can have this and other posts diving into specific details, and link them all together so curious readers have a pathway to digging deeper or broader depending on what they want

2. Perhaps the biggest problem is that the post drops the reader into the weeds too quickly, and does not properly motivate the work done: 

    - Why do we need to get Hilt working? Because the Android Architecture examples need it
    - Why do we care about the Android Architecture examples? Because they are a representative set of small Android applications
    - Why do we care about the representative set of small Android applications? Because we want to let people build Android apps with Mill, and so we need to make sure it works for common use cases
    - Why do we want people to be able to build Android apps with Mill? Because their only option is Gradle, which can be slow and confusing, and we think that Mill can provide an easier-to-use option since it already supports non-Android Java development
    - This should all be in the opening and introductory paragraphs of this post, so the reader can follow along why what we are describing is interesting, and get meaningfully invested in the achievements and successes later in the post

3. The diagram in `Connecting Everything Together` is good, but is not sufficiently set up earlier in the post
    - Perhaps we can show this diagram at the _start_ of the post, contrast it with the equivalent diagram for a normal Kotlin/Java module, and then use that to tell the reader which parts of the pipeline we are trying to fill in?
    - We could show further steps in the pipeline as well, e.g. `dex`, `d8`, `r8`, `proguard`, `android-sdk`, `cli-tools`, etc., 
    - At a glance the user would able to see "here is why an Android toolchain is more complicated than a normal Java toolchain", as well as "here is the highlighted part of the Android toolchain that this blog post is talking about

4. We need to celebrate the results of this more explicitly. 
    - For example, after the 5 technical achievements in the `Conclusion`, did that increase the proportion of the Android Architecture examples we were able to support from x/y to x+n/y? Are there examples apart from the TODO app that benefitted from this feature?
    - Can we add some interesting screenshots of the examples we can now support that we didn't support before? Screenshots are always a good hook to show "something works", and should help get Android developers interested more than dry listings of internal tools
    - In the end readers may be Android _devs_, but they won't be Android _tooling devs_, so we need to show off results that non-tooling-experts are able to understand and appreciate

5. We need a more explicit call-to-action
    - Linking to `android/hilt-sample.html` is a start, but it's too messy to attract people, even if cleaned up it is still a pretty "big" example/documentation page that mildly-interested people are unlikely to try it out 
    - Can we inline the relevant `curl`, `chmod +x mill`, and `./mill` commands in a code block in the blog post, along with any necessary installation/preparation steps, so people have something they can try immediately?
    - We can keep the link to `android/hilt-sample.html`, but hopefully our inline example would be enough to get people interested before they dive into more details

vaslabs:
Thanks for the review, I will take some time to adjust

vaslabs:
I did a re-write, one thing waiting at least from my side:

I've used `echo "//| mill-version: 1.0.2-3-e42a40" > build.mill`

because I want to wait for the release containing the (https://github.com/com-lihaoyi/mill/pull/5608)[0 pre-requisite work to be released]

Meanwhile, I'd be happy for another review!

EDIT: It can also wait for an initial overview post, I started this because it was the most advanced bit to write about

lihaoyi:
@vaslabs I took a pass editing it, maybe you can take a look see if there's more stuff to improve. 

I think the main content is in place, but still needs a bit more polish and cleanup before publishing

vaslabs:
thanks! will have another go. I wonder if some of the content is more fitting for the documentation pages

vaslabs:
I'll also start a more introductory post, I have also a better idea of the preferred style now and I guess there's no rush to merge/publish this anyway

lihaoyi:
Sounds good!

website/blog/modules/ROOT/pages/14-android-mill-support-for-hilt.adoc
```diff
@@ -0,0 +1,266 @@
...
+
+Install mill
+
+[source,bash]
+----
+curl -L https://repo1.maven.org/maven2/com/lihaoyi/mill-dist/1.0.2/mill-dist-1.0.2-mill.sh -o mill
+chmod +x mill
+echo "//| mill-version: 1.0.2-3-e42a40" > build.mill
+./mill version
+----
```
lefou:
```suggestion
[source,console]
----
> curl -L https://repo1.maven.org/maven2/com/lihaoyi/mill-dist/1.0.2/mill-dist-1.0.2-mill.sh -o mill
> chmod +x mill
> echo "//| mill-version: 1.0.2-3-e42a40" > build.mill
> ./mill version
----
```

website/blog/modules/ROOT/pages/14-android-mill-support-for-hilt.adoc
```diff
@@ -0,0 +1,266 @@
...
+
+Configure the mill build
+
+[source,bash]
+----
+curl https://raw.githubusercontent.com/com-lihaoyi/mill/6351d7f3a29dd272c9393f690a3eb82ffa2b4f41/example/thirdparty/androidtodo/build.mill >>build.mill
+----
```
lefou:
```suggestion
[source,console]
----
> curl https://raw.githubusercontent.com/com-lihaoyi/mill/6351d7f3a29dd272c9393f690a3eb82ffa2b4f41/example/thirdparty/androidtodo/build.mill >>build.mill
----
```

website/blog/modules/ROOT/pages/14-android-mill-support-for-hilt.adoc
```diff
@@ -0,0 +1,266 @@
...
+
+Start the emulator and run the app
+[source,bash]
+----
+./mill show app.createAndroidVirtualDevice
+./mill show app.startAndroidEmulator
+./mill show app.androidInstall
+./mill show app.androidRun --activity com.example.android.architecture.blueprints.todoapp.TodoActivity
+----
```
lefou:
```suggestion

[source,console]
----
> ./mill show app.createAndroidVirtualDevice
> ./mill show app.startAndroidEmulator
> ./mill show app.androidInstall
> ./mill show app.androidRun --activity com.example.android.architecture.blueprints.todoapp.TodoActivity
----
```

website/blog/modules/ROOT/pages/14-android-mill-support-for-hilt.adoc
```diff
@@ -0,0 +1,266 @@
...
+
+Run the instrumented tests and watch the app being tested inside the emulator:
+
+[source,bash]
+----
+./mill app.androidTest
+----
```
lefou:
```suggestion
[source,console]
----
> ./mill app.androidTest
----
```

Issue #5654, opened 2025-08-05
Android integration issues


vaslabs:
I'm adding this issue to track minor bugs/missing features while I'm experimenting with various apps.

Issues found so far:

- Android manifest merge fails if `xmlns:android="http://schemas.android.com/apk/res/android"` property already exists

example
```xml
<?xml version="1.0" encoding="utf-8"?>
<manifest xmlns:android="http://schemas.android.com/apk/res/android">

</manifest>
```

- generatedSourcesWithKSP fails if android APIs (from android.jar) are mentioned -> android.jar is also missing from the classpath
e.g. these are workarounds
```
  override def kspClasspath: T[Seq[PathRef]] = super.kspClasspath() ++ androidSdkModule().androidLibsClasspaths()

  override def compileClasspath: T[Seq[PathRef]] = 
    super.compileClasspath() ++ androidSdkModule().androidLibsClasspaths()
```
- Android manifest merge cache does not seem to be invalidated when the manifest file changes

vaslabs:
I'm specifically testing this project https://github.com/NicosNicolaou16/Pokedex_Compose_Multi_Module which uses hilt, latest compose stack and multiple submodules

my build.mill so far is (only database is properly setup):

```scala
//| mill-version: 1.0.2
package build

import mill.*
import androidlib.*
import kotlinlib.*
import hilt.AndroidHiltSupport
import mill.api.ModuleRef
import mill.javalib.JavaModule

object androidSdkModule0 extends AndroidSdkModule {
  def buildToolsVersion = "36.0.0"
}

object database extends AndroidHiltCompose {
  override def androidNamespace: String = "com.nicos.database"

  override def mvnDeps: T[Seq[Dep]] = super.mvnDeps() ++ Seq(
    mvn"androidx.core:core-ktx:1.16.0",
    mvn"androidx.appcompat:appcompat:1.7.1",
    mvn"com.google.android.material:material:1.12.0",
    mvn"androidx.room:room-runtime:2.7.2",
    mvn"androidx.room:room-ktx:2.7.2",
    mvn"com.google.code.gson:gson:2.13.1",
    mvn"org.jetbrains.kotlinx:kotlinx-coroutines-core:1.10.2",
    mvn"org.jetbrains.kotlinx:kotlinx-coroutines-android:1.10.2",
  )

  override def kotlinSymbolProcessors: T[Seq[Dep]] = super.kotlinSymbolProcessors() ++ Seq(
[...]

PR #5662, opened 2025-08-06
Android: fix androidCompiledModuleResources

souvlakias:
Fixes some issues as mentioned in https://github.com/com-lihaoyi/mill/issues/5654
> - androidCompiledModuleResources fail with multiple submodules
> - Resources can be passed as a Jar to avoid submodule conflicts PR
> - generatedSourcesWithKSP fails if android APIs (from android.jar) are mentioned -> android.jar is also missing from the classpath in the case of submodules (AndroidModule)


vaslabs:
This should be ready

PR #5674, opened 2025-08-07
Android: Add `kspDependencyResolver`

souvlakias:
Adds a separate resolver for kotlin/java compiler plugins as mentioned in https://github.com/com-lihaoyi/mill/issues/5654#issuecomment-3155517981


vaslabs:
this should also be ready

PR #5677, opened 2025-08-08
Android: Proguard issues fixes

souvlakias:
Fixes an issue of Proguard usage as mentioned in https://github.com/com-lihaoyi/mill/issues/5654.

> Proguard rules for Hilt need to be autogenerated along with initialisers in the android manifest provider section to avoid R8 shrinking "nullifying" the constructors

#### EDIT:
This PR does not resolve the cache invalidation issue after all 
> Proguard file changes do not invalidate androidDex/androidApk caches  

The temporary fix we initially considered would have required API changes, which are not desired.  
We will instead follow up with a PR that fixes the issue without altering the APIs.


example/androidlib/java/5-R8/build.mill
```diff
@@ -45,10 +45,16 @@ object app extends AndroidR8AppModule { // <2>
...
+    AndroidBuildTypeSettings(
+      isMinifyEnabled = true,
+      isShrinkEnabled = true,
+      proguardFiles = ProguardFiles(
+        defaultProguardFile = Some("proguard-android-optimize.txt")
+      )
+    ).withProguardLocalFiles(
       Seq(
-        moduleDir / "proguard-rules.pro"
+        PathRef(moduleDir / "proguard-rules.pro")
```
vaslabs: doesn't it work if you make `withProguardLocalFiles` pass the PathRef instead so the API doesn't change?

libs/androidlib/src/mill/androidlib/AndroidBuildTypeSettings.scala
```diff
@@ -9,14 +9,15 @@ import mill.api.JsonFormatters.pathReadWrite
...
+ * Needs to be called from a `Task.Input` if local Proguard files are used.
  */
 case class AndroidBuildTypeSettings(
     isMinifyEnabled: Boolean = false,
     isShrinkEnabled: Boolean = false,
     enableDesugaring: Boolean = true,
     proguardFiles: ProguardFiles = ProguardFiles()
 ) {
-  def withProguardLocalFiles(localFiles: Seq[os.Path]): AndroidBuildTypeSettings =
+  def withProguardLocalFiles(localFiles: Seq[PathRef]): AndroidBuildTypeSettings =
```
vaslabs: I think you can keep the os.Path here and do `localFiles.map(PathRef(_))` below

libs/androidlib/src/mill/androidlib/AndroidAppModule.scala
```diff
@@ -800,6 +800,28 @@ trait AndroidAppModule extends AndroidModule { outer =>
 
   }
 
+  def knownProguardRules: T[String] = Task {
+    // TODO need also collect rules from other modules,
```
vaslabs: this TODO should change (because Android lib modules exist now) - to pick proguard files from [[moduleDeps]] too

example/androidlib/java/5-R8/build.mill
```diff
@@ -75,7 +69,7 @@ object app extends AndroidR8AppModule { // <2>
       false
     }
 
-    override def androidReleaseSettings: T[AndroidBuildTypeSettings] = Task.Input {
+    override def androidReleaseSettings: T[AndroidBuildTypeSettings] = Task {
```
lefou: This change will result ins missing change detection for `test-proguard-rules.pro` file.

Best, it so make that file a 

```scala
def testProgurardRulesPro = Task.Source("test-proguard-rules.pro")
```

and use it below.

lefou: Same in the other locations.

libs/androidlib/src/mill/androidlib/AndroidModule.scala
```diff
@@ -141,6 +141,42 @@ trait AndroidModule extends JavaModule {
     }
   }
 
+  def providerProguardConfigRules: T[Seq[String]] = Task {
```
```suggestion
  def androidProviderProguardConfigRules: T[Seq[String]] = Task {
```

example/androidlib/java/5-R8/build.mill
vaslabs: thanks @lefou , I suggested to @souvlakias to split this PR into two this morning, do the proguard changes for hilt and providers first, I'd like to experiment a bit with this to find a better solution, as at first I made a direct translation from gradle -> mill

vaslabs:
works with the sample app and no examples are edited for this PR, from my side it's ok to go

vaslabs (approved):

PR #5678, opened 2025-08-08
Android: Ability to easily create tests for vanilla android modules

vaslabs:
During experimentation in https://github.com/com-lihaoyi/mill/issues/5654 , I've realised that there's not an easy way to create unit tests for android submodules that are neither libs (publishable, that require pom and publish configurations) nor apps.

This PR reworks the Android (unit) tests to be  based on AndroidModule

In the mill sample provided, which will be included in third party on a later PR, this makes the following setup possible:

```scala
object database extends AndroidHiltLib, AndroidKotlinModule {
  override def androidNamespace: String = "com.nicos.database"

  override def mvnDeps: T[Seq[Dep]] = super.mvnDeps() ++ Seq(
    mvn"androidx.core:core-ktx:1.16.0",
    mvn"androidx.appcompat:appcompat:1.7.1",
    mvn"com.google.android.material:material:1.12.0",
    mvn"androidx.room:room-runtime:2.7.2",
    mvn"androidx.room:room-ktx:2.7.2",
    mvn"com.google.code.gson:gson:2.13.1",
    mvn"org.jetbrains.kotlinx:kotlinx-coroutines-core:1.10.2",
    mvn"org.jetbrains.kotlinx:kotlinx-coroutines-android:1.10.2",
  )

  override def kotlinSymbolProcessors: T[Seq[Dep]] = super.kotlinSymbolProcessors() ++ Seq(
    mvn"androidx.room:room-compiler:2.7.2"
  )

  object test extends AndroidKotlinTestModule with TestModule.Junit4

}
```

`mill show database.test`

I have also removed the `AndroidTestModule.AndroidJUnit`, it seems redundant and so I cleaned up the examples that use it too


vaslabs:
this should be ready

PR #5688, opened 2025-08-11
Android: Fix Proguard Cache Invalidation (Changes Existing Examples)

souvlakias:
Fixes a cache invalidation issue as mentioned in https://github.com/com-lihaoyi/mill/issues/5654  
> Proguard file changes do not invalidate androidDex/androidApk caches.

### NOTE
This PR changes the API, so old examples will break with this.

### Description
Proguard files were previously defined with a nested object in `AndroidBuildTypeSettings`.  
We now treat the project files as `Task.Sources` in order to detect content changes, and copy the default file from the Sdk Proguard Path to the `Task.dest`.

### API Changes
```diff
-override def androidDebugSettings: T[AndroidBuildTypeSettings] = Task {
-  AndroidBuildTypeSettings(
-    isMinifyEnabled = false,
-    isShrinkEnabled = false
-  ).withDefaultProguardFile("proguard-android-optimize.txt")
-    .withProguardLocalFiles(
-      Seq(
-        moduleDir / "proguard-rules.pro"
-      )
-    )
-  )
-}

+override def androidDebugSettings: T[AndroidBuildTypeSettings] = Task {
+  AndroidBuildTypeSettings(
+    isMinifyEnabled = false,
+    isShrinkEnabled = false
+  )
+}

+def androidProjectProguardFiles = Task.Sources(
+  "proguard-rules.pro"
+)
+def androidDefaultProguardFileNames: Task[Seq[String]] = Task.Anon {
+  Seq("proguard-android-optimize.txt")
+}

### DIFF
--- a/libs/androidlib/src/mill/androidlib/AndroidR8AppModule.scala
+++ b/libs/androidlib/src/mill/androidlib/AndroidR8AppModule.scala
@@ -66,7 +66,7 @@ trait AndroidR8AppModule extends AndroidAppModule {
   def androidProjectProguardFiles: T[Seq[PathRef]] = Task.Sources()
 
   /** ProGuard/R8 rules configuration files for release target (user-provided and generated) */
-  def androidAllProguardFiles: T[Seq[PathRef]] = Task {
+  def androidProguardConfigs: T[Seq[PathRef]] = Task {
     androidDefaultProguardFiles() ++ androidProjectProguardFiles() ++ androidLibraryProguardConfigs()
   }
 
@@ -75,7 +75,7 @@ trait AndroidR8AppModule extends AndroidAppModule {
     val inheritedProguardFile = super.androidProguard()
 
     val globalProguard = Task.dest / "global-proguard.pro"
-    val files = androidAllProguardFiles()
+    val files = androidProguardConfigs()
     os.write(globalProguard, os.read(inheritedProguardFile.path))
 
     files.foreach(pg =>

### CODE REVIEW

File: libs/androidlib/src/mill/androidlib/AndroidR8AppModule.scala [hunk 1]
```diff
@@ -60,19 +59,20 @@ trait AndroidR8AppModule extends AndroidAppModule {
...
   }
 
+  /**
+   * The proguard files in this project's sources, typically
+   * in `moduleDir / "proguard-rules.pro"`.
+   * @return
+   */
+  def androidProjectProguardFiles: Seq[os.Path] = Seq.empty[os.Path]
+
+  private def androidProjectProguardSources: Task[Seq[PathRef]] = Task.Sources(
```
vaslabs: I think we can just have one that is Task.Sources instead of this private and the above, and if I'm not mistaken Task.Source works with a relative path, so someone could just do `Task.Sources("proguard-rules.pro")` or smth similar

vaslabs: alternatively, instead of path, you can use something like [this](https://github.com/com-lihaoyi/mill/blob/main/libs/javalib/src/mill/javalib/JavaModule.scala#L774) (SubPath, so the user's entries look more like [this](https://github.com/android/architecture-samples/blob/main/app/build.gradle.kts#L49) , i.e. without the need to use moduleDir)

libs/androidlib/src/mill/androidlib/AndroidR8AppModule.scala
```diff
@@ -60,19 +59,25 @@ trait AndroidR8AppModule extends AndroidAppModule {
...
+  def androidProjectProguardFiles: Seq[os.SubPath] = Seq.empty[os.SubPath]
+
+  /**
+   * Defines the sources for the proguard files.
+   * By default, uses the SubPaths from [[androidProjectProguardFiles]].
+   * @return
+   */
+  def androidProjectProguardSources: T[Seq[PathRef]] = Task.Sources(
+    androidProjectProguardFiles*
+  )
```
lefou: What is the motivation for the two-part API? Did you hit any issues with just overriding `androidProjectProguardSources` directly (which I would prefer), if you want to change the file set?

`Task.Sources` already accepts sub-paths (in addition to `Path`s), so it should be equally convenient to write.

souvlakias: Hi, I based it on [this](https://github.com/com-lihaoyi/mill/blob/main/libs/javalib/src/mill/javalib/JavaModule.scala#L774), but you’re right, it’s better to just have a single API. I’ll fix it, thanks.

lefou: I already assumed that. I really dislike the solution in `JavaModule.sourceFolders`, since it offeres two orthogonal ways to configure stuff and has potential to confuse build maintainers. It's also not inspectable with `mill show`. We should all understand, that this pattern is a workaround due to the enforced limitation to not refer to `super` when overriding `Task.Sources`, but should not be used when we can model it with tasks directly.

libs/androidlib/src/mill/androidlib/AndroidR8AppModule.scala
```diff
@@ -60,19 +59,15 @@ trait AndroidR8AppModule extends AndroidAppModule {
       .flatMap(_.proguardRules)
   }
 
+  /**
+   * The ProGuard/R8 rules configuration files for the Android project.
+   * @return
+   */
+  def androidProjectProguardFiles: T[Seq[PathRef]] = Task.Sources()
```
lefou, line 66: Is "proguard file" the default idiom used for these config files? If there are more specific terms like "proguard rules" or "proguard configs", we probably should use these.

vaslabs: I like the `androidProjectProguardRules` better :+1:

vaslabs: in gradle though they are referred as
`proguardFiles(getDefaultProguardFile("proguard-android.txt"), "proguard-rules.pro")`

libs/androidlib/src/mill/androidlib/AndroidR8AppModule.scala
```diff
@@ -60,19 +59,15 @@ trait AndroidR8AppModule extends AndroidAppModule {
...
-    val defaultProguardFile = proguardFilesFromBuildSettings.defaultProguardFile.map {
-      pf => androidProguardPath / pf
-    }
-    val userProguardFiles = proguardFilesFromBuildSettings.localFiles
-    BuildCtx.withFilesystemCheckerDisabled {
-      (defaultProguardFile.toSeq ++ userProguardFiles).map(PathRef(
-        _
-      )) ++ androidLibraryProguardConfigs()
-    }
+  def androidProguardConfigs: T[Seq[PathRef]] = Task {
```
vaslabs, line 69: maybe `androidAllProguardRules` would be a better name?

vaslabs:
apart from some minor naming changes, this looks good to me, the only thing to highlight is that it breaks compatibility, I think in the current experimental state this is allowed, just highlighting it 

@lihaoyi @lefou

souvlakias:
For the Proguard Rules Files, I think we should keep the `Files` naming like Gradle for now, if that's okay.

lefou:
The PR description looks a bit outdated, at least the diff. The PR looks good to me. Compatibility-wise, I think we only add new tasks but don't necessarily go incompatible, if we avoid the renaming from `androidProguardConfigs` to `androidAllProguardFiles` for now.

vaslabs:
> The PR description looks a bit outdated, at least the diff. The PR looks good to me. Compatibility-wise, I think we only add new tasks but don't necessarily go incompatible, if we avoid the renaming from `androidProguardConfigs` to `androidAllProguardFiles` for now.

the incompatibility is primarily with the AndroidBuildTypeSettings which no longer has the methods for adding the proguard files. 

We can keep the old name `androidProguardConfigs` though

lefou:
> the incompatibility is primarily with the AndroidBuildTypeSettings which no longer has the methods for adding the proguard files.

We probably don't have users of these APIs and maintaining some `@deprecated`'s might be not worth the trouble, but we could just `@deprecated` and ignore those fields for now and do a consolidated cleanup round a bit later, e.g. when Mill 1.1 comes out.

vaslabs:
> > the incompatibility is primarily with the AndroidBuildTypeSettings which no longer has the methods for adding the proguard files.
> 
> We probably don't have users of these APIs and maintaining some `@deprecated`'s might be not worth the trouble, but we could just `@deprecated` and ignore those fields for now and do a consolidated cleanup round a bit later, e.g. when Mill 1.1 comes out.

I would think it's not worth the trouble, and the way these were hooked in was the primary reason proguard changes were not picked up, we also have android marked as experimental, but the final decision would be up to you, if you prefer we can do the safest course with the deprecation

Issue #4473, opened 2025-02-04
Move BSP out dir to a different folder


lihaoyi:
Now that we're using Scala 3 and the IntelliJ compile server is on by default, I've been seeing intermittent errors that seem to suggest multiple compilers stomping over each's files on disks

lefou:
So, you suggesting to no longer shared any build cache between CLI and IDE. A lot of development efforts went into sharing both, but sometimes with questionable success (For example: shared compile results while keeping IDE-dedicated resuls like semanticdb files separate). We should present the pros and cons. And probable make this an easy user-selectable option instead of a hard-coded choice.

Personally, I have not much use of shared build caches (in its current state). When using `GenIdea`, we also don't share compiler build output. When I need artifacts or other build targets, I always go to the CLI. IDE is just for editing, sometimes testing (when writing the test or the code under test).

lefou:
Beside avoiding a whole class of possible issues, we also could avoid the global lock by making it out-dir specific.

lefou:
Instead of using a different out-dir, we could just change all those `bspXXX` targets that return an `UnresolvedPath`, since they potentially violate the `T.dest` contract, to only read file in foreign `T.dest` paths. Instead of peaking into the `T.dest` of there companion non-bsp tasks, they should just use their own `T.dest`. 

The would mean potential out of sync task that return an `UnresolvedPath` aren't shared between BSP and CLI, but all other tasks like source generators or practically everything that is not handled by BSP specifically, can be shared as before, as Mill takes care of their integrity.

lihaoyi:
I think having separate `out/` folders and separate Mill processes would be the simplest thing here. IIRC that's also what SBT does (or can be configured to do?) to avoid the same class of problem.

We could definitely tweak the BSP side to try and make the interference go away, but I'm not familiar enough with our BSP integration to knwo for sure if it's feasible

lihaoyi:
According to https://blog.pierre-ricadat.com/scala-3-migration-report-from-the-field, Intellij/SBT has a co fig `use separate compiler output paths` that does this

lihaoyi:
A Manual workaround to unblock intellij is to set change `Settings / Languages & Frameworks / Scala / Editor / Error Highlighting / Scala3` from `Compiler` to `Built-in`, which makes IntelliJ not make so many requests to BSP resulting in less blocking

arturaz:
Taking this on.

Will make it configurable via mill frontmatter.

PR #5689, opened 2025-08-11
Move `out/` to `.mill-bsp-out/` when running in BSP mode

arturaz:
TODO: see if the tests pass

Fixes https://github.com/com-lihaoyi/mill/issues/4473

### DIFF
--- a/core/constants/src/mill/constants/OutFiles.java
+++ b/core/constants/src/mill/constants/OutFiles.java
@@ -1,23 +1,37 @@
 package mill.constants;
 
+import java.util.Optional;
+
 /**
  * Central place containing all the files that live inside the `out/` folder
  * and documentation about what they do
  */
 public class OutFiles {
 
+  /// Allows us to override the `out/` folder from the environment via the [EnvVars#MILL_OUTPUT_DIR] variable.
   private static final String envOutOrNull = System.getenv(EnvVars.MILL_OUTPUT_DIR);
 
-  /**
-   * Default hard-coded value for the Mill `out/` folder path. Unless you know
-   * what you are doing, you should favor using [[out]] instead.
-   */
+  /// Default hard-coded value for the Mill `out/` folder path. Unless you know
+  /// what you are doing, you should favor using [#outFor] instead.
   public static final String defaultOut = "out";
 
-  /**
-   * Path of the Mill `out/` folder
-   */
+  /// Path of the Mill `out/` folder. Unless you know what you are doing, you should
+  /// favor using [#outFor] instead.
   public static final String out = envOutOrNull == null ? defaultOut : envOutOrNull;
+
+  /// Path of the Mill `out/` folder when Mill is running in BSP mode. Unless you know
+  /// what you are doing, you should favor using [#outFor] instead.
+  public static final String bspOut = ".mill-bsp-out";
+
+  /// Path of the Mill [#out] folder.
+  ///
+  /// @param bspMode if true, we are running in BSP mode, so we need to return a different path.
+  ///                If [#envOutOrNull] is set, this parameter is ignored.
+  public static String outFor(boolean bspMode) {
+    if (envOutOrNull != null) return envOutOrNull;
+    if (bspMode) return bspOut;
+    return out;
+  }
 
   /**
    * Path of the Mill "meta-build", used to compile the `build.sc` file so we can
--- a/runner/launcher/src/mill/launcher/MillLauncherMain.java
+++ b/runner/launcher/src/mill/launcher/MillLauncherMain.java
@@ -13,12 +13,14 @@ import mill.constants.OutFiles;
  * A Scala implementation would result in the JVM loading much more classes almost doubling the start-up times.
  */
 public class MillLauncherMain {
+  static final String bspFlag = "--bsp";
+
   public static void main(String[] args) throws Exception {
     boolean runNoServer = false;
     if (args.length > 0) {
       String firstArg = args[0];
       runNoServer =
-          Arrays.asList("--interactive", "--no-server", "--no-daemon", "--repl", "--bsp", "--help")
+          Arrays.asList("--interactive", "--no-server", "--no-daemon", "--repl", bspFlag, "--help")
                   .contains(firstArg)
               || firstArg.startsWith("-i");
     }
@@ -31,14 +33,16 @@ public class MillLauncherMain {
       }
     }
 
+    var bspMode = Arrays.asList(args).contains(bspFlag);
+
     if (runNoServer) {
       // start in no-server mode
-      System.exit(MillProcessLauncher.launchMillNoDaemon(args));
+      System.exit(MillProcessLauncher.launchMillNoDaemon(args, bspMode));
     } else
       try {
         // start in client-server mode
         java.util.List<String> optsArgs = new java.util.ArrayList<>();
-        optsArgs.addAll(MillProcessLauncher.millOpts());
+        optsArgs.addAll(MillProcessLauncher.millOpts(bspMode));
         Collections.addAll(optsArgs, args);
 
         ServerLauncher launcher =
@@ -51,7 +55,7 @@ public class MillLauncherMain {
                 null,
                 -1) {
               public Process initServer(Path daemonDir, Locks locks) throws Exception {
-                return MillProcessLauncher.launchMillDaemon(daemonDir);
+                return MillProcessLauncher.launchMillDaemon(daemonDir, bspMode);
               }
 
               public void preparedaemonDir(Path daemonDir) throws Exception {
@@ -59,8 +63,8 @@ public class MillLauncherMain {
               }
             };
 
-        Path daemonDir0 = Paths.get(OutFiles.out, OutFiles.millDaemon);
-        String javaHome = MillProcessLauncher.javaHome();
+        Path daemonDir0 = Paths.get(OutFiles.outFor(bspMode), OutFiles.millDaemon);
+        String javaHome = MillProcessLauncher.javaHome(bspMode);
         int exitCode = launcher.run(daemonDir0, javaHome).exitCode;
         if (exitCode == ClientUtil.ExitServerCodeWhenVersionMismatch()) {
           exitCode = launcher.run(daemonDir0, javaHome).exitCode;
--- a/integration/ide/bsp-server/src/BspServerTests.scala
+++ b/integration/ide/bsp-server/src/BspServerTests.scala
@@ -444,7 +444,9 @@ object BspServerTests extends UtestIntegrationTestSuite {
           override def onBuildPublishDiagnostics(params: b.PublishDiagnosticsParams): Unit = {
             // Not looking at diagnostics for generated sources of the build
             val keep =
-              !uriAsSubPath(params.getTextDocument.getUri).startsWith(os.sub / OutFiles.out)
+              !uriAsSubPath(
+                params.getTextDocument.getUri
+              ).startsWith(os.sub / OutFiles.outFor( /* bspMode */ true))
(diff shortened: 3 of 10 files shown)


core/constants/src/mill/constants/OutFiles.java
```diff
@@ -6,19 +6,32 @@
  */
 public class OutFiles {
 
+  /// Allows us to override the `out/` folder from the environment via the [EnvVars#MILL_OUTPUT_DIR]
```
lefou: Unusual comment marker `///`. Please make these standard JavaDoc comments. 

(I know Java 23 supports `///` for markdown comments, but we don't necessarily use nor require Java 23 tooling right now.)

core/constants/src/mill/constants/OutFiles.java
```diff
@@ -6,19 +6,32 @@
...
 
+  /// Path of the Mill `out/` folder when Mill is running in BSP mode. Unless you know
+  /// what you are doing, you should favor using [#outFor] instead.
+  public static final String bspOut = ".mill-bsp-out";
+
+  /// Path of the Mill [#out] folder.
+  ///
+  /// @param bspMode if true, we are running in BSP mode, so we need to return a different path.
+  ///                If [#envOutOrNull] is set, this parameter is ignored.
+  public static String outFor(boolean bspMode) {
```
lefou: We should use an `enum` here. `boolean` is always under-documented at the call-site and scales badly once we think we also want IDEA or Eclipse point to their own out-dirs.

arturaz: Do you mean `enum` with one case?

lefou: I count two: CLI, BSP

lefou: Or whatever we want to call the `bspMode=false` mode.

arturaz:
> It would be nice if we could opt-in/out via an env or frontmatter setup.

It seems frontmatter is hard to do because we need to know the out directory before that's read, but env variable should be possible.

lihaoyi:
> I'm not fully convinced we should default to this behavior. It would be nice if we could opt-in/out via an env or frontmatter setup.

We could make it configurable, but this should be the behavior by default. Getting your IDE blocked by your terminal and vice versa is a huge pain in the neck, and anyway IntelliJ already spawns a separate Mill process to run it's `--bsp` mode so there shouldn't be any memory footprint increase either


> It seems frontmatter is hard to do because we need to know the out directory before that's read, but env variable should be possible.

We currently use the `out/` folder as a place to cache the parsed frontmatter key-values, so that in the common case the launcher doesn't need to load SnakeYAML and other heavyweight logic that slows down classloading

I wonder if such caching is really necessary now that we use Graal native image launchers by default? Perhaps the faster launch times that Graal provides is enough to render the whole caching/optimization thing unnecessary

arturaz:
I guess ENV var vs frontmatter is whether this setting is project-wide or user-wide.

I think that it's user-wide, given that people with beefy PCs won't mind extra compute used, where as people with slower PCs might. So ENV makes more sense to me.

lihaoyi:
The thing about an ENV_VAR is that we would need to modify IntelliJ/VSCode to pass it. Possible, but not really convenient for anyone who wants to pass it in ad-hoc. If we want it to be user-configrable, it would need to be in the frontmatter

arturaz:
> The thing about an ENV_VAR is that we would need to modify IntelliJ/VSCode to pass it.

Wouldn't it be picked up from your system's environment variables?

lihaoyi:
Good question, I don't know. Could you try it out? e.g. I don't know where to configure system environment variables such that launching `Intellij IDEA` via Spotlight on a mac would pick them up.

Once you've tried it out on mac/windows/linux, we should write it up and add it to the docs, perhaps on the page `Installation and IDE Support`

lihaoyi:
Can we move `.mill-bsp-out` into `.bsp/out`? There already is a BSP folder to put BSP stuff into, so we might as well put this BSP out folder into it too to keep the top-level project dir clean

arturaz:
> Can we move `.mill-bsp-out` into `.bsp/out`? There already is a BSP folder to put BSP stuff into, so we might as well put this BSP out folder into it too to keep the top-level project dir clean

We could, but I personally have to resort to `rm -rf .bsp .metals` to fix the state of the project quite often (unfortunately) and I wouldn't want to wipe the compilation cache when running that as well.

lihaoyi:
How about putting it in `out/mill-bsp-out` then? Thst would keep the top-level directory clean while keeping the out files all together, so someone who `rm -rf out` would wipe out the BSP output as well, but `rm -rf .bsp` would not

arturaz:
If we use something with slashes (subdirectories), we have to refactor the code to avoid this:

> [7225-1] Exception in thread "main" os.PathError$InvalidSegment: [out/mill-bsp-out] is not a valid path segment. [/] is not a valid character to appear in a non-literal path segment. If you are dealing with dynamic path-strings coming from external sources, use the Path(...)/RelPath(...)/SubPath(...) constructor calls to convert them.

~Is it worth it? :)~

Edit: nevermind, turns out it's just a couple of places.

lihaoyi:
The added docs should go on the website under installation and ide support, rsther than the --help text. Most people won't be running mill --bsp themselves in the terminal and would rely on the IDE running it for them, so they'll have no reason to check the --help text to find these docs

lefou:
The new env variable should also be documented in the same location as all the others.

core/constants/src/mill/constants/EnvVars.java
```diff
@@ -27,6 +27,12 @@ public class EnvVars {
    */
   public static final String MILL_OUTPUT_DIR = "MILL_OUTPUT_DIR";
 
+  /**
+   * If set to "1", Mill will re-use the regular @{Link OutFiles#out} folder instead of
+   * using a separate one for BSP output.
+   */
+  public static final String MILL_MERGE_BSP_OUTPUT_DIR = "MILL_MERGE_BSP_OUTPUT_DIR";
```
lefou: I find the word "merge" a bit misleading, since we don't merge anything but use the same location. What about `MILL_NO_SEPARATE_BSP_OUTPUT_DIR`?

arturaz:
Yes.








 ---- On Wed, 13 Aug 2025 11:47:18 +0300  ***@***.***  wrote ----



@lefou commented on this pull request.


Just to clarify, if the MILL_OUTPUT_DIRvariable is set, it overrides whatever we default to, right?

runner/launcher/src/mill/launcher/MillLauncherMain.java
```diff
@@ -33,14 +36,16 @@ public static void main(String[] args) throws Exception {
       }
     }
 
+    var outMode = Arrays.asList(args).contains(bspFlag) ? OutFolderMode.BSP : OutFolderMode.REGULAR;
```
lihaoyi: This will trigger if any user-defined commands have `--bsp` as a flag, which may cause confusion. But maybe that's a rare enough occurrence we can ignore it for now

lefou: Since we still assume the `--bsp` flag is given in first position (see line 26 above), we should limit the enablement to the first array item too.

lihaoyi: I'm not sure that's true, IIRC intellij passes `-i --bsp` with `--bsp` in the second position

lefou: Maybe, we can hardcode that case? IMHO preferable over some occasional wrong behavior we never hear about, but which might cause a lot of trouble for a single developer soul out there.

It's a pity that we don't properly parse the CLI in the (Java) client (for performance reasons).

lihaoyi: I think whitelisting `--bsp` and `-i --bsp` as the first two args sounds reasonable. 

Alternatively, one thing we can do is if we identify a `--bsp` `-i` in the flags, we then run the argument parser in the launcher to see whether they correspond to the flags we care about. With Graal native launchers the cost may be low enough to be acceptable, but that will need to be measured

lihaoyi:
I think to move forward here:

@arturaz can you see if loading the actual `MillCliConfig.parser` has a significant impact on startup times in the JVM and native launchers? It seems to me the best approach would be to use a naive `String#contains` as the first filter, and then if `--bsp` is found use `MillCliConfig.parser` on the tokens to verify that the `--bsp` token is intended to turn on BSP mode. If the performance hit isn't too much, especially in native mode (i.e. everyone except Windows-ARM), then it's probably worth performing that two-stage parse to avoid false positives (e.g. if the user is running a `Task.Command` which takes its own `--bsp` flag)

runner/launcher/src/mill/launcher/MillLauncherMain.java
```diff
@@ -33,14 +37,30 @@ public static void main(String[] args) throws Exception {
...
+        OutFiles.mergeBspOut
+          ? "Mill is running in BSP mode and '" + EnvVars.MILL_NO_SEPARATE_BSP_OUTPUT_DIR + "' environment variable " +
+          "is set, Mill will use the regular '" + outDir + "' as the output directory. Unset this environment " +
+          "variable if you want to use a separate output directory for BSP. This will increase the CPU usage of the " +
+          "BSP server but make it more responsive."
+          : "Mill is running in BSP mode, using a separate output directory '" + outDir + "'. If you would like to " +
+          "reuse the regular `out/` directory, unset the '" + EnvVars.MILL_NO_SEPARATE_BSP_OUTPUT_DIR +
+          "' environment variable. This will reduce the CPU usage of the BSP server but make it less responsive."
+      );
+    }
```
arturaz: @lihaoyi @lefou what do you think about this help message?

arturaz:
> can you see if loading the actual `MillCliConfig.parser` has a significant impact on startup times in the JVM and native launchers?

@lihaoyi sure, but `MillCliConfig` sits in `core.internal` module, whilst we need this check in `runner.launcher` module, which doesn't have that dependency. Do you want me to extract `MillCliConfig` into it's own module? But it uses mainargs, which depends on scala-library, so we're basically throwing out the whole idea of "launcher classpath is very minimal"?

lihaoyi:
Sure let's break it out into its own module.

We kind of threw out the idea of "launcher classpath is very minimal" in the general case when we started bundling Coursier into the launcher. The current design is "fast path is minimal, slow path might not be", so coursier is only used to bootstrap and then never touched after. We can do something similar here, where the "slow path" is if the user passes the `--bsp` flag, which should basically be done by IDEs 99.9% of the time, and for most command-line use it shouldn't touch mainargs and other Scala libraries at all

lihaoyi:
We should probably have an integration test that verifies the "fast path" doesn't load unnecessary classes, similar to what Ammonite has (https://github.com/com-lihaoyi/Ammonite/blob/2fdc440b23c9bc7eb782c496c05ec1d3c10ee3d6/integration/src/test/scala/ammonite/integration/BasicTests.scala#L33). Could you add that as part of this PR?

arturaz:
Tested with:

```java
    var outMode = containsBspFlag(args) ? OutFolderMode.BSP : OutFolderMode.REGULAR;
    System.exit(0);
```

```
$ time ./mill-assembly.jar

real	0m0,076s
user	0m0,072s
sys	0m0,049s

$ time ./mill-assembly.jar --bsp

real	0m0,250s
user	0m0,448s
sys	0m0,116s

$ time ./mill-native

real	0m0,020s
user	0m0,004s
sys	0m0,020s

$ time ./mill-native --bsp

real	0m0,021s
user	0m0,007s
sys	0m0,018s
```

Issue #3259, opened 2024-07-11
Classes don't seem to be always put in `compiledClassesAndSemanticDbFiles.dest`


kasiaMarek:
Mill build server points to `.../compiledClassesAndSemanticDbFiles.dest` as the `Classes Directory`, but classes seem to be only put there when mill is explicitly asked to compile a specific build target.

### Expectation
Mill should also put classes in `compiledClassesAndSemanticDbFiles.dest` for a build target that is compiled as a dependency.

### Example
1. Have two build targets `main` and `test`, where `test` depends on `main`.
2. Compile `test`.
3. I'd expect to have classes in appropriate `.../compiledClassesAndSemanticDbFiles.dest` for both build targets.

### More context
In Metals we use `compiledClassesAndSemanticDbFiles.dest` as part of the classpath for when debugging. Right now the user needs to first open all dependency modules, so classes will appear in the correct destination. 
connected to: https://github.com/scalameta/metals/issues/6579

lefou:
Could this be some unsharpness in the BSP protocol spec? Metals wants to use some location to access JVM byte-code of a specific module and it's dependencies, but never requests the JVM byte-code to be produced for the dependencies. 

This is also a limitation of the BSP protocol. Although, sementicDB files are not needed to debug something, BSP only supports a single output location for target files of a JVM bytecode files and a semanticDB files producer. So, Mill needs to provide the already produced JVM-byte-code in a shared location with the semanticDB files, which is only available after you have requested and produced both.

We could and probably should fix this on Mill's side by letting the `copiledClassesAndSemanticbFiles` target depend on its companions of its recursive dependencies. But it would be nice to also open an issue for BSP to get that detailed expectation into the written BSP spec.

lihaoyi:
Going to call this stale given the recent overhaul of BSP support. Feel free to re-open if it still happens on 1.0.0-RC2 or later

lefou:
I'm not aware of any newly introduced logic that is solving this issue.

arturaz:
@kasiaMarek can you elaborate on whether `out/test/compiledClassesAndSemanticDbFiles.dest` should contain classes from both `main` and `test` modules or there should be separate `out/main/compiledClassesAndSemanticDbFiles.dest` and `out/test/compiledClassesAndSemanticDbFiles.dest` with their respective files?

lefou:
[...]

PR #5692, opened 2025-08-12
Make `compiledClassesAndSemanticDbFiles` transitive through `moduleDeps`

arturaz:
Now when invoking `compiledClassesAndSemanticDbFiles` it is transitive through all `moduleDeps`.

Fixes https://github.com/com-lihaoyi/mill/issues/3259


libs/javalib/src/mill/javalib/SemanticDbJavaModule.scala
```diff
@@ -146,17 +147,22 @@ trait SemanticDbJavaModule extends CoursierModule with SemanticDbJavaModuleApi
...
-    val sems = semanticDbData().path
+    val semanticDbDatum =
+      Task.sequence(semanticDbData +: moduleDeps.collect { case m: SemanticDbJavaModule =>
+        m.semanticDbData
+      })().map(_.path)
     os.list(dest).foreach(os.remove.all(_))
-    if (os.exists(sems)) os.copy(sems, dest, mergeFolders = true)
+    semanticDbDatum.foreach { data =>
+      if (os.exists(data)) os.copy(data, dest, mergeFolders = true)
+    }
```
lefou: That doesn't look right. All we need is to trigger the update of all transitive `compiledClassesAndSemanticDbFiles` tasks. We should not copy their result into each downstream dest folder.

lefou: The point is, that the IDE/BSP is already including all the various `compiledClassesAndSemanticDbFiles`  results in their classpaths. But later these files were not updated, due to the BSP spec beeing too vague. Once, we trigger all upstream tasks, it should already work as intended.

arturaz: Hm, I understood that everything needs to go to same folder, asked Kasia for clarification: https://github.com/com-lihaoyi/mill/issues/3259#issuecomment-3178600293

arturaz: I changed the behavior to one you said in https://github.com/com-lihaoyi/mill/pull/5692/commits/5359c27b91b9d4dc5af3d0b2a48b2d9cdba5ec58 , is this resolved then?

lefou (approved):
Looks good to me. Thank you!

PR #5698, opened 2025-08-13
BSP fixes

alexarchambault:
This fixes several small BSP issues, and tweaks the BSP snapshot tests too.

In particular, this makes the BSP snapshot tests more exhaustive, by testing a meta-build too, and making some tests more broad (main class request test).

This fixes issues with the buildTargetInverseSources request:
- it makes return the correct target when passed a `build.mill` file (while it doesn't return anything without the changes here)
- it ensures a compile reporter is around when evaluating tasks for that request, which is required if a `sources` task depends on a `compile` one for example (as a check requires such a reporter to be around during all task evaluations in the BSP server, so that we don't miss any diagnostic)

This PR commits can be reviewed one-by-one, and provide a bit of explanation of what each commit does in the commit message.


alexarchambault:
@lihaoyi It seems `UTEST_UPDATE_GOLDEN_TESTS=true` writes a truncated value when we give it a string with too many lines maybe. See how it truncates `integration/ide/bsp-server/resources/snapshots/workspace-build-targets.json` and adds `...` at the end of it [here](https://github.com/com-lihaoyi/mill/pull/5698/files#diff-5529b2f0117a8be119d9e8106fdcdf7765db2a6de18631503fe7fd8a21035de3).

lihaoyi:
@alexarchambault could you try updating this line upstream and adding `height=99999` to see if that solves the problem? If that fixes it can send a PR to utest and i'll cut a new release

lihaoyi:
def goldenLiteralPrinter(x: Any): String = pprint.PPrinter.BlackWhite.apply(x).plainText

lihaoyi:
https://github.com/com-lihaoyi/utest/blob/master/utest/src/utest/framework/Formatter.scala#L40

lefou:
Changes look good to me.

lihaoyi:
@alexarchambault uTest 0.9.1 should have your fix

PR #5701, opened 2025-08-14
Make sure, dynamically installed Android tools exists

lefou:
Also revalidate cached paths, so we capture any changes to these externally stored resources.

Changed various tasks result types to `os.Path`, if they only point to some directories.

Renamed some task suffixes from `Path` to `Exe` to match the fact, that these return executable files that are expected to exist.


libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala
```diff
@@ -564,4 +556,12 @@ object AndroidSdkModule {
    * Declaration of the Maven Google Repository.
    */
   val mavenGoogle: MavenRepository = MavenRepository("https://maven.google.com/")
+
+  private def toolPathRef(path: os.Path)(using TaskCtx): PathRef = {
+    os.exists(path) match {
+      case true => PathRef(path).withRevalidateOnce
+      case false => Task.fail(s"Tool at path ${path} does not exist")
```
lefou, line 563: Can we add a hint, how to fix this?

lefou:
While reviewing the `AndroidSdkModule`, I discovered that lots of tasks run the `installAndroidSdkComponents` anon task which has a complicated synchronization setup, which isn't very Mill idomatic. Instead, we should refactor it to  depends on a single `androidSdkComponents` task, which encapsulates all of the logic. Or split it up, if we need more granular setup steps. If we really need some dynamic orchestration, we should implement all of it in a worker.

Issue #5693, opened 2025-08-12
Performance regression with `ZincWorkerMain` subprocesses


lihaoyi:
If we enable per-module `jvmId`s in our Netty example build:

```diff
@@ -15,6 +15,7 @@ import mill.api.BuildCtx
 def isOSX = System.getProperty("os.name").toLowerCase.contains("mac")
 
 trait NettyBaseModule extends MavenModule {
+  def jvmId = "17"
   def javacOptions = Seq("-source", "1.8", "-target", "1.8")
 }
 trait NettyBaseTestSuiteModule extends NettyBaseModule, TestModule.Junit5 {
@@ -78,6 +79,7 @@ trait NettyModule extends NettyBaseModule {
   def testMvnDeps: T[Seq[Dep]] = Task { Seq.empty[Dep] }
 
   object test extends NettyBaseTestSuiteModule, MavenTests {
+    def jvmId = "17"
     def moduleDeps = super.moduleDeps ++ testModuleDeps
     def mvnDeps = super.mvnDeps() ++ testMvnDeps()
     def forkWorkingDir = NettyModule.this.moduleDir
```

Running `./mill clean && time ./mill __.compile` on the Netty codebase now takes ~234 seconds, where previously it took ~34 seconds. In contrast, without the per-module `jvmId`s, it takes ~10s

the 10s -> 34s slowdown from per-module `jvmId`s is due to forking, and I would expect the long-lived `ZincWorkerMain` daemon to avoid forking and thus take ~10s similar to the default behavior, so it becoming even slower at 234s is surprising. We should figure out where the slowdown is and fix it

arturaz:
So far I've pinned it down to this:

**client**
```
[2025-08-13T16:49:27.057796788] [LOGGER:DEBUG] [RPC:ZincWorker,TCP localhost/127.0.0.1:34541 -> /127.0.0.1:53234] Sending: {"$type":"Ask","id":"c0","data":{"$type":"CompileMixed","op":{...
[2025-08-13T16:49:52.071736590] [LOGGER:DEBUG] [RPC:ZincWorker,TCP localhost/127.0.0.1:34541 -> /127.0.0.1:53234] Trying to read MillRpcServerToClient[Value]
```

**server**
```
pid:647738 2025-08-13T16:49:27.117562125 [ZincWorkerTcpServer{/127.0.0.1:53234 -> /127.0.0.1:34541}] Trying to read MillRpcClientToServer[ClientToServer]
[...]

PR #5703, opened 2025-08-14
Fix Zinc worker performance regression

arturaz:
Fixes the delay when using Mill RPC infrastructure and sending large messages over it.

The cause was PipedInput/OutputStream waiting for one second when it's internal buffer fills up. Because `ProxyStream` protocol used a very small (126 byte) buffer, things got slow very fast once a large message was sent.

This PR removes the dependence on the piped streams and uses socket streams directly. It also ditches ProxyStream protocol for Mill RPC and instead sends heartbeats as empty lines.

Other changes:
- `Server` now is basic server, `ProxyStream` specific logic was moved to `ProxyStreamServer`.
- A lot more logging is now done. You can find these logs in `out/mill/javalib/JvmWorkerModule/internalWorker.dest/jvm-worker.log` and `out/mill/javalib/JvmWorkerModule/internalWorker.dest/zinc-worker/DAEMON_HASH/daemon/server.log` respectively.
- To aid debugging `DebuggingInputStream` and `DebuggingOutputStream` was created and left in the codebase if we'll need to debug something in the future.
- As a part of debugging effort `ProxyStream` was refactored to be more readable.
- `ProxyStream` protocol was changed to allow chunk sizes bigger than 126 (up to Int.MaxValue).
- `mill.util.Timed` was introduced as a utility to time code blocks.

---

Fixes https://github.com/com-lihaoyi/mill/issues/5693


lihaoyi:
* Can you update the PR description to say how this was tested? Manually is fine, e.g. if you have CLI logs showing that the time to compile a hello-world java module with `jvmId` set is no longer unreasonable, and the time to `./mill clean && ./mill __.compile` on the netty example build with a custom `jvmId` is down to close the baseline without a custom `jvmId` (+1-2 seconds due to the `ZincWorkerMain` startup overhead)

* Let's split out the `ProxyStreams` protocol change into a separate PR. Since we hope that will improve performance separately from the `Piped*Stream` changes, I'd like to see some benchmarks demonstrating that, e.g. `def foo = Task{ println("x"*1000000) }` streaming faster than it does with the prior protocol. But that's involved enough we should do that in a follow-up PR

* `mill.util.Timed` should be `private[mill]` so we can evolve it in future

libs/daemon/client/src/mill/client/ServerLauncher.java
```diff
@@ -59,23 +57,48 @@ public RunWithConnectionResult(A result, int exitCode) {
...
       Socket connection,
       Streams streams,
       boolean closeConnectionAfterClientLogic,
       Consumer<OutputStream> sendInitData,
       RunClientLogic<A> runClientLogic)
       throws Exception {
-    var socketInputStream = connection.getInputStream();
-    var socketOutputStream = connection.getOutputStream();
+    var socketInputStream = new BufferedInputStream(connection.getInputStream());
+    var socketOutputStream = new BufferedOutputStream(connection.getOutputStream());
```
lihaoyi: Is this buffering necessary? I suspect it can be removed because the socket has its own internal buffer

libs/daemon/client/src/mill/client/ServerLauncher.java
```diff
@@ -323,12 +346,15 @@ private static ServerLaunchOutputs readOutputs(Path daemonDir) throws IOExceptio
    * @return a PumperThread that processes the output/error streams from the server
    */
   static PumperThread startStreamPumpers(
-      InputStream socketInputStream, OutputStream socketOutputStream, Streams streams) {
+      InputStream socketInputStream,
+      OutputStream socketOutputStream,
+      Streams streams,
+      String name) {
```
lihaoyi: I think we don't need the `name` anymore, since there's only one callsite of this method in the entire codebase

libs/daemon/server/src/mill/server/Server.scala
```diff
@@ -246,88 +267,83 @@ abstract class Server(
...
-    /** stdout and stderr combined into one stream */
-    val currentOutErr = clientSocket.getOutputStream
-    val writtenExitCode = AtomicBoolean()
-
-    def writeExitCode(code: Int): Unit = {
-      if (!writtenExitCode.getAndSet(true)) {
-        ProxyStream.sendEnd(currentOutErr, code)
-      }
+    val clientToServer = BufferedInputStream(clientSocket.getInputStream, bufferSize)
+    val serverToClient = BufferedOutputStream(clientSocket.getOutputStream, bufferSize)
```
lihaoyi: I suspect we don't need this buffering

libs/daemon/client/src/mill/client/ClientUtil.java
```diff
@@ -8,15 +8,15 @@
...
     return 1;
   }
 
-  public static int ExitServerCodeWhenIdle() {
+  public static byte ExitServerCodeWhenIdle() {
     return 0;
   }
 
-  public static int ExitServerCodeWhenVersionMismatch() {
+  public static byte ExitServerCodeWhenVersionMismatch() {
```
lihaoyi: Why is this change necessary?

arturaz: While not necessary, I just noticed this as using a wider return type than strictly necessary, which would prevent using that `int` in contexts where a byte is needed, even if the value itself actually fits in a byte.

libs/daemon/client/src/mill/client/ServerLauncher.java
arturaz: Right now, yes, but having the name allows you to quickly identify what for these pumpers are without having to look at codebase. Is there a downside to having a name here?

libs/daemon/client/src/mill/client/ServerLauncher.java
arturaz: According to https://pzemtsov.github.io/2015/01/19/on-the-benefits-of-stream-buffering-in-Java.html it seems that buffering on the application level is still beneficial due to syscall overhead, even if kernel has it's own socket buffers.

arturaz:
> * Let's split out the `ProxyStreams` protocol change into a separate PR. Since we hope that will improve performance separately from the `Piped*Stream` changes, I'd like to see some benchmarks demonstrating that, e.g. `def foo = Task{ println("x"*1000000) }` streaming faster than it does with the prior protocol. But that's involved enough we should do that in a follow-up PR

Created https://github.com/com-lihaoyi/mill/pull/5710

lihaoyi:
@arturaz could you run a final end-to-end test on `__.compile` on the Netty codebase to make sure the timing reported is similar to that of the no-custom-`jvmId` timing?

libs/daemon/client/src/mill/client/ClientUtil.java
lihaoyi: Weren't you the one that was telling me that exit codes need to be integers and aren't limited to single byte values? If that's the case, wouldn't `int` be more correct?

libs/daemon/client/src/mill/client/ServerLauncher.java
lihaoyi: Let's add that as a comment in the code then, so people know why it's there.

If we're adding buffering, do we need to call `.flush()` manually anywhere to make the buffered data is sent downstream without unnecessary delays?

libs/daemon/client/src/mill/client/ServerLauncher.java
lihaoyi: If we're not using it, we should remove it, and add it back later as necessary. In general, we try to avoid code that might be useful _later_ hanging around in the codebase, which has the nice benefit of meaning all the code in the codebase is useful _now_ which makes it easier to understand since you know any piece of code _must_ be used

lihaoyi:
Also please check if the slowdown that @alexarchambault reported in `./mill 'integration.ide[bsp-server].packaged.daemon.testForked' mill.integration.BspServerTests.requestSnapshots` is fixed by this PR

libs/daemon/client/src/mill/client/ClientUtil.java
arturaz: Yes, in general exit codes are `int`s, but this specific exit code fits into a `byte`. I don't have strong preference either way, I can roll this back.

libs/daemon/client/src/mill/client/ServerLauncher.java
arturaz: We are using it, to add debugging info to pumper thread names.

libs/daemon/client/src/mill/client/ServerLauncher.java
arturaz: > If we're adding buffering, do we need to call `.flush()` manually anywhere to make the buffered data is sent downstream without unnecessary delays?

My RPC server always flushes when a message is written, I was under impression that the general contract of `OutputStream` is that you need to `flush()` if you want your data to be written immediately. I'll review the other use sites, but I have not noticed any delays so far.

libs/daemon/client/src/mill/client/ClientUtil.java
lihaoyi: let's roll it back then since it's unrelated to this particular PR which is already big/complicated enough

libs/daemon/client/src/mill/client/ServerLauncher.java
lihaoyi: oh i see it now, that's fine then

arturaz:
> Also please check if the slowdown that @alexarchambault reported in `./mill 'integration.ide[bsp-server].packaged.daemon.testForked' mill.integration.BspServerTests.requestSnapshots` is fixed by this PR

Without the fix: `[7402/7402] ============================ integration.ide[bsp-server].packaged.daemon.testForked mill.integration.BspServerTests.requestSnapshots =========================== 400s` and counting. cancelled to save time.
With the fix: `[7402] + mill.integration.BspServerTests.requestSnapshots 49859ms`

lihaoyi:
Looks good then!

lihaoyi:
I assume https://github.com/com-lihaoyi/mill/pull/5710 is the one we should merge first right

arturaz:
> I assume #5710 is the one we should merge first right

Yes, but I'm still running tests for Netty.

arturaz:
> @arturaz could you run a final end-to-end test on `__.compile` on the Netty codebase to make sure the timing reported is similar to that of the no-custom-`jvmId` timing?

Without JVM id: `real    0m23,308s`
With JVM id 1st time: `real    0m33,452s`
With JVM id 2nd time: `real    0m22,015s`

libs/daemon/client/src/mill/client/ServerLauncher.java
arturaz: `ProxyStreamServer` uses `PrintStream` with `autoFlush=true`, so it's fine.

lefou:
PR comments for #5710 and this one look pretty much like copy-and-paste. Can they made more significant before merging?

arturaz:
> PR comments for #5710 and this one look pretty much like copy-and-paste. Can they made more significant before merging?

They are copy-paste, I'll edit this one once #5710 will be merged.

arturaz:
> I'd like to see some benchmarks demonstrating that, e.g. def foo = Task{ println("x"*1000000) } streaming faster than it does with the prior protocol. 

With:
```scala
def shout() = Task.Command {
  println("x" * (10 * 1024 * 1024))
}
```

old:
```
real	0m0,540s
user	0m0,399s
sys	0m0,289s
````

new: 
```
real	0m0,529s
user	0m0,228s
sys	0m0,094s
```

AI overview:

*   **Real time:** There's a minor improvement, with the total elapsed time decreasing from 0.540 seconds to 0.529 seconds. This means the overall execution time that a user would perceive is slightly faster.

*   **User time:** This shows a very large improvement, dropping dramatically from 0.399 seconds to 0.228 seconds. This indicates a substantial optimization in the code itself, as it now requires much less CPU time to execute its instructions.

*   **System time:** This is the most significant change. The time spent on kernel-level operations plummeted from 0.289 seconds to just 0.094 seconds. This suggests that the new version is making far fewer or much more efficient system calls (like for I/O operations).

In summary, while the overall "real time" improvement is modest, the underlying efficiency gains are massive. The new version is significantly better optimized, requiring much less CPU time for both its own code (`user`) and for system-related tasks (`sys`). This is a clear and substantial performance enhancement.

Lower `sys` time means less time is spent in the kernel, and the dramatic drop in `user` time shows your code's logic has become much more efficient.

The reason `real` time can stay almost the same, even with those significant improvements, is that **`real` time includes time the CPU isn't actively working on your process.**

Here's a breakdown of what these timings mean and why this happens:

*   **`real` (Wall-Clock Time):** This is the total time that has passed from the moment you start the process to the moment it finishes, just like timing it with a stopwatch. It includes *everything*: the time the CPU is actively running your code, and also time spent waiting for other things.

*   **`user` (User CPU Time):** This is the amount of time the CPU spent executing your program's own code in user-space. The massive reduction here (from 0.399s to 0.228s) is a clear sign of successful code optimization.

*   **`sys` (System CPU Time):** This is the time the CPU spent executing kernel-level code on behalf of your program. This could be for tasks like reading or writing files, memory allocation, or other system calls. The huge drop you saw (from 0.289s to 0.094s) means the new version is much more efficient in how it interacts with the operating system.

### Why `real` Time Didn't Drop as Much

The total CPU time (`user` + `sys`) for your process dropped significantly:

*   **Old:** 0.399s (user) + 0.289s (sys) = **0.688s**
*   **New:** 0.228s (user) + 0.094s (sys) = **0.322s**

Even though the active CPU work was more than halved, the `real` time only saw a minor improvement. This indicates that a large portion of the execution time is spent **waiting**. Your process is likely "I/O bound," meaning its speed is limited by waiting for input/output operations to complete.

Here are the most common reasons for this discrepancy:

*   **Waiting for I/O:** The program might be waiting for data from a hard drive, a network connection, or user input. During these waits, the CPU is not actively working on your process, so `user` and `sys` time don't increase, but the `real` time clock keeps ticking.
*   **Process Scheduling:** The operating system may be giving CPU time to other running processes. While other applications are running, your process is paused, and this waiting time contributes to the `real` time.
*   **Sleeping or Deliberate Pauses:** If the code has intentional delays (like a `sleep` command), the process will be inactive, which only adds to the `real` time.

In conclusion, your optimizations have made the CPU-intensive parts of your program dramatically faster. However, the overall execution time is still dominated by waiting for external resources, which is why the `real` time hasn't seen a proportional decrease.

lihaoyi:
Please don't paste me AI overviews, if I wanted to hear what Gemini has to say about the results I can ask it myself.

lihaoyi:
The Mill build's `./mill __.compile` output is 64k, so taking your speedup on a 10mb output proportionately we're looking at a 1ms `sys` speedup for builds like Mill and 53 microsecond `real` speedup for builds like Mill. That feels close enough to the noise floor that it doesn't give me confidence that this is a real speedup, and it may well be that change is a no-op or even a slowdown. So I think we shouldn't merge this unless we can demonstrate there's some real improvement, and it's not just churn for the sake of churn

arturaz:
@lihaoyi I would argue that we should merge this not for speed but for correctness (the exit code fix), ease of understanding the code (relying on stdlib's DataInput/OutputStream instead of hand-rolled implementation), ease of understanding the protocol (having a simple 4 byte prefix for length rather than a single byte that serves as an indicator for both length and stream type) and type safety.

lihaoyi:
going to call this a wontfix for now

PR #5705, opened 2025-08-15
WIP make `mill-jvm-version` config required, avoid using system `java` unless explicitly specified 

lihaoyi:
This will make Mill builds more reproducible, and avoid the JVM-version confusion that e.g. @makingthematrix encountered recently in the discord channel.

### DIFF
--- a/runner/launcher/src/mill/launcher/MillProcessLauncher.java
+++ b/runner/launcher/src/mill/launcher/MillProcessLauncher.java
@@ -156,7 +156,15 @@ public class MillProcessLauncher {
 
   static String millJvmVersion() throws Exception {
     List<String> res = loadMillConfig("mill-jvm-version");
-    if (res.isEmpty()) return null;
+    if (res.isEmpty()) {
+      throw new Exception(
+        "mill-jvm-version not set. Please set this in" +
+          "your build header to specify the version of the JVM this project " +
+          "should use, e.g. `//| mill-jvm-version: 17`, " +
+          "`//| mill-jvm-version: temurin:17`, or `//| mill-jvm-version: system` " +
+          "if you want to use the globally-installed JVM on your machine"
+      );
+    }
     else return res.get(0);
   }
 


lefou:
Ho can a user explicitly configure to use the system-installed `java` (whatever version it is). Or in other words, how to restore the current behaviour?

Is it this?

```scala
//| mill-jvm-version: system
```

lihaoyi:
@lefou yes I believe that's the way to do it

lefou:
Since using `system` java might be the preferred way to run for some users (it's what most other (build) tools do and might be easier to integrate in existing tool chains which already handle Java versions by other means), we should have a integration test for it.

website/docs/modules/ROOT/pages/fundamentals/configuring-jvm-versions.adoc
```diff
@@ -5,5 +5,16 @@ Java/Scala/Kotlin modules, typically whatever `java` command you have available
 or what you have configured in your xref:cli/build-header.adoc[]. This page goes into
 more detail about downloading and using a custom Java home on a per-module basis.
 
+
+Note that when publishing artifacts to Maven repositories, the JVM version used to build the
+artifacts will be the minimum version that downstream projects will need to use in order to
+use your artifacts.
+
+* For libraries is best to set your `mill-jvm-version` to the lowest version possible, e.g.
+  `mill-jvm-version: 11` in this example which is the lowest version supported by Mill.
```
lefou, line 14: Is that correct? It should be best practice to configure the target version with `-release` and other compiler options.

lihaoyi: That's what we've been doing in the com-lihaoyi libraries at least. We've never used `-release`, and instead we've just relied on installing the relevant version of Java via `actions/setup-java`. If we're making that stop working by default (due to picking some Mill default Java version) then the corresponding feature would be to configure the Java version to use in `mill-jvm-version`

lefou:
Why was this merged? The PR description explicitly delays this to at least Mill 1.1.0, since it's a breaking change, but we're still in the `1.0.x` window, since we have lots of unreleased changes in `main` branch.

lefou:
I also don't think we should hardcode any version, if there is no good reason for it. Beside being a maintenance burden, software and tools should get a chance to age, esp. when used in maintenance branches. But it's hard for an older version to age, if it also maintains defaults. We should also say up front, what the update policy for those hardcoded default is. If we later bump it without any predictable pattern, we better don't provide a default at all but force the user to pick one at the start.

lefou:
The PR description has lots of discussion but is rather unclear about the effective change. Do we pick the latest Java LTS version which was available when we released the Mill version, or do we pick it dynamically?

website/docs/modules/ROOT/pages/fundamentals/configuring-jvm-versions.adoc
lefou: While this is what most small com.lihaoyi projects do, it's not what I would recommend to any serious developer. It should be clear, that the target Java version should be configured explicitly. In fact, in Mill, we explicitly control that Mill runs with Java 11 onwards by setting the respective flags (`-release`).

lihaoyi:
I figured that by merging this the next release would be 1.1.0.

What I imagined the policy was "Mill uses the latest LTS picked at the time that Mill version is released". So if the JVM version is not specified, the version used is stable, and will only change if you update Mill.

The original plan was to force the user to pick one, but decided that the friction for doing so outweighs the benefit.

lefou:
We already labelled all the other merges PRs with milestone `1.0.4`. I think it would be generally better to release everything that does not need a minor bump as `1.0.4`, then cut the `1.1.0` with only the breaking changes. That way, user can benefit from all the goodies without being required to address the breakages.

lihaoyi:
I can tag the version before this PR as 1.0.4 and update the changelog

lefou:
> I can tag the version before this PR as 1.0.4 and update the changelog

That would be great.

lefou:
> I can tag the version before this PR as 1.0.4 and update the changelog

@lihaoyi It's probably easier to revert this PR and re-apply it after we tagged 1.0.4.

lefou:
Reverted in #5861

PR #5711, opened 2025-08-18
Rapid shutdown/startup cycles fail to connect to the mill daemon

arturaz:
Noticed this in tests, where Mill daemon is being shut down and then immediately spun back up.

Due to the daemon lock being released before server port is removed the client reads the stale server port file and then fails to connect to the server.

This PR fixes the bug by ensuring server port file is deleted before daemon lock is released and adds an integration test to rapidly cycle the mill daemon.

### DIFF
--- a/integration/feature/startup-shutdown/src/StartupShutdownTests.scala
+++ b/integration/feature/startup-shutdown/src/StartupShutdownTests.scala
@@ -9,16 +9,16 @@ object StartupShutdownTests extends UtestIntegrationTestSuite {
   val tests: Tests = Tests {
     test("rapid_cycle") - integrationTest { tester =>
       if (daemonMode) {
-        val NoCycles = 7
+        val numCycles = 7
 
-        for (index <- 1 to NoCycles) {
+        for (index <- 1 to numCycles) {
           println(s"Cycle $index")
 
           val result1 = tester.eval(("resolve", "_"))
           assert(result1.isSuccess)
 
           // No point in running shutdown on the last cycle, as the test suite will shut it down.
-          if (index != NoCycles) {
+          if (index != numCycles) {
             val result2 = tester.eval("shutdown")
             assert(result2.isSuccess)
           }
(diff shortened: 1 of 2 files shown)


integration/feature/startup-shutdown/src/StartupShutdownTests.scala
```diff
@@ -0,0 +1,29 @@
...
+import mill.testkit.UtestIntegrationTestSuite
+
+import utest._
+
+object StartupShutdownTests extends UtestIntegrationTestSuite {
+  // Ensure that you can quickly cycle through startup and shutdown without failures.
+  val tests: Tests = Tests {
+    test("rapid_cycle") - integrationTest { tester =>
+      if (daemonMode) {
+        val NoCycles = 7
```
lihaoyi: Could we call this `numCycles`? `NoCycles` sounds we're saying there are not any cycles, which is not what we want

arturaz: Sure, I'll rename it when the tests pass and then merge it without running the tests :)

libs/daemon/server/src/mill/server/Server.scala
```diff
@@ -456,13 +467,25 @@ object Server {
...
+          val autoCloseable = new AutoCloseable {
+            override def close(): Unit = {
+              beforeClose()
+              l.release()
+              afterClose()
+            }
+          }
+
+          try Some(block(autoCloseable))
+          finally autoCloseable.close()
```
lihaoyi, line 493: Is it acceptable to call `.close()` multiple times? If we're not sure, we should probably guard it with a `var closed: Boolean` to make sure it only gets called once

lihaoyi: I guess it seems impossible since we call `sys.exit` in the early-exit code path

integration/feature/startup-shutdown/src/StartupShutdownTests.scala
lihaoyi: sounds good

Issue #5717, opened 2025-08-19
`semanticDbData` compiles upstream modules twice


arturaz:
When invoking `semanticDbData` on a module with dependencies, those dependencies are compiled using `compile`.

https://github.com/com-lihaoyi/mill/blob/56843c3242c5602e62e965ed8a4225656bb7241c/libs/javalib/src/mill/javalib/SemanticDbJavaModule.scala#L116-L130

https://github.com/com-lihaoyi/mill/blob/56843c3242c5602e62e965ed8a4225656bb7241c/libs/scalalib/src/mill/scalalib/ScalaModule.scala#L624-L644

https://github.com/com-lihaoyi/mill/blob/56843c3242c5602e62e965ed8a4225656bb7241c/libs/javalib/src/mill/javalib/JavaModule.scala#L711-L716

Which means that if you do `__.semanticDbData`, we could get away by just using `semanticDbData` for the module dependencies, but instead run `compile` for all of them and then run another compilation via `semanticDbData`, effectively doing twice the work.

arturaz:
After some experimentation I've managed to cut down `__.semanticDbData` time from 50s to 30s for the gatling codebase by skipping the `compile` step.

PR #5720, opened 2025-08-19
Prevent `semanticDbData` from compiling upstream modules twice

arturaz:
When invoking `semanticDbData` on a module with dependencies, those dependencies are compiled using `compile`.

Which means that if you do `__.semanticDbData`, we could get away by just using `semanticDbData` for the module dependencies, but instead run `compile` for all of them and then run another compilation via `semanticDbData`, effectively doing twice the work.

This PR eliminates the need for `compile`.

Tested manually on gatling codebase, this cuts down `__.semanticDbData` time from 50s to 30s.

---

Fixes https://github.com/com-lihaoyi/mill/issues/5717

### DIFF
--- a/example/cli/builtins/1-builtin-commands/build.mill
+++ b/example/cli/builtins/1-builtin-commands/build.mill
@@ -241,7 +241,6 @@ foo.assembly
 /** Usage
 
 > ./mill plan foo.compileClasspath
-foo.transitiveCompileClasspathTask
 foo.compileResources
 foo.unmanagedClasspath
 ...
--- a/example/fundamentals/out-dir/1-out-files/build.mill
+++ b/example/fundamentals/out-dir/1-out-files/build.mill
@@ -185,11 +185,9 @@ out/mill-daemon
       }
     },
     "foo.compileClasspath": {
-      "foo.compileClasspathTask": {
-        "foo.localCompileClasspath": {
-          "foo.unmanagedClasspath": {},
-          "foo.compileResources": {}
-        }
+      "foo.localCompileClasspath": {
+        "foo.unmanagedClasspath": {},
+        "foo.compileResources": {}
       }
     },
     ...
--- a/integration/ide/bsp-server/resources/snapshots/logging
+++ b/integration/ide/bsp-server/resources/snapshots/logging
@@ -3,8 +3,8 @@
 [1-buildInitialize] Entered buildInitialize
 [1-buildInitialize] Got client semanticdbVersion: * Enabling SemanticDB support.
 [1-buildInitialize] buildInitialize took * msec
-[bsp-init-build.mill-60] [info] compiling * Scala sources to * ...
-[bsp-init-build.mill-60] [info] done compiling
+[bsp-init-build.mill-58] [info] compiling * Scala sources to * ...
+[bsp-init-build.mill-58] [info] done compiling
 [bsp-init] SNAPSHOT
 [2-workspaceBuildTargets] Entered workspaceBuildTargets
 [2-workspaceBuildTargets] Evaluating 13 tasks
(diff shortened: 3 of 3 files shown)


libs/javalib/src/mill/javalib/JavaModule.scala
```diff
@@ -964,14 +971,19 @@ trait JavaModule
         bspLocalRunClasspath(needsToMergeResourcesIntoCompileDest)()
     }
 
+  /**
+   * [[compileClasspathFor]] for regular compilations.
+   *
+   * Keep return value in sync with [[bspCompileClasspath]].
+   */
+  def compileClasspath: T[Seq[PathRef]] = Task { compileClasspathFor(CompileFor.Regular)() }
```
lefou: Does this need to be public?

libs/scalalib/src/mill/scalalib/ScalaModule.scala
```diff
@@ -598,60 +598,70 @@ trait ScalaModule extends JavaModule with TestModule.ScalaModuleBase
...
+          }
+
+          SemanticDbJavaModule.SemanticDbData(compilationResult, semanticDbFiles)
         }
-      }
+    }
+
+  // binary compatibility forwarder
+  override def semanticDbData: T[PathRef] = Task {
+    semanticDbDataDetailed().semanticDbFiles
```
lefou: Can this call `super.semanticDbData()` instead?

libs/javalib/src/mill/javalib/SemanticDbJavaModule.scala
```diff
@@ -17,15 +17,26 @@ import mill.javalib.api.internal.{JavaCompilerOptions, ZincCompileJava}
...
   def zincReportCachedProblems: T[Boolean]
   def zincIncrementalCompilation: T[Boolean]
   def allSourceFiles: T[Seq[PathRef]]
   def compile: T[mill.javalib.api.CompilationResult]
+
+  def compileFor(compileFor: CompileFor): Task[mill.javalib.api.CompilationResult] =
+    compileFor match {
+      case CompileFor.Regular => compile
+      case CompileFor.SemanticDb => semanticDbDataDetailed.map(_.compilationResult)
+    }
```
lefou: Does this need to be public?

libs/javalib/src/mill/javalib/JavaModule.scala
arturaz: It was previously public, so changing the visibility would break binary compatibility.

libs/scalalib/src/mill/scalalib/ScalaModule.scala
arturaz: Unfortunately no, because then a forwarder is generated which breaks binary compatibility.

libs/javalib/src/mill/javalib/JavaModule.scala
lefou: Looks like I commented on the wrong method.

See https://github.com/com-lihaoyi/mill/pull/5720/files#r2284957198

libs/javalib/src/mill/javalib/SemanticDbJavaModule.scala
arturaz: It does not. I'll change it.

libs/javalib/src/mill/javalib/JavaModule.scala
```diff
@@ -964,14 +971,19 @@ trait JavaModule
...
+
   /**
    * All classfiles and resources from upstream modules and dependencies
    * necessary to compile this module.
-   *
-   * Keep in sync with [[bspCompileClasspath]]
    */
-  override def compileClasspath: T[Seq[PathRef]] = Task {
-    resolvedMvnDeps() ++ transitiveCompileClasspath() ++ localCompileClasspath()
+  override def compileClasspathFor(compileFor: CompileFor): Task[Seq[PathRef]] = Task {
```
lefou: Does this need to be public?

arturaz: No, I'll changed everything that refers to `CompileFor` to `private[mill]`, don't want to push it while tests are running.

arturaz:
`./mill example.javalib.publishing[7-native-image].local.daemon.testForked` seems flaky, as it passes successfully on my machine.

Issue #5676, opened 2025-08-08
Annotation processors with Micronaut and KSP


mallport:
Hello! I'm playing around with setting up Mill in a project with Kotlin and Micronaut. I'm having troubles setting up KSP correctly in order to finagle the annotation processors required in Micronaut (micronaut-inject-java, for example). Micronaut, when running tests that use these annotations, complains that they cannot be found:

```
[146] Failed to evaluate condition [io.micronaut.test.extensions.junit5.MicronautJunit5Extension]: @MicronautTest used on test but no bean definition for the test present. This error indicates a misconfigured build or IDE. Please check 'micronaut.processing.group' or 'micronaut.processing.annotations' in your build. Add the 'micronaut-inject-java' annotation processor to your test processor path (for Java this is the testAnnotationProcessor scope, for Kotlin kaptTest and for Groovy testCompile). See the documentation for reference: https://micronaut-projects.github.io/micronaut-test/latest/guide/, took 0.005s
```

 I see there's an example using KSP with Dagger [here](https://github.com/com-lihaoyi/mill/blob/main/example/kotlinlib/basic/7-dependency-injection/build.mill#L23), which is mostly what I'm going by.

build.mill is attached below, would really appreciate any pointers 😄

```
//| mill-version: 1.0.3

package build
import mill.*
import kotlinlib.*
import kotlinlib.ksp.KspModule

import scala.reflect.io.File

object whodat extends MicronautModule {
  def micronautVersion = "4.9.2"
  def mvnDeps: Task.Simple[Seq[Dep]] = Seq(
    mvn"io.micronaut:micronaut-http-server-netty:$micronautVersion",
    mvn"io.micronaut.serde:micronaut-serde-jackson:2.10.1",
    mvn"ch.qos.logback:logback-classic:1.5.3",
    mvn"io.micronaut.kotlin:micronaut-kotlin-runtime:4.7.0"
  )

  object test extends KspTests, TestModule.Junit5 {
    def mvnDeps = super.mvnDeps() ++ Seq(
      mvn"io.micronaut:micronaut-http-client:$micronautVersion",
      mvn"io.micronaut.test:micronaut-test-junit5:4.4.0",
      mvn"io.micronaut:micronaut-inject-kotlin:4.9.9",
      mvn"io.micronaut:micronaut-inject-java:4.9.9",
    )

    def kotlincOptions = super.kotlincOptions() ++ Seq(
      "-no-reflect",
      "-verbose"
    )

[...]

PR #5727, opened 2025-08-20
KSP 2 support

vaslabs:
TBC: This so far is a proof of concept, and I managed to make it work with dagger

Still needs work and some cleanup but feel free to comment, it would help me to come back to it,  as I'm not gonna do much for a few days due to vacations.

Motivation: https://github.com/com-lihaoyi/mill/issues/5676


libs/kotlinlib/src/mill/kotlinlib/ksp/Ksp2Module.scala
```diff
@@ -0,0 +1,152 @@
...
+
+/**
+ * Sets up the kotlin compiler for using KSP (Kotlin Symbol Processing)
+ * by plugging in the symbol-processing and symbol-processing-api dependencies.
+ *
+ * Use of kotlin-compiler-embedded is also recommended (and thus enabled by default)
+ * to avoid any classpath conflicts between the compiler and user defined plugins!
+ */
+@mill.api.experimental
+trait Ksp2Module extends KotlinModule { outer =>
```
lefou: Looks like lots of code duplication between `KspModule` and `Ksp2Module`. Probably worth to think about moving those shared tasks in a `KspBaseModule` trait.

vaslabs:
Note: for micronaut I'm getting similar errors reported [here](https://github.com/micronaut-projects/micronaut-core/issues/11553) , I may need to dig deeper eventually

vaslabs:
ok, micronaut worked as well, turned out it needed jdk home to be passed as a parameter so it could see the java sources

https://github.com/vaslabs/mill-ksp2-micronaut-test

will pass it to @souvlakias for general cleanup and enhancements until I'm back

libs/kotlinlib/src/mill/kotlinlib/ksp/Ksp2Module.scala
vaslabs: thanks, I will review this towards the end and see, potentially yes but am a bit sceptical at the moment, due to the fact that they don't have anything in common functionally apart from the "KSP", one works through the kotlin compiler, the other works through class loading the SymbolProcessing (via the KSPJvmMain)  . I still have some experimenting to do with hilt and multi-module setups and will have a clearer picture towards the end

lefou: My perception was, that anything the user needs to configure is almost identical. So, the differences are in the implementation of the working tasks and their dependencies.

vaslabs: It makes sense, I need to iterate on it and see how it comes along. It will probably start to get complicated for me when I start testing with Hilt, so I'll probably break the work into multiple PRs and I'd like to keep the 2 separate if I think it's easier to get to the end result, and consolidate towards the end. But I'm not sure yet, just sharing my thoughts

vaslabs: ok, @souvlakias made it already work for the hilt todo example and I can see a path to consolidating, basically as you said, and I imagine users can do 
`object app extends AndroidHiltSupport with KspModule` or `object app extends AndroidHiltSupport with Ksp2Module` and the aim is to have no user side changes or at least minimal depending on generatedSources setup

vaslabs:
@souvlakias a few tasks remaining:

- [x] Let's add [the micronaut example](https://github.com/vaslabs/mill-ksp2-micronaut-test) with ksp 2
- [x] Let's test with the  [pokemon multimodule example](https://github.com/vaslabs/Pokedex_Compose_Multi_Module/tree/testing-mill)
- [x] And finally let's consolidate KspModule and Ksp2Module so we keep user changes at minimal when they swap from one to the other

libs/androidlib/src/mill/androidlib/hilt/AndroidHiltSupport.scala
```diff
@@ -16,38 +15,29 @@ import mill.{T, Task}
...
+ *
+ * object app extends Ksp2Module with AndroidHiltSupport { ... }
+ * ```
  */
 @mill.api.experimental
-trait AndroidHiltSupport extends KspModule with AndroidKotlinModule {
-
-  override def kspClasspath: T[Seq[PathRef]] =
-    super.kspClasspath()
+trait AndroidHiltSupport extends KspBaseModule with AndroidKotlinModule {
```
vaslabs: instead of with we can use the scala 3 with `,`

souvlakias:
[Pokemon Mulitimodule Example](https://github.com/vaslabs-ltd/Pokedex_Compose_Multi_Module/tree/testing-mill-ksp2) with KSP2
image

vaslabs:
I dare say this is ready for review. Please have a look @lefou @lihaoyi

vaslabs:
(failures seem to be network issues , my copy for ref: https://github.com/vaslabs-ltd/mill/actions/runs/17214220450)

lefou:
> It turns out, that integration wise, it's much easier to run the symbol processor with KSP 2 via a CLI wrapper that it is for KSP 1 as a kotlin compiler plugin

I think most tools are easiest to use by wrapping their CLIs. But JVM CLI tools are typically slower to start than integrations. In Mill we typically go the extra mile to get those speedups. We should keep this in mind when designing the traits, so no API changes are needed should we later switch to a JVM-integrated implementation.

libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -14,9 +14,19 @@ import java.io.File
...
 @mill.api.experimental
-trait KspModule extends KotlinModule { outer =>
+trait KspModule extends KspBaseModule { outer =>
+
+  /**
+   * The version of the Kotlin language to use.
+   * This is used to determine the KSP version to use.
+   * Defaults to `1.9` which is compatible with KSP 1.x.
+   */
+  def kspLanguageVersion: T[String] = "1.9"
```
lefou: Can we avoid the hard-coded version here? If we need to sanity check for compatibility, we can do it in a downstream task.

libs/kotlinlib/src/mill/kotlinlib/ksp/Ksp2Module.scala
```diff
@@ -0,0 +1,142 @@
...
+ */
+@mill.api.experimental
+trait Ksp2Module extends KspBaseModule { outer =>
+
+  def kspVersion: T[String] = "2.0.2"
+  def kspJvmTarget: T[String] = "11"
+
+  def kspLanguageVersion: T[String] = "2.0"
+
+  def kspApiVersion: T[String] = "2.0"
```
lefou: Please avoid hard-coded versions. Instead, give good documentation which version to choose and do a sanity-check in a downstream task. I'm fine with deriving some versions from some user-provided task, e.g. derive the default for `kspApiVersion` from `kspVersion`.

libs/kotlinlib/src/mill/kotlinlib/ksp/KspBaseModule.scala
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Base trait for KSP (Kotlin Symbol Processing) modules.
+ * This trait defines the common interface that both KspModule and Ksp2Module implement.
+ */
+@mill.api.experimental
+trait KspBaseModule extends KotlinModule {
+
+  def kspLanguageVersion: T[String]
+
+  def kspVersion: T[String]
```
lefou: Can we have some docs here. E.g. what is the difference?

libs/kotlinlib/src/mill/kotlinlib/ksp/Ksp2Module.scala
```diff
@@ -0,0 +1,142 @@
...
+      mvn"com.google.devtools.ksp:symbol-processing-common-deps:${kotlinVersion()}-${kspVersion()}",
+      mvn"org.jetbrains.kotlinx:kotlinx-coroutines-core-jvm:1.10.2"
+    )
+  }
+
+  /**
+   * The libraries that should be passed to the KSP compiler.
+   * This defaults to the compile classpath of the module.
+   */
+  def kspLibraries: T[Seq[PathRef]] = compileClasspath()
```
lefou: Isn't that also needed in KSP1? Can we move it into `KspBaseModule`?

vaslabs:
> > It turns out, that integration wise, it's much easier to run the symbol processor with KSP 2 via a CLI wrapper that it is for KSP 1 as a kotlin compiler plugin
> 
> I think most tools are easiest to use by wrapping their CLIs. But JVM CLI tools are typically slower to start than integrations. In Mill we typically go the extra mile to get those speedups. We should keep this in mind when designing the traits, so no API changes are needed should we later switch to a JVM-integrated implementation.

indeed, [there is a way](https://github.com/google/ksp/blob/main/docs/ksp2entrypoints.md) to avoid using the jvm cli, the arguments will be the same and only the generatedSourcesWithKSP will change, as it did between KSP 1 and 2 .

lefou:
Is it possibly to decide solely on the result values of the `kspApiVersion` task (or the other version tasks) which tools to invoke? That way, we wouldn't need to have two different `trait`s.

vaslabs:
> Is it possibly to decide solely on the result values of the `kspApiVersion` task (or the other version tasks) which tools to invoke? That way, we wouldn't need to have two different `trait`s.

I'll try it out, in theory it should be possible to choose a task implementation based on the version. I wanted to avoid it in the first iteration but with the current progress I don't see why not

vaslabs:
<details>

<summary>Switching between KSP and KSP 2 test with Hilt example via androidtodo</summary>

```diff
diff --git a/example/thirdparty/androidtodo/build.mill b/example/thirdparty/androidtodo/build.mill
index ab845eaf514..36b9ad54603 100644
--- a/example/thirdparty/androidtodo/build.mill
+++ b/example/thirdparty/androidtodo/build.mill
@@ -1,12 +1,13 @@
 package build
 
 import mill.*, androidlib.*, kotlinlib.*
-import kotlinlib.ksp.Ksp2Module
+import kotlinlib.ksp.KspModule
 import hilt.AndroidHiltSupport
 
 object Versions {
   val kotlinVersion = "2.2.0"
 
+  val kspLanguageVersion = "1.9"
   val kspVersion = "2.0.2"
   val androidCompileSdk = 35
   val androidMinSdk = 26
@@ -19,8 +20,9 @@ object androidSdkModule0 extends AndroidSdkModule {
 
 // Mill configuration for the Android Todo App project.
 object app extends AndroidAppKotlinModule, AndroidR8AppModule, AndroidBuildConfig,
-      Ksp2Module, AndroidHiltSupport {
+      KspModule, AndroidHiltSupport {
 
+  override def kspLanguageVersion = Versions.kspLanguageVersion
   def kotlinVersion = Versions.kotlinVersion
   def kspVersion = Versions.kspVersion
 
@@ -107,10 +109,11 @@ object app extends AndroidAppKotlinModule, AndroidR8AppModule, AndroidBuildConfi
     mvn"com.google.dagger:hilt-android-compiler:2.56"
   )
 
-  object test extends AndroidAppKotlinTests, Ksp2Module, AndroidHiltSupport, TestModule.Junit4 {
+  object test extends AndroidAppKotlinTests, KspModule, AndroidHiltSupport, TestModule.Junit4 {
 
     def moduleDeps = super.moduleDeps ++ Seq(`shared-test`)
 
+    override def kspLanguageVersion = Versions.kspLanguageVersion
     def androidEnableCompose = true
 
     override def kspVersion = Versions.kspVersion
@@ -144,9 +147,10 @@ object app extends AndroidAppKotlinModule, AndroidR8AppModule, AndroidBuildConfi
   }
 
   object androidTest extends AndroidAppKotlinInstrumentedTests, AndroidR8AppModule,
-        Ksp2Module, AndroidHiltSupport {
+        KspModule, AndroidHiltSupport {
 
     def moduleDeps = super.moduleDeps ++ Seq(`shared-test`)
+    override def kspLanguageVersion = Versions.kspLanguageVersion
 
     def testFramework = "com.example.android.architecture.blueprints.todoapp.CustomTestRunner"
 
@@ -208,9 +212,10 @@ object app extends AndroidAppKotlinModule, AndroidR8AppModule, AndroidBuildConfi
 
 }
 
-object `shared-test` extends AndroidKotlinModule, Ksp2Module, AndroidHiltSupport {
+object `shared-test` extends AndroidKotlinModule, KspModule, AndroidHiltSupport {
 
   def moduleDeps = Seq(app)
+  override def kspLanguageVersion = Versions.kspLanguageVersion
 
   def kotlinVersion = Versions.kotlinVersion
   def kspVersion = Versions.kspVersion
```

</details>

vaslabs:
> Is it possibly to decide solely on the result values of the `kspApiVersion` task (or the other version tasks) which tools to invoke? That way, we wouldn't need to have two different `trait`s.

I tried a few permutations, I have a few issues that I think turn out to make it more ugly:

- the ksp 1 as a kotlin compiler plugin, needs compiler args, which are not needed in ksp2, and makes the user interface confusing
- it's complex to manage when the tasks run with conditional statements - although that could be doable

I think, with this issues + their experimental state, might be easier to explicitly allow for a switch.

I did cleanup the versions and the examples though, KSP 2 being the newest, seems to be working with the derived language version

lefou:
To all tooling that build upon this, having a single trait would be beneficial. We could have a KSP1/KSP2 switch as pure `def` so tasks could have different dependencies depending on the mode. 

Do whatever you think works best for you. As long as this is experimental, we are free to change it later, too.

libs/kotlinlib/src/mill/kotlinlib/ksp/Ksp2Module.scala
vaslabs: I've replaced all of them with automatic detection from the same ones as the kotlin module, but good to have the option to defined them separately

libs/kotlinlib/src/mill/kotlinlib/ksp/Ksp2Module.scala
```diff
@@ -22,9 +22,11 @@ trait Ksp2Module extends KspBaseModule { outer =>
 
   /**
    * The JVM target version for the KSP compilation step.
-   * Default is "11".
+   * Default is derived from `java.specification.version`.
    */
-  def kspJvmTarget: T[String] = "11"
+  def kspJvmTarget: T[String] = Task {
+    System.getProperty("java.specification.version")
+  }
```
lefou: This needs to be an `Task.input`.

vaslabs:
> To all tooling that build upon this, having a single trait would be beneficial. We could have a KSP1/KSP2 switch as pure `def` so tasks could have different dependencies depending on the mode.
> 
> Do whatever you think works best for you. As long as this is experimental, we are free to change it later, too.

I'd prefer to work on further consolidation in another PR then, I think we have enough here, I'm not against, I would just need more time, more experimentation and more tests to find a good combination. 

For example, ksp 1 needs:

```scala
  def kspKotlincOptions: T[Seq[String]] = Task {
    if (kspLanguageVersion().isBlank) {
      throw new RuntimeException("KSP needs a compatible language version to be set!")
    }
    kotlincOptions() ++ Seq(
      "-Xallow-unstable-dependencies",
      "-no-reflect",
      "-no-stdlib",
      "-language-version",
      kspLanguageVersion()
    )
  }
  ```
  
  but ksp 2 does not ned any of these.
  
  So a unified trait, would have them defined just in case, and they wouldn't be used for KSP 2, wouldn't that be weird? especially since ksp 1 is being phased out for kotlin >=2.2

File: libs/kotlinlib/src/mill/kotlinlib/ksp/Ksp2Module.scala (hunk 7 above)
Comment (lefou) [reply libs/kotlinlib/src/mill/kotlinlib/ksp/Ksp2Module.scala]: TBH, I don't like this default. Why is this configured explicitly, while we don't configure the JVM target version in Kotlin/Java/Scala modules? Isn't this configured via the KSP options? What happens if a user also defines `jvmId`?

File: libs/kotlinlib/src/mill/kotlinlib/ksp/Ksp2Module.scala [hunk 8]
```diff
@@ -0,0 +1,156 @@
...
+
+  /**
+   * The jars needed to run KSP 2 via `com.google.devtools.ksp.cmdline.KSPJvmMain` .
+   *
+   * The versions are computed from [[kotlinVersion]]-[[kspVersion]]
+   *
+   * For finding the right versions, also see [[https://github.com/google/ksp/releases]]
+   * For more info go to [[https://github.com/google/ksp/blob/main/docs/ksp2cmdline.md]]
+   */
+  def kspDeps: T[Seq[Dep]] = Task {
```
lefou, line 47: So these deps are rather some "internal" tooling classpath but nothing the user needs to configure?

vaslabs: yes, should we have them private?

libs/kotlinlib/src/mill/kotlinlib/ksp/Ksp2Module.scala
vaslabs: [this is needed](https://github.com/google/ksp/blob/main/docs/ksp2cmdline.md#using-ksp2-from-command-line) from the symbol processor, initially I had it hardcoded to 11, but I tested with 17 and it worked too, so I figured to have it getting it from the system property. As my understanding goes, it's not necessarily coupled with the rest of the compilation stages, it just creates sources or compiles code against that jvm target during the ksp stage, so someone could lower it for compatibility reasons (e.g. room processor might not support a target of 25? I'm just guessing).

lefou:
> > To all tooling that build upon this, having a single trait would be beneficial. We could have a KSP1/KSP2 switch as pure `def` so tasks could have different dependencies depending on the mode.
> > Do whatever you think works best for you. As long as this is experimental, we are free to change it later, too.
> 
> I'd prefer to work on further consolidation in another PR then, I think we have enough here, I'm not against, I would just need more time, more experimentation and more tests to find a good combination.
> 
> For example, ksp 1 needs:
> 
> ```scala
>   def kspKotlincOptions: T[Seq[String]] = Task {
>     if (kspLanguageVersion().isBlank) {
>       throw new RuntimeException("KSP needs a compatible language version to be set!")
>     }
>     kotlincOptions() ++ Seq(
>       "-Xallow-unstable-dependencies",
>       "-no-reflect",
>       "-no-stdlib",
>       "-language-version",
>       kspLanguageVersion()
>     )
>   }
> ```
> 
> but ksp 2 does not ned any of these.
> 
> So a unified trait, would have them defined just in case, and they wouldn't be used for KSP 2, wouldn't that be weird? especially since ksp 1 is being phased out for kotlin >=2.2


```scala
enum KspMode {
  case Ksp1, Ksp2
}

trait KspModule {

  def kspMode: KspMode = Ksp2 // default to latest

   def kspKotlincOptions: T[Seq[String]] = kspMode match {
     case Ksp1 => Task {
       if (kspLanguageVersion().isBlank) {
         throw new RuntimeException("KSP needs a compatible language version to be set!")
       }
       kotlincOptions() ++ Seq(
         "-Xallow-unstable-dependencies",
         "-no-reflect",
         "-no-stdlib",
         "-language-version",
         kspLanguageVersion()
       )
    case Ksp2 => Task {
       Seq()
    }
  }
}

vaslabs:
> > > To all tooling that build upon this, having a single trait would be beneficial. We could have a KSP1/KSP2 switch as pure `def` so tasks could have different dependencies depending on the mode.
> > > Do whatever you think works best for you. As long as this is experimental, we are free to change it later, too.
> > 
> > 
> > I'd prefer to work on further consolidation in another PR then, I think we have enough here, I'm not against, I would just need more time, more experimentation and more tests to find a good combination.
> > For example, ksp 1 needs:
> > ```scala
> >   def kspKotlincOptions: T[Seq[String]] = Task {
> >     if (kspLanguageVersion().isBlank) {
> >       throw new RuntimeException("KSP needs a compatible language version to be set!")
> >     }
> >     kotlincOptions() ++ Seq(
> >       "-Xallow-unstable-dependencies",
> >       "-no-reflect",
> >       "-no-stdlib",
> >       "-language-version",
> >       kspLanguageVersion()
> >     )
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > but ksp 2 does not ned any of these.
> > So a unified trait, would have them defined just in case, and they wouldn't be used for KSP 2, wouldn't that be weird? especially since ksp 1 is being phased out for kotlin >=2.2
> 
> ```scala
> enum KspMode {
>   case Ksp1, Ksp2
> }
> 
> trait KspModule {
> 
>   def kspMode: KspMode = Ksp2 // default to latest
> 
>    def kspKotlincOptions: T[Seq[String]] = kspMode match {
>      case Ksp1 => Task {
>        if (kspLanguageVersion().isBlank) {
>          throw new RuntimeException("KSP needs a compatible language version to be set!")
>        }
>        kotlincOptions() ++ Seq(
>          "-Xallow-unstable-dependencies",
>          "-no-reflect",
>          "-no-stdlib",
>          "-language-version",
>          kspLanguageVersion()
>        )
>     case Ksp2 => Task {
>        Seq()
>     }
>   }
> }
> ```

I meant, how the user would have it exposed as an interface despite not being used for Ksp2, we should also communicate that, like 
"This is only used with Ksp1" <- would something like that be enough?

libs/kotlinlib/src/mill/kotlinlib/ksp/Ksp2Module.scala
vaslabs: I don't know if jvm id would affect it. is there a better way to decide on a default?

libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
vaslabs: removed and added a check as well to prompt user to use <= 1.9

vaslabs:
btw, with fmt, I'm getting these issues locally too

https://github.com/com-lihaoyi/mill/actions/runs/17237231451/job/48904838973?pr=5727#step:4:797

but intermittently

vaslabs:
I've just realised something, I'll do another try to have one trait.

EDIT: The idea with the current implementation is that tooling will extend base and users will choose implementation between 1 and 2 by extending either one of KSPModule or KSP2Module. But I think we can make a sensible default decision as you've recommended, I'll give it another go

libs/kotlinlib/src/mill/kotlinlib/ksp/Ksp2Module.scala
vaslabs: I saw the other examples and followed the same pattern, kspJvmTarget needs to be passed by the user now :+1:

vaslabs:
> I've just realised something, I'll do another try to have one trait.
> 
> EDIT: The idea with the current implementation is that tooling will extend base and users will choose implementation between 1 and 2 by extending either one of KSPModule or KSP2Module. But I think we can make a sensible default decision as you've recommended, I'll give it another go

I've kept it as is for now, with this idea, tooling can use KspBaseModule and added docs on its unimplemented `generatedSourcesWithKSP` for the available choices, so the end users can make the decision. It's not perfect as we don't default for them a choice but it helps me iterate on improvements later and diagnose/fix any incoming issues separately

Hope this is ok for now, and we can revisit the consolidation later, with the optimisation as well, as we'll have the examples serve as regression and I have few ideas in mind how to do both, so I'd like to avoid extending this PR further

lihaoyi:
@vaslabs I'll mark this ready for review so we can see CI results

vaslabs:
> @vaslabs I'll mark this ready for review so we can see CI results


Thanks, I was waiting for my repo to finish :sweat_smile:

vaslabs:
Also, I was thinking, for future development, there's KSP for native and JS too, so maybe we should create some roadmap, add support for those too, and see how everything comes along, would also help us understand the integrations better and come up with potentially better solutions/interfaces in the end

libs/kotlinlib/src/mill/kotlinlib/ksp/Ksp2Module.scala
lefou: Maybe. We want a better name then, maybe `kspToolsDeps`.

lefou: So we end up also with `kspToolsDepsResolved` or `kspToolsClasspath`, which I think is more consistent with other modules.

lefou: Let keep them public, so users can use locally modified tools. Something, that Mill is really predestined for.

vaslabs: ok, will change to `kspToolsDeps` and `kspToolsClasspath`

vaslabs:
oh I didn't push the kspDeps rename, I will do on the follow up PRs

PR #5747, opened 2025-08-26
Initial Groovy setup

lostiniceland:
I have been working on Groovy support since I would like to use Mill for projects where tests are written in Spock/Groovy.

It is basically a modified version of the Kotlin support.

Whats currently working:
- Groovy compilation (tested with 4 & 5)
- Maven project layout with an additional convenience option to configure Groovy for tests only
- JUnit5 test execution
- Spock tests
- CompileStatic
- Configure Bytecode Version and Preview

If anyone would like to provide feedback I would appreciate it. 

Pull request: https://github.com/com-lihaoyi/mill/pull/5747

### DIFF
--- a/example/groovylib/basic/3-simple/build.mill
+++ b/example/groovylib/basic/3-simple/build.mill
@@ -41,8 +41,7 @@ compile(GroovyModule.scala:...)
     ...
 Inputs:
     upstreamCompileOutput
-    allJavaSourceFiles
-    allGroovySourceFiles
+    allSources
     compileClasspath
 
 > ./mill compile # compile sources into classfiles


libs/groovylib/src/mill/groovylib/GroovyModule.scala
```diff
@@ -0,0 +1,271 @@
...
+
+  override def bomMvnDeps: T[Seq[Dep]] = super.bomMvnDeps() ++ Seq(
+    mvn"org.apache.groovy:groovy-bom:${groovyVersion()}",
+  )
+
+  /**
+   * All individual source files fed into the compiler.
+   */
+  override def allSourceFiles: T[Seq[PathRef]] = Task {
+    Lib.findSourceFiles(allSources(), Seq("groovy", "java")).map(PathRef(_))
```
lefou: This should reuse `allJavaSourceFiles` and `allGroovySourceFiles`.

We might also think about moving `allJavaSourceFile` up into `JavaModule`.

libs/groovylib/src/mill/groovylib/GroovyModule.scala
```diff
@@ -0,0 +1,271 @@
...
+
+  /**
+   * Compiler Plugin dependencies.
+   */
+  def groovycPluginMvnDeps: T[Seq[Dep]] = Task { Seq.empty[Dep] }
+
+  /**
+   * The resolved plugin jars
+   */
+  def groovycPluginJars: T[Seq[PathRef]] = Task {
```
lefou:
```suggestion
  def groovyCompilerPluginJars: T[Seq[PathRef]] = Task {
```

libs/groovylib/src/mill/groovylib/GroovyModule.scala
```diff
@@ -0,0 +1,271 @@
...
+
+    val compilerDep =  mvn"org.apache.groovy:groovy:$gv"
+
+    Seq(compilerDep)
+  }
+
+  /**
+   * Compiler Plugin dependencies.
+   */
+  def groovycPluginMvnDeps: T[Seq[Dep]] = Task { Seq.empty[Dep] }
```
lefou: Should use the same prefix as the tasks above.

```suggestion
  def groovyCompilerPluginMvnDeps: T[Seq[Dep]] = Task { Seq.empty[Dep] }
```

libs/groovylib/api/src/mill/groovylib/worker/api/GroovyWorker.scala
```diff
@@ -0,0 +1,12 @@
+package mill.groovylib.worker.api
+
+import mill.api.TaskCtx
+import mill.api.Result
+import mill.javalib.api.CompilationResult
+
+trait GroovyWorker {
+
+  def compile(sourceFiles: Seq[os.Path], classpath: Seq[os.Path], outputDir: os.Path)(implicit
```
lefou: Doesn't Groovy has any compiler options to respect? I don't see a parameter for them.

libs/groovylib/package.mill
```diff
@@ -0,0 +1,50 @@
+package build.libs.groovylib
+
+// imports
+import mill.*
+import mill.contrib.buildinfo.BuildInfo
+import mill.scalalib.*
+import millbuild.*
+
+object `package` extends MillStableScalaModule with BuildInfo {
```
lefou: Probably too early to declare a "stable" API.

```suggestion
object `package` extends MillPublishScalaModule with BuildInfo {
```

libs/groovylib/package.mill
```diff
@@ -0,0 +1,50 @@
...
+  def buildInfoObjectName = "Versions"
+  def buildInfoMembers = Seq(
+    BuildInfo.Value("groovyVersion", Deps.groovyVersion, "Version of Groovy")
+  )
+
+  trait MillGroovyModule extends MillPublishScalaModule {
+    override def javacOptions = super.javacOptions() ++ {
+      val release =
+        if (scala.util.Properties.isJavaAtLeast(11)) Seq("-release", "8")
+        else Seq("-source", "1.8", "-target", "1.8")
```
lefou: Mill already requires Java 11. No need to support Java 8.

libs/groovylib/src/mill/groovylib/GroovyModule.scala
```diff
@@ -0,0 +1,269 @@
...
+   * The dependencies of this module.
+   * Defaults to add the groovy-stdlib dependency matching the [[groovyVersion]].
+   */
+  override def mandatoryMvnDeps: T[Seq[Dep]] = Task {
+    super.mandatoryMvnDeps() ++ Seq(
+      mvn"org.apache.groovy:groovy:${groovyVersion()}"
+    )
+  }
+
+  private def jvmWorkerRef: ModuleRef[JvmWorkerModule] = jvmWorker
```
lefou: I think this should not be private, otherwise, users can't customize it.

libs/groovylib/src/mill/groovylib/GroovyModule.scala
lefou: Same for other tasks. We aim for some consistency.

libs/groovylib/src/mill/groovylib/GroovyModule.scala
```diff
@@ -0,0 +1,269 @@
...
+        incrementalCompilation = true
+      ),
+      javaHome = javaHome,
+      javaRuntimeOptions = jOpts.runtime,
+      reporter = compileProblemReporter,
+      reportCachedProblems = reportOldProblems
+    )
+  }
+
+  private[groovylib] def internalReportOldProblems: Task[Boolean] = zincReportCachedProblems
```
lefou: What's the purpose of this? If it's private it can't be customized, so no need to have it at all.

libs/groovylib/src/mill/groovylib/GroovyModule.scala
```diff
@@ -0,0 +1,269 @@
...
+      BspModuleApi.LanguageId.Groovy
+    ),
+    canCompile = true,
+    canRun = true
+  )
+
+  override def prepareOffline(all: Flag): Command[Seq[PathRef]] = Task.Command {
+    (
+      super.prepareOffline(all)() ++
+        groovyCompilerClasspath()
```
lefou, line 290: Does this include potential compiler plugin dependencies?

libs/groovylib/src/mill/groovylib/GroovyModule.scala
```diff
@@ -0,0 +1,269 @@
...
+    override def groovyLanguageVersion: T[String] = outer.groovyLanguageVersion()
+    override def groovyVersion: T[String] = Task { outer.groovyVersion() }
+    override def groovycPluginMvnDeps: T[Seq[Dep]] =
+      Task { outer.groovycPluginMvnDeps() }
+      // TODO: make Xfriend-path an explicit setting
+    override def groovycOptions: T[Seq[String]] = Task {
+      // FIXME
+      outer.groovycOptions().filterNot(_.startsWith("-Xcommon-sources")) ++
+        Seq(s"-Xfriend-paths=${outer.compile().classes.path.toString()}")
+    }
```
lefou: Does Groovy has an `-Xfriend-paths` option at all?

libs/groovylib/src/mill/groovylib/TestGroovyMavenModule.scala
```diff
@@ -0,0 +1,25 @@
...
+   * The name of this module's folder within `src/`: e.g. `src/test/`, `src/integration/`,
+   * etc. Defaults to the name of the module object, but can be overridden by users
+   */
+  def testModuleName: String = moduleCtx.segments.last.value
+
+  private def testSources = Task.Sources(moduleDir / os.up / "src" / testModuleName / "groovy")
+  override def sources: T[Seq[PathRef]] = testSources()
+
+  private def testResources = Task.Sources(moduleDir / os.up / "src" / testModuleName / "resources")
+  override def resources: T[Seq[PathRef]] = testResources()
```
lefou: TBH, I don't like this. This _assumes_, the outer module is located at one level above, but effectively, Mill does not enforce it, so this is potentially using the wrong sources or it blows up.

To properly set it up, you might need to define as well the "outer" trait which holds the inner test trait.

```scala
trait JavaMavenModuleWithGroovyTests extends JavaMavenModule { outer =>
  trait GroovyMavenTests extends JavaMavenTests {
    ...
  }
}
```

lefou (commented):
Thanks for opening this PR and welcome to Mill.

I had a quick scan and left some comments where I spotted something. Overall, this looks reasonable, but there might be some more places copied from the Kotlin implementation which aren't adequate for Groovy.

libs/groovylib/src/mill/groovylib/TestGroovyMavenModule.scala
lostiniceland: Yeah, I didn't like this either but the only reason where you need this is when you already define an outer module which has another language-module and only your test sources differ. In case you need a pure Groovy test-module you can just use GroovyMavenModule.

I will try your suggestion to get rid of the assumption going up a level.

libs/groovylib/package.mill
lostiniceland: I dont know the implications of this. Can you explain what to use instead? There seems to be no "xxxBetaModule".

libs/groovylib/src/mill/groovylib/GroovyModule.scala
lostiniceland: Since there is no real Groovy compiler in the sense of Javac, Scala or Kotlin (there is Groovyc, which is just a CLI wrapper) I guess there is no need for this. The core Groovy dependency is able to compile and you can just extends the normal classpath. Guess this is due to Groovy being a dynamic language where everything is resolved at runtime anyways. The only difference could be when you use `@CompileStatic` ([doc](https://docs.groovy-lang.org/latest/html/gapi/groovy/transform/CompileStatic.html)). I will add this to the tasks to be evaluated.

libs/groovylib/package.mill
lefou: I already provided the fix as suggestion. `MillStableScalaModule` is a `MillPublishScalaModule` which also enforces binary compatibility with older releases, which we don't want.

libs/groovylib/src/mill/groovylib/GroovyModule.scala
```diff
@@ -0,0 +1,269 @@
...
+      } else {
+        // it's Java only
+        compileJava
+      }
+    }
+
+  /**
+   * Additional Groovy compiler options to be used by [[compile]].
+   */
+  def groovycOptions: T[Seq[String]] = Task { Seq.empty[String] }
```
lefou: What is this task for? @lostiniceland, you said earlier, that the groovy compiler does not take any options, so this task doesn't make sense, does it?

libs/groovylib/src/mill/groovylib/GroovyModule.scala
```diff
@@ -0,0 +1,269 @@
...
+
+  /**
+   * Aggregation of all the options passed to the Groovy compiler.
+   * In most cases, instead of overriding this Target you want to override `groovycOptions` instead.
+   */
+  def allGroovycOptions: T[Seq[String]] = Task {
+    groovycOptions()
+  }
```
lefou: This task looks obsolete, too.

libs/groovylib/src/mill/groovylib/GroovyModule.scala
```diff
@@ -0,0 +1,269 @@
...
+
+    override def groovyLanguageVersion: T[String] = outer.groovyLanguageVersion()
+    override def groovyVersion: T[String] = Task { outer.groovyVersion() }
+    override def groovycPluginMvnDeps: T[Seq[Dep]] =
+      Task { outer.groovycPluginMvnDeps() }
+      // TODO: make Xfriend-path an explicit setting
+    override def groovycOptions: T[Seq[String]] = Task {
+      // FIXME
+      outer.groovycOptions().filterNot(_.startsWith("-Xcommon-sources")) ++
+        Seq(s"-Xfriend-paths=${outer.compile().classes.path.toString()}")
```
lefou: Both tasks look obsolete.

lefou (commented):
I understand, that you copied another language module to start this implementation, but it would be helpful, if you could revise it and delete all tasks that aren't needed, like `groovycOptions`.

libs/groovylib/src/mill/groovylib/GroovyModule.scala
lostiniceland: This would result in two passes over all source files.

lostiniceland:
@lefou I've resolved most of the issues and fixed the Spock bug. 

I've also extended the use-cases and backed them by tests to support static compilation as well as a 3-stage compilation for setups where Java depends on Groovy and vice-versa.

In the next weeks I will probably find some more time to see about what configuration options are needed and how to support them.

libs/groovylib/src/mill/groovylib/GroovyModule.scala
lefou: Is this bad? It's cached, it's in parallel, and it's likely used separately any way. 

If you care, you can still implement it in a private task returning a tuple of two sets. Or you define the extentions is a shared `val`.

From a maintenance and customization perspective, defining the file extensions in two locations isn't good.

lefou:
@lostiniceland Can you resolve the merge conflicts?

lostiniceland:
@lefou I would like to pull `bomMvnDeps` up to `JavaModuleBase` in order to setup BOMs within the `TestModule.Spock` utility (I guess other testframeworks provide BOMs as well).
Would this break compatibility? If not, are there other reasons why it is not there?

lefou:
> @lefou I would like to pull `bomMvnDeps` up to `JavaModuleBase` in order to setup BOMs within the `TestModule.Spock` utility (I guess other testframeworks provide BOMs as well). Would this break compatibility? If not, are there other reasons why it is not there?

You can try, but I think it will break bin-compat.

lostiniceland:
> > @lefou I would like to pull `bomMvnDeps` up to `JavaModuleBase` in order to setup BOMs within the `TestModule.Spock` utility (I guess other testframeworks provide BOMs as well). Would this break compatibility? If not, are there other reasons why it is not there?
> 
> You can try, but I think it will break bin-compat.

I gave it a try and no additional test broke. From SemVer perspective it also shouldn't break anything as long as the trait provides a default implementation. So I will keep this for now.

lostiniceland:
@lefou Have a look at the latest [commit](https://github.com/com-lihaoyi/mill/pull/5747/commits/c82f9a2bdf8f934582200f54c34e8bfe93d800f7)

I managed to manage (pun intended) the JUnit5/Spock dependencies. Since the BOMs are only available at certain versions this will be checked.
For JUnit starting from 5.12 this has the nice effect that the junit-bom manages the platform-launcher version. So there is actually no need to specify this (but removing the property will break compatibility). 
So I've implemented it according to these rules:
1. platform-launcher version set => add launcher with specific version
2. no launcher version, but junit-bom available => add launcher without version (because managed)
3. else => don't add launcher
This should make the usage easier.

What do you think? Do you want to keep this, or should I remove it?

libs/groovylib/api/src/mill/groovylib/worker/api/GroovyWorker.scala
```diff
@@ -0,0 +1,37 @@
...
+   * In a mixed setup this will compile the Groovy sources to Java stubs.
+   */
+  def compileGroovyStubs(
+      sourceFiles: Seq[os.Path],
+      classpath: Seq[os.Path],
+      outputDir: os.Path
+  )(implicit
+      ctx: TaskCtx
+  )
+      : Result[CompilationResult]
```
lefou: What is the output of this? `.class` files or `.java` files?

lostiniceland:
Looks like I was wrong about binary compatibility though I am not quite sure what to make of this:
https://github.com/lostiniceland/mill/actions/runs/17948459255/job/51041447311

Lets say Groovylib is released with Mill 1.0.10 then I cannot use this with Mill 1.0.9. Fine. And plugins targeting 1.0.10 will also still be compatible since the new method has a default implementation.
Currently I don't really trust this test.

lostiniceland:
@lefou I think this PR is ready to get out of the draft status.

I've added examples and docs (though `./mill docs.fastPages` didn't do anything), added most relevant compiler options, and I've been using it for Spock testing in a side project for some weeks now without problems.

There is this outstanding issue with the binary-compatibility caused by adding `bomMvnDeps` to the `TestModule` base (see previous message) which I think should actually work. Maybe @lihaoyi can provide insight. If this really breaks compat, then I will have to remove the feature that a BOM is automatically added for certain Test setups.

lihaoyi:
I haven't reviewed this in detail, but let's put this in `contrib/` rather than `libs/`. That will send the right signal that it is new and experimental, and if it becomes popular we can move it into `libs/` with the Java/Scala/Kotlin toolchains

lostiniceland:
> I haven't reviewed this in detail, but let's put this in `contrib/` rather than `libs/`. That will send the right signal that it is new and experimental, and if it becomes popular we can move it into `libs/` with the Java/Scala/Kotlin toolchains

Do I have to change the package from `mill.groovylib` to `mill.contrib.groovylib`? It would be the correct way, but it just creates more work and I am quite confident that the module is in a good shape. I'd rather spent my time on adding tests for the recently added compiler options (currently only tested manually)

lefou:
Since we also have experimental `pythonolib` and `javascriptlib` not in `contrib`, I think it's ok to have a `groovylib`, as long as it is marked as experimental.

We already have the example tooling in place for `<lang>lib` modules, which is quite different to `contrib` modules.

libs/groovylib/src/mill/groovylib/GroovyModule.scala
```diff
@@ -0,0 +1,267 @@
...
+   */
+  def groovyLanguageVersion: T[String] = Task { groovyVersion().split("[.]").take(2).mkString(".") }
+
+  private def isGroovyBomAvailable: T[Boolean] = Task {
+    if (groovyVersion().isBlank) {
+      false
+    } else {
+      Version.isAtLeast(groovyVersion(), "4.0.26")(using Version.IgnoreQualifierOrdering)
+    }
+  }
```
lefou: The name is misleading. It should be `useGroovyBom` or `groovyUseBom`, since it's not checking for availability but deciding (based on the version).

libs/groovylib/src/mill/groovylib/GroovyModule.scala
```diff
@@ -0,0 +1,267 @@
...
+    allGroovySourceFiles() ++ allJavaSourceFiles()
+  }
+
+  /**
+   * Specifiy the bytecode version for the Groovy compiler.
+   * {{{
+   *   def targetBytecode = Some("17")
+   * }}}
+   */
+  def targetBytecode: T[Option[String]] = None
```
lefou: Should be `groovyCompileTargetBytecode`

lihaoyi:
> Since we also have experimental `pythonolib` and `javascriptlib` not in `contrib`, I think it's ok to have a `groovylib`, as long as it is marked as experimental.
> 
> We already have the example tooling in place for `<lang>lib` modules, which is quite different to `contrib` modules.

Sounds good to me!

libs/groovylib/src/mill/groovylib/JavaMavenModuleWithGroovyTests.scala
```diff
@@ -0,0 +1,16 @@
+package mill.groovylib
+
+import mill.*
+import mill.javalib.{JavaModule, MavenModule}
+
+/**
+ * Convenience trait for projects using Java for production and Groovy for tests in a Maven setup
+ */
```
lefou, line 26: You should add more motivation, why this might be preferable over just adding `GroovyModule` (e.g. to avoid implicit groovy deps in the module).

libs/groovylib/api/src/mill/groovylib/worker/api/GroovyWorker.scala
lostiniceland: `.java` though they are cleaned up after the last groovy compilation stage.

lefou: I think using `CompilationResult` in the return isn't appropriate here.

libs/groovylib/package.mill
lefou: Fixed it myself

libs/groovylib/api/src/mill/groovylib/worker/api/GroovyWorker.scala
lefou: Is this stub-handling somewhere documented? Or can you outline the process and what you mean with "cleaned up after the last compilation stage"?

It looks to me, this groovy stubs should reside in it's own cacheable task, since it should not need to be regenerated when just Java files changed.

lostiniceland: I will document it.

In order for Java to compile with dependencies to Groovy, It only needs the method signatures (headers). This is what the Groovy Stubs are. When you have dependencies Groovy -> Java AND Java -> Groovy you need a 3 stage compile
1. Groovy compiles to Java stubs (*.java files in the output)
2. Java compiles with the Stubs available
2.1 Cleanup stubs after Java compile (I was wrong before, it happens here)
3. Compile Groovy (now all Java dependencies are also available on the compilepath).

It should actually be very similar in the other toolchains (scala, kotlin) but maybe the external compiler handles this already internally.

lefou: There is a difference to Scala or Kotlin handling. Both are able to read Java files, so they will always compile their files first and the Java-compiler comes last. Whereas Groovy will first generate stubs, so Java can consume them and succeed with compilation, and then compile Groovy (while consuming the compiled Java classes vie the classpath). 

Nevertheless, the only thing needed is a changed classpaths and output paths. There is no need to delete stub files. Especially, since they are needed in the next compile run and don't need any regeneration at all, when nothing has changed to the Groovy files. And even if the Groovy files have changed but the resulting stubs have not, there is no need no recompile Java files at all. Both cases would be automatically handled by Mill, if we properly model it via cached tasks.

lostiniceland: Good point about Mill caching. I was using GMavenPlus (Groovy Maven Plugin) as a reference, and maybe there are some things that we could do differently. ~~For instance, I could test if the Groovy compiler can read Java files so we might be able to ditch the stubs.~~

Edit: The Groovy compiler is actually able to receive Groovy and Java sources and compile __but__ it will not consider anything passed in `javacOptions`. Thats why I choose to split the stages.

lefou: Technically, we could model the three steps as independent cacheable tasks,  as in the following pseude-code:

```scala
def groovyCompileJavaStubs: Task[PathRef]
def groovyCompileJava: Task[PathRef] // depends on groovyCompileJavaStubs
def groovyComileGroovy: Task[PathRef] // depends on groovyCompileJava
```

Unfortunately, this produces more than one out directory with final classfiles with is not what (current) Mill expects, so we need to merge-copy the final classes-directories together.

```scala
def compile: Task[CompilationResult] {
  Seq(groovyCompileGroovy(), groovyCompileJava())
    .map(pr = os.copy(pr.path, Task.dest, mergeFolders = true))
  CompileResult(..., PathRef(Task.dest))
}
```

lostiniceland: @lefou Thanks for fixing the mima setup.

EDIT: I've rebased and updated the compilation to use a dedicated (cached) folder for the stubs.

lostiniceland: Sorry, I missed your post and tried another solution. What's your take on the latest commit?
I am not sure if we split the compile into 3 public tasks: since they are implicitly dependent on another, it can easily break if the user overwrites only one.

lefou: I haven't a look at your latest changes, yet. About my suggestions: You always could hide implementation details by making task private, without loosing all the benefits. Besides, I think there is only a minimal risk of breaking anything by overriding, since each task has exactly one purpose and clear in- and outputs. Also, overriding selected task is exactly the way to customize the integration further. There is also the benefit of inspect-ability from the CLI.

lefou:
Your latest `git rebase` looks slightly gone wrong. Some commit should not be part of this PR. Better avoid `git rebease` and use `git merge` for newer changes, esp. since we are in the middle of a review and multiple users have committed to the branch.

libs/groovylib/src/mill/groovylib/GroovyModule.scala
```diff
@@ -132,6 +132,29 @@ trait GroovyModule extends JavaModule with GroovyModuleApi { outer =>
...
+    val config = GroovyCompilerConfiguration(
+      enablePreview = groovyCompileEnablePreview(),
+      targetBytecode = groovyCompileTargetBytecode(),
+      disabledGlobalAstTransformations = disabledGlobalAstTransformations()
+    )
+
+    GroovyWorkerManager.groovyWorker().withValue(groovyCompilerClasspath()) {
+      _.compileGroovyStubs(groovySourceFiles, compileCp, stubDir, config)
+    }
+  }
```
lefou: It's not obvious what happens here. This task does not return a result, with is a good indicator it's not Mill idiomatic. It seems to mutate locations locations outside of `Task.dest`, so Mill will not be able to properly manage its cached state.

lostiniceland:
> Your latest `git rebase` looks slightly gone wrong. Some commit should not be part of this PR. Better avoid `git rebease` and use `git merge` for newer changes, esp. since we are in the middle of a review and multiple users have committed to the branch.

I did a rebase since your guidelines mention that the PR commits get squashed and afaik you cannot squash properly with merges.
Which commits are you referring to? In case its gone completely wrong I can setup a fresh branch...

lefou:
In a rebased branch, we should not see commits already in the base branch. But exactly this is what can be seen here, e.g. those commits authored by others and referring to other PRs in their title.

Looks more like you rebased some commits from the base branch on top of your branch.

I still have a copy of the original branch here: https://github.com/lefou/mill/tree/groovylib-tr

Best would be to rebase your PR to this state and re-apply you last two commits on top of it. Although we typically squash PRs as part of merging them to our `main` branch, we avoid squashing while still in review, to improve the review and tracing process. This also works if the PR contains multiple merges.

lostiniceland:
That`s weird. The commits from main which should not be shown come from a merge I did (not a rebase)
grafik

I've reset to before the merge and rebased, but still those commit show up.

EDIT: now it is correct. It took some time for Github to update the PR overview after the push.

lihaoyi:
For experimental modules we should use `MillPublishScalaModule` rather than `MillStableScalaModule`

lostiniceland:
What shall we do about the failing mima tests? It is caused by [this new property](https://github.com/com-lihaoyi/mill/pull/5747/files#diff-d76d7e87e461ac86153671bad9d1b6606226fd04daa35cb2d0892efe4d8ddfafR697).
The property should not fail compatibility since it provides a default implementation (it is a minor change, not major).
Am I missing something?

If there is no solution, I have to remove this feature (automatically adding boms).

Furthermore, Android and Kotlin tests are failing due to unresolved "publish" imports
```
[9478] Caused by: utest.AssertionError: evalResult.isSuccess
[9478] evalResult: EvalResult = EvalResult(
[9478]   exitCode = 1,
[9478]   out = """[info] compiling 1 Scala source to /home/runner/work/mill/mill/out/integration/localOutFolder.dest/local-0/out/out/mill-build/compile.dest/classes ...
Error: [9478] [error] -- [E6] /home/runner/work/mill/mill/out/integration/localOutFolder.dest/local-0/out/build.mill:7:29
Error: [9478] [error] 7 ?import mill.*, kotlinlib.*, publish.*
Error: [9478] [error]   ?                            ^^^^^^^
Error: [9478] [error]   ?Not found: publish
Error: [9478] [error] one error found
```
Could that be caused by the exports groovylib is doing?

lefou:
@lihaoyi 

> For experimental modules we should use `MillPublishScalaModule` rather than `MillStableScalaModule`

Actually, I did this change but ensured all classes are annotated with `@experimental`. The motivation was to have the API docs generated, which are needed to come up with proper documentation. When we revert this, we should manually add the groovylib module to the unidoc modules, otherwise docs will be incomplete.

lihaoyi:
Maybe we should make the scaladoc generation pick up all `MillPublishScalaModule`s instead of just `MillStableScalaModule`s? Since 1.0.0 it should already only contain modules that are on the `mill-libs` classpath, so we no longer need to rely on the module type as a filter w.r.t. what users should be using or not using

lefou:
> Maybe we should make the scaladoc generation pick up all `MillPublishScalaModule`s instead of just `MillStableScalaModule`s?

Sounds good. But it doesn't work out of the box. See #5970

lostiniceland:
I assume we will be waiting for a Coursier release to resolve the docs being published for non-stable modules.

In the meantime, I tried to replicate the failing tests locally. How do you guys work with Github actions for failing tests? I tried using [act](https://github.com/nektos/act) but I get different test results. I saw that github actions also use Mill for setup but I couldnt find an easy way to replicate the steps failing in this PR.

lefou:
> I assume we will be waiting for a Coursier release to resolve the docs being published for non-stable modules.

I think #5970 is not a blocker. Meanwhile, we merged #5977, so the docs should be linkable after you update this PR.

> In the meantime, I tried to replicate the failing tests locally. How do you guys work with Github actions for failing tests? I tried using [act](https://github.com/nektos/act) but I get different test results. I saw that github actions also use Mill for setup but I couldnt find an easy way to replicate the steps failing in this PR.

I just run the failed Mill tasks locally. Sometimes, I need to switch the Java version via `.mill-jvm-version` to reproduce the issue.

lihaoyi:
Converting this to draft so as not to block CI on main

lostiniceland:
@lihaoyi @lefou I could use your help with the remaining tests. 

I can at least reproduce the issues locally
* `./mill -i -k  selective.run example.javalib.__.local.daemon` 
* `./mill -i -k  selective.run example.kotlinlib.__.local.daemon` 

But unfortunately I haven't found a way to run some tests in isolation from inside Intellij due to missing environments like `MILL_INTEGRATION_DAEMON_MODE` or `MILL_TEST_RESOURCE_DIR`.

Anyways, I see failures in the example tests for javalib/kotlinlib like `ArgumentParserException: argument -t/--text is required` or
```
test.resolvedRunMvnDeps java.lang.RuntimeException: 
Resolution failed for 2 modules:
io.micronaut.test:micronaut-test-junit5: 
not found: /home/marc/Development/projects/mill/out/contrib/artifactory/publishLocalTestRepo.dest/io/micronaut/test/micronaut-test-junit5/maven-metadata.xml
not found: /home/marc/Development/projects/mill/out/contrib/buildinfo/publishLocalTestRepo.dest/io/micronaut/test/micronaut-test-junit5/maven-metadata.xml
not found: /home/marc/Development/projects/mill/out/contrib/codeartifact/publishLocalTestRepo.dest/io/micronaut/test/micronaut-test-junit5/maven-metadata.xml
```
and
```
==== Missing expectedLine:
my.custom.property: my-prop-value
```
which I cannot relate to my changes. 

In the CI I also see `response code: 500 for URL: https://repo1.maven.org/maven2/org/fusesource/jansi/jansi/2.4.1/jansi-2.4.1.jar) while downloading https://repo1.maven.org/maven2/org/fusesource/jansi/jansi/2.4.1/jansi-2.4.1.jar`

To me it looks like main is broken, but I guess I am wrong about that.
[This is my latest run on the fork CI](https://github.com/lostiniceland/mill/actions/runs/19565789812)

lefou:
5xx errors for external resources in CI are most likely just flakiness and you should re-run these tests.

lostiniceland:
Marked as ready for review.

Can we talk about merging this? The test failures I get locally are the same when checking out main. 
Furthermore main seems to be broken anyways at the moment. See https://github.com/com-lihaoyi/mill/commits/main/ (all marked as failures)

example/groovylib/basic/1-simple/build.mill
```diff
@@ -0,0 +1,59 @@
+//// SNIPPET:BUILD
+
+package build
+import mill.*, groovylib.*
+
+object foo extends GroovyModule {
```
lefou:
```suggestion
@mill.api.experimental
object foo extends GroovyModule {
```

core/api/daemon/src/mill/api/daemon/internal/GroovyModuleApi.scala
```diff
@@ -0,0 +1,3 @@
+package mill.api.daemon.internal
+
+trait GroovyModuleApi extends JavaModuleApi
```
lefou, line 4:
```suggestion
@mill.api.experimental
trait GroovyModuleApi extends JavaModuleApi
```

libs/groovylib/src/mill/groovylib/GroovyModule.scala
```diff
@@ -0,0 +1,310 @@
...
+  trait GroovyTests extends JavaTests with GroovyModule {
+
+    override def groovyLanguageVersion: T[String] = outer.groovyLanguageVersion()
+    override def groovyVersion: T[String] = Task { outer.groovyVersion() }
+    override def bomMvnDeps: T[Seq[Dep]] = outer.bomMvnDeps()
+    override def mandatoryMvnDeps: Task.Simple[Seq[Dep]] = outer.mandatoryMvnDeps
+  }
+}
+
+object GroovyModule {
```
lefou, line 307:
```suggestion
@mill.api.experimental
object GroovyModule {
```

libs/groovylib/src/mill/groovylib/GroovyWorkerManager.scala
```diff
@@ -0,0 +1,42 @@
...
+import mill.util.ClassLoaderCachedFactory
+
+@mill.api.experimental
+class GroovyWorkerManager()(implicit ctx: TaskCtx)
+    extends ClassLoaderCachedFactory[GroovyWorker](ctx.jobs) {
+
+  def getValue(cl: ClassLoader) = GroovyWorkerManager.get(cl)
+}
+
+object GroovyWorkerManager extends ExternalModule {
```
lefou, line 16:
```suggestion
@mill.api.experimental
object GroovyWorkerManager extends ExternalModule {
```

libs/groovylib/worker/src/mill/groovylib/worker/impl/GroovyWorkerImpl.scala
```diff
@@ -0,0 +1,95 @@
...
+import mill.javalib.api.CompilationResult
+import mill.groovylib.worker.api.{GroovyCompilerConfiguration, GroovyWorker}
+import org.codehaus.groovy.control.{CompilationUnit, CompilerConfiguration, Phases}
+import org.codehaus.groovy.tools.javac.JavaStubCompilationUnit
+import os.Path
+
+import scala.jdk.CollectionConverters.*
+import scala.util.Try
+
+class GroovyWorkerImpl extends GroovyWorker {
```
lefou, line 15:
```suggestion
@mill.api.experimental
class GroovyWorkerImpl extends GroovyWorker {
```

libs/script/src/GroovyModule.scala
```diff
@@ -0,0 +1,34 @@
+package mill.script
+
+import mill.*
+import mill.api.ExternalModule
+import mill.api.Discover
+import mill.api.ScriptModule
+import mill.javalib.TestModule
+
+class GroovyModule(val scriptConfig: ScriptModule.Config) extends GroovyModule.Base {
```
lefou, line 10:
```suggestion
@mill.api.experimental
class GroovyModule(val scriptConfig: ScriptModule.Config) extends GroovyModule.Base {
```

libs/script/src/GroovyModule.scala
```diff
@@ -0,0 +1,34 @@
...
+  override lazy val millDiscover = Discover[this.type]
+}
+
+object GroovyModule {
```
lefou, line 15:
```suggestion
@mill.api.experimental
object GroovyModule {
```

lostiniceland:
> I guess we can merge this, to get early feedback. It's marked as experimental, so we can refine it incrementally.
> 
> Not so sure about the `mill.script.GroovyModule`. I'm not yet so familiar with the scripting part of Mill, so maybe @lihaoyi can have a final review of this class?

I've just added this scripting support after it showed up recently to comply with the other language-libs.

lihaoyi:
I think it looks reasonable to me, once CI is green-ish we can probably merge it as experimental

website/docs/modules/ROOT/nav.adoc
```diff
@@ -62,6 +62,8 @@
 *** xref:javascriptlib/linting.adoc[]
 *** xref:javascriptlib/publishing.adoc[]
 *** xref:javascriptlib/build-examples.adoc[]
+** xref:groovylib/intro.adoc[]
+*** xref:groovylib/testing.adoc[]
```
lihaoyi, line 66: Let's merge this into `intro.adoc`, these pages aren't long enough to be worth splitting up

lostiniceland:
> I think it looks reasonable to me, once CI is green-ish we can probably merge it as experimental

The problem is that I cannot get the CI greenish for 2 weeks now because main is broken and the things which fail are in other libs for no good reason.

core/api/daemon/src/mill/api/daemon/internal/GroovyModuleApi.scala
```diff
@@ -0,0 +1,4 @@
+package mill.api.daemon.internal
+
+@mill.api.experimental
+trait GroovyModuleApi extends JavaModuleApi
```
lefou:
```suggestion
@mill.api.daemon.experimental
trait GroovyModuleApi extends JavaModuleApi
```

lefou:
> > I think it looks reasonable to me, once CI is green-ish we can probably merge it as experimental
> 
> The problem is that I cannot get the CI greenish for 2 weeks now because main is broken and the things which fail are in other libs for no good reason.

Don't worry. Our CI suite is large and we have flaky tests from time to time. But we typically recognize if a test is a repeated unrelated flaky test or something new. Unrelated test failures don't block a feature merge.

lostiniceland:
@lefou the merge from the [scoverage update](https://github.com/com-lihaoyi/mill/pull/6338) breaks the build locally. I tried to reproduce the native-compiler errors but a clean compile fails with the following because 2.5.0 is not yet on central (at least not in the 2.13.15 folder)

```
24190] [error] dummy[org.scoverage_scalac-scoverage-plugin_3.8.0-RC3_2.5.0].resolvedMvnDeps
java.lang.RuntimeException: 
  org.scoverage:scalac-scoverage-plugin_2.13.15:2.5.0 
        Not an internal Mill module: org.scoverage:scalac-scoverage-plugin_2.13.15:2.5.0
        not found: /home/marc/.ivy2/local/org.scoverage/scalac-scoverage-plugin_2.13.15/2.5.0/ivys/ivy.xml
        not found: https://repo1.maven.org/maven2/org/scoverage/scalac-scoverage-plugin_2.13.15/2.5.0/scalac-scoverage-plugin_2.13.15-2.5.0.pom
```

lefou:
> @lefou the merge from the [scoverage update](https://github.com/com-lihaoyi/mill/pull/6338) breaks the build locally. I tried to reproduce the native-compiler errors but a clean compile fails with the following because 2.5.0 is not yet on central (at least not in the 2.13.15 folder)
> 
> ```
> 24190] [error] dummy[org.scoverage_scalac-scoverage-plugin_3.8.0-RC3_2.5.0].resolvedMvnDeps
> java.lang.RuntimeException: 
> Resolution failed for 1 modules:
> --------------------------------------------
>   org.scoverage:scalac-scoverage-plugin_2.13.15:2.5.0 
>         Not an internal Mill module: org.scoverage:scalac-scoverage-plugin_2.13.15:2.5.0
>         not found: /home/marc/.ivy2/local/org.scoverage/scalac-scoverage-plugin_2.13.15/2.5.0/ivys/ivy.xml
>         not found: https://repo1.maven.org/maven2/org/scoverage/scalac-scoverage-plugin_2.13.15/2.5.0/scalac-scoverage-plugin_2.13.15-2.5.0.pom
> ```

I must have overlooked, that they dropped support for Scala 2.13.15 in 2.4.0. We either need to update our moduledefs plugin to support newer Scala 2.13 versions or rollback the update for now.

lefou:
* https://github.com/com-lihaoyi/mill-moduledefs/pull/27
* https://github.com/com-lihaoyi/mill-moduledefs/pull/28

lostiniceland:
Not sure whats going on with the CI. After updating the Docs to latest changes the command `./mill -i website.fastPages + website.checkBrokenLinks` runs just fine locally

```
...
6514] You can browse the pages at: file:///home/xxx/dev/projects/mill/out/website/localPages.dest/site/index.html
6517/6517] ============================== website.fastPages + website.checkBrokenLinks ============================== 23s
```
But the CI fails with some broken links

lefou:
I think the tests break, since the newly introduced `bomMvnDeps` override in `TestModule.Junit5` does not call `super.bomMvnDeps` (which is due to bin-compat). This is effectively erasing the locally defined BOM dependencies for Micronaut. 

We can work around that in the tests, but other users may hit the same issue, so we better find a better solution.

IMHO, the break-surface in external binary-published APIs is low. I'm not aware of a single plugin extending the test module chain involving `TestModule.Junit5`. Hence, we probably just want accept this breakage.

@lihaoyi WDYT?

NOTE: There is an open discussion with some ideas, how to improve the bin-compat story for Mill 2.x: https://github.com/com-lihaoyi/mill/discussions/5873

lefou:
Here are the options:

1. Fix by applying a workaround in the failing tests:

```diff
 object test extends MavenTests, TestModule.Junit5 {
+  def bomMvnDeps = super[MavenTests].bomMvnDeps() ++ super[Junit5].bomMvnDeps()
```

2. Properly fix by calling `super` in `TestModule.Junit5`, but breaking bin-compat:

```diff
   override def bomMvnDeps: T[Seq[Dep]] = Task {
-    // cannot call super.bomMvnDeps because it will break mima compat
-    Seq(jupiterVersion())
+    super.bomMvnDeps() ++ Seq(jupiterVersion())
```

lihaoyi:
@lefou I think I've seen @alexarchambault calling `super[JavaModule].task` that seems to preserve bincompat, could you try that? Another option is to introduce a `private[mill] def bomMvnDeps0` in the super class that contains the actual implementation, and both superclass and subclass `def bomMvnDeps` can call it without using `super`

lefou:
I'd say, both suggestions have issues, so I'd like to avoid them at any cost.

> @lefou I think I've seen @alexarchambault calling `super[JavaModule].task` that seems to preserve bincompat, could you try that? 

Even if that preserves bin-compat, it does not fix the issue, since it's hard-wiring the call to `JavaModule.bomMvnDeps`, but not the effective `super`-trait. E.g. we known `JavaModule.bomMvnDeps` just doesn't add anything.

> Another option is to introduce a `private[mill] def bomMvnDeps0` in the super class that contains the actual implementation, and both superclass and subclass `def bomMvnDeps` can call it without using `super`

The way this was done in Mill codebase by @alexarchambault (let `bomMvnDeps` refer to `bomMvnDeps0`) would mean, I need to move the actual source of settings to the `bomMvnDeps0`, which also means, we de-facto make `bomMvnDeps` deprecated or even obsolete. While it still can be used to access the value of `bomMvnDeps0`, `bomMvnDeps` can't be used for overriding. Or do I overlook somethink?


In the end, we deal with a limitation in Scala's `trait` encoding, which we want to use in tightly constraint bin-compat setting. The only (unproven) way I currently see how we could improve this, is by preparing the to-be bin-compat code before release (means, we can only do that for Mill 2.0), e.g. by generating some placeholders-`def`s to all super-tasks of a trait in the bytecode.

lefou:
> Not sure whats going on with the CI. After updating the Docs to latest changes the command `./mill -i website.fastPages + website.checkBrokenLinks` runs just fine locally
> 
> ```
> ...
> 6514] You can browse the pages at: file:///home/xxx/dev/projects/mill/out/website/localPages.dest/site/index.html
> 6517/6517] ============================== website.fastPages + website.checkBrokenLinks ============================== 23s
> ```
> 
> But the CI fails with some broken links

I think some links broke after the latest larger documentation restructuring. I try to fix them...

lefou:
@lihaoyi Any idea, how to deal with this error message?

```
[error] build.mill.yaml:7:30
8714]   groovyCompileTargetBytecode: '11'
8714]                                ^
8714]   groovyCompileTargetBytecode Failed de-serializing config override: expected unit got string
```

The task is defined as `def groovyCompileTargetBytecode: T[Option[String]]`.

lostiniceland:
@lefou I also worked on the website restructuring yesterday (merging right now). I will run more tests over the weekend as well. And Spock 2.4 for Groovy 5 was released yesterday, so I would like to add a cross-test to the current test-suite.
I am also going over every example from the docs again to make sure all is correct.

lefou:
I think we can commit to merge this before we tag `1.1.0`. The only issue I'm currently aware of is the fact that the declarative example doesn't work when Mill runs on Java 17, but we can't yet select the `groovyCompileTargetBytecode`, which we can fix independently, given this is still experimental and it works on Java 21.

lihaoyi:
> @lihaoyi Any idea, how to deal with this error message?
> 
> ```
> [error] build.mill.yaml:7:30
> 8714]   groovyCompileTargetBytecode: '11'
> 8714]                                ^
> 8714]   groovyCompileTargetBytecode Failed de-serializing config override: expected unit got string
> ```
> 
> The task is defined as `def groovyCompileTargetBytecode: T[Option[String]]`.

I think I've seen this before as a type inference issue for `None`. Can you try replacing it with `Option.empty[T]` at the definition of this method in the code

lefou:
> > @lihaoyi Any idea, how to deal with this error message?
> > ```
> > [error] build.mill.yaml:7:30
> > 8714]   groovyCompileTargetBytecode: '11'
> > 8714]                                ^
> > 8714]   groovyCompileTargetBytecode Failed de-serializing config override: expected unit got string
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The task is defined as `def groovyCompileTargetBytecode: T[Option[String]]`.
> 
> I think I've seen this before as a type inference issue for `None`. Can you try replacing it with `Option.empty[T]` at the definition of this method in the code

Changed it.

```diff
-  def groovyCompileTargetBytecode: T[Option[String]] = None
+  def groovyCompileTargetBytecode: T[Option[String]] = Option.empty
```

Now the error message is

```
8714]   [error] build.mill.yaml:7:30
8714]   groovyCompileTargetBytecode: '11'
8714]                                ^
8714]   groovyCompileTargetBytecode Failed de-serializing config override: java.lang.Enum.<init>(java.lang.String)

```

lihaoyi:
How about `Option.empty[String]`?

lefou:
> How about `Option.empty[String]`?

That's not supported. But I tried something that works.

```diff
-  def groovyCompileTargetBytecode: T[Option[String]] = None
+  def groovyCompileTargetBytecode: T[Option[String]] = Task.apply[Option[String]] { Option.empty }
```

lefou:
Correction: Compiler hickup, `Option.empty[String]` also works.

lefou:
@lostiniceland I think I fixed everything except the test you're still working on. I leave it to you to finish your documentation changes. Please ping me, if you're done.

lostiniceland:
Test for Spock under Groovy 5 has been updated and documentation verified so far. 

I will have a look tomorrow on the last failing CI Test.

lostiniceland:
@lefou finally a green build. Thanks for the support.

PR #5751, opened 2025-08-26
Implement support for Scala 3.8

lefou:
Instead of using a scala-library-2.13.x, since Scala 3.8.0 the scala-library is now versions equally to the Scala version.


libs/javalib/api/src/mill/javalib/api/JvmWorkerUtil.scala
```diff
@@ -146,4 +150,15 @@ object JvmWorkerUtil {
...
+  /**
+   * Checks whether the version of the scala-library should be `2.13.x` or match the given `scalaVersion`
+   * @param scalaVersion The Scala version
+   * @return `true` if the scala-library version should be a `2.13.`
+   */
+  def needsScala213Library(scalaVersion: String): Boolean = {
+    val sv = minorMajorVersion(scalaVersion)
+    // Some Dotty versions and all Scala 3 versions before 3.8
+    sv.major == 0 || (sv.major == 3 && sv.minor < 8)
+  }
```
He-Pin, line 170: What about  2.12？seems 2.12 will now with a 2.13 std

lefou: No, that's not what the code does. But maybe I can improve the method name? 

`enforceScala213Library` maybe?

libs/javalib/worker/src/mill/javalib/zinc/ZincWorker.scala
```diff
@@ -106,9 +106,10 @@ class ZincWorker(
           loaderLibraryOnly = ClasspathUtil.rootLoader,
           libraryJars = Array(libraryJarNameGrep(
             compilerClasspath,
-            // we don't support too outdated dotty versions
-            // and because there will be no scala 2.14, so hardcode "2.13." here is acceptable
-            if (JvmWorkerUtil.isDottyOrScala3(key.scalaVersion)) "2.13." else key.scalaVersion
+            // if Dotty or Scala 3.0 - 3.7, use the 2.13 version of the standard library
+            if (JvmWorkerUtil.enforceScala213Library(key.scalaVersion)) "2.13."
```
He-Pin, line 110: So Scala 2.12 will not enter this branch right?

lefou: Exactly.

WojciechMazur:
Thank you for the fix
@lefou @lihaoyi  I wonder would it be possible to backport the fix to 0.12.16? 

I'm mostly interested because of Scala 3 Open Community Build which currently is broken for most of the build tools when using latest nightlies. In the OCB we use support building with multiple versions of mill: 0.9, 0.10, 0.11, 0.12 and 1.x.  We cannot force upgrades to mill 1.1.0 as it would break the build definitions itself. 
I've come up with some workaround that works for versions 0.9-0.11 https://github.com/VirtusLab/community-build3/pull/547 but I'm unable to modify the zincWorker starting with 0.12 as it would result in missing dependencies to sbt interface bridege (e.g. `xsbit.compile.Compiler`). Seems like the only feasible way would be to upgrade to at least 0.12.16 and 1.1 containing the fixes for Scala 3.8 
Edit: Apparently the fix for 0.12 seems to work now but has not worked for me locally previously for some reason), it should be also appliable to mill 1.0, but we'd prefer to simply force minimal supported patch/minor versions in 0.12 and 1.1

I also wonder when can we expect the release 1.1.0?

lefou:
> Thank you for the fix @lefou @lihaoyi I wonder would it be possible to backport the fix to 0.12.16?

The patch looks as if it should work also for Mill 0.12.x. I can't promise anything. A contributed PR could speed up the backport significantly. ;-)

> I also wonder when can we expect the release 1.1.0?

There is no exact timeline. From looking on the open PRs, I think only a small number should be merged before we cut a `1.1.0`.

lihaoyi:
Probably can expect 1.1.0 some time mid-september I think

bishabosha:
The scala3-sbt-bridge cant be resolved by adding the repo, (it doesnt try to look there) is there another setting to change?

lefou:
This is an already merged and closed PR. Can you please open new issues or discussions?

bishabosha:
@lefou I have now opened https://github.com/com-lihaoyi/mill/issues/5802

PR #5766, opened 2025-08-28
Rework Android SDK installation

lefou:
This refactors the task dependency flow of `AndroidSdkModule`. Previously, most tasks provided `PathRef`s to tools depending itself on some anonymous installation tasks which needed to synchronize their work.

Now, the installation steps are bound to single tasks, which return a handle to the installed resources. These tasks are:

* `cmdlineTools`
* `androidSdk`
* `androidNdk`

Existing tasks pointing to various resources and executables, now just depend on these installation tasks and simply forward the returned resources.

Additional on Windows, tools will be searched for with the extensions `exe` and `bat`.

This is not the final design, but a step forward in making the setup more idiomatic and readable.

### DIFF
--- a/libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala
+++ b/libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala
@@ -388,7 +388,6 @@ trait AndroidSdkModule extends Module {
         )
       }
 
-      PathRef(basePath)
       CmdlineToolsComponents(
         basePath = basePath,
         avdmanagerExe = toolPathRef(basePath / "bin/avdmanager"),


lefou:
@vaslabs, @souvlakias, this is ready for a first review round.

souvlakias:
Hi,

[Here’s ](https://github.com/com-lihaoyi/mill/compare/main...vaslabs-ltd:mill:androidsdkmodule-osname-bugfix)a draft of the workarounds I tried so far for the issues I found:

- Java version must be at least 17
- Handling of `.bat` and `.exe` extensions
- A bug in `cmdlineToolsURL` 
- A bug in `acceptLicenses` (piping with just `echo y` did not work on windows)

We could keep the last two points if you want for now.
My extensions discovery workaround is a bit similar to yours. 
For the java version, I'm sure the `require` solution isn't what we want, just listing it for reference.

lefou:
Merging, since all tests look good.

PR #5770, opened 2025-08-28
Android: Fix androidtodo and replace compose samples in ci with it

vaslabs:
This was broken accidentally since [the ksp 2 work](https://github.com/com-lihaoyi/mill/pull/5727) , I didn't realise it wasn't running in the CI.

I'd prefer to have androidtodo over jetnews, as jetnews doesn't use Hilt, and androidtodo uses both compose and hilt (and KSP), so if I may choose one I prefer androidtodo.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -156,7 +156,7 @@ jobs:
             setup-android: true
 
           - java-version: 17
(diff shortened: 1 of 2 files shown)


vaslabs:
cc: @lefou @lihaoyi

lefou (approved):

Issue #5762, opened 2025-08-27
Fix android support for windows


vaslabs:
Several android tools are not accessible via the AndroidSdkModule with the current implementation. Souvlakias is working to make the class compatible with windows. I'm opening this issue so we can track it separately

vaslabs:
This can be an opportunity to also re-write AndroidSdkModule, I've realised we haven't really touched its original design , see also @lefou's commend https://github.com/com-lihaoyi/mill/pull/5701#issuecomment-3187767441

lefou:
Do you have some details which tools are affected? Some concrete error message maybe? 

Also, I pushed PR #5766, which does a first pass at refactoring `AndroidSdkModule`.

vaslabs:
just that most of the executables on windows end in either `.bat` or `.exe`, @souvlakias already's working on fixing them and testing on a windows machine

vaslabs:
from what I see, you can go ahead with your refactoring and @souvlakias can rebase his fixes

souvlakias:
Another issue is large os.call commands (E.g. this aapt2 call has many -R arguments)
```
show 1 tasks failed
app.androidLinkedResources java.io.IOException: Cannot run program "D:\Android\Sdk\build-tools\36.0.0\aapt2.exe" (in directory "D:\Projects\Pokedex_Compose_XR\out\app\androidLinkedResources.dest"): CreateProcess error=206, The filename or extension is too long
    java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1143)
    java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1073)
    os.proc.proc$lzyINIT1$1(ProcessOps.scala:345)
    os.proc.os$proc$$_$proc$1(ProcessOps.scala:351)
    os.proc.spawn(ProcessOps.scala:355)
    os.proc.call(ProcessOps.scala:224)
    os.call$.apply(ProcessOps.scala:42)
    mill.androidlib.AndroidModule.androidLinkedResources$$anonfun$1$$anonfun$1(AndroidModule.scala:629)
    mill.api.Task$Named.evaluate(Task.scala:338)
    mill.api.Task$Named.evaluate$(Task.scala:323)
    mill.api.Task$Computed.evaluate(Task.scala:349)
java.io.IOException: CreateProcess error=206, The filename or extension is too long
    java.base/java.lang.ProcessImpl.create(Native Method)
[...]

PR #5775, opened 2025-08-29
Android: Bugfixes for Windows Support

souvlakias:
### Motivation
Windows support was largely absent due to several bugs and issues. This PR extends some fixes of https://github.com/com-lihaoyi/mill/pull/5766 and resolves additional of issues, enabling the basics of Android Development with Mill work on Windows.
### Changes / Bugfixes
- Handling of `.bat` and `.exe` extensions
- A bug in `cmdlineToolsURL` 
- A bug in `acceptLicenses` (piping with `echo y` alone did not work on windows)
- Skip `posix permission fix` of  `androidLibsClassesJarMetaInf` on Windows
- Pass file arguments with `@files.txt` on CLI calls. ([Ref](https://github.com/com-lihaoyi/mill/issues/5762#issuecomment-3232786720))

### Example Tests Coverage
⚠️✅  : Works but some tests don't pass due to expecting unix paths or commands.
- AndroidLib
	- Java 
	1,2,5,6 ✅ 
	3,4 ⚠️✅
	- Kotlin
	1,2 ✅
	3 ❌ Screenshot png does not get generated
	4 ⚠️✅
- Thirdparty
	- Androidtodo ⚠️✅
	- Android Compose✅ 
	- Android Endless Tunnel ✅ 

#### Android Todo PoW

https://github.com/user-attachments/assets/6bac61ca-35eb-48ad-813b-4e705d2dcca9

	
### Notes
The `@files.txt` solution is only applied in a few tasks for now. Will do some more testing and see if its applicable to other tasks that pass many files to a CLI call as well.


vaslabs:
I see java version 17 here too https://developer.android.com/build/jdks , isn't what we use on our Linux machines too? I've used 11 and 21 I think accidentally a few times, but I'm not sure what happens under the hood, some windows setup might not actually switching to 17? would worth some testing in both platforms to figure out what's really going on

vaslabs:
not sure about require, could be considered too strict and maybe someone would configure a really old android app or they want to do some niche experimentation, maybe a jvmEnvCheck task before running sdkmanager that can be overridden ?

souvlakias:
> not sure about require, could be considered too strict and maybe someone would configure a really old android app or they want to do some niche experimentation, maybe a jvmEnvCheck task before running sdkmanager that can be overridden ?

Yes,  a task would be better, but I'm not sure where it should sit in the hierarchy. Since it's a requirement of `sdkmanager`, the `sdkmodule` is practically unusable when Java < 17, so the check should happen early.  
The first `sdkmanager` call on a clean setup could [here](https://github.com/com-lihaoyi/mill/blob/main/libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala#L352) or [here](https://github.com/com-lihaoyi/mill/blob/main/libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala#L262), but they are not part of a task, which makes things more complicated.  
@lefou, where would you suggest adding this check?

Reference of `sdkmanager`
image

lefou:
> @lefou, where would you suggest adding this check?

Isn't `sdkmanager` just used to bootstrap the environment. In general, such bootstrapping task should be considered helpers and are not essentially to build for Android, hence it should be overridable by users. I'm pretty sure older Android version don't require JDK17 and at least from my perspective, Mill should also support these devs.

If we implement a check, we also need to consider the versions in play. Best, we print a warning with helpful instructions before we use `sdkmamnager`. If `sdkmanager` fails, the user will also recognize our warning and hopefully has some guidance how to fix.

After all, IIUC, our bootstrapping only get active if the required tools aren't already present. So manually installation should be a way to work around.

core/constants/src/mill/constants/Util.java
```diff
@@ -22,8 +22,11 @@ private static String lowerCaseOsName() {
 
   public static boolean isWindows = lowerCaseOsName().startsWith("windows");
   public static boolean isLinux = lowerCaseOsName().equals("linux");
-  public static boolean isJava9OrAbove =
-      !System.getProperty("java.specification.version").startsWith("1.");
+  public static boolean isMac = lowerCaseOsName().contains("mac");
+  public static String javaSpecificationVersion = System.getProperty("java.specification.version");
+  public static boolean isJava9OrAbove = !javaSpecificationVersion.startsWith("1.");
+  public static boolean isJava17OrAbove =
+      isJava9OrAbove && Integer.parseInt(javaSpecificationVersion) >= 17;
```
lefou: All this looks unnecessary to me. We can use the various methods of [`scala.util.Properties.*`](https://scala-lang.org/api/3.7.1/scala/util/Properties$.html#), e.g. `isMac`, `isJavaAtLeast`.

libs/androidlib/src/mill/androidlib/AndroidAppModule.scala
```diff
@@ -147,22 +148,24 @@ trait AndroidAppModule extends AndroidModule { outer =>
...
-        // Fix permissions of unzipped directories
-        // `os.walk.stream` doesn't work
-        def walkStream(p: os.Path): geny.Generator[os.Path] = {
-          if (!os.isDir(p)) geny.Generator()
-          else {
-            val streamed = os.list.stream(p)
-            streamed ++ streamed.flatMap(walkStream)
+        // Fix permissions of unzipped directories (skip on Windows)
+        if (!isWindows) {
+          // `os.walk.stream` doesn't work
```
lefou, line 153: Can you elaborate why it doesn't work?

libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala
```diff
@@ -583,6 +592,7 @@ private object AndroidNdkLock
 private object AndroidCmdlineToolsLock
 
 object AndroidSdkModule {
+  require(isJava17OrAbove, "Android SDK requires Java 17 or above to run.")
```
lefou: This is too restrictive. This should be at least on task-level, where a user can customize if needed.

libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala
```diff
@@ -622,10 +632,10 @@ object AndroidSdkModule {
     if (os.exists(path)) {
       PathRef(path).withRevalidateOnce
     } else boundary {
-      if (scala.util.Properties.isWin && path.lastOpt.isEmpty && path.ext != "bat") {
+      if (isWindows && path.ext.isEmpty) {
```
lefou: Although I won't expect it, this would not catch files with a dot in it. Let's check for the absence of the extension we support (`bat` and `exe`) instead.

libs/androidlib/src/mill/androidlib/AndroidAppModule.scala
souvlakias: I am not sure what the issue with `walk.stream` is there, I did not author that. The change here was wrapping the permissions code section the `if (!isWindows)`

core/constants/src/mill/constants/Util.java
souvlakias: Yes, I first landed on `constants.Util` which had the `isWindows`  and did not see `scala.util.Properties`. I will use that, thanks.

libs/androidlib/src/mill/androidlib/AndroidSdkModule.scala
souvlakias: I'll remove this requirement completely since `sdkmanager` does throw a verbose error.

vaslabs: yeah, I think removing it entirely is the best approach, this knowledge belongs to sdkmanager, for instance if in a year it runs on jdk 24, we shouldn't have to make changes in mill

libs/androidlib/src/mill/androidlib/AndroidAppModule.scala
vaslabs: I think this is a leftover from some other time, things have changed a lot, let's remove these comments

souvlakias:
For reference, one issue I am facing is shortening the [cli-command](https://github.com/com-lihaoyi/mill/blob/main/libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala#L345) of `generatedSourcesWithKsp2Cli`.
I did not manage to use the solution used in other CLI calls of passing `@cp.txt` instead of the whole classpath string to parameters: [-libraries](https://github.com/com-lihaoyi/mill/blob/main/libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala#L326) and [\<processorClasspath\>](https://github.com/com-lihaoyi/mill/blob/main/libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala#L337C24-L337C42)

vaslabs:
> For reference, one issue I am facing is shortening the [cli-command](https://github.com/com-lihaoyi/mill/blob/main/libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala#L345) of `generatedSourcesWithKsp2Cli`. I did not manage to use the solution used in other CLI calls of passing `@cp.txt` instead of the whole classpath string to parameters: [-libraries](https://github.com/com-lihaoyi/mill/blob/main/libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala#L326) and [<processorClasspath>](https://github.com/com-lihaoyi/mill/blob/main/libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala#L337C24-L337C42)

I think this is good enough for now, this is another motivation to implement the in-process ksp2 as @lefou suggested

PR #5777, opened 2025-08-29
feat: support Playwright testing for Scala-JS

ThijsBroersen:
I could not find any existing JSEnv which allows me to properly test against some browser features and am using [scala-js-env-playwright](https://github.com/ThijsBroersen/scala-js-env-playwright) for a while now. The current version is still a snapshot so this PR is not ready to be merged, but I would love some feedback on what is required to get this feature as part of Mill. Currently I am using a custom build mill-assembly for my projects.

Here a draft of Playwright support for Scala-js testing. 

I was also wondering why isn't Mill supporting just a raw JSEnv? Any custom JSEnv implemention now first requires Mill to include it..

### DIFF
--- a/libs/scalajslib/src/mill/scalajslib/api/JsEnvConfig.scala
+++ b/libs/scalajslib/src/mill/scalajslib/api/JsEnvConfig.scala
@@ -11,6 +11,7 @@ object JsEnvConfig {
   implicit def rwExoegoJsDomNodeJs: RW[ExoegoJsDomNodeJs] = macroRW
   implicit def rwPhantom: RW[Phantom] = macroRW
   implicit def rwSelenium: RW[Selenium] = macroRW
+  implicit def rwPlaywright: RW[Playwright] = macroRW
   implicit def rw: RW[JsEnvConfig] = macroRW
 
   private given Root_JsEnvConfig: Mirrors.Root[JsEnvConfig] =
@@ -116,4 +117,15 @@ object JsEnvConfig {
         new SafariOptions()
     }
   }
+
+  final case class Playwright(
+      browserName: String = "chromium",
+      headless: Boolean = true,
+      showLogs: Boolean = false,
+      debug: Boolean = false,
+      //   pwConfig: Config = Config(),
+      runConfigEnv: Map[String, String] = Map.empty,
+      launchOptions: List[String] = Nil,
+      additionalLaunchOptions: List[String] = Nil
+  ) extends JsEnvConfig
 }
--- a/libs/scalajslib/api/src/mill/scalajslib/worker/api/JsEnvConfig.scala
+++ b/libs/scalajslib/api/src/mill/scalajslib/worker/api/JsEnvConfig.scala
@@ -38,4 +38,15 @@ private[scalajslib] object JsEnvConfig {
     case class FirefoxOptions(headless: Boolean) extends Capabilities
     case class SafariOptions() extends Capabilities
   }
+
+  final case class Playwright(
+      browserName: String = "chromium",
+      headless: Boolean = true,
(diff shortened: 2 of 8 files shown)


lefou:
> I was also wondering why isn't Mill supporting just a raw JSEnv? Any custom JSEnv implemention now first requires Mill to include it

I agree to it. It's nice to have out-of-the-box support for popular `JSEnv`s, but this should not be a closed set. Anyway, this should be reported/fixed in a seperate issue/PR.

libs/scalajslib/src/mill/scalajslib/api/JsEnvConfig.scala
```diff
@@ -116,4 +117,140 @@ object JsEnvConfig {
...
+        )
+      given rw: RW[FirefoxOptions] = macroRW
+
+    val defaultWebkitLaunchOptions = List(
+      "--disable-extensions",
+      "--disable-web-security",
+      "--allow-running-insecure-content",
+      "--disable-site-isolation-trials",
+      "--allow-file-access-from-files"
+    )
```
lefou, line 235: Where do these options come from? Can you link some documentation to ease future maintenence?

ThijsBroersen: These are the defaults we currently have in the plugin. Shall I just add some documentation referring to the plugin repo?

lefou: Yeah, that helps maintaining them.

ThijsBroersen: I cannot find any docs related to using the `jsEnvConfig` task. Where shall we add something?

ThijsBroersen: I added refs to the plugin repo source in code docs.

lolgab:
> > I was also wondering why isn't Mill supporting just a raw JSEnv? Any custom JSEnv implemention now first requires Mill to include it
> 
> I agree to it. It's nice to have out-of-the-box support for popular `JSEnv`s, but this should not be a closed set. Anyway, this should be reported/fixed in a separate issue/PR.

The rationale to not depend on the scalajs artifact was binary compatibility. If we used scalajs-jsenvs in mill it means that if scala-jsenvs gets a version 2.0 it's a problem in Mill since it would not be binary compatible with the one we shipped. in

ThijsBroersen:
> > > I was also wondering why isn't Mill supporting just a raw JSEnv? Any custom JSEnv implemention now first requires Mill to include it
> > 
> > 
> > I agree to it. It's nice to have out-of-the-box support for popular `JSEnv`s, but this should not be a closed set. Anyway, this should be reported/fixed in a separate issue/PR.
> 
> The rationale to not depend on the scalajs artifact was binary compatibility. If we used scalajs-jsenvs in mill it means that if scala-jsenvs gets a version 2.0 it's a problem in Mill since it would not be binary compatible with the one we shipped. in

I already assumed that. However, adding new test-environments is now only possible by modifying the build tool. Would there be a path possible where this could be pulled in via a plugin perhaps? (And perhaps would should open a discussion to answer this)

ThijsBroersen:
Could there be made a decision on what is going to happen with this PR? Rolling my own Mill because I need Playwright is quite annoying.
I started a discussion wrt to any JSEnv support: #5936

lihaoyi:
At a first glance this looks reasonable. @lefou @lolgab any more concerns? If not we can merge it

lefou (commented):
Looks good to me, too. But I'm not versed with JSEnv.

Issue #5759, opened 2025-08-27
Scalameta in task does not work


jnatten:
This is probably something I'm doing wrong or not understanding, but i couldn't get this to work.
Wanted to use scalameta to build a plugin to detect som cyclical dependencies internally in my application, but i couldn't even get it to run the [example](https://scalameta.org/docs/trees/guide.html#from-strings) on scalameta's page.

[Here is a selfcontained minimal repository](https://github.com/jnatten/mill-scala-meta-example/) for what im failing to do 😅

Quafadas:
~~Very likely related;~~

~~https://github.com/com-lihaoyi/mill/issues/5745~~

lefou:
> Very likely related;
> 
> [#5745](https://github.com/com-lihaoyi/mill/issues/5745)

I don't think so. While `show` on this particual task gives an error, downstream tasks still work. E.g. in this case, we get the following classpath:

```console
> mill --meta-level 1 show resolvedMvnDeps
[
  "qref:v1:40ece282:/home/lefou/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/sourcecode_3/0.4.3-M5/sourcecode_3-0.4.3-M5.jar",
  "qref:v1:ed58d398:/home/lefou/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/mill-libs_3/1.0.4/mill-libs_3-1.0.4.jar",
  "qref:v1:385c3bc4:/home/lefou/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scalameta/scalameta_3/4.13.9/scalameta_3-4.13.9.jar",
  "qref:v1:2610e8cb:/home/lefou/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala3-library_3/3.7.1/scala3-library_3-3.7.1.jar",
  "qref:v1:297efc78:/home/lefou/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/mill-moduledefs_3/0.11.10/mill-moduledefs_3-0.11.10.jar",
  "qref:v1:66928462:/home/lefou/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/unroll-annotation_3/0.2.0/unroll-annotation_3-0.2.0.jar",
[...]

PR #5790, opened 2025-09-01
Let `MillURLClassLoader` find classes not found in shared classloader

lefou:
This fixes an issue with our current `MillURLClassLoader` implementation regarding package prefixes. The issue is, that if a users package is a sub-package of one of the pre-configured shared package prefixes, classes will only be searched in the shared classloader, but not in the `MillURLClassLoader`. This leads to `ClassNotFoundError`s although the class is present on the classpath represented as URLs. In one specific case, Scalameta, which uses the `scala.meta` package (a sub-package of `scala` which we have in our pre-configured package prefix list), could not be used in a users project.

This is a trade-of between usability and correctness. Ideally, we would not match sub-packages but exact packages, but it's likely we miss some or run into compatibility and maintenance issues whenever a single part changes, e.g. Java version or Scala versions. It might even be not exactly configurable at all given the history of packages like `javax`.

Instead, maintaining a fixed positive list of exceptions is equally bad, since it needs to be maintained as well and will most likely result in bad user experiences, esp. in environments, where closed or non-open source libraries are used. 

The solution of this PR does not require any special maintenance, but has the potential to give misleading error messages, since we won't detect classloader misconfigurations as exactly as before. While I don't expect it being an issue often, it may still happen, especially, when developing or changing more fundamental parts of Mill like the classloader separation and isolation setup. (But it is very easy to disable this feature while refactor, so this isn't such a dramatic downside.)

Added an integration test under `integration/feature/mill-classloader`.

Fix https://github.com/com-lihaoyi/mill/issues/5759

### DIFF
--- a/integration/feature/mill-classloader/resources/build.mill
+++ b/integration/feature/mill-classloader/resources/build.mill
@@ -11,6 +11,3 @@ def graphDependencies: T[String] = Task {
   val tree = program.parse[Source].get
   tree.syntax
 }
-
-
-
--- a/integration/feature/mill-classloader/src/MillURLClassLoaderTests.scala
+++ b/integration/feature/mill-classloader/src/MillURLClassLoaderTests.scala
@@ -7,7 +7,7 @@ class MillURLClassLoaderTests extends UtestIntegrationTestSuite {
 
   val tests: Tests = Tests {
     test("test") - integrationTest { tester =>
-      val result1 = tester.eval(("show" , "graphDependencies"))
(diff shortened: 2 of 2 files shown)


lihaoyi (approved):
Let's add an integration test using `scala.meta` before merging just to verify that it works and stays working

lefou:
> Let's add an integration test using `scala.meta` before merging just to verify that it works and stays working

Done. Added `integration/feature/mill-classloader` which fails before applying this PR, but succeeds after.

Issue #5789, opened 2025-09-01
Make `MillURLClassLoader` parallel capable


lefou:
In Java this means this:

```java
  static {
        ClassLoader.registerAsParallelCapable();
  }
```

But since Scala has no static initialization, we can use the solution described here: https://sbuild.org/news/2014/12/17/SBuild-0.7.7-About-Thread-Deadlock-fixes-and-parallel-ClassLoaders-in-Scala.html

PR #5791, opened 2025-09-01
Register MillURLClassLoader as parallel capable classloader

lefou:
Fix https://github.com/com-lihaoyi/mill/issues/5789

The implementation is based on this commit: https://github.com/SBuild-org/sbuild/commit/4b8efad967b8eff422eb7179729b550e09159d6b

### DIFF
--- a/core/api/daemon/src/mill/api/daemon/MillURLClassLoader.scala
+++ b/core/api/daemon/src/mill/api/daemon/MillURLClassLoader.scala
@@ -12,10 +12,10 @@ import java.net.URLClassLoader
  *
  * In Java, parallel capable classloaders need to invoke a caller sensitive
  * protected static method. Since Scala has no `static` initialization syntax,
- * the only way to can a caller-sensitive method is from within the class (instance).
+ * the only way to invoke a caller-sensitive method is from within the class (instance).
  * Since this instance isn't yet seen as parallel capable from the JVM, we need to
  * drop it and create a new instance. All that logic is nicely hidden from the user
- * by making this constructor private. Instead, the companion objects
+ * by making this constructor private. Instead, the companion object's
  * [[MillURLClassLoader.apply()]] method is used to create an instance, which
  * handles all the initialization logic.
  *
@@ -58,14 +58,14 @@ class MillURLClassLoader private (
   /**
    * This constructor is deprecated and should not be used, since it can throw a
    * [[DoNotUseThisInstanceException]]. Instead, you should use the companion
-   * objects [[MillURLClassLoader.apply()]] method.
+   * object's [[MillURLClassLoader.apply()]] method.
    *
    * If you need to use this constructor, make sure to catch the
    * [[DoNotUseThisInstanceException]] and recover by a second constructor attempt,
    * which should succeed.
    *
    * @throws DoNotUseThisInstanceException When this is the first instance to be
-   *                                       created in this JVM instance.
+   *                                       created in this JVM.
    */
   @deprecated("Use the companion object's apply method instead.", "Mill 1.1.0")
   def this(
--- a/core/api/daemon/src/mill/api/daemon/MillURLClassLoader.scala
+++ b/core/api/daemon/src/mill/api/daemon/MillURLClassLoader.scala
@@ -6,17 +6,25 @@ import java.net.URLClassLoader
  * Convenience wrapper around `java.net.URLClassLoader`. Allows configuration
  * of shared class prefixes, and tracks creation/closing to help detect leaks
  */
-class MillURLClassLoader(
+class MillURLClassLoader private (
     classPath: Iterable[java.nio.file.Path],
     parent: ClassLoader,
     sharedLoader: ClassLoader,
     sharedPrefixes: Iterable[String],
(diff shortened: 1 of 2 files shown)


lihaoyi:
What's the result of this change/ Should we expect to see performance improvements in classloading, or something else?

lefou:
In theory, this can result in better classloader performance, since Mill runs by default in parallel and also shares many classloaders between threads. Technically, the `MillURLClassLoader` is just relying on other classloaders which all are parallel capable. But since it is currently not parallel capable itself, the JVM can not rely on it and falls back to use a big lock on the classloader itself, instead of more fine-grained locking on the class name. So this should improve lock contention.

I haven't measured anything, but considered it a good thing to pick this low hanging fruit. Unfortunately, the Scala implementation is quite boilerplate-y, as a caller-sensitive static method must be called, which is quite easy in Java but almost impossible in Scala.

lefou:
When [I found the initial way how to encode the caller-sensitive static call in Scala in 2014](https://sbuild.org/news/2014/12/17/SBuild-0.7.7-About-Thread-Deadlock-fixes-and-parallel-ClassLoaders-in-Scala.html), the use of a parallel-capable classloader in SBuild was measurable, but I don't have any numbers anymore and it was with much older JVMs.

But if we look around, a custom classloader implementation in Gradle also sets this flag. Acccording to a comment on the PR, it improved the speed significantly in one case.

* https://github.com/gradle/gradle/issues/749
* https://github.com/gradle/gradle/pull/772

core/api/daemon/src/mill/api/daemon/MillURLClassLoader.scala
```diff
@@ -6,17 +6,25 @@ import java.net.URLClassLoader
...
       classPath.iterator.map(_.toUri.toURL).toArray,
       MillURLClassLoader.refinePlatformParent(parent)
     ) {
+
   import MillURLClassLoader.*
+
+  if (initParallelClassloading) {
+    java.lang.ClassLoader.registerAsParallelCapable()
+    throw new DoNotUseThisInstanceException()
+  }
```
lihaoyi: Could you add a comment what this is all about?

core/api/daemon/src/mill/api/daemon/MillURLClassLoader.scala
```diff
@@ -6,17 +6,25 @@ import java.net.URLClassLoader
  * Convenience wrapper around `java.net.URLClassLoader`. Allows configuration
  * of shared class prefixes, and tracks creation/closing to help detect leaks
  */
-class MillURLClassLoader(
+class MillURLClassLoader private (
```
lihaoyi, line 35: This is probably what's causing MIMA to fail. We can't privatize the constructor of a class that was public before

core/api/daemon/src/mill/api/daemon/MillURLClassLoader.scala
lefou: Of course.

lefou: Added lots of prosa to scaladoc.

core/api/daemon/src/mill/api/daemon/MillURLClassLoader.scala
```diff
@@ -4,19 +4,89 @@ import java.net.URLClassLoader
...
+
+  if (initParallelClassloading) {
+    // This method is caller sensitive, so we must call it from inside the class.
+    // The just created class instance is improperly initialized, hence we need to not use it.
+    // Therefore, we throw an exception.
+    java.lang.ClassLoader.registerAsParallelCapable()
+    parallelRegistered = true
+    throw new DoNotUseThisInstanceException()
+  }
+
```
lihaoyi, line 57: Could a `@static` annotation be used here? https://docs.scala-lang.org/sips/static-members.html

lihaoyi: that would let us create a static method on the `MillURLClassLoader` companion object to call `registerAsParallelCapable` without needing to create an instance of `MillURLClassLoader`

lefou: To my understanding, no. The `@static` annotation can only be used to make fields of `object`s static, which means, the field is static in the companion object, which has a different name then the companion class. If we use such a static field to register the class as a side-effect, we effectively register an unrelated class, which doesn't work in our interests.

lihaoyi: According to the SIP it works on methods, and makes the annotated object method a static method in the companion class

> Intuitively, the presence of the `@static` annotation ensures that a field/method is declared as a static member of the companion class

lefou: I must have misread that sentence. This would greatly improve the implementation, of course. I can prepare a follow-up PR.

lefou: * #5811

PR #5793, opened 2025-09-01
Read coursier env vars and Java properties via input tasks

alexarchambault:
This is a new attempt at fixing https://github.com/com-lihaoyi/mill/issues/5134. Turns out setting the context class loader when evaluating user code addresses the issue while not running into https://github.com/com-lihaoyi/mill/issues/5398 again.

---

Description from https://github.com/com-lihaoyi/mill/pull/5350 that still applies:

This makes Mill read all environment variables and Java properties used by coursier via a `Task.Input`.

In more detail, this adds a `CoursierConfigModule` external module in scalalib. It contains a single `Task.Input`, that reads all environment variables used by coursier via `Task.env` and all Java properties used by coursier via `sys.props`.

Helper tasks compute default values for various parameters (repositories, cache location, etc.). `CoursierConfigModule.coursierConfig` returns most of these parameters in a case class defined in Mill, `CoursierConfig`. For convenience, down-the-line, we pass `CoursierConfig` instances around.

Fixes https://github.com/com-lihaoyi/mill/issues/5134 (by managing coursier env vars / Java properties via `Task.Input`)

### DIFF
--- a/libs/javalib/src/mill/javalib/Lib.scala
+++ b/libs/javalib/src/mill/javalib/Lib.scala
@@ -39,7 +39,8 @@ object Lib {
       ] = None,
       resolutionParams: ResolutionParams = ResolutionParams(),
       boms: IterableOnce[BomDependency] = Nil,
-      checkGradleModules: Boolean = false
+      checkGradleModules: Boolean = false,
+      config: mill.util.CoursierConfig
   ): Result[Resolution] = {
     val depSeq = deps.iterator.toSeq
     mill.util.Jvm.resolveDependenciesMetadataSafe(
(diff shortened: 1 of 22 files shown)


libs/javalib/src/mill/javalib/Lib.scala
```diff
@@ -89,12 +119,42 @@ object Lib {
...
       resolutionParams = resolutionParams,
-      checkGradleModules = checkGradleModules
+      checkGradleModules = checkGradleModules,
+      config = config
     )
 
     res.map(_.map(_.withRevalidateOnce))
   }
 
+  // bin-compat shim
```
lihaoyi: Can we use `@com.lihaoyi.unroll` for these?

lihaoyi:
Could we add an integration test that calls `./mill` without coursier env vars, and with two different values for them, to ensure that they get picked up without needing to explicitly shut down? This test should fail on the current `main` branch

libs/javalib/src/mill/javalib/Lib.scala
alexarchambault: Should be fixed

nightscape:
@alexarchambault thanks a lot for this!
This has bitten me more than once 😉

Issue #5787, opened 2025-09-01
KSP2 In-Process mode


souvlakias:
Implement an in-process mode for KSP2 to avoid using CLI calls.  

See also:
- https://github.com/com-lihaoyi/mill/pull/5768
- https://github.com/com-lihaoyi/mill/pull/5775#issuecomment-3237514164

Creating this issue in case you’ve already started working on it, @lefou. If not, I can try implementing it.

lefou:
@souvlakias I haven't started to work on it.

souvlakias:
Hi @lefou, should KSP be an external module like in the cases of `androidlib` (e.g., `hiltTransform`, `keytool`, `manifestMerger`)? In those cases, the tasks ultimately call `Jvm.callProcess`, which for our case seems to defeat the purpose of avoiding the CLI call.

Or do we want a `ksp-worker` similar to the various `worker`s in `javalib`?

lefou:
A shared worker solution would be the preferred one, since it

* avoids starting new processes, which will safe OS resources
* avoid IO, by reusing classloader(s)
* avoid memory consumption, by sharing the worker between multiple module
* increase speed, by reusing already JIT-ed classes and further JIT-ing hot code paths

Of course, this assume the tool to run is based on JVM bytecode.

Without a proper shared worker logic, it's probably not worth the effort.  But I think the KSP tooling just depends on the KSP/Kotlin version, so it should improve in build performance and memory usage even for a single KSP module, if we re-build it frequently.

vaslabs:
with the worker, I've checked a few examples, is it the only way to do this via createClassloader -> reflection or are there other ways to not lose static checking?

lihaoyi:
It's basically classloader + reflection. That's the tradeoff we have for loading things dynamically in isolated classloaders

lihaoyi:
The only other alternative is could do subprocess + CLI-args/RPC but that's even less safe

vaslabs:
Maybe `SpringBootToolsImpl` is a good example to combine static checks with minimal reflection, so we do the implementation outside with static checks and we call by reflection "our" API instead of the symbol-processing ones, like a wrapper, so the reflection call is simpler

EDIT: looking at this again, this shouldn't be too tricky with reflection, I think there are only 2 things that are statically referenced [here](https://github.com/google/ksp/blob/main/docs/ksp2entrypoints.md)

vaslabs:
[...]

PR #5795, opened 2025-09-01
Ksp2 in program mode

vaslabs:
Related to: https://github.com/com-lihaoyi/mill/issues/5787

KspWorker accessed by creating 2 shared classloaders ( one for the ksp2 module + including deps and one with the symbol processor dependencies e.g. dagger, hilt, micronaut). The implementation is adopted from [the Google KSP 2 entry point docs](https://github.com/google/ksp/blob/main/docs/ksp2entrypoints.md)

This is working now with kotlin dagger example, Micronaut and hilt (android Todo, Pokédex multi module).

Also seems to resolved the ksp2cli issue on windows when classpaths are long enough to cause issues.

## Implementation Summary

The KspWorkerImpl acts as the KspJvmMain from the cli. It has the symbol-processing-aa, symbol-processing-api and symbol-processing-common-deps as compile only dependencies with a fixed kotlin and KSP version to avoid heavy use of reflection.

When the KspWorkerImpl runs, it needs a classpath which will contain these dependencies (typically with different versions than the ones used in its compiled deps) and instead of passing a processor classpath (like in the KspJvmMain case) a classloader is passed, which is long living in the Jvm process and has the user defined symbol processor classpath (e.g. containing dagger or hilt or Micronaut inject from the existing examples). This replaces the processor classloader creation from the official KSP 2 entry point example.

The above is summarised in the following diagram

image

### DIFF
--- a/libs/kotlinlib/package.mill
+++ b/libs/kotlinlib/package.mill
@@ -10,7 +10,7 @@ import millbuild.*
 //  because currently there is no previous artifact version
 object `package` extends MillStableScalaModule with BuildInfo {
 
-  def moduleDeps = Seq(build.libs.javalib, build.libs.javalib.testrunner, api)
+  def moduleDeps = Seq(build.libs.javalib, build.libs.javalib.testrunner, api, `ksp2-api`)
   def localTestExtraModules =
     super.localTestExtraModules ++ Seq(worker)
 
@@ -68,7 +68,10 @@ object `package` extends MillStableScalaModule with BuildInfo {
       )
   }
 
-  object ksp extends MillKotlinModule {
+  object `ksp2-api` extends MillKotlinModule
+
+  object ksp2 extends MillKotlinModule {
+    def moduleDeps = Seq(`ksp2-api`)
     override def compileMvnDeps: T[Seq[Dep]] = super.compileMvnDeps() ++ Seq(
       mvn"com.google.devtools.ksp:symbol-processing-common-deps:${Deps.kotlinVersion}-${Deps.kspVersion}",
       mvn"com.google.devtools.ksp:symbol-processing-api:${Deps.kotlinVersion}-${Deps.kspVersion}",
--- a/libs/kotlinlib/package.mill
+++ b/libs/kotlinlib/package.mill
@@ -68,4 +68,12 @@ object `package` extends MillStableScalaModule with BuildInfo {
       )
   }
 
+  object ksp extends MillKotlinModule {
+    override def compileMvnDeps: T[Seq[Dep]] = super.compileMvnDeps() ++ Seq(
+      mvn"com.google.devtools.ksp:symbol-processing-common-deps:${Deps.kotlinVersion}-${Deps.kspVersion}",
+      mvn"com.google.devtools.ksp:symbol-processing-api:${Deps.kotlinVersion}-${Deps.kspVersion}",
+      mvn"com.google.devtools.ksp:symbol-processing-aa:${Deps.kotlinVersion}-${Deps.kspVersion}"
+    )
+  }
+
 }
--- a/libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
+++ b/libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
@@ -38,7 +38,6 @@ trait KspModule extends KotlinModule { outer =>
    * [[KspModuleMode.Ksp2]] works with an internal worker that runs KSP in the same JVM as Mill. This is the recommended
    * way to run KSP 2, as it is faster than the CLI mode and doesn't have the cli limitations of Ksp2Cli (e.g. exceeding
    * character limit on Windows).
-   * @return
    */
   def kspModuleMode: KspModuleMode = KspModuleMode.Ksp2
 
@@ -310,7 +309,6 @@ trait KspModule extends KotlinModule { outer =>
    * The classpath used to run KSP 2 in-process worker mode, which is provided via
    * [[KspWorkerModule]]. It includes the KSP 2 API via [[ksp2ToolsDeps]], the user defined symbol processors
    * via [[kotlinSymbolProcessors]] and Mill's kotlinlib-ksp module with the worker that executes the SymbolProcessingProviders.
-   * @return
    */
   def ksp2InProgramToolsClasspath: T[Seq[PathRef]] = Task {
     defaultResolver().classpath(
--- a/libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
+++ b/libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
@@ -456,7 +456,6 @@ object KspWorkerModule extends KspWorkerModule {
 
   def runKsp(logLevel: String, workerClasspath: Seq[PathRef], kspArgs: Seq[String]): Unit = {
 
-    println("worker classpath is: " + workerClasspath.map(_.path).mkString(":"))
     val kspClassLoader = Jvm.createClassLoader(
       workerClasspath.map(_.path),
       getClass.getClassLoader
--- a/libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
+++ b/libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
@@ -25,6 +25,19 @@ import java.io.File
 @mill.api.experimental
 trait KspModule extends KotlinModule { outer =>
 
+  /**
+   * Controls the mechanism in which the Kotlin Symbol Processing is run.
+   * [[KspModuleMode.Ksp1]] works with the embeddable kotlin compiler and via the KSP compiler plugin.
+   * In case of Ksp1 choice, you need to set [[kspLanguageVersion]] to 1.9 or earlier, the compiler flags
+   * via [[ksp1KotlincOptions]]. The mandatory KSP plugins are automatically added via [[ksp1Plugins]].
+   *
+   * [[KspModuleMode.Ksp2Cli]] works with the KSP 2 command line tool `com.google.devtools.ksp.cmdline.KSPJvmMain`
+   * which is run in a separate JVM process.
+   * [[KspModuleMode.Ksp2]] works with an internal worker that runs KSP in the same JVM as Mill. This is the recommended
+   * way to run KSP 2, as it is faster than the CLI mode and doesn't have the cli limitations of Ksp2Cli (e.g. exceeding
+   * character limit on Windows).
+   * @return
+   */
   def kspModuleMode: KspModuleMode = KspModuleMode.Ksp2
 
   /**
@@ -291,6 +304,12 @@ trait KspModule extends KotlinModule { outer =>
     )
   }
 
+  /**
+   * The classpath used to run KSP 2 in-process worker mode, which is provided via
+   * [[KspWorkerModule]]. It includes the KSP 2 API via [[ksp2ToolsDeps]], the user defined symbol processors
+   * via [[kotlinSymbolProcessors]] and Mill's kotlinlib-ksp module with the worker that executes the SymbolProcessingProviders.
+   * @return
+   */
   def ksp2InProgramToolsClasspath: T[Seq[PathRef]] = Task {
     defaultResolver().classpath(
       Seq(
--- a/libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
+++ b/libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
@@ -3,7 +3,7 @@ package mill.kotlinlib.ksp
 import coursier.core.VariantSelector.VariantMatcher
 import coursier.params.ResolutionParams
 import mill.*
-import mill.api.{PathRef, Task}
+import mill.api.{Discover, ExternalModule, ModuleRef, PathRef, Task}
 import mill.kotlinlib.worker.api.KotlinWorkerTarget
 import mill.kotlinlib.{Dep, DepSyntax, KotlinModule, KotlinWorkerManager}
 import mill.util.Jvm
@@ -25,7 +25,7 @@ import java.io.File
 @mill.api.experimental
 trait KspModule extends KotlinModule { outer =>
 
-  def kspModuleMode: KspModuleMode = KspModuleMode.Ksp2Cli
+  def kspModuleMode: KspModuleMode = KspModuleMode.Ksp2
 
   /**
    * The version of the symbol processing library to use, which needs to be compatible with
@@ -110,6 +110,7 @@ trait KspModule extends KotlinModule { outer =>
 
   override def kotlinUseEmbeddableCompiler: Task[Boolean] = kspModuleMode match {
     case KspModuleMode.Ksp1 => Task { true }
+    case KspModuleMode.Ksp2 => Task { super.kotlinUseEmbeddableCompiler() }
     case KspModuleMode.Ksp2Cli => Task { super.kotlinUseEmbeddableCompiler() }
   }
 
@@ -118,6 +119,7 @@ trait KspModule extends KotlinModule { outer =>
    */
   def generatedSourcesWithKsp: T[GeneratedKspSources] = kspModuleMode match {
     case KspModuleMode.Ksp1 => generatedSourcesWithKsp1()
+    case KspModuleMode.Ksp2 => generatedSourcesWithKsp2()
     case KspModuleMode.Ksp2Cli => generatedSourcesWithKsp2Cli()
   }
 
@@ -289,6 +291,14 @@ trait KspModule extends KotlinModule { outer =>
     )
   }
 
+  def ksp2InProgramToolsClasspath: T[Seq[PathRef]] = Task {
+    defaultResolver().classpath(
+      Seq(
+        Dep.millProjectModule("mill-libs-kotlinlib-ksp")
+      ) ++ ksp2ToolsDeps()
+    )
+  }
+
   /**
    * The Kotlin compile task with KSP.
    * This task should run as part of the [[generatedSources]] task to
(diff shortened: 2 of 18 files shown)


vaslabs:
I think [this](https://github.com/com-lihaoyi/mill/pull/5790) is related (as a fix) to the issue I'm having but I'm gonna try to change the implementation to not wrap an extra classloader

vaslabs:
I have a working prototype, @lefou am I doing the mill module ceremonies right or shall I create an interface like the KotlinWorker example?

I've also diverted a bit from the official example to have only one classloader (not wrap an extra one) since the KspWorker classpath is plain and I don't think there will be any conflicts. 

EDIT: maybe a ksp-api module would make sense, to give proper code for the runner configuration. The improvement over kspCli here, is that we can define the log level for KSP (which helped me a lot on the first integration, but I had to recompile the official ksp jars), instead of having a `Map[String, String]` as in my PoC so far and not use the cli arguments, although it speeds up the implementation, but this can be done with proper types

On the other hand, most of the configs are kind of hidden and re-using the cli also avoids duplication when switching between the two, but the latter can be remedied with a typed config that tranforms into cli args

libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -289,6 +291,15 @@ trait KspModule extends KotlinModule { outer =>
     )
   }
 
+  def ksp2InProgramToolsClasspath: T[Seq[PathRef]] = Task {
```
vaslabs, line 311: TODO: docs

libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -25,7 +25,7 @@ import java.io.File
 @mill.api.experimental
 trait KspModule extends KotlinModule { outer =>
 
-  def kspModuleMode: KspModuleMode = KspModuleMode.Ksp2Cli
+  def kspModuleMode: KspModuleMode = KspModuleMode.Ksp2
```
vaslabs, line 42: TODO: docs

libs/kotlinlib/package.mill
```diff
@@ -68,4 +68,12 @@ object `package` extends MillStableScalaModule with BuildInfo {
       )
   }
 
+  object ksp extends MillKotlinModule {
```
lefou: Should we name this `ksp2`?

libs/kotlinlib/src/mill/kotlinlib/ksp/KspWorkerModule.scala
```diff
@@ -0,0 +1,36 @@
+package mill.kotlinlib.ksp
+
+import mill.api.{Discover, ExternalModule, PathRef}
+import mill.util.Jvm
+
+@mill.api.experimental
+trait KspWorkerModule extends ExternalModule {
```
lefou: The `extends ExternalModule` should be placed on the `object` only.

libs/kotlinlib/src/mill/kotlinlib/ksp/KspWorkerModule.scala
```diff
@@ -0,0 +1,36 @@
...
+  def runKsp(logLevel: String, workerClasspath: Seq[PathRef], kspArgs: Seq[String]): Unit
+}
+
+@mill.api.experimental
+object KspWorkerModule extends KspWorkerModule {
+
+  def runKsp(logLevel: String, workerClasspath: Seq[PathRef], kspArgs: Seq[String]): Unit = {
```
lefou: The implementation should be in the `trait` instead.

vaslabs:
@souvlakias does this mode work on windows? could you test when you get the chance

souvlakias:
> @souvlakias does this mode work on windows? could you test when you get the chance

Does work yes

libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -360,6 +390,71 @@ trait KspModule extends KotlinModule { outer =>
...
+      s"-kotlin-output-dir=${kotlin}",
+      s"-java-output-dir=${java}",
+      s"-resource-output-dir=${resources}",
+      s"-language-version=${kspLanguageVersion()}",
+      s"-incremental=true",
+      s"-incremental-log=true",
+      s"-api-version=${kspApiVersion()}",
+      processorOptions,
+      s"-map-annotation-arguments-in-java=false"
+    ) ++ ksp2Args() :+ processorClasspath
```
vaslabs: Need to clean up here, this processor classpath is not needed as we load the whole thing in the classloader (KSP tools deps + processor classpath)

libs/kotlinlib/src/mill/kotlinlib/ksp/KspWorkerModule.scala
```diff
@@ -0,0 +1,36 @@
...
+
+    val kspClassLoader = Jvm.createClassLoader(
+      workerClasspath.map(_.path),
+      getClass.getClassLoader
+    )
```
lefou: Does that mean, we re-create the classloader for every invocation? Instead, we should share the classloader, which should improve performance for repetitive invocations (watch-mode or multi-module use).

lefou (commented):
IIUC, this recreated the tools classpath for each invocation. Can you tell how fast/slow the typical KSP processing is? I think we should try to share the classloader, which should improve the performance of repetitive runs.

lefou:
@vaslabs Can you give a quick outline, what classpaths and classloaders are used and what their lifecycle is?

libs/kotlinlib/src/mill/kotlinlib/ksp/KspWorkerModule.scala
vaslabs: how would it detect classpath changes? can this be achieved by creating a classloader in a task? or could you point me to an example?

vaslabs:
> IIUC, this recreates the tools classpath/classloader for each invocation. Can you tell how fast/slow the typical KSP processing is? I think we should try to share the classloader, which should improve the performance of repetitive runs.

Thanks! I think so, I'll take a look

libs/kotlinlib/src/mill/kotlinlib/ksp/KspWorkerModule.scala
lefou: Lets first name the various classloaders and their lifecycles and then find the best implementation.

lefou: This page should cover the topic:

* https://mill-build.org/mill/extending/running-jvm-code.html

Esp. this section

* https://mill-build.org/mill/extending/running-jvm-code.html#_classloader_worker_tasks

vaslabs: ah found it, I need to have a `Worker[java.net.URLClassLoader]`, but will it be recreated if the classpath changes (e.g. the symbol processor versions are changed)

lefou:
It looks the classpath of the KSP processing tool and the KSP plugins are merge into the same classpath. That's an uncommon setup. Instead, I'd expect to have two classpaths, one for the tool itself, and one for the KSP plugins, which we later give the tool via its options.

The former is a nice candidate for sharing (and caching). The latter is probably shared with the other KSP implementations we have.

vaslabs:
> It looks the classpath of the KSP processing tool and the KSP plugins are merge into the same classpath. That's an uncommon setup. Instead, I'd expect to have two classpaths, one for the tool itself, and one for the KSP plugins, which we later give the tool via its options.
> 
> The former is a nice candidate for sharing (and caching). The latter is probably shared with the other KSP implementations we have.

I'll have another look, I've initially tried to separate them , but the symbol processors (like dagger and hilt) bring their own versions of ksp deps and I thought to give it a go and merge them into the same classpath with the dependency resolution . 

I'll do another attempt to separate them, I think I understand the problem a bit better now

lefou:
> I'll have another look, I've initially tried to separate them , but the symbol processors (like dagger and hilt) bring their own versions of ksp deps and I thought to give it a go and merge them into the same classpath with the dependency resolution .
> 
> I'll do another attempt to separate them, I think I understand the problem a bit better now

The other two implementation also have this separation, so I think it should meld smoothly with dagger and hilt.

vaslabs:
> > I'll have another look, I've initially tried to separate them , but the symbol processors (like dagger and hilt) bring their own versions of ksp deps and I thought to give it a go and merge them into the same classpath with the dependency resolution .
> > I'll do another attempt to separate them, I think I understand the problem a bit better now
> 
> The other two implementation also have this separation, so I think it should meld smoothly with dagger and hilt.

Ok, I think I've done it right and also create a ksp2-api module to bind things more properly.

Now:

- ksp2 tools classpath -> Creates KspWorker (this I followed a bit the KotlinWorker paradigm)
- ksp2-api gives proper types for calling runKsp so the only bit of reflection is in `ksp2Worker`
- the kotlin symbol processors (e.g. hilt, dagger, micronaut) classloader is created with the `ksp2Worker` classloader as its parent so the 2 can discover each other and `KotlinSymbolProcessing` works

Please let me know @lefou , I hope I haven't misunderstood completely :sweat_smile: 


I'm not 100% sure what happens to the worker when dependencies and thus the classpath changes, will do some testing (probably the classloader is not closed)

For the lifecycle, I don't understand it yet, should we cache it and tear it down? In theory it should be similar to the kotlin compiler's

EDIT: 

maybe something like setup and tear down of `ScalaJsWorker`?

EDIT:

Separated the classloader task so the worker management lifecycle (and thus classloader lifecycle) is delegated to mill

vaslabs:
reading [this](https://mill-build.org/mill/extending/running-jvm-code.html#_caching_and_re_using_jvm_subprocesses_and_classloaders) I think what's left is create some sort of Manager class (probably no caching is needed?) that stores the classpath, encapsulates the KspWorker and its classloader and on tear down it closes the classloader?

Or would it work if I just have 2 tasks like this:

```scala
  def ksp2InProgramToolsClassloader: Worker[MillURLClassloader] = Jvm.createClassLoader(
      classPath = ksp2InProgramToolsClasspath().map(_.path),
      parent = getClass.getClassLoader
    )
  def ksp2Worker: Worker[KspWorker] = Task.Worker {
  ksp2InProgramToolsClassloader()
.loadClass("mill.kotlinlib.ksp2.worker.KspWorkerImpl").getConstructor().newInstance()
      .asInstanceOf[KspWorker]
  }
```
as MillUrlClassloader is a Closeable

ah I'll do the latter for now

libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -110,6 +124,7 @@ trait KspModule extends KotlinModule { outer =>
 
   override def kotlinUseEmbeddableCompiler: Task[Boolean] = kspModuleMode match {
     case KspModuleMode.Ksp1 => Task { true }
+    case KspModuleMode.Ksp2 => Task { super.kotlinUseEmbeddableCompiler() }
```
lefou:
```suggestion
    case KspModuleMode.Ksp2 | KspMOduleMode.Ksp2Cli => Task { super.kotlinUseEmbeddableCompiler() }
```

or

```suggestion
    case _ => Task { super.kotlinUseEmbeddableCompiler() }
```

libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -289,6 +305,20 @@ trait KspModule extends KotlinModule { outer =>
     )
   }
 
+  /**
+   * The classpath used to run KSP 2 in-process worker mode, which is provided via
+   * [[KspWorkerModule]]. It includes the KSP 2 API via [[ksp2ToolsDeps]], the user defined symbol processors
+   * via [[kotlinSymbolProcessors]] and Mill's kotlinlib-ksp module with the worker that executes the SymbolProcessingProviders.
```
lefou: Is this still correct? Aren't KSP API/Tools and sysmbol processors in separated classpaths?

libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -360,6 +390,108 @@ trait KspModule extends KotlinModule { outer =>
 
   }
 
+  /**
+   * The classloader with [[ksp2InProgramToolsClasspath]], which includes the KSP 2 Worker
+   * and the user defined symbol processors via [[kotlinSymbolProcessors]]. This classloader is a
+   * parent of [[kotlinSymbolProcessorClassloader]] as classes are shared between the [[ksp2ToolsDeps]]
+   * and [[kotlinSymbolProcessors]] (symbol processors depend on the KSP API).
+   */
+  def ksp2WorkerClassloader: Worker[MillURLClassLoader] = Task.Worker {
```
lefou: Does this really need to be a `MillURLClassLoader`? Why?

libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -360,6 +390,108 @@ trait KspModule extends KotlinModule { outer =>
...
+    Jvm.createClassLoader(
+      classPath = ksp2InProgramToolsClasspath().map(_.path),
+      parent = getClass.getClassLoader
+    )
+  }
+
+  /**
+   * The in-process worker instance of the KSP 2 processor.
+   * The classloader contains [[ksp2InProgramToolsClasspath]], which includes the KSP 2 Worker
+   * and the user defined symbol processors via [[kotlinSymbolProcessors]]. The classloader used here
```
lefou: Comments seems outdate as it mentions a combined classloader.

libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
```diff
@@ -360,6 +390,108 @@ trait KspModule extends KotlinModule { outer =>
...
+   * [[kotlinSymbolProcessors]].
+   *
+   * For more info see reference implementation: [[https://github.com/google/ksp/blob/main/docs/ksp2entrypoints.md]]
+   */
+  def kotlinSymbolProcessorClassloader: Worker[MillURLClassLoader] = Task.Worker {
+    Jvm.createClassLoader(
+      kotlinSymbolProcessorsResolved().map(_.path),
+      parent = ksp2Worker().getClass.getClassLoader
+    )
+  }
```
lefou: I still don't get, why the tool itself needs to be part of this classpath. Isn't this classpath fed to the tool? So, why should I feed the path of the tool to the tool itself? This is rather unusual, and also doesn't match the logic used for KSP 1 and KSP 2 CLI.

vaslabs: The tool, similarly to the main/cli approach, has the KSP deps as its classpath, these are symbol-processor-aa deps and co (kspToolDeps). This allows it to reference the KSP APIs without knowing the actual user defined processors (e.g. dagger-compiler). 
The discovery starts from the ServiceLoader loading SymbolProcessorProvider (s) . This will discover any KSP providers from the processor classpath (the symbol processor deps, dagger, hilt, Micronaut inject kotlin etc). Then the KSP API, accepts a list of symbol processor providers, if the classloaders are independent (no parent relationship) the cast fails (in the Kotlin Symbol processing itself, not in the tool)

Is there a different/better way of doing it?

vaslabs:
Some of the documentation is indeed from my old attempts I'll iron it out. But I tested all the setups against this implementation and they are working.

To summarise:

The KspWorkerImpl acts as the KspJvmMain from the cli. It has the symbol-processing-aa and common-deps as compile only dependencies with a fixed kotlin and KSP version to avoid heavy use of reflection.

When the KspWorkerImpl runs, it needs a classpath which will contain these dependencies (typically with different versions than the ones used in its compiled deps) and instead of passing a processor classpath (like in the KspJvmMain case) a classloader is passed, which is long living in the Jvm process and has the user defined symbol processor classpath (e.g. containing dagger or hilt or Micronaut inject from the existing examples).  This replaces the processor classloader creation from the official KSP 2 entry point example.


I will do some more housekeeping tomorrow and include the above in the pr description too

libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
vaslabs: No, it's from the combined classpath implementation from before, will change, sorry

libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
vaslabs: I thought this was the type typically used. Should it just be URLClassLoader?

lefou: I guess `ClassLoader` would be enough. In the end, all we want is loading classes, and we don't care however this is managed. This gives more flexibility to our users.

libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
lefou: Got it. (I somehow don't feel comfortable with the `ksp2Worker().getClass.getClassloader`, although I can't tell right now, why.) Let's see how it works out.

vaslabs: I can actually change that now to `ksp2WorkerClassloader()` , it was before I introduced that method

libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala
vaslabs: ClassLoader doesn't implement Autoclosable or Closeable though, would that be a problem?

vaslabs:
The only thing that remains is this:

```
  /**
   * The classloader with [[ksp2InProgramToolsClasspath]], which includes the KSP 2 Worker
   * with the KSP deps from [[ksp2InProgramToolsClasspath]]. This classloader is a
   * parent of [[kotlinSymbolProcessorClassloader]] as classes are shared between the [[ksp2ToolsDeps]]
   * and [[kotlinSymbolProcessors]] (symbol processors depend on the KSP API).
   */
  def ksp2WorkerClassloader: Worker[ClassLoader] = Task.Worker {
    Jvm.createClassLoader(
      classPath = ksp2InProgramToolsClasspath().map(_.path),
      parent = getClass.getClassLoader
    )
  }
  ```
  
  I think this can be done without the parent if in the `ksp2InProgramToolsClasspath` also has the ksp2-api and should make the ksp2 fairly isolated, just in case the mill classloader has anything that conflicts with the ksp ones.
  
But for this I will need to make another level of indirection (because my current implementation has the ksp worker args in ksp2-api which mill needs in its classpath). Not sure it's worth it now

**Comment by vaslabs** [comment]
multi-module retest with latest changes

image

File: libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala (hunk 12 above)
Comment (lefou) [reply libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala]: I think what's bugging me is the fact, that we only need the KSP API on the symbol processor classpath, but instead by using the worker classpath as parent, we provide the full implementation.

Comment (vaslabs) [reply libs/kotlinlib/src/mill/kotlinlib/ksp/KspModule.scala]: I know what you mean, I think it's equivalent with the main, because I was wondering how I was getting class cast errors but the KSPJvmMain with

```kotlin
    val processorClassloader = URLClassLoader(classpath.map { File(it).toURI().toURL() }.toTypedArray())
```

didn't, and it seems that this constructor resolves in:

```java
    protected ClassLoader() {
        this(checkCreateClassLoader(), null, /*parent = */getSystemClassLoader());
    }
```

which is the classpath given to jvm (?) when it starts, so from then on , the 2 classloaders are basically the same pair in both cases

lefou: I see. Lets use `ksp2WorkerClassloader()` as parent them, which is a bit more direct.

lihaoyi:
I think this looks fine, I can merge it if you're done @vaslabs

vaslabs:
> I think this looks fine, I can merge it if you're done @vaslabs

Yes, let's merge, I can always follow up if something comes up

PR #5797, opened 2025-09-02
Implement support for Scala 3.8 (backport of #5751)

WojciechMazur:
Backports https://github.com/com-lihaoyi/mill/pull/5751 to 0.12.x branch

### DIFF
--- a/scalalib/api/src/mill/scalalib/api/ZincWorkerUtil.scala
+++ b/scalalib/api/src/mill/scalalib/api/ZincWorkerUtil.scala
@@ -65,7 +65,7 @@ trait ZincWorkerUtil {
     case _ => scalaVersion
   }
 
-  private def majorMinorVersion(version: String): ( /*major: */ Int, /*minor:*/ Int) = version match {
+  private def partialVersion(version: String): ( /*major: */ Int, /*minor:*/ Int) = version match {
     case PartialVersion(major, minor) => (major.toInt, minor.toInt)
   }
 
@@ -175,7 +175,7 @@ trait ZincWorkerUtil {
    * @return `true` if the scala-library version should be enforced to be a `2.13.`
    */
   def enforceScala213Library(scalaVersion: String): Boolean = {
-    val (major, minor) = majorMinorVersion(scalaVersion)
+    val (major, minor) = partial(scalaVersion)
     // Some Dotty versions and all Scala 3 versions before 3.8
     major == 0 || (major == 3 && minor < 8)
   }
--- a/scalalib/api/src/mill/scalalib/api/ZincWorkerUtil.scala
+++ b/scalalib/api/src/mill/scalalib/api/ZincWorkerUtil.scala
@@ -65,7 +65,7 @@ trait ZincWorkerUtil {
     case _ => scalaVersion
   }
 
-  private def binaryVersion(version: String): ( /*major: */ Int, /*minor:*/ Int) = version match {
+  private def majorMinorVersion(version: String): ( /*major: */ Int, /*minor:*/ Int) = version match {
     case PartialVersion(major, minor) => (major.toInt, minor.toInt)
   }
 
@@ -175,7 +175,7 @@ trait ZincWorkerUtil {
    * @return `true` if the scala-library version should be enforced to be a `2.13.`
    */
   def enforceScala213Library(scalaVersion: String): Boolean = {
-    val (major, minor) = binaryVersion(scalaVersion)
+    val (major, minor) = majorMinorVersion(scalaVersion)
     // Some Dotty versions and all Scala 3 versions before 3.8
     major == 0 || (major == 3 && minor < 8)
   }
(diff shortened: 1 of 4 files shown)


scalalib/api/src/mill/scalalib/api/ZincWorkerUtil.scala
```diff
@@ -65,9 +65,8 @@ trait ZincWorkerUtil {
     case _ => scalaVersion
   }
 
-  private case class Version(major: Int, minor: Int)
-  private def minorMajorVersion(version: String): Version = version match {
-    case PartialVersion(major, minor) => Version(major = major.toInt, minor = minor.toInt)
+  private def binaryVersion(version: String): ( /*major: */ Int, /*minor:*/ Int) = version match {
+    case PartialVersion(major, minor) => (major.toInt, minor.toInt)
```
lefou: I wonder what binary issue that may be, since the case class is `private`.

scalalib/api/src/mill/scalalib/api/ZincWorkerUtil.scala
```diff
@@ -65,6 +65,10 @@ trait ZincWorkerUtil {
     case _ => scalaVersion
   }
 
+  private def binaryVersion(version: String): ( /*major: */ Int, /*minor:*/ Int) = version match {
```
lefou: The name `binaryVersion` is a bit misleading, since in Scala 3, a binary version has only the major position.

scalalib/api/src/mill/scalalib/api/ZincWorkerUtil.scala
WojciechMazur: It should be effectively safe, but maybe there is either a false positive or some possible incompatibility when defining nested classes inside traits. 

https://github.com/com-lihaoyi/mill/actions/runs/17394751012/job/49374829543#step:4:308
```
[7629] Found 1 issue when checking against com.lihaoyi:mill-scalalib-api_2.13:0.11.4
[7629]  * abstract synthetic method mill$scalalib$api$ZincWorkerUtil$$Version()mill.scalalib.api.ZincWorkerUtil#Version# in interface mill.scalalib.api.ZincWorkerUtil is present only in current version
   filter with: ProblemFilter.exclude[ReversedMissingMethodProblem]("mill.scalalib.api.ZincWorkerUtil.mill$scalalib$api$ZincWorkerUtil$$Version")
```

lefou: Does making it `private[this]` change anything?

scalalib/api/src/mill/scalalib/api/ZincWorkerUtil.scala
WojciechMazur: I've renamed it to `partialVersion` so that it would be more explicit, would not clash with `xBinaryVersion` methods defined in other places, and would give a hint about. 
In the 1.x it was named `minorMajorVersion` which unfortunately was confusing when we get rid of named tuples

scalalib/api/src/mill/scalalib/api/ZincWorkerUtil.scala
WojciechMazur: Seems like it would fix it. Should I replace the current tuple variant with the private[this] alternative? 
I'm considering keeping the tuple since the `PartialVersion` would clash with existing regex, and `Version` is too opaque

lefou: Yes. I'd prefer if the patch is as close as possible to the original PR. I think `PartialVersion` was unused before and this patch fixes and reuses it.

WojciechMazur:
I think that CI on 0.12.x branch is broken. The remaining failures in CI are the same as on the last PR targeting that branch https://github.com/com-lihaoyi/mill/pull/5639

scalalib/api/src/mill/scalalib/api/ZincWorkerUtil.scala
WojciechMazur: I've made a mistake and tested mima on `scalalib` instead of `scalalib.api` subproject. It still fails when using `private[this]`, I've adjusted mima filter since that should be safe change

lihaoyi:
@WojciechMazur yes the androind and python/js jobs are broken, just ignore them

WojciechMazur:
Thank you for merging the PR. I wonder would it be possible to make a snapshot release for 0.12.16? It would really help us test mill projects in large scale in Open Community Build against Scala 3.8 nightlies and hopefully identify more issues before the cutoff.

lihaoyi:
@WojciechMazur I kicked off a publish job here https://github.com/com-lihaoyi/mill/actions/runs/17429247707, once it's done it should have a dev version you can start trying out

lihaoyi:
@WojciechMazur `0.12.15-2-561986`

Issue #5758, opened 2025-08-27
Android XR support


vaslabs:
I've been working to test mill's capabilities of building Android XR applications. While I managed to build and install it on a non-XR device, I haven't managed to actually test it yet.

It seems XR is only available for Windows and Mac. I have a Mac machine that I'm planning to test so I'm opening this issue to track progress and any issues

For reference, my current experiment is [here](https://developer.android.com/develop/xr/jetpack-xr-sdk/setup)

@souvlakias if you get the chance before me, to test this setup on your windows machine that'd be great!

vaslabs:
I've also tracked down the reason for needing:

```
  // TODO warns about classes that are shipped in devices
  override def androidR8Args = Seq("--map-diagnostics", "error", "warning")
```

most of the errors (that I realised with the XR extensions that are shipped with the devices) are because classes (such as `android.windows.sidecar.*` we usually find in jetpack compose apps) are expected to be present on supported devices.

So we are missing the functionality somewhere, that adds a rule to proguard to ignore this missing links

vaslabs:
Managed to run the XR device, but the app is crashing, seems these native libraries need to be packaged:

<img width="434" height="330" alt="Image" src="https://github.com/user-attachments/assets/e2b1097d-7aca-4d52-b37a-e4cf0ba00a1c" />

vaslabs:
we are missing functionality where we need to package any jni's with the apk

<img width="1327" height="53" alt="Image" src="https://github.com/user-attachments/assets/35ab7433-4e5f-4410-b3c7-f4573b57a056" />

vaslabs:
also, @souvlakias found an issue on windows, he's gonna be following up with a PR

vaslabs:
after fixing the native libs I get

```
java.lang.NoSuchMethodError: No virtual method createSystemRendererConnection(Landroid/app/Activity;Ljava/util/concurrent/Executor;)Lcom/android/extensions/xr/splitengine/SystemRendererConnection; in class Lcom/android/extensions/xr/XrExtensions; or its super classes (declaration of 'com.android.extensions.xr.XrExtensions' appears in /system_ext/framework/com.android.extensions.xr.jar)
```

which is either an issue with the dependency resolution, the rest of the xr dependencies, somehow getting it packaged or proguard rules not being picked from compiled deps and they do something with the method signatures.

but at the moment the 3 found issues are valid for all mainstream apks and will not dig deeper to XR for the time being

[...]

PR #5809, opened 2025-09-03
Android: Add proguard-common.txt

souvlakias:
## This PR resolves 2 issues:

- The need to add extra dependencies due to the R8 detecting internal android classes as missing ([Ref](https://github.com/com-lihaoyi/mill/issues/5758#issuecomment-3227733884))

- The need to pass `--map-diagnostics error warn` to r8 for the same reason.
Now the gradle and mill configurations look quite close to each other

### Motivation
In the thirdparty compose examples, classes such as `androidx.windows.sidecar.**` and `androidx.windows.extensions.**` are provided in runtime and R8 should not warn about them.
We found that AGP seems to [append](https://android.googlesource.com/platform/tools/base/+/refs/heads/studio-master-dev/build-system/gradle-core/src/main/java/com/android/build/gradle/ProguardFiles.java?autodive=0%2F%2F%2F%2F#145) this [common rules file](https://android.googlesource.com/platform/tools/base/+/refs/heads/studio-master-dev/build-system/gradle-core/src/main/java/com/android/build/gradle/ProguardFiles.java?autodive=0%2F%2F%2F%2F#145) in order to avoid such warnings.

### Changes
Added this file in `androidlib`'s resources and always pass it to R8.


example/thirdparty/android-compose-samples/build.mill
```diff
@@ -68,7 +64,7 @@ object JetLagged extends mill.api.Module {
       mvn"androidx.emoji2:emoji2:1.5.0",
       mvn"androidx.emoji2:emoji2-views:1.5.0",
       mvn"androidx.emoji2:emoji2-bundled:1.5.0",
-      mvn"androidx.window:window:1.4.0",
```
vaslabs, line 71: We should be able to completely remove this (window extensions)

libs/androidlib/src/mill/androidlib/AndroidR8AppModule.scala
```diff
@@ -137,6 +137,10 @@ trait AndroidR8AppModule extends AndroidAppModule {
       androidReleaseSettings()
   }
 
+  def androidR8ExtraRules: T[Seq[String]] = Task {
```
vaslabs: Guessing this is a leftover

vaslabs:
I'm suspecting the test apk must have a don't warn flag or even no minification or shrinking, but if this solves all the main apk packaging tasks to not need the error warn mapping, is a nice win

libs/androidlib/src/mill/androidlib/AndroidModule.scala
```diff
@@ -172,6 +172,16 @@ trait AndroidModule extends JavaModule { outer =>
     rules
   }
 
+  // Common Proguard Rules used by AGP
+  // Source: https://android.googlesource.com/platform/tools/base/+/refs/heads/studio-master-dev/build-system/gradle-core/src/main/resources/com/android/build/gradle/proguard-common.txt
+  def commonProguardFile: T[PathRef] = Task {
```
```suggestion
  def androidCommonProguardFile: T[PathRef] = Task {
```

vaslabs:
I have resolved the remaining issues and cleaned up some dependencies from the thirdparty compose samples, which I've also tested, probably androidtodo can be cleaned up from unnecessary dependencies too

vaslabs:
this PR resolves 2 issues:
- The need to add extra dependencies due to the R8 detecting internal android classes as missing
- The need to pass `--map-diagnostics error warn` to r8 for the same reason.

Now the gradle and mill configurations look quite close to each other

@souvlakias when you get the chance, could you update the PR description please (+with the above) and mark the PR as ready when it passes on our copy

vaslabs:
@lihaoyi @lefou this should be ready for review.

I think the failures are unrelated , bootstrap was just `Exception in thread "main" java.lang.Exception: Mill server process already present` . job's green in my copy

vaslabs:
it seems that this PR also fixes the slowness in androidtodo, there's 4 consecutive runs without a timeout, I'll keep an eye. but there were unneeded extra dependencies, which should make the example faster anyway

vaslabs:
heh, spoken too soon

Issue #5796, opened 2025-09-02
forceVersion not working as expected


arminsumo:
simple reproduction:
```scala
object MyModule extends ScalaModule {
  def scalaVersion = "2.12.20"
  def javacOptions = Seq("-source", "17", "-target", "17")
  override def mvnDeps: T[Seq[Dep]] = super.mvnDeps() ++ Seq(
    mvn"org.apache.lucene:lucene-analyzers-common:4.6.1",
    mvn"org.apache.lucene:lucene-core:4.6.0".forceVersion()
  )
}
```

**Expected behavior:** `lucene-core` should resolve to version 4.6.0 due to `.forceVersion()`

**Actual behavior:** `lucene-core` resolves to 4.6.1 instead of 4.6.0

```bash
./mill show MyModule.resolvedMvnDeps | grep lucene
  "qref:v1:20b97b2d:/Users/xxx/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/apache/lucene/lucene-analyzers-common/4.6.1/lucene-analyzers-common-4.6.1.jar",
  "qref:v1:6c091d8a:/Users/xxx/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/apache/lucene/lucene-core/4.6.1/lucene-core-4.6.1.jar"
```

**Workaround:** Using exclude works as expected:
```scala
mvn"org.apache.lucene:lucene-analyzers-common:4.6.1".exclude("org.apache.lucene" -> "lucene-core")
```

The issue appears to be that `.forceVersion()` doesn't override transitive dependencies brought in by other explicitly declared dependencies.

lihaoyi:
CC @alexarchambault who's the expert on Coursier

PR #5813, opened 2025-09-04
Fix forced versions

alexarchambault:
This fixes https://github.com/com-lihaoyi/mill/issues/5796 (forced versions not enforced) by taking them into account in `JavaModule#coursierDependency`, so that they're being enforced during dependency resolution, and are taken into account in other modules if needed.

In order to get forced versions from `mvnDeps`, `JavaModule#coursierDependency` needs to be a task. So in more detail, the current `coursierDependency` is deprecated, and a new `coursierDependencyTask` is used instead.

### DIFF
--- a/libs/javalib/src/mill/javalib/JavaModule.scala
+++ b/libs/javalib/src/mill/javalib/JavaModule.scala
@@ -478,9 +478,11 @@ trait JavaModule
   /**
    * The `coursier.Dependency` to use to refer to this module
    */
+  @deprecated("Use coursierDependencyTask instead", "Mill > 1.0.4")
   def coursierDependency: cs.Dependency =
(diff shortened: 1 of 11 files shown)


libs/javalib/src/mill/javalib/JavaModule.scala
```diff
@@ -478,9 +478,11 @@ trait JavaModule
   /**
    * The `coursier.Dependency` to use to refer to this module
    */
+  @deprecated("Use coursierDependencyTask instead", "Mill > 1.0.4")
```
lefou: `main` branch targets 1.1.0

```suggestion
  @deprecated("Use coursierDependencyTask instead", "Mill 1.1.0")
```

lihaoyi (approved):

Issue #5812, opened 2025-09-04
Improve documentation for Task.Anon (or improve its functionality)


arminsumo:
Current documentation: https://mill-build.org/mill/fundamentals/tasks.html#_anonymous_tasks

***
You can define anonymous tasks using the Task.Anon {...​} syntax. These are not runnable from the command-line, but can be used to share common code you find yourself repeating in Tasks and Commands.

Anonymous task’s output does not need to be JSON-serializable, their output is not cached, and they can be defined with or without arguments. Unlike [Tasks](https://mill-build.org/mill/fundamentals/tasks.html#_tasks) or [Commands](https://mill-build.org/mill/fundamentals/tasks.html#_commands), anonymous tasks can be defined anywhere and passed around any way you want, until you finally make use of them within a downstream task or command.
***

From this description, it is not clear why this does not work
```scala
  def taskA = Task {
    val a = 1+2
    taskB(a)() // instead, doing taskB(1+2)() does work. taskB(taskC())() also does not work.
  }

  def taskB(n: Int) = Task.Anon {
    println(Task.dest) // prints taskA.dest as expected, if there is no compilation error
    println(n)
  }

  def taskC = Task{
    1 + 2
  }
```
results in error
`Task#apply() call cannot use `val a` defined within the Task{...} block`

It seems like this limitation somehow defeats the purpose of Anonymous tasks as helper functions for reusable pieces of code, as it's too limiting to not use any locally defined variables. This leads to a lot of code duplication, because there doesn't seem to be a good way to split a `Task` into smaller pieces. It would be great if Anon task can be updated to reflect this, and also to document what the best practices are for splitting Tasks up into smaller ones when the logic is not trivial.

lefou:
In your example, in `taskA` you create an instance of `taskB` which depends on `a`. This value is only defined inside the task, so it's only known at evaluation time.  But for evaluation, all tasks need already be defined. One way to fix it is to move the definition of value `a` out of the Task body. 

```scala
  def taskA = {
    val a = 1+2
    Task {
      taskB(a)() // instead, doing taskB(1+2)() does work. taskB(taskC())() also does not work.
    }
  }

  def taskB(n: Int) = Task.Anon {
    println(Task.dest) // prints taskA.dest as expected, if there is no compilation error
    println(n)
  }

  def taskC = Task{
    1 + 2
  }
```

lefou:
Another way to fix this is to make the parameter of `taskB` itself a task.

[...]

PR #5816, opened 2025-09-05
document-more-complex-task-usecases

arminsumo:
Please open all PRs as drafts and ensure that your fork of Mill has 
`settings/actions` / `Allow all actions and reusable workflows` enabled to run CI on
your own fork of the Mill repo. Only once CI passes mark the PR as `Ready for review`
and CI will run on the main Mill repo before we merge it.


arminsumo:
I also discovered a potential bug during this. If that is actually intended behavior, I can include it in the documentation
https://github.com/com-lihaoyi/mill/issues/5812#issuecomment-3260302233

example/fundamentals/tasks/3-anonymous-tasks/build.mill
```diff
@@ -23,6 +23,9 @@ def printFileData(fileName: String) = Task.Command {
 // anywhere and passed around any way you want, until you finally make use of them
 // within a downstream task or command.
 //
+// Anonymous tasks work well when parameters are known at task definition time. Dynamically
+// generated inputs can not be used with Anonymous Tasks, as this would break caching.
```
lefou: "as this would break caching." is wrong. It's not about caching, but definition/dependency order.

Since we need to know the parameters before creating the task, but the tasks need to be created before we can start evaluation, we have a classic hen-or-egg problem.

lefou: There are three strategies to resolve this:

* define the values outside of a task, so Mill knows how to create the anonymous tasks before evaluation starts
* put the value itself into a task, so it can be properly ordered to be evaluated before it's dependencies
* Make it not a task but a normal `def`

PR #5817, opened 2025-09-06
scalapb-support-java-generation

arminsumo:
Extend the scalapb plugin to allow for generating scala, or java, or both.
Can probably be extended in the future to support arbitrary generators https://scalapb.github.io/docs/scalapbc/#loading-additional-generators-from-maven

### DIFF
--- a/contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBWorker.scala
+++ b/contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBWorker.scala
@@ -24,7 +24,8 @@ class ScalaPBWorker {
           val scalaPBCompilerClass = cl.loadClass("scalapb.ScalaPBC")
           val mainMethod = scalaPBCompilerClass.getMethod("main", classOf[Array[java.lang.String]])
           val args = otherArgs ++ generators.map { gen =>
-            val opts = if (scalaPBOptions.isEmpty || !gen.supportsScalaPBOptions) "" else scalaPBOptions + ":"
+            val opts = if (scalaPBOptions.isEmpty || !gen.supportsScalaPBOptions) ""
+            else scalaPBOptions + ":"
             s"${gen.generator}=$opts${generatedDirectory.getCanonicalPath}"
           } ++ roots.map(root => s"--proto_path=${root.getCanonicalPath}") ++ sources.map(
             _.getCanonicalPath
@@ -89,7 +90,14 @@ class ScalaPBWorker {
           Seq(ioFile)
     }
     val roots = scalaPBSources.map(_.toIO).filter(_.isDirectory)
-    compiler.compileScalaPB(roots, sources, scalaPBOptions, dest.toIO, scalaPBCExtraArgs, generators)
+    compiler.compileScalaPB(
+      roots,
+      sources,
+      scalaPBOptions,
+      dest.toIO,
+      scalaPBCExtraArgs,
+      generators
+    )
     mill.api.Result.Success(PathRef(dest))
   }
 }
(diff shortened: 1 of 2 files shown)


contrib/scalapblib/test/src/mill/contrib/scalapblib/TutorialTests.scala
```diff
@@ -70,19 +72,68 @@ object TutorialTests extends TestSuite {
...
+    object core extends TutorialModule {
+      override def scalaPBGenerators = Seq(JavaGen)
+    }
+
+    lazy val millDiscover = Discover[this.type]
+  }
+
+  object TutorialWithScalaAndJavaGen extends TutorialBase {
+    object core extends TutorialModule {
+      override def scalaPBGenerators = Seq[Generator](ScalaGen, JavaGen)
```
arminsumo, line 85: Seems like we need to explicitly define the type of the Seq here, otherwise upickle can't find the implicit reader.
although having a single value in there seems to work fine...

contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBWorker.scala
```diff
@@ -97,10 +108,25 @@ trait ScalaPBWorkerApi {
...
+case object ScalaGen extends Generator {
+  override def generator: String = "--scala_out"
+  override def supportsScalaPBOptions: Boolean = true
+}
+case object JavaGen extends Generator {
+  override def generator: String = "--java_out"
+  override def supportsScalaPBOptions: Boolean =
+    false // Java options are specified directly in the proto file
+}
+
```
lefou, line 129: Looks like this could be an enum.

lihaoyi:
@lefou could you help review this one? Thanks!

contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBModule.scala
```diff
@@ -22,6 +22,8 @@ trait ScalaPBModule extends ScalaModule {
 
   def scalaPBVersion: T[String]
 
+  def scalaPBGenerators: T[Seq[Generator]] = Seq[Generator](ScalaGen)
```
lefou, line 25: Can we have some docs here which answers: What is a generator? Which can be choosen from? What is the default?

contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBWorker.scala
lefou: Can you define `Generator` in it's own file? It also needs some docs what `supportsScalaPBOptions` means. I'd prefer to name it `supportsScalaPbOptions` (small `b`).

Are there other generators we could support? Can we add an `CustomGen` to let the user define and use other generators?

contrib/scalapblib/test/src/mill/contrib/scalapblib/TutorialTests.scala
lefou: Does making `Generator` an enum fixes the issue?

contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBWorker.scala
stellerex: > Are there other generators we could support?

Things like OpenAPI generator, probably. I'm using following configuration in one of my projects:
```
override def scalaPBAdditionalArgs: T[Seq[String]] = {
  val classifier = if (Properties.isWin) "windows-x86_64" else "linux-x86_64"
  Seq(
    s"--plugin-artifact=io.grpc:protoc-gen-grpc-java:${versions.grpc}:default,classifier=$classifier,ext=exe,type=jar",
    s"--plugin-artifact=zone.dragon.protobuf:protoc-gen-openapi:${versions.genOpenApi}:default,classifier=$classifier,ext=exe,type=jar,exclude=com.google.protobuf%*,exclude=io.grpc%*",
    s"--java_out=${generatedJavaSources().path.toIO.getCanonicalPath}",
    s"--grpc-java_out=${generatedJavaSources().path.toIO.getCanonicalPath}",
    s"--openapi_out=version=0.0.1,enum_type=string:${generatedOpenApi().path.toIO.getCanonicalPath}"
  )
}
```

lefou:
This was merged as part of #5996. Thank you, @arminsumo !

PR #5819, opened 2025-09-07
Cut down the number of Mill file suffixes to just `.mill`

lihaoyi:
`.sc` was fine as a migration bridge and `.mill.scala` was a good stopgap measure when `.mill` file support in tools wasn't there, but IntelliJ/VSCode/ScalaFmt now have support for `.mill` files so we can consider standardizing on it.

This is a breaking change for local builds but not plugins, so would go into 1.1.0


lefou:
Reverted in #5861

Issue #5714, opened 2025-08-18
Single-file Java, Scala, and Kotlin projects


lihaoyi:
The goal of this ticket is to allow people to directly run `.java`, `.scala`, and `.kotlin` files outside of a project:

`foo.java`
```java
//| mvnDeps: ["org.thymeleaf:thymeleaf:3.1.1.RELEASE"]
import org.thymeleaf.TemplateEngine;
import org.thymeleaf.context.Context;

void main(String[] args) {
  var context = new Context();
  context.setVariable("heading", args[0]);
  context.setVariable("paragraph", args[1]);
  var output = new TemplateEngine().process(
    "<div><h1 th:text=\"${heading}\"></h1><p th:text=\"${paragraph}\"></p></div>",
    context
  );
  System.out.println(output);
}
```

```console
>  ./mill Foo.java run hello world
<div><h1>hello</h1><p>world</p></div>

>  ./mill Foo.java compile

>  ./mill Foo.java show assembly
```

Mill already has the major components for this:

* Header-comment YAML syntax allowing build config to be included in the file the config applies too
* A rich set of configuration values or configuring Java, Scala, and Kotlin modules: `mvnDeps`, `javacOptions`, `scalacOptions`, `kotlincOptions`, `mill-jvm-version`, etc.
* High-quality toolchains for compiling, running, packaging and publishing these modules

This would allow Mill to compete with JBang, ScalaCLI, Ammonite Scripts, and Kotlin Scripting in the ability to write and run small single-file scripts, but with the power of a full build tool behind it for more advanced use cases: autoformatting, executable assemblies, graal native images, publishing to maven central, etc. These single-file modules could also integrate with a larger project, depending on modules defined in `.mill` build files via a `moduleDeps: [str]` YAML config, or be entirely standalone

While `./mill Foo.java`, `./mill Foo.scala`, or `./mill Foo.kotlin` may technically conflict with user-defined modules such as `Foo/java/package.mill`, in practice this should be rare enough we don't have to worry about it.

lefou:
[...]

PR #5820, opened 2025-09-07
Allow running standalone single-file scripts using Mill

lihaoyi:
First pass at https://github.com/com-lihaoyi/mill/issues/5714

Initial pass, this works:

```
$ ./mill -f Foo.java run --text hello

[6972/6972] dist.run
watching processId file (expected content = 63396): /Users/lihaoyi/Github/mill/test/out/mill-no-daemon/pid-63396/processId
============================== run --text hello ==============================
[55/55] run
[55] SLF4J: No SLF4J providers were found.
[55] SLF4J: Defaulting to no-operation (NOP) logger implementation
[55] SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
[55] <h1>hello</h1>
[55/55] ============================== run --text hello ==============================
[6972/6972] ============================== dist.run test -i -f Foo.java run --text hello ============================== 1s
```

The basic approach is 

- Allow `-f`/`--file` to be passed to specify a script file to use rather than `build.mill`
- If `-f` is passed, use an alternate shortened `MillBuildBootstrap` process that has a single build stage with a hard-coded `RootModule` based on the extension of the script: `mill.meta.ScriptModule.Java`, `.Scala`, or `.Kotlin`

In the common case where someone wants to just run a script, they can run:

```
./mill Foo.java --text hello
```

As a shorthand for the full `./mill -f Foo.java run --text hello`. We special-case the first token ending with `.java`, `.scala`, or `.kt` to support this. I expect this would be what people want to do with their scripts the vast majority of the time, so it deserves its own extra-concise syntax even if it has some theoretical overlap with user-defined tasks ending with those names

Other notes: 

- For single-file projects, we assign them a subfolder in the out folder where all the build takes place e.g. `foo/bar/Baz.kt` gets assigned `out/foo/bar/Baz.kt/` as its out folder. Normal tasks and modules won't normally have `.`s in their name, so this ensures that the single-file projects won't collide with normal module and task output files

- Some of the task outputs from `mill.javalib.publish` have been given short-hands `upickle.Reader`s for writing them in the YAML build header. I expect we will need to add more such shorthands in future

- Started with some basic doc sections on the per-language `intro.adoc` and `module-config.adoc` pages touching on usage of these single-file projects, but I expect we'll need more documentation

- For this PR, the single-file projects are entirely standalone and not connected to any enclosing Mill build, but that may be something we can integrate in future to allow single-file projects to define `moduleDeps` on the enclosing `build.mill`'s modules

Tested with a example tests


lefou:
> As a shorthand for the full `./mill -f Foo.java run --text hello`. We special-case the first token ending with `.java`, `.scala`, or `.kt` to support this. I expect this would be what people want to do with their scripts the vast majority of the time, so it deserves its own extra-concise syntax even if it has some theoretical overlap with user-defined tasks ending with those names

We probably should only do this, if there is no `build.mill` files. Otherwise, we should check for the existence of the assumed `.java` file and inform the user that she needs to use `-f`/`--file`.

lefou:
> * For single-file projects, we assign them a subfolder in the out folder where all the build takes place e.g. `foo/bar/Baz.kt` gets assigned `out/foo/bar/Baz.kt/` as its out folder. Normal tasks and modules won't normally have `.`s in their name, so this ensures that the single-file projects won't collide with normal module and task output files

This concept should allow us to also handle classic but renamed `build.mill` files. This might be useful for experimenting or testing stuff by just copying the `build.mill` and running Mill with `mill -f my-copied-build.mill`. E.g. when migrating or refactoring build stuff.

lihaoyi:
Going to merge this first and iterate further on the code in master

lihaoyi:
Decided to revert this for now in https://github.com/com-lihaoyi/mill/commit/72db1f396dd1a6f6205228e7f0aef65c7412bd18; I think it needs a bit more baking and experimentation before it's ready for people to use, and I don't want to commit to the current implementation and semantics just yet

PR #5844, opened 2025-09-12
Fix Scala.js link logs in server-client mode

lolgab:
The Scala.js toolchain calls the logger functions in a separate thread, which breaks the streams swapping Mill does.
This PR, instead of printing directly to `stderr`, adds the lines to a queue and drains them later from the main thread while waiting for the linking future to complete by polling.


lihaoyi:
do try and figure out where the previous logging was going, so we don't just patch over a problem that may come back later in a different context. I'd want to know why it breaks the stream swapping and if there's anything we can do to fix it

lolgab:
Superseded in favor of #5845

PR #5845, opened 2025-09-12
Fix Scala.js link logs in server-client mode

lolgab:
The Scala.js toolchain uses its own threads to print using the logger. `System.err` on Scala.js threads doesn't have the Mill setup to forward the output to the client.
So we store `Console.err` at the beginning of the task (in the Mill thread) which has the correct setup and use that reference to log.
This way logs always reach the terminal correctly.

Pull Request: https://github.com/com-lihaoyi/mill/pull/5845

### DIFF
--- a/libs/scalajslib/test/src/mill/scalajslib/JarPublishRunTests.scala
+++ b/libs/scalajslib/test/src/mill/scalajslib/JarPublishRunTests.scala
@@ -59,10 +59,8 @@ object JarPublishRunTests extends TestSuite {
         val log = os.read(paths.log)
         assert(
           result.evalCount > 0,
-          log.contains("node")
-          // TODO: re-enable somehow
-          // In Scala.js 1.x, println's are sent to the stdout, not to the logger
-          // log.contains("Scala.js")
+          log.contains("node"),
+          log.contains("Scala.js")
         )
       }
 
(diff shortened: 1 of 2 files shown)


lihaoyi:
My understanding is that non-forwarded output is meant to go to `out/mill-daemon/stdout` and `out/mill-daemon/stderr`, and from there is meant to be picked up by the client and send to the terminal asynchronously (and possibly out of order). Can you verify whether or not this is happening?

lefou:
Maybe, we could test this (and reproduce the original issue) by expecting the linker output in one of our example tests. That way, we hopefully could also detect if we break it again.

lolgab:
> erstanding is that non-forwarded output is meant to go to `out/mill-daemon/stdout` and `out/mill-daemon/stderr`, and from there is meant to be picked up by the client and send to the terminal asynchronously (and possibly out of order). Can you verify whether or not this is happening?

I tried on `main` and the linker output doesn't reach the `out/mill-daemon/stderr` file.

lolgab:
> Maybe, we could test this (and reproduce the original issue) by expecting the linker output in one of our example tests. That way, we hopefully could also detect if we break it again.

Thank you, I did that and was able to test. If you change `val err = Console.err` to `def err = Console.err` the tests fails.
I wonder if having the test means that I can delete the comment on `val err = Console.err`

lefou:
> > Maybe, we could test this (and reproduce the original issue) by expecting the linker output in one of our example tests. That way, we hopefully could also detect if we break it again.
> 
> Thank you, I did that and was able to test. If you change `val err = Console.err` to `def err = Console.err` the tests fails. 

Oh, I had an example test in mind, since we then can catch daemon vs. non-daemon mode differences. But looks like you already found and repaired some evidence that something is broken, according to the commented out log tests. :-)

> I wonder if having the test means that I can delete the comment on `val err = Console.err`

Let's keep it, since it provides some extra info.

lolgab:
@lefou @lihaoyi The PR is ready to be merged on my side.

PR #5855, opened 2025-09-13
Replace Yoda conditions

ShreckYe:
A follow-up of #4586. Replaces #5851.

Yoda conditions are unnecessary and seem unidiomatic in Scala. Replacing them might slightly improve readability.


ShreckYe:
To be consolidated into a single PR.

PR #5865, opened 2025-09-16
Clean up daemon/server launching code to try and stabilize it

lihaoyi:
Previously the retries of each step (1) ensure server alive (2) read port (3) connect to socket, happened sequentially with no opportunity for backtracking, which could cause problems if the server was in a bad state, e.g. if the server was alive but shutting down: 

- (1) may succeed but (2) may never become possible
- (1) may succeed and (2) may succeed but (3) may then never work

This PR changes the logic such that the retries for (1) and (2, 3) are nested: we retry the server-alive check, if that succeeds we try to read the port and connect to it, but if either of those fails we go back to retry the server alive check again. This ensures that in the server shutting down scenario, we get a chance to launch a new server if port reading or socket connection fails

### DIFF
--- a/libs/daemon/client/src/mill/client/ServerLaunchResult.java
+++ b/libs/daemon/client/src/mill/client/ServerLaunchResult.java
@@ -1,13 +1,7 @@
 package mill.client;
 
-import java.util.function.Function;
 
 public abstract class ServerLaunchResult {
-  public abstract <R> R fold(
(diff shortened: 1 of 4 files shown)


lihaoyi:
merging this since CI passed on my fork https://github.com/lihaoyi/mill-1/actions/runs/17761082333/job/50474350957

PR #5866, opened 2025-09-16
Improve the `init` `buildgen` code resolving some TODOs

ShreckYe:
Consolidated follow-ups of #4586.

A summary of changes:

1. rename `Config` to `MavenAndGradleCommonConfig` in `BuildGenUtil` to make it clearer
1. rename `optional` in `BuildGenUtil` to `renderIfArgsNonEmpty` to make it more consistent with the other function names
1. remove `moduleOptionTree` in `BuildGenBase.convert` which seems redundant
1. convert generic type parameters to abstract type members for `BuildGenBase` and its subtypes

   Also see the TODO comment which is resolved and removed and https://github.com/com-lihaoyi/mill/pull/4586/files#r1970795027. As these types `M`, `D`, and `I` are only used in the trait and its subtypes, I think parameterization is not needed here, so replacing these type parameters with type members make them more readable and also consistent with the type member `C`.

1. remove some TODOs of slight significance without resolving them and convert some into ordinary comments
1. replace Yoda conditions
1. rename `IrBuild` to `IrModuleBuild`
1. Rename `IrTrait` to `IrBaseInfo` and replace the original references of `IrBaseInfo` with `Option[IrBaseInfo]`

   The original name `IrTrait` seems too broad and is not very readable.

1. rename `extractScopedDeps` to `extractConfigurationDeps` in `GradleBuildGenMain`
1. Change `I` to `(BuildExport, Tree[Node[Option[Project]]])` in `SbtBuildGenMain` and refactor `getBaseInfo` to retrieve `buildExport.projects` directly

See the resolved and removed TODOs for more detailed explanations of the changes. Also see the added TODO comments showing some alternative renaming options and remove them before merging.

### DIFF
--- a/libs/init/buildgen/src/mill/main/buildgen/BuildGenBase.scala
+++ b/libs/init/buildgen/src/mill/main/buildgen/BuildGenBase.scala
@@ -32,7 +32,7 @@ trait BuildGenBase {
   def getModuleFqnMap(moduleNodes: Seq[Node[Module]]): ModuleFqnMap
 
   def convert(
-               input: Input,
+      input: Input,
       cfg: C,
       shared: BuildGenUtil.BasicConfig
   ): Tree[Node[BuildObject]] = {
@@ -86,7 +86,7 @@ trait BuildGenBase {
   def getSupertypes(cfg: C, baseInfo: Option[IrBaseInfo], build: Node[Module]): Seq[String]
 
   def getBaseInfo(
-                   input: Input,
+      input: Input,
       cfg: C,
       baseModule: String,
       packagesSize: Int


libs/init/buildgen/src/mill/main/buildgen/BuildGenBase.scala
```diff
@@ -4,13 +4,12 @@ import mill.main.buildgen.BuildGenUtil.{compactBuildTree, writeBuildObject}
...
-TODO Can we just convert all generic type parameters to abstract type members?
- See https://stackoverflow.com/a/1154727/5082913.
- I think abstract type members are preferred in this case.
- */
-trait BuildGenBase[M, D, I] {
+trait BuildGenBase {
+  type M
+  type D
+  type I
   type C
```
lefou: Can we have more descriptive names here?

libs/init/gradle/src/mill/main/gradle/GradleBuildGenMain.scala
```diff
@@ -269,16 +272,15 @@ object GradleBuildGenMain extends BuildGenBase.MavenAndGradle[ProjectModel, Dep]
...
+  def extractConfigurationDeps(
       project: ProjectModel,
       getModuleFqn: PartialFunction[String, String],
       cfg: Config
   ): IrScopedDeps = {
     var sd = IrScopedDeps()
     val hasTest = os.exists(os.Path(project.directory()) / "src/test")
     val _java = project._java()
-    if (null != _java) {
+    if (_java != null) {
```
lefou: Can you also check if `_java` can get a better name, without that `_` prefix?

libs/init/buildgen/src/mill/main/buildgen/BuildGenBase.scala
ShreckYe: Yeah, they are respectively `Module`, `Dependency`, `Input`, and `Config`.

ShreckYe: I also just noticed that `D`/`Dependency` is no longer used in the current implementation. Shall I remove it or just keep it for possible use and better readability?

libs/init/gradle/src/mill/main/gradle/GradleBuildGenMain.scala
ShreckYe: Yeah its type is `JavaModel`, so I guess it can be renamed to `javaModel`.

libs/init/buildgen/src/mill/main/buildgen/BuildGenBase.scala
lefou: If it's unused, remove it.

ShreckYe: I have removed `D` and renamed `M` and `I`. `C` is kept as renaming it would conflict with the `Config` nested classes. I guess it can be renamed to something like `ConfigT` if needed, but I am not sure whether this is conventional in Scala.

PR #5867, opened 2025-09-16
Ensure deprecation versions always contain the project name

lefou:
This greatly improves compiler warning readability and is best practice.


lefou:
@arturaz FYI

PR #5871, opened 2025-09-16
[config.artifactory] Read credential configs for publish

mga-sc:
Currently, the `publishArtifactory` task requires credentials to be supplied via environment variables or command-line arguments. If these are not provided, the system will fall back to searching for credentials in the Coursier configuration files.


mag-mga:
@lefou , @lihaoyi , please, review this small fix

mag-mga:
@lefou , @lihaoyi , ping

PR #5897, opened 2025-09-22
Bump Scala to 3.7.3

lihaoyi:


lefou:
I think, this requires Scala 3.7.3 support in `mill-moduledefs` first, which we ideally back-publish for the currently used version `0.11.10`.

Issue #5910, opened 2025-09-24
`selective.run` seems borked in CI


lihaoyi:
We're not seeing tests being skipped on PRs, e.g. https://github.com/com-lihaoyi/mill/pull/5909

`selective.resolve` and `selective.resolveTree` both give empty lists, but `selective.run` is running all tests. Something must be broken between them

Naively running through `selective.prepare` and `selective.run` on my laptop doesn't seem to surface the problem

lihaoyi:
Minimizing the test command to `./mill -i -k selective.run 'core.api.test'` still reproduces the problem in CI, but not locally on my laptop

lihaoyi:
Should be fixed by https://github.com/com-lihaoyi/mill/pull/5916 I think

PR #5916, opened 2025-09-24
Try to remove `COURSIER_ARCHIVE_CACHE` configuration

lihaoyi:
Inconsistent application of this environment variable seems to be the cause of https://github.com/com-lihaoyi/mill/issues/5910. Before applying it everywhere, first let's try removing it to see if it's still needed at all

### DIFF
--- a/.github/actions/post-build-selective/action.yml
+++ b/.github/actions/post-build-selective/action.yml
@@ -8,10 +8,6 @@ inputs:
     required: true
     type: string
 
-  coursierarchive:
-    default: ''
-    type: string
-
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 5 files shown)


lihaoyi:
Seems like the original failure that made us add this env var has gone away, and the native image test now passes https://github.com/com-lihaoyi/mill/actions/runs/17987654562/job/51170296930?pr=5916

lihaoyi:
@alexarchambault do you know if you changed anything in Coursier related to the `COURSIER_ARCHIVE_CACHE` path? I'm wondering if there's some change in Coursier that can explain why this `COURSIER_ARCHIVE_CACHE` override that we previously required to avoid windows path length issues seems to be no longer necessary

PR #5930, opened 2025-09-27
Free individual `IRFileCache.Cache` instances when garbage collected

lolgab:
We keep a global `IRFileCache` inside `ScalaJSWorkerImpl`. which creates individual  `IRFileCache.Cache` instances. When they are garbage collected, they leave garbage in the root IRFileCache unless you manually call `free()` on them. This uses `java.lang.ref.Cleaner` to call `free()` when they are garbage collected.


lolgab:
This can't work with Scala.js API, since the free method is on the instance that is being garbage collected.

PR #5932, opened 2025-09-27
Try using `-Ymagic-offset-header` to fixup line numbers

lihaoyi:
This flag lets us remove a lot of our sketchy user-land workarounds to fix the filename and line numbers of errors:

- `AsmPositionUpdater` to which handled runtime exception error positions
- `PositionalMapper` and `TransformingReporter` which handled compile-time error positions
- `//SOURCECODE_ORIGINAL_FILE_PATH` and `//SOURCECODE_ORIGINAL_CODE_START_MARKER`, which handled `sourcecode.File`/`sourcecode.Line` positioning

This replaces it with a single `///SOURCE_CODE_START` comment referenced by `-Ymagic-offset-header:SOURCE_CODE_START`, which hooks in early enough in the compiler pipeline that all these downstream use cases can rely on it

The new `///SOURCE_CODE_START` comment relies on character offsets unlike line numbers like the previous hacks, and so we needed to change our `CodeGen.scala` code to avoid mangling lines within the user code block. Concretely, we needed to stop translating `object package` into `abstract class package extends blahblahblah`. Instead we translate `object package` into `class  _package` (which takes up the same number of characters to avoid affecting offsets) and move the `extends blahblahblah` into the generated code above the `///SOURCE_CODE_START` header


lihaoyi:
Next blocker seems to be that Mill's codegen desugars 

```scala
object `package` extends mill.Module {
```
```scala
abstract class package_  extends _root_.mill.util.MainRootModule, mill.Module {
```

This didn't affect the old error position fixup mechanism because it was line based, but the new one is offset based and so the longer line screws up all the subsequent offsets

lihaoyi:
Seems we might not be able to merge this until Mill 2.x, as the new codegen would require that `MainModule` be turned into a proper `trait` (without extending from `class RootModule0`) in order to make the only-inherit-from-one-class requirement line up.

Fixable, but would break binary compatibility to fix

lihaoyi:
Got most tests passing with a modified codegen strategy, but will need to wait for Scala 3.8.0 to land with https://github.com/scala/scala3/pull/24301 before we can merge this since that PR will likely miss 3.7.4

lihaoyi:
Can't land this yet because https://github.com/scala/scala3/pull/24301 hasn't made it into 3.8.0-RC2

lihaoyi:
Now blocked on https://github.com/scala/scala3/pull/25706

PR #5933, opened 2025-09-28
Clean Linker caches when ScalaJSWorker is closed

lolgab:
This is a smaller step compared to #5931. Still doesn't fix all the problems with linkers, but it's enough to guarantee teardown happens correctly and Scala.js linker caches are cleaned.

Pull Request: https://github.com/com-lihaoyi/mill/pull/5933

### DIFF
--- a/libs/scalajslib/worker/1/src/mill/scalajslib/worker/ScalaJSWorkerImpl.scala
+++ b/libs/scalajslib/worker/1/src/mill/scalajslib/worker/ScalaJSWorkerImpl.scala
@@ -1,11 +1,15 @@
 package mill.scalajslib.worker
 
-import scala.concurrent.Await
-import scala.concurrent.duration.Duration
 import java.io.File
 import java.nio.file.Path
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+
+import com.armanbilge.sjsimportmap.ImportMappedIRFile
+import mill.constants.InputPumper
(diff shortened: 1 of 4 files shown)


libs/scalajslib/worker/1/src/mill/scalajslib/worker/ScalaJSWorkerImpl.scala
```diff
@@ -22,12 +26,7 @@ import org.scalajs.jsenv.{Input, JSEnv, RunConfig}
 import org.scalajs.testing.adapter.TestAdapter
 import org.scalajs.testing.adapter.TestAdapterInitializer as TAI
 
-import scala.collection.mutable
-import scala.ref.SoftReference
-import com.armanbilge.sjsimportmap.ImportMappedIRFile
-import mill.constants.InputPumper
-
-class ScalaJSWorkerImpl extends ScalaJSWorkerApi {
+class ScalaJSWorkerImpl(jobs: Int) extends ScalaJSWorkerApi {
```
lefou, line 29: Having the jobs in the ctr means, we need to drop all cached linkers (since we need to recreate the instance) when the jobs config changes, which looks wasteful to me. But after a closer look, the `jobs` value is already a ctr-arg of the `ScalaJSWorker`, which already suffers from the same issue. So, this PR looks ok to me, but we probably should overhaul the design as a follow-up.

lefou (approved):

PR #5949, opened 2025-10-03
Run examples with local repo under out/ rather than ~/.ivy2/local

alexarchambault:
This changes how Mill is run in integration tests and examples. This makes it stop publishing things under `~/.ivy2/local`, and use something like `out/dist/localRepo.dest` instead, which is the `Task.dest` of the new `dist.localRepo` task. That new repository is a local Maven repositiory rather than an Ivy one, which has a slightly different file structure, the same as Maven Central.

For that new repository to be taken into account, the Mill launcher used in examples and ITs is now a script (bat under Windows, bash everywhere else) that sets the `COURSIER_REPOSITORIES` env var, so that the new local repo is added to it, and gets automatically used when running examples and ITs.

The second commit of this PR mainly updates the file names of the locally published Mill artifacts, because of the switch from an Ivy repository to a Maven one.

### DIFF
--- a/dist/package.mill
+++ b/dist/package.mill
@@ -9,6 +9,9 @@ import mill.util.Jvm
 import millbuild.*
 import mill.api.BuildCtx
 import scala.util.Using
+import scala.util.Properties
+import java.nio.file.Files
+import java.nio.file.attribute.PosixFilePermission
 
 trait DistModule extends Module {
   // All modules that we want to aggregate as part of this `dev` assembly.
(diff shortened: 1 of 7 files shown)


lihaoyi:
What's the benefit of making these changes?

alexarchambault:
> What's the benefit of making these changes?

Sorry, forgot to mention that in the PR description. That's been useful to me when using a locally published Mill version, to develop on Mill or other projects. Running examples or ITs erases it, which is annoying, but also somewhat unexpected (running tests shouldn't erase things in the local repository, at least not things put there manually on purpose).

lihaoyi (approved):

dist/package.mill
```diff
@@ -53,16 +95,30 @@ trait DistModule extends Module {
...
+      Task.Anon {
+        val targetFile = os.Path(binFile(), BuildCtx.workspaceRoot)
+        if (os.exists(targetFile))
+          Task.log.info(s"Overwriting existing local Mill binary at ${targetFile}")
+        os.copy.over(executable().path, targetFile, createFolders = true)
+        Task.log.info(
+          s"Published ${build.dist.allPublishModules.size} modules under local repo and installed ${targetFile}"
+        )
+        targetFile
+      }
```
lefou: Looks like having two anon tasks `installLocalTask` and `installGlobalTask` could make this simpler.

alexarchambault: Should be fixed

alexarchambault:
@lefou What do you mean? It makes them sensitive to the Mill version not being `SNAPSHOT` for sure.

lefou:
> @lefou What do you mean? It makes them sensitive to the Mill version not being `SNAPSHOT` for sure.

Yeah. I mean, it would be nice if the test suite still works in a release build. But a release build will most likely not have the version `SNAPSHOT`.

lefou:
To make it clear, I really support the idea of using a local repo instead of `~/.ivy2/local`. Out of curiosity, was the choice to switch to a local Maven repository necessary for technical reasons, or is it just the more conventioal repository format? I'm asking, since using a local ivy repo would most likely avoid the issue of integration tests not working with a non-`SNAPSHOT` version. But maybe, we can find another solution, e.g. replace the version string in the golden literals.

PR #5950, opened 2025-10-03
Update remaining examples still referencing `JvmWorkerModule.jvmId`

seschwar:
With the move of `jvmId` from `JvmWorkerModule` to `JavaModule` in PR #5327 a few examples were not updated.


lihaoyi:
thanks!

Issue #5961, opened 2025-10-07
Android: Reduce size of test apk


vaslabs:
Due to using the main android sources in android instrumented tests' module deps, the main source code is packaged together.

I think we can declare it as compile module deps and have the test APK only contain the test compilation result and dependencies.

I'm not entirely sure how to do this, I've done a few failed attempts but after newest R8 improvements and findings, we can give this a proper go

vaslabs:
if we take Jetsnack as an example, the test apk is <1MB in size with gradle and doesn't have any "production" code packaged

<img width="1191" height="648" alt="Image" src="https://github.com/user-attachments/assets/1d90bb22-4650-4903-981a-e3ece90349fd" />

while the android test apk we generate with mill is 18MB

<img width="1461" height="571" alt="Image" src="https://github.com/user-attachments/assets/1d417301-ed88-4e37-abba-06b4c72de944" />

and has all the main sources + the test ones. 

So it looks like we can avoid re-packaging the main sources

vaslabs:
looking at the gradle apk generated, it has signs that R8 is erasing bytecode as some classes have empty method signatures. I'm trying to replicate the same behaviour in mill

PR #5963, opened 2025-10-08
Android: Test apk size reduction

vaslabs:
For https://github.com/com-lihaoyi/mill/issues/5961

## Implementation

- Add the outer AndroidAppModule in the compileModuleDeps instead of moduleDeps
- Fix some packaging mappings to use run resolved deps
-  Fix the compile only classpath to pass the compileModuleDeps classpaths too
- By having instrumented tests inherit the `androidDefaultProguardFileNames` from the parent module, the android tests are working. It's not clear to me why, I found it due to some trial and error. To achieve this kind of inheritance without major rewiring I've moved the method to `AndroidAppModule`

## Other fixes and cleanups
- Somehow androidtodo had a wrong namespace, maybe a find and replace mishap
- Removed an unnecessary dep from androidtodo test module

## Outcome

This avoids packaging the main (app) module dependencies and source code into the test apk.

## Result (Jetsnack)

image

## Future work

I had some difficulties reaching this point and I think the 13MB reduction is very satisfiable at this stage. In the gradle's case, it appears some classes are erased by R8 but I haven't pinpointed exactly how it does it.

### DIFF
--- a/example/thirdparty/androidtodo/build.mill
+++ b/example/thirdparty/androidtodo/build.mill
@@ -133,7 +133,6 @@ object app
     def mvnDeps = super.mvnDeps() ++ composeDeps ++ Seq(
       mvn"junit:junit:4.13.2",
       mvn"androidx.arch.core:core-testing:2.2.0",
-      mvn"org.jetbrains.kotlinx:kotlinx-coroutines-android:1.9.0",
       mvn"org.jetbrains.kotlinx:kotlinx-coroutines-test:1.9.0",
       mvn"androidx.navigation:navigation-testing:2.8.5",
       mvn"androidx.test.espresso:espresso-core:3.6.1",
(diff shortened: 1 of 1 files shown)


vaslabs:
the included sample apps are passing but I get failures in jetsnack when running the tests

vaslabs:
Managed to make it work

image

it's not in the <1MB level of gradle yet, but it's getting there.

vaslabs:
need only to fix the Android Res so leaving this as a draft for now

vaslabs:
## More Tests

Ref: https://github.com/vaslabs-ltd/compose-samples-with-mill/tree/android-test-apk-opts

### JetNews

image

### JetLagged

image

vaslabs:
I think overall it's an improvement on the build flow (some fixes to pick run deps for tests and some transitive dependencies on android resources) and reduces the android test apk by a lot (13-15MB reduction) in most of the cases.

Ready for review @lihaoyi @lefou

Issue #5862, opened 2025-09-15
Android: configurable option to transitively pull resources from dependencies


vaslabs:
In gradle [there's an option](https://github.com/vaslabs/Pokedex_Compose_Multi_Module/blob/main/gradle.properties#L23) to pull resources transitively.

We should have the option too. The current behaviour is that we pull the dependencies (so we have the whole dependency tree) and then get the resources from everything (I maybe mistaken about this behaviour so it needs clarification as well).

This behaviour also namespaces the R class of each dependency, so the code will have the fully qualified package name for referencing the R class.

I think, by having that option, we can resolve the top level dependencies only, pull less resources and thus speed up building times as well.

It might be, that we instead go through each dependency and generate the RClass within its own namespace.

Requires investigation

The motivation is:
- Reduce size of apk
- Reduce build times
- Be closer to the end apk as generated by Gradle

lihaoyi:
I guess if we're not sure what to do, we can provide an option and make it configurable as well?

vaslabs:
> I guess if we're not sure what to do, we can provide an option and make it configurable as well?

Yes, it will most likely be configurable one way or another. 
I opened the issue to help me focus on the resources, even though R8 cleans up a lot of unused resources, I think we can optimise better. I think Gradle > 8 has this option on by default too. 

I'll spend some time investigating first and I'll know more

PR #5980, opened 2025-10-15
Android: Add nonTransitiveRClasses Option

souvlakias:
https://github.com/com-lihaoyi/mill/issues/5862
This feature affects multi-module android applications. [This blog](https://stefma.medium.com/unveiling-the-reality-my-experiment-with-non-transitive-r-classes-and-compilation-avoidance-6296922db40f) helps on understanding its benefits.

The basic idea is compilation avoidance: in a dependency tree that’s at least three levels deep, modifying a node causes only that node and its parent to recompile, instead of propagating all the way up to the root.
([Test Branch](https://github.com/vaslabs-ltd/Pokedex_Compose_Multi_Module/tree/testing-mill-non-transitive-r-classes))

Default value is set to true, like AGP>=8

### DIFF
--- a/libs/androidlib/src/mill/androidlib/AndroidModule.scala
+++ b/libs/androidlib/src/mill/androidlib/AndroidModule.scala
@@ -210,8 +210,11 @@ trait AndroidModule extends JavaModule { outer =>
    * This does not include direct dependencies, meaning
    * these are only the dependencies of the dependencies.
    */
-  def androidTransitiveModuleDeps: Seq[JavaModule] =
-    transitiveModuleRunModuleDeps.filterNot(moduleDepsChecked.toSet)
+  def androidTransitiveModuleDeps: Seq[JavaModule] = {
+    val moduleDepsCheckedSet = moduleDepsChecked.toSet
+    val isDirectDependency = (m: JavaModule) => moduleDepsCheckedSet.contains(m)
+    transitiveModuleRunModuleDeps.filterNot(isDirectDependency)
+  }
 
   /**
    * Gets all the transitive compiled Android resources (typically in res/ directory)


vaslabs:
Can we have a comparison of the difference? E.g. a line count for the file input on the aapt2 compile or/and linking for each configuration

souvlakias:
> Can we have a comparison of the difference? E.g. a line count for the file input on the aapt2 compile or/and linking for each configuration
## transitiveRClasses:
image
image


## NonTransitiveRClasses:
image 
image

## Comments
The line count difference on the aapt2 link args is not significant, since the `ModuleCompiledResources` are way less compared to the `CompiledLibResources`. 
Nevertheless the goal of Compilation Avoidance is achieved here, since modules only depend on their direct dependencies.
E.g. modifying `Database`'s resources would only cause itself and `network` to recompile and not any other modules which was the case before.

libs/androidlib/src/mill/androidlib/AndroidModule.scala
```diff
@@ -205,6 +205,15 @@ trait AndroidModule extends JavaModule { outer =>
     }().flatten.distinct
   }
 
+  def androidDirectCompiledResources: T[Seq[PathRef]] = Task {
+    Task.traverse(moduleDepsChecked) {
+      case m: AndroidModule =>
+        Task.Anon(m.androidCompiledModuleResources())
```
lefou:
```suggestion
        m.androidCompiledModuleResources
```

lefou (commented):
We now have `androidDirectCompiledResources` and `androidTransitiveCompiledResources`. The latter is a super-set of the former. 

I was thinking, if it makes sense to have two disjunct sets, meaning, the transitive set contains only those resources not already in the direct resources? That way, we could easily spot what resources we potentially miss with `mill show _.androidTransitiveCompiledResources`. But it also makes reasoning a bit easier, since the direct resources are always used. Meaning, we have a more stable task dependency tree.

PR #6005, opened 2025-10-20
Re-land "Avoid using system `java` unless explicitly specified, instead use latest LTS Java version unless explicitly overridden with `system`"

lihaoyi:
Merges https://github.com/com-lihaoyi/mill/pull/5705 again. Description reproduced below:

This would make Mill builds more reproducible, and avoid the JVM-version confusion that e.g. @makingthematrix encountered recently in the discord channel.

The basic question is what to do if `//| mill-jvm-version` is not explicitly set. It seems all options have tradeoffs

1. Not providing any default and forcing the user to provide one
    * Annoying for first-time users, which have to manually set a `//| mill-jvm-version` even if they don't care about it
    * Even plenty of advanced users or large projects don't _really_ care about their JVM version, so asking them to pick one is annoying when we could just pick something for them and they'll be fine
    * If Mill aims to support non-JVM toolchains like Python or JS, then they don't care about this either and would be happy for some default to be picked
    * We'd need special handling for integration and example tests, and internal uses like `dist.run` or `dist.installLocal`, where providing an explicit JVM for every test case would be super annoying. But special handling would make them diverge from normal build files (although they already diverge, e.g. not using a bootstrap script and not requiring a `mill-version` specified)
 
2. Using some hardcoded default even if there is a `java` environmentally installed
    * Can cause confusion since people would expect the `java` they installed to take effect in the absence of other explicit configuration
    * Anyone upgrading from earlier versions of Mill could have their JVM version change silently, possibly causing published artifacts to silently require newer JVMs to run and causing downstream bytecode validation errors later
    * If we keep the default version stable, it will inevitable fall out of date over time. If we keep the default version updated, it might change the user's JVM version every time they update Mill
    * There is no "generally safe" default version to use: libraries would prefer the lowest possible version to not impose requirements on downstream users, while applications would prefer the highest possible version to take advantage of JVM improvements. `com-lihaoyi` libraries are built with 11, Spring-Boot requires 17, etc.
    * The existence of a default version that the user has to notice is incorrect and then override adds substantial user-facing complexity over forcing them to pick a version in the first place, since in the latter case at least the user would be aware of the value and it's ability to be configured

3. Using the environmental `java` installation as the default, 
    * Matches the "default" way most people work with JVMs, but that default way sucks
    * Someone who checks out a codebase on a different machine with some other machine with a different `java` installation would encounter failures. 
    * While users could always resolve this issue by setting an explicit `//| mill-jvm-version` in their build header, most users likely won't by default, won't know that they should, and when they hit weird Java version errors they won't know that this is the solution for it. e.g. all the `com-lihaoyi` projects never did this, and if we as the experts aren't doing this ourselves we can't realistically expect others to do so

We are currently at (3), but of these options I think (2) is probably the best:

(1) Adds significant increased up-front inconvenience. This can be mitigated by copy-pasteable instructions on `installation-ide.adoc`, but then people are likely to end up with the wrong JVM version anyway by following the default, and so there's not that much value in forcing it to be visible in the build file if they don't care about it

(2) Has the problem that using the non-system version is surprising, but has the advantage that people who "don't care" about their JVM version don't need to do/write/see anything, and those that do care can just set it. We can mitigate this problem by putting it in the changelog and docs where people install Mill and configure JVM versions, with instructions on how to set it 

The longer-term convenience of "Mill projects are easy for anyone to checkout and build" would definitely be very valuable, and can be a differentiator over other build tools where that is generally not the case

As the default version, the policy is to choose the latest LTS version of the JVM, which at this point in time is 21.

This is a breaking change, but it doesn't break binary compatibility, so it'll likely go into 1.1.0. Users can fall back to the prior behavior using `mill-jvm-version: system` or `mill-jvm-version: 17` as necessary


Ichoran:
Personally, I find (2) the most infuriating aspect of using scala-cli, because it ignores my system Java which is 21+ because I always want virtual threads available, without exception; and then it runs things without virtual threads unless I go to extra effort to make it happen.

It is a small itch, but it annoys me every single time because there is no reason it should be ignoring me.  I can understand automatically _upgrading_ a system JVM which is too low, but not automatically _downgrading_ a system JVM which is presumably as upgraded as it is for a reason.

So, anyway, personally I do not think (2) is better for _me_.

It may, however, be better for other people, for reasons that do not exactly make sense to me, but anyway, I can put up with it.

lefou:
@Ichoran same for me. I don't think we should enforce a default JVM version in Mill if there is a usable java version on the path. We might instead detect the fact that we rely on a system-provided JVM version and print some info message suggesting the user to explicitly configure it in the project.

Issue #6006, opened 2025-10-20
Spring Boot: Document getting started docs


vaslabs:
This is to track the status quo while I'm testing various spring boot guides against mill, see what's working and any possible issues/difficulties in setting up.

I'll accumulate all my findings in this issue and then we can create a spring boot guide focused on mill in the docs.

For potential issues/fixes/missing features I'll open different issues

vaslabs:
## Spring Initializr

- Generate a sample from https://start.spring.io/ or using IntelliJ spring boot initializr plugin
- Run `./mill init`

### What's working

- `./mill run` works
- Test trait is created, but neither `./mill test` or `./mill __.test` work, a test target doesn't seem to exist

```scala
  trait test extends MavenTests {

    def mvnDeps =
      Seq(mvn"org.springframework.boot:spring-boot-starter-test:3.5.6")

    def testSandboxWorkingDir = false
    def testParallelism = false
    def forkWorkingDir = moduleDir
  }
```

On first look it seems the test framework cannot be detected (or maybe the maven migration does not support it?)

changing the code to

```scala

  object test extends MavenTests with Junit4 {

    def mvnDeps =
      Seq(mvn"org.springframework.boot:spring-boot-starter-test:3.5.6")

    def testSandboxWorkingDir = false
    def testParallelism = false
    def forkWorkingDir = moduleDir

  }
```

does not crash, but the tests are not detected.

Working configuration needs junit platform version. The below works properly

```scala
object test extends MavenTests, Junit5 {

    override def junitPlatformVersion: T[String] = "6.0.0"

    def mvnDeps =
      Seq(mvn"org.springframework.boot:spring-boot-starter-test:3.5.6")

    def testSandboxWorkingDir = false
    def testParallelism = false
    def forkWorkingDir = moduleDir

  }
```

vaslabs:
## Setting up Spring boot with Kotlin

This one fails to run with:

```
./mill run
[59/59, 1 failed] ============================== run ==============================
1 tasks failed
[58] finalMainClass No main class specified or found
```

The reason for  the failure is due to the init not configuring the module as KotlinMavenModule.

Setting it though results in a different error

```
[68] Compiling 1 Kotlin sources to /home/vnicolaou/Downloads/demo-kotlin/out/compile.dest/classes ...
[...]

PR #6014, opened 2025-10-22
Add spring boot example for mill configuration compatible with spring  initializr for a maven kotlin project

vaslabs:
## Motivation

Related to https://github.com/com-lihaoyi/mill/issues/6006

Getting a "learning path" on how to setup spring boot projects in increasing complexity. The aim of the documentation is to follow the spring boot learning path and map it to mill.

## In this PR

This is the first small step of getting a kotlin project as downloaded from spring boot's [initializr](https://start.spring.io/) project to work with mill . While spring boot examples exist in mill, this differentiates by

- Project structure is the same as initializr
- Dependencies are up to date (and thus resulting in Junit5 failing without the platform version override) - I can also do these updates in the existing examples if you'd like
- Added an extra mill command to demonstrate how to build a jar and run the spring boot server
- Documentation scaffolding (see below)

### Documentation Scaffolding

This PR adds a special section for spring boot in both Java and Kotlin. The aim is to add more examples in increasing complexity as they appear [here](https://spring.io/projects/spring-boot)

image

The examples an in a separate example/springboot so it will be easier to the reader and to the tester (e.g. for directory grouping without affecting other javalib/kotlinlib examples, port isolation)

### Running the assembly jar

image

image

## Follow up PRs (plan)

- More examples
- More docs


lihaoyi:
Looks reasonable I think, some notes:

- Can we have a Java version of this example in addition to Kotlin? In general we should prioritize Java for the spring-boot stuff, since the vast majority of spring-boot users are on Java

- This should be wired up into the website on `website/docs/modules/ROOT/pages/kotlinlib/web-examples.adoc` 

- Since we're focusing on the Java spring-boot examples, I expect the to be numerous enough to deserve their own documentation page, which we can put at `website/docs/modules/ROOT/pages/javalib/spring-boot.adoc`

vaslabs:
> Looks reasonable I think, some notes:
> 
> * Can we have a Java version of this example in addition to Kotlin? In general we should prioritize Java for the spring-boot stuff, since the vast majority of spring-boot users are on Java
> * This should be wired up into the website on `website/docs/modules/ROOT/pages/kotlinlib/web-examples.adoc`
> * Since we're focusing on the Java spring-boot examples, I expect the to be numerous enough to deserve their own documentation page, which we can put at `website/docs/modules/ROOT/pages/javalib/spring-boot.adoc`

agreed, will add the java and start its own documentation page, I was planning to do so anyway in subsequent PRs, just wanted to get some early feedback

vaslabs:
created separated directories for both docs and code examples, added the doc sections and added some references in the docs.

the plan is to add to these examples and docs following the spring boot project structure in increasing difficulty (e.g. https://github.com/spring-projects/spring-authorization-server/tree/main/samples) (probably as we go we can exclude trivial examples, but that we can discuss later on case by case)

vaslabs:
@lihaoyi this should be ready as a first step, if you are happy, it can be merged and I'll continue adding examples and any potential mill extensions that maybe needed

lefou:
I think we should move the current Spring examples under "{language} Web Project Examples" into the new section (keep the section but refer to the other page with a link) to make the reading experience more consistent and not confuse the users.

vaslabs:
> I think we should move the current Spring examples under "{language} Web Project Examples" into the new section (keep the section but refer to the other page with a link) to make the reading experience more consistent and not confuse the users.

I'll follow up on subsequent PRs as this takes shape

PR #6017, opened 2025-10-23
Android: make `AndroidModule` a CrossModule

souvlakias:
This an experiment on making `AndroidModule` a CrossModule, in order to have `debug` and `release` variants. [Ref](https://github.com/com-lihaoyi/mill/pull/5957?notification_referrer_id=NT_kwDOBJuLtbQxOTQyNzE5NzU3Mzo3NzMwMjcwOQ#issuecomment-3418814797)
### Thoughts
- Build.mill is a bit more complex for the user this way
- Since this is just for experimentation, `androidIsDebug` remains and returns true/false according to `crossValue`

### Example:
image

### DIFF
--- a/example/androidlib/java/1-hello-world/build.mill
+++ b/example/androidlib/java/1-hello-world/build.mill
@@ -17,7 +17,7 @@ object androidSdkModule0 extends AndroidSdkModule {
 }
 
 // Actual android application
-object app extends AndroidAppModule {
+trait App extends AndroidAppModule {
   def androidSdkModule = mill.api.ModuleRef(androidSdkModule0)
   def androidMinSdk = 19
   def androidCompileSdk = 35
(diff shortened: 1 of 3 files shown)


souvlakias:
@lefou, does this implentation look good to you, or should we maybe go for the submodule approach instead?

lefou:
We should not weave the `CrossModule` nature into the `AndroidModule`, since it tightens the way how users can design their builds. 

As said before, it think it should feel more convenient for the user to model dev/prod builds via sub-modules, since dev-builds typically need less stuff to set up (e.g. no publishing, insecure signing, less optimizations) and a cross-module will require all machinery already set-up for each cross-value, while a sub-module can easily mix-in more traits only if needed. With sub-modules, it's also easier to have multiple prod-builds targeting different destinations or platforms.

For users who prefer to use a cross-module, we could provide dedicated traits, as we do with `ScalaCrossModule`.

PR #6018, opened 2025-10-23
Improve docs for single-file scripts

lihaoyi:
- Replace kotlinx-cli (which is unmaintained) with clikt
- Include Crawler example in Java and Kotlin `intro.adoc`
- Add links to the libraries used in the Crawler script
- Add some more text comparing {language} scripts with Bash
-


example/scalalib/basic/2-script-crawler/build.mill
```diff
@@ -32,3 +43,12 @@
 // These bundled libraries are only available in single-file scripts for convenience; in
 // config-based modules or programmatic modules, you would need to add them to `mvnDeps` explicitly.
 // You can also disable use of these libraries in scripts via `//| includDefaultScriptMvnDeps: false`.
+// And as we saw earlier, you can include any other libraries you with using `//| mvnDeps`.
```
lefou: you wish? 

```suggestion
// And as we saw earlier, you can include any other libraries you wish using `//| mvnDeps`.
```

Issue #3660, opened 2024-10-04
Make `out/` folder contents (more) reproducible and filesystem layout agnostic


lihaoyi:
The goal of this ticket is to make the `out/` folder contents more reproducible, such that it contains the same bytes and hashes regardless of the user's filesystem layout _outside_ of that folder. This is would allow re-using the `out/` folder as a build cache between different machines that may have the checkout in different place (e.g. `/Users/alice/my-repository` vs `/Users/charlie/my-repository`), both coarse grained (e.g. by sending over a zip file) and fine grained (via the bazel remote cache protocol)

The main thing that needs to happen is that every `os.Path` and `mill.api.PathRef` that is serialized within a "known" directory needs to be normalized to a path relative to an abstract reference to that known directory. e.g. 

* `/Users/alice/my-repository/out/foo/bar.dest/qux` should be serialized as `$WORKSPACE/out/foo/bar.dest/qux`
* `/Users/lihaoyi/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.13.14/scala-library-2.13.14.jar` should be serialized as `$COURSIER_CACHE/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/2.13.14/scala-library-2.13.14.jar`
* `/Users/alice/thing-outside-repository` should be serialized as `$HOME/thing-outside-repository`

AFAIK the necessary known roots should all be available globally (e.g. `mill.api.workspace.WorkspaceRoot.workspaceRoot`, `os.home`, `sys.env("COURSIER_CACHE")`). It should be easy enough to add to the serialization logic:

* `mill.api.PathRef` serialization https://github.com/com-lihaoyi/mill/blob/e0a2c93bfbc7bf68a71ddbc0c52afbb14e73e6f2/main/api/src/mill/api/PathRef.scala#L175-L197
* `os.Path` serialization https://github.com/com-lihaoyi/mill/blob/e0a2c93bfbc7bf68a71ddbc0c52afbb14e73e6f2/main/api/src/mill/api/JsonFormatters.scala#L27-L31

Apart from `PathRef` and `Path`, we will also need to deal with:

* Files in `out/` which are naturally non-deterministic: `mill-profile.json`, `mill-chrome-profile.json`, `mill-server/*` and `mill-no-server/*`, etc. 

[...]

PR #6021, opened 2025-10-24
[WIP] Map known root paths in cache files

lefou:
Another attempt to make Mill cache storage relocate-able.

The idea is to replace absolute paths with paths relative to pre-defined known root paths.

The current roots are:
* `$WORKSPACE` - the project directory (where the top-level `build.mill` can be found.
* `$MILL_OUT` - the Mill output directory used for caches and build-results
* `$HOME` - the user home directory

More paths come to mind:
* `$CACHE_COURSIER` - the coursier cache directory 
* `$CACHE_IVY` - the ivy cache directory
* `$CACHE_MAVEN` - the Maven cache directory

Those are currently not directly accessible in the code base. Also, their typical storage locations are already relative to `$HOME` (e.g. `$CACHE_COURSIER` vs `$HOME/.cache/coursier`), hence I left them out for now.

One design choice I made for now, is to go with a fixed set of known roots. This means, all of those must be configured before `PathRef`s can be used and de-/serialized. The alternative would be to dynamically register roots, allowing `PathRef` to be used without this restriction.

Some refactoring was necessary to streamline the configuration flow of the `$MILL_OUT` path. Before, the `out` path was constructed in many different locations, sometimes with slightly different logic. This is of course a minefield, since we have many entry points (Mill CLI, Mill Daemon, Mill BSP, Test Suites, Integration tests, ...) where those where set up. So this PR relies heavily on our good programmatic test coverage.

Attempt to fix https://github.com/com-lihaoyi/mill/issues/3660

Superseded by https://github.com/com-lihaoyi/mill/pull/6031, which maps the dynamically, which seems to be the better approach.

### DIFF
--- a/libs/javalib/api/src/mill/javalib/api/internal/JvmWorkerApi.scala
+++ b/libs/javalib/api/src/mill/javalib/api/internal/JvmWorkerApi.scala
@@ -48,8 +48,8 @@ trait JvmWorkerApi extends PublicJvmWorkerApi {
     compileJava(
       ZincCompileJava(
         upstreamCompileOutput = upstreamCompileOutput,
-        sources = sources,
-        compileClasspath = compileClasspath,
+        sources = sources.map(LocalPath.apply),
+        compileClasspath = compileClasspath.map(LocalPath.apply),
         javacOptions = jOpts.compiler,
         incrementalCompilation = incrementalCompilation
       ),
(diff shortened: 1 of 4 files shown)


lefou:
I don't know when and why we changed the output of `MainModule.show` from result as JSON to content of cache-json file, but it now means, `show` displays no longer absolute local paths but pseudo-paths with placeholders for `$WORKSPACE`,`$MILL_OUT` and `$HOME`.

lefou:
I currently hold the current `out` path in `PathRef.outPathOverride`, as there is no other accessible place where this information is available reliably. IMHO, we should expose it via `BuilCtx`, next to `BuildCtx.workspaceRoot`, since this imformation might be useful at various places and user builds. `Evavluator.currentEvaluator` also has a `outPath`, but it's often not available, and we don't need a full evaluator for just a user-configured path info.

lefou:
Incremental tests fail, but I have no clue why.

The incrementation zinc-compilation is working properly, but in case of a second compile run Mill should not run the compile task at all, due to caching. This is what's broken now, probably due to some hash changes, which I need to figure out. Most likely we somewhere mix hashes of the mapped and unmapped json data.

lefou:
Found the issue. Fixed it by implementing `PathRef.hashCode` based on the encoded path value.

lefou:
The RCP-server currently fails, due to the mapping not being optional. I started to prepare a fix introducing new types `LocalPath` and `LocalPathRef`, which might work but is rather invasive. Instead, I'd like to reassess the non-optional design in another attempt, first.

lefou:
Closing in favor of #6031

Issue #6039, opened 2025-10-28
Failing tests in `integration.manual[migrating]` module


ajaychandran:
The following tests are failing due to a mismatch in [tooling compatibility](https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html#tooling-compatibility-table):
- `MillInitSbtScalaLoggingTests`
- `MillInitSbtScoptTests`

The issue can be fixed by using ~~a Java home based on the SBT version used in the project~~ the system JDK to run the `sbt` shell.

`MillInitSbtScalaPBTests` is also failing. The underlying issue here is that a build file is generated for an invalid path.
The project uses `sbt-projectmatrix` plugin that is not supported by the conversion. It is preferable to fail fast in this case, instead of generating incorrect build files.

lefou:
@ajaychandran How can we tell the sbt importer to use a specific JVM ID?

ajaychandran:
> @ajaychandran How can we tell the sbt importer to use a specific JVM ID?

By passing the `-java-home` command line option to `sbt`.
For real world usage, the best option is using the system JDK. We already do this for the Gradle conversion where `gradle-jvm-id` option defaults to `system` and is overridden in tests.

PR #6046, opened 2025-10-29
Fixed manual integration tests for `./mill init` (#6039)

ajaychandran:
Added `--sbt-jvm-id` in `SbtBuildGenMain` to prevent errors due to JDK incompatibility.
Also, updated `MillInitTestSuite` and `MillInitSbtScalaPBTests` to handle `init` failure.

### DIFF
--- a/integration/manual/migrating/src/MillInitSbtScalaLoggingTests.scala
+++ b/integration/manual/migrating/src/MillInitSbtScalaLoggingTests.scala
@@ -5,9 +5,12 @@ import utest.*
 object MillInitSbtScalaLoggingTests extends MillInitTestSuite {
   def tests = Tests {
     test - checkImport(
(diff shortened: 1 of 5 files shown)


ajaychandran:
Note that post `init` failures occur in `MillInitSbtScalaLoggingTests` and `MillInitSbtScoptTests` due to a mismatch in JDK-Scala [compatibility](https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html#scala-compatibility-table).
Both projects use JDK 21, the Mill default, since no `jvmId` is configured at the module level.
```scala
// excerpt from MillInitSbtScalaLoggingTests
passingTasks = Seq("[2.11.12].compile"),
failingTasks = Seq("[2.12.15].compile", "[2.13.8].compile", "[3.1.2].compile")
```

PR #6050, opened 2025-10-30
Use relative script imports in `1-script/` examples

lihaoyi:
This should ensure that the imports work even if the scripts gets downloaded or copy-pasted somewhere else

### DIFF
--- a/example/javalib/basic/1-script/FooTest.java
+++ b/example/javalib/basic/1-script/FooTest.java
@@ -1,5 +1,5 @@
 //| extends: [mill.script.JavaModule.Junit4]
-//| moduleDeps: [Foo.java]
+//| moduleDeps: [./Foo.java]
 //| mvnDeps:
 //| - "com.google.guava:guava:33.3.0-jre"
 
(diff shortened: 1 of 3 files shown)


example/javalib/basic/1-script/FooTest.java
```diff
@@ -1,5 +1,5 @@
 //| extends: [mill.script.JavaModule.Junit4]
-//| moduleDeps: [Foo.java]
+//| moduleDeps: [./Foo.java]
```
lefou, line 2: Looks like we should be handle this automatically.

lihaoyi: The current design follows the Node.js model, where `./foo` is relative while `foo` is absolute. Not sure if that's the best way to do it, but after reviewing a bunch of import styles that was the one that looked best

lefou: Interesting. What exactly does absolute mean here? Since OS absolute path typically start with `/`.

lihaoyi: in our case absolute relative to the `workspaceRoot`, same in Node.js I think

lefou: That's an unusual pattern in the JVM ecosystem, IMHO, esp. in Scala, where even `import mill` means a potentially relative path, unless you write `import _root_.mill`.

PR #6053, opened 2025-10-30
Fix Scala Native link logs in server-client mode

lolgab:
The Scala Native toolchain uses its own threads to print using the logger. System.err on Scala Native threads doesn't have the Mill setup to forward the output to the client. So we store Console.err at the beginning of the task (in the Mill thread) which has the correct setup and use that reference to log. This way logs always reach the terminal correctly. Port for Scala Native of the following Scala.js related Pull Request: https://github.com/com-lihaoyi/mill/pull/5845


lefou (commented):
Looks good to me. Do you think we can add a test here too? 

We could just expect some tooling output in some of the existing  integration tests to make sure we capture the output.

lefou (approved):

lefou:
Thank you, Lorenzo!

PR #6061, opened 2025-10-30
Remove stale script module resolution code

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/6059

This code was hanging around from an earlier iteration of the config-based/script module design (https://github.com/com-lihaoyi/mill/pull/5836), where we loaded config-based modules dynamically, and caused the spurious resolution attempts and stack traces linked above. Since config-based modules are now done during codegen and not resolved dynamically, we can remove this extra resolution logic and simplify the code paths substantially

### DIFF
--- a/libs/script/src/ScriptModuleInit.scala
+++ b/libs/script/src/ScriptModuleInit.scala
@@ -151,7 +151,7 @@ object ScriptModuleInit
       scriptFileString: String,
       resolveModuleDep: String => Option[mill.Module]
   ) = {
-    val workspace = mill.api.BuildCtx.workspaceRoot
+    mill.api.BuildCtx.workspaceRoot
 
     mill.api.BuildCtx.withFilesystemCheckerDisabled {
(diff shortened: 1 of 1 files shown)


core/eval/src/mill/eval/EvaluatorImpl.scala
```diff
@@ -25,12 +25,7 @@ final class EvaluatorImpl private[mill] (
     private[mill] val allowPositionalCommandArgs: Boolean,
     private[mill] val selectiveExecution: Boolean = false,
     private val execution: Execution,
-    scriptModuleResolver: (
-        String,
-        String => Option[mill.Module],
-        Boolean,
-        Option[String]
-    ) => Seq[Result[mill.api.ExternalModule]]
+    scriptModuleResolver: (String, String => Option[Module]) => Seq[Result[ExternalModule]]
```
lefou, line 28: Could you opportunistically use a named tuple, since it make the code immediately more readable.

PR #6065, opened 2025-10-30
Allow users to disable some modules in BSP

alexarchambault:
This adds a `def enableBsp: Boolean` method on `JavaModule` (via `BspModuleApi`). Users can override this method and make it return `false` to make the BSP server hide / ignore that module.

Note that for consistency, if a module depends on another that has `enableBsp` set to false, it gets itself disabled too. Else, that might create situations where a module is loaded in BSP, but some of its dependencies would just be missing because they have `enableBsp` set to false.

### DIFF
--- a/runner/bsp/worker/src/mill/bsp/worker/BspEvaluators.scala
+++ b/runner/bsp/worker/src/mill/bsp/worker/BspEvaluators.scala
@@ -15,10 +15,6 @@ private[mill] class BspEvaluators(
     val watched: Seq[Watchable]
 ) {
 
-  private val transitiveDependencyModules0 = new ConcurrentHashMap[ModuleApi, Seq[ModuleApi]]
-  private val transitiveModulesEnableBsp0 =
-    new ConcurrentHashMap[ModuleApi, Option[Seq[BspModuleApi]]]
-
   /**
    * Compute all transitive modules from module children via moduleDirectChildren
    */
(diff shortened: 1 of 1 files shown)


alexarchambault:
This has been useful while trying to port the Scala 3 build to Mill: some modules are defined twice, once during "bootstrapping", and a second time using what was built during bootstrapping. Disabling the latter modules for BSP makes the BSP import phase faster: the first modules need to be built in order for the compiler class path of the second modules to be known, so getting BSP details about them requires compiling the first modules, which takes time. Better hide them by default.

runner/bsp/worker/src/mill/bsp/worker/BspEvaluators.scala
```diff
@@ -14,30 +15,74 @@ private[mill] class BspEvaluators(
     val watched: Seq[Watchable]
 ) {
 
+  private val transitiveDependencyModules0 = new ConcurrentHashMap[ModuleApi, Seq[ModuleApi]]
+  private val transitiveModulesEnableBsp0 =
+    new ConcurrentHashMap[ModuleApi, Option[Seq[BspModuleApi]]]
```
lihaoyi: @alexarchambault can you explain the changes here? I would have expected them to be a simple `.filter(_.enableBsp)`, not sure what the `transitiveDependencyModules` stuff is. If it's to resolve a different issue, it should be described in the PR description

core/api/daemon/src/mill/api/daemon/internal/bsp/BspModuleApi.scala
```diff
@@ -6,6 +6,14 @@ trait BspModuleApi extends ModuleApi {
   private[mill] def bspBuildTargetData: TaskApi[Option[(String, AnyRef)]]
   private[mill] def bspBuildTarget: BspBuildTarget
   private[mill] def bspDisplayName: String
+
+  /**
+   * Set this to false to make the Mill BSP server hide / ignore that module
+   *
+   * Beware that if a module depends via `moduleDeps` or `compileModuleDeps` on modules
+   * that have `enableBsp` set to false, the Mill BSP server will ignore that module too
```
lefou: the last `that` is to unspecific for me. It's not clear, what modules is also ignored.

lefou:
> This has been useful while trying to port the Scala 3 build to Mill:

Do you have a link to this port?

runner/bsp/worker/src/mill/bsp/worker/BspEvaluators.scala
alexarchambault: I added a word about that in the PR description. The goal is not to have "holes" in the module graph we expose via BSP.

core/api/daemon/src/mill/api/daemon/internal/bsp/BspModuleApi.scala
alexarchambault: Should be fixed in the last push

alexarchambault:
> Do you have a link to this port?

It's currently [here](https://github.com/alexarchambault/scala3/tree/mill-build). Beware that it's a WIP and that I might do force pushes on its branch. It also needs some other changes in Mill that I'm going to open PRs for.

PR #6068, opened 2025-10-31
Improve IDE support for scripts by scoping them per-file

lihaoyi:
Scoping them per-folder results in conflicts when there are multiple scripts in a single folder, whereas scoping them so the module root is the file itself seems to work better and allow the different scripts to each have their own classpath etc.

Screenshot 2025-10-31 at 10 34 17 AM

### DIFF
--- a/integration/ide/bsp-server-reload/resources/snapshots/broken/start/workspace-build-targets.json
+++ b/integration/ide/bsp-server-reload/resources/snapshots/broken/start/workspace-build-targets.json
@@ -5,7 +5,7 @@
         "uri": "file:///workspace/app/src/milltests/TheApp.scala"
       },
       "displayName": "app.src.milltests.TheApp.scala",
-      "baseDirectory": "file:///workspace/app/src/milltests",
+      "baseDirectory": "file:///workspace/app/src/milltests/TheApp.scala",
       "tags": [
         "library",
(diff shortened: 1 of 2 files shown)


unkarjedy:
@lihaoyi Could you also attach the screenshot of how it looks in the Project Structure?
image

lihaoyi:
@unkarjedy 

Screenshot 2025-11-01 at 2 16 29 PM
Screenshot 2025-11-01 at 2 16 27 PM
Screenshot 2025-11-01 at 2 16 26 PM
Screenshot 2025-11-01 at 2 16 24 PM
Screenshot 2025-11-01 at 2 16 23 PM
Screenshot 2025-11-01 at 2 16 21 PM

lihaoyi:
@unkarjedy here's a zip with an example project you can try once this publish job completes (https://github.com/com-lihaoyi/mill/actions/runs/18992462461). The file-level modules seems to work for java, scala, and kotlin scripts, with highlighting and navigation and such all working. 

[test.zip](https://github.com/user-attachments/files/23281110/test.zip)

Screenshot 2025-11-01 at 2 21 30 PM

Not sure if it's expected to work or not, but it does seem that it does. Perhaps you may have some more insights into why it seems to be working?

PR #6067, opened 2025-10-31
Revert "Run examples with local repo under out/ rather than ~/.ivy2/local (#5949)"

lihaoyi:
This reverts commit 2beb6c15f1c7ae5265db043f5d13bb87016330e7, as it seems to break `./mill dist.installLocalCache`

Presumably an unexpected interaction, but reverting it for now so that local testing and development isn't blocked

Seems like it causes IntelliJ BSP installation of `installLocalCache` output to fail

```
Executing /Users/lihaoyi/test/mill -i mill.bsp.BSP/install
Mill is running in BSP mode, using a separate output directory 'out/mill-bsp-out'. If you would like to reuse the regular `out/` directory, set the 'MILL_NO_SEPARATE_BSP_OUTPUT_DIR' environment variable. This will reduce the CPU usage of the BSP server but make it less responsive.
Downloading https://repo1.maven.org/maven2/com/lihaoyi/mill-runner-daemon_3/1.0.6-89-6adfd5-DIRTYc23995aa/mill-runner-daemon_3-1.0.6-89-6adfd5-DIRTYc23995aa.pom
Failed to download https://repo1.maven.org/maven2/com/lihaoyi/mill-runner-daemon_3/1.0.6-89-6adfd5-DIRTYc23995aa/mill-runner-daemon_3-1.0.6-89-6adfd5-DIRTYc23995aa.pom
Downloading https://repo1.maven.org/maven2/com/lihaoyi/mill-runner-daemon_3/1.0.6-89-6adfd5-DIRTYc23995aa/mill-runner-daemon_3-1.0.6-89-6adfd5-DIRTYc23995aa.pom.sha1
Failed to download https://repo1.maven.org/maven2/com/lihaoyi/mill-runner-daemon_3/1.0.6-89-6adfd5-DIRTYc23995aa/mill-runner-daemon_3-1.0.6-89-6adfd5-DIRTYc23995aa.pom.sha1
Exception in thread "main" java.lang.RuntimeException: coursier.error.ResolutionError$CantDownloadModule: Error downloading com.lihaoyi:mill-runner-daemon_3:1.0.6-89-6adfd5-DIRTYc23995aa
  not found: /Users/lihaoyi/.ivy2/local/com.lihaoyi/mill-runner-daemon_3/1.0.6-89-6adfd5-DIRTYc23995aa/ivys/ivy.xml
[...]

PR #6070, opened 2025-10-31
Run examples with local repo under out/ rather than ~/.ivy2/local

alexarchambault:
This changes how Mill is run in integration tests and examples. This makes it stop publishing things under `~/.ivy2/local`, and use something like `out/dist/localRepo.dest` instead, which is the `Task.dest` of the new `dist.localRepo` task. That new repository is a local Maven repositiory rather than an Ivy one, which has a slightly different file structure, the same as Maven Central.

For that new repository to be taken into account, the Mill launcher used in examples and ITs is now a script (bat under Windows, bash everywhere else) that sets the `COURSIER_REPOSITORIES` env var, so that the new local repo is added to it, and gets automatically used when running examples and ITs.

The second commit of this PR mainly updates the file names of the locally published Mill artifacts, because of the switch from an Ivy repository to a Maven one.

That's been useful to me when using a locally published Mill version, to develop on Mill or other projects. Running examples or ITs erases it, which is annoying, but also somewhat unexpected (running tests shouldn't erase things in the local repository, at least not things put there manually on purpose). This PR fixes that.

---

This adds back https://github.com/com-lihaoyi/mill/pull/5949 that was reverted in https://github.com/com-lihaoyi/mill/pull/6067, and changes `dist.installLocalCache`, so that it publishes things under `~/.ivy2/local` like `dist.installLocal` does (rather than using the local repo under `out/`, which is more likely to go away).

### DIFF
--- a/dist/package.mill
+++ b/dist/package.mill
@@ -78,25 +78,29 @@ trait DistModule extends Module {
    */
   def installLocal(binFile: String = localBinName) = {
     val binFile0 = os.Path(binFile, BuildCtx.workspaceRoot)
-    val task = installIvyLocalTask(binFile0)
-    Task.Command {
-      task()
-      PathRef(binFile0)
+    Task.Command[Unit] {
+      installIvyLocalTask(Task.Anon(binFile0))()
     }
   }
 
   val batExt = if (scala.util.Properties.isWin) ".bat" else ""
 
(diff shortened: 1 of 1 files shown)


lefou:
Can you please include the full change description of the original PR? (Otherwise git archeology is getting harder.)

PR #6071, opened 2025-10-31
Improve Spring docs

lefou:
* Add general notes about Spring and Mill
* Add `RepackageModule` docs for Scala and Kotlin with their specific examples
* Move all spring project examples to Spring landing pages


lefou:
The main change is ready, but I still try to figure out why some tests fail in CI.

Issue #6064, opened 2025-10-30
Android: Jetchat android tests are failing with non found resource


vaslabs:
## Reproduction
 
- Use steps in https://github.com/vaslabs-ltd/compose-samples-with-mill/actions/runs/18947793324/job/54103934717
- Error     

> Message 2) profileScreen_back_conversationScreen(com.example.compose.jetchat.NavigationTest)
> android.content.res.Resources$NotFoundException: String resource ID #0x7f0c002b
> Message android.content.res.Resources$NotFoundException: String resource ID #0x7f0c002b

> [1-291] Message 
> [1-291] FAILURES!!!
> [1-291] Message FAILURES!!!
> [1-291] Tests run: 10,  Failures: 2
> [1-291] Message Tests run: 10,  Failures: 2

My guess this is something we've missed from the android test apk -> base apk separation

vaslabs:
A workaround to make it work is to switch to the androidTest "fat" apk
```
    override def compileModuleDeps = Seq.empty
    override def moduleDeps = Seq(app)
```
which points that we are missing something when building against the R files.

Perhaps the packaging does not exclude the R files which later on get erased by R8 and creates the not found issue

Issue happens here

```
        composeTestRule.activity.getString(androidx.compose.ui.R.string.navigation_menu),
```

most likely a missing dependency from the android test

vaslabs:
it's not a missing dependency, the R is present in the compilation in the androidTestScope from the `androidx.compose.io.junit4/R.java`, that leaves R8 elimination

vaslabs:
the issue is fixed in #6073

vaslabs:
All compose sample tests are passing for 1.1.0-RC1
https://github.com/vaslabs-ltd/compose-samples-with-mill/actions/runs/19093693578

I'll do the rest of the examples, hopefully no other functionality is missing

PR #6073, opened 2025-10-31
Android: exclude the main apk deps from the test apk

vaslabs:
This is related to https://github.com/com-lihaoyi/mill/issues/6064

### Improvements

- Aligning base module and androidTest dependencies is [no longer necessary](https://github.com/vaslabs-ltd/compose-samples-with-mill/pull/6/files#diff-35fe6f53838971f482104d05f316647dead2c9fff13e5df4f7f2c7ad200e6a0dL95)
- Properly passing compile classpath to R8 makes the R8 workaround error -> warning mapping [not necessary](https://github.com/vaslabs-ltd/compose-samples-with-mill/pull/6/files#diff-35fe6f53838971f482104d05f316647dead2c9fff13e5df4f7f2c7ad200e6a0dL85)
- The test apk size is now greatly reduced (from >20MB to <1MB in most cases)
- With the new AndroidR8InstrumentedTestsModule , androidTests apk can also be minified as the `androidGeneratedMinifyKeepRules` is only available in R8 modules

| App | Mill Size | Gradle Size |
|------ |  ----------- | ------------------|
| Jetchat | 760KB | 1019KB | 
| JetNews|742KB|999KB|
| Architecture Samples | 4.3MB | 5.6MB |

### Fixes

- Cyclic dependency issue when compiling r classes
- Jetchat android tests

### Implementation details 

Introduced a new `AndroidR8InstrumentedTestsModule` which builds the test apk excluding the dependency tree of the base module (base apk) with 2 important tasks:

- `androidR8CompileOnlyClasspath` incorporates the whole base apk (but as a jar file)
- `androidResolvedPackagableMvnDeps` excludes the dependencies of the base module via `androidPackagableMvnDeps` and is used in place of `resolvedRunMvnDeps` . 

### Extensions

To do the dependency resolution again by excluding the dependency tree, without breaking the examples, I needed a way to pass the bom to the `defaultResolver().classpath` . Thus I extended the methods needed and fixed any MiMa issues with `@unroll` , although I haven't used it before, this is what I've found from looking around, so please let me know if this was a mistake 

### Further work

Testing a few apps, there was a case where a test dependency was pulling `com.google.android.material:material` which in turn depends on appcompat android resources, so aapt2 linking was failing. 

I tried to see if `aapt2 link` `-I` argument would be of any help, but it seems I'm missing something, it appears only one artifact can be passed to it, tried several usage variations without luck

Gradle incorporates these resources without an issue, so it means we need to find a way to link the resources against the base module resources and at the same time avoid duplicate contents

I think it's a very niche use case, as typically runtime usage of these resources will be done in the base apk. For now, I added the appcompat explicitly in the androitTest mvnDeps so the linking goes through

### DIFF
--- a/libs/androidlib/src/mill/androidlib/AndroidR8AppModule.scala
+++ b/libs/androidlib/src/mill/androidlib/AndroidR8AppModule.scala
@@ -310,10 +310,6 @@ trait AndroidR8AppModule extends AndroidAppModule { outer =>
 
       val compileOnlyClasspath = androidR8CompileOnlyClasspath()
 
-      /*
-       * FIXME on windows this will not work if it has too many files. Windows support will have to wait
-       * until we use R8 from its java API as the @file trick is not working as initially expected
-       */
       r8ArgsBuilder ++= compileOnlyClasspath.filter(_.path.ext == "jar").flatMap(compiledMvnDeps =>
         Seq(
           "--classpath",


vaslabs:
with the R8 classpath fix I'm very close to replicating the same test apk as gradle

image

lefou:
> This PR is far from functional, just posting if someone knows a few tips on the best way to exclude the dependency tree of the app module from the androidTest one

Each `JavaModule` has a `coursierProject` representing the internal coursier representation of the dependencies and constraints. Those get collected in `transitiveCoursierProjects` before resolution starts. Maybe, you can hook in there, to make the app module dep tree intransitive or something like that.

vaslabs:
> > This PR is far from functional, just posting if someone knows a few tips on the best way to exclude the dependency tree of the app module from the androidTest one
> 
> Each `JavaModule` has a `coursierProject` representing the internal coursier representation of the dependencies and constraints. Those get collected in `transitiveCoursierProjects` before resolution starts. Maybe, you can hook in there, to make the app module dep tree intransitive or something like that.

thanks, I have it somewhat functional now. The problem is this as the steps in the packaging flow goes (just thinking out loud):

1. The dependency tree is needed for compilation
2. when packaging the test apk, the dependencies that were used for the base apk need to be excluded. This is somewhat problematic to do on a jar by jar basis (I started with this approach) because the versions don't always match between what the test deps pull and what the base module pulls

This extends beyond the jars, anything like android resources, native libs, meta info files. So it seems on a first glance that this transitiveCoursierProjects can be useful, if the androidTest compileClasspath has the base apk deps so that we can compile the test code , but not package it. 

Thanks for the tip, now that I have 2 complex examples working, I can make this less hacky!

EDIT: Hm, I think after introducing the `androidR8Jar` I can use that in the android compile classpath and have no module dependency between the 2

vaslabs:
For the case of [jet-chat](https://github.com/vaslabs-ltd/compose-samples-with-mill/pull/6) I've even managed to exclude redundant things that are included in the gradle test apk

### Mill with this PR

image

### Gradle

image

vaslabs:
> EDIT: Hm, I think after introducing the androidR8Jar I can use that in the android compile classpath and have no module dependency between the 2

The problem is that I still have to compute the dependency exclusions, because the test deps might bring the same dependencies as in base module/apk and even worse , different versions.

vaslabs:
I've done some cleanup, avoiding duplication by overriding the resolvedRunMvnDeps.

I've tried a few other methods but it starts to get a bit complicated to play with coursier + manage the kotlin friend paths, I ended up having even more duplication.

I think with the latest commit, the implementation is fairly clean, but I'm open of course to further comments and suggestions to try .

In terms of technical things and results, this PR allows mill to produce tiny **test apks**.

Some result comparisons:

| App | Mill Size | Gradle Size |
|------ |  ----------- | ------------------|
| Jetchat | 760KB | 1019KB | 
| JetNews|742KB|999KB|
| Architecture Samples | 4.3MB | 5.6MB |


With the newest changes, there's a problem with androidtodo, so still not ready but will update accordingly

EDIT: I've added a small workaround for now for androidtodo and also kept backwards compatibility if the new AndroidR8InstrumentedTests is not extended


In addition to these size improvements, there's further improvement of setting up mill demonstrated [here](https://github.com/vaslabs-ltd/compose-samples-with-mill/pull/6/files#diff-35fe6f53838971f482104d05f316647dead2c9fff13e5df4f7f2c7ad200e6a0dL95) with a lot of duplicated dependencies not being needed in android test . Also the r8 error to warning mapping is [no longer needed](https://github.com/vaslabs-ltd/compose-samples-with-mill/pull/6/files#diff-35fe6f53838971f482104d05f316647dead2c9fff13e5df4f7f2c7ad200e6a0dL85)

If one wishes to reproduce these results, this branch should be built locally and then start an emulator and run `mill Jetchat.app.androidTest` .

vaslabs:
I need to figure out how to fix mima due to the added bom argument in a few methods, but apart from that , I'm open to review and comments

libs/androidlib/src/mill/androidlib/AndroidR8AppModule.scala
```diff
@@ -152,126 +162,173 @@ trait AndroidR8AppModule extends AndroidAppModule {
...
+          "--pg-conf",
+          extraRulesFile.toString
+        ) ++ androidCommonProguardFiles().flatMap(pgf => Seq("--pg-conf", pgf.path.toString))
+
+      r8ArgsBuilder ++= pgArgs
+
+      val compileOnlyClasspath = androidR8CompileOnlyClasspath()
+
+      /*
+       * FIXME on windows this will not work if it has too many files. Windows support will have to wait
```
vaslabs: This is fixed with the single jar file passed to it

PR #6083, opened 2025-11-01
Fixes for build functionality post auto-migration

ajaychandran:
Follow up to #5428 that adds several enhancements and fixes to enable build functionality.
- Set `mill-jvm-version` to `system`.
- Added support for converting BOM modules.
- Added mappings for `depManagement`, `sourcesFolders`, `sources`, `resources`, test `forkArgs`.
- Added base trait for test modules in generated base module.
- Added workaround for running Junit5 tests.
- Added mapping for Gradle dependency version range.
- Fixed Maven inter-test-module deps.
- Added support for non-cross-platform SBT modules that share the same base directory.

**Summary of code changes**
- `libs/init/package.mill`: Updated modules to target Java 11 and added mock tests.
- `libs.init.buildgen.api`: Added `ModuleSpec.Value` and `ModuleSpec.Values` to capture cross-version specific values and inter-module relationships. Merged `ModuleConfig` into `ModuleSpec`.
- `libs.init.buildgen`: Added `BuildGen` for generating meta-build types and writing build files using `ModuleSpec`.
- `libs.init.buildgen.test`: Updated `BuildGenChecker` to test implementations with a custom `JAVA_HOME`.
- `libs.init.gradle.exportplugin`: Updated `BuildModelBuilder` with enhancements/fixes.
- `libs.init.gradle`: Added workaround for non-published managed BOMs in `GradleBuildGenMain`.
- `libs.init.maven`: Added workaround to retain BOM dependencies and fixed configuring system properties in `Modeler`. Refactored `Plugins` and fixed `errorProneMvnDeps` derivation. Updated `MavenBuildGenMain` with enhancements/fixes.
- `libs.init.sbt.exportplugin`: Updated `ExportBuildPlugin` with enhancements/fixes.
- `libs.init.sbt`: Updated `SbtBuildGenMain`.


lihaoyi:
@ajaychandran if the heuristics to guess the java version from the compiler flags don't work well in practice, maybe we should just go with `mill-java-version: system` then. Trying to look at the user `JAVA_HOME` and expecting the user to set it to the right place is probably too error prone for new Mill users to do reliably

ajaychandran:
> @ajaychandran if the heuristics to guess the java version from the compiler flags don't work well in practice, maybe we should just go with `mill-java-version: system` then. Trying to look at the user `JAVA_HOME` and expecting the user to set it to the right place is probably too error prone for new Mill users to do reliably

We auto-detect the JVM version from the system JDK in `ModuleConfig.JavaHomeModule.system` and set a concrete version for `jvmId`. The user is not required to set a version anywhere.
Only the environment needs to be set up with the correct JAVA_HOME. This should already be true on a development system.

Using `mill-java-version: system` will also work. The loss of build reproducibility is acceptable I suppose.

lihaoyi:
Got it. If we want to auto-detect the JVM version, how about just shelling out to `java -version`? That should give us a version to use and would be simpler than the logic you currently have in `lazy val system`

ajaychandran:
On second thoughts, I will make the switch to the  `mill-java-version: system` option.
This will minimize changes to any CI workflows and is semantically more accurate.

ajaychandran:
@lihaoyi The PR description has been updated with the summary of code changes.

lihaoyi:
Bump on this @ajaychandran, can we get this PR ready to merge?

ajaychandran:

PR #6091, opened 2025-11-02
Skip parts of the `androidtodo` test suite to try and mitigate flakiness

lihaoyi:
This CI job isn't useful right now since it's constantly flaky, hopefully reducing the scope of the test will help stabilize it. CC @vaslabs


vaslabs:
It was also broken a few commits ago, but we're moving it into its own repo, so we'll be testing it against dev builds constantly. 

(Need to update with latest changes, my PR that was just merged fixes the Cyclic issue)

lihaoyi:
@vaslabs got it, in that case I'll just remove it for now

vaslabs:
Can just remove the action as the example is used in docs, and it's working fine locally after the latest changes

vaslabs:
I've pushed a change for this https://github.com/com-lihaoyi/mill/pull/6096 if you'd agree

PR #6097, opened 2025-11-02
Update spring boot example with bom mvn deps

vaslabs:
As the finding in [my comment](https://github.com/com-lihaoyi/mill/issues/6006#issuecomment-3477921825) I think the "idiomatic" way would be to set a bom mvn dep, to match the maven and gradle setups.

Updated the examples accordingly


lefou (approved):
Using Spring BOMs is the right thing.

Issue #6109, opened 2025-11-04
Mill seems to create a `~/.mill/cache/jline` directory on Windows


lefou:
Instead, it should use the `~/.cache/mill/jline` location.

Issue #5837, opened 2025-09-11
Change the naming structure of `mill.XXXXXX` temporary files


arturaz:
Now if there's leftover files like `mill.tsDMxN` in the filesystem it's pretty hard to remove all of them. `rm mill.*` removes `mill.bat` as well, which is unwanted.

I would suggest naming them `mill-temp.XXXXXX`, so you could easily `rm mill-temp*` them.

lefou:
Other options:

* use `.mill.XXXXXX`
* create them in the final download directory as `<final-name>.XXXXXX.part`

PR #6117, opened 2025-11-05
Cleanup bootstrap scripts

lihaoyi:
* Remove github latest-release downloader, since that's a hermeticity issue that can cause nondeterminism as new github versions are released, or even as time passes and you run the command again
* Remove `No mill version specified` warning, since now we recommend people to rely on the `MILL_DEFAULT_VERSION` in the docs 
* Remove check for system-installed Mill, which adds a bunch of complexity just to save a single download 
* Remove header comment that is already out of date and impossible to maintain
* DRY up the `sed` regexes in `mill.sh` and the `sed`-emulation code in `mill.bat`
* Re-arranged the two scripts and standardized variable names so the logic happens in the same order, so they are easier to read through side-by-side and compare
* Move mill temp download to `out/mill-bootstrap-download` to fix https://github.com/com-lihaoyi/mill/issues/5837
* Added a `dist.scripts.test` windows CI job to ensure the windows code paths are covered
* Tweaked a few windows cache paths to fix https://github.com/com-lihaoyi/mill/issues/6109

Covered by unit tests, and tested manually on my macbook pro and windows microsoft surface laptop 7

### DIFF
--- a/dist/scripts/src/mill.bat
+++ b/dist/scripts/src/mill.bat
@@ -93,6 +93,7 @@ rem without bat file extension, cmd doesn't seem to be able to run it
 
 set "MILL_NATIVE_SUFFIX=-native"
 set "MILL_JVM_SUFFIX=-jvm"
+set "MILL_FULL_VERSION=%MILL_VERSION%"
 set "MILL_DOWNLOAD_EXT=.bat"
 set "ARTIFACT_SUFFIX="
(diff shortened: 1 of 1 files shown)


lefou:
We should keep a small header in the script files indicating, where these scripts come from, including a link where it (or an update) can be downloaded from. Maybe, also an link to the documentation.

Issue #6047, opened 2025-10-29
Android Release Task/Module Implementation


souvlakias:
Creating this issue so we can track various attempts on implementing the release module for Mill's android support.
Current implementation is a hardcoded boolean task: [androidIsDebug](https://github.com/com-lihaoyi/mill/blob/main/libs/androidlib/src/mill/androidlib/AndroidModule.scala#L87).
Goal is to have a separate task or module for the release logic while also share tasks/resources between the two variants.

# 1. Cross Module [(branch)](https://github.com/com-lihaoyi/mill/compare/main...vaslabs-ltd:mill:android-cross-module-experiment1)

Have `AndroidCrossModule` / `AndroidAppCrossModule`.
Note: The linked branch has `AndroidModule` converted into a cross module but that wasn't the best choice.

### Pros

* This enables us to have two modules `Debug/Release`. Internal tasks could then determine the variant based on `crossValue`.

### Cons

* More complex api for the user

  ```scala
  trait App extends AndroidAppCrossModule {
    object test extends AndroidAppInstrumentedTests
  }
  object app extends Cross[App](AndroidBuildVariant.Debug, AndroidBuildVariant.Release)
  ```
* More complex cli for the user

  ```bash
  ./mill show app[debug].androidApk
  ./mill show app[debug].test
  ```
* Tests submodules will exist for both variants but release should not really have any.
* No shared resources between the variants

# 2. Shared Base Trait

Below are two attempts on having a base trait for the application.

Some common Pros/Cons are:

### Pros

* It requires no modification of `androidlib`'s modules.
* Tests Submodules can be defined for only the Debug variant

### Cons

* Separate resources for all tasks

## 2.1 2 Submodules [(branch)](https://github.com/com-lihaoyi/mill/compare/main...vaslabs-ltd:mill:android-release-module-exp)

Have a base trait and then instantiate 2 submodules that extend that:

```scala
trait base extends AndroidAppModule {
  override def moduleDir = super.moduleDir / os.up
  ??? // All implementation tasks
}

object app extends mill.api.Module {
  object debug extends base {
    object test extends AndroidAppTests, TestModule.Junit4 
  }
  object release extends base {
    override def androidIsDebug = false
    ??? // Other Release options
  }
}
```

### Notes

This feels like the cross module attempt but with slight differences.
- CLI would look like:
```bash
./mill show app.debug.androidApk
./mill show app.debug.test
```

[...]

PR #6126, opened 2025-11-05
Android: Variant and Release Submodules

souvlakias:
This implements https://github.com/com-lihaoyi/mill/issues/6047#issuecomment-3463884935.
### Changes
- AndroidModules now have inner Variant traits that share common tasks.
- AppModules also have Release traits which, for now, only override `androidIsDebug` to `false`.

Having a dedicated Variant for each module type *(Android, AndroidKotlin, AndroidR8, AndroidNative)* seems necessary to avoid missing configurations.

Users now need to create a `release` object that extends the corresponding `Release` trait, and then run: `./mill app.release.compile`, etc

### What's Next
Remove the `androidIsDebug` flag and instead let the `Release` traits override the relevant tasks directly.
Release configurations (keystore details) should be in the release variant too

### Concerns
This approach utilizes shared resources and may provide a better user experience with the APIs. However in my opinion, manually overriding tasks to reference outer implementations may be error-prone, both in terms of development and potential unexpected behavior if users override unshared tasks.

Maybe the shared base trait approach is something safer and more scalable, but with the cost of not sharing resources. Perhaps we could use the `outer` approach —just as an exception— to share some resource-heavy tasks.

### DIFF
--- a/example/androidlib/kotlin/1-hello-kotlin/build.mill
+++ b/example/androidlib/kotlin/1-hello-kotlin/build.mill
@@ -172,9 +172,6 @@ object app extends AndroidAppKotlinModule {
 
 */
 
-
-
-
 // To create a release apk and run it, you can use the configured release submodule.
 // Every AndroidAppModule can be configured with its own emulator, so you can easily
 // avoid signature mismatches between debug and release as follows.


example/androidlib/java/1-hello-world/build.mill
```diff
@@ -57,6 +46,18 @@ object app extends AndroidAppModule {
...
 
+  object release extends AndroidReleaseModule
+
+  /**
+   * Configuration for ReleaseKey
+   * WARNING: Replace these default values with secure and private credentials before using in production.
+   * Never use these defaults in a production environment as they are not secure.
+   * This is just for testing purposes.
+   */
+  def androidReleaseKeyName: Option[String] = Some("releaseKey.jks")
```
vaslabs: ~~I think with this change, we should remove the defaults and avoid the comment above~~

vaslabs:
@lefou what do you think about this approach? Looking at the [build variants](https://developer.android.com/build/build-variants) in general, I think the cross module might be a better fit and perhaps we give a trait to easily achieve the release targets.

Just thinking out loud: For the variants in general the users are free to create a cross module themselves, but we should have something for release imo, but in this form a lot of the weight falls on us maintaining it while perhaps a different cross module approach than our first attempt would be better, for example having a `AndroidCrossModule extends CrossModule[Android]("debug", "release")` and then an `AndroidReleaseModule extends AndroidCrossModule("release")` or something in those lines (I haven't tried it yet)

vaslabs:
Now that I got to play with spring boot under a similar concept, I'll give this another go

lefou:
From a quick glance, this approach looks good to me.

I'm not sure what you refer to with "shared base trait approach". In general, I think the variant-sub-modules should be fully capable android modules that can produce results on their own. The `outer`-references just provide the configuration (sources, resources, etc). This is, what we also do in other sub-traits, e.g. `ScalaModule` and `ScalaTests`.

vaslabs:
I added to the docs in kotlin 1-hello-kotlin example, and from the user experience perspective, it feels nice.

Let's go with this I think

vaslabs:
well, feels less nice when the build is more complex with hilt, R8 + kotlin

something like that is needed

```
object release extends AndroidHiltCompose, AndroidKotlinReleaseModule, AndroidR8ReleaseModule {
    def androidReleaseKeyName: Option[String] = Some("releaseKey.jks")

    def androidReleaseKeyAlias: T[Option[String]] = Task {
      Some("releaseKey")
    }

    def androidReleaseKeyPass: T[Option[String]] = Task {
      Some("MillBuildTool")
    }

    def androidReleaseKeyStorePass: T[Option[String]] = Task {
      Some("MillBuildTool")
    }

    override def androidEmulatorPort = "5654"
  }
  
trait AndroidHiltCompose extends AndroidHiltLib with AndroidCompose {

  override def mvnDeps: T[Seq[Dep]] = super.mvnDeps() ++ Seq(
      mvn"androidx.hilt:hilt-navigation-compose:1.2.0",
  )
}

trait AndroidCompose extends AndroidKotlinLib {
  override def androidEnableCompose: T[Boolean] = Task { true }
}
```

vaslabs:
this is a working release configuration with a complex setup

https://github.com/vaslabs/Pokedex_Compose_Multi_Module/pull/1/files

vaslabs:
on the other hand, cross module is fairly straight forward to be plugged in and I don't think we need to do anything there to support it (https://github.com/vaslabs/Pokedex_Compose_Multi_Module/pull/2)

apart from maybe some minor indirect UX improvements like making the emulator port a task so it can be derived from androidIsDebug

vaslabs:
the question is, this approach has a lot of complexity, do we take the hit? I think we don't do anything and close this PR, the cross module should be more than enough for the android variant to work out of the box by purely using Mill's cross module features

we can add some examples to aid any users and that's it imo

vaslabs:
Let's close this for the time being, we can revisit later. In the meantime we can add a cross module example

PR #6129, opened 2025-11-05
New `mill.api.opt` API for configuration data containing paths

lefou:
Add new API classes with the purpose to represent configuration data that may contain paths. All paths, even when embedded in options or text, are represented as `os.Path` type and can be therefore handled in a type-safe manner. In case, paths need to be to relocated, no guessing or vague `String.replace` logic is needed. For example, PR https://github.com/com-lihaoyi/mill/pull/6031 contains a path mapping mechanism, with which all `Opt`s would be automatically covered, too. That means if a `Opts` or it's companion types are used for `javacOptions`, `scalacOptions` and other configuration tasks, embedded paths are no longer an issue for relocatable Mill caches or remote caching in general.

In contrast to the initial proposal of this idea (https://github.com/com-lihaoyi/mill/discussions/6057) the main classes are named `Opt`/`Opts`, not `Arg`/`Args`, since we already have a `mill.api.Args` class used for other purposes.

## New API classes

`Opt` is a holder of configuration data. `Opt` can be constructed from one or multiple `String`s or `os.Path`s. When rendered as string, all parts will be concatenated.

`Opts` is a Seq-like container to be used instead of `Seq[String]`. 

`OptMap` is a type alias to `Map[String, Opt]` to provide some convenience to be used instead of `Map[String, String]`. 

Both container types, `Opts` and `OptMap` can be easily converted to a String-based container with `.toStringSeq` or `.toStringMap`. 

`OptGroup` is a group of `Opt`s that are tightly coupled, and should never be split. This is useful to model semantically bound CLI args with multi-arity like `--file <file>` or `-classpath jar2:jar2:...:jarN`. It can help to communicate intent and eases post-processing of multi-part opts.

## JSON serialization

The JSON structure is more complex than the previous `Seq[String]` equivalent, because it needs to support options, option groups and multi-parts options, which may itself consist of many strings and paths.

A rather complex `Opts` JSON serialization may look like this:

```scala
Opts(
  "-deprecated", 
  OptGroup("-classpath", Opt.mkPath(classpath, sep = ":"))
)
```

```json
[
  "-deprecated",
  [
    "-classpath",
    [
      { "path": "/path/to/jar1.jar" },
      ":",
      { "path": "/path/to/jar2.jar" },
      ":",
      { "path": "/path/to/jar3.jar" }
    ]
  ]
]
```

Since the JSON serialization for paths re-use the respective upickle-serializers, the root path mapping of PR https://github.com/com-lihaoyi/mill/pull/6031 may also work here (Resulting is paths like `$MILL_OUT/task/to/jar.dest/out.jar`).

## YAML convenience

In addition, the json reader supports reading of flat json arrays, which means, that all `Seq[String]` can easily be read as `Opts`. This was done to make working with `Opts` as convenient as `Seq[String]` not only in Scala code but also in `YAML` configuration code.

This means, `build.mill` YAML frontmatter and `build.mill.yaml` don't need any changes.

```yaml
mill-build:
  scalacOptions: ["-verbose"]
  forkEnv: { "MY_CUSOTM_ENV": "my-env-value" }
```

Although the full encoding is also supported but seldomly required.

```yaml
mill-build:
  forkArgs: [[["-javaagent:", {"path": "$WORKSPACE/agent.jar"}]]]
  forkEnv: { "MY_WORK_DIR": [{"path": "$HOME/workDir"}] }
```

## Realistic example

In POC PR https://github.com/com-lihaoyi/mill/pull/6800, you can see a refactoring of Mill to use the new ` mill.api.opt` API in all relevant tasks like `javacOptions`, `scalacOptions`, `forkArgs`, `forkEnv` and some more.

## Caveats / Feedback wanted

In this current POC, we need to add `import mill.api.opt.*`. I'm looking for ideas, how to improve this and make it a single import.

## Related issues and discussions 

* https://github.com/com-lihaoyi/mill/issues/3660 - Tracker issue
* https://github.com/com-lihaoyi/mill/pull/6031 - Stores relocatable paths
* https://github.com/com-lihaoyi/mill/discussions/6057 - Idea that drives this PR about how to store relocatable configuration options

Pull request: https://github.com/com-lihaoyi/mill/pull/6129

### DIFF
--- a/core/api/src/mill/api/opt/Opt.scala
+++ b/core/api/src/mill/api/opt/Opt.scala
@@ -81,8 +81,10 @@ object Opt {
           val elems = arr.value.map {
             case ujson.Str(opt) => opt
             case ujson.Obj(map) => upickle.read[os.Path](map("path"))
+            case _ => throw RuntimeException("Unsupported JSON data")
           }
           Opt(elems.toSeq*)
+        case _ => throw RuntimeException("Unsupported JSON data")
       }
     )
 


lefou:
Merged in latest changes from `main` branch, fixed conflicts and added a `OptMap` type alias to make working with maps containing config data more convenient.

lefou:
Going to move the binary breaking refactoring into it's own PR to get this API into Mill. 

I'm still interested into feedback, though.

lihaoyi:
Overall not really convinced that this is a viable alternative to the symlink+relative-path approach prototyped in the other PR. 

This PR adds special handling for option lists, and some limited data structures related to them, but it doesn't really handle all the other places where paths may be embedded in various data structures, text blobs, or `.dest` file contents. One way we can attempt to prove out this approach is to get the `integration/dedicated/reproducibility` tests passing using this approach. That would at least act as a smoketest that this API is sufficient to get the bare minimum reproducible builds working. 

This PR is also opt-in, v.s. the other approach which is on-by-default for OS-Lib via global configuration, which means it would be much easier for users to forget and get it wrong. And in general seems to be solving a much narrower slice of the problem that I'm not convinced is enough to get us to the goal of reproducible builds

PR #6132, opened 2025-11-06
Adding coverage validation at build time.

monksy:
This PR adds the ability for testing coverage minimums for your project, and gives you the ability to break the build when the test coverage falls below that.

It is reproducing the behavior granted by https://github.com/scoverage/sbt-scoverage guarenting coverage.

### DIFF
--- a/example/scalalib/linting/2-contrib-scoverage/build.mill
+++ b/example/scalalib/linting/2-contrib-scoverage/build.mill
@@ -65,4 +65,12 @@ Branch coverage....: 100.00%
 ...
 Statement coverage.: 10.53%
 Branch coverage....: 100.00%
+
+> sed -i 's/10.00/15.00/' build.mill
+
+> ./mill scoverage.validateCoverageMinimums
+error: Exception while validating coverage minimums. Coverage minimums violated:
+error: This project's statement coverage (10.53) did not meet the minimum desired by the project. (15.0)
+error: 1 tasks failed
+
 */


monksy:
Fixed the tests and the validations work.

monksy:
Run test /linux 11 scalalib local daemon is a test failure on HtmlScraper. That doesn't make sense since none of the code touched that. Is main broken?

example/scalalib/linting/2-contrib-scoverage/build.mill
```diff
@@ -14,6 +14,9 @@ object `package` extends ScoverageModule {
     mvn"com.lihaoyi::mainargs:0.7.7"
   )
 
+  def statementCoverageMin = Some(10.00)
```
lefou, line 17: Let's set this to 15, so we can show how a failure looks like.

example/scalalib/linting/2-contrib-scoverage/build.mill
```diff
@@ -50,6 +53,13 @@ scoverage.xmlCoberturaReport
 ...
 scoverage.xmlReport
 ...
+scoverage.validateCoverageMinimums
+
+> ./mill scoverage.validateCoverageMinimums
+...
+Coverage Minimums Met. Minimums set: {Statements: 10.00% Branches: 90.00%}
+Statement coverage.: 10.53%
+Branch coverage....: 100.00%
```
lefou, line 62: By prepending this with "error:", we can assert that the `validateCoverageMinimums` task indeed fails.

monksy:
@lihaoyi (Tagging you because I saw you're the original test creator)

The failure of this test is due to a change in the output from Wikipedia on Singapore, and that it no longer meets the requirements of the test.  (Or at least the amount of lines printed no longer does.. which fails the test). 

I'll create a PR about this and reference the fix. 

This test and these changes are not connected in any way.

lihaoyi:
@lefou do you have any other comments here? If not I'll probably merge it

example/scalalib/linting/2-contrib-scoverage/build.mill
monksy: Do we want a failure and success test or just a failure test? 

I'm not sure how we can do this without making a second copy of the test code.

lihaoyi: You can use `sed` to modify the files during the test. Grep for other examples to see how it's done

example/scalalib/linting/2-contrib-scoverage/build.mill
monksy: Looks like your change fixed this. But now we're seeing a bloken java android tests.

>  [1] selective.run 1 tasks failed
[8189] example.androidlib.java[1-hello-world].local.daemon.testForked 1 tests failed: 
  mill.testkit.UtestExampleTestSuite mill.testkit.UtestExampleTestSuite.exampleTest
> Error: Process completed with exit code 1.

I think we're good.. I think that's just a flaky test.

lihaoyi:
Thanks @monksy !

monksy:
Yay thank you for your help @lihaoyi and @lefou 

What fixed teh javalib android broken test?

lihaoyi:
I think it was just flaky, not sure

PR #6133, opened 2025-11-06
DO NOT MERGE - Test for a main build break

monksy:


monksy:
I noticed that scalalibs failed in a previous build and that most prs are not building lately. This is to kick off a test to see if main is broken or not.

PR #6134, opened 2025-11-06
`millBinPlatform` should be `1` for `1.1.0-RCx`

lolgab:
Only RCs and Ms of 1.0.0 should have the whole version as binary version
since minor versions are binary compatible with the previous minor.

Pull Request: https://github.com/com-lihaoyi/mill/pull/6134


lolgab:
The CI error seems unrelated

PR #6139, opened 2025-11-07
Fixes #6138: HtmlScraper is a flakey test. #6138

### LINKED ISSUE #6138: HtmlScraper is a flakey test.
HTMLScraper.scala is a flaky test as that it relies on a live connection to wikipedia and ordering of links. Links are subject to change. 

Ask: Save an version of the page in the resources and then reference that.

monksy:
The original code goes out to Wikipedia's singapore page. From there it would print out all of the links that it found and then verify the output. The way that HtmlScraper prints out this gets truncated and causes the test to fail. The contents on the page put the link of the Government Of Singapore beyond the amount that was printed on pprint. This also changes the source of what's being parsed to a local file. (Making this test more stable)


lihaoyi:
The truncation issue should be fixed in main

monksy:
Just out of curiousity.. where did that get fixed? 

https://github.com/com-lihaoyi/mill/pull/6137/files#diff-d17e0ced2050ef0957c5c5cf83069e4b2e2255b49835f0001601a0e99d288445 

I thought it was more of an amm print issue than a console write issue.

monksy:
Why not convert the test to use a local resource rather than a live html file? There are a lot of non-test things that could go wrong and violate the test.

Issue #6120, opened 2025-11-05
Android: add no metrics flag to emulator to avoid future cli block


vaslabs:
```
Consider using the '-metrics-collection' flag to help improve the       ##
##  emulator by sending anonymized usage data. Or use the '-no-metrics'     ##
##  flag to bypass this warning and turn off the metrics collection.        ##
##  In a future release this warning will turn into a one-time blocking     ##
##  prompt to ask for explicit user input regarding metrics collection
```

souvlakias:
Ref:
https://github.com/vaslabs-ltd/compose-samples-with-mill/actions/runs/19099646954/job/54568184072
<img width="969" height="347" alt="Image" src="https://github.com/user-attachments/assets/c58ec0fb-a68b-4d7c-b0a5-9eff3d9c23ff" />

PR #6144, opened 2025-11-07
Android: Add `-no-metrics` flag to emulator

souvlakias:
Resolves https://github.com/com-lihaoyi/mill/issues/6120
New versions of android emulator will block waiting for user input unless there is a `metrics` flag set.

### Changes
- Added `-no-metrics` to defaultArgs
- 
 ```diff
- startAndroidEmulator()
+ startAndroidEmulator(excludeDefaultArgs: Flag, extraArgs: String*)
```
Users can provide additional arguments to the `emulator` command.
There is an `excludeDefaultArgs` flag that when set, does not use the default arguments Seq.

### DIFF
--- a/website/docs/modules/ROOT/pages/android/android-initial-setup.adoc
+++ b/website/docs/modules/ROOT/pages/android/android-initial-setup.adoc
@@ -65,7 +65,7 @@ image::android/AndroidStudio.png[AndroidStudio.png]
 == 5. Run your application on an Android Virtual Device (AVD)
 
 
-=== Create an Android Virtual Device (AVD)s
+=== Create the AVD
 [source,console]
 ----
 > ./mill app.createAndroidVirtualDevice

### CODE REVIEW

File: libs/androidlib/src/mill/androidlib/AndroidAppModule.scala [hunk 1]
```diff
@@ -593,7 +593,8 @@ trait AndroidAppModule extends AndroidModule { outer =>
       androidSdkModule().emulatorExe().path.toString(),
       "-delay-adb",
       "-port",
-      androidEmulatorPort
+      androidEmulatorPort,
+      "-no-metrics"
```
vaslabs: since this is a command, and users may want to opt in for metrics, can we accept additional args from the command and append them here? if args are not provided, then we add `-no-metrics` or all of our default args (everything after `androidSdkModule().emulatorExe().path.toString()` .

alternatively we can have an opt in for emulator metrics task or an android emulator args task and put everything (everything after `androidSdkModule().emulatorExe().path.toString(),`)

vaslabs: so I think let's have a task androidEmulatorArgs and have everything that we pass to emulator exe there. If the user provides args to the command, we use those instead, if not we use the ones provided by the task.

This will give a way for advanced users to have more control on the emulator and also keep things working with default settings

lefou: How about adding two parameters, `extraArgs: Seq[String]` and `includeDefaultArgs: Flag`?

vaslabs:
could you show us a few examples on how it works now?

and maybe we should add something in the docs too, a use the android emulator via mill or something like that in the intro docs

souvlakias:
> could you show us a few examples on how it works now?
> 
> and maybe we should add something in the docs too, a use the android emulator via mill or something like that in the intro docs

Added a doc section about the avd and starting the emulator.
```
# Run with default arguments
> ./mill app.startAndroidEmulator

# Run with additional arguments
> ./mill app.startAndroidEmulator -noaudio

# Exclude default arguments
> ./mill app.startAndroidEmulator --excludeDefaultArgs -avd test_avd
```
The only *tricky* part I guess, is having to pass `--excludeDefaultArgs` before the extra args.
This is because the args are defined as `startAndroidEmulator(excludeDefaultArgs: Flag, extraArgs: String*)`. Not sure if there's another way to have `extraArgs` not as a `String*`

website/docs/modules/ROOT/pages/android/android-initial-setup.adoc
```diff
@@ -60,4 +60,45 @@ If you want to start with an example project, you can use the `init` command as
...
+# Run with default arguments
+> ./mill app.startAndroidEmulator
+
+# Run with additional arguments
+> ./mill app.startAndroidEmulator -noaudio
+
+# Exclude default arguments
+> ./mill app.startAndroidEmulator --excludeDefaultArgs -avd test_avd
+----
+
```
vaslabs, line 89: We should also mention (with a Tip?) that the additonal arguments are the emulator cli arguments that can be found [here](https://developer.android.com/studio/run/emulator-commandline)

Issue #6141, opened 2025-11-07
Android: AndroidBuildConfig issues


souvlakias:
While migrating [Jetcaster](https://github.com/android/compose-samples/tree/main/Jetcaster), `core.data` module does not compile
```
[1-129/130] Jetcaster.core.data.compile.super.kotlinlib.KotlinModule
[1-129] Compiling 26 Kotlin sources and reading 19 Java sources to /home/styl/projects/sample-apps/compose-samples-with-mill/out/Jetcaster/core/data/compile.super/kotlinlib/KotlinModule.dest/classes ...
[1-129] e: e: file:///home/styl/projects/sample-apps/compose-samples-with-mill/Jetcaster/core/data/src/main/java/com/example/jetcaster/core/data/di/DataDiModule.kt:22:40 Unresolved reference 'BuildConfig'.
```

This happens because the source file imports `import com.example.jetcaster.core.data.BuildConfig` and our generated BuildConfig does not declare a package name so the import fails.

EDIT:
Also noticed that our current implementation of BuildConfig extends AndroidAppModule. This shouldn't be the case as normal modules could have it as well.

PR #6145, opened 2025-11-07
Android: move AndroidBuildConfig to AndroidModule

souvlakias:
Resolves https://github.com/com-lihaoyi/mill/issues/6141
### Motivation
- BuildConfig was an additional trait to be extended only by AndroidAppModules, but [regular Modules](https://github.com/android/compose-samples/blob/main/Jetcaster/core/data/build.gradle.kts#L43) should have it as well
- Missing `package` field in the generated config
- There was no task to add custom fields

### Changes
- Moved BuildConfig generation to AndroidModule.
	- It is configurable with the `enableBuildConfig` flag, which defaults to true ([like gradle's](https://developer.android.com/reference/tools/gradle-api/7.4/com/android/build/api/dsl/BuildFeatures#buildConfig())).
- Added package name that defaults to `androidNamespace`
- Added a task for the config's members. (e.g. AndroidAppModule overrides this to add its `applicationID`

### DIFF
--- a/libs/androidlib/src/mill/androidlib/AndroidModule.scala
+++ b/libs/androidlib/src/mill/androidlib/AndroidModule.scala
@@ -594,7 +594,7 @@ trait AndroidModule extends JavaModule { outer =>
     Seq(
       s"boolean DEBUG = ${androidIsDebug()}",
       s"""String BUILD_TYPE = "$buildType"""",
-      s"""String LIBRARY_PACKAGE_NAME = "$androidNamespace""""
+      s"""String LIBRARY_PACKAGE_NAME = "$androidBuildInfoPackageName""""
     )
   }
 


libs/androidlib/src/mill/androidlib/AndroidBuildConfig.scala
souvlakias, line 1: Should this get marked as deprecated instead?

libs/androidlib/src/mill/androidlib/AndroidModule.scala
```diff
@@ -582,6 +571,56 @@ trait AndroidModule extends JavaModule { outer =>
...
+  def androidBuildInfoPackageName: String = androidNamespace
+
+  /**
+   * The members to include in the generated BuildConfig.java file.
+   * Format is "type NAME = value"
+   */
+  def androidBuildConfigMembers: T[Seq[String]] = Task {
+    val buildType = if (androidIsDebug()) "debug" else "release"
+    Seq(
+      s"boolean DEBUG = ${androidIsDebug()}",
```
souvlakias, line 595: Would having a `Map[String,Any]` and then parse it, be better?

libs/androidlib/src/mill/androidlib/AndroidBuildConfig.scala
vaslabs: it was experimental, so we can avoid duplications. But remember to update the examples

libs/androidlib/src/mill/androidlib/AndroidModule.scala
```diff
@@ -582,6 +571,56 @@ trait AndroidModule extends JavaModule { outer =>
...
+      "/"
+    )) / "BuildConfig.java"
+
+    os.write(destination, content, createFolders = true)
+
+    PathRef(destination)
+  }
+
+  override def generatedSources: T[Seq[PathRef]] = Task {
+    super.generatedSources() ++ Seq(generatedBuildConfig())
```
vaslabs: this [is configurable in AGP right](https://github.com/vaslabs-ltd/compose-samples-with-mill/blob/main/Jetcaster/mobile/build.gradle.kts#L87)? I mean creating a build config

libs/androidlib/src/mill/androidlib/AndroidModule.scala
```diff
@@ -582,6 +571,56 @@ trait AndroidModule extends JavaModule { outer =>
...
+   */
+  def generatedBuildConfig: T[PathRef] = Task {
+    val parsedMembers: Seq[String] = androidBuildConfigMembers().map { member =>
+      s"public static final $member;"
+    }
+    val content: String =
+      s"""
+         |package $androidBuildInfoPackageName;
+         |public final class BuildConfig {
+         |  ${parsedMembers.mkString("\n  ")}
```
vaslabs, line 613: ~~does it also need a `;`~~

libs/androidlib/src/mill/androidlib/AndroidModule.scala
vaslabs: I think it's ok for now, but `;` are missing from what I see

libs/androidlib/src/mill/androidlib/AndroidModule.scala
```diff
@@ -582,6 +571,56 @@ trait AndroidModule extends JavaModule { outer =>
...
+    )
+  }
+
+  /**
+   * Generates a BuildConfig.java file in the [[androidBuildInfoPackageName]] package
+   * This is a basic implementation of AGP's build config feature!
+   */
+  def generatedBuildConfig: T[PathRef] = Task {
+    val parsedMembers: Seq[String] = androidBuildConfigMembers().map { member =>
+      s"public static final $member;"
```
souvlakias, line 607: This always adds a `;` in case one is missing

vaslabs: ah thanks, missed it

souvlakias:
Some thoughts

Like in the case of `viewBinding` - in order to avoid having many traits - we [override](https://github.com/com-lihaoyi/mill/blob/main/libs/androidlib/src/mill/androidlib/AndroidKotlinModule.scala#L127) the generated sources based on a boolean flag. Since `viewBinding` is in AndroidKotlinModule there was no conflicts with `BuildConfig`, but if we were to support all [BuildFeatures](https://developer.android.com/reference/tools/gradle-api/7.4/com/android/build/api/dsl/BuildFeatures#summary) flags, things would get messy with this simple `override`.

I don't know if having a task like
```scala
override def generatedSources: T[Seq[PathRef]] = Task{
  super.generatedSources() ++
    Task.traverse(Seq(
        (androidGeneratedBuildConfigSources, enableBuildConfig),
		// More (task, boolean) tuples
      )) {
        case (task, enabled) =>
          if (enabled) task else Task.Anon(Seq.empty)
      }().flatten
}
```
(or something easier 😅) would be something that made sense.

vaslabs:
lgtm

vaslabs (approved):

souvlakias:
BuildConfigs for normal (non-app) modules have different fields. (e.g. `LIBRARY_PACKAGE_NAME` vs `APPLICATION_ID`)
image
Ref: https://android.googlesource.com/platform/tools/base/+/refs/heads/studio-master-dev/build-system/gradle-core/src/main/java/com/android/build/gradle/tasks/GenerateBuildConfig.kt#116

Issue #5943, opened 2025-10-01
Bumping Mill version doesn't trigger selective test execution


lihaoyi:
Noticed in https://github.com/com-lihaoyi/mill/pull/5940. We probably should conservatively trigger all tests when the Mill version changes just to be safe

We're already passing in the `classLoaderSigHash` and `javaHomeHash` when computing the `inputHash` of any task in `GroupExecution`, so that invalidates caches in the normal code path. But for selective test execution we only consider the _output_ hash of the `Taks.Input`s when deciding what to selectively execute, so changes to `classLoaderSigHash` or `javaHomeHash` get ignored

PR #6150, opened 2025-11-09
Make selective execution trigger when Mill version or JVM version is changed

lihaoyi:
We do this by including `classloaderSigHash`/`javaHomeHash` in Mill task value hash. Since selective execution triggers based on the value hashes of input tasks, this causes all input tasks to trigger when selective execution is used

Fixes https://github.com/com-lihaoyi/mill/issues/5943

### DIFF
--- a/core/exec/src/mill/exec/GroupExecution.scala
+++ b/core/exec/src/mill/exec/GroupExecution.scala
@@ -485,7 +485,7 @@ trait GroupExecution {
   }
 
   def getValueHash(v: Val, task: Task[?], inputsHash: Int): Int = {
-    (if (task.isInstanceOf[Task.Worker[?]]) inputsHash else v.##) + invalidateAllHashes
+    if (task.isInstanceOf[Task.Worker[?]]) inputsHash else v.## + invalidateAllHashes
   }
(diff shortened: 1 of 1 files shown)


core/exec/src/mill/exec/GroupExecution.scala
```diff
@@ -483,7 +485,7 @@ trait GroupExecution {
   }
 
   def getValueHash(v: Val, task: Task[?], inputsHash: Int): Int = {
-    if (task.isInstanceOf[Task.Worker[?]]) inputsHash else v.##
+    if (task.isInstanceOf[Task.Worker[?]]) inputsHash else v.## + invalidateAllHashes
```
lefou, line 488: This is hard to read. Does the `+ invalidateAllHashes` bind to both `if` branches?

lihaoyi: It only binds to the `else` branch. We don't want to bind it to the `then` branch because `inputsHash` already contains `invalidateAllHashes` when passed in, and including it twice breaks some assumptions elsewhere (haven't pinpointed exactly where)

Issue #6153, opened 2025-11-09
Potential logic error in `AndroidDataBindingsImpl`


lefou:
https://github.com/com-lihaoyi/mill/blob/5859dfa3098840b8a03a8563740bc3a7a0777441/libs/androidlib/databinding/impl/src/mill/androidlib/databinding/AndroidDataBindingImpl.scala#L84-L87

@vaslabs Could it be, this is meant to be

```diff
- os.Path(s"base/${asPath}.java").toIO 
+ os.Path(s"${base}/${asPath}.java").toIO 
```

?

Otherwise, the `base` class variable is unused.

vaslabs:
Yes , should be like that, probably works by accident anyway , will check once I'm on the pc

PR #6154, opened 2025-11-09
Fix logic error

lefou:
Discovered due an used variable warning.

Fix https://github.com/com-lihaoyi/mill/issues/6153


vaslabs:
The compose example that uses this is Jetchat, I'll give it a retest as well

EDIT: example is ok . possibly that code path is not executed in our example

PR #6160, opened 2025-11-10
Reorganize spring-boot docs

lihaoyi:
- Kill the kotlinlib/sprint-boot.adoc page. As discussed, the focus will be on Java examples
- Kill the scalalib/spring-boot.adoc page
- Inline `Spring_Header.adoc`

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -150,9 +150,6 @@ jobs:
           - java-version: 21
             millargs: "example.javalib.__.local.daemon"
 
-          - java-version: 21
-            millargs: "example.springboot.__.local.daemon"
-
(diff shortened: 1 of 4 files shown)


lefou:
Maybe, the Scala version isn't really appropriate (although in the spirit of the other duplication we have in the docs, e.g. for publishing and packaging.) But Spring is massively targeting Kotlin, as you can see in their framework docs, where they always give Kotlin examples next to Java examples. And since most Java and Kotlin examples in Mill already duplicate their Scala counterparts, I think we should continue this for the Spring examples for Kotlin. Especially, when examples get more evolved and use Java-based `@Configuration` classes, which look slightly different in Kotlin.

PR #6161, opened 2025-11-10
Fixed compilation error in BomModule example

ajaychandran:
The fix resolves the following compile error:
```console
[build.mill-59] [error] 6 │object myBom extends BomModule {
[build.mill-59] [error]   │       ^
[build.mill-59] [error]   │object creation impossible, since def compile: mill.api.Task.Simple[mill.javalib.api.CompilationResult] in trait SemanticDbJavaModule in package mill.javalib is not defined 
[build.mill-59] [error]   │(The class implements abstract override def compile:
[build.mill-59] [error]   │  mill.api.Task.Simple[mill.javalib.api.CompilationResult] in trait BomModule in package mill.javalib but that definition still needs an implementation)
```

### DIFF
--- a/example/fundamentals/dependencies/2-managed/build.mill
+++ b/example/fundamentals/dependencies/2-managed/build.mill
@@ -55,7 +55,7 @@ object baz extends JavaModule {
 
 // One can manage and publish Bill of Material (BOM) modules from Mill, by using `BomModule`:
 
-object myBom extends BomModule {
+object myBom extends JavaModule, BomModule {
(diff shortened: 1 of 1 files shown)


lefou:
I wonder why we didn't saw this in CI? Is this project not running by default?

lihaoyi:
@lefou I think it's because those examples don't have `/** Usage */` blocks so no commands are run at all to validate the code

lefou:
We probably should always implicitly run a `mill resolve __` to ensure the example build itself is valid.

Issue #6166, opened 2025-11-11
Spring boot: Example with spring-boot-configuration-processor


vaslabs:
I think it's a good example to have in mill due to [the annotation processor](https://docs.spring.io/spring-boot/specification/configuration-metadata/annotation-processor.html)

vaslabs:
This works OOTB but with a warning if we just place the dependencies in

```
[36] [info] compiling 2 Java sources to /home/vnicolaou/mill/example/javalib/springboot/5-configuration-processor/out/compile.dest/classes ...
[36] [info] Annotation processing is enabled because one or more processors were found
[36] [info]   on the class path. A future release of javac may disable annotation processing
[36] [info]   unless at least one processor is specified by name (-processor), or a search
[36] [info]   path is specified (--processor-path, --processor-module-path), or annotation
[36] [info]   processing is enabled explicitly (-proc:only, -proc:full).
[36] [info]   Use -Xlint:-options to suppress this message.
[36] [info]   Use -proc:none to disable annotation processing.
[36] [info] done compiling
[36/36] ============================== compile ============================== 4s
```

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/67b7a8c1-5777-47a2-a676-ce97934450de" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/33927023-9a2f-4620-8799-c446e9090ae7" />

vaslabs:
@lihaoyi @lefou would it be a good idea to add an annotations processor task in Java modules to make this easier?  This will make the setup for spring boot, micronaut and the likes easier than doing [this](https://github.com/com-lihaoyi/mill/blob/main/example/javalib/web/3-todo-micronaut/build.mill#L55) ceremony, which is a bit error prone if the setup involves a bom.

I think we can have a task like annotationProcessorMvnDeps that does what the examples do + the bom which is available since [this](https://github.com/com-lihaoyi/mill/pull/6073/files#diff-4054783182757e4c35f50757780d8f470e5a5bef3c697e3dffda7b8cea98d13dR254) and feeds the javacOptions task

If we go with the direction of having a new task, we'd of course need to deal with backwards compatibility, so perhaps some flag or a trait that mixes in these new tasks and javac options

lefou:
[...]

PR #6168, opened 2025-11-12
JavaModule: Annotation Processor API

vaslabs:
## Motivation

Shift the responsibility of setting up annotation processors from the user (via javacOptions) to mill (via a new annotationProcessor API).

Related to: https://github.com/com-lihaoyi/mill/issues/6166

## Changes

This is initially motivated by an upcoming example with spring-boot-configuration-processor where a warning suggested that in a future update, auto class discovery will not work without explicitly enabling processors via flags or the processor path.

Since [we've introduced](https://github.com/com-lihaoyi/mill/pull/6073/files#diff-4054783182757e4c35f50757780d8f470e5a5bef3c697e3dffda7b8cea98d13dR254) the bom option to the default resolver we can avoid misconfigurations from the users' behalf, by accepting a list of mvn deps for the processors, resolving the dependencies (by also) using boms and thus making framework/platform setups such as spring and micronaut easier.

In this spirit, I've also updated the micronaut examples to show the effect of these changes. There's a mill module (`ErrorProneModule`) that can also benefit from this, but I've not touched it. If you'd like I can update that as well to use the new API

## Improvement summary

I think now there's a 1:1 mapping from gradle/maven to mill when migrating spring or micronaut framework examples, by substituting the "plugin" or "platform" dependency such as

```
val micronautVersion: String by settings // from gradle.properties
dependencyResolutionManagement {
    repositories {
        mavenCentral()
    }
    versionCatalogs {
        create("mn") {
            from("io.micronaut.platform:micronaut-platform:${micronautVersion}")
        }
    }
}
```

with a 

```
def bomMvnDeps = Seq(mvn"io.micronaut.platform:micronaut-platform:${micronautVersion}")
```


vaslabs:
ready for review @lefou @lihaoyi 

working on that, I've noticed that the javac options passed in [KotlinModule](https://github.com/com-lihaoyi/mill/blob/main/libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala#L330) are not the same as in the [JavaModule](https://github.com/com-lihaoyi/mill/blob/main/libs/javalib/src/mill/javalib/JavaModule.scala#L863) 

Should they be in sync?

EDIT: Also, wondering if the `annotationProcessorJavacOptions` should go in `mandatoryJavacOptions` but my first guess was no as from the javadoc I think this is meant to be used by other modules?

lihaoyi:
I think this looks good to me

lihaoyi:
The `JavaModule` javac options are something to do with how annotation processors output their generated sources. I think we didn't do the same for `KotlinModule` because people don't use javac annotation processors much with kotlin

Issue #6188, opened 2025-11-15
Spring boot: ahead of time compilation support


vaslabs:
As discussed in https://github.com/com-lihaoyi/mill/issues/6006#issuecomment-3532870384

We can create a build step (probably not unlike the Hilt support for android as a concept) to support the optimisation build step as documented in https://docs.spring.io/spring-framework/reference/core/aot.html

Also should document the process and any tips for knowing it's working or any common troubleshooting steps. 

Support can probably be included in SpringBootTools so it can be mixed in with other spring boot support.

PR #6208, opened 2025-11-18
Spring boot: AOT and Native support

vaslabs:
# Spring Boot AOT and Native support

For https://github.com/com-lihaoyi/mill/issues/6188

### Main

- springBootProcessAOT: Uses [SpringApplicationAotProcessor](https://github.com/spring-projects/spring-boot/blob/main/core/spring-boot/src/main/java/org/springframework/boot/SpringApplicationAotProcessor.java) thanks to the pointers provided by @re-thc at [this discussion](https://github.com/com-lihaoyi/mill/issues/6006#issuecomment-3532870384)
- Native image creation with the native properties generated by the AOT Processor
- Native example with AOT demonstration for both JVM aot enabled and native runs

### Things in support of the scope of this PR

- Added nativeRunBackground (same implementation as the `PythonModule`) so that `mill clean nativeRunBackground` works.
- SpringBootModule which speeds up configuring spring boot applications (with spring boot version and auto configuring the bomMvnDeps)

### Running the JVM AOT optimised version

image

### Running the Non AOT optimised version with AOT enabled flag on

image

### Running with prod.run (JVM + AOT)

image

### Native run

Covered in the example, but native always looks for AOT otherwise:

image

### DIFF
--- a/core/api/src/mill/api/Cross.scala
+++ b/core/api/src/mill/api/Cross.scala
@@ -248,6 +248,11 @@ trait Cross[M <: Cross.Module[?]](factories: Cross.Factory[M]*) extends mill.api
   def get(args: Seq[Any]): M = valuesToModules(args.toList)
 
   /**
+   * Fetch the cross module corresponding to the given cross value strings
+   */
+  def lookup(args: String*): M = segmentsToModules(args.toList)
+
+  /**
    * Fetch the cross module corresponding to the given cross values
    */
   def apply(arg0: Any, args: Any*): M = valuesToModules(arg0 :: args.toList)
--- a/core/api/src/mill/api/internal/Resolved.scala
+++ b/core/api/src/mill/api/internal/Resolved.scala
@@ -9,10 +9,12 @@ private[mill] sealed trait Resolved {
    * What segments selector was used to resolve this root module. Might not be
    * `rootModule.moduleSegments` as it might be an alias that we want to preserve
    */
-  def rootModuleSegments: Segments
+  def rootModulePrefix: String
   def taskSegments: Segments
   def cls: Class[?]
-  def fullSegments: Segments = rootModuleSegments ++ taskSegments
+  def fullSegments: Segments =
+    (if (rootModulePrefix.isEmpty) Segments() else Segments.labels(rootModulePrefix)) ++
+      taskSegments
 }
 
 private[mill] object Resolved {
@@ -33,19 +35,19 @@ private[mill] object Resolved {
 
   case class Module(
       rootModule: RootModule0,
-      rootModuleSegments: Segments,
+      rootModulePrefix: String,
       taskSegments: Segments,
       cls: Class[?]
   ) extends Resolved
   case class Command(
       rootModule: RootModule0,
-      rootModuleSegments: Segments,
+      rootModulePrefix: String,
       taskSegments: Segments,
       cls: Class[?]
   ) extends Resolved
   case class NamedTask(
       rootModule: RootModule0,
-      rootModuleSegments: Segments,
+      rootModulePrefix: String,
       taskSegments: Segments,
       cls: Class[?]
   ) extends Resolved
--- a/core/eval/src/mill/eval/EvaluatorImpl.scala
+++ b/core/eval/src/mill/eval/EvaluatorImpl.scala
@@ -137,7 +137,7 @@ final class EvaluatorImpl(
               case _ => (Nil, k)
             }
 
-            val (None, Some(rest)) = ParseArgs.extractSegments(taskSel).get
+            val ("", rest) = ParseArgs.extractSegments(taskSel).get
 
             Option.when(module.moduleSegments == Segments(prefix ++ rest.value.dropRight(1))) {
               rest.last.value
--- a/core/internal/src/mill/internal/LineBufferingOutputStream.scala
+++ b/core/internal/src/mill/internal/LineBufferingOutputStream.scala
@@ -52,7 +52,9 @@ class LineBufferingOutputStream(onLineComplete: ByteArrayOutputStream => Unit)
   }
 
   override def flush(): Unit = synchronized {
-    writeOutBuffer()
+    // We ignore flushes, because we already flush on completed lines, and flushing
+    // incomplete lines can result in all sorts of confusing logs being printed
+    // writeOutBuffer()
   }
 
   override def close(): Unit = {
--- a/core/internal/test/src/mill/internal/LineBufferingOutputStreamTests.scala
+++ b/core/internal/test/src/mill/internal/LineBufferingOutputStreamTests.scala
@@ -11,7 +11,7 @@ object LineBufferingOutputStreamTests extends TestSuite {
       val lpos =
         new LineBufferingOutputStream(s => { baos.write("PREFIX".getBytes()); s.writeTo(baos) })
       for (b <- "hello\nworld\n!".getBytes()) lpos.write(b)
-      lpos.flush()
+      lpos.close()
       assert(baos.toString == "PREFIXhello\nPREFIXworld\nPREFIX!")
     }
 
@@ -30,7 +30,7 @@ object LineBufferingOutputStreamTests extends TestSuite {
         new LineBufferingOutputStream(s => { baos.write("PREFIX".getBytes()); s.writeTo(baos) })
       val arr = "hello\nworld\n!".getBytes()
       lpos.write(arr)
-      lpos.flush()
+      lpos.close()
 
       assert(baos.toString == "PREFIXhello\nPREFIXworld\nPREFIX!")
     }
--- a/core/resolve/src/mill/resolve/ParseArgs.scala
+++ b/core/resolve/src/mill/resolve/ParseArgs.scala
@@ -9,7 +9,7 @@ import scala.annotation.tailrec
 
 object ParseArgs {
 
-  type TasksWithParams = (Seq[(Option[Segments], Option[Segments])], Seq[String])
+  type TasksWithParams = (Seq[(String, Segments)], Seq[String])
 
   /** Separator used in multiSelect-mode to separate tasks from their args. */
   val MultiArgsSeparator = "--"
@@ -81,13 +81,13 @@ object ParseArgs {
   }
 
   def extractSegments(selectorString: String)
-      : Result[(Option[Segments], Option[Segments])] =
+      : Result[(String, Segments)] =
     parse(selectorString, selector(using _)) match {
       case f: Parsed.Failure => Result.Failure(s"Parsing exception ${f.msg}")
       case Parsed.Success(selector, _) => Result.Success(selector)
     }
 
-  private def selector[_p: P]: P[(Option[Segments], Option[Segments])] = {
+  private def selector[_p: P]: P[(String, Segments)] = {
     def wildcard = P("__" | "_")
     def label = P(CharsWhileIn("a-zA-Z0-9_\\-")).!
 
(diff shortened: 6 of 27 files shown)


lefou:
I don't think we want to have this in `RepackageModule`, which can be used for any kind of JVM module and does not require Spring otherwise. You might want to create a new module, and let it inherit `RepackageModule`, since it should provide the `repackagedJar` task as well.

vaslabs:
Created a spring boot module, but I think I'll split the work into multiple parts: 

1. to introduce the spring boot module as a helper for spring boot apps (minimise the code needed even further)
2. introduce the AOT for java runtime
3. support for testing
4. Graal builds

Also, I'm not entirely convinced on my approach of a child module for an optimised build, I've kind of followed one of the approaches we tried for android release

vaslabs:
> Also, I'm not entirely convinced on my approach of a child module for an optimised build, I've kind of followed one of the approaches we tried for android release

hm, probably not entirely wrong, the usage I see with maven and gradle is:

```
mvn -Pnative native:compile -Daot.profiles=prod
```

or

```
./gradlew nativeCompile -PaotProfiles=prod
```

so I think a mill equivalent with an interface like

```
./mill prod.compile
```

would be ok

vaslabs:
Updated with native example, still needs work but please let me know if the design direction makes sense

re-thc:
> although probably not worth it as this optimisation development flow wise shouldn't happen constantly, maybe just once?

Depends if you want / need to ensure it works for AOT mode. So things don't perform 1:1 with it on. So you may run it on each restart of the dev loop.

lefou:
In-process workers should in general start faster and should save more memory. So, if it's easy to have them (e.g. the entry point doesn't use problematic globals or `System.exit` calls)  and we already have the setup in the existing spring worker, we should run the tools in-process.

vaslabs:
> In-process workers should in general start faster and should save more memory. So, if it's easy to have them (e.g. the entry point doesn't use problematic globals or `System.exit` calls) and we already have the setup in the existing spring worker, we should run the tools in-process.

what I was thinking is, the classpath for the worker would be the the run classpath of the module, that would change on every source and dependency change, would that be ok anyway?

EDIT: ah just saw your other comment, should be fine then

libs/javalib/spring-boot-worker/src/mill/javalib/spring/boot/worker/impl/SpringBootToolsImpl.scala
```diff
@@ -63,6 +56,33 @@ class SpringBootToolsImpl() extends SpringBootTools {
...
+        resourceOut.toString,
+        classOut.toString,
+        groupId,
+        artifactId
+      ) ++ applicationArgs
+    )
+    process.stdout.buffered.lines().forEach(println)
+    process.stderr.buffered.lines().forEach(System.err.println(_))
+    process.waitFor()
+  }
```
lefou: The point of the `SpringBootTools` module is, that it can be shared and typically only depend on the version task. The logic implemented here isn't using anyting from this module, so it could also reside in `SpringBootModule` directly.

The preferred solution would be to run in-process and use the module's classloader, which is already shared.

lefou:
> > In-process workers should in general start faster and should save more memory. So, if it's easy to have them (e.g. the entry point doesn't use problematic globals or `System.exit` calls) and we already have the setup in the existing spring worker, we should run the tools in-process.
> 
> what I was thinking is, the classpath for the worker would be the the run classpath of the module, that would change on every source and dependency change, would that be ok anyway?

If we use a new classloader with the application classes, but have the toolchain come from the parent classloader (the one of the worker), then all spring tooling stuff should still be shared. Maybe, I'm overlooking something?

vaslabs:
> > > In-process workers should in general start faster and should save more memory. So, if it's easy to have them (e.g. the entry point doesn't use problematic globals or `System.exit` calls) and we already have the setup in the existing spring worker, we should run the tools in-process.
> > 
> > 
> > what I was thinking is, the classpath for the worker would be the the run classpath of the module, that would change on every source and dependency change, would that be ok anyway?
> 
> If we use a new classloader with the application classes, but have the toolchain come from the parent classloader (the one of the worker), then all spring tooling stuff should still be shared. Maybe, I'm overlooking something?

I think so, I'm not a 100% comfortable with how things behave when switching classloaders, from seeing around I'm assuming this is what should be done? The withClassLoader takes care of closing the old one etc if I understand correctly

```java
def springBootProcessAOT: T[PathRef] = Task {
    //....

    val args: Array[String] = ???

    mill.util.Jvm.withClassLoader(
      runClasspath().map(_.path)
    ) {
      classloader =>
        RunModule.getMainMethod(
          "org.springframework.boot.SpringApplicationAotProcessor",
          classloader
        )
          .invoke(null, args)
    }

    //....
  }
  
```

lefou:
> The withClassLoader takes care of closing the old one etc if I understand correctly

I think so. But you should make the parent classloader explicit. Since the default is `null` and also the `sharedLoader` defaults to `getClass.getClassloader`, which is probably the classloader that used to load `Jvm`, which is not the one that contains any spring tools. So, in effect, you load all spring tools every time.

You probably want something like this:

```scala
    mill.util.Jvm.withClassLoader(
      classpath = runClasspath().map(_.path),
      parent = this.getClass().getClassLoader() // assuming, `this` is the spring boot tools worker
    ) {
      classloader =>
```

Edit: This comment assumes, you implement the AOT logic in `SpringBootTools` module, but your latest changes instead implement it in the module directly, which means, there is no spring tooling API available.

libs/javalib/spring-boot-worker/src/mill/javalib/spring/boot/worker/impl/SpringBootToolsImpl.scala
lefou: I saw you removed this task. This was not my intention. I think it should stay here but use any spring tooling from the module classpath directly, in the way  as the `repackagedJar` task does.

vaslabs:
> > The withClassLoader takes care of closing the old one etc if I understand correctly
> 
> I think so. But you should make the parent classloader explicit. Since the default is `null` and also the `sharedLoader` defaults to `getClass.getClassloader`, which is probably the classloader that used to load `Jvm`, which is not the one that contains any spring tools. So, in effect, you load all spring tools every time.
> 
> You probably want something like this:
> 
> ```scala
>     mill.util.Jvm.withClassLoader(
>       classpath = runClasspath().map(_.path),
>       parent = this.getClass().getClassLoader() // assuming, `this` is the spring boot tools worker
>     ) {
>       classloader =>
> ```
> 
> Edit: This comment assumes, you implement the AOT logic in `SpringBootTools` module, but your latest changes instead implement it in the module directly, which means, there is no spring tooling API available.

Thank you!

I moved away from the spring tools, because everything needed is in the run class path  but I'll adjust your suggestion

lefou:
> I moved away from the spring tools, because everything needed is in the run class path but I'll adjust your suggestion

You mean the class `SpringApplicationAotProcessor` is already in the `runClasspath`? I have overlooked this detail. In that case, there is indeed not much value in any classloader caching. To the contrary, any issue in the app could negatively influence Mill's runtime behavior. Comparable to running `testLocal` instead of `testForked`.

vaslabs:
> > I moved away from the spring tools, because everything needed is in the run class path but I'll adjust your suggestion
> 
> You mean the class `SpringApplicationAotProcessor` is already in the `runClasspath`? 

Yes indeed!


> I have overlooked this detail. In that case, there is indeed not much value in any classloader caching. To the contrary, any issue in the app could negatively influence Mill's runtime behavior. Comparable to running `testLocal` instead of `testForked`.

I followed the runLocal

```java
def runLocalTask(mainClass: Task[String], args: Task[Args] = Task.Anon(Args())): Task[Unit] =
    Task.Anon {
      Jvm.withClassLoader(
        classPath = runClasspath().map(_.path).toVector
      ) { classloader =>
        RunModule.getMainMethod(mainClass(), classloader).invoke(null, args().value.toArray)
      }
    }
```

I'll check the testLocal

vaslabs:
> > > I moved away from the spring tools, because everything needed is in the run class path but I'll adjust your suggestion
> > 
> > 
> > You mean the class `SpringApplicationAotProcessor` is already in the `runClasspath`?
> 
> Yes indeed!
> 
> > I have overlooked this detail. In that case, there is indeed not much value in any classloader caching. To the contrary, any issue in the app could negatively influence Mill's runtime behavior. Comparable to running `testLocal` instead of `testForked`.
> 
> I followed the runLocal
> 
> ```java
> def runLocalTask(mainClass: Task[String], args: Task[Args] = Task.Anon(Args())): Task[Unit] =
>     Task.Anon {
>       Jvm.withClassLoader(
>         classPath = runClasspath().map(_.path).toVector
>       ) { classloader =>
>         RunModule.getMainMethod(mainClass(), classloader).invoke(null, args().value.toArray)
>       }
>     }
> ```
> 
> I'll check the testLocal

Ok, they look similar and I think I get it, also saw the warning on runLocal. Should I revert back to jvm spawnable like testForked ?

lefou:
> Ok, they look similar and I think I get it, also saw the warning on runLocal. Should I revert back to jvm spawnable like testForked ?

Probably yes. This ensures the run classpath isn't tainted by anything not part of the application and not anticipated by the spring tooling. Sorry for the trouble, but I think we learned something on the way.

vaslabs:
> > Ok, they look similar and I think I get it, also saw the warning on runLocal. Should I revert back to jvm spawnable like testForked ?
> 
> Probably yes. This ensures the run classpath isn't tainted by anything not part of the application and not anticipated by the spring tooling. Sorry for the trouble, but I think we learned something on the way.

No worries at all, playing with the classloaders is always fun :D 

I'll finalise everything tomorrow

libs/javalib/src/mill/javalib/spring/boot/SpringBootModule.scala
```diff
@@ -0,0 +1,156 @@
...
+
+import mill.{T, Task}
+import mill.api.{ModuleRef, PathRef}
+import mill.javalib.{Dep, DepSyntax, MavenModule, NativeImageModule, RunModule}
+
+import java.util.Properties
+import scala.util.Using
+
+@mill.api.experimental
+trait SpringBootModule extends MavenModule {
```
lefou: Can you add some docs, esp. why this needs to be a `MavenModule`?

vaslabs: I will, just to run something by you first, this is based (e.g. structure) on the spring boot starter

(https://start.spring.io/)

Should we name the module SpringBootStarterModule instead?

lefou: I mean, is it really necessary to use the Maven layout? If not, we should make it optional by not extending `MavenModule`.

It's IMHO ok when showcasing the starter projects, to explicitly add `MavenModule`.

vaslabs: it's only necessary to be compatible with the default examples of spring boot, in my head this new module would just be a helper to get started quickly without doing anything else, worrying about the structure, what maven module is and so on, just get the code from start.spring.io, extend one trait and be good to go

vaslabs: we could cover both, by having a SpringBootModule that extends JavaModule and a SpringBootStarterModule that extends SpringBootModule, MavenModule

lefou: I still don't get it, what the `SpringBootStarterModule` would provide over the `SpringBootModule`. Having concise examples is nice, but we should not compromize the overall structure and usefulness of the various modules we provide.

By using `SpringBootModule` and `MavenModule`, users learn early how we do things in Mill. It is probably less complicated then to learn what the difference between `SpringBootModule` and `SpringBootStarterModule` could be.

vaslabs:
I think I cleaned up most of the rough edges. I'm sure there will be more, but it works and we can smooth anything we find as we go along as all this functionality is under `@mill.api.experimental`

libs/javalib/src/mill/javalib/spring/boot/SpringBootModule.scala
vaslabs: agreed & done

vaslabs:
Failures seem due to flakiness + network errors, jobs are green on my copy

lefou (approved):

PR #6219, opened 2025-11-19
Android: Collect all output files from r8

souvlakias:
### Motivation
While migrating Jetcaster compose sample it came up that we don't collect all the output files from R8 but only `.dex` ones.
For example: 
image
In this case the `rome.properties` file was necessary for a module to work.

### Changes
- Collect more than just `.dex` files from R8, with an `excludePackageFiles` task (similar to [Gradle's](https://developer.android.com/reference/tools/gradle-api/8.3/null/com/android/build/api/dsl/ResourcesPackaging#getExcludes()))


libs/androidlib/src/mill/androidlib/AndroidAppModule.scala
```diff
@@ -234,6 +234,13 @@ trait AndroidAppModule extends AndroidModule { outer =>
     )
   }
 
+  /**
+   * File names to be excluded from the APK package.
+   */
+  def excludePackageResources: T[Seq[String]] = Task {
+    Seq("configuration.txt", "seeds.txt", "mapping.txt")
```
souvlakias: TODO: os-lib glob matching?

Gradle has this `exclude` functionality too, but instead takes as arguments glob strings and performs glob matching, so it's easier for the user to specify exclude rules.

vaslabs: yeah that would be ideal

vaslabs: should also prefix with android

vaslabs:
Can you update the description with some screenshots + comparisons with previous

libs/androidlib/src/mill/androidlib/AndroidAppModule.scala
```diff
@@ -234,6 +235,13 @@ trait AndroidAppModule extends AndroidModule { outer =>
     )
   }
 
+  /**
+   * Regex patterns of files to be excluded from packaging into the APK.
+   */
+  def androidExcludePackageResources: T[Seq[Regex]] = Task {
```
vaslabs: Resources may hint to "resources" , let's name it ...PackagedFiles

vaslabs:
also tested locally on a multi-module android app and looks good to me, you can mark it as ready @souvlakias if you're done

vaslabs:
I'm still testing this with some configurations

vaslabs:
Lgtm

PR #6230, opened 2025-11-20
Put back global Mill install instructions in the docs

lihaoyi:
This follows the style used by the Ammonite documentation, which seems to work well enough in practice across both Mac and Linux

CC @lefou

### DIFF
--- a/website/docs/modules/ROOT/pages/cli/installation-ide.adoc
+++ b/website/docs/modules/ROOT/pages/cli/installation-ide.adoc
@@ -11,7 +11,6 @@ a zip file containing the full example ready to use. These examples come with a 
 script you can use to immediately begin working with the project, automatically downloading and
 caching a JVM and any additional dependencies as necessary.
 
-[#_bootstrap_scripts]
 == Bootstrap Scripts
 
 The Mill example projects in this documentation come with `./mill` and `./mill.bat`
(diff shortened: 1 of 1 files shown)


website/docs/modules/ROOT/pages/cli/installation-ide.adoc
```diff
@@ -71,6 +70,21 @@ folder with one of the Mill example projects. There are a wide range of example
...
+> sudo sh -c '(curl -L {mill-download-url}/mill-dist-{mill-version}-mill.sh -o mill) > /usr/local/bin/mill && chmod +x /usr/local/bin/mill'
+```
+
+This can be useful when using Mill to run script files (written in xref:javalib/script.adoc[Java],
+xref:scalalib/script.adoc[Scala], or xref:kotlinlib/script.adoc[Kotlin]) which may not belong
+to any particular project. However, for projects that use Mill, it is recommended you use the
+xref:#_bootstrap_scripts[per-project bootstrap scripts] discussed above to ensure that anyone
+with the project codebase on disk will be able to build the project without needing a separate
+installation
+
```
lefou, line 87: There is also some middle ground, which requires no root permissions: Install in the user specific `bin` directory, which is typically `~/bin` or `.local/bin`. I think this is by far the most common way to install developer tools on any machine.

lihaoyi: Those don't work on MacOS though, without additional fiddling to set the path in your `.bashrc` and `.zshrc`. It seems `/usr/local/bin` works across both Mac and Linux

lefou: Fair point. (Probably, because Mac's aren't really used in multi-user setups.) In the end, those instructions are just ideas. Any proficient developer should know, what to do with a shell script and where to place it. The only information she needs is, "Will the script work from a `bin` directory properly?", and that question is answered here.

lefou (approved):

Issue #6232, opened 2025-11-21
Android: Regression on some android resources


vaslabs:
Found a discrepancy while testing previously working apps

<img width="421" height="178" alt="Image" src="https://github.com/user-attachments/assets/7745b120-7437-4991-a3ab-06098b16bdd7" />

The back icon in this pokemon multimodule app is a check box, it should be a left arrow.

Setting ` override def androidNonTransitiveRClass = false` makes it work as expected.

There are 2 scenarios I think:
1. When I setup the app, I didn't put the dependencies exactly as in the original project , I avoided duplicating the dependencies (including module deps) from each module. 
2. There's an issue in the way we resolve resources with the direct resources

cc @souvlakias 

I think point 1 is more likely, as is what non transitive does, will test and update the ticket

vaslabs:
indeed tested and this configuration works

https://github.com/vaslabs/Pokedex_Compose_Multi_Module/commit/aa954e7f8fca85146dffd90be9a79792040b1485

which is kind of proof non transitive is working , sorry for the bother with the issue

vaslabs:
I'm reopening this because upon investigation, I've figured the namespacing feature is missing from non transitive R class

vaslabs:
basically every module, should be able to reference the R classes of itself and with the transitive R classes on, it loses the references when it's pulled into the main app

vaslabs:
I think to fix this issue we need to emit ids from each module and feed it to what depends on it

```
--emit-ids path	Emits a file at the given path with a list of names of resource types and their ID mappings. This is suitable to use with --stable-ids.
--stable-ids outputfilename.ext	Consumes the file generated with --emit-ids containing the list of names of resource types and their assigned IDs.
This option allows assigned IDs to remain stable even when you delete or add new resources while linking.
```

Not entirely sure though, it's just suspicious that the text and icons are mixed up


EDIT: these are not needed, but they may help to highlight the issue

vaslabs:
[this](https://android.googlesource.com/platform/tools/base/+/refs/heads/mirror-goog-studio-main/build-system/gradle-core/src/main/java/com/android/build/gradle/internal/res/Aapt2ProcessResourcesRunnable.kt?autodive=0%2F%2F%2F%2F%2F%2F%2F%2F%2F%2F%2F%2F#119) seems relevant

vaslabs:
[...]

PR #6233, opened 2025-11-21
Android: Revert androidNonTransitiveRClasses

vaslabs:
I've decided to revert the whole androidNonTransitiveRClasses because it was the wrong implementation after all (see linked issue). 

To achieve the same optimisation as in gradle we need to do more work with Symbol based R generation that is not immediately available from the tools we have.

```
# Enables namespacing of each library's R class so that its R class includes only the
# resources declared in the library itself and none from the library's dependencies,
# thereby reducing the size of the R class for that library
android.nonTransitiveRClass=true
```

Related to: https://github.com/com-lihaoyi/mill/issues/6232


libs/androidlib/src/mill/androidlib/AndroidModule.scala
```suggestion
    Task.traverse(transitiveModuleRunModuleDeps) {
```

PR #6242, opened 2025-11-25
Split plugin/user-facing code to Scala3.7/Java11 to allow rest of codebase to upgrade to Scala3.8/Java17

lihaoyi:
This PR splits the Mill codebase into Java11/Scala3.7.3 and Java17/Scala3.8.0-RC1  portions

Scala3.7/Java11 portion:

* Classes visible to user plugins, mostly the transitive dependencies of `mill-libs`, as we cannot upgrade that to Scala 3.8 without breaking plugin compatibility as Scala 3.7 -> 3.8 is not forwards-compatible

* The classes needed by the code running co-located with user modules in the same process, primarily the `javalib.testrunner` code and `javalib.worker` code

Scala3.8/Java17 portion:

* `mill.daemon` and other non-user-facing processes/classes

* User `build.mill` files

This lets us continue supporting user code with `def jvmId = "11"` while requiring `mill-jvm-version: 17` for Mill itself, which is a similar requirement that [Maven](https://maven.apache.org/whatsnewinmaven4.html#required-java-version) and [Gradle](https://docs.gradle.org/current/userguide/compatibility.html) impose on their users. It also allows us to preserve compatibility with existing plugins compiled against the Scala Java11/3.7 version of Mill, as upgrading to Scala 3.8 is a bin-compat breaking change

Mostly uneventful, though the codesig logic needed to be updated to handle Scala 3.8+ lazy vals which are encoded differently than 3.7-, and some modules that use `export`s need to stay on Scala 3.7.3 due to bincompat 

Covered by updating the `example/depth/javahome/1-custom-jvms/build.mill` test to set `def jvmId = "11"` and ensure things work

### DIFF
--- a/libs/util/package.mill
+++ b/libs/util/package.mill
@@ -72,7 +72,11 @@ object `package` extends MillStableScalaModule with BuildInfo {
       "Scala 2.13 version used by some workers."
     ),
     BuildInfo.Value("millVersion", build.millVersion(), "Mill version."),
-    BuildInfo.Value("scalaVersion", Deps.scalaVersion, "Scala version used to compile build.mill files"),
+    BuildInfo.Value(
+      "scalaVersion",
+      Deps.scalaVersion,
+      "Scala version used to compile build.mill files"
+    ),
     BuildInfo.Value("millBinPlatform", build.millBinPlatform(), "Mill binary platform version.")
   )
 }


lefou:
Does that mean, Mill no longer runs on a Java 11 VM?

lihaoyi:
Yes, although since https://github.com/com-lihaoyi/mill/pull/6005 Mill already bootstraps its own JVM unless you explicitly opt out, so users shouldn't need to do any manual installation of the specific JVM version. 

The main user-facing change from this PR is that you would have to use `def jvmId` if you want your user code to run on Java 11, rather than relying on `mill-jvm-version` setting it globally for both Mill and for user code.

lihaoyi:
https://github.com/com-lihaoyi/mill/pull/6248 adds a nice error message so people who relied on `mill-jvm-version: 11` before will be told to instead use `def jvmId = "11"` to configure their modules

Issue #6246, opened 2025-11-25
mill.javalib.Dependency: Breaks when there are no dependency versions (i.e. when having BOMs)


vaslabs:
I've found this during experimentation with updating deps for android but as we are expanding the spring boot + micronaut examples to Bom based dependency configuration this is potentially going to become a problem for users.

```
 Task.Anon@2e70f260(mill.javalib.dependency.versions.VersionsFinder.classpath) 
Resolution failed for 30 modules:
--------------------------------------------
  androidx.compose.ui:ui-tooling-preview: 
        not found: /home/vnicolaou/.ivy2/local/androidx.compose.ui/ui-tooling-preview
        not found: https://repo1.maven.org/maven2/androidx/compose/ui/ui-tooling-preview/maven-metadata.xml
        No version available in (,)
```

I haven't looked at the underlying implementation much but my first inclination would be to ignore any dependencies that don't have a version and take into account bomMvnDeps as well for reporting updates.

So the end result should potentially be bomUpdate -> NewVersion, mvnDepWithVersion -> NewVersion

PR #6247, opened 2025-11-25
Fix dependency update finder for bom and diverse resolution params

vaslabs:
Fix for https://github.com/com-lihaoyi/mill/issues/6246


vaslabs:
# Test

For spring, bom and empty versions fix for Android  bom dependency and resolution params.

## Spring boot

Screenshot From 2025-11-25 18-49-58


## Android

image

Issue #6237, opened 2025-11-23
Android: Rework android resources


vaslabs:
## Problem description

While testing [Jetcaster](https://github.com/vaslabs-ltd/compose-samples-with-mill/pull/11) I've found a few issues on how we manage resources when we have a fairly complex multi-module android project.

1. The module deps are built with final Resource ids which due to the inlining upon compilation may not reflect to the app modules' final resource ids resulting in jumbled resource references with either showing wrong icons or crashes. 
2. When we compile and link the resources of the AndroidApp, the R.jar package from the submodule is effectively in the main app's package.

<img width="1801" height="384" alt="Image" src="https://github.com/user-attachments/assets/4fadd4e2-4662-413d-a996-260ad9c55c62" />

## Current workaround

There's no easy workaround but in theory the 

```scala
  def extraPackages = Task {
    Seq(
      core.data,
      core.designsystem,
      core.domain,
      glancewidget,
      core.`domain-testing`
    ).flatMap(lib => Seq("--extra-packages", lib.androidNamespace))
  }

  override def androidAaptOptions =
    super.androidAaptOptions().filterNot(_ == "--non-final-ids") ++ extraPackages()
```

should have worked. Unfortunately building the R.jar only takes into account 1 R.java file

```scala
  def androidLibsRClasses: T[Seq[PathRef]] = Task {
    // TODO do this better
    // So we have application R.java class generated by aapt2 link, which includes IDs of app resources + libs resources.
    // But we also need to have R.java classes for libraries. The process below is quite hacky and inefficient, because:
    // * it will generate R.java for the library even library has no resources declared
    // * R.java will have not only resource ID from this library, but from other libraries as well. They should be stripped.
    val rClassDir = androidLinkedResources().path / "generatedSources/java"
    val mainRClassPath = os.walk(rClassDir)
      .find(_.last == "R.java")
      .get
```

This makes the workaround non-intuitive with an additional override being needed:

```scala
  override def androidLibsRClasses = Task {
    (super.androidLibsRClasses()
      ++ os.walk(androidLinkedResources().path / "generatedSources").filter(
      os.isFile
    ).filter(_.ext == "java").map(PathRef(_))).distinct
  }
```

and this makes the Jetcaster workable

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/7dd35db8-dbea-4c8f-91cf-a1dbb6d4f355" />

## Proposal

I think we need to rework a bit the dependencies so that:

[...]

PR #6249, opened 2025-11-26
Android: Fix resource linking on multimodule projects

souvlakias:
Addresses #6237
 
This is a temporary fix for the resource linking until we properly pass submodules' resources to other modules.
 
### Changes
- Set the `--non-final-ids` flag in aapt2 link call for non-app modules
- Pass  recursiveModuleDeps' namespaces as `--extra-packages` to aapt
- Pickup all java files from `androidLinkedResources` instead of only R.java
- Set `kotlincUseBtApi` to false in `AndroidHiltSupport` ([Ref](https://github.com/com-lihaoyi/mill/issues/6355#issuecomment-3636550819))

### DIFF
--- a/libs/androidlib/src/mill/androidlib/hilt/AndroidHiltSupport.scala
+++ b/libs/androidlib/src/mill/androidlib/hilt/AndroidHiltSupport.scala
@@ -24,6 +24,11 @@ import mill.{T, Task}
 @mill.api.experimental
 trait AndroidHiltSupport extends KspModule, AndroidKotlinModule {
 
+  // Dagger does not work with the bt api
+  override def kotlincUseBtApi: T[Boolean] = Task {
+    false
+  }
+
   override def kspProcessorOptions: T[Map[String, String]] = Task {
     super.kspProcessorOptions() ++ Map(
       "dagger.fastInit" -> "enabled",


libs/androidlib/src/mill/androidlib/AndroidModule.scala
```diff
@@ -131,15 +131,23 @@ trait AndroidModule extends JavaModule { outer =>
...
-      )
-    } else {
-      Seq("--auto-add-overlay")
-    }
+    val debugOptions = Seq(
+      "--proguard-minimal-keep-rules",
+      "--debug-mode"
+    )
+
+    // Add module dependencies' namespaces as extra packages
```
vaslabs: We should add a TODO here to clean up, once we rewire the submodule sources to go to the app module

libs/androidlib/src/mill/androidlib/AndroidModule.scala
```diff
@@ -498,7 +506,8 @@ trait AndroidModule extends JavaModule { outer =>
     // * it will generate R.java for the library even library has no resources declared
     // * R.java will have not only resource ID from this library, but from other libraries as well. They should be stripped.
     val rClassDir = androidLinkedResources().path / "generatedSources/java"
-    val mainRClassPath = os.walk(rClassDir)
+    val javaSources = os.walk(rClassDir).filter(p => os.isFile(p) && p.ext == "java")
+    val mainRClassPath = javaSources
```
vaslabs: can rename this to rSources

example/androidlib/java/1-hello-world/build.mill
```diff
@@ -99,7 +99,7 @@ object app extends AndroidAppModule {
 /** Usage
 
 > ./mill show app.test
-...compiling 1 Java source...
+...compiling 2 Java sources to out/app/test/compile.dest/classes...
```
souvlakias, line 102: This was intended to match for the `compile` task and should have failed when we defaulted BuildConfig to true in #6145 as a second source was being added, but it coincidentaly matched with the `androidCompiledRClasses`
> compiling 1 Java sources to out/app/test/androidCompiledRClasses.dest/classes

Now we explicitly match for the `compile` task

vaslabs:
the androidtodo does not seem to be working with the changes

```
8106]   Test com.example.android.architecture.blueprints.todoapp.taskdetail.TaskDetailViewModelTest.activateTask 
failed: java.lang.NoClassDefFoundError: com/example/android/architecture/blueprints/todoapp/R$string, took 0.072 sec
```

vaslabs:
pokemon multimodule is also complaining:

```
Error: Missing class com.nicos.compose_ui.R$color (referenced from: int com.nicos.compose_ui.utils.extensions.ExtensionsKt.colorToInt$default(android.graphics.drawable.Drawable, android.content.Context, int, int, java.lang.Object))
599] Missing class com.nicos.compose_ui.R$drawable (referenced from: kotlin.Unit com.nicos.compose_ui.components.ToolbarKt.CustomToolbar_3IgeMak$lambda$8(kotlin.jvm.functions.Function0, androidx.compose.runtime.Composer, int))
599] Missing class com.nicos.compose_ui.R$string (referenced from: kotlin.Unit com.nicos.compose_ui.components.ComposableSingletons$GenericViewsKt.lambda__515675410$lambda$0(androidx.compose.foundation.layout.RowScope, androidx.compose.runtime.Composer, int) and 4 other contexts)
599] Missing class com.nicos.network.R$string (referenced from: java.lang.String com.nicos.network.generic_classes.HandlingError.handleErrorMessage(java.lang.Throwa
```

vaslabs:
I think we need transitive modules instead

libs/androidlib/src/mill/androidlib/AndroidModule.scala
```diff
@@ -131,15 +131,24 @@ trait AndroidModule extends JavaModule { outer =>
...
-      Seq("--auto-add-overlay")
-    }
+    val debugOptions = Seq(
+      "--proguard-minimal-keep-rules",
+      "--debug-mode"
+    )
+
+    // Add module dependencies' namespaces as extra packages
+    // TODO: cleanup once we properly pass resources from dependencies
+    val extraPackages = moduleDeps.collect {
```
vaslabs: maybe here we need transitive modules.

Also for better customisation , we can make this into a task.

`androidAaptLinkExtraPackages`

libs/androidlib/src/mill/androidlib/AndroidAppModule.scala
```diff
@@ -235,6 +235,10 @@ trait AndroidAppModule extends AndroidModule { outer =>
     )
   }
 
+  override def androidAaptOptions: T[Seq[String]] = Task {
+    super.androidAaptOptions().filterNot(_ == "--non-final-ids")
```
vaslabs: we should have a task androidAaptNonFinalIds instead, so we avoid filtering and append it [here](https://github.com/com-lihaoyi/mill/pull/6249/changes#diff-860dd1deac6289159d6fb5ae208aa73a654e2b39f0cda226b63cd920d443d1c7R149)



if it's set to true

PR #6250, opened 2025-11-26
Allow users to provide their own compiler bridge

alexarchambault:
This adds a `ScalaModule#scalaCompilerBridge` task, whose value is an `Option[PathRef]`. It allows users to pass their own compiler bridge, instead of having Mill fetch and / or compile one on its own internally.

This is useful when trying to build Scala 3 with Mill: Scala 3 builds its own bridge, that we want to use at some point in the Scala 3 build.

### DIFF
--- a/libs/scalalib/src/mill/scalalib/ScalaModule.scala
+++ b/libs/scalalib/src/mill/scalalib/ScalaModule.scala
@@ -233,6 +233,15 @@ trait ScalaModule extends JavaModule with TestModule.ScalaModuleBase
     defaultResolver().classpath(scalacPluginMvnDeps())
   }
 
+  /**
+   * Manually supplied compiler bridge
+   *
+   * The file this points at can be either a JAR containing class files, or a directory
+   * with source files.
+   *
+   * If `None`, Mill fetches and compiles if needed a compiler bridge on its own.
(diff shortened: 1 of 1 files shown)


alexarchambault:
(The CI error is a flaky test, unrelated to the changes here)

libs/scalalib/src/mill/scalalib/ScalaModule.scala
```diff
@@ -232,6 +233,8 @@ trait ScalaModule extends JavaModule with TestModule.ScalaModuleBase
     defaultResolver().classpath(scalacPluginMvnDeps())
   }
 
+  def scalaCompilerBridge: T[Option[PathRef]] = Task(None)
```
lefou, line 245: Docs please. What does `None` mean.

alexarchambault: Sorry, that's an oversight

alexarchambault: Should be fixed

lefou: Should this be a `Seq[PathRef]`, to support a multi-entry classpaths?

PR #6257, opened 2025-11-27
Remove unnecessary `new` keywords

lihaoyi:
Since Scala 3, you don't need to put `new` unless you need to explicitly call the constructor rather than a companion object `apply` method, or you are instantiating an anonymous subclass. This PR removes all the rest


lihaoyi:
Seems like Scalafix `RemoveUnused` spuriously removes imports used in creator applications

lihaoyi:
This is blocked on https://github.com/scala/scala3/issues/24562

PR #6262, opened 2025-11-28
Add MILL_BSP_OUTPUT_DIR env var and rework defaults

lefou:
* Moved the ouput path calculation logic into the constants, so that `out`, `bspOut`, `mergeBspOut` and `outDir(outMode)` act more consistently

* Added new `MILL_BSP_OUTPUT_DIR` env variable to let the user control an alternative location

* Change default for the BSP output dir from `out/mill-bsp-out` to `.bsp/mill-bsp-out`. Reasons:

  1. This places BSP related files under the specific `.bsp` directory, so user can easily clean all BSP related files in one go.
  2. Avoid nested Mill output directories. E.g. Cleaning Mill's cache and daemon state by running `rm -r out/` may break the IDE (make it unresponsive), which is often not the wanted behavior.

### DIFF
--- a/website/docs/modules/ROOT/pages/cli/installation-ide.adoc
+++ b/website/docs/modules/ROOT/pages/cli/installation-ide.adoc
@@ -384,7 +384,7 @@ A lot of other editors may work too, since {link-metals}[Metals], the Language S
 === Debugging IDE issues
 
 Mill's BSP IDE integration writes to log files under
-`out/mill-bsp/`, where you can find various information about what's
+`.bsp/mill-bsp-out/mill-bsp/`, where you can find various information about what's
 going on. It contains regular Mill output accompanied by additional BSP
 client-server communication details. This can be useful to look at if your
 IDE fails to import your Mill project


lefou:
As mentioned multiple times earlier, I think we never want to use `static final` fields in Java classes for "constant" values that might change in subsequent releases, since the compiler is allowed to inline those fields, resulting in stale values in downstream bytecode. 

Instead, we should use methods, which results in more boilerplate though.

lefou:
> As mentioned multiple times earlier, I think we never want to use `static final` fields in Java classes for "constant" values that might change in subsequent releases, since the compiler is allowed to inline those fields, resulting in stale values in downstream bytecode.
> 
> Instead, we should use methods, which results in more boilerplate though.

* Opened #6266 to address this issue

lefou:
@lihaoyi Can I get your opinion?

lihaoyi:
Looks good to me

lefou:
@lihaoyi I updated the tests, and they pass. But I'm not 100% sure why the test literals now contain `MillMiscInfo.scala` and `BuildFileImpl.scala` but didn't before. Is there any ignore-logic that now no longer applies due to the changed out path?

lihaoyi:
@lefou I think you need to add `.bsp/` to this list https://github.com/com-lihaoyi/mill/blob/main/runner/meta/src/mill/meta/MillBuildRootModule.scala#L40

Issue #5402, opened 2025-06-30
Support a task query in `TaskModule.defaultTask`


lefou:
It would be nice, if one could return a task in a sub-module from `defaultTask`.

Example:

```scala
object example extends TaskModule {

  def defaultTask() = "foo.bar.testForked"

  trait MyTest extends TestModule {
    ..
  }

  object foo extends Module {
    object bar extends MyTest
    object baz extends MyTest
  }

}
```

Bonus, if we also support a wildcard query:

```scala
  def defaultTask() = "foo._.testForked"
```

lefou:
My typical workaround is to add a forwarder task,but that's cumbersome and boilerplate-y. 

https://github.com/lefou/mill-vcs-version/blob/13359114747338d5b490499daed5b409d8f5292b/build.sc#L153-L157

```scala
// Mill 0.11
object itest extends Cross[ItestCross](millItestVersions.map(_._1)) with TaskModule {
  override def defaultCommandName(): String = "test"
  def testCached: T[Seq[TestCase]] = itest(millItestVersions.map(_._1).head).testCached
  def test(args: String*): Command[Seq[TestCase]] = itest(millItestVersions.map(_._1).head).test(args: _*)
}
```

PR #6280, opened 2025-12-01
Add a `example.javalib.basic[3-simple]` alias for integration and example tests

lihaoyi:
So we don't need to keep writing `.packaged.daemon` every time we want to run something

### DIFF
--- a/integration/package.mill
+++ b/integration/package.mill
@@ -40,7 +40,7 @@ object `package` extends mill.Module {
     }
   }
 
-  trait IntegrationTestModule extends Cross.Module[String] with DefaultTaskModule{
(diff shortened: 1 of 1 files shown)


integration/package.mill
```diff
@@ -113,6 +113,13 @@ object `package` extends mill.Module {
 
     }
 
+    // Set `packaged.daemon` as the default way to run integration and example tests
+    override def defaultTask(): String = "testForked"
+    def testForked(args: String*): Task.Command[(msg: String, results: Seq[TestResult])] =
+      Task.Command {
+        packaged.daemon.testForked(args*)()
+      }
+
```
lefou, line 122: Related: https://github.com/com-lihaoyi/mill/issues/5402

PR #6298, opened 2025-12-03
Remove `--add-default` parameter from `testForked` due to breaking semantic compatibility

lihaoyi:
This is causing problems rebootstrapping due to the `override def testForked` in `integration/package.mill` no longer working due to the change in signature.

It seems that for non-final tasks, adding parameters cannot be done safely in a binary/semantic-compatible way. That is the same limitation we hit in https://docs.scala-lang.org/sips/unroll-default-arguments.html, and I guess is just something we'll have to live with

### DIFF
--- a/libs/javalib/src/mill/javalib/TestModule.scala
+++ b/libs/javalib/src/mill/javalib/TestModule.scala
@@ -95,9 +95,6 @@ trait TestModule
     }
   }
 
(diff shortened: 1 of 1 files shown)


lefou:
I lately think, we should just migrate to a `String*` API for commands, mostly as this provides the best option for parameter evolution.

* https://github.com/com-lihaoyi/mill/discussions/2594#discussioncomment-15017493

PR #6313, opened 2025-12-04
Try and fix coursier version regex in gen-idea tests

lihaoyi:


lihaoyi:
@alexarchambault it seems the coursier resolution is causing differing libraries to be  resolved on OS-X and linux, and worse is ending up with multiple versions of some jars on the runtime classpath when resolved on OS-X. Can you take a look at https://github.com/com-lihaoyi/mill/actions/runs/19921835214/job/57112794490 and see why that might be the case? We need to make this test reproducible so that people on different machines can update the golden files, otherwise they're impossible to maintain

alexarchambault:
@lihaoyi Couldn't you have a locally published coursier version under `~/.ivy2/local` on one of these machines? The way the test data are updated is suspicious, it seems the coursier JARs originate from `~/.ivy2/local` rather then a Maven repository cached in the coursier cache.

alexarchambault:
The file names under `~/.ivy2/local` are like `coursier-cache_2.13.jar` (no version) while those from a Maven repository are like `coursier-cache_2.13-2.1.25-M19.jar`, which confuses this test.

lihaoyi:
@alexarchambault good point, that might be the reason. Let me wipe out my `.ivy2/local` cache

Issue #6310, opened 2025-12-04
Spring boot AOT mode fails with Kotlin


re-thc:
Hi, just tested 1.1.0-RC3 with the new `SpringBootModule`. Thanks for adding it. Using the "optimized" AOT mode fails with Kotlin. It works with Java.

[error] out/test/springBootProcessAOT.dest/sources/com/test/App__BeanDefinitions.java:16:64                                                     ^
cannot find symbol
symbol:   class App
location: class com.test.App__BeanDefinitions

vaslabs:
Thanks for reporting, I'll take a look today, could you also share a sample of your build.mill configuration?  Or if you have a complete example even better, I can add it in the collection

lefou:
@vaslabs Probably related: https://github.com/com-lihaoyi/mill/pull/6168#issuecomment-3528129611 (If this involves annotation processors)

vaslabs:
> [@vaslabs](https://github.com/vaslabs) Probably related: [#6168 (comment)](https://github.com/com-lihaoyi/mill/pull/6168#issuecomment-3528129611) (If this involves annotation processors)

thanks, yes, I've just started working on it! EDIT: (I mean not sure it's the annotation processors, I'll know soon)

vaslabs:
I suspect it's the directory structure (e.g. need to use `KotlinMavenModule`), I'm preparing a PR with this working example:

```scala
package build

import mill.*
import mill.javalib.*
import mill.javalib.spring.boot.SpringBootModule
import mill.javalib.TestModule.{Junit4, Junit5}
import mill.javalib.publish.*
import mill.kotlinlib.KotlinMavenModule

object `package` extends SpringBootModule, KotlinMavenModule { outer =>

  override def springBootPlatformVersion = "3.5.7"

  override def kotlinVersion = "2.2.21"
  override def artifactName: T[String] = "spring-boot-native-demo"

  override def kotlincOptions: T[Seq[String]] = super.kotlincOptions() ++ Seq("-jvm-target", "17")

  override def mvnDeps = Seq(
    mvn"com.fasterxml.jackson.module:jackson-module-kotlin:2.19.2",
    mvn"org.jetbrains.kotlin:kotlin-reflect:${kotlinVersion()}",
    mvn"org.jetbrains.kotlin:kotlin-stdlib:${kotlinVersion()}",
    mvn"org.springframework.boot:spring-boot-starter-web"
  )

[...]

PR #6315, opened 2025-12-04
Extend Native (GraalVM) support

vaslabs:
## Motivation

- Related to https://github.com/com-lihaoyi/mill/issues/6310
- Upon investigation I've discovered we can extend native support to multiple libraries using https://github.com/oracle/graalvm-reachability-metadata which is what both gradle and maven graal vm plugins are doing
- The Kotlin example is not using the maven layout per @lefou 's suggestion

## Implementation

- For every direct mvn dependency, we get the relevant directory from the graalvm-reachability-metadata.
- The metadata need to both be in run classpath and passed as a `--configurations-path`
- The `--configurations-path` is used to simplify native image options - no more scraping the properties file
- The maven deps use the resolution to get all runtime dependencies and resolve their reachability metadata
- Both finding the reachability metadata and copying them over reuse the native-build-tools functionality which uses a walk file tree logic compatible to the metadata structure. 

## Outcomes + examples

- To demonstrate this working, there's an example with Kotlin + Spring Boot + native configuration which before was failing with 
```
141] kotlin.jvm.KotlinReflectionNotSupportedError: Kotlin reflection implementation is not found at runtime. Make sure you have kotlin-reflect.jar in the classpath
```
The error was because we didn't build the native image with the reachability metadata.

- I've also added some unit tests to show some "unexpected" behaviour and the need to reuse the copying mechanism from native-build-tools instead of copying on our own, as it takes care of how the structure is in graalvm-reachability-metadata. For more information on this functionality see [here](https://github.com/graalvm/native-build-tools/blob/master/common/graalvm-reachability-metadata/src/main/java/org/graalvm/reachability/DirectoryConfiguration.java#L95)

Pull request: https://github.com/com-lihaoyi/mill/pull/6315

### DIFF
--- a/libs/javalib/src/mill/javalib/NativeImageModule.scala
+++ b/libs/javalib/src/mill/javalib/NativeImageModule.scala
@@ -225,7 +225,7 @@ trait NativeImageModule extends WithJvmWorkerModule {
    * For more information and implementation details go to [[https://github.com/graalvm/native-build-tools/blob/master/common/graalvm-reachability-metadata/src/main/java/org/graalvm/reachability/internal/FileSystemRepository.java]]
    */
   def nativeGraalVmMetadataQuery: T[MetadataQuery] = this match {
-    case m: JavaModule => Task {
+    case _: JavaModule => Task {
         val metadataPath = nativeGraalVMReachabilityMetadata().path
         MetadataQuery(
           rootPath = metadataPath,
@@ -250,10 +250,15 @@ trait NativeImageModule extends WithJvmWorkerModule {
    */
   def nativeGraalVmQueryDeps: T[Set[String]] = this match {
     case m: JavaModule => Task {
-        val metadataPath = nativeGraalVMReachabilityMetadata().path
+
+        def isValidGAVSection(value: String): Boolean = value.nonEmpty && !value.contains(':')
 
         def isValidGAV(d: coursier.core.Dependency): Boolean =
-          d.module.organization.value.nonEmpty && d.module.name.value.nonEmpty && d.versionConstraint.asString.nonEmpty
+          Seq(
+            d.module.organization.value,
+            d.module.name.value,
+            d.versionConstraint.asString
+          ).forall(isValidGAVSection)
 
         def artifactQueryGav(d: coursier.core.Dependency): String =
           s"${d.module.organization.value}:${d.module.name.value}:${d.versionConstraint.asString}"


example/kotlinlib/web/9-spring-boot-aot/build.mill
```diff
@@ -0,0 +1,75 @@
+package build
+
+import mill.*
+import mill.javalib.*
+import mill.javalib.spring.boot.SpringBootModule
+import mill.javalib.TestModule.{Junit4, Junit5}
+import mill.javalib.publish.*
+import mill.kotlinlib.KotlinMavenModule
+
+object `package` extends SpringBootModule, KotlinMavenModule { outer =>
```
lefou: Let's use the Mill default layout in this example. We already have an example for `SpringBootModule & MavenModule` in `example/javalib/springboot/6-native`, but non without `MavenModule`.

vaslabs:
Remaining issue is with native:

```
h failed: kotlin.jvm.KotlinReflectionNotSupportedError: Kotlin reflection implementation is not found at runtime. Make sure you have kotlin-reflect.jar in the classpath] with root cause
141] 
141] kotlin.jvm.KotlinReflectionNotSupportedError: Kotlin reflection implementation is not found at runtime. Make sure you have kotlin-reflect.jar in the classpath
141]    at kotlin.jvm.internal.ClassReference.error(ClassReference.kt:88) ~[na:na]
141]    at kotlin.jvm.internal.ClassReference.getMembers(ClassReference.kt:18) ~[na:na]
141]    at kotlin.reflect.full.KClasses.getFunctions(KClasses.kt:87) ~[native-executable:2.2.21-release-469]
141]    at kotlin.reflect.jvm.ReflectJvmMapping.getKotlinFunction(ReflectJvmMapping.kt:140) ~[na:na]
```

I think some META-INF or resource configuration is missing

vaslabs:
ok I found the reason for the native issue and we need a more holistic approach to fix it, so I'll keep this PR short and deal with it in another PR.

Basically, this project

https://github.com/oracle/graalvm-reachability-metadata/tree/master

has  a collection of metadata for various libraries to work with the native images.

We probably need to integrate it into the NativeImageModule in a way that users just pass a list of library names and versions (the versions don't necessarily match with whatever is going to be in maven deps from what I see) and use the files in the configurations directory

vaslabs:
Well, I think it was easier than I anticipated in my last comment. I think I can do it in this PR.

I put a [FIXME workaround](https://github.com/com-lihaoyi/mill/pull/6315/commits/687796646ae59900a97ae3090deb28a8bb33f79f#diff-5e78e9d225efc09c61e15e4b5933fc6e365fd1d4f0700fa75ac1fe415bc44430R30) just to demonstrate it working, but obviously it's going to look a bit different in the NativeImageModule + SpringBootModule.

@lefou what do you think? any particular input or consideration before I add the changes in Mill?

This will also increase the value of Mill's NativeImageModule of course , although I'm not sure how we can generalise this for every user

vaslabs:
I'm also not sure if the native image picks every file or not. If it does, I'm thinking to use resolvedRunMvnDeps and then for each get organisation + name , and copy whatever is in the metadata directory if it exists (or programmatically the other way around, copy dir if it's in resolved run mvn deps)

vaslabs:
I'm looking how much we can use this to complete the implementation https://github.com/graalvm/native-build-tools

vaslabs:
@lihaoyi @lefou I think the functionality implementation wise is there.

The only thing remaining, is I think we should have either a JavaNativeModule or make the NativeImageModule aware of mvn deps in case it is also a java module.

what do you think?

lefou:
We should move the complete logic into `NaticeImageModule`. Matching on the module type is is a valid approach for support modules. We do it in other places too, e.g. `TestModule`.

vaslabs:
there's a failure on my copy but I see the same one [in the main branch](https://github.com/com-lihaoyi/mill/actions/runs/19955354582/job/57223784958) so I think it's unrelated.

This PR should be ready for review

libs/javalib/src/mill/javalib/NativeImageModule.scala
```diff
@@ -236,7 +236,7 @@ trait NativeImageModule extends WithJvmWorkerModule {
         val deps = resolution.dependencies
           .map(d =>
             s"${d.module.organization.value}:${d.module.name.value}:${d.versionConstraint.asString}"
-          )
+          ).filter(_.split(':').length == 3) // gav requirements for graalvm metadata artifact query
```
lefou: Can we log the skipped dependencies, to assist the poor soul that tries to understand her misbehaving build?

vaslabs: Yep, but also looking at it again, I can do a better implementation

lihaoyi:
I think it looks reasonable, @lefou when you're done reviewing we can merge it

PR #6320, opened 2025-12-05
More widespread use of  Java17/Scala3.8 and narrow Java11/Scala3.7 to necessary code only

lihaoyi:
Previously we were forced to use Scala 3.7 for all modules containing `export` clauses due to https://github.com/scala/scala3/issues/24553, but since that is fixed we can migrate to 3.8 more aggressively


lihaoyi:
Blocked on https://github.com/scala/scala3/issues/24671

PR #6336, opened 2025-12-07
Ensure `must be defs defined in a Module class/trait/object body` error happens for commands

lihaoyi:
The basic issue is that we did this check in `Cacher`, but since `Command`s are not cached this check did not take effect. The solution is to move the check into `Task.scala`'s `assertTaskShapeOwner` so it happens for all named tasks

covered by unit tests


lihaoyi:
CC @lefou

PR #6353, opened 2025-12-09
Fix `./mill __.compile`

lihaoyi:


lefou:
We can revert this, once we bump the Scala version, which is blocked by https://github.com/com-lihaoyi/mill/pull/6351

Issue #6355, opened 2025-12-09
Android: androidtodo is broken


vaslabs:
Not sure since when, probably after RC2

```
8461]   /home/vnicolaou/mill/out/example/thirdparty/androidtodo/packaged/daemon/testForked.dest/sandbox/run-1/app/src/main/res/font/opensans_font.xml:18: error: attribute font (aka com.example.android.architecture.blueprints.main.test:font) not found.
8461]   /home/vnicolaou/mill/out/example/thirdparty/androidtodo/packaged/daemon/testForked.dest/sandbox/run-1/app/src/main/res/font/opensans_font.xml:18: error: attribute fontStyle (aka com.example.android.architecture.blueprints.main.test:fontStyle) not found.
8461]   /home/vnicolaou/mill/out/example/thirdparty/androidtodo/packaged/daemon/testForked.dest/sandbox/run-1/app/src/main/res/font/opensans_font.xml:18: error: attribute fontWeight (aka com.example.android.architecture.blueprints.main.test:fontWeight) not found.
8461]   /home/vnicolaou/mill/out/example/thirdparty/androidtodo/packaged/daemon/testForked.dest/sandbox/run-1/app/src/main/res/font/opensans_font.xml:22: error: attribute font (aka com.example.android.architecture.blueprints.main.test:font) not found.
8461]   /home/vnicolaou/mill/out/example/thirdparty/androidtodo/packaged/daemon/testForked.dest/sandbox/run-1/app/src/main/res/font/opensans_font.xml:22: error: attribute fontStyle (aka com.example.android.architecture.blueprints.main.test:fontStyle) not found.
8461]   /home/vnicolaou/mill/out/example/thirdparty/androidtodo/packaged/daemon/testForked.dest/sandbox/run-1/app/src/main/res/font/opensans_font.xml:22: error: attribute fontWeight (aka com.example.android.architecture.blueprints.main.test:fontWeight) not found.
...
8461]   [error] app.androidTest.androidLinkedResources
8461]   os.SubprocessException: Result of /home/vnicolaou/Android/Sdk/build-tools/35.0.0/aapt2?: 1
8461]   
8461]   os.SubprocessException$.apply(Model.scala:211)
8461]     os.proc.call(ProcessOps.scala:232)
8461]     os.call$.apply(ProcessOps.scala:42)
8461]     mill.androidlib.AndroidModule.androidLinkedResources$$anonfun$1$$anonfun$1(AndroidModule.scala:764)
[...]

PR #6356, opened 2025-12-09
Fix androidtodo

vaslabs:
Fixes the issue of instrumented tests pulling resources from the main app: https://github.com/com-lihaoyi/mill/issues/6355

Issue was after non transitive R class revert, we've left the androidTransitiveModuleDeps resulting in missing resources


vaslabs:
~~but it breaks compose samples, need to push for the proper fix instead~~

fixed both

PR #6365, opened 2025-12-10
Add asciinema recording for Mill vs Maven performance

lihaoyi:
Recorded it with `asciinema rec both.cast --overwrite ` using `screen` to split the terminal into two halves

### DIFF
--- a/website/docs/modules/ROOT/pages/index.adoc
+++ b/website/docs/modules/ROOT/pages/index.adoc
@@ -143,16 +143,16 @@ Mill is used to build many real-world projects, such as:
 https://github.com/disneystreaming/smithy-translate/blob/main/build.sc[Disney-Streaming],
 and other organizations
 
+To engage Mill with the community, you can use the channels below:
+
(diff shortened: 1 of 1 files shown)


website/docs/modules/ROOT/partials/Asciinema_Comparison.adoc
```diff
@@ -0,0 +1,19 @@
+++++
+<link rel="stylesheet" type="text/css" href="https://cdn.jsdelivr.net/npm/asciinema-player@3.10.0/dist/bundle/asciinema-player.css" />
+<script src="https://cdn.jsdelivr.net/npm/asciinema-player@3.10.0/dist/bundle/asciinema-player.min.js"></script>
```
lefou, line 3: Can we instead deliver these files from our site?

lihaoyi: Possibly, but relying on a CDN is probably the next best thing

PR #6366, opened 2025-12-10
Make use of `-language:experimental.packageObjectValues

lihaoyi:
This PR makes use of [SIP-68 Referenceable Package Objects ](https://github.com/scala/improvement-proposals/pull/100) in `.mill` files. This simplifies our codegen considerably, since we no longer need to "fake" the package objects via normal `object`s and aliasing `import`s and `lazy val`s


lihaoyi:
Probably can't land this until SIP-68 is non-experimental because of weirdness in how `@experimental` annotations are propagated

PR #6378, opened 2025-12-12
Rebootstrap on top of new-cross-syntax changes

lihaoyi:


lihaoyi:
Seems there are issues in the `.shared` example tests in this PR, with `./mill example.javalib.basic._.shared.daemon` failing but every test passing when run individually

Issue #6379, opened 2025-12-12
Android: Databinding and hilt are not compatible with incremental compilation


vaslabs:
Found this issue in our [compose samples with mill](https://github.com/vaslabs-ltd/compose-samples-with-mill/actions/runs/20160764913/job/57872441950#step:5:723) testing repo. It seems that generated sources by both ksp and databinding don't seem to work.

For now the workaround is to disable the incremental compilation with:

```
  override def kotlincUseBtApi: T[Boolean] = Task {
    false
  }
```

but maybe there's a fix for this. Opening the issue to investigate

vaslabs:
I think this is related to [this](https://github.com/com-lihaoyi/mill/pull/5785#issuecomment-3273752670) . Do you recall @lefou ?

feels like the change [here](https://github.com/com-lihaoyi/mill/pull/6360/changes#diff-6eb53290d742916b27027e19a830628c4ff8d33f60f77f0e49263677a5e60401L46) causes the same issue, I'll test it now

EDIT:

I put a relevant PR (https://github.com/com-lihaoyi/mill/pull/6381) although perhaps it needs to be done a bit differently (before it has passing an empty list in the place of sources)

EDIT2:

I think this was an issue with mixed compilation instead, the generated sources in the above cases are java files

lefou:
I guess the BT-API is not very intuitive or some kind of "grown this way".

lefou:
Closed by #6381

PR #5785, opened 2025-08-31
Use Kotlin build-tools-api to compile Kotlin code

lefou:
With Kotlin 2.2, there is a new beta build tools api (BT API) which should be used by build tools to integrate the Kotlin compiler. This PR adds support for this API and uses it by default if the module is using Kotlin 2.0+ for JVM.

Related to https://github.com/com-lihaoyi/mill/issues/5788

PR #6381, opened 2025-12-12
Kotlin fix incremental compilation for generated sources

vaslabs:
Fix for https://github.com/com-lihaoyi/mill/issues/6379

It seems that in order to get the compilation to take mixed sources (Java + Kotlin) we should pass all the files as [compiler arguments](https://github.com/JetBrains/kotlin/blob/677deda4b2f29d931161cd6d234fafbf023e5f81/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/compilerRunner/btapi/BuildToolsApiCompilationWork.kt#L111-L117)

as found by @lefou (see [this comment](https://github.com/com-lihaoyi/mill/pull/5785#issuecomment-3273752670))

This was evident in ksp setups because both hilt and dagger generate java files. But the problem is not necessarily on KSP / dependency injection frameworks, but in mixed sources in general.

### DIFF
--- a/example/kotlinlib/module/10-dependency-injection/build.mill
+++ b/example/kotlinlib/module/10-dependency-injection/build.mill
@@ -14,9 +14,6 @@ object dagger extends KspModule {
 
   def kotlincOptions = super.kotlincOptions() ++ Seq("-no-reflect", "-verbose")
 
-  // dagger doesn't play well with BtApi's incremental compilation
-  override def kotlincUseBtApi = false
-
   def mvnDeps = Seq(
     mvn"com.google.dagger:dagger:2.57"
   )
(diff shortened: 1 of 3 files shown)


vaslabs:
> This looks ok to me. @vaslabs do you see any reasons not to merge this, as you labelled the "fix" with a question mark?

I don't see a particular reason , I added the question just in case you discovered something new since then

lihaoyi:
@vaslabs could you leave a comment or update the PR description with your current understanding of the fix

vaslabs:
> @vaslabs could you leave a comment or update the PR description with your current understanding of the fix

done, I haven't done a thorough investigation to see what's happening under the hood, but this is my current understanding

PR #6409, opened 2025-12-18
Introduce `mill.api.OneOrMore` type, use it for `HeaderData#extends` to allow a single superclass to be passed

lihaoyi:
This should streamline the common case to allow a single superclass to be passed and reference the common super-classes un-qualified


core/api/src/mill/api/internal/SimpleTaskTokenReader.scala
lefou, line 1: Isn't this a binary incompatible change?

lihaoyi: It's `private[mill]', so i think its fine?

lefou: I see. Then we need to apply some problem filter.

Issue #6413, opened 2025-12-18
Current `main` branch breaks binary compatibility


lefou:
```console
> just check-mima 
mill -j 0 __.mimaReportBinaryIssues
...
2140] [error] Found 3 issue when checking against com.lihaoyi:mill-core-api_3:VersionConstraint.Lazy(1.0.0, VersionInterval(None, None, false, false), Some(Version(1.0.0)), None)
2140] [error]  * class mill.api.Cached does not have a correspondent in current version
2140]    filter with: ProblemFilter.exclude[MissingClassProblem]("mill.api.Cached")
2140] [error]  * object mill.api.Cached does not have a correspondent in current version
2140]    filter with: ProblemFilter.exclude[MissingClassProblem]("mill.api.Cached$")
2140] [error]  * class mill.api.SimpleTaskTokenReader does not have a correspondent in current version
2140]    filter with: ProblemFilter.exclude[MissingClassProblem]("mill.api.SimpleTaskTokenReader")
...
2140] [error] Found 3 issue when checking against com.lihaoyi:mill-core-api_3:VersionConstraint.Lazy(1.0.6, VersionInterval(None, None, false, false), Some(Version(1.0.6)), None)
2140] [error]  * class mill.api.Cached does not have a correspondent in current version
2140]    filter with: ProblemFilter.exclude[MissingClassProblem]("mill.api.Cached")
2140] [error]  * object mill.api.Cached does not have a correspondent in current version
2140]    filter with: ProblemFilter.exclude[MissingClassProblem]("mill.api.Cached$")
2140] [error]  * class mill.api.SimpleTaskTokenReader does not have a correspondent in current version
2140]    filter with: ProblemFilter.exclude[MissingClassProblem]("mill.api.SimpleTaskTokenReader")
2189] done compiling
2189] compiling 12 Scala sources to out/libs/javalib/compile.dest/classes ...
2189] done compiling
2331/2331, 1 failed] ============================== __.mimaReportBinaryIssues ============================== 144s
2140] [error] core.api.mimaReportBinaryIssues Failed binary compatibility check! Found 21 potential problems (filtered 147)
```

lefou:
These are false positives, as all three moved classes where `private[mill]`.

lefou:
Fixed by #6411

PR #6414, opened 2025-12-18
Add MiMa problem filters for changed private classes

lefou:
Fix https://github.com/com-lihaoyi/mill/issues/6413


lihaoyi:
I already added these as part of https://github.com/com-lihaoyi/mill/commits/main/

Issue #6075, opened 2025-11-01
Crash when importing gradle projects `ghidra` or `RxJava`


lihaoyi:
```
lihaoyi Github$ git clone --depth=1 https://github.com/NationalSecurityAgency/ghidra
Cloning into 'ghidra'...
remote: Enumerating objects: 23711, done.
remote: Counting objects: 100% (23711/23711), done.
remote: Compressing objects: 100% (17364/17364), done.
remote: Total 23711 (delta 7258), reused 12749 (delta 4783), pack-reused 0 (from 0)
Receiving objects: 100% (23711/23711), 75.14 MiB | 13.91 MiB/s, done.
Resolving deltas: 100% (7258/7258), done.
Updating files: 100% (19702/19702), done.
lihaoyi Github$ cd ghidra 

lihaoyi ghidra$ ./mill init
[1/1] init
[14/14] mill.init.InitGradleModule.init
converting Gradle build
connecting to Gradle daemon
org.gradle.tooling.BuildException: Could not fetch model of type 'BuildModel' using connection to Gradle distribution 'https://services.gradle.org/distributions/gradle-8.11.1-bin.zip'.
	at org.gradle.tooling.internal.consumer.ConnectionExceptionTransformer.transform(ConnectionExceptionTransformer.java:53)
	at org.gradle.tooling.internal.consumer.ResultHandlerAdapter.onFailure(ResultHandlerAdapter.java:42)
	at org.gradle.tooling.internal.consumer.async.DefaultAsyncConsumerActionExecutor$1$1.run(DefaultAsyncConsumerActionExecutor.java:68)
	at org.gradle.internal.concurrent.ExecutorPolicy$CatchAndRecordFailures.onExecute(ExecutorPolicy.java:64)
	at org.gradle.internal.concurrent.AbstractManagedExecutor$1.run(AbstractManagedExecutor.java:48)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
	at org.gradle.tooling.internal.consumer.BlockingResultHandler.getResult(BlockingResultHandler.java:46)
	at org.gradle.tooling.internal.consumer.DefaultModelBuilder.get(DefaultModelBuilder.java:50)
[...]

PR #6415, opened 2025-12-18
Added check for first character of a segment in `Util.backtickWrap`

ajaychandran:
Segments made of numeric characters were causing a crash for an example listed in #6075.
```
Caused by: .../ghidra/Ghidra/Test/IntegrationTest/package.mill:10: error: [dialect scala3] `)` expected but `double constant` found
  def moduleDeps = Seq(build.Ghidra.Processors.68000, build.Ghidra.Processors.8051, build.Ghidra.Processors.AARCH64, ...
```
This fix wraps such values in backticks to produce a legal identifier.

### DIFF
--- a/core/internal/src/mill/internal/Util.scala
+++ b/core/internal/src/mill/internal/Util.scala
@@ -61,7 +61,11 @@ object Util {
 
   def backtickWrap(s: String): String = s match {
     case s"`$_`" => s
-    case _ => if (encode(s) == s && !alphaKeywords.contains(s)) s
+    case _ =>
+      if (
+        encode(s) == s && !alphaKeywords.contains(s) &&
+        (s.isEmpty || Character.isJavaIdentifierStart(s.charAt(0)))
+      ) s
       else "`" + s + "`"
   }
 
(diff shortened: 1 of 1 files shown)


core/internal/src/mill/internal/Util.scala
```diff
@@ -61,7 +61,11 @@ object Util {
 
   def backtickWrap(s: String): String = s match {
     case s"`$_`" => s
-    case _ => if (encode(s) == s && !alphaKeywords.contains(s)) s
+    case _ =>
+      if (
+        encode(s) == s && !alphaKeywords.contains(s) &&
+        (s.isEmpty || Character.isJavaIdentifierStart(s.charAt(0)))
```
lihaoyi: When will `s` ever be empty?

lihaoyi:
Can you also update the PR description explaining what the problem is and why this fixes it

core/internal/src/mill/internal/Util.scala
ajaychandran: No. The redundant check has been removed.

ajaychandran:
@lihaoyi The PR title and description were updated.

Issue #6407, opened 2025-12-17
Android: Emulator configurations


souvlakias:
Two issues that came up when integrating Jetcasters `tv` and `wear` modules ([Ref](https://github.com/vaslabs-ltd/compose-samples-with-mill/pull/21/)):

## Image configuration
Currently if a user wants to change the emulator's image (e.g. for wear-os), one needs to override the whole [sdkInstallSystemImage](https://github.com/com-lihaoyi/mill/blob/main/libs/androidlib/src/mill/androidlib/AndroidAppModule.scala#L547) task.
We should have a task (e.g `androidEmulatorImage` that `sdkInstallSystemImage` picks up)

## AVD ID
The default value of [androidVirtualDeviceIdentifier](https://github.com/com-lihaoyi/mill/blob/main/libs/androidlib/src/mill/androidlib/AndroidAppModule.scala#L531) is "test"
This means that all apps (in all projects) access the same avd
Maybe we should default this to something like `<deviceID>-<api>-<arch>-avd` so there are no conflicts, and apps that need the same specs could use the same one.

vaslabs:
let's do the `androidEmulatorImage` first . Usually on a local environment, there's a long running emulator, I think the emulator names and starting it tasks are more useful in a CI rather than a local dev env

PR #6418, opened 2025-12-18
Android: Create `AndroidVirtualDevice` configuration

souvlakias:
Addresses #6407

### Motivation
- Be able to easily configure the AVD
- Have different AVD names for **different** configurations (e.g. for different architectures)

### Changes
- Instead of having many tasks for the AVD configuration, have an `AndroidVirtualDevice` object 
- The name value of the avd defaults to  `s"mill_avd_${apiVersion}_${architecture}_${deviceId}"` so apps that target the same AVD configuration can share the same AVD.

### DIFF
--- a/example/androidlib/java/1-hello-world/build.mill
+++ b/example/androidlib/java/1-hello-world/build.mill
@@ -35,7 +35,7 @@ object app extends AndroidAppModule {
   def androidReleaseKeyPass: T[Option[String]] = Task { Some("MillBuildTool") }
   def androidReleaseKeyStorePass: T[Option[String]] = Task { Some("MillBuildTool") }
 
-  override def androidVirtualDeviceIdentifier: String = "java-test"
+  override def androidVirtualDevice = super.androidVirtualDevice().withName("java-test")
 
   object test extends AndroidAppTests, TestModule.Junit4 {
     def junit4Version = "4.13.2"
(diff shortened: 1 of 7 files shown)


libs/androidlib/src/mill/androidlib/AndroidVirtualDevice.scala
```diff
@@ -0,0 +1,44 @@
...
+ * @param systemImageSource e.g. "google_apis_playstore"
+ */
+case class AndroidVirtualDevice(
+    name: String,
+    apiVersion: String,
+    architecture: String,
+    deviceId: String,
+    systemImageSource: String
+) {
+  val systemImage: String = s"system-images;${systemImageSource};${apiVersion};${architecture}"
```
vaslabs: I think the source and API version are the other way around, that's why the tests are failing

libs/androidlib/src/mill/androidlib/AndroidAppModule.scala
```diff
@@ -527,26 +527,21 @@ trait AndroidAppModule extends AndroidModule { outer =>
...
-  /** The device  id as listed from avdmanager list device. Default is medium_phone */
-  def androidDeviceId: String = "medium_phone"
-
-  /**
-   * The target architecture of the virtual device to be created by  [[createAndroidVirtualDevice]]
-   *  For example, "x86_64" (default). For a list of system images and their architectures,
-   *  see the Android SDK Manager `sdkmanager --list`.
-   */
-  def androidEmulatorArchitecture: String = "x86_64"
+  def androidVirtualDevice: T[AndroidVirtualDevice] = Task {
```
vaslabs, line 533: Need javadocs + update SDK install system image javadocs below to reference this task

Issue #6431, opened 2025-12-21
Occasional NoSuchFileException: .../out/someCommand.dest


lefou:
Unexpected exception. 

`GroupExecution.executeGroupCached` is failing to delete a dest dir. One would think this is a code path used so many times, it's already robust. 

The `dockerBuild` command isn't even using the `Task.dest` path.

```console
java.nio.file.NoSuchFileException: /<project>/out/docker/project-db/dockerBuild.dest
        at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
        at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
        at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
        at java.base/sun.nio.fs.UnixFileSystemProvider.implDelete(UnixFileSystemProvider.java:291)
        at java.base/sun.nio.fs.AbstractFileSystemProvider.delete(AbstractFileSystemProvider.java:105)
        at java.base/java.nio.file.Files.delete(Files.java:1153)
        at os.remove$all$.apply(FileOps.scala:345)
        at os.remove$all$.apply(FileOps.scala:332)
        at mill.exec.GroupExecution.executeGroupCached(GroupExecution.scala:307)
        at mill.exec.GroupExecution.executeGroupCached$(GroupExecution.scala:23)
        at mill.exec.Execution.executeGroupCached(Execution.scala:16)
        at mill.exec.Execution.evaluateTerminals$1$$anonfun$1$$anonfun$1$$anonfun$1(Execution.scala:246)
        at mill.api.daemon.Logger.withPromptLine(Logger.scala:29)
        at mill.api.daemon.Logger.withPromptLine$(Logger.scala:12)
        at mill.internal.PrefixLogger.withPromptLine(PrefixLogger.scala:21)
        at mill.exec.Execution.evaluateTerminals$1$$anonfun$1$$anonfun$1(Execution.scala:278)
        at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:503)
        at mill.exec.ExecutionContexts$.execute$$anonfun$1$$anonfun$1$$anonfun$1$$anonfun$1(ExecutionContexts.scala:67)
        at mill.exec.ExecutionContexts$.execute$$anonfun$1$$anonfun$1$$anonfun$1$$anonfun$adapted$1(ExecutionContexts.scala:68)
[...]

PR #6432, opened 2025-12-21
Only delete the dest dir when it exists

lefou:
Also log it in debug-mode.


lefou:
Don't know if this is the solution. I can no longer reproduce #6431 with `--debug`, but without I still see the same problem.

Issue #6435, opened 2025-12-21
Introduce a `Task.uncached`, which is never cached


lefou:
When we removed the ability to depend on other tasks from `Task.Input`, we disabled some use-cases which can't be implemented by other tasks. The main reason for limitting `Task.Input` was the complexity that arises from the fact that `Task.Input` results must be able to be watched. 

There are use cases, where a task needs to always run, e.g. to check some external state (e.g. a git repository or a container registry), without any special watch mechanism. Using `Task.Command` isn't suitable, as in some situations a `TaskFactory` can effectively produced more than one task instance in the graph. This not only produces unnecessary work but can also lead to concurrent access to the same resources like the `Task.dest` path of that command (e.g. #6431). 

I hereby propose to add a new `Task.uncached` (the name can be discussed), that is not cached and therefore runs once in every evaluation it takes part of. It should work as `Task` but without caching. Other features like `persistent = true` would be nice, too.

lefou:
I made a simple POC in PR #6436 and with it applied I could finally build a larger customer project with Mill 1.1 instead of Mill 0.12.

lefou:
Alternative or in addition to the introduction of `Task.Uncached`, we should fix the underlying issue in `Task.Command` (#6437) and use just non-exclusive commands.

PR #6436, opened 2025-12-22
Add new task type `Task.Uncached`

lefou:
New task type `Task.Uncached`. 

Fix https://github.com/com-lihaoyi/mill/issues/6435

This task type closes a gap in what Mill builds can cover. There was currently no task type, which could map some dependence to an external system, and produce some result only when the external system requires it. In previous Mill version, it was possible to use `Task.Input` for this, but since 1.0 `Task.Input` can no longer have dependencies to other tasks.

For example, to build and register a container image only when an external registry requires it. Or to publish an artifact only when some external repository requires it. Such a task must be able to depend on other task but must also always run, since it's not enough to passively rely on the Mill task cache.

Usage:

```scala
def foo = Task.Uncached {
  // this always runs
}

def bar = Task.Uncached(persisent = true) {
  // this always runs
}
```

### DIFF
--- a/website/docs/modules/ROOT/pages/fundamentals/tasks.adoc
+++ b/website/docs/modules/ROOT/pages/fundamentals/tasks.adoc
@@ -59,7 +59,7 @@ include::partial$example/fundamentals/tasks/6-workers.adoc[]
 === Uncached Tasks
 
 You can define unached tasks with the `Task.Uncached {...}` syntax.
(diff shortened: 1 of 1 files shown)


lihaoyi:
I think this looks fine. One thing we need to note is that `Task.Uncached` does not participate in selective execution or `--watch` triggering. That is probably fine for most use cases, but should be documented

PR #6441, opened 2025-12-23
Convert `./mill init` to generate `.mill.yaml` format, add `!append` support

lihaoyi:
* We generate `.mill.yaml` files when importing Java projects using `./mill init`. 

* We still generate Scala code in `mill-build/src/`, which lets us avoid duplication in the `.mill.yaml` files since the yaml files can `extend` the shared `trait` containing the common config

* Scala project import makes use of cross values which are currently not supported in `.mill.yaml` so we leave those as generating `.mill` files for now. 

* You can now tag top-level YAML values with `!append`, which concatenates the Seq the overridden value rather than completely replacing it. This is necessary for `.mill.yaml` Java project import to work, as we make use of `super ++` in some places. This is implemented via `parseYaml0` converting the `!append` tags into a synthetic `{"$millAppend": ...}` wrapper, which is then deserialized into an `Appendable[T]` that lets us decide later on whether to append something or not

The `!append` stuff is mentioned in some updated doc examples which demonstrate and exercise it

### DIFF
--- a/core/exec/src/mill/exec/GroupExecution.scala
+++ b/core/exec/src/mill/exec/GroupExecution.scala
@@ -398,8 +398,10 @@ trait GroupExecution {
                     val (mergedData, serializedPaths) = PathRef.withSerializedPaths {
                       (taskValue ++ yamlValue.asInstanceOf[Seq[Any]]).asInstanceOf[Any]
                     }
-                    // Write the merged result to disk cache so `show` can pick it up
-                    // Use inputsHash + yaml.## to avoid overwriting the original task cache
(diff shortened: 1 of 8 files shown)


lefou:
Is the YAML renderer an addition to the Scala renderer, or do we loose the latter? 

Asking, since IMHO we should keep the Scala renderer, e.g. to later also convert YAML projects to Scala projects, in case, users need more customizations.

lihaoyi:
Going to merge this first, there's some unresolved breakage that we'll have to look into post-merge. CC @ajaychandran

PR #6445, opened 2025-12-26
Rebootstrap

lihaoyi:


libs/javalib/src/mill/javalib/Lib.scala
```diff
@@ -114,8 +114,10 @@ object Lib {
       )
 
   def scalaConsoleMvnDeps(scalaOrganization: String, scalaVersion: String): Seq[Dep] = {
-    if (scalaVersion.startsWith("3.8.") ||
-        mill.util.Version.isAtLeast(scalaVersion, "3.8")(using mill.util.Version.MavenOrdering)) {
+    if (
+      scalaVersion.startsWith("3.8.") ||
+      mill.util.Version.isAtLeast(scalaVersion, "3.8")(using mill.util.Version.MavenOrdering)
+    ) {
```
lefou, line 120: First part looks redundant. Is this for optimization?

lihaoyi: It's to make it work for 3.8.0-RC versions, which dont seem to pass the `isAtLeast` check otherwise

lefou: Understood.

PR #6459, opened 2025-12-28
Trap `Ctrl-C` in `MillLauncherMain` and `MillNoDaemonMain` to make Ctrl-C in Scala REPL work

lihaoyi:
This allows Ctrl-C to be properly handled by the Scala REPL since it doesn't terminate the launcher process, but has the side effect of Ctrl-C no longer terminating the launcher in other scenarios as well. Somehow this isn't necessary for JShell

### DIFF
--- a/core/constants/src/mill/constants/Util.java
+++ b/core/constants/src/mill/constants/Util.java
@@ -1,8 +1,5 @@
 package mill.constants;
 
-import sun.misc.Signal;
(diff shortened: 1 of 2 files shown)


lihaoyi:
Closing this since https://github.com/scala/scala3/pull/24837 would solve the problem without any of the downsides

PR #6469, opened 2025-12-30
Change YAML path deserialization logic to default to relative paths and use `//` to reference workspace root

lihaoyi:
This should simplify the common case of paths relative to the current folder, while allowing workspace-root-relative paths with a similar syntax to other build tools like Bazel/Buck/Pants


lefou:
Good change.

PR #6473, opened 2025-12-31
Support proguard for java modules

lemony312:
Replace ScalaModule with JavaModule in Proguard trait to support both Java and Scala modules.

Since ScalaModule extends JavaModule, this change maintains backward compatibility while enabling the Proguard trait to work with pure Java projects as well.


lefou (approved):
Looks good to me. Thank you!

PR #6489, opened 2026-01-04
Avoid `String.stripMargin` in code generator templates

lefou:
Generated code may purposefully contain the symbol used as margin `|` , resulting in ill-formatted and wrongly generated code.


lefou:
The failing CI tasks seems unrelated. Ignoring it for now.

Issue #5906, opened 2025-09-23
Provide a built-in command to install/update the `mill` bootstrap script


lefou:
That way, you can use Mill to update or initially install a bootstrap script in a project directory. The command should provide some optional options, e.g. to select the script target platform (all, current, explicit).

PR #6492, opened 2026-01-05
Add `updateMillScripts` command to update the Mill bootstrap scripts

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/5906

Tested manually on OSX and Windows

### DIFF
--- a/libs/util/src/mill/util/MainModule.scala
+++ b/libs/util/src/mill/util/MainModule.scala
@@ -73,7 +73,9 @@ trait MainModule extends RootModule0, MainModuleApi {
       }
 
       if (results.isEmpty) {
-        Result.Failure(s"Failed to download Mill $version bootstrap scripts. Please check that version '$version' exists.")
+        Result.Failure(
+          s"Failed to download Mill $version bootstrap scripts. Please check that version '$version' exists."
+        )
       } else {
         Task.log.info(s"Successfully updated Mill bootstrap scripts to version $version")
         Result.Success(results)


lefou:
I find the name misleading, since it does not update the selected Mill version (in `//| mill-version` header or `.mill-version` files). I propose the name `updateMillScript` or `updateMillExe`.

We have some logic to find the latest stable version in the shell script. It should be straight forward to apply it in the command to "guess" the best (= newest) version without requiring the user to provide it.

lefou:
Since changing the command signature later is not so easy, we might as well add a `location` parameter, so the user can also use it when the want to update their `millw` script (still common) or update a shared installation (e.g. `$HOME/bin/mill`).

website/docs/modules/ROOT/pages/cli/installation-ide.adoc
```diff
@@ -476,6 +487,11 @@ or
 to override the Mill version manually. This takes precedence over the version
 specified in `./mill`, `.config/mill-version` or `.mill-version`
 
+In general, we recommend you manage your Mill version via your `./mill` bootstrap scripts and
+update it via `./mill updateMillScripts`. That is the most reliable approach, and avoids issues
+that may be caused by incompatibilities between your `./mill` script and the `mill-version`
+you override.
+
```
lefou, line 494: Personally, I don't like to manage the version in the script, and I would not recommend it to anyone. I think the script should change rarely and should be bit-wise identical to the downloadable version of the script. Especially, since I typically don't use the local script at all, even if there is one, I prefer to always use my installed version in a shared location (e.g. `$HOME/bin`). I would recommend to always manage the Mill version via a config property.

PR #6493, opened 2026-01-05
Streamline using Mill as sole source of Java installation

lihaoyi:
- Make common JDK binaries like `java`, `javac`, `javap`, etc. available on `JavaModule` and `MainModule`
- Put the `javaHome`'s `bin/` folder on the `PATH` in `allForkEnv` so the Mill `javaHome` is used when shelling out to JDK binaries
- Replace `jvmId` with `jvmVersion`, for consistency with `mill-jvm-version` and the other `{java,scala,...}Version` tasks. `jvmId` remains for backwards compat and anyone setting it should still work since `jvmVersion` forwards to it by default

Covered by additions to the `1-custom-jvms` example tests

### DIFF
--- a/example/fundamentals/javahome/1-builtin-jvm/build.mill
+++ b/example/fundamentals/javahome/1-builtin-jvm/build.mill
@@ -1,5 +1,6 @@
-// Mill comes bundled with a JDK (default `{default-mill-jvm-version}`) that it uses for itself that also acts as the default JDK
-// for any modules you define. You can use this JDK yourself via the builtin command
+// Mill comes with a managed JVM (default `{default-mill-jvm-version}`) that it
+// uses for itself that also acts as the default JVM
+// for any modules you define. You can use this JVM yourself via the builtin command
 // `./mill java`, `javac`, `javap`, `jps`, `jstack`, etc.:
 
 /** Usage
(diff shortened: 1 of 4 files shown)


libs/daemon/client/src/mill/client/lock/Locks.java
```diff
@@ -16,7 +16,7 @@ public Locks(Lock launcherLock, Lock daemonLock) {
   }
 
   /**
-   * @deprecated Use {@link #forDirectory(String)} instead, which automatically
+   * @deprecated Use `forDirectory` instead, which automatically
```
lefou, line 19: Why did you remove these links?

lefou:
My initial feeling about this is, it bloates the `MainModule` a lot. As an alternative, we could just add a single `jdkTool` task, accepting the tool name as first argument. It's help text could suggest the most common tools.

libs/daemon/client/src/mill/client/lock/Locks.java
lihaoyi: They were causing problems in publishing due to javadoc errors, not sure why

lefou: Could be this is fixed by ~~Scala 3.8.0-RC5~~ Zinc 2.0.0-M11. I think I read something about it.

* https://github.com/sbt/zinc/releases/tag/v2.0.0-M11
* https://github.com/sbt/zinc/pull/1621

PR #6497, opened 2026-01-05
Update zinc to 2.0.0-M11

lefou:


lefou:
CI error for `integration.migrating.init.packaged.daemon.testForked` can't be reproduced locally.

Issue #5091, opened 2025-05-09
Absolute paths in `package.mill` hinder isolated compilation of nested submodules


Yan-Muzi:
The package declaration requirements introduced in Mill 0.12.x (#3426) create conflicts in projects with nested submodules that need to function both as project dependencies and standalone components.

Take [XiangShan](https://github.com/OpenXiangShan/XiangShan) as an example. We have a submodule `rocket-chip`, which contains `cde` as its submodule. When `cde` is used in XianShan, `cde` directory requires `package build.rocket-chip.cde`. However, when we want to develop `rocket-chip` independently, the same `cde` directory now requires `package build.cde`.

Is there any ways to fix or bypass this?

lihaoyi:
@Yan-Muzi we don't have any first-class support for the scenario you describe. We could probably come up with something e.g. what @lefou discussed.

Yan-Muzi:
Am I correct in understanding that I may configure the top-level package as either `XiangShan` or `rocket-chip` within `cde`? This sounds OK but is rather inelegant as a submodule should not be aware of its containing directory.

If relative paths are not acceptable, is it possible to define all the modules in the root directory?

lihaoyi:
@Yan-Muzi it is possible to define all the modules in the root directory `build.mill`. Could you explain a bit how that would help?

lihaoyi:
@Yan-Muzi got it. I think we may have some sanity-check assertions that yell at you if you have a `build.mill` in a sub-folder, because it's a pretty easy mistake to make when people should create a `package.mill` folder instead. But maybe we can offer some way to silence that error

KireinaHoro:
Would like to give this a bump since right now our stuff are stuck on Mill `0.11.x` due to this.

I've been thinking about how to repurpose the [multi-file builds](https://mill-build.org/mill/large/multi-file-builds.html) support and ask the "dual use" projects (e.g. `rocket-chip` in @Yan-Muzi's example) to include most of its build logic as traits inside `package.mill`, while exporting the necessary module (object)s inside `build.mill` when they are invoked directly.

For your reference, this is how the submodule use case used to work; take [SpinalHDL](https://github.com/KireinaHoro/SpinalHDL/blob/7f7043702b1aa90f34cd7bdad86028c00d1a6653/build.sc#L81-L91) as an example, the build logic for the library is first defined as a trait (`Lib`) and then exported locally as an object (`lib`) for tasks like publishing to Maven:

```scala
[...]

PR #6504, opened 2026-01-06
Support nested build files with relaxed package naming 

Emin017:
This is an attempt to support #5091

---

- Add `allNestedBuildFileNames` to `CodeGenConstants` to include both `package.*` and `build.*` variants, enabling recognition of both types as nested build files.
- Update the code generation and file import logic in `CodeGen.scala` and `FileImportGraph.scala` to use `allNestedBuildFileNames` for identifying nested build files and allowing relaxed package declarations for these files.

---

## Relaxed Package Declarations

This update introduces **relaxed package naming** for nested mill build files, enabling the same build definition file to be used in different scenarios (as an independent project or as a reusable submodule).

Commit: 0636620ee0c

| File Type | Allowed Packages | Disallowed Packages | Examples |
|-----------|------------------|---------------------|----------|
| Root build.mill | Empty package (package build), Exact match | Relaxed match | `package build`, `package build.foo` |
| Nested build.mill | Empty package (supports independent use), Relaxed match, Exact match | None | `package build`, `package build.foo`, `package build.bar` |
| Nested package.mill | Relaxed match, Exact match | Empty package | `package build.deps`, `package build.foo.bar` |
| Other helper files | Relaxed match, Exact match | Empty package | `package build.helpers` |

### Relaxed Match Definition

A **relaxed match** occurs when the import path is either:
- A package name starting with `build*` (i.e., `package build`)
- Or if the expected path starts with the package declaration (e.g., `package foo`, and we can import `foo.bar.baz`)

```scala
val relaxedMatch =
  importSegments.startsWith("build") ||
  expectedImportSegments.startsWith(importSegments + ".")
```

This means that any package name in the form of `package build*` is valid, and we can also expect an import path like `foo.bar.baz` when we have `package foo`.

---

## Usage Example

We can now import build configuration files from subdirectories:

```scala
// build.mill
// Imports FooModule from the deps subdirectory
package build

import mill.*, scalalib.*

trait BarModule extends build.deps.foo.FooModule
```

```scala
// deps/package.mill
package build.deps
```

```scala
// deps/foo/build.mill
package build

import mill.*, scalalib.*

object `package` extends FooModule
```
```scala
// deps/foo/module.mill
package build

import mill.*, scalalib.*

trait FooModule extends ScalaModule { // trait definition in Helper Files
  def scalaVersion = "2.13.16"
}
```

If the subdirectory imports nested build files, this is also supported:

```scala
// build.mill
// Imports FooModule from the deps subdirectory
package build

import mill.*, scalalib.*

trait BarModule extends build.deps.foo.FooModule
```

```scala
// deps/package.mill
package build.deps
```

```scala
// deps/foo/build.mill
package build

import mill.*, scalalib.*

// Foo has its own helper files
import build.project.FooHelper // importing helper this way is now supported

object `package` extends FooModule
```
```scala
// deps/foo/module.mill
package build

import mill.*, scalalib.*

trait FooModule extends ScalaModule { // trait definition in Helper Files
  def scalaVersion = "2.13.16"
}
```

```scala
// deps/foo/project/package.mill
package build.project
```

```scala
// deps/foo/project/version.mill
trait FooHelper {
  // Something here
}
```
###  Limitation

This PR **does not support importing from nested `build.mill`**. Nested `build.mill` continues to be usable as an entrypoint when running Mill within that subdirectory, but reuse across directories is achieved by importing helper `.mill` files.

---

I believe that if this feature is merged, it will help smooth the migration of several open-source projects (such as [Chisel](https://github.com/chipsalliance/chisel), [SpinalHDL](https://github.com/SpinalHDL/SpinalHDL), [rocket-chip](https://github.com/chipsalliance/rocket-chip), and [Xiangshan](https://github.com/OpenXiangShan/XiangShan)) to the new version of Mill, as they currently rely on this functionality in the older version of Mill.

### DIFF
--- a/core/internal/src/mill/internal/Util.scala
+++ b/core/internal/src/mill/internal/Util.scala
@@ -369,7 +369,14 @@ object Util {
               case Some(failure) =>
                 // If there is an associated `Result.Failure` with no prefix, set the prefix to the
                 // current `keyPrefix` and prefix `error` with `key`
-                if (failure.tickerPrefix == "")
+                if (
+                  failure.tickerPrefix == "" && failure.path == null &&
+                  failure.index < 0 &&
+                  failure.error.startsWith("[error]") &&
+                  failure.error.contains("package.mill:")
+                )
+                  failure
+                else if (failure.tickerPrefix == "")
                   failure.copy(error = s"$key ${failure.error}", tickerPrefix = keyPrefix)
                 // If there is an associated `Result.Failure` with its own prefix, preserve it
                 // and chain together a new `Result.Failure` entry representing the current key


inochisa:
Hi @Emin017, I have seen your PR. Great work. And I have tested on my branch, but got a dependency error:

```
# ./mill -i dist.run ../misc-ips -i compile

56/69, 1 failed, completed] ============================== compile -explain-cyclic ============================== 2s
build.mill-56] [error] generatedScriptSources Package declaration "package build" in ext/SpinalHDL/package.mill does not match folder structure. Expected: "package build.ext.SpinalHDL"
Package declaration "package build.project" in ext/SpinalHDL/project/Version.mill does not match folder structure. Expected: "package build.ext.SpinalHDL.project"
Package declaration "package build.project" in ext/SpinalHDL/project/package.mill does not match folder structure. Expected: "package build.ext.SpinalHDL.project"
8850/8850, 1 failed, completed] ============================== dist.run xxx/misc-ips -i compile -explain-cyclic ============================== 88s
8850] [error] dist.run dist.run failed with an exception. Result of xxx/mill/out/dist/launcher.dest/run…: 1
```

My repo is at https://github.com/project-inochi/misc-ips/tree/mill-subproject

In fact, I am also trying to resolve this, but got lots of cycle dependency.

Emin017:
> My repo is at https://github.com/project-inochi/misc-ips/tree/mill-subproject
> 

@inochisa  Could you make this repo public? That way, we can see your specific configuration and debug more effectively.

inochisa:
@Emin017 My fault, now it is opened.

Emin017:
@inochisa  I took a look at your repository structure, and I have a general understanding of how the issue is occurring. We may still need to modify the configuration mechanism for searching subfolders and the assertion check mechanism in mill.

Previously, I only considered the following check and usage mechanism:
In the subfolders of our repository, there is only `build.mill` (without `package.mill`), and then we can import the traits from the sub-repository configuration into the `build.mill` in the root repository.

```scala
// build.mill
package build

import mill.*, scalalib.*

trait BarModule extends build_.deps.foo.FooModule // extends trait from deps/foo/build.mill
```
```scala
// deps/package.mill
package build.deps
```
```scala
// deps/foo/build.mill
package build

import mill.*, scalalib.*

trait FooModule extends ScalaModule {
  def scalaVersion = "2.13.16"
}

object `package` extends FooModule
```

However, looking at it now, I realize that I didn't fully consider this aspect. I'll think about making adjustments to this part. (We might also need more relaxed check rules.)

inochisa:
> @inochisa I took a look at your repository structure, and I have a general understanding of how the issue is occurring. We may still need to modify the configuration mechanism for searching subfolders and the assertion check mechanism in mill.
> 

Agreed.

> Previously, I only considered the following check and usage mechanism: In the subfolders of our repository, there is only `build.mill` (without `package.mill`), 

I think it is possible to remove this (at least in my repository). I added it as mill always complaint it is not a package.

> and then we can import the traits from the sub-repository configuration into the `build.mill` in the root repository.
> 
> ```scala
> // build.mill
> package build
> 
> import mill.*, scalalib.*
> 
> trait BarModule extends build_.deps.foo.FooModule // extends trait from deps/foo/build.mill
> ```

I also consider this, but I found it is not nature to do such a thing. Using internal `build_` is not documented and confused. In my view, it is more nature to use something like `build.deps.foo.FooModule`

> 
> ```scala
> // deps/package.mill
> package build.deps
> ```
> 
> ```scala
> // deps/foo/build.mill
> package build
> 
> import mill.*, scalalib.*
> 
> trait FooModule extends ScalaModule {
>   def scalaVersion = "2.13.16"
> }
> 
> object `package` extends FooModule
> ```
> 
> However, looking at it now, I realize that I didn't fully consider this aspect. I'll think about making adjustments to this part. (We might also need more relaxed check rules.)

In this example, the SpinalHDL has its own build lib. This is not resolved in a right way, as mill will generate `build_` class for nested build file, which will cause a loop for package import. (This is the next step after this error).

Emin017:
> In this example, the SpinalHDL has its own build lib. This is not resolved in a right way, as mill will generate `build_` class for nested build file, which will cause a loop for package import. (This is the next step after this error).

@inochisa I think Mill’s CodeGen will rewrite package name prefixes, which should prevent loops from occurring (at least in this case). Could you explain in more detail when the loop you mentioned for package imports would arise?

Emin017:
@inochisa I think I roughly understand what you mean. I’ve pushed some new commits, and you can try it based on that. The new support I added includes:

1. Added detection for the case where `build.mill` and `package.mill` coexist in the same directory, since this can lead to duplicate `package_` definitions during code generation. However, I’m not entirely sure whether Mill is supposed to support having both `build.mill` and `package.mill` present at the same time (I haven't seen this usage in the example yet), this probably needs to be confirmed with @lihaoyi.

2. By modifying `codeGen`, I added support for `build.mill` files in subdirectories to import nested configuration files themselves. For such nested `build.mill` files, a local `build` object is generated. This object “shadows” the root-level `build` alias, allowing local references like `import build.foo.Bar` to work correctly.
```scala
// deps/foo/build.mill
// This is build.mill in subdir, but also import trait from its subdir
package build

import build.foo.Bar
trait fooModule extends Bar
```
```scala
// deps/foo/bar/package.mill
package build.bar

trait Bar {
// Do something here
}
```

I’m not sure whether there are other aspects I may have overlooked. Please feel free to point them out after testing :)

inochisa:
Thanks for your update @Emin017 

Mostly, I think it is works, but I got an weird error for the build. I guess it may not related to this build and it is a new problem? I have found a similar bug here at sbt/sbt#7777. @lihaoyi, could you take a look for it? I have attached the log below.

https://gist.github.com/inochisa/aa362832d7736df7b4d427ac26841a0a

lihaoyi:
@inochisa can you try wrapping everything in an explicit `object package extends mill.Module{...}`? The last time I hit an error like that wrapping things made it go away, even though I don't really understand why

inochisa:
@lihaoyi Yeah, by adding this, it is resolved, and I got an error related to this PR. It seems like this is related to the package generation.

https://gist.github.com/inochisa/ad1353ea0d2b2a28e72f5ba2fc84fc90

Emin017:
> @lihaoyi Yeah, by adding this, it is resolved, and I got an error related to this PR. It seems like this is related to the package generation.
> 
> https://gist.github.com/inochisa/ad1353ea0d2b2a28e72f5ba2fc84fc90

@inochisa Could you sync your latest configuration to the repository? That way, I can reproduce the issue and debug.

inochisa:
@Emin017 I have updated the repository by force push.
And my test command is `./mill -i dist.run <path>/misc-ips -i compile -explain-cyclic`

Emin017:
@inochisa Can you try the latest commit? I think the previous error should have been fixed.
```shell
❯ ./mill dist.run misc-ips/ -i compile -explain-cyclic
8850] dist.run
Mill version SNAPSHOT is different than configured for this directory!
Configured version is 1.0.6 (/home/Emin/Workbench/mill/misc-ips/.mill-version)
============================== compile -explain-cyclic ==============================
build.mill-64] compile compiling 10 Scala sources to out/mill-build/compile.dest/classes ...
build.mill-64] [warn] ext/SpinalHDL/project/Version.mill:5:25
build.mill-64] import scala.collection.JavaConverters._
build.mill-64]                         ^^^^^^^^^^^^^^
build.mill-64] object JavaConverters in package scala.collection is deprecated since 2.13.0: Use `scala.jdk.CollectionConverters` instead
build.mill-64] 
build.mill-64] [warn] one warning found
build.mill-64] done compiling
69/69, completed] ========================== compile -explain-cyclic ========================== 8s
[error] Cannot resolve compile. Try `mill resolve _`, `mill resolve __.compile` to see what's available, or `mill __.compile` to run all `compile` tasks
8850/8850, 1 failed, completed] ============================== dist.run misc-ips/ -i compile -explain-cyclic ============================== 9s
8850] [error] dist.run dist.run failed with an exception. Result of /home/Emin/Workbench/mill/out/dist/launcher.dest/run…: 1
```

inochisa:
@Emin017 Confirmed

```bash
# clear && ./mill -i dist.run xxx/misc-ips -i hw.runMain io.inochisa.gen.clock.ClockDetectedVerilog

8850] dist.run
Mill version SNAPSHOT is different than configured for this directory!
Configured version is 1.0.6 (xxx/misc-ips/.mill-version)
build.mill-64] compile compiling 1 Scala source to out/mill-build/compile.dest/classes ...
build.mill-64] done compiling
build.mill-64] compiling 5 Scala sources to out/mill-build/compile.dest/classes ...
build.mill-64] [warn] ext/SpinalHDL/project/Version.mill:5:25
build.mill-64] import scala.collection.JavaConverters._
build.mill-64]                         ^^^^^^^^^^^^^^
build.mill-64] object JavaConverters in package scala.collection is deprecated since 2.13.0: Use `scala.jdk.CollectionConverters` instead
build.mill-64] 
build.mill-64] [warn] one warning found
build.mill-64] done compiling
378] hw.runMain
[Runtime] SpinalHDL dev    git head : 0a2bd7bc67e6ce86c1231386c18cfd8c216919d1
[Runtime] JVM max memory : 30208.0MiB
[Runtime] Current date : 2026.01.09 12:19:05
[Progress] at 0.000 : Elaborate components
[Progress] at 0.016 : PhaseCreateComponent
[Progress] at 0.284 : PhaseDummy
[Progress] at 0.284 : Checks and transforms
[Progress] at 0.284 : PhaseMemBlackBoxingDefault
[Progress] at 0.288 : PhaseDeviceSpecifics
[Progress] at 0.288 : PhaseApplyIoDefault
[Progress] at 0.290 : PhaseNameNodesByReflection
[Progress] at 0.292 : PhaseCollectAndNameEnum
[Progress] at 0.296 : PhaseCheckIoBundle
[Progress] at 0.298 : PhaseCheckHierarchy
[Progress] at 0.301 : PhaseAnalog
[Progress] at 0.303 : PhaseNextifyReg
[Progress] at 0.304 : PhaseRemoveUselessStuff
[Progress] at 0.312 : PhaseRemoveIntermediateUnnameds
[Progress] at 0.316 : PhasePullClockDomains
[Progress] at 0.319 : PhaseInferEnumEncodings
[Progress] at 0.327 : PhaseInferWidth
[Progress] at 0.329 : PhaseNormalizeNodeInputs
[Progress] at 0.333 : PhaseRemoveIntermediateUnnameds
[Progress] at 0.333 : PhaseSimplifyNodes
[Progress] at 0.337 : PhaseCompletSwitchCases
[Progress] at 0.337 : PhaseRemoveUselessStuff
[Progress] at 0.338 : PhaseRemoveIntermediateUnnameds
[Progress] at 0.339 : PhaseCheck_noLatchNoOverride
[Progress] at 0.348 : PhaseCheck_noRegisterAsLatch
[Progress] at 0.349 : PhaseCheckCombinationalLoops
[Progress] at 0.353 : PhaseCheckCrossClock
[Progress] at 0.357 : PhasePropagateNames
[Progress] at 0.368 : PhaseObfuscate
[Progress] at 0.368 : PhaseAllocateNames
[Progress] at 0.374 : PhaseDevice
[Progress] at 0.375 : PhaseGetInfoRTL
[Progress] at 0.376 : PhaseDummy
[Progress] at 0.377 : Generate Verilog to gen
[Progress] at 0.377 : PhaseVerilog
[Done] at 0.417
378/378, completed] ============================== hw.runMain io.inochisa.gen.clock.ClockDetectedVerilog ============================== 19s
8850/8850, completed] ============================== dist.run xxx/misc-ips -i hw.runMain io.inochisa.gen.clock.ClockDetectedVerilog ============================== 25s

```

inochisa:
@Emin017 I can also confirmed the change for adding Module for every object is not needed anymore.

Emin017:
CI Status link: [runs/20844288189](https://github.com/Emin017/mill/actions/runs/20844288189)

The relevant tests have passed, and most tests on CI have also passed. The only failure is the `"example.androidlib.__.packaged.daemon"` test (#6522 which seems unrelated to this change). Therefore, I am marking this PR as ready for review.

EDIT: I have also updated the PR description.

lefou:
> @Emin017 I have updated the repository by force push. And my test command is `./mill -i dist.run <path>/misc-ips -i compile -explain-cyclic`

The `compile` task doesn't accept parameters. You need to add `-explain-cyclic` to your `scalacOptions` by overriding `def scalacOptions`.

inochisa:
> > @Emin017 I have updated the repository by force push. And my test command is `./mill -i dist.run <path>/misc-ips -i compile -explain-cyclic`
> 
> The `compile` task doesn't accept parameters. You need to add `-explain-cyclic` to your `scalacOptions` by overriding `def scalacOptions`.

Thanks, it is useless. And I just did a copy paste from the original test, so please ignore this thing.

lihaoyi:
I think this looks great,  could you include the example page in the relevant `.adoc` file and run `website.fastPages` to make sure the documentation for this feature is readable and makes sense

example/large/multifile/12-import-subdir/build.mill
```diff
@@ -0,0 +1,7 @@
+// build.mill
```
lihaoyi: Can you make sure this test covers all cases mentioned in the PR description

- nested build.mill
- nested package.mill
- other helper files

Emin017: This has been added in a new pushed commit, including the example documentation.

example/large/multifile/12-import-subdir/build.mill
```diff
@@ -0,0 +1,50 @@
+// Mill supports nested `build.mill` files in subdirectories. Unlike `package.mill` files
```
lihaoyi: Let's make this opt-in via a flag `//| allowNestedBuildMillFiles: true`? That would help make `package.mill` the default style, while allowing nested `build.mill` style for the people who need it.

Please mention that this is mostly for compatibility with isolated sub-builds, but has the downside of making IDE support less precise

example/large/multifile/12-import-subdir/build.mill
```diff
@@ -0,0 +1,50 @@
...
+// they are located.
+//
+// This makes it easy to:
+//
+// * Keep module definitions self-contained in subdirectories that can be built independently
+// * Refactor or move modules around without updating package declarations
+//
+// === Example Project
+// The `FooModule` trait is defined in the nested `deps/foo/build.mill` file, and we can reference it
+// using the hierarchical path `build_.deps.foo.FooModule`
```
lihaoyi: Can we make this work without the `build_` prefix? In normal `build.mill`/`package.mill` files you generally never need to write `build_`, and we should try to provide the same experience here

example/large/multifile/12-import-subdir/build.mill
Emin017: This has been added in a new pushed commit.

example/large/multifile/12-import-subdir/build.mill
Emin017: I updated the code generation logic to support importing `build.deps.foo.FooModule` (instead of `build_.deps.foo.FooModule`).

**Previously generated file**
```scala
// Generated by Mill ...
package build_.deps.foo

import build_.{package_ => build}

import MillMiscInfo.*
import _root_.mill.util.TokenReaders.given
import _root_.mill.api.JsonFormatters.given

object package_ extends package_ {
  
  export build_.deps.foo.*
  
}

//SOURCECODE_ORIGINAL_FILE_PATH=.../deps/foo/build.mill
//SOURCECODE_ORIGINAL_CODE_START_MARKER
import mill.*, scalalib.*abstract class package_
    extends _root_.mill.api.internal.SubfolderModule(build.millDiscover), FooModule
```
**Newly generated file**
```scala
// Generated by Mill ...
package build_.deps.foo

import build_.{package_ => build}

import MillMiscInfo.*
import _root_.mill.util.TokenReaders.given
import _root_.mill.api.JsonFormatters.given

object _defs {
  import mill.*, scalalib.*
  
  trait FooModule extends ScalaModule {
    def scalaVersion = myScalaVersion
  }
}
export _defs._

object package_ extends package_ {
  export _defs._
  
  export build_.deps.foo.*
  
}

//SOURCECODE_ORIGINAL_FILE_PATH=.../deps/foo/build.mill
//SOURCECODE_ORIGINAL_CODE_START_MARKER
import mill.*, scalalib.*

trait FooModule extends ScalaModule {
  def scalaVersion = myScalaVersion
}

abstract class package_
    extends _root_.mill.api.internal.SubfolderModule(_root_.build_.package_.millDiscover), FooModule
```

---

**Old code generation flow**

The previous code generation pipeline now will produce invalid output such as:
```scala
import mill.*, scalalib.*abstract class package_
```
which is a syntax error.

**New code generation flow**

The new pipeline works as follows:
1. Detect whether there is any code between `lastImportEnd` and `packageObjectStart`
    - This required extending `MillScalaParser` to support the detection.
2. If such code exists:
    - Create a `_defs` wrapper:
```
baseHeader
+ object _defs { top-level definitions }
+ export _defs._
+ object package_ { export _defs._ }
+ wrapperTail
```
3. During concatenation:
    - Only keep imports in `importsPrefix`
    - Append `rewrittenScriptCode.substring(packageObjectStart)`

This results in:
```
import mill.*, scalalib.*

trait FooModule ...

abstract class package_ ...
```
which is syntactically correct.

---

However, these implementations have changes in many places across the repository, and I’m not sure whether they might introduce other issues. By comparison, supporting imports only from `build_.deps.foo` would be much simpler.

Emin017: https://github.com/Emin017/mill/actions/runs/20895535833

I think this is now fully implemented, so I'll convert this PR from draft to ready for review.

integration/failure/root-module-compile-error/src/RootModuleCompileErrorTests.scala
```diff
@@ -14,65 +14,67 @@ object RootModuleCompileErrorTests extends UtestIntegrationTestSuite {
...
         ))
       }
 
       locally {
-        assert(res.err.contains("""build.mill:8:22"""))
+        assert(res.err.contains("""build.mill:5:22"""))
         assert(res.err.contains("""Not found: unknownRootInternalDef"""))
         assert(res.err.contains("""def scalaVersion = unknownRootInternalDef"""))
-        assert(res.err.contains("""                   ^^^^^^^^^^^^^^^^^^^^^^"""))
+        assert(res.err.contains("""                     ^^^^^^^^^^^^^^^^^^^^^^"""))
```
lihaoyi: Why did these error messages change? Are the new errors correct and this is an improvement, or are the old errors correct and this is a regression?

integration/feature/inspect/src/InspectTests.scala
```diff
@@ -195,8 +195,7 @@ object InspectTests extends UtestIntegrationTestSuite {
       val core3Inspect = out("inspect").json.str
       assertGoldenLiteral(
         core3Inspect,
-        """core3(core3/package.mill:6)
-          |    Subfolder Module Scaladoc
```
lihaoyi, line 199: Why did the line number change and why did the scaladoc disappear?

runner/meta/src/mill/meta/FileImportGraph.scala
```diff
@@ -56,28 +56,64 @@ object FileImportGraph {
...
 
+        val allowNestedBuildMillFiles =
+          if (useDummy) false
+          else {
+            try {
+              val yamlHeader = mill.constants.Util.readBuildHeader(s.toNIO, s.last, true)
+              yamlHeader.contains("allowNestedBuildMillFiles") && (
+                yamlHeader.contains("allowNestedBuildMillFiles: true") ||
+                  yamlHeader.contains("allowNestedBuildMillFiles:true")
+              )
```
lihaoyi: This is not the correct way to parse data out of the YAML header

runner/meta/src/mill/meta/CodeGen.scala
```diff
@@ -358,75 +361,128 @@ object CodeGen {
...
+        s"lazy val ${backtickWrap(c)} = ${CGConst.wrapperObjectName}.${backtickWrap(c)}"
+      ).mkString("\n  ")
+      s"""
+         |object build {
+         |  $childRefs
+         |}
+         |""".stripMargin
+    } else ""
+
+    val baseHeader =
```
lihaoyi: Why are all these invasive changes to CodeGen.scala necessary? I would have thought you'd just turn off the "error on wrong package name" logic, add some logic to recognize both `package.mill` and `build.mill` files, and everything else would be unchanged

lihaoyi:
Overall the implementation looks like it broke some error reporting positions, and also seems far more invasive and complex that I would expect. Please try to minimize the necessary change and avoid adding all these extra codgen for what should be a relatively small tweak in error reporting

runner/meta/src/mill/meta/CodeGen.scala
Emin017: Sorry, I didn’t expect there to be so many invasive changes either. Originally, importing via `build_.deps` really didn’t require this much work, but if we want to support `import build.deps`, it seems we have to pay a high cost. I suggest that for now we only support importing through `build_`.

lihaoyi: Why does supporting `import build.deps` require a high cost when it already works in `package.mill` files?

Emin017: Okay, let me rethink these issues. I’ll consider redesigning these.

Emin017: > Why does supporting `import build.deps` require a high cost when it already works in `package.mill` files?

It seems that if we support `import build.deps`, it will only import members from `build_.deps.foo.package_`. Since the top-level definitions in `build.mill` are not in `package_`, they are not visible. `package.mill` can support this usage because its definitions are originally inside `object package` (which is rewritten into `package_`).

```shell
build.mill-64] compile compiling 8 Scala sources to out/mill-build/compile.dest/classes ...
build.mill-64] [error] out/mill-build/generatedScriptSources.dest/wrapped/build_/deps/foo/build.mill:13:29
build.mill-64]   export _root_.build_.deps.foo.*
build.mill-64]                             ^^^
build.mill-64] Implementation restriction: package build_.deps.foo is not a valid prefix for a wildcard export, as it is a package
build.mill-64]
build.mill-64] [error] out/mill-build/generatedScriptSources.dest/wrapped/build_/deps/foo/build.mill:14:36
build.mill-64]   export build_.deps.foo.versions_.*
build.mill-64]                                    ^
build.mill-64] Conflicting definitions:
build.mill-64] final def myScalaVersion: String in object package_ at line 13 and
build.mill-64] final def myScalaVersion: String in object package_ at line 14
build.mill-64] have the same type (): String after erasure.
build.mill-64]
build.mill-64] Consider adding a @targetName annotation to one of the conflicting definitions
build.mill-64] for disambiguation.
build.mill-64]
build.mill-64] [error] two errors found
64/69, 1 failed, completed] ============================== __.compile ============================== 3s
build.mill-64] [error] compile Compilation failed
```

Emin017:
@lihaoyi 
Hi, I dropped support for importing traits from a nested `build.mill`, because supporting that seems to require a lot of invasive changes.

The commits I just pushed implement a narrower scope: **only importing modules from nested helper `.mill` files**. Implementing just this support wouldn’t require many invasive changes. And even with just this capability, I believe it can still help (or unblock) migrating these open-source projects to the newer Mill version, since it covers the core reuse workflow they depend on.

Also, I’ve updated the PR description accordingly.

lihaoyi:
@Emin017 @inochisa I have put together an alternative implementation over at https://github.com/com-lihaoyi/mill/pull/6655, please help me review it and try it out

Emin017:
Superseded by #6655.

PR #6507, opened 2026-01-06
Scripting article

lihaoyi:


lefou:
Shouldn't we mention, that Mill also allows scripting for Groovy comparable to Java/Scala/Kotlin?

PR #6523, opened 2026-01-07
Debug Scala.js Unit Test Hang

lihaoyi:


lihaoyi:
For now just commented out that specific hanging test in https://github.com/com-lihaoyi/mill/commit/f5fa477271424088153a8806a356951808674cae. Not sure why it cannot repro on these standalone PRs

PR #6525, opened 2026-01-07
Move `.bsp/mill-bsp-out` to `.bsp/out`

lihaoyi:
repeating `bsp` is redundant, as is `mill` since we already have a convention that `out/` folders belong to Mill

### DIFF
--- a/integration/ide/bsp-server-reload/resources/snapshots/broken/start/workspace-build-targets.json
+++ b/integration/ide/bsp-server-reload/resources/snapshots/broken/start/workspace-build-targets.json
@@ -2,10 +2,10 @@
   "targets": [
     {
       "id": {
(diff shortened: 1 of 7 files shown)


lefou:
`.bsp` is used by all BSP servers, not just Mill. Hence we chose `mill-bsp-out` as it starts with the same prefix as the connection file `mill-bsp.json`, to make clear which BSP server this directory belongs to.

lihaoyi:
We used `mill-bsp-out` because it was inside `out/`, and most internal stuff in `out/` is prefixed with `mill-` to avoid conflicting with user-generated tasks and modules that may have arbitrary names. Inside `.bsp/` that constraint doesn't apply, and there aren't that many BSP servers out there so we can probably call it whatever we want without worrying about conflicts

lefou:
I was not so much worrying about conflicts, but about clarity for the users. But `.bsp/out/` is fine.

PR #6533, opened 2026-01-07
Native: minor fix on image options when no matching repository data

vaslabs:
Fixes no such file or directory exception if no dependencies or no matches for graalvm metadata


vaslabs:
fixed in https://github.com/com-lihaoyi/mill/pull/6547 as well, closing this one

PR #6537, opened 2026-01-08
Add `resolvedMvnDepsSources` task to simplify looking up source files

lihaoyi:
This should help both humans and AIs quickly find the source code of the libraries on their classpath, so they can use them without needing to dig through random github repos in the browser and trying to guess at the right version

### DIFF
--- a/libs/tabcomplete/test/src/mill/tabcomplete/TabCompleteTests.scala
+++ b/libs/tabcomplete/test/src/mill/tabcomplete/TabCompleteTests.scala
@@ -361,15 +361,15 @@ object TabCompleteTests extends TestSuite {
             "--interactive             Run Mill in interactive mode, suitable for opening REPLs and taking user input. Identical to --no-daemon. Must be the first argument.",
             "--no-build-lock           Evaluate tasks / commands without acquiring an exclusive lock on the Mill output directory",
             "--tab-complete            Runs Mill in tab-completion mode",
-            "--color                   <bool> Toggle colored output; by default enabled only if the console is interactive or FORCE_COLOR environment variable is set, and NO_COLOR environment variable is not set",
             "--no-daemon               Run without a long-lived background daemon. Must be the first argument.",
-            "--ticker                  <bool> Enable or disable the ticker log, which provides information on running tasks and where each log line came from",
             "--allow-positional        Allows command args to be passed positionally without `--arg` by default",
             "--watch                   Watch and re-run the given tasks when when their inputs change.",
             "--no-wait-for-build-lock  Do not wait for an exclusive lock on the Mill output directory to evaluate tasks / commands.",
             "--bsp                     Enable BSP server mode. Typically used by a BSP client when starting the Mill BSP server.",
+            "--help-advanced           Print a internal or advanced command flags not intended for common usage",
             "--import                  <str> Additional ivy dependencies to load into mill, e.g. plugins.",
             "--meta-level              <int> Select a meta-level to run the given tasks. Level 0 is the main project in `build.mill`, level 1 the first meta-build in `mill-build/build.mill`, etc. If negative, -1 means the deepest meta-build (boostrap build), -2 the second deepest meta-build, etc.",
+            "--jshell                  Open a JShell REPL with the classpath of the meta-level 1 build module (mill-build/). This is useful for interactively testing and debugging your build logic.",
             "--offline                 Try to work offline. This tells modules that support it to work offline and avoid any access to the internet. This is on a best effort basis. There are currently no guarantees that modules don't attempt to fetch remote sources.",
             "--keep-going              Continue build, even after build failures.",
             "--define                  <k=v> Define (or overwrite) a system property.",
@@ -380,10 +380,10 @@ object TabCompleteTests extends TestSuite {
             "--bsp-install             Create mill-bsp.json with Mill details under .bsp/",
             "--help                    Print this help message and exit.",
             "--jobs                    <str> The number of parallel threads. It can be an integer e.g. `5` meaning 5 threads, an expression e.g. `0.5C` meaning half as many threads as available cores, or `C-2` meaning 2 threads less than the number of cores. `1` disables parallelism and `0` (the default) uses 1 thread per core.",
+            "--ticker                  <bool> Enable or disable the ticker log, which provides information on running tasks and where each log line came from",
+            "--color                   <bool> Toggle colored output; by default enabled only if the console is interactive or FORCE_COLOR environment variable is set, and NO_COLOR is not set",
             "--version                 Show mill version information and exit.",
-            "--task                    <str> The name or a query of the tasks(s) you want to build.",
-            "--help-advanced           Print a internal or advanced command flags not intended for common usage",
-            "--jshell                  Open a JShell REPL with the classpath of the meta-level 1 build module (mill-build/). This is useful for interactively testing and debugging your build logic. Implies options `--meta-level 1` and `--no-server`."
+            "--task                    <str> The name or a query of the tasks(s) you want to build."
           )
         )
       }


libs/javalib/src/mill/javalib/JavaModule.scala
```diff
@@ -1091,6 +1077,32 @@ trait JavaModule
...
+   * Resolved dependency sources, unpacked into a single directory. Useful to quickly
+   * look up the sources of the dependencies on your classpath so you can find the
+   * exact source code you are compiling and running against.
+   */
+  def resolvedMvnSources: T[PathRef] = Task {
+    for (jar <- resolvedMvnDeps0(sources = true)()) {
+      val jarName = jar.path.last.stripSuffix(".jar")
+      os.unzip(jar.path, Task.dest / jarName)
+    }
+    println(s"Unpacked sources of transitive third-party dependencies into ${Task.dest.relativeTo(BuildCtx.workspaceRoot)} for browsing")
```
lefou, line 1102: Can we make this a `Task.log.debug`?

libs/util/src/mill/util/JdkCommandsModule.scala
```diff
@@ -17,38 +18,48 @@ trait JdkCommandsModule extends mill.api.Module {
...
+
   /**
    * Runs the `java` command from this module's [[jdkCommandsJavaHome]].
    * Renamed to `java` on the command line.
    */
   @Task.rename("java")
   @mainargs.main(name = "java")
   def javaRun(args: String*): Command[Unit] = Task.Command(exclusive = true) {
-    Jvm.callJdkTool("java", args, jdkCommandsJavaHome().map(_.path))
+    Task.ctx().systemExit(callJdk("java", jdkCommandsJavaHome(), args))
```
lefou, line 38: I don't think we should stop the Mill process. It is enough to signal the status code via the task success value. Same for the other occurences.

lefou: Also, `systemExit` without a reason parameter is already deprecated.

core/internal/cli/src/mill/internal/MillCliConfig.scala
```diff
@@ -10,6 +10,7 @@ case class MillCliConfig(
     @arg(name = "version", short = 'v', doc = "Show mill version information and exit.")
     showVersion: Flag = Flag(),
     @arg(
+      hidden = true,
```
lefou, line 13: Why is that?

core/internal/cli/src/mill/internal/MillCliConfig.scala
```diff
@@ -21,7 +22,7 @@ case class MillCliConfig(
            tasks and where each log line came from"""
     )
     ticker: Option[Boolean] = None,
-    @arg(name = "debug", short = 'd', doc = "Show debug output on STDOUT")
+    @arg(hidden = true, name = "debug", short = 'd', doc = "Show debug output on STDOUT")
```
lefou, line 25: Why is that hidden now? This option is useful. Why do we hide it from the user.

core/internal/cli/src/mill/internal/MillCliConfig.scala
```diff
@@ -113,6 +115,7 @@ case class MillCliConfig(
     )
     noWaitForBuildLock: Flag = Flag(),
     @arg(
+      hidden = true,
```
lefou, line 118: Please make this visible again.

libs/javalib/src/mill/javalib/JavaModule.scala
lihaoyi: I wanted to make it a `println` because this is probably going to be used manually, so having it print out the message is useful and shouldn't be hidden behind `--debug`

libs/util/src/mill/util/JdkCommandsModule.scala
lihaoyi: I wanted to propagate the existing JDK exit code, the task success/failure only lets you return 1 or 0 now and would stomp over any other code

lihaoyi:
The hidden options are still visible underunder `--help-advanced`, so they should be discoverable. But I wanted to keep `--help` short and only containing the most commonly used flags and limited to ~1 page of text on my laptop terminal, and over time it had grown beyond that so it needed to be trimmed down again

lefou:
I strongly think this is not what most users want. Those, not able to comprehend a `--help` output longer than a screen are probably using an AI to get the most important flags anyways. But users who still read themselves, have a harder time to use two CLI calls to get the info, in addition to visiting the docs page and other resources. There are so many tools that have rather long help output, I don't think we follow any example here.

lefou:
One compromise could be to have a `--help-all` (as e.g. all Qt/KDE applications have)  that shows all options, not just a sub-set.

lefou:
We could ditch the `--help-advanced` then.

lefou:
Also, if we want a shorter `--help`, we should just remove all the usage examples in favor of a man page.

lihaoyi:
`--help-all` works for me, I assume it would be `--help` and `--help-all` combined?

lefou:
IMHO, `--help` is classically reserved for listing all possible options, and we currently fail this simple task.

lihaoyi:
_"What is helpful for some newbie who needs help"_ and _"a comprehensive list of all possible flags"_ isn't always the same thing. The `--help` vs `--help-all` (or `--help-extra` or `--help -a`) distinction seems pretty common, I have lots of other CLI tools on my system have such a pattern. 
```console
lihaoyi ~$ python3 -help
usage: /opt/homebrew/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/Resources/Python.app/Contents/MacOS/Python [option] ... [-c cmd | -m mod | file | -] [arg] ...
Options (and corresponding environment variables):
-b     : issue warnings about converting bytes/bytearray to str and comparing
         bytes/bytearray with str or bytes with int. (-bb: issue errors)
-B     : don't write .pyc files on import; also PYTHONDONTWRITEBYTECODE=x
-c cmd : program passed in as string (terminates option list)
-d     : turn on parser debugging output (for experts only, only works on
         debug builds); also PYTHONDEBUG=x
-E     : ignore PYTHON* environment variables (such as PYTHONPATH)
-h     : print this help message and exit (also -? or --help)
-i     : inspect interactively after running script; forces a prompt even
         if stdin does not appear to be a terminal; also PYTHONINSPECT=x
-I     : isolate Python from the user's environment (implies -E and -s)
-m mod : run library module as a script (terminates option list)
-O     : remove assert and __debug__-dependent statements; add .opt-1 before
         .pyc extension; also PYTHONOPTIMIZE=x
-OO    : do -O changes and also discard docstrings; add .opt-2 before
         .pyc extension
-P     : don't prepend a potentially unsafe path to sys.path; also
         PYTHONSAFEPATH
-q     : don't print version and copyright messages on interactive startup
-s     : don't add user site directory to sys.path; also PYTHONNOUSERSITE=x
-S     : don't imply 'import site' on initialization
-u     : force the stdout and stderr streams to be unbuffered;
         this option has no effect on stdin; also PYTHONUNBUFFERED=x
-v     : verbose (trace import statements); also PYTHONVERBOSE=x
         can be supplied multiple times to increase verbosity
-V     : print the Python version number and exit (also --version)
         when given twice, print more information about the build
-W arg : warning control; arg is action:message:category:module:lineno
         also PYTHONWARNINGS=arg
-x     : skip first line of source, allowing use of non-Unix forms of #!cmd
-X opt : set implementation-specific option
--check-hash-based-pycs always|default|never:
         control how Python invalidates hash-based .pyc files
--help-env: print help about Python environment variables and exit
--help-xoptions: print help about implementation-specific -X options and exit
--help-all: print complete help information and exit

Arguments:
file   : program read from script file
-      : program read from stdin (default; interactive mode if a tty)
arg ...: arguments passed to program in sys.argv[1:]
```
```console
lihaoyi ~$ git --help
usage: git [-v | --version] [-h | --help] [-C <path>] [-c <name>=<value>]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p | --paginate | -P | --no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           [--super-prefix=<path>] [--config-env=<name>=<envvar>]
           <command> [<args>]

These are common Git commands used in various situations:

start a working area (see also: git help tutorial)
   clone     Clone a repository into a new directory
   init      Create an empty Git repository or reinitialize an existing one

work on the current change (see also: git help everyday)
   add       Add file contents to the index
   mv        Move or rename a file, a directory, or a symlink
   restore   Restore working tree files
   rm        Remove files from the working tree and from the index

examine the history and state (see also: git help revisions)
   bisect    Use binary search to find the commit that introduced a bug
   diff      Show changes between commits, commit and working tree, etc
   grep      Print lines matching a pattern
   log       Show commit logs
   show      Show various types of objects
   status    Show the working tree status

grow, mark and tweak your common history
   branch    List, create, or delete branches
   commit    Record changes to the repository
   merge     Join two or more development histories together
   rebase    Reapply commits on top of another base tip
   reset     Reset current HEAD to the specified state
   switch    Switch branches
   tag       Create, list, delete or verify a tag object signed with GPG

collaborate (see also: git help workflows)
   fetch     Download objects and refs from another repository
   pull      Fetch from and integrate with another repository or a local branch
   push      Update remote refs along with associated objects

'git help -a' and 'git help -g' list available subcommands and some
concept guides. See 'git help <command>' or 'git help <concept>'
to read about a specific subcommand or concept.
See 'git help git' for an overview of the system.
```
```console
lihaoyi ~$ javac -help
Usage: javac <options> <source files>
where possible options include:
  @<filename>                  Read options and filenames from file
  -Akey[=value]                Options to pass to annotation processors
  --add-modules <module>(,<module>)*
        Root modules to resolve in addition to the initial modules,
        or all modules on the module path if <module> is ALL-MODULE-PATH.
  --boot-class-path <path>, -bootclasspath <path>
        Override location of bootstrap class files
  --class-path <path>, -classpath <path>, -cp <path>
        Specify where to find user class files and annotation processors
  -d <directory>               Specify where to place generated class files
  -deprecation
        Output source locations where deprecated APIs are used
  --enable-preview
        Enable preview language features.
        To be used in conjunction with either -source or --release.
  -encoding <encoding>         Specify character encoding used by source files
  -endorseddirs <dirs>         Override location of endorsed standards path
  -extdirs <dirs>              Override location of installed extensions
  -g                           Generate all debugging info
  -g:{lines,vars,source}       Generate only some debugging info
  -g:none                      Generate no debugging info
  -h <directory>
        Specify where to place generated native header files
  --help, -help, -?            Print this help message
  --help-extra, -X             Print help on extra options
  -implicit:{none,class}
        Specify whether to generate class files for implicitly referenced files
  -J<flag>                     Pass <flag> directly to the runtime system
  --limit-modules <module>(,<module>)*
        Limit the universe of observable modules
  --module <module>(,<module>)*, -m <module>(,<module>)*
        Compile only the specified module(s), check timestamps
  --module-path <path>, -p <path>
        Specify where to find application modules
  --module-source-path <module-source-path>
        Specify where to find input source files for multiple modules
  --module-version <version>
        Specify version of modules that are being compiled
  -nowarn                      Generate no warnings
  -parameters
        Generate metadata for reflection on method parameters
  -proc:{none,only,full}
        Control whether annotation processing and/or compilation is done.
  -processor <class1>[,<class2>,<class3>...]
        Names of the annotation processors to run;
        bypasses default discovery process
  --processor-module-path <path>
        Specify a module path where to find annotation processors
  --processor-path <path>, -processorpath <path>
        Specify where to find annotation processors
  -profile <profile>
        Check that API used is available in the specified profile.
        This option is deprecated and may be removed in a future release.
  --release <release>
        Compile for the specified Java SE release.
        Supported releases: 
            8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21
  -s <directory>               Specify where to place generated source files
  --source <release>, -source <release>
        Provide source compatibility with the specified Java SE release.
        Supported releases: 
            8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21
  --source-path <path>, -sourcepath <path>
        Specify where to find input source files
  --system <jdk>|none          Override location of system modules
  --target <release>, -target <release>
        Generate class files suitable for the specified Java SE release.
        Supported releases: 
            8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21
  --upgrade-module-path <path>
        Override location of upgradeable modules
  -verbose                     Output messages about what the compiler is doing
  --version, -version          Version information
  -Werror                      Terminate compilation if warnings occur
```

lihaoyi:
One other convention I've seen is `-h` for short help and `--help` for verbose help, which `ruby` and `codex` do

libs/util/src/mill/util/JdkCommandsModule.scala
lefou: I really don't see the value in those specific tasks, as they are not generic. I could very well live with some special entry point to wrap all kind of thirdparty dev tools, since we have all infrastructure to provide them, but having a dedicated task for each of them doesn't scale. Making it generic like `mill tool java` (as I suggested before) would scale better, since we could register more tools under the hood without inventing new tasks for each, and more importantly we could special handle such entry point and properly report exit values without stopping the Mill server.

lefou:
From a convenience point, I really like the `-h` for short help and `--help` for full list.

lefou:
Mill already skips the examples in the `-h`, so I think it would be ok to show all options under `--help`.

lihaoyi:
I'd like to keep `--help` to be the one-pager summary including the common-commands cheat sheet, since that's what people are mostly likely to enter if they need help. 

We can leave the other one as `--help-advanced` or rename it to `--help-all`/`--help-extra`

libs/util/src/mill/util/JdkCommandsModule.scala
lefou: Leaving with `Task.ctx().systemExit` makes subsequent calls to the same task significatly slower.

Issue #6457, opened 2025-12-27
Android: make use of globSelectors in instrumented tests


vaslabs:
Our implementation hasn't evolved from the first two days and it runs all of the tests.

We can make use of glob selectors 

```scala
globSelectors: Task[Seq[String]]
```

and the adb Activity Manager (am) [options](https://developer.android.com/studio/test/command-line#am-instrument-options) to allow users to select classes and methods to run

cc: @souvlakias

PR #6543, opened 2026-01-08
Android: Pass task args to adb when running Instrumented tests

souvlakias:
Addresses #6457 

The initial thought was to have users use the `.testOnly` task and pass `globSelectors`, but that would require for us to discover all of the test packages,classes and functions.

Until we explore how to do that correctly, we could just pass any additional args the user provides to the test task to `adb`, for manual `adb` [configuration](https://developer.android.com/studio/test/command-line#am-instrument-options)
Example usage:
```
> ./mill app.androidTest -e class <className1>,<className2> -e package <pkgName1>,... # etc
``` 

### Changes
- Pass the `tests` args to `adb`
- Add 2 additional test classes in the `hello-world` and `hello-kotlin` examples and update their `build.mill`'s documentation to showcase the example usage

### DIFF
--- a/example/androidlib/java/1-hello-world/build.mill
+++ b/example/androidlib/java/1-hello-world/build.mill
@@ -175,7 +175,7 @@ object app extends AndroidAppModule {
 
 */
 
-// The command accepts additional parameters to filter tests using `adb` `-e` options.
+// The instrumented tests command accepts additional parameters to filter tests using `adb` `-e` options.
 // (https://developer.android.com/studio/test/command-line#am-instrument-options[Documentation found here])
 //
 // For example, to run a specific package:
--- a/example/androidlib/kotlin/1-hello-kotlin/build.mill
+++ b/example/androidlib/kotlin/1-hello-kotlin/build.mill
@@ -177,7 +177,7 @@ object app extends AndroidAppKotlinModule {
 
 */
 
-// The command accepts additional parameters to filter tests using `adb` `-e` options.
+// The instrumented tests command accepts additional parameters to filter tests using `adb` `-e` options.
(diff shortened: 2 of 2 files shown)


vaslabs:
imo we should have an androidTest dedicated example to demonstrate the usage of this. just a simple app with a few android test classes that run fake assertions and demonstrate both failures and successes

libs/androidlib/src/mill/androidlib/AndroidAppModule.scala
```diff
@@ -1069,7 +1069,9 @@ trait AndroidAppModule extends AndroidModule { outer =>
     /**
      * Runs the tests on the [[runningEmulator]] with the [[androidTestApk]]
      * against the [[androidApk]]
-     * @param args
+     *
+     * @param args Additional adb shell am instrument arguments.
+     *             See [[https://developer.android.com/studio/test/command-line#am-instrument-options]]
```
vaslabs, line 1072: maybe, instead of having another example like in my previous comment/suggestion, we can put some examples in this scaladoc.

but they won't be updated as the api evolves

vaslabs:
lgtm with the docs, have you checked website.fastPages appearance is good? I don't think I'll have time to double check today

souvlakias:
> lgtm with the docs, have you checked website.fastPages appearance is good? I don't think I'll have time to double check today

image
image

EDIT:
had to remove the *specific method* lines because the command had a `#` and it messed up with the runner's parsing

vaslabs:
> EDIT: had to remove the _specific method_ lines because the command had a `#` and it messed up with the runner's parsing

Probably needs escaping or wrapping in quotes, I'll have a look

PR #6545, opened 2026-01-08
Android: Assets collection logic

### LINKED ISSUE #6544: Android: Support assets collection
Another missing feature that came up while migrating [Jetcaster](https://github.com/vaslabs-ltd/compose-samples-with-mill/pull/21), is assets collection.
Some dependencies bring in an `assets` folder that needs to be collected, passed to `aapt2 link`, and (if needed) be present in the apk.

souvlakias:
This makes the Jetcaster.wear sample functional ([Repo](https://github.com/vaslabs-ltd/compose-samples-with-mill/pull/21))
Closes #6544 

Collects the `assets` directory from dependencies and passes them to `aapt2 link`


example/androidlib/java/5-R8/build.mill
```diff
@@ -42,7 +42,8 @@ object app extends AndroidR8AppModule { // <2>
   def androidReleaseKeyPass: T[Option[String]] = Task { Some("MillBuildTool") }
   def androidReleaseKeyStorePass: T[Option[String]] = Task { Some("MillBuildTool") }
 
-  override def androidVirtualDevice = super.androidVirtualDevice().withName("java-test")
+  override def androidVirtualDevice = super.androidVirtualDevice().withName("java-r8-test")
+  override def androidEmulatorPort = "5556"
```
lefou, line 46: Since https://github.com/com-lihaoyi/mill/pull/6486, we should be able to use  `build.example.testPort` or `PORT` env variable.

https://github.com/com-lihaoyi/mill/blob/bc6f01a3937c657ed75a46fb582eca61c175fa33/example/package.mill#L255-L262

souvlakias: The caveat with `build.example.testPort`s logic, is that android emulators require an even port number between a certain range, in order to be visible to adb (https://developer.android.com/studio/run/emulator-commandline#common)

> Note that if the port value is not even and is in the range 5554 to 5584, the virtual device will start but not be visible when you use the adb devices command, if the adb server starts after the emulator. For this reason, we recommend using an even console port number.

vaslabs: @souvlakias we can customise the testPort to follow the even port logic I think

vaslabs: try editing this https://github.com/com-lihaoyi/mill/blob/bc6f01a3937c657ed75a46fb582eca61c175fa33/example/package.mill#L143

lefou: Does that really say the ports must be between 5554 and 5584? I'm not a native English speaker, so hard to tell, but it sound like it's just a suggestion to reduce impact by other assumptions.

For the even port number, you could acquire ports consecutively (run `testPort` repeatedly), until you get an even number, and then run it once more, to ensure the port + 1 is also reserved.

souvlakias: IIUC it is a suggestion yes, to catch the case of the adb server being started after the emulator. Thus, it does concern us since the adb server will run the first time we run any `adb` command which in our cases, happens after the calls of `avd` and `emulator`

vaslabs: We should also add this explanation and the link to the emulator port method

souvlakias: Implementing this in #6560 since it's more relevant there

PR #6546, opened 2026-01-08
Fix testQuick for JUnit: normalize class names from test results

### LINKED ISSUE #4109: Fine-grained selective testing at a class-level granularity

FusionBrah:
Fixes #4109

This PR completes the `testQuick` command implementation, building on the work from #4731.

**The fix:** JUnit 4 with junit-interface reports `fullyQualifiedName` as `class.method` (e.g., `foo.FooTest.test1`), while `discoveredTestClasses` returns just the class name. This caused failed tests to not be tracked for re-running. Fixed by matching test result names against discovered classes using both exact match and prefix match.

**Changes:**
- Fix class name normalization in `testQuick` (JavaModule.scala)
- Add Java/JUnit integration test
- Add Java example with usage docs
- Update javalib testing.adoc

**Testing:**
- `integration.feature[test-quick].test` passes (existing Scala tests)
- `integration.feature[test-quick-java].test` passes (new Java tests)


FusionBrah:
Closing - needs to be rebased on current main which has significant restructuring. Will resubmit with updated implementation.

Issue #6534, opened 2026-01-07
GraalVM: Exclude native image configurations from jars if they are found in reachability metadata


vaslabs:
See https://github.com/com-lihaoyi/mill/issues/6496#issuecomment-3720376603 for more details

Issue #6496, opened 2026-01-05
GraalVM: Micronaut and Quarkus examples for native builds


vaslabs:
## Background

After the work (but not limited to) in https://github.com/com-lihaoyi/mill/pull/6359, https://github.com/com-lihaoyi/mill/pull/6315 and https://github.com/com-lihaoyi/mill/pull/6315 it should be possible now to extend our examples to more advanced configurations and look into the possibility that Quarkus Java apps are  supported as well.

## Steps

- Demonstrate (or extend graalvm support further) that micronaut native builds are possible with mill - added examples
- Add basic Quarkus examples as part of https://github.com/com-lihaoyi/mill/issues/3549
- Extend Quarkus examples to demonstrate native builds are possible

### Micronaut getting Native Images get started guide

- https://guides.micronaut.io/latest/micronaut-creating-first-graal-app.html

### Quarkus get started flow:

- https://quarkus.io/get-started/

### Building native executables with Quarkus

- https://quarkus.io/guides/building-native-image

vaslabs:
@lihaoyi are we also looking to support Quarkus cli functionality via mill? I'd at least test interoperability between mill builds and quarkus cli, atm I'm not sure whether it's needed for fully supporting Quarkus dev flows but I'll find out

(e.g. https://quarkus.io/guides/maven-tooling#dev-mode)

lihaoyi:
@vaslabs good question. I'm not familiar with the quarkus CLI, but I guess the question is how much of it is covered by Mill's existing featureset, and how much of it is not and maybe we need to support explicitly

vaslabs:
Yes, I imagine supporting it explicitly is the end goal, one reason would be to sync the JVM environment, versions etc.

In any case, I'll start to cover the base and see what we discover along the way

vaslabs:
For micronaut so far I've discovered the following.


### AOT

Micronaut, like spring boot also has AOT, running it is straightforward:

<details>

```scala
def aotConfigFile: T[PathRef] = Task {
    val file = Task.dest / "micronaut-aot.properties"
    os.write(file,
      """|logback.xml.to.java.enabled=true
         |netty.properties.enabled=true
         |deduce.environment.enabled=true
         |graalvm.config.enabled=true
         |serviceloading.native.enabled=false
         |precompute.environment.properties.enabled=true
      """.stripMargin)
    PathRef(file)
  }

  def micronautProcessAOT: T[PathRef] = Task {
    val dest = Task.dest

[...]

PR #6547, opened 2026-01-08
GraalVM: Honor the reachability metadata override flags by excluding configs inside dependency artifacts 

vaslabs:
## Motivation

Fixes https://github.com/com-lihaoyi/mill/issues/6534

Fixes micronaut native image generation (on basic examples) by honouring the override configurations and using a better alternative (`-H:ConfigurationFileDirectories=`) to pass the reachability metadata configs.

For a more detailed explanation see [this issue](https://github.com/com-lihaoyi/mill/issues/6496)

## Solution

Reachability metadata may provide [override flags](https://github.com/oracle/graalvm-reachability-metadata/blob/master/metadata/io.netty/netty-common/index.json#L25). These suggest that something needs to take precedence over something else. I didn't find a direct recommendation in any of the docs, but, I pulled out the native-image command gradle uses and found 6 `--exclude-config` flags on 6 netty artifacts.

After some trial and error, I replicated the result using this override logic. For more information on the reverse engineering part [see my comment here](https://github.com/com-lihaoyi/mill/issues/6496#issuecomment-3720197458)

## Micronaut

Micronaut, also has AOT. I focused on the GraalVM support in this PR, so I haven't added direct support to Micronaut's AOT yet. But I've provided an example, which shows how it can be added by the user.

A future PR may provide a similar module to SpringBootModule with all the utilities given out of the box.

I haven't added much in this example yet, just that it builds.

Manual test:

image

I have also added support for including application resources in the native image from the module itself (ref: https://www.graalvm.org/22.1/reference-manual/native-image/Resources/)
 

## Other fixes

With the changes to fetch artifacts, the kotlin native fails due to ignoring the resolution parameters.  A fix was added in the coursier module to pass resolutionParams on.

### DIFF
--- a/example/javalib/web/5-micronaut-native/build.mill
+++ b/example/javalib/web/5-micronaut-native/build.mill
@@ -1,4 +1,5 @@
 package build
+
 import mill.*
 import javalib.*
 import mill.api.{PathRef, Task}
@@ -7,7 +8,7 @@ import os.zip.ZipSource
 
 import java.util.jar.JarInputStream
 
-object `package` extends MavenModule { outer =>
+object `package` extends MavenModule, NativeImageModule { outer =>
 
   override def bomMvnDeps = Seq(
     mvn"io.micronaut.platform:micronaut-platform:4.10.3"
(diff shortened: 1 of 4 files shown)


vaslabs:
CI failure seems flakiness due to IO

Ready for a review @lihaoyi @lefou

example/javalib/web/5-micronaut-native/build.mill
```diff
@@ -0,0 +1,160 @@
+package build
```
lihaoyi, line 1: This should be referenced on the `javalib/web.adoc` page

vaslabs:
+ also fixing a few image option issues + fixing the resources not included in native image

libs/javalib/src/mill/javalib/CoursierModule.scala
```diff
@@ -311,6 +311,7 @@ object CoursierModule {
         deps0.map(_.dep),
         sources = sources,
         ctx = Some(ctx),
+        resolutionParams = resolutionParams,
```
lefou, line 314: Looks like a bug fix. Could you mention it in the PR description?

vaslabs: done

Issue #6427, opened 2025-12-20
Occasional exception when compiling a kotlin module


lefou:
Mill: 1.1.0-RC3

Multi-module project.

```console
> mill some-kotlin-module.compile
...
java.lang.Exception: fatal exception occurred: java.lang.NoSuchMethodError: 'org.jetbrains.kotlin.com.intellij.openapi.vfs.VirtualFileSystem org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreApplicationEnvironment.getJarFileSystem()'
        at org.jetbrains.kotlin.buildtools.internal.CompilationServiceImpl.clearJarCaches(CompilationServiceImpl.kt:139)
        at org.jetbrains.kotlin.buildtools.internal.CompilationServiceImpl.finishProjectCompilation(CompilationServiceImpl.kt:130)
        at org.jetbrains.kotlin.buildtools.internal.CompilationServiceProxy.finishProjectCompilation(CompilationServiceImpl.kt)
        at mill.kotlinlib.worker.impl.JvmCompileBtApiImpl.compile(JvmCompileBtApiImpl.scala:57)
        at mill.kotlinlib.worker.impl.KotlinWorkerImpl.compile(KotlinWorkerImpl.scala:36)
        at mill.kotlinlib.KotlinModule.$anonfun$15(KotlinModule.scala:303)
        at mill.util.CachedFactoryWithInitData.withValue(CachedFactory.scala:60)
        at mill.util.CachedFactory.withValue(CachedFactory.scala:82)
        at mill.kotlinlib.KotlinModule.kotlinCompileTask$$anonfun$1(KotlinModule.scala:363)
        at mill.api.Task$Anon.evaluate(Task.scala:436)
        at mill.exec.GroupExecution.$anonfun$18(GroupExecution.scala:445)
        at mill.exec.GroupExecution$.wrap$$anonfun$1$$anonfun$1$$anonfun$1$$anonfun$1$$anonfun$1(GroupExecution.scala:757)
        at mill.api.daemon.ClassLoader$.withContextClassLoader(ClassLoader.scala:14)
        at mill.exec.GroupExecution$.wrap$$anonfun$1$$anonfun$1$$anonfun$1$$anonfun$1(GroupExecution.scala:767)
        at mill.api.Evaluator$.withCurrentEvaluator$$anonfun$2$$anonfun$1(Evaluator.scala:146)
[...]

PR #6548, opened 2026-01-08
Properly cleanup kotlin compiler after compilation and on worker close

lefou:


lefou:
Closed, since it re-introduced issue https://github.com/com-lihaoyi/mill/issues/6427

Issue #6522, opened 2026-01-07
CI: Android test suite is failing since January 2026


lefou:
I haven't analyzed it and am not sure, it's always the same set of tests. Hence, I opened this issue to track it.

lihaoyi:
@vaslabs could you help look into the failures

vaslabs:
looks like since the emulator changes, it's picking android-35 for the image name, but there are no android-35 images (only android-34). I wonder why it didn't fail before the PR was merged, but will fix it asap

cc: @souvlakias 

No I was wrong, it seems something else broke it

vaslabs:
maybe something with parallelisation, I ran a few individually and they work, but all together are conflicting on the emulator

vaslabs:
the third party compose samples pass, what's the difference between shared and packaged daemon?

<img width="283" height="221" alt="Image" src="https://github.com/user-attachments/assets/8c50dc04-9258-4f97-9e91-9a832159d29a" />

vaslabs:
I can replicate it locally, running with shared doesn't seem to be able to read the android device configuration

lihaoyi:
`.shared` checks out consecutive runs in the same workspace folder and with the daemon living between them, `.packaged` runs in separate folders with a fresh daemon per test. 

We can swap it to `.packaged` if it doesn't work with `.shared` I guess, but it would be nice to figure out why consecutive tests running in the same folder fail since Mill is meant to handle cleanup and invalidation automatically

vaslabs:
> `.shared` checks out consecutive runs in the same workspace folder and with the daemon living between them, `.packaged` runs in separate folders with a fresh daemon per test.
> 
> We can swap it to `.packaged` if it doesn't work with `.shared` I guess, but it would be nice to figure out why consecutive tests running in the same folder fail since Mill is meant to handle cleanup and invalidation automatically

yeah, I can patch it up for now, I suspect the android setup won't play so well as the emulator management is a bit hacky and the configuration is between global/system and task directories. 

there's another issue with packaged, all emulators start more or less at the same moment and I get some OOM errors locally.

I did run 1 at a time to make sure emulators don't conflict, locally this works with

`./mill -ij1 example.androidlib.__.packaged.daemon.testForked`


[...]

PR #6551, opened 2026-01-08
Android CI: Fix concurrent android sdk download

vaslabs:
- Fixes parallel app testing conflicts on sdkmanager image download and conflicting on emulators.
- I've tried two approaches, one with syncrhonized and one with file lock. The synchronized approach was not consistently working but file lock seems to work with share.

Fixes https://github.com/com-lihaoyi/mill/issues/6522

### Implementation details

Context: A lot of this was derived through trial and error, please let me know if I missed something Mill internal related.

#### Aim of the implementation

To support the shared (`example.androidlib.__.shared.daemon.testForked`) mode for Android we need to guarantee that one SdkManager process runs at any time. SdkManager is a wrapper of a java application called SdkManagerCli so we can observe this with tools like visualvm

image

#### Funnel/Locking strategy 

First, we want to guarantee that within the same process, only one thread executes the sdkmanager installation process (different tasks that do different things with sdkmanager can access it in parallel within the same jvm process), which also relies on checking the filesystem outside of Mill's control (such as the android home directory). This was true for sometime, but it recently became unreliable especially with the shared mode. 

Thus, bringing to the second part, which is to support the only one instance of sdkmanager running guarantee across different jvm processes. 

#### Worker

The download responsibility has been moved to an external module - `AndroidSdkManagerModule`. This encapsulates a worker, with a single threaded pool, which guarantees - in process - only one execution at a time.

To guarantee across jvm processes, a file lock is used in the android home, handled by the worker.

- Using just a file lock, does not work because the same process overlaps the lock with multiple threads
- Having a single threaded pool is an easy way to avoid managing in memory locks in addition with file locks.

#### Shared daemon seems to be behaving non deterministically still

Sometimes I see this failure

image

Weirdly, the file is there. I don't  know under what conditions this happens yet.

### DIFF
--- a/example/javalib/web/5-micronaut-native/build.mill
+++ b/example/javalib/web/5-micronaut-native/build.mill
@@ -1,0 +1,148 @@
+package build
+
+import mill.*
+import javalib.*
+import mill.api.{PathRef, Task}
+import mill.util.Jvm
+import os.zip.ZipSource
+
+import java.util.jar.JarInputStream
+
+object `package` extends MavenModule, NativeImageModule { outer =>
+
+  override def bomMvnDeps = Seq(
+    mvn"io.micronaut.platform:micronaut-platform:4.10.3"
+  )
+
+  /**
+   * TODO Mill support for AOT
+   */
+  def resolvedMicronautAotCli: T[Seq[PathRef]] = defaultResolver().classpath(
+    Seq(
+      mvn"io.micronaut.aot:micronaut-aot-cli:2.9.0",
+      mvn"io.micronaut.aot:micronaut-aot-api:2.9.0",
+      mvn"io.micronaut.aot:micronaut-aot-core:2.9.0",
+      mvn"io.micronaut.aot:micronaut-aot-std-optimizers:2.9.0"
+    )
+  )
+
+  /**
+   * TODO Mill support for AOT
+   * More information on configuring Micronaut AOT can be found
+   * in [[https://micronaut-projects.github.io/micronaut-aot/latest/guide/configurationreference.html]]
+   */
+  def aotConfigFile: T[PathRef] = Task {
+    val file = Task.dest / "micronaut-aot.properties"
+    os.write(
+      file,
+      """
+        |logback.xml.to.java.enabled=true
+        |netty.properties.enabled=true
+        |deduce.environment.enabled=true
+        |serviceloading.native.enabled=false
+        |precompute.environment.properties.enabled=true
+        |sealed.property.source.enabled=true
+        |cached.environment.enabled=true
+        |graalvm.config.enabled=true
+      """.stripMargin
+    )
+    PathRef(file)
+  }
+
+  /**
+   * The settings can be found in
+   * [[https://github.com/micronaut-projects/micronaut-aot/blob/3.0.x/aot-cli/src/main/java/io/micronaut/aot/cli/Main.java]]
+   */
+  def micronautProcessAOT: T[PathRef] = Task {
+    val dest = Task.dest
+
+    val args = Seq(
+      "--classpath",
+      (runClasspath() ++ resolvedMicronautAotCli()).map(_.path).mkString(":"),
+      "--package",
+      "hello.world",
+      "--runtime",
+      "native",
+      "--config",
+      aotConfigFile().path.toString,
+      "--output",
+      dest.toString
+    )
+
+    Jvm.callProcess(
+      mainClass = "io.micronaut.aot.cli.Main",
+      mainArgs = args,
+      classPath = resolvedMicronautAotCli().map(_.path)
+    )
+
+    PathRef(dest)
+  }
+
+  override def nativeImageClasspath: Task.Simple[Seq[PathRef]] = Task {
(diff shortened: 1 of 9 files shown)


vaslabs:
if this works out, we probably should create some sort of AndroidSdkManagerWorker that handles the installations for all modules (i.e. AndroidSdkModule) . I suspect the issue would exist in other parts but it's probably more stable due to the license check and the ndk is only used in one place in the CI

vaslabs:
hm, need to dig deeper

9307] Warning: Observed package id 'tools' in inconsistent location '/home/runner/.android/sdk/tools-2' (Expected '/home/runner/.android/sdk/tools')
9307] Warning: Already observed package id 'tools' in '/home/runner/.android/sdk/tools'. Skipping duplicate at '/home/runner/.android/sdk/tools-2'
9307] Warning: Observed package id 'tools' in inconsistent location '/home/runner/.android/sdk/tools-2' (Expected '/home/runner/.android/sdk/tools')
9307] Warning: Already observed package id 'tools' in '/home/runner/.android/sdk/tools'. Skipping duplicate at '/home/runner/.android/sdk/tools-2'
9307] Warning: Observed package id 'tools' in inconsistent location '/home/runner/.android/sdk/tools-2' (Expected '/home/runner/.android/sdk/tools')
9307] Warning: Already observed package id 'tools' in '/home/runner/.android/sdk/tools'. Skipping duplicate at '/home/runner/.android/sdk/tools-2'

vaslabs:
hm, something else is (also?) wrong, for some reason, the downloads get interrupted

vaslabs:
thanks @souvlakias

lefou:
I'm not convinced we want a general purpose generic action runner in the context of the worker. Can we instead move the download jobs into the worker and use them via API? The worker then can do whatever is needed to guarantee it's working well.

vaslabs:
> I'm not convinced we want a general purpose generic action runner in the context of the worker. Can we instead move the download jobs into the worker and use them via API? The worker then can do whatever is needed to guarantee it's working well.

yep, will do it like that, the general purpose was just the quickest way to get there, but I can make the general purpose function private and expose domain methods + get the os calls in, effectively move some of the AndroidSdkModule methods to the worker

lefou:
I thought, the `AndroidSdkModule` is already a singleton and the download happens in a single task, such that there don't happen concurrent downloads to the same location. So either my mental module isn't right, or there are some tasks where we didn't get it right, yet. Or, we manually use alternative `AndroidSdkModule`s pointing to the same target location, and more then one of them.

vaslabs:
> I thought, the `AndroidSdkModule` is already a singleton and the download happens in a single task, such that there don't happen concurrent downloads to the same location. So either my mental module isn't right, or there are some tasks where we didn't get it right, yet. Or, we manually use alternative `AndroidSdkModule`s pointing to the same target location, and more then one of them.

yeah, something is weird there, the synchronized didn't seem to work, even if I replaced the file lock with one of them. I think the AndroidSdkModule is not a singleton, it's extended in every example. But even so, I'd expect the private objects could be used as locks, but apparently not.

(Even so the presence of 3 locks could cause a deadlock, which might be one of the reasons the job occasionally times out)

lefou:
So, we need to use a single ExternalModule to hold all the download logic (or if you want to use locks, to own the locks). If not we can't ensure the locks are singletons. Even some clever worker logic won't help, if the worker has multiple instances due to being bound to non-singleton `AndroidSdkModule`s.

vaslabs:
> So, we need to use a single ExternalModule to hold all the download logic (or if you want to use locks, to own the locks). If not we can't ensure the locks are singletons. Even some clever worker logic won't help, if the worker has multiple instances due to being bound to non-singleton `AndroidSdkModule`s.

Ok, I'll give it a go with the external module in the weekend, I was gonna do major work on the file lock approach anyway in the worker, might as well do the external module approach

libs/androidlib/src/mill/androidlib/AndroidSdkManagerModule.scala
```diff
@@ -0,0 +1,472 @@
...
+      tryLock.close()
+      attempts += 1
+      if (attempts > maxAttempts)
+        Task.fail(
+          s"Giving up waiting for sdkmanager. Please check if the process is slow or stuck or if the ${androidSdkManagerLockFile} is held by another process!"
+        )
+      Task.log.info(
+        s"Waiting for sdkmanager lock ($attempts/$maxAttempts)... Trying again in 5 seconds!"
+      )
+      Thread.sleep(5000)
```
vaslabs, line 429: I was thinking this could be shorter or an exponential backoff, but , we should arrive here only in cases where multiple mill jvm processes exist, in cases of same process access, it should be handled by the pool (can't enter here twice from the same process anyway). It's probably a  mill CI only flow

vaslabs:
so far , android tests run consistently green 2 times (2 in my branch, 2 here). The only failure seems to be the same on main branch

lihaoyi:
@vaslabs If you're done with this PR I'll merge it

vaslabs:
I'm done, I'll keep an eye for improvements as we go along anyway!

PR #6553, opened 2026-01-08
Add `testQuick` command for selective test execution

### LINKED ISSUE #4109: Fine-grained selective testing at a class-level granularity

FusionBrah:
## Summary
- Adds `testQuick` command that uses bytecode callgraph analysis to run only tests affected by code changes
- Re-runs previously failed tests even without code changes  
- Fixes JUnit 4 bug where `fullyQualifiedName` includes method name ("class.method" vs "class")
- Supports cross-module dependency tracking (changes in dependency module trigger tests in dependent modules)
- Works with Java, Scala, and Kotlin projects

Fixes #4109

## Test plan
- [x] Integration test: single module Java (JUnit4) - first run, skip run, code change, failed re-run, fix, final skip
- [x] Integration test: multi-module Java - changes in `bar` module trigger tests in dependent `foo.test`
- [x] Integration test: Scala module (uTest) - full testQuick cycle
- [x] Example test: `example/javalib/testing/8-test-quick`
- [x] Example test: `example/scalalib/testing/8-test-quick`

## Documentation
- Added `testQuick` section to `javalib/testing.adoc`
- Added `testQuick` section to `scalalib/testing.adoc`

### DIFF
--- a/integration/ide/build-classpath-contents/src/BuildClasspathContentsTests.scala
+++ b/integration/ide/build-classpath-contents/src/BuildClasspathContentsTests.scala
@@ -34,6 +34,8 @@ object BuildClasspathContentsTests extends UtestIntegrationTestSuite {
             "mill-core-api-java11_3-SNAPSHOT.jar",
             "mill-core-api_3-SNAPSHOT.jar",
             "mill-core-constants-SNAPSHOT.jar",
+            "mill-core-internal-cli_3-SNAPSHOT.jar",
+            "mill-core-internal_3-SNAPSHOT.jar",
             "mill-libs-androidlib-databinding_3-SNAPSHOT.jar",
             "mill-libs-androidlib_3-SNAPSHOT.jar",
             "mill-libs-daemon-client-SNAPSHOT.jar",
@@ -60,7 +62,8 @@ object BuildClasspathContentsTests extends UtestIntegrationTestSuite {
             "mill-libs-util_3-SNAPSHOT.jar",
             "mill-libs_3-SNAPSHOT.jar",
             "mill-moduledefs_3-0.12.8.jar",
-            "mill-runner-autooverride-api_3-SNAPSHOT.jar"
+            "mill-runner-autooverride-api_3-SNAPSHOT.jar",
+            "mill-runner-codesig_3-SNAPSHOT.jar"
           )
         )
         assert(millLocalClasspath == Nil)


integration/feature/test-quick/src/TestQuickTests.scala
```diff
@@ -0,0 +1,142 @@
...
+      // Modify bar() to return 2 - this will cause FooTest2 to fail
+      modifyFile(
+        workspacePath / "foo" / "src" / "Foo.java",
+        _.replace("return 1;", "return 2;")
+      )
+
+      // Third run: FooTest2 should run and fail (bar() changed and affects FooTest2)
+      val thirdRun = eval("foo.test.testQuick")
+      assert(!thirdRun.isSuccess)
+      assert(thirdRun.out.contains("FooTest2"))
```
lihaoyi, line 33: These tests are too coarse and do not assert that the things that shouldn't run, don't. Please assert against the full set of test outputs and make sure it matches exactly what you expect

integration/feature/test-quick/src/TestQuickTests.scala
```diff
@@ -0,0 +1,142 @@
+package mill.integration
```
lihaoyi, line 1: This can be a unit test rather than an integration test

integration/feature/test-quick/src/TestQuickTests.scala
```diff
@@ -0,0 +1,142 @@
...
+
+import mill.testkit.UtestIntegrationTestSuite
+import utest._
+
+object TestQuickTests extends UtestIntegrationTestSuite {
```
lihaoyi, line 6: Please include test cases that cover the edge cases where codesig does not work well. See the original codesig PR for details

integration/feature/test-quick/src/TestQuickTests.scala
```diff
@@ -0,0 +1,142 @@
...
+      )
+
+      // Third run: FooTest2 should run and fail (bar() changed and affects FooTest2)
+      val thirdRun = eval("foo.test.testQuick")
+      assert(!thirdRun.isSuccess)
+      assert(thirdRun.out.contains("FooTest2"))
+      // FooTest1 may or may not run depending on callgraph - it shouldn't since qux() didn't change
+
+      // Fourth run: FooTest2 should re-run (it failed, so it should re-run even without changes)
+      // This is the CRITICAL test for the JUnit bug fix
```
lihaoyi, line 37: This comment and many others are useless and demonstrate a profound lack of understanding. If you don't understand what the code is doing, and your AI assistant doesn't either, you probably shouldn't be opening a PR

example/scalalib/testing/8-test-quick/build.mill
```diff
@@ -0,0 +1,41 @@
+// The `testQuick` command provides fine-grained selective test execution by
```
lihaoyi, line 1: This example test doesn't actually test anything

FusionBrah:
No worries. Sorry for wasting your time on this.

PR #6555, opened 2026-01-09
Add possibility to add scripts and override entrypoint in dockerfile

gunnarvelle:
Adds uploading of scripts to docker image, and also override the entrypoint with reference to the script. 

Can be used with
```
  object docker extends DockerConfig {
    def exposedPorts = Seq(8080)
    def runApp: T[PathRef] = Task.Source { RelPath("../run-app.sh") }
    def scripts = Task {
      Seq(runApp() -> "run-app.sh")
    }
    def entrypoint = "/run-app.sh"
  }
```


contrib/docker/src/mill/contrib/docker/DockerModule.scala
```diff
@@ -108,6 +108,21 @@ trait DockerModule { outer: JavaModule =>
...
 
+    /**
+     * Scripts to include in the resulting docker image. Sequence of (PathRef source, target path inside image).
+     * The target paths are used for the COPY instructions in the Dockerfile.
+     *
+     * Note: users should provide the sources via `Task.Source(...)` so Mill can track file inputs, e.g.
+     * `def runApp = Task.Source { RelPath("../run-app.sh") }`
+     * `def scripts = Task { Seq(runApp() -> "run-app.sh") }`.
+     */
+    def scripts: T[Seq[(PathRef, String)]] = Seq.empty[(PathRef, String)]
```
lihaoyi: - Should it be typed as `Seq[(PathRef, os.SubPath)]`, to make the types a bit stricter?

- Is `scripts` the right name for this? It seems like it can be used for any arbitrary file right?

gunnarvelle: Fair points. Will update the pr.

contrib/docker/src/mill/contrib/docker/DockerModule.scala
```diff
@@ -160,13 +180,31 @@ trait DockerModule { outer: JavaModule =>
...
 
-    final def build = Task {
+    final def build: Task.Simple[Seq[String]] = Task {
       val dest = Task.dest
       val env = dockerEnv()
 
       val asmPath = outer.assembly().path
       os.copy(asmPath, dest / asmPath.last)
 
+      // Copy scripts (source PathRef -> target). Only files are supported.
```
lihaoyi, line 190: Looks like the term `script` here and below should be replaced by `files`?

contrib/docker/src/mill/contrib/docker/DockerModule.scala
```diff
@@ -160,13 +180,31 @@ trait DockerModule { outer: JavaModule =>
...
+      files().foreach { case (srcRef, targetStr) =>
+        val srcPath = srcRef.path
+        val dstPath = dest / targetStr
+
+        if (!os.exists(srcPath))
+          Task.fail(s"Docker script source does not exist: $srcPath")
+        else if (os.isDir(srcPath))
+          Task.fail(s"Docker script source is a directory (folder copying not supported): $srcPath")
```
lihaoyi, line 198: How hard would it be to support folders? Would it just be a matter of enumerating the files via `os.walk`? If so we might as well support it

contrib/docker/src/mill/contrib/docker/DockerModule.scala
```diff
@@ -108,6 +108,21 @@ trait DockerModule { outer: JavaModule =>
      */
     def executable: T[String] = "docker"
 
+    /**
+     * Files to include in the resulting docker image. Sequence of (PathRef source, target path inside image).
+     * The target paths are used for the COPY instructions in the Dockerfile.
```
lihaoyi, line 113: Since we're using the term `COPY` a lot, in relation to how this is implemented, should we just call it `def copy` instead? That would remove one layer of indirection

PR #6560, opened 2026-01-09
Android: create an `AndroidEmulatorWorker`

### LINKED ISSUE #6528: Android: Make emulator process known to mill
### API changes

```
./mill app.startAndroidEmulator
./mill app.stopAndroidEmulator
```

to

```
./mill app.androidEmulator
./mill clean app.androidEmulator
```

### Implementation example

See the MillBackgroundWrapper example [here](https://github.com/com-lihaoyi/mill/blob/b11da92b66518eee046d1308ab55d42358eae77b/libs/javalib/src/mill/javalib/NativeImageModule.scala#L71)

(Maybe) Related to https://github.com/com-lihaoyi/mill/issues/6522

souvlakias:
### Motivation
Have mill be aware of the emulator process

### Api Changes
Instead of using `{start,stop}AndroidEmulator` tasks to start and stop the emulator:
Have one task: `androidEmulator` that calls the `AndroidEmulatorWorker` worker which handles the `adb kill` logic when stopped. e.g. with `./mill clean app.androidEmulator`

Closes #6528

### Example port configurations
This also changes the Emulator port configurations in the examples to use a unique port set from `build.example` as suggested in https://github.com/com-lihaoyi/mill/pull/6545#discussion_r2679437314
Since it is [recommended](https://developer.android.com/studio/run/emulator-commandline#common)for emulators to use even ports between `5544 - 5584`:
- Have an atomic integer starting at `5554`
- Each example gets an even number through `EMULATOR_PORT`

### DIFF
--- a/example/androidlib/java/1-hello-world/build.mill
+++ b/example/androidlib/java/1-hello-world/build.mill
@@ -36,7 +36,7 @@ object app extends AndroidAppModule {
   def androidReleaseKeyStorePass: T[Option[String]] = Task { Some("MillBuildTool") }
 
   override def androidVirtualDevice = super.androidVirtualDevice().withName("java-test")
-  override def androidEmulatorPort: String = System.getenv.getOrDefault("PORT", "5554")
+  override def androidEmulatorPort: String = System.getenv.getOrDefault("EMULATOR_PORT", "5554")
 
   object test extends AndroidAppTests, TestModule.Junit4 {
     def junit4Version = "4.13.2"
@@ -126,7 +126,6 @@ object app extends AndroidAppModule {
 > ./mill show app.androidEmulator
 
 > ./mill show app.adbDevices
-...emulator-5554...device...
 
 > ./mill show app.it
 ...
--- a/example/androidlib/kotlin/1-hello-kotlin/build.mill
+++ b/example/androidlib/kotlin/1-hello-kotlin/build.mill
@@ -39,7 +39,7 @@ object app extends AndroidAppKotlinModule {
   def androidReleaseKeyPass: T[Option[String]] = Task { Some("MillBuildTool") }
   def androidReleaseKeyStorePass: T[Option[String]] = Task { Some("MillBuildTool") }
   override def androidVirtualDevice = super.androidVirtualDevice().withName("kotlin-test")
-  override def androidEmulatorPort: String = System.getenv.getOrDefault("PORT", "5556")
+  override def androidEmulatorPort: String = System.getenv.getOrDefault("EMULATOR_PORT", "5556")
 
   override def androidIsDebug: T[Boolean] = Task { false }
 
--- a/example/package.mill
+++ b/example/package.mill
@@ -144,10 +144,15 @@ object `package` extends Module {
     def testExclusive = true
 
     /**
-     * Unique Android emulator port for this example test module.
-     * Each module claims an even port by incrementing the shared counter by two.
+     * Unique and even Android emulator port for this example test module.
+     * Each module claims the port by incrementing the shared counter by two.
      */
-    override val testPort: Int = androidPortCounter.getAndAdd(2)
+    val emulatorPort: Int = androidPortCounter.getAndAdd(2)
+
+    override def forkEnv = super.forkEnv() ++ Map(
+      "EMULATOR_PORT" -> emulatorPort.toString
+    )
+
   }
 
   trait ExampleCrossModuleJava extends ExampleCrossModule {
--- a/example/androidlib/java/5-R8/build.mill
+++ b/example/androidlib/java/5-R8/build.mill
@@ -43,7 +43,7 @@ object app extends AndroidR8AppModule { // <2>
   def androidReleaseKeyStorePass: T[Option[String]] = Task { Some("MillBuildTool") }
 
   override def androidVirtualDevice = super.androidVirtualDevice().withName("java-r8-test")
-  override def androidEmulatorPort: String = System.getenv.getOrDefault("PORT", "5554")
+  override def androidEmulatorPort: String = System.getenv.getOrDefault("EMULATOR_PORT", "5554")
   override def androidIsDebug: T[Boolean] = Task { false }
 
   def androidProjectProguardFiles = Task.Sources(
--- a/example/androidlib/java/6-native-libs/build.mill
+++ b/example/androidlib/java/6-native-libs/build.mill
@@ -35,7 +35,7 @@ object app extends AndroidNativeAppModule { // <1>
   def androidReleaseKeyStorePass: T[Option[String]] = Task { Some("MillBuildTool") }
 
   override def androidVirtualDevice = super.androidVirtualDevice().withName("native-test")
-  override def androidEmulatorPort: String = System.getenv.getOrDefault("PORT", "5554")
+  override def androidEmulatorPort: String = System.getenv.getOrDefault("EMULATOR_PORT", "5554")
 
   def androidExternalNativeLibs = Task { // <2>
     Seq(
--- a/example/androidlib/kotlin/2-compose/build.mill
+++ b/example/androidlib/kotlin/2-compose/build.mill
@@ -29,7 +29,7 @@ object app extends AndroidAppKotlinModule {
   def androidEnableCompose = true
 
   override def androidVirtualDevice = super.androidVirtualDevice().withName("kotlin-compose-test")
-  override def androidEmulatorPort: String = System.getenv.getOrDefault("PORT", "5556")
+  override def androidEmulatorPort: String = System.getenv.getOrDefault("EMULATOR_PORT", "5556")
 
   def androidIsDebug = true
 


libs/androidlib/src/mill/androidlib/AndroidAppModule.scala
```diff
@@ -638,49 +636,36 @@ trait AndroidAppModule extends AndroidModule { outer =>
...
-  def stopAndroidEmulator: T[String] = Task {
-    val emulator = runningEmulator()
-    os.call(
-      (androidSdkModule().adbExe().path, "-s", emulator, "emu", "kill")
-    )
-    emulator
-  }
-
   /** The emulator port where adb connects to. Defaults to 5554 */
   def androidEmulatorPort: String = "5554"
```
vaslabs, line 663: need to add a comment here informing users about adb recommendations

vaslabs:
This lgtm, you can mark it as ready imo, the failure is the same as in main

example/androidlib/java/1-hello-world/build.mill
```diff
@@ -164,7 +164,7 @@ object app extends AndroidAppModule {
 // The android tests (existing typically in androidTest directory, aka instrumented tests)
 // typically run on an android device.
 // The createAndroidVirtualDevice command creates an AVD (Android Virtual Device)
-// and the startAndroidEmulator command starts the AVD. The it task runs the android tests
-// against the available AVD. The stopAndroidEmulator command stops the AVD and the
+// and the androidEmulator command starts the AVD. The it task runs the android tests
```
lefou, line 167:
```suggestion
// and the androidEmulator command starts the AVD. The `it` task runs the android tests
```

example/androidlib/kotlin/1-hello-kotlin/build.mill
```diff
@@ -167,7 +166,7 @@ object app extends AndroidAppKotlinModule {
 // The android tests (existing typically in androidTest directory, aka instrumented tests)
 // typically run on an android device.
 // The createAndroidVirtualDevice command creates an AVD (Android Virtual Device)
-// and the startAndroidEmulator command starts the AVD. The it task runs the integration tests
-// against the available AVD. The stopAndroidEmulator command stops the AVD and the
+// and the androidEmulator command starts the AVD. The it task runs the integration tests
```
lefou, line 169:
```suggestion
// and the androidEmulator command starts the AVD. The `it` task runs the integration tests
```

example/package.mill
```diff
@@ -142,6 +142,17 @@ object `package` extends Module {
 
   trait ExampleCrossModuleAndroid extends ExampleCrossModule {
     def testExclusive = true
+
+    /**
+     * Unique and even Android emulator port for this example test module.
+     * Each module claims the port by incrementing the shared counter by two.
```
lefou, line 148: Please add, why this "by two" is necessary to complete the picture.

libs/androidlib/src/mill/androidlib/AndroidEmulatorWorker.scala
```diff
@@ -0,0 +1,78 @@
...
+    os.call(
+      (
+        adbPath.path,
+        "-s",
+        emulatorId,
+        "emu",
+        "kill"
+      )
+    )
+  }
```
lefou, line 76: I think we should keep track of the fact, that we have started the emulator and not stopped it yet. Only then, we should stop it. Otherwise, there is a chance we stop someone's else emulator, e.g.. one started from a different branch.

libs/androidlib/src/mill/androidlib/AndroidAppModule.scala
```diff
@@ -605,7 +603,7 @@ trait AndroidAppModule extends AndroidModule { outer =>
    * @param extraArgs Additional arguments to pass to the emulator
    * @return The log line that indicates the emulator is ready
    */
-  def startAndroidEmulator(
+  def androidEmulator(
```
lefou, line 606: Why is this command necessary? Can't we use the worker directly?

website/docs/modules/ROOT/pages/android/android-initial-setup.adoc
```diff
@@ -102,7 +102,7 @@ This will install apk built by the `app.androidApk` command.
 === Stop the emulator
 [source,console]
 ----
-> ./mill app.stopAndroidEmulator
+> ./mill clean app.androidEmulator
```
lefou, line 105: Is this correct? Shouldn't this clean the worker instead?

souvlakias:
Hi, thanks for the feedback

> It's not that clear to me why we need a command and a worker. 

Don't we need the `Command` task in order for it to be a command, accept and process the named parameters and then call the worker?

> Also, in the examples, the command is `clean`ed, while I'd have expected that the worker needs to be `clean`ed.

After testing `clean`ing the command and seeing that the worker's `close` was called, I thought that this is an intended design in how tasks work. (that if a task calls a worker, it would be fine to `clean` the caller)

lefou:
> Hi, thanks for the feedback
> 
> > It's not that clear to me why we need a command and a worker.
> 
> Don't we need the `Command` task in order for it to be a command, accept and process the named parameters and then call the worker?

In none of the examples, there were any parameters supplied to the command and worker, hence I overlooked the fact that we support parameters at all. I need to review, what they do and how they affect the worker, since the worker state can and should not depend on these parameters, since they are not proper inputs to the worker. The only way you can pass those parameters to the worker are by calling methods on the worker itself. So, your command can only provide a way to interact with the worker from the CLI. But it should not be necessary to implement any build logic using the worker.

> > Also, in the examples, the command is `clean`ed, while I'd have expected that the worker needs to be `clean`ed.
> 
> After testing `clean`ing the command and seeing that the worker's `close` was called, I thought that this is an intended design in how tasks work. (that if a task calls a worker, it would be fine to `clean` the caller)

This can't be right. In your case, the initialized worker is a dependency to the command. Cleaning the command will not and should not affect the worker at all. Of course you can clean callers of the worker, but that should not have any effects to the worker.

lefou:
IIUC, the worker is managing a single emulator, but delays the start until `startEmulator` is called. Therefore the potentially provided parameters of the command find their way into the running emulator state. It unclear in which way the emulator process is reused by subsequent `androidEmulator` runs. It looks like the worker is spawning a new emulator each time but not properly cleaning up. 

Can you describe, how the command and the worker are supposed to interact and how and when reuse/cleanup/shutdown happens?

souvlakias:
In the implementation we have, if one subsequently calls `startAndroidEmulator`, the second call will fail since the emulator is already running. Calling `stopAndroidEmulator` will call `adb` to stop the emulator. (without checking if one is running)

Similarly, in the new implementation, when calling `androidEmulator` twice, both will reach the worker and the call that the worker does the second time will fail. 
If then one calls `clean app.androidEmulator`, the `close` method of the worker is called. (I'm saying this after manual test and observation, I am not sure if that should not have been the case since you said:
> Cleaning the command will not and should not affect the worker at all


</br>
</br>


Regarding
> I think we should keep track of the fact, that we have started the emulator and not stopped it yet

We could have a `var started: Boolean` for this and only when `true` call the `adb stop`.
We could also use this to avoid failures on subsequent `androidEmulator` commands. When true, either return "Already started", or restart it. (I think restarting makes more sense)

souvlakias:
Found why the `clean` works: https://github.com/com-lihaoyi/mill/blob/main/libs/util/src/mill/util/MainModule.scala#L266

I'm not sure if the `startsWith` was intentional there to support such functionality. 
In practice, I think it's a good feature to have 😅 (but maybe a safer logic than just common prefix which could accidentally match with many tasks)

lihaoyi:
@lefou if you don't have any other comments I'll merge this

lefou:
IMHO, the comments I already gave aren't resolved yet. The solution works by accident, but isn't robust enough. The documented `clean` command isn't the correct one to stop the worker. We either need more documentation (to the code or docs, not the PR discussion) about the limitation of this approach or fix them.

vaslabs:
I think the worker might not fit what we are trying to do. In my ideal situtation, we want to keep the commands as documented as it fits other mill use cases , so `./mill androidEmulator` and `./mill clean androidEmulator` works as `./mill run clean runBackground` for instance. 

But we need a robust way to clean. The difference between the run background, is that simply calling process.destroy is not the same as stopping the emulator in AGP , which calls the adb emu kill command, which takes care of saving the emulator state.

To move things forward, I propose to split this PR , do the ports so we get that in in one PR .

The second PR, we can either find a way to do what we aim with the worker, or if it's not possible a `MillBackgroundWrapper` specific for the adb emulator lifecycle.

vaslabs:
tbh, we started the emulator work with the motivation to improve the CI by letting mill keep track of the emulator in a similar fashion that we do with run background. But since that seems fixed now, there's less need for that, the only motivation left, to have consistency with "clean" .

@souvlakias I need to think more about this personally, so perhaps , let's just do the emulator port and park everything else for later

lefou:
There is no need to rush this, then. I think the worker is still the best way to shutdown external process and resources, since the `AutoCloseable.close` gives us the required callback to act on. Only limitation is currently, that a stopping Mill process is currently not closing all running workers (TODO: add issue ref), so we miss an opportunity to cleanup. 

But from what I learned in this PR is, that we probably want more that one emulator, e.g. when using user-provided command properties to start em up. Question is, do we want to stop all other or not. That decides, if the worker needs to manage more than one or not.

vaslabs:
> But from what I learned in this PR is, that we probably want more that one emulator

This is a question to be answered as well. There are use cases where you may need more than one emulator (to test an app with various android api levels, screen sizes, device types). There's also daily development where you may quickly run the app or instrumented tests.

Currently, we somewhat cover both, you can spin up various emulators by allowing users to inject arguments to the android emulator command. You can also use a different emulator for running the app to running the android tests.

What I suspect, is that we'll need to support (I mean, make it easy) an uber integration test, where users can run the instrumented tests against various emulators. I need to do some user research on this to see what people are actually doing...

vaslabs:
Let's close this for the time being

PR #6566, opened 2026-01-12
Add shared BuildGen interface, add --declarative options for Maven and Gradle converters

lefou:
When converting Maven or Gradle builds, users can now choose between YAML- or Scala-based Mill project with the `--declarative` option. Default is `true` and will generate the YAML project files.


lefou:
The test failure is unrelated and already present in `main`.

https://github.com/com-lihaoyi/mill/actions/runs/20915191287/job/60087563096

lihaoyi (approved):

Issue #6043, opened 2025-10-28
`mill init`  should provide a `--help` option


lefou:
Otherwise, it's hard to know which options are available.

E.g. The page https://mill-build.org/mill/migrating/auto-migrating.html shows an example with a `--jvm-id` option, but when I try to use it, it fails with:

```
Unknown arguments: "--jvm-id" "11"
Expected Signature: run-import
Imports an SBT build located in the current working directory.
  --custom-sbt <str>  path to sbt executable
  --merge             merge generated build files
  --no-meta           disable generating meta-build files
```

The options is probably only available on the Gradle importer, but who knows? 

Since there is some kind of help text available as part of the error message, we should also make it available via `--help`.

lefou:
See also
* https://github.com/com-lihaoyi/mill/issues/6039

ajaychandran:
> Otherwise, it's hard to know which options are available.
> 
> E.g. The page https://mill-build.org/mill/migrating/auto-migrating.html shows an example with a `--jvm-id` option, but when I try to use it, it fails with:
> 
> ```
> Unknown arguments: "--jvm-id" "11"
> Expected Signature: run-import
> Imports an SBT build located in the current working directory.
>   --custom-sbt <str>  path to sbt executable
>   --merge             merge generated build files
>   --no-meta           disable generating meta-build files
> ```
> 
> The options is probably only available on the Gradle importer, but who knows?
> 
> Since there is some kind of help text available as part of the error message, we should also make it available via `--help`.

The actual examples do not use `--jvm-id` anymore. I suppose the documentation views will be updated in the next release.
The other sections are a little dated and could use a touch up.

The help message displayed is the default one logged by `mainargs.Parser(this).runOrExit`. I am not sure what needs to change here. Could you elaborate?

lefou:
This issue isn't solved.

> The help message displayed is the default one logged by `mainargs.Parser(this).runOrExit`. I am not sure what needs to change here. Could you elaborate?

What looks like a "help-message" is the output of an "error-message". There is no way to see the "help-message" without causing an error. You can't see it with `mill inspect` and also not with adding `--help` to the command. Whether this is a generic issue or specific to the init command, I can't tell right now.

ajaychandran:
Below is the output using the latest version of Mill.

```sh
fs2$ ./mill init --help
No mill version specified.
You should provide a version via a '//| mill-version: ' comment or a '.mill-version' file.
Using mill version 1.1.0-RC1
[...]

PR #6567, opened 2026-01-12
Add explicit options for `init` command and --projectDir support

lefou:
* Allow explicit selection of the migration source with --maven, --gradle, --sbt.
* Allow explicit template selection with --giter8 option.
* Allow explicit example selection with --example option.
* Add a --projectDir option to the Maven, Gradle and sbt migrators to be able to use a distinct directory for the generated files.
* Add a --help to the init command

Fix https://github.com/com-lihaoyi/mill/issues/6043

Pull request: https://github.com/com-lihaoyi/mill/pull/6567


lefou:
Unfortunately, `mainargs` does not support parsing commands, so we can't fully model a complete set of accepted options for `init`.

lihaoyi (approved):

PR #6570, opened 2026-01-13
Make declarative module deps resolved dynamically

lihaoyi:
This further reduces the need for re-compilation during editing of `.mill.yaml` files

We always code-gen `def moduleDefs`, `def compileModuleDefs`, etc definitions in the generated code, and move the actual contents of these to a `mill/module-deps-config.json` file computed and placed on the classpath. 

The code-gen are hardcoded to call `resolveModuleDeps`, which (a) reads the `moduleDeps` info from `module-deps-config.json`, and calls a `superMethod` macro that calls the super method (if any), and combines them to form the final result.

As `module-deps-config.json` is a classpath resource, on any change the classloader and module tree are invalidated and re-instantiated. While this still does have a cost, we save on the re-_compiling_ of the generated build files, which is much more expensive.

After this, the only thing that would cause the generated Scala code to be re-compiled is changes to the `extends` clauses.


lefou:
Does it detect cycles?

lihaoyi:
> Does it detect cycles?

Yes it seems to work via the existing circular dependency checks we have on JavaModule

PR #6582, opened 2026-01-14
Ensure changes in `mill-version` or `mill-jvm-version` trigger selective execution for all tasks

lihaoyi:
Previously we only considered `mill-version`/`mill-jvm-version` during cache checks in the executor, but those cache checks would never get triggered during selective execution because we would skip those sub-graphs entirely. This PR re-uses the same `invalidateAllHashes` hash in the selective execution logic to ensure that we select everything if `mill-version`/`mill-jvm-version` change.

Tested via an integration test that forcibly mangles `invalidateAllHashes` in the selective execution data and checks that everything invalidates without any code or build changes


lihaoyi:
Superseded by https://github.com/com-lihaoyi/mill/pull/6583

PR #6587, opened 2026-01-14
Prefix all main classes with `Mill` to make it easier to spot them in `jps`

lihaoyi:
This way it's clear that all the `Mill*` process belong to Mill in some shape or form.

Also improved the naming of Scala script main classes, so `foo-bar.scala`'s classname changed from `foo$minusbar_MillScriptMain` to `MillScriptMain_foo_bar_scala`, which follow the `Mill*` prefix convention of other Mill main classes and avoids having `$encoded` characters in the name

### DIFF
--- a/libs/script/asm-worker/src/mill/script/asm/AsmWorkerImpl.scala
+++ b/libs/script/asm-worker/src/mill/script/asm/AsmWorkerImpl.scala
@@ -8,7 +8,7 @@ object AsmWorkerImpl {
     // Find the MillScriptMain_ class name to forward to
     val targetClassName = os.list(os.Path(classesDir))
       .map(_.last)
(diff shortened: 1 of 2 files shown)


libs/init/maven/src/mill/main/maven/MillMavenBuildGenMain.scala
```diff
@@ -6,7 +6,7 @@ import org.apache.maven.model.{Developer as MvnDeveloper, License as MvnLicense,
 
 import scala.jdk.CollectionConverters.*
 
-object MavenBuildGenMain {
+object MillMavenBuildGenMain {
```
lefou: `MillBuildGenMavenMain`

PR #6588, opened 2026-01-14
Android: Assign each example a unique even port in CI

souvlakias:
(Split from #6560)
### Motivation
Configure the examples running in CI to use a unique port set from `build.example`
It is [recommended](https://developer.android.com/studio/run/emulator-commandline#common) for emulators to use even ports between `5544 - 5584` in order for them to be discoverable to `adb`.

Specifically, if `adb` server starts after an emulator that listens outside of the suggested range `adb` will not discover it. 
In our case, the `adb` server starts the first time `adb` is called which happens after starting the emulator.

### Changes
- Have an atomic integer starting at `5554`
- Each example gets an even number through `EMULATOR_PORT` envvar
- 2 bugfixes:
	- `stopAndroidEmulator` needs to be a Command
    - when starting the emulator, also match `Successfully loaded snapshot` string

Pull request: https://github.com/com-lihaoyi/mill/pull/6588

### DIFF
--- a/contrib/sbom/readme.adoc
+++ b/contrib/sbom/readme.adoc
@@ -5,7 +5,6 @@ This plugin creates Software Bill of Materials (SBOM)
 
 This module has some limitations at the moment:
 
-- Minimal SBOM, various properties of libraries are missing. e.g. the license.
 - Only JVM ecosystem libraries are reported.
 - Only the CycloneDX JSON format is supported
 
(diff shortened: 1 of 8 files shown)


lefou:
Please add the bugfix comment to the PR description, since that becomes the final squash-commit comment.

lihaoyi:
Can we re-use the `PORT` env var introduced in https://github.com/com-lihaoyi/mill/pull/6486? Or is that specific port range necessary?

souvlakias:
> Or is that specific port range necessary?

As far as I understand it, yes, if we want to avoid any timing issues with the adb not discovering the emulators we should stick to the range adb suggests

vaslabs:
There's still some flakiness with these ports, let's run the pipeline one more time, I'm updating the branch

PR #6593, opened 2026-01-15
Add support for `@nonBootstrapped` commands that run without the Mill bootstrap process completing

lihaoyi:
Commands annotated with `@nonBootstrapped` are run as early in the `MillBuildBootstrap` process as possible, speeding them up on clean repos and allowing them to run even if your `build.mill` is broken. This is useful for tasks such as `./mill version`, `./mill init`, `./mill shutdown`, etc. that do not depend on the current checkout.

To implement this in `MillBuildBootstrap`, at every meta-level we attempt to resolve the user query. If it resolves to tasks that are all `@nonBootstrapped`, we cut the bootstrapping process early and just run those tasks. `@nonBootstrapped` is an annotation rather than a runtime value like `persistent` so we do not need to instantiate the tasks to perform that check, instead we can look at the annotation using reflection. This is important since this check happens every command, so we want to minimize the actual logic that gets run to perform the check

`./mill version` is no longer an effective way of validating the build compiles, for now I changes all our tests to use `./mill resolve _` instead

Users can also define their own `@nonBootstrapped` tasks that behave similarly.

### DIFF
--- a/runner/daemon/src/mill/daemon/MillBuildBootstrap.scala
+++ b/runner/daemon/src/mill/daemon/MillBuildBootstrap.scala
@@ -7,7 +7,7 @@ import mill.api.daemon.internal.{
   PathRefApi,
   RootModuleApi
 }
-import mill.api.{BuildCtx, Evaluator, Logger, PathRef, Result, SelectMode, SystemStreams, Val}
+import mill.api.{BuildCtx, Logger, PathRef, Result, SelectMode, SystemStreams, Val}
 import mill.constants.CodeGenConstants.*
 import mill.constants.OutFiles.OutFiles.{millBuild, millRunnerState}
 import mill.internal.Util
--- a/core/api/src/mill/api/Discover.scala
+++ b/core/api/src/mill/api/Discover.scala
@@ -23,7 +23,7 @@ final class Discover(val classInfo: Map[Class[?], Discover.ClassInfo]) {
    * Looks up the class hierarchy to find if any parent class has this task marked as nonBootstrapped.
    */
   private[mill] def isNonBootstrapped(cls: Class[?], name: String): Boolean = {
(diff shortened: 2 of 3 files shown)


lefou:
I'm currently using `mill --meta-level -1 <cmd>` in the same spirit, but the explicit annotation is even better. We probably also want to annotate commands in the external modules for formatters and jdk tool runners.

runner/daemon/src/mill/daemon/MillBuildBootstrap.scala
```diff
@@ -81,205 +82,161 @@ class MillBuildBootstrap(
...
+              nestedState.frames.headOption.fold(Map())(_.buildOverrideFiles)
+
+          // Short-circuit for @nonBootstrapped tasks:
+          // When root build.mill has compile errors, tasks marked @nonBootstrapped
+          // (like `version`, `clean`, `shutdown`) can still run on the meta-build.
+          // This is only possible when:
+          // - No explicit --meta-level (targeting root build at depth 0)
+          // - A compiled meta-build exists (nestedState.frames.nonEmpty)
+          val canPotentiallyShortCircuit =
+            requestedMetaLevel.isEmpty && nestedState.frames.nonEmpty
```
lefou: Isn't it enough to check, that the effective `requestedMetaLevel` is not lower than the current. I think `--meta-level -1|<max-level>` should still allow a short-circuit.

lihaoyi: This whole file is a mess right now, I'm trying my best to clean it up but I'm sure there's a lot of tangled logic

lefou: I feel you and totally agree. I spend way to many hours there too.

Issue #6572, opened 2026-01-13
Don't use `~/.cache/mill/download/` for `mill-completion.sh`


lefou:
This directory is managed by the `mill.sh` for caching Mill binary artifacts. Missing files will be detected and re-downloaded, which is what users expect for cache locations. The `mill-completion.sh` script on the other hand was explicitly installed by the user. Once it's removed, it will not re-appear automatically, ruining the CLI DX after a cache cleanup.

PR #6595, opened 2026-01-15
Change default location of the shell completion script

lefou:
Preferred location is now `$XDG_DATA_HOME/mill/completion/mill-completion.sh`,
or `~/.local/share/mill/completion/mill-completion.sh`.

The previous location under `$XDG_CACHE_HOME` (`~/.cache`) should only be used for non-essential cached files, which can be removed at any time by the user.

Fix https://github.com/com-lihaoyi/mill/issues/6572

Pull request: https://github.com/com-lihaoyi/mill/pull/6595


lefou:
CI test failure looks unrelated.

Issue #4596, opened 2025-02-20
`init <Giter8 template>` does not work anymore (version 0.12.8)


carlosrogue:
When checking https://github.com/com-lihaoyi/mill-scala-hello.g8 I realize this does not work anymore:
`mill init com-lihaoyi/mill-scala-hello.g8` but like this `MILL_VERSION=0.11.11 mill version`. With my local mill version `0.12.8` I have to run it as interactive and it works: `mill -i init com-lihaoyi/mill-scala-hello.g8`.
Should the `-i` flag be added to the init module docs: Run `mill -i init <Giter8 template>` or should the run mode when is giter8 template be internally changed?

lihaoyi:
Do you have an error message we can see?

carlosrogue:
```shell
$ mill init com-lihaoyi/mill-scala-hello.g8
No mill version specified.
You should provide a version via '.mill-version' file or --mill-version option.
Using mill version 0.12.8
[build.mill-61/65] compile
[build.mill-61] [info] compiling 1 Scala source to /home/carlosrogue/Projects/2025/g8/out/mill-build/compile.dest/classes ...
[build.mill-61] [info] done compiling
[1/1] init
[1] Creating a new project...
[1] A minimal Scala project built with Mill 
[1] 
[1] name [Scala Hello World]: [1] 
[1] Interrupting...
[1] 
[1/1] ================ init com-lihaoyi/mill-scala-hello.g8 =============== 19s
1 tasks failed
init java.lang.Exception: 1 tasks failed
mill.scalalib.giter8.Giter8Module.init os.SubprocessException: Result of /home/carlosrogue/.cache/coursier/jvm/openjdk@1.17.0/bin/java…: 1

    os.proc.call(ProcessOps.scala:230)
    mill.util.Jvm$.callProcess(Jvm.scala:95)
    mill.scalalib.giter8.Giter8Module.$anonfun$init$1(Giter8Module.scala:36)
    mill.define.Task$TraverseCtx.evaluate(Task.scala:225)
    mill.main.MainModule.$anonfun$init$1(MainModule.scala:603)
    mill.define.Task$TraverseCtx.evaluate(Task.scala:225)
```

carlosrogue:
@lihaoyi something like this should succeed in `MillInitTests`:
```scala
    test("Mill init works for g8 templates without explicit interactive") - integrationTest { tester =>
      import tester._
      // pass "example" to eval from stdin
[...]

PR #6600, opened 2026-01-16
Automate re-running of interactive tasks in `--no-daemon` mode to remove need for `-i`

lihaoyi:
This PR removes the need for manually passing in `-i` for commands that need to be interactive, replacing it with an `interactive = true` flag on the command. When run in daemon mode, Mill's daemon skips these interactive commands, passes them back to the launcher, which then runs them in a subsequent `MillNoDaemonMain` subprocess.

This sequential-run strategy has a bunch of edge cases. We tweaked the logging output to try and make it less obvious there's two runs, but there are still issues like `mill-profile.log`/`mill-chrome-profile.log`/etc. being over-written by the subsequent run and be generally useless, or re-runs not working with evaluator commands such as `show`. But that can be fixed in future work

Covered by an integration test checking that `interactive = true` commands are run with `MillNoDaemonMain` even when no `-i`/`--interactive`/`--no-daemon` flag is passed, while normal commands are still run with `MillDaemonMain`

Fixes https://github.com/com-lihaoyi/mill/issues/4596

### DIFF
--- a/libs/scalalib/src/mill/scalalib/ScalaModule.scala
+++ b/libs/scalalib/src/mill/scalalib/ScalaModule.scala
@@ -2,7 +2,7 @@ package mill
 package scalalib
 
 import mill.util.JarManifest
-import mill.api.{BuildCtx, DummyInputStream, ModuleRef, PathRef, Result, Task}
+import mill.api.{BuildCtx, ModuleRef, PathRef, Result, Task}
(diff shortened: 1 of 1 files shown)


lihaoyi:
Superseded by https://github.com/com-lihaoyi/mill/pull/6602

PR #6602, opened 2026-01-16
Make Mill Launcher/Daemon protocol use `libs/rpc/`, move `console`/`repl`/`jshell` subprocesses to MillLauncherMain

lihaoyi:
This PR replaces the traditional `ProxyStreams`-based interface between the Mill launcher and daemon with a `libs/rpc`-based interface. This allows us to extend the interface to let the daemon code call back into the launcher during execution, which we use to allow the daemon to run subprocesses like the Scala REPL or JShell in the launcher where it has direct access to the user's raw IO streams. 

This allows the user to do `./mill repl` or `./mill foo.jshell` without needing `-i`, since even if the bulk of Mill evaluation happens on the daemon, the `repl` or `jshell` tasks can be send back to the launcher to run via `Jvm.callInteractiveProcess` or `Jvm.runInteractiveCommand`. These methods send RPCs back to the launcher specifying the subprocess configuration and block until the launcher has finished running the subprocess and returns and exit code.

The downside is that it means we would need to run Scala code in the launcher, which would make the non-native launcher have worse startup latencies on Windows-ARM which Graal does not support (~250ms -> 500ms). Graal native launcher latencies are not affected (~50ms), and the slowdown for Windows-ARM users is probably small enough to be tolerable

Steady-state performance probably also suffers, since Mill's launcher/daemon now exchange line-delimited-JSON rather than raw bytestreams, but that can probably be optimized over time and Mill typically isn't logging huge volumes anyway

### DIFF
--- a/core/constants/package.mill
+++ b/core/constants/package.mill
@@ -15,7 +15,8 @@ object `package` extends MillStableJavaModule with BuildInfo {
     BuildInfo.Value("millVersion", build.millVersion(), "Mill version."),
     BuildInfo.Value("millBinPlatform", build.millBinPlatform(), "Mill binary platform version."),
     BuildInfo.Value("millDocUrl", Settings.docUrl),
-    BuildInfo.Value("coursierJvmIndexVersion", Deps.coursierJvmIndexVersion)
+    BuildInfo.Value("coursierJvmIndexVersion", Deps.coursierJvmIndexVersion),
+    BuildInfo.Value("defaultJvmVersion", build.defaultMillJvmVersion())
   )
 
   def jvmId = "11"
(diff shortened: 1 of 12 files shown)


lefou:
> The downside is that it means we would need to run Scala code in the launcher, which would make the non-native launcher have worse startup latencies on Windows-ARM which Graal does not support (~250ms -> 500ms). Graal native launcher latencies are not affected (~50ms), and the slowdown for Windows-ARM users is probably small enough to be tolerable

On the upside, we could use the same CLI parser in the client, making code more consistent and removing strange requirements like `-i` must given as the first arg.

PR #6610, opened 2026-01-17
Rebootstrap for real

lihaoyi:
Follow up to https://github.com/com-lihaoyi/mill/pull/6609, attempting to follow the newer convention of relying on the bootstrap scripts to define the mill version


lefou:
This breaks (not only) my workflow! 

I don't like this "new" workflow. You claim it being a convention, but it's not one you find elsewhere. It's against good practices separating configuration changes from downloaded executable code. It breaks my personal workflow and the workflow of all users who prefer to use a "locally installed" mill script instead of a `./mill` script.

lefou:
The fact that the current version of the `mill` script does not have a identifying header and you want users to edit that files makes it impossible to anyone to find out which versions it was derived from, since you can't fingerprint it. How are users exepected to maintain it, how should they detect, they need to update the mill script? That workflow is flawed.

lefou:
* Opened #6618

Issue #6637, opened 2026-01-20
Android: Androidtodo unit tests are broken


vaslabs:
```
411-3] Test com.example.android.architecture.blueprints.todoapp.tasks.TasksViewModelTest.clearCompletedTasks_clearsTasks failed: java.lang.NoClassDefFoundError: com/example/android/architecture/blueprints/todoapp/R$string, took 0.004 sec
```

PR #6638, opened 2026-01-20
Fix unit tests of androidtodo

vaslabs:
Fixes https://github.com/com-lihaoyi/mill/issues/6637

by introducing

`def androidIncludeAndroidResources: Boolean = false`

as equivalent of AGP's

`testOptions.unitTests { isIncludeAndroidResources = true }`

As in the original configuration of architecture samples that can be found [here](https://github.com/android/architecture-samples/blob/main/app/build.gradle.kts#L63)

### Manual test

image

### Example test

image


vaslabs:
this androidtodo has broken a few times without us noticing, let's take a note to at least enable the job only on our fork @souvlakias

lefou:
> this androidtodo has broken a few times without us noticing, let's take a note to at least enable the job only on our fork @souvlakias

You could prepare a CI workflow that runs on a schedule, but not with each merge/PR.

Issue #6157, opened 2025-11-10
Add a header to bootstrap script with some infos


lefou:
Since #6117, the bootstrap script don't have any header comment. This makes working with them inconvenient, esp. when maintaining older projects, and you want to know where the script is coming from or which version it is.

Script should contain the following info:
* Mill Project + URL
* A link to the download page
* A link to the online documentation of the script (essentially what was removed as part of #6117)
* A version

lefou:
* Fixed by https://github.com/com-lihaoyi/mill/pull/7510

PR #6647, opened 2026-01-21
Check glibc version on Linux in bootstrap script

jackkoenig:
If the version is too old, fall back to the JVM client.

This is an alternative resolution to https://github.com/com-lihaoyi/mill/discussions/5568. Statically linking things is a pain and has pitfalls. A simple alternative for users on older Linux distributions is just to have the bootstrap script automatically fall back to JVM client if glibc is too old.

### DIFF
--- a/dist/scripts/package.mill
+++ b/dist/scripts/package.mill
@@ -21,7 +21,7 @@ object `package` extends mill.Module { scripts =>
       "mill-download-cache-unix" -> "~/.cache/mill/download",
       "mill-download-cache-win" -> "%USERPROFILE%\\\\.mill\\\\download",
       "template-file" -> templateFile().path.relativeTo(BuildCtx.workspaceRoot).toString,
-      "glibc-version" -> glibcVersion(),
+      "glibc-version" -> glibcVersion()
     )
     def substitutionMarkers: T[(String, String)] = ("{{{", "}}}")
     def inRepoDir: Task[os.SubPath] = Task.Anon { os.sub / finalName() }


jackkoenig:
This is technically failing CI, but I think it's failing in the same way that its parent commit is.

lefou:
Looks reasonable, but you need to apply this change to `dist/scripts/src/mill.sh`, as `mill` is generated from it. 

Related:
* https://github.com/com-lihaoyi/mill/issues/6157

jackkoenig:
> Looks reasonable, but you need to apply this change to dist/scripts/src/mill.sh

Thanks for the tip, I've updated that script instead. I also updated the determination of the glibc version to be a substitution so that (as long as you're running on linux), the glibc version will be kept in sync with the glibc version on the system. This means that as long as the native binary is published on the same version of linux as the bootstrap script, the version will always be in sync.

lihaoyi:
@jackkoenig can you manually run some tests to measure how much latency the extra subprocess calls take in the happy and non-happy paths? It probably is fine, but I don't want to find we're accidentally calling some 100ms+ subprocesses early in the bootstrap script

jackkoenig:
Good point @lihaoyi, will do.

jackkoenig:
@lihaoyi here's the data:

I benchmarked by adding at the top:
```
start=$(date +%s.%N)
```
and at the bottom right before the actual `exec "${MILL}"`
```
end=$(date +%s.%N)
elapsed=$(echo "$end - $start" | bc)
echo "${elapsed}"
  
exit 0
```

**All units in milliseconds**

"old" is the current bootstrap script, "new" is the bootstrap script regenerated with the changes in this PR.

  | RHEL8[^1] |   | MBP M2 Pro |   | Ubuntu 24.04[^2]  | 
-- | -- | -- | -- | -- | -- | --
  | old | new | old | new | old | new
Average | 2.71 | 3.32 | 9.96 | 9.54 | 2.02 | 5.30
Stddev | 0.29 | 0.14 | 1.02 | 3.04 | 0.72 | 1.99

So it is noticeably slower, I'm not stoked about the 2 => 5.3 ms average on a typical Ubuntu system (at least as best as I can model via Docker). Fast enough though maybe?

[^1]: x86_64, old glibc
[^2]: Ubuntu 24.04 is a Docker container running on my MBP with M2 Pro so is ARM64

lihaoyi:
should be fast enough I think

lihaoyi:
thanks!

PR #6654, opened 2026-01-23
Scala 3.8.1

lihaoyi:


lihaoyi:
Green in https://github.com/lihaoyi/mill-1/actions/runs/21270312211

integration/feature/gen-idea/resources/extended/idea/libraries/scala_SDK_3_0_2.xml
```diff
@@ -1,7 +1,7 @@
 <component name="libraryTable">
-    <library name="scala-SDK-3.0.2" type="Scala">
+    <library name="scala-SDK-<version>" type="Scala">
         <properties>
-            <language-level>Scala_3_0</language-level>
+            <language-level>Scala_3_X</language-level>
```
lefou, line 4: I think, that's something we want to assert without a wildcard. It should change infrequently, but a correct language level is important for correct IDE support.

PR #6664, opened 2026-01-23
Fix Documentation Commands Reference

gamlerhart:
Fix the documentation to use 'website.fastPages' and `website.localPages`


lefou:
Thank you!

PR #6668, opened 2026-01-24
Generalize `AllEvaluators` logic

lihaoyi:
This PR is an extension of https://github.com/com-lihaoyi/mill/pull/6593, allowing the hardcoded `mill.idea/` and `mill.eclipse/` support to be removed from `MillMain0` and instead replaced by commands annotated with `@allEvaluatorsCommand`


lefou:
While the PR seems reasonable, it introduced an alternative access to the Evaluator. The question is, how it differs from evaluator commands (receiving the evaluator via a parameter injection). Why do we have two different ways to access them?

Issue #6651, opened 2026-01-22
Android: Android tests (dexing) on Jetpack compose apps seem to fail on Macs (M4)


vaslabs:
I'm still gathering information for this

vaslabs:
hm, this seems about case sensitivity:

androidx.compose.material.icons.outlined.AddchartKt actually defines class androidx.compose.material.icons.outlined.AddChartKt

so it may exist on windows as well

vaslabs:
On Linux (btrfs):

<img width="1326" height="109" alt="Image" src="https://github.com/user-attachments/assets/48ba3398-8e19-4cfe-ae43-021cf02367d5" />

Maybe this is just an issue with the old compose bom version that we still have in the example

 mvn"androidx.compose:compose-bom:2024.12.01"

we can update versions and check

vaslabs:
I'm able to reproduce this on an M2 as well (the architecture doesn't seem relevant, just the file system case insensitive mode), I'm working on a fix

PR #6686, opened 2026-01-26
Fixes building the main android jar for android tests compile classpath on case insensitive systems

vaslabs:
Fixes https://github.com/com-lihaoyi/mill/issues/6651

### Solution

Instead of building the android Jar by passing a minify directory to R8 (and thus having it explode the classfiles in the directory), we pass it a zip as an output path, to make it independent of filesystem case-sensitivity.

### Mac Test

Screenshot 2026-01-26 at 2 36 28 PM

Same issue should exist on Windows or any other case insensitive system


vaslabs:
we could do the same for dex mode, but I find it very convenient to inspect dex files directly without having a zip in the middle for the androidDex job.

Maybe something for later

vaslabs:
failure should be unrelated. should be ready for review + I tested on a mac (added screenshot), androidTest.androidTestApk now works and the tests run

Issue #6696, opened 2026-01-27
`def foo: T[Option[V]] = Task{ None }` cannot be properly overridden by YAML config


lihaoyi:
```scala
def jlinkModuleVersion: T[Option[String]] = Task { None }
```
```yaml
jlinkModuleVersion: "1.0"
```

This fails with

```
jlinkModuleVersion: "1.0"

foo.jlinkModuleVersion Failed de-serializing config override: expected unit got string
```

The problem is that `Task { None }` infers `Task[None.type]{ None }` rather than `Task[Option[String]]`, which ends up having the wrong uPickle deserializer.

Workaround is to use

```scala
def jlinkModuleVersion: T[Option[String]] = Task { Option.empty[String] }
```

PR #6700, opened 2026-01-27
Make `Task[V]` invariant

lihaoyi:
Should fix https://github.com/com-lihaoyi/mill/issues/6696 by avoiding the pathological `None.type` type parameter inference. Converting all the `Option.empty`s back to `None`s seems to not break things with this change

Not sure if this will break MIMA or other backwards compatibility constraints, but at least it seems to pass compile


lihaoyi:
Seems like MIMA is happy so probably going to merge this

lihaoyi:
Thinking about this, having `Task[V]` be invariant may be more correct because `Task[None.type]` is not behaviorally a subclass of `Task[Option[String]]`: `Task[Option[String]]` can accept either `"value"` or `null` as JSON input, while `Task[None.type]` can only accept `null`. 

So a covariant `Task[+V]` treating `Task[None.type]` as an acceptable substitute for `Task[Option[String]]` is incorrect, and an invariant `Task[V]` requiring an exact type param match for `Task[Option[String]]` is the only correct variance

PR #6715, opened 2026-01-29
Rename `.shared` tests to `.fast`, make them the default when running tests outside of CI.

lihaoyi:
With this PR `integration.foo.bar` or `example.foo.bar.qux` defaults to the same mode as is tested in CI: some are `.fast` (renamed from `.shared`), some are `.packaged`, `.native`, `.proguard`, etc.. This should make it much easier to run the right tests when testing locally, avoiding divergence compared to what ends up running in production, and for the `.fast` tests it makes it a lot faster to run them locally (e.g. `./mill -j1 example.javalib.basic._` goes from 120s to 40s)

### DIFF
--- a/libs/init/test/src/mill/init/InitModuleTests.scala
+++ b/libs/init/test/src/mill/init/InitModuleTests.scala
@@ -90,12 +90,13 @@ object InitModuleTests extends TestSuite {
         errStream = new PrintStream(OutputStream.nullOutputStream(), true)
       ).scoped { evaluator =>
 
-        val results = evaluator.evaluator.execute(
-          Seq(initmodule.init(Some("scalalib/basic/3-simple")))
-        ).executionResults
+        val results = evaluator
+          .evaluator
+          .execute(Seq(initmodule.init(Some("scalalib/basic/1-simple"))))
+          .executionResults
 
         val expected =
-          "Downloading example from https://repo1.maven.org/maven2/com/lihaoyi/mill-dist/SNAPSHOT/mill-dist-SNAPSHOT-example-scalalib-basic-3-simple.zip."
+          "Downloading example from https://repo1.maven.org/maven2/com/lihaoyi/mill-dist/SNAPSHOT/mill-dist-SNAPSHOT-example-scalalib-basic-1-simple.zip."
 
         // Make sure the download URL logged to the terminal looks reasonable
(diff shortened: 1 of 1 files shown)


integration/package.mill
```diff
@@ -170,14 +174,14 @@ object `package` extends mill.Module {
...
+    // Set `fast.daemon` as the default way to run integration and example tests
     override def defaultTask(): String = "testForked"
     def testForked(args: String*): Task.Command[(msg: String, results: Seq[TestResult])] =
       Task.Command {
-        packaged.daemon.testForked0(args*)()
+        fast.daemon.testForked0(args*)()
       }
 
-    object shared extends IntegrationLauncherModule {
+    object fast extends IntegrationLauncherModule {
```
lefou, line 184: Can we have Scaladoc explaining the key properties of this module?

integration/package.mill
```diff
@@ -219,12 +223,27 @@ object `package` extends mill.Module {
...
-  object bootstrap extends Cross[IntegrationCrossModule](build.listCross)
+  object dedicated extends Cross[DedicatedIntegrationCrossModule](build.listCross)
+  object multimode extends Cross[MultimodeIntegrationCrossModule](build.listCross)
+  object bootstrap extends Cross[BootstrapIntegrationCrossModule](build.listCross)
   object migrating extends Cross[MigrationIntegrationCrossModule](build.listCross)
   object manual extends Cross[IntegrationCrossModule](build.listCross)
+
+  trait DedicatedIntegrationCrossModule extends IntegrationCrossModule {
+    override def testForked(args: String*): Task.Command[(msg: String, results: Seq[TestResult])] =
+      Task.Command { proguard.daemon.testForked0(args*)() }
```
lefou, line 234: This looks like boilerplate we could avoid with 
* https://github.com/com-lihaoyi/mill/issues/5402

integration/package.mill
lihaoyi: It's covered in the `developer.adoc` file, so hopefully people and bots will see it. I `@included` it in CLAUDE.md and it seems the bot is aware of all the stuff in that file.

lefou: I was more thinking about our human users. I mean, `mill inspect` does not magically read `developer.adoc`. How about "/** Fast integration test suite. See developer.adoc for details. */".

PR #6716, opened 2026-01-29
feat: Implement Library Shading Support (Mill #3815)

pantha704:
## Summary
Implemented library shading support for `JavaModule` in Mill (Issue #3815). This allows users to include dependencies in their main JAR file while relocating their packages to avoid classpath conflicts.

## Changes Made
- Added `shadedIvyDeps` configuration to specify dependencies to be shaded.
- Added `shadeRules` configuration to specify relocation rules.
- Updated `jar` task to use `Assembly.create` (via `jarjar-abrams`) when shading is enabled.
- Updated `resolvedMvnDeps` to include shaded dependencies for compilation (so user code can reference them) but excludes them from the published POM (since they are bundled).
- Updated `runClasspath` to include the shaded JAR instead of raw dependencies when shading is active.
- Added `ReadWriter` to `Assembly.Rule` to support upickle serialization required for Mill's task caching.
- Updated `dependencies.adoc` with a "Shading / Relocating Dependencies" section and usage example.

## Verification
- Created `integration/feature/shading` test suite.
- Verified locally with `mill integration.feature.shading.test`.

### DIFF
--- a/integration/feature/shading/resources/build.mill
+++ b/integration/feature/shading/resources/build.mill
@@ -3,12 +3,12 @@ import mill._, javalib._
 import mill.javalib.Assembly
 
 object hello extends JavaModule {
-  def ivyDeps = Seq(
-    ivy"commons-io:commons-io:2.11.0"
+  def mvnDeps = Seq(
+    mvn"commons-io:commons-io:2.11.0"
   )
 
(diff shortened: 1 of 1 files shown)


lihaoyi:
Doesn't look promising

PR #6721, opened 2026-01-30
Clean up publishing docs

lihaoyi:
- Delete the `== Publishing to Sonatype Maven Central` example since it just duplicated the other sections
- Delete `Instructions for Publishing to Maven Central via Legacy OSSRH (Deprecated)` since it's been deprecated for 6 months and it's super confusing to duplicate the earlier section
- Add a `VcsVersionModule` that makes use of `VcsVersion`, to allow usage in `.mill.yaml` files
- Add an example demonstrating usage of `VcsVersionModule`
- Drop reference to https://github.com/ckipp01/mill-ci-release and https://github.com/lefou/mill-vcs-version, which have been largely upstreamed into Mill itself


website/docs/modules/ROOT/pages/extending/thirdparty-plugins.adoc
lefou, line 1: We typically don't remove these but document here, where a replacement can be found, e.g. link to in-repo docs. This helps users migrating or looking up stuff they remembered.

Issue #6412, opened 2025-12-18
Mill Scala Scripts don't work for Scala 2


lihaoyi:
```
compiling 3 Scala sources to out/mill-build/compile.dest/classes ...
8714]   done compiling
8714]   Compiling compiler interface...
8714]   compiling 1 Scala source to out/Foo.scala/compile0.dest/classes ...
8714]   [error] Foo.scala:17:1
8714]   type main = mainargs.main; def _millScriptMainSelf = this; object _MillScriptMain { def main(args: Array[String]): Unit = this.getClass.getMethods.find(m => m.getName == "main" && m.getParameters.map(_.getType) == Seq(classOf[Array[String]]) && m.getReturnType == classOf[Unit]) match{ case Some(m) => m.invoke(_millScriptMainSelf, args); case None => mainargs.Parser(_millScriptMainSelf).runOrExit(args) }}
8714]   ^
8714]   expected class or object definition
8714]   
8714] [error] Foo.scala:17:28
8714]   type main = mainargs.main; def _millScriptMainSelf = this; object _MillScriptMain { def main(args: Array[String]): Unit = this.getClass.getMethods.find(m => m.getName == "main" && m.getParameters.map(_.getType) == Seq(classOf[Array[String]]) && m.getReturnType == classOf[Unit]) match{ case Some(m) => m.invoke(_millScriptMainSelf, args); case None => mainargs.Parser(_millScriptMainSelf).runOrExit(args) }}
8714]                              ^
8714]   expected class or object definition
8714]   
8714] [error] two errors found
8714]   [error] Foo.scala:compile0 Compilation failed
```

PR #6723, opened 2026-01-30
Add a nice error for Scala 2 scripts which aren't supported

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/6412


.codex/config.toml
```diff
@@ -0,0 +1,16 @@
+# Example codex/config.toml. Doesn't work in repo, you need to copy it to
```
alexarchambault, line 1: Seems this added a file that's specific to your machine

PR #6726, opened 2026-01-30
Try and clean up publishing infrastructure

lihaoyi:
Attempts to fix https://github.com/com-lihaoyi/mill/issues/5538. 

> Before
>
> - Publishing data split across two representations:
>     - PublishInfo (rich metadata) used for local Ivy/M2 publishing.
>     - PublishData (artifact + filename mapping only) used for remote publishing.
> - The remote path only saw filenames, so it had to reconstruct metadata by parsing the filenames, which is fragile and
>   duplicates logic.
> - File selection logic (pom/jar/sources/docs/extras) existed in more than one place, including custom logic in Android.
>
> After
>
> - PublishData is still the main carrier, but now optionally includes pom and publishInfos, so rich metadata can flow all
>   the way to remote publishing.
> - Remote Maven publishing now uses publishInfos/pom when present, falling back to filename parsing only when they’re
>   missing.
> - The payload map construction is centralized via PublishInfo metadata, and publishArtifactsPayload/
>   publishArtifactsDefaultPayload now share a helper so logic isn’t duplicated.
> - Android publishing defers to the shared default payload logic.
> - Construction of PublishData is unified in one helper (publishArtifactsData) for both publishArtifacts and
>   publishSonatypeCentral.
> - Pattern matches in contrib publishers updated to the new case class shape.
>
> Why this is an improvement
>
> - The metadata is preserved end‑to‑end when available, eliminating the need for filename heuristics in the common case.
> - The source of truth for “what is published” is unified, so users have fewer places to override and fewer implicit
>   dependencies on internal paths.
> - The fallback heuristic still exists, so legacy callers remain supported.
>
> Notes on API shape
>
> - PublishData now has optional metadata fields (pom, publishInfos).
> - The old Map‑based constructor is kept as a JVM apply (for bincompat) but exposed as applyFromMap in Scala due to
>   unroll macro constraints, and it’s deprecated to steer callers to the richer metadata path.
>
> Key codepaths updated
>
> - PublishModule.publishArtifacts and SonatypeCentralPublishModule.publishSonatypeCentral now build the same PublishData
>   structure via publishArtifactsData.
> - MavenPublish uses PublishData.publishInfos/pom when present, otherwise falls back to filename parsing.
> - PublishInfo is still the metadata type and is now carried in PublishData.
>

### DIFF
--- a/libs/javalib/src/mill/javalib/MavenPublish.scala
+++ b/libs/javalib/src/mill/javalib/MavenPublish.scala
@@ -48,21 +48,24 @@ private[mill] trait MavenPublish {
   ): Unit = {
     val uri = if (isSnapshot) snapshotUri else releaseUri
     val payloadAsMap = publishData.payloadAsMap
+    // TODO 2.0.0: Remove filename-based metadata inference once PublishData always carries publishInfos.
     val publishInfos =
       if (publishData.publishInfos.nonEmpty) publishData.publishInfos
       else {
-        payloadAsMap.iterator.map { case (name, pathRef) =>
-          publish.PublishInfo.IvyMetadata
-            .parseFromFile(name.toString, publishData.meta.id, publishData.meta.version)
-            .toPublishInfo(pathRef)
-        }.toList
+        payloadAsMap
+          .map { case (name, pathRef) =>
+            publish.PublishInfo.IvyMetadata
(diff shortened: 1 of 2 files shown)


lihaoyi:
Overall I think this is probably a bit too messy to do while trying to preserve backwards compat, probably need to postpone this to Mill 2.0.0

PR #6734, opened 2026-02-01
Test selectiveInputs

lihaoyi:
- This PR should not run the Kotlin tests, despite touching `MillLauncherMain.scala`, because of `selectiveInputs`

- This PR should run the Javascript tests, because we touched `javascriptlib` source files directly


lihaoyi:
Seems to work!

lihaoyi:
Actually maybe not, some tests that should be running like example.scalalib._ don't seem to be running

Issue #6762, opened 2026-02-03
Incorrect scala-library Resolution (Mixed Scala 2.13 + 3.8+ module dep)


jilen:
For `moduleA`(scala 2.13) depends on `moudleB(scala 3.8+)`

The compileClasspath of moduleA is
```
/org/scala-lang/scala-library/2.13.18/scala-library-2.13.18.jar
/org/scala-lang/scala3-library_3/3.8.1/scala3-library_3-3.8.1.jar
```
This is not correct, `scala-library` should also be `3.8.1`
since the `scala3-library` after 3.8  is empty, and its contents are moved to `scala-library`

lihaoyi:
The fix for now would pin it to `/org/scala-lang/scala-library/2.13.18/scala-library-2.13.18.jar`, since in general setting `scalaVersion` pins the `scala-*` jars to that version regardless of transitive dependencies. We may want to change this in future, but that is how things are now

Issue #6405, opened 2025-12-17
GenIdea: Multiple scala-libraries end up in meta-build module and break it


lefou:
It seems, having multiple scala-libraries on the classpath confuses IDEA.

```console
> rm -r .idea

> mill mill.idea.GenIdea/
mill mill.idea.GenIdea/ 
1] version
1.1.0-RC3
1/1] ============================== mill.idea.GenIdea/ ==============================
Analyzing modules ...
Cleaning obsolete IDEA project files ...
Writing 1292 IDEA project files to /home/lefou/work/Comfis/comfis-mill-1/.idea ...
...

> cat .idea/mill_modules/mill-build.iml | grep -E "scala.*-(library|SDK)"
        <orderEntry type="library" name="scala-SDK-3.8.0-RC2" level="project"/>
        <orderEntry type="library" scope="PROVIDED" name="scala-library-2.13.12.jar" level="project"/>
        <orderEntry type="library" name="scala-library-3.8.0-RC2.jar" level="project"/>
        <orderEntry type="library" scope="PROVIDED" name="scala3-library_3-3.3.3.jar" level="project"/>
        <orderEntry type="library" name="scala3-library_3-3.8.0-RC2.jar" level="project"/>
```

In the image below, you can see that all type from Mill are red. Code navigation doesn't work.

<img width="803" height="625" alt="Image" src="https://github.com/user-attachments/assets/ba194112-b3de-4c45-ac73-65cf9ff9fcdd" />

Once I remove the unwanted versions manually, the red error marker vanish (mostly).

```diff
         <orderEntry type="library" scope="RUNTIME" name="scala-collection-compat_2.13-2.13.0.jar" level="project"/>
         <orderEntry type="library" name="scala-collection-compat_3-2.12.0.jar" level="project"/>
-        <orderEntry type="library" scope="PROVIDED" name="scala-library-2.13.12.jar" level="project"/>
         <orderEntry type="library" name="scala-library-3.8.0-RC2.jar" level="project"/>
         <orderEntry type="library" name="scala-reflect-2.13.15.jar" level="project"/>
         <orderEntry type="library" scope="RUNTIME" name="scala-xml_2.13-2.4.0.jar" level="project"/>
         <orderEntry type="library" name="scala-xml_3-2.4.0.jar" level="project"/>
-        <orderEntry type="library" scope="PROVIDED" name="scala3-library_3-3.3.3.jar" level="project"/>
[...]

PR #6763, opened 2026-02-03
Fix Scala 3.8 scala-library resolution

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/6762 and fixes https://github.com/com-lihaoyi/mill/issues/6405. Attempts to clean up the conditional logic around selecting the appropriate scala library artifacts


lefou:
> > * The Scala 3.8+ runtime standard library is now scala-library at the Scala 3 version. The correct resolution behavior is:
> >   
> >   * For Scala 3.0–3.7: depend on scala3-library (which brings scala-library 2.13 transitively).
> >   * For Scala 3.8+: depend on scala-library directly at the Scala 3 version.

Can we have a rough copy of that algo in as a comment in the code?

lihaoyi:
I tried to clean up the code and add comments, hopefully it's a bit less opaque than it was before

Issue #6770, opened 2026-02-03
GraalVM Native Image launcher eats some Java options


jackkoenig:
It appears that the GraalVM Native image launcher will consume some options such that they are not available to user code.

For example, for a Mill build using ScalaTest, one might pass options to ScalaTest via `-D<key>=<value>`:
```
./mill myproject.test -Dfoo=bar
```
If you are using the native image launcher (which is the default), `-Dfoo=bar` does not make it to ScalaTest. However, if you use the JVM launcher, it does.

lihaoyi:
Thanks @jackkoenig! Would be great if you could minimize the issue and open an upstream ticket in https://github.com/oracle/graal as well

jackkoenig:
@lihaoyi It looks like there is an existing option to tell Native Image to not parse the options: `-H:-ParseRuntimeOptions` (see https://github.com/oracle/graal/issues/779#issuecomment-436274256). So I think the fix is to use that in Mill's native image generation

lihaoyi:
@jackkoenig great! Could you try it out by adding the flag here https://github.com/com-lihaoyi/mill/blob/23f07adf868f354ceb777863f94392514140706e/dist/package.mill#L507 and building a `dist.native.installLocal` executable? If it works you can send a PR

PR #6774, opened 2026-02-04
Add -H:-ParseRuntimeOptions to native image options

jackkoenig:
Fixes https://github.com/com-lihaoyi/mill/issues/6770

I tested this on my M2 Pro Macbook Pro and confirmed it fixes the issue with propagating `-D<key>=<foo>` arguments to Scalatest.

### DIFF
--- a/dist/package.mill
+++ b/dist/package.mill
@@ -516,7 +516,10 @@ object `package` extends MillJavaModule with Proguard with DistModule {
       // Set UTF-8 encoding to fix Unicode character display issues on Windows
       "-Dfile.encoding=UTF-8",
       "-Dsun.stdout.encoding=UTF-8",
-      "-Dsun.stderr.encoding=UTF-8"
+      "-Dsun.stderr.encoding=UTF-8",
+      // Prevent Native Image runtime from consuming runtime options like -D<key>=<value>
+      // https://github.com/com-lihaoyi/mill/issues/6770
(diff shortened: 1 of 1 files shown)


lihaoyi:
@jackkoenig I released `1.1.1-6-480a23` including this PR if you would like to start using it

jackkoenig:
Thanks @lihaoyi!

if-sc:
The `-H:-ParseRuntimeOptions` flag added here breaks SSL certificate handling for system properties like `-Djavax.net.ssl.trustStore/`.

This flag changes the timing of argument processing in the Native Image launcher. It defers parsing of runtime options until after daemon-runner initialization, whereas they were previously parsed before. This means SSL trustStore settings and other critical JVM properties don't apply early enough to daemon-runner setup, causing SSL handshake failures.

Rather than using a blanket flag that defers all argument processing, we need a more targeted fix that:
- Parses system properties early (before daemon-runner loading) as before
- Selectively allows downstream tools to receive test-specific properties through an alternative mechanism

This would restore SSL certificate functionality while maintaining Scalatest argument propagation.

@lihaoyi consider reverting this change

lihaoyi:
@if-sc can you provide a minimal repro of what broke? that would help us understand the problem you are describing

lefou:
@if-sc This PR is already merged. Please open a new issue.

PR #6782, opened 2026-02-05
Fix "Scala Native linking crash when setting `SourceLevelDebuggingConfig.Enabled`"

lihaoyi:
Fixes https://github.com/com-lihaoyi/mill/issues/6780. Vibe coded, I have no idea why using the `_2.13` jars is necessary, and neither does codex it seems. But it does make the stack trace and hang go away

### DIFF
--- a/libs/scalanativelib/src/mill/scalanativelib/ScalaNativeModule.scala
+++ b/libs/scalanativelib/src/mill/scalanativelib/ScalaNativeModule.scala
@@ -53,10 +53,16 @@ trait ScalaNativeModule extends ScalaModule with ScalaNativeModuleApi { outer =>
       case v @ ("0.4.0" | "0.4.1") =>
         Result.Failure(s"Scala Native $v is not supported. Please update to 0.4.2+")
       case version =>
+        // Workaround for https://github.com/com-lihaoyi/mill/issues/6780:
+        // Scala Native `tools_3` 0.5.10 can crash during linking when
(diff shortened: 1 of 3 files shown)


lihaoyi:
@lolgab could you test this out see if it fixes your original repro? It's kind of strange, but if it fixes the issue and doesn't break anything I guess we can just merge it???

lolgab:
@lihaoyi It works! There might be a bug in the Scala 3 version of the toolchain.
I noticed a warning:
```
117] [warn] Failed to resolve Scala sources for NIR symbols defined in /Users/lorenzo/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/scala-native/nativelib_native0.5_3/0.5.10/nativelib_native0.5_3-0.5.10.jar - they would be unavailable in debugger. You can try to add custom custom source directory or jars to config and try again.
```
but it can be fixed later since it happens only with `SourceLevelDebuggingConfig.Enabled` and it didn't work at all before.

lihaoyi:
@lolgab could you open a ticket on the scala-native repo about this? Just so it doesn't get buried and forgotten, since presumably despite our workaround it's still a problem

lolgab:
@lihaoyi It was a bug in Scala Native and it's being worked on right now for the update to Sbt 2.0: https://github.com/scala-native/scala-native/issues/4772#issuecomment-3858991125

PR #6804, opened 2026-02-08
Correct error message suggesting `jvmId` instead of `jvmVersion`

joan38:
Correct error message for JVM


lefou:
Bad commit message. Please review / edit before merging a PR!

joan38:
Ah I did not realize the PR message is going on the commit message.

PR #6815, opened 2026-02-10
Integrate Jib Docker builder into contrib/docker

nightscape:
Add Jib (Google's container image builder) as the default Docker image build backend in contrib/docker. Jib builds images without requiring a Docker daemon or CLI, producing optimized layered images directly.

DockerConfig now uses Jib by default. The existing Dockerfile-based approach is available via ClassicDockerConfig for users who need Dockerfile features like RUN instructions and VOLUME mounts.

New files:
- JibTypes.scala: Type definitions (JibImageFormat, JibSourceImage, etc.)
- JibLogging.scala: Bridges Mill logger to Jib's event system
- JibBuild.scala: Core Jib build logic (layers, container params, main class detection)

### DIFF
--- a/contrib/docker/readme.adoc
+++ b/contrib/docker/readme.adoc
@@ -3,7 +3,9 @@
 
 Automatically build docker images from your mill project.
 
-Requires the docker CLI to be installed.
+By default, images are built using https://github.com/GoogleContainerTools/jib[Jib],
+which does not require a Docker daemon or CLI. To use the classic Dockerfile-based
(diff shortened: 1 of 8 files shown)


nightscape:
@lihaoyi @lefou ready for a first review 👍

sideeffffect:
Is this derived from https://github.com/GeorgOfenbeck/mill-docker ?
There was a fatal issue(s?) with packaging a multi-module project to an image
 * https://github.com/GeorgOfenbeck/mill-docker/pull/4

Does the code in this PR work well with multi-module projects?

PR #6819, opened 2026-02-11
Ensure local test repo has precedence over local Ivy 2 repo in ITs

alexarchambault:
This gets rid of the need of `MILL_LOCAL_TEST_REPO` and `MILL_USER_TEST_REPO` from https://github.com/com-lihaoyi/mill/pull/5514, and manages those repos solely via the `COURSIER_REPOSITORIES` environment variable. That way, the repositories from `mill-repositories` of https://github.com/com-lihaoyi/mill/pull/6713 automatically get precedence over the former repos.

This addresses issues I've been seeing locally, where some artifacts under `~/.ivy2/local` were being used by the JvmWorker, while they shouldn't.

### DIFF
--- a/runner/launcher/src/mill/launcher/CoursierClient.scala
+++ b/runner/launcher/src/mill/launcher/CoursierClient.scala
@@ -61,7 +61,7 @@ object CoursierClient {
     val millRepositories0 = millRepositories.sorted
 
     val cacheKey =
-      s"${BuildInfo.millVersion}:${overridesRepos.reverse.mkString(":")}:${millRepositories0.mkString("|")}"
+      s"${BuildInfo.millVersion} ${overridesRepos.reverse.mkString("|")}|${millRepositories0.mkString("|")}"
 
     cached[Seq[os.Path]](
       cacheFile = cacheDir(outMode) / "mill-daemon-classpath",


dist/package.mill
```diff
@@ -308,7 +308,9 @@ object `package` extends MillJavaModule with Proguard with DistModule {
 
     // Use the unified localRepo instead of individual publishLocalTestRepo paths
     override def localTestOverridesEnv = Task {
-      Seq("MILL_LOCAL_TEST_REPO" -> localRepo().path.toString)
+      val repos = Seq(localRepo().path.toNIO.toUri.toASCIIString) ++
+        Seq(Task.env.getOrElse("COURSIER_REPOSITORIES", "ivy2Local|central"))
+      Seq("COURSIER_REPOSITORIES" -> repos.mkString("|"))
```
lihaoyi, line 313: Is there a reason to join/split by `|` to later join by `:`, rather than passing them joined by `:` to begin with?

alexarchambault: You mean, in `CoursierClient.resolveMillDaemon`? We can change the separator there too.

`|` is used because it's the separator that `COURSIER_REPOSITORIES` uses (so that it's fine with URLs).

alexarchambault: Just changed that

Issue #3549, opened 2024-09-15
Add hello-world and todo-mvc Quarkus examples


lihaoyi:
Mill's example test suite has examples for hello-world and todo-mvc projects using in Java using Spring-Boot and Micronaut built using Mill (https://github.com/com-lihaoyi/mill/tree/main/example/javalib/web). They're not super fleshed out, but they exist.

We should add similar hello-world/todo-mvc examples for the Quarkus web framework, which according to the IntelliJ dev survey is relatively popular (between Spring-Boot and Micronaut https://www.jetbrains.com/lp/devecosystem-2023/java/)

lolgab:
I'm giving a shot to this.

lefou:
The `import`-scope Maven dependencies are called BOM - Bill of Material. There are open requests for it in Mill and coursier:

* https://github.com/com-lihaoyi/mill/issues/1975
* https://github.com/coursier/coursier/issues/1390

I think @alexarchambault will work on it in the next months (https://github.com/com-lihaoyi/mill/discussions/3496).

llvee:
@lolgab Did you complete this? If not maybe I will give this a shot in 24 hours from now.

llvee:
@lolgab Would you like some help with it? Have you made a lot of progress?

omarsoufiane:
hello @lihaoyi,
I'm the founder of [Bountyhub](https://bountyhub.dev), a new platform dedicated exclusively to managing bounties on GitHub issues. Our platform is designed to provide greater transparency between bounty creators and solvers, ensuring that both sides have clear visibility throughout the process. people who use your projects can also create bounties on your repositories. 

we would be very happy if you install our github app and add this badge to your readme:
[![Bounties on BountyHub](https://img.shields.io/badge/Bounties-on%20BountyHub-yellow)]( https://bountyhub.dev/bounties?repo=com-lihaoyi/mill)
it will redirect contributors to the bounties created for this repository (try it by clicking on the badge above).

Looking forward to hearing your thoughts.

lolgab:
No, I haven't. Feel free to work on it 👍

lihaoyi:
Bumping this up from 500 to 1000USD

himanshumahajan138:
@lihaoyi Working on this (Have Read Docs and Seems it can be done)

himanshumahajan138:
> quarkus, in its maven and gradle plugins, does some code generation which seems to be necessary to create the `main` method to run applications.

[...]

PR #6820, opened 2026-02-11
Quarkus test support

vaslabs:
Part of https://github.com/com-lihaoyi/mill/issues/3549 and https://github.com/com-lihaoyi/mill/issues/6496

## Quarkus Test support feature

- QuarkusTests provides the ability to run the unit tests against the relevant modules via serializing a test app model.
- The discovery needs both the quarkus junit and the quarkus bootstrap deps in the runtime classpath to function. At this moment, I'm not completely sure why the `DiscoverJunit5Tests` ZincOp is not working but it works with running a forked jvm with the classpath of the runtime classpath (so the QuarkusTest annotation is detected and works).
- Added a `testDiscoverRuntimeOptions` for easier integration with the TestModule instead of overriding the whole discoveredTestClasses.
-  Updated the example with the test run
- Added the quarkus example in the web docs

## Other fixes

I've noticed that springboot todo started failing with a docker environment not found error. Probably some update in docker as I updated the deps and (at least locally) it seems fine now:
- Fixed javalib springboot todo example docker failure
- Fixed kotlinlib springboot todo example docker failure

## What is not supported yet

- Native builds & tests
- Multi-module Quarkus builds.

## Future PRs

For the native builds tests, I've started a separate work, I think it's close, just missing something to cause the augmented result to generate a native build too.

Also, quarkus has a lot into it, so perhaps in the future we can have a dedicated page like we do with SpringBoot

### DIFF
--- a/example/kotlinlib/web/8-todo-spring-boot/build.mill
+++ b/example/kotlinlib/web/8-todo-spring-boot/build.mill
@@ -32,10 +32,9 @@ object `package` extends KotlinModule {
 
   object integration extends HelloTests {
     def mvnDeps = super.mvnDeps() ++ Seq(
-      mvn"org.testcontainers:testcontainers:1.18.0",
-      mvn"org.testcontainers:junit-jupiter:1.18.0",
-      mvn"org.testcontainers:postgresql:1.18.0",
-      mvn"org.postgresql:postgresql:42.6.0"
+      mvn"org.testcontainers:junit-jupiter:1.21.4",
+      mvn"org.testcontainers:postgresql:1.19.8",
+      mvn"org.postgresql:postgresql:42.7.3"
     )
   }
 }


vaslabs:
I've managed to make tests work with a caveat, we need to implement a new `discoveredTestClasses` for Quarkus. the classloader that handles the discovery somehow loses the AugmentActionImpl with

```scala
// TODO implement a new discoveredTestClasses for Quarkus
    override def discoveredTestClasses: T[Seq[String]] = Task {
      val worker = jvmWorker().internalWorker()
      worker.apply(
        mill.javalib.api.internal.ZincOp.DiscoverJunit5Tests(
          runClasspath().map(_.path),
          testClasspath().map(_.path),
          classesDir()
        ),
        javaHome().map(_.path),
        javaRuntimeOptions = forkArgs()
      )
    }
```    

The above passes the quarkus bootstrapping because the fork args 
```
    override def forkArgs = Task {
      Seq(
        s"-Dquarkus-internal-test.serialized-app-model.path=${quarkusSerializedAppModel().path}"
      )
    }
```

cause the worker to spin a new process but the classloader is not behaving properly resulting in a

```
1-141] Caused by: java.lang.NoSuchMethodException: io.quarkus.runner.bootstrap.AugmentActionImpl.<init>(io.quarkus.bootstrap.app.CuratedApplication,java.util.List)
1-141]  at java.base/java.lang.Class.getConstructor0(Class.java:3763)
1-141]  at java.base/java.lang.Class.getConstructor(Class.java:2444)
1-141]  at io.quarkus.bootstrap.app.CuratedApplication.createAugmentor(CuratedApplication.java:129)
```

even though I manually tested that this constructor can be found in `DiscoverJunit5Tests`

For now I stubbed it and I'll try to implement a fresh one for Quarkus and either I'll figure out where the conflict happens or we use the fresh for quarkus modules

vaslabs:
with the latest commit we are there for the quarkus test support (at least junit example working)

I'll add scaladocs and add the example into the mill docs and if I have time and the native-tests work out I'll add those too otherwise on a future PR.

The only trouble was separating the classpaths/making the discovery classpath work, but thankfully it can accept java opts, which I introduced with a new `testDiscoverRuntimeOptions`

Please have an initial review if you have the time @lihaoyi @lefou and I'll take your feedback in mind when I finalise the PR.

libs/javalib/api/src/mill/javalib/quarkus/ApplicationModelWorker.scala
```diff
@@ -1,21 +1,26 @@
...
+import mill.api.JsonFormatters.pathReadWrite
 
 @experimental
 trait ApplicationModelWorker extends AutoCloseable {
   def quarkusBootstrapApplication(
       applicationModelFile: os.Path,
       destRunJar: os.Path,
-      jar: os.Path
+      jar: os.Path,
+      isTest: Boolean
```
vaslabs: Unused will remove

example/javalib/web/6-quarkus-getting-started/build.mill
```diff
@@ -1,12 +1,15 @@
...
+// Quarkus support is an ongoing effort. Native tests are not supported yet.
+
 package build
 
 import mill.*, javalib.*
 
 import mill.javalib.quarkus.QuarkusModule
 
-object `package` extends QuarkusModule, MavenModule {
+object `package` extends QuarkusModule, MavenModule { outer =>
```
vaslabs: outer not needed, leftover from testing

libs/javalib/quarkus-worker/src/mill/javalib/quarkus/ApplicationModelWorkerImpl.scala
```diff
@@ -125,23 +129,32 @@ class ApplicationModelWorkerImpl extends ApplicationModelWorker {
       builder
     }
 
-    val dependencies = appModel.dependencies.map(toResolvedDependencyBuilder)
+    // TODO distinct should not be needed, caller needs fixing
+    val dependencies = appModel.dependencies.distinct.map(toResolvedDependencyBuilder)
```
vaslabs: need to fix this

PR #6822, opened 2026-02-12
Add selective.reset command

joan38:
Our CI does the follow before running all tests with `selective.run`:
```
  Try(sh("git reset --hard Last-Successful-Release")) match
    case Success(_) => ev.evaluate("selective.prepare")
    case Failure(_) => os.write(workspaceRoot / "out" / "mill-selective-execution.json", "")
  sh("git reset --hard $SHA")
```

We can also see that the `mill-selective-execution.json` file is being created in the mill CI:
https://github.com/com-lihaoyi/mill/blob/fe778e0fdd0b770cb0cc89f74c9c2581112e5a41/.github/actions/pre-build-setup/action.yml#L49-L50

So I propose to have an actual task for creating an empty `mill-selective-execution.json` so that we can run `selective.run`.


lihaoyi:
could you flesh out the PR description, I have no idea what this is about

joan38:
@lihaoyi Done.

Are the tests failing because of my change?

lihaoyi:
@joan38 probably not, likely just flakiness

joan38:
@lihaoyi @lefou do we have agreement on adding this feature?

lihaoyi:
@joan38 I still don't quite understand your PR description

joan38:
Basically this PR is adding the `selective.reset` command that simply creates the `out/mill-selective-execution.json` file.

### Why the need for this command?

We cannot run `selective.run` without that json file. It's prompting to run `selective.prepare`.
But there are instances were we cannot prepare because we don't know the relative change. Therefore we would like to run everything.

The documentation suggests:
> selective.run relies on an out/mill-selective-execution.json file generated by selective.prepare in order to work, and will report an error if that file is missing. **You can also zero out that file** to explicitly tell selective.run to run all given tasks non-selectively, which is convenient if you want to conditionally disable selective execution (e.g. perhaps you want to perform selective execution on pre-merge on pull requests but not post-merge on the main branch)

The "zero out that file" can be done with `selective.reset`.


If you have a better task name, I'll take it.

lihaoyi:
Maybe something like `selective.prepare --empty` or `--none`? `selective.reset` doesn't really fit what you describe, since it's not resetting anything to a previous state

joan38:
Like this?

joan38:
@lihaoyi the `empty` flag does not play very well with the var arg of tasks.
Should I rollback to the initial version? Or you have a better idea?

lihaoyi:
Maybe a different name then? `selective.prepareEmpty` perhaps?

PR #6833, opened 2026-02-16
Android koin example

ABeesettiPaleBlue:
Added [koin/examples/androidx-samples](https://github.com/InsertKoinIO/koin/tree/main/examples/androidx-samples) to thirdparty examples (Closes #6652)

image

PR #6841, opened 2026-02-17
Android: Generate test_config.properties

### LINKED ISSUE #6620: Android: Support Robolectric unit tests
While the latest version of Mill makes [Jetcaster.wear](https://github.com/vaslabs-ltd/compose-samples-with-mill/pull/21) runnable, we've discovered that Roboelectric unit tests need additional work.

Basically, it seems that a similar flow with instrumented tests (build wise) is needed, as these tests run in a JVM instead of an emulator - that to me implies that what we have in the instrumented tests classpath, needs to be in the unit tests classpath as well. So far, we just took care of only jvm relevant artifacts/assets.

Running the jetcaster unit test gives:

```
Test com.example.jetcaster.NavigationTest.launchAndNavigate failed: java.lang.RuntimeException: Unable to resolve activity for Intent { act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] cmp=org.robolectric.default/com.example.jetcaster.MainActivity } -- see https://github.com/robolectric/robolectric/pull/4736 for details, took 0.915 sec
801]     at org.robolectric.android.internal.RoboMonitoringInstrumentation.startActivitySyncInternal(RoboMonitoringInstrumentation.java:101)
801]     at org.robolectric.android.internal.LocalActivityInvoker.startActivity(LocalActivityInvoker.java:36)
801]     at org.robolectric.android.internal.LocalActivityInvoker.startActivity(LocalActivityInvoker.java:41)
801]     at androidx.test.core.app.ActivityScenario.launchInternal(ActivityScenario.java:367)
801]     at androidx.test.core.app.ActivityScenario.launch(ActivityScenario.java:203)
801]     at androidx.test.ext.junit.rules.ActivityScenarioRule.lambda$new$0(ActivityScenarioRule.java:77)
801]     at androidx.test.ext.junit.rules.ActivityScenarioRule.before(ActivityScenarioRule.java:112)
```

## What does that mean?

It seems to me the namespace of the app Roboelectric is looking is different (`org.robolectric.default`) from what we give in the manifest (user package) and so the activity cannot be found.

There's relevant information in the PR logged by the error, but we need to investigate which information is relevant to mill and probably some reverse engineering will be needed.

My suspicion is that the plugin:

`alias libs.plugins.roborazzi` (io.github.takahirom.roborazzi) takes care of the manipulation of the manifest or other resources.

In addition, the gradle option:

```
    testOptions {
        unitTests {
            includeAndroidResources true
        }
    }
```

feels relevant  ~~but we already include the resources in the unit tests (but maybe not the flat files?)~~

EDIT latter introduced in https://github.com/com-lihaoyi/mill/pull/6638

souvlakias:
Closes #6620 

This adds to the support for the `isIncludeAndroidResources` that was introduced in https://github.com/com-lihaoyi/mill/pull/6638, by generating the `com/android/tools/test_config.properties` file as described [here (agp docs)](https://developer.android.com/reference/tools/gradle-api/8.3/null/com/android/build/api/dsl/UnitTestOptions#getIsIncludeAndroidResources()).

This also makes roboelectric unit tests work (Ref: [jetcaster.wear](https://github.com/vaslabs-ltd/compose-samples-with-mill/blob/main/Jetcaster/package.mill#L442) and the [koin example](https://github.com/com-lihaoyi/mill/pull/6833))

### DIFF
--- a/libs/androidlib/src/mill/androidlib/AndroidModule.scala
+++ b/libs/androidlib/src/mill/androidlib/AndroidModule.scala
@@ -945,19 +945,26 @@ trait AndroidModule extends JavaModule { outer =>
     }
 
     /**
-     * Generates a Java properties file required when [[androidIncludeAndroidResources]] is true,
-     * containing necessary information for the Android resource processing in unit tests.
-     *
-     * Properties and expected name on the classpath are defined at
-     * [[https://developer.android.com/reference/tools/gradle-api/8.3/null/com/android/build/api/dsl/UnitTestOptions#getIsIncludeAndroidResources()]]
+     * The properties of the generated test configuration file for Android unit tests.
      */
-    def androidGeneratedTestConfigSources: T[Seq[PathRef]] = Task {
-      val properties: Map[String, String] = Map(
+    def androidTestConfigProperties: T[Map[String, String]] = Task {
+      Map(
         "android_custom_package" -> outer.androidNamespace,
         "android_merged_manifest" -> outer.androidMergedManifest().path.toString,
         "android_resource_apk" -> (outer.androidLinkedResources().path / "apk" / "res.apk").toString,
         "android_merged_assets" -> outer.androidTransitiveMergedAssets().path.toString
       )
+    }
+
+    /**
+     * Generates a Java properties file required when [[androidIncludeAndroidResources]] is true,
+     * containing necessary information for the Android resource processing in unit tests.
+     *
+     * Expected name on the classpath and properties are defined at
+     * [[https://developer.android.com/reference/tools/gradle-api/8.3/null/com/android/build/api/dsl/UnitTestOptions#getIsIncludeAndroidResources()]]
+     */
+    def androidGeneratedTestConfigSources: T[Seq[PathRef]] = Task {
+      val properties = androidTestConfigProperties()
 
       val content = properties.map { case (key, value) =>
         s"$key=$value"


libs/androidlib/src/mill/androidlib/AndroidModule.scala
```diff
@@ -930,8 +944,36 @@ trait AndroidModule extends JavaModule { outer =>
...
 
+    /**
+     * Generates a Java properties file required when [[androidIncludeAndroidResources]] is true,
+     * containing necessary information for the Android resource processing in unit tests.
+     *
+     * Properties and expected name on the classpath are defined at
+     * [[https://developer.android.com/reference/tools/gradle-api/8.3/null/com/android/build/api/dsl/UnitTestOptions#getIsIncludeAndroidResources()]]
+     */
+    def androidGeneratedTestConfigSources: T[Seq[PathRef]] = Task {
+      val properties: Map[String, String] = Map(
```
vaslabs: we can make this into a separate task so we have more flexibility debugging/fixing issues in the future and allow for easier workarounds

vaslabs:
does it also fix the tests in https://github.com/com-lihaoyi/mill/pull/6833 ? (maybe need to add include android resources as true there too)

ah yeah it does, you mentioned it, sorry

I thought it might have issues with [mocked default values](https://developer.android.com/reference/tools/gradle-api/8.3/null/com/android/build/api/dsl/UnitTestOptions#getIsReturnDefaultValues()) but maybe it's not hitting it

vaslabs:
amazing work, nice one @souvlakias

Issue #6850, opened 2026-02-18
Unhandled exception with relative path


PapyChacal:
Hello,

We are facing a boggling issue when bumping from mill 1.0.0-RC1 to 1.1.X on a Scala3 project. To start with a tentatively short description, we see:

```
1867]   An unhandled exception was thrown in the compiler.
1867]   Please file a crash report here:
1867]   https://github.com/scala/scala3/issues/new/choose
1867]   For non-enriched exceptions, compile with -Xno-enrich-error-messages.
1867] 
1867] 
1867]      while compiling: <no file>
1867]         during phase: parser
1867]                 mode: Mode()
1867]      library version: version 3.8.1
1867]     compiler version: version 3.8.1
1867]             settings: -Wunused List(imports, imports) -classpath /home/runner/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-lang/scala-library/3.8.1/scala-library-3.8.1.jar:/home/runner/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/fastparse_3/3.1.1/fastparse_3-3.1.1.jar:/home/runner/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/sourcecode_3/0.4.0/sourcecode_3-0.4.0.jar:/home/runner/.cache/coursier/v1/https/repo1.maven.org/maven2/com/lihaoyi/geny_3/1.1.0/geny_3-1.1.0.jar:/home/runner/work/scair/scair/utils/compile-resources:/home/runner/work/scair/scair/out/utils/scoverage/compile.dest/classes:/home/runner/work/scair/scair/core/compile-resources:/home/runner/work/scair/scair/out/core/scoverage/compile.dest/classes -color never -coverage-out /home/runner/work/scair/scair/out/core/scoverage/data.dest -d /home/runner/work/scair/scair/out/core/scoverage/compile.dest/classes -explain true -sourceroot /home/runner/work/scair/scair
1867] 
Error: 1867] [error] ## Exception when compiling 43 sources to /home/runner/work/scair/scair/out/core/scoverage/compile.dest/classes
1867] java.lang.IllegalArgumentException: requirement failed: core/src/clair/Macros.scala is not an absolute path
1867] scala.Predef$.require(Predef.scala:394)
1867] os.Path.<init>(Path.scala:574)
1867] os.Path$.apply(Path.scala:497)
[...]

PR #6851, opened 2026-02-18
Handle relative path in dottyStyleMessage.

PapyChacal:
Proposed as potential solution to my own issue:
- #6850
In a nutshell, this line is crashing whole builds for simple warnings, in some very specific cases. I figured that maybe the absolute path expectation was just arbitrary?

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -185,8 +185,8 @@ def listCross(using ctx: mill.api.ModuleCtx) = BuildCtx.watchValue {
 
 val dummyDeps: Seq[Dep] = Seq(
   Deps.DocDeps.millScip,
-  Deps.semanticDbJava,
-  Deps.semanticDBscala,
+  Deps.semanticDbJava_runtime,
+  Deps.semanticDBscala_runtime,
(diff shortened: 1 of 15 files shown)


libs/javalib/worker/src/mill/javalib/zinc/TransformingReporter.scala
```diff
@@ -82,7 +82,12 @@ private object TransformingReporter {
     InterfaceUtil.jo2o(pos.sourcePath()) match {
       case None => message
       case Some(path) =>
-        val absPath = os.Path(path)
+        val absPath = os.FilePath(path) match {
+          case abs: os.Path => abs
+          // Assume relative paths are relative to the current workspaceRoot
+          case rel: os.RelPath => workspaceRoot / rel
+          case sub: os.SubPath => workspaceRoot / sub
+        }
```
alexarchambault, line 90: Same as this?
```suggestion
        val absPath = os.Path(path, workspaceRoot)
```

PR #6852, opened 2026-02-18
Move visualize to a separate worker

re-thc:
Was looking into cold start performance and visualize downloads a lot of libs before anything happens (and likely isn't used in a lot of paths).

Visualize needs its own dedicated thread and queues etc also.

This moves it to a worker.

### DIFF
--- a/runner/idea/src/mill/idea/GenIdeaImpl.scala
+++ b/runner/idea/src/mill/idea/GenIdeaImpl.scala
@@ -436,14 +436,29 @@ class GenIdeaImpl(
       os.sub / s"${name.replaceAll("""[-.:]""", "_")}.${ext}"
     }
 
+    def shouldFallbackSourcesToClasses(path: os.Path): Boolean = {
+      val jarName = path.last
+      jarName.startsWith("scala-library-") ||
+      jarName.startsWith("scala3-library_3-") ||
+      jarName.startsWith("scala3-library_sjs1_3-") ||
+      jarName.startsWith("scala-reflect-") ||
+      jarName.startsWith("scala-xml_") ||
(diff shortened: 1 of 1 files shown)


lihaoyi:
Looks great, just need to regenerate the golden test files and I'll merge it

re-thc:
@lihaoyi I've found this confusing with what happens with scala3_library and so I'm not sure if I should have those changed or not. I've generated and ran it locally lots of times and tried to fix it.

I suspect the issue is coursier/cache-action and the selective tests. The what the cache returns across CI runs and how that impacts what GenIdea gets picked up is the issue.

What would be the fix?

lihaoyi:
@re-thc I have no idea what you are asking

libs/util/src/mill/util/VisualizeModule.scala
```diff
@@ -55,53 +142,17 @@ object VisualizeModule extends ExternalModule {
...
-        }
+    evaluator.resolveTasks(tasks, SelectMode.Multi).flatMap { rs =>
+      planTasks match {
+        case Some(allRs) => callVisualizeModule(rs, allRs)
+        case None => callVisualizeModule(rs, rs)
+      }
     }
   }
 
-  // TODO: Remove bin-compat shim in Mill 2.x
```
lefou, line 67: Can you preserve this TODO, otherwise we risk to overlook to remove this deprecation again.

re-thc: Updated

re-thc:
@lihaoyi see if this fix is acceptable for CI

lihaoyi:
Thanks @re-thc !

PR #6853, opened 2026-02-19
Move Bsp to a separate worker

re-thc:
Was looking into cold start performance and Bsp downloads a lot of libs before anything happens (and likely isn't used in a lot of paths).

This moves it to a worker.


runner/daemon/src/mill/daemon/MillMain0.scala
```diff
@@ -528,7 +522,7 @@ object MillMain0 {
       outLock: Lock,
       bspLogger: Logger,
       daemonDir: os.Path
-  ): (BspServerHandle, BuildClient) = {
+  ): (BspServerHandle, AnyRef) = {
```
lefou: Can we introduce some type here? `AnyRef` screams for trouble and doesn't help at all when reading the code.

runner/daemon/src/mill/daemon/BspIdeaWorkerSupport.scala
```diff
@@ -0,0 +1,188 @@
...
+import coursier.core.Repository
+import coursier.{Dependency, Module, ModuleName, Organization, VersionConstraint}
+import mill.api.daemon.internal.{CompileProblemReporter, EvaluatorApi}
+import mill.api.daemon.internal.bsp.BspServerHandle
+import mill.api.{Logger, MillException, Result, SystemStreams}
+import mill.javalib.api.JvmWorkerUtil
+import mill.client.lock.Lock
+import mill.util.{BuildInfo, Jvm}
+
+private object BspIdeaWorkerSupport {
```
lefou: Maybe replace the `BspIdea` with `Ide`.

lihaoyi:
Can you report what the difference in the number of libs downloaded is with and without this PR, and their total size in mb/gb?

re-thc:
@lihaoyi 

On main: 73.79MB
On this PR: 66.02MB

So saved 7.4MB (that's about 10%)

Big changes: `jgit`, `guava`, `gson`, `bsp4j`, `lsp4j`

It's not just download size and download time. Coursier also has to hash and analyze them.

lefou:
> @lihaoyi
> 
> On main: 73.79MB On this PR: 66.02MB
> 
> So saved 7.4MB (that's about 10%)
> 
> Big changes: `jgit`, `guava`, `gson`, `bsp4j`, `lsp4j`
> 
> It's not just download size and download time. Coursier also has to hash and analyze them.

If not for the size, I find i good thing to not have these deps on the default classpath.

lihaoyi:
@lefou this stuff is on the daemon classpath, not on the `build.mill` classpath, so apart from the download size this change wouldn't affect users at all

re-thc:
Context is I'm trying to migrate from Gradle. On cold start it's easily 8x slower on a build. Pretty significant. This triggers on CI or other use cases. Did some profiling and trying to reduce it. So the download size is a thing.

A separate topic, but the caching and remote caching has come up in other topics and so far it doesn't really resolve it so I'm trying to tackle cold starts 1st.

runner/daemon/src/mill/daemon/BspIdeaWorkerSupport.scala
re-thc: Done

runner/daemon/src/mill/daemon/MillMain0.scala
re-thc: Updated

lihaoyi:
Thanks @re-thc !

PR #6854, opened 2026-02-19
Kapt support

re-thc:
Kapt is a Kotlin compiler plugin and/or Kotlin plugin for "old" Java and pre-ksp use cases. It's slower and in "maintenance" mode but still available.

It's needed to port existing Java code (still relying on annotation processors) and many things still don't have Ksp support.

Not to mention Ksp 2 still has some hard bugs that aren't fixed yet (so it's not strictly worse).

### DIFF
--- a/libs/kotlinlib/src/mill/kotlinlib/kapt/KaptModule.scala
+++ b/libs/kotlinlib/src/mill/kotlinlib/kapt/KaptModule.scala
@@ -30,7 +30,11 @@ trait KaptModule extends KotlinModule { outer =>
    * KAPT compiler plugin dependency.
    */
   def kaptPluginMvnDeps: T[Seq[Dep]] = Task {
-    Seq(mvn"org.jetbrains.kotlin:kotlin-annotation-processing-embeddable:${kaptVersion()}")
+    if (kotlinUseEmbeddableCompiler()) {
+      Seq(mvn"org.jetbrains.kotlin:kotlin-annotation-processing-embeddable:${kaptVersion()}")
+    } else {
+      Seq(mvn"org.jetbrains.kotlin:kotlin-annotation-processing:${kaptVersion()}")
+    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


re-thc:
The CI failure doesn't look related to this PR? Please confirm.

lihaoyi:
@vaslabs can you help review this

libs/kotlinlib/src/mill/kotlinlib/kapt/KaptModule.scala
```diff
@@ -0,0 +1,198 @@
...
+   * Kotlin compiler plugin version used by KAPT.
+   * Defaults to the same version as [[kotlinVersion]].
+   */
+  def kaptVersion: T[String] = kotlinVersion()
+
+  /**
+   * KAPT compiler plugin dependency.
+   */
+  def kaptPluginMvnDeps: T[Seq[Dep]] = Task {
+    Seq(mvn"org.jetbrains.kotlin:kotlin-annotation-processing-embeddable:${kaptVersion()}")
```
vaslabs: is there a non embeddable dependency to choose if the user overrides?
```
  override def kotlinUseEmbeddableCompiler: Task[Boolean] = Task { true }
```

libs/kotlinlib/src/mill/kotlinlib/kapt/KaptModule.scala
```diff
@@ -0,0 +1,198 @@
...
+          "-classpath",
+          compileCp.mkString(File.pathSeparator)
+        ),
+        when(kotlinExplicitApi())(
+          "-Xexplicit-api=strict"
+        ),
+        allKotlincOptions(),
+        Seq(
+          "-P",
+          s"plugin:org.jetbrains.kotlin.kapt3:aptMode=$aptMode",
```
vaslabs: could these be in `apClasspathOpts` ?

vaslabs:
lgtm on first glance, put comments on some minor nits

@souvlakias if you can also have a look at this

libs/kotlinlib/src/mill/kotlinlib/kapt/KaptModule.scala
re-thc: Updated

libs/kotlinlib/src/mill/kotlinlib/kapt/KaptModule.scala
re-thc: Updated

PR #6855, opened 2026-02-19
Android: Create Scheduled CI Workflow

### LINKED ISSUE #6845: Setup a CI workflow that runs on schedule for androidtodo
> this androidtodo has broken a few times without us noticing, let's take a note to at least enable the job only on our fork @souvlakias

You could prepare a CI workflow that runs on a schedule, but not with each merge/PR.

_Originally posted by @lefou in https://github.com/com-lihaoyi/mill/issues/6638#issuecomment-3776447004_
            
            
cc: @souvlakias

souvlakias:
### Adds 
- A `Scheduled Android Tests` workflow that will run periodically (every week?), in order to avoid testing certain examples on each merge/PR. 
For now this includes `androidtodo` and `android-endless-tunnel` (native example)

### Tweaks
- Makes `selective.prepare` not run when `event_name` is `scheduled` or `workflow_dispatch`
- Exclude a flaky instrumented test class of `androidtodo` from running on CI. Will look into it later on

Closes #6845

### DIFF
--- a/.github/actions/pre-build-setup/action.yml
+++ b/.github/actions/pre-build-setup/action.yml
@@ -53,7 +53,7 @@ runs:
       shell: bash
 
     - run: ./mill -k selective.prepare ${{ inputs.prepareargs }}
-      if: ${{ (github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'run-all-tests')) || github.repository != 'com-lihaoyi/mill' || github.event_name == 'schedule' }}
+      if: ${{ (github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'run-all-tests')) || github.repository != 'com-lihaoyi/mill' }}
       shell: ${{ inputs.shell }}
 
     - uses: actions/upload-artifact@v4.6.0
--- a/.github/actions/pre-build-setup/action.yml
+++ b/.github/actions/pre-build-setup/action.yml
@@ -53,7 +53,7 @@ runs:
       shell: bash
 
     - run: ./mill -k selective.prepare ${{ inputs.prepareargs }}
-      if: ${{ (github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'run-all-tests')) || github.repository != 'com-lihaoyi/mill' }}
+      if: ${{ (github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'run-all-tests')) || github.repository != 'com-lihaoyi/mill' || github.event_name == 'schedule' }}
       shell: ${{ inputs.shell }}
 
     - uses: actions/upload-artifact@v4.6.0
--- a/.github/actions/pre-build-setup/action.yml
+++ b/.github/actions/pre-build-setup/action.yml
@@ -53,7 +53,7 @@ runs:
       shell: bash
 
     - run: ./mill -k selective.prepare ${{ inputs.prepareargs }}
-      if: ${{ (github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'run-all-tests')) || github.repository != 'com-lihaoyi/mill' }}
+      if: ${{ ((github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'run-all-tests')) || github.repository != 'com-lihaoyi/mill') && github.event_name != 'schedule' && github.event_name != 'workflow_dispatch' }}
       shell: ${{ inputs.shell }}
 
     - uses: actions/upload-artifact@v4.6.0
--- a/.github/workflows/scheduled-android-tests.yml
+++ b/.github/workflows/scheduled-android-tests.yml
@@ -14,8 +14,15 @@ env:
   FORCE_COLOR: '1'
 
 jobs:
+  build-linux:
+    uses: ./.github/workflows/pre-build.yml
+    with:
+      java-version: 21
+      os: ubuntu-latest
+      shell: bash
+
   android:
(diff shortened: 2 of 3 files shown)


.github/workflows/scheduled-android-tests.yml
```diff
@@ -0,0 +1,39 @@
...
+
+  android:
+    needs: build-linux
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: androidtodo
+            millargs: example.thirdparty.androidtodo.packaged.daemon
+            setup-android: true
```
vaslabs, line 33: We can also add android native, I think it's not in the run tests pipeline

vaslabs:
Is there a test run of this somewhere to make sure it works?

vaslabs:
Also let's remove the emulator run, and keep the tests + also focus the docs on the use case of the example , Koin for dependency injection and robolectric tests

souvlakias:
> Is there a test run of this somewhere to make sure it works?

No, I'm not sure how to test if the scheduled workflow works tbh. Maybe merge it into a local fork's main and see

vaslabs:
> > Is there a test run of this somewhere to make sure it works?
> 
> No, I'm not sure how to test if the scheduled workflow works tbh. Maybe merge it into a local fork's main and see

Yes let's do that

souvlakias:
> > No, I'm not sure how to test if the scheduled workflow works tbh. Maybe merge it into a local fork's main and see
> 
> Yes let's do that

Merged into our forks main and it triggered succesfully https://github.com/vaslabs-ltd/mill/actions/workflows/scheduled-android-tests.yml
Tests passed ~~but there's seems to be a bug in our CI for a few days, where the android apps' tests do not actually run but still pass~~, but didn't not run probably because due to the selective resolve
image
will have to investigate on this, but its a different issue

.github/workflows/scheduled-android-tests.yml
```diff
@@ -0,0 +1,42 @@
+name: Scheduled Android Tests
+
+on:
+  schedule:
+    # 03:00 every Sunday
+    - cron: '0 3 * * 0'
```
souvlakias: also, this was chosen at random. If another time or frequency is more reasonable let me know so I update accordingly

vaslabs:
> > > No, I'm not sure how to test if the scheduled workflow works tbh. Maybe merge it into a local fork's main and see
> > 
> > 
> > Yes let's do that
> 
> Merged into our forks main and it triggered succesfully https://github.com/vaslabs-ltd/mill/actions/workflows/scheduled-android-tests.yml Tests passed but there's seems to be a bug in our CI for a few days, where the android apps' tests do not actually run but still pass image will have to investigate on this, but its a different issue

It didn't run , probably because of the selective? Not sure

vaslabs:
updated the schedule job, to not use selective, seems [to be working now](https://github.com/vaslabs-ltd/mill/actions/runs/22276968468/job/64440650563). I think it still needs a build job to not build mill twice but gonna leave it to you @souvlakias  

EDIT: I think it's better to split it docs-koin/scheduled, this dual scope can delay things

.github/actions/pre-build-setup/action.yml
```diff
@@ -53,7 +53,7 @@ runs:
       shell: bash
 
     - run: ./mill -k selective.prepare ${{ inputs.prepareargs }}
-      if: ${{ (github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'run-all-tests')) || github.repository != 'com-lihaoyi/mill' }}
+      if: ${{ (github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'run-all-tests')) || github.repository != 'com-lihaoyi/mill' || github.event_name == 'schedule' }}
```
vaslabs: I added this for experimentation, can be undone

souvlakias:
> EDIT: I think it's better to split it docs-koin/scheduled, this dual scope can delay things

Split #6881

souvlakias:
Current version (use of existing workflow scripts and avoid selective prepare) seems to work:
https://github.com/vaslabs-ltd/mill/actions/runs/22404694012

vaslabs:
> Current version (use of existing workflow scripts and avoid selective prepare) seems to work: https://github.com/vaslabs-ltd/mill/actions/runs/22404694012

cool, now we need to figure out why androidtodo consistently times out, could be the port

souvlakias:
> cool, now we need to figure out why androidtodo consistently times out, could be the port

It seems that a test class of instrumented tests is problematic in CI and causes to timeout. I didn't manage to reproduce locally.
If we [exclude](https://github.com/com-lihaoyi/mill/blob/f0f78f517c280607621a113d48cb73632db89633/example/thirdparty/androidtodo/build.mill#L281) that class it works https://github.com/vaslabs-ltd/mill/actions/runs/22861753577

vaslabs:
This is ready, right?

souvlakias:
> This is ready, right?

The workflow should be ready, but `androidtodo`'s run will fail 😅
I guess, we can exclude the flaky class and try to fix it later on

vaslabs:
> > This is ready, right?
> 
> The workflow should be ready, but `androidtodo`'s run will fail 😅 I guess, we can exclude the flaky class and try to fix it later on

let's exclude it, we want to know periodically that the android todo test, androidTest, and main app run is wired properly

vaslabs (approved):

vaslabs:
@lihaoyi if you are happy with this scheduled task on the repo, it's ready to merge, otherwise we can close it and setup an external repo like we do with https://github.com/vaslabs-ltd/compose-samples-with-mill

Issue #6634, opened 2026-01-20
Android: Attach sources automatically


vaslabs:
Right now, the android aar libs that we transform, have sources but are not carried to the intelliJ settings.

For example:

<img width="1819" height="1025" alt="Image" src="https://github.com/user-attachments/assets/f69f714e-7dea-4250-aa73-566f4435ee77" />

After manual attachement

<img width="1819" height="1025" alt="Image" src="https://github.com/user-attachments/assets/3c8d27a4-5da1-46ad-955f-028c0953643c" />

<img width="1819" height="1025" alt="Image" src="https://github.com/user-attachments/assets/f76c4142-2ee6-4470-93bf-f7fc9ecf1a3d" />

souvlakias:
Bsp and GenIdea do discover the sources jars through
[bspBuildTargetDependencySources](https://github.com/com-lihaoyi/mill/blob/main/libs/javalib/src/mill/javalib/bsp/BspJavaModule.scala#L50) 

<img width="735" height="279" alt="Image" src="https://github.com/user-attachments/assets/53b1d9f5-b60b-4bd2-b1c4-7b824a76433e" />

and an unused [externalSources](https://github.com/com-lihaoyi/mill/blob/main/libs/javalib/src/mill/javalib/idea/GenIdeaModule.scala#L72)

<img width="716" height="307" alt="Image" src="https://github.com/user-attachments/assets/317a4fb7-d56d-46ad-8cef-9480a52f147c" />

Im not sure yet how the linking with the classes jars should be done

vaslabs:
> Bsp and GenIdea do discover the sources jars through [bspBuildTargetDependencySources](https://github.com/com-lihaoyi/mill/blob/main/libs/javalib/src/mill/javalib/bsp/BspJavaModule.scala#L50)
> 
[...]

PR #6856, opened 2026-02-19
Android: Show class sources in IDE

souvlakias:
Addresses #6634 

This is a **POC** on linking the classes jar with their respective sources.jar using bsp
### The problem
Since android deps are aars, we extract them into a Task.dest and pass the found `.jar`s into the classpath.
Bsp sees the `sources.jar`s that exist in `~/.cache/coursier` and doesnt pair the two.

### How it works
- When unpacking an aar that we want to include in the classpath, also copy it's sources.jar in the extract directory
- Create a `BspAndroidModule` that overrides `bspBuildTargetDependencySources` to show to the sources.jars that exist next to the jars that we pass in the classpath

### Tweaks
- Changed `resolvedMvnDeps0` from `private` to `private[mill]` so `AndroidModule` has access to it, in order to get the Sources Jars
- `androidResolvedMvnDeps` (which is what we pass to compileclasspath), now needs to use a helper task that returns `Seq[UnpackedDep]]`, in order for the bsp module to find `_.sourcesJar` from that helper task (that exist in the same Task.dest with the ones in classpath)

### Notes
This may not be the best approach. Another solution could be to use GenIdea, and create xmls that link the two jars for each dep.

https://github.com/user-attachments/assets/9382dcee-52fd-457d-8cc3-7a115de6e7ed

### DIFF
--- a/libs/javalib/src/mill/javalib/JavaModule.scala
+++ b/libs/javalib/src/mill/javalib/JavaModule.scala
@@ -1089,7 +1089,7 @@ trait JavaModule
     compileResources() ++ unmanaged
   }
 
-  private def resolvedMvnDeps0(sources: Boolean) = Task.Anon {
+  private[mill] def resolvedMvnDeps0(sources: Boolean) = Task.Anon {
     millResolver().classpath(
       Seq(
         BoundDep(
@@ -1127,17 +1127,13 @@ trait JavaModule
     resolvedMvnDeps0(sources = false)()
   }
 
-  def resolvedMvnSourceDeps: T[Seq[PathRef]] = Task {
-    resolvedMvnDeps0(sources = true)()
-  }
-
   /**
    * Resolved dependency sources, unpacked into a single directory. Useful to quickly
    * look up the sources of the dependencies on your classpath so you can find the
    * exact source code you are compiling and running against.
    */
   def resolvedMvnSources: T[PathRef] = Task {
-    for (jar <- resolvedMvnSourceDeps()) {
+    for (jar <- resolvedMvnDeps0(sources = true)()) {
       val jarName = jar.path.last.stripSuffix(".jar")
       os.unzip(jar.path, Task.dest / jarName)
     }
--- a/libs/androidlib/src/mill/androidlib/AndroidModule.scala
+++ b/libs/androidlib/src/mill/androidlib/AndroidModule.scala
@@ -411,7 +411,7 @@ trait AndroidModule extends JavaModule { outer =>
     extractAarFiles(
       androidAarMvnDeps().map(_.path),
       transformDest,
-      resolvedMvnSourceDeps().map(_.path)
+      resolvedMvnDeps0(sources = true)().map(_.path)
     )
   }
 


libs/androidlib/src/mill/androidlib/bsp/BspAndroidModule.scala
```diff
@@ -0,0 +1,40 @@
...
+import java.nio.file.Path
+
+import mill.Task
+import mill.api.daemon.internal.internal
+import mill.api.ModuleCtx
+import mill.androidlib.AndroidModule
+import mill.javalib.bsp.BspModule
+import mill.api.JsonFormatters.given
+
+trait BspAndroidModule extends mill.javalib.bsp.BspJavaModule {
```
vaslabs, line 13: `@experimental`

vaslabs:
this looks reasonable to me but I'm not familiar with that area, we'll need some help reviewing to see if it's in the right direction @lihaoyi @lefou  

Also if you could attach a small video recording with the sources working @souvlakias , would be great

souvlakias:
If it's in the right direction, it might be a good time to organize all of the tasks that extract aars into using a helper task that extracts all of them instead of every task extracting to its own Task.dest. 

(I mean that I just messed with `androidResolvedMvnDeps` for the POC for now, but it could be a good choice to update the rest of the relevant tasks like `androidUnpackArchives` and their Run variants)

If it would be out of scope, this could be done in a different PR

vaslabs:
> If it's in the right direction, it might be a good time to organize all of the tasks that extract aars into using a helper task that extracts all of them instead of every task extracting to its own Task.dest.
> 
> (I mean that I just messed with `androidResolvedMvnDeps` for the POC for now, but it could be a good choice to update the rest of the relevant tasks like `androidUnpackArchives` and their Run variants)
> 
> If it would be out of scope, this could be done in a different PR

I think this is an important fix, we can follow up with optimisations on follow up prs

libs/javalib/src/mill/javalib/JavaModule.scala
```diff
@@ -1127,13 +1127,17 @@ trait JavaModule
     resolvedMvnDeps0(sources = false)()
   }
 
+  def resolvedMvnJarSources: T[Seq[PathRef]] = Task {
+    resolvedMvnDeps0(sources = true)()
+  }
```
lefou: Why is this required to show source JARs in the IDE?

souvlakias: `ressolvedMnvDeps0` is private in JavaModule, so this serves as a way for the AndroidModule to easily get the `sources.jar`s

lefou: Maybe we can just change `resolvedMvnDeps0` to `private[mill]` and avoid another public task?

If not, I find the name unlucky, since `Sources` suffix is typically not associated archives but plain files. Instead, we could name it `resolvedMvnSourceJars` (If it is guaranteed to always be jar files) or `resolvedMvnSourceDeps`.

lefou: I saw you renamed it. So why not the first option?

souvlakias: If you mean the private[mill] vs having a public task, both felt acceptable so I just went with the clearer api.
About the name, im not sure about the "only jars" guarantee so the latter suggestion (`resolvedMvnSourceDeps`) seemed ok too.

If you prefer to avoid the public task, yes I can change to private[mill]

vaslabs: from my point of view, let's not increase the surface of the API if not necessary @souvlakias (e.g. especially since JavaModule is not experimental), it is convenient to run it directly and check the output, but since this looks straightforward functionality to me, `private[mill]` is preferable for me

souvlakias: changed to `private[mill]`

vaslabs:
Any objections merging this? @lihaoyi @lefou

PR #6857, opened 2026-02-20
Use KotlinToolchains (fixes #6793)

### LINKED ISSUE #6793: Address feedback to our Kotlin BTAPI impl
Feedback from @ALikhachev submitted via https://youtrack.jetbrains.com/issue/KT-62087/How-to-enable-incremental-compilation-via-CLI-or-generic-build-tool#focus=Comments-27-13289556.0-0

Key points:

* Classpath contains to many and likely inconsistent jars
* BT API should always be used with the embeddable compiler, so we need to restrict it's applicability further (when `kotlinUseEmbeddableCompiler = true`)
* Should we change `KotlinModule.kotlinUseEmbeddableCompiler` default to `true`?
* Review the classpath snapshot logic, since we likely using it wrong an need to compute it ourselves

Full message:

> Thanks, I am interested in taking a look at your implementation and will do this soon. Please be informed the API with the entry point CompilationService was more a prototype which gave us some insights and based on the experience gained, we have deprecated it since Kotlin 2.3.0. The new design is available with the entry point at org.jetbrains.kotlin.buildtools.api.KotlinToolchains. It's still marked as experimental, but we are putting effort into stabilizing it. Mostly it's safe to use already and major changes to the API shape are not expected. Worth mentioning the new API entry point keeps compatibility with the previous versions also. For this, in addition to the org.jetbrains.kotlin:kotlin-build-tools-impl, the artifact org.jetbrains.kotlin:kotlin-build-tools-compat of the version matching the API version should be added at runtime. (https://github.com/JetBrains/kotlin/blob/ce9089dc11b0b8bd9274750c070cab70ffba0096/compiler/build-tools/kotlin-build-tools-compat/README.md)
>
> The exceptions you are getting java.lang.NoSuchMethodError: 'org.jetbrains.kotlin.com.intellij.openapi.vfs.VirtualFileSystem org.jetbrains.kotlin.cli.jvm.compiler.KotlinCoreApplicationEnvironment.getJarFileSystem()' might be related to having mixed compiler artifacts in the classpath. It's recommended to have an isolated classloader for the compiler.
By mixed compiler artifacts, I mean we have a confusing set of published artifacts, like org.jetbrains.kotlin:kotlin-compiler and org.jetbrains.kotlin:kotlin-compiler-embeddable (that patterns repeats for many artifacts). We would like to address this confusion in some future ([KT-33805](https://youtrack.jetbrains.com/issue/KT-33805/Distribute-single-version-of-the-compiler-rather-than-regular-and-embeddable-variants) Distribute single version of the compiler (rather than regular and -embeddable variants)). The Build Tools API relies on the -embeddable artifacts.
>
> Here's an example of the recommended approach in the Kotlin Maven plugin with isolated classloader: https://github.com/JetBrains/kotlin/blob/fecfa653742616daa054309116e0e99a4d20f5a0/libraries/tools/kotlin-maven-plugin/src/main/java/org/jetbrains/kotlin/maven/K2JVMCompileMojo.java#L252
> 
> Also, I noticed
> 
> // Create approach parameters for classpath snapshot-based incremental compilation.
      // Pass empty list for newClasspathSnapshotFiles to let the API compute snapshots from classpath JARs.
      // See: https://github.com/JetBrains/kotlin/blob/v2.1.20/libraries/tools/kotlin-maven-plugin/src/main/java/org/jetbrains/kotlin/maven/K2JVMCompileMojo.java#L356
      val approachParams = new ClasspathSnapshotBasedIncrementalCompilationApproachParameters(
        java.util.Collections.emptyList(),
        (incrementalCachePath / "shrunk-classpath-snapshot.bin").toIO
      )
> 
> It's actually not working as you expect. It's a not yet implemented thing in the Kotlin Maven plugin. We do not compute the snapshots automatically, we expect you to compute them explicitly and pass to the API. The reason behind this is that many build systems have concepts of caching, and dependencies are quite often reused between the project modules, so instead of reinventing caches here, we allow explicit control on the snapshotting so that you could use your build system caching mechanisms.
> 
> > One issue we had was the finishProjectCompilation. We didn't find the proper event to invoke it in a multi-module scenario. We got occasional exceptions like reported here and went with simply never calling it.
> 
> It's not dangerous to not call it, but recommended to avoid resource leaks. Though, calling it after each compilation would also be not very efficient, as it will drop compiler caches. We have the same problem in the Kotlin Maven plugin at the moment, so there we drop the caches after each compilation, but I hope we will invent a solution there later.

re-thc:
Switch Kotlin BT API JVM compile path to KotlinToolchains.

This addresses all comments in #6793 and the related youtracks.

This ensure incremental compiling works and to only use BT API from 2.3 since the old 1 is "experimental" / "deprecated". Otherwise fall back to CLI.

Checked for KSP / KAPT compatibility.

### DIFF
--- a/libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
+++ b/libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
@@ -130,7 +130,6 @@ trait KotlinModule extends JavaModule with KotlinModuleApi { outer =>
   def kotlinCompilerMvnDeps: T[Seq[Dep]] = Task {
     val useEmbeddable = kotlinUseEmbeddableCompiler()
     val kv = kotlinVersion()
-    val btApiVersion = kotlinBuildToolsApiVersion()
     val isOldKotlin = Seq("1.0.", "1.1.", "1.2.0", "1.2.1", "1.2.2", "1.2.3", "1.2.4")
       .exists(prefix => kv.startsWith(prefix))
 
@@ -139,8 +138,7 @@ trait KotlinModule extends JavaModule with KotlinModuleApi { outer =>
       else mvn"org.jetbrains.kotlin:kotlin-compiler:${kv}"
 
     val btApiDeps = when(kotlincUseBtApi() && useEmbeddable)(
-      mvn"org.jetbrains.kotlin:kotlin-build-tools-api:$btApiVersion",
-      mvn"org.jetbrains.kotlin:kotlin-build-tools-compat:$btApiVersion",
+      mvn"org.jetbrains.kotlin:kotlin-build-tools-api:$kv",
       mvn"org.jetbrains.kotlin:kotlin-build-tools-impl:$kv"
     )
 
@@ -156,18 +154,6 @@ trait KotlinModule extends JavaModule with KotlinModuleApi { outer =>
     )
 
     Seq(compilerDep) ++ btApiDeps ++ scriptCompilerDeps
-  }
-
-  /**
-   * Version of the Kotlin Build Tools API implementation classes used by Mill.
-   * KotlinToolchains is available starting at 2.3.0; for older compiler versions
-   * we keep using a 2.3.x API together with kotlin-build-tools-compat.
-   */
-  def kotlinBuildToolsApiVersion: T[String] = Task {
-    val kv = kotlinVersion()
-    if (Version.parse(kv).isAtLeast(Version.parse("2.3.0"))(using Version.IgnoreQualifierOrdering))
-      kv
-    else "2.3.0"
   }
 
   /**
--- a/mill-build/src/millbuild/Deps.scala
+++ b/mill-build/src/millbuild/Deps.scala
@@ -220,12 +220,10 @@ object Deps {
   val sonatypeCentralClient = mvn"com.lumidion::sonatype-central-client-requests:0.6.0"
   val kotlinVersion = "2.1.20"
   val kspVersion = "2.0.1"
-  val kotlinBuildToolsApiVersion = "2.3.0"
+  val kotlinBuildToolsApiVersion_api = "2.3.0"
   val kotlinCompiler = mvn"org.jetbrains.kotlin:kotlin-compiler:$kotlinVersion"
-  val kotlinBuildToolsApi =
-    mvn"org.jetbrains.kotlin:kotlin-build-tools-api:$kotlinBuildToolsApiVersion"
-  val kotlinBuildToolsCompat =
-    mvn"org.jetbrains.kotlin:kotlin-build-tools-compat:$kotlinBuildToolsApiVersion"
+  val kotlinBuildToolsApi_api =
+    mvn"org.jetbrains.kotlin:kotlin-build-tools-api:$kotlinBuildToolsApiVersion_api"
   val kotlinBuildToolsImpl = mvn"org.jetbrains.kotlin:kotlin-build-tools-impl:$kotlinVersion"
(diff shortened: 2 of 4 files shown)


lefou:
There are merge conflicts.

re-thc:
> There are merge conflicts.

Fixed

lihaoyi:
@lefou could you help review this one?

libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -395,11 +419,12 @@ trait KotlinModule extends JavaModule with KotlinModuleApi { outer =>
 
   /**
    * Enable use of new Kotlin Build API (Beta).
-   * Enabled by default for Kotlin 2.1+ for JVM.
+   * Enabled by default for Kotlin 2.3+ when using the embeddable compiler.
```
lefou, line 408: We should add that this setting is only ever considered, when the embedded compiler is used.

libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -156,6 +158,18 @@ trait KotlinModule extends JavaModule with KotlinModuleApi { outer =>
     Seq(compilerDep) ++ btApiDeps ++ scriptCompilerDeps
   }
 
+  /**
+   * Version of the Kotlin Build Tools API implementation classes used by Mill.
+   * KotlinToolchains is available starting at 2.3.0; for older compiler versions
+   * we keep using a 2.3.x API together with kotlin-build-tools-compat.
+   */
```
lefou: I don't get this comment. Why do we need to have some version and use some compat API, when we don't need it at all, since we no longer support incremental compilation for Kotlin 2.1 and 2.2?

mill-build/src/millbuild/Deps.scala
```diff
@@ -220,8 +220,12 @@ object Deps {
   val sonatypeCentralClient = mvn"com.lumidion::sonatype-central-client-requests:0.6.0"
   val kotlinVersion = "2.1.20"
   val kspVersion = "2.0.1"
+  val kotlinBuildToolsApiVersion = "2.3.0"
```
lefou: To help in maintenance, this should be named `kotlinBuildToolsApiVersion_api`, so we know it is a lower-bound version we should not bump when updates are available. Same for the derived dependencies.

https://github.com/com-lihaoyi/mill/blob/e5a5d9c443c9b175f74043cc018d803c0f35bead/mill-build/src/millbuild/Deps.scala#L9-L13

re-thc:
@lefou so the design of compat was meant to allow older versions with the older APIs to still use the new API but honestly I'm not seeing it working so let's go with the simple way. < 2.3 go with the old style and >= 2.3 this new style. Then everyone is happy.

mill-build/src/millbuild/Deps.scala
re-thc: Updated

libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
re-thc: It was meant to bridge but I could never actually get it to work, so my bad there (or maybe Kotlin). I assume their aim was for 2.1 / 2.2 to combat but doesn't exactly compile for me. Updated.

libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
re-thc: Reverted the old usage, so all good

PR #6860, opened 2026-02-21
Error prone plugins: support NullAway and Picnic

re-thc:
Makes Error prone more robust by using 2 commonly supported plugins:

NullAway: helps with JSpecify and to find null errors
Picnic: applies a set of additional rules to ErrorProne that catches many common bugs


lihaoyi:
@re-thc can you elaborate on the PR description what this PR is all about?

re-thc:
@lihaoyi done. These are useful plugins that are often used to enhance ErrorProne and catch more errors.

lihaoyi:
@re-thc thanks!

Issue #4998, opened 2025-04-25
Review and potentially fix `TaskCtx.env` handling when on Windows


lefou:
On Windows, environment variables don't need to be cases-sensitive, hence a simple lookup for a key might not be enough. 

For example the `scala.sys.env` map has a `withDefault` override that tried for find the variable more vigorous. `Task.ctx.get("UserPROFILE") should return a value, even is there is only a key for `UserProfile`. 

See also this review comment thread: https://github.com/com-lihaoyi/mill/pull/4917#pullrequestreview-2793944318

PR #6867, opened 2026-02-22
Support case-insensitive behavior for `TaskCtx.env` under Windows

lefou:
Fix https://github.com/com-lihaoyi/mill/issues/4998

### DIFF
--- a/core/api/java11/src/mill/api/TaskCtx.scala
+++ b/core/api/java11/src/mill/api/TaskCtx.scala
@@ -39,7 +39,6 @@ trait TaskCtx extends TaskCtx.Dest
  */
 object TaskCtx {
 
-
   private[mill] class Impl(
       val args: IndexedSeq[?],
       dest0: () => os.Path,


gamlerhart:
Contra argument:
It is quite easy break builds with Windows <-> Unix-ish OS. Like having a build work by accident on Window but not others.So, I'm not sure if there Mill itself should encourage 'OS-differences'.
We probably never 100% replicate the Windows behavior as well. eg. Windows probably uses the user local for the case-sensitive matching. That local might be different for the JVM process. So, the edge cases now become harder to debug.

So, the alternative to this fix would be to document that Mill is case sensitive in env variables.

lefou:
Thing is, JVM already resolves env variable with `System.getenv` is a case-insensitive manner, since that's what the OS is requiring it. Under Windows, it's legal to use `PATH` which was defined as `Path`. That means, Mill can fail in such cases, while probably most other JVM tools will handle it properly by using `System.getenv`. Detecting those issue, esp. since it's not reproducible when using `System.genenv` in a small check script or whatever debug mechanism, is really hard.

lefou:
> Windows probably uses the user local for the case-sensitive matching.

If you can find source for that or can clarify it is not locale-specific, that would be great.

gamlerhart:
PS: According to Microsoft the order / case-sensitivity for case variable is not local dependent: https://learn.microsoft.com/en-us/windows/win32/procthread/changing-environment-variables

In reality there seem odd edge cases (eg: https://nullprogram.com/blog/2023/08/23/). But that so out of scope of Mills concerns.

PR #6868, opened 2026-02-22
Use existing Path constructor in error reporting.

PapyChacal:
As per https://github.com/com-lihaoyi/mill/pull/6851#discussion_r2834819115


alexarchambault (approved):
Thanks!

PR #6873, opened 2026-02-23
Avoid using CoursierModule#mapDependencies

alexarchambault:
This works around an issue I [raised in the past](https://github.com/com-lihaoyi/mill/pull/5070). `CoursierModule#mapDependencies` makes dependency resolution less "standard": resolutions relying on it can't be easily done with the coursier CLI or from other build tools for comparison.

Mill using it by default can surprise users trying to make sense of some resolutions, and debug them outside of Mill. This PR makes Mill stop relying on it by default, and makes it use `CoursierModule#resolutionParams` instead.

Only one use of `mapDependencies` can't be handled with `resolutionParams` for now: using a custom Scala organization. AFAIK, this was only useful in the past with the Typelevel fork of the Scala compiler. This compiler fork has been discontinued for a while, and I'm not sure anyone needs to customize the scala organization. So this PR simply drops the ability to customize it.

Lastly, this PR makes sure test modules inherit `resolutionParams` from their main module

### DIFF
--- a/libs/javalib/src/mill/javalib/CoursierModule.scala
+++ b/libs/javalib/src/mill/javalib/CoursierModule.scala
@@ -85,7 +85,7 @@ trait CoursierModule extends mill.api.Module {
    * Override this to customize the set of dependencies.
    *
    * Using this makes it harder to make sense of dependency resolutions and to reproduce those
-   * resolutions via the coursier command-line.
+   * resolutions via the coursier command-line or from other build tools.
    */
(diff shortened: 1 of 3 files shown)


lefou:
Considering, that this will break all user modules that rely on `mapDependencies` this is a breaking change. I wonder what the next Mill version should be then. At least `1.2`. Binary compatibility isn't compromised.

alexarchambault:
@lefou That's not a breaking change, `mapDependencies` is still there and can be used by Mill users. This only makes Mill rely less on it internally / by default.

PR #6874, opened 2026-02-23
Report *.mill files as "wrapped sources" via BSP

alexarchambault:
This makes Mill report its `*.mill` files as "wrapped sources" in BSP via the `buildTarget/wrappedSources` request. This can help BSP clients pass to their presentation compiler the same sources that Mill passes to the compiler. In order to get more accurate infos from it, in particular for the `build.*` stuff in the build (to refer to things defined in other `*.mill` files).

Quick tests on a minimal project show that this doesn't break the currently existing features in Metals, while I can make good use of it in [my own LSP server](https://github.com/alexarchambault/plasmon).

Pull request: https://github.com/com-lihaoyi/mill/pull/6874

### DIFF
--- a/libs/javalib/src/mill/javalib/JavaModule.scala
+++ b/libs/javalib/src/mill/javalib/JavaModule.scala
@@ -899,7 +899,7 @@ trait JavaModule
    * files) are not reported to BSP clients, and must be handled by them in an ad hoc fashion.
    */
   @internal
-  private[mill] def wrappedSources: T[Seq[(original: PathRef, generated: PathRef)]] = Task(Nil)
+  private[mill] def wrappedSources: T[Seq[(original: PathRef, generated: PathRef)]] = Task(Seq())
 
   /**
    * Path to sources generated as part of the `compile` step, eg.  by Java annotation


lefou:
I can't find the BSP spec of the `wrappedSources` request, but I could find these pull requests and reported issues:

* https://github.com/com-lihaoyi/mill/pull/3195
* https://github.com/build-server-protocol/build-server-protocol/pull/673

I still don't get it, why the Scala-CLI developers never tried to upstream the request. This is not at all how Open Sources works. The current state of the BSP project is unmaintained and the repo is kind-of unmaintainable. But I still would prefer to upstream the wrappedSources request into the spec. 

Maybe we can just fork the BSP project from its last stable release tag, which should be still on sbt? Or we can just take maintainership, revert the Bazel migration and add the wrappedSources request directly?

libs/javalib/src/mill/javalib/JavaModule.scala
```diff
@@ -870,6 +870,37 @@ trait JavaModule
...
+   * on a single line.
+   *
+   * Without this separator, these sources are not passed to BSP clients.
+   *
+   * As of writing this, further changes in the original sources prior to adding them
+   * in the generated ones (like removing package directives, which Mill does to its build
+   * files) are not reported to BSP clients, and must be handled by them in an ad hoc fashion.
+   */
+  @internal
+  private[mill] def wrappedSources: T[Seq[(original: PathRef, generated: PathRef)]] = Task(Nil)
```
lefou: For consistence with existing code.

```suggestion
  private[mill] def wrappedSources: T[Seq[(original: PathRef, generated: PathRef)]] = Task(Seq())
```

runner/meta/src/mill/meta/MillBuildRootModule.scala
```diff
@@ -223,21 +235,8 @@ trait MillBuildRootModule()(using rootModuleInfo: RootModule.Info) extends Boots
...
-        generatedScriptSources().wrapped
-
-    val candidates =
-      Lib.findSourceFiles(allMillSources, Seq("scala", "java") ++ buildFileExtensions.asScala.toSeq)
-
-    // We need to unlist those files, which we replaced by generating wrapper scripts
-    val filesToExclude = Lib.findSourceFiles(scriptSources(), buildFileExtensions.asScala.toSeq)
-
-    candidates.filterNot(filesToExclude.contains).map(PathRef(_))
-  }
```
lefou, line 240: Why can this be removed? How does it behave in IDEA?

lefou:
This change makes `*.mill` files more or less "official" source files. Shouldn't we report that to the BSP client somehow? E.g. we could add "Mill" to the supported Languages.

runner/meta/src/mill/meta/MillBuildRootModule.scala
alexarchambault: It can be removed because the default implementation from `JavaModule` now returns the same result, handling `*.mill` files and their generated counterparts the same way (that is, returning only the latter in `allSourceFiles`)

alexarchambault:
> This change makes `*.mill` files more or less "official" source files. Shouldn't we report that to the BSP client somehow? E.g. we could add "Mill" to the supported Languages.

I don't know, editing `.mill` files in VSCode has been working fine for me, I'm not sure anything needs to be changed here. Don't know about IntelliJ, but I think many people use it fine to work on Mill and on `.mill` files.

libs/javalib/src/mill/javalib/JavaModule.scala
alexarchambault: Sure, why not

lefou (approved):

PR #6875, opened 2026-02-23
Don't assume LinkageError-s to be fatal exceptions

alexarchambault:
This makes Mill handle `LinkageError`s (`NoClassDefFoundError` and all) just like standard "non fatal" exceptions. While that kind of error means something went badly wrong in a standard application or server (with a single flat class path and no manual class loading), it's not as bad in Mill. Mill does quite some class loading, and that kind of error can happen, and should be handled / reported just like any other exception IMO.

This PR changes that by adding a custom `mill.api.daemon.internal.NonFatal` extractor, to be used as a substitute to the `NonFatal` provided by the standard library. This also adds a helper method `NonFatal.Try`, meant to be a substitute to `Try.apply`, using our new `NonFatal` rather than the standard library one.

Pull request: https://github.com/com-lihaoyi/mill/pull/6875

### DIFF
--- a/core/api/daemon/src/mill/api/daemon/internal/NonFatal.scala
+++ b/core/api/daemon/src/mill/api/daemon/internal/NonFatal.scala
@@ -8,7 +8,7 @@ import scala.util.{Failure, Success, Try}
  */
 object NonFatal {
 
-  def attempt[T](f: => T): Try[T] =
+  def Try[T](f: => T): Try[T] =
     try Success(f)
     catch {
       case NonFatal(ex) =>
--- a/core/exec/src/mill/exec/ExecutionContexts.scala
+++ b/core/exec/src/mill/exec/ExecutionContexts.scala
@@ -128,7 +128,7 @@ object ExecutionContexts {
       val runnable = new PriorityRunnable(
         priority = priority,
         run0 = () => {
-          val result = NonFatal.attempt(logger.withPromptLine {
+          val result = NonFatal.Try(logger.withPromptLine {
             os.dynamicPwdFunction.withValue(() => makeDest()) {
               mill.api.SystemStreamsUtils.withStreams(logger.streams) {
                 t(logger)
--- a/libs/rpc/src/mill/rpc/MillRpcClient.scala
+++ b/libs/rpc/src/mill/rpc/MillRpcClient.scala
@@ -82,7 +82,7 @@ object MillRpcClient {
     }
 
     def handleServerMessage(msg: ServerToClient): Unit = {
-      val response = NonFatal.attempt(currentServerMessageHandler(msg))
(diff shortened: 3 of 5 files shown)


core/api/daemon/src/mill/api/daemon/internal/NonFatal.scala
```diff
@@ -0,0 +1,29 @@
...
+import scala.util.control.{NonFatal => StdNonFatal}
+import scala.util.{Failure, Success, Try}
+
+/**
+ * Same as [[scala.util.control.NonFatal]], but also considers LinkageError-s to be non fatal
+ */
+object NonFatal {
+
+  extension (t: Try.type)
+    def millNonFatal[T](f: => T): Try[T] =
```
lihaoyi: Let's use a normal static method here rather than an extension method

We shouldn't use an extension method unless we have some particular DSL or something we want to implement, and I don't see that's the case here

core/api/daemon/src/mill/api/daemon/internal/NonFatal.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import scala.util.control.{NonFatal => StdNonFatal}
+import scala.util.{Failure, Success, Try}
+
+/**
+ * Same as [[scala.util.control.NonFatal]], but also considers LinkageError-s to be non fatal
+ */
+object NonFatal {
+
+  def attempt[T](f: => T): Try[T] =
```
lefou: Is there some naming convention you're following here? Can't we name it just `try`?

lolgab: `try` is a keyword, you can't use it as a method name.

lefou: Yeah, ``` `try` ``` is ugly. What about `Try`, with a leading upper case?

If that feels like betrayal, we could also add a inner object `Try` with an `apply` method.

I'm mostly concerned about the code reading experience. `NonFatal.attempt` simply don't suggest it's the equivalent to `scala.util.Try` but with custom `NonFatal` settings.

alexarchambault: Maybe `MillTry`, so that we can import it and use it like
```scala
MillTry {
  ...
}
```

lefou: IMHO, `NonFatal.Try` is good enough. Otherwise, why not `MillNonFatal`?

alexarchambault: Just went for `NonFatal.Try`

PR #6878, opened 2026-02-24
Add JavaModule.jvmOptions and deprecate -J support in javacOptions

### LINKED ISSUE #6806: Add `JavaModule.jvmOptions` and deprecate `-J` support in `javacOptions`
Instead of the hidden capability to forward JVM options via `javacOptions`, we should explicitly configure the options in a dedicated task. 

Related:
* https://github.com/com-lihaoyi/mill/pull/6803

re-thc:
Add JavaModule.jvmOptions and deprecate -J support in `javacOptions`

Fixes #6806

### DIFF
--- a/libs/javalib/src/mill/javalib/SemanticDbJavaModule.scala
+++ b/libs/javalib/src/mill/javalib/SemanticDbJavaModule.scala
@@ -124,7 +124,8 @@ trait SemanticDbJavaModule extends CoursierModule with SemanticDbJavaModuleApi
 
     Task.log.debug(s"effective javac options: ${javacOpts}")
 
-    val (javacCompilerOptions, legacyRuntimeOptions) = JavaModule.splitJavacAndRuntimeOptions(javacOpts)
+    val (javacCompilerOptions, legacyRuntimeOptions) =
+      JavaModule.splitJavacAndRuntimeOptions(javacOpts)
     if (legacyRuntimeOptions.nonEmpty) {
       Task.log.warn(
         "`-J` options in `javacOptions` are deprecated; use `jvmOptions` instead" +


libs/javalib/src/mill/javalib/JavaModule.scala
```diff
@@ -321,6 +321,21 @@ trait JavaModule
...
+   * JVM options passed to the Java compiler worker process.
+   *
+   * Prefer this over `javacOptions` for JVM flags such as `-D`, `--add-opens`, and
+   * `-X` options.
+   */
+  def jvmOptions: T[Seq[String]] = Task { Seq.empty[String] }
+
+  protected[javalib] def splitJavacAndRuntimeOptions(
+      options: Seq[String]
+  ): (Seq[String], Seq[String]) = {
```
lefou: Should return a named tuple and reside in the companion object.

Also, looks like some code duplication with SemanticDbJavaModule. Not sure, if this is supposed to override the other, in which case I'd prefer to use the `override` keyword explicitly, but the code looks identical.

libs/javalib/src/mill/javalib/errorprone/ErrorProneModule.scala
lefou, line 1: Not sure why these changes are here, they look unrelated.

libs/javalib/package.mill
```diff
@@ -55,7 +55,10 @@ object `package` extends MillStableScalaModule {
       // This was `private[mill]`, package private doesn't have the JVM bytecode equivalent, so mima can't check it.
       ProblemFilter.exclude[Problem]("mill.javalib.PublishModule.checkSonatypeCreds"),
       ProblemFilter.exclude[Problem]("mill.javalib.publish.SonatypeHelpers.getArtifactMappings"),
-      ProblemFilter.exclude[Problem]("mill.javalib.publish.PublishInfo.parseFromFile")
+      ProblemFilter.exclude[Problem]("mill.javalib.publish.PublishInfo.parseFromFile"),
+      ProblemFilter.exclude[Problem](
```
lefou: We need a comment here explaining why this is ok. It obviously does not fall into the "This was `private[mill]` before"-section. Also, it looks unrelated.

re-thc: Fixed, reverted this part

libs/javalib/src/mill/javalib/JavaModule.scala
re-thc: Updated

libs/javalib/src/mill/javalib/errorprone/ErrorProneModule.scala
lefou: No explanation, esp. not in the final merge commit. Too bad.

Issue #6871, opened 2026-02-23
Selective Execution Invalidates resolvePublishDependency Task When Adding Modules To A Build


maxstreese:
When adding a new module to an existing Mill build that is independent of any other build modules, Mill invalidates the `resolvePublishDependency` Task. This can lead to many more tasks getting selectively resolved/run than you would like.

The below Bash script illustrates the issue and can be executed for reproduction:
```bash
curl -L https://repo1.maven.org/maven2/com/lihaoyi/mill-dist/1.1.2/mill-dist-1.1.2-mill.sh -o mill

sudo chmod +x mill

cat <<'EOF' > build.mill
//| mill-version: 1.1.2
//| mill-jvm-version: temurin:25

package build

import mill.*
import mill.scalalib.*

object `package` extends Module:

  object foo extends ScalaModule:
    override def scalaVersion = "3.8.1"
EOF

./mill selective.prepare

cat <<'EOF' >> build.mill

  object bar extends ScalaModule:
    override def scalaVersion = "3.8.1"
EOF

./mill selective.resolveTree __.compile
```

The output of the final `selective.resolveTree` command yields:
```json
{
  "def build_.package_$bar$#resolvePublishDependency()mill.api.Task": {
    "external mill.scalalib.ScalaModule": {
      "call mill.scalalib.ScalaModule.compile$(mill.scalalib.ScalaModule)mill.api.Task$Simple": {
        "def build_.package_$foo$#compile()mill.api.Task$Simple": {
          "foo.compile": {}
        }
      }
    }
  },
  "def build_.package_.<clinit>()void": {
    "call build_.package_.<clinit>()void": {
      "def build_.package_$#<init>()void": {
        "bar.compile": {}
      }
    }
  }
}
```

From my perspective as a user, this is unexpected. As someone who is in the process of setting up Mill for a large monorepo, this is also quite problematic because it means that when I onboard another application module to the build, my selective CI workflows would publish all existing apps/libs, which is exactly what I would like to prevent by using selective execution.

maxstreese:
@lihaoyi any assessment here? Is this expected? Am I doing something obviously wrong? For my monorepo use case described above this is quite the obstacle unfortunately.

lihaoyi:
[...]

PR #6880, opened 2026-02-24
Try to improve precision of codesig callgraph analysis

lihaoyi:
Attempts to resolve https://github.com/com-lihaoyi/mill/issues/6871. 

>⏺ Summary of Key Algorithm and Data Structure Changes
>                                                                                                                          
>  1. Per-call-site precise type analysis (LocalSummary.scala)                                                             
>   
>  New dependency: asm-analysis (ASM's dataflow analysis framework).                                                       
>                                                                  
>  New algorithm: analyzeMethodArgTypes runs ASM's Analyzer with a custom BasicInterpreter that preserves precise object
>  types on the operand stack (the default collapses all references to one REFERENCE_VALUE). For each method invoke
>  instruction, it records:
>
>  - callSiteReceiverType: Map[MethodCall, JCls] — the precise type of this on the stack at virtual/special call sites
>  - callSiteArgTypes: Map[MethodCall, Set[JCls]] — precise types of reference-typed arguments on the stack
>
>  These are stored in MethodInfo and flow into ReachabilityAnalysis for narrowing.
>
>  Optimization: Methods without any MethodInsnNode skip the expensive Analyzer entirely. The ClassNode tree is parsed once
>   and shared between MyClassVisitor (call graph extraction) and Analyzer (stack type analysis).
>
>  2. Compound ExternalClsCall(dest, narrow) nodes (ReachabilityAnalysis.scala)
>
>  Old structure: One ExternalClsCall(cls) node per loaded external class. Call nodes connected to ExternalClsCall nodes.
>  ExternalClsCall(X) had edges to ExternalClsCall(parent(X)), forming an external class hierarchy tree. Callback methods
>  were resolved per-level, with ancestor methods reached transitively through parent edges.
>
>  New structure: ExternalClsCall(dest, narrow) is a compound key. dest determines which callback method signatures are
>  collected (by walking ancestors of dest in externalClassLocalDests). narrow determines which local subtypes receive
>  those callbacks (subtypes of narrow via typeToLocalSubtypes). No ExternalClsCall → ExternalClsCall edges exist — the
>  hierarchy walk is done eagerly inside each node's resolution.
>
>  Key algorithm — narrowForDest: For each externalDest at a call site, collects all precise types (receiver + args) and
>  filters to those that are subtypes of dest (via getTransitiveAncestors(pt).contains(dest)). Subtypes that match become
>  the narrow values; if none match, falls back to dest itself (no narrowing, same as old behavior). This naturally handles
>   the fact that externalDests conflates receiver and arg types — the subtype check separates them since a receiver type
>  won't be a subtype of an unrelated arg-derived dest.
>
>  Node creation is demand-driven: Only (dest, narrow) pairs actually referenced by call sites are created, rather than one
>   node per loaded external class. This typically produces fewer graph nodes.
>
>  Edge movement: External callback edges moved from Call → ExternalClsCall to LocalDef → ExternalClsCall. This is because
>  the precise types needed for narrowing live on the enclosing MethodInfo, not on the Call node.
>
>  3. Precomputed lookup tables (ReachabilityAnalysis.scala)
>
>  - callbackMethodsPerDest: Map[Cls, Set[MethodSig]] — Callback methods depend only on dest, not narrow. Precomputed once
>  per unique dest, shared across all ExternalClsCall nodes with the same dest.
>  - typeToLocalSubtypes: HashMap[Cls, Set[Cls]] — Reverse ancestor index mapping each type to its local subtypes. Built by
>   iterating all local classes and indexing them under each of their ancestors. Enables O(1) lookup during ExternalClsCall
>   resolution.
>  - ancestorCache: HashMap[Cls, Set[Cls]] — Memoized getTransitiveAncestors BFS results, shared between pair computation,
>  callbackMethodsPerDest, and typeToLocalSubtypes.
>
>  4. transitiveCallGraphValues dedup fix (ReachabilityAnalysis.scala)
>
>  Old: If multiple nodes in an SCC had edges to the same upstream SCC group, that group's hash was reduced in multiple
>  times. With reduce = _ + _, transitive hashes depended on edge count, not just reachability.
>
>  New: A BitSet tracks seen upstream groups, ensuring each contributes exactly once. This makes transitive hashes depend
>  only on reachability, reducing spurious invalidations when edges change between already-connected nodes.
>
>  Also replaced nodeGroups: Map with nodeGroupsArray: Array[Int] for O(1) group lookup.
>
>  5. Iterative Tarjans (Tarjans.scala)
>
>  Converted recursive DFS to iterative using an explicit call stack (three parallel flat arrays: stackNode, stackChildIdx,
>   stackIsRoot). Prevents StackOverflowError on deep graphs. Same algorithm, same results.
>
>  6. Reverse graph construction (spanningInvalidationTree)
>
>  Replaced groupMap (allocates millions of tuples) with a two-pass approach: first pass counts in-degrees to pre-allocate
>  arrays, second pass fills them. Same result, fewer allocations.
>
>  7. Parallel LocalSummary processing (LocalSummary.scala, JvmModel.scala)
>
>  SymbolTable.Table changed from LinkedHashMap to ConcurrentHashMap with computeIfAbsent. LocalSummary.apply accepts an
>  optional TaskCtx and processes class files in parallel batches using fork.async.
>
>  8. Lazy val hash stability (LocalSummary.scala)
>
>  More robust handling of Scala 3's VarHandle-based lazy val initialization sequences in <clinit>, using regex patterns
>  ($lzy\\d+$, $lzy\\d+\\$lzyHandle$) and a state machine (lazyNameSeenInClinit) to skip the entire
>  lookup()/findVarHandle/PUTSTATIC sequence.
>
>  Test changes
>
>  - Test 15 (15-static-method): Removed false positive Hello#<init>() → Foo#toString. The old code conservatively fanned
>  out from ExternalClsCall(Object) to all local subtypes of Object. The new compound key ExternalClsCall(Object, Hello)
>  correctly narrows to Hello's subtypes, and Hello doesn't override toString.
>  - Tests 19, 20: New regression tests for precise narrowing of external constructor receivers and static method argument
>  types.
>

### DIFF
--- a/runner/codesig/src/mill/codesig/LocalSummary.scala
+++ b/runner/codesig/src/mill/codesig/LocalSummary.scala
@@ -219,16 +219,22 @@ object LocalSummary {
 
                 // For non-static calls, capture the receiver's precise type.
                 // This is the most important signal for narrowing virtual dispatch.
-                if (invokeType != InvokeType.Static && argStartIdx > 0)
+                if (invokeType != InvokeType.Static && argStartIdx > 0) {
                   preciseType(frame.getStack(argStartIdx - 1))
                     .foreach(callReceiverTypes += call -> _)
+                }
 
-                val refArgs = desc.args.indices.collect {
-                  case j if desc.args(j).isInstanceOf[JCls] =>
-                    preciseType(frame.getStack(argStartIdx + j))
-                }.flatten.toSet
+                val refArgs = desc.args.indices
+                  .collect {
+                    case j if desc.args(j).isInstanceOf[JCls] =>
+                      preciseType(frame.getStack(argStartIdx + j))
+                  }
+                  .flatten
+                  .toSet
+
                 if (refArgs.nonEmpty) callArgTypes += call -> refArgs
               }
+
             case _ =>
           }
 
--- a/runner/codesig/src/mill/codesig/ReachabilityAnalysis.scala
+++ b/runner/codesig/src/mill/codesig/ReachabilityAnalysis.scala
@@ -44,12 +44,11 @@ class CallGraphAnalysis(
     if (subtypes.nonEmpty) subtypes else Iterable(dest)
   }
 
-  // Collect all (externalDest, narrowType) pairs needed as ExternalClsCall proxy nodes.
-  // externalDest determines which callback methods are collected (from its ancestors),
-  // narrowType determines which local subtypes are included in the fan-out.
-  // For each externalDest, we find precise types (from both receiver and arg positions)
-  // that are subtypes of that dest, and use them as narrowing. This is uniform across
-  // receiver and arg positions — both narrow the same way based on their bytecode type.
+  // Collect all (dest, narrow) pairs needed as ExternalClsCall proxy nodes, then
+  // transitively expand to include (ancestor, narrow) pairs so that each node can
+  // chain to its parent. This avoids duplicating callback method edges: each node
+  // only handles methods declared at its own level, reaching ancestor methods
+  // transitively through ExternalClsCall → ExternalClsCall parent edges.
   val externalClsCallPairs: Set[(JType.Cls, JType.Cls)] = {
     val pairs = collection.mutable.Set.empty[(JType.Cls, JType.Cls)]
     for ((_, methodInfo) <- methods) {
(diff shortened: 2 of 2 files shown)


lihaoyi:
This seems to increase the runtime of `def codeSignatures` from ~1.3s to ~7s on the Mill repo. So although it would be nice to have the precision improvements, it's probably not worth the performance hit unless we can speed it up further

lihaoyi:
Overhead seems down to 1-2s, may be ok to merge now

PR #6881, opened 2026-02-24
Android: Add Koin example to docs and CI

souvlakias:
### Motivation
- Document the support of Robolectric tests
- Have them in CI

### Adds
- The Koin example (#6833) (only it's Robolectric unit tests) to the CI
- A relevant doc page

### DIFF
--- a/example/thirdparty/koin-androidx-samples/build.mill
+++ b/example/thirdparty/koin-androidx-samples/build.mill
@@ -101,26 +101,3 @@ object examples extends mill.api.Module {
   ]
 }
 */
-
-// Create an AVD, start the emulator, install the APK and run the app on the emulator
-/**
-
-> ./mill show examples.androidx-samples.createAndroidVirtualDevice
-...DeviceId: medium_phone...
-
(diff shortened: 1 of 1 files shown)


example/thirdparty/koin-androidx-samples/build.mill
```diff
@@ -70,10 +70,41 @@ object examples extends mill.api.Module {
...
+      "selector": "CheckModulesTest.Verify Module Configuration",
+      "duration": ...,
+      "status": "Success"
+    }
+  ]
+}
+*/
+
+// Create an AVD, start the emulator, install the APK and run the app on the emulator
+/**
```
souvlakias:
```diff
- /** Usage
+ /**
```
Avoiding the `Usage` here so it doesnt run on CI, but the relevant commands can still be found in the documentation.
Downside is that this will show as a comment block though

Image
Not sure if there's an alternative

vaslabs: we can remove it entirely, it's repetitive in all the examples anyway

vaslabs:
lgtm

PR #6883, opened 2026-02-25
Use C1 for no-server (faster cold start)

re-thc:
This change is only for "--no-server".

Since this is a single-run mode the startup time to analyze and compile C2 takes longer than the run itself.

On a project like Ammonite the cold start before is 29s and after is 24s (tested locally). Pretty big savings.

On a Spring hello world example before is 7.5s and after is 5.9s.

Big savings, pretty useful for cases like CI or other single use cases.

### DIFF
--- a/runner/launcher/src/mill/launcher/MillProcessLauncher.scala
+++ b/runner/launcher/src/mill/launcher/MillProcessLauncher.scala
@@ -29,7 +29,13 @@ object MillProcessLauncher {
 
     val userPropsSeq = ClientUtil.getUserSetProperties().map { case (k, v) => s"-D$k=$v" }.toSeq
 
-    val cmd = millLaunchJvmCommand(runnerClasspath, outMode, workDir, millRepositories) ++
+    val cmd = millLaunchJvmCommand(
+      runnerClasspath,
+      outMode,
+      workDir,
+      millRepositories,
+      extraJvmOpts = Seq("-XX:TieredStopAtLevel=1")
+    ) ++
       userPropsSeq ++
       Seq(mainClass, processDir.toString, outMode.asString, useFileLocks.toString) ++
       loadMillConfig(ConfigConstants.millOpts, workDir) ++
@@ -192,7 +198,8 @@ object MillProcessLauncher {
       runnerClasspath: Seq[os.Path],
       outMode: OutFolderMode,
       workDir: os.Path,
-      millRepositories: Seq[String]
+      millRepositories: Seq[String],
+      extraJvmOpts: Seq[String] = Seq.empty
   ): Seq[String] = {
     val millProps = sys.props.toSeq
       .filter(_._1.startsWith("MILL_"))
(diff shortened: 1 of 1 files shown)


lihaoyi:
Not sure if I want to include this kind of performance tuning, since I assume it would really depend on the workload of what task you are running? If so then it's hard to say if it's really a win, or if it's just a sideways change

lefou:
Maybe, we can have a runtime configuration option for that, so users can customize to their need? We already have `mill-jvm-opts` which controls the JVM, although the documentation isn't that clear about which specific Mill JVM processess do use this flag. We could add  more fine-grained control like `mill-daemon-jvm-opts` and `mill-nodaemon-jvm-opts`.

lihaoyi:
putting this config in `mill-jvm-opts` should already work if I'm not mistaken

lefou:
> putting this config in `mill-jvm-opts` should already work if I'm not mistaken

But I assume, @re-thc doesn't want to apply the settings to a long-running daemon JVM, only to the shorter-running non-daemon JVM.

lihaoyi:
mill-jvm-opts should work for both daemon and non-daemon modes

lefou:
> mill-jvm-opts should work for both daemon and non-daemon modes

Exactly. That why it's probably not the configuration the OP want to use to configure the non-daemon JVM by e.g. adding it to the project, as it likely will make the Mill daemon slower.

re-thc:
I haven't found a use case where this is worse. The more modules involved, the more classes involve i.e. bigger project the bigger overhead. So unless there's a case where overhead is low but time taken is still so huge that C2 benefits then ok.

Having said that I can't prove it so happy to go with @lefou 's suggestion on making it a config if all agrees. Then we can leave both options so no accidents occur. I'd not remember to add jvm args just for specific use cases.

As a developer, Mill can be hot, daemon and is relatively fast. 

However if I do use Mill in CI it's very hard to cache as it's not split up. My runs are bad. If I have to run Mill remotely or in docker to test different environments the same applies. Many other examples but Mill is not optimized for non-human continuous use case right now. When I mean not optimized a Mill run of 30s has Gradle taking 3s and Gradle is broken enough. The same run theoretically could take 1s.

lefou:
Maybe, just maybe, we could experiment with an optional build-file cache feature, which can be checked into the project repository, so that builds become instantaneous. The vague idea includes some dedicated cache location containing a single jar, which is prepared by Mill in the background and used, when appropriate. Once we have remote caching support, this is probably obsolete.

re-thc:
@lefou I did extensive profiling and I see 2 main issues.

1. The non-client (daemon or not) has to download a lot in the default path. If I build any project this happens (regardless of modules used). Coursier scanning, hashing, etc then also come up. This set is part of out and there's no way to just cache it. So either this should just be part of the download that gets the server/daemon going or at least be isolated to be cache-able.

2. Coursier / Zinc compiler has to do a lot to get started. Ignoring user defined modules or 3rd party modules that can be dynamic, there are system modules e.g. JavaModule that ALWAYS exists. They always have to be compiled again. If there was a way to cache to separately (as above) or even ship the precompiled jar for just that it would speed things up.

With these 2 we'd be at least 50-75% better than where we are today.

PR #6885, opened 2026-02-26
Bump scala to 3.8.2

lihaoyi:


TotallyAvailable:
@lihaoyi ***Please*** remember to set the GITHUB_TOKEN permissions, 'Secret Source' & 'persist-credentials' for **all** workflows to the ***absolute minimum*** required to avoid using default **write-all** tokens.

This post has nothing to do with the original PR and should only be seen as a Heads-up.  Direct mentions are not the "nicest" way of doing this, I get it. Apologies for that.

PR #6905, opened 2026-03-05
Run at most one BSP server per build client

alexarchambault:
This makes Mill run at most one BSP server per build client type in a given project. This identifies build clients with their "display name" in the build initialize request, and ensures only one build server for this display name runs at a time.

If two build servers run at a time for build clients with the same display name, it is likely the first one is a zombie that's not needed anymore. Yet, it keeps watching the build sources, and tries to re-compile the build every time its sources change. Build servers hinder each other in such a scenario, so this PR makes this more unlikely, by having the second build server kill the first one upon startup.

Knowing if a build server already runs is handled the same way as locking when running tasks, except we also put the display name in the lock file name and its accompanying "active" info file name, and we default to killing the other Mill process here, rather than waiting for it to release the lock.

The build client display name is taken into account here, so that users can still have a Mill BSP server running for IntelliJ and another one running for VSCode.

### DIFF
--- a/core/constants/src/mill/constants/OutFiles.java
+++ b/core/constants/src/mill/constants/OutFiles.java
@@ -128,9 +128,23 @@ public class OutFiles {
     public final String millOutLock = "mill-out-lock";
 
     /**
+     * Lock file used for exclusively running the Mill BSP server for that lock id
+     */
+    public final String millBspLock(String lockId) {
+      return "mill-bsp-" + lockId + "-lock";
+    }
+
+    /**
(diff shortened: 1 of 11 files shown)


lefou:
This is only relevant to instances sharing the same `BSP_OUTPUT_DIR`, right?

alexarchambault:
> This is only relevant to instances sharing the same `BSP_OUTPUT_DIR`, right?

Yes, the lock files for this live under the output directory that ended up being picked (`.bsp/out` by default here)

Issue #6908, opened 2026-03-06
AAR files in classpath for android modules


GoogeTan:
Mill gives aar files in a class path which results in LSP's being unable to properly work. To be precise metals think that module can not be compiled due to import errors which lead to classes from aar dependencies.  

Problem was tested on mill versions 1.0.6 and 1.1.0. 

Steps to reproduce:
+ Clone [minimal example](https://github.com/GoogeTan/mill-arr-bug-showcase) 
+ Run `./mill mill.bsp.BSP/install`
+ Run `./mill __.compile` to check that everything compiles. 
+ Run `python tester.py` to check that classpath contains aar files.
+ Open metal doctor to see phantom errors 

P.s. the tester script is vibecoded because I was not sure how to call BSP protocol

lihaoyi:
CC @vaslabs

vaslabs:
I'll take a look, we haven't looked at mixing scala + android modules yet, probably something is not right there, give me some time. Thanks for the reproducible repo

vaslabs:
I can see the import issue and the error from metals, but not the aars in the classpath, on either version (below is with 1.0.6)

<details>

```
error id: file://<WORKSPACE>/android/skia/src/main/scala/example/example/MainActivity.scala:skia.
file://<WORKSPACE>/android/skia/src/main/scala/example/example/MainActivity.scala
empty definition using pc, found symbol in pc: skia.
semanticdb not found
empty definition using fallback
non-local guesses:
	 -org/jetbrains/skia/org/jetbrains/skia.
	 -org/jetbrains/skia.
	 -scala/Predef.org.jetbrains.skia.
offset: 133
uri: file://<WORKSPACE>/android/skia/src/main/scala/example/example/MainActivity.scala
text:
```scala
package example.example

import android.app.Activity
import android.os.Bundle
import android.widget.TextView
import org.jetbrains.ski@@a.*

class MainActivity extends Activity:
  override def onCreate(savedInstanceState: Bundle): Unit =
    super.onCreate(savedInstanceState)

    // A simple view to prove the activity loads
    val textView = new TextView(this)
    textView.setText("Hello from Scala 3 on Android!")

    setContentView(textView)

  def drawSomething(): Unit = 
    val surface = Surface.Companion.makeRasterN32Premul(100, 100)
    val canvas = surface.getCanvas()
    
    val paint = new Paint()
    paint.setColor(0xFFFF0000) // ARGB Red
    
[...]

Issue #6624, opened 2026-01-19
Android/Kotlin: Some IDE highlighting is broken


vaslabs:
For instance in architecture samples:

<img width="1326" height="778" alt="Image" src="https://github.com/user-attachments/assets/b1078edc-e961-4a52-998c-f1e5d3ca1650" />

This should work through the kotlinc -Xplugin parameter passing the compose plugin, but it seems a "bspBuildTargetKotlincOptions" is missing

vaslabs:
this maybe fairly complex,

if something like this

```scala
  @internal
  private[mill] def bspKotlinBuildTargetTask: Task[KotlinBuildTarget] = Task.Anon {
    KotlinBuildTarget(
      languageVersion = kotlinLanguageVersion(),
      apiVersion = kotlinApiVersion(),
      kotlincOptions = allKotlincOptions(),
      jvmBuildTarget = bspJvmBuildTargetTask()
    )
  }

  @internal
  override def bspBuildTargetData: Task[Option[(String, AnyRef)]] = Task.Anon {
    Some((KotlinBuildTarget.dataKind, bspKotlinBuildTargetTask()))
  }
```

doesn't work (ref: https://github.com/JetBrains/intellij-bsp/blob/main/protocol/src/main/kotlin/org/jetbrains/bsp/protocol/KotlinBuildTarget.kt)


Maybe there's also something here https://github.com/JetBrains/hirschgarten

souvlakias:
Noting a (failed?) [attempt](https://github.com/vaslabs-ltd/mill/commit/6908b80491ec45651d5e1bcdc6c6f989f95b44d0) (for basically kotlin bsp support):

The [KotlinBuildTarget](https://github.com/JetBrains/intellij-bsp/blob/main/protocol/src/main/kotlin/org/jetbrains/bsp/protocol/KotlinBuildTarget.kt) of the jetbrains bsp extension seems indeed promising. It also has a test that helps see how it processes it internally: ([link](https://github.com/JetBrains/intellij-bsp/blob/77a8f7f3abbdb2a4dc03d57bc7081944638a28f4/src/test/kotlin/org/jetbrains/plugins/bsp/magicmetamodel/impl/workspacemodel/impl/updaters/transformers/ModuleDetailsToJavaModuleTransformerTest.kt#L295))

A KotlinBuildTarget, doesnt exist in the bsp4j library, so I tried creating a java object with the fields the extension expects for Kotlin
[...]

PR #6916, opened 2026-03-08
Android: mill.idea/

vaslabs:
## Context

Part of https://github.com/com-lihaoyi/mill/issues/6624

This is part of improving the IDE integration (and for this PR in particular intelliJ) with the Android support in mill.

After a long and hard look at BSP , I think it may be a better way to at least tackle first direct idea support. This PR does not resolve the whole issue but makes a small step in that direction.

## Provided in the PR

The PR fixes the missing dependencies from the .idea/ configuration files which were ignored because, the GenIdea uses the resolvedMvnDeps which have the AARs which are filtered out. 

## Results

This sets up the AARs properly with sources

image

## Limitations

- Still , this does not support mixing mill's Android support with Scala (related to https://github.com/com-lihaoyi/mill/issues/6908 ) , we'll need to cover a lot of missing features for that since the androidlib has not got any scala context .
- It does not fix the wider IDE highlighting issue but it makes a small step towards that direction, for the `mill.idea/` approach

### DIFF
--- a/libs/androidlib/src/mill/androidlib/idea/GenIdeaModule.scala
+++ b/libs/androidlib/src/mill/androidlib/idea/GenIdeaModule.scala
@@ -1,0 +1,29 @@
+package mill.androidlib.idea
+
+import mill.androidlib.AndroidModule
+import mill.api.daemon.experimental
+import mill.api.daemon.internal.internal
+import mill.api.{ModuleCtx, Task}
+
+@experimental
+trait GenIdeaModule extends mill.javalib.idea.GenIdeaModule {
+
+  override private[mill] def extDependencies = Task {
+    super.extDependencies() ++ javaModuleRef().androidUnpackedAarMvnDeps().flatMap(_.classesJar)
+  }
+
+  def javaModuleRef: mill.api.ModuleRef[AndroidModule]
+
+}
+
+@internal
+object GenIdeaModule {
+  trait Wrap(javaModule0: AndroidModule) extends mill.api.Module {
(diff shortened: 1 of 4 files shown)


vaslabs:
cc: @souvlakias 

I think this might be more doable than bsp, regardless of this PR being merged, maybe have a look for a wider Kotlin/ Android support with the gen idea approach

vaslabs:
scala support will have to come after we build support for basic android + kotlin , there will be a lot of issues there I'm expecting, not exclusive to IDE support

libs/androidlib/src/mill/androidlib/idea/GenIdeaModule.scala
```diff
@@ -0,0 +1,29 @@
...
+import mill.androidlib.AndroidModule
+import mill.api.daemon.experimental
+import mill.api.daemon.internal.internal
+import mill.api.{ModuleCtx, Task}
+
+@experimental
+trait GenIdeaModule extends mill.javalib.idea.GenIdeaModule {
+
+  override private[mill] def extDependencies = Task {
+    super.extDependencies() ++ javaModuleRef().androidUnpackedAarMvnDeps().flatMap(_.classesJar)
```
vaslabs, line 12: ~~better use androidResolvedMvnDeps here~~
ah no, that task will not have sources

vaslabs, line 12: should exclude aars from `super.extDependencies()`

vaslabs:
Merged into https://github.com/com-lihaoyi/mill/pull/6918

Issue #6907, opened 2026-03-06
`KtlintModule` ignores module sources


viluon:
`KtlintModule` runs the linter on the entire module directory, instead of just the module sources.

The root cause is an interaction between the filtering done inside `ktlintAction`:
https://github.com/com-lihaoyi/mill/blob/e3e0beb29d4f337769318ebb6789b2bdb92e3d6e/libs/kotlinlib/src/mill/kotlinlib/ktlint/KtlintModule.scala#L123-L127

and how `ktlintAction` is invoked from the `ktlint` `Command`:
https://github.com/com-lihaoyi/mill/blob/e3e0beb29d4f337769318ebb6789b2bdb92e3d6e/libs/kotlinlib/src/mill/kotlinlib/ktlint/KtlintModule.scala#L21-L29

The issue is that `sources()` are typically directories. `ktlintAction` never walks them, yet it filters out anything that doesn't have a `kt` or `kts` extension. When the resultant `Seq` ends up being empty, the ktlint CLI defaults to linting everything in the current working directory.

This default behaviour is a nasty surprise, IMO `ktlintAction` should take care to never hit this path by checking that the `Seq` is not empty. Of course it also needs to walk the directories first.

lefou:
Thanks for the report. I agree, this is not the correct behavior. I opened PR #6917 to address this.

PR #6917, opened 2026-03-09
Ktlint: properly scan for source files and detect missing provided files

lefou:
Instead of using the provided paths as-is, we traverse the given paths for source files, that we can format.

If no paths were given, we now fail the action, otherwise `ktlint` would fall back to using the current working directory, which is not what we want. Not finding any actual source files is ok, though.

Fix https://github.com/com-lihaoyi/mill/issues/6907

Pull request: https://github.com/com-lihaoyi/mill/pull/6917

### DIFF
--- a/libs/kotlinlib/src/mill/kotlinlib/ktlint/KtlintModule.scala
+++ b/libs/kotlinlib/src/mill/kotlinlib/ktlint/KtlintModule.scala
@@ -106,14 +106,14 @@ object KtlintModule extends ExternalModule with KtlintModule with DefaultTaskMod
   )(using ctx: mill.api.TaskCtx): Seq[os.Path] = {
     val sourceFiles = Lib.findSourceFiles(pathsToFormat, Seq("kt", "kts"))
       // skip formatting single-file projects since Palantir Format messes up the header block
-      .filter(!os.read(_).startsWith("//|"))
+      .filterNot(p => p.ext == "kt" && os.read(p).startsWith("//|"))
     if (sourceFiles.isEmpty) {
       // no files found
       if (pathsToFormat.isEmpty) Task.fail("No paths selected.")
       else {
-        // The sources simple didn't contain any formattable source file,
-        // which is probably ok for a freshly set up project
-        // we just return, since ktlint defaults to format the current working dir
+        // Nothing to do here, which is ok for a freshly set up project.
+        // Please note that ktlint tends to default to format the current working
+        // dir when no source paths are given, which is not what we want.
         ctx.log.info(s"No kotlin sources found.")
         return Seq()
       }


libs/kotlinlib/src/mill/kotlinlib/ktlint/KtlintModule.scala
```diff
@@ -99,15 +99,30 @@ object KtlintModule extends ExternalModule with KtlintModule with DefaultTaskMod
...
-      filesToFormat: Seq[PathRef],
+      pathsToFormat: Seq[PathRef],
       config: Option[PathRef],
       options: Seq[String],
       classPath: Seq[PathRef]
-  )(using ctx: mill.api.TaskCtx): Unit = {
+  )(using ctx: mill.api.TaskCtx): Seq[os.Path] = {
+    val sourceFiles = Lib.findSourceFiles(pathsToFormat, Seq("kt", "kts"))
+      // skip formatting single-file projects since Palantir Format messes up the header block
+      .filterNot(p => p.ext == "kt" && os.read(p).startsWith("//|"))
```
viluon, line 109: this may work better extracted to a task to avoid recomputation when `pathsToFormat` change but `sourceFiles` do not

lefou: What do you mean? The `pathsToFormat` come from commands, which are not cached, so I don't see any way a cached task can be of any help here.

viluon: yeah, it's not really a fault of this PR, it's just that the way it's written (via commands) makes `KtlintModule` difficult to use from a task.

In Gradle with [the ktlint plugin](https://github.com/JLLeitschuh/ktlint-gradle), ktlint runs as part of `build`, which is a task that's normally cached. Calling one of the Mill commands in a task is overkill because the whole thing is recomputed even when something like `src/main/resources/` changes. `ktlintAction` is private, so in my build scripts I reimplemented it and wrapped with tasks (not commands) to fit nicely into a `build`-like task with proper caching.

When the scanning and filtering of source files happens internally, there's no clean way to reuse the logic from a task while maintaining fine-grained caching. But as I said, it's not a fault of this PR, just a sadness with `KtlintModule` as a whole.

lefou: I see. `KtlintModule` is designed to be used from CLI. The name is a bit unfortunate therefore. It's still possible to cache results, e.g. with a worker and a persistent cache, but that's not the scope of this PR.

lefou:
Not sure about the failing `integration.dedicated` test. Looks unrelated to me. It also fails in other PRs. 

Edit: Looks like just flakiness.

PR #6918, opened 2026-03-09
GenIdea: Fix Android Sources + Kotlin Facet Generation

souvlakias:
Part of #6624, includes #6916

## Android Sources
- In `GenIdeaImpl.scala` : Pickup up sources jars from dep directory 
- Create `androidlib.GenIdeaModule` for overrides:
	- `extDependencies`
	- add `androidLinkedResources` to the generatedSources passed down to the IDE
	(these aren't in regular `generatedSources`)

### Motivation
Since Kotlin support is missing in BSP, the IDE can't know about potential `kotlinc` flags, which results to broken highlighting (e.g. the compose plugin for android apps)

### Solution
Create an experimental `KotlinIdeaModule` that fills in the [facet](https://www.jetbrains.com/help/idea/facet-page.html) [field](https://github.com/com-lihaoyi/mill/blob/189e3160d882cdddcc2b975d3adcad582793af05/core/api/daemon/src/mill/api/daemon/internal/idea/ResolvedModule.scala#L14) of KotlinModules

### Proper usage
If a project was initially loaded with BSP, facets still don't get recognized right after running `GenIdea`
So a way to get this to work is to ignore BSP and:
- make sure `.bsp` and `.idea` are deleted
- Run `genIdea` in the directory 
- Open the project (don't import, so it isn't required to select BSP)
- Works

Note: Testing on `androidtodo` thirdparty example

Co-authored by: @vaslabs

### DIFF
--- a/contrib/jmh/src/mill/contrib/jmh/JmhModule.scala
+++ b/contrib/jmh/src/mill/contrib/jmh/JmhModule.scala
@@ -37,7 +37,7 @@ trait JmhModule extends JavaModule {
   def mvnDeps = super.mvnDeps() ++ Seq(mvn"org.openjdk.jmh:jmh-core:${jmhCoreVersion()}")
 
   def runJmh(args: String*) =
-    Task.Command {
+    Task.Command(exclusive = true) {
       val (_, resources) = generateBenchmarkSources()
       Jvm.callProcess(
         mainClass = "org.openjdk.jmh.Main",
--- a/contrib/scoverage/src/mill/contrib/scoverage/ScoverageModule.scala
+++ b/contrib/scoverage/src/mill/contrib/scoverage/ScoverageModule.scala
@@ -218,7 +218,10 @@ trait ScoverageModule extends ScalaModule { outer: ScalaModule =>
       }
     }
 
-    override def skipIdea = true
+    // Disable this module in BSP clients by default
+    override def enableBsp: Boolean = false
+    // Skip this module in IntelliJ IDEA by default
+    override def skipIdea: Boolean = true
   }
 
   trait ScoverageTests extends ScalaTests {
--- a/integration/dedicated/bsp-server/resources/snapshots/build-targets-output-paths-synthetic-root.json
+++ b/integration/dedicated/bsp-server/resources/snapshots/build-targets-output-paths-synthetic-root.json
@@ -1,0 +1,39 @@
+{
+  "items": [
+    {
+      "target": {
+        "uri": "file:///workspace/mill-synthetic-root-target"
+      },
+      "outputPaths": [
+        {
+          "uri": "file:///workspace/.idea/",
+          "kind": 2
+        },
+        {
+          "uri": "file:///workspace/out/",
+          "kind": 2
+        },
+        {
+          "uri": "file:///workspace/.bsp/",
+          "kind": 2
+        },
+        {
+          "uri": "file:///workspace/.bloop/",
+          "kind": 2
+        },
+        {
+          "uri": "file:///workspace/.project/",
+          "kind": 2
+        },
+        {
+          "uri": "file:///workspace/.classpath/",
+          "kind": 2
+        },
+        {
+          "uri": "file:///workspace/.settings/",
+          "kind": 2
+        }
+      ]
+    }
+  ]
+}
--- a/integration/dedicated/bsp-server/resources/snapshots/build-targets-output-paths.json
+++ b/integration/dedicated/bsp-server/resources/snapshots/build-targets-output-paths.json
@@ -92,41 +92,6 @@
     },
     {
       "target": {
-        "uri": "file:///workspace/mill-synthetic-root-target"
-      },
-      "outputPaths": [
-        {
-          "uri": "file:///workspace/.idea/",
-          "kind": 2
-        },
-        {
-          "uri": "file:///workspace/out/",
-          "kind": 2
-        },
-        {
-          "uri": "file:///workspace/.bsp/",
-          "kind": 2
-        },
-        {
-          "uri": "file:///workspace/.bloop/",
-          "kind": 2
-        },
-        {
-          "uri": "file:///workspace/.project/",
-          "kind": 2
-        },
-        {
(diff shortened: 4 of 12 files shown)


vaslabs:
We can always feature flag this either with a ideaUseKotlinFacets or with another trait

libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -498,6 +499,72 @@ trait KotlinModule extends JavaModule with KotlinModuleApi { outer =>
...
+                          )
+                        ),
+                        Element(
+                          "stringArg",
+                          attributes = Map(
+                            "name" -> "apiVersion",
+                            "arg" -> kotlinApiVersion()
+                          )
+                        ),
+                        Element(
```
vaslabs: could make a helper function for this like `stringArgElement(name, arg)`

vaslabs:
there's a binary compatibility issue, so I suggest we make a new trait for this and someone can mix it in , maybe something like a KotlinIdeaModule and mark it as experimental

vaslabs:
+ if bsp is removed, it doesn't seem the dependencies are working properly, you'll need parts of my PR

https://github.com/com-lihaoyi/mill/pull/6916/changes#diff-2ea10356e02638938089e7d77940a3bd567a0b0a04bf2391ac3114fe4e0e79b4R12

perhaps we can merge them together anyway

vaslabs:
the proper steps to setup the project with idea are:

- make sure .bsp and .idea are deleted
- run `mill mill.idea/`
- open the project (just simply file -> open) not new project from existing sources

souvlakias:
> * if bsp is removed, it doesn't seem the dependencies are working properly, you'll need parts of my PR

Yes, I tested on top of #6916 so the sources were included in the xmls

vaslabs:
> > * if bsp is removed, it doesn't seem the dependencies are working properly, you'll need parts of my PR
> 
> Yes, I tested on top of #6916 so the sources were included in the xmls

merge them together here and I'll close my PR

vaslabs:
I tested this with architecture samples ->
```
// Mill configuration for the Android Todo App project.
object app
  extends AndroidAppKotlinModule, KotlinIdeaModule,
    AndroidR8AppModule,
    AndroidHiltSupport {
```

and it's working

Screenshot From 2026-03-12 09-21-05

Screenshot From 2026-03-12 09-20-25

vaslabs:
let's add some docs @souvlakias for android and wrap it up

EDIT:

ah sorry, need to fix this too ->

image

R sources could be passed here

image

souvlakias:
> R sources could be passed here
> image

I tried passing them to the `sources` section but IDE only picked them up if I added them to the `dependencies` section
image
The xml entry for that is:
```xml
<orderEntry type="module-library">
  <library>
    <CLASSES>
      <root url="file://$MODULE_DIR$/../../out/app/androidCompiledRClasses.dest/classes" />
		...OR...
	  <root url="jar://$MODULE_DIR$/../../out/app/androidProcessedResources.dest/R.jar!/" />
    </CLASSES>
    <JAVADOC />
    <SOURCES />
  </library>
</orderEntry>
  ```
  
  EDIT: my bad i was passing classes instead of sources

**Comment by vaslabs** [comment]
> > R sources could be passed here
> > image
> 
> I tried passing them to the `sources` section but IDE only picked them up if I added them to the `dependencies` section image The xml entry for that is:
> 
> ```
> <orderEntry type="module-library">
>   <library>
>     <CLASSES>
>       <root url="file://$MODULE_DIR$/../../out/app/androidCompiledRClasses.dest/classes" />
> 		...OR...
> 	  <root url="jar://$MODULE_DIR$/../../out/app/androidProcessedResources.dest/R.jar!/" />
>     </CLASSES>
>     <JAVADOC />
>     <SOURCES />
>   </library>
> </orderEntry>
> ```

have you looked into android facets? maybe there's something there

File: libs/androidlib/src/mill/androidlib/idea/GenIdeaModule.scala [hunk 2]
```diff
@@ -0,0 +1,38 @@
...
+
+  override private[mill] def extDependencies = Task {
+    super.extDependencies().filter(_.path.ext != "aar")
+      ++ javaModuleRef().androidUnpackedAarMvnDeps().flatMap(_.classesJar)
+  }
+
+  override private[mill] def moduleGeneratedSources = Task {
+    val superSources = super.moduleGeneratedSources()
+    val rSourcesDirs =
+      Seq(javaModuleRef().androidLinkedResources().path / "generatedSources").map(PathRef(_))
```
souvlakias, line 25: Since these sources aren't in normal `generatedSources` task, this passes them down to the IDE so R highlighting works

vaslabs: You can add this as a scaladoc comment

vaslabs:
I'll say the only major think left is testing and verifying what happens when:

## For Android

 
- just bsp -> we know the issue - we can just document it in a troubleshooting section for android
- just mill.idea/ -> this should work but with the limitation of needing to run it again everytime deps or android resources change?
- combining bsp + mill.idea/ -> does this work and resolves both issues or is making things worse?

### Minor things

- Add the `KotlinIdeaModule` in the examples that need it (architecture-samples and compose-samples ?)

souvlakias:
> just mill.idea/ -> this should work but with the limitation of needing to run it again everytime deps or android resources change?

- For changes to resources, just running the `compile` task will do, since that calls `androidLinkedResources` where IDE gets `R.java` from
- For changes to deps, it may be needed to run `genIdea` again yes. (sometimes the `compile` task will update stuff too, e.g. if a dep is removed) 

> combining bsp + mill.idea/ -> does this work and resolves both issues or is making things worse?

I think that it does work but don't really see the benefits, since BSP also needs to resync the project for all changes to be updated

Afterall the `genIdea` approach seems better than expected 😅

vaslabs:
This is ready for review @lihaoyi @lefou

PR #6927, opened 2026-03-12
Don't return synthetic root results when not asked to in buildTargetSources

alexarchambault:
This fixes a bug in the BSP server, in `buildTargetSources`, where it as returning results for the "synthetic root module" even these were not requested.


alexarchambault:
The "synthetic root module" being a module added for IntelliJ to handle root modules fine.

PR #6928, opened 2026-03-12
Add Task.Command#makeExclusive

alexarchambault:
This adds a `makeExclusive` helper on `Task.Command`, that returns a copy of the command with `exclusive` set to true.

This can be helpful if users want to make existing Mill commands exclusive, like
```scala
def testForked(args: String*) =
  super.testForked(args*).makeExclusive
def publishLocal(localIvyRepo: String = null, sources: Boolean = true, doc: Boolean = true, transitive: Boolean = false) =
  super.publishLocal(localIvyRepo, sources, doc, transitive).makeExclusive
```

This can be useful in some contexts, like if the tests do fancy termnal stuff, or when publishing locally many modules at a time while passing `--transitive=true` (if several command runs try to publish a given transitive dependency at the same time, they sometimes clash, with one writing files that the other doesn't expect to exist).

### DIFF
--- a/core/api/src/mill/api/Task.scala
+++ b/core/api/src/mill/api/Task.scala
@@ -536,6 +536,25 @@ object Task {
     override def asCommand: Some[Command[T]] = Some(this)
     // FIXME: deprecated return type: Change to Option
     override def writerOpt: Some[Writer[?]] = Some(writer)
+
+    /**
+     * Make this command exclusive
+     *
+     * This ensures this command doesn't run concurrently with other commands.
+     */
+    def makeExclusive: Command[T] =
+      if (exclusive) this
+      else
+        new Command[T](
+          inputs,
+          evaluate0,
+          ctx0,
+          writer,
+          isPrivate,
+          exclusive = true,
+          persistent,
+          selectiveInputs0
+        )
   }
   class Worker[T](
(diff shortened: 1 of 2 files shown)


lihaoyi:
Maybe call it withExclusive to follow the "with*" convention often used for these copy-update methods?

core/api/src/mill/api/Task.scala
```diff
@@ -536,6 +536,25 @@ object Task {
...
     // FIXME: deprecated return type: Change to Option
     override def writerOpt: Some[Writer[?]] = Some(writer)
+
+    /**
+     * Make this command exclusive
+     *
+     * This ensures this command doesn't run concurrently with other commands.
+     */
+    def makeExclusive: Command[T] =
+      if (exclusive) this
```
lefou, line 546: Not sure about this. We always want to define a new task when it is named but anchored in a new location (another module or inheritance level), otherwise the context is wrong.

alexarchambault:
Closing this one for now, ~it seems there are already straightforward ways to achieve that (making an existing command exclusive), like~

```scala
def publishLocal(…): Task.Command[Unit] =
  Task.Command(exclusive = true) {
    super.publishLocal(…)()
  }
```

PR #6933, opened 2026-03-13
Allow customising linker OutputDirectory

Quafadas:
Companion to #6932 . 

I think this is the minimally invasive change which enables this. 

Tests use a copy (taken from scala js repo) of the `MemOutputDirectory.scala` to check that the linking functions write into a `CustomOutput` writer as expected. 

Whether it's a good idea or not is up for discussion; it seems to be philosophically compatible with `ScalaJSConfigModule`. My motivation is to experiment with post-processing the linker outputs in a plugin. Link -> re-write -> Hash -> refresh loop without disk IO overhead. 

Ideally; I think it would have been possible to do that experiment without changing mill. That seems to me both an argument for and against this PR.


Quafadas:
To me, the failing test does not look related to this PR?

PR #6934, opened 2026-03-13
fix: catch NoClassDefFoundError during test discovery

klejejs:
## Problem

Mill's test discovery crashes with `NoClassDefFoundError` when the test classpath contains libraries that use Scala 3 inline methods with synthetic type references.

Scala 3 inline methods can generate synthetic type references in method signatures that don't have corresponding class files — e.g. package namespace types like `io.bullet.borer.internal`. When `TestRunnerUtils.matchFingerprints` calls `cls.getDeclaredMethods()` / `cls.getMethods()`, the JVM tries to resolve these types and throws `NoClassDefFoundError`.

## Reproducer

Using the [tresql](https://github.com/mrumkovskis/tresql) project (`mill-build` branch):

```bash
git clone -b mill-build https://github.com/mrumkovskis/tresql.git && cd tresql
./mill _.test  # Scala 3.3.7 tests fail during discovery
```

Error:
```
NoClassDefFoundError: io/bullet/borer/internal
  at java.lang.Class.getDeclaredMethods0(Class.java:-2)
  at mill.javalib.testrunner.TestRunnerUtils$.matchFingerprints
```

The `io.bullet.borer.internal` type is a Scala 3 compiler synthetic for the `internal` package namespace, generated in `Encoder`/`Decoder` inline method stubs. No such class file exists in the borer JAR.

sbt handles this gracefully because its test discovery uses zinc analysis rather than JVM reflection.

## Fix

Wrap class loading and fingerprint matching in `TestRunnerUtils` with try/catch for `NoClassDefFoundError` and `LinkageError`, skipping classes that can't be fully resolved during reflection.

## Test plan
- [x] Verified fix resolves the tresql reproducer — all 3 Scala versions pass
- [ ] Existing Mill test suite

### DIFF
--- a/libs/javalib/testrunner/src/mill/javalib/testrunner/TestRunnerUtils.scala
+++ b/libs/javalib/testrunner/src/mill/javalib/testrunner/TestRunnerUtils.scala
@@ -63,7 +63,9 @@ import scala.math.Ordering.Implicits.*
                 // sbt-jupiter-interface ignores fingerprinting since JUnit5 has its own resolving mechanism
                 Some((cls, fingerprints.head))
               } else if (
-                Modifier.isAbstract(cls.getModifiers) || cls.isInterface || publicConstructorCount > 1
+                Modifier.isAbstract(
+                  cls.getModifiers
+                ) || cls.isInterface || publicConstructorCount > 1
               ) {
                 None
               } else {
--- a/libs/javalib/testrunner/src/mill/javalib/testrunner/TestRunnerUtils.scala
+++ b/libs/javalib/testrunner/src/mill/javalib/testrunner/TestRunnerUtils.scala
@@ -77,10 +77,14 @@ import scala.math.Ordering.Implicits.*
               }
             } catch {
               // Scala 3 inline methods can generate synthetic type references in method
-              // signatures that don't have corresponding class files (e.g. package namespace
-              // types). When the JVM reflects on such classes via getDeclaredMethods/getMethods,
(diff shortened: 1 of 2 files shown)


lefou (commented):
Looks good to me. 

Personally, I'd prefer to log these kind of gracefully ignored cases, as there are no other ways to communicate that there were issues, but looks like we don't have any logger around.

It would be nice to also have a test case, that reproduces the specific issue and it's fix?

klejejs:
> Looks good to me.
> 
> Personally, I'd prefer to log these kind of gracefully ignored cases, as there are no other ways to communicate that there were issues, but looks like we don't have any logger around.
> 
> It would be nice to also have a test case, that reproduces the specific issue and it's fix?

Thank you for your review!

Similarly as other examples in this file, I added `System.err.println` to these cases.
Also, added some relevant test cases.

alexarchambault:
Shouldn't this be opt-in / disabled by default? And enabled via a `ignoreTestDiscoveryNoClassDefFound` method on `TestModule` say, that would default to false. These changes could make some test suites being ignored, and users might not check the output for the logged line added here, especially on CIs.

And we could print a message about how to enable this via the method above if it's not enabled, but keep throwing until it's enabled.

lefou:
@alexarchambault Do you have an example of a test suite that would be ignored but shouldn't?

klejejs:
Closing as I don't have time to work on this now.

Issue #6944, opened 2026-03-18
coursierCacheCustomizer logger is overridden by CoursierTickerResolutionLogger


rkophs:
While extending the `JavaModule` trait, I override the `coursierCacheCustomizer` to include a logger in order to export some details/metrics around Coursier download performance and errors like so:

```scala
override def coursierCacheCustomizer : 
  Task[Option[FileCache[coursier.util.Task] => FileCache[coursier.util.Task]]] = Task.Anon {
    Some { (cache: FileCache[coursier.util.Task]) => cache.withLogger(...)}
}
```

However, it appears [this line here ](https://github.com/com-lihaoyi/mill/blob/94d79d8d33be6c2cca963e6d0c3ad6f2ce5244c0/libs/util/java11/src/mill/util/Jvm.scala#L573) overrides the logger that we provide. I can appreciate why the CoursierTickerResolutionLogger is necessary, but maybe we can use a delegate pattern such that a wrapper logger delegates to the `CoursierTickerResolutionLogger` and the user-provided logger in succession? So the Jvm code becomes:

```scala
pipe { cache =>
    coursierCacheCustomizer.fold(cache)(c => c.apply(cache))
  }
  .pipe { cache =>
    ctx.fold(cache) { c =>
      val tickerLogger = new CoursierTickerResolutionLogger(c)
      val existingLogger = cache.logger  // user-provided logger, if any
      val composedLogger = existingLogger match {
        case CacheLogger.nop => tickerLogger
        case existing        => new CompositeCacheLogger(Seq(tickerLogger, existing))
      }
      cache.withLogger(composedLogger)
    }
  }
```
I'm happy to post a PR if this is a direction we want to go in :)

alexarchambault:
Maybe we could do the opposite, that is having Mill do [its own customizations](https://github.com/com-lihaoyi/mill/blob/94d79d8d33be6c2cca963e6d0c3ad6f2ce5244c0/libs/util/java11/src/mill/util/Jvm.scala#L572-L578) first, then [call the user customizer](https://github.com/com-lihaoyi/mill/blob/94d79d8d33be6c2cca963e6d0c3ad6f2ce5244c0/libs/util/java11/src/mill/util/Jvm.scala#L569-L571).

That way, it'd be up to the user customizer to erase Mill's logger or compose it alongside another logger, say.

rkophs:
Thanks y'all! I'm going to close this out since the associated PR was merged

PR #6945, opened 2026-03-19
Fix coursierCacheCustomizer being overridden by CoursierTickerResolutionLogger (#6944)

rkophs:
Addresses https://github.com/com-lihaoyi/mill/issues/6944

While extending the `JavaModule` trait, I override the `coursierCacheCustomizer` to include a logger in order to export some details/metrics around Coursier download performance and errors like so:

```scala
override def coursierCacheCustomizer : 
  Task[Option[FileCache[coursier.util.Task] => FileCache[coursier.util.Task]]] = Task.Anon {
    Some { (cache: FileCache[coursier.util.Task]) => cache.withLogger(...)}
}
```

However, it appears [this line here ](https://github.com/com-lihaoyi/mill/blob/94d79d8d33be6c2cca963e6d0c3ad6f2ce5244c0/libs/util/java11/src/mill/util/Jvm.scala#L573) overrides the logger that we provide

The suggested fix is to swap the ordering so that the user has more flexibility choose to:
 - override the logger 
 - or compose in the existing `CoursierTickerResolutionLogger` (via the parameterized `cache.logger` in `coursierCacheCustomizer`) so as to delegate to both.

### DIFF
--- a/libs/javalib/test/src/mill/javalib/CoursierParametersTests.scala
+++ b/libs/javalib/test/src/mill/javalib/CoursierParametersTests.scala
@@ -6,7 +6,7 @@ import mill.testkit.{TestRootModule, UnitTester}
 import mill.util.TokenReaders.*
 import utest.*
 
-import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.atomic.AtomicBoolean
 
 object CoursierParametersTests extends TestSuite {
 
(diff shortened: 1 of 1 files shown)


lihaoyi:
@alexarchambault I'll let you review and merge this one, since you'd be more familiar with this part of the codebase

PR #6957, opened 2026-03-23
Android: Create an `android-sources.jar` for Idea Highlighting

### LINKED ISSUE #6956: Android: Include sources of `android.jar`
Currently sources of `android.jar` are not available in the IDE:

Image

Image

</br></br>

`android.jar` is a dependency that comes from sdk `platforms` package.
It's sources should come from the `sources;<android_version>` sdk package
AndroidStudio Reference:

Image

souvlakias:
To close #6956 

Add a flag that defaults to `true` except in CI that:
- Includes `sources;<android_version>` in the needed packages
- Copies the sources directory into a `android-sources.jar` next to `android.jar`
This works, since current implementation of GenIdea looks for sources based on their path, and fallbacks to searching for `-sources.jar` next of the respective classes jar.

</br>
A minor issue is that unresolved references inside the jars appear as syntax errors in IDEA
image

Android Studio seems to suppress the highlighting errors: 
image


libs/androidlib/src/mill/androidlib/AndroidSdkManagerModule.scala
```diff
@@ -269,6 +269,7 @@ trait AndroidSdkManagerModule extends ExternalModule {
         "platform-tools", // adb
         s"build-tools;${buildToolsVersion()}",
         s"platforms;${platformsVersion()}",
+        s"sources;${platformsVersion()}", // sources for the platform (android.jar)
```
vaslabs: should we control this with a flag? we don't need sources in the CI for example

souvlakias: added a flag that defaults to `true` except in CI

vaslabs:
this is ready @lihaoyi

PR #6959, opened 2026-03-23
Add Task.Command#makeExclusive

alexarchambault:
New attempt at https://github.com/com-lihaoyi/mill/pull/6928

This adds a `makeExclusive` helper on `Task.Command`, that can be used to make a copy of the command with `exclusive` set to true.

This can be helpful if users want to make existing Mill commands exclusive, like
```scala
def testForked(args: String*) =
  super.testForked(args*).makeExclusive(true)
def publishLocal(localIvyRepo: String = null, sources: Boolean = true, doc: Boolean = true, transitive: Boolean = false) =
  super.publishLocal(localIvyRepo, sources, doc, transitive).makeExclusive(true)
```

This can be useful in some contexts, like if the tests do fancy terminal stuff and cannot run in parallel because of that, or when publishing locally many modules at a time while passing `--transitive=true` (if several command runs try to publish a given transitive dependency at the same time, they sometimes clash, with one writing files that the other doesn't expect to exist already).

### DIFF
--- a/core/api/src/mill/api/Task.scala
+++ b/core/api/src/mill/api/Task.scala
@@ -542,7 +542,7 @@ object Task {
      *
      * Changing it to true ensures this command doesn't run concurrently with other commands.
      */
-    def withExclusive(value: Boolean)(implicit ctx: ModuleCtx): Command[T] =
+    def makeExclusive(value: Boolean)(implicit ctx: ModuleCtx): Command[T] =
       if (exclusive == value) this
       else
         new Command[T](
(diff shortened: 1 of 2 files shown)


lefou:
> ```scala
> def testForked(args: String*) =
>   super.testForked(args*).withExclusive(true)
> def publishLocal(localIvyRepo: String = null, sources: Boolean = true, doc: Boolean = true, transitive: Boolean = false) =
>   super.publishLocal(localIvyRepo, sources, doc, transitive).withExclusive(true)
> ```

Is the resulting binary even calling the `super` task? 

I think I'm open to the idea of this PR, but have issue with the name, which claims to be a harmless property changer, while what we get is a potentially new command with an copied implementation. I'd prefer a name that make this more clear, but I currently don't have an idea. Maybe there are some good naming patterns in the copy-on-write domain?

alexarchambault:
I initially named it `makeExclusive` in https://github.com/com-lihaoyi/mill/pull/6928. I don't mind either way.

alexarchambault:
Just changed it to `makeExclusive` again

PR #6963, opened 2026-03-25
Update zinc to 2.0.0-M16

lefou:


lihaoyi:
Seems like this bumps the upstream zinc version and drops java 11 support, so we may not be able to upgrade unless we special case Java 11-17 to run Javac in a subproces

lihaoyi:
Superseded by https://github.com/com-lihaoyi/mill/pull/7103

PR #6958, opened 2026-03-23
Android: Scala support

souvlakias:
Initial work on supporting Android Scala projects for JVM-target apps. See #6949 and https://github.com/com-lihaoyi/mill/issues/6949#issuecomment-4110203223

## Additions
- `AndroidScalaModule` and `AndroidAppScalaModule`
- Rename androidlib's `GenIdeaModule` to `GenIdeaAndroidModule` for clarity
- A `GenIdeaAndroidScalaModule` and a `BspAndroidScalaModule` to be used by the new Scala Modules to solve the diamond problem of `AndroidModule` and `ScalaModule` using different idea and bsp modules internally
- A `hello-scala` example that builds scala android app (for JVM target), and a doc entry 

## Later work
- Support building scala android apps for native target

PR #6964, opened 2026-03-26
Android: Minor cleanup and dexing/shrinking fixes

souvlakias:
## Changes

Make `androidLinkedResources` return a structure instead of just its `Task.dest`
```scala
case class AndroidLinkedResources(
  apk: PathRef,
  generatedSourcesDir: PathRef,
  proguardRulesFile: PathRef
)
```
so downstream tasks can safely use these fields.

**Legacy code cleanup:**
- Proguard rules that are meant for d8 ([main-dex-rules](https://github.com/com-lihaoyi/mill/blob/main/libs/androidlib/src/mill/androidlib/AndroidModule.scala#L858)  and [mainDexPlatformRules](https://github.com/com-lihaoyi/mill/blob/main/libs/androidlib/src/mill/androidlib/AndroidModule.scala#L30)) were needed for api <21 which we can ignore as no modern builds require it

**R8 fixes:**
- Generate proguard keep rules for R8 with AAPT2 (`--proguard` flag), pick them up and pass them to R8 
- Add `androidR8ExtraRules` task to be able to configure R8 with more ease from `build.mill`
- The logic of passing the classpath with `--classpath @<filename_with_cp_entries>` was accidentally ignored, so this adds this back. (This is helpful because it makes the cli command more compact)

## Future work
Gradle seems to use D8 by default for faster (debug) builds and only uses R8 when minification or optimization is enabled (release builds). ([Ref](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/gradle-core/src/main/java/com/android/build/gradle/internal/TaskManager.kt;l=1723?q=maybeCreateJavaCodeShrinkerTask&ss=android-studio%2Fplatform%2Ftools%2Fbase))
In the current implementation, mixing in `AndroidR8AppModule` always enables R8, regardless of build type.
We could do the same as gradle, determine which tool to use based on the build settings, and even completely drop the R8 trait

Additionally, if we do keep the trait, some Proguard-related tasks currently reside in `AndroidModule` and `AndroidAppModule` but are only relevant to R8 and could be moved into `AndroidR8AppModule`

### DIFF
--- a/example/thirdparty/androidtodo/build.mill
+++ b/example/thirdparty/androidtodo/build.mill
@@ -107,7 +107,7 @@ object app
     mvn"com.google.dagger:hilt-android-compiler:2.56"
   )
 
-  object test extends KspTests, AndroidAppKotlinTests, AndroidHiltSupport, TestModule.Junit4,
+  object test extends KspTests, AndroidHiltSupport, AndroidAppKotlinTests, TestModule.Junit4,
         KotlinIdeaModule {
 
     def moduleDeps = super.moduleDeps ++ Seq(`shared-test`)
--- a/libs/androidlib/src/mill/androidlib/AndroidAppBundle.scala
+++ b/libs/androidlib/src/mill/androidlib/AndroidAppBundle.scala
@@ -28,7 +28,7 @@ trait AndroidAppBundle extends AndroidAppModule with JavaModule {
    */
   def androidBundleZip: T[PathRef] = Task {
     val dexFile = androidDex().path
-    val resFile = androidLinkedResources().path / "apk/res.apk"
+    val resFile = androidLinkedResources().apk.path
     val baseDir = Task.dest / "base"
     val appDir = Task.dest / "app"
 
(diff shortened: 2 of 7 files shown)


vaslabs:
I thought the main Dex task was to configure which classes are loaded first from a multi Dex build. How does this change? Also, I thought this was only relevant for d8 builds?

I don't remember exactly, but if there's a difference how we configure for d8 and r8, we should override the class + we need an example to demonstrate why this is a problem imo

souvlakias:
> I thought the main Dex task was to configure which classes are loaded first from a multi Dex build. How does this change? Also, I thought this was only relevant for d8 builds?
> 
> I don't remember exactly, but if there's a difference how we configure for d8 and r8, we should override the class + we need an example to demonstrate why this is a problem imo

👍 I'll look more into it

vaslabs:
Lgtm, for good measure test with compose samples and android Todo

vaslabs:
For the future work, let's open an issue to track. Iirc I was also under the impression that d8 was used for debug, but couldn't see evidence when running the examples, I may have missed something. Let's discuss over a new issue though

souvlakias:
> Lgtm, for good measure test with compose samples and android Todo

- Compose Samples [pass](https://github.com/vaslabs-ltd/compose-samples-with-mill/actions/runs/24053825690/job/70155486154)
- androidtodo's unit tests fail, ~~but I'm having some trouble finding out what the cause is~~

EDIT: from manual testing, it seems that they broke with #6958. Was a trait mixing order issue, fixed.

vaslabs (approved):

PR #6965, opened 2026-03-27
Add Docker config, Docker credential helper and Google ADC credential configurations for Jib image builder in contrib/docker

jirihausner:
This PR expands the authentication capabilities of the Jib image builder in contrib/docker by introducing multiple credentials retrieval methods beyond environment variables while maintaining backwards compatibility. 

Changed files:

- **JibBuild.scala**: Added `JibCredentialConfig` sealed trait with four credential types and backward-compatible constructors and accessors around original `credentialsEnvironment` parameter of `DockerDaemonImage` and `RegistryImage` case classes
- **JibTypes.scala**: Refactored credential retrieval using `CredentialRetrieverFactory` via unified`configuredCredentialRetriever` method


jirihausner:
@lihaoyi @lefou ready for review whenever you have time, thanks!

stevechy:
Thank you! I was just looking into this module. This is a separate issue but with the current env var setup it seems like the credentials are getting cached or maybe tied to the daemon sometimes?  Using the docker style `config.json` auth would be easier.

nafg:
Why are credentials involved in building, rather than pushing? How do I push to a private registry?

stevechy:
> Why are credentials involved in building, rather than pushing? How do I push to a private registry?

The build task will build and push. To add a private registry you add the host in the qualified name of the jibTargetImage task.  

I think it's similar to calling containerize in the jib-core lib. https://github.com/GoogleContainerTools/jib/tree/master/jib-core#examples

PR #6967, opened 2026-03-28
Scala scripting blog post

lihaoyi:


website/blog/modules/ROOT/pages/20-making-scala-scripting-actually-good.adoc
```diff
@@ -0,0 +1,610 @@
...
+without first running `javac`. Consider this small Java program
+
+.Test.java
+```java
+import java.util.*;
+import java.util.stream.Collectors;
+
+public class Test {
+    public static void main(String[] args) {
+        Map<String, List<String>> graph = Map.of(
```
gamlerhart, line 102: While the larger point is correct, I think showing Java 8-style, without 'var' etc makes Java look 'worse' than it is. 

I would may use the Java best case here, with var' and without class (compact source, since Java 25). Avoiding 1 round of bike-shedding discussion on Reddit/hacker news =)

PR #6968, opened 2026-03-28
Add OWASP Dependency Check Contrib

gamlerhart:
Motivation: In many projects there is some policy that requires to scan dependencies for vulnerable versions.

This adds support for the OWASP dependency check,
that allows to scan dependencies against vulnerability databases: https://dependency-check.github.io/DependencyCheck/dependency-check-cli/index.html

The implementation is deliberately based on the
OWASP dependency check CLI.
This allows us to use the existing documentation
for all the flags: https://dependency-check.github.io/DependencyCheck/dependency-check-cli/arguments.html. It avoids having a separate set of flags naming conventions.


contrib/dependencycheck/test/src/mill/contrib/sbom/DependencyCheckModuleTests.scala
```diff
@@ -0,0 +1,111 @@
...
+import mill.contrib.dependencycheck.{DependencyCheckJavaModule, DependencyCheckModule}
+import mill.testkit.{TestRootModule, UnitTester}
+import utest.{TestSuite, Tests, test}
+import mill.api.{Discover, PathRef, Task}
+import mill.*
+import mill.javalib.*
+import os.{Path, pwd}
+
+object TestModule extends TestRootModule {
+  // TODO: Discussion: Run these tests in Mill CI?
```
gamlerhart: Question/Discussion. Running Dependency Check is somewhat a flaky business, as it fetches data from the internet and checks against only services.

So, in my experience it sometimes fails on services being down, or API's changing etc.

Plus it needs credentials of the OSS repository to actually work.

Therefore, I made the tests skip running, unless you explicitly configure a OSSINDEX credentials via environment.

contrib/owaspdependencycheck/src/mill/contrib/owaspdependencycheck/OwaspDependencyCheckModule.scala
```diff
@@ -0,0 +1,103 @@
...
+
+  def dependencyCheckClassLoader: Worker[java.net.URLClassLoader] = Task.Worker {
+    mill.util.Jvm.createClassLoader(dependencyCheckClasspath().map(_.path))
+  }
+
+  def worker: Worker[DependencyCheckInstance] =
+    Task.Worker { new DependencyCheckInstance(dependencyCheckClassLoader()) }
+
+  private[dependencycheck] class DependencyCheckInstance(cl: ClassLoader) extends AutoCloseable {
+    val depencencyCheckCli = cl.loadClass("org.owasp.dependencycheck.App")
```
gamlerhart, line 91: Currently I load the CLI into the Mill process via class loader. If this proofs to be a issue, we still can switch to launch an extra process in the future.

contrib/dependencycheck/src/mill/contrib/dependencycheck/DependencyCheckModule.scala
```diff
@@ -0,0 +1,103 @@
...
+   * @return
+   */
+  def dependencyCheckFailTask: Boolean = true
+
+  case class DependencyCheckResult(reportFiles: Seq[PathRef], exitCode: Int)
+      derives upickle.ReadWriter {
+    def success: Boolean = exitCode == 0
+  }
+
+  def dependencyCheck(): Task.Command[DependencyCheckResult] = Task.Command(exclusive = true) {
```
gamlerhart: I assume:  Task.Command(exclusive = true) garantee that only one scan runs?

Otherwise, I'll add a lock to the scanning. Because the OWASP dependency scan isn't thread safe at all.

lihaoyi:
`dependencycheck` is a bit of a generic name for what this does. How about `owaspcheck` or similar?

PR #6988, opened 2026-04-06
Android support desugaring

irodotos7:
## Before desugaring

image

## After desugaring 

image


vaslabs:
I think we also need an example that demonstrates the use case, with old APIs

vaslabs:
looks reasonable but some checks are failing

souvlakias:
This still needs some work to get it all working. More specifically: creating and including the compiled desugar classes in the apk. (creating the `desugar_lib.dex` can be done using the [L8 cli](https://r8.googlesource.com/r8/+/refs/heads/main/src/main/java/com/android/tools/r8/L8.java#195) tool)

Also, an issue is that [AndroidBuildSettings](https://github.com/com-lihaoyi/mill/blob/main/libs/androidlib/src/mill/androidlib/AndroidBuildTypeSettings.scala) that already has the `enableDesugaring` flag, is only defined when extending the R8 trait, but this is a feature that also concerns D8, so it's a bit weird to have 2 flags

vaslabs:
I've done some experimentation with this and I see no point in supporting it. Check my latest commits:
1. for some reason, I can even build with jvm version 7 and androidMinSdk 25. Even with java 8 syntax.
2. mill does not work with global java version < 11 , at best you can run testLocal for unit tests, but the rest wrappers require new versions of java
3. You cannot go below android 25, newest android devices reject the apps as not secure. Having someone building apps for < 25 would be an extreme edge case in my opinion and we can spend those resources elsewhere
4. I am not sure L8 is required to build this (in that case we'll need to have the android tools available in an android l8 worker. But again, I couldn't make the app break with sdk 25 and java 7.

I suggest we close https://github.com/com-lihaoyi/mill/issues/6976 and in the case where someone comes up with a reproducible example that demonstrates a problem and requires desugaring we revisit again

vaslabs:
@irodotos7 please close this PR

lefou:
The current PR description isn't specific enough. If this is still relevant, please explain the issue or write, what the images are suspected to show. Also explain the fix this PR applies.

vaslabs:
> The current PR description isn't specific enough. If this is still relevant, please explain the issue or write, what the images are suspected to show. Also explain the fix this PR applies.

we were experimenting with desugaring https://github.com/com-lihaoyi/mill/pull/6988#issuecomment-4698222556

but I can't replicate a specific issue. In theory, if you try to build for android sdk <= 26 , you'll need desugaring, but I couldn't replicate it . I think there's no need to push forward to support this (at least for now)

PR #6990, opened 2026-04-08
Add resolveProtoc task to ScalaPBModule for Coursier-based protoc resolution

lemony312:
## Summary

- Adds opt-in `resolveProtoc` and `scalaPBProtocVersion` tasks to `ScalaPBModule`
- Resolves a platform-specific native `protoc` binary via Coursier, cached as a Mill task
- Default behavior unchanged (`scalaPBProtocPath` still defaults to `None` / embedded Java protoc)

## Motivation

ScalaPB's embedded Java protoc (`os72/protoc-jar`) downloads and extracts the protoc binary at **runtime** inside `ScalaPBC.main()`. When multiple `ScalaPBModule`s compile in parallel, each invocation races to download/extract the same binary to the same location concurrently, causing intermittent failures.

By resolving protoc as a **Mill task**, the binary is downloaded once via Coursier, copied into `Task.dest`, and cached. Parallel modules that depend on `resolveProtoc` are serialized by Mill's evaluator on the resolution step, then all use the same cached result. This eliminates the race condition at the task graph level rather than requiring runtime locking or retry logic.

## Usage

Users opt in by overriding `scalaPBProtocPath`:

```scala
object myModule extends ScalaPBModule {
  override def scalaPBProtocPath = Task { Some(resolveProtoc().path.toString) }
}
```

`scalaPBProtocVersion` can also be overridden to pin a specific protoc version:

```scala
override def scalaPBProtocVersion = "3.21.12"
```

## Test plan

- [ ] `extractsVersionFromScalaPB` — reflective version extraction from ScalaPB classpath
- [ ] `resolvesExecutableBinary` — Coursier resolution produces an existing executable binary
- [ ] `protocPathUsesResolvedBinary` — opt-in wiring from `resolveProtoc` through `scalaPBProtocPath`
- [ ] `protocFlagReachesCompileOptions` — `--protoc=` flag propagates to compile options
- [ ] `compilesWithResolvedProtoc` — end-to-end `.proto` compilation using the resolved native binary

Run with: `./mill contrib.scalapblib.test`


lemony312:
This PR addresses the issue reported in #7009 — parallel `ScalaPBModule` compilation causes intermittent failures due to the embedded protoc binary extraction race condition.

contrib/scalapblib/src/mill/contrib/scalapblib/ScalaPBModule.scala
```diff
@@ -77,7 +77,75 @@ trait ScalaPBModule extends ScalaModule {
...
+  /**
+   * Resolves a platform-specific `protoc` binary via Coursier.
+   *
+   * This avoids the embedded Java protoc that ScalaPB bundles, which is prone to
+   * race conditions when multiple modules compile in parallel (each tries to
+   * download/extract the same binary concurrently).
+   *
+   * By resolving protoc as a Mill task, the binary is downloaded once and cached.
+   */
+  def resolveProtoc: T[PathRef] = Task {
```
lihaoyi: Let's prefix this task with `scalaPB` for consistency with the other tasks on this module

lemony312: https://github.com/com-lihaoyi/mill/pull/6990/changes/12f23785c8a4382631c48c51d715a2ab2c84313d

PR #6998, opened 2026-04-12
Share zinc worker classloaders across meta-build levels

lihaoyi:
Among all the various tasks we can possibly share, these are especially worth sharing because 

1. Every build i guaranteed to duplicate them in the meta-build to compile the build files, and in the main build to compile user modules
2. They are expensive: both in memory footprint to keep classloaders around, and in initialization time

Empirically, this shaves about ~4s off a cold `rm -rf out && time ./mill core.api.compile` on the Mill repo. Not a huge speedup, but not nothing


core/api/daemon/src/mill/api/daemon/ClassLoaderCache.scala
```diff
@@ -0,0 +1,91 @@
...
+            _ = timerField.setAccessible(true)
+            timerOpt0 = timerField.get(module)
+            getOrElseMethod <- timerOpt0.getClass.getMethods.find(_.getName == "getOrElse")
+            timer <-
+              Option(getOrElseMethod.invoke(timerOpt0, null).asInstanceOf[java.util.Timer])
+          } {
+            timer.cancel()
+          }
+        } catch {
+          case _: LinkageError => ()
```
lefou, line 85: This case should have an comment

PR #7012, opened 2026-04-14
OWASP Dependency Check Improvements

gamlerhart:
Improvements for the OWASP Dependency check:

- Removing OfflineModule trait: Its untested and barely makes sense.
- Avoid triggering compling the JavaModule, by only scanning the mvnDeps true instead of the runClasspath


lefou:
> * Removing OfflineModule trait: Its untested and barely makes sense.

I disagree. The support makes sence but the correct implementation was missing. The `prepareOffline` command needs to be overridden and ensure the `dependencyCheckClasspath` is resolved and present.

gamlerhart:
@lefou I'll take a look and will add it =).

PR #7017, opened 2026-04-15
scalajslib: fix LinkerInput equality so the linker cache hits

### LINKED ISSUE #7016: scalajslib: fastLinkJS linker cache never hits (regression since 1.1.3)
## Summary

Since Mill 1.1.3, the inner `ScalaJSLinker` `CachedFactory` inside `ScalaJSWorkerImpl` never reuses its cached `clearableLinker` across `fastLinkJS` / `fullLinkJS` invocations. Every run creates a fresh linker with empty optimizer/emitter caches. Symptom: `Optimizer: Batch mode: true` and `Class/Method tree cache stats: reused: 0 -- invalidated: N` on every invocation regardless of what actually changed between runs.

On a large Scala.js project this turns every dev-loop link into a cold link.

## Affected versions

| Version | Status |
|---|---|
| ≤ 1.1.2 | **unaffected** — pre-refactor, `LinkerInput` used Mill's structural API types |
| 1.1.3 | **first affected release** |
| 1.1.4, 1.1.5, `main` | affected |

## Regression introduced in

Commit `e5a5d9c44`, PR #6872 ("Allow users to fully customize Scala.js linker config via a new module"), released in 1.1.3.

Before that PR, `LinkerInput` held Mill's own worker-API types (`mill.scalajslib.worker.api.ModuleKind`, `ESFeatures`, `ModuleSplitStyle`, `OutputPatterns`, etc.) — all structural case classes / sealed enums, so field-wise equality worked. The refactor collapsed those into a single `config: sjs.StandardConfig` without noticing that the replacement type has reference equality, silently breaking the cache.

## Proposed fix

Scala.js already ships a structural fingerprint for exactly this use case — [`StandardConfig.fingerprint`](https://github.com/scala-js/scala-js/blob/main/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala#L317):

```scala
def fingerprint(config: StandardConfig): String = Fingerprint.fingerprint(config)
```

Minimal fix — override `equals`/`hashCode` on `LinkerInput`:

```scala
private case class LinkerInput(
    dest: Either[File, sjs.OutputDirectory],
    config: sjs.StandardConfig
) {
  override def equals(other: Any): Boolean = other match {
    case that: LinkerInput =>
      dest == that.dest &&
        sjs.StandardConfig.fingerprint(config) ==
          sjs.StandardConfig.fingerprint(that.config)
    case _ => false
  }
  override def hashCode(): Int =
    (dest, sjs.StandardConfig.fingerprint(config)).hashCode
}
```

Happy to open a PR.

<details>
<summary><b>Minimal reproduction</b></summary>

`build.mill`:
```scala
package build
import mill.*, scalalib.*, scalajslib.*

object app extends ScalaJSModule {
  def scalaVersion = "3.3.4"
  def scalaJSVersion = "1.19.0"
}
```

`app/src/Main.scala`:
```scala
object Main {
    val xs = (1 to 100).toList
    val mapped = xs.map(_ * 2).filter(_ % 3 == 0)
    val grouped = mapped.groupBy(_ % 5).view.mapValues(_.sum).toMap
    println(s"sum = ${mapped.sum}, groups = $grouped")
  }
}
```

Run `fastLinkJS` three times, mutating the source between runs so the outer Mill task cache doesn't short-circuit and hide the inner-cache bug:

```bash
./mill clean app
./mill app.fastLinkJS 2>&1 | grep -E "Batch mode|cache stats|Optimizing"
printf '\n// bump 1\n' >> app/src/Main.scala
./mill app.fastLinkJS 2>&1 | grep -E "Batch mode|cache stats|Optimizing"
printf '\n// bump 2\n' >> app/src/Main.scala
./mill app.fastLinkJS 2>&1 | grep -E "Batch mode|cache stats|Optimizing"
```

**Actual** — three byte-identical runs:
```
Optimizer: Batch mode: true
Optimizer: Optimizing 2933 methods.
Emitter: Class tree cache stats: reused: 0 -- invalidated: 283
Emitter: Method tree cache stats: reused: 0 -- invalidated: 907
```

**Expected** — runs 2 and 3 should show `Batch mode: false` and `reused: 282 -- invalidated: 1` (only the one class whose source changed).

</details>

<details>
<summary><b>Root cause</b></summary>

`libs/scalajslib/worker/1/src/mill/scalajslib/worker/ScalaJSWorkerImpl.scala#L28-L31`:

```scala
private case class LinkerInput(
    dest: Either[File, sjs.OutputDirectory],
    config: sjs.StandardConfig
)
```

`LinkerInput` is a case class → equality is field-wise via `==`. But `sjs.StandardConfig` is declared as `final class` ([StandardConfig.scala#L24](https://github.com/scala-js/scala-js/blob/main/linker-interface/shared/src/main/scala/org/scalajs/linker/interface/StandardConfig.scala#L24)) with **no `equals`/`hashCode` override**, so it falls back to reference identity.

Mill builds a fresh `StandardConfig` on every `fastLinkJS` invocation via `ScalaJSConfig.config(...)` → `sjs.StandardConfig().withX(...).withY(...)`. Each builder call goes through `StandardConfig#copy(...)` which allocates a new instance. Two structurally identical calls produce reference-distinct `StandardConfig`s → `LinkerInput#equals` returns false → `CachedFactoryBase#getOrCreate` misses the `unusedCache` lookup ([CachedFactoryBase.scala#L79-L81](https://github.com/com-lihaoyi/mill/blob/main/libs/util/java11/src/mill/util/CachedFactoryBase.scala#L79-L81)) → `setup` → `createLinker` → `StandardImpl.clearableLinker(config)` → brand-new linker with empty optimizer/emitter state.

The orphaned previous linker sits in `unusedCache` and is never reused.

</details>

<details>
<summary><b>Direct proof (independent of Mill, runs in scala-cli)</b></summary>

```scala
//> using scala 2.13.14
//> using dep "org.scala-js::scalajs-linker-interface:1.19.0"

import java.io.File
import org.scalajs.linker.interface.{StandardConfig, Semantics, ModuleKind, ModuleSplitStyle, OutputDirectory}

object DirectProof {
  case class LinkerInput(dest: Either[File, OutputDirectory], config: StandardConfig)

  def buildConfig(): StandardConfig =
    StandardConfig()
      .withSemantics(Semantics.Defaults)
      .withOptimizer(true)
      .withSourceMap(true)
      .withModuleKind(ModuleKind.NoModule)
      .withModuleSplitStyle(ModuleSplitStyle.FewestModules)

    val dest = Left(new File("/tmp/repro-out"))
    val input1 = LinkerInput(dest, buildConfig())
    val input2 = LinkerInput(dest, buildConfig())
    println(s"input1 == input2                : ${input1 == input2}")
    println(s"input1.config eq input2.config  : ${input1.config eq input2.config}")
    println(s"fingerprints equal              : ${StandardConfig.fingerprint(input1.config) == StandardConfig.fingerprint(input2.config)}")
  }
}
```

Output:
```
input1 == input2                : false
input1.config eq input2.config  : false
fingerprints equal              : true
```

</details>

nguyenyou:
Fixed https://github.com/com-lihaoyi/mill/issues/7016

## How was this tested

I tested it manually with this minimal repro. `reused`, and `invalidated` show if the cache hit.

<details>
<summary><b>Minimal reproduction</b></summary>

`build.mill`:
```scala
package build
import mill.*, scalalib.*, scalajslib.*

object app extends ScalaJSModule {
  def scalaVersion = "3.3.4"
  def scalaJSVersion = "1.19.0"
}
```

`app/src/Main.scala`:
```scala
object Main {
    val xs = (1 to 100).toList
    val mapped = xs.map(_ * 2).filter(_ % 3 == 0)
    val grouped = mapped.groupBy(_ % 5).view.mapValues(_.sum).toMap
    println(s"sum = ${mapped.sum}, groups = $grouped")
  }
}
```

Run `fastLinkJS` three times, mutating the source between runs so the outer Mill task cache doesn't short-circuit and hide the inner-cache bug:

```bash
./mill clean app
./mill app.fastLinkJS 2>&1 | grep -E "Batch mode|cache stats|Optimizing"
printf '\n// bump 1\n' >> app/src/Main.scala
./mill app.fastLinkJS 2>&1 | grep -E "Batch mode|cache stats|Optimizing"
printf '\n// bump 2\n' >> app/src/Main.scala
./mill app.fastLinkJS 2>&1 | grep -E "Batch mode|cache stats|Optimizing"
```

**Actual** — three byte-identical runs:
```
Optimizer: Batch mode: true
Optimizer: Optimizing 2933 methods.
Emitter: Class tree cache stats: reused: 0 -- invalidated: 283
Emitter: Method tree cache stats: reused: 0 -- invalidated: 907
```

**Expected** — runs 2 and 3 should show `reused: 282 -- invalidated: 1` (only the one class whose source changed).

</details>

## ❌ Mill 1.1.5

Screenshot 2026-04-16 at 06 49 58

## ✅ Mill 1.1.2

Screenshot 2026-04-16 at 06 51 53

## ✅ Mill 1.1.5 with the fix

Screenshot 2026-04-16 at 07 12 44

### DIFF
--- a/libs/scalajslib/worker/1/src/mill/scalajslib/worker/ScalaJSWorkerImpl.scala
+++ b/libs/scalajslib/worker/1/src/mill/scalajslib/worker/ScalaJSWorkerImpl.scala
@@ -38,8 +38,8 @@ class ScalaJSWorkerImpl(jobs: Int) extends ScalaJSWorkerApi with ScalaJSConfigWo
     override def equals(other: Any): Boolean = other match {
       case that: LinkerInput =>
         dest == that.dest &&
-          sjs.StandardConfig.fingerprint(config) ==
-            sjs.StandardConfig.fingerprint(that.config)
+        sjs.StandardConfig.fingerprint(config) ==
+          sjs.StandardConfig.fingerprint(that.config)
       case _ => false
     }
     override def hashCode(): Int =
--- a/libs/scalajslib/worker/1/src/mill/scalajslib/worker/ScalaJSWorkerImpl.scala
+++ b/libs/scalajslib/worker/1/src/mill/scalajslib/worker/ScalaJSWorkerImpl.scala
@@ -28,7 +28,23 @@ class ScalaJSWorkerImpl(jobs: Int) extends ScalaJSWorkerApi with ScalaJSConfigWo
   private case class LinkerInput(
       dest: Either[File, sjs.OutputDirectory],
       config: sjs.StandardConfig
-  )
+  ) {
+    // sjs.StandardConfig is a `final class` with reference equality, so the
+    // auto-generated case-class equals would never match across fastLinkJS
+    // invocations (each invocation builds a fresh StandardConfig via
+    // ScalaJSConfig.config(...)). That made the ScalaJSLinker CachedFactory
+    // miss every time, discarding the optimizer/emitter incremental state.
+    // Use StandardConfig.fingerprint for a structural comparison instead.
+    override def equals(other: Any): Boolean = other match {
+      case that: LinkerInput =>
+        dest == that.dest &&
+          sjs.StandardConfig.fingerprint(config) ==
+            sjs.StandardConfig.fingerprint(that.config)
+      case _ => false
+    }
+    override def hashCode(): Int =
+      (dest, sjs.StandardConfig.fingerprint(config)).hashCode
+  }
   private def minorIsGreaterThanOrEqual(number: Int) = ScalaJSVersions.current match {
     case s"1.$n.$_" if n.toIntOption.exists(_ < number) => false
     case _ => true
--- a/libs/scalajslib/worker/1/src/mill/scalajslib/worker/ScalaJSWorkerImpl.scala
+++ b/libs/scalajslib/worker/1/src/mill/scalajslib/worker/ScalaJSWorkerImpl.scala
@@ -35,15 +35,15 @@ class ScalaJSWorkerImpl(jobs: Int) extends ScalaJSWorkerApi with ScalaJSConfigWo
     // ScalaJSConfig.config(...)). That made the ScalaJSLinker CachedFactory
     // miss every time, discarding the optimizer/emitter incremental state.
     // Use StandardConfig.fingerprint for a structural comparison instead.
+    // The fingerprint is cached so CachedFactory lookups (hashCode + equals
+    // on each candidate) don't rebuild the string repeatedly.
+    private lazy val configFingerprint: String = sjs.StandardConfig.fingerprint(config)
     override def equals(other: Any): Boolean = other match {
       case that: LinkerInput =>
-        dest == that.dest &&
-        sjs.StandardConfig.fingerprint(config) ==
-          sjs.StandardConfig.fingerprint(that.config)
+        dest == that.dest && configFingerprint == that.configFingerprint
       case _ => false
     }
-    override def hashCode(): Int =
-      (dest, sjs.StandardConfig.fingerprint(config)).hashCode
+    override def hashCode(): Int = (dest, configFingerprint).hashCode
   }
   private def minorIsGreaterThanOrEqual(number: Int) = ScalaJSVersions.current match {
     case s"1.$n.$_" if n.toIntOption.exists(_ < number) => false


lihaoyi:
How was this tested? This should be included in the PR description for future reference

nguyenyou:
@lihaoyi Just updated the PR description with how it was tested.

libs/scalajslib/worker/1/src/mill/scalajslib/worker/ScalaJSWorkerImpl.scala
```diff
@@ -28,7 +28,23 @@ class ScalaJSWorkerImpl(jobs: Int) extends ScalaJSWorkerApi with ScalaJSConfigWo
...
+    // ScalaJSConfig.config(...)). That made the ScalaJSLinker CachedFactory
+    // miss every time, discarding the optimizer/emitter incremental state.
+    // Use StandardConfig.fingerprint for a structural comparison instead.
+    override def equals(other: Any): Boolean = other match {
+      case that: LinkerInput =>
+        dest == that.dest &&
+        sjs.StandardConfig.fingerprint(config) ==
+          sjs.StandardConfig.fingerprint(that.config)
+      case _ => false
+    }
```
HollandDM, line 45: Be careful, there is no indicator proving that if fingerprint equals, then the object will be equals
Should be fine to use it for hashCode thou

nguyenyou: Good catch to flag it, but I think it's safe. The `Fingerprint` trait's Scaladoc explicitly documents it as *"an injective one-way serialization representing the object"*, and `StandardConfigFingerprint` builds from every field of `StandardConfig`. So `fingerprint(a) == fingerprint(b)` does imply structural equality by contract.

The Scala.js sbt plugin itself relies on this: it stores `StandardConfig.fingerprint(...)` as a task key and compares it with `!=` across runs to detect config changes ([ScalaJSPluginInternal.scala#L198-L229](https://github.com/scala-js/scala-js/blob/main/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPluginInternal.scala#L198-L229)) — same use case as here.

I did push a follow-up that caches the fingerprint in a `lazy val` on `LinkerInput` so each `CachedFactory` probe (hashCode + equals per candidate) doesn't rebuild the string.

HollandDM: Then it is good to use this

PR #7018, opened 2026-04-15
Restore Offline Support for OWASP Dependency Check

gamlerhart:
From discussion on https://github.com/com-lihaoyi/mill/pull/7012, restore the offline capability for the OWASP Dependency Check

### DIFF
--- a/contrib/owaspdependencycheck/src/mill/contrib/owaspdependencycheck/OwaspDependencyCheckModule.scala
+++ b/contrib/owaspdependencycheck/src/mill/contrib/owaspdependencycheck/OwaspDependencyCheckModule.scala
@@ -1,10 +1,11 @@
 package mill.contrib.owaspdependencycheck
 
+import mainargs.Flag
 import mill.*
 import mill.api.*
 import mill.javalib.*
 
-trait OwaspDependencyCheckModule extends Module {
+trait OwaspDependencyCheckModule extends Module with OfflineSupportModule {
 
   /**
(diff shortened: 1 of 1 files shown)


gamlerhart:
@lefou Added offline support. A Module that extends OwaspDependencyCheckModule will now also download the OWASP depencency scanner when prepareOffline is used on it.

contrib/owaspdependencycheck/src/mill/contrib/owaspdependencycheck/OwaspDependencyCheckModule.scala
```diff
@@ -34,6 +35,10 @@ trait OwaspDependencyCheckModule extends Module {
     def success: Boolean = exitCode == 0
   }
 
+  override def prepareOffline(all: Flag): Command[Seq[PathRef]] = Task.Command {
+    super.prepareOffline(all)() ++ OwaspDependencyCheckWorker.dependencyCheckClasspath()
+  }
+
```
lefou, line 41: While this is essentially what we want, the responsibilities seem to not match perfectly. We should make the worker itself a `OfflineSupportModule` and just trigger its `prepareOffline` task from here.

lefou: I'll apply this is a subsequent PR.

lefou:
Thank you!

PR #7022, opened 2026-04-16
Fix #7015: stabilize Zinc MiniSetup for Scala 3 across TTY transitions

### LINKED ISSUE #7015: Scala 3 incremental compile breaks when alternating clients with different `colored` (TTY) state
## TL;DR

On Scala 3, `ZincWorker` conditionally prepends `-color:never` to `scalacOptions` based on the current client's `colored` flag (`libs/javalib/worker/src/mill/javalib/zinc/ZincWorker.scala:470`). `colored` is per-RPC, so a TTY client and a non-TTY client produce two different `scalacOptions` for the same module. That gets baked into Zinc's `MiniSetup`, and any `MiniSetup` mismatch is a full Zinc invalidation — so alternating a TTY shell and a non-TTY caller (CI, editor "run command", an agent's bash) cold-recompiles the whole module on every source edit.

Mill itself is not wiping `compile.dest`; Zinc is. Mill's `inputsHash` does not include `colored`, so the mismatch is invisible to Mill and sits latent on disk until the next edit forces the compile body to re-run.

## Minimal repro

`build.mill`:

```scala
package build
import mill._, mill.scalalib._
object repro extends ScalaModule { def scalaVersion = "3.8.3" }
```

With any Scala 3 source file under `repro/src/`, starting from a clean `out/`:

1. Ask [Claude Code](https://www.anthropic.com/claude-code) to run `./mill repro.compile` (it executes the command via its `Bash` tool). Full compile. The baked `MiniSetup` now has `-color:never`.
2. Edit one source file.
3. From an **interactive zsh terminal**: `./mill repro.compile` ⇒ **full cold rebuild of every source**, not just the edited one.

## Root cause

`ZincWorker.compileInternal` ([libs/javalib/worker/src/mill/javalib/zinc/ZincWorker.scala:470](https://github.com/com-lihaoyi/mill/blob/a3e22d654db9d251b7b07c4e2a27b39583d60f57/libs/javalib/worker/src/mill/javalib/zinc/ZincWorker.scala#L470-L477)):

```scala
val addColorNeverOption = Option.when(
  !localConfig.logPromptColored &&
    compilers.scalac().scalaInstance().version().startsWith("3.") &&
    !scalacOptions.exists(_.startsWith("-color:"))
) { "-color:never" }
val finalScalacOptions = addColorNeverOption.toSeq ++ scalacOptions
```

`logPromptColored` comes from `log.prompt.colored`, which is recomputed per client RPC from `mainInteractive`/`env` at [runner/daemon/src/mill/daemon/MillMain0.scala](https://github.com/com-lihaoyi/mill/blob/a3e22d654db9d251b7b07c4e2a27b39583d60f57/runner/daemon/src/mill/daemon/MillMain0.scala#L170-L173). The injected option is never fed back into Mill's input hashing — it is observed inside the worker `apply` and written straight into Zinc's `MiniSetup` analysis file. Next call, `MiniSetup` comparison fails, Zinc invalidates everything.

## Workaround

Pin `scalacOptions` in `build.mill` so the condition `!scalacOptions.exists(_.startsWith("-color:"))` is always false:

```scala
def scalacOptions = Seq("-color:never")
```

Verified end-to-end on the minimal repro.

## Suggested upstream fix

Runtime logger state should not leak into compiler `MiniSetup`. Cleanest fix: pre-normalize `scalacOptions` before handing them to Zinc so `-color:*` is excluded from the `MiniSetup` comparison key (but still applied at compile time). Alternatives: always inject `-color:never` unconditionally, or never inject it and handle color purely at log-rendering time.

---

<details>
<summary><b>Environment</b></summary>

- Mill `1.1.5`
- Scala `3.8.3`
- macOS 25 (Apple Silicon)
- Zulu JDK 21.0.10 (downloaded via Coursier by Mill)
- Single Mill background daemon shared across both clients — PID stable, verified via `out/mill-daemon/processId` and `ps`

</details>

<details>
<summary><b>Full repro cycle with observed output</b></summary>

Two clients against the same daemon:

- **Client B** — non-TTY subprocess (I used Claude Code's `Bash` tool; equivalent to a CI runner, an editor "run shell command" action, or a nested script). `mainInteractive = false` ⇒ `colored = false`.
- **Client T** — interactive zsh terminal. `mainInteractive = true` ⇒ `colored = true`.

Starting from a clean `out/` with 5 trivial source files in the module:

| # | Client | Action | Mill output | Notes |
|---|---|---|---|---|
| 1 | B | `./mill repro.compile` | `compiling 5 Scala sources` | Cold start. Bakes `colored=false` `MiniSetup` (with `-color:never`) into `out/repro/compile.dest/zinc`. |
| 2 | — | edit one line | — | |
| 3 | B | `./mill repro.compile` | `compiling 1 Scala source` | Same TTY state as step 1. `MiniSetup` matches, true incremental. |
| 4 | T | `./mill repro.compile` (no edit) | *(no `compiling …` line — task body skipped)* | **Smoking gun.** Mill's `inputsHash` is unchanged, so the compile task body is not re-run. The `colored` mismatch sits latent on disk, invisible to Mill. |
| 5 | — | edit one line | — | |
| 6 | T | `./mill repro.compile` | `compiling 5 Scala sources` | Edit forces body to re-run. Worker now sees `colored=true`, builds `scalacOptions` *without* `-color:never`, hands them to Zinc; Zinc compares against stored `MiniSetup` (with `-color:never`), declares mismatch, cold-recompiles all 5. Every `.class`, `.tasty`, and the `zinc` analysis file get new inodes, but bytes for unchanged files are byte-identical. |

Step 4 is the part that distinguishes this from a Mill-cache-invalidation story: Mill's input layer is genuinely happy across clients. The corruption is purely in the on-disk Zinc analysis and stays latent until the next edit.

</details>

<details>
<summary><b>Fix verification (same cycle, with workaround)</b></summary>

After adding `def scalacOptions = Seq("-color:never")` and clearing `out/`:

| # | Client | Action | Mill output |
|---|---|---|---|
| 1 | B | `./mill repro.compile` | `compiling 5 Scala sources` |
| 2 | — | edit one line | — |
| 3 | B | `./mill repro.compile` | `compiling 1 Scala source` |
| 4 | T | `./mill repro.compile` (no edit) | *(no-op cache hit)* |
| 5 | — | edit one line | — |
| 6 | T | `./mill repro.compile` | **`compiling 1 Scala source`** |

Step 6 is the key: with `scalacOptions` pinned, the `ZincWorker.scala:474` conditional is bypassed for both clients, `MiniSetup` is identical regardless of `colored`, and Zinc's incremental analysis stays valid across the TTY transition.

</details>

<details>
<summary><b>Evidence it's Zinc invalidation, not a Mill <code>compile.dest</code> wipe</b></summary>

- `compile` is declared `Task(persistent = true)` at `libs/scalalib/src/mill/scalalib/ScalaModule.scala:329`, so its dest should be preserved across runs.
- The only `os.remove.all(paths.dest)` call in `core/exec` / `core/eval` / `runner` is at `core/exec/src/mill/exec/GroupExecution.scala:311`, gated by `!labelled.persistent`. Mill never deletes `compile.dest` for a persistent task.
- After the cross-TTY run, inodes of every `compile.dest/classes/**/*.class` and `*.tasty` change, but for unchanged sources the new bytes are byte-identical to the old bytes (deterministic Scala 3 output). That is consistent with Zinc rewriting the analysis via temp-file + atomic rename (`ConsistentFileAnalysisStore`), not with Mill wiping the directory.
- `out/mill-profile.json` for the cross-TTY run shows `valueHashChanged: true` only on the edit-induced chain `repro.sources → repro.allSources → repro.allSourceFiles → repro.compile`. Every other upstream task, including the `T.input` ones (`coursierEnv`, `useFileLocks`, `zincLogDebug`, `zincReportCachedProblems`, `compileResources`), reports `valueHashChanged: false`. Mill's input graph is clean.
- Same daemon PID across both clients, verified via `out/mill-daemon/processId` and `ps`.

</details>

<details>
<summary><b>Extended root cause walk-through</b></summary>

Mill recomputes `colored` on every client RPC at `runner/daemon/src/mill/daemon/MillMain0.scala:168`:

```scala
val noColorViaEnv    = env.get("NO_COLOR").exists(_.nonEmpty)
val forceColorViaEnv = env.get("FORCE_COLOR").exists(_.nonEmpty)
val colored = config.color.getOrElse(
  (mainInteractive || forceColorViaEnv) && !(noColorViaEnv || config.bsp.value)
)
```

`mainInteractive` and `env` are per-client RPC fields filled from `init.interactive` and `init.env` at `runner/server/src/mill/server/MillDaemonServer.scala:147,149`. Interactive zsh ⇒ `colored = true`; non-TTY subprocess ⇒ `colored = false`.

`colored` flows into `PromptLogger`, then `log.prompt.colored`, then `JvmWorkerImpl` builds a fresh `LocalConfig` per call at `libs/javalib/worker/src/mill/javalib/worker/JvmWorkerImpl.scala:31`:

```scala
val zincCtx = ZincWorker.LocalConfig(
  dest             = ctx.dest,
  logDebugEnabled  = log.debugEnabled,
  logPromptColored = log.prompt.colored,   // per-client TTY state
  workspaceRoot    = mill.api.BuildCtx.workspaceRoot
)
```

Then `ZincWorker.compileInternal` builds `finalScalacOptions` with the conditional `-color:never` (see snippet in the root-cause section above) and passes them into Zinc's `incrementalCompiler.inputs(...)`, which bakes them into `MiniSetup` and persists them in the analysis store at `Task.dest/zinc`. On the next compile, sbt-zinc compares old vs new `MiniSetup`; `scalacOptions` mismatch is a standard full-invalidation trigger, so every source is cold-recompiled. `ConsistentFileAnalysisStore` writes the analysis via temp-file + atomic rename, so even byte-identical analyses get a new inode.

`-color:never` is injected inside the worker `apply`, after Mill has finished computing `inputsHash`, and is never fed back into Mill's input graph. So `inputsHash` / `valueHash` for every upstream task is identical across clients, and Mill's cache layer never knows anything is wrong. The only thing that triggers re-running the compile body is the legitimate source edit — but once it re-runs, Zinc decides the `MiniSetup` is incompatible and goes cold.

</details>

<details>
<summary><b>Related code references</b></summary>

All paths relative to `com-lihaoyi/mill` at tag `1.1.5`:

- `libs/javalib/worker/src/mill/javalib/zinc/ZincWorker.scala:470` — Scala-3-only `-color:never` injection **(root cause)**
- `libs/javalib/worker/src/mill/javalib/worker/JvmWorkerImpl.scala:31` — `LocalConfig.logPromptColored` plumbed in
- `runner/daemon/src/mill/daemon/MillMain0.scala:168` — per-client `colored` resolution
- `runner/server/src/mill/server/MillDaemonServer.scala:147` — `mainInteractive` / `env` populated from RPC `init`
- `libs/scalalib/src/mill/scalalib/ScalaModule.scala:329` — `def compile = Task(persistent = true) { ... }`
- `core/exec/src/mill/exec/GroupExecution.scala:311` — only `os.remove.all(paths.dest)` site, gated by `!labelled.persistent`

</details>

nguyenyou:
Fixes #7015.

1. **Test** — repro on `main`: compile `--color=false`, edit, recompile `--color=true`, assert unchanged `.class` not rewritten.
2. **Fix** — drop `!localConfig.logPromptColored &&` so `-color:never` is injected unconditionally for Scala 3.

## Update

Current fix regresses color. `integration.dedicated.full-run-logs` fails: `-color:never` strips ANSI from Scala 3 errors even when user passes `--color=true`. Needs a proper fix (filter `-color:*` from `MiniSetup` hash, or move color to reporter layer). Leaving direction to @lihaoyi.

### DIFF
--- a/integration/feature/zinc-incremental-scala3-color/resources/app/src/App.scala
+++ b/integration/feature/zinc-incremental-scala3-color/resources/app/src/App.scala
@@ -1,0 +1,7 @@
+package app
+
+object App {
+  def main(args: Array[String]): Unit = {
+    println(models.Foo("hello").greet)
+  }
+}
--- a/integration/feature/zinc-incremental-scala3-color/resources/app/src/models/Foo.scala
+++ b/integration/feature/zinc-incremental-scala3-color/resources/app/src/models/Foo.scala
@@ -1,0 +1,5 @@
+package models
+
+case class Foo(name: String) {
+  def greet: String = s"hello, $name"
+}
--- a/integration/feature/zinc-incremental-scala3-color/resources/build.mill
+++ b/integration/feature/zinc-incremental-scala3-color/resources/build.mill
@@ -1,0 +1,9 @@
+package build
+// Issue https://github.com/com-lihaoyi/mill/issues/7015
+import mill.*
+import mill.scalalib.*
+
+object app extends ScalaModule {
+  def scalaVersion = "3.8.3"
+  def zincIncrementalCompilation = true
+}
--- a/integration/feature/zinc-incremental-scala3-color/src/ZincIncrementalScala3ColorTests.scala
+++ b/integration/feature/zinc-incremental-scala3-color/src/ZincIncrementalScala3ColorTests.scala
@@ -1,0 +1,64 @@
+package mill.integration
+
+import mill.testkit.UtestIntegrationTestSuite
+
+import utest.*
+
+// Regression test for issue https://github.com/com-lihaoyi/mill/issues/7015
+//
+// On Scala 3, ZincWorker conditionally prepends `-color:never` to scalacOptions
+// based on the current client's `colored` flag. That option is baked into
+// Zinc's MiniSetup, so alternating clients with different TTY state (e.g. a
+// non-TTY CI run followed by an interactive shell) makes the MiniSetup
+// comparison fail on the next edit and cold-rebuilds every source in the
+// module. Here we simulate the two clients with `--color=false` / `--color=true`
+// and assert that sources which were NOT edited are not recompiled.
+object ZincIncrementalScala3ColorTests extends UtestIntegrationTestSuite {
+  val tests: Tests = Tests {
+    test("colorFlagPreservesIncremental") - integrationTest { tester =>
+      import tester.*
+
+      val appSrc = workspacePath / "app/src/App.scala"
+      val classes = workspacePath / "out/app/compile.dest/classes"
+      val appClass = classes / "app/App.class"
+      val fooClass = classes / "models/Foo.class"
+
+      // Client B (non-TTY): cold compile with colored=false.
+      // MiniSetup gets baked with `-color:never` in scalacOptions.
+      val firstRun = tester.eval(("--color=false", "app.compile"))
+      assert(firstRun.isSuccess)
+      assert(Seq(classes, appClass, fooClass, appSrc).forall(os.exists))
+
+      val appClassInfo1 = os.stat(appClass)
+      val fooClassInfo1 = os.stat(fooClass)
+
+      // Edit only App.scala. Foo.scala is untouched.
+      os.write.append(appSrc, "\n ")
+
+      // Client T (TTY): compile with colored=true.
+      // Worker builds scalacOptions WITHOUT `-color:never`, hands them to
+      // Zinc. Zinc compares against stored MiniSetup (with `-color:never`),
+      // declares a mismatch, and cold-recompiles every source — including
+      // Foo.scala, which was not edited.
+      val secondRun = tester.eval(("--color=true", "app.compile"))
+      assert(secondRun.isSuccess)
+
+      val appClassInfo2 = os.stat(appClass)
+      val fooClassInfo2 = os.stat(fooClass)
+
+      // App.scala was edited, so its .class is expected to be rewritten.
+      // We use mtime instead of ctime because Scala 3 compile is fast enough
+      // that both runs can land inside the same wall-clock second, and
+      // os-lib's ctime formatter is second-granularity; mtime has nanosecond
+      // precision.
+      assert(appClassInfo1.mtime != appClassInfo2.mtime)
+
+      // Foo.scala was NOT edited. With a correct incremental compile this
+      // .class should be untouched. This assertion currently FAILS and is
+      // the reproduction of issue #7015 — Zinc cold-rebuilds Foo.class
+      // because of the MiniSetup mismatch caused by the `-color:never`
+      // injection being tied to per-client TTY state.
+      assert(fooClassInfo1.mtime == fooClassInfo2.mtime)
+    }
+  }
+}
--- a/libs/javalib/worker/src/mill/javalib/zinc/ZincWorker.scala
+++ b/libs/javalib/worker/src/mill/javalib/zinc/ZincWorker.scala
@@ -467,9 +467,12 @@ class ZincWorker(jobs: Int, useFileLocks: Boolean = false) extends AutoCloseable
       }
     }
 
+    // Inject `-color:never` unconditionally for Scala 3 so the stored Zinc
+    // MiniSetup stays stable across clients with different TTY state. Making
+    // this depend on per-RPC `logPromptColored` caused full cold rebuilds on
+    // every TTY transition — see https://github.com/com-lihaoyi/mill/issues/7015
     val addColorNeverOption = Option.when(
-      !localConfig.logPromptColored &&
-        compilers.scalac().scalaInstance().version().startsWith("3.") &&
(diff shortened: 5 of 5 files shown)


nguyenyou:
@lihaoyi this PR is mostly for the failing regression test — curious what you think about it and the direction of the fix.

On our side we've already applied the `def scalacOptions = Seq("-color:never")` workaround from the issue, so we're unblocked. Flagging mainly because the bug is hard to notice — from experience, you just edit one file, Mill recompiles a bunch, and it leaves the wrong impression that Mill's caching is flaky. Other Scala 3 users alternating TTY / non-TTY clients against the same daemon are probably hitting this too without realizing the cause.

lihaoyi:
Your PR description says this is not the correct fix, so closing this

PR #7023, opened 2026-04-17
Make `OwaspDependencyCheckModule` worker configurable

lefou:
Since the worker is bound to a specific version, it should not be hardcoded but user configurable.

The default version of the `org.owasp.dependency-check-cli` artifact is now sourced from the main build and can be scanned for updates (like all other tool versions used by Mill).


gamlerhart:
Thanks.

PR #7034, opened 2026-04-24
Android: Only have `AndroidLibModule` for libraries

souvlakias:
Motivation: see https://github.com/com-lihaoyi/mill/issues/6908#issuecomment-4303507575

Reviewing our current implementation, realized that there is no need for us to have different variants for 
`AndroidLibModule`, neither an `AndroidLibTests`.
  - Specifically, the old `AndroidLibTests` was missing important required overrides (e.g. test module BSP location), so having users use the expected `AndroidTestModule` also fixes the developer experience.

### Changes
- Remove`AndroidLibKotlinModule`
- Update existing examples and add a Scala lib example
- Add a doc section about publishing (that includes the Kotlin lib example)


example/androidlib/scala/2-sum-lib-scala/lib/src/test/scala/com/example/lib/SumUnitTest.scala
```diff
@@ -0,0 +1,11 @@
+package com.example.lib
+
+import org.junit.Test
+import org.junit.Assert._
+
+class SumUnitTest {
```
vaslabs: I wonder what the android scala community uses most, if it's scalatest, we can use that instead of junit

souvlakias: There aren't many examples of scala android apps so I'm not sure what the preferred framework is.
But yes we can have this lib example use scalatest (or utest) to showcase the usage.

souvlakias: Update: Made the unit tests run on `utest`. What do you think?

souvlakias:
I realised that some inconsistencies came from `AndroidLibTests` not extending `AndroidTestModule`. 
By extending that, there isn't really any point left for us having the different traits afterall, since there are no overrides needed.

Thus, I think, we could just keep the base `AndroidLibModule` and completely drop the other variants (e.g. `AndroidLibKotlinModule`.
Usage would be the expected:
```scala
object lib extends AndroidKotlinModule, AndroidLibModule {
  object test extends AndroidKotlinTestModule {}
}
```
(See: https://github.com/vaslabs-ltd/mill/commit/e94133646d37a8a231cdedd9b6056bbdd0ab5e6b)

vaslabs:
> I realised that some inconsistencies came from `AndroidLibTests` not extending `AndroidTestModule`. By extending that, there isn't really any point left for us having the different traits afterall, since there are no overrides needed.
> 
> Thus, I think, we could just keep the base `AndroidLibModule` and completely drop the other variants (e.g. `AndroidLibKotlinModule`. Usage would be the expected:
> 
> ```scala
> object lib extends AndroidKotlinModule, AndroidLibModule {
>   object test extends AndroidKotlinTestModule {}
> }
> ```
> 
> (See: [vaslabs-ltd@e941336](https://github.com/vaslabs-ltd/mill/commit/e94133646d37a8a231cdedd9b6056bbdd0ab5e6b))

sounds good, if it simplifies things and doesn't break anything, let's do it

vaslabs (approved):

example/androidlib/scala/2-sum-lib-scala/lib/src/test/scala/com/example/lib/SumUnitTest.scala
lefou: Not an Android dev, but regarding test frameworks, I'd prefer scalatest due to being more flexible over utest when writing test matrices.

souvlakias: Updated to scalatest

lefou (approved):
Looks good to me.

PR #7036, opened 2026-04-24
Fix some `Task`s not returning `Task.Simple`

souvlakias:
Fixes https://github.com/com-lihaoyi/mill/issues/6979#

Some `Task`s were accidentally returning `Task` instead of `Task.Simple`, which caused the following:
- Can't be called by cli
- Broke `visualizePlan` for such tasks that depended on their `super`s (see: https://github.com/com-lihaoyi/mill/issues/6979#issuecomment-4312458654)

### DIFF
--- a/example/thirdparty/android-compose-samples/build.mill
+++ b/example/thirdparty/android-compose-samples/build.mill
@@ -83,7 +83,7 @@ object JetLagged extends mill.api.Module {
     )
 
     def androidEnableCompose = true
-    override def kotlinUseEmbeddableCompiler: Task[Boolean] = Task { true }
+    override def kotlinUseEmbeddableCompiler: T[Boolean] = Task { true }
 
     def androidSdkModule = mill.api.ModuleRef(androidSdkModule0)
 
@@ -101,7 +101,7 @@ object JetLagged extends mill.api.Module {
       override def androidEnableCompose = true
 
       // TODO consider defaulting this to the parent app value
-      override def kotlinUseEmbeddableCompiler: Task[Boolean] = Task {
+      override def kotlinUseEmbeddableCompiler: T[Boolean] = Task {
         true
       }
 
@@ -193,7 +193,7 @@ object JetNews extends mill.api.Module {
     )
 
     def androidEnableCompose = true
-    override def kotlinUseEmbeddableCompiler: Task[Boolean] = Task { true }
+    override def kotlinUseEmbeddableCompiler: T[Boolean] = Task { true }
 
     def androidSdkModule = mill.api.ModuleRef(androidSdkModule0)
 
@@ -212,7 +212,7 @@ object JetNews extends mill.api.Module {
       override def androidEnableCompose = true
 
       // TODO consider defaulting this to the parent app value
-      override def kotlinUseEmbeddableCompiler: Task[Boolean] = Task {
+      override def kotlinUseEmbeddableCompiler: T[Boolean] = Task {
         true
       }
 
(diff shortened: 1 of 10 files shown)


souvlakias:
I noticed that [this test](https://github.com/com-lihaoyi/mill/blob/main/core/api/test/src/mill/api/MacroErrorTests.scala#L48) exists, so I guess this ascription may have an intentional usecase afterall. 
I went ahead and only fixed tasks in code I was familiar with, so I wouldn't break anything

Some others still remain:
```bash
$ grep -RInP '((?:final\s+)?(?:private|protected)?\s*def\s+\w+\s*(?:\([^)]*\))?\s*:\s*)Task\[((?:[^\]]|\](?!\s*=))*)\]\s*=\s*Task\s*\{' . --exclude-dir out

./mill-build/src/millbuild/MillPublishJavaModule.scala:101:  override def publishLocalTestRepo: Task[PathRef] = Task {
./mill-build/src/millbuild/MillJavaModule.scala:87:  def ciJavacOptions: Task[Seq[String]] = Task {
./contrib/sbom/src/mill/contrib/sbom/CycloneDXJavaModule.scala:21:  def sbomComponents: Task[Seq[Component]] = Task {
./libs/javalib/src/mill/javalib/PublishModule.scala:624:  def publishLocalTestRepo: Task[PathRef] = Task {
./libs/javalib/src/mill/javalib/spotless/SpotlessModule.scala:63:  def spotlessFormats: Task[Seq[Format]] = Task {
./libs/javascriptlib/src/mill/javascriptlib/TestModule.scala:601:    private def mkConfig: Task[TestResult] = Task {
./core/api/test/src/mill/api/MacroErrorTests.scala:53:            def x(): Task[Int] = Task {1}
./core/exec/test/src/mill/exec/ExecutionTests.scala:59:        def taskA: Task[String] = Task { taskB() }
./core/exec/test/src/mill/exec/ExecutionTests.scala:61:      def taskB: Task[String] = Task { taskC() }
./core/exec/test/src/mill/exec/ExecutionTests.scala:63:    def taskC: Task[String] = Task { module.nested.taskA() }
./example/extending/jvmcode/1-subprocess/build.mill:22:  def groovyClasspath: Task[Seq[PathRef]] = Task {
./example/extending/jvmcode/2-classloader/build.mill:15:  def groovyClasspath: Task[Seq[PathRef]] = Task {
./example/extending/jvmcode/3-worker/build.mill:20:def groovyClasspath: Task[Seq[PathRef]] = Task {
```

lefou (approved):

lefou:
Looks good to me. Don't know about the test. I typically leave a code-comment for a later dev/me when I have no idea what some code is about.

souvlakias:
@lefou are the binary issues for KotlinModule's `kotlinUseEmbeddableCompiler` okay for this, or should I remove the change so this PR just changes AndroidLib ?

Or maybe deprecate it and have a new task?

lefou:
> @lefou are the binary issues for KotlinModule's `kotlinUseEmbeddableCompiler` okay for this, or should I remove the change so this PR just changes AndroidLib ?
> 
> Or maybe deprecate it and have a new task?

We probably need to revert the offending changes, unfortunately. To work around the `cli`-issues (which should be addressed properly nevertheless) we should try to fix the rhs, e.g. by changing `Task` to `Task.Anon`.

If Kotlin would be a dedicated plugin, we could just merge and bump the major version, but that's unfortunately not possible being so tightly coupled to Mill releases.

This is one of the reasons, why I don't like that Mill's support for various tools and language is directly coupled to Mill releases.

lefou:
When reverting, please leave a in-code `TODO` comment for maintainers.

souvlakias:
Closing since
- #6955 

includes these fixes (and the KotlinModule ones), and fixed
- #6979 :)

PR #7047, opened 2026-04-28
Update semanticdb-scalac to 4.16.1

lefou:


lefou:
@alexarchambault I think you added the `SemanticdbProcessor`. There seems to be a unhandled case due to the semanticdb API update. Do you have an idea what's needed here?

alexarchambault:
@lefou I think the commit I pushed should do

lefou:
Thank you @alexarchambault !

PR #7051, opened 2026-04-28
Fix `_` and `__` wildcard support for pre-compiled modules

lihaoyi:


lihaoyi:
Failure is an infra issue, merging it

PR #7057, opened 2026-04-30
Optmize performance for cylic check

jilen:
List.contains is slow on large number of modules
```
benchmark                  (depCount)  Mode  Cnt        Score        Error  Units
ModuleUtilsJmh.newSetList          10  avgt   20      809.918 ±    164.825  ns/op
ModuleUtilsJmh.newSetList         100  avgt   20     8839.895 ±    852.083  ns/op
ModuleUtilsJmh.newSetList        1000  avgt   20   113960.451 ±  14368.900  ns/op
ModuleUtilsJmh.oldList             10  avgt   20      530.315 ±     44.856  ns/op
ModuleUtilsJmh.oldList            100  avgt   20    33012.497 ±   2543.791  ns/op
ModuleUtilsJmh.oldList           1000  avgt   20  3065535.100 ± 230869.212  ns/op
```
https://gist.github.com/jilen/b9dce075d3c0ce9f2fb4a16838fc4e40

### DIFF
--- a/libs/javalib/src/mill/javalib/internal/ModuleUtils.scala
+++ b/libs/javalib/src/mill/javalib/internal/ModuleUtils.scala
@@ -21,14 +21,15 @@ object ModuleUtils {
   def recursive[T](name: String, start: T, deps: T => Seq[T]): Seq[T] = {
 
     @tailrec def rec(
-        seenModules: List[T],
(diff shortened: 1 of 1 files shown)


lefou:
The use of a `LinkedHashSet` should do the same but avoid the second data structure.

PR #7059, opened 2026-05-01
Attempt to fix inactive annotation processor detection

### LINKED ISSUE #7058: Inactive annotation processors can disable Zinc incremental compilation
Hi,

TLDR: after upgrading to Mill `1.1.5-44-d4c442`, many modules started warning that Zinc incremental compilation was disabled because annotation processors lack incremental metadata. It looks like inactive javac processors on the compile classpath can be treated as active, especially for mixed Scala/Java modules and Scala-only modules.

Warning:

```text
Disabling zinc incremental compilation because annotation processors do not declare supported
incremental metadata.
Mill only reads incremental annotation processor metadata from the active processor path
and the compile classpath.
To enable incremental annotation processing, use processors that publish
META-INF/gradle/incremental.annotation.processors, or add a patched jar/directory or
compile-time classpath entry with that metadata.
```

This did not happen for us before the recent incremental annotation processing changes in #6999.

From debugging, I think the issue is:

1. When no explicit `-processorpath` is set, Mill uses the compile classpath as the processor path.
2. Mill discovers processors from `META-INF/services/javax.annotation.processing.Processor`.
3. For mixed Scala/Java modules, the source annotation index appears to try parsing `.scala` files with javac.
4. If indexing fails, all discovered processors are treated as active.
5. For Scala-only modules, javac processors should not be relevant, but processors on the compile classpath can still trigger the warning.

I plan to send a PR with regression tests for:

- mixed Java/Scala sources with an unrelated processor on the compile classpath
- Scala-only sources with a processor on the compile classpath

Opening the issue first because there may be multiple reasonable fixes. The local fix I tested was to skip AP detection when there are no Java sources, and only pass `.java` files into the annotation index.

Thanks.

### LINKED ISSUE #7063: Scala-only edits can full-recompile mixed modules with javac AP tracking
TLDR: Mill’s incremental javac annotation-processing tracker appears to snapshot Scala sources too. In mixed Scala/Java modules with an aggregating/unknown AP output, an unrelated Scala-only edit can make Mill ignore Zinc’s previous analysis and full-recompile the module.

Symptom:

```text
mixed Scala/Java module
+ active javac annotation processor with incremental metadata
+ AP-generated output with aggregating/unknown ownership
+ edit unrelated .scala file
=> whole module recompiles
```

Likely cause:

```text
Scala edit
  -> AP source snapshot changes
  -> sourceSetDidChange = true
  -> aggregating/Unknown AP product exists
  -> requiresFullRecompile = true
  -> Zinc previous analysis ignored
```

Suggested fix: javac AP source stamps / previous-stamp comparison / AP ownership tracker should use Java sources only, since javac annotation processors do not process Scala sources.

nguyenyou:
Fixes #7058.
Fixes #7063.

This PR keeps normal Zinc incremental compilation enabled when javac annotation processors are present but unsupported by Mill's AP tracker.

Changes:

- Missing javac AP incremental metadata disables only Mill AP tracking, not Zinc incremental compilation.
- Inactive processor detection and javac AP tracking consider Java sources only.

#7063:

- Scala-only edits in mixed Scala/Java modules should not invalidate javac AP outputs or force full-module recompiles.

Validation:

```bash
./mill libs.javalib.worker.test.testOnly mill.javalib.zinc.IncrementalAnnotationProcessingDetectionTests
```

### DIFF
--- a/libs/javalib/worker/src/mill/javalib/zinc/IncrementalAnnotationProcessing.scala
+++ b/libs/javalib/worker/src/mill/javalib/zinc/IncrementalAnnotationProcessing.scala
@@ -129,9 +129,8 @@ private[mill] object IncrementalAnnotationProcessing {
       log: Logger.Actions
   ): Mode = {
     val javaSources = sources.filter(_.ext == "java")
-    if (
-      !incrementalCompilation || javacOptions.contains("-proc:none") || javaSources.isEmpty
-    ) Mode.None
+    if (!incrementalCompilation || javacOptions.contains("-proc:none") || javaSources.isEmpty)
+      Mode.None
     else {
       val processorPath = parsePathOption(javacOptions, "-processorpath", "--processor-path")
         .map(_.map(os.Path(_, os.pwd)))
--- a/libs/javalib/worker/src/mill/javalib/zinc/IncrementalAnnotationProcessing.scala
+++ b/libs/javalib/worker/src/mill/javalib/zinc/IncrementalAnnotationProcessing.scala
@@ -128,10 +128,9 @@ private[mill] object IncrementalAnnotationProcessing {
       incrementalCompilation: Boolean,
       log: Logger.Actions
   ): Mode = {
+    val javaSources = sources.filter(_.ext == "java")
     if (
-      !incrementalCompilation || javacOptions.contains("-proc:none") || !sources.exists(
-        _.ext == "java"
-      )
+      !incrementalCompilation || javacOptions.contains("-proc:none") || javaSources.isEmpty
     ) Mode.None
     else {
       val processorPath = parsePathOption(javacOptions, "-processorpath", "--processor-path")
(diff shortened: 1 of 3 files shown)


nguyenyou:
Checked the two failed jobs. `SpotlessTests.nestedModule` passes locally with `./mill libs.scalalib.test.testOnly mill.scalalib.spotless.SpotlessTests`. The `example.scalalib.script.2-json-api-client` failure is a Wikipedia HTTP 429. Looks flaky/external rather than caused by this PR.

lihaoyi:
@nguyenyou I think let's just remove the `Disabling zinc incremental compilation because annotation processors do not declare supported incremental metadata.` logic entirely. Having things on the classpath mysteriously turn off incremental compilation is confusing, and anyway people can manually disable it with `zincIncrementalCompilation: false` if they need to

nguyenyou:
@lihaoyi  Thanks, followed your guidance and removed the logic that disables Zinc incremental compilation when annotation processors do not declare supported incremental metadata.

I also found a separate related bug and filed #7063. In mixed Scala/Java modules with javac AP tracking enabled, a Scala-only edit could still force a full-module Zinc recompile because the AP tracker was snapshotting Scala sources too. I added a regression test for that case as well, and fixed the tracker to consider Java sources only.

PR #7069, opened 2026-05-04
Split out `globalExclusive = true` flag from vanilla `exclusive = true`

lihaoyi:
Follow up from https://github.com/com-lihaoyi/mill/pull/6955

`exclusive = true` makes the command exclusive only within this launcher, which is used for commands that you want to run single-threaded per launcher e.g. to avoid interleaving with other tasks in the terminal logs, whereas `globalExclusive = true` takes the global exclusive write lock which is useful for tasks which actually do global side effects that should not be run concurrently even with other launchers.

Most `exclusive = true` tasks remain unchanged, but those like `clean`, `bspClean`, `scalafmt`, `init`, etc. are marked  like `globalExclusive = true` as part of this PR

Also includes some code to make sure we release the locks when doing nested evaluations, so e.g. `show` evaluating tasks  uses the locking strategy of the tasks it evaluates


lihaoyi:
Seems CI mostly passed on my fork https://github.com/lihaoyi/mill-1/actions/runs/25316708936/job/74216990258

PR #7076, opened 2026-05-05
Use Java 25 by default

lihaoyi:
As part of the Mill 1.1.0 minor version bump we increased the default JVM version from 17 to 21, as the current LTS JVM version . As we prepare for Mill 1.2.0, this PR bumps it to 25 which is latest LTS JVM version


lihaoyi:
Seems like a bunch of stuff breaks on 25, might want to hold off on this for a bit more

lefou:
We should aim for flawless Java 25 support in Mill 1.2.0. Maybe, we can add a failing-allowed CI workflow with Java 25 to make issues more visible?

PR #7077, opened 2026-05-05
Simplify readwrite lock blocking message 

lihaoyi:
No need for it to contain the task name in the multi-line prompt, since it appears on the prompt line which already includes the task name. Other scenarios like the BSP logs or when `--ticker false` (e.g. in integration tests) we continue to show the task name


lihaoyi:
CI passed on the fork https://github.com/lihaoyi/mill-1/actions/runs/25371818333/job/74403177834

Issue #7056, opened 2026-04-30
Decoupling of Mill core and libs bundles


lefou:
As of now, Mill releases always contain the core tool as well as a lot of bundled language and tool libraries. While it is very convenient to develop and use, it also limits how fast each part can evolve. Providing some degree of binary compatibility means, we can't adapt specific languages at a different pace as the core runtime, and vice versa.

But essentially, this tight coupling isn't necessary. The Mill launcher, daemon, CLI, and the various integrations like `javalib`, `scalalib` or `kotlinlib`, to name a few, don't need to be coupled that tight. Especially, the various language libs could have a faster release speed when it comes to binary compatibility. We could improve the user experience faster by cutting new minor and major versions and deprecate blocking stuff faster. Since Mill 1.0 release, we have dropped wanted changes multiple times because of binary compatibility issues in the modules, but not in the core APIs of Mill. We also added a lot of cruft and workarounds to keep some compatibility but at the price of clarity and maintainability. Given, that Mill modules are Scala traits and changes to traits often require recompilation for all downstream derives, we are technically limited.

I therefore propose to split up the language modules (Java, Scala, Kotlin, ScalaJS, Scala Native, Groovy) from the Mill core platform (runner, evaluator, cli, launcher, daemon, and the core API including tasks and modules). This means we would have different versioning and binary compatibility guarantees for each.

I think for most of the issues I typically witnessed it's sufficient to keep the various JVM languages in the shared release, so we essentially end up with three release bundles:

* Mill core
* Mill libs
* Mill contrib

Having the freedom to bump Mill libs from e.g. `1.1.5` to `2.0` to apply cleanups and improvements, also gives users the freedom to stay on Mill libs `1.x` for a while and choose the best date for the switch. And in case they need to stay on Mill libs `1.x`, they could still update Mill core to the newest compatible version to  profit from improvements and fixes.

lefou:
So, how should the usability story look like? We want to keep the batteries included experience.

We no longer want to implicitly add the `mill-libs` dependency equivalent to the classpath. Instead, it should be explicitly added to `mvnDeps` of the meta build. By default, we add it with a preconfigured version. but if a user adds `mill-libs` with a newer (yet compatible) version, coursier should reconcile the version to use that instead.

lihaoyi:
[...]

PR #7082, opened 2026-05-05
Test and fix IDEA `*.iml` file generation for Kotlin

viluon:
In Kotlin, a module can only access its own `internal` definitions and `internal` definitions from its "friend" modules, configured by the `-Xfriend-paths=` compiler option.

Previously, `.iml` files generated by `mill mill.idea/` for `KotlinIdeaModule`s were incorrect:
- compilation output ended up somewhere in `internalGenIdea/ideaCompileOutput.dest/classes`, but the `-Xfriend-paths=` compiler option still referenced `compile.dest/classes` directories
- IDEA highlighted attempts to access `internal` definitions from friend modules as errors because the `<additionalVisibleModuleNames>` element was missing from the IML configuration
- the generated modules defaulted to inheriting the project configuration, which made any customisations they introduced irrelevant
- `kotlinExplicitApi` was handled inconsistently and was also missing from the compiler options listed in IML files. I moved it to mandatory compiler options and removed the ad-hoc handling, making the code shorter overall.

This PR fixes these issues by introducing the higher-level `kotlinFriendModules` setting to `KotlinModule`, from which the compiler options and IDEA metadata are generated. It also adds an integration test for `KotlinIdeaModule`s.

- [x] once CI passes mark the PR as `Ready for review` (and CI will run on the main Mill repo before we merge it)

### DIFF
--- a/core/api/daemon/src/mill/api/daemon/internal/IdeUtils.scala
+++ b/core/api/daemon/src/mill/api/daemon/internal/IdeUtils.scala
@@ -1,6 +1,6 @@
-package mill.scalalib.internal
+package mill.api.daemon.internal
 
-import mill.api.daemon.Segments
+import mill.api.daemon.{Segment, Segments}
 import mill.api.*
 
 import java.nio.file.Path
--- a/core/api/daemon/src/mill/api/daemon/internal/IdeUtils.scala
+++ b/core/api/daemon/src/mill/api/daemon/internal/IdeUtils.scala
@@ -3,8 +3,6 @@ package mill.api.daemon.internal
 import mill.api.daemon.{Segment, Segments}
 import mill.api.*
 
-import java.nio.file.Path
-
 object IdeUtils {
 
   /**
@@ -13,21 +11,13 @@ object IdeUtils {
    * @param path If the  segments yield no result, use the Mill Module folder name from this path.
    * @see [[Module.moduleSegments]]
    */
-  def moduleName(p: Segments, path: Path = null): String = {
-    val name = p.value
-      .foldLeft(StringBuilder()) {
-        case (sb, Segment.Label(s)) if sb.isEmpty => sb.append(s)
-        case (sb, Segment.Cross(s)) if sb.isEmpty => sb.append(s.mkString("-"))
-        case (sb, Segment.Label(s)) => sb.append(".").append(s)
-        case (sb, Segment.Cross(s)) => sb.append("-").append(s.mkString("-"))
-      }
-      .mkString
-      .toLowerCase()
-
-    // If for whatever reason no name could be created based on the module segments, create a
-    // generic one from the Mill Module folder name. Users can rename the project inside the
-    // IDE if they like.
-    if (name.isBlank && path != null) path.getFileName.toString
-    else name
-  }
+  def moduleName(p: Segments): Option[String] = Some(p.value
+    .foldLeft(StringBuilder()) {
+      case (sb, Segment.Label(s)) if sb.isEmpty => sb.append(s)
+      case (sb, Segment.Cross(s)) if sb.isEmpty => sb.append(s.mkString("-"))
+      case (sb, Segment.Label(s)) => sb.append(".").append(s)
+      case (sb, Segment.Cross(s)) => sb.append("-").append(s.mkString("-"))
+    }
+    .mkString
+    .toLowerCase()).filterNot(_.isBlank)
 }
--- a/core/api/daemon/src/mill/api/daemon/internal/idea/Element.scala
+++ b/core/api/daemon/src/mill/api/daemon/internal/idea/Element.scala
@@ -8,10 +8,14 @@ import com.lihaoyi.unroll
  * @param name The XML element name
  * @param attributes The optional XML element attributes
  * @param childs The optional XML child elements.
+ * @param childsOrText The optional XML child elements or text content.
  */
 final case class Element(
     name: String,
     attributes: Map[String, String] = Map(),
-    childs: Seq[Element] = Seq(),
+    @deprecated(
+      "Using Elements only is insufficient to represent PCData in XML. Use childsOrText instead.",
+      since = "Mill after 1.1.6"
+    ) childs: Seq[Element] = Seq(),
     @unroll childsOrText: Seq[Element | String] = Seq()
 )
--- a/core/api/daemon/src/mill/api/daemon/internal/idea/Element.scala
+++ b/core/api/daemon/src/mill/api/daemon/internal/idea/Element.scala
@@ -7,7 +7,7 @@ import com.lihaoyi.unroll
  *
  * @param name The XML element name
  * @param attributes The optional XML element attributes
- * @param childs The optional XML child elements.
+ * @param childs Deprecated. The optional XML child elements.
  * @param childsOrText The optional XML child elements or text content.
  */
 final case class Element(
--- a/core/api/daemon/src/mill/api/daemon/internal/idea/Element.scala
+++ b/core/api/daemon/src/mill/api/daemon/internal/idea/Element.scala
@@ -1,7 +1,10 @@
 package mill.api.daemon.internal.idea
 
+import com.lihaoyi.unroll
+
 /**
  * Encoding of an Idea XML configuration fragment.
+ *
  * @param name The XML element name
  * @param attributes The optional XML element attributes
  * @param childs The optional XML child elements.
@@ -9,5 +12,6 @@ package mill.api.daemon.internal.idea
 final case class Element(
     name: String,
     attributes: Map[String, String] = Map(),
-    childs: Seq[Element] = Seq()
+    childs: Seq[Element] = Seq(),
+    @unroll childsOrText: Seq[Element | String] = Seq()
 )
--- a/libs/kotlinlib/package.mill
+++ b/libs/kotlinlib/package.mill
@@ -19,7 +19,7 @@ object `package` extends MillStableScalaModule with BuildInfo {
   }
 
   def moduleDeps =
-    Seq(build.libs.javalib, build.libs.scalalib, build.libs.javalib.testrunner, api, `ksp2-api`)
+    Seq(build.libs.javalib, build.libs.javalib.testrunner, api, `ksp2-api`)
   def localTestExtraModules =
     super.localTestExtraModules ++ Seq(worker)
 
--- a/libs/kotlinlib/package.mill
+++ b/libs/kotlinlib/package.mill
@@ -18,7 +18,7 @@ object `package` extends MillStableScalaModule with BuildInfo {
     override def selectiveInputs = Seq(sources, resources) ++ kotlinSelectiveInputs
   }
 
-  def moduleDeps = Seq(build.libs.javalib, build.libs.javalib.testrunner, api, `ksp2-api`)
+  def moduleDeps = Seq(build.libs.javalib, build.libs.scalalib, build.libs.javalib.testrunner, api, `ksp2-api`)
   def localTestExtraModules =
     super.localTestExtraModules ++ Seq(worker)
 
--- a/libs/kotlinlib/package.mill
+++ b/libs/kotlinlib/package.mill
@@ -18,7 +18,8 @@ object `package` extends MillStableScalaModule with BuildInfo {
     override def selectiveInputs = Seq(sources, resources) ++ kotlinSelectiveInputs
   }
 
-  def moduleDeps = Seq(build.libs.javalib, build.libs.scalalib, build.libs.javalib.testrunner, api, `ksp2-api`)
+  def moduleDeps =
+    Seq(build.libs.javalib, build.libs.scalalib, build.libs.javalib.testrunner, api, `ksp2-api`)
   def localTestExtraModules =
     super.localTestExtraModules ++ Seq(worker)
 
--- a/libs/kotlinlib/src/mill/kotlinlib/KotlinIdeaModule.scala
+++ b/libs/kotlinlib/src/mill/kotlinlib/KotlinIdeaModule.scala
@@ -1,7 +1,7 @@
 package mill.kotlinlib
 
 import mill.api.daemon.internal.idea.{Element, JavaFacet}
-import mill.api.{Task, TaskCtx, experimental}
+import mill.api.{BuildCtx, Task, TaskCtx, experimental}
 import mill.api.daemon.internal.IdeUtils
 
 private lazy val FriendPathsPattern = "^-Xfriend-paths=(.+)$".r
@@ -31,24 +31,19 @@ trait KotlinIdeaModule extends KotlinModule {
         case _ => false
       }
 
-      if (kotlinFriendModules.nonEmpty) {
-        val redirectedFriendPaths = Task.traverse(kotlinFriendModules)(friend =>
-          Task.Anon {
-            friend.genIdeaInternalExt().ideaCompileOutput().path
-          }
+      if (kotlinFriendModulesChecked.nonEmpty) {
+        val redirectedFriendPaths = Task.traverse(kotlinFriendModulesChecked)(friend =>
+          Task.Anon { friend.genIdeaInternalExt().ideaCompileOutput().path }
         )()
-
-        if (redirectedFriendPaths.nonEmpty) {
-          options += redirectedFriendPaths.map(path =>
-            "$MODULE_DIR$/../../" + (path.relativeTo(Task.ctx().workspace)).toString
-          ).mkString("-Xfriend-paths=", ",", "")
-        }
+        options += redirectedFriendPaths
+          .map(path => "$MODULE_DIR$/../../" + (path.relativeTo(BuildCtx.workspaceRoot)).toString)
+          .mkString("-Xfriend-paths=", ",", "")
       }
 
       options.mkString(" ")
     }
 
-    lazy val friendElements = kotlinFriendModules
+    lazy val friendElements = kotlinFriendModulesChecked
       .flatMap(friend => IdeUtils.moduleName(friend.moduleSegments))
       .map(friendName => Element("friend", childsOrText = Seq(friendName)))
 
--- a/libs/kotlinlib/src/mill/kotlinlib/KotlinIdeaModule.scala
+++ b/libs/kotlinlib/src/mill/kotlinlib/KotlinIdeaModule.scala
@@ -48,6 +48,10 @@ trait KotlinIdeaModule extends KotlinModule {
       options.mkString(" ")
     }
 
+    lazy val friendElements = kotlinFriendModules
+      .flatMap(friend => IdeUtils.moduleName(friend.moduleSegments))
+      .map(friendName => Element("friend", childsOrText = Seq(friendName)))
+
     val facets = ideaConfigVersion match {
       case 4 => {
         Seq(
@@ -61,15 +65,10 @@ trait KotlinIdeaModule extends KotlinModule {
                 "useProjectSettings" -> "false"
               ),
               childs = Seq(
-                if (kotlinFriendModules.nonEmpty) {
+                if (friendElements.nonEmpty) {
                   Element(
                     "additionalVisibleModuleNames",
-                    childs = kotlinFriendModules.map(friend =>
-                      Element(
-                        "friend",
-                        childsOrText = Seq(IdeUtils.moduleName(friend.moduleSegments))
-                      )
-                    )
+                    childs = friendElements
                   )
                 } else null,
                 Element(
--- a/libs/kotlinlib/src/mill/kotlinlib/KotlinIdeaModule.scala
+++ b/libs/kotlinlib/src/mill/kotlinlib/KotlinIdeaModule.scala
@@ -2,6 +2,7 @@ package mill.kotlinlib
 
 import mill.api.daemon.internal.idea.{Element, JavaFacet}
 import mill.api.{Task, TaskCtx, experimental}
+import mill.scalalib.internal.IdeUtils
 
 private lazy val FriendPathsPattern = "^-Xfriend-paths=(.+)$".r
 
@@ -72,7 +73,7 @@ trait KotlinIdeaModule extends KotlinModule {
                     childs = kotlinFriendModules.map(friend =>
                       Element(
                         "friend",
-                        childsOrText = Seq(friend.moduleSegments.render.toLowerCase())
+                        childsOrText = Seq(IdeUtils.moduleName(friend.moduleSegments))
                       )
                     )
                   )
--- a/libs/kotlinlib/src/mill/kotlinlib/KotlinIdeaModule.scala
+++ b/libs/kotlinlib/src/mill/kotlinlib/KotlinIdeaModule.scala
@@ -1,7 +1,9 @@
 package mill.kotlinlib
 
-import mill.api.{Task, experimental}
 import mill.api.daemon.internal.idea.{Element, JavaFacet}
+import mill.api.{Task, TaskCtx, experimental}
+
+private lazy val FriendPathsPattern = "^-Xfriend-paths=(.+)$".r
 
 /**
  * Adds a JavaFacet to the generated .iml file with the Kotlin compiler options.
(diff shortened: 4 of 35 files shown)


viluon:
Marking as ready for review now, since none of the CI failures seem like the fault of this PR.

image

To be clear, these are ugly hacks (particularly the crude `internalGenIdea/ideaCompileOutput.dest/classes` replacement) and there are more bugs to be fixed around here but at least it works to some extent, which sadly can't be said about `KotlinIdeaModule` as of 1.1.6.

lihaoyi:
CC @lefou if you want to review this, I'm not that familiar with the GenIDEA Kotlin code paths

souvlakias:
Nice adding the `additionalVisibleModuleNames` element. Checking out what other stuff can be passed, we can also add the `dependsOnModuleNames` ([ref1](https://github.com/JetBrains/kotlin/blob/4996b96b6df70ffc6e219576099d1fc8c5e25bda/jps/jps-common/src/org/jetbrains/kotlin/config/KotlinFacetSettings.kt#L171), [ref2](https://github.com/JetBrains/kotlin/blob/4996b96b6df70ffc6e219576099d1fc8c5e25bda/jps/jps-common/src/org/jetbrains/kotlin/config/facetSerialization.kt#L141)) for `moduleDeps` if that additionally helps the IDE somehow

runner/idea/src/mill/idea/GenIdeaImpl.scala
```diff
@@ -795,6 +796,18 @@ class GenIdeaImpl(
...
+      else {
+        <component name="FacetManager">
+          {
+          for (facet <- facets) yield <facet type={facet.`type`} name={facet.name}>
+            {ideaConfigElementTemplate(facet.config)}
+          </facet>
+        }
+        </component>
+      }
+    }
```
lefou: Let's revert this. IDEA is constantly reordering the IML/XML files (e.g. attributes), making this a moot argument. Instead, I'd rather keep the general part (common for all Java-like modules) before the optional addition. Esp. since the ordering doesn't seem to change anything in the logic.

viluon:
Good catch @souvlakias. IMLs generated by Gradle contain way more information. I still haven't fixed explicit API settings and gone through the various other things in a Gradle IML module (it seems to provide a higher level interface for Kotlin compiler flags, along with the raw CLI options).

I would like to tackle these things in follow-up PR(s) to get some incremental improvement ASAP. I have a production project waiting for this.

libs/kotlinlib/src/mill/kotlinlib/KotlinIdeaModule.scala
```diff
@@ -20,6 +22,30 @@ trait KotlinIdeaModule extends KotlinModule {
...
+          case opt => KotlinCompilerOptions(friendPaths = Seq.empty, unrecognised = Seq(opt))
+        })
+      }
+
+    lazy val redirectedFriendPaths = parsedKotlincOptions.friendPaths.map { path =>
+      val moduleDirPath = "$MODULE_DIR$/../../" + os.Path(path).relativeTo(Task.ctx().workspace)
+      if (!moduleDirPath.endsWith("/compile.dest/classes")) {
+        Task.ctx().log.warn(
+          s"Module ${implicitly[ModuleCtx].segments.render} has a non-standard friend path: $moduleDirPath. Generated files may be inaccurate."
+        )
```
lefou: Instead of this path mangling, which is error-prone (esp. when dealing with a bespoken module), we should introduce a `kotlinFriendModules: Seq[KotlinModule]` method, as discussed on Discord. This is also more in line with how Bazel handles friend modules with the `associates` attribute (https://github.com/bazel-contrib/rules_kotlin/pull/465).

souvlakias:
@viluon Yeah the extra element isn't urgent (since we don't know for sure how it helps), it can be added later.
I think the important things for this PR are in the comments I added above

core/api/daemon/src/mill/api/daemon/internal/idea/Element.scala
```diff
@@ -11,3 +11,9 @@ final case class Element(
     attributes: Map[String, String] = Map(),
     childs: Seq[Element] = Seq()
 )
+
+object Element {
+
+  /** Create an Element representing a text node (rendered as PCDATA, not a child element). */
+  def text(value: String): Element = Element("#text", attributes = Map("value" -> value))
+}
```
lefou: What is the exact issue this addition solves?

viluon: Currently, it's impossible to express `<friend>module</friend>` (AFAICT). It's only possible to express `<friend><module/></friend>`.

This is a hacky workaround that preserves the public API.

runner/idea/src/mill/idea/GenIdeaImpl.scala
viluon: Ah, I wasn't aware of this IDEA behaviour. I agree that it doesn't make sense to try to match it if the order is unstable. In my testing with 2026.1.1, it was stable and reproducible, however. Matching IDEA's ordering made it much easier to compare IML files from multiple sources (Gradle, IDEA project structure, Mill).

libs/kotlinlib/src/mill/kotlinlib/KotlinIdeaModule.scala
viluon: What is the plan for such a change? From my PoV, that would require downstream adoption along with a minor version bump (at least), right? I mean, passing `-Xfriend-paths` explicitly should then be discouraged, maybe deprecated.

(Feel free to redirect me to docs if Mill stability guarantees / migration philosophy is documented somewhere, or to a previous discussion on the matter.)

runner/idea/src/mill/idea/GenIdeaImpl.scala
viluon: Take this with a grain of salt given the unreliability of LLMs, but a quick search with Copilot led to [this](https://github.com/JetBrains/intellij-community/blob/cc800aacf38c5ba774a3b45284f7b7017fb0f7c9/platform/projectModel-impl/src/com/intellij/configurationStore/SchemeManagerIprProvider.kt#L116). It seems that components in the project model configuration are sorted lexicographically before they're saved.

<details>
<summary>Gemini 3.1 explanation, grounded by GitHub search</summary>

In IntelliJ IDEA, the serialization logic (primarily driven by `JDOMUtil` and the project model components) is designed to be deterministic to minimize version control conflicts. 

Here is how the order of XML containers in `*.iml` files is decided:

1. **`<component>` tags:** Sorted alphabetically by their `name` attribute.
2. **`<orderEntry>` tags:** Ordered sequentially based on the exact classpath order defined in the module's settings (Project Structure -> Modules -> Dependencies). This order is semantic and dictates the compilation and runtime classpath.
3. **`<content>` and `<sourceFolder>` tags:** Sorted alphabetically by their `url` attribute (lexicographical order of the directory paths).
4. **XML Attributes:** Inside any given tag, attributes are generally sorted alphabetically by their key name.

In short: structural elements that don't affect build semantics are sorted lexicographically (by name/URL) for stable persistence, while elements like `<orderEntry>` maintain their explicit, user-defined sequence because order matters to the compiler.

</details>

core/api/daemon/src/mill/api/daemon/internal/idea/Element.scala
lefou: I see. Maybe, we could just use `name=""` and require that `childs` must be empty in that case?

lefou: The preferred solution that's not binary compatible would be `childs: Seq[Element|String]`, I guess.

lefou: Maybe we can add an alternative to `childs`? 

```scala
final case class Element(
    name: String,
    attributes: Map[String, String] = Map(),
    childs: Seq[Element] = Seq(),
    @unroll childsOrText: Seq[Element|String] = Seq()
)
```

libs/kotlinlib/src/mill/kotlinlib/KotlinIdeaModule.scala
lefou: Sure, in SemVer, adding any new API should require a minor version bump, but due to the many modules a Mill release contains (Mill core vs Mill libraries all have the same number, see also #7056), we often allowed new tasks (with sensible defaults) even in micro versions. 

That being said, the next planned version for the `main` branch is `1.2.0`, a minor version bump, so we can perfectly add a new method with a default implementation.

viluon: In that case I'm happy to implement that tomorrow morning (starting 9:00ish GMT+2) and maybe have a look at some of the other concerns with `KotlinIdeaModule`s as well.

core/api/daemon/src/mill/api/daemon/internal/idea/Element.scala
viluon: I don't think the text variant can have more than one entry, can it? I.e. it's rather `Seq[Element]|String`.

FWIW I think the proper model is what I implemented earlier, making `Element` a sum type of two variants. If you're willing to consider a change of an API with `internal` in the package name not breaking, we could bring that back.

viluon: If instead the hacky workaround would stay, I don't mind revising it however you wish, `name=""` seems like a better choice than a special reserved string.

lefou: Although, the classes reside under `internal` (and are probably excluded from any mima-checks), they still appear in public stable API and thus, must be kept binary compatible. Otherwise users can't use them to customize Mill.

I think `Seq[Element|String]` is the better fit, since you always can insert text freely in XML before, between and after a child element. Two consecutive strings are simply appended.

runner/idea/src/mill/idea/GenIdeaImpl.scala
lefou: If we want to apply this, we should ask IDEA folks directly or find some reference. The auto completed answer isn't valid enough for my taste, although it sounds plausible. Nevertheless, I'd like to decouple such changes from this PR.

viluon: Alright, I'll split this commit off into its own PR. I also raised a [question](https://intellij-support.jetbrains.com/hc/en-us/community/posts/35352485668882-Order-of-elements-in-IML-files) on the community forums, we'll see.

core/api/daemon/src/mill/api/daemon/internal/idea/Element.scala
viluon: I didn't think of mixing text and children, you're right, it's also what the `sealed trait` approach allowed.

runner/idea/src/mill/idea/GenIdeaImpl.scala
viluon: Removed in [this force-push](https://github.com/com-lihaoyi/mill/compare/d19445718cccb54eb9503fd6f5869b210ec53c3d..28c1a42a84799b6178823ae6be545d8eb524b4df), I opened #7084 to fix this separately (pending an answer from JetBrains).

libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -510,11 +519,7 @@ trait KotlinModule extends JavaModule with KotlinModuleApi { outer =>
     override def kotlinVersion: T[String] = Task { outer.kotlinVersion() }
     override def kotlincPluginMvnDeps: T[Seq[Dep]] =
       Task { outer.kotlincPluginMvnDeps() }
-      // TODO: make Xfriend-path an explicit setting
-    override def kotlincOptions: T[Seq[String]] = Task {
-      outer.kotlincOptions().filterNot(_.startsWith("-Xcommon-sources")) ++
```
viluon, line 515: this filtering removal is a potentially breaking change, I need to bring that back

libs/kotlinlib/src/mill/kotlinlib/KotlinIdeaModule.scala
viluon: You can now review the approach here: 3e43e67cffa5298a52b1cfed188a99663cc0aae1

A proper adjustment of the `KotlinIdeaModule` changes is still pending.

libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -399,10 +399,19 @@ trait KotlinModule extends JavaModule with KotlinModuleApi { outer =>
       }
     }
 
+  def friendModules: Seq[KotlinModule] = Seq.empty[KotlinModule]
```
lefou: We should prefix all specific options with `kotlin` and scaladoc is needed.

```suggestion
  def kotlinFriendModules: Seq[KotlinModule] = Seq.empty[KotlinModule]
```

libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -510,10 +519,9 @@ trait KotlinModule extends JavaModule with KotlinModuleApi { outer =>
     override def kotlinVersion: T[String] = Task { outer.kotlinVersion() }
     override def kotlincPluginMvnDeps: T[Seq[Dep]] =
       Task { outer.kotlincPluginMvnDeps() }
-      // TODO: make Xfriend-path an explicit setting
+    override def friendModules: Seq[KotlinModule] = super.friendModules ++ Seq(outer) ++ outer.friendModules
     override def kotlincOptions: T[Seq[String]] = Task {
-      outer.kotlincOptions().filterNot(_.startsWith("-Xcommon-sources")) ++
-        Seq(s"-Xfriend-paths=${outer.compile().classes.path.toString()}")
+      super.kotlincOptions().filterNot(_.startsWith("-Xcommon-sources"))
```
lefou: If you know why we filter this, please add a code-comment.

viluon: I unfortunately do not, I just wanted to preserve the former behaviour.

core/api/daemon/src/mill/api/daemon/internal/idea/Element.scala
viluon: So the relevant commit has been reworked. As far as I could tell, the only read usage is in `GenIdeaImpl.scala`.

I haven't reworked the write usages of `childs` nor have I deprecated `childs`. I suggest to do so in a follow-up PR. Happy to add another commit here if you'd rather sweep this out sooner.

libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
viluon: [Fixed](https://github.com/com-lihaoyi/mill/compare/8053011418c5b3a9cb2629ed9e387a42b9555c79..9eac14fef79f50b05a3934eeba2a9fa756f46e91).

libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -532,10 +547,10 @@ object KotlinModule {
     override def kotlinVersion: T[String] = Task { outer.kotlinVersion() }
     override def kotlincPluginMvnDeps: T[Seq[Dep]] =
       Task { outer.kotlincPluginMvnDeps() }
-    // TODO: make Xfriend-path an explicit setting
+    override def kotlinFriendModules: Seq[KotlinModule] =
+      super.kotlinFriendModules ++ Seq(outer) ++ outer.kotlinFriendModules
```
lefou: Is this option meant to be transitive? If yes, we should calculate the transitive set anyways and don't need to hand-roll it here.


```suggestion
      super.kotlinFriendModules ++ Seq(outer)
```

viluon: no, I don't think friend modules should be transitive. Tests are a special case though, they should match the visibility of the main module. That's at least what I would naively expect but I concede I haven't tried this in a Gradle project.

viluon: actually, it seems that test source sets in Gradle don't inherit the friend paths of the main source set, so I'll remove that anyway

viluon: [Fixed](https://github.com/com-lihaoyi/mill/compare/0b36499da9065e0bc3b1165b49daf70cabc6d6e4..04b46b32b909213fcf510559b6b849db31bd91d9).

libs/kotlinlib/src/mill/kotlinlib/KotlinIdeaModule.scala
viluon: `KotlinIdeaModule` now works directly with `kotlinFriendModules`. I'll resolve this conversation to make keeping track of the latest concerns easier. Feel free to unresolve if there's anything I forgot to address here.

core/api/daemon/src/mill/api/daemon/internal/IdeUtils.scala
```diff
@@ -0,0 +1,32 @@
...
+        case (sb, Segment.Cross(s)) => sb.append("-").append(s.mkString("-"))
+      }
+      .mkString
+      .toLowerCase()
+
+    // If for whatever reason no name could be created based on the module segments, create a
+    // generic one from the Mill Module folder name. Users can rename the project inside the
+    // IDE if they like.
+    if (name.isBlank && path != null) path.getFileName.toString
+    else name
```
lefou: I don't think we want this. What are the reasons the module name is empty? Could only be the empty-segments case. Let's deal on the caller site with this. We should return an `Option[String]`, to make this explicit and easy handleable (`getOrElse`) on the caller side.

libs/scalalib/src/mill/scalalib/internal/IdeUtils.scala
```diff
@@ -0,0 +1,32 @@
+package mill.scalalib.internal
```
lefou: The location doesn't make sense, it's to wide downstream. What should a `javalib` caller do?

We should move this to `mill.api.daemon.internal`.

core/api/daemon/src/mill/api/daemon/internal/idea/Element.scala
```diff
@@ -1,13 +1,17 @@
...
  * @param name The XML element name
  * @param attributes The optional XML element attributes
  * @param childs The optional XML child elements.
  */
 final case class Element(
     name: String,
     attributes: Map[String, String] = Map(),
-    childs: Seq[Element] = Seq()
+    childs: Seq[Element] = Seq(),
+    @unroll childsOrText: Seq[Element | String] = Seq()
```
lefou, line 20: API docs needed. `childs` should be deprecated, with a meaningful reason and version "Mill after 1.1.6"

runner/eclipse/package.mill
```diff
@@ -4,6 +4,6 @@ import mill.*
 import millbuild.*
 
 object `package` extends MillPublishScalaModule {
-  def moduleDeps = Seq(build.libs.util, build.core.internal)
+  def moduleDeps = Seq(build.libs.util, build.libs.scalalib, build.core.internal)
```
lefou: Indicator, that somethings wrong. This dependency in not wanted.

libs/kotlinlib/src/mill/kotlinlib/kapt/KaptModule.scala
```diff
@@ -122,9 +122,6 @@ trait KaptModule extends KotlinModule { outer =>
           "-classpath",
           compileCp.mkString(File.pathSeparator)
         ),
-        when(kotlinExplicitApi())(
-          "-Xexplicit-api=strict"
-        ),
```
lefou, line 127: The fact that we need to calc additional compiler args here is fishy. Whatever these are, those are likely also needed in the IDE, so we should aim to handle them all in `mandatoryKotlincOptions`.

Please, add a TODO here, so we don't forget.

libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
```diff
@@ -399,6 +396,14 @@ trait KotlinModule extends JavaModule with KotlinModuleApi { outer =>
...
+  /**
+   * Modules whose internal declarations are visible to this module.
+   *
+   * Drives `-Xfriend-modules` for Kotlin/JS and `-Xfriend-paths` elsewhere.
+   * See [`friendPaths`](https://kotlinlang.org/api/kotlin-gradle-plugin/kotlin-gradle-plugin-api/org.jetbrains.kotlin.gradle.tasks/-base-kotlin-compile/friend-paths.html) for the Gradle equivalent.
+   */
+  def kotlinFriendModules: Seq[KotlinModule] = Seq.empty[KotlinModule]
+
   /**
    * Additional Kotlin compiler options to be used by [[compile]].
```
lefou, line 422: Should we enforce, that `kotlinFriendModules` are also in `moduleDeps`?

libs/kotlinlib/package.mill
```diff
@@ -18,7 +18,8 @@ object `package` extends MillStableScalaModule with BuildInfo {
     override def selectiveInputs = Seq(sources, resources) ++ kotlinSelectiveInputs
   }
 
-  def moduleDeps = Seq(build.libs.javalib, build.libs.javalib.testrunner, api, `ksp2-api`)
+  def moduleDeps =
+    Seq(build.libs.javalib, build.libs.scalalib, build.libs.javalib.testrunner, api, `ksp2-api`)
```
lefou: Should not be necessary

libs/scalalib/src/mill/scalalib/internal/IdeUtils.scala
viluon: Fixed in 6f4a458ea9c6c9a4ee2b29bf70792b5092c3f377.

libs/kotlinlib/package.mill
viluon: Fixed in a2d328481f765d92f5c1f31bd20546642c88effc.

runner/eclipse/package.mill
viluon: Fixed in a2d328481f765d92f5c1f31bd20546642c88effc.

core/api/daemon/src/mill/api/daemon/internal/IdeUtils.scala
viluon: Fixed in ac693b38ba172233b78e7f32a946f474756a2dc0.

libs/kotlinlib/src/mill/kotlinlib/kapt/KaptModule.scala
viluon: Fixed in a35568ab4acb5e11fc9d893f78b54db36ac74006.

core/api/daemon/src/mill/api/daemon/internal/idea/Element.scala
viluon: Good point, fixed in 3e5320a4a516b8e25c6e513edbbe5e454fff23bd.

libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
viluon: maybe, but I'm not sure how to achieve that. Something like
```scala
  require(kotlinFriendModules.toSet.subsetOf(moduleDeps.toSet), "All kotlinFriendModules must also be declared in moduleDeps")
```
in the primary constructor? I pushed that in 60c444bab8a986661c5e14b164f13f0884b6895b but idk if it's a good solution, I'm not well-versed in the order of evaluation of Scala mixins so there may be unforeseen issues.

lefou: We could just check at the use side, or adapt the `JavaModule.moduleDepsChecked` approach. The nicest thing would be to just override `moduleDepsChecked` and hook into it, but unfortunately it was made final.

viluon: Ok, moved in aee69bae30358b57a72cb93ac25ef3c9d919b1fa. Is that a good approach in your opinion? It's a bit sad that `kotlinFriendModules` have to be reevaluated multiple times but using a local `val` is illegal in a Task body.

lefou (commented):
The PR looks good so far. Some points though:

* We should add some explicit documentation about the new `kotlinFriendModules` task and its default value in `KotlinTests`.

* The `example.thirdparty.arrow` test is currently failing, which could be related to the changes of `kotlincOptions`.

* I already pushed some changes to address the reported MiMa issue and added the `lazy val kotlinFriendModulesChecked` for internal reuse.

viluon:
Tests are green again with acfc1db5586d67e727ea68ef56df1ffeb94dd8cd -- I noticed that `-Xfriend-paths` was listed as unrecognized in the logs of `example.thirdparty.arrow` tests for Kotlin/JS.

Do you think we could merge this as-is with a docs update to follow afterwards in a separate PR? I'm not sure where/how to document the friend modules option, I think extending an example for `kotlinlib` is probably the right thing to do, but I'm very confused as to why [the example for `moduleDeps`](https://github.com/com-lihaoyi/mill/blob/65e176956d21c9fcbcec9e4bf66f1f59ad8e511e/example/kotlinlib/basic/2-module-deps/build.mill.yaml) contains no docs despite what's [rendered on the website](https://mill-build.org/mill/dev-1.1.6-7-15943a/kotlinlib/intro.html#_multi_module_projects).

lefou:
Great! I think we can update the docs in a follow-up PR and merge this. Thank you for it!

The example tests for `javalib`, `scalalib` and `kotlinlib` try to share the same structure and if the names match, the one in `scalalib` is used as a template, and the language specific one only provides partial snippets and additions.

viluon:
I see, then I shall look into extending the docs tomorrow. I would also like to take a stab at supporting Kotlin compiler plugins in the `.iml` files, I noticed earlier that the Gradle equivalent has a special section for those (each compiler plugin can have its own options, even).

Issue #6077, opened 2025-11-01
Make `./mill init` work for spring-boot-examples project


lihaoyi:
https://github.com/in28minutes/spring-boot-examples

There are a bunch of small example projects in this repo, we should make sure that for each one we can do `./mill init`, `./mill __.compile`, and `./mill __.test` and have it all work as much as possible

lihaoyi:
CC @ajaychandran

lihaoyi:
@ajaychandran I saw you linked this in https://github.com/com-lihaoyi/mill/pull/6083, but it's not obvious whether it was resolved. Can we close this issue?

ajaychandran:
@lihaoyi This remains unresolved. A mapping for `spring-boot-starter-test` is required to enable test functionality.

Issue #5571, opened 2025-07-21
Add heuristics to translate common tools and plugins during `./mill init` from SBT, Maven, and Gradle


lihaoyi:
`./mill init` should do a best-effort to try and identify usages of tools and plugins in the existing build tool that Mill has support for, and translate them to the Mill equivalent. We won't be able to identify 100% of tools, and the tools we do identify we won't be able to translate 100% precisely, but even a best-effort detection and translation would help anyone using `./mill init` more quickly get to a working state.

Concretely this means:

Java
- [ ] ErrorProne
- [ ] Checkstyle
- [ ] PMG
- [ ] PalantirJavaFormat
- [ ] Spotless
- [ ] Jacoco
- [ ] RevAPI
- [ ] Lombok

Scala
- [ ] Scalafix
- [ ] Scalafmt
- [ ] Scoverage
- [ ] Mima

Kotlin
- [ ] Detekt
- [ ] KtLint
- [ ] KtFmt
- [ ] Kover

For example, importing the https://github.com/mockito/mockito codebase currently results in a failing due to the `def javacOptions = Seq("-Xplugin:ErrorProne ", "-XDcompilePolicy=simple")` config not working right

lihaoyi:
CC @ajaychandran

ajaychandran:
> For example, importing the https://github.com/mockito/mockito codebase currently results in a failing due to the def javacOptions = Seq("-Xplugin:ErrorProne ", "-XDcompilePolicy=simple") config not working right

The underlying issue here is that the `jvmId` configured is not compatible with the version of `com.google.errorprone:error_prone_core` dependency. This failure is observed in other tests as well.
- `MillInitGradleEhcache3Tests`: The JDK 17 requirement is enforced with a [runtime check](https://github.com/ehcache/ehcache3/blob/6c754832b0c473fee5806cb9a875010489ef1df4/build.gradle#L32).
- `MillInitGradleFastCsvTests`: [Requires JDK 25](https://fastcsv.org/guides/contribution/).
- `MillInitGradleMockitoTests`: Compiled for Java 11 but the project is published using `21.0.8 (Azul Systems, Inc.)`, based on the [JAR](https://repo1.maven.org/maven2/org/mockito/mockito-core/5.20.0/mockito-core-5.20.0.jar) manifest.

In all these cases, the correct `jvmId` cannot be determined during conversion. Defining a setting that was never configured leads to unsurprising behavior, from the user's perspective.

[...]

PR #7093, opened 2026-05-08
`mill init`: Detect spring boot projects

souvlakias:
Part of:
- https://github.com/com-lihaoyi/mill/issues/6077
  - 16/22 examples here pass the `./mill init && ./mill test`

and
- #5571 

Adds detection of Springboot project and applies `SpringBootModule`. 
The `spring-start` example is generated by https://start.spring.io/

### Steps - Maven
- Detect SpringBoot project for maven modules that inherit from  `spring-boot-starter-parent`
- Use `rawModel` to extract a clean `depencyManagement` without deps coming from spring's BOM
  - See the diff on the new expected files for the cleaner `depManagement`  
- Extend `SpringBootModule` and add `springBootPlatformVersion`

### Steps - Gradle
- Detect the spring-boot plugin and its version
  - Plugin version detection is a bit hacky since the [Plugin](https://docs.gradle.org/current/javadoc/org/gradle/api/Plugin.html) api doesn't provide anything 

#### Notes
As mentioned this is a small starting point for extending the `init` support, will follow up in later PRs for different configurations and frameworks


libs/init/maven/test/resources/spring-start/pom.xml
vaslabs, line 1: We need to double check this maven is reproducible by https://start.spring.io .

souvlakias: Yes, the current version in this PR is generated by https://start.spring.io/

vaslabs:
I think this should be good enough as a starting point to rework and have an ootb experience with modern spring boot configurations.

The idea is we break down the work in multiple PRs as we add init support for various configurations and frameworks (spring boot, quarkus and micronaut)

wdyt @lihaoyi @lefou ?

lihaoyi:
@vaslabs sounds good to me

lihaoyi:
@vaslabs could you help review this? If it looks good to you I'll go ahead and merge it

souvlakias:
In the meantime also added SpringBoot detection for Gradle projects, wasn't that complex.

Btw, when running `mill init` on a gradle project, I noticed a weird hang (1-2 minutes) after the `build.mill` files are generated. Will investigate ...

vaslabs (approved):

PR #7099, opened 2026-05-13
Fix outdated doc comment

viluon:
`protected` is misplaced here:
image

Immediately marking the PR as ready as there are no code changes.


lefou (approved):
Nice catch. In contrast, there is a TODO on `RootModule0.millDiscover`, to make it protected, which is a bin-compat change, so it needs to wait for next major release.

PR #7103, opened 2026-05-14
Upgrade zinc

lihaoyi:
This version of zinc requires Scala3.8.2/Java17+ (https://github.com/sbt/zinc/issues/1696), so we add a special case to fork Javac when compiling Java code with Java<17. Since Scalac runs in memory, we can't compile Scala with versions of Scala <17 after this PR. Which might be fine by this point since the community is moving to Java 17.


lihaoyi:
Currently blocked on https://github.com/sbt/zinc/pull/1698, but once that is published we should be able to land this CC @lefou

PR #7107, opened 2026-05-15
fix(javalib): avoid duplicate non-parallel test runners

### LINKED ISSUE #7106: `testParallelism=false` can reuse the same test dest across runners
## TL;DR

Since #5630, non-parallel tests also use `runTestQueueScheduler`.

That scheduler can still create multiple runners for one test group, but with `testParallelism=false` all runners use the same `groupFolder`.

So a later runner can fail with:

```text
java.nio.file.FileAlreadyExistsException: .../testForked.dest/testargs
```

## Introduced

Bisected to:

- #5630
- commit `628e766d2d1085846ee5b6447199da97edeaaa0a`
- first released in `1.0.3`

`1.0.2` and earlier do not contain this path.

## Why

Runner count is still based on `jobs` / `numTests`:

```scala
processIndex <- 0 until Math.max(Math.min(jobs, numTests), 1)
```

But non-parallel runners share the same folder:

```scala
if (testParallelism && filteredClassCount != 1) groupFolder / workerLabel
else groupFolder
```

Shape:

```text
runner 0 -> groupFolder/testargs
runner 1 -> groupFolder/testargs
```

## Expected

When `testParallelism=false`, spawn only one runner per test group.

## Suggested Fix

Cap subprocess count to `1` unless test parallelism is actually enabled:

```scala
if (testParallelism && filteredClassCount != 1)
  Math.max(Math.min(jobs, numTests), 1)
else
  1
```

nguyenyou:
Fixes #7106.

Commit structure:

1. Add failing regression for non-parallel queue scheduler reusing `testForked.dest`.
2. Cap non-parallel scheduler to one runner per group.

Validation:

- Commit 1 fails with `FileAlreadyExistsException: .../testForked.dest/testargs`.
- Final branch passes focused scheduler tests and existing UTest/ScalaTest runner tests.

Regression setup:

- Test module sets `testParallelism = false`.
- It has one test group with two classes.
- `UnitTester(..., threads = Some(2))` lets the old scheduler create two runner futures.
- A tiny fake test framework writes one extra class marker into `test-classes` during `Runner.done()`.

Old behavior:

```text
runner 0 -> testForked.dest/testargs
runner 0 drains tests, then leaves one marker
runner 1 sees marker, reuses testForked.dest
runner 1 writes testargs again -> FileAlreadyExistsException
```

Fixed behavior: non-parallel mode creates one runner per group, so no second runner reuses the same dest.

### DIFF
--- a/libs/javalib/src/mill/javalib/TestModuleUtil.scala
+++ b/libs/javalib/src/mill/javalib/TestModuleUtil.scala
@@ -190,9 +190,14 @@ final class TestModuleUtil(
     testClassQueueFolder
   }
 
-  def jobsProcessLength(numTests: Int) = {
-    val cappedJobs = Math.max(Math.min(Task.ctx().jobs, numTests), 1)
-    (cappedJobs, cappedJobs.toString.length)
+  def jobsProcessLength(filteredClassCount: Int, numTests: Int) = {
+    val processCount = TestModuleUtil.testSubprocessCount(
+      testParallelism = testParallelism,
+      filteredClassCount = filteredClassCount,
+      jobs = Task.ctx().jobs,
+      numTests = numTests
+    )
+    (processCount, processCount.toString.length)
   }
 
   def runTestQueueScheduler(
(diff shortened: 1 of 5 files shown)


lihaoyi:
Thanks @nguyenyou, I kicked off a publish job so in an hour or so you should have a dev version of Mill you can use https://github.com/com-lihaoyi/mill/actions/runs/25916729330

PR #7111, opened 2026-05-16
Fix Spotless RelPathRef decoding

odenix:
Fix Spotless RelPathRef decoding

Spotless RelPathRef writes by delegating to PathRef, producing values like
`ref:v0:...:/path/to/config`. The corresponding reader receives this decoded
string and passes it to `upickle.read[PathRef](t)`, which treats `t` as raw JSON text.

That fails because `ref:v0:...:/path/to/config` without quotes is not valid JSON.
The fix is to replace `t` with `ujson.Str(t)`, so `t` is read as a JSON string value
rather than as raw JSON text.

This fixes Spotless task cache misses for options such as a non-default
`ScalaFmt.configFile`, where Mill wrote `out/spotlessFormats.json` but could not
read it back on the next run.

### DIFF
--- a/libs/javalib/src/mill/javalib/spotless/Format.scala
+++ b/libs/javalib/src/mill/javalib/spotless/Format.scala
@@ -293,7 +293,7 @@ object Format {
         def write0[V](out: Visitor[?, V], v: RelPathRef) =
           summon[ReadWriter[PathRef]].write0(out, v.ref)
         def mapNonNullsFunction(t: String) =
-          if t.startsWith("ref:") then RelPathRef(upickle.read[PathRef](t))
+          if t.startsWith("ref:") then RelPathRef(upickle.read[PathRef](ujson.Str(t)))
           else t
       }
   }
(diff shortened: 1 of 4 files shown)


lihaoyi:
Please explain what this is fixing and why this is the right fix

odenix:
@lihaoyi I've prepared an end-to-end repro demonstrating the cache issue:

https://github.com/odenix/mill-spotless-repro

The PR also includes a unit test that fails before my fix with:

```
ujson.ParseException: expected json value got "r" at index 0
```
and passes with the fix applied.

lihaoyi:
@odenix please write a proper PR description

odenix:
Fixed commit and PR messages.
Made a tiny tweak to the unit test.

odenix:
Looking at CI, some follow-up fixes are probably required.

odenix:
With my two follow-up fixes, the following commands succeed locally (WSL2, Ubuntu 24.04, ext4):
```
./mill libs.scalalib.test mill.scalalib.spotless.SpotlessTests
./mill libs.javalib.test
```

I've never had a successful run of `./mill libs.scalalib.test `, but this also happens on `main` and seems unrelated.

lihaoyi:
Thanks @odenix , I kicked off a publish with your commit so in an hour or two you'll have a dev release you can try out https://github.com/com-lihaoyi/mill/actions/runs/26114532308

PR #7117, opened 2026-05-18
`mill init`: Detect Quarkus + fixes

souvlakias:
Part of:
- #5571 

This PR detects the Quarkus plugin and applies `QuarkusModule`.
### Example/Test addition:
- [quarkus-quickstarts](https://github.com/quarkusio/quarkus-quickstarts) in example migrating. This runs `__.compile` on all suprojects and tests in just 2.
- A simple `getting-started` example in `libs.init.gradle`.

#### Notes on Gradle migration:
`QuarkusModule` requires some publish information which for Maven worked easily since `init` always extends `PublishModule` and they can be picked up from there. 
This PR makes it so when running `init` on Gradle and quarkus is detected, it also extends `PublishModule` and includes artifact/pom information. (See added example)

### Additional fixes / tweaks:
- Calls `System.exit` when importing Gradle builds so it doesn't hang for another minute until helper deamons stop
- Removes a redundant `MavenBuildGenTests` object
- Fixes some yaml string quoting (issues with `${x}` placeholders)


libs/init/buildgen/api/src/mill/main/buildgen/ModuleSpec.scala
```diff
@@ -13,6 +13,7 @@ case class ModuleSpec(
     alias: Option[String] = None,
     moduleDir: Value[String] = Value(),
     springBootPlatformVersion: Value[String] = Value(),
+    quarkusPlatformVersion: Value[String] = Value(),
```
lefou: A more general thing, that applies not only to this location: We are within the same major version, so we should always make changes in a least-disruptive way. That means, we should add new parameters to the end, not in-between existing parameters. This helps on the caller side, but also the code writer, esp. when having multiple branches.

If there is need to group some parameters or comunicate context, use the scaladoc.

libs/init/maven/test/src/mill/main/maven/MavenBuildGenTests.scala
```diff
@@ -52,13 +52,9 @@ trait MavenBuildGenTests extends TestSuite {
...
 }
 
-object MavenBuildGenTests extends MavenBuildGenTests {
-  override def expectedDir: os.SubPath = "expected"
-}
-
 object MavenBuildGenYamlTests extends MavenBuildGenTests {
   override def expectedDir: os.SubPath = "expected"
-  override def extraArgs = Seq("--declarative", "true")
+  // No extra args, ensure default is declarative (YAML)
```
lefou: Can we please keep some ``` this is equivalent to `--declarative true` ```, for documentation purposes? I think this flag in not well covered in docs yet.

libs/init/buildgen/api/src/mill/main/buildgen/ModuleSpec.scala
souvlakias: Alright :+1:

souvlakias:
Merged with main. I removed the new addition of picking up annotationProcessorMvnDeps from maven builds since there were some conflicts with latest errorProne changes in main. Will follow up in separate PR. 
~~Will mark as ready after fork's CI passes~~

vaslabs:
this lgtm I think, I'll update the branch once again but the failures look to be due to flakiness

souvlakias:
@lefou @lihaoyi, are you happy to merge this?

lefou (approved):
Looks good to me.

PR #7119, opened 2026-05-19
fix(javalib): add batch task mode for forked tests

### LINKED ISSUE #7113: Weaver GlobalResource loses forkEnv under test scheduler
## TLDR

Regression since **Mill 1.0.3**: Mill calls `runner.tasks(...)` during scheduler pre-filtering in the Mill/Zinc worker, before the forked test JVM is launched. For Weaver, `runner.tasks(...)` initializes `GlobalResource`, so `forkEnv` is not visible there.

Then the queue scheduler calls `runner.tasks(...)` one claimed class at a time, so Weaver's `GlobalResource` `TaskDef` and resource-using suite `TaskDef` are split apart.

Minimal repro: https://github.com/nguyenyou/mill-weaver-forkenv-global-resource-repro

## Root Cause

Current path:

```text
TestModuleUtil.runTests
  |
  |-- ZincOp.GetTestTasks
  |     |
  |     `-- TestRunnerUtils.getTestTasks0
  |           |
  |           `-- runner.tasks(...)   // runs in Mill/Zinc worker, no forkEnv
  |
  `-- forked test JVM
        |
        `-- queue scheduler calls runner.tasks([one claimed class])
```

For most frameworks, `runner.tasks(...)` is effectively cheap task materialization. For Weaver, it is semantically meaningful: global resources are collected and initialized from the full `TaskDef` batch.

So there are two related scheduler issues:

1. `runner.tasks(...)` runs outside the forked JVM, so framework setup cannot see `forkEnv`.
2. The queue scheduler splits related `TaskDef`s, so Weaver sees the suite without the global resource.

## Regression / Affected Versions

Bisect points to:

- PR: https://github.com/com-lihaoyi/mill/pull/5630
- Commit: `628e766d2d1085846ee5b6447199da97edeaaa0a`
- Commit title: `Consolidate old TestModuleUtil#runTestDefault code path into runTestQueueScheduler`

First affected stable release: **Mill 1.0.3**

Affected:

- `1.0.3+`
- all current `1.1.x` releases, including `1.1.5`

Not affected:

- `0.12.x`
- `1.0.0` through `1.0.2`

User report: `0.12.16` works, `1.1.5` fails.

## Expected

Framework setup and `TaskDef` batching should happen inside the forked test JVM with `forkEnv` applied, or Mill should preserve full selected `TaskDef` batch semantics for frameworks that require it.

## Actual

Mill pre-filters by invoking framework task creation in a worker without `forkEnv`, then the forked queue runner invokes framework task creation per claimed class.

nguyenyou:
## TLDR

Add opt-in `testBatchFrameworkTasks` for sbt-testing frameworks that need the full selected `TaskDef` batch in one `Runner.tasks(...)` call.

Fixes #7113.

## Why

Mill's queue scheduler calls `Runner.tasks(...)` per claimed class.

That is fine for most frameworks, but not for frameworks that wire related `TaskDef`s together during `Runner.tasks(...)`.

Example: Weaver `GlobalResource` needs to see the resource `TaskDef` and suite `TaskDef`s together, inside the forked JVM where `forkEnv` is available.

## What Changed

- Add `def testBatchFrameworkTasks: T[Boolean] = false`
- When true:
  - skip queue pre-filtering outside the forked test JVM
  - run each fork group as one batch inside the forked JVM
  - pass the full selected `TaskDef` batch to `Runner.tasks(...)`
- Keep default behavior unchanged for all frameworks

## Tradeoff

`testBatchFrameworkTasks = true` disables intra-group queue parallelism.

`testForkGrouping` still works: separate groups can still run separately.

## Safety

Default is `false`, so existing queue behavior remains unchanged.

The batch path reuses the same fork group folder naming as the queue path. Empty batches use an empty queue so framework setup/done still runs without selecting all tests.

## Tests

- `./mill libs.javalib.compile`
- `./mill integration.feature.test-task-batch`
- `./mill libs.javalib.test mill.javalib.TestModuleUtilTests`

Regression fixture:
- fake sbt-testing framework fails if `forkEnv` is missing in `Runner.tasks(...)`
- fails if resource and suite `TaskDef`s are split
- opts in with `testBatchFrameworkTasks = true`

### DIFF
--- a/integration/feature/test-task-batch/resources/app/test/src/com/example/BatchFramework.scala
+++ b/integration/feature/test-task-batch/resources/app/test/src/com/example/BatchFramework.scala
@@ -1,0 +1,118 @@
+package com.example
+
+import sbt.testing.Event
+import sbt.testing.EventHandler
+import sbt.testing.Fingerprint
+import sbt.testing.Framework
+import sbt.testing.Logger
+import sbt.testing.OptionalThrowable
+import sbt.testing.Runner
+import sbt.testing.Selector
+import sbt.testing.Status
+import sbt.testing.SubclassFingerprint
+import sbt.testing.SuiteSelector
+import sbt.testing.Task
+import sbt.testing.TaskDef
+
+trait BatchSuite
+trait BatchResource
+
+object SharedResource extends BatchResource
+object BatchSuiteTest extends BatchSuite
+
+class BatchFramework extends Framework {
+  def name(): String = "batch-framework"
+
+  def fingerprints(): Array[Fingerprint] = Array(
+    BatchFramework.ResourceFingerprint,
+    BatchFramework.SuiteFingerprint
+  )
+
+  def runner(
+      args: Array[String],
+      remoteArgs: Array[String],
+      testClassLoader: ClassLoader
+  ): Runner = new BatchRunner(args, remoteArgs)
+
+  def slaveRunner(
+      args: Array[String],
+      remoteArgs: Array[String],
+      testClassLoader: ClassLoader,
+      send: String => Unit
+  ): Runner = new BatchRunner(args, remoteArgs)
+}
+
+object BatchFramework {
+  object SuiteFingerprint extends SubclassFingerprint {
+    def isModule(): Boolean = true
+    def requireNoArgConstructor(): Boolean = true
+    def superclassName(): String = "com.example.BatchSuite"
+  }
+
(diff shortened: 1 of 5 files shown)


lihaoyi:
Thanks @nguyenyou , kicked off a dev release here for you to try out https://github.com/com-lihaoyi/mill/actions/runs/26208549926

PR #7121, opened 2026-05-19
Contrib: Document how to Owasp Dependency Check Version

gamlerhart:
- Document how to override the version, by creating your own worker
- Turn val -> def to avoid any initialization weirdness


contrib/owaspdependencycheck/src/mill/contrib/owaspdependencycheck/OwaspDependencyCheckModule.scala
```diff
@@ -28,7 +28,7 @@ trait OwaspDependencyCheckModule extends Module, OfflineSupportModule {
   /**
    * The worker is version specific. If another version is required it can be customized here.
    */
-  val owaspDependencyCheckWorker: ModuleRef[OwaspDependencyCheckWorker] =
```
gamlerhart, line 31: I'm always nervous to override. So I made it a more conservative def.

lefou (approved):
Thank you!

PR #7123, opened 2026-05-20
Update autofix-ci/action to 1.3.4

lefou:
Switch from cryptic commit hash-based to release tag-based version.


lefou:
Supersedes https://github.com/com-lihaoyi/mill/pull/7030 which uses a commit hash instead of a release tag.

PR #7128, opened 2026-05-21
fix(kotlinlib): remove stale Kotlin CLI class files

### LINKED ISSUE #7126: Kotlin compilation leaves stale class files behind
See
- #7125 

this is particularly bad when renaming/deleting tests, it effectively poisons the cache.

nguyenyou:
TLDR: Track Kotlin CLI output files and remove the previous Kotlin outputs before recompiling, so deleted Kotlin sources cannot leave stale `.class` files that still get discovered as tests.

This uses `-Xreport-output-files` for Kotlin CLI versions that support it. Kotlin documents compiler option metadata via the [compiler-options schema](https://kotlinlang.org/docs/compiler-reference.html#schema-for-compiler-options); the `kotlin-compiler-arguments-description` schema lists `Xreport-output-files` as reporting source-to-output mappings, introduced in Kotlin 1.1.3. Older CLI versions fall back to a clean classes dir.

The Kotlin Build Tools API path is unchanged because Kotlin IC already handles stale output cleanup.

Closes #7126

Validation:
- `./mill --no-server libs.kotlinlib.test mill.kotlinlib.HelloKotlinTests`
- `./mill --no-server libs.kotlinlib.test 'mill.kotlinlib.MixedHelloWorldTests.failures'`
- Re-ran the external stale-test MRE with locally built Mill

### DIFF
--- a/libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
+++ b/libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
@@ -340,7 +340,16 @@ trait KotlinModule extends JavaModule with KotlinModuleApi { outer =>
           s"Compiling ${kotlinSourceFiles.size} Kotlin sources ${extra}to ${classes.relativeTo(BuildCtx.workspaceRoot)} ..."
         )
 
-        val compilerArgs: Seq[String] = Seq(
+        val supportsCliOutputTracking =
+          Version.parse(kotlinVersion())
+            .isAtLeast(Version.parse("1.1.3"))(using Version.IgnoreQualifierOrdering)
+
+        val useBtApi =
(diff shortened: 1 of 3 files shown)


viluon:
Does this negatively impact incremental compilation with Kotlin CLI?

nguyenyou:
Good catch. The first version did make every invalidated Kotlin CLI compile start from a clean classes dir.

I revised it to use Kotlin CLI’s `-Xreport-output-files`: Mill records the Kotlin outputs from each successful CLI compile, and before the next CLI compile removes only those previously recorded Kotlin outputs. Kotlin CLI itself only reports the outputs; Mill does the cleanup.

That should avoid stale class files without wiping the whole output directory. The Build Tools API path is unchanged because Kotlin IC already handles this cleanup.

PR #7131, opened 2026-05-22
Refine test class fingerprinting - exclude local classes

### LINKED ISSUE #7094: ScalaTest discovery should ignore inner classes
I'm testing extensions to ScalaTest (using ScalaTest) with suites such as:
```scala
class AssertionsTests extends AnyFunSuite:
   test("assertExpr"):
      class Tests extends AnyFunSuite:
         test("success"):
            assertExpr(Seq(1, 2)):
               List(1) ::: List(2)
         test("failed"):
            assertExpr(42, ": not good"):
               40 + 1
      end Tests

      val suite = Tests()
      assert(suite.run(Some("success"), silent).succeeds())
      assert(!suite.run(Some("failed"), Args(R)).succeeds())
      R.lastEvent match
         case Some(ev: TestFailed) => assert(ev.message == "Expected 41, but got 42 : not good")
         case other                => fail(s"unexpected: $other ")
```
The Mill task `test` runs `AssertionsTests`, as it should, but also the inner class `Tests` (and counts the test `"failed"` as a failure).
Adding the setting `def discoverTestsWithZinc = true` solves the problem, but I think inner classes should always be ignored during discovery.
(Maybe it's a ScalaTest (or `sbt`) issue, but I find it interesting that `discoverTestsWithZinc` has an impact.)

lefou:
Added a new test case adapted from issue comment https://github.com/com-lihaoyi/mill/issues/7094#issuecomment-508575575, reproducing the issue with false detection of a local class.

Refined test class fingerprinting. Especially detect local classes, which are defined inside other methods. Those should not be counted as tests.

Fix https://github.com/com-lihaoyi/mill/issues/7094

### DIFF
--- a/libs/javalib/testrunner/src/mill/javalib/testrunner/TestRunnerUtils.scala
+++ b/libs/javalib/testrunner/src/mill/javalib/testrunner/TestRunnerUtils.scala
@@ -56,13 +56,24 @@ import scala.math.Ordering.Implicits.*
           listClassFiles(base).map { path =>
             val cls = cl.loadClass(path.stripSuffix(".class").replace('/', '.'))
             val publicConstructorCount =
-              cls.getConstructors.count(c => Modifier.isPublic(c.getModifiers))
+              cls.getConstructors.count(c => Modifier.isPublic(c.getModifiers()))
 
             if (framework.name() == "Jupiter") {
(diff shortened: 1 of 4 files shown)


lefou:
Looks like mockito integration tests isn't happy with the changes...

PR #7133, opened 2026-05-24
Fixed mappings for ErrorProneModule in auto-migration

### LINKED ISSUE #6827: Maven migration issues
Hi

We tried to migrate our Maven project to Mill system and found two small issues. Here's original block of Maven configuration:

```
<compilerArgs>
	<arg>-XDcompilePolicy=simple</arg>
	<arg>-Xplugin:ErrorProne -Xep:EmptyBlockTag:OFF  -Xep:JavaTimeDefaultTimeZone:OFF</arg>
	<arg>--should-stop=ifError=FLOW</arg>						
</compilerArgs>
```

And Mill converted that into:
```
javacOptions: [--release, 21, --should-stop=ifError=FLOW]
errorProneOptions: ["-Xep:EmptyBlockTag:OFF", , "-Xep:JavaTimeDefaultTimeZone:OFF"]
```

1) Error #1 - **--should-stop** is not Java compiler option, it's Error Prone option.
2) Error #2 - There are two space characters between -Xep:EmptyBlockTag:OFF and -Xep:JavaTimeDefaultTimeZone:OFF. And probably because of that errorProneOptions contains two consecutive commas.

Mill version: 1.1.2

ajaychandran:
- Revised mappings to include unmanaged [back door compiler options](https://github.com/openjdk/jdk/blob/ef0851d8adbb834e1cd5aff5b3b973b953e57e2d/src/jdk.compiler/share/classes/com/sun/tools/javac/main/Option.java#L653) in `javacOptions` instead of `errorProneEnableJavacOptions`.
- Handle consecutive delimiters when parsing whitespace delimited values. This change was applied in other places where this pattern appears.
- For Maven projects,
  - Added support for loading [errorProneOptions from file](https://errorprone.info/docs/flags#maven).
  - Added mappings for `classifier` and `type` attributes for `errorProneMvnDeps`.

Resolves #6827

### DIFF
--- a/integration/migrating/init/resources/golden/gradle/fast-csv
+++ b/integration/migrating/init/resources/golden/gradle/fast-csv
@@ -55,14 +55,18 @@ Module dependencies of lib.test:
     "--release",
     "17",
     "-Xlint:all",
-    "-Werror"
+    "-Werror",
+    "-XDshould-stop.ifError=FLOW",
+    "-XDshouldStopPolicyIfError=FLOW"
   ],
   "lib.test.javacOptions": [
     "-source",
     "25",
     "-target",
     "25",
-    "-parameters"
+    "-parameters",
+    "-XDshould-stop.ifError=FLOW",
+    "-XDshouldStopPolicyIfError=FLOW"
   ]
 }
 null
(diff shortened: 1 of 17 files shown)


lefou:
@ajaychandran Why is it needed to replace the dedicated options with their "back door" version?

ajaychandran:
> @ajaychandran Why is it needed to replace the dedicated options with their "back door" version?

Nothing is being replaced. The backdoor options, ones that start with `-XD`, were being assigned to `errorProneEnableJavacOptions`. This is incorrect  because these are not specific to the Error Prone plugin even though the plugin may be using them. 

This is different from the first point mentioned in the bug report that incorrectly states that `--should-stop` is an Error Prone option. The PR description has been updated to avoid this confusion.

PR #7134, opened 2026-05-24
Android: Use java 21 in examples

souvlakias:
#7076 which bumped to Java 25, correctly updated the Android examples to use Java21 except these two. 
I noticed that their [weekly scheduled CI run](https://github.com/com-lihaoyi/mill/actions/runs/26351968056) failed.


vaslabs:
we have the same issue here https://github.com/vaslabs-ltd/compose-samples-with-mill/

I think in android we need to enforce the version that's compatible

otherwise users will have to add it manually https://github.com/vaslabs-ltd/compose-samples-with-mill/blob/main/build.mill#L1

PR #7152, opened 2026-05-28
Fix Kotlin incremental compilation

### LINKED ISSUE #7148: Kotlin incorrect incremental compilation invalidation
Kotlin incremental compilation doesn't properly invalidate when a module dependency adds a `sealed` inheritance hierarchy member. For the repro, see
- #7152 

This is on
```
Mill Build Tool version 1.1.6-33-43d8dd
java.version: 25.0.2
java.vendor: Azul Systems, Inc.
java.home: /home/viluon/.cache/coursier/arc/https/cdn.azul.com/zulu/bin/zulu25.32.17-ca-jdk25.0.2-linux_x64.tar.gz/zulu25.32.17-ca-jdk25.0.2-linux_x64
file.encoding: UTF-8
os.name: Linux
os.version: 7.0.9
os.arch: amd64
```

viluon:
Fix https://github.com/com-lihaoyi/mill/issues/7148

The original problem was that Mill assumed the BT API would compute a classpath snapshot if it's called with an empty one. This is not the case, it's up to Mill to compute this snapshot for correct incremental compilation.

We store these snapshots for every element of the classpath under a `classpath-snapshots/hashCode` directory in the `KotlinWorkerManager.kotlinWorker`'s `Task.dest`, where `hashCode` is the hash code of the tools classloader. This keeps them separate should there be multiple classloaders with different snapshotting implementations.

The unfortunate side effect is that we don't reuse snapshots across JVM restarts. Optimising this is left for future work.


libs/kotlinlib/worker/src/mill/kotlinlib/worker/impl/JvmCompileBtApiImpl.scala
```diff
@@ -53,6 +54,24 @@ class JvmCompileBtApiImpl() extends Compiler {
...
+    args.sliding(2)
+      .collectFirst {
+        case Seq("-classpath" | "-cp", classpath) => classpath
+      }
+      .toSeq
+      .flatMap(_.split(java.io.File.pathSeparator).iterator)
+      .filter(_.nonEmpty)
+      .map(os.Path(_, ctx.workspace))
+      .filter(os.exists)
+  }
```
lefou: Instead of post-processing a string-ly typed args list, which we constructed ourselves on the caller side, we could just pass the real classpath files to the worker `compile` method. While this is a larger change, it's IMHO the more correct solution.

lefou (commented):
Is there some more information how snapshot files are calculated? If these snapshots are reproducible and only depend on the classpath entry content, creating and storing them in each module sounds like some significant overhead. We already have the `PathRef.sig`s for each classpath entry (in the caller side, e.g. `KotlinModule.compileClasspath`), so we could as well store the snapshot files in the context of the worker and look them up based on their `PathRef.sig`, to avoid rescanning of the same entries over and over. This requires a change to the `KotlinWorker.compile` method, so that it additionally accepts the full classpath.

viluon:
Thanks for the review @lefou, I'll take your advice into account but the patch needs to be reworked anyway, it was written against something closer to 1.1.6 and doesn't work with the new relative path stuff. Also, I think adding the test to `kotlinlib` is wrong, it should be an integration test instead.

lefou (commented):
Thank you for the update. Looks good at a first glance. I will do a proper review but it may take a day longer (I'm a voluntary firefighter and we currently have a large fire in a local incinerator plant).

libs/kotlinlib/api/src/mill/kotlinlib/worker/api/KotlinWorker.scala
```diff
@@ -14,7 +14,9 @@ trait KotlinWorker {
       target: KotlinWorkerTarget,
       useBtApi: Boolean,
       args: Seq[String],
-      sources: Seq[os.Path]
+      sources: Seq[os.Path],
+      classpath: Seq[PathRef] = Nil,
+      classpathSnapshotCache: Option[os.Path] = None
```
lefou: At least the new parameters should have some ScalaDoc.

It looks to me, the `classpathSnapshotCache` should be managed as an implementation detail of the worker. The current design (as parameter) makes it hard to reason about, why and which cache instance should be used. It's also error-prone, there are no guards that prevent mixing up caches from unrelated workers.

libs/kotlinlib/worker/src/mill/kotlinlib/worker/impl/JvmCompileBtApiImpl.scala
```diff
@@ -53,52 +55,95 @@ class JvmCompileBtApiImpl() extends Compiler {
...
+              snapshottingOperation,
+              executionPolicy,
+              kotlinLogger
+            )
+            val tmpFile = os.temp(dir = snapshotCacheDir, suffix = ".snapshot")
+            snapshot.saveSnapshot(PathRef.toAbsNioPath(tmpFile).toFile)
+            os.move(tmpFile, snapshotFile, atomicMove = true, replaceExisting = true)
+          }
+          PathRef.toAbsNioPath(snapshotFile)
+        }
```
lefou: The snapshot file creation and cache logic should probably be moved into a dedicate `private def`.

libs/kotlinlib/worker/src/mill/kotlinlib/worker/impl/Compiler.scala
```diff
@@ -1,11 +1,13 @@
...
-import mill.api.TaskCtx
+import mill.api.{PathRef, TaskCtx}
 
 trait Compiler {
   def compile(
       args: Seq[String],
-      sources: Seq[os.Path]
+      sources: Seq[os.Path],
+      classpath: Seq[PathRef] = Nil,
+      classpathSnapshotCache: Option[os.Path] = None
```
lefou: There is no need to have the `snapshotCache` in each compile call. It should be enough to pass it to the class constructor, if the specific impl does support a cache at all.

libs/kotlinlib/api/src/mill/kotlinlib/worker/api/KotlinWorker.scala
viluon: Please check if the new comments suffice.

libs/kotlinlib/worker/src/mill/kotlinlib/worker/impl/Compiler.scala
viluon: Agreed, now in the constructor of `JvmCompileBtApiImpl`.

libs/kotlinlib/src/mill/kotlinlib/KotlinWorkerManager.scala
```diff
@@ -28,7 +28,7 @@ object KotlinWorkerManager extends ExternalModule {
       ) + ".impl." + classOf[KotlinWorker].getSimpleName() + "Impl"
 
     val impl = toolsClassLoader.loadClass(className)
-    val worker = impl.getConstructor().newInstance().asInstanceOf[KotlinWorker]
+    val worker = impl.getConstructor(classOf[TaskCtx]).newInstance(ctx).asInstanceOf[KotlinWorker]
```
lefou: Using this task context is not correct. `KotlinWorkerManager` is a shared external module and the task context, coming from `KotlinWorkerManager.kotlinWorker`, is identical for different classloaders. This may result in concurrent write access and in the worst case, different workers with different snapshot algorithms overwriting each others result.

I suggest two changes:

1. In `KotlinWorkerImpl`, instead of using a task context, accept an explicit cache path.

2. In `KotlinWorkerManager`, construct a unique cache path per classloader. 

There is a huge downside. There is no stable hash for the classloader, which means, after stopping the worker, all snapshots need to be regenerated. We also pile up unused snapshot files. 

What we want is a unique cache directory derived from the classpath, that is loaded into the classloader, but the currently used types (`ClassLoader`) don't allow a direct access to this information.

lefou: We can either try to match the runtime type of the classlaoder and access the used classpath, or we could register the classpath used to create a classloader in some runtime state. 

It may be ok for this PR to just live with this shortcoming and deal with these optimisations later. It should already perform well for the typical use case (many modules using the same kotlin version and thus the same worker).

libs/kotlinlib/api/src/mill/kotlinlib/worker/api/KotlinWorker.scala
lefou: They do, thanks.

libs/kotlinlib/src/mill/kotlinlib/KotlinWorkerManager.scala
viluon: I agree this is bad, though concurrent writes should be safe as long as all implementations agree (the snapshot is generated in a temporary file that's moved atomically to the target location). But I realize this is insufficient if there are different snapshot algorithms at play.

Could we tackle point 2 with a heuristic solution? E.g. query the Kotlin version in the classloader or something and use that as the key? I would like to preserve classpath snapshots over worker restarts, notably for CI.

lefou: Looks like we already have the classpath, but the `ClassLoaderCachedFactory.setup` method is abstracting it away. Maybe we can access it by overriding `setup`.

viluon: how about making `classloaderCache` `protected`? I would like to add the `key` to the `getValue` call here:
```scala
override def setup(key: Seq[PathRef]) = {
  val cl = classloaderCache.getOrCreate(key, e)
  val bridge = getValue(cl)

  bridge
}
```

viluon: Fixed in d4415c5bd563034f2d593c2f8b57f3bd0d122a0d, please have a look.

libs/kotlinlib/worker/src/mill/kotlinlib/worker/impl/JvmCompileBtApiImpl.scala
```diff
@@ -53,52 +63,67 @@ class JvmCompileBtApiImpl() extends Compiler {
...
+        PathRef.toAbsNioPath(destinationDirectory)
+      )
 
     compilationOperation.getCompilerArguments().applyArgumentStrings(args.asJava)
 
     val snapshotIcOptions = compilationOperation.createSnapshotBasedIcOptions().tap { options =>
       options.set(
         JvmSnapshotBasedIncrementalCompilationOptions.ROOT_PROJECT_DIR,
-        ctx.workspace.toNIO
+        PathRef.toAbsNioPath(ctx.workspace)
```
lefou, line 93: NOTE: Workaround issue #7207

lefou (commented):
Looks good to me! Can you please update the PR description. 

I'll merge after CI is green.

libs/kotlinlib/worker/src/mill/kotlinlib/worker/impl/JvmCompileBtApiImpl.scala
viluon: should I add that as a comment? It's used in quite a few places and there's already a comment at line 78

lefou: Nope, I just wanted to collect the reference in the other issue.

viluon:
One last thing @lefou, the classpath snapshots get pretty large, dozens to hundreds of MB for large projects. They'd stay in `out/` indefinitely. There's no point in keeping them on disk if the JVM shuts down since the classloader hashes will be different.

I believe 12ec37c1a24cd4cb7fad39d108694b5078938c9c fixes that.

lefou:
> One last thing @lefou, the classpath snapshots get pretty large, dozens to hundreds of MB for large projects. They'd stay in `out/` indefinitely. There's no point in keeping them on disk if the JVM shuts down since the classloader hashes will be different.
> 
> I believe [12ec37c](https://github.com/com-lihaoyi/mill/commit/12ec37c1a24cd4cb7fad39d108694b5078938c9c) fixes that.

External modules don't support clean, but workers do. You need to apply the cleanup in `KotlinWorkerImpl` by implementing `AutoCloseable` and only clear the specific cache sub-path.

lefou:
> > One last thing @lefou, the classpath snapshots get pretty large, dozens to hundreds of MB for large projects. They'd stay in `out/` indefinitely. There's no point in keeping them on disk if the JVM shuts down since the classloader hashes will be different.
> > I believe [12ec37c](https://github.com/com-lihaoyi/mill/commit/12ec37c1a24cd4cb7fad39d108694b5078938c9c) fixes that.
> 
> External modules don't support clean, but workers do. You need to apply the cleanup in `KotlinWorkerImpl` by implementing `AutoCloseable` and only clear the specific cache sub-path.

Sorry, I was wrong. I overlooked that your `close` method is on the `class`, not the `object`.

lefou:
Thank you, @viluon !

PR #7188, opened 2026-06-01
Fix #7118: expose top-level precompiled YAML modules under BuildCtx.rootModule

### LINKED ISSUE #7118: BuildCtx.rootModule returns isolated/dummy root in YAML precompiled module workspaces, breaking cross-module discovery
[#7090](https://github.com/com-lihaoyi/mill/pull/7090/changes) introduced BuildCtx.rootModule so plugins can walk the module tree without threading an Evaluator. The docstring says to cast to mill.api.Module and use moduleInternal.modules to traverse the tree.

This breaks for workspaces built entirely from precompiled package.mill.yaml/build.mill.yaml modules.

### Repro Environment

https://github.com/rkophs/mill-scratch-tests/tree/main

**Layout:**
```
custom_module/                       # publishes com.rkophs:custom-module_3:0.1.0
  src/com/rkophs/mill/SimpleJavaModule.scala
precompiled_tests/
  build.mill.yaml
  foo/package.mill.yaml               # extends com.rkophs.mill.SimpleJavaModule
  bar/package.mill.yaml               # extends com.rkophs.mill.SimpleJavaModule
```

**SimpleJavaModule.scala:**
```scala
class SimpleJavaModule(val scriptConfig: PrecompiledModule.Config)
    extends JavaModule with PrecompiledModule {

  override lazy val millDiscover = mill.api.Discover[this.type]

  lazy val allModuleSegments =
    BuildCtx.rootModule.asInstanceOf[Module]
      .moduleInternal.modules.map(_.moduleSegments)

  def listModules = Task { allModuleSegments.map(_.render) }

  override def moduleDeps: Seq[JavaModule] = {
    def sibling(name: String): Seq[JavaModule] =
      BuildCtx.rootModule
        .asInstanceOf[Module]
        .moduleInternal
        .modules
        .collect { case m: SimpleJavaModule if m.moduleSegments.last.value == name => m }
        .toSeq

    moduleSegments.last.value match {
      case "bar" => sibling("foo")
      case _ => Seq.empty
    }
  }

  def listModuleDeps = Task {
    moduleDeps.map(_.moduleSegments.render)
  }
}
```

**Steps to reproduce**
```
cd custom_module && mill __.publishLocal
cd ../precompiled_tests && mill show __.listModules && ../mill show __.listModuleDeps
```

**Observed**
```
$ ../mill show __.listModules
1] show
{
  "bar.listModules": [
    "",
    "selective"
  ],
  "foo.listModules": [
    "",
    "selective"
  ]
}
1/1, SUCCESS] ../mill show __.listModules

$ ../mill show __.listModuleDeps
1] show
{
  "bar.listModuleDeps": [],
  "foo.listModuleDeps": []
}
1/1, SUCCESS] ../mill show __.listModuleDeps 1s
```

**Expected**

Something like `["foo", "bar", "selective"]` — a workspace-wide view. Each precompiled module's `BuildCtx.rootModule` points at its own isolated root, so siblings are invisible.

**Additional diagnosis**

We tried `mill.api.internal.PrecompiledModuleRef` as an alternative cross-module discovery surface, expecting `PrecompiledModuleRef.cache: mutable.Map[os.Path, Module]` to give a workspace-wide view of all loaded precompiled modules.

**Two friction points:**

1. `private[mill]` visibility. Both `PrecompiledModuleRef.cache` and `PrecompiledModuleRef.resolveModuleRef` are marked `private[mill]`. They cannot be called from a third-party module like `SimpleJavaModule` directly — even compiling against them produces cache can only be accessed from package mill.
2. I tried it with reflection temporarily to keep testing. Reaching the cache via `Class.forName("mill.api.internal.PrecompiledModuleRef$").getMethod("cache").invoke(...)` compiles and runs, but the cache is empty by the time tasks are evaluated, and also by the time overrides like `moduleDeps` resolve during the bootstrapping phase. From inspection of the bytecode docstring and behavior, the cache is only populated when one module's YAML configModuleDeps trigges `PrecompiledModuleRef.apply` (the codegen path). Sibling top-level precompiled modules with no explicit YAML cross-reference never register, so there's nothing to discover via the cache.

----
Would it be possible to either:

1. Have BuildCtx.rootModule in YAML-only precompiled workspaces point at a real root that contains all top-level precompiled modules as children, OR
2. Provide a separate API for cross-precompiled-module enumeration.

rkophs:
## Summary

Fixes [#7118](https://github.com/com-lihaoyi/mill/issues/7118).

When a workspace's only build configuration is precompiled YAML modules (i.e. every top-level `*.mill.yaml` declares `mill-experimental-precompiled-module: true`), codegen produces no `build_.package_`, so `BuildFileCls` falls back to `DummyModule` — and on `main` today, that fallback has no children at all. `mill resolve _` lists none of the user's precompiled modules, and `BuildCtx.rootModule.moduleInternal.modules` only sees the synthetic root.

This PR makes those depth-1 precompiled YAMLs reachable as direct children of the root module, in both the "no compiled root" case and the "compiled root mixed with precompileds" case.

The implementation has two halves:

1. **Single source of truth for discovery.** `CodeGen` already walks the workspace via `BuildFileDiscovery.walkBuildFiles` and parses every YAML header to build `parsedYamlHeaderData` / `precompiledModulePaths` (`runner/meta/src/mill/meta/CodeGen.scala`). It now also identifies "orphan" precompileds — depth-1 precompiled YAMLs with no compiled sibling at root that would otherwise codegen-alias them — and serializes them into a classpath resource `mill/precompiled-modules.json`, alongside the existing `mill/module-deps-config.json`. No new workspace walker, no runtime filesystem probe.

2. **Additive append on `MainRootModule`.** `MainRootModule` (the common parent of every generated root `package_` and the `DummyModule` fallback) mixes in `DynamicModule` and overrides `moduleDirectChildren` to `super ++ orphanPrecompiledChildren`. Orphans are read from the JSON resource and instantiated via the existing `PrecompiledModuleRef.apply` path, with sibling `moduleDeps` refs resolved through a name → entry map that shares `PrecompiledModuleRef.cache` and its cycle-detection contract. The fallback `DummyModule` shrinks back to a one-line `extends MainRootModule`; it inherits the append automatically. So:
   - **No compiled root:** `DummyModule` is loaded, gets the orphans appended → siblings show up under `resolve _`.
   - **Compiled root:** the codegen-emitted `build_.package_` is loaded; its codegen aliases handle depth-1 precompileds, and the orphan list is empty → behavior unchanged.

A small `PrecompiledModule.all` API is added on `mill.api` for plugin/user code that wants the discovered set without the `BuildCtx.rootModule.asInstanceOf[Module].moduleInternal.modules.collect { case m: PrecompiledModule => m }` dance.

## Why the orphan resource is depth-1 only

`ResolveCore` materializes `DynamicModule` children using `Segments.labels(child.moduleSegments.last.value)` (`core/resolve/src/mill/resolve/ResolveCore.scala:519`) — so for a child whose `moduleSegments` are e.g. `["foo", "bar"]`, only `"bar"` survives the resolve handoff. Depth-2+ orphans would therefore render as the wrong name (and collide with depth-1 siblings of the same last segment). `CodeGen` filters them out at the source rather than letting them be silently mis-attached. Supporting deeper orphans cleanly would require synthesizing intermediate `DynamicModule` wrappers per missing path segment — a deliberate follow-up, called out in code comments.

## Known pre-existing limit not addressed here

A precompiled root yaml (`mill-experimental-precompiled-module: true` on `build.mill.yaml` itself) is silently dropped on `main` today: codegen excludes it from the main loop, `BuildFileCls` falls back to `DummyModule`, and the yaml's `extends:`/`moduleDeps:`/nested object blocks are never honored. This PR doesn't change that behavior — depth-1 sibling precompileds are now visible (which they weren't on `main`), but a root precompiled yaml's own config is still ignored. Worth surfacing as a loud error in a follow-up.

## Tests

- `integration.feature.precompiled-root-codegen` (5 sub-tests): all-precompiled workspace — root + depth-1 siblings, cross-sibling `moduleDeps`, `BuildCtx.rootModule` walk, `PrecompiledModule.all` discovery.
- `integration.feature.precompiled-moduledeps-override` (2 sub-tests): programmatic `moduleDeps` override walking `BuildCtx.rootModule`.
- `integration.feature.precompiled-nested-relationships` (3 sub-tests, new): mixed-tree workspace covering codegen-alias path at depth (precompiled grandchild under compiled `package.mill`) and cross-shape `moduleDeps` (precompiled YAML depending on a compiled Scala sibling resolved via `PrecompiledModuleRef.resolveModuleRef` on `build_.package_`).
- `./mill __.compile` clean.

## Test plan

- [ ] CI passes all precompiled-related tests (`precompiled-root-codegen`, `precompiled-moduledeps-override`, `precompiled-nested-relationships`).
- [ ] CI passes the `example.large.precompiled.10-precompiled-modules` example test (compiled-root + precompiled siblings + cross-sibling deps), which exercises the codegen-alias side of the augmented `MainRootModule.moduleDirectChildren`.
- [ ] No regressions in the broader `__.compile` / non-precompiled `example.*` suites.


rkophs:
@lihaoyi let me know your thoughts on this 😄

PR #7189, opened 2026-06-02
Always clean the classes dir when using the Kotlin CLI compiler

### LINKED ISSUE #7126: Kotlin compilation leaves stale class files behind

lefou:
Since the compile task is now persistent and does not clean itself before the compiler runs,
we need to clean the classes destination dir explicitly.

This does not affect the use of the BT-API, which already tracks its inputs and outputs and is supposed to cleanup stale files.

Fix https://github.com/com-lihaoyi/mill/issues/7126

### DIFF
--- a/libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
+++ b/libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
@@ -305,8 +305,8 @@ trait KotlinModule extends JavaModule with KotlinModuleApi { outer =>
       val classes = dest / "classes"
 
       val useBtApi = kotlincUseBtApi() && kotlinUseEmbeddableCompiler()
-      if(!useBtApi) {
-        // Non BT-API compiler is not incremental and does not keep track of older files, 
+      if (!useBtApi) {
+        // Non BT-API compiler is not incremental and does not keep track of older files,
         // so we always need to start fresh.
         os.remove.all(classes)
       }
--- a/libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
+++ b/libs/kotlinlib/src/mill/kotlinlib/KotlinModule.scala
@@ -303,6 +303,13 @@ trait KotlinModule extends JavaModule with KotlinModuleApi { outer =>
       val ctx = Task.ctx()
       val dest = ctx.dest
       val classes = dest / "classes"
+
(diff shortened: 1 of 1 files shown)


viluon:
As with https://github.com/com-lihaoyi/mill/pull/7128 my question is how does this affect incremental compilation performance?

lefou:
Using the CLI compiler is always slow. Maybe, it is somewhat faster when you don't clear the destination dir, but then you have to do extra work to make sure it is correct, which we currently don't. Any real incremental compilation support is only accessible via the BT API, which is not affected by this PR.

viluon:
I have not experienced a significant compilation speedup after switching to the BT API but I concede I didn't benchmark this carefully. Are you sure the CLI compiler doesn't support incremental compilation at all?

The issue of clearing stale classes really only comes up with classpath scanning, which is somewhat rare. A notable exception is unit tests, which can be selectively cleaned beforehand.

I think this change is ultimately correct, I'm just afraid that it leaves Mill Kotlin users with no real option for incremental compilation (see #7148).

lefou:
> I have not experienced a significant compilation speedup after switching to the BT API but I concede I didn't benchmark this carefully. Are you sure the CLI compiler doesn't support incremental compilation at all?

It's not the speedup by compiling stuff but by not-compiling stuff. The CLI compiler always compiles all files, which is a huge difference when you have lots of source files. (E.g. 60+ seconds to compile a single module.) Esp. after small edits where the BT-API could choose to compile just a single file. 

> The issue of clearing stale classes really only comes up with classpath scanning, which is somewhat rare. A notable exception is unit tests, which can be selectively cleaned beforehand.

That's not quite correct. While it surfaces in classpath scanning and testing, stale files is always an issue, e.g. those can end up in released JARs. A API refactoring can get completely unnoticed by a downstream user while upstream isn't even aware of the old API still being present. Later it suddently breaks, just because some release-pipeline cleaned its caches.

> I think this change is ultimately correct, I'm just afraid that it leaves Mill Kotlin users with no real option for incremental compilation (see #7148).

Now, that is the consequence of choosing a vendor-driven ecosystem like Kotlin. The initial incremental compilation support in Kotlin 1 was never exposed to users outside IDEA and Gradle. I'm not even sure there would be some public API yet, if not driven by demand from external folks like  Bazel or Mill. And it still feels like a stepchild that has teething troubles.

PR #7196, opened 2026-06-03
Add `--replay-logs` option to show logs of cached tasks

### LINKED ISSUE #7193: Add a new `--replay-logs` option to output the logs of cached tasks
With this enabled, a mill build run on a cached build should be visibly more like a fresh build, except being faster. This makes task caching feature more transparent. It helps especially with remote-caching to get a better feeling of the current state of the build. If the developer never sees a recent build output (since all comes from a remote cache), they may not be aware of any warnings a task might show.

I'm not aware of some precedent feature in other tools, so we need to find a good name for it.

My first thought was `--replay-logs` or just `replay`. WDYT?

lefou:
If using the new `--replay-logs` option, the persistent logs of cached tasks will be printed to STDERR. This make task caching more transparent -- the visual output of Mill then also contains the output of cached tasks and looks identical to a fresh build.

It helps especially with remote-caching to get a better feeling of the current state of the build. If the developer never sees a recent build output (since all comes from a remote cache), they may not be aware of any warnings a task might show.

Technically, we could differentiate between locally and remotely cached tasks, and `--replay-logs` could accept a second argument like `none, all, local, remote`. But I'm not sure we want/need this, so I didn't implemented it for now.

Tested manually with and without remote caching (to a local directory).

Fix https://github.com/com-lihaoyi/mill/issues/7193

Pull request: https://github.com/com-lihaoyi/mill/pull/7196

### DIFF
--- a/core/exec/src/mill/exec/GroupExecution.scala
+++ b/core/exec/src/mill/exec/GroupExecution.scala
@@ -409,6 +409,11 @@ trait GroupExecution {
               }
             )
           }
+          
+          // when requested, forward the cached logs to STDERR
+          def doReplayLog() = if (replayLogs && os.exists(paths.log)) {
+            os.read.stream(paths.log).writeBytesTo(logger.streams.err)
+          }
 
           taskLocks.readThenWrite { scope =>
             loadCachedOrWorker(
@@ -417,11 +422,7 @@ trait GroupExecution {
             ) match {
               case Some(res) =>
                 taskLocks.retainRead(scope)
-
-                if (replayLogs && os.exists(paths.log)) {
-                  os.read.stream(paths.log).writeBytesTo(logger.streams.err)
-                }
-
+                doReplayLog()
                 LockUpgrade.Decision.Complete(res)
               case None =>
                 LockUpgrade.Decision.Escalate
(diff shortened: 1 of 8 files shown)


core/exec/src/mill/exec/GroupExecution.scala
```diff
@@ -470,7 +472,13 @@ trait GroupExecution {
                   if (remoteMaterialized) taskLocks.markTaskWritten()
                   else taskLocks.currentVersion
                 taskLocks.retainDowngraded(scope, version)
+
+                if(replayLogs && os.exists(paths.log)) {
+                  os.read.stream(paths.log).writeBytesTo(logger.streams.err)
+                }
```
lefou: This works for freshly retrieved remote cached tasks, but not once they are locally cached. Different from what I expected.

website/docs/modules/ROOT/pages/cli/flags.adoc
```diff
@@ -83,13 +83,15 @@ Options:
...
   --no-daemon            Run without a long-lived background daemon.
+  --replay-logs          Replay logs for cached tasks.
   --ticker <bool>        Enable or disable the ticker log, which provides information on running
                          tasks and where each log line came from
   -v --version           Show mill version information and exit.
   -w --watch             Watch and re-run the given tasks when their inputs change.
   task <str>...          The name or a query of the tasks(s) you want to build.
 
 See documentation at https://mill-build.org for more details
+
```
lefou:
```suggestion
```

lefou:
CI now green except one job timing out without errors, which looks unrelated.

PR #7200, opened 2026-06-04
`mill init`: Pick up Annotation Processors

### LINKED ISSUE #7199: `mill init` Pick up Annotation Processors
Currently `mill init` doesn't pick up annotation processor deps explicitly. 
In reality, the logic is there in maven migration ([Plugins.scala](https://github.com/com-lihaoyi/mill/blob/f398ed287763be01d433de05ccc0d0d953cce939/libs/init/maven/src/mill/main/maven/Plugins.scala)), in order to pick up Error Prone Deps, but aren't passed as `annotationProcessorsMvnDeps`

[Example](https://github.com/com-lihaoyi/mill/blob/f398ed287763be01d433de05ccc0d0d953cce939/example/migrating/javalib/6-maven-quarkus/build.mill#L11) where it's needed

souvlakias:
To close #7199

#### Maven
Reworked `Plugins.scala` a bit.
##### The ErrorProne *problem*  
Consider the following `pom.xml` segment:
```xml
<configuration>
    <compilerArgs>
        <arg>-Xplugin:ErrorProne</arg>
        <arg>-XepOpt:NullAway:AnnotatedPackages=com.enterprise.app</arg>
    </compilerArgs>
    <annotationProcessorPaths>
        <path>
            <!-- Core errorProneDep -->
            <groupId>com.google.errorprone</groupId> 
            <artifactId>error_prone_core</artifactId>
            <version>2.31.0</version> <!-- We want this -->
        </path>
        <path>
            <!-- Additional errorProneDep -->
            <groupId>com.uber.nullaway</groupId> 
            <artifactId>nullaway</artifactId>
            <version>0.11.0</version>
        </path>
        <path>
            <!-- Any other annotation processors -->
            <groupId>org.projectlombok</groupId> 
            <artifactId>lombok</artifactId>
            <version>1.18.24</version>
        </path>
    </annotationProcessorPaths>
</configuration>
```
Ideally we need to find out which of these are error prone related and override [errorProneDeps](https://github.com/com-lihaoyi/mill/blob/main/libs/javalib/src/mill/javalib/errorprone/ErrorProneModule.scala#L65) with only them. The existing logic picked up all annotation processors it could find.

The solution in this PR, only extracts the `com.google.errorprone:error_prone_core` from `allAnnotationProcessors` as the override of `errorProneDeps` (this is important for its version) 
And since [ErrorProneModule](https://github.com/com-lihaoyi/mill/blob/main/libs/javalib/src/mill/javalib/errorprone/ErrorProneModule.scala#L76) already passes other annotation Processors in the classpath it shouldn't be a problem if other Error Prone processors (e.g. `com.uber.nullaway:nullaway`) are in there.

#### Gradle
For Gradle, things are simple since the processors are distinguished by `annotationProcessor(...)` and `errorprone(...)`.


libs/init/maven/src/mill/main/maven/Plugins.scala
```diff
@@ -20,42 +25,31 @@ class Plugins(model: Model) {
...
-            `type` = _type,
-            excludes = excludes
-          )
-        }
-    } yield spec.withErrorProneModule(errorProneMvnDeps = epMvnDeps, errorProneOptions = epOptions)
+  def errorProneMvnDeps: Seq[MvnDep] =
+    allAnnotationProcessors.filter(isErrorProneCore)
+
+  def annotationProcessorsMvnDeps: Seq[MvnDep] =
+    allAnnotationProcessors.filterNot(isErrorProneCore)
```
vaslabs, line 40: This is very limiting, perhaps we can have 2 methods with mandatoryAnnotationProcessors that is used by mill plugins and an annotationProcessorsMvnDeps that's the current and used by users. This pattern I've seen elsewhere with mandatoryMvnDeps for example

souvlakias: I've updated the PR's description so the rationale on the filtering of errorProne is more clear

vaslabs (approved):

Issue #7207, opened 2026-06-10
Mill breaks `os.Path` guarantee to always contain a absolute path


lefou:
See [documentation of `os.Path`](https://github.com/com-lihaoyi/os-lib#os-path).

> OS-Lib uses strongly-typed data-structures to represent filesystem paths. The two basic versions are:
>
> *   [os.Path](https://github.com/com-lihaoyi/os-lib#os-path): an absolute path, starting from the root
>
> *   [os.RelPath](https://github.com/com-lihaoyi/os-lib#os-relpath): a relative path, not rooted anywhere
>
> *   [os.SubPath](https://github.com/com-lihaoyi/os-lib#os-subpath): a sub path, without any .. segments, not rooted anywhere

The issue results in instances of `os.Path` that return a relative path from `toString`. 

Here are some reports being a result of this issue:

* #7206 
* https://github.com/com-lihaoyi/mill/pull/7204#discussion_r3369275991
* https://github.com/com-lihaoyi/mill/issues/7285
* https://github.com/com-lihaoyi/mill/issues/7299
* https://github.com/com-lihaoyi/mill/issues/7354

lefou:
I think, we should revert this situation completely. While this can be worked around, it's neither obvious to a user nor does it results in user trust against Mill or os-lib.

lihaoyi:
I think we're going to have to go ahead with this in the interest of remote caching and reproducible builds. I agree it's gross and weird and not ideal, but fundamentally OS-Lib's every-path-is-absolute data model from 2018 is completely at odds with the idea of reproducible builds and remote caching. The other alternatives are to 

1. Never have reproducible builds in Mill 
2. Have every task manually serialize things with special helpers to support reproducible builds
3. Use a completely different filesystem library, or overhaul OS-Lib to the point of un-recognizability

None of these are great options, and I think living with some weirdness around path serialization is probably the least-bad option out of all of these

lefou:
I think those features (reproducible builds, remote caching) are unrelated to whether `os.Path` properly represent absolute paths or not. 

The blurred `os.Path` semantics is a property of the current implementation, which was merged without significant review feedback. But it is IMHO not necessary.

Instead of softening `os.Path` and make it worthless, we should introduce a new `ProjectPath` type that holds a project-relative path, that is easily convertible to and from `os.Path`. The JSON serializers used by Mill can explicitly use `ProjectPath` while in code there is no longer a doubt which path representation is used.

[...]

PR #7286, opened 2026-06-17
fix(javalib): resolve relative Zinc product ids

### LINKED ISSUE #7285: Mill 1.2.0-RC1: JVM compile crash when Zinc product id is a relative/aliased path
## TLDR

Mill `1.2.0-RC1` can crash during JVM compile when Zinc incremental annotation-processing persistence sees a product id like:

```text
mill-workspace/out/app/compile.dest/classes/com/example/Foo.class
```

`IncrementalAnnotationProcessing.persist` passes that directly to `os.Path(...)`, but `os.Path` requires an absolute path.

## Error

```text
java.lang.IllegalArgumentException: requirement failed:
mill-workspace/out/app/compile.dest/classes/com/example/Foo.class is not an absolute path

  at os.Path.<init>(Path.scala:707)
  at os.Path$.apply(Path.scala:630)
  at mill.javalib.zinc.IncrementalAnnotationProcessing$.$anonfun$8(IncrementalAnnotationProcessing.scala:203)
  at mill.javalib.zinc.IncrementalAnnotationProcessing$.persist(IncrementalAnnotationProcessing.scala:202)
  at mill.javalib.zinc.ZincWorker.compileInternal(ZincWorker.scala:554)
```

## What seems to be happening

Mill/Zinc can now produce reproducible product ids using path aliases such as:

```text
mill-workspace/out/...
../mill-workspace/out/...
BASE/out/...
```

But `IncrementalAnnotationProcessing.persist` treats `product.id` as an absolute filesystem path.

So the failure is basically:

```text
alias path  -> os.Path(...)
              expects absolute path
              crashes
```

## Expected behavior

Mill should resolve relative/aliased product ids back under the compile `classesDir` before converting them to `os.Path`.

## Validation

I patched Mill locally to resolve non-absolute Zinc product ids relative to `classesDir`.

With that patched Mill executable:

```text
mill show app.jvm.zincIncrementalCompilation
# true

mill app.jvm.compile
# SUCCESS
```

So this does not require disabling Zinc incremental compilation.

nguyenyou:
## TLDR

Fixes #7285 by resolving relative/aliased Zinc product ids under `classesDir` before converting them to `os.Path`.

## What changed

- Add a regression test for a Zinc product id shaped like `mill-workspace/out/.../classes/example/Foo.class`
- Resolve non-absolute analysis product ids relative to the compile classes directory
- Preserve existing behavior for absolute product ids

## Test

```text
./mill libs.javalib.worker.test.testOnly mill.javalib.zinc.IncrementalAnnotationProcessingTrackerTests
```

The new test fails before the fix with `is not an absolute path`, and passes after the fix.

### DIFF
--- a/libs/javalib/worker/src/mill/javalib/zinc/IncrementalAnnotationProcessing.scala
+++ b/libs/javalib/worker/src/mill/javalib/zinc/IncrementalAnnotationProcessing.scala
@@ -4,7 +4,7 @@ import mill.api.daemon.Logger
 import sbt.internal.inc.Analysis
 import xsbti.{PathBasedFile, VirtualFile, VirtualFileRef}
 import java.io.File
-import java.nio.file.Path
+import java.nio.file.{Path, Paths}
 import java.util.Optional
 import java.util.jar.JarFile
 import javax.annotation.processing.Processor
@@ -200,7 +200,7 @@ private[mill] object IncrementalAnnotationProcessing {
       tracker: CompileTracker
   ): Unit = {
     val managedProducts = analysis.relations.allProducts.iterator.flatMap { product =>
-      val path = os.Path(product.id)
+      val path = analysisProductPath(product.id, classesDir)
       Iterator.single(path) ++ auxiliaryClassFileExtensions.iterator.map { ext =>
         path / os.up / s"${path.last.stripSuffix(".class")}.$ext"
       }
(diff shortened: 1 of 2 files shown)


libs/javalib/worker/src/mill/javalib/zinc/IncrementalAnnotationProcessing.scala
```diff
@@ -230,6 +230,35 @@ private[mill] object IncrementalAnnotationProcessing {
     )
   }
 
+  private def analysisProductPath(productId: String, classesDir: os.Path): os.Path = {
+    val productNio = Paths.get(productId).normalize()
+    if (productNio.isAbsolute) os.Path(productNio)
+    else {
+      val productSegments = productNio.iterator().asScala.map(_.toString).toVector
+      val classesSegments = classesDir.wrapped.toAbsolutePath.normalize().iterator().asScala
```
lefou, line 238: This is a code smell, but not your fault. I'm just marking it here, for a later reviewer. See issue #7207.  `os.Path` should be by design always absolute.

lefou:
Thank you!

PR #7289, opened 2026-06-18
contrib/codeartifact: split dotted group into path segments

### LINKED ISSUE #7288: contrib/codeartifact: publishCodeartifact fails with InvalidSegment for any dotted group id
**Mill version:** 1.1.6 (the code is unchanged on `main` as of this writing)

**Module:** `mill-contrib-codeartifact`

### Description

`publishCodeartifact` throws `os.PathError$InvalidSegment` and uploads nothing whenever the module's `pomSettings.organization` (group id) contains a `.` — i.e. essentially every real Maven coordinate (`com.example`, `org.foo`, …).

### Reproduction

1. A module mixing in `CodeartifactPublishModule` with a normal dotted organization:
   ```scala
   object `package` extends ScalaModule with CodeartifactPublishModule {
     def pomSettings = PomSettings(organization = "com.example", /* … */)
     def codeartifactUri = "https://…/maven/my-repo"
     def codeartifactSnapshotUri = codeartifactUri
   }
   ```
2. `mill mymodule.publishCodeartifact --credentials "$CODEARTIFACT_AUTH_TOKEN"`

### Actual

```
[error] mymodule.publishCodeartifact
os.PathError$InvalidSegment: [com/example] is not a valid path segment.
[/] is not a valid character to appear in a non-literal path segment.
  os.BasePath$.fail$1(Path.scala:179)
  mill.contrib.codeartifact.CodeartifactPublisher.basePublishPath(CodeArtifactPublisher.scala:26)
  mill.contrib.codeartifact.CodeartifactPublisher.versionPublishPath(CodeArtifactPublisher.scala:30)
  mill.contrib.codeartifact.CodeartifactPublisher.publishAll(CodeArtifactPublisher.scala:58)
  …
```

### Expected

The artifacts publish to `com/example/<artifact>/<version>/…` in the CodeArtifact Maven repo.

### Root cause

`CodeArtifactPublisher.scala`, `basePublishPath`:

```scala
private def basePublishPath(artifact: Artifact) =
  os.SubPath(Vector(
    artifact.group.replace(".", "/"),  // "com.example" -> "com/example"
    artifact.id
  ))
```

The `os.SubPath(Vector[String])` constructor treats each vector element as a **single** path segment, but `"com/example"` contains `/`, which os-lib rejects. The string is never split into `com` / `example` segments.

### Suggested fix

Use the `os.SubPath(String)` parser, which splits on `/`:

```scala
private def basePublishPath(artifact: Artifact) =
  os.SubPath(artifact.group.replace(".", "/")) / artifact.id
```

Verified: this produces the correct `com/example/<id>/<version>/…` layout and publishes successfully.

raskam:
Fixes #7288.

`CodeartifactPublisher.basePublishPath` builds the publish path as `os.SubPath(Vector(artifact.group.replace(".", "/"), artifact.id))`. The `Vector` constructor treats each element as a single path segment, but `group.replace(".", "/")` produces a string containing `/` (e.g. `com/example`), which os-lib rejects with `InvalidSegment`. As a result `publishCodeartifact` fails for any module whose group id contains a dot — i.e. essentially every real Maven coordinate.

This uses the `os.SubPath(String)` constructor instead, which parses `/` into segments:

```scala
os.SubPath(artifact.group.replace(".", "/")) / artifact.id
```

The resulting path is identical for a single-segment group and correct (`com/example/<id>/…`) for a dotted one. Verified against a live CodeArtifact repository: a module with a dotted `organization` now publishes successfully where it previously threw `InvalidSegment`.

### DIFF
--- a/contrib/codeartifact/src/mill/contrib/codeartifact/CodeArtifactPublisher.scala
+++ b/contrib/codeartifact/src/mill/contrib/codeartifact/CodeArtifactPublisher.scala
@@ -21,10 +21,7 @@ class CodeartifactPublisher(
   )
 
   private def basePublishPath(artifact: Artifact) =
-    os.SubPath(Vector(
-      artifact.group.replace(".", "/"),
-      artifact.id
-    ))
+    os.SubPath(artifact.group.replace(".", "/")) / artifact.id
 
   private def versionPublishPath(artifact: Artifact) =
     basePublishPath(artifact) / artifact.version


lefou:
Thank you!

PR #7306, opened 2026-06-22
Make Kotlin classpath snapshotting persistent

lefou:
Follow up on https://github.com/com-lihaoyi/mill/issues/7148

Use a stable cache path for classpath snapshots, based on the worker classpath.

Change worker cleanup to only auto-remove non-stable snapshot cache paths.

Add an overloaded version of `ClassLoaderCachedFactory.getValue` exposing the actual classpath, since this is what we need to create a stable hash for the classloader. Deprecate the other `getValue` overload to perspectively have only one of them.

Pull request: https://github.com/com-lihaoyi/mill/pull/7306

### DIFF
--- a/libs/util/src/mill/util/ClassLoaderCachedFactory.scala
+++ b/libs/util/src/mill/util/ClassLoaderCachedFactory.scala
@@ -20,7 +20,8 @@ abstract class ClassLoaderCachedFactory[T](jobs: Int)(using e: sourcecode.Enclos
   @deprecated("Use getValue(ClassLoader, Seq[PathRef] instead", "Mill after 1.2.0-RC1")
   def getValue(cl: ClassLoader): T
   // default impl to ensure binary compatibility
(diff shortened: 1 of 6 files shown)


libs/util/src/mill/util/ClassLoaderCachedFactory.scala
```diff
@@ -15,10 +17,15 @@ abstract class ClassLoaderCachedFactory[T](jobs: Int)(using e: sourcecode.Enclos
     sharedPrefixes = Seq("sbt.testing.", "mill.api.daemon.internal.TestReporter")
   )
 
+  @deprecated("Use getValue(ClassLoader, Seq[PathRef] instead", "Mill after 1.2.0-RC1")
```
lefou:
```suggestion
  @deprecated("Use getValue(ClassLoader, Seq[PathRef]) instead", "Mill after 1.2.0-RC1")
```

PR #7310, opened 2026-06-24
Update nativeImageOptions to enable HTTP and HTTPS

lakako:
Some enterprise intranet services still operate over HTTP without HTTPS support.   
This update adds protocol flexibility to ensure image loading works in both HTTP and HTTPS environments.


lefou (approved):
Looks good to me. Thank you!

PR #7311, opened 2026-06-24
Add a note to input tasks about uncached tasks

lefou:


example/fundamentals/tasks/4-inputs/build.mill
```diff
@@ -136,3 +136,13 @@ def myEnvTask = Task {
...
+
+// [NOTE]
+// ====
+// `Task.Input` can't have task dependencies!
+// This is mainly because input tasks can trigger re-evaluation when run with `--watch` or if you run selective execution,
+// and task dependency may make this mechanism slow and in some situations incorrect.
+//
+// If you need to have dependencies in a task that always avaluates but is not cached, you can use xref:cli/tasks.adoc#_uncached_tasks[`Task.Uncached`].
+// `Task.Uncached` is new in Mill 1, since in older versions, input tasks could depend on other tasks.
+// ====
```
lefou, line 151:
```suggestion

// [NOTE]
// ====
// `Task.Input` can't have task dependencies!
// This is mainly because input tasks can trigger re-evaluation when run with
// `--watch` or if you run selective execution, and task dependencies may make
// this mechanism slow and in some situations incorrect.
//
// If you need to have dependencies in a task that always evaluates but is not
// cached, you can use xref:cli/tasks.adoc#_uncached_tasks[`Task.Uncached`].
// `Task.Uncached` was added to Mill 1. In older versions input tasks could
// depend on other tasks.
// ====
```

PR #7313, opened 2026-06-24
Test and fix IDEA `*.iml` file generation for Kotlin (backport #7082)

lefou:
In Kotlin, a module can only access its own `internal` definitions and
`internal` definitions from its "friend" modules, configured by the
`-Xfriend-paths=` compiler option.

Previously, `.iml` files generated by `mill mill.idea/` for
`KotlinIdeaModule`s were incorrect:
- compilation output ended up somewhere in
`internalGenIdea/ideaCompileOutput.dest/classes`, but the
`-Xfriend-paths=` compiler option still referenced
`compile.dest/classes` directories
- IDEA highlighted attempts to access `internal` definitions from friend
modules as errors because the `<additionalVisibleModuleNames>` element
was missing from the IML configuration
- the generated modules defaulted to inheriting the project
configuration, which made any customisations they introduced irrelevant
- `kotlinExplicitApi` was handled inconsistently and was also missing
from the compiler options listed in IML files. I moved it to mandatory
compiler options and removed the ad-hoc handling, making the code
shorter overall.

This PR fixes these issues by introducing the higher-level
`kotlinFriendModules` setting to `KotlinModule`, from which the compiler
options and IDEA metadata are generated. It also adds an integration
test for `KotlinIdeaModule`s.

Backport of pull request: https://github.com/com-lihaoyi/mill/pull/7082

### DIFF
--- a/core/api/daemon/src/mill/api/daemon/internal/IdeUtils.scala
+++ b/core/api/daemon/src/mill/api/daemon/internal/IdeUtils.scala
@@ -1,0 +1,23 @@
+package mill.api.daemon.internal
+
+import mill.api.daemon.{Segment, Segments}
+import mill.api.*
+
+object IdeUtils {
+
+  /**
+   * Create the module name (to be used by an IDE) for the module based on its segments.
+   *
+   * @param path If the  segments yield no result, use the Mill Module folder name from this path.
+   * @see [[Module.moduleSegments]]
+   */
+  def moduleName(p: Segments): Option[String] = Some(p.value
+    .foldLeft(StringBuilder()) {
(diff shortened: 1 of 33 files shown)


viluon:
Idk if this was a good thing to back port, it's adding to the public API, so technically requires a minor version bump per semver.

lefou:
Next minor version `1.2.0` is already taken (e.g. `1.2.0-RC1` already released) by the `main` branch which contains some more disrupting changes such as #7207, so I try to bring some real lib improvements to the stable 1.1.x branch. Although it adds API, it's still backwards binary compatible, so I think this is a ok-ish compromise.

PR #5977, opened 2025-10-15
Generate API docs from all modules (include non-stable modules)

lefou:

PR #7334, opened 2026-06-30
Support Kotlin 2.4 incremental compilation

viluon:
> [!WARNING]
> Assisted by Copilot, Claude Opus 4.8.

In the Kotlin Build Tools API, `createJvmCompilationOperation` has been deprecated ([source](https://github.com/JetBrains/kotlin/blob/fd00198088a7b3a3413750890ae8999dce5d22ae/compiler/build-tools/kotlin-build-tools-impl/src/main/kotlin/org/jetbrains/kotlin/buildtools/internal/jvm/JvmPlatformToolchainImpl.kt#L22-L30), see [KT-81790](https://youtrack.jetbrains.com/issue/KT-81790) for context), making Mill's kotlinlib incompatible with Kotlin 2.4. This PR adopts the new builder interface for 2.4 and above, calling it via reflection.

- [x] once CI passes, mark the PR as `Ready for review`


lefou:
> In the Kotlin Build Tools API, `createJvmCompilationOperation` has been deprecated ([source](https://github.com/JetBrains/kotlin/blob/fd00198088a7b3a3413750890ae8999dce5d22ae/compiler/build-tools/kotlin-build-tools-impl/src/main/kotlin/org/jetbrains/kotlin/buildtools/internal/jvm/JvmPlatformToolchainImpl.kt#L22-L30), see [KT-81790](https://youtrack.jetbrains.com/issue/KT-81790) for context), making Mill's kotlinlib incompatible with Kotlin 2.4. This PR adopts the new builder interface for 2.4 and above, calling it via reflection.

Deprecation isn't removal. Hence, I doubt kotlinlib incremental compilation is incompatible with Kotlin 2.4. But I haven't looked into it in detail.

If it is in fact necessary to use another code path, I'd avoid reflective invocation. We should be able to class-load a different worker implementation (based on the Kotlin version), and have some dedicated worker for Kotlin 2.4+.

viluon:
> Deprecation isn't removal.

In this case, the methods were deprecated in the implementation of the [`JvmPlatformToolchain` interface](https://github.com/JetBrains/kotlin/blob/fd00198088a7b3a3413750890ae8999dce5d22ae/compiler/build-tools/kotlin-build-tools-api/src/main/kotlin/org/jetbrains/kotlin/buildtools/api/jvm/JvmPlatformToolchain.kt#L36), but they were removed completely from this interface as part of [KT-82390](https://youtrack.jetbrains.com/issue/KT-82390).

Before this patch, Kotlin 2.4 compilation via the build tools API using the embedded compiler fails with a `NoSuchMethodError`:
```
java.lang.NoSuchMethodError: 'org.jetbrains.kotlin.buildtools.api.jvm.operations.JvmCompilationOperation org.jetbrains.kotlin.buildtools.api.jvm.JvmPlatformToolchain.createJvmCompilationOperation(java.util.List, java.nio.file.Path)'
  mill.kotlinlib.worker.impl.JvmCompileBtApiImpl.compile(JvmCompileBtApiImpl.scala:86)
  mill.kotlinlib.worker.impl.KotlinWorkerImpl.compile(KotlinWorkerImpl.scala:41)
  mill.kotlinlib.KotlinModule.$anonfun$18(KotlinModule.scala:302)
  mill.util.CachedFactoryBase.withValue(CachedFactoryBase.scala:170)
  mill.util.CachedFactory.withValue(CachedFactory.scala:51)
  mill.kotlinlib.KotlinModule.kotlinCompileTask$$anonfun$1(KotlinModule.scala:385)
  mill.api.Task$Anon.evaluate(Task.scala:564)
```

Thanks for the classloading suggestion, I'll pivot to that.

libs/kotlinlib/api/src/mill/kotlinlib/worker/api/KotlinWorker.scala
```diff
@@ -13,13 +13,16 @@ trait KotlinWorker {
...
+   * @param kotlinVersion The Kotlin version being compiled against. Selects the Build Tools API
+   *                      backend, since the API changed incompatibly in 2.4.0.
    * @param args Compiler arguments.
    * @param sources Source files to compile.
    * @param classpath Compilation classpath, useful for incremental compilation.
    */
   def compile(
       target: KotlinWorkerTarget,
       useBtApi: Boolean,
+      kotlinVersion: String,
```
lefou: This doesn't look like what we want. Instead the `KotlinWorkerManager.kotlinWorker` should load the correct worker implementation.

libs/kotlinlib/package.mill
```diff
@@ -70,6 +71,20 @@ object `package` extends MillStableScalaModule with BuildInfo {
       )
   }
 
+  // Dedicated Build Tools API backend for Kotlin 2.4+, compiled against the 2.4 API generation.
+  object `worker-btapi2` extends MillPublishScalaModule {
```
lefou: Let's name it `worker-btapi-2-4`

mill-build/src/millbuild/Deps.scala
```diff
@@ -241,6 +241,11 @@ object Deps {
   val kotlinCompiler = mvn"org.jetbrains.kotlin:kotlin-compiler:$kotlinVersion"
   val kotlinBuildToolsApi_api =
     mvn"org.jetbrains.kotlin:kotlin-build-tools-api:$kotlinBuildToolsApiVersion_api"
+  // The 2.4.0 Build Tools API dropped the legacy operation factories in favour of builders; the
+  // `worker-btapi2` module compiles the dedicated Kotlin 2.4+ backend against this generation.
+  val kotlinBuildToolsApiVersion_api2 = "2.4.0"
+  val kotlinBuildToolsApi_api2 =
+    mvn"org.jetbrains.kotlin:kotlin-build-tools-api:$kotlinBuildToolsApiVersion_api2"
```
lefou:
```suggestion
  val kotlinBuildToolsApi24Version_api = "2.4.0"
  val kotlinBuildTools24Api_api =
    mvn"org.jetbrains.kotlin:kotlin-build-tools-api:$kotlinBuildToolsApi24Version_api"
```

libs/kotlinlib/worker/src/mill/kotlinlib/worker/impl/KotlinWorkerImpl.scala
```diff
@@ -47,9 +44,31 @@ class KotlinWorkerImpl(
...
     if (!classpathSnapshotCacheIsStable) {
       os.remove.all(classpathSnapshotCache)
     }
   }
+
+  private def usesBuilderApi(kotlinVersion: String): Boolean = {
+    val Seq(major, minor) =
+      kotlinVersion.split("[.-]").take(2).flatMap(_.toIntOption).toSeq.padTo(2, 0)
+    major > 2 || (major == 2 && minor >= 4)
+  }
```
lefou: Let's invert the logic and check for the deprecated case(s), e.g. `< 2.4`. We can reason better about what we fully know.

It will result in a more maintainable and consistent codebase.

lefou (commented):
Let's see, if your approach works. It looks compact.

My idea was to explicitly load the correct worker into the tools classpath, based on the known Kotlin version. Deciding up-front makes later guessing/testing unnecessary, ~~but results in a more invasive change. I assume, the currently used `ClassLoaderCachedFactory` isn't up to that change, as it can't hold the Kotlin version as part of the cache key~~.

lefou:
I think instead of probing for the right Kotlin version, you just need to adapt the `KotlinModule.kotlinCompilerClasspath` and load the correct worker artifact. Something like:

```diff
   def kotlinCompilerClasspath: T[Seq[PathRef]] = Task {
+    val Array(major, minor) = kotlinVersion().split("[.]").take(2).map(_.toIntOption)
+    val workerDep = if(major.exists(_ < 2) || (major.exists(_ == 2) && minor.exists(_ < 4))) {
+      "mill-libs-kotlinlib-worker"
+    } else {
+      "mill-libs-kotlinlib-worker-2-4"
+    }
+
     val deps = kotlinCompilerMvnDeps() ++ Seq(
-      Dep.millProjectModule("mill-libs-kotlinlib-worker")
+      Dep.millProjectModule(workerDep)
     )
```

lefou:
I took the opportunity to rework the worker loading a bit, so it better aligns with other modules. The worker module is now a cross-module with cross-value "1" supporting older Kotlin versions starting from 1.0 to 2.3 and "2-4" supporting Kotlin 2.4+.

I also excluded the worker from the api docs generator, as it caused errors. Something we previously also saw in the Maven worker, https://github.com/com-lihaoyi/mill/pull/5977/#issuecomment-3407301062 .

lefou:
@viluon If you don't have any objections, I'll merge.

viluon (commented):
LGTM

PR #7340, opened 2026-07-01
`mill init`: Interpolate maven rawModel properties

### LINKED ISSUE #7326: mill init fails for a spring-boot-examples project
The generated build fails for one of the [spring-boot-examples](https://github.com/in28minutes/spring-boot-examples) that is migrated using Mill `1.2.0-RC1`.

```
spring-boot-basic-microservice/spring-boot-microservice-eureka-naming-server$./mill init
converting Maven build
removing existing build files ...
writing build.mill.yaml
build.mill.yaml-14] mill.init.InitMavenModule/init formatting Mill build files
build.mill.yaml-14] Formatting 1 Scala sources
parsed config (v3.8.5): ../mill-workspace/out/mill-build/mill.init.InitMavenModule/buildGenScalafmtConfig.dest/.scalafmt.conf
file excluded: ../mill-workspace/build.mill.yaml
build.mill.yaml-14] init completed, run "mill resolve _" to list available tasks
1/1, SUCCESS] ./mill init 1s

spring-boot-basic-microservice/spring-boot-microservice-eureka-naming-server$./mill show bomMvnDeps
1] show
[
  "org.springframework.cloud:spring-cloud-dependencies:${spring-cloud.version}"
]
```

A Maven property is unresolved because the `rawModel` used is not interpolated. This was fixed in #7174 but the changes were reverted in #7265 for excluding member dependencies of `spring-boot-dependencies` BOM.

The member dependencies were being included due to POM inheritance. The project extends `spring-boot-starter-parent` that in turn extends `spring-boot-dependencies` instead of declaring it a BOM dependency. The resulting `effectiveModel` does not contain the BOM-member relationship and so, there is no easy way to exclude the members of any given BOM.

In short, the changes in #7174 should be restored so that `init` generates a working build in the general case.

souvlakias:
To fix #7326

### The problem
When migrating Spring Projects, in order to avoid bloated generated build files by including everything that the effective model inherits, we get the dependency management from the rawModel instead.
This, however, leaves some properties uninterpolated, for example `${spring-cloud.version}`.

### Fix
Add an extra step that finds these properties with regex and then and replaces them with the corresponding values from the effective model.

We think it's preferable to reason through and hack around any issues arising from using the raw model than to generate build files containing hundreds of dependencies that ultimately come from a BOM.

### DIFF
--- a/libs/init/maven/test/resources/expected/spring-start/build.mill.yaml
+++ b/libs/init/maven/test/resources/expected/spring-start/build.mill.yaml
@@ -4,6 +4,8 @@ extends: [spring.boot.SpringBootModule, MavenModule, PublishModule]
 springBootPlatformVersion: 4.0.6
 mvnDeps:
 - org.springframework.boot:spring-boot-starter:4.0.6
+bomMvnDeps:
+- org.springframework.cloud:spring-cloud-dependencies:2023.0.3
 artifactName: demo
 pomSettings:
   organization: com.example
--- a/libs/init/maven/src/mill/main/maven/Modeler.scala
+++ b/libs/init/maven/src/mill/main/maven/Modeler.scala
@@ -22,11 +22,16 @@ class Modeler(
   private val PropertyRef = "\\$\\{([^}]+)\\}".r
 
   private def interpolateValue(value: String, properties: Properties): String =
-    PropertyRef.replaceAllIn(value, m =>
-      Matcher.quoteReplacement(Option(properties.getProperty(m.group(1))).getOrElse(m.matched))
+    PropertyRef.replaceAllIn(
+      value,
+      m =>
+        Matcher.quoteReplacement(Option(properties.getProperty(m.group(1))).getOrElse(m.matched))
     )
 
-  private def interpolateDependencyVersions(model: org.apache.maven.model.Model, properties: Properties): Unit = {
+  private def interpolateDependencyVersions(
+      model: org.apache.maven.model.Model,
+      properties: Properties
+  ): Unit = {
     Option(model.getDependencies).foreach(_.asScala.foreach { dep =>
       Option(dep.getVersion).foreach(v => dep.setVersion(interpolateValue(v, properties)))
     })
@@ -62,8 +67,9 @@ class Modeler(
       val depMgmt1 = Option(result1.getEffectiveModel.getDependencyManagement).map(_.clone)
       val result2 = builder.build(request, result1)
       val interpolationProps = Properties()
-      systemProperties.forEach((k, v) => interpolationProps.put(k, v))
       result2.getEffectiveModel.getProperties.forEach((k, v) => interpolationProps.put(k, v))
+      // Match Maven precedence: externally supplied/system properties override model properties.
+      systemProperties.forEach((k, v) => interpolationProps.put(k, v))
       interpolateDependencyVersions(result2.getRawModel, interpolationProps)
       // Restore dep mgmt from Phase 1 since Phase 2 substitutes BOM deps with their components.
       depMgmt1.foreach(result2.getEffectiveModel.setDependencyManagement)
--- a/libs/init/maven/src/mill/main/maven/Modeler.scala
+++ b/libs/init/maven/src/mill/main/maven/Modeler.scala
@@ -8,6 +8,7 @@ import org.eclipse.aether.repository.{LocalRepository, RemoteRepository}
 import org.eclipse.aether.supplier.RepositorySystemSupplier
 
 import java.io.File
+import java.util.regex.Matcher
 import java.util.Properties
 import scala.jdk.CollectionConverters.*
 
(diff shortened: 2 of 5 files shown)


lefou:
I agree, detecting substitutions and interpolating them explicitly is better than generating a highly overspecified build.

ajaychandran:
Another thing to consider is that using the `rawModel` results in all configuration in the POM hierarchy being ignored, including values in the `<dependencyManagement>` section.

souvlakias:
> Another thing to consider is that using the `rawModel` results in all configuration in the POM hierarchy being ignored, including values in the `<dependencyManagement>` section.

Valid point. Would a linear pass over the `rawModels` of models above in the hierarchy, and picking up their bom/deps make sense? (except the `spring` one).

The alternative that you suggested in the linked issue, sounds a bit too technical for this case I think. 
> The DefaultModelBuilderFactory from the maven-model-builder API allows for customizing the [model building steps](https://maven.apache.org/ref/3.9.16/maven-model-builder/index.html). To satisfy the spring usecase, an InheritanceAssembler can be provided to exclude the dependencies from spring-boot-

libs/init/maven/test/resources/expected/spring-start/build.mill.yaml
```diff
@@ -4,6 +4,8 @@ extends: [spring.boot.SpringBootModule, MavenModule, PublishModule]
 springBootPlatformVersion: 4.0.6
 mvnDeps:
 - org.springframework.boot:spring-boot-starter:4.0.6
+bomMvnDeps:
+- org.springframework.cloud:spring-cloud-dependencies:2023.0.3
```
vaslabs, line 8: so in this case it will be fine because the mvnDeps versions are hardcoded, but SpringBootModule also overrides bomMvnDeps, perhaps we should move that to mandatoryBomMvnDeps to separate it from user provided bomMvnDeps and also avoid using !append here

vaslabs: ideally we should detect when the version is resolved via bom and not have it in mvn dep, but out of scope for this PR I guess

ajaychandran:
IMO, the real issue is that `spring-boot-dependencies` appears in the POM hierarchy instead of being declared a BOM dependency. Should the migration process be responsible for correcting such problems in the input build?

I would also prefer if the implementation was not compromised for a cosmetic requirement (removal of "bloated" dependencies).

> Valid point. Would a linear pass over the rawModels of models above in the hierarchy, and picking up their bom/deps make sense? (except the spring one).

Is this simpler than using a custom `InheritanceAssembler` that delegates to the default instance in all but one case?

vaslabs:
> IMO, the real issue is that `spring-boot-dependencies` appears in the POM hierarchy instead of being declared a BOM dependency. Should the migration process be responsible for correcting such problems in the input build?
> 
> I would also prefer if the implementation was not compromised for a cosmetic requirement (removal of "bloated" dependencies).
> 
> > Valid point. Would a linear pass over the rawModels of models above in the hierarchy, and picking up their bom/deps make sense? (except the spring one).
> 
> Is this simpler than using a custom `InheritanceAssembler` that delegates to the default instance in all but one case?

tbf, there are two lines of work here that conflict. The `SpringBootModule` is responsible for making the user experience similar as adding a few deps and the spring boot plugin in maven or gradle. Imo it is not the responsibility of the migration to exhaust all possibilities to make a framework based application work. SpringBootModule is also quite new so we should make the responsibility lines more clear now that it exists.

Nevertheless, I'd prefer to move on with small steps and improve both migration and SpringBootModule on an example basis , rather than an exhaustive solution on the migration side. 

For instance, handling the spring-boot-dependencies (or any peculiarities of a spring specific setup) correctly should be the responsibility of `SpringBootModule` imo .

souvlakias:
As fair as i understand, the motive of heuristically detecting and applying mill's `SpringBootModule` is indeed partly cosmetic and moves in favor of developer experience when viewing the build scripts. This is why it sounds reasonable to hack around the final effective model 😅

> > Valid point. Would a linear pass over the rawModels of models above in the hierarchy, and picking up their bom/deps make sense? (except the spring one).

> Is this simpler than using a custom `InheritanceAssembler` that delegates to the default instance in all but one case?

Regarding this, im not really sure. Ill try to explore both options

libs/init/maven/test/resources/expected/spring-start/build.mill.yaml
souvlakias: opened https://github.com/com-lihaoyi/mill/issues/7345 for tracking. will open a PR soon

souvlakias:
opened
- https://github.com/com-lihaoyi/mill/pull/7349

as a first draft for using the custom maven filters. 
It does seem like a better solution, in which case I can close this PR if we proceed with that strategy.

souvlakias:
https://github.com/com-lihaoyi/mill/pull/7349 solved the issue

PR #7342, opened 2026-07-01
Scheduled migration workflow

### LINKED ISSUE #7324: Add scheduled tasks to verify all mill init
Because we verify against bill example projects for mill init such as [spring examples](https://github.com/in28minutes/spring-boot-examples/) and [quarkus-quickstart](https://github.com/quarkusio/quarkus-quickstarts) , compiling all of the projects in one go takes a long time and more often than not,  such example tests in CI time out.

My suggestion is:

- Add a configurable parameter to example tester to parametarise the timeout from the hardcoded 15 minutes (in `UtestExampleTestSuite`)
- Add scheduled actions for all big example/quickstart projects
   - Run the scheduled actions for migration examples on all platforms (mac/linux/windows) and for all modules and submodules in an example / quickstart project
- Have only a small subset of them run on every PR/main change like done [here](https://github.com/com-lihaoyi/mill/pull/7296/changes#diff-46a4fb00e3a3354f777ff969e02a977871ed987d89343fff77ee40ffab8f82d4R13) as a smoke test.           
    - Perhaps add another parameter in ExampleTester so some commands run only on schedule, if they are on CI, e.g. `# scheduled` but on non CI (locally) they also run so we don't miss running tests locally
    - select the examples in a way that covers different features, e.g. codegen (quarkus), reflection (springboot) , native etc.

souvlakias:
Successful test run: https://github.com/vaslabs-ltd/mill/actions/runs/30640323209

### Adds
- A configurable timeout envar `MILL_TEST_TIMEOUT_MINUTES` (Keeping the default 15 minutes)
- Support for scheduled-only runs
  - A new prefix filter `scheduled`. These commands run locally but not in CI, unless the `MILL_TEST_SCHEDULED` envvar is set (from the scheduled runs)
- `scheduled-migration-tests.yml` to run `example.migrating._._` weekly on all platforms
  - Its timeout is 60 minutes.  
  -  windows run uses `packaged.daemon` to work around an issue with oslib (see https://github.com/com-lihaoyi/mill/pull/7342#issuecomment-4970650939)
- Also updated some migration examples to run their heavy runs (like `__.compile`) only on schedule.

To close https://github.com/com-lihaoyi/mill/issues/7324


.github/workflows/run-tests.yml
```diff
@@ -165,7 +165,7 @@ jobs:
           - millargs: "example.thirdparty.koin-androidx-samples.packaged.daemon"
             setup-android: true
 
-          - millargs: "example.migrating.scalalib._"
+          - millargs: "example.migrating._._"
```
lefou: If find `scalalib._` already strange, but `_._` is a call for trouble. It would be good, if we could identify the exact task name and use that, e.g. `_._.testForked`.

souvlakias:
This still needs some tweaking and investigation for everything to pass. 
I made https://github.com/vaslabs-ltd/mill/tree/scheduled-ci-tests-testrun to have it run on push and see how it goes. Some tests fail on CI but pass locally.

souvlakias:
Testrun [here](https://github.com/vaslabs-ltd/mill/actions/runs/29344235891)
For the windows runs, I'm hitting this issue from os-lib:
- https://github.com/com-lihaoyi/os-lib/issues/404

on the removal of the `.git` folder by https://github.com/com-lihaoyi/mill/blob/dc1f603548f468c5bd306cd50c48e64b2f16f5cd/testkit/src/mill/testkit/IntegrationTesterBase.scala#L69 which is readonly.

At first I thought it was a file lock issue with some git process, but its the permissions afterall

vaslabs:
> Testrun [here](https://github.com/vaslabs-ltd/mill/actions/runs/29344235891) For the windows runs, I'm hitting this issue from os-lib:
> 
> * [removing directory on OneDrive fails with `AccessDeniedException` os-lib#404](https://github.com/com-lihaoyi/os-lib/issues/404)
> 
> on the removal of the `.git` folder by
> 
> https://github.com/com-lihaoyi/mill/blob/dc1f603548f468c5bd306cd50c48e64b2f16f5cd/testkit/src/mill/testkit/IntegrationTesterBase.scala#L69
> 
> which is readonly.
> At first I thought it was a file lock issue with some git process, but its the permissions afterall

can we only have it for whatever works for the time being and open issues for the rest to work on?

souvlakias:
Update: 
- changed windows run to use `packaged.daemon` to work around the issue
- made the quarkus `__.compile` to only run on linux

Successful test run: https://github.com/vaslabs-ltd/mill/actions/runs/30640323209

Will mark this as ready after the fork's `Run Tests` finishes

vaslabs (approved):

vaslabs:
We'll need an issue to fix Quarkus on windows. Lgtm

PR #7349, opened 2026-07-02
`mill init` Spring detection fixes

souvlakias:
### Motivation

<details>
<summary>To close https://github.com/com-lihaoyi/mill/issues/7326 . </summary>

`spring-boot-examples/spring-boot-basic-microservice/spring-boot-microservice-eureka-naming-server/build.mill.yaml`

```yaml
mill-version: 1.1.2-301-492680
mill-jvm-version: system
extends: [spring.boot.SpringBootModule, MavenModule, PublishModule]
springBootPlatformVersion: 4.0.0
mvnDeps:
- org.springframework.cloud:spring-cloud-starter-netflix-eureka-server:4.3.0
bomMvnDeps:
- org.springframework.cloud:spring-cloud-dependencies:2025.0.0
artifactName: spring-boot-microservice-eureka-naming-server
pomSettings:
  description: Microservices with Spring Boot and Spring Cloud - Eureka Naming Server
  organization: com.in28minutes.springboot.microservice.eureka.naming.server
publishVersion: 0.0.1-SNAPSHOT
repositories: ["https://repo.spring.io/milestone"]

object test:
  extends: [SpringBootTestsModule, MavenTests, TestModule.Junit5]
  mvnDeps:
  - org.springframework.boot:spring-boot-starter-test
  testParallelism: false
  testSandboxWorkingDir: false
```
```console
> ./mill __.compile   
90] test.compile compiling 1 Java source to out/test/compile.dest/classes ...
90] done compiling
90/90, SUCCESS] mill __.compile
```

</details>

Avoids the direct use of `rawModel` for `depManagement` and also the need for
- https://github.com/com-lihaoyi/mill/pull/7340

since relying on just `rawModel` could lose inherited properties.

Additionally, current approach only detected from the `spring-boot-starter-parent`, but it's valid to use spring boot without the parent. ([example](https://github.com/spring-projects/spring-boot/blob/fb9794a2d346036a6adfa74cc3f1253010d63ea0/build-plugin/spring-boot-maven-plugin/src/docs/antora/modules/maven-plugin/examples/using/no-starter-parent-override-dependencies-pom.xml#L35)). 

### Additions
- Detect spring-boot usage from BOM declaration too.
The issue with this is that the BOM gets removed from the `effectiveModel`, so this had to be a hybrid approach, meaning it first picks up the BOM from the `rawModel` and then resolves (if needed) the version property from the `effectiveModel`. 
	- Also add a test for such usage
- Detect spring-boot usage from any spring-boot dep.
	- We have [3-maven-complete-large](https://github.com/com-lihaoyi/mill/blob/main/example/migrating/javalib/3-maven-complete-large/build.mill) for this, it has many subprojects that use springboot. 
	With this PR, they cleanly get the `SpringBootModule` too, and versions that should come from the dep BOM get stripped.

### Changes
- Remove the logic of using `rawModel` for `dependencyManagement`
- Filter the `effectiveModel` based on the location the info came from

### DIFF
--- a/libs/init/maven/src/mill/main/maven/Modeler.scala
+++ b/libs/init/maven/src/mill/main/maven/Modeler.scala
@@ -1,7 +1,7 @@
 package mill.main.maven
 
 import mill.api.PathRef
-import org.apache.maven.model.*
+import org.apache.maven.model.{DependencyManagement, Model}
 import org.apache.maven.model.building.*
 import org.apache.maven.model.composition.DependencyManagementImporter
 import org.apache.maven.model.inheritance.InheritanceAssembler
@@ -118,10 +118,13 @@ class FilteringInheritanceAssembler(delegate: InheritanceAssembler)
       request: ModelBuildingRequest,
       problems: ModelProblemCollector
   ): Unit = {
-    val parentGroupId = Option(parent.getGroupId).orElse(Option(parent.getParent).map(_.getGroupId)).getOrElse("")
+    val parentGroupId =
+      Option(parent.getGroupId).orElse(Option(parent.getParent).map(_.getGroupId)).getOrElse("")
     val parentArtifactId = parent.getArtifactId
 
-    if (parentGroupId == SpringBoot.GroupId && parentArtifactId == SpringBoot.DependenciesArtifactId) {
+    if (
+      parentGroupId == SpringBoot.GroupId && parentArtifactId == SpringBoot.DependenciesArtifactId
+    ) {
       val parentClone = parent.clone()
       parentClone.setDependencyManagement(null)
       delegate.assembleModelInheritance(child, parentClone, request, problems)
@@ -140,7 +143,8 @@ class FilteringDependencyManagementImporter(delegate: DependencyManagementImport
       problems: ModelProblemCollector
   ): Unit = {
     // Filter out Spring Boot BOMs by checking if their source location points to a Spring Boot BOM or parent.
-    val filteredSources = if (sources == null) null else {
+    val filteredSources = if (sources == null) null
+    else {
       sources.asScala.filterNot { dm =>
         Option(dm.getDependencies).flatMap(_.asScala.headOption).exists { dep =>
           val location = dep.getLocation("")
@@ -154,6 +158,11 @@ class FilteringDependencyManagementImporter(delegate: DependencyManagementImport
         }
       }.asJava
     }
-    delegate.importManagement(target, filteredSources.asInstanceOf[java.util.List[? <: DependencyManagement]], request, problems)
+    delegate.importManagement(
+      target,
+      filteredSources.asInstanceOf[java.util.List[? <: DependencyManagement]],
+      request,
+      problems
+    )
   }
 }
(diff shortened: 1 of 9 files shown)


souvlakias:
While this adds detection from BOM too, I'm exploring if we should detect and apply `SpringBootModule` from any spring related dep too.

For example the [3-maven-complete-large](https://github.com/com-lihaoyi/mill/blob/main/example/migrating/javalib/3-maven-complete-large/build.mill) example we have, has spring projects but they do not declare nor parent nor bom.

Test branch and commit ([here](https://github.com/vaslabs-ltd/mill/commit/e95bbae1701a3472951fea7156b883f911d3f858)) (adds on top of this one)
Seems functional. Will look a bit more into it and possibly include it in this PR too. Alternatively, could add this in another PR after this

libs/init/maven/src/mill/main/maven/Modeler.scala
```diff
@@ -92,3 +103,66 @@ object Modeler {
...
+          } else ""
+          val isSpringBOM = sourceName.contains(SpringBoot.DependenciesArtifactId) ||
+            sourceName.contains(SpringBoot.ParentArtifactId)
+          isSpringBOM
+        }
+      }.asJava
+    }
+    delegate.importManagement(
+      target,
+      filteredSources.asInstanceOf[java.util.List[? <: DependencyManagement]],
```
vaslabs: I don't remember if asInstanceOf blows up if it's filteredSources is null (from 146)

vaslabs: looks alright
```scala
> scala
Welcome to Scala 3.7.0 (17, Java OpenJDK 64-Bit Server VM).
                                                                                                                                                                                             
scala> null.asInstanceOf[List]
-- [E056] Syntax Error: ---------------------------------------------------------------------------------------------------------------------------------------------------------------------
1 |null.asInstanceOf[List]
  |^
  |Missing type parameter for List
1 error found
                                                                                                                                                                                             
scala> null.asInstanceOf[List[Int]]
val res0: List[Int] = null
                                                                                                                                                                                             
scala> 
```

vaslabs:
output wise , this looks much better to me

ajaychandran:
@souvlakias Another approach would be to enable location tracking in `DefaultModelBuildingRequest` and exclude dependencies by  `Dependency.getInputLocation("")`.

souvlakias:
> @souvlakias Another approach would be to enable location tracking in `DefaultModelBuildingRequest` and exclude dependencies by `Dependency.getInputLocation("")`.

okay, I will look into it too, thanks

souvlakias:
Update: Was already using `getLocation` in the custom `DependencyManagementImporter` I was trying but yes, we can do that check in the final effective model rather than in the middle of the import; meaning we can avoid using that and the `InheritanceAssembler`

Pushed some logic of applying `SpringBootModule` as I described in above comment. It seems functional with [3-maven-complete-large](https://github.com/com-lihaoyi/mill/blob/main/example/migrating/javalib/3-maven-complete-large/build.mill)

Still needs some checking and cleaning up. Will update PR description too

souvlakias:
I think this is ready for review

vaslabs:
Lgtm, you can remove the WIP from the PR description

vaslabs (approved):

vaslabs:
@lefou @ajaychandran I think this looks good, wdyt? the plan after merging this, is to create a baseline of tests with https://github.com/com-lihaoyi/mill/pull/7342 (still in progress)

PR #7357, opened 2026-07-03
Change `Result.create` to accept a by-name parameter

### LINKED ISSUE #7356: Result.create has useless try?
value is not a by-name parameter so is this a mistake?
https://github.com/com-lihaoyi/mill/blob/1a9b2cb9bea7e23e321fe9d7193bdade68b161df/core/api/daemon/src/mill/api/daemon/Result.scala#L27-L31

lefou:
Fix https://github.com/com-lihaoyi/mill/issues/7356


core/api/daemon/src/mill/api/daemon/Result.scala
```diff
@@ -24,7 +25,10 @@ sealed trait Result[+T] {
   }
 }
 object Result {
-  implicit def create[T](value: T): Result[T] =
+  @deprecated("Use the overload accepting a by-name parameter", "Mill after 1.1.7")
+  @publicInBinary
+  private[daemon] def create[T](value: T): Result[T] = Success(value)
```
bishabosha, line 31: i think this isnt tasty compatible to hide visibility if an inline method called this previously

bishabosha: if no one ported tasty mima to a mill plugin then this works well for me: https://github.com/bishabosha/scala-object-notation/blob/main/tastyMima.mill

lefou: Should we keep it public then, but keep the deprecation and remove the `implicit`?

lefou: From reading the [docs](https://scala-lang.org/api/scala/annotation/publicInBinary.html), it looks safe to me. 

Compiling `core.api.daemon` with `-WunstableInlineAccessors` isn't complaining.

lefou: I think we can ignore it for now. We don't expect users to use a different Scala version, when building plugins. Also, when experimentally testing your plugin, I see lots of unrelated issues. But I'm also not sure how to use it correctly. I just added it to `MillStableScalaModule` and ran `mill __.tastyMimaReportIssues`.

lefou: But it's reporting this:

```
mill-core-api-daemon_3-1.1.7: The symbol mill.api.daemon.Result.create has a more restrictive visibility qualifier in current version: before: public; after: package[mill.api.daemon]
  filter with: ProblemMatcher.make(ProblemKind.RestrictedVisibilityChange, "mill.api.daemon.Result.create")
```

bishabosha: i think it would mean that any inline method published in a plugin that happened to use this method would get an error when you try to call it from the newer mill

lefou: So, this the risk is rather small. 

What we gain is no ambiguity between `create(T)` and `create(=> T)`.

What we risk is someone hitting this issue, which is IMHO unlikely. If someone reports it, we can either

1. advice to workaround by not using `inline`
2. release a fix by making `def create[T](T)` public again

lefou: But I still don't get it. If this is a runtime issue, why isn't it enough to have the older `def` public in bytecode? Isn't tasty only required, when it comes to compilation?

bishabosha: when you call a function in an inline method, the TASTy then records the function signature, and therefore the public `create` will have a signature that says `scala.Function0` instead of `java.lang.Object` so it wont resolve and you get compiletime error.

E.g. for a published mill plugin being used in a newer version of mill

lefou: So this is all about compile-time, not runtime. Both `create` functions exists (accepting `scala.Function0` and `java.lang.Object`) at compile and runtime in the newer binary. Mill API is not forward compatible. We don't expect plugins build against a specific Mill version to work with an older Mill version.

bishabosha: ~No, I specifically said the other way around - making `create[T](t: T)` private hides it from inline type-checking, so plugin that was published against old Mill API (using that method in an inline function) can't work in the new Mill.~
Edit: ignore its incorrect

bishabosha: I apologise, i tested this exact scenario and `@publicInBinary` specifically does make the visibility public for the purpose of checking after inlining so this is all clear! (which the scaladoc hints at but does not explicitly say)

lefou: Thank you very much for your support!

lefou:
Not sure this is ready for merge. Some tests that check for error reporting in meta build fail.

Issue #7367, opened 2026-07-08
Mill init for Kotlin Projects (Gradle)


souvlakias:
Importing Gradle projects with `mill init` does not have Kotlin detection yet.

Kotlin detection is also needed for 
- https://github.com/com-lihaoyi/mill/issues/7366

since most android projects use Kotlin.

vaslabs:
can close this now and open individually for every case, more will appear once we start with android (e.g. ksp)

PR #7386, opened 2026-07-16
`mill init`: Gradle Kotlin Support & Refactor 

souvlakias:
First work on:
- https://github.com/com-lihaoyi/mill/issues/7367

## Summary
This PR adds support for Kotlin based Gradle projects, including kompiler options and compiler plugins. 
Additionally, it refactors BuildGenModule to separate Gradle data extraction and ModuleSpec configuration logics, to prepare the ground for upcoming Android detection.

### Refactor

- Extract gradle `Project` interactions into a single `extractJvmData` function which returns `ExtractedJvmData`
  - Additional configuration elements have their own dataclass (e.g. `KotlinData`, `FrameworkData`) 
- Have base configuration logic in `configureBaseJvmModule`
- Then for java/kotlin, `configureJvmModule` uses the extracted data and configures framework specfic logic and adds TestModule.
  - Todo? have the base logic of the tesmodule be added in `configureBaseJvmModule`

- Android configuration could be a `configureAndroidModule` that uses the base config.

### Kotlin support
- For Kotlin data, we use reflection for:
	- KotlinVersion: by calling `getPluginVersion`.
	If this fails, it fallbacks to implementation version and strips any `-release` suffix ( e.g. `2.4.0-release-281`)
	- KotlinOptions: call `getCompilerOptions -> getFreeCompilerArgs`
- For the compiler plugins we use the`kotlinCompilerPluginClasspathMain` config, and exclude the `kotlin-scripting-compiler-embeddable`
- Fix `kotlin-test` resolution:
  -  `kotlin-test` is a multiplatform pom and doesn't include the needed Classes. The actual variant is picked up from the resolved `testRuntimeClasspath` configuration. (e.g. find `kotlin-test-junit5`)
- Added a `kotlin-multiplatorm` test in gradle tests, that was generated with Idea. It features both kotlin options and compiler plugin detection.


vaslabs:
would it make sense if we had a completely separate PR for BuildGenModule do you think? I don't have strong opinions on this, it just might be easier to follow incrementally and make sure there are no regressions with just the refactoring changes

souvlakias:
> would it make sense if we had a completely separate PR for BuildGenModule do you think? I don't have strong opinions on this, it just might be easier to follow incrementally and make sure there are no regressions with just the refactoring changes

At this point I think splitting would be more painful😅
But if needed, then I think the refactor should come before adding the kotlin addition (in case you meant the other way round)

vaslabs:
> > would it make sense if we had a completely separate PR for BuildGenModule do you think? I don't have strong opinions on this, it just might be easier to follow incrementally and make sure there are no regressions with just the refactoring changes
> 
> At this point I think splitting would be more painful😅 But if needed, then I think the refactor should come before adding the kotlin addition (in case you meant the other way round)

yes, I meant refactor first. but I think it's ok to proceed from my side.

let's fix whatever issues exist with golden examples and so on and get it ready

souvlakias:
> let's fix whatever issues exist with golden examples and so on and get it ready

Yes, the tests locally succeed, but the [tests in CI](https://github.com/vaslabs-ltd/mill/actions/runs/29485478315/job/87589040900?pr=7386) are failing.
~~I'm not sure why this is happening, seem to be an inconsistency with the `ProjectBaseModule/Test` generation on local vs CI run.~~

Update: The test `src` directory wasn't being commited.

vaslabs:
as this is the first targetted effort to make mill for kotlin, would love to see your input

@lefou @lihaoyi

libs/init/gradle/exportplugin/src/mill/main/gradle/BuildModelBuilder.scala
```diff
@@ -36,11 +36,372 @@ class BuildModelBuilder(ctx: GradleBuildCtx, objectFactory: ObjectFactory, works
...
+          .toSeq.distinct
+      }
+
+      // org.jetbrains.kotlin:kotlin-test is a multiplatform POM that lacks JVM classes.
+      // We find the actual platform variant (e.g. kotlin-test-junit5)
+      // resolved by Gradle's test runtime classpath.
+      val kotlinTestResolvedNameOpt = {
+        testConfigs.find(_.getName == "testRuntimeClasspath")
+          .flatMap { config =>
+            Try(config.getResolvedConfiguration.getResolvedArtifacts.asScala).toOption
```
vaslabs: this can fail silently

vaslabs: I think there's no reason for a Try here, if it fails it should crash and give whatever error information to the user, what do you think @souvlakias ?

Jason-123-cyber (approved):
MERGE

libs/init/gradle/exportplugin/src/mill/main/gradle/BuildModelBuilder.scala
souvlakias: This, and a few lines above, are the only 2 places where we call and trigger gradle's resolved configuration, which might fail for whatever reason (dependency resolution, incomplete build, etc) and throw an exception. 

I think its valid to have a Try, but maybe we could print a short log like `Gradle Resolution failed, can't detect kotlin compiler plugins` or something, in the case of a failure

vaslabs: yes, we should at least log something, so the user knows what's going on

vaslabs: yes, we should at least log something, so the user knows what's going on

vaslabs (approved):

PR #7432, opened 2026-08-11
Introduce `mandatoryBomMvnDeps` for module overrides

### LINKED ISSUE #7345: Create mandatoryBomMvnDeps for module overrides
For `mvnDeps`, a `mandatoryMvnDeps` exists for modules to override freely, so that users don't need to append to `super.mvnDeps`.

This doesn't exist for `bomMvnDeps`, and could potentially cause issues with users (or automigration tool) overriding the needed BOMs that modules add.
Example modules that need this:
```console
> grep -r -E "override def bomMvnDeps"
libs/javalib/src/mill/javalib/spring/boot/SpringBootModule.scala:  override def bomMvnDeps: T[Seq[Dep]] = Seq(
libs/javalib/src/mill/javalib/quarkus/QuarkusModule.scala:  override def bomMvnDeps: T[Seq[Dep]] = Task {
libs/javalib/src/mill/javalib/TestModule.scala:    override def bomMvnDeps: T[Seq[Dep]] = Task {
libs/groovylib/src/mill/groovylib/GroovyModule.scala:  override def bomMvnDeps: T[Seq[Dep]] = super.bomMvnDeps() ++
```

souvlakias:
Closes #7345

### Changes (Binary compatible)
- Adds `mandatoryMvnDeps` in JavaModule **+** `allBomMvnDeps` 
- Makes Groovy, Quarkus and Springboot modules override that instead of `bomMvnDeps`

### Future changes (Break compatibility)
- Also propagate to `JavaTests/JavaTests0` and  `TestModule.Junit5/Spock`
- Commit [here](https://github.com/vaslabs-ltd/mill/commit/86c9118fc086e0a539ffd5734b2ca32061186d7d) 
- Mima run [here](https://github.com/vaslabs-ltd/mill/actions/runs/31490369921/job/93777242916)


souvlakias:
@lefou wdyt about the changes that break binary compatibility? When would be a good time to include them, given that a minor version change is coming up? Or can we only break compatibility with a major version change?

lefou:
In principle, we can only break binary compatibility in major versions. 

See also

* https://github.com/com-lihaoyi/mill/issues/7056

souvlakias:
oh `example.javalib` doesn't pass sorry. 
I didn't notice cause fork actions always fail due to the [test-dist](https://github.com/vaslabs-ltd/mill/actions/runs/31490366323/job/93775159947) job failing and I was expecting an overall failure 😅️

souvlakias:
Added an `allBomMvnDeps` (like `allMvnDeps`)

vaslabs:
Looks good to me, I guess we can do the java tests at a later point

vaslabs (approved):

PR #7433, opened 2026-08-11
fix(mill): Avoid watching directory probes (backport #7173)

nguyenyou:
## Summary

Backport the directory-probe handling from #7173 to the Mill 1.1.x branch.

`ScriptModuleInit.resolveScriptModule` previously registered every candidate path as a watch, including directories. Because `PathRef` recursively digests directories, resolving a module namespace could hash a large source tree during every Mill invocation.

Directory candidates are now excluded from `evalWatch`. Script-file and missing-file candidates continue to be watched, preserving `staticBuildOverrides` invalidation behavior.

A focused regression test covers both cases.

## Performance

Profiling a repeated warm no-op build with unchanged sources showed:

- Total invocation: approximately 1.5 seconds
- Recursive directory digest: approximately 953 milliseconds

This change removes that directory-digest work from module resolution.

## Reference

This follows the implementation already merged for the Mill 1.2.x line:

- #7173
- https://github.com/com-lihaoyi/mill/commit/69e06ebfd9810533b2fd2e6f2d0cdac5650824c3


lefou:
Thank you!

Issue #7366, opened 2026-07-08
Mill init for Android (gradle only)


vaslabs:
I think we are in a good place to set this up for all known examples we have managed so far and we may even find gaps in `AndroidModule`s

Cover:

- compose samples (jetpack compose)
- hilt (architecture samples)
- data bind view
- proguard settings
- R8/D8 configs

We can do this in multiple PRs and note where things can be simplified (e.g. merge kotlin and java, r8 and d8 etc)

cc: @souvlakias

souvlakias:
👍 We can start with pure Java Android projects. And then move on to Kotlin ones when we implement the basics of  
- #7367

souvlakias:
> note where things can be simplified (e.g. merge kotlin and java)

Thinking about this, it may be indeed needed, at least for newer project using **AGP 9.0+**. With newer builds, Kotlin compilation is no longer explicitly indicated by the presence of the `kotlin-android` plugin, so relying on that plugin alone would not be sufficient.

Source: See https://developer.android.com/build/migrate-to-built-in-kotlin
> Android Gradle plugin 9.0 introduces built-in Kotlin support and enables it by default. That means you no longer have to apply the org.jetbrains.kotlin.android (or kotlin-android) plugin 
> 1. Remove the kotlin-android plugin
> ...

Merging the functionality of `AndroidKotlinModule` into `AndroidModule` would be beneficial. This would align with the direction taken by AGP and also avoid having to distinguish between the two based solely on whether the Kotlin plugin is applied.

lefou:
What would be the impact (of merging with Kotlin), if someone mixes the `AndroidModule` with `ScalaModule` or ~~`GradleModule`~~`GroovyModule`?

souvlakias:
> What would be the impact (of merging with Kotlin), if someone mixes the `AndroidModule` with `ScalaModule` or `GradleModule`?

I realized this later and forgot to edit. `AndroidScalaModule` currently needs the plain `AndroidModule`, so maybe we should do something like:
- `AndroidModule` -> `AndroidBaseModule` (so `AndroidScalaModule` could extend this internally)
- `AndroidKotlinModule` -> `AndroidModule` (Adjust whats needed so it works for Java and Kotlin, and users should use this for both languages)

vaslabs:
It might not be best to change the paradigm for Android, since other mill modules are using a similar structure. Let's start on mill init and see how it goes, so we'll know what "metrics" or "patterns" we want to get improvements against

[...]

PR #7446, opened 2026-08-21
`mill init`: Initial Android support

souvlakias:
Part of #7366 
This PR adds the initial  support for Android projects. The included examples are single-module Android Applications

### Additions 
Previous refactor PR came in handy for these new additions to be cleanly added.
- Extracts `android {}` config via reflection. Stores to the new `AndroidAppData` case class.
- Applies `AndroidKotlinModule` or `AndroidAppKotlinModule`
- Emits one shared `androidSdkModule0` at the build root (to avoid having each module declare one)
- Fixes a preexisting bug  where the repositories check had a non-read side-effect.

### An issue - YAML builds
No support for Declarative builds yet, See https://github.com/com-lihaoyi/mill/issues/7366#issuecomment-5355440260
The logic here, just ignores android modules when `declarative=true`

### Future work
- Apply `androidLibModule` when publishing information is available
- Multi-Module examples
- Compose, hilt, data/view binding, proguard/R8 config
- Signing configuration tasks? (didn't add these for now)
- YAML support

### DIFF
--- a/libs/init/buildgen/api/src/mill/main/buildgen/ModuleSpec.scala
+++ b/libs/init/buildgen/api/src/mill/main/buildgen/ModuleSpec.scala
@@ -102,7 +102,8 @@ case class ModuleSpec(
     androidTargetSdk: Value[Int] = Value(),
     androidVersionCode: Value[Int] = Value(),
     androidVersionName: Value[String] = Value(),
-    androidBuildToolsVersion: Value[String] = Value()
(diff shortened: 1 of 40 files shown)


vaslabs:
Let's open an issue to support declarative for android and a few for the rest of the future work.

I think we should tackle declarative first as we'll probably need to do design changes (luckily everything on android is marked as experimental)

vaslabs (approved):

PR #7454, opened 2026-08-24
Rescue undiscovered but explicitly specified test classes

### LINKED ISSUE #7450: Enhancement request: testOnly can be used to run ScalaTest suites annotated with @DoNotDiscover
Mill's `testOnly` command appears to filter only the suites in `discoveredTestClasses`.
Consequently, a ScalaTest suite annotated with `@DoNotDiscover` cannot be selected explicitly:

```sh
./mill example.test.testOnly TestGroup
```

This fails with:

```text
Test selector does not match any test: TestGroup
Run discoveredTestClasses to see available tests
```

### Minimal example

```scala
import org.scalatest.DoNotDiscover
import org.scalatest.funsuite.AnyFunSuite

@DoNotDiscover
class TestGroup extends AnyFunSuite:
  test("example") {
    assert(true)
  }
```

### Desired behavior

```sh
./mill example.test
# Does not run TestGroup.

./mill example.test.testOnly TestGroup
# Runs TestGroup.
```

### Why this is useful

ScalaTest documents `@DoNotDiscover` as excluding a suite from automatic discovery while still allowing it to be run when explicitly selected:

> ScalaTest will run classes annotated with `DoNotDiscover` if asked to explicitly; it just won't discover them.

At present, Mill's `testOnly` collapses these two concepts: a suite excluded from automatic discovery cannot be explicitly selected.

### Suggested behavior

For an exact class-name selector, if the class is a valid test-framework suite but is absent solely because it is not discoverable, Mill could pass it to the framework as an explicitly selected suite.

### Environment

- Mill: 1.1.8
- ScalaTest: 3.2.20

souvlakias:
To close #7450

**ES = explicitly specified**

### Summary
`TaskDef`'s 3rd argument is `boolean explicitlySpecified`, which is used by some frameworks like ScalaTest to know if a test class was explicitly intended to run. The previous `classFilter: Class[?] => Boolean` just did the glob matching and didn't store any information on what the user requested (with `testOnly`).

This PR creates `class ClassFilter` that includes the old `classFilter: Class[?] => Boolean` and also provides `def isExplicitlySpecified(className: String): Boolean` to determine if a test class was **ES**.

### Problems encountered / Workarounds
- The batch execution logic resolves some globs, so I added a new field to `TestArgs` called `rawSelectors` to be passed to the final `ClassFilter` when the tests run. 
- Marking normally discoverable classes that **were** **ES** as **ES** raised some issues with ScalaTest ignoring filters like `-n`/`-l`. Thus, we leave normally discoverable classes untouched (`_explicitly_specified=false`)

This also adds an `integration.feature` test with the scenario of #7450 (`scalatest.DoNotDiscover`)

### DIFF
--- a/integration/feature/test-do-not-discover/resources/build.mill
+++ b/integration/feature/test-do-not-discover/resources/build.mill
@@ -1,0 +1,14 @@
+package build
+
+// Reproduction of issue https://github.com/com-lihaoyi/mill/issues/7450
+
+import mill.*
+import mill.scalalib.*
+
+object foo extends ScalaModule {
+  def scalaVersion = "3.8.2"
+
+  object test extends ScalaTests with TestModule.ScalaTest {
+    def mvnDeps = Seq(mvn"org.scalatest::scalatest:3.2.19")
+  }
+}
--- a/integration/feature/test-do-not-discover/resources/foo/test/src/NormalSuite.scala
+++ b/integration/feature/test-do-not-discover/resources/foo/test/src/NormalSuite.scala
@@ -1,0 +1,9 @@
+package foo
+
+import org.scalatest.funsuite.AnyFunSuite
+
+class NormalSuite extends AnyFunSuite {
+  test("example") {
+    assert(true)
+  }
+}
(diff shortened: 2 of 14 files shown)


souvlakias:
~~not sure if the failures are related.. probably they are, ill see~~

Remaining failures seem unrelated

PR #7461, opened 2026-08-27
Android: Support Declarative Builds

### LINKED ISSUE #7452: Android: Support declarative (YAML) builds
### Motivation
Have declarative build support for Android projects, for general usecase and for when migrating from gradle builds with `mill init`

### Problem

Android modules need to wire up a shared `AndroidSdkModule` via:
```scala
def androidSdkModule = mill.api.ModuleRef(build.androidSdkModule0)
```
Currently there's now way to pass this in declarative (YAML) builds.

#### Non-blocking problem
Our `androidlib` examples also do:
```scala
override def androidVirtualDevice = super.androidVirtualDevice().withName("java-test")
override def androidEmulatorPort: String = System.getenv.getOrDefault("EMULATOR_PORT", "5554")
```
so this means we can't directly convert them to YAML to showcase examples, but general builds (converted from gradle) will be okay

### Solution ?
Declarative builds have support `Seq[ModuleDep]` by finding such tasks in:
https://github.com/com-lihaoyi/mill/blob/07d23ac50ad014218e6d6347b6cc948d75d8dc8a/core/api/src/mill/api/internal/HeaderData.scala#L3-L15 

### Options
1. Since resolution is name-based currently (and not type-based), we can't directly reuse the `Seq[T]` logic of `moduleDeps`. We would need to add the `androidSDkModule` as another special case and implement whatever is needed (codegen/resolver) for it.

2. Explore how feasible a type-based resolution is, so we can generalize for any traits needing `ModuleRef` or `Seq[Module]`

cc @vaslabs

souvlakias:
Closes #7452 
### Problems
- [x] Be able to declare a `ModuleRef[AndroidSdkModule]` for  `def androidSdkModule`
- [x] Some `AndroidModule` fields like `androidNamespace: String` are plain scalar fields (not tasks, not part of the taskgraph) and aren't picked up by declarative builds
- [x] Mima issues

### Changes
- Generalize the logic of `ModuleDepsResolver` a bit so it can also resolve a `ModuleRef`
  - Added an **exception for the mima check** for this one (internal api)
  ⚠️ Please verify this, since I'm not 100% sure
- Add `androidSdkModule` to the hardcoded keys that get picked up and handled separately.
- Converted scalar fields of `AndroidModule` to now be tasks so they can be configured from YAML. 
  **Left as scalar**:
  - `enableBuildConfig`, `androidIncludeAndroidResources`, `androidEnableViewBinding`: We kept these as non-tasks so the task graph would be computed with those flags in mind, meaning we avoid evaluating irrelevant branches (which could fail when some tasks weren't configured). 
  **Non blocking** for (most?) basic YAML builds
  - `androidReleaseKeyName`: `Task.Source` can't evaluate `Task`
  **Blocking** for YAML builds that would want to build the **release** APK
- Also add an example `androidlib.kotlin.1-hello-kotlin-declarative` and document it


souvlakias:
>  Some `AndroidModule` fields like `androidNamespace: String` are plain scalar fields (not tasks, not part of the taskgraph) and aren't picked up by declarative builds

@vaslabs Don't know if these were intended to be scalar fields, but I think we can safely convert them to tasks (e.g. `T[String]`... I'll verify

**Update**: Thinking about it, we did deliberately choose to have some non-task fields like `androidIncludeAndroidResources` to have something like:
https://github.com/com-lihaoyi/mill/blob/d3c76ff2e5f1b8284d991982241bb19e1207d232/libs/androidlib/src/mill/androidlib/AndroidModule.scala#L972-L979

but maybe its better to just have it as a task? 🤔️

vaslabs:
> > Some `AndroidModule` fields like `androidNamespace: String` are plain scalar fields (not tasks, not part of the taskgraph) and aren't picked up by declarative builds
> 
> @vaslabs Don't know if these were intended to be scalar fields, but I think we can safely convert them to tasks (e.g. `T[String]`... I'll verify
> 
> **Update**: Thinking about it, we did deliberately choose to have some non-task fields like `androidIncludeAndroidResources` to have something like:
> 
> https://github.com/com-lihaoyi/mill/blob/d3c76ff2e5f1b8284d991982241bb19e1207d232/libs/androidlib/src/mill/androidlib/AndroidModule.scala#L972-L979
> 
> but maybe its better to just have it as a task? 🤔️

I am not entirely sure, I think these booleans being tasks may cause problems e.g. with computing the task graph, both branches would be executed, I remember some things like that back then. but give it a go .

Kotlin would have similar things if that's the case? how is it handle there (e.g. with the ksp settings)

souvlakias:
> I am not entirely sure, I think these booleans being tasks may cause problems e.g. with computing the task graph, both branches would be executed, I remember some things like that back then. but give it a go .

Indeed, noting:
- `enableBuildConfig`, `androidIncludeAndroidResources`, `androidEnableViewBinding`: We kept these as non-tasks so the task graph would be computed with those flags in mind, meaning we avoid evaluating irrelevant branches (which could fail when some tasks weren't configured). 
  **Non blocking** for basic YAML builds
- `androidReleaseKeyName`: `Task.Source` can't evaluate `Task`
  **Blocking** for YAML builds that would want to build the release APK

Thus, keeping these as `scalar` fields for now

souvlakias:
What's now left are some [mima](https://github.com/vaslabs-ltd/mill/actions/runs/33157916741/job/98807235118) issues, due to renamings.

As I understand, these can be added as exceptions since they are internal tooling. Ref:
https://github.com/com-lihaoyi/mill/blob/82ecc427f0f2e8e5fd2dd6f42402b048a38e7013/mill-build/src/millbuild/MillStableJavaModule.scala#L104-L106

So either add as exceptions or revert the renamings

Update: added the `ProblemFilter.exclude` exceptions

vaslabs:
These test failures look unrelated, right @souvlakias ? It's passed on our fork I think

souvlakias:
> These test failures look unrelated, right @souvlakias ? It's passed on our fork I think

Yes I think so too

vaslabs (approved):

PR #7469, opened 2026-08-28
Contrib: SBOM Add purl

gamlerhart:
Add the `purl` to the generated SBOM,
as many tools (eg. SonarQube) require it.

purl reference: https://github.com/package-url/purl-spec


lefou:
Thank you!

PR #7472, opened 2026-08-29
Fix PGP signing with exported secret subkeys

### LINKED ISSUE #7052: Make PGP signing for Maven (Central) publishing more robust
I ran into two issues while setting up `publishSonatypeCentral`:

First, I couldn't find a way to locally test PGP signing in isolation, as I'm used to from Gradle. Instead, I had to run `publishSonatypeCentral` with fake Sonatype credentials, which isn't ideal.

Second, Mill can fail with an NPE when the signing key is exported with `gpg --export-secret-subkeys`.

As far as I know, using a dedicated signing subkey is a good practice. The primary key remains the long-lived identity key, while signing is done with a replaceable subkey. For CI, this also avoids putting the primary private key into CI secrets.

I exported the subkey as follows:

    gpg --armor --export-secret-subkeys <signing-subkey-id> > signing-subkey.asc

Such an export still includes the primary key metadata, but with dummy private key material. The primary key can still be marked signing-capable.

Mill appears to select the first `PGPSecretKey` where `key.isSigningKey` is true. With this kind of export, that may be the dummy primary key. Private key extraction then returns `null`, and signing fails with:

    java.lang.NullPointerException: Cannot invoke "org.bouncycastle.openpgp.PGPPrivateKey.getPublicKeyPacket()" because "<parameter1>" is null
      org.bouncycastle.openpgp.operator.bc.BcPGPKeyConverter.getPrivateKey(Unknown)
      org.bouncycastle.openpgp.operator.bc.BcPGPContentSignerBuilder.build(Unknown)
      org.bouncycastle.openpgp.PGPSignatureGenerator.init(Unknown)
      mill.javalib.pgp.worker.PgpSignerWorker.signDetached(PgpSignerWorker.scala:51)

I worked around this by explicitly selecting the signing subkey:

    override def sonatypeCentralGpgArgs = Task {
      "--local-user,<signing-subkey-id>"
    }

It would be good if Mill either skipped signing-capable keys whose private key cannot be extracted, or failed with a clear error message.

Mill Build Tool version 1.1.5-18-e69f7b
java.version: 21.0.10
java.vendor: Eclipse Adoptium
file.encoding: UTF-8
os.name: Windows 11
os.version: 10.0
os.arch: amd64

lihaoyi:
Vibe Coded

Skip signing-capable secret-key stubs that contain no private material, while preserving explicit key selection and extraction failures. Add a GnuPG exported-subkey fixture covering automatic selection and clear error behavior.

Fixes #7052


lefou:
This PR is not merged to main

PR #7476, opened 2026-08-29
Document SemanticDB version configuration

### LINKED ISSUE #4901: Documentation is missing info about `ScalaModule.semanticDbVersion`
We provide defaults for this version, so users typically aren't aware of that config task. But if they don't fix, users see resolution errors without further details.

* [ ] We should cover this topic in the documentation. Currently, a page search for semanticdb doesn't bring anything.
* [ ] Maybe we can improve this specific resolution message.

lihaoyi:
Vibe Coded

Explain the Scala, Java, environment, and explicit-override version-selection rules with a runnable mixed-source example. Correct the Java SemanticDB failure message to name semanticDbJavaVersion and cover the selection and diagnostic behavior.

Fixes #4901


lefou:
This PR is not merge to main

PR #7479, opened 2026-08-29
Fix SemanticDB paths with custom JVMs

### LINKED ISSUE #7376: Mill 1.2.0-RC1 builds incorrect semanticDbData when jvmVersion is specified in ScalaModule
Reproducible example:

build.mill:
```scala
//| mill-version: 1.2.0-RC1-42-dc1f60
package build

import mill.*, scalalib.*

object `package` extends ScalaModule {
  def scalaVersion = "3.8.4"
  def jvmVersion = "21"
}
```

src/main.scala:
```scala
@main def main(): Unit = {
  println("Hello World")
}
```

Running `./mill semanticDbData` produces an empty directory at `out/semanticDbDataDetailed.dest/data/META-INF/semanticdb/out/mill-daemon/mill-workspace/src/`

Removing the `jvmVersion` line yields the expected result: `out/semanticDbDataDetailed.dest/data/META-INF/semanticdb/src/main.scala.semanticdb`

This looks like relative paths are being incorrectly resolved in a separate JVM process.

lihaoyi:
Vibe Coded

Resolve reproducible workspace aliases before Zinc exposes source paths to compiler plugins, and pass SemanticDB a resolved absolute sourceroot. This keeps forked compiler workers in the same path coordinate system as local workers.

Cover Scala 2 and Scala 3 custom-JVM modules while retaining the existing default-JVM and incremental SemanticDB coverage.

Fixes #7376

### DIFF
--- a/libs/javalib/worker/src/mill/javalib/zinc/ZincWorker.scala
+++ b/libs/javalib/worker/src/mill/javalib/zinc/ZincWorker.scala
@@ -430,8 +430,12 @@ class ZincWorker(jobs: Int, useFileLocks: Boolean = false) extends AutoCloseable
     val classpath = (compileClasspathPaths.iterator ++ Some(classesDir))
       .map(path => converter.toVirtualFile(path.toNIO))
       .toArray
+    // Compiler plugins compare virtual source paths with options such as SemanticDB's sourceroot.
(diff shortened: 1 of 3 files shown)


lefou:
This PR is not merged to main

Issue #4481, opened 2025-02-05
Include sub-modules in `inspect <module>`


lefou:
There is currently no way to find sub-modules. `resolve` always lists tasks.

lihaoyi:
Doesn't `resolve` list both tasks and submodules?

lefou:
> Doesn't `resolve` list both tasks and submodules?

Yes, when used with wildcards. But there is no way to differentiate between modules and tasks in the result, and the list can become quite long.

lefou:
Fixed by https://github.com/com-lihaoyi/mill/pull/7508

PR #7508, opened 2026-08-31
Show direct submodules in inspect output

lihaoyi:
inspect exposed a module's tasks and metadata but omitted its immediate child modules, so callers could not discover the module hierarchy from the inspection result.

Include direct child modules in inspect output while preserving the existing task output. Unit and integration coverage exercise nested modules and modules that also define tasks.


lefou:
Fix https://github.com/com-lihaoyi/mill/issues/4481

PR #7509, opened 2026-08-31
Document SemanticDB version configuration

### LINKED ISSUE #4901: Documentation is missing info about `ScalaModule.semanticDbVersion`

lihaoyi:
The Scala and Java SemanticDB version-selection hooks were undocumented, and the Java failure message named a task that does not exist.

Document the defaults, environment-requested minimums, and explicit overrides. Add a checked mixed-language example and focused tests, and correct the Java diagnostic to name semanticDbJavaVersion.

Fix https://github.com/com-lihaoyi/mill/issues/4901

### DIFF
--- a/example/scalalib/module/19-semanticdb-versions/build.mill
+++ b/example/scalalib/module/19-semanticdb-versions/build.mill
@@ -1,0 +1,38 @@
+// Mill uses SemanticDB metadata for IDE features and other source-code tooling.
+// It provides default plugin versions, so most builds do not need any SemanticDB
+// configuration.
+//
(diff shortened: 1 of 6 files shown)


lefou:
Fix https://github.com/com-lihaoyi/mill/issues/4901

PR #7510, opened 2026-08-31
Restore provenance headers to bootstrap scripts

### LINKED ISSUE #6157: Add a header to bootstrap script with some infos

lihaoyi:
Generated POSIX and Windows bootstrap scripts did not identify their project, download source, documentation, or generated Mill version.

Add that metadata to both script templates using canonical build settings, remove substitutions no longer consumed by either template, and verify the exact generated headers.

* Fix https://github.com/com-lihaoyi/mill/issues/6157

### DIFF
--- a/dist/scripts/package.mill
+++ b/dist/scripts/package.mill
@@ -10,6 +10,9 @@ import millbuild.*
 
 object `package` extends mill.Module { scripts =>
 
+  private val bootstrapDocumentationUrl =
+    s"${Settings.docUrl}/mill/cli/installation-ide.html#_bootstrap_scripts"
+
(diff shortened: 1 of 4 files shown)


lefou:
* Fix https://github.com/com-lihaoyi/mill/issues/6157

PR #7517, opened 2026-08-31
Allow Java modules to override BSP dependency sources

### LINKED ISSUE #7140: bspBuildTargetDependencySources aggregate causing hang when sources cant be found (via metals build import)
bspBuildTargetDependencySources will always fail in my project as i am using a private repo that does not have any sources. I would like to override this so that there is no long wait times.

another diagnosis from Claude Code, suggesting that exceptions are escaping so a promise never completes:

> That's the key insight. The task evaluators stopped 7+ minutes ago (err.log last write 12:03, now 12:10) but Mill BSP still hasn't replied to Metals — mill-active.json still claims IDE:buildTargetDependencySources is in flight, and Metals keeps heartbeating. So this isn't slow downloads — it's a deadlock/hang in Mill's BSP server when every bspBuildTargetDependencySources task fails.
> 
> Let me grab a thread dump from the BSP JVM to point at the actual stuck frame.
> 
>   Ran 2 shell commands
> 
> Confirmed it's a deadlock in Mill BSP itself, not slow downloads. The thread dump pins the cause:
> 
> "mill-bsp-evaluator" #33  WAITING (parking)
>   at scala.concurrent.Await.result(package.scala:209)
>   at mill.exec.Execution.evaluateTerminals$1$$anonfun$2(Execution.scala:352)
>   at mill.exec.Execution.execute0$$anonfun$1(Execution.scala:362)
>   ...
>   at mill.bsp.worker.MillBuildServer.evaluate(MillBuildServer.scala:424)
>   at mill.bsp.worker.MillBuildServer.handlerTasks(MillBuildServer.scala:238)
> 
> The BSP handler is doing Await.result(...) on a Future inside Execution.evaluateTerminals and never returns. When every per-module bspBuildTargetDependencySources task throws (Coursier 404 on -sources.jar), Mill 1.1.6's executor doesn't fail-fast back to the aggregating Promise — the futures wrap individual task outcomes, but the orchestrator keeps Await-ing for a value that the failure path never produces. The 30 task failed lines stopped flowing into err.log at 12:03; no new work has happened in the ~8 min since. Metals is fine — its periodic buildTarget/compile (8..17) are just heartbeats while it waits on response id 7.

lihaoyi:
BSP dependency-source reporting resolved sources for every Maven dependency with no module-level escape hatch, so dependencies that do not publish sources could leave clients waiting indefinitely.

Expose bspMvnDependencySources on JavaModule and consume it from BSP, allowing modules to omit or replace unavailable sources. Android supplies its AAR source jars through the same hook, and BSP tests cover an explicit override.

* Fix https://github.com/com-lihaoyi/mill/issues/7140

### DIFF
--- a/integration/dedicated/bsp-server/src/BspServerTests.scala
+++ b/integration/dedicated/bsp-server/src/BspServerTests.scala
@@ -30,6 +30,92 @@ object BspServerTests extends UtestIntegrationTestSuite {
     super.workspaceSourcePath / "project"
 
   def tests: Tests = Tests {
+    test("dependencySourcesWithoutSourceArtifact") - integrationTest { tester =>
+      import tester.*
+
+      def publishLocalArtifact(name: String, withSources: Boolean): Unit = {
+        val artifactDir = workspacePath / "repo/example" / name / "1.0"
+        os.makeDir.all(artifactDir)
+        os.write(
+          artifactDir / s"$name-1.0.pom",
+          s"""<project>
+             |  <modelVersion>4.0.0</modelVersion>
+             |  <groupId>example</groupId>
+             |  <artifactId>$name</artifactId>
+             |  <version>1.0</version>
+             |</project>
+             |""".stripMargin
+        )
+        os.write(artifactDir / s"$name-1.0.jar", "binary")
+        if (withSources) os.write(artifactDir / s"$name-1.0-sources.jar", "sources")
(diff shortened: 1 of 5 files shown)


lefou:
* Fix https://github.com/com-lihaoyi/mill/issues/7140

PR #7521, opened 2026-08-31
Notify BSP clients after build-definition changes

### LINKED ISSUE #7409: New Mill project in IDEA problems
I created new Mill project by hand, with basic Spring Boot dependencies. And opened it in IDEA as BSP project. IDEA constantly can't see dependencies in editor and shows false errors in code. Invalidating cache and restarting every 10 minutes help. Is there anything that can be done to fix it?

lihaoyi:
BSP target snapshots omit endpoint data such as compile classpaths and Java homes. After the reported dependency or JVM configuration edits, Mill could reload the build without sending didChange, leaving IDEA with stale model data.

Carry staleness from the previous build-definition watch set into change detection and conservatively invalidate retained targets after a real build edit. Preserve pending notifications through broken-build recovery while ignoring watches discovered during startup, with reload coverage for dependencies and JVM configuration.

* Fix https://github.com/com-lihaoyi/mill/issues/7409

### DIFF
--- a/integration/dedicated/bsp-server-reload/resources/project/build.mill.yaml.configuration-base
+++ b/integration/dedicated/bsp-server-reload/resources/project/build.mill.yaml.configuration-base
@@ -1,0 +1,5 @@
+mill-jvm-version: system
+extends: JavaModule
+jvmVersion: 17
+mvnDeps:
+  - org.slf4j:slf4j-api:2.0.16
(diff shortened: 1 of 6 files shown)


lefou:
* Fix https://github.com/com-lihaoyi/mill/issues/7409

PR #7524, opened 2026-08-31
Discover non-public overrides for task path mapping

### LINKED ISSUE #7080: `mill.idea/` hangs (Mill after 1.1.6)
I think this happened after 
- #6955 

Repro:
```
$ cd example/androidlib/java/1-hello-world
$ rm -rf out/ && mill mill.idea/              
```
Then it hangs
```                                                                   
build.mill-68] compile compiling 1 Scala source to out/mill-build/compile.dest/classes ...
build.mill-68] done compiling
73/73+] mill mill.idea/ 8s
Analyzing modules ...
^C%                                                                         
```

lihaoyi:
Fix #7080

Fine-grained locking distinguishes an override from its inherited implementation through the task override map. Discover only recorded CLI-visible tasks, so the package-private overrides in GenIdeaAndroidModule were absent and both task nodes received the same output path. mill.idea then waited for a lock already held by its own dependency graph.

Record concrete overridable named tasks separately from CLI entrypoints and use that metadata only for task path mapping. Follow ModuleRef result types so internal IDEA wrapper modules are discovered without evaluating the references, while excluding abstract and truly private methods and preserving the existing ClassInfo constructor.

Cover visibility and compatibility rules, the Android task graph paths, and the original mill.idea daemon reproduction under a strict timeout.

### DIFF
--- a/core/api/src/mill/api/Discover.scala
+++ b/core/api/src/mill/api/Discover.scala
@@ -52,9 +52,15 @@ final class Discover(val classInfo: Map[Class[?], Discover.ClassInfo]) {
 object Discover {
   final class ClassInfo(
       val entryPoints: Seq[mainargs.MainData[?, ?]],
-      val declaredTasks: Seq[TaskInfo]
+      val declaredTasks: Seq[TaskInfo],
+      @com.lihaoyi.unroll val pathTaskNames: Seq[String] = null
   ) {
     lazy val declaredTaskNameSet = declaredTasks.map(_.name).toSet
+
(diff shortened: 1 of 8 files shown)


lefou:
Looks like this PR causes new failures in the test suite.

* https://github.com/com-lihaoyi/mill/issues/7562
* https://github.com/com-lihaoyi/mill/issues/7561

Not sure, why CI didn't caught it, probably due to selective testing.

@lihaoyi, could you have a look?

PR #7536, opened 2026-09-07
JMH plugin: use forkArgs when invoking JMH

asflierl:
My use case:
The txt output of JMH depends on the Java default Locale for number formats, e.g. on a German system we'd get `1,234` while on a US system that would be `1.234`. Furthermore, the only way to set this to anything else than the system Locale on some systems (e.g. MacOS) is by passing it as properties (`user.language` and `user.country`) to the JVM at startup.

For my use case, but also in general, I think it is valuable to use the module's `forkArgs` when the plugin starts the main JMH JVM - in a similar fashion to what #3957 did.


asflierl:
I don't think any of the test failures in CI have anything to do with my change. 🤔 💭

PR #7297, opened 2026-06-20
Deduplicate reverse dependency tree to fix OOM errors in coursier

lefou:
This is currently not optional, since the non-de-duplicated trees tend to either take very long to render or break before finish with an out of memory error.

Fix https://github.com/com-lihaoyi/mill/issues/6823

PR #7548, opened 2026-09-09
Deduplicate the non-inverted dependency tree to fix `showMvnDepsTree` hanging

h2000:
Fix https://github.com/com-lihaoyi/mill/issues/6823

`showMvnDepsTree` never returns on large dependency graphs. This makes the non-inverted tree deduplicate repeated nodes, the same way #7297 already did for the inverted one.

## Root cause

Two compounding problems, both in coursier's non-inverted tree renderer.

**1. The renderer walks paths, not nodes.** `Tree.customRender0` only guards against cycles (a path-local `ancestors` set); any node reachable by several routes is fully re-expanded at each one. The output therefore grows with the number of distinct *paths* through the graph, not with the number of dependencies.

Measured on a ~460-dependency Kotlin/Spring/Vaadin resolution (`spring-boot-starter-web` + `vaadin-spring-boot-starter` + `jackson-module-kotlin` + `kotlin-reflect` + `kotlinx-coroutines-core-jvm`):

| | |
|---|---|
| distinct nodes in the graph | 153 |
| nodes visited by the current renderer | **> 20,000,000** (capped — still growing) |
| nodes visited with deduplication | 1,192 |

It does not finish, and with a bounded heap it OOMs — matching the `OutOfMemoryError` reported in the issue.

**2. Each re-expansion is expensive.** `children` is deliberately a `def`, so every visit recomputes it, including this block:

```scala
// coursier DependencyTree.scala:160-167
val globalOverrides = resolution.projectCache0
  .get(dependency.moduleVersionConstraint)
  .map(_._2.overrides.filter((_, v) => v.global).flatten.toSeq)
  .getOrElse(Nil)
  .collect {
    case (k, v) if resolution.dependencySet.containsModule(k.fakeModule) => ...
  }
```

`DependencySet.containsModule` is `grouped.exists(_._1.module == mod)` — a linear scan of the whole dependency set, run once per global override, per visit. That is the `containsModule` frame at the top of both thread dumps in the issue (`DependencyTree.scala:164/165` matches these exact lines in 2.1.25-M26, which both 1.1.8 and 1.1.9 use).

**Why `cs` is fast on the same coordinates.** Those global overrides only exist when Gradle Module metadata is read: `global` is set exclusively via `Overrides#enforceGlobalStrictVersions`, reached only from `Resolution.scala:2112` for BOM imports carrying `endorseStrictVersions`, which in turn only `GradleModule.scala:102` sets. Mill's `KotlinModule` (and `GroovyModule`, `AndroidModule`) sets `checkGradleModules = true`, so Mill resolves those where a plain `cs resolve -t` does not. Confirmed directly on the same coordinates:

| | global overrides |
|---|---|
| plain resolution | **0** |
| `checkModule = true` + Kotlin variant attributes | **297** |

The reproduction in the issue is a Kotlin module (`cloud.vaadin extends AcmeKotlinModule`), which is why it is hit so hard.

Problem 1 alone is enough to hang; problem 2 turns each of those millions of visits into a full dependency-set scan.

## Fix

Coursier's `Print.dependencyTree0` exposes `deduplicateNodes` only for the inverted direction (`reverseDeduplicateNodes`, added in coursier/coursier#3671 and adopted in #7297). Rather than wait on an upstream release, `mill.javalib.internal.DepsTreeRenderer` renders the non-inverted tree in Mill, mirroring coursier's private `Print#render` for node labels.

A dependency reached by more than one path is expanded once and referenced elsewhere. Every occurrence, the expanded one included, carries a `(*n)` reference so the expansion can be found:

```
├─ com.lihaoyi:upickle_2.13:1.4.0
│  ├─ com.lihaoyi:ujson_2.13:1.4.0
│  │  └─ com.lihaoyi:upickle-core_2.13:1.4.0 (*1)
│  │     └─ com.lihaoyi:geny_2.13:0.6.10
│  ├─ com.lihaoyi:upack_2.13:1.4.0
│  │  └─ com.lihaoyi:upickle-core_2.13:1.4.0 (*1)
│  └─ com.lihaoyi:upickle-implicits_2.13:1.4.0
│     └─ com.lihaoyi:upickle-core_2.13:1.4.0 (*1)
└─ org.scala-lang:scala-library:2.13.10
```

Numbering requires knowing in advance which nodes are referenced later, so only those get a number — otherwise the ~90% of nodes that appear once would all be tagged. A pre-pass finds the nodes reached by more than one path; numbers are then assigned in printed order.

### Why the same coordinates can carry different numbers

Nodes are identified as coursier identifies them, by the `Dependency` they were reached with rather than by module and version, because `DependencyTree.Node#children` is computed from that `Dependency`. Two occurrences that differ only in *exclusions* therefore have genuinely different subtrees.

On the Spring/Vaadin resolution: 506 distinct nodes, 104 of which render to a label that some other node also renders to, and **26 of those have a different child set** — e.g. `io.netty:netty-transport-native-kqueue:4.1.135.Final` appears both with no exclusions (5 children) and with 6 exclusions (0 children). Collapsing by rendered label would point the reader at an expansion that is not theirs.

Checked the other direction too: widening the key to `(dependency, excluded, endorsed)` — the exact fields `children` reads — yields the same 506 classes, so coursier's identity is not too fine either. Documented in the scaladoc and in `library-deps.adoc`.

## Behaviour change

Output is unchanged except that a dependency appearing in several places is expanded only the first time, and repeated dependencies carry a `(*n)` reference. For graphs with no repeated nodes the output is byte-identical to before. Five expected trees in `CrossVersionTests` were updated; every change is a collapsed repeat, nothing else.

`DepsTreeRenderer` lives in `mill.javalib.internal` and is marked `@internal`, following `mill.javalib.internal.PublishModule`. No public API changed.

The inverted tree (`--inverse`, `--whatDependsOn`) still uses coursier's plain `(*)`, since it goes through `Print.dependencyTree0(reverseDeduplicateNodes = true)`. Making that consistent means reimplementing the `ReverseModuleTree` rendering as well; happy to fold that in if wanted.

## Verification

**On the reproduction from the issue** ([lefou/repro-mill-showMvnDepsTree-issue-6823](https://github.com/lefou/repro-mill-showMvnDepsTree-issue-6823), `cloud.vaadin`):

| | before | after |
|---|---|---|
| `showMvnDepsTree` | never returns | **6.8s** warm, 3272 lines, 463 references |
| `--inverse` | (already fixed by #7297) | 2.4s |
| `--whatDependsOn kotlin-stdlib` | did not return | 2.4s |

`--whatDependsOn` returning for heavily depended-on dependencies addresses the "works for some dependencies but not others" symptom reported in the issue.

**Output fidelity** — rendered via this code path vs. `Print.dependencyTree0(reverse = false)`:

| dependency | identical | lines (new → old) | time |
|---|---|---|---|
| `org.slf4j:slf4j-api:1.7.35` | yes | 1 / 1 | — |
| `com.lihaoyi::os-lib:0.9.1` | yes | 4 / 4 | — |
| `com.lihaoyi::upickle:1.4.0` | dedup only, same node set | 8 / 10 | — |
| `org.apache.spark::spark-sql:4.1.1` | dedup only, same node set | 1,292 / 51,248 | 21ms / 131ms |

**Tests**: `libs.scalalib.test.testOnly mill.scalalib.CrossVersionTests` 7/7, `integration.feature[inspect].testForked` 1/1, `mill.scalalib.scalafmt.ScalafmtModule/scalafmt --check` clean.

## Not addressed

The `resolvedMvnDeps` hangs mentioned later in the issue thread are a separate cause — nothing in that path renders a dependency tree.


lefou (commented):
Very glad you found the issue and also provide a solution, since this issue is really my major blocker to upgrade a larger project to Mill 1.x. I'll test it and report later.

I'd like to have the `DepsTreeRenderer` class in a `mill.javalib.internal` package, to make its internal purpose even more clear. Although already an `private[mill]` class, in the resulting bytecode it ends up being public.

How hard would it be, to replace the `(*)` duplication marker with a numbered reference like `(*1)`, so that it is easier to find the matching fully expanded tree?

lefou:
> How hard would it be, to replace the `(*)` duplication marker with a numbered reference like `(*1)`, so that it is easier to find the matching fully expanded tree?

That implies, that the first full sub-tree also needs the marker.

h2000:
Thanks — both done in b479af8b44.

**`mill.javalib.internal`** — moved, and marked `@internal`, following `mill.javalib.internal.PublishModule`. Good catch on `private[mill]` not surviving into the bytecode.

**Numbered references** — done, `(*)` is now `(*n)`, and as you noted the expanded occurrence carries the number too:

```
├─ com.lihaoyi:upickle_2.13:1.4.0
│  ├─ com.lihaoyi:ujson_2.13:1.4.0
│  │  └─ com.lihaoyi:upickle-core_2.13:1.4.0 (*1)
│  │     └─ com.lihaoyi:geny_2.13:0.6.10
│  ├─ com.lihaoyi:upack_2.13:1.4.0
│  │  └─ com.lihaoyi:upickle-core_2.13:1.4.0 (*1)
│  └─ com.lihaoyi:upickle-implicits_2.13:1.4.0
│     └─ com.lihaoyi:upickle-core_2.13:1.4.0 (*1)
└─ org.scala-lang:scala-library:2.13.10
```

Not hard, but it did need a restructure. Coursier's `Tree.customRender0(deduplicateNodes = true)` hardcodes the `(*)` string, so the renderer no longer delegates to it; it walks the tree itself (same prefix/`├─`/`└─` logic). Numbering also needs to know *in advance* which nodes get referenced later — otherwise every node would get a number, including the ~90% that appear once — so there is now a pre-pass that finds the nodes reached by more than one path. Numbers are then assigned in printed order, so `(*1)` is the first repeated dependency you meet reading top to bottom.

**One thing worth flagging**, because it looks like a bug at first glance: the same coordinates can appear with *different* numbers.

```
│  │  ├─ com.fasterxml.jackson.core:jackson-annotations:2.22 (*2)
│  │  │  │  ├─ com.fasterxml.jackson.core:jackson-annotations:2.22 (*6)
```

I looked into whether to collapse those, and they should not be. Nodes are identified by the `Dependency` they were reached with, not by module+version, because `DependencyTree.Node#children` is computed from that `Dependency` — so two occurrences differing only in *exclusions* have genuinely different subtrees. On the Spring/Vaadin resolution: 506 distinct nodes, 104 of which render to a label some other node also renders to, and **26 of those really do have a different child set**. For example `io.netty:netty-transport-native-kqueue:4.1.135.Final` appears both with no exclusions (5 children) and with 6 exclusions (0 children). Merging by rendered label would point you at an expansion that is not yours.

I also checked the other direction — whether coursier's node identity is *too* fine and could be safely widened to `(dependency, excluded, endorsed)`, which is the exact set of fields `children` reads. It yields the same 506 classes on that resolution, so there is nothing to gain; the renderer keeps coursier's identity. Both the scaladoc and the user docs now explain this.

**Still using `(*)`, not `(*n)`: the inverted tree.** `--inverse` and `--whatDependsOn` go through coursier's `Print.dependencyTree0(reverseDeduplicateNodes = true)` (from #7297), which emits the plain marker. Making that consistent means reimplementing the `ReverseModuleTree` rendering too — happy to do it in this PR if you'd like, I left it out to keep the change contained.

Test fixtures and `library-deps.adoc` updated; `CrossVersionTests` 7/7, `integration.feature[inspect]` 1/1, scalafmt clean. On your reproduction the output is unchanged at 3272 lines, with 463 reference numbers, each used at least twice.

lefou:
Great, thanks. Looks like the numbered version is more correct, since otherwise, a user might choose to assume a wrong tree, so I think using the `(*n)` version is superior to the plain `(*)` marker, even if it takes longer.

lefou:
> The inverted tree (`--inverse`, `--whatDependsOn`) still uses coursier's plain `(*)`, since it goes through `Print.dependencyTree0(reverseDeduplicateNodes = true)`. Making that consistent means reimplementing the `ReverseModuleTree` rendering as well; happy to fold that in if wanted.

I'd rather not do this in this PR. But for consistency, this might come as a separate PR. 

Are there also correctness implications where a numbered reference is more correct? Like in the non-inverted tree case, where the same dependency (or rather a dependency with the same printed name but different exclusions) might have multiple different trees.

lefou (approved):

h2000:
**CI: 35 of 36 green, one job hit the 60-minute timeout.**

`linux (example.androidlib._._, true)` was cancelled, not failed — it ran 14:27:58 → 15:32:58 and was killed by the `timeout-minutes: 60` in `post-build-selective.yml`. I don't have rights to re-run it; could someone kick it?

I checked whether it could plausibly be caused by this PR, and it can't:

- **The job doesn't reach the changed code.** `DepsTreeRenderer` is only called from `JavaModule#renderDepsTree`, which is only reachable from `mvnDepsTree`/`showMvnDepsTree`. There is no `MvnDepsTree` usage anywhere under `example/`, so `example.androidlib._._` never enters this path.
- **The same job passed on the previous run of this PR**, in 22 minutes ([run 34359447572](https://github.com/com-lihaoyi/mill/actions/runs/34359447572)), against the same code path — the second commit only moved the class between packages and changed the elision marker.
- The job carries `setup-android: true`, so it downloads the Android SDK before doing any Mill work.

For reference, the first run of this PR also shows as `cancelled` at the workflow level, but that one was mine — pushing the review commit cancelled it via `cancel-in-progress: true`. In that run every job including `example.androidlib` had already finished green.

The three jobs that do cover this change all passed on the current head (b479af8b44):

| job | result |
|---|---|
| `linux ('libs.scalalib.__.test')` | pass — includes the updated `CrossVersionTests` fixtures |
| `linux ('integration.feature._')` | pass — includes the `inspect` test that runs `showMvnDepsTree` |
| `lint-autofix` | pass — scalafmt plus the autofix check over the package move |

h2000:
> Are there also correctness implications where a numbered reference is more correct? Like in the non-inverted tree case, where the same dependency (or rather a dependency with the same printed name but different exclusions) might have multiple different trees.

Sorry, this slipped past me before the merge. Short answer: **no — the inverted tree does not have that hazard**, so a `(*n)` follow-up there would be purely for navigability, not correctness.

The reason is structural. `ReverseModuleTree.fromDependencyTree` builds its dependee map keyed by `Module` alone:

```scala
val dependees = new mutable.HashMap[Module, mutable.HashSet[(Module, VersionConstraint, Boolean, Boolean)]]
```

Every `Dependency` for a given module — different exclusions included — is merged into one dependee set while the map is being built. A node's subtree is then looked up from that map, so it is a function of `module`, and every rendered label contains the module. Two inverted-tree nodes that print the same therefore always expand to the same thing. The forward tree distinguishes nodes by the full `Dependency` precisely because `DependencyTree.Node#children` is computed from it, which is where the exclusion-dependent subtrees come from; the inverted tree has already collapsed that distinction before rendering starts.

Checked it rather than trusting the reading, on the same Spring/Vaadin resolution:

| | |
|---|---|
| distinct inverted-tree nodes | 765 |
| labels backed by more than one node | 132 |
| of those, with a differing subtree | **0** |

Compare the non-inverted tree on the same resolution: 506 nodes, 104 shared labels, **26** with a genuinely different subtree.

So the two directions are safe for different reasons — the non-inverted one because it keys on the `Dependency` and numbers the references, the inverted one because there is only ever one subtree per module to point at. Happy to do the consistency follow-up whenever it is wanted.

PR #7519, opened 2026-08-31
Preserve Quarkus executable dependency metadata

lihaoyi:
Quarkus gRPC code generation requires platform-specific protoc artifacts selected by type and classifier. Mill omitted executable artifacts and modeled every resolved file as an unclassified jar.

Include executable artifacts in Quarkus resolution, preserve their exact coordinates in the application model, avoid scanning executables as jars, and pass Quarkus its build directory. Tests verify classifier preservation and execute gRPC source generation.

* Fix https://github.com/com-lihaoyi/mill/issues/7331

PR #7552, opened 2026-09-10
Fix/quarkus exe artifact resolution and dependencies with missing os placeholders

### LINKED ISSUE #7331: Quarkus: Fix protoc dependency
CodeGen fails for Grpc examples on Quarkus CodeGen Worker (e.g. on Linux x64) with missing protoc dependency (protobuf compiler)

```console
io.quarkus.bootstrap.prebuild.CodeGenException: Failed to find com.google.protobuf:protoc:linux-x86_64:exe among dependencies
```

Reproducible example: https://github.com/quarkusio/quarkus-quickstarts/tree/main/grpc-plain-text-quickstart

vaslabs:
fixes https://github.com/com-lihaoyi/mill/issues/7331 and other dependency issues found in quarkus quick examples

Made with initial vibes and then some handcrafted changes

## Summary

Fixes two dependency-resolution regressions in `QuarkusModule` introduced by #7519
("Preserve Quarkus executable dependency metadata"), both surfacing while running
`example.migrating.javalib.6-maven-quarkus` and while building real-world Quarkus
projects (e.g. the `quarkus-quickstarts` repo) with Mill:

1. Coursier fails to resolve dependencies whose Maven coordinates rely on
   `${os.detected.classifier}`-style properties (e.g. `io.netty:netty-tcnative`, a
   transitive, optional dependency of `io.netty:netty-handler`), because that
   property is normally injected by the `os-maven-plugin`/`osdetector` Maven/Gradle
   build extensions, which Coursier has no equivalent for.
2. Dependencies published with a Maven packaging type other than the literal `jar`
   (e.g. `jakarta.ws.rs:jakarta.ws.rs-api`, packaged as an OSGi `bundle`) were
   silently dropped from Quarkus's runtime/deployment classloaders during
   `quarkusCodeGen`, causing `ClassNotFoundException`/`NoClassDefFoundError` for
   basically any Quarkus app using `quarkus-rest`.
3. Enabled all quarkus examples (removed the `rm grpc-plain-text-quickstart/package.mill.yaml  grpc-tls-quickstart/package.mill.yaml` . They are all in a working state now

## Problem 1: `${os.detected.classifier}` left unresolved

`io.netty:netty-handler`'s published POM declares its native TLS dependency as:

```xml
<dependency>
  <groupId>${project.groupId}</groupId>
  <artifactId>${tcnative.artifactId}</artifactId>
  <classifier>${tcnative.classifier}</classifier>
  <optional>true</optional>
</dependency>
```

where `tcnative.classifier` is itself defined as `${os.detected.classifier}` - a
property that real Maven builds only get a concrete value for because the
`os-maven-plugin` build extension injects it before dependency resolution starts.
Coursier has no such extension mechanism, so it only ever sees the literal,
unresolved template text, and dependency resolution fails with:

```
not found: https://repo1.maven.org/maven2/io/netty/netty-tcnative/2.0.77.Final/netty-tcnative-2.0.77.Final-${os.detected.classifier}.jar
```

This is not new to `QuarkusModule`, but was previously invisible for most modules
because Mill's default artifact-type filter never asked Coursier to fetch this
particular dependency's artifact type. `QuarkusModule`'s wider dependency graph
resolution (needed to correctly model `protoc`/`protoc-gen-grpc-java` executables
for gRPC code generation) exposed it.

### Fix

Add `OsDetector`, a small port of the `os-maven-plugin` OS/arch normalization
algorithm (`os.detected.name`, `os.detected.arch`, `os.detected.bitness`,
`os.detected.classifier`), and wire it into `CoursierModule#resolutionParams` via
`ResolutionParams#addProperties`. This makes the properties available to every
module that resolves dependencies through Coursier (not just Quarkus modules), the
same way a real Maven build with the `os-maven-plugin` extension would.

This introduces a new OSDetectorModule with a task `osDetectedMvnProperties:`  mixed in the overriden CoursierModule's `resolutionParams` . It is only mixed in by default in QuarkusModule.

Note: this only fixes the classifier-substitution half of the problem. Coursier
also has a related, separate bug where the `optional` flag on such a dependency is
lost inside `QuarkusModule`'s newer `VariantSelector`-based resolution path, so a
truly-missing (rather than merely mis-classified) optional artifact would still
fail hard. That is tracked upstream in coursier/coursier and fixed by
https://github.com/coursier/coursier/commit/b472a5a8a15418a7ea607d1acd15dd783b76c8cf,
which is not yet in a released Coursier version Mill can depend on.

## Problem 2: jar-equivalent packaging types dropped from Quarkus's classloaders

As part of #7519, `QuarkusModule` started reporting each dependency's real Coursier
artifact type to Quarkus's `ApplicationModel` (previously every dependency was
modeled as an unclassified jar). Coursier treats several Maven packaging types as
ordinary jars for classpath purposes - `bundle`, `eclipse-plugin`, `hk2`, `orbit`,
`scala-jar`, `klib`, `maven-plugin` (see `Resolution.defaultTypes`) - since plenty
of real-world libraries use these packaging types despite being perfectly normal
jars (e.g. `jakarta.ws.rs:jakarta.ws.rs-api` is packaged as an OSGi `bundle`).

Quarkus's own `ArtifactCoords.isJar()`, however, only recognizes the literal type
`jar`/`test-jar` when deciding what to add to its classloaders
(`CuratedApplication#processCpElement`). Any dependency typed as `bundle` (or one
of the other jar-equivalent types) was therefore silently excluded from both the
runtime and deployment classloaders, surfacing as:

```
java.lang.NoClassDefFoundError: jakarta/ws/rs/BadRequestException
  org.jboss.resteasy.microprofile.config.ServletConfigSource.getValue(...)
  ...
  io.quarkus.deployment.CodeGenerator.initAndRun(CodeGenerator.java:77)
  mill.javalib.quarkus.ApplicationModelWorkerImpl.quarkusCodeGen(...)
```

during `quarkusCodeGen`, for essentially any Quarkus module depending on
`quarkus-rest` (directly or transitively) - reproduced with, among others,
`opentelemetry-quickstart`, `mongodb-quickstart`, `mongodb-panache-quickstart`,
`liquibase-mongodb-quickstart`, `kafka-avro-schema-quickstart`, and
`micrometer-quickstart` from the `quarkus-quickstarts` repo.

This is what causes [the failure in javalib tests](https://github.com/com-lihaoyi/mill/actions/runs/34403911553/job/102644004413) in the current main branch as well

### Fix

Normalize any artifact type Coursier already treats as jar-equivalent
(`artifactTypes()`, i.e. the same set used before adding the new `exe` type) down
to the literal `"jar"` before handing it to Quarkus's `ApplicationModel`. Only
types genuinely outside that set - like `exe`, used for platform-specific tool
executables such as `protoc` - are passed through unchanged, preserving the
original intent of #7519.

## Testing

- New `OsDetectorTests` covering the OS/arch normalization table.
- Extended `QuarkusModuleTests.dependencyCoordinates` with an assertion on a
  `bundle`-packaged dependency (`jakarta.ws.rs-api`), verified to fail without the
  fix and pass with it.
- `./mill libs.javalib.test` - all green.
- `./mill __.compile` - clean.
- Manually reproduced and verified the fix end-to-end against a real checkout of
  `quarkus-quickstarts` (not part of this repo) using a locally-built Mill
  assembly:
  - `grpc-plain-text-quickstart` (`quarkusDependencies`, `quarkusCodeGen`) - was
    failing with the `netty-tcnative` error, now succeeds, with `netty-tcnative`'s
    classifier correctly resolved to e.g. `linux-x86_64`.
  - `opentelemetry-quickstart`, `mongodb-quickstart`, `mongodb-panache-quickstart`,
    `liquibase-mongodb-quickstart`, `kafka-avro-schema-quickstart`,
    `micrometer-quickstart` (`quarkusCodeGen`) - were failing with the
    `jakarta.ws.rs.BadRequestException` error, now all succeed.
- `./mill example.migrating.javalib.6-maven-quarkus` works without deleting any projects now (grpc examples work)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

## External Module OsDetectorModule provided

One can do
```console
> ./mill-proguard.jar show mill.javalib.OsDetectorModule/
Mill version 1.1.0-RC4-509-02ebb8-DIRTYdbfe71c9 is different than configured for this directory!
Configured version is 1.2.0-RC1 (../mill-workspace/build.mill)
1] show
{
  "os.detected.name": "linux",
  "os.detected.arch": "x86_64",
  "os.detected.bitness": "64",
  "os.detected.classifier": "linux-x86_64"
}
1/1, SUCCESS] mill show mill.javalib.OsDetectorModule/
```

I also suspect, people will want to build native builds of other architectures . This could be done by overriding the task, but this area in general is something we have not yet explored in regards to quarkus and potentially the rest of the native stack.  

## Future work

I think these changes in their entirety or big portion of them can be ported to coursier. If people agree, I'll create a ticket and follow-up but until then, we can host them in mill

On this note, the optional changes have been merged to coursier (https://github.com/coursier/coursier/commit/b472a5a8a15418a7ea607d1acd15dd783b76c8cf) so we could clean up some recent work around this area, which was also motivated by mill's quarkus dependency failures

Also opened an issue to explore cross platform builds more


vaslabs:
should be ready, android test failure should be unrelated, seems flakiness from emulator not getting started properly or on time

lefou (commented):
Looks mostly good. I'm not so sure about the `CoursierModule` addition. I think, build-platform dependent stuff should be the exception, and bloating almost every module (`JavaModule extends CoursierModule`) with utility stuff isn't that cool. What do you think of moving `CoursierModule.osDetectedMvnProperties` into a new `trait OsDetector`? It's contained, easy mix-able but without adding to each unrelated module?

vaslabs:
> Looks mostly good. I'm not so sure about the `CoursierModule` addition. I think, build-platform dependent stuff should be the exception, and bloating almost every module (`JavaModule extends CoursierModule`) with utility stuff isn't that cool. What do you think of moving `CoursierModule.osDetectedMvnProperties` into a new `trait OsDetector`? It's contained, easy mix-able but without adding to each unrelated module?

I initially had it in JavaModule, but I think it should be how coursier should work in the first place, I mean, to me it looks like this property placeholder was missed , it should be available to coursier resolution . Ultimately, this needs to go to the coursier repo imo.

I think having a new trait would increase the steps in having a correct build, if you mean we let the users mix it in, unless I misunderstood.

vaslabs:
alternatively, if you don't like the change in coursier module, I can create the trait as you said and mix it in QuarkusModule, I think practically, frameworks that do their own thing with native builds are more likely to hit that issue than using reachability-metadata

lefou:
Yeah, my idea implied, that `QuarkusModule` mixes `OsDetector` in, to profit from the extra properties substitutions. Others modules, that need it too, can also mix it in.

The thing is, the more dynamic properties a shared resources like a remote dependency specification like a Maven POM can contain, the less likely is a reproducible build output on a different machine, which is a downside. So choosing to depend on some built-time state should be a deliberate choice of the user.

vaslabs:
> Yeah, my idea implied, that `QuarkusModule` mixes `OsDetector` in, to profit from the extra properties substitutions. Others modules, that need it too, can also mix it in.
> 
> The thing is, the more dynamic properties a shared resources like a remote dependency specification like a Maven POM can contain, the less likely is a reproducible build output on a different machine, which is a downside. So choosing to depend on some built-time state should be a deliberate choice of the user.

makes sense, thanks, will update

libs/javalib/src/mill/javalib/OsDetector.scala
```diff
@@ -0,0 +1,65 @@
...
+ * `os.detected.classifier`, `os.detected.bitness`).
+ *
+ * A number of published artifacts (e.g. `com.google.protobuf:protoc`,
+ * `io.grpc:protoc-gen-grpc-java`, `io.netty:netty-tcnative*`) reference a platform-specific
+ * classifier via a `${os.detected.classifier}`-style placeholder in their consumers' POMs,
+ * expecting one of those extensions to inject the property at build time. Coursier has no
+ * equivalent extension mechanism, so without injecting the same properties ourselves, such
+ * placeholders are left as literal, unresolvable text and dependency resolution fails.
+ */
+object OsDetector {
```
lefou: We should make this the companion object of `OsDetectorModule`.

```suggestion
object OsDetectorModule {
```

libs/javalib/src/mill/javalib/OsDetectorModule.scala
```diff
@@ -0,0 +1,26 @@
+package mill.javalib
+
+import coursier.params.ResolutionParams
+import mill.T
+import mill.api.Task
+
+@mill.api.daemon.experimental
```
lefou: No need to make this experimental.


```suggestion
```

lefou (commented):
Requested some minor changes. 

I was thinking if we could make `OsDetectorModule` an external module and provide a `detect` command to show all supported properties and their values on the current platform.

vaslabs:
> Requested some minor changes.
> 
> I was thinking if we could make `OsDetectorModule` an external module and provide a `detect` command to show all supported properties and their values on the current platform.

I think we can yeah.

Should it follow actually the same convention as Offline Module?
```
object OfflineSupport extends ExternalModule {
  override def millDiscover: Discover = Discover[this.type]
```

lefou:
Yes. This is the pattern all external modules need to follow.

vaslabs:
ok, I picked a few things from here and there. In the end I kept OsDetector object but made it private, it makes unit tests much easier, otherwise I think I'll need an evaluator.

If you still want it gone, let me know, I'll move everything to the module and change the test

libs/javalib/src/mill/javalib/OsDetectorModule.scala
```diff
@@ -21,6 +32,24 @@ trait OsDetectorModule extends CoursierModule {
...
+ */
+object OsDetectorModule extends ExternalModule, DefaultTaskModule {
+  lazy val millDiscover: Discover = Discover[this.type]
+
+  def defaultTask(): String = "detect"
+
+  /**
+   * Outputs the `os.detected.*` properties for the JVM currently running Mill.
+   */
+  def detect: T[Map[String, String]] = Task {
```
lefou: This should be a command, to capture the current system, which is otherwise not tracked as input in this cached task.

lefou (commented):
That makes sense, no need need to make testing unnecessary complicated.

I think we don't need to make the `OsDetector` private. Instead, it should be a `trait` with a singleton instance kept as companion object to allow derivations.

vaslabs:
done, I think there's still work to do in this area, I have to do some exploration. For example, building arm binaries with quarkus on an x64 machine, while technically possible (by overriding the os task or the coursier resolution params), need to test and see how quarkus devs are doing this. but this should not be in the scope of this PR .

you were right to push for moving os detection out from coursier module, I think there's more into that and would probably have some weird effects on users that do custom native builds on x64 for arm or the other way around

vaslabs:
I think it hits a dead lock and times out with the command :grimacing: 

```
05784/15379+] mill show __.compile 61s
1] show 61s
1-00051] mill.javalib.OsDetectorModule/detect 57s blocked on write lock PID 71782 'show __.compile'
1-00081] mill.javalib.OsDetectorModule/detect 57s blocked on write lock PID 71782 'show __.compile'
1-00121] mill.javalib.OsDetectorModule/detect 57s blocked on write lock PID 71782 'show __.compile'
1-00151] mill.javalib.OsDetectorModule/detect 57s blocked on write lock PID 71782 'show __.compile'
1-00191] mill.javalib.OsDetectorModule/detect 57s blocked on write lock PID 71782 'show __.compile'
1-00221] mill.javalib.OsDetectorModule/detect 57s blocked on write lock PID 71782 'show __.compile'
1-00261] mill.javalib.OsDetectorModule/detect 57s blocked on write lock PID 71782 'show __.compile'
1-00291] mill.javalib.OsDetectorModule/detect 56s blocked on write lock PID 71782 'show __.compile'
1-00331] mill.javalib.OsDetectorModule/detect 56s blocked on write lock PID 71782 'show __.compile'
1-00361] mill.javalib.OsDetectorModule/detect 56s blocked on write lock PID 71782 'show __.compile'
1-00401] mill.javalib.OsDetectorModule/detect 56s blocked on write lock PID 71782 'show __.compile'
1-00431] mill.javalib.OsDetectorModule/detect 56s blocked on write lock PID 71782 'show __.compile'
1-00471] mill.javalib.OsDetectorModule/detect 56s blocked on write lock PID 71782 'show __.compile'
```

I'll put a workaround for now to use the OsDetector directly in the `OsDetectorModule`

lefou:
I think depending on a command from a task is the issue here. The workaround looks fine to me.

libs/javalib/src/mill/javalib/OsDetectorModule.scala
```diff
@@ -0,0 +1,57 @@
...
+   * normally injected by the `os-maven-plugin`/`osdetector` build extensions - resolve to the
+   * current platform instead of being left as literal, unresolvable text.
+   *
+   * See [[OsDetector]].
+   */
+  def osDetectedMvnProperties: T[Map[String, String]] = Task {
+    // TODO using the OsDetectorModule.detect command seems to cause a timeout
+    // If the cause is found, this can be switched to reuse the command.
+    OsDetector.detect()
+  }
```
lefou: Since this task effectively depends on some system properties, it should be uncached or an input task. Otherwise we risk stale values, e.g. when the project is remote-cached or located on a shared/portable drive.

lefou:
Thank you @vaslabs !

PR #7557, opened 2026-09-12
fix(javalib): discover package-access main methods on Java 25+

### LINKED ISSUE #7549: Java 25: automatic main-class discovery misses package-access static main methods
## Problem description

With Java 25 (which is the default), Mill 1.2.0-RC1 compiles a package-access `static void main(String[] args)`, but `run` fails with `No main class specified or found`.
Adding only `public` to the method makes automatic discovery work.

[Java SE 25 JLS §12.1.4](https://docs.oracle.com/javase/specs/jls/se25/html/jls-12.html#jls-12.1.4) permits a candidate main method to have public, protected, or package access, and to be static or an instance method.
This applies to explicitly declared classes as well as compact source files.

## Environment

- Mill: 1.2.0-RC1
- Java: 25

## Minimal reproduction

- `src/Main.java`:

    ```java
    public class Main {
        static void main(String[] args) {
            System.out.println("Hello from main");
        }
    }
    ```
- `build.mill.yaml`:

   ```yaml
   extends: JavaModule
   ```

- `./mill run`:

   ```
   ./mill --ticker=false run
   compiling 1 Java source to out/compile.dest/classes ...
   done compiling
   [error] finalMainClass No main class specified or found
   ```

## Workaround

Adding the `public` modifier makes the problem go away.
However, it is flagged as redundant by IntelliJ IDEA, which can be a cause for confusion (e.g., with students).

arimu1:
## Summary

- Extend ClassGraph main-method discovery to accept non-private `static void main(String[])` methods when Mill runs on Java 25+, matching JLS 12.1.4 (public, protected, or package access).
- Call `ignoreMethodVisibility()` so ClassGraph exposes package-access `main` methods in bytecode scans.
- Add a `RunTests` regression with the issue’s minimal `Main` snippet (JDK 25+ discovers and runs; older JDKs still expect no discovery).

Fixes #7549

## Test plan

- [x] `./mill libs.javalib.test mill.javalib.RunTests` with `JAVA_HOME` set to JDK 25 (16/16 passed, including `packageAccessStaticMainDiscovery`)

### DIFF
--- a/libs/javalib/classgraph-worker/src/mill/javalib/classgraph/impl/ClassgraphWorkerImpl.scala
+++ b/libs/javalib/classgraph-worker/src/mill/javalib/classgraph/impl/ClassgraphWorkerImpl.scala
@@ -5,11 +5,23 @@ import java.io.File
 import scala.jdk.CollectionConverters.*
 import scala.util.Using
 
-import io.github.classgraph.ClassGraph
+import io.github.classgraph.{ClassGraph, MethodInfo}
 import mill.api.TaskCtx
 import mill.javalib.classgraph.ClassgraphWorker
 
 class ClassgraphWorkerImpl() extends ClassgraphWorker {
+
+  /** Java 25+ allows public, protected, or package-access static {@code main} methods (JLS 12.1.4). */
+  private def allowsNonPublicMainMethods: Boolean =
+    Runtime.version().feature() >= 25
+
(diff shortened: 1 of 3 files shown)


lefou:
The failing test looks unrelated.

lefou:
Thank you!

PR #7563, opened 2026-09-14
Android: Aggregate all dirs in android resources, fix for not being processed by the data binding worker under a single dir

### LINKED ISSUE #7530: AndroidKotlinModule.androidCompiledModuleResources ignores additional resource directories when viewBinding/dataBinding is enabled
## Problem

When `androidEnableViewBinding=true` or `androidEnableDataBinding=true`, `AndroidKotlinModule` overrides `androidCompiledModuleResources` but only processes `androidResources().head` — the first resource directory. Any additional resource directories added by subclasses via overriding `androidResources` are completely ignored.

This affects two tasks:

1. **`androidProcessedLayoutXmls`** — hardcodes `androidResources().head.path.toString` as `resInputDir`
2. **`androidCompiledModuleResources`** — only compiles `androidProcessedLayoutXmls().path / "resources"`, which is derived from the first resource dir

## Related

- Related to #5654 (Android integration issues) — that issue reported `androidCompiledModuleResources fail with multiple submodules`, which was fixed in PR #5662. However, PR #5662 only addressed **submodule dependency** resource compilation, not the case where a subclass overrides `androidResources` to add additional resource directories.
- PR #5662 changed `AndroidModule.androidCompiledModuleResources` to return `Seq[PathRef]` and added `androidTransitiveCompiledResources()`, but did not modify `AndroidKotlinModule`'s override which still hardcodes `head`.

## Reproduction

```scala
import mill._
import mill.androidlib._
import mill.kotlinlib._

object app extends AndroidAppKotlinModule {
  def androidSdkModule = mill.api.ModuleRef(AndroidSdkModule)
  def kotlinVersion = "2.1.20"
  def androidCompileSdk = 35
  def androidNamespace = "com.example.app"
  def androidEnableViewBinding = true  // key setting

  // Subclass adds an extra resource directory
  override def androidResources: T[Seq[PathRef]] = Task {
    super.androidResources() ++ Seq(PathRef(Task.dest / "res"))
  }
}
```

The extra resources in `Task.dest / "res"` will be present in `androidResources()` output but will **not** be compiled by `androidCompiledModuleResources`. AAPT2 link will then fail or produce an APK missing those resources.

## Expected Behavior

`androidProcessedLayoutXmls` and `androidCompiledModuleResources` should iterate over **all** directories returned by `androidResources()`, not just `.head`.

## Suggested Fix

In `AndroidKotlinModule.scala`:

**`androidProcessedLayoutXmls`** — change:
```scala
resInputDir = androidResources().head.path.toString,
```
to iterate over all resource dirs (or at minimum, process the first dir and pass the rest through unchanged).

**`androidCompiledModuleResources`** — change:
```scala
val moduleResources = Seq(androidProcessedLayoutXmls().path / "resources")
```
to also include aapt2-compiled output from the remaining resource directories not processed by `androidProcessedLayoutXmls`.

Note: `androidProcessedLayoutXmls` only needs to process layout XMLs (for binding code generation), so it's acceptable for it to only handle the first/main resource dir. But `androidCompiledModuleResources` must compile **all** resource dirs with AAPT2, since non-layout resources (raw, values, drawables, etc.) in additional dirs still need to be compiled into flat files for the link step.

## Environment

- Mill version: 1.1.6 (also reproduced on `main` branch)
- `androidEnableViewBinding = true`
- Workaround: Override `androidCompiledModuleResources` in `build.mill` to manually compile extra resource dirs with AAPT2

vaslabs:
Fixes https://github.com/com-lihaoyi/mill/issues/7530

Introduces a `androidProcessedLayoutInputDir` task which accumulates all the declared top level androidResources dirs into a staged/res Task dir which in turn is passed to the layout xml processor as the layout input dir. This way all Source dirs in androidResources can be qualified to be processed by view binding.

Tested with [Jetchat](https://github.com/vaslabs-ltd/compose-samples-with-mill/tree/main/Jetchat) for regressions 

Tests: https://github.com/vaslabs-ltd/compose-samples-with-mill/actions/runs/34835915734


libs/androidlib/src/mill/androidlib/AndroidKotlinModule.scala
```diff
@@ -86,6 +86,30 @@ trait AndroidKotlinModule extends KotlinModule with AndroidModule { outer =>
       .asInstanceOf[AndroidDataBindingWorker]
   }
 
+  /**
+   * Gathers all resources top directories as defined in [[androidResources]]
+   * under a single directory to be passed to the xml layout processor
+   * via [[androidProcessedLayoutXmls]]. If there are conflicting files
+   * the task fails.
+   * @return
```
lefou:
```suggestion
```

lefou (approved):
Looks good to me.

lefou:
Since copying can be a costly operation and multiple resource dirs might be the exception, we could count the actual resources and if it's only a single pathref, we could just forward it without copying. 

I think we do this in other places too, e.g. when aggregating compiled classes and semanticdb files in the BSP server.

vaslabs:
> Since copying can be a costly operation and multiple resource dirs might be the exception, we could count the actual resources and if it's only a single pathref, we could just forward it without copying.
> 
> I think we do this in other places too, e.g. when aggregating compiled classes and semanticdb files in the BSP server.

good idea, I always forget copying is expensive hashtag btrfs life :sweat_smile:

libs/androidlib/src/mill/androidlib/AndroidKotlinModule.scala
```diff
@@ -86,6 +86,35 @@ trait AndroidKotlinModule extends KotlinModule with AndroidModule { outer =>
...
+   * Gathers all resources top directories as defined in [[androidResources]]
+   * under a single directory to be passed to the xml layout processor
+   * via [[androidProcessedLayoutXmls]]. If there are conflicting files
+   * the task fails. If there is only one directory in androidResources,
+   * it is passed as the input dir so copying is avoided.
+   */
+  def androidProcessedLayoutInputDir: T[PathRef] = Task {
+    val resInputDir = Task.dest / "staged/res"
+    os.makeDir.all(resInputDir)
+    val qualifiedResDirs = androidResources().filter(pr => os.exists(pr.path) && os.isDir(pr.path))
```
lefou, line 100: What happens with resources that are not dirs? Are these even allowed (and supported somewhere)? I think we should at least warn, if there are some.

vaslabs: will put a warning, we should expect top level dirs

libs/androidlib/src/mill/androidlib/AndroidKotlinModule.scala
```diff
@@ -96,7 +96,14 @@ trait AndroidKotlinModule extends KotlinModule with AndroidModule { outer =>
...
     os.makeDir.all(resInputDir)
-    val qualifiedResDirs = androidResources().filter(pr => os.exists(pr.path) && os.isDir(pr.path))
+    val qualifiedResDirsPredicate: PathRef => Boolean =
+      pr => os.exists(pr.path) && os.isDir(pr.path)
+    val qualifiedResDirs = androidResources().filter(qualifiedResDirsPredicate)
+    val nonQualifiedResDirs = androidResources().filterNot(qualifiedResDirsPredicate)
+    nonQualifiedResDirs.foreach {
+      pr =>
+        Task.log.warn(s"Dropped $pr because it's not a directory or it does not exist")
+    }
```
lefou: We should not warn for non-existing dirs, as this is expected. We conveniently  default to resource dirs, even if they are not necessary.

vaslabs:
will wait for the fix to go in https://github.com/com-lihaoyi/mill/pull/7577 as without it, the pipelines will keep failing

PR #7565, opened 2026-09-15
fix: ignore shebang before YAML frontmatter

### LINKED ISSUE #6708: Accept and ignore a shebang (`#!`) before a YAML frontmatter (`//|`)
Java since version 11 allows a shebang in source files. Also Kotlin scripts (`.kts`) accept it. To allow Java/Kotlin scripts to also being used as Mill single file scripts Mill needs to accept (and ignore) a shebang in the first line, but still read the YAML frontmatter starting in the second line.

A second use case would be self-executable single file scripts starting with 

```Java
#!/usr/bin/env mill
```

r3wretrhy:
Java 11+ and Kotlin scripts allow a shebang on the first line. Mill YAML frontmatter (`//|`) currently has to start at line 1, so a self-executable

```
#!/usr/bin/env mill
//| mill-version: 1.2.3
```

is rejected.

Skip a leading `#!` line when reading the header (and when reading `mill-version` from the same file). A shebang does not let YAML comments appear after other code.

Fixes #6708


lefou (commented):
Looks good to me. Thank you!

PR #7570, opened 2026-09-16
Omit checksums for Sonatype signatures

alexarchambault:
This makes Mill stop adding checksum files for signatures when publishing things to Sonatype. Basically, these are `.asc.md5` and `.asc.sha1` files. These aren't required to publish things to Sonatype. They also unnecessarily increase the file count of releases, which is a problem with the new Maven Central Publishing Limits.


lefou (approved):
Looks good to me. Thank you!

